PR #1023, opened 2020-12-20
Client: make fetcher more typesafe

jochem-brouwer:
This PR is very WIP. It currently makes the Fetcher a bit more typesafe, I am running into some type problems now with the HeaderFetcher and will need to do a bit more research what the types are intended to be under the hood and possibly have to introduce a third type on Fetcher.ts.

All `any` is now removed in Fetcher/BlockFetcher, need to fix HeaderFetcher, this has some nasty types which don't work with the type system I setup.

PR #1017, opened 2020-12-15
Client: add VM execution (old)

holgerd77:
Ok, i did some back-and-forth experimentation on where/how to integrate VM execution in the client and I think this should finally do it.

This PR first removes the HeaderFetcher -> BlockFetcher inheritance dependency. This frees the Fetcher classes to now have a clean object relationship to the respective synchronizers (so BlockFetcher <-> FullSynchronizer, HeaderFetcher <-> LightSynchronizer) and subsequently allows for a direct integration of the VM execution into the BlockFetcher, the VM (another time 😄 ) also further moved into this class.

In the `BlockFetcher.store()` function the functionality from `Chain.putBlocks()` is then drawn in and decomposed (this should be no problem since this class is going away anyhow with the wrapper class removal @jochem-brouwer is planning). 
This allows for atomic alternate block execution and blockchain storage (I've also drawn in the three line `Blockchain.putBlocks()` functionality) and assures that only successfully executed upon blocks are stored in the chain.

PR is not running yet. Unit tests are passing but integration tests need some modification since the mock setup now triggers some validation checks to fail along the `vm.runBlock()` run, not sure how to fix this yet.

Client run is also triggering the following error at the moment once a block with transactions is hit: `ERROR [12-16|00:31:58] Error: sender doesn't have enough funds to send tx. The upfront cost is: 996205183388591000 and the sender's account only has: 0` (another error `ERROR [12-16|00:30:46] TypeError: Cannot read property 'map' of undefined
    at BlockFetcher.request (/EthereumJS/ethereumjs-vm/packages/client/dist/lib/sync/fetcher/blockfetcher.js:41:62)` seems unrelated to this PR and might rather be introduced along the type improvements from @ryanio in some previous PRs, at least that's my current unproven suspicion).

So this is rather open to be picked up and further continued.

PR #1026, opened 2020-12-22
Add EIP 2565, ModExp precompile gas cost

jochem-brouwer:
This PR adds [EIP 2565](https://eips.ethereum.org/EIPS/eip-2565), ModExp precompile gas cost, which is planned for Berlin. The EIP looks more complex than it seems to be, the only change however is to the `multComplexity` method, see the [reference](https://github.com/holiman/go-ethereum/blob/2c99023b68c573ba24a5b01db13e000bd9b82417/core/vm/contracts.go#L311) implementation of Geth.

This PR also seems to resolve a bug when the result of ModExp is 0. The output is expected to be of length `modulusLength`, an input to the precompile. Originally, it returned one byte with zeros.

PR #1028, opened 2021-01-02
Client: add VM execution

holgerd77:
Continuation of #1017 

This has been rebased on top of #1023 and is working apart from one integration test failing (FullSync, this is going into an endless loop in the `while` loop in `sync.start()`).

Will give some more comments on the changes later the day.

We should merge relatively soon otherwise this is getting a bit out of hands. 😋


holgerd77:
Some more debugging: integration test failure is due to FullSynchronizer.best() not delivering a peer on the `sync()` call.

holgerd77:
This difficulty check `(!best && td.gte(this.chain.blocks.td))` in `best()` is failing now (in contrast to before).

holgerd77:
Ok, tests are passing and client run is working both initially and resumed upon an existing chain state, so this is now ready for review. 😄 

The devp2p changes from 8bfec9d44eee93a80c9d2ae66209f113830b0fa5 are technically independent from this PR, I did this yesterday because sync start did take so long and often was unreliable so it got very much a hazzle to further debug the PR. The change is subdividing the peer `findNeighbours`refresh calls, so now calls for all the peers are not done any more at the same time (given by the `interval` option) but calls are distributed (still keeping the `interval` time for a dedicated peer) throughout the time period with a segregation by a `peer.id` derived selector, I had this change in mind already for quite some time. This leads to a more distributed networking traffic (avoiding networking spikes switching with no-activity periods) and leads to a faster sync pick-up.

holgerd77:
Rebased this.

holgerd77:
This is now open for review (and merge), #1030 should subsequently automatically update its base.

holgerd77:
Rebased this on top of #1026

holgerd77:
Hi @jochem-brouwer, good morning 😀 , then only thing I would insist upon: can we please get this merged before you start rewriting the cache mechanism for trie? This is otherwise getting a bit hairy.

This PR shouldn't be too hard to review, especially since the core logic was rewritten by yourself, some explanations:

---

After rebasing upon your fetcher type improvements in 1fe5e03d1fe7e13be6833cb867bf27e1faff359f I did the following:

- The state DB is properly closed d9846846c9d670855a54ba9d092eca7d06ed7bd5 (line 248 in `fullsync.ts`, actually needs to be expanded to the the `close()` line, addition seems to have been done in a commit before that) in `FullSynchronizer.stop()`, follow-up correction in c7e32c86b7bbf651df83921cefcd8a3105145b2a
- State and chain DBs are better differentiated by renaming the original `db` -> `chainDB` and aligning state DB to also be an option in `node.ts` analog to `db`, 67aaf12b36cefdd0371f538128ee8fafd0f65dd8 and baf5c3309fa58838f61e9455db90233b178624ad
- There is now a more consistent and chain/state separating directory layout with c7079d9afb8fc66715571284e6595ac36608fcb5 so that we have the network data in folders like:
  - `/[HOME_DIR]/Ethereum/ethereumjs/mainnet/chain`
  - `/[HOME_DIR]/Ethereum/ethereumjs/mainnet/state`
  - `/[HOME_DIR]/Ethereum/ethereumjs/rinkeby/chain`
  - `/[HOME_DIR]/Ethereum/ethereumjs/rinkeby/state`
  - ...
- Devp2p DPT changes from d3b5856d799a183a2502ad39c8328679340467a3 are already explained in https://github.com/ethereumjs/ethereumjs-vm/pull/1028#issuecomment-753491184. This is in the peer discovery mechanism, `findNeighbours()` looks for others peers to connect to based on the existing (initially: bootstrap) peers. The call to `refresh()` is now done 10 times more often with a subdivided refresh interval by `const refreshIntervalSubdivided = Math.floor((options.refreshInterval || ms('60s')) / 10)`. In `refresh()` there is then a selector extracted from the peer ID with `const selector = buffer2int(peer.id.slice(0, 1)) % 10`, so this just gives a number 0..9 for each peer. On a refresh call there are then only peers called where this selector matches a looping 0..9 counter, `selector === this._refreshIntervalSelectionCounter`. All this is just to distribute the network traffic a bit better, before there were all calls together in one moment in time (every `refreshInterval` seconds). Hope this is a sufficient explanation for review. Peer discovery in the client is still working as before with this (actually: substantially better, therefore the change)
- The remaining commits are just some added tests and some minor fixes

packages/client/lib/client.ts
```diff
@@ -51,11 +58,13 @@ export default class EthereumClient extends events.EventEmitter {
...
             config: this.config,
-            db: options.db,
+            chainDB: options.chainDB,
+            stateDB: options.stateDB,
           })
         : new LightEthereumService({
             config: this.config,
-            db: options.db,
+            chainDB: options.chainDB,
+            stateDB: options.stateDB,
```
jochem-brouwer, line 67: Do we need a stateDB in LightEthereumService?

packages/client/lib/config.ts
```diff
@@ -184,15 +184,25 @@ export class Config {
...
-   * Returns the directory for storing the client sync data
+   * Returns the directory for storing the client chain data
    * based on syncmode and selected chain (subdirectory of 'datadir')
    */
-  getSyncDataDirectory(): string {
-    const syncDirName = this.syncmode === 'light' ? 'lightchaindata' : 'chaindata'
-    const chain = this.common.chainName()
-    const networkDirName = chain === 'mainnet' ? '' : `${chain}/`
+  getChainDataDirectory(): string {
+    const networkDirName = this.common.chainName()
```
jochem-brouwer, line 191: How would this work for private chains?

packages/client/lib/config.ts
```diff
@@ -184,15 +184,25 @@ export class Config {
...
+    const chainDataDirName = this.syncmode === 'light' ? 'lightchain' : 'chain'
```
jochem-brouwer, line 192: In principle, the data which light/full reports should be the same, right? Only light will report a subset of the full data (IIRC Light only downloads Block Headers and not Block Bodies / state - right? Block Headers can also be used in full sync again, don't have to re-request those).

packages/client/lib/config.ts
```diff
@@ -184,15 +184,25 @@ export class Config {
...
 
-    const dataDir = `${this.datadir}/${networkDirName}ethereumjs/${syncDirName}`
+    const dataDir = `${this.datadir}/${networkDirName}/${chainDataDirName}`
```
jochem-brouwer, line 194: I think I like this approach, since you now ensure that if someone accidentally uses the same data dir for a different chain than it was initially used, they won't start using the "wrong" data.

packages/client/lib/sync/fullsync.ts
```diff
@@ -1,34 +1,105 @@
...
+
       this.vm = new VM({
         common: this.config.common,
         blockchain: this.chain.blockchain,
+        stateManager,
       })
     } else {
       this.vm = this.config.vm
       //@ts-ignore blockchain has readonly property
       this.vm.blockchain = this.chain.blockchain
```
jochem-brouwer, line 48: Just noticing this here, we don't enforce that VM uses the `StateManager` which is attached to the `stateDB`. Not sure if I/we (?) intended this.

packages/client/lib/sync/sync.ts
```diff
@@ -47,6 +52,7 @@ export abstract class Synchronizer extends EventEmitter {
 
     this.pool = options.pool
     this.chain = options.chain
+    this.stateDB = options.stateDB
```
jochem-brouwer, line 55: I don't think we need `stateDB` here, since it does not seem we use it anywhere?

packages/vm/lib/runBlock.ts
```diff
@@ -149,17 +149,26 @@ export default async function runBlock(this: VM, opts: RunBlockOpts): Promise<Ru
   }
 
   // Checkpoint state
-  await state.checkpoint()
+  // (only do checkpointing when txs are executed)
+  // TODO: determine if checkpointing is needed in a block
+  // context at all or CPs in runTx() are sufficient
```
jochem-brouwer, line 154: Yeah, we want this - if we run a bad block, we want to rollback the DB to the version how it was before running the bad block.

packages/vm/lib/runBlock.ts
```diff
@@ -149,17 +149,26 @@ export default async function runBlock(this: VM, opts: RunBlockOpts): Promise<Ru
...
+    await state.checkpoint()
+  }
   let result
   try {
     result = await applyBlock.bind(this)(block, opts)
   } catch (err) {
-    await state.revert()
+    if (block.transactions.length > 0) {
+      await state.revert()
+    }
```
jochem-brouwer, line 164: I guess this will work as I cannot think of a case where the block has 0 txs and throws (in that case, we'd need to rollback the `coinbase` account to the previous block)

packages/vm/lib/runBlock.ts
```diff
@@ -149,17 +149,26 @@ export default async function runBlock(this: VM, opts: RunBlockOpts): Promise<Ru
...
     throw err
   }
 
   // Persist state
-  await state.commit()
+  if (block.transactions.length > 0) {
+    await state.commit()
+  }
```
jochem-brouwer, line 171: If we have 0 transactions (thus no checkpoint) then the DB is flushed immediately? (Is `coinbase` account always updated on disk when running the block?)

jochem-brouwer (commented):
I got some questions, but nothing alarming. Looks good! 😄

packages/client/lib/client.ts
holgerd77: No, you are right, I will remove. Will keep the member association in `EthereumService` though, might be more flexible for future use cases (respectively generally not such a big question and we can also move later to `FullEthereumService` if we decide to).

packages/client/lib/config.ts
holgerd77: Hmm, I haven't spent too much time thinking about custom chains yet TBH. On a first look it is possible though to provide a chain name in `Common.forCustomChain()` and this would then be used for the folder name.

packages/client/lib/config.ts
holgerd77: Haven't spent that much time thinking about `lightsync` either 😋 , since we don't have a focus on this right now. On first thought I would assume that mixing this in one folder is not what one would want. If you start with a full sync to Block x, then continue to Block y with a light sync, all the block bodies from x to y would be missing? 🤔

Anyhow. This is really a simple change if we would want to combine, nothing which would need to be decided now, I just took this over "as is" from the existing implementation.

jochem-brouwer: Maybe you are right - it is not a problem for this PR anyways, just wanted to mention it 😄

packages/client/lib/sync/fullsync.ts
holgerd77: Yes, noticed this as well at some point and decided not to put too much attention right now, but this is a bit an open field where we need to put a bit more continued thought, not so pressing though I would say.

packages/client/lib/sync/sync.ts
holgerd77: I put it there and not directly in `FullSynchronizer` since I thought this might be more flexible if we want to use `stateDB` in other synchronizers. Do you want me to move over to `FullSynchronizer` directly? Don't have a strong opinion on this myself, think these are all things which can be easily evolved and adopted further down the road.

packages/vm/lib/runBlock.ts
holgerd77: Ok, have removed the comment part. 😛

packages/client/lib/sync/sync.ts
jochem-brouwer: I have no strong opinion either, it is OK to keep it here for now, if we focus on the light client then we can remove it if it turns out it is unnecessary.

packages/vm/lib/runBlock.ts
holgerd77: You know what? This extra case with the zero txs will get somewhat unnecessary with the better checkpointing mechanism anyhow and no need to take in extra risks here (haven't thought about the `coinbase` account e.g.), I will just remove this part of the commit.

packages/vm/lib/runBlock.ts
holgerd77: see above

holgerd77:
@jochem-brouwer ok, just pushed the changes, ready for re-review. 😀

packages/client/lib/sync/fullsync.ts
```diff
@@ -1,34 +1,105 @@
...
   public vm: VM
+  public runningBlocks: boolean
 
-  private blockFetcher: BlockFetcher | null
+  private stopSyncing: boolean
+  private vmPromise?: Promise<void>
+
+  // Tracking vars for log msg condensation on zero tx blocks
+  private NUM_ZERO_TXS_PER_LOG_MSG = 50
+  public zeroTxsBlockLogMsgCounter: number = 0
```
ryanio, line 26: nit: I don't think need the `number` type annotation since it's being initialized with `0`

packages/devp2p/src/dpt/dpt.ts
```diff
@@ -55,6 +55,7 @@ export class DPT extends EventEmitter {
   private _kbucket: KBucket
   private _server: DPTServer
   private _refreshIntervalId: NodeJS.Timeout
+  private _refreshIntervalSelectionCounter: number = 0
```
ryanio, line 58: same nit here for `number` type annotation, I believe not needed when already initialized with a number

jochem-brouwer (approved):
Yup, great!! 😄

jochem-brouwer:
I assume that first this one gets moved into master and then the checkpoint one immediately after?

holgerd77:
@jochem-brouwer thanks for the review 😄 , yes, I will now first merge here.

PR #1029, opened 2021-01-03
Devp2p DPT Type Improvements

holgerd77:


holgerd77:
Ok, open for review, commit messages should be relatively self-descriptive. 😄

packages/devp2p/src/dpt/kbucket.ts
```diff
@@ -1,56 +1,58 @@
...
 
-  static getKeys(obj: Buffer | string | KObj): string[] {
+  static getKeys(obj: Buffer | string | PeerInfo): string[] {
     if (Buffer.isBuffer(obj)) return [obj.toString('hex')]
     if (typeof obj === 'string') return [obj]
 
     const keys = []
     if (Buffer.isBuffer(obj.id)) keys.push(obj.id.toString('hex'))
-    if (obj.address && obj.port) keys.push(`${obj.address}:${obj.port}`)
+    //if (obj.address && obj.port) keys.push(`${obj.address}:${obj.port}`)
```
ryanio, line 46: should this remain commented out?

packages/client/lib/net/server/rlpxserver.ts
```diff
@@ -122,17 +122,17 @@ export class RlpxServer extends Server {
...
-      } catch (e) {
-        this.error(e)
-      }
+      return this.dpt!.bootstrap(bootnode)
     })
-    await Promise.all(promises)
+    try {
+      await Promise.all(promises)
+    } catch (e) {
+      this.error(e)
```
ryanio, line 134: codecov is warning that this line isn't covered, it's not *so* important, but could add a quick test forcing a fast timeout to ensure the error is caught and propagated properly

ryanio (approved):
left two quick comments but otherwise looks great! much better clarity with the improved types :)

packages/client/lib/net/server/rlpxserver.ts
holgerd77: I will leave this for now but generally that's a good idea - also left a comment on the codecov/patch discussion in the chat. Will likely be good if we get to some work mode where we take coverage more serious again (I at least started in #1028 to add some substantial tests and had a look at codecov/patch, took me quite some time but I think I finally had some mental breakthrough in my `testdouble` understanding 😄  ).

holgerd77: I've directly taken the occasion and made the `codecov/project` check mandatory again, together with the other library's `test-*` checks (e.g. `test-trie`), think we still had some gap here due to uncertainties in the CI setup for some time (these "run everything or run selected" discussions). And we can of course later always adopt again if some CI setup change comes up.

packages/devp2p/src/dpt/kbucket.ts
holgerd77: Yes, there is just no property `port` on peer objects - just `updPort` and `tcpPort`, so the line in its current form has no effect and I couldn't decide on selecting for one of the other port properties, in fact I am not understanding well enough what this line is doing TBH. So I left to indicate that there was something which might have had some intention. 😄  Could have left a comment though.

Will also leave to not revoke the approval but rather merge here now.

PR #1030, opened 2021-01-04
Trie: Rework Checkpointing Mechanism

holgerd77:
Along the work on the VM execution in the client #1028 it became pretty clear that the current trie checkpointing mechanism is one of our main bottlenecks (if not: THE bottleneck) when removing the checkpointing in `VM.runBlock()` on zero transactions 6f64fa69d5e6e42b15af82b9ceaa98ab26fe6841 (so: where is no additional tx checkpointing -> no checkpointing applied at all) increased processing performance by a factor of 10-100 (and rather the upper bound, in fact processing log messages had to be batched in 50 block chunks (before: 1 log msg per block) in 45c9e9d77084e65f2ce4f456d39bc5ccfa79642c and log output is still coming somewhat faster-paced than before.

Currently the trie checkpointing mechanism is copying the whole state db on checkpointing which is extremely resource intense and not sustainable. This PR will experiment with a more fine-grained approach by creating an operations stack which can be reverted on a `trie.revert()` and - simply - deleted on a `trie.commit()`.

This first PR push which is just including the first commit 2926256 which removes all the `ScratchDB` related logic from `CheckpointTrie` (so basically: all the checkpointing functionality itself) is for a first test to see what kind of tests are failing within this constellation.

Interestingly enough ALL (!!) tests from `checkpoint.spec.ts` are still passing, lol. 😜  Couldn't believe this as well at first glance but double checked and really seems to be the case, even the one single `revert` related test is not triggering anything.

ARgh.


holgerd77:
Woohoo, this is working and it was surprisingly easy, just one day of work. 😄

All tests are passing and this brings similar speed increases (10-50x) as seen in the 0-tx blocks processing.

This is a client run before (with 1 block at a time):

https://user-images.githubusercontent.com/931137/103559314-39764a00-4eb6-11eb-851f-2398a91b7fbc.mov

And this after (with 50 blocks at a time respectively per log msg):

https://user-images.githubusercontent.com/931137/103559397-5ad73600-4eb6-11eb-8025-96dabe34cbf9.mov

holgerd77:
Depends on #1028

packages/vm/lib/state/stateManager.ts
```diff
@@ -161,7 +161,8 @@ export default class DefaultStateManager implements StateManager {
     const account = await this.getAccount(address)
     const storageTrie = this._trie.copy(false)
     storageTrie.root = account.stateRoot
-    storageTrie._checkpoints = []
+    //@ts-ignore
```
ryanio, line 164: just curious what this `@ts-ignore` is for, wondering if we could solve it without.

side note, i wish there was a finer grained `@ts-ignore {rule}` like `eslint-ignore-next-line {rule}`, would be much more informative :)

packages/trie/test/db.spec.ts
```diff
@@ -1,29 +1,31 @@
...
+tape('DB tests', (t) => {
   const db = new DB()
 
-  const k = Buffer.from('foo')
-  const v = Buffer.from('bar')
+  const k = Buffer.from('k1')
+  const v = Buffer.from('v1')
+  const k2 = Buffer.from('k2')
+  const v2 = Buffer.from('v2')
+  //const k3 = Buffer.from('k3')
```
ryanio, line 11: commented out, if unused can remove?

jochem-brouwer:
So a general question here - if you remove the checkpointing logic, then the checkpointing tests still pass? (I got the feeling I am missing something here). That implies that checkpointing is not tested thoroughly, right?

packages/trie/src/checkpointTrie.ts
```diff
@@ -71,15 +44,13 @@ export class CheckpointTrie extends BaseTrie {
    * parent checkpoint as current.
    */
   async revert(): Promise<void> {
+    if (!this.isCheckpoint) {
+      throw new Error('trying to revert when not checkpointed')
+    }
+
     await this.lock.wait()
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
ryanio, line 52: can this `no-unnecessary-condition` line be removed?

packages/trie/src/db.ts
```diff
@@ -55,23 +106,79 @@ export class DB {
...
    * Performs a batch operation on db.
    * @param opStack A stack of levelup operations
    */
   async batch(opStack: BatchDBOp[]): Promise<void> {
-    await this._leveldb.batch(opStack, ENCODING_OPTS)
+    if (this.isCheckpoint) {
+      for (const op of opStack) {
+        if (op.type === 'put') {
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+          if (!op.value) {
```
ryanio, line 170: is this `!op.value` check for safety so in case it isn't present it doesn't run a del operation?

packages/trie/src/db.ts
```diff
@@ -55,23 +106,79 @@ export class DB {
...
+            continue
+          }
+          await this.put(op.key, op.value)
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+        } else if (op.type === 'del') {
```
ryanio, line 175: is this `no-unnecessary-condition` for `if (op.type === 'del')`? is it because it would otherwise just be covered by the `else` clause because type is either `put` or `del`? if so i think i like it more as-is. otherwise could change to `} else {` and just leave a comment below `// type: del`

packages/trie/test/stream.spec.ts
```diff
@@ -156,21 +156,21 @@ tape('db stream test', function (tester) {
...
+  } as any*/
 
   it('should populate trie', async function (t) {
     trie.checkpoint()
     await trie.batch(ops)
     t.end()
   })
 
-  it('should only fetch nodes in the current trie', function (t) {
+  /*it('should only fetch nodes in the current trie', function (t) {
```
ryanio, line 173: still commented out

holgerd77:
Rebased this.

packages/vm/lib/state/stateManager.ts
holgerd77: Ah, this was still from the old trie version not being updated in Visual Studio Code (Visual Studio has its problems here or is it just me? 🤔 ) and I was blindly taking this over. So this is not necessary, will remove (next push). Thanks for the catch!

packages/trie/src/checkpointTrie.ts
holgerd77: Yes, have removed.

packages/trie/src/db.ts
holgerd77: Took this code over from `Trie.batch()` and thought it might be safer to leave (but not interrupt the `batch()` run, might be better on the `DB` level to just do the plain `level` operations "as-is", will remove.

packages/trie/src/db.ts
holgerd77: I could remove the first `no-unnecessary-condition` here but linter is complaining on the second. So I will leave half-as-is. 😄

packages/trie/test/stream.spec.ts
holgerd77: This test doesn't work any more since `_scratchReadStream` related, will remove.

packages/trie/test/db.spec.ts
holgerd77: Have removed.

holgerd77:
Have rebased this and done the fixes.

@jochem-brouwer that's correct, there was not a single checkpoint test which would fail when the functionality was removed, so these tests had not very much of an effect (apart from maybe testing that the added functionality is not introducing additional failures to the base functionality).

A bit strange, but as one can see, these things can also happen. 😛

Checkpointing is now better covered with the new DB-related checkpointing tests. To further increase trust in the mechanism I've now also expanded these tests to run in a Trie context with the last commit.

holgerd77:
@alcuadrado @ryanio Actually we DO have some kind of threshold for patch, don't we? If "95% of diff hit" is enough to pass? 🤔

ryanio (dismissed):
great, lgtm!

jochem-brouwer:
I'll have to give this a more in-depth look later, but a general comment here would be that I find it surprising that MPT does not have a cache just like `StateManager` has. I was under the impression that MPT had this as well. If we want to optimize this in the future, we should introduce a cache - this would make checkpointing probably faster too as this would decrease the amount of I/O operations. If you would revert a cache where none of the cache items were written to disk, you have to do nothing and just delete all of the batch operations currently scheduled to be committed. If a cache item would be written to disk, then we would indeed have to revert it. 

An alternative approach would be to only save the internal roots of the MPT. If you would then revert a MPT, you would simply set the root of the MPT to the old root. The downside of this would be that you would save junk nodes on the disk, which could possibly exhaust the storage space rather quickly. (I'm gonna take a look how other clients do this).

In general this is a very cool PR 👍 😄

packages/trie/src/db.ts
```diff
@@ -55,23 +106,75 @@ export class DB {
...
+          type: 'put',
+          key: key,
+          value: oldValue,
+        })
+      }
+    }
     await this._leveldb.put(key, val, ENCODING_OPTS)
+    // In CP mode add del to the revert ops stack
+    if (this.isCheckpoint) {
+      revertOps.push({
```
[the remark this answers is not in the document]
jochem-brouwer: (The `key` does not change during this method call)

packages/trie/src/db.ts
```diff
@@ -55,23 +106,75 @@ export class DB {
...
    * @param keys
    */
   async del(key: Buffer): Promise<void> {
+    const revertOps: BatchDBOp[] = []
+    // In CP mode check for an old value to be put
+    // on the revert ops stack
+    if (this.isCheckpoint) {
+      const oldValue = await this.get(key)
+      if (oldValue !== null) {
+        revertOps.push({
```
jochem-brouwer, line 146: Since we only put one operation in the `revertOps` array, can't we immediately call `addCPRevertOperation`?

packages/trie/src/db.ts
```diff
@@ -27,9 +34,53 @@ export class DB {
...
+    }
+    return root
+  }
+
+  /**
+   * Reverts the latest checkpoint
+   */
+  async revert() {
+    const { root, revertOps } = this.checkpoints.pop()!
+    await this.batch(revertOps.reverse())
```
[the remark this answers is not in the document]
jochem-brouwer: This seems to be reversed because of the logic in the `put` method; we first want to `del` the key and then `put` the old value in the corresponding key - right?

jochem-brouwer (commented):
Some questions, nothing alarming (I only reviewed the changes in the `trie` package)

holgerd77:
@jochem-brouwer oh yeah, a cache here is a great idea. I had this in mind when we were brainstorming around performance improvement ideas, but funnily enough I didn't bring this together when working on this here. Feel free to rework or throw away everything I've done here (not in this PR though! 😋 ), I am not attached to this at all and would be glad if this PR would trigger further optimizations. The switch here is definitely a big improvement but on thinking about it an in-memory cache is definitely the more optimal solution. I think we should give this a really high priority, since these optimizations on the MPT are so much felt throughout the whole (minimally EthereumJS) ecosystem.

packages/trie/src/db.ts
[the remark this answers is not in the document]
holgerd77: Yeah, correct, this is just a design decision, the alternative would have been to always prepend to the array which didn't seem right to me.

packages/trie/src/db.ts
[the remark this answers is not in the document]
holgerd77: Not sure, might not be necessary, I just wanted to play more on the safe side in doubt and not in the optimization side (this is already in by overall design-change), I was generally a bit afraid about introducing some edge cases I might not have thought through.

Open for further optimizations though, might be worth more to directly put the thinking in the caching solution though.

packages/trie/src/db.ts
holgerd77: This is for the case that `leveldb.del` might throw for some reason, then the revert operation shouldn't be added to the revert stack.

jochem-brouwer:
Yep, the idea is to do this in a follow up PR. Besides a few questions I am fine with this one in general 😄

jochem-brouwer (dismissed):
After your comments, LGTM! I'll start with the cache design probably tomorrow.

holgerd77:
@jochem-brouwer thanks, I would want to merge #1028 before though 😄  .

holgerd77:
(this one is targeted towards that branch)

holgerd77:
Ah, the base branch changed here not on merge of #1028 but at the moment I deleted the branch over there, that's interesting.

Process has dismissed the reviews here though, so this would need a renewed approval. //cc @jochem-brouwer @ryanio or everyone else

jochem-brouwer:
Your force push is due to the rebase (because of some changes of the VM full sync branch), right?

holgerd77:
@jochem-brouwer not 100% sure, after merging #1028 there was still an "Update branch" button here. So I rebased the branch locally towards master and then force-pushed to be sure that everything (hopefully) is correct.

holgerd77:
@jochem-brouwer scrolled through all the changes, looks everything correct though.

jochem-brouwer (approved):
Also looks correct to me!

PR #1031, opened 2021-01-05
Client Network Improvements

holgerd77:
This PR does the following:

- Fixes a severe devp2p bug not sending the `nextForkBlock` correctly on ETH.sendStatus() msg
- Improves RLPx disconnect reason debug output
- Updates Goerli bootnodes

### DIFF
--- a/packages/devp2p/src/eth/index.ts
+++ b/packages/devp2p/src/eth/index.ts
@@ -41,7 +41,7 @@ export class ETH extends EventEmitter {
     // Set forkHash and nextForkBlock
     if (this._version >= 64) {
       const c = this._peer._common
-      this._hardfork = c.hardfork() ? (c.hardfork() as string) : this._hardfork
+      this._hardfork = c.hardfork() ? c.hardfork() : this._hardfork
       // Set latestBlock minimally to start block of fork to have some more
       // accurate basis if no latestBlock is provided along status send
       this._latestBlock = c.hardforkBlock(this._hardfork)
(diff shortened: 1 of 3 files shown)


packages/devp2p/src/eth/index.ts
```diff
@@ -208,7 +208,7 @@ export class ETH extends EventEmitter {
         this._latestBlock = status.latestBlock
       }
       const forkHashB = Buffer.from(this._forkHash.substr(2), 'hex')
-      const nextForkB = Buffer.from(this._nextForkBlock.toString(16), 'hex')
+      const nextForkB = int2buffer(this._nextForkBlock)
```
jochem-brouwer, line 211: I don't really see how this is a fix to a bug? What is changed here?

holgerd77: I am just a simple man and still not so fluent in `Buffer` operations after all these years 😜 . The only thing I can say here is that the old code delivered an empty `Buffer`, just re-tested this with Node to be sure:

```shell
> let nextForkBlock = 5
undefined
> Buffer.from(nextForkBlock.toString(16), 'hex')
<Buffer >
```

holgerd77:
@alcuadrado @ryanio this is a good PR example where `codecov/patch` gets a bit annoying in the current setup.

packages/devp2p/src/eth/index.ts
ryanio: ha! here is some explanation: https://github.com/nodejs/node/issues/24491#issuecomment-440085063

ryanio (approved):
lgtm

PR #1032, opened 2021-01-06
Add Clique Support

holgerd77:
This PR aims to add clique support as defined in [EIP-225](https://eips.ethereum.org/EIPS/eip-225) throughout the library stack, so likely touching `Common`, `Block`, `Blockchain`, eventually also `Client`.

Some references:

- [Go Ethereum Implementation](https://github.com/ethereum/go-ethereum/blob/master/consensus/clique/clique.go)

Still WIP.

Note: mention `Block` bug fix from def4177ff4db54790f9693177c9881c4f1f5d501 in release notes


holgerd77:
Ok, made some good progress here today and we at least have got some substantial start here, this will likely take some time to mature though, there are also still larger parts missing, in particular the signer voting (respectively the extraction of an up-to-date signer list from the voting info from the blocks, read the EIP if you are interested) and the signature validation, both likely for implementation in `Blockchain`.

If someone has a good idea where to get or how to easily generate PoA specific blockchain test data - so a file like [testdata.json](https://github.com/ethereumjs/ethereumjs-vm/blob/master/packages/block/test/testdata/testdata.json) but with Rinkeby or Goerli blocks - that would really helpful, got a bit stuck on this front and at the end deactivated some tests for now.

jochem-brouwer:
@holgerd77 We only care about the blocks in this testdata right? The transactions in them don't really matter? If that's the case we can just pull the first ~10 blocks from Goerli/Rinkeby using web3 I'd say?

holgerd77:
@jochem-brouwer I had some problems with missing transactions as well, e.g. on running (testing) the general `Block.validate()` function, then the tx trie validation would fail. So optimally this would be including the tx data (if not so be it).

holgerd77:
This is now functionally ready I would say, namely this PR implements the following at this point:

- `Common` now includes a PoA consensus configuration for the PoA clique networks (Rinkeby and Goerli)
- `Block` performs various header field checks which are specific to Clique (`timestamp` diff, format of `extraData` which has a central role on the protocol)
- Various helper functions for `Block` like `cliqueIsEpochTransition()` to comfortably request PoA related data from a block
- `Block`: Working signature validation (the signature is included in the last 65 `extraData` bytes and called `extraSeal`)
- `Blockchain`: extracting initial signers from the genesis block (also included in `extraData` and re-included for validation every `EPOCH_TRANSITION` blocks (so that block is then also called an "epoch transition block" with some special properties), data structure for making updated signer states persistent in DB, the structure also prepares for re-orgs where some last signer states might be deleted (deletion on re-orgs not yet implemented, should not be difficult though with the preparations done)
- `Blockchain`: PoA clique block signature validation when `validateConsensus` is activated
- `Blockchain`: added the voting mechanism (voting - to add a new signer or remove an existing one - is done repurposing some block header fields like `coinbase` and `nonce`), vote persistence in DB  (also re-org ready), most of the test cases from the [EIP](https://eips.ethereum.org/EIPS/eip-225)
- `VM`: made PoA clique ready (the `COINBASE` opcode returns the block signer from the `extraData` field on clique, since the `coinbase` header field has some other purpose and is not containing the actual signer)

There is some further work to be done and for example some test cases to be finished and/or reactivated, this also needs some polishing in general, so I would expect this to take another 1-2 weeks at least until we get this merged here.

Since the changes are so broad I will nevertheless put this open this for review already - together with adding a "do-not-merge" label - since it will make sense to not do an all-at-once review here but rather give some continued thoughts and polish together. You are also very invited to open PRs towards this branch if you want to continue to work here or make something better.

I have some dedicated questions and remarks as well, will put these directly along the related code parts.

Cheers 🥳  👾  🌻

Holger

ryanio:
I've rebased on master and added some commits, here are the summary of changes:

* Tests added for coverage:
  * Test invalid extra data length (pow, poa)
  * Test invalid signer list (indivisible by 20)
  * Test timestamp diff between blocks is lower than PERIOD
* Renamed `_checkClique()` to `_requireClique()` for increased clarity
* Used `Address` instead of `Buffer` for clique signer types
* Block Header:
  * Add helper method `hashClique()`
  * Renamed for clarity `cliqueSignatureToAddress()` -> `cliqueSigner(): Address`
* Blockchain: 
  * Catch `NotFoundError` for clique key db reads
  * Add method helper: `cliqueSaveSigners(genesisBlock)`
  * Replaced `const CLIQUE_EMPTY_BENEFICIARY = Buffer.alloc(20)` with `Address.isZero()`\
  * Completed todo: validate signer lists on epoch transition blocks

---

Pending TODOs I will be working on next
  * Test zero beneficiary (coinbase) on epoch transitions
  * Test non-zero mixHash
  * CLIQUE_SIGNER_HISTORY_BLOCK_LIMIT is unused?

---

One future TODO for myself: I found having to write a lot of `address1.toBuffer().equals(address2.toBuffer()` so I would like to add `Address.equals(address: Address)` to `ethereumjs-util`, which would shorten many lines to `address1.equals(address2)`.

holgerd77:
Great, I'm ok with all the changes, yes, CLIQUE_SIGNER_HISTORY_BLOCK_LIMIT is not yet used, this should be interested into the updateSignerStates method (at least that was my intention, came over it though) to not endlessly bloat the list of signer states on ongoing updates.

packages/blockchain/src/index.ts
```diff
@@ -622,22 +622,28 @@ export default class Blockchain implements BlockchainInterface {
...
+        // set total difficulty in the current context scope
+        console.log('gogo1')
+        if (this._headHeaderHash) {
+          currentTd.header = await this.getTotalDifficulty(this._headHeaderHash)
+        }
+        console.log('gogo2')
+        if (this._headBlockHash) {
+          currentTd.block = await this.getTotalDifficulty(this._headBlockHash)
+        }
+        console.log('gogo3')
```
holgerd77: Ok, seems we are all debugging in the same way, lol. 😂

packages/blockchain/src/index.ts
```diff
@@ -457,34 +622,46 @@ export default class Blockchain implements BlockchainInterface {
...
-      const parentTd = await this.getTotalDifficulty(header.parentHash, blockNumber.subn(1))
-      td.iadd(parentTd)
+        // calculate the total difficulty of the new block
+        const parentTd = await this.getTotalDifficulty(header.parentHash, blockNumber.subn(1))
+        td.iadd(parentTd)
+      }
 
-      // save block and total difficulty to the database
+      // save total difficulty to the database
       dbOps = dbOps.concat(DBSetTD(td, blockNumber, blockHash))
```
holgerd77, line 654: So should this always be executed or could this also be put into the ethash-only section? 🤔

packages/blockchain/src/index.ts
```diff
@@ -401,7 +400,7 @@ export default class Blockchain implements BlockchainInterface {
 
   private async cliqueUpdateVotes(header?: BlockHeader) {
     // Block contains a vote on a new signer
-    if (header && !header.coinbase.toBuffer().equals(CLIQUE_EMPTY_BENEFICIARY)) {
+    if (header && !header.coinbase.isZero()) {
```
holgerd77, line 411: These kind of changes are really a proof that it makes so much sense to work on these kind of PRs together! 😄

holgerd77: (so everyone has his special open-eyes-sections)

packages/blockchain/src/index.ts
```diff
@@ -435,15 +434,16 @@ export default class Blockchain implements BlockchainInterface {
...
+          } else if (nonce.equals(CLIQUE_NONCE_DROP)) {
             activeSigners = activeSigners.filter((signer: Buffer) => {
               return !signer.equals(beneficiary)
             })
             // Discard votes from removed signer
             this._cliqueLatestVotes = this._cliqueLatestVotes.filter((vote: CliqueVote) => {
               return !vote[1][0].equals(beneficiary)
             })
+          } else {
+            throw new Error('Invalid nonce for clique block: ' + nonce)
```
holgerd77: Nice. 👍

packages/blockchain/src/db/operation.ts
```diff
@@ -21,10 +21,10 @@ export enum DBTarget {
   HashToNumber,
   NumberToHash,
   TotalDifficulty,
-  CliqueSignerStates,
-  CliqueVotes,
   Body,
   Header,
+  CliqueSignerStates,
+  CliqueVotes,
```
holgerd77, line 27: Totally ok with this change but out of curiosity: is the change of ordering here a (potential) problem?

packages/block/test/header.spec.ts
```diff
@@ -90,36 +90,143 @@ tape('[Block]: Header functions', function (t) {
...
+    extraData = Buffer.concat([Buffer.alloc(32), Buffer.alloc(65), Buffer.alloc(20), Buffer.alloc(21)])
+    const epoch = new BN(common.consensusConfig().epoch)
+    header = BlockHeader.fromHeaderData({ ...data, number: epoch, extraData }, opts)
+    try {
+      await header.validate(blockchain)
+      st.fail(testCase)
+    } catch (error) {
+      st.equals(error.message, 'invalid signer list length in extraData, received signer length of 41 (not divisible by 20)', testCase)
+    }
+
```
holgerd77, line 201: Nice test additions. 👍

packages/block/src/header.ts
```diff
@@ -616,8 +616,8 @@ export class BlockHeader {
   /**
    * Returns the signer address
    */
-  cliqueSignatureToAddress(): Address {
-    this._requireClique('cliqueSignatureToAddress')
+  cliqueSigner(): Address {
```
holgerd77, line 616: Yeah, that's really a nice simplification, didn't look through that on first iteration.

packages/block/src/block.ts
```diff
@@ -172,7 +172,7 @@ export class Block {
    * transition block and should therefore be used
    * in conjunction with `cliqueIsEpochTransition()`
    */
-  cliqueEpochTransitionSigners(): Buffer[] {
+  cliqueEpochTransitionSigners(): Address[] {
```
holgerd77, line 175: Yes, these `Address` changes are also nice.

holgerd77 (commented):
Reviewed up till 893a0b48fab621330968710613b5e212dadd0a52

Everything ok. 😄

packages/blockchain/src/index.ts
ryanio: probably can be moved inside, I wasn't sure if it would have a problem retrieving it later with a key NotFoundError, but I will check.

packages/blockchain/src/db/operation.ts
ryanio: I was debugging some db issues and thought it might have been, but in the end I don't think it was a problem.

I kept it this way though because otherwise it changed the enum integers for `Body, Header` and figured it would be better to maintain the existing numbering (and also found it a bit easier to read having them separate at the bottom)

packages/blockchain/src/index.ts
ryanio: added a commit for this (https://github.com/ethereumjs/ethereumjs-monorepo/pull/1032/commits/805a97361d7a174167160a3375b0a1cc0eff975c)

ryanio:
okay I believe I've completed all remaining TODOs! all tests are passing.

The last task is to handle reorgs by removing added signers and votes for deleted blocks, which I will add soon.

I also wanted to ask if we should additionally validate the `difficulty` field, I don't think I saw any logic for that yet:
> difficulty: Contains the standalone score of the block to derive the quality of a chain.
Must be DIFF_NOTURN (1) if BLOCK_NUMBER % SIGNER_COUNT != SIGNER_INDEX
Must be DIFF_INTURN (2) if BLOCK_NUMBER % SIGNER_COUNT == SIGNER_INDEX

Another validation we can add is that `ommersHash` is equal to  `UNCLE_HASH` (`Keccak256(RLP([]))`), I don't think it's critical, but a requirement of the spec.

holgerd77:
@ryanio I honestly didn't understand this whole turn-no-turn part of the spec. 😋  If you get it just implement, at the moment it is not taken in (as I also perceived this as non-mission-critical, but of course would be good to have).

The uncle validation is automatically done in [Block.validateUnclesHash()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/block.ts#L241) which is comparing against `(Keccak256(RLP([]))` if the list of uncles (as checked) is empty.

holgerd77:
Have added a new factory helper method `Blockchain.fromBlocksData()` here 72e6fb2d78975d5b38e615ab0154203581cebec4 for testing in the client but also for general usage, this might also be helpful for the tests here. Eventually in a subsequent PR since this needs some time to get merged in along with the other changes on client execution (will give some description along the PR later on).

packages/block/src/header.ts
```diff
@@ -477,6 +526,103 @@ export class BlockHeader {
...
+    this._requireClique('cliqueExtraVanity')
+    return this.extraData.slice(0, CLIQUE_EXTRA_VANITY)
+  }
+
+  /**
+   * Returns extra seal data
+   * (only clique PoA, throws otherwise)
+   */
+  cliqueExtraSeal(): Buffer {
+    this._requireClique('cliqueExtraSeal')
```
jochem-brouwer, line 568: We can possible write these fancy `_requireClique` stuff as a [decorator](https://www.typescriptlang.org/docs/handbook/decorators.html#parameter-decorators). I have never used this before, but looks like these decorators are meant for exactly this!

ryanio: great idea! I will do that

holgerd77:
@ryanio can you please remove one of the labels "WIP" and "Ready for review" here? Always confused when I recheck here. 😅  So what's the status?

ryanio:
@holgerd77 ah yes sorry almost there, targeting to finish today, the last things on my list are:

* adding last test cases in EIP-225
* creating a method decorator to replace `_requireClique()` as suggested by jochem above
  * thinking `@requireConsensus('clique')`
* re-org support to remove removed blocks from `_cliqueLatestSignerStates` and `_cliqueLatestVotes`
  * and some test cases to ensure it's functioning properly

packages/block/src/header.ts
ryanio: ok so I tried to implement this but unfortunately the decorator is executed when the class is defined, not instantiated, so it’s not possible to check against the set common for something like `@requireConsensus('clique')`

ryanio:
ok, I believe I've completed everything and this is now ready for review!

I will give it one last look over tomorrow to see if I can add any helpful additional clarifying comments or typedocs, but meanwhile it is open for review.

Please let me know if anything can be improved or clarified. Thanks!

packages/block/src/from-rpc.ts
```diff
@@ -44,5 +44,5 @@ export default function blockFromRpc(blockParams: any, uncles: any[] = [], optio
 
   const uncleHeaders = uncles.map((uh) => blockHeaderFromRpc(uh, options))
 
-  return Block.fromBlockData({ header, transactions, uncleHeaders })
+  return Block.fromBlockData({ header, transactions, uncleHeaders }, options)
```
ryanio, line 47: This is a bug fix 🥴 I found common wasn't being passed properly to the returned block in `blockFromRpc` when using it for the [goerli block test](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1032/files#diff-5e26767037750597317a815e2bc28ff13e85897dde318118ad5a392d9ba10ddaR105).

ryanio:
One thing we can consider adding in the future is a helper to create a block with a certain clique signature seal, otherwise we currently have to manually put together the `extraData` which can be a bit tedious. For now this might only be useful for making test data/cases (we do this in `test/blockchain/clique.spec.ts:addNextBlock`) until we start producing signed blocks in the client.

ryanio:
This is also one last implementation detail in EIP-225 for the client when broadcasting clique blocks, noting here for subsequent work:
> ### Authorization strategies
> As long as signers conform to the above specs, they can authorize and distribute blocks as they see fit. The following suggested strategy will however reduce network traffic and small forks, so it’s a suggested feature:
> * If a signer is allowed to sign a block (is on the authorized list and didn’t sign recently).
>   * Calculate the optimal signing time of the next block (parent + `BLOCK_PERIOD`).
>   * If the signer is in-turn, wait for the exact time to arrive, sign and broadcast immediately.
>   * If the signer is out-of-turn, delay signing by `rand(SIGNER_COUNT * 500ms)`.
>
> This small strategy will ensure that the in-turn signer (who’s block weighs more) has a slight advantage to sign and propagate versus the out-of-turn signers. Also the scheme allows a bit of scale with the increase of the number of signers.

holgerd77:
> One thing we can consider adding in the future is a helper to create a block with a certain clique signature seal, otherwise we currently have to manually put together the `extraData` which can be a bit tedious. For now this might only be useful for making test data/cases (we do this in `test/blockchain/clique.spec.ts:addNextBlock`) until we start producing signed blocks in the client.

Yes, I thought this as well and already wanted to add, just couldn't come up with an API I was satisfied with so I put this aside. Think we should add this rather sooner than later, I would assume this is a somewhat common use case also useful beside our own signed block creation one.

holgerd77 (commented):
Ok, have reviewed the rest of the commits, thanks Ryan, really great PR with these substantial functionality additions and completions and extensive test additions.

Have some comments (likely) to address, will nevertheless give this a go to have this huge PR out of the way.

I will also merge with admin merge for procedural reasons, since Ryan has already reviewed my initial code and I have now done the review of the remaining code changes, so technically this should make sense, hope that is ok.

packages/common/src/types.ts
```diff
@@ -27,6 +27,10 @@ export interface Chain {
   consensus?: {
     type: string
     algorithm: string
+    clique: {
+      period: number
+      epoch: number
+    }
```
holgerd77, line 33: This was too strict and breaks typing in non-PoA contexts, have fixed this in 6fe4a1ddab7c78a5551dce139de4b3e988f2cd7b.

packages/block/src/types.ts
```diff
@@ -130,4 +130,5 @@ export interface JsonHeader {
 
 export interface Blockchain {
   getBlock(hash: Buffer): Promise<Block>
+  cliqueActiveSigners(): Address[]
```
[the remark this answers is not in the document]
ryanio: ah I needed this for the `validateCliqueDifficulty` method which needs to calculate if the signer is in or out of turn. I thought it would be good in the block header next to the pow difficulty check method, but you are right that it adds to the blockchain interface here. I can move it over to blockchain and leave the interface here as before.

packages/blockchain/src/index.ts
```diff
@@ -460,12 +467,21 @@ export default class Blockchain implements BlockchainInterface {
         }
       }
     }
+
+    // trim length to CLIQUE_SIGNER_HISTORY_BLOCK_LIMIT
+    const length = this._cliqueLatestVotes.length
+    const limit = this.CLIQUE_SIGNER_HISTORY_BLOCK_LIMIT
+    if (length > limit) {
+      this._cliqueLatestVotes = this._cliqueLatestVotes.slice(length - limit, length)
+    }
```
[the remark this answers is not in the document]
ryanio: right now all votes are being cleared on epoch transitions: https://github.com/ethereumjs/ethereumjs-monorepo/blob/6721321fb46e8543eeb1027d11fdd8cdf8841994/packages/blockchain/src/index.ts#L822-L826

so I can implement the above logic you suggested instead. 

Is maybe 50 blocks a safe enough waiting period? I have no personal experience to get a feel for how long reorgs could possibly get on poa chains.

ryanio:
> > One thing we can consider adding in the future is a helper to create a block with a certain clique signature seal, otherwise we currently have to manually put together the `extraData` which can be a bit tedious. For now this might only be useful for making test data/cases (we do this in `test/blockchain/clique.spec.ts:addNextBlock`) until we start producing signed blocks in the client.
> 
> Yes, I thought this as well and already wanted to add, just couldn't come up with an API I was satisfied with so I put this aside. Think we should add this rather sooner than later, I would assume this is a somewhat common use case also useful beside our own signed block creation one.

just thinking about this, how about a new BlockOption:
```
/**
 * Provide a [Address, privateKey] to seal clique block with this signer.
 * If present, will use maximum 32 bytes of the passed-in extraData, then sealed.
 * Will throw if provided on non-clique chains.
 */
cliqueSigner?: [Address, Buffer]
```
then I can create a helper method that would easily concat it with the passed-in (or blank) extraData.

packages/common/src/types.ts
ryanio: ok great, I was wondering, it might be helpful if all the options were optional, at least for my use in `forCustomChain` , I only wanted to set `epoch` to `3`, but I had to provide all the other values:

https://github.com/ethereumjs/ethereumjs-monorepo/blob/6721321fb46e8543eeb1027d11fdd8cdf8841994/packages/blockchain/test/clique.spec.ts#L535-L548

I understand this is a limitation of how the code is set up in `forCustomChain` though, since it's using the spread operator, it does not allow for deep nesting replacement of single values, it would replace the whole block provided. Not sure if there is a better technique we could use.

packages/blockchain/src/index.ts
```diff
@@ -678,10 +678,15 @@ export default class Blockchain implements BlockchainInterface {
...
-            // Add eventual new header vote on non-epoch transition blocks
+
+            // validate checkpoint signers towards active signers
+            const checkpointSigners = header.cliqueEpochTransitionSigners()
+            const activeSigners = this.cliqueActiveSigners()
+            for (const cSigner of checkpointSigners) {
+              if (!activeSigners.find(aSigner => aSigner.toBuffer().equals(cSigner.toBuffer()))) {
+                throw new Error('checkpoint signer not found in active signers list: ' + cSigner.toString())
+              }
+            }
```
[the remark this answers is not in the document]
ryanio: oh right good point, I will change to a strict equality check of the same order, that would be much better.

packages/block/src/types.ts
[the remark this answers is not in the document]
holgerd77: Just rechecked, so if this is used for typing in `Block` functions like `async validate(blockchain: Blockchain): Promise<void>` then this change is definitely breaking on backwards-compatibility, so we should find another solution here.

Maybe alternatively just leave the interface as-is and manually check in `validateCliqueDifficulty()` for `blockchain.cliqueActiveSigners()` to be present and if not throw? And additionally add some notes on the `blockchain` consuming methods in `Block` that these methods needs a clique supporting `Blockchain` object if run with a PoA setting? 🤔

Not sure if this is the best/optimal solution though.

packages/blockchain/src/index.ts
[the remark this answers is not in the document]
holgerd77: I don't actually think it's too tricky: if there is a reorg, then all vote and signer state entries with block numbers higher than the reorg block number will be deleted, the signers from the  last singer state stack item will be become the new active signers. Everything else from that moment on the continues (restarts) as before.

If I don't get this wrong, votes need to be kept relatively long in this scenario, so i think "EPOCH_LENGTH + REORG_SAFE_PERIOD" block numbers - so something like 30050, since if a reorg is happening from let's say 30020 -> 29950 and there is voting taking place in these last 50 blocks before the epoch transition, theoretically all existing vote from this 0-29999 epoch might be relevant for the pre-epoch voting. Does this make sense?

For Signer States I would think - all states from the last REORG_SAFE_PERIOD blocks + minimally one state - should be enough if I am not forgetting something.

But I think we can also generally be a bit more generous here - e.g. take a somewhat longer REORG_SAFE_PERIOD of 100 or something or keep more signer states, nothing lost here, this doesn't take that much space and eventually the date might be interesting for analytical purposes.

packages/common/src/types.ts
holgerd77: Wouldn't be a fan if we loosen this here in general since the type safety of these chain and HF definition JSON files is so underdeveloped already and should rather be improved towards the other side at some PiT. I get the point though. Allow `any` here for the function signature as the most simple/naive solution? 😜  Not sure, if there is something better, this should of course be preferred.

holgerd77:
> > > One thing we can consider adding in the future is a helper to create a block with a certain clique signature seal, otherwise we currently have to manually put together the `extraData` which can be a bit tedious. For now this might only be useful for making test data/cases (we do this in `test/blockchain/clique.spec.ts:addNextBlock`) until we start producing signed blocks in the client.
> > 
> > 
> > Yes, I thought this as well and already wanted to add, just couldn't come up with an API I was satisfied with so I put this aside. Think we should add this rather sooner than later, I would assume this is a somewhat common use case also useful beside our own signed block creation one.
> 
> just thinking about this, how about a new BlockOption:
> 
> ```
> /**
>  * Provide a [Address, privateKey] to seal clique block with this signer.
>  * Will throw if provided on non-clique chains.
>  */
> cliqueSigner?: [Address, Buffer]
> ```
> 
> then I can create a helper method that would easily concat it with the passed-in (or blank) extraData.
> 
> (edit: actually looks like we only need the privateKey, not the address)

I think that's an option, at some point I also thought about adding an option to the constructor, do you think this is ok from a security PoV to request the private key from the signer? This was my main reason to hesitate, but rather based on some blurry feeling than some more solid analysis here.

P.S.: not sure if this was clever - at least it gives a neutral perspective - just to let you know that I answered these questions here *before* looking at your follow-up clique PR #1074.

packages/block/src/types.ts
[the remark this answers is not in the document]
ryanio: that sounds like a good solution for now, I like keeping the validation in the block header since it's a logical place to validate the difficulty (although it requires the context of blockchain).

i can be convinced to move it but let's see how the solution looks this way. (added in commit: https://github.com/ethereumjs/ethereumjs-monorepo/pull/1074/commits/d58a901b649940b89c0002941f0813c1a2b6cfd2)

packages/common/src/types.ts
ryanio: yeah, agree, in general would love some more type safety in common coming from the json files.

what I meant though is I would love to do this:

```
 const common = Common.forCustomChain( 
   'rinkeby',
   { consensus: { clique: { epoch: 3 } } }, 
   'chainstart' 
 )
```

however the current way the code is set up, then` { consensus: type, algorithm, clique: { period } }` would all be undefined. we would need some more sophisticated logic for deeply replacing nested values, rather than the simple solution of using the spread operator ([here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/55f640003c9488fe9285d2c86b0c49bce643d0b5/packages/common/src/index.ts#L89)).

holgerd77: Ah ok, thanks for the clarification.

PR #968, opened 2020-11-26
Add support for EWC/Volta Networks

matextrem:
These changes add support for `EWC/Volta` networks. More info about this chain can be found [here.](https://energyweb.org/technology/energy-web-chain/)

Issue #905, opened 2020-10-07
Tiny Fixes Meta Issue


holgerd77:
This is a meta issue to collect small fixes not worth a dedicated PR. Please collect in the TODO list below and check or delete from the TODO list once a PR is up (so not only when it is merged but on push):

#### common

- [ ] Minimum HF for EIP-3198 (BASEFEE opcode) is still set to `london` and it is therefore not possible to run in isolation. This should be switched to `berlin` and instead 1559 added as a required EIP (with `"requiredEIPs": [1559]`). This will allow to instantiate with `berlin` and then just `1559` and `3198` being activated
- [ ] Rename `tests/customChains.ts` -> `tests/customChains.spec.ts` for being included in the test runner
- [ ] Expose `_calcNextForkHash()` method (-> `calcNextForkHash()`, include in API docs)

#### devp2p

- [x] Update `devp2p` [README](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/devp2p) header, remove very old `node-devp2p` package references (directly in first section)

#### client

- [ ] Add suggestion from https://github.com/ethereumjs/ethereumjs-monorepo/pull/1428#discussion_r697614565
- [ ] Add suggestion from https://github.com/ethereumjs/ethereumjs-monorepo/pull/1538/files#r734001330

#### tx

- [x] Inline default `baseFee` value in `getUpfrontCost` ([link](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1148#discussion_r637437606)) (added [here](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1272/files#r641885256))
- [ ] Add note to access list API being experimental, see #1267 
- [ ] 2930 uses yParity, 1559 uses v (from @jochem-brouwer, not sure about the context, needs look into the code base)
- [ ] Update all internal monorepo references from deprecated `tx.transactionType` to `tx.type`

#### block

- [x] Block import package name is wrong in the v3.3.0 [release note](https://github.com/ethereumjs/ethereumjs-monorepo/releases/tag/%40ethereumjs%2Fblock%403.3.0) example: update from `ethereumjs-block` to `@ethereumjs/block`
  - [x] also add `import { BN } from 'ethereumjs-util'` for complete example
  - [x] add example with `calcNextBaseFee()` to block README.

#### VM

[...]

PR #1034, opened 2021-01-07
Common: Custom Chain File Integration

holgerd77:
This PR will be adding functionality for a more convenient and flexible way to integrate custom chains into Common instances,  will push a more complete code base here throughout the day.

This is an early push since I also improved some typings along the way on functions like `hardfork()` or `consensusType()` and I was not sure if this might have side effects, so I wanted to run the additional tests from CI.


holgerd77:

To add on the missing explanation: this PR introduces a more flexible way to add custom chains with a new optional `customChains` constructor option where an arbitrary number of chains can be passed in. This should allow for a range of new use cases, e.g. if people want to build some library/tools integrating some of our libraries or just Common and supporting a range of their own custom chains (or just: some other chains than the ones we support).

The PR furthermore improves on the overall custom chain situation and combines all the custom chain tests into a dedicated test file and comes with a revised README section. Hope this all sets us better up for custom chain support.

One special thing: I added a new (still optional, backwards compatibility) parameter `defaultHardfork` to the chain files. This allows - e.g. on custom chains - to define by-chain default HFs and is especially useful for chains which just don't include the currently defined default HF, this would otherwise be a constant source of error with initialization throwing if people not explicitly pass an existing HF for their custom chain. For now the solution is fully backwards compatible, pointing to `istanbul` for all existing supported chains. We might want to leverage this ourselves in the future since this can allow us to solve situations which lately occurred with `muirGlacier` where we needed to stick to `istanbul` on all chains as default while the PoW chains could have moved on to `muirGlacier`. With the new option we have at least the possibility for a more fine grained approach if we want to. 😄

jochem-brouwer:
Will give this a review in a minute. Just a comment here:

> where we needed to stick to istanbul on all chains as default while the PoW chains could have moved on to `muirGlacier`

This only applies to ETH mainnet as this delays the difficulty bomb by a certain amount and is thus dependent on the height of a chain. (I'm not sure if ETC has something similar)

packages/common/README.md
```diff
@@ -104,12 +104,41 @@ files like `mainnet.json` in the `chains` directory, or to the `Chain` type in [
...
+const customChainParams = { name: 'custom', chainId: 123, networkId: 678 }
+const customChainCommon = Common.forCustomChain('mainnet', customChainParams, 'byzantium')
+```
+
+### Initialize using customChains Array
+
+A second way for custom chain initialization is to use the `customChains` constructor option. This
+option comes with more flexibility and allows for an activation-independent initialization (e.g. to
+initialize with `mainnet` activated but with an additional `myCustomChain` available) as well as
+initializing with an arbitrary number of your own custom chain configuration objects:
```
jochem-brouwer, line 135: From this description I am not entirely sure what this does. I can initialize a common which is set to the default `mainnet`, but I can also pass an array with extra chain parameters - how do I activate them? Should I activate them by setting it as hardfork?

packages/common/src/chains/index.ts
```diff
@@ -1,16 +1,42 @@
...
+  }
+  if (customChains) {
+    for (const chain of customChains) {
+      const name = chain.name
+      names[chain.chainId.toString()] = name
+      chains[name] = chain
+    }
+  }
+
+  chains['names'] = names
```
jochem-brouwer, line 34: Hmm. So custom chains are treated on a different foot than "normal" chains. Can we either move these custom chains into the special field of `chains.names` or disallow the custom chain name `names`? (Very unlikely to happen, but it will probably mess up the internal logic if someone uses this name).

packages/common/src/index.ts
```diff
@@ -79,17 +93,19 @@ export default class Common {
...
+        return initializedChains[name]
       }
 
       throw new Error(`Chain with ID ${chain} not supported`)
     }
 
-    if (chainParams[chain]) {
-      return chainParams[chain]
+    if (initializedChains[chain]) {
+      return initializedChains[chain]
```
jochem-brouwer, line 108: Ok, so I can only get a custom chain by name, not by ID, right?

packages/common/src/index.ts
```diff
@@ -79,17 +93,19 @@ export default class Common {
     })
   }
 
-  private static _getChainParams(chain: string | number): Chain {
+  private static _getChainParams(chain: string | number, customChains?: Chain[]): Chain {
```
jochem-brouwer, line 96: Here `customChains` is an argument - can we not use `this.customChains` and remove the argument?

packages/common/src/types.ts
```diff
@@ -16,11 +16,18 @@ export interface Chain {
   name: string
   chainId: number
   networkId: number
+  // TODO: make mandatory in next breaking release
+  defaultHardfork?: string
```
jochem-brouwer, line 20: I was about to suggest this to indeed make `defaultHardfork` mandatory.

packages/common/tests/data/energyWebChain.json
```diff
@@ -0,0 +1,79 @@
+{
+  "name": "energyWebChain",
```
jochem-brouwer, line 2: Awesome test data 😄

packages/common/tests/customChains.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    st.end()
+  })
+
+  t.test('customChains parameter: initialization', (st) => {
+    let c = new Common({ chain: 'mainnet', hardfork: 'byzantium', customChains: [testnet] })
+    st.equal(c.chainName(), 'mainnet', 'customChains, chain set to supported chain')
+    st.equal(c.hardforkBlock(), 4370000, 'customChains, chain set to supported chain')
+
+    c.setChain('testnet')
```
jochem-brouwer, line 83: Ah OK so this is how it works - could this be added in the README?

jochem-brouwer (commented):
Some questions, but looks great in general 😄

packages/common/src/chains/mainnet.json
```diff
@@ -2,6 +2,7 @@
   "name": "mainnet",
   "chainId": 1,
   "networkId": 1,
+  "defaultHardfork": "istanbul",
```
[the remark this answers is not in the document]
holgerd77: No, this would be breaking if we change the default HF along the road. We would need to wait until the next major release for VM and Common.

packages/common/README.md
```diff
@@ -104,12 +104,41 @@ files like `mainnet.json` in the `chains` directory, or to the `Chain` type in [
 
 ### Working with private/custom chains
 
-There are two ways to set up a common instance with parameters for a private/custom chain:
+There are two distinct APIs availale for setting up custom(ized) chains.
```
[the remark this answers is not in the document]
holgerd77: Updated (next push).

packages/common/README.md
holgerd77: Have rewritten this, hope this makes it more clear.

packages/common/src/chains/index.ts
holgerd77: Are you maybe misreading the code here (or I get your message wrong)? Custom chains are just normally added to the `names` dict here and the other entries are not affected in any way.

Generally I would want to move away from this `chains/index.ts` file altogether - therefore the deprecation notes. I would already want to move this to logic to the main `index.ts` file but then I kept it like this also for some backwards compatibility considerations.

packages/common/src/index.ts
holgerd77: No, since the method is static and custom chains from the instance are not available yet. This is a private method though, so should be unproblematic.

packages/common/src/index.ts
holgerd77: No, that's the same logic as before here - `chainParams` -> `initializedChains` is just a renaming. The code for returning from a number is some lines above and still works for baked-in as well as custom chains. In fact baked-in chains and custom chains are treated exactly the same throughout the whole library I would assume (if I am not missing something).

packages/common/src/types.ts
holgerd77: If someone is using this type this could otherwise break things unfortunately.

packages/common/tests/customChains.ts
holgerd77: Have added.

packages/common/tests/data/energyWebChain.json
holgerd77: This is taken from #968 😄 , I would actually want to wait for a quick ok from @matextrem the author until we merge here.

holgerd77:
Ok, did the changes and rebased.

jochem-brouwer:
Added a `blocked` label to await @matextrem's approval for the testdata.

packages/common/src/chains/index.ts
jochem-brouwer: At line 30 it sets `chains[name]` to `chain` - so if you'd use the `name` "names", then it'd first set `chains.names` to the chain and then later overwrite it. Not sure if that causes problems?

packages/common/src/index.ts
jochem-brouwer: Ah - woops - did not see that this was a `static` method.

packages/common/src/index.ts
jochem-brouwer: Yup you are right, I misread this code.

packages/common/src/types.ts
jochem-brouwer: Yep but we can do this in the next major release (maybe add this to the V6 tracking issue?)

jochem-brouwer (dismissed):
Will approve but lets await @matextrem's approval for the test data.

packages/common/README.md
```diff
@@ -104,12 +104,48 @@ files like `mainnet.json` in the `chains` directory, or to the `Chain` type in [
...
+If you want to initialize a `Common` instance with a single custom chain which is then directly activated
+you can pass a dictionary - conforming to the parameter format described above - with your custom chain 
+values to the constructor using the `chain` parameter or the `setChain()` method, here is some example:
+
+```typescript
+import myCustomChain from './[PATH]/myCustomChain.json'
+const common = new Common({ chain: myCustomChain })
+```
+
+If you just want to change some certain parameters on a chain configuration it can also be conveniened to use
```
cgewecke, line 120:
```suggestion
If you just want to change certain parameters on a chain configuration it can also be convenient to use 
```

cgewecke (dismissed):
LGTM!

holgerd77:
@matextrem Is it ok if we use your examples here? I would otherwise push another commit tomorrow and remove again.

holgerd77:
Ok, I now replaced the test files here with some more neutral onces. This would need another approval for merge.

cgewecke (approved):
Test file changes LGTM. 

Re-approving 🙂

ryanio:
ah this didn't make it: https://github.com/ethereumjs/ethereumjs-monorepo/pull/1034#discussion_r553517646 would be a good typo to fix

holgerd77:
@ryanio have added to #905

PR #1035, opened 2021-01-07
Trie: cache CheckpointTrie in memory

jochem-brouwer:
This PR: 

- Seperates the DB used by `BaseTrie` and the extended DB used by `CheckpointTrie` (and `SecureTrie`) into two DBs; the base `DB` and the `CheckpointDB`. 
- All checkpoint operations are cached into memory. This makes reverting operations very easy (just throw away the latest cache) and also does not write to disk (so less I/O operations)

### DIFF
--- a/packages/trie/src/checkpointDb.ts
+++ b/packages/trie/src/checkpointDb.ts
@@ -1,0 +1,162 @@
+import { LevelUp } from 'levelup'
+import { DB } from './db'
+const level = require('level-mem')
+
+export const ENCODING_OPTS = { keyEncoding: 'binary', valueEncoding: 'binary' }
+
+export type Checkpoint = {
+  keyValueMap: Map<Buffer, Buffer | null>,
+  root: Buffer
+}
+
+export type BatchDBOp = PutBatch | DelBatch
+export interface PutBatch {
+  type: 'put'
+  key: Buffer
+  value: Buffer
+}
+export interface DelBatch {
+  type: 'del'
+  key: Buffer
+}
+
+/**
+ * DB is a thin wrapper around the underlying levelup db,
+ * which validates inputs and sets encoding type.
+ */
+export class CheckpointDB extends DB {
+  public checkpoints: Checkpoint[]
+
+  /**
+   * Initialize a DB instance. If `leveldb` is not provided, DB
+   * defaults to an [in-memory store](https://github.com/Level/memdown).
+   * @param leveldb - An abstract-leveldown compliant store
+   */
+  constructor(leveldb?: LevelUp) {
+    super(leveldb)
+    // Roots of trie at the moment of checkpoint
+    this.checkpoints = []
+  }
+
+  /**
+   * Is the DB during a checkpoint phase?
+   */
+  get isCheckpoint() {
+    return this.checkpoints.length > 0
+  }
+
+  /**
+   * Adds a new checkpoint to the stack
+   * @param root
+   */
+  checkpoint(root: Buffer) {
+    this.checkpoints.push({ keyValueMap: new Map<Buffer, Buffer>(), root })
+  }
+
+  /**
+   * Commits the latest checkpoint
+   */
+  async commit() {
+    const { keyValueMap } = this.checkpoints.pop()!
+    if (!this.isCheckpoint) {
+      // This was the final checkpoint, we should now commit and flush everything to disk
+      const batchOp: BatchDBOp[] = []
+      keyValueMap.forEach(function(value, key) {
+        if (value === null) {
+          batchOp.push({
+            type: 'del',
+            key
+          })
+        } else {
+          batchOp.push({
+            type: 'put',
+            key,
+            value
+          })
+        }
+      })
+      await this.batch(batchOp)
+    } else {
+      // dump everything into the current (higher level) cache
+      const currentKeyValueMap = this.checkpoints[this.checkpoints.length - 1].keyValueMap
+      keyValueMap.forEach(function(value, key){ 
+        currentKeyValueMap.set(key, value)
+      })
+
+    }
+
+  }
+
+  /**
+   * Reverts the latest checkpoint
+   */
+  async revert() {
+   const { root } = this.checkpoints.pop()!
+   return root
+  }
+
+  /**
+   * Retrieves a raw value from leveldb.
+   * @param key
+   * @returns A Promise that resolves to `Buffer` if a value is found or `null` if no value is found.
+   */
+  async get(key: Buffer): Promise<Buffer | null> {
+    // Lookup the value in our cache. We return the latest checkpointed value (which should be the value on disk)
+    for (let index = this.checkpoints.length; index >= 0; index--) {
+      const value = this.checkpoints[index].keyValueMap.get(key)
+      if (value != undefined) {
+        return value
+      }
+    }
+    // Nothing has been found in cache, look up from disk
+    // TODO: put value in cache.
+    return await super.get(key)
+  }
+
+  /**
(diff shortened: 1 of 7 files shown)


packages/trie/src/checkpointDb.ts
```diff
@@ -0,0 +1,162 @@
...
+export type BatchDBOp = PutBatch | DelBatch
+export interface PutBatch {
+  type: 'put'
+  key: Buffer
+  value: Buffer
+}
+export interface DelBatch {
+  type: 'del'
+  key: Buffer
+}
```
ryanio, line 21: these can be imported from `db.ts`

jochem-brouwer: Ah yep thanks, I actually copied the original and then removed stuff. Should indeed import from the original 👍

packages/trie/src/checkpointDb.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * Writes a value directly to leveldb.
+   * @param key The key as a `Buffer`
+   * @param value The value to be stored
+   */
+  async put(key: Buffer, val: Buffer): Promise<void> {
+    if (this.isCheckpoint) {
+      // put value in cache
+      this.checkpoints[this.checkpoints.length - 1].keyValueMap.set(key.toString('hex'), val)
```
jochem-brouwer, line 112: There was a weird bug where, if you would `set` the value to just the raw `key.toString()`, then if you'd use `Buffer.from(key)` you'd get a different buffer. So I have settled with hex encoding here, but it should be faster/better if we can just use the raw buffer so we don't have to convert it (maybe has to do with some encoding issue?). Very weird.

jochem-brouwer:
Hmm, weird that it crashes on a random blockchain test while passing thousands of others 🤔

jochem-brouwer:
Very weird as the Trie changes apparently found some edge case in the blockchain tests which are unrelated to the changes in Trie. Have fixed.

packages/trie/src/checkpointDb.ts
ryanio: I think hex encoding is the right option here.

I found in the Buffer [docs](https://nodejs.org/api/buffer.html#buffer_buf_tostring_encoding_start_end) that it says:
> If encoding is 'utf8' and a byte sequence in the input is not valid UTF-8, then each invalid byte is replaced with the replacement character U+FFFD.

so I wonder if that could be related to the bug you encountered.

Another idea I had is that base64 encoding would actually output a shorter string for the key, but I'm unsure if that would make a significant difference in the performance of the map. How large could these checkpoint maps conceivably get?

packages/trie/src/checkpointDb.ts
```diff
@@ -0,0 +1,156 @@
+import { LevelUp } from 'levelup'
+import { DB, BatchDBOp } from './db'
+
+export const ENCODING_OPTS = { keyEncoding: 'binary', valueEncoding: 'binary' }
```
ryanio, line 4: this can also be imported from `./db`

packages/trie/src/secure.ts
```diff
@@ -89,8 +89,11 @@ export class SecureTrie extends CheckpointTrie {
    * @param includeCheckpoints - If true and during a checkpoint, the copy will contain the checkpointing metadata and will use the same scratch as underlying db.
    */
   copy(includeCheckpoints = true): SecureTrie {
-    const trie = super.copy(includeCheckpoints)
-    const db = trie.db.copy()
-    return new SecureTrie(db._leveldb, this.root)
+    const db = this.db.copy()
+    const secureTrie = new SecureTrie(db._leveldb, this.root)
+    if (includeCheckpoints && this.isCheckpoint) {
+      secureTrie.db.checkpoints = this.db.checkpoints.slice()
```
ryanio, line 95: is this `slice` for a (shallow) copy?

I've never seen `slice` without any params, in ES6 you could use the spread syntax: `[...this.db.checkpoints]`

packages/trie/src/checkpointDb.ts
```diff
@@ -0,0 +1,156 @@
...
+          })
+        }
+      })
+      await this.batch(batchOp)
+    } else {
+      // dump everything into the current (higher level) cache
+      const currentKeyValueMap = this.checkpoints[this.checkpoints.length - 1].keyValueMap
+      keyValueMap.forEach(function (value, key) {
+        currentKeyValueMap.set(key, value)
+      })
```
ryanio, line 74: more of a nit than anything else, but this could be written shorthand in ES6:
`keyValueMap.forEach((value, key) => currentKeyValueMap.set(key, value))`

ryanio:
overall this looks really awesome, nice inline comments that provide explanation.

just left a few small comments in review, otherwise will approve 👍

packages/trie/src/secure.ts
jochem-brouwer: Yep this copies the array. I took this from the existing implementation of `CheckpointTrie.copy` - but I agree it looks a bit confusing. Will improve.

packages/trie/src/checkpointDb.ts
jochem-brouwer: These checkpoint maps can get huge, since we checkpoint the trie before the block and then flush it when the block is done. The only thing I'm pretty scary of is when we hit an extremely bad block which hits a lot of trie items and caches these all, this could probably run into some memory issues... 🤔 

The problem I have with this hex encoding is that we actually convert these raw bytes to strings. So if you'd use one byte (let's say `0xff`) then internally it will be represented as `"ff"` which is 2 bytes, so this doubles the strain on memory. I'll take a look at this encoding, maybe if I explicitly pass `utf8` as encoding it works.

holgerd77:
Have updated the branch here.

@jochem-brouwer have you run this PR in the client? How does this behave?

Otherwise once you find some time to address the comments from @ryanio I think we can get this merged relatively soon.

jochem-brouwer:
Hi @holgerd77, thanks for the rebase, I just tested on client and it seems to have improved execution speed.

image


@ryanio ready for re-review 😄

packages/vm/tests/BlockchainTestsRunner.ts
```diff
@@ -73,6 +73,10 @@ export default async function runBlockchainTest(options: any, testData: any, t:
     common,
   })
 
+  // Need to await the init promise: in some tests, we do not run the iterator (which awaits the initPromise)
+  // If the initPromise does not finish, the `rawHead` of `blockchain.meta()` is still `undefined`.
+  await blockchain.initPromise
```
ryanio, line 78: Do you think there is any chance users of the blockchain library may run into this?

If so then as you previously suggested we can make an async public factory method to generate the blockchain and then make the constructor private to reduce the chance of this happening. For example libp2p uses `.create()` as an async method (although they don't restrict the use of the constructor, in their case if you supply a PeerId then it doesn't need to be async and the constructor can be used)

ryanio (approved):
lgtm!

ryanio:
oh if you wanted to include an update to the changelog i will also re-approve 👍

holgerd77 (approved):
I will merge here since I would want to rebase further work on top, hope that's ok. Gave this just a brief look, looks great also with this new `CheckpointDB` subclass, makes a lot of sense.

I also tested in the client with block containing transactions. I also have the impression that this has gotten significantly faster again.

If one is looking at the time diffs from the logs on execution there is also some evidence that this is the case, here is a run from `master`:

image

And here one using the code base from this PR:

image

Seems to be more execution logs per second on the second run. Just a very rough indicator though of course.

Any, @jochem-brouwer, nice work! 😄

PR #1036, opened 2021-01-08
Network improvements

holgerd77:


holgerd77:
Ah, forgot a bit about this PR, this is actually ready for review already for some days. 😄 

As mentioned yesterday in the call, the initial goal here was to improve network connection quality on a level that there is a reliable connection to non-mainnet networks possible, before it was rather not possible at all any more to connect to e.g. `Rinkey` or `Goerli`. I didn't quite get there but the networking quality is generally improved with the PR for all networks and also the non-mainnet networks got better and now it is - sometimes - possible to connect after 1-2 minutes. As mentioned in the call yesterday a way to improve from some other angle would be here to add DNS discovery support, will open a dedicated issue on that.

This PR does the following:

- Some additional RLPx type improvements along 75d224b86c7c746db1a11eccdac420e78dfb5c83
- Analogue to the latest discovery changes from #1031 the `refill()` calls for RLPx are subdivided to better distribute networking traffic, this perceptibly improves the connection 838d6c4a0d8f3b714ace2d8381739b217748d23a
- There was a bug in the client reporting a repeated first block number on executions (in case you already wondered, this was just a logging error), this has been fixed in 88e5f039eb124f03f7dda88047b6e19f1f1e4195
- "devp2p -> connection reliability: fixed an error in DPT not properly banning old peers and replacing with a new peer on KBucket ping" (from the commit message): this is a super important bug fix, before the ban functionality for discovery for peers just didn't work due to a generic catch block shading the error messages. This lead to bad peers just being used for connections over and over again. Bad. There are some more of these generic catch blocks (so not containing any code) in the code base. It's a bit less trivial to fix though than one would think on first thought since this needs some thought on well-balancing how to instead handling potential errors, since there is some mix of errors which are just expected, some which one might want to know about and some which are severe. And we have to decide how to react properly (re-throw, emit an error event, display a debug message, do nothing) without polluting the client log e.g.. 06f31ea42fe5617b784ec4fd07de21fe1221ffdd
- There was another traffic cluster caused along neighbour retrieval from the DPT server in `_onServerPeers`, this has been smoothened in c110b70b86b09d53f117b73568390e846b33cf31

packages/devp2p/src/rlpx/rlpx.ts
```diff
@@ -259,17 +267,27 @@ export class RLPx extends EventEmitter {
...
+      // Randomly distributed selector based on peer ID
+      // to decide on subdivided execution
+      const selector = buffer2int((item.peer.id! as Buffer).slice(0, 1)) % 10
+      if (selector === this._refillIntervalSelectionCounter) {
+        this._connectToPeer(item.peer)
+        return false
+      } else {
+        // Still keep peer in queue
+        return true
+      }
```
cgewecke, line 290: Out of curiosity, do you know why this technique improves connection reliability? 

With this change the connection attempts are spaced out and for any given peer there's a 1 in 10 chance it will be made?

cgewecke (approved):
This looks good to me! 

(Just have a question about the how the refill interval changes work.)

packages/devp2p/src/rlpx/rlpx.ts
holgerd77: This distributes the networking traffic over time, so there's not all connections hammered out at once - then followed by a longer non-networking-activity - and subsequently many of the replies from the peers also coming in at a similar PIT (in the old version). Instead the traffic gets out and in more frequently and distributed. I am not 100% sure about the reliability impacts in a narrower sense, so e.g. if this directly prevents connection losses by avoiding concurrent network handling, my deeper networking experience is too limited for that. But this definitely leads to a healthier protocol flow - this long waiting period is avoided - becoming e.g. very apparent when the last peer connection gets lost and there is another 30 second wait - but there is always some activity/connection reattempts going on. I would also think that these distributes better on available protocol respectively implementation-given resource restrictions like the maximum number of allowed peer connections or the number of slots for the queue, but not completely sure here.

However benefit reasons are distributed: this is one of the changes which can be really "felt" when running the client, so the first connection pickup after start is just directly and apparently getting so much faster that it gets a bit out of question if there "might be" some positive effect. 😄

PR #1037, opened 2021-01-09
Generate receipt roots in runBlock

alcuadrado:
Hey guys,

We found a bug in Hardhat network yesterday, that IMO is a bug in `ethereumjs-vm`, so I fixed it :)

I think that if a user calls `vm.runBlock` with the `generate: true`, is because they don't want to validate the block but to  modify the state and update (`generate`) the fields of the block header that can only be computed after executing the txs. If this is the actual intention, then `block.header.receiptTrie` was missing.

This PR just updates that header field, and adds a test for it.

Also, as you can see in the test, using `runBlock` with `generate: true` is really really inconvenient. I don't get why the new `Block` wasn't added to `RunBlockResult`. You can only access it with an event listener. The good news is that adding it is a backwards compatible change. Should I create a PR doing it?

### DIFF
--- a/packages/vm/lib/runBlock.ts
+++ b/packages/vm/lib/runBlock.ts
@@ -171,7 +171,7 @@ export default async function runBlock(this: VM, opts: RunBlockOpts): Promise<Ru
     const bloom = result.bloom.bitvector
     block = Block.fromBlockData({
       ...block,
-      header: { ...block.header, stateRoot, bloom },
+      header: { ...block.header, stateRoot, bloom, receiptTrie: result.receiptRoot },
     })
   } else {
     if (result.receiptRoot && !result.receiptRoot.equals(block.header.receiptTrie)) {
--- a/packages/vm/tests/api/runBlock.spec.ts
+++ b/packages/vm/tests/api/runBlock.spec.ts
@@ -1,13 +1,14 @@
 import tape from 'tape'
-import { Address, BN, rlp } from 'ethereumjs-util'
+import { Address, BN, rlp, KECCAK256_RLP } from 'ethereumjs-util'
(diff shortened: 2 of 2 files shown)


alcuadrado:
btw, I loved these lines updating the block @ryanio https://github.com/ethereumjs/ethereumjs-vm/pull/1037/files#diff-31eaac2b514be99694029f78d9049fe0c99a28d5bb01060644fee69fbe203926R172-R175

holgerd77:
@alcuadrado thanks for the PR, did you guys update to VM v5 and so will you get the update on a normal release? If so, how were your update experiences?

alcuadrado:
Not yet, I also fixed it in a v4 fork. We are currently on holiday seasons, so won't be able to migrate this month. I'll report our experience doing it.

jochem-brouwer:
Great find - you are indeed right that the `receiptsRoot` should also be generated, but also the field `gasUsed` and `logsBloom` should be generated - right?

You are also right that `runBlock` is very annoying for this use case. However, we have an issue which proposes a new static method which actually handles this generating stuff. I think I'm in favor of that - it makes more sense to call some method like `generateBlock` (I don't think this is the best name for the method though, maybe something like `generateBlockFields`?) which handles setting these 4 vars (`stateRoot`, `gasUsed`, `logsBloom` and `receiptsRoot`) and returns the actual "verified" block.

oozayranci:
> btw, I loved these lines updating the block @ryanio https://github.com/ethereumjs/ethereumjs-vm/pull/1037/files#diff-31eaac2b514be99694029f78d9049fe0c99a28d5bb01060644fee69fbe203926R172-R175

thanks for this reply!

PR #1038, opened 2021-01-09
Fix block gas limit validation

alcuadrado:
Another bug found, another fix :)

This PR changes how the block gas limit is validated, as what was previously implemented didn't match what mainnet does.

Before this PR, `runBlock` would check if a tx's gas limit was higher than the gas let (i.e. block gas limit - gas used so far) before running it.  What is actually done on mainnet is first running the tx, and then validating if it consumed too much gas.

This change is really important, as mainnet has lots of blocks where the sum of their txs' gas limits is larger than the block gas limit.

### DIFF
--- a/packages/vm/lib/runBlock.ts
+++ b/packages/vm/lib/runBlock.ts
@@ -253,11 +253,6 @@ async function applyTransactions(this: VM, block: Block, opts: RunBlockOpts) {
   for (let txIdx = 0; txIdx < block.transactions.length; txIdx++) {
     const tx = block.transactions[txIdx]
 
-    const gasLimitIsHigherThanBlock = block.header.gasLimit.lt(tx.gasLimit.add(gasUsed))
-    if (gasLimitIsHigherThanBlock) {
-      throw new Error('tx has a higher gas limit than the block')
-    }
-
     // Run the tx through the VM
     const { skipBalance, skipNonce } = opts
     const txRes = await this.runTx({
@@ -270,6 +265,11 @@ async function applyTransactions(this: VM, block: Block, opts: RunBlockOpts) {
 
     // Add to total block gas usage
     gasUsed = gasUsed.add(txRes.gasUsed)
+
+    if (block.header.gasLimit.lt(gasUsed)) {
+      throw new Error(`The block consumed more gas than its gas limit after running tx ${txIdx}`)
+    }
+
     // Combine blooms via bitwise OR
     bloom.or(txRes.bloom)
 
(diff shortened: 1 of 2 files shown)


packages/vm/lib/runBlock.ts
```diff
@@ -253,11 +253,6 @@ async function applyTransactions(this: VM, block: Block, opts: RunBlockOpts) {
   for (let txIdx = 0; txIdx < block.transactions.length; txIdx++) {
     const tx = block.transactions[txIdx]
 
-    const gasLimitIsHigherThanBlock = block.header.gasLimit.lt(tx.gasLimit.add(gasUsed))
```
alcuadrado, line 256: Note that validating that the tx's gas limit isn't larger than the block's is done in runTx.

jochem-brouwer:
Are you sure about this statement? `What is actually done on mainnet is first running the tx, and then validating if it consumed too much gas.`

I am pretty certain that it is not allowed to run a Tx which has a higher gas limit than the current gas remaining in the block. How did you find this bug?

jochem-brouwer:
image

From the yellow paper.

alcuadrado:
> I am pretty certain that it is not allowed to run a Tx which has a higher gas limit than the current gas remaining in the block. How did you find this bug?

By re-running mainnet forks using hardhat network's fork. I'll find the block tomorrow.

The part of the YP you posted seems to indicate the contrary. I'm confused now.

jochem-brouwer:
It would be nice to see this block, because it might also indicate that we are charging too much gas on one of the txs in the. block 😄

alcuadrado:
> It would be nice to see this block, because it might also indicate that we are charging too much gas on one of the txs in the. block 😄

ohhhh! that has to be the reason! Thanks @jochem-brouwer! I "discovered" this while debugging a gas cost discrepancy 🤦‍♂️

The error was in our custom state manager for Hardhat Network's fork mode.

jochem-brouwer:
Hi @alcuadrado, if it's indeed an internal hardhat bug, could you then close this PR? 😄

alcuadrado:
I couldn't find the block I used, not reproduce it. Closing this. Will get back if I ever find it. Thanks @jochem-brouwer

holgerd77:
@alcuadrado didn't closely read the code or dug into the discussion, but could we eventually repurpose/keep some of the test cases you wrote here? You added so many lines here (and we are generally under-covered regarding the VM API), would be nice if something can be used in some way. 😄

PR #1039, opened 2021-01-09
Add an option to disable the block gas limit check in runTx

### LINKED ISSUE #1014: eth_call can't be implemented without copy-pasting code form the VM
I just found a bug in Hardhat Network's implementation of `eth_call`, which I'm currently fixing it, and realized that I'll have to copy paste a bunch fo code from the VM to do it.

Our implementation of `eth_call` uses `vm.runTx`, instead of `vm.runCall`, as [the latter can fail](https://github.com/ethereumjs/ethereumjs-vm/issues/613). But `runTx` wasn't really designed for this, and I'm hitting a limitation.

When you execute `eth_call`, you can run your call in the context of a block, and it should be as close as possible to running a tx as the last one in the block. This means that you need to keep things like `block.timestamp` and `block.gasLimit` in sync. 

`eth_call` also accepts a gas limit param, which doesn't need to be capped by `block.gasLimit`. Actually, it's normally higher. For example, Geth accepts up to 25000000 regardless of the block gas limit.

My problem is that `runTx` checks that the tx's gas limit is <= than the block's one [here](https://github.com/ethereumjs/ethereumjs-vm/blob/60ed11bbae151b26b151598f0a52c813f719ec6c/packages/vm/lib/runTx.ts#L65-L67). So, I need to do something [like this](https://github.com/ethereumjs/ethereumjs-vm/blob/60ed11bbae151b26b151598f0a52c813f719ec6c/packages/vm/lib/runTx.ts#L70-L79) myself, but `_runTx` is not exported, so I'd need to copy everything.

As a "workaround" I can modify the block, setting its gas limit to the `eth_call`'s one right before execution, but that behavior is slightly incorrect.

In the meantime, I'll implement that workaround, but it'd be great to have a way to implement this properly. We are not on v5 yet, so even if we somehow fix this, I'll need the workaround for about a month.

alcuadrado:
This PR implements @jochem-brouwer's suggestion in #1014 

closes #1014

### DIFF
--- a/packages/vm/lib/runTx.ts
+++ b/packages/vm/lib/runTx.ts
@@ -27,6 +27,12 @@ export interface RunTxOpts {
    * If true, skips the balance check
    */
   skipBalance?: boolean
+
+  /**
+   * If true, skips the validation of the tx's gas limit
+   * agains the block's gas limit.
+   */
+  skipBlockGasLimitValidation?: boolean
 }
 
 /**
@@ -66,7 +72,10 @@ export default async function runTx(this: VM, opts: RunTxOpts): Promise<RunTxRes
   // create a reasonable default if no block is given
   opts.block = opts.block ?? Block.fromBlockData({}, { common: opts.tx.common })
 
-  if (opts.block.header.gasLimit.lt(opts.tx.gasLimit)) {
+  if (
+    opts.skipBlockGasLimitValidation !== true &&
+    opts.block.header.gasLimit.lt(opts.tx.gasLimit)
+  ) {
     throw new Error('tx has a higher gas limit than the block')
   }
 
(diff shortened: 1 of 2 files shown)


alcuadrado:
codecov is failing. I guess the positive case wasn't already tested.

jochem-brouwer:
Codecov randomly fails or passes (still haven't found out why) so this is not a problem 😄

holgerd77 (approved):
That's a really nice and clean test setup, looks good, will merge.

holgerd77:
(ah, short nit, just writing to you since you'll likely do reocurring PRs from time to time: please always add labels referencing the packages touched, otherwise we might forget on release notes)

Issue #1041, opened 2021-01-12
isSigned returning true for unsigned transactions


alexfmpe:
Creating a transaction with `fromTxData` **without** the signature fields leads to them being initialized 
```
v: <BN: 0>,
r: <BN: 0>,
s: <BN: 0>,
```

which makes `isSigned` return `true`
https://github.com/ethereumjs/ethereumjs-vm/blob/3588bd8dbc744eec1b58c4014cef3838d0a539aa/packages/tx/src/transaction.ts#L359
 
```
> !!0
false
> !!(new BN(0))
true
```

ryanio:
ah good catch, they should indeed not be initialized if missing, so these lines in the constructor:

```typescript
    this.v = new BN(toBuffer(v))
    this.r = new BN(toBuffer(r))
    this.s = new BN(toBuffer(s))
```

should change to:

```typescript
    this.v = v ? new BN(toBuffer(v)) : undefined
    this.r = r ? new BN(toBuffer(r)) : undefined
    this.s = s ? new BN(toBuffer(s)) : undefined
```

holgerd77:
Oh, that's an important one, I'll add a respective label here and we should release relatively soon once we have merged the respective PR from @ryanio (thanks 😄 ). Think we are generally at a point where it makes sense to target a new round of releases after the initial v5 releases.

holgerd77:
Closed by #1042

PR #1042, opened 2021-01-12
Fix isSigned returning true for unsigned txs

ryanio:
This PR fixes `tx.isSigned()` as reported in #1041. It ensures that `v,r,s` aren't initialized in the constructor if undefined.

In writing some extra test cases, I also found the same bug in `fromValuesArray` that would initialize `v,r,s` to `BN(0)` if missing, so that is also fixed.

### DIFF
--- a/packages/tx/src/transaction.ts
+++ b/packages/tx/src/transaction.ts
@@ -59,6 +59,8 @@ export default class Transaction {
 
     const [nonce, gasPrice, gasLimit, to, value, data, v, r, s] = values
 
+    const emptyBuffer = Buffer.from([])
+
     return new Transaction(
       {
         nonce: new BN(nonce),
@@ -66,10 +68,10 @@ export default class Transaction {
         gasLimit: new BN(gasLimit),
         to: to && to.length > 0 ? new Address(to) : undefined,
         value: new BN(value),
-        data: data || Buffer.from([]),
-        v: v ? new BN(v) : undefined,
-        r: r ? new BN(r) : undefined,
-        s: s ? new BN(s) : undefined,
+        data: data ?? emptyBuffer,
+        v: !v?.equals(emptyBuffer) ? new BN(v) : undefined,
+        r: !r?.equals(emptyBuffer) ? new BN(r) : undefined,
+        s: !s?.equals(emptyBuffer) ? new BN(s) : undefined,
       },
       opts
     )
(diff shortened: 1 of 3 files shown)


packages/tx/src/transaction.ts
```diff
@@ -59,17 +59,19 @@ export default class Transaction {
...
         to: to && to.length > 0 ? new Address(to) : undefined,
         value: new BN(value),
-        data: data || Buffer.from([]),
-        v: v ? new BN(v) : undefined,
-        r: r ? new BN(r) : undefined,
-        s: s ? new BN(s) : undefined,
+        data: data ?? emptyBuffer,
+        v: !v?.equals(emptyBuffer) ? new BN(v) : undefined,
+        r: !r?.equals(emptyBuffer) ? new BN(r) : undefined,
+        s: !s?.equals(emptyBuffer) ? new BN(s) : undefined,
```
holgerd77, line 74: I find these lines somewhat hard to read, replicated on REPL to trust them. No blocker though.

holgerd77 (approved):
LGTM

packages/tx/src/transaction.ts
jochem-brouwer: Is the question mark necessary? What does it do? The `!` implies that it is defined, so I don't see what this `?` should do?

ryanio: I agree that this is not the easiest to read, but I found it the best option for this particular use. I am happy to change if there is a better suggestion.

> Is the question mark necessary? What does it do? The `!` implies that it is defined, so I don't see what this `?` should do?

The `!` negates the equals expression to ensure that it doesn't match `emptyBuffer`. It would imply that it is defined if it was in the position of the `?`. The `?` is optional chaining so that if it *is* `undefined`, it will revert to the `false` clause of the ternary and output `undefined`.

If ternary expressions could accept `&&` clauses, then It could be written as: `v && !v.equals(emptyBuffer) ? new BN(v) : undefined`

jochem-brouwer: Hi, the bug is actually not fixed (I should have noticed this earlier especially after your comment) - indeed, if `v` does not exist, the `?` operator makes it `false`, but it is then negated - so it is true. There's always a BN created.
```
      let r = undefined
      console.log(!r?.equals(emptyBuffer) ? new BN(r) : undefined)
```

Prints `<BN 0>`

jochem-brouwer: CC #1048

ryanio: hmmm oh man I am just testing this again, I have to check why the new tests I added in this PR are passing okay then...

I am also wondering why in the comment above I was confused because ternary expression *do* accept `&&` clauses (must have been confused with something else), so this example code I wrote might be more reasonable to read:
 
`v && !v.equals(emptyBuffer) ? new BN(v) : undefined`

I saw you updated in your PR to:

`r !== undefined && !r?.equals(emptyBuffer) ? new BN(r) : undefined,`

but in that case the optional chaining `?` is not needed anymore (since it's already being checked for undefined) so this is a lot more understandable:

`r !== undefined && !r.equals(emptyBuffer) ? new BN(r) : undefined,`

jochem-brouwer: Ah yes good point! I still find it very hard to read though, so if there's a better alternative that'd be great :)

ryanio: I opened a PR to fix this: https://github.com/ethereumjs/ethereumjs-monorepo/pull/1077

PR #1048, opened 2021-01-17
Implement EIP2718 and EIP2930

jochem-brouwer:
This PR intends to implement [EIP2718](https://eips.ethereum.org/EIPS/eip-2718) and [EIP2930](https://eips.ethereum.org/EIPS/eip-2930) which are scheduled for YoloV3/Berlin.

I have decided to just dump the progress here and it is extremely WIP and at this point also extremely messy.

I started around trying to come up with a good design but kept running into problems. I've decided to create a first proof of concept and from that point identify which things we should change to come up with a consistent design. I've also created this PR so I can dump my thoughts on (mainly) the Tx implementation.

It is likely extremely hard or even impossible to make this PR backwards compatible, but it is worth giving it a try.

Here are some things I noticed: 

- It would make a lot of sense to reject certain behavior on a Transaction which is not signed. For instance, we can call `hash()` on a non-signed Transaction, but this makes no sense and actually returns a garbage hash (we could change this to return a raw hash which we would use to sign - that would make some sense).  I've also considered to create two classes; a SignedTransaction and an UnsignedTransaction, but I'm not really sure here as this means that for every transaction type which we introduce we always have to introduce two classes (the signed one could extend the unsigned one though..?)
- We should come up with some structure which places the data which we have to use at a single place. For instance, the `raw()` and the `hash()` method use the same copied code it seems. I tried creating an abstract `BaseTransaction` class which enforces some common logic, but the class itself would then have to implement the raw buffers to sign/hash/rlp-encode/etc.

**Please do not review this yet, it is very messy at this point.** I will use this PR to keep track of any stuff I notice along the way. This PR will likely end up to be a refactor of Tx which is hopefully backwards-compatible. If that's not the case, we can keep this PR and keep rebasing it until we decide we want to release a new Tx version.

[Handy: Go-Ethereum implementation (also WIP)](https://github.com/ethereum/go-ethereum/pull/21502)

[Extra decisions made on encoding transactions](https://github.com/ethereum/pm/issues/231#issuecomment-739998185)

The intention is that this PR pushes breaking changes to the Tx package. It also refactors the Tx package. It intends to explicitly distinguish between unsigned and signed transactions.

TODOs:

- [x] Make tx types a bit more consistent. 
- [x] Write docs, and especially update what `hash()`, `serialize()`, `raw()` of Tx are supposed to produce
- [x] Integrate EIP2930 in VM
- [x] Add transaction hash tests, for both unsigned and signed transaction


jochem-brouwer:
In general if I'd design this from scratch I would create a `TransactionFactory` which either returns a `LegacyTransaction` (see EIP2718 for the definition - `LegacyTransaction` is essentially the current Transaction class) or a `TypedTransaction`. We'd have to come up with some common interface for both of these since `TypedTransaction` can be arbitrary. They would at least require `hash`, `sign` and `verify` methods. The VM would then have to change behavior based upon what `Transaction` type we are using. This `EIP2718` seems rather easy, but it seems very complex to implement as we have to change types everywhere and change behavior based upon the Transaction type.

jochem-brouwer:
Just temporarily fixed a circular dependency problem in bf63c74. It does move two classes within one file, which I had seperated out earlier. I think it is solvable in two files. The problem was that the `Signed*` tx variants imported their `Unsigned*` variant (because `Signed* extends Unsigned*`), but `Unsigned*` also imports their `Signed*` variants, because their static constructors return the right type based on the input.

jochem-brouwer:
There's something which I find rather weird, which I stumbled upon this refactor. In [raw()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/9a3b800cec7f772bb55471237806b39d1b4a1929/packages/tx/src/transaction.ts#L331), which we use to `serialize()` a transaction, we use empty buffers for the `v`,`r`,`s` values in case the tx is not signed. Why? It seems more logical to me to not include these buffers if the tx is not signed. Or is this part of some sort of consensus thing? Worth researching.

jochem-brouwer:
Another thing: we can check if the transaction has enough "upfront cost" (checks if gas limit is at least the cost to actually execute the transaction: the data fee plus the base fee). We can check this in the constructor (but we don't do this yet). Would it not make sense to verify this immediately in the constructor and throw if the provided limit is incorrect?

packages/common/src/eips/2930.json
```diff
@@ -0,0 +1,22 @@
+{
+    "name": "EIP-2929",
+    "comment": "Optional access lists",
+    "url": "https://eips.ethereum.org/EIPS/eip-2930",
+    "status": "Draft",
+    "minimumHardfork": "berlin",
```
holgerd77, line 6: This field is rather meant as "what HF needs to be activated so that all technical preconditions are met so that the EIP can be used". Berlin generally doesn't make sense here since the HF doesn't really exist yet, so the VM would throw with this setting. I think `istanbul` would be an appropriate setting here.

packages/common/src/eips/2930.json
```diff
@@ -0,0 +1,22 @@
...
+      "accessListStorageKeyCost": {
+        "v": 1900,
+        "d": "Gas cost per storage key in an Access List transaction"
+      },
+      "accessListAddressCost": {
+        "v": 2400,
+        "d": "Gas cost per storage key in an Access List transaction"
+      }
+    },
+    "requiredEIPs": [2718, 2929],
```
holgerd77, line 18: Some nit: such a field such be defined higher in the parameter list - after `minimumHardfork` e.g., and not between this gas and config parameter dictionaries.

Apart from that: such a parameter has the problem that this doesn't include (at least in the implementation presented here) EIPs included in (encapsulated by) hardforks. So this only works if the dependent EIPs are then at some point included in the same HF.

I would prefer if we give this some more thought before we introducing such an option here (or maybe we directly find the solution along this discussion), so that this would also work if e.g. the dependent EIPs are added to a separate HF (so here: 2718, 2929).

This whole question here in the library of the relation of HFs and EIPs is generally a complex one and needs some careful thought.

holgerd77: (sorry, I know this is a bit a side track and you are interested in more general feedback 😄 , just since we are already on it)

Update: so this is highly connected to the two HF file formats with the new format now being used for the first time for [berlin](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/hardforks/berlin.json) and allowing to destructure the hardfork files and place the parameters in the respective EIP files (instead of clustering them in the HF file).

Maybe we'll target to continue this for the other HFs relatively soon, since this is generally the cleaner solution and allows for more flexibility in using this library? 🤔  For this use case here we can then also iterate through the HF definitions and see if the EIPs being required are present *either* in the `eips` parameter *or* in one of the active HFs?

With this perspective ahead I would say we can then also eventually leave the implementation here since - in this case - all EIPs affected will be integrated in the same HF. And latest with including the final version of `berlin` we can do this HF file decomposition.

Thoughts?

packages/tx/src/transactionFactory.ts
```diff
@@ -0,0 +1,69 @@
...
+   * If transactionID is undefined, return the LegacyTransaction class.
+   * Based on the `signed` boolean, return either the signed version or the unsigned version.
+   * Note that the static methods of the unsigned versions automatically return the signed version,
+   * if that data exists in the parameters.
+   * @param transactionID
+   * @param signed
+   * @param common
+   */
+  public static getTransactionClass(
+    transactionID: number = 0, // Transaction ID 0 is a special type; it is the Legacy Transaction
```
holgerd77, line 52: Reading from the EIP definitions "[Backwards Compatibility](https://eips.ethereum.org/EIPS/eip-2718#implementation)" section, transaction ID for legacy transactions is rather in a range then being a single number, so just checking for `0` should not be enough I would assume?

holgerd77:
Hi @jochem-brouwer, just had a first look, thanks for this excellent base work on such a wide ranging topic! 😀 🚀  With these preparations we should really be able to finish this in a structurally clean way providing a solid ground for working with transaction types in the years ahead, cool. 👍 

The main concern I have based on the current state of the implementation: while transaction types theoretically and *by EIP* are conceptualized as being *opaque* (so they can contain any behavior, properties,...) in practice the new `EIP-2930` transaction still shares 80%+ of its behavior with the legacy transactions. I would therefore have a strong tendency to give both these classes a shared base class, otherwise we copy far too much code. This could then also serve as some interface (abstract base class) where shared function signatures are defined and we can use this (hopefully) for typing on a lot of consumer functionality, otherwise we would have to do a lot of switches in the rest of the code base like (if txType === x use tx function a); if txType === y use tx function b)). @jochem-brouwer what are your thoughts here, were your questions on abstract classes in the chat also targeted on solving stuff around this? What do the others think?

I think since these updates will likely strongly affect our further design choices, we should give these questions some special attention, we can also discuss along our call on Wednesday. And I will also post a related question in the R&D Discord (Eth 1.x), coming a bit more from the research angle to complement.

Some more practical note, somewhat related to the topic from above: for this being able to be properly reviewed, we need to come to a commit-change-path with substantially smaller diffs, otherwise this will not really be reviewable with work being done on such a sensitive library like `tx`. If we decide to share more code this will already ease the situation. But this can also mean to split the PR into two PRs for each EIP. Let's see.

Putting this as a side note here first to not discuss everything at once: I am not yet convinced by this `Unsigned`/`Singed` transaction class split. Maybe for another discussion round. 😄 

But again: fantastic start, just becoming aware how big of a change (for Ethereum) this actually is. 😎 🤩 🥳

jochem-brouwer:
Hi @holgerd77 , I actually intend to also come up with an abstract base class for transactions. You are right that these transactions currently share a lot of code so it makes a lot of sense to integrate what these classes have in common in a base class. 

I will try to motivate this unsigned/signed stuff during our call.

packages/tx/src/transactionFactory.ts
jochem-brouwer: Ah, I think we are thinking about this in a different way - you are proposing to also include all numbers >= `0x80` here? So you can just check the first byte of the RLP-encoded transaction?

packages/tx/src/transactionFactory.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export default class TransactionFactory {
+  // It is not possible to instantiate a TransactionFactory object.
+  private constructor() {}
+
+  public static fromRawData(rawData: Buffer, transactionOptions: TxOptions): Transaction {
+    const common = transactionOptions.common ?? DEFAULT_COMMON
+    if (rawData[0] <= 0x7f) {
+      // It is an EIP-2718 Typed Transaction
+      if (!common.eips().includes(2718)) {
```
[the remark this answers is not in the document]
jochem-brouwer: Yep this makes a lot of sense!

packages/tx/src/transactionFactory.ts
holgerd77: I am not proposing anything here, I just thought from a first look that there might be the possibility that you have misread the EIP at that point, but I didn't look into your code too closely. So if you are still confident that the implementation is correct on a second look everything should be ok.

holgerd77:
For reference: https://github.com/ethereum/tests/pull/772 (Access list example test case)

jochem-brouwer:
Very weird, when I `git mv unsignedLegacyTransaction.ts LegacyTransaction.ts` git deletes the file and just dumps all the changes in a new file, so doesn't seem to be tracking any progress. It worked OK for `EIP2930Transaction` 🤔

jochem-brouwer:
Some notes to self:

Have a lot of byte types all around. Should investigate which are needed and descriptively name those. In particular, it seems that we need the hash of a transaction (this is included in a block header), the serialized transaction (this is included in a block body), the hash of the transaction (should be the hash of the serialized transaction), and the raw hash of the transaction (this is the hash of the transaction, but without the signature fields, possibly also RLP-encoded).

jochem-brouwer:
Rebased upon master.

Fixed the bug in #1042

holgerd77:
> Very weird, when I `git mv unsignedLegacyTransaction.ts LegacyTransaction.ts` git deletes the file and just dumps all the changes in a new file, so doesn't seem to be tracking any progress. It worked OK for `EIP2930Transaction` 🤔

I think you need to do this in two commits, one with moving location and the other one with just renaming, otherwise git can't follow allow (think I can remember that I had a similar case).

jochem-brouwer:
> > Very weird, when I `git mv unsignedLegacyTransaction.ts LegacyTransaction.ts` git deletes the file and just dumps all the changes in a new file, so doesn't seem to be tracking any progress. It worked OK for `EIP2930Transaction` 🤔
> 
> I think you need to do this in two commits, one with moving location and the other one with just renaming, otherwise git can't follow allow (think I can remember that I had a similar case).

I will try this, but it still does not make much sense; if this would work and then I would squash these two commits together, it should lead to the same result. I also committed changes and `git mv`'d `EIP2930Transaction`. I will retry to see if it works. Maybe it has to be a fresh commit.

jochem-brouwer:
I am having some problems coming up with a good type for this access list item (see 26b7d34). The actual type is of `(Buffer | Buffer[])[][] ` (and it would be great if we could somehow ensure that the 0-index is `Buffer` and the 1-index is `Buffer[]`). However this clogs up the package a lot, since now in `fromValuesArray` we have to explicitly cast these types.

I want to actually enable a second type on `TxData`, where it is possible to create an array, where each item is of type `{address: string, storageKeys: string[]}` (this is inspired by the Geth implementation). At runtime, in the constructor, if this is the type, then it will be converted immediately to the Buffer array (since we need this for stuff like hashes, serializing, etc.). It will also be possible to convert a raw tx (so with a Buffer) to this more readable object with some getters (like `accessListToJSON`). 

I am mainly a bit worried that this explicit `(Buffer | Buffer[])[][] ` type is nice in terms of (better) type safety, but it makes the code somewhat less readable. Thoughts? @ryanio @cgewecke @holgerd77

jochem-brouwer:
Actually, I think I got a fix for it, since I can cast the values array as a certain type, so it immediately knows what types they are. The problem in general is that if you have an Array which has multiple types, then TypeScript does not know what type it should choose for each individual item; we can however just state that 0-index is `Buffer`, and 1-index is `Buffer[]`, for example. Then the type would be `[Buffer, Buffer[]]`.

jochem-brouwer:
I experimented a bit with some getters to allow a getter to get the `accessListJSON` stuff, but I will remove this soon, if users want to get the access list, they will have to call the `toJSON` method. 

I have to convert the `toJSON` so it creates the right data, but once we do, we are (currently) compatible with what type Geth returns; 

image

(From Eth R&D)

packages/tx/src/EIP2930Transaction.ts
```diff
@@ -0,0 +1,385 @@
...
+        for (let index = 0; index < item.storageKeys.length; index++) {
+          storageItems.push(toBuffer(item.storageKeys[index]))
+        }
+        newAccessList.push([addressBuffer, storageItems])
+      }
+      usedAccessList = newAccessList
+    } else {
+      usedAccessList = accessList ?? []
+      // build the JSON
+      const json: AccessList = []
```
jochem-brouwer, line 152: This stuff will be removed from the constructor

packages/tx/src/EIP2930Transaction.ts
```diff
@@ -0,0 +1,385 @@
...
+    const accessListJSON = []
+    for (let index = 0; index < this.accessList.length; index++) {
+      const item: any = this.accessList[index]
+      const JSONItem: any = ['0x' + (<Buffer>item[0]).toString('hex')]
+      const storageSlots: Buffer[] = item[1]
+      const JSONSlots = []
+      for (let slot = 0; slot < storageSlots.length; slot++) {
+        const storageSlot = storageSlots[slot]
+        JSONSlots.push('0x' + storageSlot.toString('hex'))
+      }
```
jochem-brouwer, line 296: TODO: make this Geth-compatible (convert to `AccessList`)

packages/block/test/block.spec.ts
```diff
@@ -599,5 +599,51 @@ tape('[Block]: block functions', function (t) {
...
+      const rawData = require(`${fileHead}${i}.json`)
+      const rlp = Buffer.from(rawData.rlp, 'hex')
+      const block = Block.fromRLPSerializedBlock(rlp, { common })
+      st.ok('succesfully decoded block')
+      mock.putBlock(block)
+      try {
+        await block.validate(mock)
+        st.pass("succesfully verified block")
+      } catch (e) {
+        st.fail("block verification failed")
```
jochem-brouwer, line 644: Currently fails; transaction trie is incorrect. Seems like the factory converts some EIP2930 transactions right, but on other ones, they fail to import the `v`, `r`, `s` values, which thus makes the transaction payload the wrong value.

jochem-brouwer:
Pushed a few bugfixes and expanded tests; now EIP2930 blocks are fully validated, and the txs produce the expected hashes. The bugfixes are rather easy, but were hard to spot. A bit of a trial-and-error process 😝

holgerd77:
Hi @jochem-brouwer, just skimmed over all the code changes here once again, this is really outstanding work what you are doing here! 😄

holgerd77:
More `EIP-2930` tests: https://github.com/ethereum/tests/pull/774

jochem-brouwer:
c7e802f Bit in the dark here, since Geth's hash implementation which I built this PR on turned out to be incorrect. I checked their PR but cannot find a test case which I can actually use to verify that our hash implementation is correct (it *should* be correct, but I'd rather be sure).

holgerd77:
@jochem-brouwer think we will likely find out soon if the hashes are correct once we run the Yolo sync and then eventually take test blocks or txs from Yolo, in doubt we might be able to live without test cases for now, this should be ok since we'll stay in non-released mode for another 1-3 weeks anyhow.

Would be great if you can prioritize cleaning up the PR relatively soon so that we have some stable ground to build some PRs upon this one. 😄  Currently I don't dare yet since I fear a complete rebase and commit reordering of this base PR too much 😋 , would likely get too hairy to get an on-top PR updated then. So would be great if we have a stable base here where no further rebasing/squashing/whatever of the existing commits is necessary.

holgerd77:
(do you think this might be doable until Monday or Tuesday?)

jochem-brouwer:
Yeah this is possible, I aim to get this done by tomorrow night. I forgot I also have to do a rebase process here, and some commit cherry picking, which will likely take some time. I'll focus on implementing the VM side now.

jochem-brouwer:
I'm trying to implement it in VM, but I think our implementation of EIP2929 is wrong. Since every time we do a [sub-call](https://github.com/ethereumjs/ethereumjs-monorepo/blob/ad9def9eba0cb223dfcf913f0f1b855da041375e/packages/vm/lib/evm/opcodes/functions.ts#L1038) (see [EEI](https://github.com/ethereumjs/ethereumjs-monorepo/blob/ad9def9eba0cb223dfcf913f0f1b855da041375e/packages/vm/lib/evm/eei.ts#L495)-related code) we [create a new interpreter](https://github.com/ethereumjs/ethereumjs-monorepo/blob/ad9def9eba0cb223dfcf913f0f1b855da041375e/packages/vm/lib/evm/evm.ts#L395), which in turn creates an [empty set](https://github.com/ethereumjs/ethereumjs-monorepo/blob/ad9def9eba0cb223dfcf913f0f1b855da041375e/packages/vm/lib/evm/interpreter.ts#L94) of both accessed addresses and storage slots. EIP2929 specifies the "warm" slots/addresses are transaction-wide, not call-wide.

I think the fix is easy, we track these slots in the EEI. Will open a separate PR for this one.

cgewecke:
@jochem-brouwer Yes that sounds right to me ... I remember being a bit uncertain about that distinction.

jochem-brouwer:
I planned to have this one done today, but since the EIP2929-related work on #1124 took much longer than anticipated, I will aim to get this done tomorrow.

holgerd77:
@jochem-brouwer no stress, these EIP-2929 fixes really have been an in-between burner, just take the time you need. If it's the weekend, so be it.

jochem-brouwer:
Rebased on master. I feared a rebase hell, but there weren't a lot of conflicts surprisingly 😄  Will continue here tomorrow.

holgerd77:
You're a machine. 🤖  😀

holgerd77:
Great that the rebase worked out well! 😄

spacesailor24:
Thank you for your work on this :slightly_smiling_face: made my job easier

jochem-brouwer:
Okay, some extra stuff done today:

- Added updated tests for EIP2930 transactions. We now comply with how Geth serializes/signs/hashes these transactions
- Removed the old test data, since this data was incorrect
- Fixed some tests (seems like VM API still fails)
- Implemented EIP2930 in VM (minimal changes!)
- Added an EIP2930 test in VM

What's left to do:

- (Temporary) integrate https://github.com/ethereum/tests/pull/774 so we can also verify VM is correct
- Add block tests for EIP2930
- Some cleanup work

I don't want to merge this until we know we are correct and have to do some cleanup work, it's still in a bit of a messy state. There's also a lot of test ground we should cover.

holgerd77:
@jochem-brouwer Currently reviewing this a bit, I would do some fixes directly and then push here. Let me know if this is not ok or you're working in parallel.

packages/common/src/eips/2930.json
holgerd77: Fixed (next push).

packages/vm/lib/runTx.ts
```diff
@@ -97,6 +98,23 @@ export default async function runTx(this: VM, opts: RunTxOpts): Promise<RunTxRes
   debug('-'.repeat(100))
   debug(`tx checkpoint`)
 
+  // Is it an Access List transaction?
+  if (opts.tx.transactionType == 1 && this._common.isActivatedEIP(2929)) {
```
holgerd77, line 102: Wrong equality operator `==` (also in `runBlock`), this check is also not sufficient respectively not working atm since - currently - the `LegacyTransaction` is also returning type 1 and therefore the cast will be wrong (this is currently breaking the API tests along the `eip2929.ts` test file).

jochem-brouwer: LegacyTransaction should have type 0, weird that I messed this up.

jochem-brouwer:
Okay, have fixed the legacy transaction type.

Here are the commands to run EIP2930 tests: (I updated .gitsubmodules, but not sure if this works, does not work for me).

To manually checkout EIP2930 tests;

```
cd packages/ethereum-tests
git remote set-url origin git@github.com:qbzzt/tests.git
git pull
git checkout eip2930
```

To run tests:

`npm run test:blockchain -- --dir=GeneralStateTests/stEIP2930 --fork=Berlin`

jochem-brouwer:
Okay, will stop here, these are the errors I'm getting:

```
Error: Invalid transaction. Only expecting 6 values (for unsigned tx) or 9 values (for signed tx).
    at Function.fromValuesArray (/Volumes/Ethereum/ethereumjs-vm/packages/tx/src/legacyTransaction.ts:49:13)
    at Function.fromBlockBodyData (/Volumes/Ethereum/ethereumjs-vm/packages/tx/src/transactionFactory.ts:75:32)
    at Function.fromValuesArray (/Volumes/Ethereum/ethereumjs-vm/packages/block/src/block.ts:96:28)
    at DBManager.getBlock (/Volumes/Ethereum/ethereumjs-vm/packages/blockchain/src/db/manager.ts:171:18)
    at /Volumes/Ethereum/ethereumjs-vm/packages/blockchain/src/index.ts:1237:31
    at Blockchain.runWithLock (/Volumes/Ethereum/ethereumjs-vm/packages/blockchain/src/index.ts:407:21)
    at Blockchain.initAndLock (/Volumes/Ethereum/ethereumjs-vm/packages/blockchain/src/index.ts:394:12)
    at Blockchain._iterator (/Volumes/Ethereum/ethereumjs-vm/packages/blockchain/src/index.ts:1218:12)
    at VM.runBlockchain (/Volumes/Ethereum/ethereumjs-vm/packages/vm/lib/runBlockchain.ts:18:10)
    at runBlockchainTest (/Volumes/Ethereum/ethereumjs-vm/packages/vm/tests/BlockchainTestsRunner.ts:129:7)
    ```
    
    This is very weird, it seems some transactions are decoded as a Buffer, and others are decoded as an array. In both cases, they are EIP2930 transactions 🤔

**Comment by jochem-brouwer** [comment]
Yay! `npm run test:blockchain -- --dir=GeneralStateTests/stEIP2930 --fork=Berlin`

```
1..396
# tests 396
# pass  396

# ok
```

**Comment by jochem-brouwer** [comment]
Eh, weird, the blockchain tests pass for EIP2930, but the state tests don't...

File: packages/tx/src/baseTransaction.ts [hunk 9]
```diff
@@ -0,0 +1,193 @@
...
+      gasLimit: this.gasLimit,
+      value: this.value,
+    }
+
+    this.validateExcdeedsMaxInteger(validateCannotExceedMaxInteger)
+
+    this.common =
+      (txOptions.common &&
+        Object.assign(Object.create(Object.getPrototypeOf(txOptions.common)), txOptions.common)) ??
+      DEFAULT_COMMON
```
holgerd77, line 44: `DEFAULT_COMMON` won't work atm for the `EIP2930Transaction` without the activation of the EIPs. Not sure, do we need to separate the Common objects here and instantiate one with `eips: [2718, 2930]` or can we use the one with the EIP activations also for the `LegacyTransaction`? Or will this have side-effects? 🤔

holgerd77: Ah, likely best to directly set to `berlin` I guess.

holgerd77:
Ok, won't do anything further now. Will likely concentrate on the tests tomorrow morning (between 8-12am), just had a look at the current [coverage](https://app.codecov.io/gh/ethereumjs/ethereumjs-monorepo/compare/1048/tree/packages/tx/src).

jochem-brouwer:
Okay, state tests now work. It is likely that the test format for EIP2930 will change, and since CI currently auto-updates to the new version of the EIP2930 branch, it is likely that at some point the tests will fail due to the new format.

jochem-brouwer:
@holgerd77 I cannot view the coverage, says I am unauthorized (403 error).

packages/tx/src/baseTransaction.ts
jochem-brouwer: If we set default common to Berlin, then these EIPs are activated by default right?

Also a side note: I think we should move `DEFAULT_COMMON` to Common, so we can have package-wide "default" common objects. I also think it would be helpful if we add a `copy` method to Common.

jochem-brouwer:
Great, tests pass, only codecov refuses to report on project status?

packages/tx/src/EIP2930Transaction.ts
jochem-brouwer: This comment is probably outdated, I think it is fine in the constructor, it doesn't make much sense I think to create a private method here to handle this access list conversion.

packages/tx/src/legacyTransaction.ts
```diff
@@ -50,73 +44,65 @@ export default class Transaction {
...
+        {
+          nonce: new BN(nonce),
+          gasPrice: new BN(gasPrice),
+          gasLimit: new BN(gasLimit),
+          to: to && to.length > 0 ? new Address(to) : undefined,
+          value: new BN(value),
+          data: data ?? emptyBuffer,
+          v: v !== undefined && !v.equals(emptyBuffer) ? new BN(v) : undefined,
+          r: r !== undefined && !r.equals(emptyBuffer) ? new BN(r) : undefined,
+          s: s !== undefined && !s.equals(emptyBuffer) ? new BN(s) : undefined,
```
jochem-brouwer, line 71: TODO: check if this is compliant with the latest check to these v/r/s values, this might have gone wrong on the rebase.

packages/tx/src/EIP2930Transaction.ts
```diff
@@ -0,0 +1,385 @@
...
+  get yParity() {
+    return this.v
+  }
+
+  public static fromTxData(txData: TxData, opts: TxOptions = {}) {
+    return new EIP2930Transaction(txData, opts)
+  }
+
+  // Instantiate a transaction from the raw RLP serialized tx. This means that the RLP should start with 0x01.
+  public static fromRlpSerializedTx(serialized: Buffer, opts: TxOptions = {}) {
```
jochem-brouwer, line 74: This is actually a bit misleading. This method states that it should be an RlpSerialized transaction - but you actually have to input an invalid RLP (buffer starts with 0x01). I am not super sure here what to do, I like to have these static constructors with the same name on both this class and on LegacyTransaction. Thoughts?

packages/tx/src/transactionFactory.ts
jochem-brouwer: Currently anything >= 0x80 and <= 0xff return a LegacyTransaction.

packages/vm/tests/GeneralStateTestsRunner.ts
```diff
@@ -34,6 +34,13 @@ function parseTestCases(
       tx.gasLimit = testData.transaction.gasLimit[testIndexes['gas']]
       tx.value = testData.transaction.value[testIndexes['value']]
 
+      if (tx.accessLists) {
+        tx.accessList = testData.transaction.accessLists[testIndexes['data']]
+        if (tx.chainId == undefined) {
+          tx.chainId = 1
```
jochem-brouwer, line 40: This might look weird, but these state tests don't have chain ID in their tests, so have added manually (also raised a point about this on the tests PR)

packages/tx/src/EIP2930Transaction.ts
ryanio: hmmm that's interesting, I do see the value in keeping the same name as Legacy, however if you are technically correct maybe it could be named `fromSerializedTx` and add `fromRlpSerializedTx` as an alias for compatibility.

packages/tx/src/EIP2930Transaction.ts
ryanio: it might be nice to move this logic to a private method, just to keep the constructor light and readable

packages/tx/src/EIP2930Transaction.ts
```diff
@@ -0,0 +1,391 @@
...
+    }
+  }
+
+  public isSigned(): boolean {
+    const { yParity, r, s } = this
+    return yParity !== undefined && !!r && !!s
+  }
+
+  public hash(): Buffer {
+    // TODO add decorator
```
ryanio, line 325: if this comment is for `!this.isSigned()`, unfortunately decorators can't be dynamic

packages/tx/src/EIP2930Transaction.ts
```diff
@@ -0,0 +1,391 @@
...
+    if (!asList) {
+      return Buffer.concat([Buffer.from('01', 'hex'), rlp.encode(base)])
+    } else {
+      return base
+    }
+  }
+
+  /**
+   * Returns the encoding of the transaction. For typed transaction, this is the raw Buffer.
+   * In LegacyTransaction, this is a Buffer array.
```
ryanio, line 281: why different between Legacy and Typed?

packages/tx/src/EIP2930Transaction.ts
```diff
@@ -0,0 +1,391 @@
...
+    const base = [
+      bnToRlp(this.chainId),
+      bnToRlp(this.nonce),
+      bnToRlp(this.gasPrice),
+      bnToRlp(this.gasLimit),
+      this.to !== undefined ? this.to.buf : Buffer.from([]),
+      bnToRlp(this.value),
+      this.data,
+      this.accessList,
+    ]
```
ryanio, line 229: can this be DRY'd to `this.raw().slice(0, 7)`

packages/tx/src/baseTransaction.ts
ryanio: > I also think it would be helpful if we add a `copy` method to Common.

that's a great idea

packages/tx/src/baseTransaction.ts
```diff
@@ -0,0 +1,195 @@
...
+  /**
+   * Checks if the transaction has the minimum amount of gas required
+   * (DataFee + TxFee + Creation Fee).
+   */
+  /**
+   * Checks if the transaction has the minimum amount of gas required
+   * (DataFee + TxFee + Creation Fee).
+   */
+  validate(): boolean
+  /* eslint-disable-next-line no-dupe-class-members */
```
ryanio, line 119: hm that's weird if you were getting lint errors for `no-dupe-class-members`, the current `transaction.ts` doesn't seem to have it.

packages/tx/src/transactionFactory.ts
ryanio: lgtm, i see the logic for checking that a few lines down on line 107.

holgerd77:
@jochem-brouwer did you log into Codecov with GitHub?

packages/tx/src/baseTransaction.ts
holgerd77: Have been thinking about this a bit, this brings again up the compatibility question. So I guess we can't set to `berlin` yet, the VM uses the default blocks/txs and this will therefore get in an inconsistent state if the VM and the other libs are still on `istanbul`.

Might be a bit counterintuitive on first thought, but I guess the technically clean way to do it is to keep default on plain `istanbul` - even without the EIPs, since this is also a switch on the rule set - and then just let this throw on default when an `EIP2930Transaction` is instantiated without an explicit `common`. I think this makes it clear for everyone and should be acceptable if properly communicated (in the constructor docs).

Together with the `london` implementation the time then might be there for another major release series and we then switch directly to `london`.

WDYGT? 🤔

`DEFAULT_COMMON` to Common: that's an interesting idea, we should be careful with this though and make sure it is properly thought through and we don't introduce any redundancy - potential source for inconsistencies here - we already have the `DEFAULT_HARDFORK` there. We might then also want to do `DEFAULT_CHAIN` set to `mainnet` (likely?). Not 100% sure but supportive on first thought, likely some case for a separate issue though.

holgerd77:
For now from your friendly manual reporter 😄  :

image

holgerd77:
@jochem-brouwer started to rework the tests and separate generic base tests from the legacy tests, so that they can be run on all current and future tx types. Currently breaking due to a minor thing, will push nevertheless, think that it will be valuable to have this as a common ground, will continue later the day or tomorrow.

holgerd77:
Ah, and one small [fix](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1048/commits/87ff80798028cdc1e9ef4a7edd76d4d1203413a9) to initialize the `chainId` with the chain ID from Common if not passed in - sorry, wanted to give this a separate commit but forgot. This was to align the constructor behavior of the different Tx classes (EIP2930 would throw on passing in a plain array to `fromTxData()`, Legacy tx not).

packages/tx/src/EIP2930Transaction.ts
jochem-brouwer: In a block, a legacy transaction is encoded as a buffer array, while a typed transaction is a raw buffer. The typed transaction starts with a byte which represents the transaction type, and then we concatenate the transaction payload, which is usually the rlp encoding of the data.

packages/tx/src/baseTransaction.ts
jochem-brouwer: Yeah I was also surprised!

holgerd77:
@jochem-brouwer ah, this reviewing by writing tests is actually really nice 😄 , will continue for another hour or so.

One question I had along f68932c: I am confused on the [EIP2930Transaction.raw()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/eip2718-eip2930/packages/tx/src/EIP2930Transaction.ts#L258) method: is this RLP encoded or is this not? And which values are RLP encoded and why? Also note that this only works for the roundtrip in the tests (so: raw() -> fromValuesArray(), see commit) if the method is called with `true`, not sure if this will cause some usability problem on the user side respectively diminish tx type class interoperability.

packages/tx/src/baseTransaction.ts
jochem-brouwer: I think I agree here - I don't think it makes sense to default it to Berlin, especially since it is not yet activated on mainnet. It would be very weird to default the VM to currently non-activated new logic.

About creating a `DEFAULT_COMMON` in Common: this is not in the scope of this PR I'd say, we can open an issue for it and then discuss this. I will open an issue.

jochem-brouwer: Issue #1128 

Again, it would make sense that we default to Istanbul. It is therefore not possible to create `EIP2930Transaction` if you don't pass an explicit Common to it with the EIPs activated.

jochem-brouwer:
@holgerd77 Ah right I should log in to CodeCov.

The `raw` method returns a Buffer. We use this method in our [block package](https://github.com/ethereumjs/ethereumjs-monorepo/blob/93c1110eb4c1932f8982813c3b86244c9ac3672c/packages/block/src/block.ts#L157). We also use it in the [blockchain package](https://github.com/ethereumjs/ethereumjs-monorepo/blob/93c1110eb4c1932f8982813c3b86244c9ac3672c/packages/blockchain/src/db/helpers.ts#L46).

This raw method thus defines what value is put in the block body. It should therefore comply with EIP2718/EIP2930. This states that for typed transactions, the payload should be `TransactionType || TransactionPayload`. Here, the `TransactionPayload` is the rlp-encoding of the transaction data (like `chainId`, `value`, etc.). Since we put a 0x01 byte at the start of this RLP encoding, it invalidates the RLP encoding. This is different from the Legacy transaction, where we actually put a Buffer[] in block with the values (we can directly read them), for TypedTransactions we first need to decode the RLP-encoding and then read it.

You are right that these roundtrips should work, I'm gonna take a look if there's an easy way to fix them.

jochem-brouwer:
Ah wait the test passes because you use `true`. If we default it to return lists, then the transactions get encoded the wrong way in the block, and we thus either run into decoding problems, or we run into problems where the TransactionsTrie is incorrect.

packages/tx/src/EIP2930Transaction.ts
```diff
@@ -170,7 +170,7 @@ export class EIP2930Transaction extends BaseTransaction<EIP2930Transaction> {
       this.AccessListJSON = json
     }
 
-    this.chainId = new BN(toBuffer(chainId))
+    this.chainId = chainId ? new BN(toBuffer(chainId)) : new BN(this.common.chainId())
```
jochem-brouwer, line 173: Yep this is a great addition!

packages/tx/src/EIP2930Transaction.ts
```diff
@@ -172,7 +172,7 @@ export class EIP2930Transaction extends BaseTransaction<EIP2930Transaction> {
 
     this.chainId = chainId ? new BN(toBuffer(chainId)) : new BN(this.common.chainId())
     this.accessList = usedAccessList
-    this.v = v ? new BN(v) : undefined
+    this.v = v ? new BN(toBuffer(v)) : undefined
```
jochem-brouwer, line 175: This is fine, just as a note here; I did not use `toBuffer` this because `v` can only be 0 or 1.

jochem-brouwer:
Okay, added some tests to increase coverage and some minor improvements. Thanks for writing the other tests @holgerd77 😄  Have also fixed some points raised by @ryanio 👍

packages/tx/src/EIP2930Transaction.ts
jochem-brouwer: Just added this on both transactions!

packages/tx/src/EIP2930Transaction.ts
```diff
@@ -0,0 +1,385 @@
...
+  Buffer,
+  Buffer,
+  Buffer,
+  AccessListBuffer,
+  Buffer?,
+  Buffer?,
+  Buffer?
+]
+
+export class EIP2930Transaction extends BaseTransaction<EIP2930Transaction> {
```
jochem-brouwer, line 41: Should we rename this class to `AccessListTransaction`?

jochem-brouwer:
Yeah we have a problem here. `ecrecover` takes for `v` a number, not a BN. We cannot convert `v` to a number, since BN rejects this (number too big). I will make a hotfix here but we should fix it ASAP.

packages/tx/src/EIP2930Transaction.ts
holgerd77: Had to let this sink in a bit. I generally think that we should avoid naming things after EIPs since it is likely hard for users not having dealt with an EIP so intensely to just remember by number (I do have this very often myself). In this transaction case I wonder if we will always be able to find short descriptive (enough) names.

I think it is valuable to broaden the scope here to include [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) in the discussion. There we have 2-3 points:

- `AccessLists` are *also*  included in this transaction type (I actually wonder if we are not running into a whole lot of dependency hell with these tx types being so generic (so what tx type is depending on what other type and so on?), another discussion though)
- *This* (so: `EIP-1559`) is actually an EIP people very much *do* refer to by number

When counting these things in I currently therefore have the tendency to name by *both* the EIP and something descriptive, think this would give the best compromise of staying unambiguous and giving some hint on the functionality at the same time, or is this overblown? 🤔

Therefore some naming first-shots:

- `EIP2930AccessListTransaction`
-`EIP1559FeeMarketTransaction`

We could also shorten to:

- `EIP2930AccessListTx`
- `EIP1559FeeMarketTx`

(I think I would have some preference for the first since `Tx` is coming in so short on the second and gets lost a bit). But I have to say I am generally not completely settled yet on the whole subject.

Would think that @ryanio might have a good intuition here? Other opinions?

packages/tx/src/EIP2930Transaction.ts
holgerd77: Just as a note: this was nevertheless not done just for cosmetics, since otherwise the `fromTxData()` constructor would break due to allowing these `BNLike` values as an input. This would then break for a `string` input (as happened for my test data).

packages/tx/src/EIP2930Transaction.ts
jochem-brouwer: I like it. Can we put the descriptive name first and then the EIP? So `AccessListTransactionEIP2930`, or `AccessListEIP2930Transaction` (I think I like the latter).

holgerd77:
> Ah wait the test passes because you use `true`. If we default it to return lists, then the transactions get encoded the wrong way in the block, and we thus either run into decoding problems, or we run into problems where the TransactionsTrie is incorrect.

Yes, I "fixed" this and added `true` to the `raw()` call to make this roundtrip work.

Would it be a way to fix the whole situation by allowing both formats for `EIP2930Transaction.fromValuesArray()` and therefore make this round trip work again and also comply with the tx encoding in blocks? 🤔

packages/tx/src/EIP2930Transaction.ts
holgerd77: Yeah, that's likely even better with the descriptive name first, and - also yes - would put `Transaction` last, this is otherwise confusing.

Maybe nevertheless wait for another voice here until we change.

cgewecke: I like this too. Really good.

holgerd77:
I wonder if we want to dare to think about merging this in here? 🤔

This now had some eyes over it and will also somewhat naturally stay in continued review during the next 2-3 weeks by being worked upon and evolved (a bit like with the Clique PR). And at the same time things get a bit hairy here with this big PR and this would ease continued work.

Side note: would have been optimal if we could have been done some rebase and commit-cleanup here, but I have the impression that this gets more and more unrealistic with all the follow-up PRs. So maybe we just keep this here "as is", should nevertheless remain the general goal to have PRs in a somewhat cleaner state at some point, always challenging though.

jochem-brouwer:
I don't think we should merge yet, we can keep this branch up-to-date with master and keep things clean, just like how I now have 2 PRs on top of this one which wants to get merged in here (#1129, #1132).

You are right that the commit history is very messy, I actually intended to clean this at some point but shelved it because I wanted to continue the fun part of the PR, not the not-so-fun part (the cleaning). 

I will start to clean this up now. (With a big backup of course)

holgerd77:
@jochem-brouwer Ok, then we'll keep this open for now. 👍

holgerd77:
(I wonder though, is it actually even possible to rebase PRs with follow-up PRs without the follow-up PRs getting lost because they loose their connection points? 🤔  Also really be careful with your backups I guess - so in doubt backup the whole repository directory and not just the branch, otherwise the follow-up work might also be half-lost)

jochem-brouwer:
Okay that's a good point, will backup those branches as well. I don't think it gets lost though, because I doubt that git deletes the original commits in case you do a rebase.

packages/tx/src/EIP2930Transaction.ts
ryanio: great idea, descriptive name first seems friendly :)

jochem-brouwer:
Have opened a temporary separate branch on #1133 to see if CI passes. 

I rebased this one into 31 commits, and mainly removed the entire process here where I first split up the files in signed/unsigned variants (this history is now gone). If I force push here, then it should be ready-to-review on a per-commit basis.

holgerd77:
@jochem-brouwer great stuff, thanks for going this extra mile! 😄

jochem-brouwer:
Ignore this commit, I was on the wrong branch 🤦

jochem-brouwer:
Very scary force push 😅 

All right, rebased this one, ready for review! 🎉

packages/block/src/block.ts
```diff
@@ -91,7 +91,7 @@ export class Block {
     const transactions = []
     for (const txData of txsData || []) {
       transactions.push(
-        Transaction.fromValuesArray(txData, {
+        TransactionFactory.fromBlockBodyData(txData, {
```
holgerd77, line 94: Was thinking a bit about this breaking part in `Block`.

So if I get this right the *only* breaking thing here would be `Transaction` now being a type in the main constructor, right?

Wouldn't we already back on the compatibility track if we rename this new type `Transaction` -> `TypedTransaction` (not sure, or another name, I would think though that this name is justified even if it encompassed the old transaction format, to be discussed) and then `LegacyTransaction` back to `Transaction`?

I actually like this in general (we might still want to rename back `Transaction` -> `LegacyTransaction` on a general breaking-release-round on all libraries, not sure), since having the type named `TypedTransaction` more clearly indicates that a function now allows for a typed tx as an input.

packages/tx/src/eip2930Transaction.ts
```diff
@@ -0,0 +1,385 @@
...
+  // EIP-2930 alias for `r`
+  get senderR() {
+    return this.r
+  }
+
+  // EIP-2930 alias for `v`
+
+  get yParity() {
+    return this.v
+  }
```
holgerd77, line 68: Not sure, are these new aliases really worth to be introduced or do they not just complicate working with signatures since it confuses on generic usage if someone gets used to use e.g. `yParity` instead of `v`. I would have a tendency to just remove (or - not sure - add everything to `BaseTransaction` and at some point make `v`, `r`, `s` private further down the road and therefore have theses aliases just for all tx types).

packages/tx/src/eip2930Transaction.ts
```diff
@@ -0,0 +1,385 @@
...
+    this.r = r ? new BN(toBuffer(r)) : undefined
+    this.s = s ? new BN(toBuffer(s)) : undefined
+
+    if (!this.chainId.eq(new BN(this.common.chainId().toString()))) {
+      throw new Error('The chain ID does not match the chain ID of Common')
+    }
+
+    if (this.v && !this.v.eqn(0) && !this.v.eqn(1)) {
+      throw new Error('The y-parity of the transaction should either be 0 or 1')
+    }
```
holgerd77, line 189: If we decide to move up this check should then also be in `BaseTransaction`.

packages/tx/src/eip2930Transaction.ts
```diff
@@ -0,0 +1,385 @@
...
+      value: bnToHex(this.value),
+      data: '0x' + this.data.toString('hex'),
+      accessList: accessListJSON,
+    }
+  }
+
+  public isSigned(): boolean {
+    const { yParity, r, s } = this
+    return yParity !== undefined && !!r && !!s
+  }
```
holgerd77, line 317: Also: remove here.

packages/tx/src/eip2930Transaction.ts
```diff
@@ -0,0 +1,385 @@
...
+    if (!this.isSigned()) {
+      throw new Error('Cannot call hash method if transaction is not signed')
+    }
+
+    return keccak256(this.serialize())
+  }
+
+  public getMessageToVerifySignature(): Buffer {
+    return this.getMessageToSign()
+  }
```
holgerd77, line 329: If you want to have this name as a better working alias I think it's cleaner to:

- replace `getMessageToSign` with `getMessageToVerifySignature` in both tx classes (so also the legacy tx), add the implementations there
- Remove this call-in function here, add a reverse-one in the `BaseTransaction` (so name `getMessageToSign` and call into the `getMessageToVerifySignature()` functions
- Mark this `getMessageToSign` function in `BaseTransaction` as deprecated, add a removal note to the v6 release notes

This should work, right?

packages/tx/src/eip2930Transaction.ts
```diff
@@ -0,0 +1,385 @@
...
+
+    const { yParity, r, s } = this
+    if (yParity === undefined || !r || !s) {
+      throw new Error('Missing values to derive sender public key from signed tx')
+    }
+
+    try {
+      return ecrecover(
+        msgHash,
+        yParity.toNumber() + 27, // Recover the 27 which was stripped from ecsign
```
holgerd77, line 354: I am confused here, so is `yParity` an alias to `v` or does this have a difference of `27`?

holgerd77 (commented):
Did another round of review comments here.

packages/vm/lib/runTx.ts
```diff
@@ -97,6 +98,23 @@ export default async function runTx(this: VM, opts: RunTxOpts): Promise<RunTxRes
   debug('-'.repeat(100))
   debug(`tx checkpoint`)
 
+  // Is it an Access List transaction?
+  if (opts.tx.transactionType === 1 && this._common.isActivatedEIP(2929)) {
```
holgerd77, line 102: I think this backwards things applies to the VM as well (then `tx` in `RunTxOpts` just gets expanded by also allowing the new `2930Transaction`) and then we would just need some additional checks in places like here if the `opts.tx.transactionType` property exists before checking for the value.

packages/vm/lib/runBlock.ts
```diff
@@ -328,30 +331,45 @@ async function applyTransactions(this: VM, block: Block, opts: RunBlockOpts) {
...
+      encodedReceipt = encode(Object.values(txReceipt))
+    } else if (tx.transactionType === 1) {
       txReceipt = {
-        status: txRes.execResult.exceptionError ? 0 : 1, // Receipts have a 0 as status on error
+        status: txRes.execResult.exceptionError ? 0 : 1,
         ...abstractTxReceipt,
-      } as PostByzantiumTxReceipt
-      const statusInfo = txRes.execResult.exceptionError ? 'error' : 'ok'
-      receiptLog += ` status=${txReceipt.status} (${statusInfo}) (>= Byzantium)`
+      } as EIP2930Receipt
```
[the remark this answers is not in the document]
jochem-brouwer: I wanted to be explicit here, but we can theoretically also use the `PostByzantiumReceipt` type.

packages/vm/lib/runBlock.ts
```diff
@@ -328,30 +331,45 @@ async function applyTransactions(this: VM, block: Block, opts: RunBlockOpts) {
...
+
+      encodedReceipt = Buffer.concat([Buffer.from('01', 'hex'), encode(Object.values(txReceipt))])
```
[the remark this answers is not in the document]
jochem-brouwer: Yes this entire dynamic type thing sounds like a great addition!

packages/block/src/block.ts
jochem-brouwer: We can do this, but this still implies we have to use the `TransactionFactory` in `Block`, and we also have to use the `TypedTransactions` as the type for the `transactions` field in the block body. We can indeed export the `LegacyTransaction` as a `Transaction`, this would mean a non-breaking release for `Transaction`, but due to the type changes on `Block` they are breaking. (I am not entirely sure if they are completely breaking, since all methods of the old `Transaction` are available on both `EIP2930Transaction` and `LegacyTransaction`, so maybe it is not strictly breaking?)

packages/tx/src/eip2930Transaction.ts
jochem-brouwer: I added this to be consistent with the EIP, it is basically an utility method.

packages/tx/src/eip2930Transaction.ts
jochem-brouwer: This specific check? `LegacyTransaction` can have any `v` value, it depends on the chain id? So if we would move this exact check to `BaseTransaction`, we break `LegacyTransaction`.

packages/tx/src/eip2930Transaction.ts
jochem-brouwer: We can indeed put this one in `BaseTransaction`.

packages/tx/src/eip2930Transaction.ts
jochem-brouwer: The EIP states that `yParity` is either `0` or `1`. `ecsign` internally adds 27 to the output of `secp256k1`. Thus when we get the signature's `v` value, we either get `27` or `28`, to get the right `v` value we subtract 27. But in order to use `ecrecover` (which expects a number >=27) we have to "recover" the original output value from `ecsign`.

packages/block/src/block.ts
holgerd77: The `TransactionFactory` in `Block` is not a problem. With a new `Block` release the `Transaction` dependency is also bumped, so the new `Block` version *for sure* works together with the new `Transaction` version release. So that is no problem.

Compatibility problems only arise on direct API usage, there, the only thing we have is the constructor:

```typescript
constructor(
    header?: BlockHeader,
    transactions: Transaction[] = [],
    opts: BlockOptions = {}
  ) 
```

This is the only part which we need to make sure that it still takes in `Transaction` objects from the current library version, which I think it should when we have a type `EIP2930Transaction | Transaction` (we should test this though, this would be an application for @cgewecke using the functionality from #1118.

If there is the possibility we should very much do this. This is so so crucial that we remain backwards-compatible (especially if it's just a case about a simple rename).

packages/tx/src/eip2930Transaction.ts
holgerd77: Ok, I've read this again a bit also in [EIP-155](https://eips.ethereum.org/EIPS/eip-155). So if I get correctly, both `v` (in legacy) and `yParity` (in EIP2930) are referrring to the `v` value (so are indeed aliases), it is just the case that different `v` values for the different transaction types are used.

If this is the case (correct me if I am wrong) I would pledge to:

1. Move

```typescript
public readonly v?: BN
public readonly r?: BN
public readonly s?: BN
```

to `BaseTransaction`.

2. Drop at least the `s` and `r` aliases in `EIP2930Transaction`, so `senderS()` and `senderR()`. If people uses these this will only lead to the situation that they are making their code incompatible to work with a `LegacyTransaction`, which is unnecessary and rather annoying in doubt.

3. Eventually also drop the `yParity()` alias, not sure. Might be somewhat nice to reflect the differences, but at the end same argumentation as in 2.

4. Move as much generic signature functionality as possible into `BaseTransaction`

packages/vm/lib/runBlock.ts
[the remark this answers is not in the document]
holgerd77: Not sure if we are talking about the same things here, I was just suggesting to replace `Buffer.from('01', 'hex')` with `Buffer.from([tx.transactionType])` (or similar) here. 😄

packages/block/src/block.ts
cgewecke: I will rebase #1118 against this an open as draft.

holgerd77: Cool! 😄  Can you do the suggested renaming (see above) also along for test purposes? So this would be needed to see if this will work.

Any opinion if we now might want to merge this in here? I would be still a fan, think this would ease follow-up development. Doesn't mean that we take this out of review. We can just do follow-up fixes on future PRs.

holgerd77:
Guys, I would want to merge this in here now. I think this is stable enough to do selected further work upon, and this is otherwise blocking too much other work like the other mid-sized outstanding PR like the RPC one from @ryanio. I will also help on further discussion frame to have some blank slate again so that it gets easier again to pick selected aspects to discuss.

holgerd77 (approved):
LGTM, further work to be done in subsequent PRs.

PR #1049, opened 2021-01-18
Test PR on the renamed repository

holgerd77:
Test PR to see if the normal commit workflow still functions after renaming the repo from:

`ethereumjs-vm` -> `ethereumjs-monorepo`

Seems to work everything, but there is also a warning displayed that we should rename.


holgerd77:
Ok, everything working normally, CI is also running, will close.

Issue #1024, opened 2020-12-21
Breaking Releases - Meta Issue (VM v6, others) (old)


holgerd77:
Follow-up on #907
Work currently done towards: `develop` (we might switch to `master` at some point)

Release planning for the next major releases (VM `v6` and other major releases), please use this to add breaking changes planned as well as deprecation tasks which should be executed upon.

**Planned Release Date** (Timeframe): Latest somewhat before Merge HF (~Summer 2022), eventually earier

### Ecosystem (updated: 2022-01-13)

- [ ] `important`, `not-urgent`: 2-3 day team sprint for error messages -> enums
  - -> Can be done on a per-library basis
  - -> 2022-01-13, @holgerd77: concrete approach not clear, some non-breaking pre-work done, needs discussion
- [ ] `needs discussion + evaluation`: Switch to BigInt, #1651 
  - -> 2022-01-13, @holgerd77: Scope not yet defined, we will do some first trials like switching the VM stack implementation from storing `BN.js` instances to `BigInt` and see how it goes (and performes)

### Various Libraries (updated: 2022-01-13)

- ~Make all constructors from libraries with async constructor methods (VM, Blockchain, Block, Tx ?) protected~, needs strong discussion due to potential inheritance issues
  - -> Updated opinion (@holgerd77): inheritance issue would have been to grave on switching to `private`, `protected` will need some renewed discussion
- `needs discussion + evaluation`: Switch to `peerDependencies` library inclusion, see #1200, some experimentation done in #1402
  - -> doubts from @holgerd77, @ryanio (needs manual installation (if understood correctly), strong intervention in current installation process)
  - -> from @ryanio: npm 7 installs peerDependencies [by default](https://github.blog/2021-02-02-npm-7-is-now-generally-available/#peer-dependencies) so this might be more reasonable to accomplish now.

### VM

- [ ] `somewhat important`: Add `StateManager` to exports in main `index.js` file (see [here](https://github.com/ethereumjs/ethereumjs-vm/pull/1017#discussion_r546813522) for an example for the current suboptimal import referencing `dist`, make `VM` export non-default
- [ ] `nice-to-have`: Remove deprecated `state` constructor option
[...]

PR #1065, opened 2021-01-20
Client: Warning on not possible further block execution

holgerd77:
My own client (blockchain) DB is currently in a state where block execution is not further done due to a somewhat inconsistent db state. In the client this gets noticeable by the execution messages not further showing up, there are only the block import logs displayed.

I could identify the root cause of the no-execution behavior: in the `Blockchain` iterator function there is no [block found](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/blockchain/src/index.ts#L769) in the DB following the current iterator head. Therefore no blocks are executed. I have no idea though how and when the blockchain DB got into this state. I do know though that I've been there (in this state) a couple of times before with block execution stopping.

This PR is approaching this on a low level by not changing any logic or behavior but as a first step providing feedback when block execution is not further processing. For this it:

- Expands the `Blockchain.iterator()` interface to allow for a `number` value to be returned (so now allowing `Promise<void | number>`, keeping `void` is for backwards compatibility) and changing our own interface implementation in the blockchain class to return the number of blocks which have been actually iterated. This will likely generally be useful for a number of use cases.
- The `VM.runBlockchain()` method now similarly also returns the number of iterated blocks
- This allows in the client to give an instant warning feedback message together with the current latest head hash and number and stop further block execution

Output looks like this:

image

This should allow to better tracing/debugging this in the future and there is now a clear consumer feedback if this behavior occurs.

We might want to decide if we want to further act upon this case, a possibility would be to set the latest blockchain head to the VM iterator head, so that sync will restart from there. I hesitated to directly integrate since it is not clear what corrupted the DB in the first place. We can decide this independently from this PR though.

### DIFF
--- a/packages/blockchain/src/index.ts
+++ b/packages/blockchain/src/index.ts
@@ -41,7 +41,7 @@ export interface BlockchainInterface {
    * @param onBlock - Function called on each block with params (block: Block,
    * reorg: boolean)
    */
-  iterator(name: string, onBlock: OnBlock): Promise<void>
+  iterator(name: string, onBlock: OnBlock): Promise<void | number>
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/blockchain/src/index.ts
```diff
@@ -737,55 +737,55 @@ export default class Blockchain implements BlockchainInterface {
...
-      }
-
-      if (maxBlocks && maxBlocks < 0) {
-        throw 'If maxBlocks is provided, it has to be a non-negative number'
-      }
+  private async _iterator(name: string, onBlock: OnBlock, maxBlocks?: number): Promise<number> {
+    return await this.initAndLock<number>(
+      async (): Promise<number> => {
+        const headHash = this._heads[name] || this._genesis
+        let lastBlock: Block | undefined
```
holgerd77, line 753: I've also renamed some variable names for better clarity, not sure why this is show with such a large gap here, might be a bit confusing, but code flow remains the same here.

packages/blockchain/src/index.ts
```diff
@@ -737,55 +737,55 @@ export default class Blockchain implements BlockchainInterface {
...
-      while (maxBlocks !== blocksRanCounter) {
-        try {
-          const block = await this._getBlock(blockNumber)
+        if (!headHash) {
+          return 0
+        }
 
-          if (!block) {
-            break
-          }
```
holgerd77, line 771: This could be removed since `_getBlock()` is not allowing for an `undefined` or `null` return but is just throwing a `NotFoundError` when no block is found (which is caught later on).

Apart from this code flow is the same here as well (just switched the `if` clause order in the `catch` block for better readability, that's all)

packages/vm/lib/runBlockchain.ts
```diff
@@ -5,13 +5,13 @@ import VM from './index'
 /**
  * @ignore
  */
-export default async function runBlockchain(this: VM, blockchain?: Blockchain, maxBlocks?: number) {
```
holgerd77, line 8: I would argue that return value can be changed this way here (remaining non-breaking), since no one will have done something with the `void` return value before and just used `await vm.runBlockchain()`. So the updated return value is more of an optional extra thing now which now can be used if makes sense.

ryanio: great, sounds good, could the same reasoning be used for the return type of `iterator`? I think I would find `void | number` confusing.

holgerd77: Hmm, not for the interface I guess, since this would mean that if someone implements that interface it would break their implementation. 😕  I've added this (remove `void` return type part) to #1024 though, and the `iterator` implementation itself has no void return type.

holgerd77: When thinking about it I'm afraid that it's rather the other way around and we need to also return `Promise<void | number>` in the VM for the non-breaking release. Otherwise if someone keeps their own `Blockchain` implementation with the `void` return and uses this for the `VM.runBlockchain()` method, this will break on `TypeScript` since the method is expecting a chained return value of `number` for the output.

jochem-brouwer:
It would be nice to figure out how the DB got corrupted. It could possibly be that we do not explicitly flush the blockchain DB once we kill the client, so part of it stays in memory? We'd have to explicitly flush the DB once we get a kill signal.

packages/client/lib/sync/fullsync.ts
```diff
@@ -19,7 +19,7 @@ export class FullSynchronizer extends Synchronizer {
   public runningBlocks: boolean
 
   private stopSyncing: boolean
-  private vmPromise?: Promise<void>
+  private vmPromise?: Promise<void | number>
```
jochem-brouwer, line 22: This is a `Promise<number>` right?

packages/vm/lib/runBlockchain.ts
```diff
@@ -5,13 +5,17 @@ import VM from './index'
...
+export default async function runBlockchain(
+  this: VM,
+  blockchain?: Blockchain,
+  maxBlocks?: number
+): Promise<void | number> {
```
jochem-brouwer, line 12: This returns a `Promise<number>`?

packages/client/lib/sync/fullsync.ts
```diff
@@ -76,7 +76,16 @@ export class FullSynchronizer extends Synchronizer {
       while (!newHead.equals(oldHead) && !this.stopSyncing) {
         oldHead = newHead
         this.vmPromise = this.vm.runBlockchain(this.vm.blockchain, 1)
-        await this.vmPromise
+        const numExecuted = (await this.vmPromise) as number
```
jochem-brouwer, line 79: `as number` not necessary once the types are updated.

packages/blockchain/src/index.ts
```diff
@@ -737,55 +737,55 @@ export default class Blockchain implements BlockchainInterface {
...
-          const reorg = lastBlock ? lastBlock.hash().equals(block.header.parentHash) : false
-          lastBlock = block
-          await onBlock(block, reorg)
-          blockNumber.iaddn(1)
-          blocksRanCounter++
-        } catch (error) {
-          if (error.type !== 'NotFoundError') {
-            throw error
+        const headBlockNumber = await this.dbManager.hashToNumber(headHash)
+        const nextBlockNumber = headBlockNumber.addn(1)
```
cgewecke, line 764:
```suggestion
        const nextBlockNumber = headBlockNumber.iaddn(1)
```


...not sure this change would be meaningful but `iaddn` is used in the loop below to increment `nextBlockNumber`

cgewecke (approved):
LGTM!!

jochem-brouwer:
Doesn't this bug also indicate we have a problem in our block fetcher? If the 'bad block' gets deleted from the chain, then we should re-request the block from our peers (by block number, not by block hash). So, if a block gets deleted, send a request to our peers to retrieve the deleted block number(s).

jochem-brouwer:
What will happen if we are at the tip of the chain? Won't we then get this warning, which should not get logged, as there simply is no valid block available at this point in time?

holgerd77:
Need to merge for some subsequent work, will answer the remaining questions.

packages/client/lib/sync/fullsync.ts
holgerd77: No, we need to keep `void` along the call chain for now since this would otherwise break `Blockchain` interface compatibility, see discussion with Ryan. We can update on next breaking releases though.

holgerd77:
@jochem-brouwer I think "tip of the chain" behavior generally needs some dedicated work on its own - once we reach a tip 😋 - would very much expect this not to work on first run.

PR #1027, opened 2021-01-01
Upgrade libp2p

ryanio:
This PR upgrades all `libp2p` packages to latest. These updates include:

* deprecated `peer-info` module refactored out
  * `PeerId` and `multiaddr`used  instead
* `libp2p` callbacks converted to promises
* `bootnode` type refactored out
  * `multiaddr` used instead
* libp2p connection encryption: `secio` (deprecated) swapped for `noise`
* `pull-stream` refactored to [async iterables](https://github.com/libp2p/js-libp2p/blob/master/doc/STREAMING_ITERABLES.md)
* various type improvements
* misc. fixes:
  * `package.json`: fix `ethereumjs` bin path
  * `cli.ts`: use `args.datadir` for `datadir`
  *  `browser/index.ts`: fix init logic

PR #1067, opened 2021-01-21
Client: Save node key

jochem-brouwer:
This PR ensures that we save the node key to disk. This is saved in the root directory of the current network. This way, we ensure that we will re-use the enode, such that we can quickly reconnect, or add our client as a bootnode. This would not be possible if it is not saved, since then our enode keeps changing every time we reboot the client.

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -175,11 +175,14 @@ async function run() {
   }
 
   const common = new Common({ chain, hardfork: 'chainstart' })
+  const datadir = args.datadir ?? Config.DATADIR_DEFAULT
+  const key = await Config.getClientKey(datadir, common)
   const config = new Config({
     common,
     syncmode: args.syncmode,
     lightserv: args.lightserv,
-    datadir: args.datadir,
+    datadir,
+    key,
     transports: args.transports,
     bootnodes: args.bootnodes ? parseMultiaddrs(args.bootnodes) : undefined,
     multiaddrs: args.multiaddrs ? parseMultiaddrs(args.multiaddrs) : undefined,
--- a/packages/client/lib/config.ts
+++ b/packages/client/lib/config.ts
@@ -294,4 +294,11 @@ export class Config {
     if (option !== undefined) return option
     return this.chainCommon.chainName() === 'mainnet'
   }
+  
+  getNetworkDir(): string {
+    const networkDirName = this.common.chainName()
+    const dataDir = `${this.datadir}/${networkDirName}`
+
+    return dataDir
+  }
 }
--- a/packages/client/lib/config.ts
+++ b/packages/client/lib/config.ts
@@ -294,7 +294,7 @@ export class Config {
     if (option !== undefined) return option
     return this.chainCommon.chainName() === 'mainnet'
   }
-  
+
   getNetworkDir(): string {
     const networkDirName = this.common.chainName()
     const dataDir = `${this.datadir}/${networkDirName}`
--- a/packages/client/lib/config.ts
+++ b/packages/client/lib/config.ts
@@ -1,9 +1,12 @@
 import Common from '@ethereumjs/common'
 import VM from '@ethereumjs/vm'
+import { genPrivateKey } from '@ethereumjs/devp2p'
 import Multiaddr from 'multiaddr'
 import { getLogger, Logger } from './logging'
 import { Libp2pServer, RlpxServer } from './net/server'
 import { parseTransports } from './util'
+import type { LevelUp } from 'levelup'
+const level = require('level')
 
 export interface ConfigOptions {
   /**
@@ -39,6 +42,13 @@ export interface ConfigOptions {
    * Root data directory for the blockchain
    */
   datadir?: string
+
+  /**
+   * Private key for the client.
+   * Use return value of `await Config.getClientKey(datadir, common)`
+   * If left blank, a random key will be generated and used.
+   */
+  key?: Buffer
 
   /**
    * Network transports ('rlpx' and/or 'libp2p')
@@ -172,6 +182,7 @@ export class Config {
   public readonly vm?: VM
   public readonly lightserv: boolean
   public readonly datadir: string
+  public readonly key: Buffer
   public readonly transports: string[]
   public readonly bootnodes?: Multiaddr[]
   public readonly multiaddrs?: Multiaddr[]
@@ -199,6 +210,7 @@ export class Config {
     this.bootnodes = options.bootnodes
     this.multiaddrs = options.multiaddrs
     this.datadir = options.datadir ?? Config.DATADIR_DEFAULT
+    this.key = options.key ?? genPrivateKey()
     this.rpc = options.rpc ?? Config.RPC_DEFAULT
     this.rpcport = options.rpcport ?? Config.RPCPORT_DEFAULT
     this.rpcaddr = options.rpcaddr ?? Config.RPCADDR_DEFAULT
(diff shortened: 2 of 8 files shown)


ryanio:
> It seems that this whole bootnodes option is already present - if you use the transports option you can add this as a comma separated list. I don't think I like that.

right, I don't think we should add a `bootnodes` option to the CLI, as `transports` can be used to define additional bootnodes, or otherwise the code added is redundant. Like so: `--transports rlpx:bootnodes=bootnode1,bootnode2,bootnode3`. This is already in the client readme, but under the libp2p key, so I could see why it could be confusing, we can add this additional documentation to show how to add bootnodes to rlpx.

You can see in #1027 I actually parse them as `multiaddrs` to pass around the client, which is a powerful type to use. The only thing is I haven't added yet is the enode ID (can be added as an extra key to multiaddr so e.g. `/ip4/${ip}/tcp/${port}/enode/${id}`) but that isn't being passed to devp2p at this point anyway. (it should, so it only connects to the peer if it has the same id, otherwise could be a compromised node). In my newly added `parseMultiaddrs` function it can parse the `id@ip:port` format so we don't need that additional code added in `config.ts` for parsing.

jochem-brouwer:
I think you are right. I will check if this is possible, since it might be that this string parser might choke upon the address string format.

jochem-brouwer:
Just checked:

`npm run client:start -- --transports rlpx:bootnodes=d860a01f9722d78051619d1e2351aba3f43f943f6f00718d1b9baa4101932a1f5011f16bb2b1bb35db20d6fe28fa0bf09636d26a87d31de9ec6203eeedb1f666@18.138.108.67,22a8232c3abc76a16ae9d6c3b164f98775fe226f0917b0ca871128a74a8e9630b458460865bab457221f1d448dd9791d24c4e5d88786180ac185df813a68d4de@3.209.45.79:30303`

(First two bootnodes from mainnet Common). This parses the bootnodes wrong:

```
{
  name: 'rlpx',
  options: {
    bootnodes: 'd860a01f9722d78051619d1e2351aba3f43f943f6f00718d1b9baa4101932a1f5011f16bb2b1bb35db20d6fe28fa0bf09636d26a87d31de9ec6203eeedb1f666@18.138.108.67',
    '22a8232c3abc76a16ae9d6c3b164f98775fe226f0917b0ca871128a74a8e9630b458460865bab457221f1d448dd9791d24c4e5d88786180ac185df813a68d4de@3.209.45.79:30303': undefined
  }
}
```

jochem-brouwer:
I can't come up with a consistent format scheme here, will open an issue.

packages/client/lib/net/server/rlpxserver.ts
```diff
@@ -51,6 +52,21 @@ export class RlpxServer extends Server {
   constructor(options: RlpxServerOptions) {
     super(options)
 
+    if (this.key === undefined) {
```
ryanio, line 55: nice, looks good, can you put this logic in a helper func in `config.ts` named something like `getClientKey(): Buffer` that does the dual get/save logic? Then I can reuse in `libp2pserver.ts` in my pr.

And then here it can just be `this.key = getClientKey()`

jochem-brouwer: Yep - will do! Do you have any suggestion to fix the browser tests not passing? It doesn't have access to this `fs-extra` stuff, how do we fix this problem in general? We have a lot of `fs`-related operations right?

ryanio: hmmm yeah browser won't have access to fs-related modules.

In the client readme it says:

> That's it! Now, you should start seeing headers being downloaded to the local storage of your browser. Since IndexDB is being used, even if you close and re-open the browser window, the headers you've downloaded will be saved.

Maybe we can do something similarly like if we are in the browser we can save the key to localStorage and get it from there as well.

I wonder if there is a package that can abstract away this `fs || localStorage` logic with a higher level api. (briefly googling around, found https://github.com/jvilk/BrowserFS although haven't looked further if it would be a good choice for our usage here)

holgerd77: Can we maybe just use leveldb for this as well? Is another separate DB instance for configuration too heavy?

holgerd77:
@jochem-brouwer can you give this another look at some PIT?

jochem-brouwer:
Will make this one merge-ready ASAP @holgerd77 😄

holgerd77:
@jochem-brouwer can you do an update here?

holgerd77:
(if time permits, not super urgent, but would be nice to get this finished)

jochem-brouwer:
@holgerd77 Yep, sorry forgot about this one. Will probably do today, but by end of this weekend the latest.

holgerd77:
No problem. 😄

jochem-brouwer:
Just rebased this one. If we want to introduce this node key into a new database (which we currently don't have) I am facing the problem where I should put the database. The logical place for this database would be to be in Config. Would it make sense to add a global config DB here? This would depend on the data dir, so we don't use the same config for e.g. mainnet and rinkeby.

packages/client/lib/config.ts
```diff
@@ -282,4 +282,11 @@ export class Config {
     if (option !== undefined) return option
     return this.chainCommon.chainName() === 'mainnet'
   }
+
+  getNetworkDir(): string {
+    const networkDirName = this.common.chainName()
+    const dataDir = `${this.datadir}/${networkDirName}`
+
+    return dataDir
```
holgerd77, line 290: Please use the same naming scheme here as in the other directory methods - so `getNetworkDirectory()` and sort this in above along with the other methods, a bit misplaced here.

holgerd77: Can you also use this method for directory creation in both of the other two methods?

holgerd77:
@jochem-brouwer yeah, I think it makes sense to add a global config DB and the `Config` class should be good place I guess. 😄 

> This would depend on the data dir, so we don't use the same config for e.g. mainnet and rinkeby.

I am not sure here, we'll likely have both global configuration parameters as well as chain-specific ones. Would it be a way to simplify and use one DB for both, and then prefix keys, so e.g. `general:config_key1` (or `global` or `config`), `rinkeby:config_key2`, or similar?

jochem-brouwer:
If we want to have global config and also chain-specific config, would it then make sense to introduce a global database and also a chain-specific one?

holgerd77:
For me having 5 databases for 20 config parameters sound a bit overblown, but I'm also really no expert here. Would be happy on further voices.

ryanio:
I rebased this and added a client config leveldb in the network directory for getting and setting the client key.

packages/client/bin/cli.ts
```diff
@@ -175,11 +175,14 @@ async function run() {
   }
 
   const common = new Common({ chain, hardfork: 'chainstart' })
+  const datadir = args.datadir ?? Config.DATADIR_DEFAULT
```
holgerd77, line 178: Won't make this a blocker, but this shouldn't be necessary since `datadir` has a local CLI-specific default (first was afraid that this would actually *change*  the default, but this is not the case)? 🤔

holgerd77 (approved):
Nice, LGTM! 😄  Will directly use this to more comfortably debug #1147, hehe 😛  (really annoying to always copy over a new enode ID on every new test run)

holgerd77:
Ah, and also tested this locally and can confirm that this works, client key / network URL is perceived.

packages/client/bin/cli.ts
```diff
@@ -175,11 +175,14 @@ async function run() {
...
+  const key = await Config.getClientKey(datadir, common)
   const config = new Config({
     common,
     syncmode: args.syncmode,
     lightserv: args.lightserv,
-    datadir: args.datadir,
+    datadir,
+    key,
```
holgerd77, line 185: I wonder though if we want to make the name of this option a bit more expressive?

holgerd77:
@ryanio on a closer look I think this is logically wrong, the client key from the DB shouldn't be re-passed in to `Config`, this makes not much sense. I stumbled upon this since this is currently breaking when using a custom `datadir` with "IO error: datadir-dev2/mainnet/config/LOCK: No such file or directory".

When trying to fix this (with `ensureDirSync` in cli.ts) this get's unnecessarily complex since one is still acting in a static context where `Config` is not existing yet and one would therefore need to manually construct the directory paths.

Will rework this a bit along #1147.

holgerd77:
@ryanio Update: ok, I am slowly getting why you are doing this, seems this has again this `leveldb` creation problem (or at least I am stumbling atm upon that I don't want to introduce the `fs-extra` dependency to config). So I will just give this a hot-fix right now on the mentioned PR.

I would still assume that this is structurally not correct and we should give this some additional thought (it just shouldn't be necessary to have access to the configuration database to actually *create*  a `Config` object). Maybe we can rework on the addition of the next config option or something (or alternatively directly doing now).

ryanio:
@holgerd77 ah thanks, yeah I had some challenges with the async nature of saving or getting the client key from the db, so I had to move the async part out of the config/server instantiation.

Issue #1066, opened 2021-01-21
Client: "Error: invalid transaction trie" leading to corrupted state DB


holgerd77:
Here is another one, this might be the last (occurring during the first 200.000 or so mainnet blocks, not related to some specific blocks though):

```shell
INFO [01-21|18:26:09] Imported blocks count=700 number=62001 hash=3c1e0ffd... peers=8
INFO [01-21|18:26:09] Executed blocks count=50 first=57643 hash=3c9230be... last=57692 hash=308a3a08... with txs=8
ERROR [01-21|18:26:11] Error: invalid transaction trie
    at Block.validateData (/EthereumJS/ethereumjs-vm/packages/block/dist/block.js:179:19)
    at async Block.validate (/EthereumJS/ethereumjs-vm/packages/block/dist/block.js:162:9)
    at async VM.applyBlock (/EthereumJS/ethereumjs-vm/packages/vm/dist/runBlock.js:132:13)
    at async VM.runBlock (/EthereumJS/ethereumjs-vm/packages/vm/dist/runBlock.js:70:18)
    at async /EthereumJS/ethereumjs-vm/packages/vm/dist/runBlockchain.js:27:13
    at async /EthereumJS/ethereumjs-vm/packages/blockchain/dist/index.js:597:21
    at async Blockchain.runWithLock (/EthereumJS/ethereumjs-vm/packages/blockchain/dist/index.js:188:27)
    at async Blockchain.initAndLock (/EthereumJS/ethereumjs-vm/packages/blockchain/dist/index.js:176:16)
    at async Blockchain._iterator (/EthereumJS/ethereumjs-vm/packages/blockchain/dist/index.js:576:9)
    at async VM.runBlockchain (/EthereumJS/ethereumjs-vm/packages/vm/dist/runBlockchain.js:10:5)
INFO [01-21|18:26:11] Imported blocks count=100 number=62701 hash=9f1d35d9... peers=9
INFO [01-21|18:26:11] Imported blocks count=50 number=62801 hash=f144c086... peers=9
```

This one occurs during some - I would say - random PiT during block execution. Execution then stops and also doesn't resume on restarting the client, after merging of #1065 we will have some better output here. This is not related to a specific block, I re-ran the execution with a state DB backup and on the second run it passed the blocks where it broke the first time. So this is likely rather some race condition.

holgerd77:
[...]

PR #1068, opened 2021-01-22
Client: refactor execution

holgerd77:
Ready for review, see description below.


holgerd77:
Ok, ready. 😄 

Some summary on this PR: this work here was initially triggered when I wanted to add some additional HF logic to the execution logic in `FullSynchronizer` and realized that this is not sustainable since the synchronizer class got more and more bloated and mixed up with the execution logic.

I have therefore in a first step along 3351298162b44e9a9789710906aaa0206f43a2cc, 763306155613b11ce8d50f723a5f0ad0d62fd44d and (later) c437c282dbb038f304895c22931c173a738ea473 extracted the execution code parts from `FullSynchronizer` and created an own execution module, which has it's own base class `Execution` and a subclass `VMExecution`, following the structural setup of the other client modules. This should help us to write cleaner execution logic code and better separate the functionality. It also - maybe this is the main point - decouples the execution from the sync logic, so this should prepare both for adding additional forms of execution (e.g. some form of `StatelessExecution`) as well as using the same execution modules for different types of sync (e.g. beam sync).

Note that the current structure of the execution classes is rather meant to be some start, so feel free to evolve on this along future work.

On a second step I have then integrated the execution code from `VM.runBlockchain()` directly in the execution module here 413a3925ca62873fc87048ef0eb7884e4e9eb1e7 following the suggestion from @jochem-brouwer in this discussion https://github.com/ethereumjs/ethereumjs-monorepo/issues/1066#issuecomment-764861502. This gives us more control over the code since we now have both the blockchain `iterator` run and the enclosing `try/catch` block in the client which we will need to solve issues like #1066 (not yet integrated in this PR) as well as generally handle execution failures appropriately. This integration was a bit more tricky (a least for me) than I thought, my initial goal here has been to make this functionally equivalent to the version before. Also feel free here to evolve on this code part.

With these changes from above the current `testdouble` based testing of VM execution with mocking several functions came finally to its limits. I've therefore added a new static constructor factory `Blockchain.fromBlocksData()` (would `fromBlockData()` be a better name?) in ed3f007cd76cbeee1a737308f736e8d7899a9e70 which allows for one-line creation of `Blockchain` objects and which I have used in ed3f007cd76cbeee1a737308f736e8d7899a9e70 and a3c32cc2ff0a4caf533348b3d109a97d4fcf4f64 to switch to a real-data based testing for VM execution in the client. Note that test coverage can still be improved, rather a start here.

On a4683e6066731ae99b688c65c2b3342af3e30bd8 and 87bc66219fc2a9c90825c9730424c7c84c14e76b I've finally added HF switch logic for both the chain download and the execution. There are now two common instances `chainCommon` and `execCommon` created based upon the `common` instance provided (this is still needed to allow e.g. to pass in a custom chain, see one of the tests for an example). These two common instances are then updated by the synchronizer (`chainCommon`) and the execution class (`execCommon`) to independently track the respective active hardforks to correctly submit block download requests (the `forkhash` changes on HF switches) as well as VM executions. All this is rather some preparation and will likely need some real world testing and some refinement. We can do this (better) once #1032 is merged with the early-on occuring HF switches on Rinkeby and Goerli.

Ok. Open for review. 😄

holgerd77:
Phew, the fight along a5421b862ea92e8197e6c121510497375bc4b3e3 can't be seen in this commit. 😜

(actually a whole morning)

packages/client/lib/sync/fullsync.ts
```diff
@@ -182,6 +102,14 @@ export class FullSynchronizer extends Synchronizer {
     const count = height.sub(first).addn(1)
     if (count.lten(0)) return false
 
+    const nextForkBlock = this.config.chainCommon.nextHardforkBlock()
+    if (nextForkBlock) {
+      if (first.gten(nextForkBlock)) {
+        this.config.chainCommon.setHardforkByBlockNumber(first.toNumber())
+        this.hardfork = this.config.chainCommon.hardfork()
+      }
+    }
```
cgewecke, line 111: 👍

cgewecke (approved):
This all looks excellent to me! Great. 💯 

Just one question about how common works in FullSync's `nextBlockFork` section...

I think the `fromBlocksData` name makes sense / is correct.

packages/client/lib/sync/fullsync.ts
holgerd77: No, not here since the HF switches on block download and block execution occur independently. A HF change for execution is handled along the [VMExecution.run()](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1068/files#diff-3ff0e2525d3eb1dd38e375f27736fe31ad55e351e512355106ac144739aa6b78R102) method.

holgerd77:
Rebased this on top of the clique PR #1032 - needed to add a small typing fix. I will also merge this one in (with admin merge), feel the very strong need to streamline things a bit due to the current extra constraints and PR has been reviewed by @cgewecke, let me nevertheless know if you think this is inappropriately deviating from the protocol.

Then we have both of the large PRs out of the way and can concentrate on doing "normally" scoped work again 😄  and @jochem-brouwer (and eventually others) has (have) a solid and stable base for further work on #1048.

Let me nevertheless know though if you have final comments on this PR, happily doing follow-up PRs on this doing latest fixes or changes.

PR #1074, opened 2021-01-29
Clique small fixes / follow-up work

ryanio:
This PR addresses some further work from review in #1032:

* New block option to provide a privateKey to seal a block: `cliqueSigner?: Buffer`
* Exactly compare checkpointSigners list with activeSigners list
* Keep signer states for `CLIQUE_SIGNER_HISTORY_BLOCK_LIMIT` as number of blocks, not number of last states
* Don't reset votes on epoch blocks to keep in case of reorgs, but only count non-stale votes in `cliqueUpdateVotes()` by checking if the vote's blockNumber >= last epoch block.
* Explicitly check for availability of `blockchain.cliqueActiveSigners()` in `BlockHeader.validateCliqueDifficulty()` to not add to block's minimum blockchain interface ([discussion](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1032#discussion_r565997187))
* Simplify `Address.equals(address)` checks with ethereumjs-util [v7.0.8](https://github.com/ethereumjs/ethereumjs-util/releases/tag/v7.0.8)


holgerd77:
Can this PR get an update? Is this generally ready for review?

ryanio:
@holgerd77 yes now ready for review, thanks!

packages/block/src/header.ts
```diff
@@ -238,6 +240,11 @@ export class BlockHeader {
       this.difficulty = this.canonicalDifficulty(options.calcDifficultyFromHeader)
     }
 
+    // If cliqueSigner is provided, seal block with provided privateKey.
+    if (options.cliqueSigner) {
+      this.extraData = this.cliqueSealBlock(options.cliqueSigner)
+    }
```
holgerd77, line 246: Nice, also like that the private key is not residing in the library, was a bit worried that we introduce some new potential security risks here.

packages/block/src/header.ts
```diff
@@ -673,9 +671,9 @@ export class BlockHeader {
    */
   cliqueVerifySignature(signerList: Address[]): boolean {
     this._requireClique('cliqueVerifySignature')
-    const signerAddress = this.cliqueSigner().toBuffer()
+    const signerAddress = this.cliqueSigner()
     const signerFound = signerList.find((signer) => {
-      return signer.toBuffer().equals(signerAddress)
+      return signer.equals(signerAddress)
```
holgerd77, line 676: Ah, yes, that's really a much nicer API! 😄

holgerd77 (approved):
Looks all good, thanks Ryan! 👍

PR #1077, opened 2021-02-01
Tx: fix assignment of v,r,s in fromValuesArray

ryanio:
This PR is a follow-up fix to #1042, thanks to @jochem-brouwer for [noticing](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1042#discussion_r566187908) this, where the logic in checking for values of `v,r,s` in `fromValuesArray` was flawed.

This bug would have been caught by `_validateTxV` a long time ago (prior to #1041) if the function did not return if `v` was zero. The proper behavior in this package is for those values to be undefined if unavailable, zero is not a valid value as it's not being checked anywhere else for zero, only undefined.


packages/tx/src/transaction.ts
```diff
@@ -69,9 +69,9 @@ export default class Transaction {
         to: to && to.length > 0 ? new Address(to) : undefined,
         value: new BN(value),
         data: data ?? emptyBuffer,
-        v: !v?.equals(emptyBuffer) ? new BN(v) : undefined,
-        r: !r?.equals(emptyBuffer) ? new BN(r) : undefined,
-        s: !s?.equals(emptyBuffer) ? new BN(s) : undefined,
+        v: v && !v.equals(emptyBuffer) ? new BN(v) : undefined,
```
jochem-brouwer, line 72: This won't work if `v` is 0. This is problematic in EIP-2930.

holgerd77 (dismissed):
LGTM, argh, this does need another release I guess? 😋

packages/tx/src/transaction.ts
jochem-brouwer, line 72:
```suggestion
        v: (v !== undefined) && !v.equals(emptyBuffer) ? new BN(v) : undefined,
```

holgerd77:
Ah, just seeing that @jochem-brouwer is also reviewing right now with some remarks, will then leave this to him to give an approval since he has generally some better insight here

packages/tx/src/transaction.ts
jochem-brouwer: Although I am not sure if there exists legacy transactions with a `v` value of 0... They do exist on EIP-2930 transactions, so it can also be "fixed" there, but I am generally more in favor to be correct here; if one inputs any value (also for `r` and `s`) which is 0, then it should create a `BN(0)`.

ryanio: ah hm okay, then maybe I shouldn't remove the validation for `v` to be zero in this commit? https://github.com/ethereumjs/ethereumjs-monorepo/pull/1077/commits/14581b0296745e31b75bcadbe7b9232d17d80ac6

I think I'd prefer that tx v3 continues to support legacy transactions only, and then v4 can support EIP-2930. These accidental `BN(0)` bugs are not fun and I don't think v,r,s should ever be initialized with `BN(0)`, only `undefined` for unsigned transactions, otherwise we need to add more checks around the library (like in `isSigned`). What do you think?

holgerd77 (approved):
LGTM

I read this correctly from the comments that the bug initially addressed was not fixed at all, so this needs as urgent a release as the initial PR, right?

packages/tx/src/transaction.ts
holgerd77: > Although I am not sure if there exists legacy transactions with a v value of 0... They do exist on EIP-2930

Are you maybe mixing this up with this v - 27 thing (never understand that completely), I can't imagine that EIP-2930 is changing on the signature format?

Maybe @alcuadrado can have a short look here as well

ryanio:
@holgerd77 yes, another bug fix release for this PR would be great, sorry for the troubles and thanks :) the first PR fixed one half of it (static constructor `fromTxData`) but `fromValuesArray` was still flawed and should now be fixed.

Issue #1076, opened 2021-01-29
Client -> Mainnet: Consensus Error on block 226522 "invalid receipt trie"


holgerd77:
Yeah, we have our first consensus error - a real repeatable one and not some of this race condition stuff. 😅  I find this somewhat of a relief to now come to the "real stuff" after all this infrastructure code writing since this was one of the main reasons the client was build for. 🥳

Since this one is relatively early - on block 226522 - I would expect more to come further down the line, but we'll see.

Just pasting the stack trace here, haven't dug deeper into what's wrong:

```shell
ERROR [01-29|23:10:05] Error: invalid receiptTrie
    at VM.runBlock (/EthereumJS/ethereumjs-vm/packages/vm/dist/runBlock.js:88:19)
    at async /EthereumJS/ethereumjs-vm/packages/client/dist/lib/sync/execution/vmexecution.js:101:21
    at async /EthereumJS/ethereumjs-vm/packages/blockchain/dist/index.js:909:21
    at async Blockchain.runWithLock (/EthereumJS/ethereumjs-vm/packages/blockchain/dist/index.js:262:27)
    at async Blockchain.initAndLock (/EthereumJS/ethereumjs-vm/packages/blockchain/dist/index.js:250:16)
    at async Blockchain._iterator (/EthereumJS/ethereumjs-vm/packages/blockchain/dist/index.js:891:16)
    at async VMExecution.run (/EthereumJS/ethereumjs-vm/packages/client/dist/lib/sync/execution/vmexecution.js:147:28)
    at async Chain.<anonymous> (/EthereumJS/ethereumjs-vm/packages/client/dist/lib/sync/fullsync.js:32:17)
```

Curious to see how the debugging of such an error goes (we'll likely need stack traces for this from another client?).

jochem-brouwer:
This is exciting! 😄 

Optimally, we would get stack traces. However, (in this case), we can quickly figure out which transaction goes wrong: to do that, we download each receipt of the block (we can do that over `web3`), and compare those to our outputs. These are very helpful, as they contain, for instance, `gasUsed`. So we can figure out which transaction(s) is/are wrong, and in most cases `gasUsed` is the problem.

jochem-brouwer:
Well we don't have to do that in this case, [since the block only has one transaction](https://etherscan.io/tx/0x8420dbe268209c4e921bf6b78bc9fb959d9d9553fd16aacb9dd8cad955dbe729) 😅

[...]

PR #1080, opened 2021-02-02
VM Debugging

holgerd77:
I am so tired of VM debugging where it becomes extremely hard to follow on the control flow that I would want to add some debugging analog to what is done in the `devp2p` and `client` packages. This should make it significantly easier for us to track bugs like in #1076.

This is a first test commit to gather some feedback, I would continue with this tomorrow if there is no significant objection or critics.

This produces output like the following if activated with e.g. `DEBUG=vm:tx ts-node test.ts`:

image

(so only the upper part, JSON below is just a `console.log()` from the script run)


cgewecke:
100% in favor 🙂

holgerd77:
Have added some first debug loggers `vm:tx`, `vm:tx:gas` and `vm:state` which can be run all at once with `DEBUG=vm:*,vm:tx:gas ts-node test.ts` and I can assure you this will be game-changing for both debugging and educational purposes! 😀  😁  😀

Here is a sample run:

Bildschirmfoto 2021-02-03 um 11 33 51

holgerd77:
Ok, have now also adding a debug logger for opcode based logging, this comes in really really handy, e.g. by just logging on all the `SSTORE` processings with `DEBUG=vm:tx,vm:evm,vm:ops:sstore,vm:*:gas ts-node test.ts` //cc @jochem-brouwer: 😄 

image

holgerd77:
Ok, I'll stop here, have also added some additional README instructions, this is now ready for review. 😄 

@jochem-brouwer I'll try to add an additional `vm:block` logger during next week if you think this is a useful addition which should be prioritized since it will get useful for our client runs.

Please note though that this is meant to be a general purpose debugging helper not necessarily (only) meant to be for our client runs but for all sorts of development and user activity (adding new opcodes, testing some functionality changes in the `evm.ts` code, debugging and/or understanding tx runs,...). Should be very usable for our client testing as well though, as loggers can be arbitrarily combined as necessary. 😁  I also very much assume that we will add additional loggers over time.

jochem-brouwer:
I have a few suggestions after just reading the readme: (can be in a subsequent PR)

- It is very helpful if we know what the current address is we are running operations on. Just the depth is not enough. If we SSTORE, we want to know on which address we are storing. (Same with CREATE)
- It is also very helpful to know if a `CALL*` / `CREATE*` succeeds or not. So, if the depth changes and gets lower (so a call returns), then we should check the top of the stack. If this is nonzero, the call/create has succeeded. If this is not the case, then the call/create has failed. (This thus also means that any storage writes / code deposits did not succeed and are reverted). This is somewhat tricky, since if you call an address which has no code, we don't change the depth (it immediately succeeds, and is usually just used to transfer ether).

In general what would also help is to verify our `step` event against the data which is reported if we call `debug_traceTransaction` on geth. Then we can figure out exactly at which step our results diverge from that of Geth.

holgerd77:
Hi Jochem, thanks for the suggestions, just an answer on the first one on a first round:

If I get this right the operations address on a `CALL` is the `to` address and on `CREATE` the created address, right? Both are available in the `SSTORE`-(or other bytecode)-surrounding log messages. Correct me if I am wrong?

So e.g.

```shell
vm:evm New message caller=0x147013436bd5c7def49a8e27c7fba8ac2b9dfe1f gasLimit=276872 to=0x9ca228250f9d8f86c23690074c2b96d5f5479f79 value=0 delegatecall=no +0ms
...and then closing with
vm:evm message checkpoint +96ms
```

And for CREATE

```shell
vm:evm New message caller=0x9ca228250f9d8f86c23690074c2b96d5f5479f79 gasLimit=243805 to= value=0 delegatecall=no +0ms
  vm:evm Message CREATE execution (to undefined) +0ms
  vm:evm Reduce sender (0x9ca228250f9d8f86c23690074c2b96d5f5479f79) balance (-> 40000000000000000) +1ms
  vm:evm Generated CREATE contract address 0x8398ff6c618e9515468c1c4b198d53666cbe8462 +1ms
... and closing again with
vm:evm message checkpoint committed +0ms#
```

holgerd77:
Isn't point 2 (CALL / CREATE succeeds or not) included in the message results? 🤔

So e.g.:

```shell
vm:evm Received message results gasUsed=106040 execResult: [ gasUsed=106040 exceptionError= returnValue=0000000000000000000000008398ff6c618e9515468c1c4b19... gasRefund=undefined ] +11ms
```

packages/vm/lib/evm/interpreter.ts
```diff
@@ -193,6 +197,29 @@ export default class Interpreter {
...
+      gas: '0x' + eventObj.gasLeft.toString('hex'),
+      gasCost: '0x' + eventObj.opcode.fee.toString(16),
+      stack: hexStack,
+      depth: eventObj.depth,
+    }
+
+    if (!(nameLC in this.opDebuggers)) {
+      this.opDebuggers[nameLC] = createDebugLogger(`vm:ops:${nameLC}`)
+    }
+    this.opDebuggers[nameLC](JSON.stringify(opTrace))
```
cgewecke, line 221: This is really nice.

@jochem-brouwer You mentioned wanting to diff the geth / ethereumjs traces - would the most convenient thing be to have two files with JSONs of the respective traces? 

If so debug has [custom formatters][1] which might help with this. 

I think you could collect the steps into a trace and then write to `ejs.vm.<txhash>.json`, when `DEBUG=vm:savetrace` (or something)

[1]: https://www.npmjs.com/package/debug#custom-formatters

packages/vm/lib/evm/evm.ts
```diff
@@ -297,12 +331,15 @@ export default class EVM {
...
+        debug(`Not enough gas or code size not allowed (>= Homestead)`)
         result = { ...result, ...OOGResult(message.gasLimit) }
       } else {
         // we are in Frontier
+        debug(`Not enough gas or code size not allowed (Frontier)`)
         if (totalGas.sub(returnFee).lte(message.gasLimit)) {
           // we cannot pay the code deposit fee (but the deposit code actually did run)
-          result = { ...result, ...COOGResult(totalGas.sub(returnFee)) }
+          //result = { ...result, ...COOGResult(totalGas.sub(returnFee)) }
+          result.gasUsed = totalGas.sub(returnFee)
```
cgewecke, line 342: Out of curiosity - this is a simplification e.g the result is already formatted as COOGResult? 

Should [COOGResult][1] be removed in this PR?

[1]: https://github.com/ethereumjs/ethereumjs-monorepo/blob/80093e6f650427d3cf25d15f2b12fd7e3948c803/packages/vm/lib/evm/evm.ts#L78-L92

cgewecke (approved):
Looks great to me! 💯

packages/vm/lib/evm/evm.ts
holgerd77: Oh oh oh, this actually sneaked in along doing debugging work for #1076, this shouldn't be here. 😏

@jochem-brouwer just to make you aware, can you please take care of this along working on #1081 respectively along some rebase since you are touching this code anyhow? Thanks!

packages/vm/lib/evm/interpreter.ts
holgerd77: I would guess this is a good suggestion since comparing might be easiest on a per file basis with some linux CLI tools? @jochem-brouwer WDYT?

PR #1084, opened 2021-02-04
VM: Add type definitions for logs

LogvinovLeon:


holgerd77:
Hi there, this new data structure for logs is definitely an improvement, unfortunately we are at least weeks - if not months - away from a next major VM release, and we wouldn't want to merge in breaking changes atm. Is it ok if we leave this PR open for now (will mark as "blocked") and we come back to this once a major release is in sight?

LogvinovLeon:
Sure

holgerd77:
Ok, have added a label and added to #1024 to include in the next breaking release v6

holgerd77:
(ah, the tests should pass before)

LogvinovLeon:
Ok, change of plans here. It appears, that this logs format is actually not random, but the same that is used for RLP encoding in receipt trie. Therefore - no need to change it. I've just kept it like that and typed it as it is. No breaking changes any more ;)

packages/vm/lib/evm/types.ts
```diff
@@ -0,0 +1 @@
+export type Log = [Buffer, Buffer[], Buffer]
```
ryanio, line 1: great PR! unfortunately we are not on TypeScript 4.0 yet where we can [label tuple elements](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#labeled-tuple-elements) but meanwhile I think this would be helpful:

```suggestion
export type Log = [Buffer, Buffer[], Buffer] // [address, topics, data]
```

packages/vm/lib/evm/types.ts
```diff
@@ -0,0 +1 @@
+export type Log = [Buffer, Buffer[], Buffer] // [address, topics, data]
```
ryanio, line 1: Ok just looking one last time, I think a proper doc here should be better for editor hinting 🙂 then I can give a final approval and we can merge this in!

```suggestion
/**
 * Logs that the contract emitted.
 * Format: [address, topics, data]
 */
export type Log = [Buffer, Buffer[], Buffer]
```

ryanio:
Thanks for updating the PR to not introduce any breaking changes, I have removed the blocked label (and removed the entry from the v6 release planning doc since this can be merge in now).

ryanio (approved):
lgtm!

packages/vm/lib/evm/types.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Logs that the contract emitted.
```
ryanio, line 2: plural -> singular for clarity

```suggestion
 * Log that the contract emitted.
```

ryanio (approved):

PR #1087, opened 2021-02-07
Client & dDevp2p connection improvements

holgerd77:


holgerd77:
Phew, this was a large battle (half of the day) for somewhat smaller changes, in sum this improves both debugging and reliability of a client connection though. A bit laborious nevertheless.

Following changes:

---

4063ffda and 1e94410 switches to use the `sendMethod()` API function from the devp22 ETH and LES protocol classes in `RlpxSender`, before an internal `_send` method was used. This moved around all the checks in e.g. `ETH.sendMessage()` and - that's my main reason of implementing - skipped all the debug messages on `ETH` message sending (apart from the `STATUS` send), which was a source of confusion in the client for a long time when run with the `DEBUG` option.

Now `DEBUG` output with `eth:*` activated looks like this:

Bildschirmfoto 2021-02-07 um 14 44 48

E.g. `Send GET_BLOCK_HEADERS...` and similar messages are new and the outgoing communication gets a lot more transparent.

I had to adopt the `LES.sendMessage()` signature for this respectively unify with the `ETH.sendMessage()` method. This should be ok since the next `devp2p` release will be breaking anyhow and - apart from that - is the LES message not really looked for atm anyhow (which we should mention in the README actually).

---

bd6d9fc sets the `MINPEERS_DEFAULT` from 2 to 1. This is an easy change which we should do - at least for now where we do a lot of debugging - since this directly improves the sync pick-up on the PoA networks.

---

aea0bfa sets `started` to `true` before bootstrapping. This allows start() methods from other services to resolve earlier and also makes the sync pickup more reliable. Before peers bootstrapped didn't come into action because other services were not available yet. This might not be solved completely but at least gets better with this small change.

cgewecke (approved):
Great! All LGTM.

@holgerd77 Do you know what the original reasoning was behind having MINPEERS set to 2?

holgerd77:
Hi guys, thanks for the quick review, that helps me on a continued development! 😄 

@cgewecke I think having MINPEERS on 2 was just some general felt "engineering correctness" 😋 , I think I remember some exchange I had on this with Vinay 2 years ago already if it might make sense to lower.

I now had the chance to get a bit more of a feeling on your DNS changes, thanks again for this extensive PR! 👍  The connection on the PoA networks has definitely improved while on mainnet I am not sure if the connection wasn't rather more stale with the old v4 discovery version.

I think it might be generally a good thing to add an explicit option to the `DPT` class with some enum values or something to select which discovery types should be activated, WDYT? Then we can experiment with this a bit more, for mainnet we might want to deactivate DNS discovery (eventually) from within the client. Also wonder if this explicit `shouldFindNeighbours` option is then still necessary any more it would be *not* under the assumptions that a) DNS discovery is in ALL cases not benefiting from it and b) v4 discovery can't in ALL cases not live without it. My guess is both is the case, but not 100% sure.

One thing I also noticed is that discovery - I think mostly for the PoA networks - now either gets to a connection relatively quickly or - in the case *if* it tries longer - gets to some kind of stale state with close to no movement at all after some time with the peertable count staying stable and with not much of a chance for a connection still getting activated. I guess by this distributed design of DNS discovery (and not relying on some very selected bootstrap peers) this should normally not happen (or at least should be possible to avoid?)? Maybe we need another trigger in the client somewhere (similar (or added to) the "Retriggering bootstrap" mechanism? @cgewecke can you maybe generally do another round focusing on connection reliability and do some final polishing where you stumble upon things? Think this should likely be worth doing.

holgerd77:
(note: please always always read my posts on site, there is close to no one I don't need to correct on proof-reading after posting (stale -> stable, lol 😂 ).

cgewecke:
@holgerd77 

> The connection on the PoA networks has definitely improved while on mainnet I am not sure if the connection wasn't rather more stable with the old v4 discovery version.

> Maybe we need another trigger in the client somewhere (similar (or added to) the "Retriggering bootstrap" mechanism?

Ah great! Will investigate / reproduce on my end. (I had a similar impression that things were smoother on mainnet with the previous setup fwiw, but here's a lot of variability. Sometimes it's quick / sometimes slow)

As currently written, discovery is fetching `maxPeers` additional candidates from DNS (e.g on top of bootnodes) [during bootstrap][4]. And bootstrap [should be retriggered every ~60s if there are no peers][1]

It's also continuously fetching new candidates in the background as part of the [DPT refresh cycle][2] (~12 every 3 seconds by default)

Maybe this is not enough? It could be findNeighbours makes many more requests than this. I think we're also restricting ourselves to geth clients only with DNS discovery so some part of the network is invisible.

> gets to some kind of stale state with close to no movement at all after some time with the peertable count staying stable and with not much of a chance for a connection still getting activated (maybe on my setup for 20-30% of the cases I would say)

One possibility is that it's exhausting the DNS list (which might not be that big). For example [rinkeby.io][3] suggests there are only ~50 active peers in the network. 

Anyway will take another look today. Thanks for the in-depth review of this!  😄   



[1]: https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/lib/net/peerpool.ts#L202-L227
[2]: https://github.com/ethereumjs/ethereumjs-monorepo/blob/05366ae9b802b6c725244b4c302c135f4ddb95c0/packages/devp2p/src/dpt/dpt.ts#L234-L266
[3]: https://www.rinkeby.io/#stats
[4]: https://github.com/ethereumjs/ethereumjs-monorepo/blob/05366ae9b802b6c725244b4c302c135f4ddb95c0/packages/client/lib/net/server/rlpxserver.ts#L154-L155

Issue #1086, opened 2021-02-07
Client: PoA Clique (Goerli) run "Error: could not find parent header"


holgerd77:
When the client is run on PoA with the new clique changes in it gets the following error "Error: could not find parent header" thrown in the `BlockHeader.validate()` function when trying to store the blocks in the blockchain via `BlockFetcher.store()`.

I did some debugging this morning: this is due the the `hash` of the blocks being wrong respectively the `BlockHeader.hash()` function not delivering the correct hash, here is some output:

```shell
1 e26ba58f7923693693f3b6279b53bb29e17d6c7d1779bf2c793c14c969abf660 p:bf7e331f7f7c1dd2e05159666b3bf8bc7a8a3a9eb1d518969eab529dd9b88c1a 18
1 e26ba58f7923693693f3b6279b53bb29e17d6c7d1779bf2c793c14c969abf660 p:bf7e331f7f7c1dd2e05159666b3bf8bc7a8a3a9eb1d518969eab529dd9b88c1a 19
2 14db95de34b269dbbdae0d6b68d57e737270e98ebc6455716858cecf524fdd1f p:8f5bab218b6bb34476f51ca588e9f4553a3a7ce5e13a66c660a5283e97e9a85a $ 1
2 14db95de34b269dbbdae0d6b68d57e737270e98ebc6455716858cecf524fdd1f p:8f5bab218b6bb34476f51ca588e9f4553a3a7ce5e13a66c660a5283e97e9a85a 2
2 14db95de34b269dbbdae0d6b68d57e737270e98ebc6455716858cecf524fdd1f p:8f5bab218b6bb34476f51ca588e9f4553a3a7ce5e13a66c660a5283e97e9a85a 3
```

So these are the first two blocks, first hash is the wrongly calculated hash, second hash is the (correct) parent hash.

Root cause is likely that the `Common` instance is not set on all block or header creation occurrences in the client, specifically on [receiving block headers](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/lib/net/protocol/ethprotocol.ts#L60) in the ETH and LES protocol implementations. Since the clique hash calculation is deviating from the ethash block hash calculation (for clique the seal in `extraData` is removed from hashing) the hash will be wrong.

Further investigation, can't test so well yet because my Goerli connections are still not that good (my relatively bad internet connection might further contribute to this).

holgerd77:
[...]

PR #1088, opened 2021-02-08
Client/Block: Clique PoA fixes

holgerd77:


holgerd77:
I am as close as one can get: 😄 

Bildschirmfoto 2021-02-08 um 16 25 51

holgerd77:
Yeah, yeah, yeah

https://user-images.githubusercontent.com/931137/107247510-a474ec80-6a31-11eb-9317-879464fcb9ba.mov

🎉  🎉  🎉  🥳

holgerd77:
Got it, client is now syncing with the Goerli network. 😄 

This is now ready for review, change summary can be taken from the commit messages.

packages/client/lib/net/protocol/ethprotocol.ts
```diff
@@ -56,8 +56,6 @@ const messages: Message[] = [
     name: 'BlockHeaders',
     code: 0x04,
     encode: (headers: BlockHeader[]) => headers.map((h) => h.raw()),
-    decode: (headers: BlockHeaderBuffer[]) =>
-      headers.map((h) => BlockHeader.fromValuesArray(h, {})),
```
ryanio, line 60: Could we just add common here instead of removing the decode function?

`headers.map((h) => BlockHeader.fromValuesArray(h, { common: this.chain._common }))`

This commit (https://github.com/ethereumjs/ethereumjs-monorepo/pull/1088/commits/2ca45cda37865b15e5da650ce85162729314a6b3) feels like it's moving around and adding code to handle the buffers that was otherwise nicely handled here.

holgerd77: I had this implemented before (as announced here https://github.com/ethereumjs/ethereumjs-monorepo/issues/1086#issuecomment-774657947) but then reverted it because I had to pass `config` along every single `decode` method due to typing and the add another round of //es-lint-ignore comments since it wasn't used most of the time. So I decided that this would at the end be the cleaner way, especially since this was in an inconsistent state anyhow with the Block decoding also working without a decode method like now done here for the `BlockHeader` decoding. Wouldn't be a big fan of reverting, but if you have got a cleaner way of doing this feel free to push here, no emotional connections to this code 😋 , just would want this to work "somehow".

holgerd77: Maybe this can be readded (then also together with the Block decoding) along some refactor at some point towards some more type friendly handling of this whole ETH/LES message call handling?

holgerd77:
First time sync stops on Goerli is with `Error: invalid clique difficulty (block number=5281 hash=cca147be20ed1c416cfc8b533d15827764a216eef5d2c67f472ee9d143538926)`, so there still seems to be something wrong with this difficulty-in-turn, difficulty-out-turn check in [BlockHeader.validateCliqueDifficulty()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/header.ts#L399), difficulty of the respective block [5281](https://goerli.etherscan.io/block/5281) is `1`.

holgerd77:
(side question: so what do you guys think? I would love to add such additional information as in the error message above (with the block number and hash) also to the other block/header error messages (and also expand to other libraries), for now I just kept it to the new error messages to not change anything existing here. Do you think this is acceptable?)

packages/client/lib/net/protocol/ethprotocol.ts
ryanio: oh yeah passing common into decode would be a lot of passing around. can't we use the common from `chain` in `EthProtocolOptions`?

> Maybe this can be readded (then also together with the Block decoding) along some refactor at some point towards some more type friendly handling of this whole ETH/LES message call handling?

yes for sure, I will start thinking about this. I think if we inline the `messages` const into methods on the actual class then it would allow for seamless type info without needing to define the additional type helper `EthProtocolMethods`.

ryanio:
yeah I love the additional context for the error messages, I wonder if we can stick it in some meta field on the error object to not disturb the prior messages, but then I'm not sure if it would output on a `toString()`. another idea could be to make some kind of error factory that would try to insert this data so we don't have to add it to each error individually.

in general i think the more we use enums for error messages gives us more flexibility to not worry about disturbing the string content. it would also free us to more easily fix typos or improve wording/clarity incrementally. perhaps we could do a sprint to swap error strings for enums in v6. i found geth's code base also does a really good job of defining and documenting their error enums ([example 1](https://github.com/ethereum/go-ethereum/blob/053ed9cc847647a9b3ef707d0efe7104c4ab2a4c/core/vm/errors.go#L24), [example 2](https://github.com/ethereum/go-ethereum/blob/master/consensus/clique/clique.go#L75)).

ryanio:
> Maybe it's the order in blockchain of the `validate` call and the signer updates? 🤔

That could definitely be it, it looks like `validate` is run before `cliqueUpdateVotes` and `cliqueUpdateLatestBlockSigners` in `_putBlockOrHeader`.

packages/client/lib/net/protocol/ethprotocol.ts
holgerd77: Ah, yes, you are so right, I didn't see the option that `messages` could be relatively easily inlined into the `Protocol` classes. Have removed the old commits via rebase and done this now in 08ffcd8. I had to add a `no-invalid-this` lint ignore rule here since eslint was complaining, but I checked with a `console.log()` in the decode message and `this` is properly resolved and the `common` object is present and not set to `undefined` or something.

This is really much nicer now. 😄

holgerd77:
(side note for reviewers: the clique error from above is independent from this PR - just noted it down for convenience since I looked a bit further - so this remains open for review)

holgerd77:
@ryanio that's a great idea with the error sprint, have added this to #1024. 😄

holgerd77:
@ryanio regarding error annotation this should already be better I guess a738424 ? I think we should be safe to at least already update the already-dynamic error messages which add custom data to a static prefix message (like the `invalid transactions: ...` error from the `Block` class.

holgerd77:
This still needs a review as well.

holgerd77:
(@ryanio thanks for reviewing #1089)

holgerd77:
Guys, I will do an admin merge here, otherwise this will get bloated up too much and more and more difficult to review, and I need these changes for continued work this morning. I did a self-review of the remaining changes, I think all are local and unproblematic, the fixed clique hash function change has already be confirmed by the geth team, and the reactivated total difficulty is also obvious if looking at e.g. an [Etherscan block](https://rinkeby.etherscan.io/block/55) where the total difficulty is still present. The common addition to the ETH/LES protocols respectively the block creations has already been reviewed and discussed with @ryanio.

Fee free to do a post-merge review if you are not satisfied with something, I'll happily integrate into a follow-up PR.

PR #1089, opened 2021-02-09
Block, Tx: always copy Common on instantiation / Fix HF consistency bugs

holgerd77:
I think we can safe us from a load of problems when the `Common` instance in block and tx is just copied over. After months of thinking about this I am pretty much sure this is the right thing to do. Both txs and blocks just DO have a certain HF state - and if this is attached to an outer `Common` which - e.g. in the case of our client with `chainCommon` - might move on or change, this will lead to consistency problems if the respective Tx or Block objects are still around.

PR was triggered when running the client on Rinkeby which broke at bloke 55 with:

image

This is now fixed with this PR - as always replaced by another error though (`Invalid receipt trie`). Original root cause was the block being executed on the wrong HF though.

I think these `Common`-copy-fix can go in as a bug fix, this will very much rather help prevent unexpected behavior within the community using the library, maybe there might be some rather unintended use cases (which should rather be discouraged) of the library where this might break, but I am pretty sure the benefits weight significantly more here.

In the client as well as in the blockchain library I then switched to `hardforkByBlockNumber` block and header instantiations. This should once and for all - together with the change from above - free us from all consistency problems in this regard.

I also fixed bugs along the way in `Block` where the tx objects have been instantiated with the wrong HF in the static constructors when the `hardforkByBlockNumber` option has been used for block instantiation.

Note that this PR is targeted towards the `client-block-poa-fixes` branch.

### DIFF
--- a/packages/block/src/block.ts
+++ b/packages/block/src/block.ts
@@ -25,7 +25,11 @@ export class Block {
     // parse transactions
     const transactions = []
     for (const txData of txsData || []) {
-      const tx = Transaction.fromTxData(txData, opts as TxOptions)
+      const tx = Transaction.fromTxData(txData, {
+        ...opts,
+        // Use header common in case of hardforkByBlockNumber being activated
+        common: header._common,
+      } as TxOptions)
       transactions.push(tx)
     }
 
@@ -34,7 +38,10 @@ export class Block {
     for (const uhData of uhsData || []) {
       const uh = BlockHeader.fromHeaderData(uhData, {
         ...opts,
-        // Disable this option here (all other options carried over), since this overwrites the provided Difficulty to an incorrect value
+        // Use header common in case of hardforkByBlockNumber being activated
+        common: header._common,
+        // Disable this option here (all other options carried over), since this overwrites
+        // the provided Difficulty to an incorrect value
         calcDifficultyFromHeader: undefined,
       })
       uncleHeaders.push(uh)
@@ -65,7 +72,13 @@ export class Block {
     // parse transactions
     const transactions = []
     for (const txData of txsData || []) {
-      transactions.push(Transaction.fromValuesArray(txData, opts))
+      transactions.push(
+        Transaction.fromValuesArray(txData, {
+          ...opts,
+          // Use header common in case of hardforkByBlockNumber being activated
+          common: header._common,
+        })
+      )
     }
 
     // parse uncle headers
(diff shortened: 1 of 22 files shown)


holgerd77:
This is the next failure now (independent from this PR):

image

Block number [55](https://rinkeby.etherscan.io/block/55) on Rinkeby.

holgerd77:
Here is the debug output from the VM:

image

holgerd77:
Have found another state changing error in 749b93f where we were still updating the miner account in `runTx()` also on PoA chains. Block from above is still failing though, now with another (very likely) still invalid state root (if it is not something else from the receipt):

```shell
{
  stateRoot: <Buffer 9e 15 ba 99 5a a6 a3 4f 60 f4 4e 16 4d a3 de aa 75 ab e7 bb eb af cc b4 fd 44 47 b9 99 25 d1 6d>,
  gasUsed: <Buffer 52 08>,
  bitvector: <Buffer 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 ... 206 more bytes>,
  logs: []
}
```

holgerd77:
Ok, I've now added two additional loggers `vm:state` and `vm:block` here, this should - on a first round - very much complete the picture. 😄  //cc @jochem-brouwer 

Output as following:

image

Still open for review.

holgerd77:
But back to the failing tx on Rinkeby (the above was just some other testing log output).

So here is the log:

image

Actually this is really a dead-simple transaction, which really puzzles me. Just a simple transfer of value. The only thing I can think of right now is something going wrong due to the very large initial account balance of `0x31b98d14007bdee637298086988a0bbd31184523`.

holgerd77:
Got it. 😄 

The fee reward didn't need to be removed completely (have removed the commit doing this) but instead the fees have to be provided to the signer account. VM execution is now going forward. I have also added additional backwards-compatibility checks for adding the block clique functionality in the VM as well as some reasonable default for returning a signer on default blocks (zero address), otherwise e.g. the VM would break when be run on a default block.

holgerd77:
Argh, guys, can someone please review this? Tomorrow I would add yet another PR on top of these two! 😀😅

ryanio (approved):
lgtm! nice fixes, and wow this new debug output looks really amazing

Issue #1090, opened 2021-02-11
Client -> Rinkeby: Consensus Error on Block 14182 (invalid receiptTrie)


holgerd77:
At least we made it up to block 14182 after the fixes from #1089 - that's something. To relativize a bit, there were only 10 or so transactions along the way. 😂  😂  😂

This one is significantly more complex, here is the output:

![image](https://user-images.githubusercontent.com/931137/107617074-84227900-6c4f-11eb-81e5-10cf9837f771.png)

![image](https://user-images.githubusercontent.com/931137/107617141-9ef4ed80-6c4f-11eb-8acc-82fee56745fd.png)

Block on Rinkeby: [14182](https://rinkeby.etherscan.io/block/14182)

Some first observations and thoughts:

- Shouldn't be related to #1076, since this one here is on `SpuriousDragon` and #1076 is fixing a pre-Homestead bug (that said, @jochem-brouwer you mentioned that you need to fix other bugs. Have these already been fixed and are these independent from the Homestead HF and might play a role here?)
- I have to say that this early breaking on PoA puzzles me since mainnet was running so smoothly for such a long period of blocks: this might be an indicator that this has something to do with the PoA implementation as well, can't think of anything atm though
- Gas used matches (with `1339314`), being an indicator that there is likely nothing wrong along opcode execution I guess? 🤔

Ok. We'll see. 😄  😋

jochem-brouwer:
Hi @holgerd77, first of all these debug messages look great!

The problem in #1076 is that this particular transaction seems to trigger multiple, possibly unrelated bugs. The problem is there that a wrong value is written to a storage slot. Therefore the state root is wrong after execution. It might be that the bug here is also caused by that bug, it could be a bug in an opcode which has not changed during Frontier -> SpuriousDragon.

holgerd77:
@jochem-brouwer thanks for the quick response - yeah- always difficult to spot. I then still wonder though why such generic bugs have not been triggered on mainnet with this load of transactions? 🤔  This might be an indicator that wrong behavior is either a) PoA change or b) HF related (since on mainnet we are still running everything on `Frontier`).

But of course might also be some incidence and a very early transaction has triggered the same bug (probability nevertheless points in the other direction I guess).

(argh, I have got a signifcant problem with my "b" on the keyboard, really annoying 😜 )

holgerd77:
[...]

PR #1091, opened 2021-02-11
Client, VM -> Rinkeby: Fix consensus error on block 14182 (invalid receipt trie)

holgerd77:
Fixing #1090 

I will actually not start working on new consensus bugs without adding a new meaningful debug tool. 😂

This one from this first commit 8d0afb7 is actually super-handy: it adds a new CLI option (mainly for internal usage, but of course also can be leveraged by consumers) `debugCode` which generates a dynamic script on a VM block execution failure. 

So this can be run with:

```shell
npm run client:start -- --network=rinkeby --debugCode
```

Script looks something like this and will be logged out to the console:

```typescript
/**
 * Script for locally executing a block in the EthereumJS VM,
 * meant to be used from packages/vm directory within the
 * https://github.com/ethereumjs/ethereumjs-monorepo repository.
 *
 * Block: 14182
 * Hardfork: spuriousDragon
 *
 * Run with: DEBUG=vm:*:*,vm:*,-vm:ops:* ts-node [SCRIPT_NAME].ts
 *
 */

const level = require('level')
import VM from './lib'
import { SecureTrie as Trie } from '@ethereumjs/trie'
import { DefaultStateManager } from './lib/state'
import Blockchain from '@ethereumjs/blockchain'

const main = async () => {
  const common = new Common({ chain: 'rinkeby', hardfork: 'spuriousDragon' })
  const block = Block.fromRLPSerializedBlock(Buffer.from('f91692f9025ba0b477241819c2222f...', 'hex'), { common })

  const stateDB = level('/Ethereum/ethereumjs/rinkeby/state')
  const trie = new Trie(stateDB)
  const stateManager = new DefaultStateManager({ trie, common })
  // Ensure we run on the right root
  stateManager.setStateRoot(Buffer.from('135fae6d4909a072d77707a797769afc72f6239a72173e0f411c7700af66ac26', 'hex'))


  const chainDB = level('/Ethereum/ethereumjs/rinkeby/chain')
  const blockchain = await Blockchain.create({
    db: chainDB,
    common,
    validateBlocks: true,
    validateConsensus: false,
  })
  const vm = new VM({ stateManager, blockchain, common })

  await vm.runBlock({ block })
}

main()
```

The script works out of the box, has all the correct values filled in (HFs, directories, block RLP) and can directly be copied over into the VM folder and then executed to reliably reproduce the error locally. 😄

### DIFF
--- a/packages/client/lib/util/debug.ts
+++ b/packages/client/lib/util/debug.ts
@@ -1,0 +1,65 @@
+import { Block } from '@ethereumjs/block'
+import { VMExecution } from '../sync/execution'
+
+/**
+ * Generates a code snippet which can be used to replay an erraneous block
+ * locally in the VM
+ *
+ * @param block
+ */
+export async function debugCodeReplayBlock(execution: VMExecution, block: Block) {
+  const code = `
+/**
+ * Script for locally executing a block in the EthereumJS VM,
+ * meant to be used from packages/vm directory within the
+ * https://github.com/ethereumjs/ethereumjs-monorepo repository.
+ * 
+ * Block: ${block.header.number}
+ * Hardfork: ${execution.hardfork}
+ * 
+ * Run with: DEBUG=vm:*:*,vm:*,-vm:ops:* ts-node [SCRIPT_NAME].ts
+ * 
+ */
+
+const level = require('level')
+import Common from '@ethereumjs/common'
+import { Block } from '@ethereumjs/block'
+import VM from './lib'
+import { SecureTrie as Trie } from '@ethereumjs/trie'
+import { DefaultStateManager } from './lib/state'
+import Blockchain from '@ethereumjs/blockchain'
+
+const main = async () => {
+  const common = new Common({ chain: '${execution.config.execCommon.chainName()}', hardfork: '${
+    execution.hardfork
+  }' })
+  const block = Block.fromRLPSerializedBlock(Buffer.from('${block
+    .serialize()
+    .toString('hex')}', 'hex'), { common })
+
+  const stateDB = level('${execution.config.getStateDataDirectory()}')
+  const trie = new Trie(stateDB)
+  const stateManager = new DefaultStateManager({ trie, common })
+  // Ensure we run on the right root
+  stateManager.setStateRoot(Buffer.from('${(
+    await execution.vm.stateManager.getStateRoot(true)
+  ).toString('hex')}', 'hex'))
+
+
+  const chainDB = level('${execution.config.getChainDataDirectory()}')
+  const blockchain = await Blockchain.create({
+    db: chainDB,
+    common,
+    validateBlocks: true,
+    validateConsensus: false,
+  })
+  const vm = new VM({ stateManager, blockchain, common })
+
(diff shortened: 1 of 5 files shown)


jochem-brouwer:
Hi @holgerd77, please check this PR #1093. I wanted to fetch the Geth JSON trace from an archive node, but sadly the eth dev ops node seems to be down.

I am currently spinning up a Rinkeby archive node to get the trace. I'll upload it to the PR, so you can directly compare traces and see exactly at which point the two transactions diverge 😄

holgerd77:
@jochem-brouwer great 👍

jochem-brouwer:
Will merge the debug stuff here (you can remove these debug files later when ready for review) so we can compare stack traces.

holgerd77:
@jochem-brouwer Ok. 😄 

The CI got stuck at some point, I will merge this in by admin merge so that we can continue.

jochem-brouwer:
`debug.ts` is [not](https://github.com/ethereumjs/ethereumjs-monorepo/blob/7c4c5b939600a6c1039983ee99f53ce9a65de621/packages/client/lib/util/debug.ts#L11) a valid script?

holgerd77:
No, not yet. You need to run the client with the new option `npm run client:start -- --network=rinkeby --debugCode` the final script will then printed out to the console on VM block execution error with the correct values filled in.

holgerd77:
(so this will work for all future debugging cases)

packages/client/lib/util/debug.ts
```diff
@@ -0,0 +1,65 @@
...
+  const block = Block.fromRLPSerializedBlock(Buffer.from('${block
+    .serialize()
+    .toString('hex')}', 'hex'), { common })
+
+  const stateDB = level('${execution.config.getStateDataDirectory()}')
+  const trie = new Trie(stateDB)
+  const stateManager = new DefaultStateManager({ trie, common })
+  // Ensure we run on the right root
+  stateManager.setStateRoot(Buffer.from('${(
+    await execution.vm.stateManager.getStateRoot(true)
```
jochem-brouwer, line 45: Ah yes, I should have spotted this earlier. You should get the state root of the previous block, and set the state managers' root to the root of that block. (Can probably just get it from vm's blockchain)

packages/client/lib/util/debug.ts
```diff
@@ -0,0 +1,65 @@
...
+  const code = `
+/**
+ * Script for locally executing a block in the EthereumJS VM,
+ * meant to be used from packages/vm directory within the
+ * https://github.com/ethereumjs/ethereumjs-monorepo repository.
+ * 
+ * Block: ${block.header.number}
+ * Hardfork: ${execution.hardfork}
+ * 
+ * Run with: DEBUG=vm:*:*,vm:*,-vm:ops:* ts-node [SCRIPT_NAME].ts
```
jochem-brouwer, line 20: Should be `export DEBUG`. If I don't use export then debug is not triggered.

holgerd77: Ah, no, export should not be needed. Your are likely putting an `&&` in between the commands?

PR #1093, opened 2021-02-11
VM: debug tools

jochem-brouwer:
Debug tools for VM which helps debugging consensus bugs.


holgerd77:
Ah, I am actually not sure what you are doing here, why do you want to merge these heavy bug-specific stack traces into master? And can we eventually stay in the `scripts` folder and not add just another folder `tools`here + can you please add a one-line in `developer.md` on usage? Thanks!

holgerd77:
(if you just want to distribute the stack traces maybe simply put in a gist? I think we shouldn't use master to arbitrarily root some stuff temporarily)

jochem-brouwer:
Just to clarify; I am not sure if I want to merge this tool, but just wanted to share it in case it is handy for these consensus bugs. You are right that we should not dump in raw JSON traces into the master branch.

holgerd77:
Ah, thanks for the update! 😄  👍

jochem-brouwer:
Will close since I don't have the intention to merge this.

PR #1097, opened 2021-02-12
Add node discovery mode flags & define network-based defaults

cgewecke:
PR:

+ Adds two boolean cli flags: --discV4 and --discDns to enable/disable these discovery modes
+ Configures default discovery modes as below

| Network | V4 | DNS |
|-----|-----|-----|
| mainnet | ✅  | ❌ |
| rinkeby | ❌  | ✅  |
| goerli | ❌  | ✅  |
| ropsten | ❌  | ✅  |

I've also delegated DNS peer addition in the dpt refresh cycle to the`onServerPeers` handler (now renamed `_addPeerBatch`). This method staggers peer addition with a 200ms delay per peer and distributes the background work of  querying peers more evenly.


holgerd77 (approved):
Thanks Chris, this looks good and very clean, also tested locally - yes - this *does* improve mainnet sync pickup, think this further confirms the intuition we both had here.

Will merge.

Issue #969, opened 2020-11-27
Cases where changing the fork in Common invalidates block/tx


jochem-brouwer:
In [this specific test](https://github.com/ethereumjs/ethereumjs-vm/pull/966/files#diff-1119148a3aee9540929fd72987c533cd2da4a81aa3f7006a072b1e44718d337bR143) I noticed that I had to specifically pass on the Common object which is used by the VM in order to get the right results. I would assume that any package which is used downstream (tx/block) would use the Common of the VM in case it is ran. But this is not the case. One can create a VM which points to `Istanbul`, and then instantiate a block which uses `Chainstart`. This leads to the very weird situation where the EVM executes rules according to `Istanbul` (e.g. gas calculations), but the block will validate the rules against `Chainstart`. This sounds like a very bug-prone feature and it seems very inconsistent to me. 

I'd say a way to fix this is to call `setHardfork` on the tx/block before we run it and thus set the common to the same hardfork as the one the VM is using. But this would be a breaking change as now suddenly the Common of tx/block can switch forks when you run it. We could also explicitly communicate that if you want to ensure "correctness", then you need to always pass on the common of the VM in case you create a tx/block.

One other nasty problem here is that in the constructor of block/tx there are some validations/calculations which explicitly depend upon the fork. In `block` it checks DAO extra data, and if `calcDifficultyFromHeader` is set to `true` then it also calculates difficulty (which could thus change if you are on a different fork due to the difficulty bomb). In `tx` it also verifies that the `v` value of the signature is correct.

holgerd77:
From my perspective this is neither a bug nor an inconsistency but is a by-design choice.

Both the VM and e.g. a block are independent entities. If you create a VM with common set to `istanbul`, you get a VM comforming to the `istanbul` rules. If you instantiate a block e.g. retrieved by RPC from `byzantium` times you should instantiate it with `byzantium`.

Then you can decide how to use these independent entities. If you then run this `byzantium` block in the `istanbul` VM this is prone to get somewhat wrong. And this won't get any better if you automatically switch `common` from the block. In this example you would force a `Byzantium` block into an `istanbul` behavior, likely even more error prone than the original behavior might be.

Therefore you've got these two "operation modes".

1. If you are in a static HF context you create an outer `Common` and then use this for both instantiations.
[...]

PR #1101, opened 2021-02-12
VM/Client: fix HF switching. Fix TangerineWhistle update bug.

### LINKED ISSUE #1090: Client -> Rinkeby: Consensus Error on Block 14182 (invalid receiptTrie)

jochem-brouwer:
Closes #1090. Cherry-picked from #1094.

Very nasty bug. There are two problems:

(1) Client does not update the opcodes of VM when we change hard forks. In practice, some stuff happens though: if we update a context variable which we were already using (like tx data gas cost for nonzero data) then this is used (as we dynamically fetch this from the current Common). However, if a HF introduces new opcodes, then these are not added to VM and are thus not accessible. 

(2) We never updated TangerineWhistle gas cost changes in the first place. If you instantiate a Common with Chainstart, then switch to TangerineWhistle, and even called `_updateOpcodes()` before, then gas costs are *not* updated in the VM.

(Side note: also weird that tests did not cover this. On the other hand, there are not that many tests where we check HF switching).

### DIFF
--- a/packages/client/lib/sync/execution/vmexecution.ts
+++ b/packages/client/lib/sync/execution/vmexecution.ts
@@ -49,6 +49,7 @@ export class VMExecution extends Execution {
     const blockNumber = headBlock.header.number.toNumber()
     this.config.execCommon.setHardforkByBlockNumber(blockNumber)
     this.hardfork = this.config.execCommon.hardfork()
+    this.vm._updateOpcodes()
     this.config.logger.info(`Initializing VM execution hardfork=${this.hardfork}`)
   }
 
@@ -112,6 +113,7 @@ export class VMExecution extends Execution {
                 `Execution hardfork switch on block number=${blockNumber} hash=${hash} old=${this.hardfork} new=${hardfork}`
               )
               this.hardfork = this.config.execCommon.setHardforkByBlockNumber(blockNumber)
+              this.vm._updateOpcodes()
             }
             await this.vm.runBlock({ block, root: parentState })
             txCounter += block.transactions.length
--- a/packages/vm/lib/evm/opcodes/codes.ts
+++ b/packages/vm/lib/evm/opcodes/codes.ts
@@ -191,11 +191,27 @@ const opcodes = {
 
 // Array of hard forks in order. These changes are repeatedly applied to `opcodes` until the hard fork is in the future based upon the common
 // TODO: All gas price changes should be moved to common
+// If the base gas cost of any of the operations change, then these should also be added to this list.
+// If there are context variables changed (such as "warm slot reads") which are not the base gas fees,
+// Then this does not have to be added.
 const hardforkOpcodes = [
   {
     hardforkName: 'homestead',
     opcodes: {
       0xf4: { name: 'DELEGATECALL', isAsync: true }, // EIP 7
+    },
+  },
+  {
+    hardforkName: 'tangerineWhistle',
+    opcodes: {
+      0x54: { name: 'SLOAD', isAsync: true },
+      0xf1: { name: 'CALL', isAsync: true },
+      0xf2: { name: 'CALLCODE', isAsync: true },
+      0x3b: { name: 'EXTCODESIZE', isAsync: true },
+      0x3c: { name: 'EXTCODECOPY', isAsync: true },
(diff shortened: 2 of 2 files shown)


packages/client/lib/sync/execution/vmexecution.ts
```diff
@@ -49,6 +49,7 @@ export class VMExecution extends Execution {
     const blockNumber = headBlock.header.number.toNumber()
     this.config.execCommon.setHardforkByBlockNumber(blockNumber)
     this.hardfork = this.config.execCommon.hardfork()
+    this.vm._updateOpcodes()
```
jochem-brouwer, line 52: Side note: we can also launch VM with the `selectHardforkByBlockNumber: true` option, then this stuff is handled in the VM. We would only not directly see in Client that we are switching HFs. This would generally go well if we always start the Common with Chainstart, and then let VM check per block number (via this option) if we should change it.

jochem-brouwer:
Also please check that Client now processes the wrong rinkeby block the right way. It does for me.

If client refuses to run the 14182 block, and you have already executed block 14181 previously, then you can temporarily change `client/lib/sync/execution/vmexecution.ts`, add this to `open()`:

```typescript
  async open(): Promise<void> {
    await this.vm.blockchain.setIteratorHead('vm', Buffer.from('b477241819c2222fe45f63300cc2b55760e08f626461f7d37deddcfcc6055def','hex'))
    const headBlock = await this.vm.blockchain.getIteratorHead()
    const blockNumber = headBlock.header.number.toNumber()
    this.config.execCommon.setHardforkByBlockNumber(blockNumber)
    this.hardfork = this.config.execCommon.hardfork()
    this.vm._updateOpcodes()
    this.config.logger.info(`Initializing VM execution hardfork=${this.hardfork}`)
  }
  ```
  
  This b477... hex is the hash of block 14181.

packages/client/lib/sync/execution/vmexecution.ts
holgerd77: Great find: already thought some time ago: would it make sense for such uses cases to make the `Common` also a class inheriting from `EventEmitter` and then submit an event `hardforkChanged` whenever a HF change has been triggered?

Then the VM can listen to this internally and itself do a call to `_updateOpcodes()`. Calling this from the outside is not so optimal and likely other users will stumble upon this as well.

(making a class an EventEmitter should be a non-breaking change I would say?)

jochem-brouwer: Yeah, this would make a lot of sense. I think at the moment our packages are mostly built around non-changing Common hardforks. We also have this issue for instance: #969. Although I'm not sure if reacting on events would solve that particular problem in an elegant way 🤔 .

jochem-brouwer: And it would indeed be non-breaking, since you only add methods. (Unless I'm missing some very specific things which might be broken if you suddenly start extending classes).

holgerd77 (approved):
Will give this an approval independently from this EventEmitter question, we can discuss separately and eventually update here along a subsequent PR.

jochem-brouwer:
@holgerd77 can you check that Rinkeby block now processes on your node?

holgerd77:
Synced from scratch and can confirm that the sync is going on. 🎉 🎉 🎉

PR #1103, opened 2021-02-12
Client, Blockchain: Fix Goerli Clique Difficulty Bug

### LINKED ISSUE #1100: Client -> Goerli: Sync error "invalid clique difficulty" (block 5281)
I should really stop with the lately accustomed habit of adding new discussions to existing threads, could hardly refind this one (and was already convincing me that it likely never happened, lol 😂 )

But at the end I spotted here https://github.com/ethereumjs/ethereumjs-monorepo/pull/1088#issuecomment-775498167 on a closed PR. 😄 

So taking over the first few lines, otherwise let's continue the investigation here:

> First time sync stops on Goerli is with Error: invalid clique difficulty (block number=5281 hash=cca147be20ed1c416cfc8b533d15827764a216eef5d2c67f472ee9d143538926), so there still seems to be something wrong with this difficulty-in-turn, difficulty-out-turn check in BlockHeader.validateCliqueDifficulty(), difficulty of the respective block 5281 is 1.

holgerd77:
Fixes #1100

Got it, cause for the difficulty bug was the ordering of the signers mentioned in the issue. 😄 

Now there is an additional clique bug triggered some blocks down the road though, now with a pretty complex voting scenario. Here is the output:

image

### DIFF
--- a/packages/blockchain/src/index.ts
+++ b/packages/blockchain/src/index.ts
@@ -511,7 +511,6 @@ export default class Blockchain implements BlockchainInterface {
       // Do two rounds here, one to execute on a potential previously reached consensus
       // on the newly touched beneficiary, one with the added new vote
       for (let round = 1; round <= 2; round++) {
-
         // See if there is a new majority consensus to update the signer list
         const lastEpochBlockNumber = header.number.sub(
           header.number.mod(new BN(this._common.consensusConfig().epoch))
@@ -521,7 +520,7 @@ export default class Blockchain implements BlockchainInterface {
         let consensus = false
 
         // AUTH vote analysis
-        if (round === 1 || round === 2 && nonce.equals(CLIQUE_NONCE_AUTH)) {
+        if (round === 1 || (round === 2 && nonce.equals(CLIQUE_NONCE_AUTH))) {
           let beneficiaryVotesAUTH = this._cliqueLatestVotes.filter((vote) => {
             return (
               vote[0].gte(lastEpochBlockNumber) &&
@@ -552,7 +551,7 @@ export default class Blockchain implements BlockchainInterface {
           }
         }
         // DROP vote
-        if (round === 1 || round === 2 && nonce.equals(CLIQUE_NONCE_DROP)) {
+        if (round === 1 || (round === 2 && nonce.equals(CLIQUE_NONCE_DROP))) {
           let beneficiaryVotesDROP = this._cliqueLatestVotes.filter((vote) => {
             return (
               vote[0].gte(lastEpochBlockNumber) &&
@@ -569,9 +568,9 @@ export default class Blockchain implements BlockchainInterface {
             consensus = true
             // Drop signer
             activeSigners = activeSigners.filter((signer) => !signer.equals(beneficiary))
-            // Discard votes from removed signer
             this._cliqueLatestVotes = this._cliqueLatestVotes.filter(
-              (vote) => !vote[1][0].equals(beneficiary)
+              // Discard votes from removed signer and for removed signer
+              (vote) => !vote[1][0].equals(beneficiary) && !vote[1][1].equals(beneficiary)
             )
             debug(
               `[Block ${header.number.toNumber()}] Clique majority consensus (DROP ${beneficiary})`
--- a/packages/blockchain/src/index.ts
+++ b/packages/blockchain/src/index.ts
@@ -488,10 +488,8 @@ export default class Blockchain implements BlockchainInterface {
 
     await this.dbManager.batch(dbOps)
     // Output active signers for debugging purposes
-    let i = 0
-    for (const signer of this.cliqueActiveSigners()) {
+    for (const [i, signer] of this.cliqueActiveSigners().entries()) {
       debug(`Clique signer [${i}]: ${signer}`)
-      i++
     }
   }
 
--- a/packages/blockchain/src/index.ts
+++ b/packages/blockchain/src/index.ts
@@ -488,8 +488,10 @@ export default class Blockchain implements BlockchainInterface {
 
     await this.dbManager.batch(dbOps)
     // Output active signers for debugging purposes
-    for (const [i, signer] of this.cliqueActiveSigners().entries()) {
+    let i = 0
+    for (const signer of this.cliqueActiveSigners()) {
       debug(`Clique signer [${i}]: ${signer}`)
+      i++
     }
   }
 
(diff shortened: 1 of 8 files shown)


holgerd77:
(PR can also be merged if convenient, as you like)

jochem-brouwer:
There's a blockchain test failing? I suspect that in this case this is one of these cases where the test is actually wrong, and not the code?

holgerd77:
Ah. No? Where are you reading this information from?

jochem-brouwer:
[This](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1103/checks?check_run_id=1888723885) CI?

holgerd77:
Ah, non, that are just some test failures I haven't fixed yet caused by the clique fixes on the first bug. That's easily fixable and totally unrelated to the second bug I have posted here. This is again about clique voting and signer updates, still something wrong in the implementation there.

packages/blockchain/src/index.ts
```diff
@@ -483,6 +487,12 @@ export default class Blockchain implements BlockchainInterface {
     dbOps.push(DBOp.set(DBTarget.CliqueSignerStates, rlp.encode(formatted)))
 
     await this.dbManager.batch(dbOps)
+    // Output active signers for debugging purposes
+    let i = 0
+    for (const signer of this.cliqueActiveSigners()) {
+      debug(`Clique signer [${i}]: ${signer}`)
+      i++
+    }
```
ryanio, line 495: nice little es6 syntax destructuring trick for getting the index of an iteration in a for-of loop:
```suggestion
    for (const [i, signer] of this.cliqueActiveSigners().entries()) {
      debug(`Clique signer [${i}]: ${signer}`)
    }
```

jochem-brouwer: Wow, cool 😄

holgerd77:
Two things I noticed here:

1) The implementation of the vote count with [alreadyVoted](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/blockchain/src/index.ts#L526) doesn't seem correct to me. A vote from the past may not have been affected the majority consensus. If there has been a change on the votes (or the signer list) in between a double vote, this second vote may very well cause a new consensus situation, it just shouldn't been counted twice. So completely ignoring this second (or third, fourth,...) vote is likely not correct. Also from the spec:

> Only the latest proposal per target beneficiary is kept from a single signer.

This contradicts the current implementation.

2) I had the case when testing with Goerli that blockchain sync and VM execution were (as they often are) out of sync. When there is a change on the active signers in between, the block validation triggered in the VM will not work correctly for the part using the blockchain object (so e.g. the clique difficulty validation). I would suggest that we therefore skip the block validation in the VM since this is redundant anyhow with the blocks having already been validated on sync.

//cc @ryanio

ryanio:
I added some commits to resolve the failing blockchain tests and rebased on master.

---

> 1. The implementation of the vote count with [alreadyVoted](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/blockchain/src/index.ts#L526) doesn't seem correct to me. A vote from the past may not have been affected the majority consensus. If there has been a change on the votes (or the signer list) in between a double vote, this second vote may very well cause a new consensus situation, it just shouldn't been counted twice. So completely ignoring this second (or third, fourth,...) vote is likely not correct. Also from the spec:
> 
> > Only the latest proposal per target beneficiary is kept from a single signer.
> 
> This contradicts the current implementation.

Ok, I can work on this.

> 2. I had the case when testing with Goerli that blockchain sync and VM execution were (as they often are) out of sync. When there is a change on the active signers in between, the block validation triggered in the VM will not work correctly for the part using the blockchain object (so e.g. the clique difficulty validation). I would suggest that we therefore skip the block validation in the VM since this is redundant anyhow with the blocks having already been validated on sync.

Ah ok, should we configure the client's blockchain if (poa) then turn validateBlocks off?

holgerd77:
> I added some commits to resolve the failing blockchain tests and rebased on master.

Thanks for fixing the tests. 😄 

> > This contradicts the current implementation.
> 
> Ok, I can work on this.

Also experimented a bit on this yesterday, came up with the following code:

```typescript
let beneficiaryVotes
// AUTH vote
if (nonce.equals(CLIQUE_NONCE_AUTH)) {
  beneficiaryVotes = this._cliqueLatestVotes.filter(
    (vote) =>
      vote[0].gte(lastEpochBlockNumber) &&
      vote[1][1].equals(beneficiary) &&
      vote[1][2].equals(CLIQUE_NONCE_AUTH)
  )
  // Majority consensus
  if(beneficiaryVotes.length - numAlreadyVoted >= limit) {
    // Authorize new signer
    activeSigners.push(beneficiary)
    activeSigners.sort((a, b) => {
      // Sort by buffer size
      return a.toBuffer().compare(b.toBuffer())
    })
    // Discard votes for added signer
    this._cliqueLatestVotes = this._cliqueLatestVotes.filter(
      (vote) => !vote[1][1].equals(beneficiary)
    )
  }
// DROP vote
} else {
  beneficiaryVotes = this._cliqueLatestVotes.filter(
    (vote) =>
      vote[0].gte(lastEpochBlockNumber) &&
      vote[1][1].equals(beneficiary) &&
      vote[1][2].equals(CLIQUE_NONCE_DROP)
  )
  // Majority consensus
  if(beneficiaryVotes.length - numAlreadyVoted >= limit) {
    // Drop signer
    activeSigners = activeSigners.filter((signer) => !signer.equals(beneficiary))
    // Discard votes from removed signer
    this._cliqueLatestVotes = this._cliqueLatestVotes.filter(
      (vote) => !vote[1][0].equals(beneficiary)
    )
  }
}
if (beneficiaryVotes.length - numAlreadyVoted >= limit) {
  debug(
    `[Block ${header.number.toNumber()}] Clique majority consensus -> update signer states`
  )
  const newSignerState: CliqueSignerState = [header.number, activeSigners]
  await this.cliqueUpdateSignerStates(newSignerState)
}
```

So replacing the respective code parts before. This *did* solve the first Goerli failure on block 5310 posted in the first screenshot. I went into some new issues some blocks further though (seems clique edge cases have been really triggered on purpose along these first blocks, which ultimately is a good thing 😄 ). Also some additional tests were failing. So I hesitated pushing directly. Might serve as some inspiration though.

Maybe I'll also find some time to continue to work on this during the day.

> Ah ok, should we configure the client's blockchain if (poa) then turn validateBlocks off?

I would think we can generally turn this off for the VM since locks are already validated in the blockchain class and this is just executing the same code twice. Or am I missing something?

packages/blockchain/src/index.ts
```diff
@@ -483,6 +487,10 @@ export default class Blockchain implements BlockchainInterface {
     dbOps.push(DBOp.set(DBTarget.CliqueSignerStates, rlp.encode(formatted)))
 
     await this.dbManager.batch(dbOps)
+    // Output active signers for debugging purposes
+    for (const [i, signer] of this.cliqueActiveSigners().entries()) {
+      debug(`Clique signer [${i}]: ${signer}`)
+    }
```
holgerd77, line 493: Yeah, this is really nice stuff. 😄

packages/blockchain/test/clique.spec.ts
```diff
@@ -44,61 +44,61 @@ tape('Clique: Initialization', (t) => {
...
     publicKey: Buffer.from(
-      'd3e3d2b722e325bfc085ff5638a112b4e7e88ff13f92fc7f6cfc14b5a25e8d1545a2f27d8537b96e8919949d5f8c139ae7fc81aea7cf7fe5d43d7faaa038e35b',
+      'ca0a55f6e81cb897aee6a1c390aa83435c41048faa0564b226cfc9f3df48b73e846377fb0fd606df073addc7bd851f22547afbbdd5c3b028c91399df802083a2',
       'hex'
     ),
   }
 
   const C: Signer = {
-    address: new Address(Buffer.from('8458f408106c4875c96679f3f556a511beabe138', 'hex')),
+    address: new Address(Buffer.from('83c30730d1972baa09765a1ac72a43db27fedce5', 'hex')),
```
holgerd77, line 59: Lol, I also started fixing the tests yesterday but I went along with changing the order of the signers in the test cases. This is obviously the significantly better idea. 😎

holgerd77:
@ryanio ok, I've reworked this significantly, I still not have got 1-2 corner cases under control unfortunately. I nevertheless think that the reworked structure provides a better ground for further refinements, I therefore dared to push.

One test is failing (again), nevertheless this better takes this touched-account scenario in consideration, should be a relatively simple fix, then things should be working (hopefully).

Feel free to pick up.

holgerd77:
(phew. but hard stuff. 😕 )

holgerd77:
(maybe keep a local copy of your branch just in case, I am getting so confused by this stuff here)

holgerd77:
On this latest state the tests pass again, but Goerli run fails at:

image

holgerd77:
Yeah, yeah, yeah, got it. 🎉

image

Really spent another 3-4 hours debugging on this, would have never imagined.

What a journey, phew

ryanio (approved):
nice work, great, lgtm!

packages/blockchain/src/index.ts
```diff
@@ -498,71 +508,115 @@ export default class Blockchain implements BlockchainInterface {
...
+        const beneficiaryVotesDROP: Address[] = []
+        for (const vote of votes) {
+          const num = beneficiaryVotesDROP.filter((voteCMP) => {
+            return voteCMP.equals(vote[1][0])
+          }).length
+          if (num === 0) {
+            beneficiaryVotesDROP.push(vote[1][0])
+          }
+        }
+        let numBeneficiaryVotesDROP = beneficiaryVotesDROP.length
```
holgerd77, line 579: @ryanio thanks for review, wasn't so sure about here how to do this elegantly, with a direct additional `filter()` or `reduce()` call, if you know a better way feel free to rewrite on some subsequent PR (then we really make sure that it still syncs these first 6.000 or so Goerli blocks though, a bit trick. I was wondering if we should extract the case which was failing for us before the rewrite and add (propose) as an additional test case to the EIP. I am planning to do a PR with some suggestions for improvements and clarifications anyhow once we are stable (might be more or less: now, we'll see, likely though), have taken notes along the lines. If you have something to add feel free to send me. We can alternatively also fork the EIP repo to `ethereumjs` and do the PR together).

PR #1105, opened 2021-02-12
Blockchain: if passed header only validate header

ryanio:
This PR adds some logic to `blockchain._putBlockOrHeader` where if a header is passed in then it will only validate the header and not the entire block since the additional data needed would be missing (txs, uncles). This was discovered when trying to sync the client on rinkeby in light mode.


holgerd77 (approved):
LGTM

jochem-brouwer (approved):
Makes a lot of sense!

ryanio:
thanks!

PR #1112, opened 2021-02-15
Common: Add EventEmitter and hardforkChanged event

### LINKED ISSUE #1102: Extend Common with EventEmitter
We currently have no way to react if our Common updates to a new hard fork. For instance, the VM should update the opcodes, and in some cases transactions and blocks which were valid in a specific hardfork, are not valid anymore (#969). It would therefore make sense to extend Common with EventEmitter, so when we change the hard fork, then each individual component can react individually. 

Related discussion in #1101 .

holgerd77:
Fixes #1102


holgerd77:

I originally just wanted to add 1-2 tests for the opcode changes after the HF event is received, but when I looked into the existing tests I found this semi-exposed static `getOpcodesForHF(common: Common)` method from `evm/opcodes/codes.ts` so out of place that I added a `VM.getActiveOpcodes()` to the API which then directly calls with the correct VM common context, see b3177e1.

Then this can be used in a controlled way. I just checked, since I was wondering if the above static function is exposed to the docs: we actually DO expose all kinds of weird stuff to the [documentation](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/docs/README.md). We should relatively urgently give this an update here, otherwise people will start to use this more and more eventually which would not be too healthy.

ryanio (approved):
nice and simple, lgtm!

jochem-brouwer (commented):
Really clean and the tests added are great!

Issue #1107, opened 2021-02-14
Client, Devp2p: Stale behavior on DNS sync


holgerd77:
It still happens for me that client sync gets into some stale sync, here is a screenshot from Goerli with `peertablesize` staying completely static and no sync resuming for over 10 minutes:

![image](https://user-images.githubusercontent.com/931137/107876019-a91f2200-6ec3-11eb-9604-6a72e2f3b943.png)

This should be given some priority on improving the situation, since it prevents long client runs where one can hope that the client significantly progressed after going away for several hours (and therefore being able to sync a large chunks of blocks and progress towards the head of a chain).

It might very well be that my current slow internet connections contributes to the connection/sync generally being not so well. I think this is some kind of separate problem though, since a sync *does*  resume relatively quickly (within 0-2 minutes) when the client is stopped and restarted. So it should be possible to have a similar refresh behavior directly triggered by the client (along the `Retriggering bootstrap` call e.g..) if nothing helps any more.

holgerd77:
Addressed on a basic level with #1113 and not a totally apparent problem any more. If issues arise again or are still present, a more specific issue can be opened instead.

PR #1113, opened 2021-02-15
client -> restart rlpx server on peerpool re-bootstrap

cgewecke:
#1107 

This is a naive, stop-gap solution to the problem described in #1107. Its only virtue is that it "works"  (allows the client to continue to sync.) 

But am not happy :/  - that destroying and rebooting the server breaks the discovery deadlock suggests something is wrong.

Some notes:

+ Initially, I thought the peer banlist might be too inflexible for small networks and tested that hypothesis by resetting it in the re-triggered bootstrap. This had no effect. 

+ Some peers disconnect from us with USELESS PEER, after making a `GetBlockBodies` request that fails with `Key not found in database`. In principle, all the peers acquired via DNS should be serving the relevant network (like goerli) ... this should be looked into further. 
```
DEBUG [02-15|10:00:47] Peer added: id=8ec54986 address=163.172.89.89:30303 transport=rlpx protocols=eth inbound=false 
DEBUG [02-15|10:00:52] Error handling message (eth:GetBlockBodies): Key not found in database [Hw8�
                                                                                                   �豫�EŜ�P>Vҝ��s�[_C�]
INFO [02-15|10:00:53] Imported blocks count=50 number=233854 hash=2629f267... hardfork=petersburg peers=1 
```

+ Sometimes when the connections are fragile, the peer count drops to zero while the connection seems to still be live / importing blocks. Very intermittent. Possibly just reflects a lag between reporting block activity and network connection processing.

```
INFO [02-14|12:40:36] Imported blocks count=200 number=207110 hash=.... peers=0
```

Sorry this is a bit inconclusive at the moment - am hoping that this patch provides a way to debug things further. The connection dropping happens non-deterministically on a scale of minutes so ... it's tricky.


packages/client/lib/sync/fetcher/fetcher.ts
```diff
@@ -348,7 +348,7 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable
       this.config.logger.debug(
         `Task timed out for peer (banning) ${JSON.stringify(job.task)} ${job.peer}`
       )
-      this.pool.ban(job.peer!, 300000)
+      this.pool.ban(job.peer!, this.banTime)
```
cgewecke, line 351: Default banTime is `60000`

cgewecke:
(I've re-run all the jobs but the cov report is still wrong. There's no diff for the changes in this PR.)

cgewecke:
Was experimenting with this a bit more...
+ synced Ropsten from 0 to ~170k, mostly w/ 2 peers. 
+ then...had a sequence of `key not found` errors while handling `GetBlockBodies` reqs
+ on server reboot the same peers repeatedly disconnected with `USELESS_PEER`

On shutdown and restart, the same peers immediately connected and began serving blocks (for a while, until the above cycle began again)

ryanio:
> * then...had a sequence of `key not found` errors while handling `GetBlockBodies` reqs

that's interesting, so who is making these requests? the peers we are syncing with or other peers? sounds like we need to solve the problem here. at first thought we would catch the NotFound error and fail gracefully with an empty/null/error response to the peer.

> * peers dropped
> * on server reboot the same peers repeatedly disconnected with `USELESS_PEER`

that makes sense if it crashed on the GetBlockBodies request we would time out and get marked.

cgewecke:
> that makes sense if it crashed on the GetBlockBodies request we would time out and get marked.

Yes, agree - I think that's what's happening. Am currently trying debug via the `chain` module methods to see if I can figure out where the throw happens. If we don't have something in the DB I think it's fine, we just need to send back an empty array or whatever.

holgerd77 (approved):
LGTM, will merge (eventually extract other things from discussions along the thread into separate issue if still planned to be worked on or kept for reference).

Issue #1111, opened 2021-02-15
VM: allow API testing with different core dependency versions


holgerd77:
We currently have the problem of a very tight coupling of our monorepo dependencies (see dependency relationship [graph](https://github.com/ethereumjs/ethereumjs-monorepo#package-dependency-relationship) for an overview), so that a breaking major release update within one of the monorepo dependencies leads to necessary major release updates in all up-stream library dependants.

This becomes particularly apparent within the VM where most of the libraries are used in conjunction. To ease the coupling here we have discussed to allow for more than one major version of a library to be passed in as an API parameter as one potential solution (and then documented along the function signature).

The newly reworked tx library with the #1048 typed transaction integration (which requires a major tx release v4) is a good candidate for a first application here. To prepare for an integration a test setup should be integrated so that some/all/selected [API tests](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/vm/tests/api) can be run with a different dependency version, e.g. by providing a different `package.json` file or something.

In our specific case one test scenario is to run the test with the `@ethereumjs/tx` version 3 from [npm](https://www.npmjs.com/package/@ethereumjs/tx), the other scenario is to run - as it is done now - the library with the version from the `master` branch (respectively a feature branch) (not sure - open for alternative approaches - if it is e.g. significantly easier to run against a to be created `ethereumjs/tx/v3` branch or something).

This setup should prepare to execute a test against the [VM.runTx()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/lib/runTx.ts) function with on the two runs passing the two mentioned `tx` versions as a `tx` parameter to the function.

Old `v3` txs should then likely be wrapped early on by a `v4` tx with some `Tx.fromV3Transaction()` function (not necessarily part of this PR).

holgerd77:
We shifted away here gain and instead refocused on not too do breaking releases to often. There is a switch to peer dependencies planned #1200 which should bring some ease here.

Will close.

PR #1118, opened 2021-02-18
Add backwards-compatibility test fixture and run in CI for `tx`

cgewecke:
#1111

> a test setup should be integrated so that some/all/selected API tests can be run with a different dependency version, e.g. by providing a different package.json file or something.

This PR's strategy is to swap an npm published version of a sub-dependency  (ex: `tx`) into the monorepo, replacing the current branch's version. Then everything is re-linked and vm API tests are run with the new version. 

Have added a shell script that can execute this process generically for any sub-dependency/target pair. It restores the monorepo to its original state after executing so in principle it can be used locally but have added an env variable safety check to dissuade people from doing this accidentally. 

It should also be possible to execute repeatedly in CI in the `vm-backwards-compatibility` job for different subpackages (mpt, etc)

Usage:
```
$ ./scripts/check-compatibility.sh [options]

> OPTIONS (all are required)
> -t <test>       monorepo package whose tests we'd like to run (ex: vm)
> -d <dependency> monorepo package dependency we want to swap out (ex: tx)
> -n <npm>        npm package name to swap in (ex: @ethereumjs/tx)
> -v <version>    npm package version to swap in (ex: 3.0.2)
> -s <script>     script to "npm run" to execute the tests
```

The script flow for this case is 
+ Move `packages/tx` to a safe place (`stash/`)
+ Create a `tx` folder in the project root, cd in and npm install @ethereumjs/tx there
+ Move the contents of `tx/node_modules/@ethereumjs/tx` out 2 levels so that the folder looks like
```
tx/
  dist/
  dist.browser/
  node_modules/
  package.json
  etc...
```
+ Move `tx/` into `packages/`
+ Re-link everything
+ Run vm tests
+ Unspool all the above

### DIFF
--- a/.github/workflows/vm-pr.yml
+++ b/.github/workflows/vm-pr.yml
@@ -39,7 +39,7 @@ jobs:
         working-directory: ${{github.workspace}}
 
       - run: npm run coverage
-      
+
       - uses: codecov/codecov-action@v1
         with:
           file: ${{ env.cwd }}/coverage/lcov.info
@@ -48,7 +48,7 @@ jobs:
       - run: npm run test:API
       - run: npm run test:API:browser
       - run: npm run lint
-  
+
   vm-state:
     runs-on: ubuntu-latest
     strategy:
@@ -132,9 +132,9 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        # Args to pass to the tester. Note that some have splitted the slow tests and only 
+        # Args to pass to the tester. Note that some have splitted the slow tests and only
         #   running those: these are only on forks where that is applicable (see PR #489 for numbers on these)
-        
+
         # Tests were splitted with --dir and --excludeDir to balance execution times below the 9min mark.
         args: [
           '--fork=Istanbul --dir=GeneralStateTests/stTimeConsuming --expected-test-amount=15561',
@@ -174,9 +174,9 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        # Args to pass to the tester. Note that some have splitted the slow tests and only 
+        # Args to pass to the tester. Note that some have splitted the slow tests and only
         #   running those: these are only on forks where that is applicable (see PR #489 for numbers on these)
-        
+
         # Tests were splitted with --dir and --excludeDir to balance execution times below the 9min mark.
         args: [
           '--fork=Berlin --expected-test-amount=33',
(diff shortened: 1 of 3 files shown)


cgewecke:
In [this CI run][1] a "verification" log line is injected into the npm sourced package's Transaction constructor (before swapping) to demonstrate that the substitution takes place as expected.

Screen Shot 2021-02-17 at 6 55 56 PM

[1]: https://github.com/ethereumjs/ethereumjs-monorepo/pull/1118/checks?check_run_id=1923858793

holgerd77:
Great, generally stay in close sync with @jochem-brouwer on this, still not sure if this will be necessary on this round or if #1048 will stay generally backwards-compatible, but independently from that it's great to have this, we'll see if it comes to application directly.

One question I have is how to solve the TypeScript signature typing for txs elegantly so that `runTx(this: VM, opts: RunTxOpts)` don't complain on typing when passed in a v3 Tx, not sure about the design here, just to mention that this needs to be thought along as well.

cgewecke:
Rebased

cgewecke:
This test sort of works...

Screen Shot 2021-03-04 at 7 38 19 PM

Maybe `runTx.ts:102` should get an explicit cast to `EIP2930Transaction`?

```diff
-  if (opts.tx.transactionType === 1 && this._common.isActivatedEIP(2929)) {
+  if ((opts.tx as EIP2930Transaction).transactionType === 1 && this._common.isActivatedEIP(2929)) {
```


A more correct version of this test might look like the setup in #1134 and a simple mock project whose dependencies are:
```
@ethereumjs/tx: 3.0.2
@ethereumjs/vm: master
  |- @ethereumjs/tx: master
```
...with tests similar to `vm/test/api/runTx.ts` (but not actually run inside the vm package). 

Seems like the implementation is pretty non-breaking to me though...

holgerd77:
> This test sort of works...
> 
> Screen Shot 2021-03-04 at 7 38 19 PM
> 
> Maybe this line should get an explicit cast to `EIP2930Transaction`?

A cast wouldn't work here, since this actually *should* work without this being an `EIP2930Transaction`, but a simple property check should do on such occurrences, see comment here https://github.com/ethereumjs/ethereumjs-monorepo/pull/1048/files#r587391549.

cgewecke:
Am going to close here and pursue my suggestion for a more realistic test above as part of reviewing #1138. This test isn't designed correctly.

holgerd77:
Since we now already have two use cases for this (testing the tx-related VM methods, testing the `Block` constructor on passing in tx instances), maybe it is possible to design this in a way not tied too much to a specific package but in a relative generic way which would work for all the packages? 🤔

Not sure, just a thought, if it gets too complicated definitely also viable to copy-paste e.g. some GH actions code from one actions file to another or something.

PR #1119, opened 2021-02-18
VM: Only expose API method on docs

### LINKED ISSUE #1115: VM: only expose API methods on docs
We currently expose all kind of weird stuff in the VM [documentation](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/docs/README.md), this was found along the work on #1112. We should relatively urgently give this an update here, otherwise people will start to use this more and more eventually which would not be too healthy.

If someone works on this please give a short reasoning on every removal.

holgerd77:
Fixes #1115 

PR fixes the issue from above.

Small deviations from the suggestions from here https://github.com/ethereumjs/ethereumjs-monorepo/issues/1115#issuecomment-780183672:

- The `evm` folder has been completely removed recursively, there is actually nothing officially exposed
- I kept `state/index` to draw some attention to the information what is exposed on the directory level

Resulting docs look as following:

image

I find this direct exposure of the `runBlock.ts`,... files still not optimal, from an API PoV it would suffice to just expose the `lib/index.ts`, than there would be a clearer entrypoint to the API since all these `run*` methods are reexported (as well as documented) from there. Did some experimentation with removing the `lib/run*.ts` files from documentation, but this leads (similarly with the state manager) to the option types (e.g. `RunBlockOpts`) not being followable/clickable (this is actually still the case for the methods as they are displayed along the main `VM` module in the `lib/index.ts` docs), so information about these types is only accessible if going through the e.g. `lib/runBlock.ts` docs.

So still room for improvement (StateManager docs are also not very much cleaned up with a lot of private methhods in). This is nevertheless already a substantial improvement, would stop here for now.


packages/vm/docs/classes/_index_.vm.md
```diff
@@ -43,7 +43,7 @@ This class is an AsyncEventEmitter, please consult the README to learn how to us
 
 \+ **new VM**(`opts`: [VMOpts](../interfaces/_index_.vmopts.md)): *[VM](_index_.vm.md)*
 
-*Defined in [index.ts:139](https://github.com/ethereumjs/ethereumjs-vm/blob/master/packages/vm/lib/index.ts#L139)*
+*Defined in [index.ts:147](https://github.com/ethereumjs/ethereumjs-vm/blob/master/packages/vm/lib/index.ts#L147)*
```
ryanio, line 46: hm i thought the ethereumjs-vm links would be updated to ethereumjs-monorepo

ryanio (approved):
lgtm!

packages/vm/docs/classes/_index_.vm.md
holgerd77: 🤔

PR #1121, opened 2021-02-19
Update run-solidity-contract readme instructions

ryanio:
This PR updates the readme instructions for `run-solidity-contract`  removing the old `npm run build:dist` step since the monorepo packages are now automatically built post-install following `npm i` on root.


holgerd77 (approved):
LGTM

Issue #727, opened 2020-04-24
Common: Add isPrecompiled method


holgerd77:
The `isPrecompiled` method is currently [being removed](https://github.com/ethereumjs/ethereumjs-util/pull/242) from the `ethereumjs-util` library, since the library has no notion of a "hardfork" and the method is therefore pretty error-prone.

Instead `isPrecompiled` might be a better fit in the `Common` class and give out the respective answers depending on the HF set on the common instance.

Not sure if this is needed or if there is a strong demand, just drop this here for some eventual discussion and/or remembrance.

jochem-brouwer:
This should have been exposed in [this PR](https://github.com/ethereumjs/ethereumjs-vm/pull/783). It is trivial, I'll add it.

jochem-brouwer:
Actually it is not trivial where we should expose this function. Should we expose this on top of the VM? This only has things like `runCall`, `runBlock`, etc. Feels a bit weird to put this "utility" function there. It feels more natural to put it in the EVM, but the VM does not appear to have a pointer to the EVM? Where should we expose this function?

I think it is good to think about this, as we might get demand for some more of these utility-like functions. We need to know where to place those.

holgerd77:
Ah - attention! 😄  - this is an issue associated to the `Common` library, so "here" means "Common" in this case.

So this would be a utility function in `Common` to return the precompile status of an account based on some precompile info which would need to be added to the HF files.

Not sure if it's worth to be added right now tbh, would be nice to have but also some significant effort to integrate.

jochem-brouwer:
Oops - I didn't see that this was actually the Common package.

We have a slight problem here as currently the hardforks are hard-coded (hah!) in `index.ts` of [precompiles](https://github.com/ethereumjs/ethereumjs-vm/blob/55c4a1eb56e3439e2364e9ee067e8da0331de9f1/packages/vm/lib/evm/precompiles/index.ts#L43). It would make more sense to let the precompiles call into Common: `getPrecompile` actually has access to `Common` now so this should not be a problem.

I currently don't know how we should do this in a clean way. We could maybe create a map in common where we loop over each fork, latest first, and either return a precompile (I guess this should be an enum so we can support different precompiles at the same address if this happens at some point (?)) or we return that no precompile exists at this address? And then build the trivial `isPrecompiled` on top of that.

jochem-brouwer:
[...]

PR #1124, opened 2021-02-22
Fix EIP2929

### LINKED ISSUE #1123: EIP2929 problems
There's a problem with the current [EIP2929](https://eips.ethereum.org/EIPS/eip-2929) implementation. We currently track which addresses/slots are tracked on a message level (so each CALL creates a new message), while we should track it on transaction level.

Since EEI and Interpreter are re-instantiated each time we execute a new message, we need to do it at a higher level (e.g. EVM). The best place to me seems to place this in `StateManager`, we can track these things the same way as almost the same way as we track `touched` accounts. The only problem I have with this one is that we might bloat the StateManager a lot with the EIP2929 stuff.

Thoughts?

This is rather important, since EIP2930 especially depends on how these storage slots / addresses are tracked, so creating EIP2930 on the current EIP2929 implementation would end up that we do this work twice. So let's first at least figure out how we should fix EIP2929 and then proceed to implement EIP2930.

jochem-brouwer:
Closes #1123

The problem with the current implementation of EIP2929 is that the warm storage slots / addresses are only tracked per internal message, not on the entire transaction as the EIP implies. This PR moves the warm storage tracker from the Interpreter to the StateManager. This is done, because we also have to bookkeep if an internal message reverts. If that is the case, then extra added warm slots in these reverted internal messages, are now considered cold for the remainder of the transaction, instead of warm.

Since we explicitly revert/commit/checkpoint on StateManager, this is the logical place for this cache. 

I also added an extra rule here, which I don't recall having removed, which also adds any created contract to the warm address list.

However, there are some problems here: (which is why this PR is WIP)

(1) We also use this checkpoint/commit/revert mechanism when running a block. Currently, this means warm storages are tracked during the entire block, which we don't want. Fix? Maybe we can explicitly clear the slots at the end of a transaction, and at the start of a transaction we also require an empty starting point?
(2) There's also a problem with runCall/runCode: this does never checkpoint state manager, and thus the current cache is the empty list. The current logic actually expects that there is a Map/Set available. Fix: we can probably initialize the default value as the empty Map, this should work fine. If we combine this with the fix above this should be OK. **This is also the reason why the API tests file for EIP2929, since this uses runCode**.

(I have written this down to also sit down and figure out what my current problems are, but came along with some thoughts for fixes along the way, I think this be fine. We also use this "end of tx" logic for the touched accounts.)

Current tests added as API tests: 

- [x] Load a cold storage slot. Now call the current contract, reload the storage slot (it is now warm). (I should add the correct gas number here, but verified that this works)

API tests to add:

- [x] Call the current contract. Load the storage slot. Now revert this internal transaction. In the main message, re-load the storage slot. It should charge cold SLOAD cost twice.
- [x] Query an external account. Call current contract, query external account again (it is now warm). 
- [x] Call current contract. Query external account. Revert internal transaction. Query external account (charges cold costs twice).

These do of course not cover everything, but give a rough idea that the implementation is correct.


packages/vm/lib/evm/evm.ts
```diff
@@ -272,6 +273,10 @@ export default class EVM {
       }
     }
 
+    if (this._vm._common.eips().includes(2929)) {
+      this._state.addWarmAddress(message.to.buf)
```
jochem-brouwer, line 277: Note: this is in the wrong place, we should do this before the checkpoint in the new message (in case that the creation fails, then the address should /still/ be in the warm list)

packages/vm/tests/api/EIPs/eip-2929.spec.ts
```diff
@@ -211,4 +254,18 @@ tape('EIP 2929: gas cost tests', (t) => {
...
+  tape('ensure storage slots are tracked transaction-wide', async (t) => {
+    // load same storage slot twice (also in inner call)
+    await runCodeTest('60005460003415601357600080808080305AF15B00', 1, t)
+    // call to contract, load slot 0, revert inner call. load slot 0 in outer call.
+    //await runCodeTest('3415600B57600054600080FD5B600080808080305A60005400', 1, t)
+
+    // call to address 0xFFFF..FF
+    //const callFF = '6000808080806000195AF1'
+    //await runCodeTest(callFF + '60003415601C57600080808080305AF15B00', 1, t)
+    //await runCodeTest('3415601557' + callFF + '600080FD5B600080808080305A60005400', 1, t)
```
jochem-brouwer, line 267: These tests are WIP tests as described in message with the PR.

jochem-brouwer:
Small note, for #1048, the VM side of logic is now extremely easy, we can first `checkpoint`, and then just dump in the entire transaction list in the state manager (we mark them as warm).

holgerd77:
@jochem-brouwer great analysis and integration, thanks so much Jochem! 😄  👍 

Yeah, I think for (1) it should be ok to just clear after a tx? 🤔  Maybe just a generic public method `StateManager.clearWarmAccounts()` which is then called in `VM.runTx()` after tx execution?

Short side note: I would have a slight preference for `warmed` instead of `warm` for all the method names (so e.g. -> `addWarmedAccount()`, no strong opinions here though, just to note.

(2) Would think that the suggested solution should work on first thought. Will give additional comments here if I get some additional insights here.

packages/vm/lib/state/stateManager.ts
```diff
@@ -542,6 +575,71 @@ export default class DefaultStateManager implements StateManager {
...
+   */
+
+  /**
+   * Returns true if the address is warm in the current context
+   * @param address - The address (as a Buffer) to check
+   */
+  isWarmAddress(address: Buffer): boolean {
+    for (let i = this._accessedStorage.length; i >= 0; i--) {
+      const currentMap = this._accessedStorage[this._accessedStorage.length - 1]
+      if (currentMap.has(address.toString('hex'))) {
```
holgerd77, line 589: If I read the code correctly there seems to be no reason for this reverse order traversing (?) and this could then be simplified to `for (const currentMap of this._accessedStorage)`.

packages/vm/lib/state/stateManager.ts
```diff
@@ -542,6 +575,71 @@ export default class DefaultStateManager implements StateManager {
...
+   * Returns true if the slot of the address is warm
+   * @param address - The address (as a Buffer) to check
+   * @param slot - The slot (as a Buffer) to check
+   */
+  isWarmStorage(address: Buffer, slot: Buffer): boolean {
+    const addressKey = address.toString('hex')
+    const storageKey = slot.toString('hex')
+
+    for (let i = this._accessedStorage.length; i >= 0; i--) {
+      const currentMap = this._accessedStorage[this._accessedStorage.length - 1]
```
holgerd77: Same here, loop can be simplified.

packages/vm/lib/evm/evm.ts
holgerd77: There will likely be a reason why you nevertheless put this here along with giving a comment, but it is not obvious too me (too difficult for some reasons?). Could you clarify?

packages/vm/lib/runTx.ts
```diff
@@ -121,6 +122,21 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
   const caller = tx.getSenderAddress()
   debug(`New tx run hash=${opts.tx.hash().toString('hex')} sender=${caller.toString()}`)
 
+  if (block._common.eips().includes(2929)) {
+    // Add origin and precompiles to warm addresses
+    for (let addressInt = 1; addressInt <= numPrecompiles; addressInt++) {
+      // Check if precompile exists on the current Common of the Block
+      if (getPrecompile(new Address(Buffer.from(addressInt)), block._common)) {
+        state.addWarmAddress(new Address(Buffer.from(addressInt)))
```
holgerd77: This is not so beautiful to re-create this address space for the precompiles, we already have the `precompiles` const in `precompiles/index.ts` with all the addresses available, and this is also already exported.

Can you please use this here for address creation? Then there is also no need to do this extra `numPrecompiles` export.

packages/vm/lib/state/interface.ts
```diff
@@ -32,4 +32,8 @@ export interface StateManager {
   accountExists(address: Address): Promise<boolean>
   cleanupTouchedAccounts(): Promise<void>
   clearOriginalStorageCache(): void
+  addWarmAddress(address: Buffer): void
+  isWarmAddress(address: Buffer): boolean
+  addWarmStorage(address: Buffer, slot: Buffer): void
+  isWarmStorage(address: Buffer, slot: Buffer): boolean
```
holgerd77, line 38: These methods unfortunately can't be added here yet to be mandatory, since this would be a breaking change.

Just sitting here realizing that this gets - again - a bit trickier from the compatibility angle. Suggestion: we add these here as optional and then make it very clear in the code docs /release notes, that - if people want to run EIP2929 (and likely also 2930) functionality AND have their custom state manager implementation - they need to implement these methods.

Ug. Likely also always (?) a good advice for people to do their custom implementations by inheriting from the default state manager. Then this has a higher chance that a custom state manager will continue to work on such kind of updates.

packages/vm/lib/evm/opcodes/EIP2929.ts
```diff
@@ -78,13 +72,12 @@ export function adjustSstoreGasEIP2929(
...
 
-  const keyStr = key.toString('hex')
-  const address = runState.eei.getAddress().toString()
+  const address = runState.eei.getAddress().buf
   const warmRead = runState._common.param('gasPrices', 'warmstorageread')
   const coldSload = runState._common.param('gasPrices', 'coldsload')
 
   // @ts-ignore Set Object is possibly 'undefined'
-  if (runState.accessedStorage.has(address) && runState.accessedStorage.get(address).has(keyStr)) {
+  if (runState.stateManager.isWarmStorage(address, key)) {
```
holgerd77, line 80: I think the @ts-ignore line from above can be removed.

holgerd77 (commented):
Really nice and clean process of doing this transition here! 👍  Have left some comments.

packages/vm/tests/api/EIPs/eip-2929.spec.ts
```diff
@@ -55,6 +56,48 @@ tape('EIP 2929: gas cost tests', (t) => {
...
+    const account = await vm.stateManager.getAccount(address)
+    await vm.stateManager.putAccount(
+      address,
+      Account.fromAccountData({ ...account, balance: initialBalance })
+    )
+
+    const result = await vm.runTx({ tx })
+
+    st.ok(result.gasUsed.toNumber() == expectedGasUsed)
+  }
```
holgerd77, line 99: That's a cool test helper function here. 😄

packages/vm/lib/state/stateManager.ts
jochem-brouwer: I did this in reverse order since this is an optimization. Imagine that we are rather deep in our call stack, say 10 levels deep. If we now access a storage slot on the current address (assuming  that this is the first time we are at this address) then we actually mark the slot as warm in the end of the array. If we traverse the array in normal order, we first have to check the 9 other maps before figuring out that the 10th map (at current address) has marked it as warm. (It is a small optimization)

packages/vm/lib/runTx.ts
jochem-brouwer: Should have left a comment here. This is for forward compatibility. Geths implementation only marks current active precompiles to be warm. If we do not implement something like I created here (I agree it is not beautiful currently) then we get a consensus bug, for instance if we call a BLS precompile. Geth would mark it as cold, while we mark it as warm address. (GetPrecompile checks if precompile is available at current HF)

packages/vm/lib/evm/evm.ts
jochem-brouwer: Yeah sorry this is a bit weird. I realized somewhere around pushing that this might be wrong, checked it, and realized that it is indeed wrong. It is more of a reminder comment that I should fix it :)

packages/vm/lib/state/interface.ts
jochem-brouwer: Yep you are right. I only fear that if I remove this, then TypeScript is going to complain about possible non existent methods...

packages/vm/lib/runTx.ts
holgerd77: Ah ok, thanks. Can you then either:

a) add an additional `getPrecompiles(common: Comon)` function to `precompiles/index.ts` just looping over `precompiles` and return the active ones analog to `getPrecompile()? or

b) minimal solution: add least do the `numPrecompiles` count dynamically by doing a `precompiles.length + 1` in `precompiles/index.ts`, so that we do not introduce a relatively error prone new static constant there

This whole logic really doesn't belong there and we should at some point move to `Common` (not now though likely).

packages/vm/lib/evm/evm.ts
holgerd77: 😄

packages/vm/lib/state/interface.ts
holgerd77: Yeah, that's what we have to live with and what the `!` operator is for. Unfortunately no way around that we just can't do such an interface change along the road without a major VM release. 😕

packages/vm/lib/runTx.ts
jochem-brouwer: I will go for option (a). #727 is related here

packages/vm/lib/state/interface.ts
jochem-brouwer: Sadly we cannot use the ! on methods. (I tried). Now we have to resort to explicit casting or casting the state manager as any =(

jochem-brouwer:
I think I resolved all issues above. Am pretty confident in the implementation currently. 

Just some small remarks:

(1) Before we join YoloV3, it makes sense to at least get all Berlin tests passing. The basics are now covered by some API tests, but I'm certain that I've missed some edge cases here.
(2) I had to convert the original EIP2929 tests to use `runTx` instead of `runCode`. This made me realize that `runCode` and `runCall` are not very consistent with other parts of the package. For instance, `runCode` doesn't do any checkpointing, and therefore the entire EIP2929 logic does not work. It makes sense to either deprecate both methods, or to internally wrap these things to use `runTx` internally (I think we should do the latter). With slightly fancy tricks this should not be too hard to do.

jochem-brouwer:
PR run fails but should be easy fix (will do ASAP): 

```
2021-02-22T15:10:23.8023909Z not ok 486 should fail with appropriate error
2021-02-22T15:10:23.8025394Z   ---
2021-02-22T15:10:23.8026181Z     operator: ok
2021-02-22T15:10:23.8027047Z     expected: true
2021-02-22T15:10:23.8028772Z     actual:   false
2021-02-22T15:10:23.8030461Z     at: <anonymous> (/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/packages/vm/tests/api/runTx.spec.ts:36:10)
2021-02-22T15:10:23.8031905Z     stack: |-
2021-02-22T15:10:23.8032958Z       Error: should fail with appropriate error
2021-02-22T15:10:23.8095393Z           at Test.assert [as _assert] (/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/node_modules/tape/lib/test.js:228:54)
2021-02-22T15:10:23.8098744Z           at Test.bound [as _assert] (/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/node_modules/tape/lib/test.js:80:32)
2021-02-22T15:10:23.8100543Z           at Test.assert (/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/node_modules/tape/lib/test.js:347:10)
2021-02-22T15:10:23.8102057Z           at Test.bound [as ok] (/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/node_modules/tape/lib/test.js:80:32)
2021-02-22T15:10:23.8103540Z           at /home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/packages/vm/tests/api/runTx.spec.ts:36:10
2021-02-22T15:10:23.8104409Z           at runMicrotasks (<anonymous>)
2021-02-22T15:10:23.8105187Z           at processTicksAndRejections (internal/process/task_queues.js:97:5)
2021-02-22T15:10:23.8106043Z           at runNextTicks (internal/process/task_queues.js:66:3)
2021-02-22T15:10:23.8106740Z           at processImmediate (internal/timers.js:434:9)
2021-02-22T15:10:23.8107230Z   ...
2021-02-22T15:10:23.8107723Z not ok 487 error should include "enough funds"
2021-02-22T15:10:23.8108322Z   ---
2021-02-22T15:10:23.8108671Z     operator: ok
2021-02-22T15:10:23.8109076Z     expected: true
2021-02-22T15:10:23.8109464Z     actual:   false
2021-02-22T15:10:23.8110600Z     at: <anonymous> (/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/packages/vm/tests/api/runTx.spec.ts:44:10)
2021-02-22T15:10:23.8111534Z     stack: |-
2021-02-22T15:10:23.8112012Z       Error: error should include "enough funds"
2021-02-22T15:10:23.8113256Z           at Test.assert [as _assert] (/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/node_modules/tape/lib/test.js:228:54)
2021-02-22T15:10:23.8114822Z           at Test.bound [as _assert] (/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/node_modules/tape/lib/test.js:80:32)
2021-02-22T15:10:23.8116356Z           at Test.assert (/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/node_modules/tape/lib/test.js:347:10)
2021-02-22T15:10:23.8117867Z           at Test.bound [as ok] (/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/node_modules/tape/lib/test.js:80:32)
2021-02-22T15:10:23.8119363Z           at /home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/packages/vm/tests/api/runTx.spec.ts:44:10
2021-02-22T15:10:23.8120252Z           at runMicrotasks (<anonymous>)
2021-02-22T15:10:23.8121034Z           at processTicksAndRejections (internal/process/task_queues.js:97:5)
2021-02-22T15:10:23.8121899Z           at runNextTicks (internal/process/task_queues.js:66:3)
2021-02-22T15:10:23.8122962Z           at processImmediate (internal/timers.js:434:9)
2021-02-22T15:10:23.8123449Z   ...
```

packages/vm/lib/evm/opcodes/EIP2929.ts
```diff
@@ -44,11 +46,12 @@ export function accessStorageEIP2929(runState: RunState, key: Buffer, isSstore:
   const baseFee = !isSstore ? runState._common.param('gasPrices', 'sload') : 0
   const address = runState.eei.getAddress().buf
 
-  const slotIsCold = !runState.stateManager.isWarmStorage(address, key)
+  const slotIsCold = !(<EIP2929StateManager>runState.stateManager).isWarmedStorage(address, key)
```
holgerd77, line 49: Nice and thought-through solution, cool! 👍

packages/vm/tests/api/EIPs/eip-2929.spec.ts
```diff
@@ -263,16 +259,34 @@ tape('EIP 2929: gas cost tests', (t) => {
...
-    //await runCodeTest('3415600B57600054600080FD5B600080808080305A60005400', 1, t)
+    await runCodeTest('341515600D57600054600080FD5B600080808080305AF160005400', 25374, t)
 
     // call to address 0xFFFF..FF
-    //const callFF = '6000808080806000195AF1'
-    //await runCodeTest(callFF + '60003415601C57600080808080305AF15B00', 1, t)
-    //await runCodeTest('3415601557' + callFF + '600080FD5B600080808080305A60005400', 1, t)
+    const callFF = '6000808080806000195AF1'
+    // call address 0xFF..FF, now call same contract again, call 0xFF..FF again (it is now warm)
+    await runCodeTest(callFF + '60003415601B57600080808080305AF15B00', 23909, t)
```
holgerd77, line 283: Out of interest: how did you generate these bytcode sequences? 🤔

holgerd77:
@jochem-brouwer Have updated this branch since I needed the latest changes from #1125 for some experimentation, hope that's ok, don't forget to pull the branch changes (and if you didn't and have already done some substantial amount of work: just force-push, maybe announce here before though).

jochem-brouwer:
Hi @holgerd77 , have not done extra work. Feel free to review this one, I dont think that fixing the tests will require big changes.

packages/vm/tests/api/EIPs/eip-2929.spec.ts
jochem-brouwer: One of my (I guess) somewhat esoteric hobbies is writing raw EVM assembly. I do this from time to time, and getting more "fluent" in raw EVM 😄  It is very cool to figure out how you can create a minimal implementation which does what it should do. The [minimal proxy contract](https://eips.ethereum.org/EIPS/eip-1167), for instance, is also very interesting. As a side note: I hope that at some point we get a new [gas golf contest](https://g.solidity.cc/). They allow raw assembly too 😄

jochem-brouwer:
Will mark this as ready when I've fixed the tests. (They should all pass, since 2718/2930 is not yet in the suite)

holgerd77:
Just running another test case which is a bit more fitting: `DEBUG=vm:*,vm:*:*,-vm:ops:* ts-node ./tests/tester --state --test='callcallcallcode_001' --fork=Berlin` (you can of course also leave the `DEBUG` stuff).

I am pretty sure that this is just some basic error in the gas calculation in the `opcodes/EIP2929.ts` methods, since gas costs should differ and they do not atm.

jochem-brouwer:
Which gas cost difference are we talking about here? If I checked against Berlin and Istanbul, there's a difference.

jochem-brouwer:
Got another one which has to do with DELEGATECALL.

jochem-brouwer:
I will just push these small fixes, will squash them down the road.

holgerd77:
Ah, yes, you are right, at least in the test mentioned above there is a gas cost difference.

holgerd77:
Don't want to push here concurrently, but this  `berlin` reference in the test runner [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/tests/GeneralStateTestsRunner.ts#L65) with this manual EIP-2537 injection might also affect the outcome, maybe you can remove.

jochem-brouwer:
Ah yes I should remove it anyways, it might indeed interfere since it marks BLS precompiles as warm. Should be cold.

I found something related to SSTORE gas which is in EIP but we did not implement it. So this one might be huge.

packages/vm/tests/api/EIPs/eip-2929.spec.ts
holgerd77: Lol, I absolutely need more hobbies I can brag about

Maybe you can give a course at some PiT 😁  (far from fluent assembly writer)

jochem-brouwer: Oh yeah - that sounds fun 😄 , I can definitely prepare something at some PiT 👍

jochem-brouwer:
Ok, this took a bit longer than expected. (This EIP is also much more complex than anticipated)

Test I used was a (at first glance) simple looking one; [sstore_gasLeft_d2g0v0_Berlin](https://github.com/ethereum/tests/blob/develop/BlockchainTests/GeneralStateTests/stSStoreTest/sstore_gasLeft.json). We charge, in most cases, wrong gas for SSTOREs, which is also why so many tests fail. 

Currently SSTORE code looks a bit weird; we first charge for SSTORE gas, and then "access" the slot in EIP2929. This is because we have the sentry gas check in EIP2200: it checks if at least 2300 gas is available; if not, then go OOG. If we first deduct the cold SLOAD gas (2100 gas), then while we might have something like 3000 gas available, then at the sentry check we go OOG, which should not be the case.

Also a note from EIP2929 (from "SSTORE changes" section):  

> Note: The constant `SLOAD_GAS` is used in several places in EIP 2200, e.g `SSTORE_SET_GAS - SLOAD_GAS`. Implementations that are using composite definitions have to ensure to update those definitions too.

We pre-calculated these values in common, so we should re-calculate them for Berlin.

jochem-brouwer:
> Don't want to push here concurrently, but this `berlin` reference in the test runner [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/tests/GeneralStateTestsRunner.ts#L65) with this manual EIP-2537 injection might also affect the outcome, maybe you can remove.

Ah, sorry, I thought this EIP was the BLS one, it is actually one currently included in Berlin. I have removed it anyways, since it is already in the Common of Berlin (so the whole operation should be a no-op).

Cool, failing tests down from 3621 to just 40! 💥 🎉

jochem-brouwer:
Okay, gave this some more look, but seems I'm stuck here (or I'm too tired).

<details closed>
<summary>Here is a list of failing tests</summary>
2021-02-22T23:21:21.6556433Z Errors thrown in file: JUMPDEST_Attack test: JUMPDEST_Attack_d0g0v0_Berlin:
2021-02-22T23:21:21.6557820Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6558907Z 	correct last header block
2021-02-22T23:21:21.6560601Z Errors thrown in file: JUMPDEST_AttackwithJump test: JUMPDEST_AttackwithJump_d0g0v0_Berlin:
2021-02-22T23:21:21.6562116Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6563187Z 	correct last header block
2021-02-22T23:21:21.6565237Z Errors thrown in file: static_ABAcalls0 test: static_ABAcalls0_d1g0v0_Berlin:
2021-02-22T23:21:21.6566648Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6567718Z 	correct last header block
2021-02-22T23:21:21.6569170Z Errors thrown in file: static_ABAcalls2 test: static_ABAcalls2_d1g0v0_Berlin:
2021-02-22T23:21:21.6572168Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6574719Z 	correct last header block
2021-02-22T23:21:21.6576877Z Errors thrown in file: static_ABAcalls3 test: static_ABAcalls3_d1g0v0_Berlin:
2021-02-22T23:21:21.6577953Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6578745Z 	correct last header block
2021-02-22T23:21:21.6580231Z Errors thrown in file: static_CallRecursiveBomb0 test: static_CallRecursiveBomb0_d0g0v0_Berlin:
2021-02-22T23:21:21.6581527Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6582320Z 	correct last header block
2021-02-22T23:21:21.6583564Z Errors thrown in file: static_callCreate3 test: static_callCreate3_d0g0v0_Berlin:
2021-02-22T23:21:21.6584670Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6585469Z 	correct last header block
2021-02-22T23:21:21.6587077Z Errors thrown in file: static_callcallcodecall_ABCB_RECURSIVE test: static_callcallcodecall_ABCB_RECURSIVE_d0g0v0_Berlin:
2021-02-22T23:21:21.6588477Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6589276Z 	correct last header block
2021-02-22T23:21:21.6590908Z Errors thrown in file: static_callcallcodecall_ABCB_RECURSIVE2 test: static_callcallcodecall_ABCB_RECURSIVE2_d0g0v0_Berlin:
2021-02-22T23:21:21.6592318Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6593097Z 	correct last header block
2021-02-22T23:21:21.6594874Z Errors thrown in file: static_callcallcodecallcode_ABCB_RECURSIVE test: static_callcallcodecallcode_ABCB_RECURSIVE_d0g0v0_Berlin:
2021-02-22T23:21:21.6596376Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6597161Z 	correct last header block
2021-02-22T23:21:21.6598967Z Errors thrown in file: static_callcallcodecallcode_ABCB_RECURSIVE2 test: static_callcallcodecallcode_ABCB_RECURSIVE2_d0g0v0_Berlin:
2021-02-22T23:21:21.6600560Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6601348Z 	correct last header block
2021-02-22T23:21:21.6602952Z Errors thrown in file: static_callcodecallcall_ABCB_RECURSIVE test: static_callcodecallcall_ABCB_RECURSIVE_d0g0v0_Berlin:
2021-02-22T23:21:21.6605738Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6611081Z 	correct last header block
2021-02-22T23:21:21.6612767Z Errors thrown in file: static_callcodecallcall_ABCB_RECURSIVE2 test: static_callcodecallcall_ABCB_RECURSIVE2_d0g0v0_Berlin:
2021-02-22T23:21:21.6614407Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6615184Z 	correct last header block
2021-02-22T23:21:21.6616837Z Errors thrown in file: static_callcodecallcall_ABCB_RECURSIVE2 test: static_callcodecallcall_ABCB_RECURSIVE2_d0g0v1_Berlin:
2021-02-22T23:21:21.6618238Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6619022Z 	correct last header block
2021-02-22T23:21:21.6620791Z Errors thrown in file: static_callcodecallcallcode_ABCB_RECURSIVE test: static_callcodecallcallcode_ABCB_RECURSIVE_d0g0v0_Berlin:
2021-02-22T23:21:21.6622324Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6623112Z 	correct last header block
2021-02-22T23:21:21.6624892Z Errors thrown in file: static_callcodecallcallcode_ABCB_RECURSIVE2 test: static_callcodecallcallcode_ABCB_RECURSIVE2_d0g0v0_Berlin:
2021-02-22T23:21:21.6626432Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6627203Z 	correct last header block
2021-02-22T23:21:21.6629013Z Errors thrown in file: static_callcodecallcallcode_ABCB_RECURSIVE2 test: static_callcodecallcallcode_ABCB_RECURSIVE2_d0g0v1_Berlin:
2021-02-22T23:21:21.6630544Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6631328Z 	correct last header block
2021-02-22T23:21:21.6633090Z Errors thrown in file: static_callcodecallcodecall_ABCB_RECURSIVE test: static_callcodecallcodecall_ABCB_RECURSIVE_d0g0v0_Berlin:
2021-02-22T23:21:21.6634589Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6635377Z 	correct last header block
2021-02-22T23:21:21.6637258Z Errors thrown in file: static_callcodecallcodecall_ABCB_RECURSIVE2 test: static_callcodecallcodecall_ABCB_RECURSIVE2_d0g0v0_Berlin:
2021-02-22T23:21:21.6638801Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6639571Z 	correct last header block
2021-02-22T23:21:21.6641364Z Errors thrown in file: static_callcodecallcodecall_ABCB_RECURSIVE2 test: static_callcodecallcodecall_ABCB_RECURSIVE2_d0g0v1_Berlin:
2021-02-22T23:21:21.6642906Z 	Error: invalid receiptTrie
2021-02-22T23:21:21.6643941Z 	correct last header block
</details>

I just started at the top, [JUMPDEST_Attack](https://github.com/ethereum/tests/blob/develop/BlockchainTests/GeneralStateTests/stSpecialTest/JUMPDEST_Attack.json). It's a simple test, there's a contract which simply keeps calling itself, and then stops (it's also filled with a lot of JUMPDESTS for some reason, but these can never be reached). We charge 65866 gas, while the test expects 65850 gas (you can check this via the files: check the coinbase of the block. After the transaction it has 2 eth + 65850 wei. The single transaction has a gas price of `1`, so when we subtract the coinbase reward (2 eth) we get the gas used, which is 65850). There's thus a 16 gas difference. I have no idea where this comes from. The only thing I could think of is setting the "max call gas" in the CALL opcode to the wrong value, so I moved the access address method after the maxCallGas, but this did not change anything. We might want to get a trace from retesteth to figure out what is wrong. Just based on the names, the other failing tests also seem rather complex, so there's probably some edge case going wrong here. 

Will stop here.

holgerd77:
I will likely push to this branch within the next 1-2 hours, let me know if you are working in parallel here.

holgerd77:
If I run the test in the logger with `DEBUG=vm:*,vm:*:*,-vm:ops:* ts-node ./tests/tester --blockchain --file='JUMPDEST_Attack' --fork=Berlin` I get totally other gas values than the ones you cited above:

image

So `251671`. Not sure, are we talking about different tests?

Anyhow, I'll likely not dig deeper into this specific test.

holgerd77:
I thought I could serve you best by adding a specific gas logger in 2e7c765. 😄  This can be used like the following `DEBUG=vm:*:gas ts-node ./tests/tester --blockchain --file='chainId' --fork=Berlin` and provides output like the following:

image

This works also for refunds and the gas specifics of different EIPs (EIP-1283, EIP-2200, EIP-2929).

jochem-brouwer:
Did you recompile Common? Because it looks like you are on Istanbuls gas schedule on your picture. 

Thanks for the debug stuff! 😄  I think I have to get a trace of this transaction, since I doubt I'm gonna figure out whats going wrong by looking at these individual gas costs. But I will try anyways.

holgerd77:
Yeah, actually I am not sure if this will work in this specific case but over time it should be helpful nevertheless on debugging gas related stuff.

Second yeah, you're right, I've forgotten to re-bootstrap after merging the release PR into `master` and doing the releases. 😜

holgerd77:
Currently running the following test `DEBUG=vm:*:gas ts-node ./tests/tester --blockchain --file='static_callCreate3' --fork=Berlin`. There the `warmstorageread` gas usage is actually getting negative (?):

image

Can this be right?

I generally have to say that I am not understanding the role of this `basefee` yet passed to e.g. `accessAddressEIP2929()` and than used in code parts like:

```typescript
} else if (baseFee !== undefined && baseFee > 0) {
    console.log(runState._common.param('gasPrices', 'warmstorageread'))
    console.log(baseFee)
    runState.eei.useGas(
      new BN(runState._common.param('gasPrices', 'warmstorageread') - baseFee),
      'EIP-2929 -> warmstorageread'
    )
  }
```

Why is this subtracted at all for `useGas()`? The EIP is just stating:

> Otherwise, charge WARM_STORAGE_READ_COST gas.

🤔

holgerd77:
Update: ah, just seeing from the screenshot that this is double-applied and first added and then subtracted again, likely for reasons of implementation simplification, but actually a bit for the cost of readability. Hmm. And this might have side effects to due to these steps happening in two separate PiTs in the control flow.

holgerd77:
The test from above actually also has such a slightly different gasUsage:

Actual: `137237`
Expected: `137217`

So just a difference of `20` (this trick with the `coinbase` account is actually gold. 😁 )

jochem-brouwer:
Hi @holgerd77, I also had to do a double take here when I saw that it is negative, but this is correct. In total, for a CALL to a warm account, we charge the WARM gas (100 gas). Internally, what we do is we take the warm gas and the base call gas value (700), and subtract these. So we get -600. Later on (or earlier, but doesnt really matter) we charge the 700 gas, so net we charge 100 gas, which is correct according to the EIP.

jochem-brouwer:
I am currently setting up retesteth and geth to get a trace of this transaction, but geth seems to have changed how we should start it in "retesteth" mode, and I cannot find any docs how I should run it 😢

holgerd77:
I think if we do this ping-pong "first subtract the base fee, then add it again" thing we might operate on the wrong values for things like `maxCallGas`. I think we should likely do this properly and simply exclude the respective CALL* and other opcodes from the basefee directly in `runStep()` in the interpreter.

jochem-brouwer:
I'm not really sure if I like adding these exceptions to the base fee. You are right that it /might/ mess up maxCallGas, but since this operation is done before calling maxCallGas it is correct. I think this approach is elegant.

jochem-brouwer:
(Or well, I am assuming it is correct - might also not be the case of course since this could be the culprit 😅  )

holgerd77:
I have the impression that is no exception or something but just the direct implementation of the EIP (there is just no basefee for these opcodes, right? Correct me if I am wrong).

This otherwise doesn't play well with this specification from the EIP:

> Note: For call-variants, the 100/2600 cost is applied immediately (exactly like how 700 was charged before this EIP), i.e: before calculating the 63/64ths available for entering the call.

And for example for the CREATE opcode the code looks like this:

```typescript
// '0xf0' range - closures
  // 0xf0: CREATE
  [
    0xf0,
    async function (runState: RunState) {
      if (runState.eei.isStatic()) {
        trap(ERROR.STATIC_STATE_CHANGE)
      }

      const [value, offset, length] = runState.stack.popN(3)

      subMemUsage(runState, offset, length)
      let gasLimit = new BN(runState.eei.getGasLeft())
      gasLimit = maxCallGas(gasLimit, runState.eei.getGasLeft(), runState)

      let data = Buffer.alloc(0)
      if (!length.isZero()) {
        data = runState.memory.read(offset.toNumber(), length.toNumber())
      }

      accessAddressEIP2929(runState, runState.eei.getAddress())
      const ret = await runState.eei.create(gasLimit, value, data)
      runState.stack.push(ret)
    },
  ]
```

This seems to be wrong? 🤔  Or is it just enough to switch the order here?

jochem-brouwer:
This indeed looks wrong. Weird that tests do not pick this up (it seems that it is not picked up).

jochem-brouwer:
> I think if we do this ping-pong "first subtract the base fee, then add it again" thing we might operate on the wrong values for things like `maxCallGas`. I think we should likely do this properly and simply exclude the respective CALL* and other opcodes from the basefee directly in `runStep()` in the interpreter.

What we could maybe do here is not explicitly add these exceptions in code, but just mark these methods to use 0 gas in Common, and handle the base fee in the functions themselves?

holgerd77:
> > I think if we do this ping-pong "first subtract the base fee, then add it again" thing we might operate on the wrong values for things like `maxCallGas`. I think we should likely do this properly and simply exclude the respective CALL* and other opcodes from the basefee directly in `runStep()` in the interpreter.
> 
> What we could maybe do here is not explicitly add these exceptions in code, but just mark these methods to use 0 gas in Common, and handle the base fee in the functions themselves?

👍  That sounds like a good way to handle it.

jochem-brouwer:
Yup - found the bug with help of the [trace](http://retesteth.ethdevops.io/web/index.php?gtests=stSpecialTest&singletest=JUMPDEST_Attack&clients=t8ntool&snetwork=Berlin&vmtrace=vmtrace&verbosity=none&trd=&trg=&trv=)! 

It has indeed to do with the base gas! At some point, if we have less than 700 gas (base call fee) then we go OOG when we deduct the fee. However, we actually charge 100 gas, so we should not go out of gas. 

I will fix this with the method described above (base gas to 0). 

In retrospect, I don't know why I haven't realized this earlier..!

jochem-brouwer:
Berlin tests pass, have just added MuirGlacier to CI to ensure that it still works. Ready for review!

holgerd77:
@jochem-brouwer yeah, thanks for going so straightly through this. Again - as I wrote in the other thread - really take some moment if you need to rebalance your time, these kind of in-between tasks just do large re-shuffles one just can't really plan in. 🙂

packages/vm/lib/evm/opcodes/EIP2929.ts
```diff
@@ -22,16 +27,16 @@ export function accessAddressEIP2929(runState: RunState, address: Address, baseF
...
+    if (chargeGas) {
       runState.eei.useGas(
-        new BN(runState._common.param('gasPrices', 'coldaccountaccess') - baseFee),
+        new BN(runState._common.param('gasPrices', 'coldaccountaccess')),
         'EIP-2929 -> coldaccountaccess'
       )
     }
     // Warm: (selfdestruct beneficiary address reads are not charged when warm)
-  } else if (baseFee !== undefined && baseFee > 0) {
+  } else if (chargeGas && !isSelfdestruct) {
```
holgerd77, line 37: That's a lot better than these implicit assumptions from before! 👍

packages/vm/lib/evm/opcodes/functions.ts
```diff
@@ -940,6 +940,8 @@ export const handlers: Map<number, OpHandler> = new Map([
 
       const [value, offset, length] = runState.stack.popN(3)
 
+      accessAddressEIP2929(runState, runState.eei.getAddress(), false)
+
```
holgerd77, line 944: Just as some post-review question: is the order with `subMemUsage()` relevant here? Since in all the other cases this is the other way around?

holgerd77 (approved):
So great, all checks pass, congrats! 😄 

LGTM

Feel free to merge once this fits for you.

packages/vm/lib/evm/opcodes/functions.ts
jochem-brouwer: Good question, and also a bit sloppy of me. But I just explicitly checked; the memory gas cost is deterministic and does not depend on the current gas available, so the order does not matter. I don't really want to trigger a new CI run, so I will merge here.

jochem-brouwer:
Thanks for the quick review @holgerd77 😄

PR #1126, opened 2021-02-22
VM, Common: updated EIP checks / added EIP-2565, EIP-2929 to Berlin HF

holgerd77:
This PR is to get an in-between overview on where we are with the berlin test runs - still excluding EIP-2718 and EIP-2930 - PR is branched off from #1124.

@jochem-brouwer feel free to merge this one in if useful or we can otherwise merge post #1124 merge, as you like.

### DIFF
--- a/.github/workflows/vm-pr.yml
+++ b/.github/workflows/vm-pr.yml
@@ -54,7 +54,7 @@ jobs:
     strategy:
       matrix:
         fork: [
-          'MuirGlacier',
+          'Berlin',
           'Istanbul'
         ]
       fail-fast: false
@@ -137,7 +137,9 @@ jobs:
         # Tests were splitted with --dir and --excludeDir to balance execution times below the 9min mark.
         args: [
           '--fork=Istanbul --dir=GeneralStateTests/stTimeConsuming --expected-test-amount=15561',
-          '--fork=Istanbul --excludeDir=stTimeConsuming --expected-test-amount=19817'
+          '--fork=Istanbul --excludeDir=stTimeConsuming --expected-test-amount=19817',
+          '--fork=Berlin --dir=GeneralStateTests/stTimeConsuming',
+          '--fork=Berlin --excludeDir=stTimeConsuming'
         ]
       fail-fast: false
     steps:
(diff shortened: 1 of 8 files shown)


jochem-brouwer:
Very disappointing to see that there are so many tests failing. Will merge.

jochem-brouwer:
Thanks Holger! 😄

holgerd77:
I would assume that there is something still very basic wrong, maybe even just the HF setup, no need to panic or something (yet, hehe 😜 ).

We'll see.

holgerd77:
I did an isolated run of a single test with `DEBUG=vm:*,vm:*:*,-vm:ops:* ts-node ./tests/tester --state --test='invalidAddr' --data=0 --gas=0 --value=0 --fork=Berlin` and compared the results from `Berlin` (failing)  and `Istanbul` (passing).

State root is expected to be different in the test file but when run through the VM state root (as well as the whole execution flow) keeps the same both run on `Istanbul` and `Berlin`.

Not sure what EIP is expected to change the state root here? I've checked, this is not yet in the new transaction format.

jochem-brouwer:
What does this do again? It seems that this only runs tests where data/gas/value is 0? (Maybe this filter is implemented wrong).

I just found a bug (which is probably unrelated). Let's continue this discussion on #1124. Am fixing bugs on a per-test basis.

holgerd77:
Just give a short answer here to one aspect: don't be confused by this data/gas/value selection, this is just selecting the specific test case, nothing else. So if there is a section like this in the test file:

```json
{
                    "indexes" : {
                        "data" : 0,
                        "gas" : 0,
                        "value" : 0
                    },
                    "hash" : "0x012502a75bf4fcefbef2850b492f038ed9863832c6a09e84fc1bce079e33a6f5",
                    "logs" : "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347"
                }
```

These values are just indexes and pointing to the first (0-index) data value in the "transaction" section. So this just a way to run an isolated test case if there are many in one file, nothing else.

jochem-brouwer:
Ah that's interesting, I didn't know what these were in ethereum/tests. Thanks for clarifying. I was actually wondering if we had this feature 😄

PR #1130, opened 2021-02-27
Client: Add 6 new RPC methods

ryanio:
This PR adds new RPC methods:

1. `eth_getTransactionCount`
1. `eth_getBalance`
1. `eth_getCode`
1. `eth_getStorageAt`
1. `eth_call`
1. `eth_estimateGas`

For `eth_call` and `eth_estimateGas`, the code uses a copy of the vm as to not make any changes.

This PR also:

1. Adds new rpc input validators: `address`, `blockOption`, `transaction`
1. Adds basic tests for all rpc input validators
1. Improves the rpc test helper code
1. Improves some of the rpc and validator typedocs

### DIFF
--- a/packages/client/lib/rpc/modules/eth.ts
+++ b/packages/client/lib/rpc/modules/eth.ts
@@ -117,9 +117,18 @@ export class Eth {
   async call(params: [RpcCallTx, string]) {
     const [transaction, blockOpt] = params
 
+    if (!this._vm) {
+      throw new Error('missing vm')
+    }
+
+    // use a copy of the vm in case new blocks are executed,
+    // and to not make any underlying changes during the call
+    const vm = this._vm.copy()
+
     if (blockOpt !== 'latest') {
-      const latest = await this.blockNumber()
-      if (blockOpt !== latest) {
+      const latest = await vm.blockchain.getLatestHeader()
+      const number = latest.number.toString(16)
+      if (blockOpt !== `0x${number}`) {
         return {
           code: INVALID_PARAMS,
           message: `Currently only "latest" block supported`,
@@ -127,17 +136,12 @@ export class Eth {
       }
     }
 
-    if (!this._vm) {
-      throw new Error('missing vm')
-    }
-
     if (!transaction.gas) {
       // If no gas limit is specified use the last block gas limit as an upper bound.
-      const latestHeader = await this._chain.getLatestHeader()
-      transaction.gas = latestHeader.gasLimit as any
+      const latest = await vm.blockchain.getLatestHeader()
+      transaction.gas = latest.gasLimit as any
     }
 
-    const vm = this._vm.copy()
     const txData = { ...transaction, gasLimit: transaction.gas }
     const tx = Transaction.fromTxData(txData, { common: vm._common, freeze: false })
 
@@ -171,9 +175,17 @@ export class Eth {
   async estimateGas(params: [RpcCallTx, string]) {
     const [transaction, blockOpt] = params
 
+    if (!this._vm) {
+      throw new Error('missing vm')
+    }
+
+    // use a copy of the vm in case new blocks are executed
+    const vm = this._vm.copy()
+
     if (blockOpt !== 'latest') {
-      const latest = await this.blockNumber()
-      if (blockOpt !== latest) {
+      const latest = await vm.blockchain.getLatestHeader()
+      const number = latest.number.toString(16)
+      if (blockOpt !== `0x${number}`) {
         return {
           code: INVALID_PARAMS,
           message: `Currently only "latest" block supported`,
@@ -181,17 +193,12 @@ export class Eth {
       }
     }
 
-    if (!this._vm) {
-      throw new Error('missing vm')
-    }
-
     if (!transaction.gas) {
       // If no gas limit is specified use the last block gas limit as an upper bound.
-      const latestHeader = await this._chain.getLatestHeader()
-      transaction.gas = latestHeader.gasLimit as any
+      const latest = await this._chain.getLatestHeader()
+      transaction.gas = latest.gasLimit as any
     }
 
-    const vm = this._vm.copy()
     const txData = { ...transaction, gasLimit: transaction.gas }
     const tx = Transaction.fromTxData(txData, { common: vm._common, freeze: false })
 
@@ -220,9 +227,17 @@ export class Eth {
   async getBalance(params: [string, string]) {
     const [addressHex, blockOpt] = params
 
+    if (!this._vm) {
+      throw new Error('missing vm')
+    }
+
+    // use a copy of the vm in case new blocks are sync'd
+    const vm = this._vm.copy()
+
     if (blockOpt !== 'latest') {
-      const latest = await this.blockNumber()
-      if (blockOpt !== latest) {
+      const latest = await vm.blockchain.getLatestHeader()
+      const number = latest.number.toString(16)
+      if (blockOpt !== `0x${number}`) {
         return {
           code: INVALID_PARAMS,
           message: `Currently only "latest" block supported`,
@@ -230,12 +245,8 @@ export class Eth {
       }
     }
 
-    if (!this._vm) {
-      throw new Error('missing vm')
-    }
-
     const address = Address.fromString(addressHex)
-    const account: Account = await (this._vm.stateManager as any).getAccount(address)
+    const account: Account = await vm.stateManager.getAccount(address)
     return `0x${account.balance.toString(16)}`
   }
 
@@ -294,9 +305,17 @@ export class Eth {
   async getCode(params: [string, string]) {
     const [addressHex, blockOpt] = params
 
+    if (!this._vm) {
+      throw new Error('missing vm')
+    }
+
+    // use a copy of the vm in case new blocks are sync'd
+    const vm = this._vm.copy()
+
     if (blockOpt !== 'latest') {
-      const latest = await this.blockNumber()
-      if (blockOpt !== latest) {
+      const latest = await vm.blockchain.getLatestHeader()
+      const number = latest.number.toString(16)
+      if (blockOpt !== `0x${number}`) {
         return {
           code: INVALID_PARAMS,
           message: `Currently only "latest" block supported`,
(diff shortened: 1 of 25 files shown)


packages/client/lib/rpc/modules/eth.ts
```diff
@@ -39,24 +77,174 @@ export class Eth {
...
+    const tx = Transaction.fromTxData(txData, { common: vm._common, freeze: false })
+
+    // set from address
+    const from = transaction.from ? Address.fromString(transaction.from) : Address.zero()
+    tx.getSenderAddress = () => {
+      return from
+    }
+
+    const { execResult } = await vm.runTx({ tx })
+    return bufferToHex(unpadBuffer(execResult.returnValue))
```
ryanio, line 155: is `unpadBuffer` here correct? otherwise the test contract [`returns (address addr)`](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1130/files#diff-6f5214b59c0bdda424d9c3634ff8af4d09e8bfdb28aa0b67af962e625c18434fR33) return value comes back as 32 bytes instead of 20.

without `unpadBuffer` it would return here: `0x000000000000000000000000ccfd725760a68823ff1e062f4cc97e1360e8d997`

packages/client/lib/rpc/modules/eth.ts
```diff
@@ -39,24 +73,178 @@ export class Eth {
...
+
+    const vm = this._vm.copy()
+    const txData = { ...transaction, gasLimit: transaction.gas }
+    const tx = Transaction.fromTxData(txData, { common: vm._common, freeze: false })
+
+    // set from address
+    const from = transaction.from ? Address.fromString(transaction.from) : Address.zero()
+    tx.getSenderAddress = () => {
+      return from
+    }
```
jochem-brouwer, line 149: Smart solution!

packages/client/lib/rpc/modules/eth.ts
jochem-brouwer: I am 99% sure that we don't want to unpad it. We want to return the actual return value. The ABI handlers will then decode the raw buffer. For instance, if ABI knows it is an address which is returned, then it will expect 32 bytes as output and will only return the last 20 bytes.

packages/client/lib/rpc/modules/eth.ts
```diff
@@ -39,24 +73,178 @@ export class Eth {
...
+
+    const { gasUsed } = await vm.runTx({
```
jochem-brouwer, line 205: Have to look this up, but `gasUsed` is the total execution gas right, so we did not subtract the gas refund yet?

packages/client/lib/rpc/modules/eth.ts
```diff
@@ -39,24 +73,178 @@ export class Eth {
...
+        }
+      }
+    }
+
+    if (!this._vm) {
+      throw new Error('missing vm')
+    }
+
+    const address = Address.fromString(addressHex)
+    const account: Account = await (this._vm.stateManager as any).getAccount(address)
```
jochem-brouwer, line 239: I think also here (and also other variants) we want to also copy VM. Between the `await this.blockNumber()` and actually getting the account, VM might execute a block, and therefore return the wrong value (it is not `latest`). What we could do is make an explicit VM copy in the constructor, and then use semaphores to ensure we are not running two queries on the copied VM twice.

packages/client/lib/rpc/types.ts
```diff
@@ -0,0 +1,8 @@
+export interface RpcCallTx {
+  from?: string
+  to: string
```
jochem-brouwer, line 3: Can this also be `null`? Or is this the empty string in RPC? (If we want to create a contract)

packages/client/test/rpc/eth/getCode.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+  const vm = (service!.synchronizer as FullSynchronizer).execution.vm
+
+  // genesis address with balance
+  const address = Address.fromString('0xccfd725760a68823ff1e062f4cc97e1360e8d997')
+
+  // sample contract from https://ethereum.stackexchange.com/a/70791
+  const data =
+    '0x608060405234801561001057600080fd5b506040516020806100ef8339810180604052602081101561003057600080fd5b810190808051906020019092919050505080600081905550506098806100576000396000f3fe6080604052600436106039576000357c010000000000000000000000000000000000000000000000000000000090048063a2a9679914603e575b600080fd5b348015604957600080fd5b5060506066565b6040518082815260200191505060405180910390f35b6000548156fea165627a7a72305820fe2ba3506418c87a075f8f3ae19bc636bd4c18ebde0644bcb45199379603a72c00290000000000000000000000000000000000000000000000000000000000000064'
+  const code =
+    '0x6080604052600436106039576000357c010000000000000000000000000000000000000000000000000000000090048063a2a9679914603e575b600080fd5b348015604957600080fd5b5060506066565b6040518082815260200191505060405180910390f35b6000548156fea165627a7a72305820fe2ba3506418c87a075f8f3ae19bc636bd4c18ebde0644bcb45199379603a72c0029'
```
jochem-brouwer, line 50: Where is the code actually put in state? Can't seem to find it.

jochem-brouwer (commented):
Some questions, great work in general! 😄

packages/client/lib/rpc/modules/eth.ts
ryanio: ah that makes sense, ok i will update, thanks!

packages/client/lib/rpc/modules/eth.ts
ryanio: it looks like it is the final amount, docs say `gasUsed = totalGasConsumed - gasRefund`

packages/client/lib/rpc/modules/eth.ts
ryanio: hm good point. i'm not sure about copying in the constructor because then it would only be on the block that the rpc server was created. i think we can figure out a solution here though.

packages/client/lib/rpc/types.ts
ryanio: well this is for `eth_call` which is not meant for contract creation, in the [wiki docs](https://eth.wiki/json-rpc/API#eth_call) `to` is required.

for `eth_estimateCall` it could be optional though, so maybe i'll rename the type to `RpcTx` and `to` can be optional. I have already written a `requiredField` validator for `eth_call` that is enforced as middleware 👍

packages/client/test/rpc/eth/getCode.spec.ts
ryanio: just a few lines down a transaction is constructed and run with this `data` as the contract deployment

packages/client/lib/rpc/modules/eth.ts
ryanio: updated in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1130/commits/6a48e434664ddc8a6c9597cd4fb8697864f1c921

packages/client/lib/rpc/modules/eth.ts
ryanio: decided to copy the vm first, let me know what you think, https://github.com/ethereumjs/ethereumjs-monorepo/pull/1130/commits/d4849c0dcfe426d9bfa1a69ec2aaa1932532194e

packages/client/lib/rpc/types.ts
ryanio: fixed in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1130/commits/487d85ee67b7d400ee8f4720160afdd1d5bdc813

packages/client/lib/rpc/modules/eth.ts
cgewecke: Copying makes sense to me as well... maybe in subsequent PR that adds support for specifying which block to execute the calls at this could be re-visited though. 

Just peeked at how ganache and hardhat handle this and it seems like
+ ganache [creates a new vm][1] 
+ hardhat [sets/unsets the StateManager state root][2] 

[1]: https://github.com/trufflesuite/ganache-core/blob/8a7ba780a664e32b2210e32c3bfebe359e7cfc37/src/chains/ethereum/ethereum/src/blockchain.ts#L782-L816
[2]: https://github.com/nomiclabs/hardhat/blob/22fef6a9b828b6f2432822e35484eb4f2b53dfdf/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts#L1422-L1437

cgewecke (dismissed):
This all looks awesome to me. Really nice! 

Approving but maybe another person could sweep over it?

holgerd77:
Rebased this.

holgerd77:
Rebased this.

packages/client/lib/rpc/modules/eth.ts
ryanio: @cgewecke awesome thanks for those links, i will take a look! i thought it would be good to introduce some kind of `stateAt(blockNumber): StateManager` functionality but didn't want to go that deep on this PR.

holgerd77:
Rebased this.

packages/client/lib/rpc/modules/eth.ts
```diff
@@ -39,24 +72,189 @@ export class Eth {
...
+   *       * data (optional) - Hash of the method signature and encoded parameters.
+   *   2. integer block number, or the string "latest", "earliest" or "pending"
+   * @returns The return value of the executed contract.
+   */
+  async call(params: [RpcTx, string]) {
+    const [transaction, blockOpt] = params
+
+    if (!this._vm) {
+      throw new Error('missing vm')
+    }
```
holgerd77, line 122: Some nit: this error message - if thrown - will likely lead to confusion by the user and should be replaced on occasion with something more descriptive like "client instance is running in a sync mode where this RPC call is not supported" (attention! first shot! 😄  ) or something.

No blocker though.

holgerd77 (approved):
Just becoming fully aware how great this PR is 😀  👍 , thanks a lot Ryan. I never digged so deeply into the various functionality exposed by the the various RPC calls and especially the implemented ones are indeed super-interesting.

Will merge here since this has also already been viewed upon by @jochem-brouwer and @cgewecke.

Issue #553, opened 2019-07-13
How to initialize the VM for a custom network with a custom genesis block


alcuadrado:
I'm creating this issue to document how to do this, as it's a somewhat intricate process that we should simplify, or at least create an example for it.

```ts
import Common from "ethereumjs-common";
import { toBuffer, bufferToHex, privateToAddress } from "ethereumjs-util";
import BN from "bn.js";
import VM from "ethereumjs-vm";
import Account from "ethereumjs-account";
import Blockchain from "ethereumjs-blockchain";
import { promisify } from "util";
import { StateManager } from "ethereumjs-vm/dist/state";
const Trie = require("merkle-patricia-tree/secure.js");

const BLOCK_GAS_LIMIT = 8000000;

interface GenesisAccount {
  balance: string;
  privateKey: string;
}

function getCurrentTimestamp(): number {
  return Math.ceil(new Date().getTime() / 1000);
}

async function createCustomNetworkVM() {
  const genesisAccounts: GenesisAccount[] = [
    {
      balance: "1000000000000000000123",
      privateKey:
        "0xe331b6d69882b4cb4ea581d88e0b604039a3de5967688d3dcffdd2270c0fd109"
    }
  ];

  // To create a custom network, we need a Common object with our custom
  // parameters.
  //
  // Depeding on which EthereumJS library you are using, you may want to set all
  // the parameters, or just a few.
  //
  // We are going to use the blokchain module, which creates its genesis block
  // based on the Common's params.
  //
  // One of those params is the root of the initial state trie. That trie
  // contains the initial set of accounts.
  //
  // We would normally use a StateManager to create such a state, but it
  // requires a Common, and we are doing this to create a Common. The solution
  // is to go deeper and directly use a Trie.
  const stateTrie = new Trie();
  const putIntoStateTrie = promisify(stateTrie.put.bind(stateTrie));
  for (const acc of genesisAccounts) {
    const balance = new BN(acc.balance);

    // We skip accounts with 0 balance, there's no need to include them.
    if (balance.eqn(0)) {
      continue;
    }

    const account = new Account({ balance: balance });
    const pk = toBuffer(acc.privateKey);
    const address = privateToAddress(pk);

    await putIntoStateTrie(address, account.serialize());
  }


  // Mimic precompiles activation. Not sure if this is necessary
  for (let i = 1; i <= 8; i++) {
    await putIntoStateTrie(
[...]

PR #1132, opened 2021-02-28
Client: keep syncing at tip of chain

jochem-brouwer:
This PR intends to keep the fullsync-syncer syncing at the tip of the chain. (Note: current branch target is the YoloV3 branch)

- Add `enqueueTask` to `fetcher`. This allows one to manually enqueue a task, and possibly restart the fetcher
- Add an option to not destroy the fetcher when it has processed all tasks
- Optimize full sync: in case no peers are available, then wait 5 seconds and try again, to get a best peer (this possibly fixes the re-bootstrap at some points)
- Fullsync `sync` method never resolves.
- Enqueue new blocks when the best peer announces new blocks

This is very WIP. Here are some things which need to be resolved: 

- [ ] Add logic when there is a gap in requested blocks (possibly, we just change the block fetcher, such that we can just update the highest block number on-demand, and not add all this extra logic)
- [ ] Add tests
- [ ] Fix the error as described below

Currently, it seems to randomly work for some blocks (when we are at tip of chain, i.e. on YoloV3, every 15 seconds it imports and executes a new block), but then at some point it throws:

```
  devp2p:eth Received NEW_BLOCK_HASHES message from 3.9.20.133:30303: e6e5a0dcc51743c5579b7bcdfadda2209ee2280e16e3e065d74333ba9ee8... +12s
  devp2p:eth Send GET_BLOCK_HEADERS message to 3.9.20.133:30303: c7830150d5018080 +5ms
  devp2p:eth Received BLOCK_HEADERS message from 3.9.20.133:30303: f9025cf90259a0d28799ec1625a5d9853afc3e10f415dd05071191af1bbf... +15ms
  devp2p:eth Send GET_BLOCK_BODIES message to 3.9.20.133:30303: e1a0dcc51743c5579b7bcdfadda2209ee2280e16e3e065d74333ba9ee836... +5ms
  devp2p:eth Received BLOCK_BODIES message from 3.9.20.133:30303: c3c2c0c0 +15ms
ERROR [02-28|18:27:21] Error [ERR_STREAM_PUSH_AFTER_EOF]: stream.push() after EOF
    at readableAddChunk (_stream_readable.js:260:30)
    at BlockFetcher.Readable.push (_stream_readable.js:213:10)
    at BlockFetcher.dequeue (/Volumes/Ethereum/ethereumjs-vm/packages/client/lib/sync/fetcher/fetcher.ts:163:17)
    at BlockFetcher.success (/Volumes/Ethereum/ethereumjs-vm/packages/client/lib/sync/fetcher/fetcher.ts:219:14)
    at /Volumes/Ethereum/ethereumjs-vm/packages/client/lib/sync/fetcher/fetcher.ts:268:44
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
  devp2p:eth Send GET_BLOCK_HEADERS message to 3.9.20.133:30303: c7830150d0048080 +28ms
  ```

### DIFF
--- a/packages/client/lib/sync/fetcher/fetcher.ts
+++ b/packages/client/lib/sync/fetcher/fetcher.ts
@@ -55,7 +55,7 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable 
   protected finished: number // number of tasks which are both processed and also finished writing
   protected running: boolean
   protected reading: boolean
-  private destroyOnFinish: boolean // Destroy the fetcher once we are finished processing each task.
+  private destroyWhenDone: boolean // Destroy the fetcher once we are finished processing each task.
 
   private _readableState?: {
     // This property is inherited from Readable. We only need `length`.
@@ -93,7 +93,7 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable 
     this.finished = 0
     this.running = false
     this.reading = false
-    this.destroyOnFinish = options.destroyWhenDone ?? true
+    this.destroyWhenDone = options.destroyWhenDone ?? true
   }
 
   /**
@@ -340,7 +340,7 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable 
       }
     }
     this.running = false
-    if (this.destroyOnFinish) {
+    if (this.destroyWhenDone) {
       this.destroy()
     }
   }
--- a/packages/client/lib/sync/fetcher/fetcher.ts
+++ b/packages/client/lib/sync/fetcher/fetcher.ts
@@ -26,6 +26,9 @@ export interface FetcherOptions {
 
   /* Retry interval in ms (default: 1000) */
   interval?: number
+
+  /* Destroy the fetcher once we are done */
+  destroyWhenDone?: boolean
 }
 
 /**
@@ -52,6 +55,8 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable 
   protected finished: number // number of tasks which are both processed and also finished writing
   protected running: boolean
   protected reading: boolean
+  private destroyOnFinish: boolean // Destroy the fetcher once we are finished processing each task.
+
   private _readableState?: {
     // This property is inherited from Readable. We only need `length`.
     length: number
@@ -88,6 +93,7 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable 
     this.finished = 0
     this.running = false
     this.reading = false
+    this.destroyOnFinish = options.destroyWhenDone ?? true
   }
 
   /**
@@ -153,6 +159,29 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable 
         return
       }
       f = this.out.peek()
+    }
+  }
+
+  /**
+   * Enqueues a task. If autoRestart is true, and Fetcher is not running, then restart the fetcher.
+   * @param task
+   * @param autoRestart
+   */
+  enqueueTask(task: JobTask, autoRestart = false) {
+    if (!this.running && !autoRestart) {
+      return
+    }
+    const job: Job<JobTask, JobResult, StorageItem> = {
+      task,
+      time: Date.now(),
+      index: this.total++,
+      state: 'idle',
+      peer: null,
+    }
+    this.in.insert(job)
+    if (!this.running && autoRestart) {
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      this.fetch()
     }
   }
 
@@ -300,17 +329,8 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable 
       return false
     }
     this.write()
-    this.tasks().forEach((task: JobTask) => {
-      const job: Job<JobTask, JobResult, StorageItem> = {
-        task,
-        time: Date.now(),
-        index: this.total++,
-        state: 'idle',
-        peer: null,
-      }
-      this.in.insert(job)
-    })
     this.running = true
+    this.tasks().forEach((task: JobTask) => this.enqueueTask(task))
     while (this.running) {
       if (!this.next()) {
         if (this.finished === this.total) {
@@ -319,7 +339,10 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable 
         await this.wait()
       }
     }
-    this.destroy()
+    this.running = false
+    if (this.destroyOnFinish) {
+      this.destroy()
+    }
   }
 
   /**
--- a/packages/client/lib/sync/fullsync.ts
+++ b/packages/client/lib/sync/fullsync.ts
@@ -98,60 +98,7 @@ export class FullSynchronizer extends Synchronizer {
     // eslint-disable-next-line no-async-promise-executor
     return await new Promise(async () => {
       if (!peer) return false
-      peer!.on('message', (message: any) => {
-        if (message.name === 'NewBlockHashes') {
-          let min: BN = new BN(-1)
-          const data: any[] = message.data
-          const blockNumberList: string[] = []
-          data.forEach((value: any) => {
-            const blockNumber: BN = value[1]
-            blockNumberList.push(blockNumber.toString())
-            if (min.eqn(-1) || blockNumber.lt(min)) {
-              min = blockNumber
-            }
-          })
-          if (min.eqn(-1)) {
-            return
-          }
-          const numBlocks = blockNumberList.length
 
-          // check if we can request the blocks in bulk
-          let bulkRequest = true
-          const minCopy = min.clone()
-          for (let num = 1; num < numBlocks; num++) {
-            min.iaddn(1)
-            if (!blockNumberList.includes(min.toString())) {
-              bulkRequest = false
-              break
-            }
-          }
-
-          if (bulkRequest) {
-            // FIXME
-            // eslint-disable-next-line @typescript-eslint/no-use-before-define
-            blockFetcher.enqueueTask(
-              {
-                first: minCopy,
-                count: numBlocks,
-              },
-              true
-            )
-          } else {
-            data.forEach((value: any) => {
-              const blockNumber: BN = value[1]
-              // FIXME
-              // eslint-disable-next-line @typescript-eslint/no-use-before-define
-              blockFetcher.enqueueTask(
-                {
-                  first: blockNumber,
-                  count: 1,
-                },
-                true
-              )
-            })
-          }
-        }
-      })
       const latest = await this.latest(peer)
       if (!latest) return false
       const height = new BN(latest.number)
@@ -173,6 +120,7 @@ export class FullSynchronizer extends Synchronizer {
         destroyWhenDone: false,
       })
       const blockFetcher = <BlockFetcher>this.blockFetcher
+      this.addNewBlockHandlers(peer, blockFetcher)
 
       this.config.events.on(Event.SYNC_FETCHER_FETCHED, (blocks) => {
         blocks = blocks as Block[]
--- a/packages/client/lib/sync/fullsync.ts
+++ b/packages/client/lib/sync/fullsync.ts
@@ -95,46 +95,103 @@ export class FullSynchronizer extends Synchronizer {
    * @return Resolves when sync completed
    */
   async syncWithPeer(peer?: Peer): Promise<boolean> {
-    if (!peer) return false
-    const latest = await this.latest(peer)
-    if (!latest) return false
-    const height = new BN(latest.number)
-    const first = this.chain.blocks.height.addn(1)
-    const count = height.sub(first).addn(1)
-    if (count.lten(0)) return false
+    // eslint-disable-next-line no-async-promise-executor
+    return await new Promise(async () => {
+      if (!peer) return false
+      peer!.on('message', (message: any) => {
+        if (message.name === 'NewBlockHashes') {
+          let min: BN = new BN(-1)
+          const data: any[] = message.data
+          const blockNumberList: string[] = []
+          data.forEach((value: any) => {
+            const blockNumber: BN = value[1]
+            blockNumberList.push(blockNumber.toString())
+            if (min.eqn(-1) || blockNumber.lt(min)) {
+              min = blockNumber
+            }
+          })
+          if (min.eqn(-1)) {
+            return
+          }
+          const numBlocks = blockNumberList.length
 
-    this.config.logger.debug(
-      `Syncing with peer: ${peer.toString(true)} height=${height.toString(10)}`
-    )
+          // check if we can request the blocks in bulk
+          let bulkRequest = true
+          const minCopy = min.clone()
+          for (let num = 1; num < numBlocks; num++) {
+            min.iaddn(1)
+            if (!blockNumberList.includes(min.toString())) {
+              bulkRequest = false
+              break
+            }
+          }
 
-    this.blockFetcher = new BlockFetcher({
-      config: this.config,
-      pool: this.pool,
-      chain: this.chain,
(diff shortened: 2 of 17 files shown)

### CODE REVIEW

**Comment by holgerd77** [comment]
Rebased this one respectively cherry-picked the new commit.

**Comment by jochem-brouwer** [comment]
Will take a look at the tests ASAP, they probably fail because this Promise in full sync now never resolves.

File: packages/client/lib/sync/fullsync.ts [hunk 1]
```diff
@@ -99,60 +99,119 @@ export class FullSynchronizer extends Synchronizer {
...
+          const first = new BN(blocks[0].header.number)
+          const hash = short(blocks[0].hash())
+          this.config.logger.info(
+            `Imported blocks count=${blocks.length} number=${first.toString(
+              10
+            )} hash=${hash} hardfork=${this.hardfork} peers=${this.pool.size}`
+          )
+        })
+      await this.blockFetcher.fetch()
+    })
```
holgerd77, line 202: Haven't looked at this in detail, but just as a general remark, this is a lot of code in bulk, would be great if we can split this up a bit.

jochem-brouwer:
Yolo's bootnode is down =(

holgerd77:
Yes, I guess Yolo is shut down completely 😕, a bit sad, right? Maybe we want to get to a situation quickly where we can spin up our own networks between two plus clients, then we can do some proper integrated testing.

jochem-brouwer:
Yeah, I think we need some tooling to import custom genesis blocks so we can setup a development environment to sync with private nodes.

holgerd77:
Rebased this.

packages/client/lib/sync/fetcher/fetcher.ts
```diff
@@ -88,6 +93,7 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable
     this.finished = 0
     this.running = false
     this.reading = false
+    this.destroyOnFinish = options.destroyWhenDone ?? true
```
holgerd77, line 96: This will always resolve to `true`, will fix.

holgerd77: Update: ah, I am incorrect, just tested, sorry.

Will just keep as is.

holgerd77:
Short writeup on how to get to a quick test scenario with two local clients:

```shell
# 1. Start a dev1 client to fill in a first few hundred blocks, cut off quickly (goerli works more streamlined than mainnet)
npm run client:start -- --network=goerli --datadir=datadir-dev1

# 2. Restart dev1 client in a useful debug scenario and with discovery switched off, keep running
DEBUG=devp2p:ETH npm run client:start:dev1 -- --datadir=datadir-dev1

# 3. Start dev2 client from another window to connect to dev1 client, should sync these first couple of hundred blocks
DEBUG=devp2p:ETH npm run client:start:dev1 -- --datadir=datadir-dev1
```

Should lead to something like this:

grafik

acolytec3:
> Short writeup on how to get to a quick test scenario with two local clients:
> 
> ```shell
> # 1. Start a dev1 client to fill in a first few hundred blocks, cut off quickly (goerli works more streamlined than mainnet)
> npm run client:start -- --network=goerli --datadir=datadir-dev1
> 
> # 2. Restart dev1 client in a useful debug scenario and with discovery switched off, keep running
> DEBUG=devp2p:ETH npm run client:start:dev1 -- --datadir=datadir-dev1
> 
> # 3. Start dev2 client from another window to connect to dev1 client, should sync these first couple of hundred blocks
> rm -Rf ./datadir-dev2 && DEBUG=devp2p:ETH npm run client:start:dev1 -- --datadir=datadir-dev1
> ```
> 
> Should lead to something like this:
> 
> grafik

Nice! I've been trying to conceive of how to approach this.

packages/client/lib/sync/fullsync.ts
```diff
@@ -95,69 +92,62 @@ export class FullSynchronizer extends Synchronizer {
...
-        ? `basefee=${blocks[0].header.baseFeePerGas} `
-        : ''
-      this.config.logger.info(
-        `Imported blocks count=${blocks.length} number=${first.toString(
-          10
-        )} hash=${hash} ${baseFeeAdd}hardfork=${this.config.chainCommon.hardfork()} peers=${
-          this.pool.size
-        }`
+    // eslint-disable-next-line no-async-promise-executor
+    return await new Promise(async () => {
```
ryanio, line 96: hm there is no resolve or reject on this promise

packages/client/lib/sync/lightsync.ts
```diff
@@ -74,43 +71,47 @@ export class LightSynchronizer extends Synchronizer {
    * @return Resolves when sync completed
    */
   async syncWithPeer(peer?: Peer): Promise<boolean> {
-    if (!peer) return false
-    const height = new BN(peer.les!.status.headNum)
-    const first = this.chain.headers.height.addn(1)
-    const count = height.sub(first).addn(1)
-    if (count.lten(0)) return false
+    // eslint-disable-next-line no-async-promise-executor
+    return await new Promise(async () => {
```
ryanio, line 75: no resolve or reject here either

packages/client/lib/sync/fullsync.ts
ryanio: hm, maybe this is a place where `Event.SYNC_SYNCHRONIZED` would be useful, we could resolve when received if the typedoc above is still accurate: `* @return Resolves when sync completed`

ryanio:
Ok I was able to accomplish quite a few things, please browse my added commits for details on each one.

The integration tests still aren't fully working for me yet, my last commit helped but now `[Integration:FullSync]` is hanging after finishing, have to stop for today but I can continue tomorrow.

packages/client/lib/sync/sync.ts
```diff
@@ -149,75 +149,70 @@ export abstract class Synchronizer {
...
-        }
-
-        // Check if new sync target height can be set
-        if (!this.syncTargetHeight || blockNumber.gt(this.syncTargetHeight)) {
-          newSyncHeight = blockNumber
-        }
-      })
-      if (min.eqn(-1)) {
-        return
+  handleNewBlockHashes(data: [Buffer, BN][]) {
```
holgerd77, line 162: Ah, thanks, that's a lot better! 😄

packages/client/lib/sync/fullsync.ts
```diff
@@ -93,15 +93,15 @@ export class FullSynchronizer extends Synchronizer {
    */
   async syncWithPeer(peer?: Peer): Promise<boolean> {
     // eslint-disable-next-line no-async-promise-executor
-    return await new Promise(async () => {
-      if (!peer) return false
+    return await new Promise(async (resolve, reject) => {
+      if (!peer) return resolve(false)
```
holgerd77, line 97: Still no `Promise` expert, what would happen if we skip this whole `Promise` wrapping? 🤔

Generally this whole Promise resolving has gotten less important with the extraction of the sync status determination by listening and acting to chain updates (I already thougth about just returning `Promise<void>` on this function).

We just need to make sure that in the [Sync.start()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/lib/sync/sync.ts#L94) call no concurrent `sync()` calls are triggered.

packages/client/test/sync/lightsync.spec.ts
```diff
@@ -71,11 +72,16 @@ tape('[LightSynchronizer]', async (t) => {
       chain,
     })
     sync.best = td.func<typeof sync['best']>()
+    sync.latest = td.func<typeof sync['latest']>()
     td.when(sync.best()).thenReturn({ les: { status: { headNum: new BN(2) } } } as any)
+    td.when(sync.latest(td.matchers.anything())).thenResolve({ number: new BN(2) })
```
holgerd77, line 77: Some annotation since I am also reading up stuff: [td.matchers.anything()](http://testdouble.github.io/testdouble.js/5-stubbing-results.html#tdmatchersanything)

packages/client/test/sync/sync.spec.ts
```diff
@@ -34,6 +35,16 @@ tape('[Synchronizer]', async (t) => {
...
+    sync.syncTargetHeight = new BN(1)
+    setTimeout(() => {
+      // eslint-disable-next-line no-extra-semi
+      ;(sync as any).chain._blocks = {
+        latest: { hash: () => Buffer.from([]) },
+        td: new BN(0),
+        height: new BN(1),
+      }
+      config.events.emit(Event.CHAIN_UPDATED)
+    }, 100)
```
holgerd77, line 50: Ah, pretty cool fix.

holgerd77:
@ryanio thanks for the continuation here, really great additions! 👍 😄 

I already wondered if it wouldn't be a good fit if we directly start with the integration of some simplified form of the block builder here. The idea would be to just do PoA for now and only produce empty blocks (maybe nevertheless using the `VM.buildBlock()` function to have this framing right from the beginning).

Guess this would help us a lot with testing (so: the practical testing by setting things up with two local clients as described [here](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1132#issuecomment-899445740)).

WDYT?

Guess the procedure would be - roughly - as follows:

1. Run the client with a custom chain and the (one, e.g. dev1) client set as the single validator of the chain (just live-thinking: this might get a bit tricky and we are not quite there yet since we a) can't easily start a CLI client with a custom `Common` and b) can't use a custom genesis file for a custom chain, see #553, might give this some thought later the day)
2. Start a second client waiting for the blocks from client 1
3. Client 1 waiting for the chain to be synced (this will likely already need some revised mechanism on determining the sync status, since there is just no `CHAIN_UPDATED` event. Update: the block timestamp is likely a lot more direct and better fitting candidate, we should replace the existing `syncTargetHeight` mechanism (can likely be completely removed I guess?) and comparison with a last block timestamp comparison, which likely also needs some revised triggers
4. If client is "in turn" (as in "clique in turn") it should produce an (empty) block and send it out clique `period` seconds after the last block timestamp with `NEW_BLOCK_HASHES` (and - for now or eventually later - `NEW_BLOCK`)
5.

holgerd77:
Just thinking a bit about the genesis case: this is likely a special case to handle. So if the signer is "in turn" on genesis (a bit dangerous to ensure that this really only happen once and not accidentally in between, if e.g. a client is started and connect to a peer with an empty blockchain), it should take genesis values from the genesis block in `Common`, add its own timestamp and then broadcast this block as the very first block.

Is this correct?

holgerd77:
(sorry, hope I am not opening up on too many things at once here, if we want to do things simultaneously we can very much divide into manageable sub tasks and it's very much ok if this whole thing here takes another 2-3 weeks or so)

holgerd77:
Update: or maybe we keep the `syncTargetHeight` as some general concept, this information is likely also nice to have and display on CLI on updates, but nevertheless switch to timestamp comparison on sync status determination.

holgerd77:
Rebased this.

packages/client/lib/sync/fullsync.ts
holgerd77: Update: tested this with removing the Promise wrapper, this then stops at another test in `sync.spec.ts` unit test, couldn't investigate further but still not completely getting why (if) this is needed.

holgerd77:
Ok, tests are processing further with the latest changes in 28bf3b4, still not finishing the `# should sync with best peer` tests in the `[Integration:LightSync]` tests (`npm run tape -- test/integration/lightsync.spec.ts`).

packages/client/lib/sync/sync.ts
```diff
@@ -76,7 +76,16 @@ export abstract class Synchronizer {
...
-      if (this.syncTargetHeight && this.chain.blocks.height.gte(this.syncTargetHeight)) {
+      let targetHeight = this.syncTargetHeight
+      if (!targetHeight) {
+        const best = this.best()
+        if (best?.eth) {
+          targetHeight = new BN(best.eth.status.latestBlock)
+        } else if (best?.les) {
+          targetHeight = new BN(best.les.status.headNum)
+        }
+      }
```
holgerd77: I guess we want to have this for directly setting the sync target height on a first iteration.

Have given this a bit more thought. The timestamp from the block is actually also not a good indicator for having the chain synced. For example for a stalled chain with 3 participants and 2 signers where we are the one signer and the other signer dropped some time ago (and we therefore would be on charge to produce a next block) there just won't be any new `NEW_BLOCK_HASHES` triggers that would update the chain and bring us in a synced state so that we could finally decide to start on block producing.

I guess with this combination (setting to the peer height and then soon after to the block height from `NEW_BLOCK_HASHES`) we should already cover a good flavour of scenarios (in the "normal" case there should also somewhat soon be a `NEW_BLOCK_HASHES` message coming in).

One scenario which is not covered yet though: how to decide in a two peer setup where we are the block producer and the other peer is just waiting for us that we "are synced" and therefore should actually start/continue producing blocks? 🤔

holgerd77: Have added this in ccc0e09.

holgerd77: Ok, I guess this b24dc4a is my first take on this, to wait for some time if we find a best peer with a higher target height (and in the meantime there is also the chance that some `NEW_BLOCK_HASHES` come in) and otherwise set the sync target height to the local DB height and emit a `SYNCHRONIZED` event.

This should now also cover this 2 peer scenario and give us the basics to start on block producing.

(really tricky edge cases here though, I wonder e.g. if we should add an explicit flag to allow block production from genesis onwards to not accidentally start new chains if there is no network connectivity)

holgerd77:
Phew, I had the most intense debug session on this I guess I ever had this year, roughly three hours straight and testing close to everything to see why these tests are hanging in `test/integration/fullsync.spec.ts`.

I did try to get hold of the hanging processes by using `ps aux` and then going down with `lsof -p {process.pid}` which I got from this [leaked-handles](https://www.npmjs.com/package/leaked-handles) package, but this was not very informative.

I manually injected some `console.log(obj)` into the `node_modules/leaked-handles` sources to get the full stream objects hanging and not only this summarized version. This at least revealed that these were actually two `WriteStream` and one `ReadStream` handles hanging, still not sure if the TCP hint from `leaked-handles` was therefore generally misguiding or if a TCP stream is also a `WriteStream`.

I digged super deep into the `mocks/mockserver.ts` and `mocks/network.ts` code and entered debug output in the respective `destroy()` methods to see if all streams got destroyed. Seems that everything is good there.

Finally I went down all the commits. The complete hanging was actually already there after the first commit from Jochem starting the work here.

---

So as some side note and (re-)learning: we really really should do client work and commits on a very granular basis and extremely regularly run the tests. This is getting so hard otherwise to track down things, especially since there often additionally come things like timing/handlers/unresolved Promises issues into play.

---

I am now at least at a point where the tests pass every second time or so. Lol.

Hope that this can be resolved with some timer or so. Very much open for everyone to be investigated and continued.

I would have a strong tendency to then merge this PR once we have got the tests reliably running. This is otherwise getting too extensive and I think we have got a good basis here. This can then already be used to continue work e.g. on the tx pool or the block builder and things from here can also still later be revised or evolved. These kind of changes from here will also get easier to debug once we have the local chain sync functionality (so tx pool + block builder) in place. That will allow us to very easily simulate various edge cases locally.

packages/client/test/sync/sync.spec.ts
ryanio: yeah a little hacky but was the best way I could find to easily override for now

packages/client/lib/sync/fullsync.ts
ryanio: I believe it's part of some promise chain that awaits for results, but I'm sure we could find an alternative architecture, I think it is probably a bit redundant now since it just awaits the SYNC'd event we can probably just listen for that wherever this promise is functionally being used.

PR #1134, opened 2021-03-03
Publish monorepo to virtual registry for e2e testing

cgewecke:
Adds CI infrastructure (mostly ported from Web3.js) which makes it possible to install any monorepo package at a given state into a 3rd party target so its tests can be run, validating the build. 

Hardhat would be a good candidate when it upgrades to v5.

This works by lerna publishing all the packages to a virtual npm registry in CI, after incrementing them a `minor` step and tagging them @e2e. 

There's also a script which saves a json map of the published packages and their new versions to `resolutions.json`.  This can be injected into a target project's package.json under the `resolutions` key in order to tell yarn to coerce all @ethereumjs dependencies to their virtually published version when installing. (See [yarn selective dependency resolutions][1])

In the absence of a target, have mocked installation in the CI validation step to show that new versions are being installed and their monorepo sub-dependencies are *also* virtually published versions.

For example installing `@ethereumjs/vm@e2e` in blank project results in this package.json
```
{
"name": "test-publish",
"version": "1.0.0",
 ...
"dependencies": {
  "@ethereumjs/vm": "^5.2.0" // latest version is actually 5.1.0
 }
}
```

Inspecting the package.json at `node_modules/@ethereum/vm` shows the transitive @ethereumjs dependencies are also at new, post-latest versions. Conversely, packages not hosted by the virtual registry have been fetched from "real" npm. 
```
{
  "name": "@ethereumjs/vm",
  "version": "5.2.0",
  "description": "An Ethereum VM implementation",
  ...
  "dependencies": {
    "@ethereumjs/block": "^3.2.0",
   "@ethereumjs/blockchain": "^5.2.0",
   "@ethereumjs/common": "^2.2.0",
   "@ethereumjs/tx": "^3.1.0", 
   "async-eventemitter": "^0.2.4",
   "core-js-pure": "^3.0.1",
   "debug": "^2.2.0",
   "ethereumjs-util": "^7.0.8",
   "functional-red-black-tree": "^1.0.1",
   "mcl-wasm": "^0.7.1",
   "merkle-patricia-tree": "^4.2.0",
   "rustbn.js": "~0.2.0",
   "util.promisify": "^1.0.1"
  },
  ...
}
```

The resolutions.json object looks like this:
```
{
  "@ethereumjs/block": "3.2.0",
  "@ethereumjs/blockchain": "5.2.0",
  "@ethereumjs/client": "0.1.0",
  "@ethereumjs/common": "2.2.0",
  "@ethereumjs/devp2p": "3.1.0",
  "@ethereumjs/ethash": "1.1.0",
  "merkle-patricia-tree": "4.2.0",
  "@ethereumjs/tx": "3.1.0",
  "@ethereumjs/vm": "5.2.0"
}
```

(This output is all visible in the CI logs - click on view raw logs in the top right hand corner to see complete output.)

One consideration here is that when there are breaking changes, the target tests may fail. This likely shouldn't be a `required` job. 

[1]: https://classic.yarnpkg.com/en/docs/selective-version-resolutions/


ryanio:
looks really great. nice explanation in the opening post, i think we should include some of that info in a section of the readme or developer.md for future reference.

cgewecke:
Added a note about this under the README's existing verdaccio section in 3d18f55

cgewecke:
Tried to get this compiling for a couple hours and didn't get anywhere...

Began looking at [conditional type formulae][1] hoping that will let TS understand there's adequate runtime type checking in the code. 

Block package also looks like it needs help. Will revisit in the morning...

[1]: https://www.typescriptlang.org/docs/handbook/2/conditional-types.html

holgerd77:
I have the impression that we have this potential API passing-in compatibility problem all over the place, was just not so aware of this until now. 😋

The most prominent one likely being `Common` passed in as an API parameter on all libraries.

So seems we should be a bit aware of that. On the other hand it just seems to not make much sense to pick this out "right now" and make this a blocker kind of problem. So I guess we should really rather concentrate on the E2E compatibility challenge with 3rd-party consuming libraries.

cgewecke:
Added an E2E target which uses the vm: https://github.com/NoahZinsmeister/as-you-permit. (This is an [EIP-2612][1] implementation written by the Uniswap lead dev and will hopefully be stable.)

[1]: https://eips.ethereum.org/EIPS/eip-2612

Tests pass.

In the initial CI run, added a "proof" log line in the vm constructor to verify that the target tests are actually using the vm at the latest commit. 

https://github.com/ethereumjs/ethereumjs-monorepo/pull/1134/checks?check_run_id=2090212226#step:7:167

Screen Shot 2021-03-11 at 2 08 48 PM

cgewecke:
(I had to regenerate the root package-lock.json during a rebase across adding ethereumjs-util to the monorepo. The line count change seems a little crazy though.)

holgerd77 (approved):
That's really super-cool and looks great, really nice to have this so generically integrated, seems we can really easily expand here. We'll likely want to see if we want to run this on every PR - execution time is for the moment ok with ~5 min but will further go up with eventually added new E2E tests - but we'll see. Might also be enough to ran nightly or so.

Anyhow, great work Chris! 😄 

Will approve and merge.

Welbert5753 (approved):

PR #1136, opened 2021-03-04
Remove @types/bn.js from block,blockchain,trie,tx,vm packages

cgewecke:
Fixes a TS compilation failure on install relating to ethereumjs-util 7.0.9. 

PR: 
+ removes @types/bn.js as a dependency everywhere
+ imports `bn.js` via ethereumjs-util everywhere (for clarity about its source)

Idea is:
+ bn.js [is not a direct dependency in any package][1] - it always comes from ethereumjs-util
+ @types/bn.js [is a direct dependency][2] of ethereumjs-util
+ monorepo should let ethereumjs-util manage the bn types

[1]: https://github.com/ethereumjs/ethereumjs-monorepo/search?p=1&q=bn.js
[2]: https://github.com/ethereumjs/ethereumjs-util/blob/b100d13f827a3d495d3de85f3e5d6756749b514f/package.json#L88

### DIFF
--- a/packages/block/package.json
+++ b/packages/block/package.json
@@ -42,7 +42,6 @@
     "@ethereumjs/common": "^2.1.0",
     "merkle-patricia-tree": "^4.1.0",
     "@ethereumjs/tx": "^3.0.2",
-    "@types/bn.js": "^4.11.6",
     "ethereumjs-util": "^7.0.8"
   },
   "devDependencies": {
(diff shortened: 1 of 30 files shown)


ryanio (approved):
lgtm, nice simplification :)

holgerd77:
Ah, yeah, that's great. I wasn't aware that we have so many direct BN imports here. 😄  👍

PR #1141, opened 2021-03-06
EIP-2718/EIP-2930 Typed Tx Improvements and Tests

holgerd77:


holgerd77:
Ok, this is now ready for review! 😄 

I tried to make this as unopinionated as possible and mostly integrate things already discussed, nevertheless @jochem-brouwer let me know if you are not going along with some changes.

Think this can best be reviewed per-commit, most commit messages + the change diff should speak for themselves. I reduced the added API complexity on various cases and e.g. removed the `rawTxHash()` alias or the `getChainId()` function. I think we should rather go low here and not unnecessarily add on API complexity along this already huge change set where we might later be at a point where it becomes hard to take exposed functionality back.

The single one block with a somewhat more substantial change is some refactoring of the `TxData` types in 
d3d1a17. This was going along with a lot of type data field redundancy which is now removed. I was also convinced after some reflection that this generic `TxData` type (from before the commit) on the TransactionFactory methods where all new  parameter fields from new transaction types will be thrown in will likely not scale very well and is also suboptimal from a type safety perspective. So I think it is needed that we have distinct data types for the different txs.

jochem-brouwer (approved):
Did a per-commit review, I am OK with all changes!

PR #754, opened 2020-05-27
Implement EIP-2315: subroutines

cgewecke:
#737 

Implements [EIP-2315][1] which introduces three new opcodes and a subroutine "return stack" to help separate control flow logic from computational operations on the main VM stack. This provides some efficiency benefits and is convenient for static analysis tools that require clear control flow semantics.

A detailed analysis of the EIP can be found at [ethereum-magicians here][2]

[2]: https://ethereum-magicians.org/t/eip-2315-simple-subroutines-for-the-evm-analysis/4229
[1]: https://eips.ethereum.org/EIPS/eip-2315

+ Added tests include the EIP examples plus a couple from the besu and geth implementations.
+ Have made the Stack class's overflow threshold configurable because the return stack's maximum height is spec'd at 1023 (1 less than normal)

(In draft pending updates to the ethereum tests (and common Berlin hardfork opcodes?))

PR #1142, opened 2021-03-08
Update ethereum-tests submodule to 8.0.0 / Remove EIP-2315 from berlin

holgerd77:
This is a simple PR which updates the `ethereum-tests` submodule to the `Berlin` release [8.0.0](https://github.com/ethereum/tests/releases/tag/8.0.0) with a removed `EIP-2315` HF. `EIP-2315` is also removed here in `Common` in the `berlin` HF definition file.

I originally also wanted to add the block numbers for `berlin` along this PR. But I don't trust these numbers yet since the `Ropsten` HF would be so close and there is no corresponding `go-ethereum` release out yet, see also https://github.com/ethereum/pm/issues/248#issuecomment-792573651.

### DIFF
--- a/packages/common/src/chains/goerli.json
+++ b/packages/common/src/chains/goerli.json
@@ -70,8 +70,8 @@
     },
     {
       "name": "berlin",
-      "block": null,
-      "forkHash": null
+      "block": 4460644,
+      "forkHash": "0x757a1c47"
     }
   ],
   "bootstrapNodes": [
--- a/packages/common/src/chains/mainnet.json
+++ b/packages/common/src/chains/mainnet.json
@@ -72,8 +72,8 @@
     },
     {
       "name": "berlin",
-      "block": null,
-      "forkHash": null
+      "block": 12244000,
+      "forkHash": "0xeb440f6"
     }
   ],
   "bootstrapNodes": [
--- a/packages/common/src/chains/rinkeby.json
+++ b/packages/common/src/chains/rinkeby.json
@@ -70,8 +70,8 @@
     },
     {
       "name": "berlin",
-      "block": null,
-      "forkHash": null
+      "block": 8290928,
+      "forkHash": "0x6910c8bd"
     }
   ],
   "bootstrapNodes": [
--- a/packages/common/src/chains/ropsten.json
+++ b/packages/common/src/chains/ropsten.json
@@ -72,8 +72,8 @@
     },
     {
       "name": "berlin",
-      "block": null,
-      "forkHash": null
+      "block": 9812189,
+      "forkHash": "0xa157d377"
     }
   ],
   "bootstrapNodes": [
--- a/packages/common/src/hardforks/berlin.json
+++ b/packages/common/src/hardforks/berlin.json
@@ -3,5 +3,5 @@
   "comment": "HF targeted for July 2020 following the Muir Glacier HF",
   "url": "https://eips.ethereum.org/EIPS/eip-2070",
   "status": "Draft",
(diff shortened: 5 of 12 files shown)


holgerd77:
Note that this PR is also running with the `type: test all hardforks` label.

holgerd77:
Some tests failing, feel very free to directly push here. Will stop for now.

jochem-brouwer:
Have given this a quick glance, [this](https://github.com/ethereumjs/ethereumjs-monorepo/blob/03d00fdc41086aa7fd9dc4869453704f3ab69dc1/packages/vm/lib/evm/opcodes/codes.ts#L246) is probably the problem. But I don't have time today to actually check it.

jochem-brouwer:
This should fix the problem. Have also moved EIP2315 tests from `berlin` directory into `EIPs` directory.

@holgerd77 as for why your tests passed locally: you probably forgot to rebuild Common, so if you would create a Berlin common it automatically activates EIP2315.

holgerd77:
@jochem-brouwer lol, good point 😄  (yes, you're right)

holgerd77:
@jochem-brouwer we can now also include the [Berlin HF blocks](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/berlin.md) here along since this is now settled. Do you might want to directly do, don't want to push here in between?

jochem-brouwer:
Ah right, good point, will add them here and push.

holgerd77:
@jochem-brouwer cool! 👍 

Have also cross-checked the [EIP-2315 implementation](https://github.com/ethereumjs/ethereumjs-monorepo/pull/754/files), I think we've all covered now.

jochem-brouwer:
Right, I don't have the fork hashes, and sadly we also don't have an utility module to generate these fork hashes (do we want this?). I have asked on R&D if someone has the hashes for me.

holgerd77:
@jochem-brouwer ah, you are not reading my messages 😋  - also wrote about this in the Yolov3 PR - you can use the Common._calcForkHash() method for that.

holgerd77:
(so put in the block numbers and then run with a Common set to `berlin` from console)

jochem-brouwer:
Ah I missed this message, just figured this out as well that we actually have this feature 😅

jochem-brouwer:
Found a possible problem: we do [not pad](https://github.com/ethereumjs/ethereumjs-monorepo/blob/94651f7b42bb24a34dcc7fdd4d2daadb568d7c23/packages/common/src/index.ts#L608) the fork hashes, which is why for instance [this](https://github.com/ethereumjs/ethereumjs-monorepo/blob/94651f7b42bb24a34dcc7fdd4d2daadb568d7c23/packages/common/src/chains/kovan.json#L26) kovan hash is of 7 characters (9 characters with 0x prefixed). This is problematic if we do `Buffer.from(hash.slice(2))`, you'd get the Buffer '10ffe5`, where we expect Buffer `010ffe56`. 

I don't know if this is an issue, but it is not consistent with other packages, we usually left pad these strings or Buffers.

Anyways, have added Berlin blocks to Common, see [the EF blog](https://blog.ethereum.org/2021/03/08/ethereum-berlin-upgrade-announcement/) for block numbers. Also got the fork hashes from Geth's [fork hash tests](https://github.com/ethereum/go-ethereum/blob/44c0bb2b44ca71790ccf8a842eccaa72619a2a02/core/forkid/forkid_test.go). And I also learned we can actually calculate these hashes 😅

holgerd77 (commented):
Ok, double-checked the numbers, LGTM

PR #1143, opened 2021-03-08
Trie: fix benchmarks

ryanio:
This PR applies a small fix to the trie benchmarks code for the conditional `auto-push` field. It also keeps a max limit of 30 previous runs of data, and upgrades the `checkout` step from v1 to v2.


holgerd77 (approved):
LGTM

PR #1144, opened 2021-03-10
Tx: Small fixes and improvements

ryanio:
This PR applies a few small fixes and improvements to the new tx library with EIP2718 and 2930 support:

* `AccessListEIP2930Transaction` - removes `asList` parameter in `raw()`, preferring to use `serialize()`
* `TransactionFactory` - for clarity, renames `fromRawData` to `fromSerializedData`
* For clarity, renames `_validateExceedsMaxInteger(validateCannotExceedMaxInteger)` to `_validateCannotExceedMaxInteger(values)`
* Renames internal usage of `fromRlpSerializedTx` to `fromSerializedTx`
* Introduces `common.copy()` for easier deep copy functionality.

### DIFF
--- a/packages/block/src/block.ts
+++ b/packages/block/src/block.ts
@@ -149,12 +149,14 @@ export class Block {
   }
 
   /**
-   *  Returns a Buffer Array of the raw Buffers of this block, in order.
+   * Returns a Buffer Array of the raw Buffers of this block, in order.
    */
   raw(): BlockBuffer {
     return [
       this.header.raw(),
-      this.transactions.map((tx) => <Buffer[]>tx.raw()),
+      this.transactions.map((tx) =>
+        'transactionType' in tx && tx.transactionType > 0 ? tx.serialize() : tx.raw()
+      ) as Buffer[],
       this.uncleHeaders.map((uh) => uh.raw()),
     ]
   }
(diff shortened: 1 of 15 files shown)


packages/block/src/block.ts
```diff
@@ -149,12 +149,14 @@ export class Block {
...
-   *  Returns a Buffer Array of the raw Buffers of this block, in order.
+   * Returns a Buffer Array of the raw Buffers of this block, in order.
    */
   raw(): BlockBuffer {
     return [
       this.header.raw(),
-      this.transactions.map((tx) => <Buffer[]>tx.raw()),
+      this.transactions.map((tx) =>
+        'transactionType' in tx && tx.transactionType > 0 ? tx.serialize() : tx.raw()
+      ) as Buffer[],
```
holgerd77, line 159: Yes, this makes sense. I think this is the right thing to do to get the whole raw and serialized data setup consistent again and avoid this unlucky splitting of the `tx.raw()` method.

We just should be aware here that this introduces another tx type induced version dependency between the libraries, so block serialization will now be wrong when an old block version is used upon a tx-type-containing tx release (but so be it).

I guess we minimally need to communicate the set of library versions to be used when using typed txs very clearly if we want to stay in the minor release realm (as we've done with the clique changes, these were still a bit better encapsulated by the `Common` `consensusType` setting though). 🤔 

//cc @cgewecke

holgerd77 (approved):
This all looks great, thanks for the continued improvements and diving deep here. 😄  👍

holgerd77:
Argh. Accidentally hit the "close" button and need to wait for the tests again. 😛

Issue #1069, opened 2021-01-22
Client: parse multiple bootnodes in transports CLI options


jochem-brouwer:
When we want to add multiple bootnodes to our client, we have a problem if we want to add multiple bootnodes. For instance,

`npm run client:start -- --transports rlpx:bootnodes=d860a01f9722d78051619d1e2351aba3f43f943f6f00718d1b9baa4101932a1f5011f16bb2b1bb35db20d6fe28fa0bf09636d26a87d31de9ec6203eeedb1f666@18.138.108.67,22a8232c3abc76a16ae9d6c3b164f98775fe226f0917b0ca871128a74a8e9630b458460865bab457221f1d448dd9791d24c4e5d88786180ac185df813a68d4de@3.209.45.79:30303`

Results in 

```
{
  name: 'rlpx',
  options: {
    bootnodes: 'd860a01f9722d78051619d1e2351aba3f43f943f6f00718d1b9baa4101932a1f5011f16bb2b1bb35db20d6fe28fa0bf09636d26a87d31de9ec6203eeedb1f666@18.138.108.67',
    '22a8232c3abc76a16ae9d6c3b164f98775fe226f0917b0ca871128a74a8e9630b458460865bab457221f1d448dd9791d24c4e5d88786180ac185df813a68d4de@3.209.45.79:30303': undefined
  }
}
```

This is because our parser first [splits](https://github.com/ethereumjs/ethereumjs-monorepo/blob/9a3b800cec7f772bb55471237806b39d1b4a1929/packages/client/lib/util/parse.ts#L34) on `,` and then splits the remainder on `=`. Transports themselves are first splitted on `:`, this is problematic, as we use it in the node format. We would need a new format to support a bootnodes array. We could maybe use an escape character like `"`, which then implies we don't split upon the contents between two `"`?

jochem-brouwer:
Hmm I think that an escape character as `"` would work - right? Then we use the raw string as bootnode option, and this will then get passed to `parseBootnodes` down the road? @ryanio 

Then we'd use `npm run client:start -- --transports rlpx:"bootnode1,bootnode2"` with bootnodeN of format `id@ip:port`.

ryanio:
ah good catch, hm can we just split on the transports "rlpx:" and "libp2p:" first?

I don't think we'll be adding more transports at a rate that would be unreasonable to just manually check for them.

cgewecke:
[...]

PR #1145, opened 2021-03-10
Client: split CLI transports options for bootnodes and multiaddrs

ryanio:
This PR splits the CLI transports options for bootnodes and multiaddrs. Closes https://github.com/ethereumjs/ethereumjs-monorepo/issues/1069.


holgerd77 (approved):
Great, LGTM! 👍  This has also the strong advantage now that these options are not hidden any more in this parsing logic but are now explicitly exposed e.g. along `ethereumjs --help`.

PR #973, opened 2020-12-01
Client monorepo transition

holgerd77:
This PR transitions the [ethereumjs-client](https://github.com/ethereumjs/ethereumjs-client) library to the monorepo, closes #892 following #971 

Same procedure as for the devp2p library - now with a bit more comprise commit history 😄 . I had to do some temporary fixes along the way (mainly around some type imports, see added TODO comments) which I didn't want to fix and rather concentrate on the main process. Nothing though which couldn't be addressed separately after merge I would say.

Ok, really excited to get this going. 😄 

Happy beaconchain start everyone! 🥳  🎉 🚀

PR #1146, opened 2021-03-11
Util monorepo transition

holgerd77:
This PR transitions the [ethereumjs-util](https://github.com/ethereumjs/ethereumjs-util) library into the monorepo following #973.

Note that the original repository shouldn't be archived since there are branches for Util v4, v5 and v6 existing (current major version: v7) which might need bug fix releases at some point.


holgerd77:
Ok, this is now working. 😀

Ready for review.

packages/util/.gitignore
```diff
@@ -0,0 +1,72 @@
...
+build/Release
+
+# Dependency directory
+# https://docs.npmjs.com/misc/faq#should-i-check-my-node-modules-folder-into-git
+node_modules
+package-lock.json
+
+# Build folder
+dist
+
```
ryanio, line 72: We can clear this file since all these entries are in the root [`.gitignore`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/.gitignore)

```suggestion
```

ryanio (dismissed):
nice, looks great!

i think the util docs need to be regenerated to update the permalinks to the code located in the monorepo, but i'm sure that would be fine to do before the next release.

ryanio:
I'm noticing in the ci that `trie-benchmarks` is having some issues in the latest run with `Cannot find module 'ethereumjs-util' or its corresponding type declarations.`, I find it a bit weird though, the `test-trie` step seems to be running fine which has the same cache, setup, and build, so not sure what's up.

ryanio (approved):

holgerd77:
Ok, feel free to merge from my side. 🙂

holgerd77:
Respectively I've just merged. 😀

PR #1148, opened 2021-03-14
Implement EIP1559 Fee Market + EIP3198 BaseFee

jochem-brouwer:
This PR implements [EIP 1559](https://eips.ethereum.org/EIPS/eip-1559).

This will probably go along with a refactor of the typed transactions. I noticed that both access lists EIP2930 and the fee market EIP1559 transactions share a lot of the same logic, which we can encapsulate in an (abstract?) class. I have merely copied a lot of code around which should not be what we want.

Very WIP.

Note: have to rollback 42d83d0d45d5776d7fcd03df019dc83d51a13f31

Some TODOs which are not related to this EIP

- [ ] Address issue with `runBlock`/`runTx` editing `gasUsed` and originally reporting wrong clique miner (clean this up and add a test)
- [ ] Figure out the RLPx 0x04 packet issue with the Nethermind node.
- [ ] Figure out why we fail to connect to Besu


holgerd77:
@jochem-brouwer cool, great that you gave this a start! 😄  👍  🎉

Yeah, it will very much make sense to further abstract here. Let's please give it a cut here, maybe you've got some time during the day and jump on a quick call (let me know when in chat)?

It actually get's hyper-tricky here. I guess the main problem is that the concept of inheritance just doesn't work well with this very abstract typed transaction EIP concept where a tx - theoretically - could contain anything. This might work for now - also e.g. regarding the question what is currently contained in `BaseTransaction`, but this is rather out of "randomness", since the feature set is just for this current set of typed transactions randomly fitting.

But one can already see where this leads to: in EIP-1559 there is e.g. no `gasPrice` and -whop - the whole `BaseTransaction` gets somewhat invalid from an API PoV. If there will be another tx type - e.g. with a new signature theme removing the v,r,s values - whop - base transaction respectively the whole inheritance chain will break again.

Not sure if this overloads us for now: but I at least theoretically wonder if e.g. some kind of module concept (gas module, signature module, access list module, eventually serialization module) might serve us better here and will be more future proof.

And a base transaction (or however it's called) might contain the information what kind of modules are activated, like so:

```typescript

class BaseTransaction {
 
  modules: {
    gas: 'EIP1559',
    accessLists: true,
    signature: 'ECDSA', // future
    serialization: 'RLP',
  }
```

Something like this would also allow us to cleaner react in third-party code (VM, Block,...) to features available. The current approach will otherwise further down the road lead to a conditional hell like either always uncleanly just request for some thought-to-be-active properties (`if (tx.gasPrice) { // do something... }`) or (not better) adding to an ever growing chain of tx types (`if (typeof tx in [AccessListTx, EIP1559Tx,... ])`. All not really scaling.

Again, let's just jump on the call sometime. If we do this later the day eventually the others want to join as well.

holgerd77:
For reference: [Base fee tests](https://github.com/hyperledger/besu/pull/1407/files?file-filters%5B%5D=.json#diff-e149bb446ef0addba7d952f215a8538ec330f2d8a2eabc22f87189014d329721) in Besu

holgerd77:
Nethermind EIP-1559 implementation: https://github.com/NethermindEth/nethermind/pull/2341

jochem-brouwer:
Rebased this one

jochem-brouwer:
Yeah this design is getting rather complex already, I can fix this build error but the idea I got in mind might not be a really great design - will take a look at the other implementations first to get some inspiration here.

holgerd77:
@jochem-brouwer yeah, this is really tricky to answer and what level we want/need to refactor the whole library for this. In doubt just put this aside for now for a couple of days as some suggestion 😋 , sometimes this helps as well and we can pick this question up on our call and discuss together. If you would have alternatively some time for #1132 that would e.g. also be cool if we get this going a bit again. 😄

jochem-brouwer:
Okay, added some very basic VM tests, but at this point I cannot continue until I have more test data. I will probably create some very simple EIP1559 tests to just see that it produces *something* (so I can sign a transaction for instance), but I cannot be sure if we are correct.

jochem-brouwer:
Rebased upon master.

jochem-brouwer:
Ok great, tests pass, now trying to connect to Aleut.

jochem-brouwer:
We are only downloading 1 block at a time since for some reason the bootnode either sends invalid data, or we have a bug in devp2p. It tries to decode RLP message bodies where RLP states it should decode ~33k bytes, but the block is actually only 1023 bytes long. (The message body itself is 1024 bytes).

jochem-brouwer:
Ok, this commit 8c112f0 fixes the consensus bug at block 55. The problem had nothing to do with EIP1559, it had to do with `runBlock`/`runTx` in VM: `runBlock` now feeds `runTx` an edited block than the block which is actually being ran; the `gasUsed` field is changed and used to keep track of the cumulative gas used. This commit is a hotfix and might be changed in the future (probably removing editing the block and just passing along the block which is actually being ran in the context). 

We should add a test to ensure that clique blocks with txs pay the right miner, which is not tested in VM currently.

There also still is a problem with the Nethermind bootnode. If we request more than 1 block (might be ok to request 2 or 3) - lets say the original 50, then we get a RLP decode error on devp2p.

With these fixes we can sync up to block 3704:

```
ERROR [04-26|22:49:16] Error: invalid transactions: errors at tx 0: tx unable to pay base fee (block number=3704 hash=ee3a234b297c792ffbecc5f90042f053a14bb6760ec4058e61bca3f20d88c292)     at BlockHeader._error (/Volumes/Ethereum/ethereumjs-monorepo/packages/block/src/header.ts:857:15)
    at Block.validateData (/Volumes/Ethereum/ethereumjs-monorepo/packages/block/src/block.ts:272:25)
    at Block.validate (/Volumes/Ethereum/ethereumjs-monorepo/packages/block/src/block.ts:258:16)
    at /Volumes/Ethereum/ethereumjs-monorepo/packages/blockchain/src/index.ts:883:9
    at Blockchain.runWithLock (/Volumes/Ethereum/ethereumjs-monorepo/packages/blockchain/src/index.ts:410:21)
    at Blockchain._putBlockOrHeader (/Volumes/Ethereum/ethereumjs-monorepo/packages/blockchain/src/index.ts:854:5)
    at Blockchain.putBlock (/Volumes/Ethereum/ethereumjs-monorepo/packages/blockchain/src/index.ts:809:5)
    at Blockchain.putBlocks (/Volumes/Ethereum/ethereumjs-monorepo/packages/blockchain/src/index.ts:795:7)
    at Chain.putBlocks (/Volumes/Ethereum/ethereumjs-monorepo/packages/client/lib/blockchain/chain.ts:274:5)
    at BlockFetcher.store (/Volumes/Ethereum/ethereumjs-monorepo/packages/client/lib/sync/fetcher/blockfetcher.ts:84:5)
    ```

**Comment by jochem-brouwer** [comment]
Another node; we only connect to Nethermind, other 3 bootnodes are up, have to figure out what is the cause of this. http://eip1559-testnet.ops.pegasys.tech:3001/

File: packages/block/src/block.ts [hunk 1]
```diff
@@ -223,9 +223,15 @@ export class Block {
   validateTransactions(stringError: true): string[]
   validateTransactions(stringError = false) {
     const errors: string[] = []
-
+    const self = this
     this.transactions.forEach(function (tx, i) {
```
ryanio, line 227: i'm sure you know, but if you update the forEach to arrow syntax you can save yourself needing to do `self = this`

```suggestion
    this.transactions.forEach((tx, i) => {
```

packages/block/src/header.ts
```diff
@@ -165,6 +178,25 @@ export class BlockHeader {
     return BlockHeader.fromHeaderData(headerData, opts)
   }
 
+  public static getCommon(opts: BlockOptions = {}, blockNumber: BN = ZERO_BN): Common {
+    let common
+    if (opts.common) {
+      common = Object.assign(Object.create(Object.getPrototypeOf(opts.common)), opts.common)
```
ryanio, line 184: nice, we have `common.copy()` now as a shortcut

packages/block/src/header.ts
```diff
@@ -165,6 +178,25 @@ export class BlockHeader {
...
+      const chain = 'mainnet' // default
+      if (opts.initWithGenesisHeader) {
+        common = new Common({ chain, hardfork: 'chainstart' })
+      } else {
+        // This initializes on the Common default hardfork
+        common = new Common({ chain })
+      }
+    }
+    if (opts.hardforkByBlockNumber) {
+      common.setHardforkByBlockNumber(blockNumber.toNumber())
```
ryanio, line 195: this accepts BN now

```suggestion
      common.setHardforkByBlockNumber(blockNumber)
```

holgerd77:
"Another node; we only connect to Nethermind"

You're too much in @jochem-brouwer. 😂  I've taken the freedom to correct. 😄

packages/block/src/block.ts
holgerd77: @ryanio ah, I personally had actually no idea. That's a super-valuable hint. 😄  (this workaround is really annoying and not very beautiful).

packages/client/lib/blockchain/chain.ts
```diff
@@ -110,7 +110,7 @@ export class Chain extends EventEmitter {
         db: options.chainDB,
         common: this.config.chainCommon,
         validateBlocks: true,
-        validateConsensus: false,
+        validateConsensus: true,
```
holgerd77, line 113: This is putting too much load on the PoW sync, I guess what we can do is just to activate for PoA by including a conditional switch along the `common` instance setting from above.

holgerd77:
@ryanio hmm, still not sure if #1222 is fixing the problem, I did a rebase here locally and removed 42d83d0 (so mainly the constant bootstrap retriggering, setting `maxPerRequest` to `1` is not relevant yet in this context here) - also did another `npm run bootstrap` locally from within this branch to make sure - and yet again the `Send BlockHeaders` message is not to be seen. 😕

Phew, tricky thing.

ryanio:
@holgerd77 this is tricky indeed, I rebased this branch on master (didn't touch the commit you specified) and was able to recreate the issue again.

I have a feeling it's still some kind of promise race condition we haven't handled properly. As a test I added a 5s timeout after starting the servers but before calling bootstrap:

```typescript
    ...
    await Promise.all(this.config.servers.map((s) => s.start()))

    // wait 5s
    function timeout(ms: number) {
      return new Promise(resolve => setTimeout(resolve, ms));
    }
    await timeout(5000)

    await Promise.all(this.config.servers.map((s) => s.bootstrap()))
    ...
```

and that worked for me! I got a `Send GET_BLOCK_HEADERS` line...so I will keep tracing this async code and see if there is something we are missing or not being awaited properly. I just tried again with `1s` and that even worked. I also tried with `setImmediate` with no timeout period but that didn't work.

We have the `@typescript-eslint/no-floating-promises` rule turned off in `packages/devp2p/.eslintrc.js` so I have a hunch that it may be some floating promise causing the problem.

holgerd77:
@ryanio great, thanks for this deep analysis, maybe we can live with the time for immediate continuation here, but would be good if we can trace this down nevertheless sooner or eventually later.

jochem-brouwer:
EIP3198: 8f6806d

jochem-brouwer:
Sent some EIP3198 txs to Aleut, client syncs correctly.

jochem-brouwer:
Okay, last push should fix the tests.

Note that Common's `hardforkGteHardfork` now has an extra condition: this ensures that if `hardfork2` is unknown, it will return `false` instead of `true`. This was originally problematic for custom chains. If we would introduce a new HF, this would automatically get activated on these custom chains, unless these custom chains explicitly stated that the block number of that specific block was `null`. Note that also `hardforkGteHardfork('chainstart', 'blablablablabla')` was originally always true; I do not think this should be the case. (This also results in that any new EIP we introduce originally automatically were activated on custom chains which did not explicitly `null` these HFs)

See bae28c3

packages/vm/lib/runBlock.ts
```diff
@@ -247,7 +247,16 @@ async function applyTransactions(this: VM, block: Block, opts: RunBlockOpts) {
...
 
-    const gasLimitIsHigherThanBlock = block.header.gasLimit.lt(tx.gasLimit.add(gasUsed))
+    let maxGasLimit
+    if (this._common.isActivatedEIP(1559)) {
+      maxGasLimit = block.header.gasLimit.muln(
+        this._common.param('gasConfig', 'elasticityMultiplier')
+      )
+    } else {
+      maxGasLimit = block.header.gasLimit
+    }
```
holgerd77, line 257: 👍  😄

holgerd77:
Pretty cool, great work Jochem! 😄  👍  🎉

holgerd77:
Rebased this.

packages/common/src/eips/1559.json
```diff
@@ -0,0 +1,25 @@
...
+    "comment": "Fee market change for ETH 1.0 chain",
+    "url": "https://eips.ethereum.org/EIPS/eip-1559",
+    "status": "Review",
+    "minimumHardfork": "london",
+    "gasConfig": {
+      "baseFeeMaxChangeDenominator": {
+        "v": 8,
+        "d": "Maximum base fee change denominator"
+      },
+      "elasticityMultiplier": {
```
holgerd77, line 13: I would name this `gasTargetElasticityMultiplier`, otherwise it lacks semantics if just read on its own (just learning from the experience with other parameter names from common - some inherited from former devs - where it is just not possible to get the context when reading).

packages/common/src/eips/1559.json
```diff
@@ -0,0 +1,25 @@
...
+        "v": 2,
+        "d": "Maximum block gas target elasticity"
+      },
+      "initialBaseFee": {
+        "v": 1000000000,
+        "d": "Initial base fee on first EIP1559 block"
```
holgerd77, line 19: The description from the EIP (in the code example) states for this base fee:

> # default to 1,000,000,000 for blocks before INITIAL_FORK_BLOCK_NUMBER

Does this match with description here respectively with the semantics (my beloved word for the day 😝 ) you applied this constant for in the code (didn't have a look yet)? I have to say I have my problems to interpret this comment from above myself, what does "before" mean in this context? Blocks before shouldn't have a base fee at all I guess? 🤔  Or is this read like "before it is implicitly always `initialBaseFee` and on the first EIP1559 block the base fee is actually calculated for the first time? Then this would still conflict with the comment from the code here saying - if I read correctly - that this fee is applied "as is" on the first block and there is not yet some calculation.

Just writing this so extensively here since I think this is important to have this really clear - also in conjunction with other clients - if the fee calculation starts at `EIP1159 block` or `EIP1159 block + 1.

packages/block/src/header.ts
```diff
@@ -558,13 +579,85 @@ export class BlockHeader {
         throw new Error('uncle block has a parent that is too old or too young')
       }
     }
+
+    if (this._common.isActivatedEIP(1559)) {
+      if (!this.baseFeePerGas) {
+        throw new Error('EIP1559 block has no base fee field')
+      }
+      let parentBaseFee = parentHeader.baseFeePerGas
+      const isInitialEIP1559Block = this._common.getEIPActivationBlockNumber(1559)!.eq(this.number)
```
holgerd77, line 588: This is at the end a question of the hardfork and the respective block can be retrieved directly with the existing `common.hardforkBlock()` function and the context can still be given with the variable name, so nothing would be lost here by using the following I would say:

```typescript
const isInitialEIP1559Block = this._common.hardforkBlock('london').eq(this.number)
```

Not some super strong opinion here, but I guess there is no need for this additional `getEIPActivationBlockNumber()` function to be introduced and this can be removed from `Common`. Or co you have got a preference to keep for some reason? Is there any value to be able to retrieve this directly by EIP number?

holgerd77 (commented):
Just sending out some first comments on the wire. 😛

packages/common/src/chains/mainnet.json
```diff
@@ -74,6 +74,11 @@
       "name": "berlin",
       "block": 12244000,
       "forkHash": "0xeb440f6"
```
holgerd77, line 76: Just a side note that this berlin forkhash doesn't seem to be correct with this missing 0 at the beginning of the byte sequence. I ran into some mainnet connection problems in some other context I can't remember with the other client expecting a `0x0eb440f6`. Just mentioning here since this might have caused some other bugs and/or needed fixes on this specific PR.

holgerd77: Additional note (we might want to open a dedicated issue to discuss): apart from an eventual direct fix here this might also need a correction of the `forkHash` creation logic in `Common`.

packages/common/src/eips/3198.json
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "EIP-3198",
+    "number": 3198,
+    "comment": "BASEFEE opcode",
+    "url": "https://eips.ethereum.org/EIPS/eip-3198",
+    "status": "Review",
+    "minimumHardfork": "london",
```
holgerd77, line 7: Same here, minimum hardfork: `berlin`.

For EIP dependency please use the `requiredEIPs` property instead, see e.g. the `2930.json` definition file with `"requiredEIPs": [2718, 2929]`.

packages/common/src/hardforks/london.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "london",
+  "comment": "HF targeted for ??? following the Berlin fork",
```
holgerd77, line 3: You can update with "July 2021" here I guess. 😀

packages/common/src/index.ts
```diff
@@ -220,6 +220,26 @@ export default class Common extends EventEmitter {
...
+        const hf = hfChanges[1]
+        if (hf['name'] === hardfork.name && 'eips' in hf) {
+          if (hf['eips'].includes(eip)) {
+            const fork = this._getHardfork(hardfork.name)
+            return new BN(fork.block)
+          }
+        }
+      }
+    }
+  }
```
holgerd77, line 241: Can eventually be removed, see comment on usage on `BlockHeader.validate()`.

packages/common/src/index.ts
```diff
@@ -468,7 +488,7 @@ export default class Common extends EventEmitter {
       if (hf['name'] === hardfork2) posHf2 = index
       index += 1
     }
-    return posHf1 >= posHf2
+    return posHf1 >= posHf2 && posHf2 !== -1
```
holgerd77, line 491: Ok, I thought about this a bit longer on eventual side effects and also in-between hardforks which didn't activate on some chains (e.g. `muirGlacier` on the PoA chains) and after some back-and-forth I cautiously think I can go along with this change, still a bit unsure though. An in between thought I had was if we should rather throw if one of the two HFs provided are not present. This will likely make things more unstable though I guess and might cause unintentional throws where the code would have otherwise proceeded without errors.

Nevertheless writing this down so explicitly here so that others might give this some thought as well if needed.

packages/devp2p/src/eth/index.ts
```diff
@@ -217,7 +217,10 @@ export class ETH extends EventEmitter {
         this._latestBlock = latestBlock
       }
       const forkHashB = Buffer.from(this._forkHash.substr(2), 'hex')
-      const nextForkB = this._nextForkBlock.toArrayLike(Buffer)
+      const nextForkB = this._nextForkBlock.eqn(0)
+        ? Buffer.from('', 'hex')
+        : this._nextForkBlock.toArrayLike(Buffer)
+
```
holgerd77, line 223: Just to confirm here, so this only applies to genesis block comparisons, if I read the code correctly?

holgerd77 (commented):
Round 2.

packages/vm/lib/runTx.ts
```diff
@@ -245,16 +255,31 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
...
+  let inclusionFeePerGas
+  if (this._common.isActivatedEIP(1559)) {
+    const baseFee = block.header.baseFeePerGas
+    const EIP1559Data = tx.getEIP1559Data()
+    inclusionFeePerGas = BN.min(
+      EIP1559Data.maxInclusionFeePerGas,
+      EIP1559Data.maxFeePerGas.sub(baseFee!)
+    )
+    gasPrice = inclusionFeePerGas.add(baseFee!)
+  } else {
```
holgerd77, line 268: Ok, I have to admit here that I am not able here to completely grasp how an old legacy transaction integrates here on gas price calculation. However can we skip this `getEIP1559Data()` method introduction - which seems very much redundant and unnecessary to me also for an EIP-1559 tx - and instead operate with a `tx.type` switch?

Guess this should be relatively easily doable as well? 🤔

holgerd77: (same argument for the second application of this function in `BlockHeader.validateTransactions()`)

holgerd77 (commented):
Round 3.

packages/common/src/eips/1559.json
jochem-brouwer: Yeah this is indeed confusing. It seems like the spec says that on the first block we should actually calculate the base fee and that it is thus not 1e9. But we actually assume that the first block has base fee 1e9 and this is also enforced in the block package. Since we synced to Aleut just fine this is weird and that could imply that other clients do not follow the spec (?). Will ask on R&D.

packages/common/src/chains/mainnet.json
jochem-brouwer: Huh this is weird, I thought I fixed this, also had a similar problem in Yolo I think.

packages/common/src/index.ts
jochem-brouwer: Yeah I agree this is a very big change. However, if we start enforcing these things that means that every time we implement a new EIP and people use their custom chains then if they update their packages it suddenly does not work anymore, since they have to explicitly state that they do not support the EIP. I would say that (before this change) it was an opt-out behavior, but it should actually be opt-in: we assume that if EIPs are not present in some common, they are not active.

packages/devp2p/src/eth/index.ts
jochem-brouwer: No, this means that if we are on the final fork currently known (so on Aleut, at block 10 we switch to London, and then after that there is no hardfork), we report the empty buffer. If you instead do `this._nextForkBlock.toArrayLike(Buffer)`, then the reported Buffer is `Buffer.from('00', 'hex')`. If we send this over the wire, other nodes (at least Geth) will disconnect.

packages/tx/src/types.ts
```diff
@@ -2,6 +2,7 @@ import { BN, AddressLike, BNLike, BufferLike, PrefixedHexString } from 'ethereum
 import Common from '@ethereumjs/common'
 import { default as Transaction } from './legacyTransaction'
 import { default as AccessListEIP2930Transaction } from './eip2930Transaction'
+import { FeeMarketEIP1559Transaction } from '.'
```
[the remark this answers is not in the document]
jochem-brouwer: Yep, will change, this probably went unnoticed since VSCode auto-imports suggestions.

packages/vm/lib/runTx.ts
jochem-brouwer: Yep we can do this, I somewhat got this idea from the EIP which wants to normalize txs, I can add a switch statement. (This does mean we should not forget to add new tx types to this switch statement in case we add new txs)

jochem-brouwer:
General comment: should really add more EIP1559 tests to Tx, our coverage of tx is now almost -20% (!!)

jochem-brouwer:
Quick update here: we cannot merge this yet because the EIP spec is most likely going to change.

packages/common/src/eips/3198.json
holgerd77: Will resolve conversations here when doing the local changes and then later push. I have now added `2930` as the sole required EIP, I guess this should do it with `2930` itself having `2718` and `2929` as EIP requirements, which are then also enforced automatically.

packages/common/src/eips/1559.json
holgerd77: On a second thought having this in the `gasConfig` section maybe that's enough context. Will leave as-is.

packages/devp2p/src/eth/index.ts
holgerd77: Ah ok, thanks for clarification. We had this initially wrong along receiving the status message as well.

packages/tx/src/types.ts
```diff
@@ -134,6 +138,21 @@ export interface AccessListEIP2930TxData extends TxData {
   type?: BNLike
 }
 
+export interface FeeMarketEIP1559TxData extends AccessListEIP2930TxData {
```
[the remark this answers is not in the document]
holgerd77: Have tested this `never` type, this actually works: 😄 

image

packages/tx/src/eip2930Transaction.ts
```diff
@@ -138,53 +138,26 @@ export default class AccessListEIP2930Transaction extends BaseTransaction<Access
...
+    this.accessList = accessListData.accessList
+    this.AccessListJSON = accessListData.AccessListJSON
+    // Verify the access list format.
+    AccessLists.verifyAccessList(this.accessList)
 
     this.chainId = chainId ? new BN(toBuffer(chainId)) : this.common.chainIdBN()
-    this.accessList = usedAccessList
+    this.gasPrice = new BN(toBuffer(gasPrice === '' ? '0x' : gasPrice))
+
+    this._validateCannotExceedMaxInteger({ gasPrice: this.gasPrice })
```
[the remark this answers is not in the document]
holgerd77: Have updated.

packages/block/src/header.ts
holgerd77: Update: ok, just seeing `Block` library `header.ts` that there is the following code where I cannot make too much out of it directly:

```typescript
// Small hack to hack in the activation block number
// (Otherwise there would be need for a custom chain only for testing purposes)
common.getEIPActivationBlockNumber = function (eip: number) {
  if (eip == 1559) {
    return new BN(1)
  }
}
```

Will leave this for now. Still think method should likely be removed though.

packages/vm/lib/runTx.ts
holgerd77: Ok, this was the most delicate fix for me, relatively confident that I translated this correctly, but please have a somewhat careful look on the changes at bb7e3b6e.

packages/tx/src/types.ts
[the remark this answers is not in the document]
holgerd77: This had some interesting cascade: I added this in 85c7efc, which caused the type casting in `TransactionFactory.fromTxData()` not to work any more, which I think is a good thing since this was pretty loose on casting everything to an AL tx, which is neither accurate or future-proof. I then reworked this code with type-specific casts and at the end also deprecated the `getTransactionClass()` method.

Might be a case for discussion, but I would be a strong proponent. We are not using this internally with this change any more and this method generally has a somewhat strange API, I would assume that this won't be much used by external users.

packages/tx/src/eip2930Transaction.ts
```diff
@@ -157,7 +157,7 @@ export default class AccessListEIP2930Transaction extends BaseTransaction<Access
     this.chainId = chainId ? new BN(toBuffer(chainId)) : this.common.chainIdBN()
     this.gasPrice = new BN(toBuffer(gasPrice === '' ? '0x' : gasPrice))
 
-    this._validateCannotExceedMaxInteger({ gasPrice: this.gasPrice })
+    this._validateCannotExceedMaxInteger({ r: this.r, s: this.s, gasPrice: this.gasPrice })
```
jochem-brouwer, line 156: Have to check but is this not validated in `BaseTransaction`?

packages/vm/lib/runTx.ts
```diff
@@ -147,11 +147,7 @@ export default async function runTx(this: VM, opts: RunTxOpts): Promise<RunTxRes
   await state.checkpoint()
 
   // Is it an Access List transaction?
-  if (
-    'transactionType' in opts.tx &&
-    opts.tx.transactionType === 1 &&
-    this._common.isActivatedEIP(2929)
-  ) {
+  if (opts.tx.transactionType !== 0 && this._common.isActivatedEIP(2929)) {
```
jochem-brouwer, line 151: Just noticed but this does not seem correct anymore, what has EIP 2929 to do with this? I think this was intended to be EIP 2718, to check if typed txs are activated.

holgerd77: I guess this was from the "early days" of implementation where you added a lot of EIP checks 😀, guess to be on the safe side in doubt. I think we can (and should) move over to limit these checks to the absolute minimum. Often things are already checked on Common instantiation and are then redundant in other places and therefore not necessary.

In this case here I would say that we can completely drop the check and just go with `if (opts.tx.transactionType !== 0)` since the`EIP-2718` check has already happened in Common if correctly activated with `EIP-2930`. Otherwise the code will throw in the follow-up lines.

packages/block/src/header.ts
```diff
@@ -165,6 +178,25 @@ export class BlockHeader {
...
+      } else {
+        // This initializes on the Common default hardfork
+        common = new Common({ chain })
+      }
+    }
+    if (opts.hardforkByBlockNumber) {
+      common.setHardforkByBlockNumber(blockNumber)
+    }
+    return common
+  }
```
holgerd77, line 198: This is a bit a strange function to have publicly exposed, I wouldn't want to do this - at least not without prior additional discussion - I currently not see the direct use case respectively this would lead to strange patterns like:

```typescript
const common = new Common()
const common2 = BlockHeader.getCommon({ common })
const blockHeader = BlockHeader.fromHeaderData({}, { common2 })
``` 

At least my fear is that people will do these kind of things. Will switch to an implementation where this is private.

holgerd77: Ok, this a199138 should very much dot it without changing the semantics too much and without the need for this `getCommon()` extraction, sorry if this feels a bit like a roll-back of work.

We should really really urgently follow the pattern from tx in `Block` and do a centralized default value setting, passing in values as a dict in the constructor instead of the values just one-by-one (this is already noted for v6) and then just having `BlockHeader.fromHeaderData()` as an alias to the constructor. The problems here with the need for having the final common in the static methods comes at the end from this defragmentation and doing fragmented checks in the various static constructors which should just be done once in the main constructor.

Anyhow: please note that this change also introduces a more explicit default value for `baseFeePerGas`, before this was somewhat implicitly set to `0` with `opts.common.isActivatedEIP(1559) ? new BN(toBuffer(baseFeePerGas)) : undefined`. I now have set this to `1`, but totally unsure here, what would actually be a good default value here? 🤔

holgerd77:
I have added the gas target checks being applied in the `BlockHeader.validate()` method in the `validateGasLimit()` method, this method would otherwise be shift to an undefined respectively useless behavior and now just continues to work post-1559.

This comes a bit to the expense of the expressiveness of the error messages, which I guess is bearable. Generally I guess we should move to directly throw within this function with more appropriate error messages, will add this to the v6 breaking change notes.

holgerd77:
Have streamlined the `BlockHeader.getBaseFee()` API in 28b8644, I guess we don't want to expose this with the need for passing in this single header fields.

holgerd77:
I guess we should further refine the getBaseFee() API and think the creation context along e.g. for the VM block builder, so we should minimally make this a static method I guess. //cc @ryanio

holgerd77:
Update: Ah, I think I got it, this makes more sense from an API PoV 62184de. There is just no need to call this on the block itself - no parameters are used - and the former implementation was also out of sync with the common instance using `common` from one block ahead (so the `common` from `this` while the `common` from the parent block would be the one setting the stage.

This should now be better usable in a block creation context, at least if one has the parent block as an object (if this is not usable yet in block builder we might combine all those efforts and decompose again to the single parameters and switch over to a `static public calcNextBaseFee(gasUsed, gasLimit, baseFee)` API.

packages/block/src/header.ts
```diff
@@ -442,21 +458,32 @@ export class BlockHeader {
...
+        gasLimit.gte(this._common.paramByHardfork('gasConfig', 'minGasLimit', hardfork))
+      )
+      // EIP-1559 gasTarget check
+    } else {
+      const parentGasTarget = parentBlockHeader.gasLimit
+      // check if the block changed the gas target too much
+      return (
+        this.gasLimit.lte(parentGasTarget.add(parentGasTarget.divn(1024))) &&
+        this.gasLimit.gte(parentGasTarget.sub(parentGasTarget.divn(1024)))
+      )
```
jochem-brouwer, line 485: Currently in a discussion but we can probably simplify this a lot; the logic is actually the same as the pre-1559 logic except that the "effective gas limit" is now the gasTarget. (So gas limit divided by 2)

holgerd77:
Rebased this on top of #1240 and #1239

packages/block/src/header.ts
holgerd77: Hmm, I would assume that this check should still be in place also with the spec adoption but then with `const parentGasTarget = parentBlockHeader.gasLimit / 2` (what would be the rounding here?)? 🤔

packages/block/test/eip1559block.spec.ts
```diff
@@ -0,0 +1,289 @@
...
+    try {
+      await block.validate(blockchain)
+      st.fail('should throw')
+    } catch (e) {
+      st.ok(e.message.includes('unable to pay base fee'), 'threw with right error')
+    }
+
+    st.end()
+  })
+})
```
holgerd77, line 289: Really great test lineup here! 😄

holgerd77 (commented):
Round 4.

Ok, I have now reviewed everything `Common` and `Block` related, will directly push change requests I had after commenting here.

Will continue the review for another 30-45 min, not sure if I'll do another push.

packages/vm/lib/runTx.ts
holgerd77: Will remove the redundant check on 2929 from here.

packages/vm/lib/runTx.ts
```diff
@@ -63,6 +66,14 @@ export interface RunTxOpts {
...
    */
   reportAccessList?: boolean
+
+  /**
+   * Optional clique Address: if the consensus algorithm is on clique,
+   * and this parameter is provided, use this as the beneficiary of transaction fees
+   * If it is not provided and the consensus algorithm is clique, instead
+   * get it from the block using `cliqueSigner()`
+   */
+  cliqueBeneficiary?: Address
```
holgerd77, line 76: Would it make sense to expand this option and allow to pass in a miner of clique beneficiary independent from the consensus type?

holgerd77:
Ok, I am also basically through with the `VM` changes.

Will stop for now.

packages/vm/lib/runTx.ts
```diff
@@ -136,12 +147,9 @@ export default async function runTx(this: VM, opts: RunTxOpts): Promise<RunTxRes
...
   await state.checkpoint()
 
-  // Is it an Access List transaction?
-  if (
-    'transactionType' in opts.tx &&
-    opts.tx.transactionType === 1 &&
-    this._common.isActivatedEIP(2929)
-  ) {
+  // Typed transaction specific setup tasks
+  if (opts.tx.transactionType !== 0 && this._common.isActivatedEIP(2718)) {
```
holgerd77, line 151: Ok, removing the EIP-2929 check was switching the error message - which I think is not good - so I added an EIP-2718 check instead. Also refined the surrounding comments a bit so that it gets more encouraged to use this whole section generally for type specific checks in the future and not only the AL things. This should fit better with the initial check on the type + the EIP check.

holgerd77:
@jochem-brouwer Ok, I can proudly say that I have now reviewed all the source code parts of all touched libraries including `tx`. 😄 

Please take everything I pushed along as "applied suggestions", so everything up for discussion and for some subsequent adaptation (or even some revert). Hope most things make sense though.

There are still 3-4 open things but all of this is "covered" by open comments here on the PR, I've also not set my "viewed" checkmark on where I still think there is something to be settled. Apart from that we should be very much good to go from a review PoV. 😀  (will nevertheless still makes a lot of sense if someone else wants to have an additional look from another perspective).

Ok. Then now there are the spec adoptions left, and likely more tests. Eventually I will also find some time tomorrow to write some more test cases.

But we are closing in I guess. 🎉  🥳

packages/block/src/block.ts
```diff
@@ -223,9 +229,20 @@ export class Block {
...
+    this.transactions.forEach((tx, i) => {
       const errs = <string[]>tx.validate(true)
+      if (self._common.isActivatedEIP(1559)) {
+        if (tx.transactionType === 2) {
+          if ((<FeeMarketEIP1559Transaction>tx).maxFeePerGas.lt(self.header.baseFeePerGas!)) {
+            errs.push('tx unable to pay base fee (EIP-1559 tx)')
+          }
+        } else {
+          if ((<Transaction>tx).gasPrice.lt(self.header.baseFeePerGas!)) {
+            errs.push('tx unable to pay base fee (non EIP-1559 tx)')
```
ryanio, line 242: removing `self` and using `this` now that the function was updated to arrow syntax

```suggestion
    this.transactions.forEach((tx, i) => {
      const errs = <string[]>tx.validate(true)
      if (this._common.isActivatedEIP(1559)) {
        if (tx.transactionType === 2) {
          if ((<FeeMarketEIP1559Transaction>tx).maxFeePerGas.lt(this.header.baseFeePerGas!)) {
            errs.push('tx unable to pay base fee (EIP-1559 tx)')
          }
        } else {
          if ((<Transaction>tx).gasPrice.lt(this.header.baseFeePerGas!)) {
            errs.push('tx unable to pay base fee (non EIP-1559 tx)')
```

packages/block/src/header.ts
```diff
@@ -187,7 +196,8 @@ export class BlockHeader {
     extraData: Buffer,
     mixHash: Buffer,
     nonce: Buffer,
-    options: BlockOptions = {}
+    options: BlockOptions = {},
+    baseFeePerGas?: BN
```
ryanio, line 200: I find this a bit inconsistent/confusing that `baseFeePerGas` is after the opts. I understand trying to minimize breaking changes though. Perhaps in the next round of breaking changes we can update this to `constructor(blockData, opts)` like the tx package so we can continue to add/remove/modify block fields while maintaining a consistent order of params.

packages/block/src/header.ts
```diff
@@ -208,6 +218,12 @@ export class BlockHeader {
       this._common.setHardforkByBlockNumber(number.toNumber())
     }
 
+    if (this._common.isActivatedEIP(1559) && baseFeePerGas === undefined) {
+      this.baseFeePerGas = new BN(1)
```
ryanio, line 222: is this default `1` value defined in the spec somewhere? can't seem to locate it. just wanted to verify it would be a proper default value. if undefined do we maybe want to set it to `INITIAL_BASE_FEE` instead? (in the spec as `1000000000`)

packages/block/src/header.ts
ryanio: I agree not adding an additional method to common if not needed, @jochem-brouwer what do you think about using @holgerd77's suggestion of `hardforkBlock`:

```typescript
const isInitialEIP1559Block = this._common.hardforkBlock('london').eq(this.number)
```

for testing purposes you can probably still stub it in the test suite itself.

packages/block/src/header.ts
```diff
@@ -546,13 +577,73 @@ export class BlockHeader {
...
+        const initialBaseFee = new BN(this._common.param('gasConfig', 'initialBaseFee'))
+        if (!this.baseFeePerGas!.eq(initialBaseFee)) {
+          throw new Error('Initial EIP1559 block does not have initial base fee')
+        }
+      } else {
+        // check if the base fee is correct
+        const expectedBaseFee = parentHeader.calcNextBaseFee()
+
+        if (!this.baseFeePerGas!.eq(expectedBaseFee)) {
+          throw new Error('Invalid block: base fee not correct')
```
ryanio, line 601: should we add some code coverage for these lines?

packages/block/src/header.ts
```diff
@@ -569,6 +660,12 @@ export class BlockHeader {
       this.mixHash,
       this.nonce,
     ]
+
+    if (this._common.isActivatedEIP(1559)) {
+      rawItems.push(unpadBuffer(toBuffer(this.baseFeePerGas)))
```
ryanio, line 665: since `baseFeePerGas` is already `BN` we don't need to use `toBuffer`, we can do:

```suggestion
      rawItems.push(unpadBuffer(this.baseFeePerGas.toArrayLike(Buffer)))
```

or use the utility function I added to util (we can rename this to `bnToUnpaddedBuffer` as per your issue https://github.com/ethereumjs/ethereumjs-util/issues/284 and maintain an alias from the old name for backwards compat):

```suggestion
      rawItems.push(bnToRlp(this.baseFeePerGas))
```

packages/common/src/eips/1559.json
ryanio: this is the issue you mentioned jochem that the spec was updated to match all the clients' technically wrong but matching implementation, right? 😄 https://github.com/ethereum/EIPs/pull/3556

packages/devp2p/src/eth/index.ts
ryanio: nice fix. I think I would prefer if `_nextForkBlock` was null or undefined rather than zero for this check if there is no next fork coming, but that is a small nit. I'm not exactly sure what other repercussions that may have in the code base, `0` could very much be the better choice other than this buffer conversion issue.

packages/tx/src/eip1559Transaction.ts
```diff
@@ -0,0 +1,353 @@
...
+  public readonly chainId: BN
+  public readonly accessList: AccessListBuffer
+  public readonly AccessListJSON: AccessList
+  public readonly maxInclusionFeePerGas: BN
+  public readonly maxFeePerGas: BN
+
+  /**
+   * EIP-2930 alias for `r`
+   */
+  get senderR() {
```
ryanio, line 27: is there a reason to continue to add these aliases? i think holger alternatively suggested to just move them to baseTransaction so they are usable by all.

packages/tx/src/eip1559Transaction.ts
```diff
@@ -0,0 +1,353 @@
...
+    return new FeeMarketEIP1559Transaction(txData, opts)
+  }
+
+  /**
+   * Instantiate a transaction from the serialized tx.
+   *
+   * Note: this means that the Buffer should start with 0x01.
+   */
+  public static fromSerializedTx(serialized: Buffer, opts: TxOptions = {}) {
+    if (!serialized.slice(0, 1).equals(TRANSACTION_TYPE_BUFFER)) {
```
ryanio, line 55: nice check, i think we can add some coverage for this along with the other coverage check warnings below

packages/tx/src/types.ts
```diff
@@ -126,14 +145,31 @@ export interface AccessListEIP2930TxData extends TxData {
...
+   */
+  maxInclusionFeePerGas?: BNLike
+  /**
+   * The maximum total fee
+   */
+  maxFeePerGas?: BNLike
 }
 
+/**
+ * Buffer values array for EIP2930 transaction
```
ryanio, line 169: not just for EIP2930, right?

```suggestion
 * Buffer values array for transactions
```

packages/block/src/header.ts
```diff
@@ -208,6 +218,12 @@ export class BlockHeader {
       this._common.setHardforkByBlockNumber(number.toNumber())
     }
 
+    if (this._common.isActivatedEIP(1559) && baseFeePerGas === undefined) {
+      this.baseFeePerGas = new BN(1)
+    } else {
+      this.baseFeePerGas = undefined
```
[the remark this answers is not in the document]
holgerd77: Yes, I added this when refactoring out the public `BlockHeader.getCommon()` class in 502b9b2, since the solution before needed a final `Common` instance before instantiation which has stronger side effects and is not desirable. The exact length check was then not possible any more and this setting to `undefined` was therefore for the case where a `baseFeePerGas` was passed in for a non-1559 block.

I will update though with a check to just throw here in this case, this is likely a lot better than implicitly overwrite. And then we completely match the functionality we had before with Jochem's implementation without the `Common` side effects.

(will do on next push).

packages/block/src/header.ts
```diff
@@ -469,13 +497,16 @@ export class BlockHeader {
...
+    // Consensus type dependent checks
+    // PoW/Ethash
     if (this._common.consensusAlgorithm() !== 'clique') {
       if (
         this.extraData.length > this._common.paramByHardfork('vm', 'maxExtraDataSize', hardfork)
       ) {
         const msg = 'invalid amount of extra data'
         throw this._error(msg)
       }
+      // PoA/Clique
```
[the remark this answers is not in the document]
holgerd77: I've always had my problems placing comments on `if/else` clauses, this might be a good convention 🙂 , will update.

packages/block/src/header.ts
holgerd77: We generally have to lift up coverage significantly for the PR - yes. Will therefore mark as resolved here, we just should generally look carefully through the Codecov reports.

packages/block/src/block.ts
holgerd77: I think Jochem decided somewhat explicitly against that solution since it would produce a lot of overhead with again the need for some factory class and all that stuff similarly to what has been done for the tx library, I would agree that it's not worth the effort right now for just a single field. Would be very much desirable though nonetheless here to come to a cleaner solution.

packages/block/src/header.ts
holgerd77: Update: this actually has side effects on instantiation and some tests fail after the change. Will revert locally since I haven't got the time to dig deeper, I think a revert is safe though since this just prevents some other checks in the  main constructor (which would throw) to be reached. Feel free to dig deeper here though.

packages/tx/src/types.ts
holgerd77: Ah, forgot to update, thanks, this is actually *only* for the legacy txs - since we still have got this situation that we have all things "legacy" just named "Transaction" or "Tx" and I wanted to remain consistent. So even more important to comment correctly. Will update.

packages/block/src/header.ts
holgerd77: Missed the comment here, ok, I will the remove the added method from `Common`.

packages/vm/lib/runTx.ts
holgerd77: Ah, this is still an interesting/important question it would be good to get additional opinions on and not just let this slip through. //cc @jochem-brouwer @ryanio

jochem-brouwer: I am not super sure what you mean by this. The clique beneficiary is actually not present in the header of the block (miner is always set to the zero address IIRC). To get the actual clique beneficiary, you have to get the address which signed the extraData of the block.

packages/tx/src/eip1559Transaction.ts
jochem-brouwer: I have no strong opinion here, what we want to do in the end is see what other client use for field names and also start using those. Alternatively, we could also use the names as specified in the EIPs.

packages/vm/lib/runTx.ts
holgerd77: I mean if we should allow here to generally add the `minerOrBeneficiary` as this option and not only for the clique case.

This option is used in the current form:

```typescript
if (this._common.consensusType() === 'pow') {
    miner = block.header.coinbase
  } else {
    // Backwards-compatibilty check
    // TODO: can be removed along VM v5 release
    if (opts.cliqueBeneficiary) {
      miner = opts.cliqueBeneficiary
    } else {
      if ('cliqueSigner' in block.header) {
        miner = block.header.cliqueSigner()
      } else {
        miner = Address.zero()
      }
    }
  }
```

So it also might make sense to use this in the PoW case and if present than set the `miner` to the miner from this option and not to the eventual default value from `block.header.coinbase`.

Does this makes sense? I am not saying that we *should* do this, just saying that this might be a worthwhile generalization of the option which we should at least consider.

holgerd77:
Yeah yeah, everything in the green! 😄 

The last 5% were easy, this was actually just adding the 1559 tx test file to the test runner, which had been forgotten before. 😂

Mr Brouwer, will hand this over back to you, Sir. 🙂

The test work I've done was mainly broad-layouted ground work, so reworking various test files and added missing - mostly generic, sometimes somewhat specific - cases (and this was already enough work, close to two workdays actually for bringing such a large PR from a 50% diff to within the threshold, impressive experience, wouldn't have thought this would take *soooo* long).

What it did *not* yet is adding some very targeted EIP-1559 test cases around the edge cases, last spec discussions,... around the EIP. This should likely still be expanded along the work on the latest/last (hopefully) spec updates.

holgerd77:
Ok, this is now syncing the current Baikal chain (up to the latest block 14408). 😄  🎉

holgerd77:
Guys, I will now merge this in here, tests are passing and I've given all code parts another final look. I have also written extensive tests for the tx library to lift the coverage up again to a decent level and have given all changes in the various libraries at least a rough and often somewhat thorough review.

We also do not have an imminently urgent release, at the same time more and more is worked "around" this one huge PR and all the other smaller work is blocked to a greater extend.

Everyone can still do reviews on parts where he feels the need, then please integrate your change requests along follow-up PRs.

Hope you can go along, but I have once again the feeling that we are at some point where things are not getting better by pushing this ahead. 😋  😀

Thanks @jochem-brouwer for this great work here! 🎉

holgerd77 (approved):
LGTM

holgerd77:
Phew, 86 commits (and all very much decent ones, and not "fixed linting", "minor changes", "forgot to add x" 😛 ).

packages/vm/lib/runTx.ts
ryanio: hey, catching up on this thread now.

ok so I see this problem was introduced when I moved the tx receipt generation to runTx, where we need a block with the current accurate gasUsed to generate the right receipt.

from @jochem-brouwer ([comment link](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1148#issuecomment-827139618)):
> Ok, this commit 8c112f0 fixes the consensus bug at block 55. The problem had nothing to do with EIP1559, it had to do with runBlock/runTx in VM: runBlock originally fed runTx an edited block than the block which is actually being ran; the gasUsed field is changed and used to keep track of the cumulative gas used. This commit is a hotfix and might be changed in the future (probably removing editing the block and just passing along the block which is actually being ran in the context).
> 
> We should add a test to ensure that clique blocks with txs pay the right miner, which is not tested in VM currently.

I can foresee future problems with this "modified block" being passed in and think it would be good to go back to the original block without any fancy meddling. Thinking about it with a fresh perspective now, what do you guys think if I add a `blockGasUsed` parameter to runTx instead, expressly for accurate tx receipt generation? Then we don't have modify the block for this reason.

I'm also confused which part of the code this todo references now:
`// TODO: can be removed along VM v5 release`
could you help clarify @holgerd77? I can add to the PR I would like to open for the work above.

holgerd77: > Thinking about it with a fresh perspective now, what do you guys think if I add a blockGasUsed parameter to runTx instead, expressly for accurate tx receipt generation? Then we don't have modify the block for this reason.

I think that's a good idea.

The TODO was for `if ('cliqueSigner' in block.header)`, this was the discussion that this would break if used with a `Block` library which hasn't included the clique changes yet. (I think v5 might be a mistake and this should be v6?)

> In conclusion, if my suggestion above is acceptable, we can remove this newly introduced cliqueBeneficiary option in runTx in favor of a blockGasUsed parameter and i can add some tests to ensure the beneficiary is correctly derived.

Yeah, sounds great, leaves me somewhat relieved.

😄

packages/tx/src/types.ts
```diff
@@ -139,6 +139,10 @@ export interface AccessListEIP2930TxData extends TxData {
 }
 
 export interface FeeMarketEIP1559TxData extends AccessListEIP2930TxData {
+  /**
+   * The transaction's gas price.
+   */
+  gasPrice?: never
```
jochem-brouwer, line 157: Yup this is great!

packages/tx/src/eip2930Transaction.ts
jochem-brouwer: Is checked; https://github.com/ethereumjs/ethereumjs-monorepo/blob/a1e7a34d989c7991454c06567573608f33caecbf/packages/tx/src/baseTransaction.ts#L69

packages/tx/src/eip1559Transaction.ts
```diff
@@ -196,6 +196,9 @@ export default class FeeMarketEIP1559Transaction extends BaseTransaction<FeeMark
    * @param baseFee The base fee of the block
    */
   getUpfrontCost(baseFee?: BN): BN {
+    if (!baseFee) {
+      baseFee = new BN(0)
```
jochem-brouwer, line 201: Could have defaulted this in the method argument.

packages/tx/typedoc.js
```diff
@@ -4,7 +4,10 @@ module.exports = {
   plugin: 'typedoc-plugin-markdown',
   readme: 'none',
   gitRevision: 'master',
-  exclude: 'test/**/*.ts',
+  exclude: [
+    'src/util.ts',
```
jochem-brouwer, line 8: Why do we exclude this?

packages/block/test/eip1559block.spec.ts
```diff
@@ -22,10 +22,14 @@ const genesis = Block.fromBlockData({})
...
 // Small hack to hack in the activation block number
 // (Otherwise there would be need for a custom chain only for testing purposes)
-common.getEIPActivationBlockNumber = function (eip: number) {
-  if (eip == 1559) {
+common.hardforkBlockBN = function (hardfork: string | undefined) {
+  if (hardfork === 'london') {
     return new BN(1)
+  } else if (hardfork === 'dao') {
+    // Avoid DAO HF side-effects
+    return new BN(99)
```
jochem-brouwer, line 31: What are these side effects?

packages/client/lib/blockchain/chain.ts
```diff
@@ -103,14 +103,19 @@ export class Chain extends EventEmitter {
...
 
     this.blockchain =
       options.blockchain ??
       new Blockchain({
         db: options.chainDB,
         common: this.config.chainCommon,
+        hardforkByHeadBlockNumber: true,
         validateBlocks: true,
-        validateConsensus: true,
+        validateConsensus,
```
jochem-brouwer, line 118: I know we had this discussion earlier, but why disable the consensus validation? Now everyone can feed us bogus PoW blocks which we will then execute. I understand that the PoW validation takes time, but we should really validate it in our client.

ryanio: could we solve this by using [node-ethash](https://github.com/ethereumjs/node-ethash#performance-node-ethash-vs-ethashjs) in the client?

packages/tx/src/eip1559Transaction.ts
ryanio: good idea, i've added it to [Tiny Fixes Meta Issue](https://github.com/ethereumjs/ethereumjs-monorepo/issues/905)

ryanio: also on the two lines underneath, baseFee doesn't need the non-null assertion operator (`!`)

packages/client/lib/blockchain/chain.ts
jochem-brouwer: Hmm that's interesting, I didn't even know we had this. The performance boost is insane, we should definitely look into this (possibly we should create a general Ethash interface if this is not yet done, and then let our users choose downstream if they want to use Ethashjs or node-ethash)

packages/tx/typedoc.js
holgerd77: The API of this feels too unfinished to me and I would like to keep this internal for now and give this some more thought and have some more discussions around it. I think no one every really reviewed these code parts (including me) and this would expose a huge new API surface.

Some things I would like to discuss (but which doesn't need to be discussed now, just for the mention):

- Is this the best way to expose this functionality or might there be better - eventually more integrated - solutions (e.g. to re-bind the methods to the affected tx classes, just as some first idea)?
- Relationship of this `getAccessListData()` (returning `AccessListJSON` as one of two items) and the `getAccessListJSON()` method
- Should `getDataFeeEIP2930()` be exposed (I have the base feeling that methods which need a `Common` as an input often point to some suboptimality in design - we had these cases in the past - can't really point towards a concrete solution thought atm yet)? 

I will actually open this as an issue, otherwise it will likely be too tempting to start the discussion here. 😄

packages/tx/src/eip2930Transaction.ts
holgerd77: Yes, I moved this from `EIP2930Transaction` to `BaseTransaction` some commits later on. 😄

packages/block/test/eip1559block.spec.ts
holgerd77: Tests were breaking due to the DAO HF being triggered on some test block(s) where block validation would then fail.

packages/client/lib/blockchain/chain.ts
holgerd77: I have no idea in what state the `note-ethash` library is actually. Will open a dedicated issue on this.

PR #1149, opened 2021-03-15
Fix devp2p browser compatibility 

cgewecke:
When the client is launched in the browser, an error is thrown from devp2p because Node's `dns` module is not automatically polyfilled. @ryanio and I looked for a standard replacement without success.

PR creates a shim class in devp2p which is loaded if dns isn't require-able. Its methods throw with an error advising user to disable dns discovery via the DPTOptions.

Loading the client browser example looks ok:

Screen Shot 2021-03-15 at 10 22 18 AM

Also checked that DNS discovery at the terminal still works:
```
cgewecke$ DEBUG=devp2p:dns:dns npm run client:start -- --discDns=true discV4=false
...
INFO [03-15|10:49:43] Listener up transport=rlpx url=enode://8110042d9871729a456b53a99c3ce0e20d1f55aebd578b75c8f9c95ee00992e2eb0ae7ce047002a86e06c554d8189d1ccd9a7c9e001a57bb20400f798ff477d2@[::]:30303 
INFO [03-15|10:49:43] Listener up transport=libp2p url=/ip4/127.0.0.1/tcp/50580/ws/p2p/12D3KooWPUB4TPcEQmaf1uY2vxWeuK8FodLfZZLWMieYQ7kz8XK3 
  devp2p:dns:dns got new peer candidate from DNS address=162.239.57.201 +0ms
  devp2p:dns:dns got new peer candidate from DNS address=107.185.23.77 +705ms
  devp2p:dns:dns got new peer candidate from DNS address=89.136.52.246 +1s
```


ryanio (approved):
nice solution, lgtm!

cgewecke:
Merging because this is pre-req to verify that other fixes are not breaking the browser example. 

(Am happy to revisit here if this could be improved though...)

PR #1150, opened 2021-03-15
Common: add BN support

### LINKED ISSUE #914: BN support in Common
Lots of methods in Common have to do with numbers. We sometimes convert BNs to numbers (for instance block numbers) and feed those to Common. Common should support inputs where this input is of type BN and should also handle arithmetic as BN. We can keep supporting numbers by internally converting it to BN and then feeding it to the same method which handles the input as BN.

ryanio:
This PR adds big number support to Common by updating functions with `number` input types to `BNLike`.

See [this comment](https://github.com/ethereumjs/ethereumjs-monorepo/issues/914#issuecomment-794557862) for the full list of functions changed.

For functions containing `number` outputs, alternative methods have been added (e.g. `chainId` -> `chainIdBN`) and the original methods have been deprecated with suggestion to use the BN alternatives for large number support.

Closes #914


holgerd77:
@ryanio oh sorry, not sure if you are still working on this here, just updated the branch (by merge), if this conflicts with anything you do locally, just force-push.

ryanio:
@holgerd77 no problem, thanks, yes just updating some of the dependent packages now based on these new changes, will follow as additional commits

packages/common/src/index.ts
```diff
@@ -188,17 +191,17 @@ export default class Common extends EventEmitter {
    * @param blockNumber
    * @returns The name of the HF
    */
-  getHardforkByBlockNumber(blockNumber: number): string {
+  getHardforkByBlockNumber(blockNumber: BNLike): string {
+    blockNumber = toType(blockNumber, TypeOutput.BN)
```
holgerd77, line 195: Oh, yeah, that plays out really nice here. 😄

packages/common/src/index.ts
```diff
@@ -442,9 +448,8 @@ export default class Common extends EventEmitter {
   hardforkGteHardfork(
     hardfork1: string | null,
     hardfork2: string,
-    opts?: hardforkOptions
+    opts: hardforkOptions = {}
   ): boolean {
-    opts = opts !== undefined ? opts : {}
     const onlyActive = opts.onlyActive === undefined ? false : opts.onlyActive
     hardfork1 = this._chooseHardfork(hardfork1, opts.onlySupported)
```
holgerd77, line 454: Side note, just because I am seeing it here: we might want to consider to remove onlySupported and/or onlyActive HF options on the next breaking release - this is not really used (maybe in the VM, but I would think this is rather not necessary to be managed in Common) - have added a reminder note to v6 Release planning.

packages/tx/src/legacyTransaction.ts
```diff
@@ -210,8 +210,9 @@ export default class Transaction extends BaseTransaction<Transaction> {
    * Process the v, r, s values from the `sign` method of the base transaction.
    */
   protected _processSignature(v: number, r: Buffer, s: Buffer) {
+    const vBN = new BN(v)
     if (this._unsignedTxImplementsEIP155()) {
-      v += this.common.chainId() * 2 + 8
+      vBN.iadd(this.common.chainIdBN().muln(2).addn(8))
```
holgerd77, line 215: 👍

holgerd77 (approved):
Looks great, also with this deep-reaching updates on the consuming libraries! 😄  👍  🎉 

Will approve and merge.

packages/client/lib/net/protocol/ethprotocol.ts
```diff
@@ -136,7 +136,7 @@ export class EthProtocol extends Protocol {
   encodeStatus(): any {
     // TODO: add latestBlock for more precise ETH/64 forkhash switch
     return {
-      networkId: this.chain.networkId,
+      networkId: this.chain.networkId.toNumber(),
```
holgerd77, line 139: This change is currently breaking the client CLI start on Yolo v3.

holgerd77: Just checked, for RLPx the `networkId` can actually be removed completely from here, since the `devp2p` library is not using this passed in `networkId` any more but [using the networkId from Common directly](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/devp2p/src/eth/index.ts#L197).

Not sure about libp2p though.

PR #1151, opened 2021-03-16
Berlin Releases

holgerd77:


holgerd77:
91 files changed and counting. Ugh. This is getting really dense.

holgerd77:

(which doesn't mean that we could not take 1-3 other changes and/or fixes in, it would just be good if this would be reviewed early on so that this can then easily be re-approved upon an eventual rebase)

holgerd77:
Files changed: 119. Argh. 😁

packages/block/CHANGELOG.md
```diff
@@ -6,9 +6,35 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
+#### EthereumJS Libraries - Typed Transactions Readiness
+
+If you are using this library in conjunction with other EthereumJS libraries make sure to minimally have the following library versions installed for typed transaction support:
+
+- `@ethereumjs/common` `v2.2.0`
+- `@ethereumjs/tx` `v3.1.0`
+- `@ethereumjs/block` `v3.2.0`
+- `@ethereumjs/blockchain` `v5.2.0`
+- `@ethereumjs/vm` `v5.2.0`
+
```
cgewecke, line 33: Nice 👍

packages/common/src/eips/2930.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "EIP-2929",
+    "name": "EIP-2930",
```
cgewecke, line 2: Nice catch!

packages/tx/README.md
```diff
@@ -17,7 +17,75 @@ Note: this `README` reflects the state of the library from `v3.0.0` onwards. See
...
+      ]
+    }
+  ],
+  "type": "0x01"
+}
+
+const tx = AccessListEIP2930Transaction.fromTxData(txData, { common })
+```
+
+A mechanism to generate access lists from tx data based on a certain network state in not part of this library.
```
cgewecke, line 80: Is there any advice that can be given about how to identify storage slots?

cgewecke (approved):
This is huge! Everything LGTM 💯  

Really exciting release I think. Looks great! 

(Down to just a handful of test failures in the Hardhat upgrade - am not anticipating any last minute surprises coming from over there fwiw.)

packages/tx/CHANGELOG.md
```diff
@@ -6,16 +6,68 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
+}
+
+const tx = AccessListEIP2930Transaction.fromTxData(txData, { common })
+```
+
+Please note that the default HF is still set to `istanbul`. You therefore need to explicitly set the `hardfork` parameter for instantiating a `Transaction` instance with a `berlin` HF activated.
+
+The now called "legacy" transactions are still supported and can be used as before by using the `Transaction` class. If the type of a tx is only known at runtime there is a new `TransactionFactory` class introduced for your convenience. This factory class decides on the tx type based on the input data and uses the corresponding tx type class for instantiation.
+
+For more guidance on how to use the new tx types and the tx factory have a look at the [README](./README.md) of this library which has perceived an extensive update along with this release.
```
ryanio, line 53:
```suggestion
For more guidance on how to use the new tx types and the tx factory have a look at the [README](./README.md) of this library which has received an extensive update along with this release.
```

packages/tx/README.md
ryanio, line 80:
```suggestion
A mechanism to generate access lists from tx data based on a certain network state is not part of this library.
```

ryanio: @cgewecke it is possible to derive the storage keys for a contract from its own code (check out [this example](https://eth.wiki/json-rpc/API#example-14)), however getting the access list for a more complex transaction can be tricky as it can change based on the latest network state.

packages/tx/README.md
```diff
@@ -44,7 +112,27 @@ const signedTx = tx.sign(privateKey)
...
+
+const common = new Common({ chain: 'mainnet', hardfork: 'berlin' })
+
+const txData = {} // Use data from the different tx type examples
+const tx = TransactionFactory.fromTxData(txData, { common })
+```
+
+The correct tx type class for instantiation will then be chosen on runtime based on the data provided as an input.
+
+`TransactionFactory` support the following static constructor methods:
```
ryanio, line 131:
```suggestion
`TransactionFactory` supports the following static constructor methods:
```

ryanio:
looks great! found just a few small typos but other than that great job adding thorough notes to the changelogs and readmes :)

packages/tx/README.md
ryanio: I definitely think we should introduce some tools to help generate access lists though, I expect that it will be requested if it helps lower the cost of transactions. Perhaps it's better suited at the library layer in web3.js and ethers where there is a connection to an Ethereum node.

Do we think there will be JSON-RPC endpoints or helper tools from wallets like MetaMask to generate contract transactions containing access lists? Etherscan would actually be in a really good position to generate them when creating a tx on their "write contract" page.

holgerd77: I think we'll relatively for sure add this as a feature to the VM - see also https://github.com/ethereumjs/ethereumjs-monorepo/issues/1152 - since this is just dropping out there on the sideline and *will* likely also have it's use cases, not sure if it is applicable for these kind of usages since there is still this "where to get the state from" problem?

Would this be another case for an RPC-based `StateManager` (this was in discussion sometime ago somewhere 😛 ) retrieving all the state by RPC on demand? This would be really cool in general, thinking about this once again (and actually might also work for use cases like MetaMask I would say for generating access lists?).

holgerd77:
@ryanio thanks 🥳 , do we want to add your block generation method in the release as well or do you rather want to give this a bit more time?

holgerd77:
BlockBuilder still needs some work #1158, I would then continue here with the planned release schedule and target a release for tomorrow. Can someone give this a final review? 😄

holgerd77:
(respectively eventually just re-approve, was already reviewed by @cgewecke but I dismissed by applying some typo corrections (these situation are always a bit annoying, on the one side on does not want to leave this in the code, on the other hand it's a bit laborious to have this whole process triggered again. Not sure what's the solution for this)

cgewecke (approved):
🚀

304s304:
> []()

holgerd77:
Just published the following releases on npm:

- `@ethereumjs/common` [v2.2.0](https://github.com/ethereumjs/ethereumjs-monorepo/releases/tag/%40ethereumjs%2Fcommon%402.2.0)
- `@ethereumjs/tx` [v3.1.0](https://github.com/ethereumjs/ethereumjs-monorepo/releases/tag/%40ethereumjs%2Ftx%403.1.0)
- `@ethereumjs/block` [v3.2.0](https://github.com/ethereumjs/ethereumjs-monorepo/releases/tag/%40ethereumjs%2Fblock%403.2.0)
- `@ethereumjs/blockchain`[v5.2.0](https://github.com/ethereumjs/ethereumjs-monorepo/releases/tag/%40ethereumjs%2Fblockchain%405.2.0)
- `@ethereumjs/vm` [v5.2.0](https://github.com/ethereumjs/ethereumjs-monorepo/releases/tag/%40ethereumjs%2Fvm%405.2.0)

PR #863, opened 2020-09-07
Chain setup API / Centralize default HF / Common options dict / Block HFbyBlockNumber & initWithGenesisHeader options

holgerd77:
This PR prepares for the rework of EIP support as a follow-up on #856 

I moved Common to be instantiated with an option dictionary - like we do in most of the other libraries too - should be a good occasion to do this now since we are right before a breaking release anyhow.

This prepares for a new to be added `eips` option, things would otherwise slowly getting too ugly with all the optional parameters chained. It also makes the library more future proof by preparing the ground for future non-breaking option additions.

First push here to see if all tests in related consuming libraries pass.

PR #1153, opened 2021-03-17
Make genesis params settable for any hardfork in block headers

cgewecke:
PR removes a restriction introduced in #863 that disallows commons with custom genesis settings unless their hardforks are set to `chainstart`. 

This is blocking integration of the `block` module in the Hardhat testrpc because they set a custom common with special gas limits, chainId, etc as an initial step on launch, but need to run on arbitrary hardforks.

https://github.com/nomiclabs/hardhat/blob/579d44a105d6a8323d270ea55876c6694c212316/packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeCommon.ts#L20-L42

Perhaps this restriction should be preserved but gated behind a flag - not sure. Am opening in the simplest form for further discussion.

### DIFF
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -209,11 +209,6 @@ export class BlockHeader {
     }
 
     if (options.initWithGenesisHeader) {
-      if (this._common.hardfork() !== 'chainstart') {
-        throw new Error(
-          'Genesis parameters can only be set with a Common instance set to chainstart'
-        )
-      }
       number = new BN(0)
       if (gasLimit.eq(DEFAULT_GAS_LIMIT)) {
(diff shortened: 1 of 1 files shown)


holgerd77:
On thinking about this I have the impression this restriction is just plain wrong and is rather a bug fix. I hadn't it on the radar that several hardforks can be applied together on one block with the special case of this happening on the genesis block (as it is the case for `goerli` e.g.).

So seems to safe to me to merge "as is".

holgerd77:
(also recently had to fix some similar - or at least somewhat related bug - on some hardfork determination logic in `Common`)

holgerd77 (approved):
LGTM

PR #1156, opened 2021-03-18
Client: small BN fixes

ryanio:
This PR is a follow-up to #1150 that fixes some BN cases where toNumber should not be used in case of large number network ids like yolov3.

### DIFF
--- a/packages/client/lib/net/protocol/lesprotocol.ts
+++ b/packages/client/lib/net/protocol/lesprotocol.ts
@@ -174,10 +174,10 @@ export class LesProtocol extends Protocol {
     }
 
     return {
-      networkId: this.chain.networkId.toNumber(),
+      networkId: this.chain.networkId.toArrayLike(Buffer),
       headTd: this.chain.headers.td.toArrayLike(Buffer),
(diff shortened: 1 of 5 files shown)


packages/client/lib/net/protocol/lesprotocol.ts
```diff
@@ -174,10 +174,10 @@ export class LesProtocol extends Protocol {
     }
 
     return {
-      networkId: this.chain.networkId.toNumber(),
+      networkId: this.chain.networkId.toArrayLike(Buffer),
       headTd: this.chain.headers.td.toArrayLike(Buffer),
       headHash: this.chain.headers.latest?.hash(),
-      headNum: this.chain.headers.latest?.number,
+      headNum: this.chain.headers.latest?.number.toArrayLike(Buffer),
```
ryanio, line 180: I believe encoding this BN as a Buffer is more correct, according to the [les protocol messages spec](https://github.com/ethereum/devp2p/blob/master/caps/les.md#status-0x00).

holgerd77 (approved):
Nice, LGTM! 😄

PR #1158, opened 2021-03-20
VM: Add vm.buildBlock() and BlockBuilder API

### LINKED ISSUE #1018: New VM API method for Block generation VM.generateBlock()
The [VM.runBlock()](https://github.com/ethereumjs/ethereumjs-vm/blob/master/packages/vm/lib/runBlock.ts) method is meant to run on existing blocks for validation and the API is not really meant for creating a new block, there is some functionality which goes into the direction with the `generate` option, but this functionality is generally problematic by leading to a situation where the block immutability is violated or some unclean workaround needs to be added - see also exchange about this on https://github.com/ethereumjs/ethereumjs-vm/pull/974.

I would suggest to add a new API function which dedicatedly covers the case of Block generation, something like:

```typescript
VM.generateBlock(parentBlock: Block, transactions: Transaction[], options: Opts)
```

This would explicitly targets the case where a new block is just not existing yet but needs to be calculated and processed with Txs from the mempool or whereever. This would also prepare for future client functionality on block creation.

Eventually this can be implemented reusing functionality from `VM.runBlock()`.

ryanio:
This PR adds `vm.buildBlock()` for building a block and adding transactions one at a time. 

It can be used like so:

```typescript
const blockBuilder = await vm.buildBlock({ parentBlock })
const txResult = await blockBuilder.addTransaction(tx)
const block = await blockBuilder.build()
```

It creates a checkpoint on the StateManager and modifies the state as transactions are run. The checkpoint is committed on `build()` or discarded with `revert()`.

Closes #1018.

This PR also:
* includes fixes for `runBlock` with `generate: true` by setting the header fields for gasUsed, logs bloom, receiptTrie, and transactionsTrie. (closes #1037)
* moves the tx receipt generation logic in `runBlock` to a function inside the same file called `generateTxReceipt`, which is then exported and used in BlockBuilder for generating the receiptTrie.

WIP, still finishing some last spec tests but sharing for initial review.

### DIFF
--- a/packages/vm/lib/buildBlock.ts
+++ b/packages/vm/lib/buildBlock.ts
@@ -71,7 +71,7 @@ export class BlockBuilder {
   /**
    * Throws if the block has already been built or reverted.
    */
-  checkStatus() {
+  private checkStatus() {
     if (this.built) {
       throw new Error('Block has already been built')
     }
--- a/packages/vm/lib/buildBlock.ts
+++ b/packages/vm/lib/buildBlock.ts
@@ -1,0 +1,250 @@
+import { Address, BN, toBuffer } from 'ethereumjs-util'
+import { encode } from 'rlp'
+import { BaseTrie as Trie } from 'merkle-patricia-tree'
+import { TypedTransaction } from '@ethereumjs/tx'
+import { Block, BlockOptions, HeaderData } from '@ethereumjs/block'
+import VM from '.'
+import { RunTxResult } from './runTx'
+import Bloom from './bloom'
+import { generateTxReceipt, calculateMinerReward, rewardAccount } from './runBlock'
+
+/**
+ * Options for building a block.
+ */
+export interface BuildBlockOpts {
+  /**
+   * The parent block
+   */
+  parentBlock: Block
+
+  /**
+   * The block header data to use.
+   * Defaults used for any values not provided.
+   */
+  headerData?: HeaderData
+
+  /**
+   * The block options to use.
+   */
+  blockOpts?: BlockOptions
+}
+
+/**
+ * Options for sealing a block.
+ */
+export interface SealBlockOpts {
+  /**
+   * For PoW, the nonce.
+   * Overrides the value passed in the constructor.
+   */
+  nonce?: Buffer
+
+  /**
+   * For PoW, the mixHash.
+   * Overrides the value passed in the constructor.
+   */
+  mixHash?: Buffer
+
+  /**
+   * For PoW, the difficulty.
+   * Overrides the value passed in the constructor.
+   */
+  difficulty?: BN
+
+  /**
+   * For PoA, the private key for the clique signer.
+   * Overrides the value passed in the constructor.
+   * If not provided, the block will not be sealed.
+   */
+  cliqueSigner?: Buffer
+}
+
+export class BlockBuilder {
+  private readonly vm: VM
+  private blockOpts: BlockOptions
+  private headerData: HeaderData
+  private transactions: TypedTransaction[] = []
+  private transactionResults: RunTxResult[] = []
+  private checkpointed = false
+  private reverted = false
+  private built = false
+
+  constructor(vm: VM, opts: BuildBlockOpts) {
+    this.vm = vm
+    this.blockOpts = { ...opts.blockOpts, common: this.vm._common }
+
+    this.headerData = {
+      ...opts.headerData,
+      parentHash: opts.headerData?.parentHash ?? opts.parentBlock.hash(),
+      number: opts.headerData?.number ?? opts.parentBlock.header.number.addn(1),
+      gasLimit: opts.headerData?.gasLimit ?? opts.parentBlock.header.gasLimit,
+    }
+  }
+
+  /**
+   * Throws if the block has already been built or reverted.
+   */
+  checkStatus() {
+    if (this.built) {
+      throw new Error('Block has already been built, please use a new BlockBuilder')
+    }
+    if (this.reverted) {
+      throw new Error('State has already been reverted, please use a new BlockBuilder')
+    }
+  }
+
+  /**
+   * Returns the cumulative gas used by the transactions added to the block.
+   */
+  gasUsed(): BN {
+    return this.transactionResults.reduce(
+      (accumulator: BN, result: RunTxResult) => accumulator.iadd(result.gasUsed),
+      new BN(0)
+    )
+  }
+
+  /**
+   * Calculates and returns the transactionsTrie for the block.
+   */
+  private async transactionsTrie() {
+    const trie = new Trie()
+    for (const [i, tx] of this.transactions.entries()) {
+      await trie.put(encode(i), tx.serialize())
+    }
+    return trie.root
+  }
+
+  /**
+   * Calculates and returns the logs bloom for the block.
+   */
+  private bloom() {
+    const bloom = new Bloom()
+    for (const txResult of this.transactionResults) {
+      // Combine blooms via bitwise OR
+      bloom.or(txResult.bloom)
+    }
+    return bloom.bitvector
+  }
+
+  /**
+   * Calculates and returns the receiptTrie for the block.
+   */
+  private async receiptTrie() {
+    const gasUsed = new BN(0)
+    const receiptTrie = new Trie()
+    for (const [i, txResult] of this.transactionResults.entries()) {
+      const tx = this.transactions[i]
+      gasUsed.iadd(txResult.gasUsed)
+      const { encodedReceipt } = await generateTxReceipt.bind(this.vm)(tx, txResult, gasUsed)
+      await receiptTrie.put(encode(i), encodedReceipt)
+    }
+    return receiptTrie.root
+  }
+
+  /**
+   * Rewards the miner.
+   */
+  private async rewardMiner() {
+    const minerReward = new BN(this.vm._common.param('pow', 'minerReward'))
+    const reward = calculateMinerReward(minerReward, 0)
+    const coinbase = this.headerData.coinbase
+      ? new Address(toBuffer(this.headerData.coinbase))
+      : Address.zero()
+    await rewardAccount(this.vm.stateManager, coinbase, reward)
+  }
+
+  /**
+   * Run and add a transaction to the block being built.
+   * Please note that this modifies the state of the VM.
+   * Throws if the transaction's gasLimit is greater than
+   * the remaining gas in the block.
+   */
+  async addTransaction(tx: TypedTransaction) {
+    this.checkStatus()
+
+    if (!this.checkpointed) {
+      await this.vm.stateManager.checkpoint()
+      this.checkpointed = true
+    }
+
+    // According to the Yellow Paper, a transaction's gas limit
+    // cannot be greater than the remaining gas in the block
+    const blockGasLimit = new BN(toBuffer(this.headerData.gasLimit))
+    const blockGasRemaining = blockGasLimit.sub(this.gasUsed())
+    if (tx.gasLimit.gt(blockGasRemaining)) {
+      throw new Error('tx has a higher gas limit than the remaining gas in the block')
+    }
+
+    const result = await this.vm.runTx({ tx })
+    this.transactions.push(tx)
+    this.transactionResults.push(result)
+    return result
+  }
+
+  /**
+   * Reverts the checkpoint on the StateManager to reset the state from any transactions that have been run.
+   */
+  async revert() {
+    this.checkStatus()
+    if (this.checkpointed) {
+      await this.vm.stateManager.revert()
+      this.reverted = true
+    }
+  }
+
+  /**
+   * This method returns the finalized block.
+   * It also:
+   *  - Assigns the reward for miner
+   *  - Commits the checkpoint on the StateManager
+   *  - Sets the tip of the VM's blockchain to this block
+   * Optionally seals the block with params:
+   *  - PoW: nonce, mixHash, and difficulty validated with the block number by ethash
+   *  - PoA: seals the block with the private key of the clique signer if provided
+   */
+  async build(sealOpts?: SealBlockOpts) {
+    this.checkStatus()
+    const blockOpts = this.blockOpts
+    const consensusType = this.vm._common.consensusType()
+
+    await this.rewardMiner()
+    await this.vm.stateManager.commit()
+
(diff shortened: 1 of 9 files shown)


packages/vm/lib/buildBlock.ts
```diff
@@ -0,0 +1,250 @@
...
+
+    await this.rewardMiner()
+    await this.vm.stateManager.commit()
+
+    const stateRoot = await this.vm.stateManager.getStateRoot(false)
+    const transactionsTrie = await this.transactionsTrie()
+    const receiptTrie = await this.receiptTrie()
+    const bloom = this.bloom()
+    const gasUsed = this.gasUsed()
+    const timestamp = this.headerData.timestamp ?? Date.now()
```
jochem-brouwer, line 218: Date.now() is not an int

packages/vm/lib/buildBlock.ts
```diff
@@ -0,0 +1,250 @@
...
+      receiptTrie,
+      bloom,
+      gasUsed,
+      timestamp,
+    }
+
+    if (consensusType === 'pow') {
+      headerData.nonce = sealOpts?.nonce ?? headerData.nonce
+      headerData.mixHash = sealOpts?.mixHash ?? headerData.mixHash
+      headerData.difficulty = sealOpts?.difficulty ?? headerData.difficulty
```
jochem-brouwer, line 233: Difficulty is calculated deterministically and is based upon the timestamp of the block. You are not allowed to set your own difficulty. (I think there is a difficulty method exported from the block package)

packages/vm/lib/buildBlock.ts
ryanio: thanks, actually it is a number but in ms so I will divide by 1000 for seconds. fixed in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1158/commits/a2226ccb0722bf331b531a75d204078aabebc86b

packages/vm/lib/buildBlock.ts
ryanio: thanks, yes you are right the proper way would be to use the blockOption `calcDifficultyFromHeader: parentHeader`. removed in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1158/commits/f045c49d6a5e494232982c4e8f146cb68d7a64c5

packages/vm/lib/buildBlock.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    // According to the Yellow Paper, a transaction's gas limit
+    // cannot be greater than the remaining gas in the block
+    const blockGasLimit = new BN(toBuffer(this.headerData.gasLimit))
+    const blockGasRemaining = blockGasLimit.sub(this.gasUsed())
+    if (tx.gasLimit.gt(blockGasRemaining)) {
+      throw new Error('tx has a higher gas limit than the remaining gas in the block')
+    }
+
+    const result = await this.vm.runTx({ tx })
```
holgerd77, line 172: If I am not misreading the code I think what we are not taking into account here yet at all is this miner state update in [runTx()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/lib/runTx.ts#L273) based on the coinbase account.

Here the `runTx()` call is not passing in a `Block` object (which is by design I guess), but then the `coinbase` used will be wrong. Not sure what's the solution here, maybe a dedicated `miner` (`coinbase`?) option in `runTx()` or - broader - a similar optional `headerData` (or: `blockHeaderData` for clarity) option as introduced here? 🤔

holgerd77 (changes requested):
Hi @ryanio, thanks for opening this, this looks really great, the API on this is really beautiful! 😄 

I've stumbled upon one bigger thing (the `miner` reward question on `runTx()`) which would have caused the post-txs state root to be wrong, this would need some addressing.

I think we can't integrate this PR into the `berlin` releases any more, this is just too much in the open so close to the releases, hope that's ok. But also considering the complexity of the introduced code here I generally think it makes sense to first test this code in the client (which just is a super-natural fit) before releasing, then we can really be pretty sure that everything is working properly.

We can then target another minor VM release with these changes together with an eventual first client release in 2-3 weeks or so. How does that sound?

packages/vm/lib/buildBlock.ts
ryanio: ah right, thanks. I was wondering why my spec test didn't catch this as I was getting a consistent final stateRoot, but it's because the coinbase defaults to the zero address so if I set the buildBlock coinbase to a non-zero address then it would have thrown. I will see if I can construct a minimal block to pass in for runTx, not only for this, but in case other things are added to runTx that are dependent on the passed-in block.

ryanio:
ok this should be ready for review now :)

I noticed that if a PoA block wasn't instantiated with the `extraData` field or wasn't at least length 97 (vanity + seal) it would return an invalid `cliqueSigner()` address since the signature would be derived incorrectly.
I fixed this in the block header constructor by padding `this.extraData` to the minimum required length. I previously had this inside the `cliqueSealBlock` function but it had to be set on `this.extraData` for `this.cliqueSigHash()` to use it correctly.

packages/vm/lib/buildBlock.ts
```diff
@@ -0,0 +1,241 @@
...
+   * For PoW, the nonce.
+   * Overrides the value passed in the constructor.
+   */
+  nonce?: Buffer
+
+  /**
+   * For PoW, the mixHash.
+   * Overrides the value passed in the constructor.
+   */
+  mixHash?: Buffer
```
jochem-brouwer, line 46: Is it correct that we cannot seal blocks with clique using this new block builder API?

packages/vm/lib/buildBlock.ts
```diff
@@ -0,0 +1,241 @@
...
+      await this.vm.stateManager.checkpoint()
+      this.checkpointed = true
+    }
+
+    // According to the Yellow Paper, a transaction's gas limit
+    // cannot be greater than the remaining gas in the block
+    const blockGasLimit = new BN(toBuffer(this.headerData.gasLimit))
+    const blockGasRemaining = blockGasLimit.sub(this.gasUsed())
+    if (tx.gasLimit.gt(blockGasRemaining)) {
+      throw new Error('tx has a higher gas limit than the remaining gas in the block')
```
jochem-brouwer, line 162: If this throws, then we should revert the state manager checkpoint right?

packages/vm/lib/buildBlock.ts
```diff
@@ -0,0 +1,241 @@
...
+   * which is validated along with the block number and difficulty by ethash.
+   * For PoA, please pass `blockOption.cliqueSigner` into the buildBlock constructor,
+   * as the signer will be awarded the txs amount spent on gas as they are added.
+   */
+  async build(sealOpts?: SealBlockOpts) {
+    this.checkStatus()
+    const blockOpts = this.blockOpts
+    const consensusType = this.vm._common.consensusType()
+
+    if (consensusType === 'pow') {
```
jochem-brouwer, line 203: Should we explicitly disallow clique here for now?

packages/vm/tests/api/buildBlock.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+    const blockchain = await Blockchain.create({ genesisBlock, common })
+    const vm = await VM.create({ common, blockchain })
+
+    // add balance for tx
+    await vm.stateManager.putAccount(signer.address, Account.fromAccountData({ balance: 100000 }))
+
+    const blockBuilder = await vm.buildBlock({
+      parentBlock: genesisBlock,
+      headerData: { difficulty: 2 },
+      blockOpts: { cliqueSigner, freeze: false },
```
jochem-brouwer, line 171: Ah okay we actually support clique.

packages/vm/lib/buildBlock.ts
jochem-brouwer: We do support it.

packages/vm/lib/buildBlock.ts
```diff
@@ -0,0 +1,241 @@
...
+      await this.rewardMiner()
+    }
+
```
jochem-brouwer, line 206: If we use clique, is the miner rewarded?

jochem-brouwer (commented):
Some comments and questions

packages/vm/lib/buildBlock.ts
ryanio: as discussed in the call, I think this is okay since the consumer of the API can decide if the tx doesn't fit if they would rather add another one, or build to finish the block. if they want to revert the whole block they can use `revert()`.

packages/vm/lib/buildBlock.ts
ryanio: there are no block rewards in clique poa. the signer will get the txs amount spent on gas (applied in `runTx`)

packages/vm/tests/api/buildBlock.spec.ts
ryanio: yep :)

packages/vm/tests/api/buildBlock.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+    const blockBuilder = await vm.buildBlock({ parentBlock: genesis })
+    const gasLimit = genesis.header.gasLimit.addn(1)
+    const tx = Transaction.fromTxData({ gasLimit }, { common })
+    try {
+      await blockBuilder.addTransaction(tx)
+    } catch (error) {
+      if (error.message.includes('tx has a higher gas limit than the remaining gas in the block')) {
+        st.pass('error thrown')
+      } else {
+        st.fail('should throw')
```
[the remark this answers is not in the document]
ryanio: ah thanks, good catch (pun intended ;) fixed with https://github.com/ethereumjs/ethereumjs-monorepo/pull/1158/commits/dbe386b0c9d9870b2b21fac3fc3c4a89071cd6fc

packages/vm/lib/buildBlock.ts
```diff
@@ -0,0 +1,241 @@
...
+    if (this.reverted) {
+      throw new Error('State has already been reverted')
+    }
+  }
+
+  /**
+   * Returns the cumulative gas used by the transactions added to the block.
+   */
+  gasUsed(): BN {
+    return this.transactionResults.reduce(
```
alcuadrado, line 87: I'd accumulate this explicitly as txs are executed, instead of computing it on demand. Otherwise, when you go adding txs, you end up with a quadratic operation.

alcuadrado: The reason for this is that to select a txs we need to call this method, as we can only add txs that fit in `blockGasLimit - builder.gasUsed()`.

packages/vm/lib/buildBlock.ts
```diff
@@ -0,0 +1,241 @@
...
+      const { encodedReceipt } = await generateTxReceipt.bind(this.vm)(tx, txResult, gasUsed)
+      await receiptTrie.put(encode(i), encodedReceipt)
+    }
+    return receiptTrie.root
+  }
+
+  /**
+   * Adds the block miner reward to the coinbase account.
+   */
+  private async rewardMiner() {
```
alcuadrado, line 134: I hadn't realized that this was necessary here 🤯

packages/vm/lib/buildBlock.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    // According to the Yellow Paper, a transaction's gas limit
+    // cannot be greater than the remaining gas in the block
+    const blockGasLimit = new BN(toBuffer(this.headerData.gasLimit))
+    const blockGasRemaining = blockGasLimit.sub(this.gasUsed())
+    if (tx.gasLimit.gt(blockGasRemaining)) {
+      throw new Error('tx has a higher gas limit than the remaining gas in the block')
+    }
+
+    const blockData = { header: this.headerData, transactions: this.transactions }
```
alcuadrado, line 165: The transactions list is not really necessary here, as the EVM can't observe it. Not that it's bad having it either, but I'm just making this comment in case this can lead to some kind of optimization or something.

packages/vm/lib/runBlock.ts
```diff
@@ -206,11 +206,15 @@ export default async function runBlock(this: VM, opts: RunBlockOpts): Promise<Ru
   // Given the generate option, either set resulting header
   // values to the current block, or validate the resulting
   // header values against the current block.
-  if (generateStateRoot) {
+  if (generateFields) {
     const bloom = result.bloom.bitvector
+    const gasUsed = result.gasUsed
```
alcuadrado, line 211: Thanks for this fix! Looks good.

alcuadrado (commented):
This looks great, @ryanio! We'll use it as soon as it's released!

I mostly focused this review on the new API and the fix to `runBlock`. Didn't get into the tests, nor the Clique things.

packages/vm/lib/buildBlock.ts
```diff
@@ -0,0 +1,241 @@
...
+      parentHash: opts.headerData?.parentHash ?? opts.parentBlock.hash(),
+      number: opts.headerData?.number ?? opts.parentBlock.header.number.addn(1),
+      gasLimit: opts.headerData?.gasLimit ?? opts.parentBlock.header.gasLimit,
+    }
+  }
+
+  /**
+   * Throws if the block has already been built or reverted.
+   */
+  checkStatus() {
```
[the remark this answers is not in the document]
ryanio: ah yes thanks, fixed in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1158/commits/02bca5a79322df66e6d72aa3969f40f54553dacb

packages/vm/lib/buildBlock.ts
ryanio: good idea, updated in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1158/commits/eb197ab68b74a61a52b1c61931599fc9610a5b40

packages/vm/lib/buildBlock.ts
ryanio: got it, yeah, i wanted to pass in the most representative block in case `runTx` gets updated in the future and uses something that wasn't anticipated.

holgerd77:
Note: please do not merge before #1170, we can discuss, but atm I am planning to release access list generation as a bugfix release very soon, since I think this is relatively time pressing and should be out before the mainnet HF and at the same time I think it would be good to give the PR here some additional days of practical application test and integration within the client.

PR #1159, opened 2021-03-22
devp2p: implement ETH/65

holgerd77:


holgerd77:
Ok, this is directly ready for review, TBH the major part of the work actually will start when implementing the protocol messages itself in the client and answer them.

Since we also have various other messages not yet implemented I guess it is ok though to integrate, then we have at least the devp2p part ready.

Tested this in the client, new protocol version is used and the new message types are coming in:

image

ryanio (approved):
nice, looks good

Issue #1161, opened 2021-03-23
Case sensitivity results in Error: Cannot find module './eip2930Transaction'


dstarke:
When requiring version 3.1.0 in environments with certain case sensitive file systems, I get an error:
`Error: Cannot find module './eip2930Transaction'`

The error does not occur on my Mac, but does occur on the Linux system on my continuous integration server.

Investigating the downloaded package for version 3.1.0 reveals that there is a file named 'EIP2930Transaction.js' in the dist folder, but no file named 'eip2930Transaction.js', which I believe is the immediate source of the error. This is somewhat confusing because there is a file named 'eip2930Transaction.ts' in the src folder in the repository, and that is the capitalization that the require statement in index.ts uses, but evidently the build system changes things when generating the .js versions.

I think renaming the file in the src folder and changing the require statement in index.ts to both use the version with capital letters would fix the issue, but I don't know enough about your build system to be sure.

holgerd77:
Hi there, thanks for reporting. Yes, just checked, the `dist` folder contains indeed an `EIP2930Transaction.js` file (+ .d.ts and .js.map files also with capital letters). I've actually no idea where this is coming from and why this is happening, guess we'll need to investigate soon, this seems to be a severe build error.

holgerd77:
Just for reference: have tried switching the Node/npm versions for building from `14.16.0` to `12.20.0`, same result.

holgerd77:
Dropping some eventually related TypeScript issues here: https://github.com/Microsoft/TypeScript/issues/2011

holgerd77:
This does fix the problem, have checked the `dist` folder and the problematic code part, see:

![image](https://user-images.githubusercontent.com/931137/112205034-aaefaa00-8c14-11eb-8fde-34c4d4d6486f.png)

holgerd77:
@dstarke `v3.1.1` has been released, would be great if you can give this a try. 😄

dstarke:
`v3.1.1` has fixed the problem in my environment. Thank you so much for the quick response

holgerd77:
@dstarke thanks a lot for reporting, being able to fix this so early has safed us from a lot of trouble! 😋

PR #1162, opened 2021-03-23
New Tx v3.1.1 Release

holgerd77:
Fixes https://github.com/ethereumjs/ethereumjs-monorepo/issues/1161

I would not do releases on the upstream libraries again, I guess if we are quick here this should be ok.


holgerd77:
(if anyone find issues on review, please push directly)

holgerd77:
Or would you guys prefer the suggested fix from #1161 ?

cgewecke (approved):
Super weird bug...will casing definitely be ok with this change? What is causing TS to mess this up?

cgewecke:
Ok sorry - I see you checked the dist locally in #1161. 

Doesn't look like the E2E caught this... maybe if the build/publish happens on linux there's no capitalization.

holgerd77:
@cgewecke thanks for approving, yes, did a few more checks, pretty sure this is ok now. Ok, then I'll directly do the release before this is distributed more broadly.

holgerd77:
(will wait for the tests though of course)

cgewecke:
Closing in favor #1163

PR #1166, opened 2021-03-25
VM: Refactor runTx() API tests / add Typed Transaction tests

holgerd77:
I've switched from TxPool to access list generation for today. 😄 

On a first round I will add some tests to `runTx()` API tests and also clean up there a bit. On a first commit 4163c40 I've removed the existing `setup()` function, I am pretty sure that this is not doing anything useful (maybe some relic which was necessary on some old structure setup) and just bloating up the code. This is also confusing to read and hinders adding more flexible test setups directly by e.g. changing the `StateManager` or `Common`. Nevertheless pushing early here, so if I've overlooked something here let me know.

Generally I will see how far I will come, I think I'll give some tests some more space on a first round, generally in need I have the impression.

### DIFF
--- a/packages/vm/tests/api/runTx.spec.ts
+++ b/packages/vm/tests/api/runTx.spec.ts
@@ -19,74 +19,139 @@ const TRANSACTION_TYPES = [
 const common = new Common({ chain: 'mainnet', hardfork: 'berlin' })
 common.setMaxListeners(100)
 
-tape('runTx() -> successful API usage', async (t) => {
-  t.test('simple run (unmodified options)', async (t) => {
+tape('runTx() -> successful API parameter usage', async (t) => {
+  async function simpleRun(vm: VM, msg: string) {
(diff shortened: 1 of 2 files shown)


holgerd77:
Side note along the way: `TransactionFactory` is also a bit dangerous, wanted to add typed txs to the mix in the VM `runTx()` API tests and accidentally initialized with `txType` (instead of `type`) using for `fromTxData()` and this then created a normal legacy tx and otherwise went completely unnoticed. We might want to add additional checks there to better prevent these kind of development errors.

holgerd77:
Ok, this has turned into a general VM runTx() API test refactor, will stop here for now and open for review! 😄 

Sorry for the wrong branch name, should be no problem though since we delete anyhow after merge.

ryanio (approved):
looks great, nice clean up and added cases :)

PR #1167, opened 2021-03-26
Blockchain: New Release v5.2.1 / Fixed Common custom chain error / VM API test rework and custom chain test additions

### LINKED ISSUE #1164: Custom chainId example not working
The example for a custom chain common from [https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/common](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/common) is not working when using it for creating a VM instance

```
const customChainParams = { name: 'custom', chainId: 123, networkId: 678 }
const customChainCommon = Common.forCustomChain('mainnet', customChainParams, 'byzantium');
const vm = new VM({ common: customChainCommon  })
```

Its returning the following error message:

UnhandledPromiseRejectionWarning: Error: Chain with name 123 not supported

holgerd77:
Fixes #1164 

I did some further test cleanup on the runBlockchain() and constructor tests on first round.

Some context on these `setup()` functions: just could remember that this was introduced by Sina years ago. This was his very first work on the libraries - adding some tests here - and he wasn't quite sure if he should test the functions (e.g. `runBlockchain()`) directly or run in a VM context, back then I also couldn't answer properly.

Now it's very clear that there is no need to run directly since the exposed API is just always going through a VM instance (in fact it is counter-productive, since this is setting up something which is just not used in production and might rather introduce new unnecessary problems or even circumvent going through some to-be-tested functional parts).


holgerd77:
Ok, ready, admittedly a lot of code changes for a one-line bug fix. 😛

Ready for review. 😄 

I've directly added release notes and a version bump for `Blockchain`, as stated in the issue I guess this issue is relatively urgent to fix and might become more prevalent if not addressed in time.

holgerd77:
Ah, and I also did a check on `new Common()` instantiations in other monorepo libraries where a similar error might occur. This seems to be the only place where this type of error can be triggered by not copying over the whole `Common` instance with eventually changed properties.

holgerd77:
I've also added a new `clean` command to all `package.json` files here and appended it to the `prepublishOnly` hook, I find it too dangerous to leave this in any longer after the latest experiences.

The solution is some I hope good compromise. I didn't want to build directly, since I guess this would significantly increase development build times if TypeScript always have to compile all the files again. At the same time I also didn't want to just add to `prepublishOnly` since it would be good to at least integrate this a bit into the everyday workflow to at least occasionally do a clean build to test things. Therefore the `clean` command, this at least makes it a bit easier and goes lighter off hand than to write the `rm -Rf ./dist && rm -Rf ./dist.browser` manually. 😄

packages/blockchain/src/index.ts
```diff
@@ -304,10 +304,8 @@ export default class Blockchain implements BlockchainInterface {
     }
 
     if (!genesisBlock) {
-      const common = new Common({
-        chain: this._common.chainIdBN(),
-        hardfork: 'chainstart',
-      })
+      const common = this._common.copy()
```
ryanio, line 307: ahh much better :)

ryanio (approved):
nice, looks great, really good test refactoring.

i noticed some newly added @ts-ignores, i think they are fine, but in the future it would be nice to add an extra comment notating what it is needed for, then it should be easier to update or remove in the future when revisiting.

holgerd77:
Just published as [v5.2.1](https://github.com/ethereumjs/ethereumjs-monorepo/releases/tag/%40ethereumjs%2Fblockchain%405.2.1) on npm.

holgerd77:
@ryanio yeah, that makes sense with the @ts-ignore comments.

Issue #1157, opened 2021-03-19
[VM] Incorrect type of ExecResult.exceptionError


alcuadrado:
I mentioned this bug some time ago, and never got to the root cause of it, so I'm opening this issue describing it.

`ExecResult` has this field `exceptionError?: VmError`. This is incorrect, as sometimes the error has a different type. This was breaking Hardhat in very weird ways.

A way to reproduce it is to use a custom `StateManager` that throws on, say, `putContractStorage`. Somehow this is watched and ends up there, and the types lie to you.

The interpreter has this TODO btw `// TODO: Throw on non-VmError exceptions`. I assume it's the same thing.

alcuadrado:
Also, IMO changing that type is a breaking change. Resolving that TODO and throwing on non-`VmError`s is probably also one and may break quite a few things.

holgerd77:
Hi there,
thanks for reporting. Had a first look: a first obvious place for this to happen is [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/lib/evm/evm.ts#L319) in the EVM class where we assign this eventual `errorMessage` object which is `any` and therefore not guarded by TypeScript.

holgerd77:
Update: ah, you are right, this can only result from within the `Interpreter` where the `exceptionError` is taken over from the `Interpreter` [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/lib/evm/evm.ts#L450) within the `EVM` class.

Mentioned TODO can be found [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/lib/evm/interpreter.ts#L122) in the `Interpreter` code.

holgerd77:
I followed the call chain and what errors are guarded where by `try / catch`, is there any reason to not just throw along the mentioned code from the TODO respectively check for a `VMError` and throw on all other cases? 🤔

```typescript
let err
// Iterate through the given ops until something breaks or we hit STOP
while (this._runState.programCounter < this._runState.code.length) {
  const opCode = this._runState.code[this._runState.programCounter]
  this._runState.opCode = opCode
  await this._runStepHook()

  try {
    await this.runStep()
  } catch (e) {
    // STOP is not an exception
    if (e.error !== ERROR.STOP) {
      err = e
    }
    // TODO: Throw on non-VmError exceptions
    break
  }
}

return {
  runState: this._runState,
[...]

PR #1168, opened 2021-03-26
Test VM exceptionError bug

holgerd77:
This is testing behavior along a potential bug fix for #1157, do not merge.


holgerd77:
@alcuadrado ok, this doesn't affect VM test run behavior at all. I would cautiously say we could eventually merge this. From my understanding of the whole coherences I would assume that the effect of this would essentially be that if you guys or we directly in the VM have a bug in the code (e.g. in a custom state manager), this would directly break and trickle up and not be disguised any more and so the effect from this would be that it will get easier to have these kind of failures fixed? Is this correct?

ryanio:
What's the status of this PR? I'm thinking we can add a test to ensure that non-VmErrors propagate through as expected, then it can be ready to merge in?

Codecov didn't give any coverage warning on the newly added lines in this PR so I'm wondering if the code path is already being covered, I'd be interested to see.

holgerd77:
@ryanio I think this can be merged.

The suggestion on setting up a test is great 😄 , what would be good setup here?

I was thinking about subclassing `Interpreter` and add some fitting function which just throws, but it is impressively complex (or rather: laborious) to get this setup going respectively do an instantiation with all this `message` setup, I think we should really simplify here at some point as far as possible and generally make it easier to instantiate an independent `EVM` separate from all this outer environmental processing logic stuff brought in in `runTx()`, `runBlock()` and the like.

The other way for adding a test here might be the usage of the `testdouble` library which we already use in the client.

Any preference here respectively suggestion?

holgerd77:
@ryanio short *ping* (see last question) 😄  (also feel free to add a test yourself if you have an easy idea here)

ryanio:
rebased this, ok let me know what you think of the added tests. tried to keep it as simple as possible and used @alcuadrado's original suggestion in the issue to throw an error from `stateManager.putContractStorage()`

packages/vm/tests/api/runCode.spec.ts
```diff
@@ -50,3 +51,47 @@ tape('VM.runcode: initial program counter', (t) => {
...
+
+    try {
+      await vm.runCode(runCodeArgs)
+      st.fail('should throw error')
+    } catch (e) {
+      st.ok(e.toString().includes('Test'), 'error thrown')
+    }
+    st.end()
+  })
+})
```
holgerd77, line 97: Ah, yeah, really nice test cases! 😄

holgerd77 (commented):
Yeah, nice, this looks really good, will merge. 😄  👍 

Also a somewhat really core-centered change (like the recent and discussed `StateManager` changes) with some potential to cause some reactions once released, but I guess very much net-positive since either a) people will discover own bugs in their customizations which were previously hidden or b) people might report bugs we still have in the VM.

So really excited about this change, long overdue! 🎉

(will merge by admin-merge since technically reviewed by @ryanio by continuing the work)

Issue #1152, opened 2021-03-17
VM: generate EIP-2930 Access Lists


holgerd77:
I wondered along writing release notes how to actually generate access lists for an `EIP-2930` transaction. I guess the only way is to actually run the code through the VM? Would it make sense to just return by standard access lists on VM runs as an additional return value of `VM.runCode()` or something? 🤔

jochem-brouwer:
I don't really like expanding `runCode` here, I think we should deprecate `runCode` and `runCall`, because they do not run some code which `runTx` does (runTx deletes empty accounts while runCode/runCall does not, and I think they do add them to the touched list but never wipe this (?)). 

In general: there is no such thing as **the** access lists. For instance on a DeFi project at the start of a block you could run a tx and see whatever orders you touch when you buy something - but when it actually gets mined, the order might be gone and you might buy into another order, so from that PoV your access lists could be considered incomplete.

I think we should add an option to `runTx`, something like `reportAccessLists`, this would deep copy the access lists before wiping the list and add it to the return value of `runTx`. We could also do this by default, but this would mean we have to deep copy this access list every time, which is not desirable.

holgerd77:
I am using `runCode` all the time and I think it has some value to be able to just do code runs without the overhead tx execution brings, so I think this has some value (`runCall` not sure). And if `runCode` is not running some code it actually *should*, then this code parts should rather be refactored? 🤔

Anyhow, side discussion. 🙂  Thanks for the evolved suggestion here on the access list generation, think that makes sense.

PR #1170, opened 2021-03-27
VM: Generate Access Lists in runTx()

holgerd77:
Ok, here is the access list generation PR fixing #1152. This needs significantly more tests, but this should generally do what is expected. I just tried to generate the following super-simple access list with this:

```shell
{
  gasUsed: <BN: 52f8>,
  execResult: {
    gasUsed: <BN: 0>,
    exceptionError: undefined,
    returnValue: <Buffer >,
    gasRefund: <BN: 0>
  },
  bloom: Bloom {
    bitvector: <Buffer 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 ... 206 more bytes>
  },
  amountSpent: <BN: 2068e0>,
  accessList: [
    {
      address: 'be862ad9abfe6f22bcb087716c7d89a26051f74c',
      storageKeys: []
    }
  ]
}
```

Which obviously is not enough yet. 😄  Functionality for more complex AL is there but should be manually tested or optimally by adding additional test cases. 

For now I've added the functionality to `StateManger.clearWarmedAccounts()` in a non-breaking way. On a second thought along pushing I rather think it might be better and more flexible to just add a distinct method for that which can be called separately (which then still can be called somewhat along the clearing-call in our context).

I also added a new helper function `Address.isPrecompileOrSystemAddress()` to Util which checks an address to be a precompile or system address following [EIP-1352](https://eips.ethereum.org/EIPS/eip-1352).

### DIFF
--- a/packages/util/test/address.spec.ts
+++ b/packages/util/test/address.spec.ts
@@ -75,6 +75,14 @@ tape('Address', (t) => {
     st.end()
   })
 
+  t.test('should provide correct precompile check', (st) => {
+    const precompile = Address.fromString('0x0000000000000000000000000000000000000009')
+    st.true(precompile.isPrecompileOrSystemAddress(), 'should detect precompile address')
+    const nonPrecompile = Address.fromString('0x990ccf8a0de58091c028d6ff76bb235ee67c1c39')
+    st.false(nonPrecompile.isPrecompileOrSystemAddress(), 'should detect non-precompile address')
+    st.end()
+  })
+
   t.test('should generate address for CREATE2', (st) => {
     for (const testdata of eip1014Testdata) {
       const { address, salt, initCode, result } = testdata
--- a/packages/vm/lib/runTx.ts
+++ b/packages/vm/lib/runTx.ts
@@ -1,7 +1,12 @@
 import { debug as createDebugLogger } from 'debug'
-import { Address, BN } from 'ethereumjs-util'
+import { Address, BN, toBuffer } from 'ethereumjs-util'
 import { Block } from '@ethereumjs/block'
-import { AccessListItem, AccessListEIP2930Transaction, TypedTransaction } from '@ethereumjs/tx'
+import {
+  AccessListItem,
+  AccessListEIP2930Transaction,
+  TypedTransaction,
+  AccessList,
+} from '@ethereumjs/tx'
 import VM from './index'
 import Bloom from './bloom'
 import { default as EVM, EVMResult } from './evm/evm'
@@ -40,6 +45,15 @@ export interface RunTxOpts {
    * agains the block's gas limit.
    */
   skipBlockGasLimitValidation?: boolean
+
+  /**
+   * If true, adds a generated EIP-2930 access list
+   * to the `RunTxResult` returned.
+   *
+   * Option works with all tx types. EIP-2929 needs to
+   * be activated (included in `berlin` HF).
+   */
+  reportAccessList?: boolean
 }
 
 /**
@@ -58,6 +72,11 @@ export interface RunTxResult extends EVMResult {
    * The amount of gas as that was refunded during the transaction (i.e. `gasUsed = totalGasConsumed - gasRefund`)
    */
   gasRefund?: BN
+
+  /**
+   * EIP-2930 access list generated for the tx (see `reportAccessList` option)
+   */
+  accessList?: AccessList
 }
 
 export interface AfterTxEvent extends RunTxResult {
@@ -111,10 +130,10 @@ export default async function runTx(this: VM, opts: RunTxOpts): Promise<RunTxRes
     const castedTx = <AccessListEIP2930Transaction>opts.tx
 
     castedTx.AccessListJSON.forEach((accessListItem: AccessListItem) => {
-      const address = Buffer.from(accessListItem.address.slice(2), 'hex')
+      const address = toBuffer(accessListItem.address)
       state.addWarmedAddress(address)
       accessListItem.storageKeys.forEach((storageKey: string) => {
-        state.addWarmedStorage(address, Buffer.from(storageKey.slice(2), 'hex'))
+        state.addWarmedStorage(address, toBuffer(storageKey))
       })
     })
   }
@@ -123,6 +142,10 @@ export default async function runTx(this: VM, opts: RunTxOpts): Promise<RunTxRes
     const result = await _runTx.bind(this)(opts)
     await state.commit()
     debug(`tx checkpoint committed`)
+    if (this._common.isActivatedEIP(2929) && opts.reportAccessList) {
+      result.accessList = state.clearWarmedAccounts(true)
+      console.log(result)
+    }
     return result
   } catch (e) {
     await state.revert()
--- a/packages/vm/lib/runTx.ts
+++ b/packages/vm/lib/runTx.ts
@@ -52,6 +52,9 @@ export interface RunTxOpts {
    *
    * Option works with all tx types. EIP-2929 needs to
    * be activated (included in `berlin` HF).
+   *
+   * Note: if this option is used with a custom `StateManager` implementation
+   * the `generateAccessList()` method must be implemented.
    */
   reportAccessList?: boolean
 }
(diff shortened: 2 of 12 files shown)


packages/util/test/address.spec.ts
```diff
@@ -75,6 +75,14 @@ tape('Address', (t) => {
     st.end()
   })
 
+  t.test('should provide correct precompile check', (st) => {
+    const precompile = Address.fromString('0x0000000000000000000000000000000000000009')
+    st.true(precompile.isPrecompileOrSystemAddress(), 'should detect precompile address')
```
jochem-brouwer, line 80: Hmm this EIP is actually a little bit confusing. It does specify which address range could be assumed to be system precompiles. However, for access lists, we are only allowed to push precompiles which actually have code to the access lists. There was a bug in Besu where they also included BLS addresses to the access lists, which does not match the EIP. What is weird is that EIP2930 in combination with EIP1352 is not very consistent.

packages/vm/lib/runTx.ts
```diff
@@ -123,6 +142,10 @@ export default async function runTx(this: VM, opts: RunTxOpts): Promise<RunTxRes
     const result = await _runTx.bind(this)(opts)
     await state.commit()
     debug(`tx checkpoint committed`)
+    if (this._common.isActivatedEIP(2929) && opts.reportAccessList) {
+      result.accessList = state.clearWarmedAccounts(true)
```
jochem-brouwer, line 146: I know this is WIP, but this currently only supports the accessed items for the root transaction right? It will not report access lists if you `*CALL*` into another address.

jochem-brouwer: Hmm it might actually report it, but it will not report if we `REVERT` one of these inner calls, since that removes the accessed items from warmed storage.

packages/util/test/address.spec.ts
holgerd77: Hmm, I'm just using this to *remove* precompile addresses, so there should be no problem? 🤔

packages/vm/lib/runTx.ts
holgerd77: Hmm, does this happen very often in practice and do you think it's an important miss?

holgerd77: (I can add extra logic for it though)

jochem-brouwer: Well, internal reverts sometimes happen, and if you don't put these storage slots on the access lists then you are going to get charged to touch a cold account multiple times. So if we want to generate "correct" access lists (we want to be friendly and provide the complete list of the slots and accounts we are going to add in the access list), then we should also include the reverted slots. 

This is actually a bit mind-boggling... There is an edge case here. What if an internal call reverts, but if we add it to the access list then it does not revert? This might also change the behavior of the call; we might now touch other slots. I think we *could* try to rebuild the access lists based upon whatever we just touched, but I think we might get in a flip/flop situation where each time you edit the access list, you get another access list which we already considered before 🤔  (Hope this reasoning makes sense)

jochem-brouwer: I will ask on R&D

jochem-brouwer: Checkout geth's implementation; https://github.com/ethereum/go-ethereum/pull/22550

They cap the count of attempts to generate an access list at 10.

packages/util/test/address.spec.ts
holgerd77: "which actually have code to the access lists"

What do you mean by that actually? 😀

holgerd77:
Hi @jochem-brouwer, thanks for digging in so deeply here, really cool! 😄  I won't do this edge case implementation here since I simply don't have the development resources for this given all our other tasks atm, instead I've added a note here along the implementation. Since this is not breaking anything if an access list might not be complete on some edge cases (+ also this 10 -run-limit from geth seems also conceptually be rather a practical solution than a 100%-correct implementation) and it is not yet clear how much this whole access list feature will be used, I think this should be acceptable.

I have added the additional structure though for `revert()`, also did some further functional cleanup on the folding logic in the `StateManager` and added a set of new tests dedicatedly for `StateManager.createAccessList()`.

Ah, yes, and this is now a separate function, haven't added to the official interface yet though for compatibility reasons. Instead `runTx()` (with `reportAccessList` option activated) will throw if used with a custom state manager not implementing this method.

Ok, ready for review. 😄

holgerd77:
Hi @ryanio, could you have a look at the latest [VM API test run](https://github.com/ethereumjs/ethereumjs-monorepo/runs/2220019585?check_suite_focus=true), which is failing for the karma browser run? I could just add the `examples` folder to the `exclude` directive and also add this `downlevelIteration` compiler option, but I guess you are better at judging the overall situation and see why this might be happening in the first place and if I e.g. introduced some structures which are problematic from a browser perspective or something.

packages/util/test/address.spec.ts
jochem-brouwer: It means that once we setup the initial access list, we add the caller and the target to the access list, but we also add every precompile which has attached code to it. So, Besu made a mistake and also added the BLS precompile addresses. However, in Berlin, we do not have BLS, so there is no precompiled code at these BLS addresses and they should thus not be added to the access lists.

packages/vm/lib/state/stateManager.ts
```diff
@@ -407,7 +423,10 @@ export default class DefaultStateManager implements StateManager {
     // setup cache checkpointing
     this._cache.revert()
     this._storageTries = {}
-    this._accessedStorage.pop()
+    const lastItem = this._accessedStorage.pop()
+    if (lastItem) {
+      this._accessedStorageReverted.push(lastItem)
```
jochem-brouwer, line 428: This will degrade performance, right? If we don't want to get the access lists, then we also should not copy these items since it is unnecessary?

packages/vm/lib/state/stateManager.ts
```diff
@@ -642,9 +661,57 @@ export default class DefaultStateManager implements StateManager {
...
+    const folded = new Map([...mergedStorage[0].entries()].sort())
+
+    // Transfer folded map to final structure
+    const accessList: AccessList = []
+    folded.forEach((slots, addressStr) => {
+      const address = Address.fromString(`0x${addressStr}`)
+      if (!address.isPrecompileOrSystemAddress()) {
+        const storageSlots = Array.from(slots)
+          .map((s) => `0x${s}`)
+          .sort()
```
jochem-brouwer, line 705: Sorting this is fine, but it is not really necessary, right?

jochem-brouwer (commented):
Some general remarks, looks great in general, nice tests 👍

packages/vm/lib/state/stateManager.ts
holgerd77: I would be a fan of sorting (even if it comes to some performance costs), I think this gives a better overview if one want to manually check the outcome. And I guess in this use case performance is a bit secondary (just an unproven assumption reflecting on how I imagine people using this stuff).

packages/vm/lib/state/stateManager.ts
holgerd77: Ah, yes, you are right. With this openly usable `generateAccessList()` function we might nevertheless want to leave as-is, but haven't thought about this aspect yes.

ryanio (approved):
lgtm!

jochem-brouwer:
Question here, do the reported access lists sometimes also report the sender address or the `to` address (or a precompile)? Because these should not be included on the access lists, they are already added to the lists internally by default.

holgerd77:
@jochem-brouwer ah, actually `to` is likely not a good candidate to remove since there might be storage slots added as well? 🤔  Would it be a problem if we then just keep both of the addresses in?

jochem-brouwer:
Ah you are right. We only don't want to put `to` in if the storage slots there are empty. But I do not think there is a situation where this can occur. It is not really a problem if we would put the `to` address in with no slots, but that means the user has to pay more gas than necessary.

PR #1172, opened 2021-03-29
Temporarily disable e2e tests (due to intermittent failures)

cgewecke:
The current E2E test seems to be failing randomly with a build error (from the target repo's own build script)
```
'dist/tokens/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48.json'
Error: ENOENT: no such file or directory, unlink 'dist/tokens/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48.json'
```

I don't think this is worth digging into because this target was POC and doesn't test much of the monorepo.

Will open a PR re-enabling when Hardhat has successfully integrated the new libraries - they have a huge test suite which touches most of the code here.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -17,31 +17,18 @@ jobs:
     strategy:
       matrix:
         node-version: [15]
-    steps:
-      - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v1
-        with:
-          node-version: ${{ matrix.node-version }}
-      - uses: actions/checkout@v2
-        with:
-          submodules: recursive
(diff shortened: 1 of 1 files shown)


ryanio (approved):
sounds good :)

holgerd77:
I am still getting email notification on this, now with:

```shell
.github/workflows/e2e-tests.yml#L1
No steps defined in `steps` for the following jobs: virtual-npm-publish
```

😜

Just discovered that the workflow can also be disabled on UI (not sure if admin-only?) here https://github.com/ethereumjs/ethereumjs-monorepo/actions/workflows/e2e-tests.yml, have done this now for this workflow.

PR #1173, opened 2021-03-29
Stop throwing when non-hex-prefixed addresses are passed to util `isValid..` methods

cgewecke:
PR changes `isValidAddress`, `isValidChecksumAddress` and `isZeroAddress` to return `false` when given a non-hex-prefixed string (or non-string) input.

This proposal comes from a Hardhat integration [review discussion][1]. 

Basic idea is that it's unexpected behavior for boolean validators like this to throw. 

[1]: https://github.com/nomiclabs/hardhat/pull/1337#discussion_r600056630

### DIFF
--- a/packages/util/CHANGELOG.md
+++ b/packages/util/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 (modification: no type change headlines) and this project adheres to
 [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [7.1.0] - TBD
+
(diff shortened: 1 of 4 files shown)


ryanio:
reasoning seems valid to me, although would this be considered a breaking change for existing usage?

alcuadrado:
> reasoning seems valid to me, although would this be considered a breaking change for existing usage?

I think it can be seen as a bugfix that won't lead to breaking code unless you were doing something extremely weird. My reasoning is that you already have to handle both `false` and an exception, and they both had the same semantics, so they should be doing the same.

ryanio:
sounds good, @cgewecke do you mind adding a changelog entry

cgewecke:
@ryanio Updated! Thanks.

ryanio (approved):
lgtm!

PR #1179, opened 2021-03-31
Fix default value for empty access lists

### LINKED ISSUE #1178: Tx: AccessListEIP23930Transaction initialization not working with default AL value
Instantiation of AL txs is not working when no accessList value is provided, following error:

image

Fix is I guess to replace `emptyBuffer` with the empty list, this fix should be added to the #1174 releases.

jochem-brouwer:
Closes #1178

### DIFF
--- a/packages/tx/src/eip2930Transaction.ts
+++ b/packages/tx/src/eip2930Transaction.ts
@@ -23,6 +23,9 @@ import {
   N_DIV_2,
 } from './types'
 
+const emptyAccessList: AccessList = []
+const emptyBuffer = Buffer.from([])
+
 /**
  * Typed transaction with optional access lists
  *
@@ -109,8 +112,6 @@ export default class AccessListEIP2930Transaction extends BaseTransaction<Access
 
     const [chainId, nonce, gasPrice, gasLimit, to, value, data, accessList, v, r, s] = values
 
-    const emptyBuffer = Buffer.from([])
-
     return new AccessListEIP2930Transaction(
       {
         chainId: new BN(chainId),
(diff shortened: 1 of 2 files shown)


packages/tx/src/eip2930Transaction.ts
```diff
@@ -137,7 +138,11 @@ export default class AccessListEIP2930Transaction extends BaseTransaction<Access
    * varying data types.
    */
   public constructor(txData: AccessListEIP2930TxData, opts: TxOptions = {}) {
-    const { chainId, accessList } = txData
+    // Have to split up the const/let part to make eslint happy
+    const { chainId } = txData
+    let { accessList } = txData
+
+    accessList = accessList ?? emptyAccessList
```
holgerd77, line 145: This is not necessary since in all cases default values are handled in the code below.

packages/tx/test/eip2930.spec.ts
```diff
@@ -318,4 +318,14 @@ tape('[AccessListEIP2930Transaction]', function (t) {
...
+
+  t.test('should create and sign transactions without passing access list value', (st) => {
+    const tx = AccessListEIP2930Transaction.fromTxData({}, { common })
+    const signed = tx.sign(pKey)
+
+    st.deepEqual(tx.accessList, [])
+    st.deepEqual(signed.accessList, [])
+
+    st.end()
+  })
```
holgerd77, line 330: Cool, thanks for an additional test on this! 😄

holgerd77 (approved):
LGTM

packages/tx/src/eip2930Transaction.ts
jochem-brouwer: Yep.

jochem-brouwer:
Ok, thanks for directly pushing 😄

PR #1180, opened 2021-03-31
Implement EIP 1186

### LINKED ISSUE #1078: Trie: Add support for EIP-1186 (eth_getProof)
As discussed in our last call 21 we might want to add support for [EIP-1186](https://eips.ethereum.org/EIPS/eip-1186) to the trie library.

On first thought I think the most obvious thing to integrate might be proof compatibility for the [Trie.verifyProof()](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/trie#merkle-proofs) method, so to allow for the method to "understand" the proofs delivered by `eth_getProof`.

This would need some closer look though how to best structurally integrate. There also might be other functionality useful to add in this context.

jochem-brouwer:
Closes #1078. Implements [EIP1186](https://github.com/ethereum/EIPs/issues/1186)

TODOs

- [ ] Implement tests for StateManager
- [ ] Implement RPC method in client
- [ ] Add client tests
- [ ] Figure out what to do when `verifyProof` from `trie` returns `null` (probably edge case; only happens on empty tries)


holgerd77:
Didn't dig deeper here (sorry, didn't answer on your initial question on this yet), but do you know that [this](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/trie#merkle-proofs) functionality exists?

jochem-brouwer:
Yep, this PR internally calls into it. The goal of this PR is to implement the EIP and to actually export the merkle proofs as specified in the EIP, and also open it as endpoint in the client. 

Based upon the EIP, it might be that our current proofs are not valid (not sure yet).

holgerd77:
Rebased this.

holgerd77:
Ok, this has gotten some renewed interest with a concrete use case. 😄

Lion from ChainSafe asked in the #lightclient R&D Discord channel on this, functionality would be needed to complement to Lightclient functionality in Lodestar, here some extract from the chat:

> To recap, the goal is to demo a minimum viable browser based light-client that shows recent user data (no execution and no p2p networking, view only)
> 
> Specifically a browser based app (either UI or extension) that communicates via HTTP with:
> - consensus client to get head + execution payload state root (dedicated REST endpoints on the beacon node, pending > standardization) This part will be covered by @chainsafe/lodestar-light-client
> - execution client to get proofs (with eth_getProof) for the user input address(es)
> 
> If EthereumJS full node is able to serve eth_getProof requests it could also be the backend for the execution part. But > the blocker so far is parsing the eth_getProof responses in the browser

@jochem-brouwer I took another look into the code base since I wanted to determine if we really need to expose this through the `StateManager`, since Lodestar (and likely other use cases as well) would only need the proof functionality (and not the whole VM). But I guess you are right, this is really pretty connected to the higher level `StateManager` functionality. Guess for now this should be ok to force Lodestar (or - as I just read - it will be a dedicated library for testing) to import the whole VM. Maybe we want to take this as some trigger though and more seriously move forward with extracting the `StateManager` from the VM (which is actually planned for a long time).

Independently from the TODO list above (would say we do not need to also implement the `eth_getProof` endpoint atm (we very well can though, then Lodestar can also use our client as a backend for the experiement): is this PR somwhat ready for merge?

I guess we want to integrate this in a way into the `StateManager` that we are not adding this functionality to the interface and also not use in the VM directly.

jochem-brouwer:
Closing this in favor of #1590. Note: the name has changed since the EIP is stagnant and also not up to date, the goal is to implement the eth_getProof RPC endpoint in the client.

PR #1174, opened 2021-03-30
VM v5.3.0 (Access List Gen & Block Builder) / Util v7.0.10 / Tx v3.1.2 Releases

holgerd77:

PR #1183, opened 2021-03-31
VM: fix access list generation bugs in runTx()

holgerd77:
PR fixes two issues:

- Sender addresses are not included in access lists any more, `to` addresses only when storage slots have been accessed (both discussed with @jochem-brouwer)
- Added `0x` to the returned addresses in access lists (impressive that this slipped both through my eyes on development and on review, since this is really present on the large amount of tests 😛 )

I first wasn't sure if to remove the addresses in `runTx()` and leave the function signature of `StateManager.generateAccessList()` simple or add directly. I've now chosen for a very generic solution. I think this might be useful for some unforeseen use cases and eventual access list optimizations (and if not it's not a great loss or overblow either).


holgerd77:
To be included in #1174

packages/vm/lib/state/stateManager.ts
```diff
@@ -699,12 +705,17 @@ export default class DefaultStateManager implements StateManager {
     const accessList: AccessList = []
     folded.forEach((slots, addressStr) => {
       const address = Address.fromString(`0x${addressStr}`)
-      if (!address.isPrecompileOrSystemAddress()) {
+      const check1 = address.isPrecompileOrSystemAddress()
```
jochem-brouwer, line 708: Ah, this is not correct, due to my comment earlier. Only precompiles which have precompiled code ("active precompiles") are already in the access list by default. It will also remove 0x00..00ffff for instance. 

The actual precompiles which you want to remove can be found here: https://github.com/ethereumjs/ethereumjs-monorepo/blob/2c711c76d3371cad32617390e14bc7c4936ffcb8/packages/vm/lib/runTx.ts#L194

holgerd77: I don't think this is a practical problem, even [EIP-1352](https://eips.ethereum.org/EIPS/eip-1352) itself builds on the extremely low probability that this address space is currently in use:

> Due to the extremely low probability (and lack of adequate testing possibilities) no explicit checks should be added to ensure that external transaction signing or the invoking of the CREATE instruction can result in a precompile address.

packages/vm/lib/runTx.ts
```diff
@@ -153,8 +153,13 @@ export default async function runTx(this: VM, opts: RunTxOpts): Promise<RunTxRes
     await state.commit()
     debug(`tx checkpoint committed`)
     if (this._common.isActivatedEIP(2929) && opts.reportAccessList) {
+      const { tx } = opts
+      // Do not include sender address in access list
+      const removed = [tx.getSenderAddress()]
+      // Only include to address on present storage slot accesses
+      const onlyStorage = tx.to ? [tx.to] : []
       // @ts-ignore method is not yet part of the interface for backwards-compatibility reasons
```
alcuadrado, line 161: I think this will lead to some very uncomfortable patterns. For example, we use `StateManager` everywhere, as we have two different implementations.

What would be better is making this new method optional. That requires less type casting, but rather checking `!== undefined` before using it.

holgerd77: How do I make a method optional in an interface? 🤔

packages/vm/lib/state/stateManager.ts
holgerd77: (and precompiles which don't exist yet won't be on the initial access list anyhow, or am I missing something?)

packages/vm/lib/runTx.ts
holgerd77: Ah, ok, got it https://stackoverflow.com/questions/27522973/typescript-optional-function-in-interface, I somehow was convinced that this is not possible, will update.

packages/vm/lib/state/stateManager.ts
jochem-brouwer: The edge case here is if someone calls `BALANCE` on 0x00..00ffff. Due to the current implementation, the item will be removed from the access list. But, as you mentioned, since this address is not on the initial access list, we will actually pay the cold read account gas cost for it. If we instead had kept it on the access list, we would have this ~10% gas discount.

jochem-brouwer: This is not a super big problem since this will probably never happen in practice, but it is not "correct".

holgerd77: Hmm, I am not convinced that it's worth it but have updated 80a9e19, want to get this release out. 😄

holgerd77: (so would be great if you can also judge on the solution on @alcuadrado 's request and approve if you are generally satisfied, now with `berlin` so close to `mainnet` I think every day counts)

jochem-brouwer (approved):
Yep, I like this, both @alcuadrado's suggestion. Let's not merge until he also approves. Have approved.

holgerd77:
@jochem-brouwer I'll merge here but won't do the final releases until he approves, hope that's ok as well. I just would want to have the release PR ready so that I can do the releases tomorrow and hopefully have an approval, I am otherwise not available any more over (most) of the Easter days.

jochem-brouwer:
Ah yes, that is OK 👍

PR #1184, opened 2021-04-02
util, devp2p: add dist.browser to package.json

ryanio:
This PR:
* adds `dist.browser` to `package.json` for `util` and `devp2p`
* adds missing `types` key to `package.json` for `devp2p`
* updates some urls to new monorepo locations
* moves some more important fields to the top, urls to bottom
* normalizes some other fields across all packages


packages/client/package.json
```diff
@@ -2,16 +2,28 @@
...
+    "client",
+    "blockchain",
+    "light sync",
+    "full sync"
+  ],
+  "engines": {
+    "node": ">=8.0.0"
+  },
+  "files": [
+    "dist"
```
holgerd77, line 19: Is this intentional to exclude `bin` here?

holgerd77 (approved):
Looks good, thanks, this is fitting well since I was planning to do a `devp2p` release during the week (unless anyone has any objections). 😄

packages/client/package.json
ryanio: yes that folder is not needed anymore since it now only contains the typescript source file. the actual built file we are interested in (that the `package.json` key for `bin: ethereumjs` points to) is now at `dist/bin/cli.js`

PR #1185, opened 2021-04-05
vm: add tx receipt to RunTxResult, allow BlockBuilder to build a block with zero txs

ryanio:
This PR:
* Adds `receipt` to `RunTxResult`
  * Moves the tx receipt generation logic from `runBlock` to `runTx`
  * Requested by @alcuadrado
* Fixes `BlockBuilder` to allow building a block with zero txs
* `AccessListEIP2930Transaction`: adds alias `type` for `transactionType` so it can be interpreted correctly when passed to `fromTxData`
* `TransactionFactory`: For extra safety, adds `toBuffer` on `txData.type` in case a 0x-prefixed hex string is passed in
* Adds types test from @alcuadrado to ensure that the actual objects can be safely used as their data types
* Adds test cases for legacy and access list transactions to `runBlock`
* `BlockBuilder`: Moves the `stateManager.commit` to after putting the block in blockchain in case it throws on validating

### DIFF
--- a/packages/block/src/types.ts
+++ b/packages/block/src/types.ts
@@ -1,6 +1,6 @@
 import { AddressLike, BNLike, BufferLike } from 'ethereumjs-util'
 import Common from '@ethereumjs/common'
-import { TxData, JsonTx } from '@ethereumjs/tx'
+import { AccessListEIP2930TxData, TxData, JsonTx } from '@ethereumjs/tx'
 import { Block } from './block'
 import { BlockHeader } from './header'
 
@@ -93,7 +93,7 @@ export interface BlockData {
    * Header data for the block
    */
   header?: HeaderData
-  transactions?: Array<TxData>
+  transactions?: Array<TxData | AccessListEIP2930TxData>
   uncleHeaders?: Array<HeaderData>
 }
 
--- a/packages/tx/src/baseTransaction.ts
+++ b/packages/tx/src/baseTransaction.ts
@@ -47,7 +47,7 @@ export abstract class BaseTransaction<TransactionObject> {
     } else {
       this._type = 0
     }
-    
+
     const toB = toBuffer(to === '' ? '0x' : to)
     const vB = toBuffer(v === '' ? '0x' : v)
     const rB = toBuffer(r === '' ? '0x' : r)
--- a/packages/tx/src/baseTransaction.ts
+++ b/packages/tx/src/baseTransaction.ts
@@ -8,7 +8,13 @@ import {
   ecsign,
   publicToAddress,
 } from 'ethereumjs-util'
-import { TxData, TxOptions, JsonTx, AccessListEIP2930ValuesArray } from './types'
+import {
+  TxData,
+  TxOptions,
+  JsonTx,
+  AccessListEIP2930ValuesArray,
+  AccessListEIP2930TxData,
+} from './types'
 
 /**
  * This base class will likely be subject to further
@@ -18,6 +24,8 @@ import { TxData, TxOptions, JsonTx, AccessListEIP2930ValuesArray } from './types
  * It is therefore not recommended to use directly.
  */
 export abstract class BaseTransaction<TransactionObject> {
+  private readonly _type: number
+
   public readonly nonce: BN
   public readonly gasLimit: BN
   public readonly gasPrice: BN
@@ -30,8 +38,16 @@ export abstract class BaseTransaction<TransactionObject> {
   public readonly r?: BN
   public readonly s?: BN
 
-  constructor(txData: TxData, txOptions: TxOptions = {}) {
+  constructor(txData: TxData | AccessListEIP2930TxData, txOptions: TxOptions = {}) {
     const { nonce, gasLimit, gasPrice, to, value, data, v, r, s } = txData
+
+    const type = (txData as AccessListEIP2930TxData).type
+    if (type !== undefined) {
+      this._type = new BN(toBuffer(type)).toNumber()
+    } else {
+      this._type = 0
+    }
+    
     const toB = toBuffer(to === '' ? '0x' : to)
     const vB = toBuffer(v === '' ? '0x' : v)
     const rB = toBuffer(r === '' ? '0x' : r)
@@ -59,9 +75,40 @@ export abstract class BaseTransaction<TransactionObject> {
   }
 
   /**
-   * Checks if the transaction has the minimum amount of gas required
-   * (DataFee + TxFee + Creation Fee).
+   * Returns the transaction type
    */
+  get transactionType(): number {
+    return this._type
+  }
+
+  /**
+   * Alias for `transactionType`
+   */
+  get type() {
+    return this.transactionType
+  }
+
+  /**
+   * EIP-2930 alias for `r`
+   */
+  get senderR() {
+    return this.r
+  }
+
+  /**
+   * EIP-2930 alias for `s`
+   */
+  get senderS() {
+    return this.s
+  }
+
+  /**
+   * EIP-2930 alias for `v`
+   */
+  get yParity() {
+    return this.v
+  }
+
   /**
    * Checks if the transaction has the minimum amount of gas required
    * (DataFee + TxFee + Creation Fee).
--- a/packages/tx/src/eip2930Transaction.ts
+++ b/packages/tx/src/eip2930Transaction.ts
@@ -38,6 +38,11 @@ export default class AccessListEIP2930Transaction extends BaseTransaction<Access
   get transactionType(): number {
     return 1
   }
+  
+  // Alias for transactionType
+  get type() {
+    return this.transactionType
+  }
 
   // EIP-2930 alias for `s`
   get senderS() {
--- a/packages/tx/src/eip2930Transaction.ts
+++ b/packages/tx/src/eip2930Transaction.ts
@@ -35,30 +35,6 @@ export default class AccessListEIP2930Transaction extends BaseTransaction<Access
   public readonly accessList: AccessListBuffer
   public readonly AccessListJSON: AccessList
 
-  get transactionType(): number {
-    return 1
-  }
-
-  // Alias for transactionType
-  get type() {
-    return this.transactionType
-  }
-
-  // EIP-2930 alias for `s`
-  get senderS() {
-    return this.s
-  }
-
-  // EIP-2930 alias for `r`
-  get senderR() {
-    return this.r
-  }
-
-  // EIP-2930 alias for `v`
-  get yParity() {
-    return this.v
-  }
-
   /**
    * Instantiate a transaction from a data dictionary
    */
@@ -143,7 +119,7 @@ export default class AccessListEIP2930Transaction extends BaseTransaction<Access
   public constructor(txData: AccessListEIP2930TxData, opts: TxOptions = {}) {
     const { chainId, accessList } = txData
 
-    super(txData, opts)
+    super({ ...txData, type: 1 }, opts)
 
     // EIP-2718 check is done in Common
     if (!this.common.isActivatedEIP(2930)) {
--- a/packages/tx/src/eip2930Transaction.ts
+++ b/packages/tx/src/eip2930Transaction.ts
@@ -38,7 +38,7 @@ export default class AccessListEIP2930Transaction extends BaseTransaction<Access
   get transactionType(): number {
     return 1
   }
-  
+
   // Alias for transactionType
   get type() {
     return this.transactionType
--- a/packages/tx/src/transactionFactory.ts
+++ b/packages/tx/src/transactionFactory.ts
@@ -1,4 +1,4 @@
-import { BN } from 'ethereumjs-util'
+import { BN, toBuffer } from 'ethereumjs-util'
 import Common from '@ethereumjs/common'
 import { default as Transaction } from './legacyTransaction'
 import { default as AccessListEIP2930Transaction } from './eip2930Transaction'
@@ -25,7 +25,7 @@ export default class TransactionFactory {
       // Assume legacy transaction
       return Transaction.fromTxData(<TxData>txData, txOptions)
     } else {
-      const txType = new BN(txData.type).toNumber()
+      const txType = new BN(toBuffer(txData.type)).toNumber()
       return TransactionFactory.getTransactionClass(txType, common).fromTxData(
         <AccessListEIP2930TxData>txData,
         txOptions
--- a/packages/vm/lib/buildBlock.ts
+++ b/packages/vm/lib/buildBlock.ts
@@ -4,9 +4,9 @@ import { BaseTrie as Trie } from 'merkle-patricia-tree'
 import { TypedTransaction } from '@ethereumjs/tx'
 import { Block, BlockOptions, HeaderData } from '@ethereumjs/block'
 import VM from '.'
-import { RunTxResult } from './runTx'
 import Bloom from './bloom'
-import { generateTxReceipt, calculateMinerReward, rewardAccount } from './runBlock'
+import { RunTxResult } from './runTx'
+import { calculateMinerReward, rewardAccount, encodeReceipt } from './runBlock'
 
 /**
  * Options for building a block.
@@ -117,7 +117,7 @@ export class BlockBuilder {
     for (const [i, txResult] of this.transactionResults.entries()) {
       const tx = this.transactions[i]
       gasUsed.iadd(txResult.gasUsed)
-      const { encodedReceipt } = await generateTxReceipt.bind(this.vm)(tx, txResult, gasUsed)
+      const encodedReceipt = encodeReceipt(tx, txResult.receipt)
       await receiptTrie.put(encode(i), encodedReceipt)
     }
     return receiptTrie.root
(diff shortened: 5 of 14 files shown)


packages/block/src/types.ts
```diff
@@ -93,7 +93,7 @@ export interface BlockData {
    * Header data for the block
    */
   header?: HeaderData
-  transactions?: Array<TxData>
+  transactions?: Array<TxData | AccessListEIP2930TxData>
```
alcuadrado, line 96: This may look like a breaking change, but it's not, as an `AccessListEIP2930TxData` is a `TxData`, right?

packages/tx/src/transactionFactory.ts
```diff
@@ -25,7 +25,7 @@ export default class TransactionFactory {
       // Assume legacy transaction
       return Transaction.fromTxData(<TxData>txData, txOptions)
     } else {
-      const txType = new BN(txData.type).toNumber()
+      const txType = new BN(toBuffer(txData.type)).toNumber()
```
alcuadrado, line 28: good catch! This is really tricky.

packages/vm/lib/runBlock.ts
```diff
@@ -84,49 +83,6 @@ export interface RunBlockResult {
   receiptRoot: Buffer
 }
 
-/**
```
alcuadrado, line 87: Technically removing these is a breaking change. To avoid it you can re-export what's now in runTx.

packages/vm/lib/runBlock.ts
```diff
@@ -420,66 +378,17 @@ export async function rewardAccount(
 }
 
 /**
- * Generates the tx receipt and returns { txReceipt, encodedReceipt, receiptLog }
+ * Returns the encoded tx receipt.
  */
-export async function generateTxReceipt(
```
alcuadrado, line 425: This is also a breaking change. Maybe just deprecate it and leave it here?

packages/vm/lib/runTx.ts
```diff
@@ -383,3 +407,64 @@ function txLogsBloom(logs?: any[]): Bloom {
...
+
+/**
+ * Returns the tx receipt.
+ */
+export async function generateTxReceipt(
+  this: VM,
+  tx: TypedTransaction,
+  txResult: RunTxResult,
+  blockGasUsed: BN,
+  debug?: Function
```
alcuadrado, line 419: Why this param?

alcuadrado:
I'd also add a test that does `runBlock` with a legacy and an access list transaction. That would help ensure that the problem with recreating txs doesn't come back.

packages/block/src/types.ts
ryanio: yes it simply extends `TxData` with 3 new optional fields: chainId, accessList, type.

packages/vm/lib/runTx.ts
ryanio: If a `debug` instance is passed in, it will do `debug(log)` for the tx receipt debug log message.

packages/vm/lib/runBlock.ts
ryanio: thanks, done in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1185/commits/a6c622e486246765953425c06ef9a1b427bf536f

packages/vm/lib/runBlock.ts
ryanio: good idea, done in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1185/commits/a6c622e486246765953425c06ef9a1b427bf536f

packages/vm/lib/runBlock.ts
```diff
@@ -419,16 +384,31 @@ export async function rewardAccount(
...
+    return encoded
+  }
+
+  const type = intToBuffer(tx.transactionType)
+  return Buffer.concat([type, encoded])
+}
+
 /**
  * Generates the tx receipt and returns { txReceipt, encodedReceipt, receiptLog }
+ * @deprecated Please use the new `generateTxReceipt` located in runTx.
```
holgerd77, line 403: Without having taken a look: a just-in-case reminder that these kind of things should be added to the v6 planning issue so that we will not forget.

packages/vm/lib/runTx.ts
holgerd77: Ah, let's please not start now to add `debug` flags to every single method, this will get messy soon and should be handled more carefully, especially since this is not specially urgent or something.

Have opened a corresponding issue on the overall eventual problem set: #1186

holgerd77:
Updated the branch here (have tested locally with a rebase before).

packages/vm/lib/runTx.ts
ryanio: @holgerd77 i agree, this was just a minimally invasive way for me to keep the debug log without too much additional complexity. do you have a suggestion on an alternative? i thought it'd be mostly harmless since it's optional.

packages/vm/lib/runBlock.ts
ryanio: thanks, added

packages/vm/lib/runTx.ts
holgerd77: I am not completely getting the context, what does "keep" mean here? Can't this just run in the normal `debug` context we are using all over the library?

ryanio: @holgerd77 take a look at this runTx `generateTxReceipt` function, there is a debug log message assembled inside with some of the tx receipt information.

packages/tx/src/eip2930Transaction.ts
```diff
@@ -40,6 +40,11 @@ export default class AccessListEIP2930Transaction extends BaseTransaction<Access
   get transactionType(): number {
     return 1
   }
+  
+  // Alias for transactionType
+  get type() {
+    return this.transactionType
+  }
```
holgerd77, line 47: If we add this here, I guess we also want to add this alias to the legacy tx where we [added type 0](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/legacyTransaction.ts#L19) (or better: move both getters up to `BaseTransaction` and set the tx type dynamically (e.g. on a private member `_type` which is then accessed from the getters) when calling the super class. Then we are sure that these aliases are available in all subsequent tx type implementations).

packages/vm/lib/runBlock.ts
```diff
@@ -84,49 +83,6 @@ export interface RunBlockResult {
...
- */
-export interface PostByzantiumTxReceipt extends TxReceipt {
-  /**
-   * Status of transaction, `1` if successful, `0` if an exception occured
-   */
-  status: 0 | 1
-}
-
-// EIP290Receipt, which has the same fields as PostByzantiumTxReceipt
-export interface EIP2930Receipt extends PostByzantiumTxReceipt {}
```
holgerd77, line 128: I hate to say this, but moving the receipts from here is likely a breaking change - if people use this importing from `lib/runBlock` then this will break now. So we should likely re-export here and add a deprecation note here and a removal note on the v6 issue?

packages/vm/lib/runTx.ts
holgerd77: I would still prefer here to just use the `debug` object generated in the `runTx.ts` module and not pass things over: just tested this by removing the parameter from the function signature and then run `DEBUG=vm:* npx tape -r ts-node/register --stack-size=1500 ./tests/api/buildBlock.spec.ts` and this works fine:

image

(or am I missing a usage context where this is not available?)

packages/tx/src/eip2930Transaction.ts
holgerd77: (I think it is super-crucial that we unify as much of the API as possible, so that we are not getting lost on the common API set (which I have the impression we already do a bit to some extend))

holgerd77: I am e.g. still struggling with these `v`, `r`, `s` aliases `senderS()`, `senderR()` and `yParity()` which is another form of API diffusion. Still wonder if we should directly deprecate this again in favor of a unified `v`, `r` and `s` access which works across all tx formats.

packages/vm/tests/api/types.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+    // Access List tx
+    const accessListTx: RequiredExceptOptionals<
+      AccessListEIP2930TxData,
+      OptionalTxFields
+    > = AccessListEIP2930Transaction.fromTxData({}, { common })
+    st.ok(accessListTx, 'accessList tx')
+
+    st.end()
+  })
+})
```
holgerd77, line 43: Super-impressive new type of tests (I am barely able to decipher this though TBH 😬 , have some intuition what it does though 😋 )

holgerd77: Update: ah, maybe you *can* add some 2-3 comment here upon the `type` definition and the appliance to give some hints how this actually works for fellow devs.

packages/vm/lib/runBlock.ts
```diff
@@ -12,6 +12,13 @@ import type { RunTxResult } from './runTx'
 import type { TxReceipt } from './types'
 import * as DAOConfig from './config/dao_fork_accounts_config.json'
 
+// For backwards compatibility from v5.3.0,
+// TxReceipts are exported. These exports are
+// deprecated and may be removed soon, please
+// update your imports to the new types file.
+import { PreByzantiumTxReceipt, PostByzantiumTxReceipt, EIP2930Receipt } from './types'
+export { PreByzantiumTxReceipt, PostByzantiumTxReceipt, EIP2930Receipt }
```
holgerd77, line 20: Ah, just seeing, ignore the other comment. Have added an additional note to https://github.com/ethereumjs/ethereumjs-monorepo/issues/1024.

holgerd77: (I am so so looking forward towards the next round of breaking releases 😛 , nevertheless think we need to have some patience there)

packages/vm/tests/api/runBlock.spec.ts
```diff
@@ -346,3 +346,70 @@ tape('runBlock() -> API return values', async (t) => {
...
+    // set `from` to a genesis address with existing balance
+    const address = Address.fromString('0xccfd725760a68823ff1e062f4cc97e1360e8d997')
+    tx.getSenderAddress = () => {
+      return address
+    }
+
+    await simpleRun(vm, [tx])
+    t.end()
+  })
+})
```
holgerd77, line 415: So, for the release notes: did `runBlock()` previously not work with typed txs or are these just additional tests for safety reasons?

packages/vm/lib/buildBlock.ts
```diff
@@ -233,7 +229,13 @@ export class BlockBuilder {
     const blockData = { header: headerData, transactions: this.transactions }
     const block = Block.fromBlockData(blockData, blockOpts)
     await this.vm.blockchain.putBlock(block)
```
holgerd77, line 231: I generally find it questionable if it is always desired that the generated `Block` is always added to the VM blockchain, I e.g. wonder if this will play well with our current client design (sorry, didn't stumble upon this earlier). Do we might want to at least add a flag `addToBlockchain` or something so that this can be made optional?

holgerd77 (commented):
Great PR in general 😀 , have some further remarks though which needs to be addressed.

Realistically I think it is too rushed to still do a release here (and also including the tx bug reports) still until the end of the week, but we can target early next week for a new release.

holgerd77:
Update: ah, took a step back here. Maybe we *can* do a release tomorrow, but very much depends on the subsequent discussion and decisions here.

Generally #1189 is a lot more pressing, in doubt we should do a bugfix release in between tomorrow.

packages/tx/src/eip2930Transaction.ts
ryanio: good idea I will move the type alias to the BaseTransaction so it will always be available.

yeah, the `senderS` `senderR` and `yParity` are from the [EIP-2930 spec](https://eips.ethereum.org/EIPS/eip-2930), I'm not sure why the naming was changed from the legacy spec. We could move those aliases to the BaseTransaction as well.

packages/vm/lib/runBlock.ts
ryanio: https://github.com/ethereumjs/ethereumjs-monorepo/pull/1185#discussion_r609558696 👍

packages/vm/tests/api/types.spec.ts
ryanio: yeah this was from @alcuadrado who whipped up this magic in the discord 🙂 i will add some comments as dev notes.

packages/vm/lib/runTx.ts
ryanio: ohh I see what you mean now, sorry it took me a second, I didn't realize `const debug` was created in the global scope, so that makes it easy it can just be used, I will remove the func param 🙂

packages/vm/tests/api/runBlock.spec.ts
ryanio: there was a bug in the block constructor would recreate a AccessListEIP2930Transaction as a legacy transaction (due to the missing `type` field, solved by the newly introduced alias). these tests were added to ensure the bug was fixed, and so it won't be reintroduced in the future (https://github.com/ethereumjs/ethereumjs-monorepo/pull/1185#issuecomment-813735865)

packages/vm/lib/buildBlock.ts
ryanio: My reasoning for this is that I wanted the `BlockBuilder` to be ready to build another block as soon as the prior one is finished. I didn't want to the consumer of the API to be responsible for keeping track of what needs to be done in between, preferring for it to be fully self resolving. Also, there are validation checks in `putBlock` that I think are important if they throw then the block wasn't valid from the perspective of the blockchain.

I see what you mean from the perspective of the client though, after having built a block we may not necessarily want to insert it into the blockchain right away?

I know @alcuadrado had to remove his own `putBlock` to accommodate this. @alcuadrado having done that, would you prefer this step remain like this? Or do you think it is a good suggestion from @holgerd77 to add an option for  `addToBlockchain`? Would you prefer it to be default true or false?

packages/tx/src/eip2930Transaction.ts
alcuadrado: Just a small note: These are called `r`, `s`, and `v` in the RPC

packages/vm/lib/buildBlock.ts
alcuadrado: I found it unexpected, but not critical. The fact that creating a block modifies the state already makes this API generate side effects, so using it to create a block that won't be will always require some special reasoning.

packages/tx/src/eip2930Transaction.ts
ryanio: i added this in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1185/commits/2ea2ebc4095530e5e63a5cc983ec752c9d663a71, please review for me

packages/vm/tests/api/types.spec.ts
ryanio: added in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1185/commits/b16fd4daa910c76bf19b10866d71b1f4e33192df, let me know if the comment explains sufficiently

packages/vm/lib/runTx.ts
ryanio: fixed in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1185/commits/8cb3c3ca0addcb3c40a33e4504a6c8ea13e601ea

packages/tx/src/eip2930Transaction.ts
holgerd77: Ok, if these name remain r, s, v even in the RPC I will deprecate the `senderR`, `senderS` and `yParity` getters. This was introduced to correctly match the spec names, but I guess we loose more here than we gain by bloating the API. Will push here before review.

@ryanio changes from your additions look good, thanks! 😄

packages/vm/tests/api/types.spec.ts
holgerd77: Yeah, that's good for now.

packages/vm/lib/buildBlock.ts
holgerd77: Ok, then we can always add such an option later if requested from whatever direction.

holgerd77 (approved):
LGTM, thanks a lot! 😄

packages/tx/src/baseTransaction.ts
```diff
@@ -59,9 +75,40 @@ export abstract class BaseTransaction<TransactionObject> {
...
+   * EIP-2930 alias for `s`
+   */
+  get senderS() {
+    return this.s
+  }
+
+  /**
+   * EIP-2930 alias for `v`
+   */
+  get yParity() {
```
alcuadrado, line 108: I don't think this makes sense here. yParity is `1` or `0`, and `v` can have other values in the legacy txs.

holgerd77: @ryanio if you have time it would be great if you move over again (then all three alias methods for consistency) to the `EIP2930`tx class in the release PR.

ryanio: @holgerd77 sure, done in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1192/commits/6078234526d79721e7bc9b41e0faa7fa1bbe9dfb

holgerd77: thanks, sorry for the back-and-forth

PR #1187, opened 2021-04-06
Client: central event bus implementation

### LINKED ISSUE #1079: Client: Centralized Event Bus
We likely want to add more events over time in the client, there was already an idea for a new 'badpeer' event here https://github.com/ethereumjs/ethereumjs-monorepo/pull/1075#discussion_r567690721 and likely others will follow.

For this not more and more bloating the code base by the need of chaining these events from class to class I described the idea of a centralized event bus (some calls ago I guess).

The idea is to add a parameter `events` to the `Config` class (which is already available everywhere), which inherits from the `Event` class and can therefore emit and receive events. In turn the `Event` inheritance on all other classes would be removed.

It then get's possible to receive all events in all classes without chaining.

To give some additional orientation I would also add some namespaces to the event names like:

- 'fetched' -> 'fetcher:fetched'
- 'added' -> 'peer:added'
- ...

Since I am not 100% sure if this idea might have side effects I am not anticipating properly (does this have other advantages I am not seeing right now if the class who deals with the logic is the EventHandler itself) I am dropping this here. Would be happy for some judgement - @cgewecke might be a particularly good candidate here, other opinions welcome as well of course - since I would want to start with this relatively soon (next week or so).

holgerd77:
Closes #1079


ryanio:
Just pushed a commit up, trying something simple:

```typescript
export enum Events {
  EXECUTION_ERROR = 'execution:error',
}
export declare interface EventBus {
  emit(event: Events.EXECUTION_ERROR, args: (error: Error) => void): boolean;

  on(event: Events.EXECUTION_ERROR, listener: (error: Error) => void): this;
}
export class EventBus extends EventEmitter {}
```

To test the type safety, changing from this (correct):

```typescript
this.config.events.emit(Events.EXECUTION_ERROR, error)
```

to this:

```typescript
this.config.events.emit(Events.EXECUTION_ERROR, 'test')
```

results in this ts error:

```
Argument of type 'string' is not assignable to parameter of type '(error: Error) => void'.ts(2345)
```

ryanio:
I added another commit to dedupe the parameters so they only need to be written once, introducing `EventParams`.
I also renamed the enum to singular `Event`, as that seems to be the preferred grammar for enums.

```typescript
export interface EventParams {
  [Event.EXECUTION_ERROR]: (error: Error) => void,
}
```

thus `EventBus` looks like:

```typescript
export declare interface EventBus {
  emit(event: Event.EXECUTION_ERROR, args: EventParams[Event.EXECUTION_ERROR]): boolean

  on(event: Event.EXECUTION_ERROR, listener: EventParams[Event.EXECUTION_ERROR]): this
}
```

holgerd77:
@ryanio thanks for picking this up, great! 😄 

I think we need to be more consistent on a naming scheme and just simply add all the namespaces from folders and classes so that it remains/gets transparent where events have its origin, have added this in 49894fc.

Also a question if we want to keep the `string` part of the enum `Event`: `SYNC_EXECUTION_VM_ERROR = 'sync:execution:vm:error'`? Is this necessary respectively helpful or does this just introduce redundancy?

I've tried to apply this solution to a second event from the fetcher in 6712805. This is already getting tricky though due to the use of generics, I am not quite sure how to properly type in this context? Is this easily doable without a significant refactoring effort here? 🤔

holgerd77:
image

ryanio:
Ok I was able to simplify the types even further with some generic typing:

```typescript
export declare interface EventBus<T extends Event> {
  emit(event: T, args: EventParams[T]): boolean
  on(event: T, listener: EventParams[T]): this
}
export class EventBus<T extends Event> extends EventEmitter {}

```

Now we don't have to add new `emit` and `on` every time for new enum values. 

The last thing I would like to figure out is if we can automatically type this `EventBusType` type from all the enum values instead of manually writing the union, but haven't found a good solution yet.

```typescript
export type EventBusType = EventBus<Event.SYNC_EXECUTION_VM_ERROR> &
  EventBus<Event.SYNC_FETCHER_FETCHED>
```

packages/client/lib/sync/fetcher/fetcher.ts
```diff
@@ -264,7 +264,7 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable
       try {
         await this.store(result)
         this.finished++
-        this.emit('fetched', result)
+        this.config.events.emit(Event.SYNC_FETCHER_FETCHED, result as any)
```
holgerd77, line 267: Just had another short look here, do we now always have to cast arguments on `emit` as `any` and the typing benefits are just on the listening side or is this some special case here? 🤔

ryanio: This one is a special case since `result` is of type `StorageItem[]` that is  `Block[] | BlockHeader[]`.

We can see if we can improve this local code to better infer `result` as `Block[] | BlockHeader[]` before emitting it, then we wouldn't need the `any`.

holgerd77: Ok, but generally we can go ahead here with this PR, right?

holgerd77:
Rebased this.

holgerd77:
Hi @ryanio, I wanted to give this another look today but got again lost in this TypeScript magic and directly stumbled upon things where I am not sure how to continue:

1. Types are still not inferred for me on the emit side, e.g. for `this.config.events.emit(Event.SYNC_EXECUTION_VM_ERROR, error)` I get `(local var) error: any` on mouseover
2. I wanted to start with the `Blockchain` `updated` event (in `chain.ts`), this has zero arguments and is (therefore) not working, tried with `[Event.CHAIN_UPDATED]: () => void` and `this.config.events.emit(Event.CHAIN_UPDATED)`, this then errors with "Expected 2 arguments, but got 1.", tried to change to `[Event.CHAIN_UPDATED]: undefined`, also with no luck

Could you please give this a prioritized look and also do 2-3 iterations (so: additional events) yourself and see if this really holds? I am starting to get a bit frustrated here, this is my 2nd or 3rd try and I am always getting stuck and I would need this for various things to continue (the tx pool e.g.).

Thanks! 😄

holgerd77:
Hi @ryanio, I wanted to give this another look today but got again lost in this TypeScript magic and directly stumbled upon things where I am not sure how to continue:

1. Types are still not inferred for me on the emit side, e.g. for `this.config.events.emit(Event.SYNC_EXECUTION_VM_ERROR, error)` I get `(local var) error: any` on mouseover
2. I wanted to start with the `Blockchain` `updated` event (in `chain.ts`), this has zero arguments and is (therefore) not working, tried with `[Event.CHAIN_UPDATED]: () => void` and `this.config.events.emit(Event.CHAIN_UPDATED)`, this then errors with "Expected 2 arguments, but got 1.", tried to change to `[Event.CHAIN_UPDATED]: undefined`, also with no luck

Could you please give this a prioritized look and also do 2-3 iterations (so: additional events) yourself and see if this really holds? I am starting to get a bit frustrated here, this is my 2nd or 3rd try and I am always getting stuck and I would need this for various things to continue (the tx pool e.g.).

Thanks! 😄

ryanio:
@holgerd77 ok pushed a commit!

1. The `error` from a catch clause has `any` type annotation since technically anything can be thrown, not just an `Error`. We could cast the type as an `Error` before doing the `.emit()`, or do `.emit(_, error as Error)` but I don't see any harm in the current setup. I'm not getting any errors on this line locally.
2. ah yes we haven't tried one with no args yet, this one took me a while but I updated the `emit` and `on` definitions to rest params and it seems to be guarding everything properly for me now. If you can give it a try and let me know if you come across any issues.

Btw, when creating a new event, have to:
1. Add to `Event` enum
2. Append to `EventBusType`: `& EventBus<Event.NEW_EVENT_NAME>`

I can continue adding some more event types and see if any other issues arise. Do you have any suggestion on which events to add next? I can get started on some related ones.

holgerd77:
@ryanio thanks! 😄 

No, no specific suggestions, if you want to continue a bit, just pick the next events you stumble upon. 😛

acolytec3:
I'm just dipping my toes into this with the `PEER_CONNECTED` commit. I need to look more at the `libp2pserver` and see if this same event can be used there as well.

packages/client/test/net/peer/libp2ppeer.spec.ts
```diff
@@ -31,7 +32,7 @@ tape('[Libp2pPeer]', async (t) => {
   t.test('should connect to peer', async (t) => {
     const config = new Config({ loglevel: 'error' })
     const peer = new Libp2pPeer({ config })
-    peer.on('connected', () => {
+    config.events.on(Event.PEER_CONNECTED, () => {
```
ryanio, line 35: so if all peers are emitting on this central bus then the event needs to return the peer or peerId for comparison and return if it isn't the right peer

packages/client/lib/sync/sync.ts
```diff
@@ -100,7 +101,9 @@ export abstract class Synchronizer extends EventEmitter {
     }, this.interval * 30)
     while (this.running) {
       try {
-        if (await this.sync()) this.emit('synchronized')
+        if (await this.sync()) {
+          this.config.events.emit(Event.SYNC_SYNCHRONIZED)
```
acolytec3, line 105: @ryanio What is this event doing?  The name seems to indicate it's emitted when the chain is fully synced and we're at the tip, but I didn't think we had anything implemented around this and what we're putting in the logs is just "Sychronized" without saying what was synchronized.

ryanio: Yep the `sync()` promise resolves when finished syncing with the peer. For example see the small blockchains set up and synced in the integration tests.

acolytec3: What do you think about adding some additional detail to the log message, maybe something like `Synchronized chain with best peer: ${peer.address} at height ${this.chain.blocks.height`, to give clarity to this event?

ryanio: yeah that's a good idea!

acolytec3:
@ryanio @holgerd77 I think this one is ready for review.  I believe I've got all the events accounted for except for the sender ones.  My logic in excluding the sender events is that the `Sender` class doesn't really have any special identifying properties that would be helpful in tracking those events if they were sent to a centralized eventbus so as far as I can tell there isn't a way to track sender messages back to individual sender objects so would require a great amount of rework for very little gain (I think).  One more to-do will be updating the typedocs once we settle on the final set of events/api.

packages/client/lib/net/peerpool.ts
```diff
@@ -62,11 +63,13 @@ export class PeerPool extends EventEmitter {
     if (this.opened) {
       return false
     }
-    this.config.servers.map((s) => {
-      s.on('connected', (peer: Peer) => {
+    // TODO: Do we still need this map now that we have the centralized event bus?
```
ryanio, line 66: I don't think the map is needed any more, as it would subscribe multiple times

packages/client/lib/net/peerpool.ts
```diff
@@ -62,11 +63,13 @@ export class PeerPool extends EventEmitter {
...
+    this.config.servers.map(() => {
+      this.config.events.on(Event.PEER_CONNECTED, (peer: any) => {
```
ryanio, line 68: do we need this `any` casting on `peer`? I'm seeing it in a lot of places, ideally it would automatically infer the type itself.

If the type is too complicated (looks like it's currently `[Libp2pPeer | RlpxPeer | Devp2pRlpxPeer | MockPeer | Peer]`) could we just simplify it to `Peer` since they all inherit from that?

packages/client/lib/net/peer/rlpxpeer.ts
```diff
@@ -112,27 +112,26 @@ export class RlpxPeer extends Peer {
...
+      this.config.events.emit(Event.PEER_ERROR, error, '')
     })
     this.rlpx.once('peer:added', async (rlpxPeer: Devp2pRlpxPeer) => {
       try {
         await this.bindProtocols(rlpxPeer)
-        this.emit('connected')
+        this.config.events.emit(Event.PEER_CONNECTED, rlpxPeer)
       } catch (error) {
-        this.emit('error', error)
+        this.config.events.emit(Event.PEER_ERROR, error, '')
```
ryanio, line 122: what's this third param supposed to be? the peer ID? I'm seeing a lot of empty strings

packages/client/lib/types.ts
```diff
@@ -1,15 +1,98 @@
...
+export interface EventParams {
+  [Event.CHAIN_UPDATED]: []
+  [Event.SYNC_EXECUTION_VM_ERROR]: [Error]
+  [Event.SYNC_FETCHER_FETCHED]: [Block[] | BlockHeader[]]
+  [Event.SYNC_SYNCHRONIZED]: [BN]
+  [Event.SYNC_ERROR]: [Error]
+  [Event.SYNC_FETCHER_ERROR]: [Error, any, any]
+  [Event.PEER_CONNECTED]: [Libp2pPeer | RlpxPeer | Devp2pRlpxPeer | MockPeer | Peer]
+  [Event.PEER_DISCONNECTED]: [Libp2pPeer | RlpxPeer | Devp2pRlpxPeer | MockPeer | Peer]
+  [Event.PEER_ERROR]: [Error, string]
```
ryanio, line 42: Unfortunately we are not on TypeScript 4.x yet where we can have [labeled tuple elements](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#labeled-tuple-elements), so this section is looking a bit confusing without context of what each type is supposed to represent (although most can be inferred, I think it would be good to have them spelled out to reduce confusion). Can we add a comment after each one labeling what the types are supposed to represent, and after our TS is upgraded we can inline them into the actual code (which should be great for autocomplete). I can help add them too :)

e.g.

`[Event.SYNC_SYNCHRONIZED]: [BN] // [headBlockNum]`

`[Event.PEER_ERROR]: [Error, string] // [error, peerId]`

then after our TS is `>=v4` we'll be able to do:

`[Event.SYNC_SYNCHRONIZED]: [headBlockNum: BN]`

`[Event.PEER_ERROR]: [error: Error, peerId: string]`

ryanio:
looks really great @acolytec3, nice work! just a few comments from me but otherwise this is looking spiffy 👍

holgerd77:
Pretty cool! 🎉

packages/client/lib/net/peerpool.ts
acolytec3: I struggled with this one because it popped up over and over again.  I think I just figured it out after you asked this question but the problem appears to arise when I use that `Devp2pRlpxPeer` type that's borrowed from `devp2p` since I guess that's a different type entirely from `Peer | RlpxPeer | Libp2pPeer` now that I look at it.  It looks like maybe my issue is really where I'm passing that `Devp2pRlpxPeer` with the event (as [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/121a617f94791477f25174c912ba5b066922cc01/packages/client/lib/net/peer/rlpxpeer.ts#L120)).  If I read this right, I think I can safely change `rlpxPeer` to `this` and then I can remove that `Devp2pRlpxPeer` type and that will remove the need for any of this funny typecasting.  Will experiment and let you know.

packages/client/lib/net/peer/rlpxpeer.ts
acolytec3: I've re-worked this slightly and revised the `PEER_ERROR` event so it passed the Error object and then the `Peer` object.  The original string there was an artifact of legacy code that I hadn't properly dealt with before.  I've updated all the existing uses of this event to reflect this new definition.

packages/client/lib/types.ts
acolytec3: I've updated all of them except for the `SYNC_FETCHER_ERROR` one.  I looked that one again and I'm honestly not sure what the second and third parameters are for.  The event is emitted [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/57a142d0b9d5ce6c9ba2390eec07c470939b1931/packages/client/lib/sync/fetcher/fetcher.ts#L249) and it's not obvious to me what those two boolean values tell you.   Any ideas?

acolytec3:
I've tried to address all the comments.  The major changes are below:
- Added a new `SERVER_ERROR` event that is emitted by servers when their `error` method is called.
- Revised the parameter list in the `PEER_ERROR` event to reflect what we're really doing
- Revised the types on the `PEER_CONNECTED` and `PEER_DISCONNECTED` events to just reflect `PEER` since we don't need any of the specific properties added by the subtypes that were originally included.

Questions:
- Do you think we need the `CLIENT_LISTENING` and `SERVER_LISTENING` events since they seem sort of duplicative (like [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/57a142d0b9d5ce6c9ba2390eec07c470939b1931/packages/client/lib/client.ts#L86))?  I'm not sure if we need that level of granularity or not.
- What are the second and third parameters on the `SYNC_FETCHER_ERROR` event for (emitted [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/57a142d0b9d5ce6c9ba2390eec07c470939b1931/packages/client/lib/sync/fetcher/fetcher.ts#L249))?

Looks like there is a new integration test error that I'll track down and fix this afternoon.  Otherwise, we're getting there!

ryanio:
Looks great, thanks!

> * Do you think we need the `CLIENT_LISTENING` and `SERVER_LISTENING` events since they seem sort of duplicative (like [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/57a142d0b9d5ce6c9ba2390eec07c470939b1931/packages/client/lib/client.ts#L86))?  I'm not sure if we need that level of granularity or not.

Hm I could go either way, @holgerd77 wdyt?

> * What are the second and third parameters on the `SYNC_FETCHER_ERROR` event for (emitted [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/57a142d0b9d5ce6c9ba2390eec07c470939b1931/packages/client/lib/sync/fetcher/fetcher.ts#L249))?

Looks like the second param is the `task` and the third param is the `peer`. 

Nice work with the labels, I think that's way more understandable to look at now :)

holgerd77:
> > Do you think we need the CLIENT_LISTENING and SERVER_LISTENING events since they seem sort of duplicative (like here)? I'm not sure if we need that level of granularity or not.
> 
> Hm I could go either way, @holgerd77 wdyt?

I didn't look so very deep in the code, but on thinking about it I do not even know what `CLIENT_LISTENING` would even mean (listening to what?) (or is there some semantics which make sense in the Ethereum Client context?). From my understanding that's a dedicated property of a server that *a server* can (and should at some point) listen to requests to then give some proper answer I guess.

On the other hand for the servers we have (networking, RPC, other?) we should have dedicated events I guess.

Not sure if this makes sense, don't have the time to dig deeper. In doubt we can also rework selected event types later on, nothing fixed for eternity.

acolytec3:
@ryanio Alrighty, I removed the `CLIENT_LISTENING` and `CLIENT_ERROR` events and just left the server level equivalents to emit the appropriate logging.  I think we might be ready for final review at this point.  All the tests are passing, I think the integration tests are generally testing the correct stuff, and it all seems to be working (I think).

ryanio:
ok great, thanks for all your help on this @acolytec3! i will give this a final review now and if it looks good I will merge it in :)

acolytec3:
@ryanio Hmm, looks like the linter caught something. Wonder why that didn't show up locally. Let me fix that before any merging.

ryanio (approved):
LGTM!

PR #1190, opened 2021-04-08
Tx: fix AccessListEIP2930Transaction constructor bug with v=0

### LINKED ISSUE #1189: Transaction constructor fails to initialize falsy values
This is pattern that is probably present in other libs, at least block. The constructor uses non-boolean values as conditions to decide wether to initialize a field or no, so it's impossible to initialize them to falsy values.

Here's a way to reproduce it

```ts
const Common = require("@ethereumjs/common").default;
const { AccessListEIP2930Transaction } = require("@ethereumjs/tx");

const tx = new AccessListEIP2930Transaction(
  { v: 0 }, // 0 is a valid v value in eip 2930
  { common: new Common({ chain: "mainnet", hardfork: "berlin" }) }
);

console.log(tx.v) // this is undefined, but should be BN(0)
```

holgerd77:
Fixes #1189 

Ok, this should do it, have tested on the console and can confirm that this fixes the bug.

Have also unified the default value setting along the road since this would otherwise generally be dangerous to introduce inconsistencies and hard-to-track bugs in the future which just occur on a single constructor or so.

I have done pretty atomic commits to allow for a careful review. I have done these single transitions pretty cautiously though.

### DIFF
--- a/packages/tx/src/baseTransaction.ts
+++ b/packages/tx/src/baseTransaction.ts
@@ -33,6 +33,7 @@ export abstract class BaseTransaction<TransactionObject> {
   constructor(txData: TxData, txOptions: TxOptions = {}) {
     const { nonce, gasLimit, gasPrice, to, value, data, v, r, s } = txData
     const toB = toBuffer(to)
+    const vB = toBuffer(v)
     const rB = toBuffer(r)
     const sB = toBuffer(s)
 
@@ -43,7 +44,7 @@ export abstract class BaseTransaction<TransactionObject> {
     this.value = new BN(toBuffer(value))
     this.data = toBuffer(data)
 
-    this.v = v ? new BN(toBuffer(v)) : undefined
+    this.v = vB.length > 0 ? new BN(vB) : undefined
     this.r = rB.length > 0 ? new BN(rB) : undefined
     this.s = sB.length > 0 ? new BN(sB) : undefined
 
--- a/packages/tx/src/baseTransaction.ts
+++ b/packages/tx/src/baseTransaction.ts
@@ -33,17 +33,19 @@ export abstract class BaseTransaction<TransactionObject> {
   constructor(txData: TxData, txOptions: TxOptions = {}) {
     const { nonce, gasLimit, gasPrice, to, value, data, v, r, s } = txData
     const toB = toBuffer(to)
+    const rB = toBuffer(r)
+    const sB = toBuffer(s)
 
     this.nonce = new BN(toBuffer(nonce))
     this.gasPrice = new BN(toBuffer(gasPrice))
     this.gasLimit = new BN(toBuffer(gasLimit))
-    this.to = toB.length > 0 ? new Address(toBuffer(to)) : undefined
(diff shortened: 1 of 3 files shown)


holgerd77:
Just seeing that CI ist failing, don't have time any more now, if someone wants to pick this up during the day feel free to push additional commits.

fvictorio:
FWIW, this seems to fix the tests:

```diff
diff --git a/packages/tx/src/baseTransaction.ts b/packages/tx/src/baseTransaction.ts
index fd863097..a02dfe07 100644
--- a/packages/tx/src/baseTransaction.ts
+++ b/packages/tx/src/baseTransaction.ts
@@ -32,10 +32,10 @@ export abstract class BaseTransaction<TransactionObject> {
 
   constructor(txData: TxData, txOptions: TxOptions = {}) {
     const { nonce, gasLimit, gasPrice, to, value, data, v, r, s } = txData
-    const toB = toBuffer(to)
-    const vB = toBuffer(v)
-    const rB = toBuffer(r)
-    const sB = toBuffer(s)
+    const toB = toBuffer(to === "" ? "0x" : to)
+    const vB = toBuffer(v === "" ? "0x" : v)
+    const rB = toBuffer(r === "" ? "0x" : r)
+    const sB = toBuffer(s === "" ? "0x" : s)
 
     this.nonce = new BN(toBuffer(nonce))
     this.gasPrice = new BN(toBuffer(gasPrice))
```

That code is terribly ugly though, and I'm not familiar enough with this project to know the proper way to do that.

holgerd77:
@fvictorio thanks a lot for having a look here! 😄 

Also tested one failing test case here. Yes, seems that this was just accidentally working before. At some point we got stricter with the `toBuffer` method in the `Util` library and require a `0x` prefix for hex strings which should be converted, so consequently also the "0x" string which then converts to an empty buffer. In the implementation before this was "shadowed" by the initial conversions to `BN`. I have the impression this was also a bug before though, since `new BN('')` translates to `<BN: 0>` which then gets the zero Buffer on conversion with `toBuffer(new BN(''))`, so `<Buffer 00>`.

Hmm. Anyhow, I think your fix is actually the best we can do here (and it's actually not that ugly). I would pledge here to stay a bit loose for now (the alternative would be to fix on the testdata consumption in the VM test runner, since the strings without `0x` are just likely coming from the test data). But since this was implemented here before with also accepting "" we might want to allow this for (some) convenience.

Will directly push after writing here.

holgerd77:
Ok, have pushed. Have added this fix for the other values in `BaseTransaction` as well.

cgewecke (approved):
Looks good to me!

ryanio (approved):
looks good, thanks for the input everyone.

these bugs have been pretty serious for us and hard to detect. it would be nice to add some more robust tests with varying data types to ensure the end values always result in what we expect.

@alcuadrado also mentioned in chat:

> i think a linter rule to forbid non-boolean conditions is a must for a project this critical

I found [this rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md) which may be appropriate.

PR #1198, opened 2021-04-12
Remove all the debug traces from the vm

alcuadrado:
We are receiving complains from our users about the release candidate version we published running >6x slower than our previous version. 

The main difference between these versions is that the RC version uses v5, instead of v4, making it the first consumer of v5.

This performance regression is aligned with that @jochem-brouwer reported when experimenting with removing the traces and running the standard tests, so I instructed one of our users to remove them from their `node_modules`, and not only the performance regression was gone, but the test run actually faster.

This PR removes all the debug traces from the VM. I didn't remove them from other packages to keep the PR small, and because we don't use them, so it's not as urgent.

@holgerd77, do you think we can get this published today? We really need to publish a final version of our Berlin support on Wednesday, or otherwise there won't be any way to test smart contracts against Berlin, and we can't publish something with a regression this significant.

### DIFF
--- a/packages/vm/lib/evm/interpreter.ts
+++ b/packages/vm/lib/evm/interpreter.ts
@@ -1,4 +1,3 @@
-import { debug as createDebugLogger } from 'debug'
 import { Account, Address, BN } from 'ethereumjs-util'
 import Common from '@ethereumjs/common'
 import { StateManager } from '../state/index'
@@ -65,9 +64,6 @@ export default class Interpreter {
   _state: StateManager
   _runState: RunState
   _eei: EEI
-
-  // Opcode debuggers (e.g. { 'push': [debug Object], 'sstore': [debug Object], ...})
-  private opDebuggers: any = {}
 
   constructor(vm: any, eei: EEI) {
     this._vm = vm // TODO: remove when not needed
@@ -142,7 +138,7 @@ export default class Interpreter {
     }
 
     // Reduce opcode's base fee
-    this._eei.useGas(new BN(opInfo.fee), `${opInfo.name} (base fee)`)
+    this._eei.useGas(new BN(opInfo.fee))
     // Advance program counter
     this._runState.programCounter++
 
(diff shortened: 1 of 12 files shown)


packages/vm/lib/evm/interpreter.ts
```diff
@@ -191,27 +187,6 @@ export default class Interpreter {
       codeAddress: this._eei._env.codeAddress,
     }
 
-    // Create opTrace for debug functionality
-    let hexStack = []
-    hexStack = eventObj.stack.map((item: any) => {
-      return '0x' + new BN(item).toString(16, 0)
-    })
-
-    const name = eventObj.opcode.name
```
jochem-brouwer, line 200: Is it possible to regenerate this data in the `step` event? We have access to this `eventObj` in the event right? Then it is OK if all this is removed in this context.

alcuadrado: Yes, you can generate the same data.

holgerd77 (approved):
Ok (short version, have a long version too 😄 , but not enough time right now).

holgerd77:
I worked several hours on this `debug` integration and crafted this very carefully (only the UX/DevEx side, not the performance side), also for things like the associated documentation. I also personally need this for VM debugging and am very much convinced that this is highly useful for others to understand the VM (there are other cases of course where the functionality is not needed at all, e.g. for HardHat). So I would really love to see that we can reintegrate at some point in a performance sensitive way.

Totally understand though that we needed to remove here asap due to timing considerations.

packages/vm/lib/evm/interpreter.ts
jochem-brouwer: This change here changes the structure of the step event as mentioned in #1199

alcuadrado: Where's the change? I really don't get it.

jochem-brouwer: If one would listen to `step` like this;

```
vm.on('step', (obj: any) => {
// do stuff
let stack = obj.hexStack
// do stuff with stack (is now undefined)
})
```

Then this code now does not work anymore, since `hexStack` is not available anymore (and this `opTrace` thing is also not available anymore).

alcuadrado: That object is not passed to the step handler. Take a look at line 232:

```ts
return this._vm._emit('step', eventObj)
```

Only `eventObj` is passed, which I didn't modify.

Issue #1181, opened 2021-03-31
Unsettled promises shouldn't be stored.


alcuadrado:
When upgrading to v5 I discovered that there are multiple promises being stored, which is pretty dangerous and should be avoided. I've seen these two, but there's probably more:

* [`mclInitPromise` in the VM](https://github.com/ethereumjs/ethereumjs-monorepo/blob/486b992019572eb252ab93f150416a37f0d8162a/packages/vm/lib/index.ts#L24)
* [`initPromise` in the Blockchain](https://github.com/ethereumjs/ethereumjs-monorepo/blob/486b992019572eb252ab93f150416a37f0d8162a/packages/blockchain/src/index.ts#L127)

The reason this is dangerous is that they may be rejected before they get to be handled. In the case of `Blockchain`, there's [some code](https://github.com/ethereumjs/ethereumjs-monorepo/blob/486b992019572eb252ab93f150416a37f0d8162a/packages/blockchain/src/index.ts#L192-L194) that I believe tries to avoid it, but throwing in the `catch` of a promise just leads to a new one that's also rejected.

This is particularly problematic in Node 15+, as unhandled promise rejections now kill the process, even if the promise was stored. For example, this leads to a crash:

```js
class C {
  constructor() {
    this.promise = new Promise((resolve, reject) => {
      this.reject = reject;
    });
  }
}

const c = new C();
c.reject();

// This callback is meant to be executed in another event loop task
// which never gets to excecute. Node 15+ kills the process at the 
// end of a task that resulted in an unhandled promise rejection.
setImmediate(() => console.log("Here"));
```

How can these be done instead?

* In the case of `mclInitPromise`: I'd move its initialization into the opcode/precompile handlers that need them. Those are already async.
* In the case of `Blockchain`: Same as the VM, have an `init` method be called and await before every operation.

holgerd77:
👍  I often had some uneasy feelings about this when seeing this code parts but couldn't put it into words what would have bothered me since I am not so firm with promises, thanks for this write-up. Would say that we should change here soon, we also have this pattern in some parts of the client.

jochem-brouwer:
In both cases we could make the constructor of these classes `private` and instead provide a static constructor which awaits these initialization promises. (For VM we already have this, `VM.create`)

alcuadrado:
Yeah, I really like that pattern. Even more than the `await this.init()` one.

ryanio:
[...]

PR #1201, opened 2021-04-12
VM, Blockchain: fix storing unsettled promises

ryanio:
#1181

This PR resolves storing unsettled promises in `VM`, `Blockchain` by preferring to use `VM.create` and `Blockchain.create` instead of `new VM` and `new Blockchain` as this lets us safely run async init code.

`VM`: `mcl` is now initialized inside `VM.init()` and has no stored promise.
`Blockchain`: `init` is run during `Blockchain.create()` and has no stored promise.

Both classes have boolean `_isInitialized` and check internally to make sure they do not init more than once.

Sometimes in the client a `Blockchain` or `VM` is set up in a constructor. In this case, the `init` is added to the classes' `open` method.


holgerd77:
This PR has somewhat conflicting readyness settings, what's the status?

ryanio:
I don't think I can make any of the changes I would have liked in this PR without any breaking changes, since moving from stored to unstored promises is a breaking change itself (for example if anyone was using or relying on them).

I think we can put this PR on the back burner until we are ready to merge breaking changes for v6 where we can make some broader changes and refactor the async init code.

holgerd77:
@ryanio ok, have added to the v6 planning notes and added a "blocked" label here.

ryanio:
Phew, rebased this, that was not trivial :)

Holger I know you decided against this approach since private constructors block inheritance, so I am going to think more about it, maybe we just make them deprecated and note they are not private for the reason of inheritance but their use is strongly discouraged over the async methods.

holgerd77:
> Holger I know you decided against this approach since private constructors block inheritance, so I am going to think more about it, maybe we just make them deprecated and note they are not private for the reason of inheritance but their use is strongly discouraged over the async methods. (edit: ok I undid the private constructor and made sure both had deprecation messages)

That's what I also suggested yeah, but I have to admit that I am also still unsure. On the other hand it *would* be a huge win to have this somewhat ensured that devs are not unsafely calling into this. But I *do* am somewhat anxious on this that we might have not considered all use cases here and might put some unpractical restrictions on users by doing so.

Some additional thoughts I had here:

1. We can actually also use `protected` instead of `private`. This might generally be the better fit, not sure why I didn't think about this yet. Just tested and works as expected:

grafik

2. This change is actually a type "we-can-try-and-see-if-we-get-through-with-it" change. 😜 So on a breaking release it is actually both safe to introduce as well as easy to later remove. So *if* we are at least somewhat convinced that this is a good thing and assume that the positive outcome overweight the negatives we can actually try (then eventually version 1. with `protected`) and - while not pretty - we can still revert in some worst case scenario of a substantial amount of substantial negative community feedback.

packages/blockchain/src/types.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * The blockchain only initializes succesfully if it has a genesis block. If
+   * there is no block available in the DB and a `genesisBlock` is provided,
+   * then the provided `genesisBlock` will be used as genesis. If no block is
+   * present in the DB and no block is provided, then the genesis block as
+   * provided from the `common` will be used.
+   */
+  genesisBlock?: Block
+}
```
holgerd77, line 98: Where is this huge type file coming from? Is this a forgotten artifact from the interface PR which also happened around that time?

ryanio: ah no I was just moving them from `index.ts` because it was getting so long. I will undo this though because I see the benefit of keeping it along with the source code now.

ryanio:
oh nice, `protected` seems great. I will add that.

alcuadrado:
IMO protected constructors with async `create` can be great, as long as all the operation of the class are async or it's expected to be used in an async context. It can lead to much cleaner code, while it may lead to the consumer having to do the lazy `init()` thing, but I guess is not probable and ok in this context.

ryanio:
@alcuadrado great, thanks!

that's weird, typescript is erroring out (on local and ci) when I added `protected` in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1201/commits/2b91fe583a490cae38c6ec2553d1796201d8334a, I tried to google around and couldn't find an answer to this, I tried upgrading typescript to latest to see if there were any fixes as well and still same error:

```
> @ethereumjs/vm@5.5.2 build
> ../../config/cli/ts-build.sh

[Node build] Using tsconfig.prod.json
> tsc --build ./tsconfig.prod.json
[Node build] Working...
/Users/ryanghods/dev/ethereumjs-monorepo/packages/vm/node_modules/typescript/lib/tsc.js:35107
                else if (type.symbol.flags & (32 | 64)) {
                                     ^
TypeError: Cannot read property 'flags' of undefined
    at getBaseTypes (/Users/ryanghods/dev/ethereumjs-monorepo/packages/vm/node_modules/typescript/lib/tsc.js:35107:38)
    at typeHasProtectedAccessibleBase (/Users/ryanghods/dev/ethereumjs-monorepo/packages/vm/node_modules/typescript/lib/tsc.js:49112:29)
    at typeHasProtectedAccessibleBase (/Users/ryanghods/dev/ethereumjs-monorepo/packages/vm/node_modules/typescript/lib/tsc.js:49140:20)
    at isConstructorAccessible (/Users/ryanghods/dev/ethereumjs-monorepo/packages/vm/node_modules/typescript/lib/tsc.js:49157:25)
    at resolveNewExpression (/Users/ryanghods/dev/ethereumjs-monorepo/packages/vm/node_modules/typescript/lib/tsc.js:49085:22)
    at resolveSignature (/Users/ryanghods/dev/ethereumjs-monorepo/packages/vm/node_modules/typescript/lib/tsc.js:49357:28)
    at getResolvedSignature (/Users/ryanghods/dev/ethereumjs-monorepo/packages/vm/node_modules/typescript/lib/tsc.js:49375:26)
    at checkCallExpression (/Users/ryanghods/dev/ethereumjs-monorepo/packages/vm/node_modules/typescript/lib/tsc.js:49441:29)
    at checkExpressionWorker (/Users/ryanghods/dev/ethereumjs-monorepo/packages/vm/node_modules/typescript/lib/tsc.js:51593:28)
    at checkExpression (/Users/ryanghods/dev/ethereumjs-monorepo/packages/vm/node_modules/typescript/lib/tsc.js:51510:38)
```

packages/blockchain/src/types.ts
holgerd77: Ah, I guess on Blockchain we likely also want to restructure the exports and use the `index.ts` file just for exporting things and move code and types to dedicated files, a bit as we are planning for the VM (see #1024 VM TODOs).

If you agree respectively have no reservations you might directly want to do in this PR since this is breaking already?

holgerd77:
Am I correct that we would now redo (respectively: re-point) this towards `develop` and - also - that this actually *hasn't* been reopened (and eventually already merged) towards `develop` yet?

ryanio:
this hasn't been updated or merged into anything yet, a good candidate i can pick up when we are ready to bring it in :)

PR #556, opened 2019-07-23
Stateless execution prototype

s1na:
The goal of this PR is to prototype something akin to the [stateless client](https://ethresear.ch/t/the-stateless-client-concept/172) concept. There are two "roles":

- **Verifier** a light-client who wishes to process a future tx/block, but only has access to the current state root, and stores none of the actual state.
- **Prover** a full-node who runs a tx/block and generates all the necessary merkle proofs for the verifier to execute the tx. This includes proofs for accounts that were read during execution, codes for executed contracts and proofs for contract storage keys that were read.

The prover currently works by hooking the `StateManager` and generating proofs on invocation of `getAccount`, `getContractCode` and `getContractStorage`.

The verifier currently works by storing all received trie nodes in a new trie's db (with their hashes as keys), and using that trie in a new `StateManager`, and then running the tx.

To test this, I'm running the `GeneralStateTests` suite, running each case twice. Once to generate all the proofs, and a second time to test the verifier given proofs generated during the first run. There are still some edge cases that need to be fixed (38 cases are failing).

This also depends on a merkle-patricia-tree release (specifically https://github.com/ethereumjs/merkle-patricia-tree/pull/82, and probably https://github.com/ethereumjs/merkle-patricia-tree/pull/94).

PR #1205, opened 2021-04-14
Add wit protocol

ryanio:
WIP. This PR adds [`wit/0`](https://github.com/ethereum/devp2p/blob/master/caps/wit.md) to devp2p and client.


ryanio:
To get the trie hash nodes for a witness, I am thinking of adding a new option `reportTrieHashNodes` to `trie.get` that if true will collect and return the list of trie hash nodes accessed during the `get` call.

I haven't decided yet if it should return the info along the trie results (so if the boolean is true the return type contains the trieHashNodes data) or use the pattern in the VM to use an EventEmitter to emit the additional information, which I like but currently the trie doesn't have any EventEmitter usage so it would have to be newly added, but could then emit e.g. `trie.emit('trieHashNodes', queryKey, trieHashNodes)`

holgerd77:
@ryanio I am sceptical if an EventEmitter is a fitting choice here. If you are doing this you are detaching the the processing code on the caller side from the `get()` call, so there will be some structure like:

```typescript

trie.on('trieHashNodes', () => { // process trie node hashes, here it might be somewhat more difficult to grasp on the context }

// Build up some context, where are these hashes used for
trie.get(key, true) // would it be oversized to use an options dict here?
```

You loose on the context here or you have to manually rebuild, so I guess from that PoV it would be easier to have:

```typescript

const hashes = trie.get(key, true)
// here we have the context directly (why where these hashes requested) and can directly do something with them
```

Not sure, also don't want to directly push too much in this direction, but let's have some exchange on this first on the up and downsides.

So what are the concrete upsides? Is this easier to implement to not have this tracked or something?

ryanio:
that's a great point thanks for the feedback. i'll write some of this code today and we can evaluate it further based on how it looks.

i'm still digesting how to collect *all* the witnesses. the [wit spec](https://github.com/ethereum/devp2p/blob/master/caps/wit.md) says:

> Using the wit protocol, peers ask each other for the list of trie node hashes read during the execution of a particular block. This includes the following data:
>
> * Storage nodes
> * Bytecodes
> * Account nodes
> * Read during EVM execution
> * Read during transaction validation
> * Read during block reward calculation
> * Nodes read when generating the final state root (i.e. sometimes deleting data requires a trie refactor that reads nearby trie nodes)
> 
> The trie node hashes which are generated at the end of the block from existing data are not included. For example, the final state root hash is not included.

I asked @s1na and he said:
> Aha I was once working on sth similar and got the list of node hashes by monkey-patching trie._db.get as in [here](https://github.com/ethereumjs/ethereumjs-monorepo/pull/556/files#diff-67fff7fbd19b21c3a782aad2c3f9568e0c239fd5a7d05e523e9d971a54e69948R53)

holgerd77:
This looks great already. 😄

ryanio:
this is rather outdated now with the underlying codebase having changed. it can be used in the future to extract some of the wit logic, but not too much was connected and working, will close for now.

PR #1208, opened 2021-04-19
VM: Small fixes in events spec file

emersonmacro:
One of the tests in `events.spec.ts` intends to check that `EVMResult` is emitted after running a message, but the test was actually listening for the `beforeMessage` event. Changed the test to listen for `afterMessage`.

Also included a couple small copy-edits in an explanatory comment further down in the file.


ryanio (approved):
nice, lgtm, thanks for opening this PR :)

PR #1213, opened 2021-04-21
VM: New test for runTX return values

emersonmacro:
Added a new test for default return values in `runTx.spec.ts` (testing the non-optional fields in `RunTxResult`).

I'm still getting my head wrapped around the code and testing practices for this project, so feedback is most definitely welcome.

### DIFF
--- a/packages/vm/tests/api/runTx.spec.ts
+++ b/packages/vm/tests/api/runTx.spec.ts
@@ -306,7 +306,7 @@ tape('runTx() -> runtime errors', async (t) => {
 
 // TODO: complete on result values and add more usage scenario test cases
 tape('runTx() -> API return values', async (t) => {
-  t.test('simple run,common return values', async (t) => {
+  t.test('simple run, common return values', async (t) => {
     for (const txType of TRANSACTION_TYPES) {
       const vm = new VM({ common })
       const tx = getTransaction(vm._common, txType.type, true)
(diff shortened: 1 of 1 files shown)


holgerd77:
Have updated this branch.

holgerd77:
@emersonmacro thanks for the PR, this looks actually pretty good already, great! 😄 

Will leave this open for now for some discussion around the `runTx()` result format.

packages/vm/tests/api/runTx.spec.ts
```diff
@@ -334,4 +334,49 @@ tape('runTx() -> API return values', async (t) => {
...
+      t.deepEqual(
+        res.bloom.bitvector,
+        Buffer.from('00'.repeat(256), 'hex'),
+        `runTx result -> bloom.bitvector -> should be empty (${txType.name})`
+      )
+      t.deepEqual(
+        res.receipt.gasUsed,
+        res.gasUsed.toArrayLike(Buffer),
+        `runTx result -> receipt.gasUsed -> result.gasUsed as Buffer (${txType.name})`
+      )
```
holgerd77, line 368: This is actually not so beautiful that we here return the `gasUsed` as a `Buffer` and on the upper level as a `BN` value. 🙄  Should we add this to the list of breaking changes for v6 #1024 for an update (to `BN` I guess) or is there some reasoning here I overlook? //cc @ryanio

ryanio: ah yes I think we should update it to return a `BN`.

in general here I'd say that instead of using `t.deepEqual` we should use `BN`'s `eq` method, so for example `t.ok(res.gasUsed.eq(new BN(res.receipt.gasUsed)))`

I also like to do this with `Buffer`s as well: `t.ok(buffer1.equals(buffer2))`

holgerd77: Is this encompassing more checks or stricter or what's the reasoning for this?

holgerd77 (approved):
LGTM, thanks!

PR #1214, opened 2021-04-21
Tx: New hashMessage parameter for getMessageToSign()

### LINKED ISSUE #1188: Unsigned transaction serialization not exposed for EIP-155 use
When signing with a ledger app that parses the transaction itself, it is necessary to produce a serialization of a EIP-155 unsigned transaction. `_getMessageToSign` computes that serialization, but only returns the hash
https://github.com/ethereumjs/ethereumjs-monorepo/blob/00fc597a817edda5dfb0673027ead3a21dd309d4/packages/tx/src/legacyTransaction.ts#L137-L154

whereas `raw` always treats the transaction as signed 
https://github.com/ethereumjs/ethereumjs-monorepo/blob/00fc597a817edda5dfb0673027ead3a21dd309d4/packages/tx/src/legacyTransaction.ts#L112-L124

The current workaround is to reimplement `_getMessageToSign` downstream but without the hashing.

holgerd77:
Fixes #1188

### DIFF
--- a/packages/tx/src/eip2930Transaction.ts
+++ b/packages/tx/src/eip2930Transaction.ts
@@ -278,9 +278,11 @@ export default class AccessListEIP2930Transaction extends BaseTransaction<Access
   /**
    * Returns the serialized unsigned tx (hashed or raw), which is used to sign the transaction.
    *
-   * @param hashMessage - Return hashed message if set to true
+   * @param hashMessage - Return hashed message if set to true (default: true)
    */
-  getMessageToSign(hashMessage = true) {
+  getMessageToSign(hashMessage: false): Buffer[]
+  getMessageToSign(hashMessage?: true): Buffer
+  getMessageToSign(hashMessage = true): Buffer | Buffer[] {
     const base = this.raw().slice(0, 8)
     const message = Buffer.concat([Buffer.from('01', 'hex'), rlp.encode(base as any)])
     if (hashMessage) {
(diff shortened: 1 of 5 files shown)


ryanio:
looks good! I just added a commit to return the correct type based on the hashMessage param

packages/tx/src/eip2930Transaction.ts
```diff
@@ -278,9 +278,11 @@ export default class AccessListEIP2930Transaction extends BaseTransaction<Access
   /**
    * Returns the serialized unsigned tx (hashed or raw), which is used to sign the transaction.
    *
-   * @param hashMessage - Return hashed message if set to true
+   * @param hashMessage - Return hashed message if set to true (default: true)
    */
-  getMessageToSign(hashMessage = true) {
+  getMessageToSign(hashMessage: false): Buffer[]
+  getMessageToSign(hashMessage?: true): Buffer
+  getMessageToSign(hashMessage = true): Buffer | Buffer[] {
```
ryanio, line 285: I am still getting familiar with function overloading in TypeScript to better infer the returned type, so I'm a bit perplexed why I needed to add this final `: Buffer | Buffer[]` when the same exact line in [`legacyTransaction.ts:156`](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1214/commits/c1f14fed9751bdb495d508938331b038ba567e4c#diff-739632585d309f4741161671245094379f440cf7ed2d8047dc24cd16ced6e712R156) was fine without it (error was: `This overload signature is not compatible with its implementation signature.`)

holgerd77:
@ryanio oh yeah, great, that's definitely better, was feeling a bit bad about the return type! 😀

Can you approve?

ryanio (approved):

PR #1216, opened 2021-04-21
VM: remove StateManager.getStateRoot() force parameter and throwing behavior

### LINKED ISSUE #1209: VM: do not throw on StateManager.getStateRoot() on uncommitted checkpoints
Throwing on [StateManager.getStateRoot()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/a4af0b13c941292180a2077dfc3588a446d833aa/packages/vm/lib/state/stateManager.ts#L446) on uncommitted checkpoints has been reported by @yann300 from Remix and @alcuadrado from Hardhat as problematic behavior. From a functional perspective I guess we could just simply remove this since along checkpointing changes written to the trie are handled by the trie commit/revert logic.

From a performance perspective though this is dangerous behavior (I vaguely remember that this might have been the reason for an introduction in the first place, we really really should code-comment these kind of motivations whenever we think of it). Every `getStateRoot()` call triggers a cache flush (so that the trie is up-to-date) which is very costly, and this on a relatively "innocent" looking function. My fear is that people will start using this extensively and wonder why things get slow.

Several ideas for approaches:

1. We do it anyhow and document on the code docs
2. We'll keep the current behavior but change own calls like in [VM.runBlock()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/a4af0b13c941292180a2077dfc3588a446d833aa/packages/vm/lib/runBlock.ts#L144) now being done with `false` to a more foregiving call with `true`. This should at least address the Remix case.
3. We remove throwing and see if we can rewrite the function logic to be more clever on the cache flush or something, maybe this can be done in a non-performance-draining way

Thoughts? 🤔

holgerd77:
Fixes #1209


holgerd77:
This is *not* solving the delete-or-not-delete question yet, would like to address this in a separate PR (or someone else can).

holgerd77:
Just did a local API:browser test run here, tests are hanging (for at least 2+ minutes) in Firefox at '# should build a valid block' and CPU load is going to 100%. Not sure what's happening yet (and if this has to do with changes in this PR).

holgerd77:
Update: just observed that API tests were failing on `master` as well but were succeeding on the release PR on the tx library I just opened. So this might hint towards some performance issue along the mentioned test sometimes leading to a timeout. Not sure if this just is due to "normal" browser behavior or if there is some performance bottleneck within the Block Builder code which eventually needs to be adressed.

ryanio:
> Update: just observed that API tests were failing on `master` as well but were succeeding on the release PR on the tx library I just opened. So this might hint towards some performance issue along the mentioned test sometimes leading to a timeout. Not sure if this just is due to "normal" browser behavior or if there is some performance bottleneck within the Block Builder code which eventually needs to be addressed.

just looked into it, it was the use of `vm.stateManager.generateCanonicalGenesis()` that was quite slow. I was only using it to have some addresses with balance. I opened #1220 to remove it and instead just add balance directly to the specific accounts needed for the txs.

ryanio (approved):
lgtm

holgerd77:
@ryanio thanks for reviewing, also had another look here with fresh eyes and some distance and still feel confident on the change and the execution. 😄  Nevertheless wanted to give a short mention that this is a somewhat bigger change where things can go wrong on a greater scheme (introducing a malfunctioning cache e.g.), so this is one of these PRs where we should take some extra care.

Ok, words for the week. 😛  😅

Will merge.

ryanio:
I just realized that we should also update `state/interface.ts` for `getStateRoot`. Does removing the param lead to any breaking changes? Should we instead mark it with an underscore `_force`, and note and mark the param as deprecated and remove in next set of breaking changes?

holgerd77:
@ryanio ah no, I had thought this along. The parameter in the interface is just "out of effect" and simply does nothing - so it can remain for now, also added a note on v6 planning along #1024 with a reminder to remove from the interface. And our implementation proves to comply with the interface since we directly implement the interface. So on the `DefaultStateManager` we could still call with `getStateRoot(true)` if we want, this just simply has no effect.

Or is something flawed with this argumentation? 🤔

PR #1218, opened 2021-04-22
devp2p: New Release v4.0.0

holgerd77:
Side note: have realized that the docs of this library are still in "manual mode", so there is no doc generation but the docs are in README and needs to be updated manually (which is usually not done).

I can vaguely remember that - from the structure of the library - doc generation is somewhat challenging. We should nevertheless try to get to some setup here, this will otherwise always stay in some low-quality-documentation mode.


ryanio:
I have already added typedoc, I just found the documentation in the README well laid out so I didn't want to remove it. Above the last "Developer" section I added a "Docs" section that says: "For a complete API reference see the generated documentation." with a link to `./docs`

If the README is getting outdated and we are happy with the quality of the generated docs, we can remove it and just point to the generated docs, or if it just needs some shape up, I can do that too.

holgerd77:
Ah, sorry, overlooked that.

ryanio (dismissed):
np, you could regenerate the docs if you would like (just tested it on local and it ran fine), otherwise looks great!

holgerd77:
@ryanio ok, I've done that, also made the note on the API docs more prominent in the README, this is otherwise really easy to overlook, but accompanied with a note that there is also some manual entrypoint introduction following (would also say that the docs are currently not yet in the state to replace this since extremely noisy, first step would likely be to remove the `_` methods and properties I guess?).

Have additionally included the missed tx docs here and rebased. Will admin-merge and release after CI run since already approved and no further code additions.

PR #1219, opened 2021-04-22
Trie: new deleteFromDB option (default: false)

### LINKED ISSUE #1210: Trie: do not delete values from DB by default
As discussed extensively around the Hardhat `StateManager` bug found and analyzed by @alcuadrado and @fvictorio our current behavior in the Trie to delete nodes from disk on delete operations is error-prone and likely not desired, since this imposes the risk of setting a state root to an older root - in the context of the VM with `StateManager.setStateRoot()` but also directly on the trie - go wrong (without any notice).

This deleting functionality also doesn't count in as pruning yet since this is just a very small portion of the package with the big majority of state bloat occuring just through orphaned nodes by trie updates which just have no connection into the (current) trie any more. 

There nevertheless is likely a use case for this functionality if people want to to do proper state hygiene and manually want to clean up their trie along their code. So I think we shouldn't remove completely.

My current suggestion is to introduce a new option dict for the trie (defaulting to `{}` to keep this backwards-compatible) with a new option `performDbDeletes` (open for a better name). My current preference is to default this to `false` and then not do any VM changes, I guess this goes along as a bug fix and not something breaking and should not have broader side effects and this is at the end the behavior people want (I guess 😋 ). Open though for argumentation in the other direction, then we have to adopt our VM-internal trie instantiations to use this new option and are depending on a VM release here.

Implementation should be accompanied by some tests showing that trie updates with deletions are resulting in the exact same state (so: having the same root) with or without this option being set.

On another thought I would say (having had another look at `CheckpointDb.ts`) that this shouldn't have effects on the CP operations - including `revert`. Might need another confirmation though and eventually should be accompanied by some tests to make sure.

holgerd77:
Fixes #1210 

I am coming more and more to the conclusion that this might be the only change we need. We shouldn't comment out stuff (or remove) directly in the DB manager to let people the freedom to delete themselves if they wish, this is what would be expected from the direct DB class.

Some historical reference in between: `_formatNode` has been moved from `CheckpointTrie` to `BaseTrie` [here](https://github.com/ethereumjs/merkle-patricia-tree/commit/40b02ee1c84f5dcda7c1f295699b1244fdbaf22f#diff-a21bc7ca183ffb0c3fc3c310436fe3ceb155b33debfda6e132ae2afb42724a8e).

All Trie delete operations - both from `del()` and from `batch()` - are routed through this `_formatNode` method, so this should shield all possible high-level trie deletes.

I am unsure about the default setting of a potential new option `deleteFromDB`. My tendency is to switch to this keep-behavior to shield users from an unwanted delete behavior. However I am not sure about the side effects. This *will* lead to larger state growth - also e.g. in our client - if used for continuous trie operations. This would be an argument to start with remaining the delete-behavior (so set `deleteFromDB` to `true` by default) but then instantiate the default trie in the VM with `false`.

Hmm, very much unsure about these decisions.

Nevertheless this is my proposal what I would currently have a tendency towards.

### DIFF
--- a/packages/trie/src/baseTrie.ts
+++ b/packages/trie/src/baseTrie.ts
@@ -39,24 +39,30 @@ export type FoundNodeFunction = (
  * The API for the base and the secure interface are about the same.
  */
 export class Trie {
-  /** The backend DB */
-  db: DB
   /** The root for an empty trie */
   EMPTY_TRIE_ROOT: Buffer
-  private _root: Buffer
   protected lock: Semaphore
+
+  /** The backend DB */
+  db: DB
+  private _root: Buffer
+  private _deleteFromDB: boolean
 
   /**
    * test
    * @param db - A [levelup](https://github.com/Level/levelup) instance. By default (if the db is `null` or
    * left undefined) creates an in-memory [memdown](https://github.com/Level/memdown) instance.
    * @param root - A `Buffer` for the root of a previously stored trie
+   * @param deleteFromDB - Delete nodes from DB on delete operations (disallows switching to an older state root) (default: `false`)
    */
-  constructor(db?: LevelUp | null, root?: Buffer) {
+  constructor(db?: LevelUp | null, root?: Buffer, deleteFromDB: boolean = false) {
     this.EMPTY_TRIE_ROOT = KECCAK256_RLP
     this.lock = new Semaphore(1)
+
     this.db = db ? new DB(db) : new DB()
     this._root = this.EMPTY_TRIE_ROOT
+    this._deleteFromDB = deleteFromDB
+
     if (root) {
       this.setRoot(root)
     }
@@ -382,7 +388,7 @@ export class Trie {
   }
 
   /**
-   * Deletes a node from the database.
+   * Deletes a node from the trie.
    * @private
    */
   async _deleteNode(k: Buffer, stack: TrieNode[]): Promise<void> {
@@ -566,11 +572,13 @@ export class Trie {
       // is applied twice (performance)
       const hashRoot = keccak(rlpNode)
 
-      if (remove && this.isCheckpoint) {
-        opStack.push({
-          type: 'del',
-          key: hashRoot,
-        })
+      if (remove) {
+        if (this._deleteFromDB) {
+          opStack.push({
+            type: 'del',
+            key: hashRoot,
+          })
+        }
       } else {
         opStack.push({
           type: 'put',
@@ -586,7 +594,8 @@ export class Trie {
   }
 
   /**
-   * The given hash of operations (key additions or deletions) are executed on the DB
+   * The given hash of operations (key additions or deletions) are executed on the trie
+   * (delete operations are only executed on DB with `deleteFromDB` set to `true`)
    * @example
    * const ops = [
    *    { type: 'del', key: Buffer.from('father') }
--- a/packages/trie/src/baseTrie.ts
+++ b/packages/trie/src/baseTrie.ts
@@ -569,10 +569,17 @@ export class Trie {
   ): Buffer | (EmbeddedNode | null)[] {
     const rlpNode = node.serialize()
 
-    // TODO: analyze why this length check is here, 2021-04-29
-    // If removed various tests fail. However this prevents certain delete operations
-    // when deleteFromDB is activated
-    // Related test: `index.spec.ts`, "setting back state root (deleteFromDB)"
+    /**
+     * TODO: analyze why this length check is here, 2021-04-29
+     *
+     * Hint from https://eth.wiki/fundamentals/patricia-tree:
+     * "When one node is referenced inside another node, what is included is H(rlp.encode(x)),
+     * where H(x) = sha3(x) if len(x) >= 32 else x and rlp.encode is the RLP encoding function."
+     *
+     * If removed various tests fail. However this prevents certain delete operations
+     * when deleteFromDB is activated.
+     * Related test: `index.spec.ts`, "setting back state root (deleteFromDB)"
+     */
     if (rlpNode.length >= 32 || topLevel) {
       // Do not use TrieNode.hash() here otherwise serialize()
       // is applied twice (performance)
--- a/packages/trie/src/baseTrie.ts
+++ b/packages/trie/src/baseTrie.ts
@@ -123,7 +123,8 @@ export class Trie {
   }
 
   /**
-   * Stores a given `value` at the given `key`.
+   * Stores a given `value` at the given `key` or do a delete if `value` is empty
+   * (delete operations are only executed on DB with `deleteFromDB` set to `true`)
    * @param key
    * @param value
    * @returns A Promise that resolves once value is stored.
@@ -149,7 +150,8 @@ export class Trie {
   }
 
   /**
-   * Deletes a value given a `key`.
+   * Deletes a value given a `key` from the trie
+   * (delete operations are only executed on DB with `deleteFromDB` set to `true`)
    * @param key
    * @returns A Promise that resolves once value is deleted.
    */
@@ -567,6 +569,10 @@ export class Trie {
   ): Buffer | (EmbeddedNode | null)[] {
     const rlpNode = node.serialize()
 
+    // TODO: analyze why this length check is here, 2021-04-29
+    // If removed various tests fail. However this prevents certain delete operations
+    // when deleteFromDB is activated
+    // Related test: `index.spec.ts`, "setting back state root (deleteFromDB)"
     if (rlpNode.length >= 32 || topLevel) {
       // Do not use TrieNode.hash() here otherwise serialize()
       // is applied twice (performance)
--- a/packages/trie/src/baseTrie.ts
+++ b/packages/trie/src/baseTrie.ts
@@ -569,17 +569,6 @@ export class Trie {
   ): Buffer | (EmbeddedNode | null)[] {
     const rlpNode = node.serialize()
 
-    /**
-     * TODO: analyze why this length check is here, 2021-04-29
-     *
-     * Hint from https://eth.wiki/fundamentals/patricia-tree:
-     * "When one node is referenced inside another node, what is included is H(rlp.encode(x)),
-     * where H(x) = sha3(x) if len(x) >= 32 else x and rlp.encode is the RLP encoding function."
-     *
-     * If removed various tests fail. However this prevents certain delete operations
-     * when deleteFromDB is activated.
-     * Related test: `index.spec.ts`, "setting back state root (deleteFromDB)"
-     */
     if (rlpNode.length >= 32 || topLevel) {
       // Do not use TrieNode.hash() here otherwise serialize()
       // is applied twice (performance)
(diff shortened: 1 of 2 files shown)


packages/trie/src/baseTrie.ts
```diff
@@ -566,11 +572,13 @@ export class Trie {
...
-          type: 'del',
-          key: hashRoot,
-        })
+      if (remove) {
+        if (this._deleteFromDB) {
+          opStack.push({
+            type: 'del',
+            key: hashRoot,
+          })
+        }
```
holgerd77, line 581: Note that this change is also changing the semantics of this conditional clause. Before if - when using a `CheckpointTrie` - being not within a checkpoint the `remove` calls would have been transitioned into a `put` operation. I am not seeing the logic of this and would assume this is falsy behaviour with no side effects (or maybe with side effects, not 100% sure), since the node references are removed from the trie anyhow (so it has no effect if the value is saved (again) in the DB. I wouldn't exclude the possibility though that I am overlooking something here.

ryanio: that is a good observation and indeed weird

ryanio:
this is looking good to me. should we add a test setting the stateRoot back and accessing a "deleted" value? then it will succeed with deleteFromDb=false and fail (return null?) with deleteFromDb=true

holgerd77:
@ryanio test addition was a great suggestion since this revealed that this fix is actually *not* fixing the problem. I added a test here d7d7c38f4a419da60c63ac2515bf152d4380b16c and this is - against assumption - also returning `v1` with `deleteFromDB=true` version when setting back the state root.

I tracked a bit with `console.log` and delete operations are actually *not* added in `_formatNode` since the `rlpNode.length >= 32` condition is not satisfied on delete calls and therefore no `delete` is added. My first conclusion here is that we then have this delete behavior really only on checkpointing when we do the `commit` calls.

First-thought alternative suggestion to fix would therefore be to add a `doDeletes=false` option to the `CheckpointTrie.commit()` method. But I also have to think this further through and let this sink in a bit. Suggestions, some alternative analysis or reasoning and/or additional remarks welcome.

holgerd77:
Ok, I would now open this up for review. 😄 

As discussed in the call we now keep (respectively: set) the no-delete option as `default` and generally do this as a general option for the `Trie` and no option on `CheckpointTrie.commit()` or something.

I've now found out why the tests failed, this is due to this `rlpNode.length >= 32` check in the `_formatNode()` function. I was not able to "fix" this directly since I don't understand where this is coming from or what it does, various tests fail though if this is removed. So I added as a TODO to further analyze. The worst thing which can happen here atm that there are some values less deleted when the `deleteFromDB` option is active. So I would regard this as an acceptable fix for now if we don't get to the root here on this, since overall the status quo is significantly improved by this PR respectively the HardHat bug and overall potentially dangerous delete behavior prevented.

Overall this PR has the following properties: I followed all the potential delete paths on the Trie methods, so various entry points which lead to either `del()` calls directly, pushing `del` operations an a DB stack, `put()` operations with an empty value and putting `put` operations with an empty value on a DB stack. All these paths end up on the `_formatNode` direct `del` DB stack addition, so all of this is prevented on the Trie with the PR. I also added some additional code docs to make this explicit. This also goes for the Trie subclasses CPTrie and SecureTrie, which all call into the super classes.

On the other hand the operations on the `DB` wrappers directly are *not* shielded - which I would argue should also be the case since the DB classes should *just*  do the DB operations if called, so people still have the chance to directly operate on the DB.

This makes me think respectively remember: we do the following direct DB operation in the `StateManager` within the VM:

```typescript
async putContractCode(address: Address, value: Buffer): Promise<void> {
    const codeHash = keccak256(value)

    if (codeHash.equals(KECCAK256_NULL)) {
      return
    }

    await this._trie.db.put(codeHash, value)

    account.codeHash = codeHash
    await this.putAccount(address, account)
  }
```

Is this still something dangerous in this context, e.g. when a numm value is passed? I have to say I am generally not understanding this code part and why the DB is accessed directly in this context and this is not done through the higher level Trie interface.

packages/trie/src/baseTrie.ts
```diff
@@ -39,24 +39,30 @@ export type FoundNodeFunction = (
...
+
   /**
    * test
    * @param db - A [levelup](https://github.com/Level/levelup) instance. By default (if the db is `null` or
    * left undefined) creates an in-memory [memdown](https://github.com/Level/memdown) instance.
    * @param root - A `Buffer` for the root of a previously stored trie
+   * @param deleteFromDB - Delete nodes from DB on delete operations (disallows switching to an older state root) (default: `false`)
    */
-  constructor(db?: LevelUp | null, root?: Buffer) {
+  constructor(db?: LevelUp | null, root?: Buffer, deleteFromDB: boolean = false) {
```
ryanio, line 58: I suggest we introduce a `TrieOptions` interface here like our other monorepo packages to remain consistent, just because I feel we could add more options in the future and would be easier to manage.

ryanio:
Regarding `putContractCode`, I think the reason the underlying db is used to `put` is because otherwise it would be routed through `SecureTrie.put()` which would keccak the `key` a second time, and `codeHash` is already keccak'd.

I am wondering if we change from `this._trie.db.put(codeHash, value)` to `this._trie.put(value, value)` if it would then properly keccak the key. But I am also wondering if it was bypassing the checkpointing or cache mechanisms on purpose? What if the state is reverted?

---

Regarding the `rlpNode.length >= 32` check in `_formatNode()`, I believe it's because it's used in `updateNode` when creating a new branch or extension node and if the length is >= 32 it needs to be further formatted as it can't be stored as is.

I found on [eth.wiki](https://eth.wiki/fundamentals/patricia-tree) it says:
> When one node is referenced inside another node, what is included is `H(rlp.encode(x))`, where `H(x) = sha3(x) if len(x) >= 32 else x` and `rlp.encode` is the RLP encoding function.

packages/trie/src/baseTrie.ts
holgerd77: Had planned this in for v6 and added a note #1024 since we already have these two optional parameters floating around (so `db` and `root`, or would you keep these separate for some reasons?). Or would you add this option nevertheless now and just do the additional inclusions on v6 (might be the more flexible solution)?

holgerd77:
@ryanio are there any code changes/updates you would recommend me to do derived from the last comment you made (`putContractCode` / greater-32-check)? 🤔  Or should we do this `putContractCode` check in a separate PR?

packages/trie/src/baseTrie.ts
ryanio: ah okay nice, I think that's fine we can do it for v6 and keep this as is 👍

ryanio:
@holgerd77 hm, thinking about it more, putContractCode must use the underlying db so it doesn't format the nodes via `baseTrie.put`. I started looking through the YP to see where it explains to store contract code and it says (bottom of page 3):

> The account state, σ[a], comprises the following four fields:
> ...
> **codeHash:** The hash of the EVM code of this account—this is the code that gets executed should this address receive a message call; it is immutable and thus, unlike all other fields, cannot be changed after construction.
> All such code fragments are contained in the state database under their corresponding hashes for later retrieval. 

The last line explains why we use the underlying db instead of formatting the trie, so I think the current code is correct.

The greater-than-32 check also seems correct to me and doesn't need any code changes.

ryanio:
Rebased the branch on master.

packages/trie/src/baseTrie.ts
```diff
@@ -561,16 +569,22 @@ export class Trie {
   ): Buffer | (EmbeddedNode | null)[] {
     const rlpNode = node.serialize()
 
+    // TODO: analyze why this length check is here, 2021-04-29
+    // If removed various tests fail. However this prevents certain delete operations
+    // when deleteFromDB is activated
+    // Related test: `index.spec.ts`, "setting back state root (deleteFromDB)"
```
ryanio, line 575: Copying this answer from my prior comment, on [eth.wiki](https://eth.wiki/fundamentals/patricia-tree) it says:
> When one node is referenced inside another node, what is included is `H(rlp.encode(x))`, where `H(x) = sha3(x) if len(x) >= 32 else x` and `rlp.encode` is the RLP encoding function.

alcuadrado:
As mentioned on Discord, this PR doesn't fix the bug, at least not in Balancer's codebase.

[This repo](https://github.com/alcuadrado/balancer-core-v2) can be used as a test. You need to check out the `state-manager-bug`.

To use the latest/released version of the state manager, just run `yarn` and `yarn test`. 58 tests will fail.

To try with a local build of the vm, run `yarn`, `yarn link` the vm, and `yarn test`. 58 tests are also failing.

holgerd77:
Ok, I've now pushed a last update on the TODO comment with the suggestions from Ryan. I guess with the latest analysis done by @alcuadrado and @ryanio we can now merge here and do a release? 😄 

Can someone please review and approve?

packages/trie/src/baseTrie.ts
```diff
@@ -561,16 +569,29 @@ export class Trie {
...
+     * TODO: analyze why this length check is here, 2021-04-29
+     *
+     * Hint from https://eth.wiki/fundamentals/patricia-tree:
+     * "When one node is referenced inside another node, what is included is H(rlp.encode(x)),
+     * where H(x) = sha3(x) if len(x) >= 32 else x and rlp.encode is the RLP encoding function."
+     *
+     * If removed various tests fail. However this prevents certain delete operations
+     * when deleteFromDB is activated.
+     * Related test: `index.spec.ts`, "setting back state root (deleteFromDB)"
+     */
```
gabrocheleau, line 582: Hey!
So, as mentioned in the docs you linked, I believe the length check is necessary because the `key` is only the `hashRoot` if the RLP encoded `value` is >= 32. 

And actually, in the case where `rlpNode.length < 32`, I don't think we even store a key-value pair, since the value is just referenced directly (`H(x)` is reversible if `rlpNode.length < 32`) so it's not necessary to store anything, which makes me think it would just not get added to the trie in the first place.

holgerd77:
@jochem-brouwer @ryanio Can this get a final review please? 😄

jochem-brouwer (dismissed):
Yup this looks great!

ryanio:
I would still like to remove this TODO before merging: https://github.com/ethereumjs/ethereumjs-monorepo/pull/1219#discussion_r629558920

holgerd77:
@ryanio could you do? I am somewhat on family holidays during the next days until Monday.

packages/trie/src/baseTrie.ts
ryanio, line 582:
```suggestion
```

packages/trie/test/index.spec.ts
```diff
@@ -273,3 +289,46 @@ tape('it should create the genesis state root from ethereum', function (tester)
     t.end()
   })
 })
+
+tape('setting back state root (deleteFromDB)', async (t) => {
+  const k1 = Buffer.from('1')
+  // Testing with longer value due to `rlpNode.length >= 32` check in `_formatNode()`
+  // see TODO note over there
```
ryanio, line 296:
```suggestion
  /* Testing with longer value due to `rlpNode.length >= 32` check in `_formatNode()`
   * Reasoning from https://eth.wiki/fundamentals/patricia-tree:
   * "When one node is referenced inside another node, what is included is `H(rlp.encode(x))`,
   * where `H(x) = sha3(x) if len(x) >= 32 else x`"
   */
```

ryanio (approved):
lgtm!

PR #1220, opened 2021-04-22
VM: Improve buildBlock test performance

ryanio:
This PR improves the test performance for `buildBlock.spec.ts` after seeing that it was running very slow in karma.

The bottleneck was `vm.stateManager.generateCanonicalGenesis()` which I was using to set up addresses with balances to use for txs. Instead I am now just adding balance to the specific accounts needed, which drastically reduced the execution time.

### DIFF
--- a/packages/vm/tests/api/utils.ts
+++ b/packages/vm/tests/api/utils.ts
@@ -1,4 +1,4 @@
-import { Account, BN } from 'ethereumjs-util'
+import { Account, Address, BN } from 'ethereumjs-util'
 import Blockchain from '@ethereumjs/blockchain'
 import VM from '../../lib/index'
 import { VMOpts } from '../../lib'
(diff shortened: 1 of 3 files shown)


packages/vm/tests/api/utils.ts
```diff
@@ -12,6 +12,13 @@ export function createAccount(nonce: BN = new BN(0), balance: BN = new BN(0xfff3
   return new Account(nonce, balance)
 }
 
+export async function setBalance(vm: VM, address: Address, balance = new BN(100000000)) {
+  const account = createAccount(new BN(0), balance)
+  await vm.stateManager.checkpoint()
+  await vm.stateManager.putAccount(address, account)
+  await vm.stateManager.commit()
```
holgerd77, line 19: Out of interest: what happens without this outer `checkpoint()` and `commit()` wrapping? Or is this just to play safe (does this add anything here on this?)?

holgerd77 (approved):
Thanks for having a look here, yes, this makes a lot of sense, `generateCanonicalGenesis()` is pretty heavy, especially on `mainnet`.

PR #1221, opened 2021-04-23
Common/Client: add Aleut network

jochem-brouwer:
This PR adds the [Aleut test network](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/client-integration-testnets/aleut.md). 

Fork hashes / state root have been manually integrated, also explicitly checked that the reported genesis hash is correct.

Cannot connect to Aleut for some reason, though.


jochem-brouwer:
Discovered a concurrency problem in client, if this is fixed then we connect to Aleut, so this spec is ready to get merged.

holgerd77 (approved):
LGTM

PR #1222, opened 2021-04-23
Client: bootstrap() extraction

holgerd77:
@ryanio could you please have a look into this and see how this can be done better? 😄


packages/client/lib/client.ts
```diff
@@ -106,8 +106,16 @@ export default class EthereumClient extends events.EventEmitter {
...
-    await Promise.all(this.services.map((s) => s.start()))
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    Promise.all(this.services.map((s) => s.start()))
+      .then(() => {
+        return Promise.all(this.config.servers.map((s) => s.start()))
+      })
+      .then(() => {
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        Promise.all(this.config.servers.map((s) => s.bootstrap()))
+      })
```
ryanio, line 117: would this achieve the same?

```suggestion
    await Promise.all(this.services.map((s) => s.start()))
    await Promise.all(this.config.servers.map((s) => s.start()))
    await Promise.all(this.config.servers.map((s) => s.bootstrap()))
```

ryanio: i believe libp2p runs its own bootstrap, so if the `bootstrap()` method is only available for `RlpxServer`, we could add a simple guard to the last line, something like: `if('bootstrap' in s) { s.bootstrap() }`

holgerd77: No, this is just not working, this was my naive (is it?) approach as well on first try. You can test this I guess now also on `master` with `Aleut`. If you use this clean `await` chaining version of yours and do a `DEBUG=devp2p:* npm run client:start -- --network=aleut` run you should see no `Send GET_BLOCK_HEADERS` response from our side. If you run with this bad hacky code from my test fix you should see a response together with some follow-up processing (going wrong with "could not decode message BlockHeaders"):

image

My Promise-feeling unfortunately is so bad that it seems that I don't get this right. Or maybe this just needs some stronger sequentiality with some artificial delays with a timer or so due to the networking stuff? Not sure if this makes sense but we can also try.

Anyhow: I am a bit stuck how to fix this.

ryanio: thanks for the info, ok I'll keep looking into it and report back, I am curious now too :)

ryanio: I found the problem, we needed to await the `Devp2pRLPx` initialization, fixed in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1222/commits/0297fd5093d21a0831ddab9d90e2471c4e2e9b42

holgerd77:
@ryanio thanks for having a look here! 😄  👍 

Just pulled the branch and tested with `DEBUG=devp2p:* npm run client:start -- --network=aleut` and there is still no answer from our side on the `GET_BLOCK_HEADERS` request, are you sure this fixes the problem?

Apart from this test: I am generally wondering what this `Devp2pRLPx` initialization does or changes (just for my understanding) since I am not seeing any internal async calls? Or is this due to the code in the event handlers?? (does this make sense? This should be independent I guess?)

ryanio:
> @ryanio thanks for having a look here! 😄 👍
> 
> Just pulled the branch and tested with `DEBUG=devp2p:* npm run client:start -- --network=aleut` and there is still no answer from our side on the `GET_BLOCK_HEADERS` request, are you sure this fixes the problem?

Hmmm, I am seeing it on my side with the current branch, does [this run](https://user-images.githubusercontent.com/22116/116129722-485c5480-a67f-11eb-8d3a-a17ba21f856d.png) look correct?

> Apart from this test: I am generally wondering what this `Devp2pRLPx` initialization does or changes (just for my understanding) since I am not seeing any internal async calls? Or is this due to the code in the event handlers?? (does this make sense? This should be independent I guess?)

All I did was wrap `RlpxServer.initRlpx()` with a Promise that resolves when `Devp2pRLPx` emits [`listening`](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1222/files#diff-19e192bdb67554a044d649c4ae2139939b007937c2472fa6f21a97cf8ca8f784R295). Otherwise it would  continue without being ready for connections.

holgerd77:
@ryanio I'll do a new `bootstrap` from my local root folder and then report back here, maybe there is something wrong on my side.

holgerd77 (commented):
This (a `bootstrap` run) indeed has solved the problem, impressive! 😄 

Will do another admin-merge here, again: reviewed by @ryanio by continuing the work respectively I've now reviewed the continued work from Ryan.

PR #1229, opened 2021-04-26
VM: fix subfolder coverage

### LINKED ISSUE #1225: VM: Subfolders like state or evm missing from Coverage report
All subfolders below `lib` are [missing](https://codecov.io/gh/ethereumjs/ethereumjs-monorepo/tree/master/packages/vm/lib/) from the Codecov coverage report.

This is something between unfortunate and serious, since having a decent coverage from e.g. `state` or `evm` is very much crucial for a functioning VM. It should be examined why this is not displayed respectively taken into account properly.

ryanio:
Fixes #1225


holgerd77 (approved):
Ah, cool, thanks for the fix 😄 , looks good, just checked the new folders on Codecov! 👍

Issue #1227, opened 2021-04-26
Client: RPC Tests Coverage Improvements


holgerd77:
The coverage of the client RPC tests - especially for the `eth` and `web3` calls - are pretty low with `eth` < 70% and `web3` < 80%, see [Codecov report](https://codecov.io/gh/ethereumjs/ethereumjs-monorepo/tree/master/packages/client/lib/rpc/modules).

This should be significantly improved. 😋

holgerd77:
@Ghorbanian hi there, thanks for watching out for open issues! 😄 

I had a closer look here, I guess this is not so much needed any more since it improved significantly by 1-2 PRs already been done on this and at least some remaining cases might be to hard to do and not worth the effort. Didn't have such a close look though and there will very likely still room for useful improvements, so if you stumble upon code coverage gaps which you think should be filled feel free to give it a go. I will nevertheless close this issue here for now.

Another test case which would be neat: we introduced a new `debug` option for the VM and the VM is doing a lot of console debug output if an environment variable `DEBUG` is set. I just realized that these switches are completely skipped by the tests, see e.g. [runBlock.ts](https://codecov.io/gh/ethereumjs/ethereumjs-monorepo/src/master/packages/vm/src/runBlock.ts) coverage. It would be neat to have at least one generic test here which calls the VM on a higher level (likely best: `VM.runBlock()` and tricking the VM to think `DEBUG` is set, so that all (or a lot of) the debug messages actually get triggered. This would substantially improve the test coverage, but would also be useful to be tested by itself, since there is always the chance that someone adds some code to a debug message generation which breaks the debug flow.

One bonus thing here would be to deactivate `console.log` just for this test, so that the test output doesn't get bloated with all these debug messages.

If you want you can give this one a try, also see [this docs section](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/vm#understanding-the-vm) for some context on VM debug. Otherwise I can also just open a dedicated issue on this, then someone else can just pick this up.

Ghorbanian:
> @Ghorbanian hi there, thanks for watching out for open issues! smile
> 
[...]

PR #1235, opened 2021-04-30
Added client rpc eth tests for unsupported block params

emersonmacro:
Partially addresses #1227 for `eth` calls. In particular, added tests for unsupported `block` params (i.e. block !== `latest`). Hopefully will help out with the code coverage report.

### DIFF
--- a/packages/client/test/rpc/eth/call.spec.ts
+++ b/packages/client/test/rpc/eth/call.spec.ts
@@ -93,10 +93,7 @@ tape(`${method}: call with unsupported block argument`, async (t) => {
     gasLimit: `0x${new BN(53000).toString(16)}`,
   }
 
-  const req = params(
-    method,
-    [{ ...estimateTxData, gas: estimateTxData.gasLimit }, 'pending']
-  )
+  const req = params(method, [{ ...estimateTxData, gas: estimateTxData.gasLimit }, 'pending'])
   const expectRes = (res: any) => {
     const msg = 'should return error if block argument is not "latest"'
     if (res.body.result.message === 'Currently only "latest" block supported') {
--- a/packages/client/test/rpc/eth/call.spec.ts
+++ b/packages/client/test/rpc/eth/call.spec.ts
@@ -74,3 +74,36 @@ tape(`${method}: call with valid arguments`, async (t) => {
   }
   baseRequest(t, server, req, 200, expectRes)
 })
+
+tape(`${method}: call with unsupported block argument`, async (t) => {
+  const blockchain = await Blockchain.create()
+
+  const client = createClient({ blockchain, includeVM: true })
+  const manager = createManager(client)
+  const server = startRPC(manager.getMethods())
+
+  // genesis address with balance
+  const address = Address.fromString('0xccfd725760a68823ff1e062f4cc97e1360e8d997')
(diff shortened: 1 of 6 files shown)


ryanio:
Hey @emersonmacro thanks for your PR, looks great.

I was planning on adding support for more block params soon. Basically if a block is passed in, we need to copy the VM, set the stateRoot back to the stateRoot of the specified block in the StateManager, then run the query (e.g. getBalance would use getAccount.balance).

I was planning on working on this soon, possibly next week, but if you want to give it a shot before then, feel free.

Otherwise I think we can still merge this PR in because we will still not have support for `pending` (we are just starting on our TxPool implementation) so I can update these tests for `pending` instead of `!== 'latest'` in a future PR.

emersonmacro:
Hi @ryanio thanks for the feedback, and sounds good. I'm still new to the codebase so I've been easing into it by working on some tests. I'd be in favor of merging this PR if that's ok, and I may take a stab at the change you described - it might take me a while lol, so you'll probably get to it before I get it working. But the steps you laid out are super helpful. Cheers!

ryanio (approved):
lgtm!

ryanio (approved):

emersonmacro:
Hi @ryanio thanks for merging!

I started looking into the code change you described, for passing in an actual block number. If you have a minute, I'd be interested in your feedback on this: https://github.com/emersonmacro/ethereumjs-monorepo/commit/fea486a295f504b4779695e20cd42b40b8b1d128.

I'm guessing there's a better approach than what I have there, but I was wondering if I'm on the right track.

ryanio:
@emersonmacro looks like a great start. i would probably start with a utility function that accepts the string block param and returns a block, that can be used in the other methods. first it can check if it's "earliest" (gets the genesis block) or "latest" (gets the latest block). if it "pending" then it can return the unsupported error. next if it is a hex block number it can return that. then as a final catch it can return an error that it didn't understand the input.

some great tests to add would be to verify that it can correctly get values at prior block numbers.

emersonmacro:
@ryanio Makes sense, thanks. I started working on an approach based on your description, but this week is really busy, so no worries if you want to move forward on it. I will try to keep working on it though.

ryanio:
@emersonmacro ok thanks for the update! my plate is filling up too so we can see who can get to it first 😄 do you want to open a branch with what you have so far so I can better follow along?

emersonmacro:
@ryanio For sure - I pushed up an update to the draft branch on my fork. I don't think what I have right now is very good, and typescript is complaining about a couple things... Still a work in progress :). https://github.com/ethereumjs/ethereumjs-monorepo/compare/master...emersonmacro:draft-support-block-param

Also, did you mean opening a branch on the main ethereumjs repo? Happy to do so but I didn't think I'd have write access to open one, but I might be wrong on that. I'm still kind of new to working with forks.

ryanio:
great, thanks, just a branch on your own fork is totally fine for now.

ryanio:
looks good, I would add "pending" as a case and return the unsupported error directly. then if the string doesn't start with "0x" you can return an error that says something like input cannot be parsed as a 0x-prefixed hex number or should be "earliest"|"latest".

PR #1236, opened 2021-04-30
Client: Aleut initial connection reliability test fix

holgerd77:
Ok, I was at least able to track down the connection reliability problems on Aleut, so the non-response to the `GET_BLOCK_HEADERS` message from the Nethermind client. This is *not* due to the early bootstrap but happens after due to the time interval between the peer handshake triggered in protocol binding `BoundProtocol` and the final setup and addition of the peer to the peerpool.

I tested refactoring out the handshake and add later in `PeerPool`. This *does* reliably solve the Aleut connection problem, so connection starts immediately with this. However the solution has too many side effects, errors are now not shielded properly any more and generally mainnet is not working any more with this solution. Lol. It is also not desirable likely to add peers to the pool *before* the handshake completes, since this would cause a lot of add-ban bouncing.

Anyhow. At least one step closer.

We might also want to pursue another approach, like buffering ETH requests if peer connection is still build up (maybe we add another flag to the `Peer` class) and then do the response once ready.

### DIFF
--- a/packages/client/lib/net/peer/libp2ppeer.ts
+++ b/packages/client/lib/net/peer/libp2ppeer.ts
@@ -29,7 +29,7 @@ export interface Libp2pPeerOptions extends Omit<PeerOptions, 'address' | 'transp
  *
  * new Libp2pPeer({ id, multiaddrs, protocols })
  *   .on('error', (err) => console.log('Error: ', err))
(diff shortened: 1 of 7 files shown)


holgerd77:
Closed in favor of #1237

PR #1237, opened 2021-04-30
Client: Simple message queue to handle unhandled ETH messages during handshake

holgerd77:
Follow-up solution to initial ETH messages being lost, replaces #1236

### DIFF
--- a/packages/client/lib/net/peer/peer.ts
+++ b/packages/client/lib/net/peer/peer.ts
@@ -46,6 +46,13 @@ export class Peer extends events.EventEmitter {
   protected transport: string
   protected protocols: Protocol[]
   private _idle: boolean
+
+  /*
+    If the peer is in the PeerPool.
+    If true, messages are handled immediately.
+    If false, adds incoming messages to handleMessageQueue,
+    which are handled after the peer is added to the pool.
+  */
   public pooled: boolean = false
 
   // Dynamically bound protocol properties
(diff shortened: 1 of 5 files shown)


packages/client/lib/net/peer/peer.ts
```diff
@@ -46,6 +46,7 @@ export class Peer extends events.EventEmitter {
   protected transport: string
   protected protocols: Protocol[]
   private _idle: boolean
+  public pooled: boolean = false
```
ryanio, line 49: PR looks really good and it worked for me on a test run with aluet. I will push a commit to fix the failing tests.

I would like some more clarity on the naming of this `pooled` variable, is there a better name we could use that's more self explanatory? Maybe like "ready" or "handshakeComplete"?

holgerd77: Hmm, I like the `pooled` as in "is in the pool", which is actually what this variable describes (it is set to `pooled` when added to the pool and set to `pooled=false` when removed from the pool. This `handshakeComplete` is just one characteristics which comes along, but this variable could also be used for user use cases (not sure if there are any though).

Thanks for dealing with the tests! 😄

ryanio: ok sounds good :) I will add a small typedoc to explain that then

holgerd77:
Looks good! @ryanio can you review? 😄

ryanio (approved):

PR #1240, opened 2021-05-02
Implement EIP3541

jochem-brouwer:
Implements [EIP3541](https://github.com/ethereum/EIPs/blob/8aa8b36070c28808f32d1e6465e71e9d0f2b8d2f/EIPS/eip-3541.md). 

~~Verify that if we deploy 0xEF on 3541 activated commons we should consume all gas (this is most likely the case).~~ (confirmed on ACD)

### DIFF
--- a/packages/vm/lib/evm/evm.ts
+++ b/packages/vm/lib/evm/evm.ts
@@ -92,11 +92,11 @@ export function COOGResult(gasUsedCreateCode: BN): ExecResult {
   }
 }
 
-export function EIP3541ViolationResult(gasLimit: BN): ExecResult {
+export function INVALID_BYTECODE_RESULT(gasLimit: BN): ExecResult {
   return {
     returnValue: Buffer.alloc(0),
     gasUsed: gasLimit,
-    exceptionError: new VmError(ERROR.EIP3541_VIOLATION),
+    exceptionError: new VmError(ERROR.INVALID_BYTECODE_RESULT),
   }
 }
 
@@ -328,7 +328,7 @@ export default class EVM {
         this._vm._common.isActivatedEIP(3541) &&
         result.returnValue.slice(0, 1).equals(Buffer.from('EF', 'hex'))
       ) {
-        result = { ...result, ...OOGResult(message.gasLimit) }
+        result = { ...result, ...INVALID_BYTECODE_RESULT(message.gasLimit) }
       } else {
         result.gasUsed = totalGas
       }
--- a/packages/vm/lib/evm/evm.ts
+++ b/packages/vm/lib/evm/evm.ts
@@ -92,6 +92,14 @@ export function COOGResult(gasUsedCreateCode: BN): ExecResult {
   }
 }
 
+export function EIP3541ViolationResult(gasLimit: BN): ExecResult {
+  return {
+    returnValue: Buffer.alloc(0),
+    gasUsed: gasLimit,
+    exceptionError: new VmError(ERROR.EIP3541_VIOLATION),
+  }
+}
+
 export function VmErrorResult(error: VmError, gasUsed: BN): ExecResult {
   return {
     returnValue: Buffer.alloc(0),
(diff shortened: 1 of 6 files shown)


jochem-brouwer:
Ok also added extensive tests here, so I am pretty sure that we are correct here as well 😄

packages/vm/lib/evm/evm.ts
```diff
@@ -92,6 +92,14 @@ export function COOGResult(gasUsedCreateCode: BN): ExecResult {
   }
 }
 
+export function EIP3541ViolationResult(gasLimit: BN): ExecResult {
```
holgerd77, line 95: Maybe I am missing something, but where is this function actually called? 😜

packages/vm/lib/exceptions.ts
```diff
@@ -16,6 +16,7 @@ export enum ERROR {
   INVALID_BEGINSUB = 'invalid BEGINSUB',
   INVALID_RETURNSUB = 'invalid RETURNSUB',
   INVALID_JUMPSUB = 'invalid JUMPSUB',
+  EIP3541_VIOLATION = 'eip3541 violation: cannot put contracts starting with 0xEF bytecode',
```
holgerd77, line 19: Maybe we can use something more semantically expressive here - a bit analog to the last discussions we had coming to the conclusion to rather avoid EIP numbers for names? 🤔

Suggestion (not sure yet though):

- `RESTRICTED_STARTING_BYTECODE`

Or similar.

holgerd77: (same a bit for the function name from above)

packages/vm/tests/api/EIPs/eip-3541.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    created = result.createdAddress
+
+    code = await vm.stateManager.getContractCode(created!)
+
+    st.ok(code.length > 0, 'did deposit code')
+
+    st.end()
+  })
+
+  t.test('should not deploy contracts starting with 0xEF using CREATE', async (st) => {
```
holgerd77, line 65: These header are not matching the test cases applied with both the success and failure cases, maybe just drop the "should not" here (and other occurrences where it applies)

holgerd77 (commented):
Hi Jochem,
had a first look here, thanks a lot for taking on these two additional EIPs as well, you are really our hero atm! 😄  Have some comments, generally the whole PR looks great though with these extensive tests and everything.

packages/vm/lib/exceptions.ts
axic: In some place seen `INVALID_CODE` or `INVALID_BYTECODE`, and in EVMC we also had `CONTRACT_VALIDATION_FAILED` from earlier, that may be also a good one in the future.

holgerd77: Ah yeah, that's better, my suggestion is likely still to specific, I like `INVALID_BYTECODE`. 😄

holgerd77: But `CONTRACT_VALIDATION_FAILED` might be a bit better targeted, so eventually more suited.

axic: Probably `INVALID_CODE`/`INVALID_BYTECODE` seems to be the best for now, and if any "real validation" is rolled out later (such as EIP-3540), then the `CONTRACT_VALIDATION_FAILED` error could be introduced in addition.

jochem-brouwer:
Updated and rebased. Ready for re-review @holgerd77 @ryanio

packages/vm/lib/evm/evm.ts
jochem-brouwer: Heh. Oops.

packages/vm/tests/api/EIPs/eip-3541.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+      gasLimit: 1000000,
+      nonce: 1,
+    }).sign(pkey)
+
+    result = await vm.runTx({ tx: tx1 })
+    created = result.createdAddress
+
+    code = await vm.stateManager.getContractCode(created!)
+
+    st.ok(code.length > 0, 'did deposit code')
```
holgerd77, line 44: Nice test case! 😄

packages/vm/tests/api/EIPs/eip-3541.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    result = await vm.runTx({ tx: tx2 })
+    created = result.createdAddress
+
+    code = await vm.stateManager.getContractCode(created!)
+
+    st.ok(code.length > 0, 'did deposit code')
+
+    st.end()
+  })
```
holgerd77, line 63: Important counter case (with all the last 1559 experiences 😛 ).

holgerd77 (approved):
Yeah, really nice, looks good to me. 🙂

axic:
Could you run ethereumjs no this PR by any chance to confirm everything is in order: https://github.com/ethereum/tests/pull/835 ?

holgerd77:
@axic @jochem-brouwer Ok, I am through (see comments on linked PR), two test runs which might need some further investigation:

`undefinedOpcodeFirstByte`: all test runs with the EIP deactivated are also passing
`Opcodes_TransactionInit`: all test cases except one with the EIP deactivated are also passing

This might either hint to the test cases being non-testing things but alternatively also to an error in our implementation (we might apply the EIP-3541 condition to a pre-EIP-3541 state).

Likely need to stop for now and pass this on.

PR #1245, opened 2021-05-09
Implement EIP3554

jochem-brouwer:
This [EIP](https://eips.ethereum.org/EIPS/eip-3554) is scheduled for London (please note that the Eth1 spec is outdated, it points to an EIP where the writer is currently not available, so a new EIP was created).

This also cleanups some code in block where the delay logic was copied (>= Byzantium). It is now a parameter in Common.

### DIFF
--- a/packages/common/src/hardforks/muirGlacier.json
+++ b/packages/common/src/hardforks/muirGlacier.json
@@ -3,5 +3,13 @@
   "comment": "HF to delay the difficulty bomb",
   "url": "https://eips.ethereum.org/EIPS/eip-2384",
   "status": "Final",
-  "eips": []
+  "gasConfig": {},
+  "gasPrices": {},
+  "vm": {},
+  "pow": {
+    "difficultyBombDelay": {
+      "v": 9000000,
+      "d": "the amount of blocks to delay the difficulty bomb with"
+    }
+  }
 }
--- a/packages/common/tests/eips.spec.ts
+++ b/packages/common/tests/eips.spec.ts
@@ -58,4 +58,16 @@ tape('[Common]: Initialization / Chain params', function (t: tape.Test) {
 
     st.end()
   })
+
+  t.test('returns the right block delay for EIP3554', function (st) {
+    for (const fork of ['muirGlacier', 'berlin']) {
+      const c = new Common({ chain: 'mainnet', hardfork: fork })
+      let delay = c.param('pow', 'difficultyBombDelay')
+      st.equal(delay, 9000000)
+      c.setEIPs([3554])
+      delay = c.param('pow', 'difficultyBombDelay')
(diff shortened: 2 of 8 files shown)


packages/common/src/hardforks/muirGlacier.json
```diff
@@ -3,5 +3,13 @@
   "comment": "HF to delay the difficulty bomb",
   "url": "https://eips.ethereum.org/EIPS/eip-2384",
   "status": "Final",
-  "eips": []
```
jochem-brouwer, line 6: Had to remove this field, otherwise the difficulty bomb delay as reported by `param()` in Common yields the Constantinople one (probably a bug?).

holgerd77: Hmm, interesting.

holgerd77 (approved):
LGTM, cool, all EIPs together!!! 🎉

Congrats. 😄

packages/common/tests/eips.spec.ts
```diff
@@ -58,4 +58,16 @@ tape('[Common]: Initialization / Chain params', function (t: tape.Test) {
...
+    for (const fork of ['muirGlacier', 'berlin']) {
+      const c = new Common({ chain: 'mainnet', hardfork: fork })
+      let delay = c.param('pow', 'difficultyBombDelay')
+      st.equal(delay, 9000000)
+      c.setEIPs([3554])
+      delay = c.param('pow', 'difficultyBombDelay')
+      st.equal(delay, 9500000)
+    }
+    st.end()
+  })
```
holgerd77, line 72: Ah, this test is actually somewhat misplaced - just noticing on rebasing - this test file is just for testing generic EIP switch logic, and not the EIPs themselves (actually all about this library is about "EIPs" 😛 ).

So this would rather go to `params.spec.ts` since this is testing a parameter switch. No drama, but we should move on occasion to not get more blurry here.

ryanio: added this as an item to Tiny Fixes Meta Issue (#905)

holgerd77: Ah, I had already updated this along my test work at some point (will remove again from the list). 😋

holgerd77: (actually: I did a lot of explicit code changes, updates, "code decisions" on my work on the 1559 PR, so it would be very helpful if you could also take some time short-term and go especially through the commits that I have done on top of Jochem's work (I have already reviewed the most part of this): https://github.com/ethereumjs/ethereumjs-monorepo/pull/1148/commits

This would help us a lot to keep our release schedule and reviewing this huge 1559 PR is like doing 10-20 other "normally-sized" PRs we do on other occasions. 😛  )

PR #1247, opened 2021-05-09
EIP-3541 ethereum/tests Run

holgerd77:
This is a test run of a our EIP-3541 implementation from #1240 towards the tests implemented at https://github.com/ethereum/tests/pull/835.

DO NOT MERGE!!!

@axic I noted that our test setup is very much not prepared to adhoc integrate these invented fork names, I couldn't imminently get this work to run on CI.

I will instead take this PR as a base and do manual tests runs of the different files.


holgerd77:
`ts-node ./tests/tester --state --customStateTest=../ethereum-tests/GeneralStateTests/stCreate2/CREATE2_EOF1.json --fork=Berlin+3541`

image

Counter-test by setting `const eips: number[] = [3541]` to `const eips: number[] = []` in `GeneralStateTestsRunner`:

pass: 0
fail: 4

holgerd77:
`ts-node ./tests/tester --state --customStateTest=../ethereum-tests/GeneralStateTests/stBadOpcode/undefinedOpcodeFirstByte.json --fork=Berlin+3541`

```shell
1..114
# tests 114
# pass  114
```

Counter test:

```shell
1..114
# tests 114
# pass  114
```

Actually also pass, this doesn't seem to be right I guess? 🤔

holgerd77:
`ts-node ./tests/tester --state --customStateTest=../ethereum-tests/GeneralStateTests/stCreate2/CREATE2_EOF1_loop.json --fork=Berlin+3541`

```shell
1..1
# tests 1
# pass  1
```

Counter test:

```shell
# tests 1
# pass  0
# fail  1
```

Everything ok here.

holgerd77:
`ts-node ./tests/tester --state --customStateTest=../ethereum-tests/GeneralStateTests/stCreateTest/CREATE_EOF1.json --fork=Berlin+3541`

```shell
1..4
# tests 4
# pass  4
```

Counter test:

```shell
# tests 4
# pass  0
# fail  4
```

Everything ok here.

holgerd77:
`ts-node ./tests/tester --state --customStateTest=../ethereum-tests/GeneralStateTests/stCreateTest/CREATE_EOF1_loop.json --fork=Berlin+3541`

```shell
1..1
# tests 1
# pass  1
```

Counter test:

```shell
# tests 1
# pass  0
# fail  1
```

Everything ok here.

holgerd77:
`ts-node ./tests/tester --state --customStateTest=../ethereum-tests/GeneralStateTests/stTransactionTest/CreateTransactionEOF1.json --fork=Berlin+3541`

```shell
# tests 4
# pass  4
```

Counter test:

```shell
# tests 4
# pass  0
# fail  4
```

Everything ok here.

holgerd77:
`ts-node ./tests/tester --state --customStateTest=../ethereum-tests/GeneralStateTests/stTransactionTest/Opcodes_TransactionInit.json --fork=Berlin+3541`

```shell
1..130
# tests 130
# pass  130
```

```shell
ok 128 the state roots should match
ok 129 the state roots should match
not ok 130 the state roots should match

1..130
# tests 130
# pass  129
# fail  1
```

On counter test run only the last failing? 🤔  Is this correct?

axic:
@gumb0 can you comment when you get a chance please?

holgerd77:
For `undefinedOpcodeFirstByte` actually the `post` sections for `Berlin` and `Berlin+3541` seem to be indentical.

holgerd77:
Ah, ok, have found the following comment from @gumb0 on the original PR:

"This could be merged separately already, because EIP-3541 doesn't change this behavior."

😄

holgerd77:
Have found this comment for `Opcodes_TransactionInit` on the tests PR:

"I didn't care to make all possible invalid opcodes here.

And this also can be merged independently, because this behavior is not changing yet."

This actually doesn't explain yet why 1 test is failing for us on `berlin`. If the test case is new for `berlin` then this might be a bug in our implementation. If the test case is old (we passed before) this might be a newly introduced bug in the tests.

(there is also always the third possibility that there is something wrong in our test setup 😄 )

holgerd77:
It's the `data=129` test failing (so with `data` set to `0x6000600060006000600073b94f5374fce5edbc8e2a8697c15331677e6ebf0b61c350f15060ef60005360016000f3`( which can be run in isolation with `ts-node ./tests/tester --state --customStateTest=../ethereum-tests/GeneralStateTests/stTransactionTest/Opcodes_TransactionInit.json --fork=Berlin+3541 --data=129`.

jochem-brouwer:
Found the problem, we are deploying a contract with 0xEF as code.

packages/vm/tests/GeneralStateTestsRunner.ts
```diff
@@ -66,7 +66,7 @@ async function runTestCase(options: any, testData: any, t: tape.Test) {
   }
 
   const state = new Trie()
-  const hardfork = options.forkConfigVM
+  const hardfork = 'berlin'
```
jochem-brouwer, line 69: ?? Now we run all in Berlin and thus the --fork param is ignored?

jochem-brouwer:
Ran this;

```
ts-node ./tests/tester --state --customStateTest=../ethereum-tests/GeneralStateTests/stCreate2/CREATE2_EOF1.json --fork=Berlin+3541
ts-node ./tests/tester --state --customStateTest=../ethereum-tests/GeneralStateTests/stBadOpcode/undefinedOpcodeFirstByte.json --fork=Berlin+3541
ts-node ./tests/tester --state --customStateTest=../ethereum-tests/GeneralStateTests/stCreate2/CREATE2_EOF1_loop.json --fork=Berlin+3541
ts-node ./tests/tester --state --customStateTest=../ethereum-tests/GeneralStateTests/stCreateTest/CREATE_EOF1.json --fork=Berlin+3541
ts-node ./tests/tester --state --customStateTest=../ethereum-tests/GeneralStateTests/stCreateTest/CREATE_EOF1_loop.json --fork=Berlin+3541
ts-node ./tests/tester --state --customStateTest=../ethereum-tests/GeneralStateTests/stTransactionTest/CreateTransactionEOF1.json --fork=Berlin+3541
ts-node ./tests/tester --state --customStateTest=../ethereum-tests/GeneralStateTests/stTransactionTest/Opcodes_TransactionInit.json --fork=Berlin+3541

```

Result: all pass.

Counter test, all fail except

`ethereum-tests/GeneralStateTests/stBadOpcode/undefinedOpcodeFirstByte.json`: all pass
`ethereum-tests/GeneralStateTests/stTransactionTest/Opcodes_TransactionInit.json`: all pass except the final test (deploy 0xEF contract and store value in some other contract)

CC @axic @gumb0

packages/vm/tests/GeneralStateTestsRunner.ts
jochem-brouwer: Ah ok this PR is supposed to be hacky :)

jochem-brouwer:
Ah ok I think I got a bit confused by the process here.

@holgerd77 Yes that one test fails is correct, I reproduced this as well; in your final test a 0xEF contract is deployed. For Berlin this is fine (it is allowed) but for Berlin+3541 this is not OK as it is an invalid contract. So it is correct that this is the only test which fails on Berlin+3541.

holgerd77:
@jochem-brouwer I have to say I can't completely "read" this explanation. First - these are not "my tests", these are the tests from @gumb0 submitted to the `ethereum/tests` repo.

Also: if there is a test case failing, there definitely *something behaves* wrong. Either in:

a) Our code
b) our test setup
c) test test itself

What are you suggesting and with what argumentation? 😄 

(again: I'm just not able to interpret your current answer).

jochem-brouwer:
I just passed all tests on `Berlin+EIP3541`. This runs on a Common with Berlin as hardfork and 3541 activated as EIP. 

So this is the test in question right?

`ts-node ./tests/tester --state --customStateTest=../ethereumtests/GeneralStateTests/stTransactionTest/Opcodes_TransactionInit.json --fork=Berlin+3541 --data=129`

Except now in the counter test we run this with no EIPs (so EIP 3541 is not activated). We still run it in the context of Berlin+3541. Then this test indeed fails. This is because:

- This test creates  a contract
- In the deploy code of creating this contract, it calls to another contract.
- This contract stores `1` in storage slot `0`.
- Now we return to the deploy code
- This `MSTORE8`s `0xEF` at memory slot 0
- We return memory [0:1] (so we return 0xEF)
- This thus deploys a contract with code 0xEF

If we activate 3541, this creation fails, and thus the storage slot write gets reverted.
If we don't activate it, then the creation does not fail, and thus the storage slot gets written. Therefore, there is a difference between the test where we activated the EIP and where we did not; since we are running the test with the same expected results, at least one has thus fail (and in this case the one where we don't activate the EIP thus fails, which is correct per the spec).

holgerd77:
@jochem-brouwer ah, yes, you're right, sorry, I was confused as well. So I guess the tests created are all ok. 😄

gumb0:
Confirming that `tests/GeneralStateTests/stBadOpcode/undefinedOpcodeFirstByte.json` should pass for both EIP3541 activated and not. (I'll remove it from https://github.com/ethereum/tests/pull/835, the improved version is in https://github.com/ethereum/tests/pull/836, you could check that, too)

In `Opcodes_TransactionInit.json` one new test is also supposed to behave without change after fork:
https://github.com/ethereum/tests/pull/835/files#diff-767d260c1cb56c8ea5548501176577874ca06d50339aa65fc236dfc37e3cb20eR373
and another one changes behavior with EIP3541:
https://github.com/ethereum/tests/pull/835/files#diff-767d260c1cb56c8ea5548501176577874ca06d50339aa65fc236dfc37e3cb20eR374
I think you've already figured this out correctly.

holgerd77:
@gumb0 cool, thanks for the feedback here! 👍

holgerd77:
Results have been picked up on the ethereum/tests side, will close.

PR #1249, opened 2021-05-10
Add baikal

jochem-brouwer:
Adds baikal to common. Don't merge this, it changes the London spec, which is not what we want currently (should probably have setup another hardfork with name baikal/aleut or something).

The genesis block is correct; the hash equals the expected hash. (Tried to sync but client does not seem to connect to the bootnode).


holgerd77 (approved):
Discussed with @jochem-brouwer - London change is not a problem any more - will merge.

PR #1253, opened 2021-05-13
blockchain: fix _putBlockOrHeader and add a test

samlior:
According to [eip-225](https://eips.ethereum.org/EIPS/eip-225), the difficulty of the inturn block should be 2, and the difficulty of the noturn block should be 1.  And the client should choose the fork with the larger total difficulty. So when `consensusType` is `poa`, total difficulty also needs to be judged.

---

One test will fail(reorg.spec.ts, `should correctly reorg a poa chain and remove blocks from clique snapshots`), but I think this is a problem with the test itself, because I didn't see any logic about timestamp in the blockchain. The previous test passed only because the `consensusType` is `poa`, and under the old logic , `poa` will cause the later `putBlock` block to always `_rebuildCanonical`. If you swap the order of `putBlocks` as below, this test will always fail.

```ts
// blockchain/test/reorg.spec.ts line163
await blockchain.putBlocks([block1_high, block2_high])
const head_high = await blockchain.getHead()

await blockchain.putBlocks([block1_low, block2_low])
const head_low = await blockchain.getHead()
```


ryanio:
hey @samlior, thanks for the PR and reasoning, looks good.

good point about the reorg tests. that's my bad and I should have paid more attention there. how about adding a `block3_high` to make it a higher total difficulty chain?

samlior:
What does `block3_high` mean? A new field of the block? @ryanio

ryanio:
A third block for the high chain so it has a higher total difficulty, e.g.:

```typescript
const low_chain = [block1_low, block2_low]
const high_chain = [block1_high, block2_high, block3_high]
```

samlior:
In theory, add `block3_high` can make the test correct. But in fact the test still fails. This is because there is a bigger problem. When a reorg occurs, for clique, in the old logic, it is just deleted the clique snapshots with height greater than or equal to `blockNumber + 1` in `_deleteCanonicalChainReferences`. But the correct logic should be (I think) to find the common ancient header, delete the clique snapshots with height greater than or equal to the `ancientHeader.blockNumber + 1`. And then, to all blocks between (ancientHeader.blockNumber, newHeader.blockNumber] Re-run clique logic to create clique snapshots.

ryanio:
sorry I forgot to get back to you on this! I will try to review soon, this is a really important one.

samlior:
If you can review this PR quickly, I would be very grateful. Because my project depends on this project, and it is about to release a new version. Now I can only fix this problem in a very "ugly" way. @ryanio

ryanio:
yes I will review today! apologies for the delay

ryanio (approved):
thanks @samlior, this looks great, the new src code is well written and concise and the updated tests look good. thanks for your contributions here!

ryanio:
I'm not sure when we'll have a new blockchain release but if it is important for your project we could prioritize a small bug fix release for you. cc @holgerd77

samlior:
In fact, this is very important for my project, I would be very very very grateful if you could prioritize a small bug fix release. @holgerd77

holgerd77:
@holgerd77 yes, we can do, we have one other PR to settle (#1315) and then I can do a release round including the blockchain library. Think this should be doable during the week.

holgerd77:
Hi @samlior, ok, this is now out on npm 😄, sorry, it took a couple of days longer since we were a bit unsure on how to combine or not combine the release with some other work. Hope it still suffices for your project, thanks a lot for the contribution!

PR #1254, opened 2021-05-13
block: Check and convert difficulty param to hex string

### LINKED ISSUE #726: from-rpc does not convert difficulty right if difficulty is a number
Geth's web3.eth.getBlock returns difficulty as a number. One has to manually convert this to hex before feeding it to from-rpc. 
```
rom_RPC without difficulty to hex conversion 0x32333032393139373438313334363734 actual difficulty: 2302919748134674 number 9643516
from_RPC with difficulty to hex conversion 0x82e7e4988a712 actual difficulty: 0x82e7e4988a712 number 9643516
```

If you feed these to blockchain you get an Invalid difficulty error.

Ghorbanian:
Applies a fix for the `difficulty` parameter of the `block` package as discussed in issue #726.

### DIFF
--- a/packages/trie/src/baseTrie.ts
+++ b/packages/trie/src/baseTrie.ts
@@ -39,24 +39,30 @@ export type FoundNodeFunction = (
  * The API for the base and the secure interface are about the same.
  */
 export class Trie {
-  /** The backend DB */
-  db: DB
   /** The root for an empty trie */
   EMPTY_TRIE_ROOT: Buffer
-  private _root: Buffer
   protected lock: Semaphore
+
+  /** The backend DB */
+  db: DB
+  private _root: Buffer
+  private _deleteFromDB: boolean
 
   /**
    * test
    * @param db - A [levelup](https://github.com/Level/levelup) instance. By default (if the db is `null` or
    * left undefined) creates an in-memory [memdown](https://github.com/Level/memdown) instance.
    * @param root - A `Buffer` for the root of a previously stored trie
+   * @param deleteFromDB - Delete nodes from DB on delete operations (disallows switching to an older state root) (default: `false`)
    */
-  constructor(db?: LevelUp | null, root?: Buffer) {
+  constructor(db?: LevelUp | null, root?: Buffer, deleteFromDB: boolean = false) {
     this.EMPTY_TRIE_ROOT = KECCAK256_RLP
     this.lock = new Semaphore(1)
+
     this.db = db ? new DB(db) : new DB()
     this._root = this.EMPTY_TRIE_ROOT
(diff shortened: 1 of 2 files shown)


ryanio:
hey, thanks for the PR!

could you add some code to `from-rpc.spec.ts` to cover both scenarios?

i am thinking it would be great to also add a test case with an exact output from geth's `web3.eth.getBlock(x)` like jochem did to ensure all other values output with the expected values.

feel free to add fixing for Transaction's `value` and `gasPrice` as identified in https://github.com/ethereumjs/ethereumjs-monorepo/issues/726#issuecomment-695329772 into this PR as well.

I am wondering since you used optional chaining in the ternary operator if it will execute the falsy expression when difficulty is undefined:

```typescript
difficulty?.slice(0, 2) === '0x' ? difficulty : parseInt(difficulty, 10).toString(16),
```

ts-node:
```typescript
> let difficulty
undefined
> difficulty?.slice(0, 2) === '0x' ? difficulty : parseInt(difficulty, 10).toString(16)
'NaN'
```
 
adding tests for undefined would be good too, learning from past experience 😅 

you would also want to output `'0x' + parseInt(difficulty, 10).toString(16)`

thinking about how to accomplish this PR then, maybe you can move `difficulty` from `const` to `let` then you can use the ternary without the optional operator:

```typescript
const { .. } = blockParams
let { difficulty } = blockParams

if (difficulty) {
  difficulty = difficulty.slice(0, 2) === '0x' ? difficulty : '0x' + parseInt(difficulty, 10).toString(16)
}
```

that's not *so* elegant, so feel free to suggest another way

ok last thought, could define a small helper function like `toHex`:

```typescript
/*
 * Converts a string containing either a hex string or integer value into a hex string.
 */
function toHex(value?: string) {
  if (!value) return undefined
  if (value.slice(0, 2) === '0x') return value
  return '0x' + parseInt(value, 10).toString(16)
}
```

then just:

```typescript
difficulty: toHex(difficulty)
```

ryanio:
hey, nice adding a return error message.

hm did you run `npm run build` or `npm run build:util` on root after adding the code to util? lerna needs to run on root to build and link the packages correctly so the dependent packages can pick them up.

Ghorbanian:
> hey, nice adding a return error message.
> 
> hm did you run `npm run build` or `npm run build:util` on root after adding the code to util? lerna needs to run on root to build and link the packages correctly so the dependent packages can pick them up.

I did try running `npm run build` from root after running `npm run clean` in order to re-link all package dependencies, but I run into the same `TS2305` error. I have recreated a minimal viable example at PR #1258 (I accidentally closed this one) in case you are interested in giving it a closer look.

PR #1255, opened 2021-05-14
tx: fix getMessageToSign return type for typed txs

### LINKED ISSUE #1251: Tx: Potential EIP-2930 getMessageToSign() Typing Bug
Something is not correct with the overload typing of [getMessageToSign()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/eip2930Transaction.ts#L283) on EIP-2930 compatible txs.

When looking at the implementation this method should return a `Buffer` for all the `hashMessage=false` cases for EIP-2930 txs.

However TypeScript is assuming this to be `Buffer[]`, this can e.g. be tested by adding the following to the respective `eip2930.spec.ts` test case:

```typescript
//@ts-ignore
console.log(unsignedTx.getMessageToSign(false).toString('hex'))
```

So this needs a `@ts-ignore` to work (when run this then outputs the correct hex string) since TypeScript is assuming this to be a `Buffer[]`.

ryanio:
Fixes #1251


holgerd77 (approved):
LGTM

PR #1256, opened 2021-05-14
Fix prettier lint

ryanio:
This PR runs lint:fix with prettier [2.3.0](https://prettier.io/blog/2021/05/09/2.3.0.html#highlights)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,3483 @@
 				"jsonstream2": "^1.1.2",
 				"make-dir": "^1.3.0",
 				"minimist": "^1.2.0",
+<<<<<<< HEAD
+=======
+				"ms": "^0.7.1",
+				"multiaddr": "^8.1.2",
+				"node-dir": "^0.1.16",
+				"nyc": "^14.0.0",
+				"peer-id": "^0.14.3",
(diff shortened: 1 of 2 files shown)


holgerd77:
Hi @ryanio, thanks for working this out with the analysis why the linting went wrong at some point, cool! 😄 

I am very much afraid that this will get a big hazzle to rebase this huge #1148 EIP-1559 PR on top of this though. I remember that I had such a situation before and it gets super laborious to manually fuddle all the changes together with the  correct linting on a rebase, and this will likely break many times before done so this is likely not worth the effort.

I will put this on "blocked" for now, let's please either wait until the 1559-PR is merged - this should just be days away - and then do the linting on top and for these few days ignore the broken linting on the other open PRs. Or we could alternatively also open this PR towards the 1559 branch.

ryanio:
for sure, thanks, no problem. this would be easy to add as an additional commit to an existing PR if that would be better, just run `npm i && npm run lint:fix` at root.

holgerd77:
I now had integrated the linting fixes into the 1559 PR, so not sure if this (or parts from here) still needs to be merged.

PR #1258, opened 2021-05-17
[WIP] convert difficulty param to hex string

Ghorbanian:
Minimum viable example of approach to implement `toHex` helper function to solve issue #726 . Attempting to build and run test throws a `TS2305` error. The compiler does not recognize the newly created helper function that is to be committed as a part of this PR.


packages/block/src/header-from-rpc.ts
```diff
@@ -1,6 +1,6 @@
 import { BlockHeader } from './header'
 import { BlockOptions } from './types'
-
+import {toHex} from 'ethereumjs-util/src/helpers'
```
ryanio, line 3: ah so you only need to import it from `'ethereumjs-util'`, not the path.

also, please follow proper [import order](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md) so `import { toHex} from 'ethereumjs-util'` would be at the top of the file with no extra lines

packages/util/src/helpers.ts
```diff
@@ -1,5 +1,21 @@
 import { isHexString } from 'ethjs-util'
 
+/**
+ * Checks if a number string is a hex string, if not, converts and returns it
```
ryanio, line 4: suggestion, this feels more clear

```suggestion
 * Returns a 0x-prefixed hex number string from a hex string or string integer.
```

packages/util/src/helpers.ts
```diff
@@ -1,5 +1,21 @@
 import { isHexString } from 'ethjs-util'
 
+/**
+ * Checks if a number string is a hex string, if not, converts and returns it 
+ * @param {string} input string to check, convert, and return
```
ryanio, line 5: typedoc can infer the type
```suggestion
 * @param input string to check, convert, and return
```

packages/util/src/helpers.ts
```diff
@@ -1,5 +1,21 @@
...
+ */
+ export const toHex = function (input?: string) {
```
ryanio, line 7: trying to think of a less ambiguous name, what do you think of `numberToHex`?

```suggestion
 export const numberToHex = function (input?: string) {
```

packages/util/src/helpers.ts
```diff
@@ -1,5 +1,21 @@
...
+  if (!input) return undefined
+  if (!isHexString(input)) {
+    try {
+      return '0x' + parseInt(input, 10).toString(16)
+    } catch {
+      const msg = `This method only supports number or 0x-prefixed hex strings but input was: ${input}`
+      throw new Error(msg)
+    }
+  }
```
ryanio, line 16: this is just personal preference for less nesting

```suggestion
  if (isHexString(input)) return input
  try {
    return '0x' + parseInt(input, 10).toString(16)
  } catch {
    const msg = `This method only supports number or 0x-prefixed hex strings but input was: ${input}`
    throw new Error(msg)
  }
```

ryanio:
thanks for the PR! looks good, i left some notes i hope they are helpful

packages/block/src/header-from-rpc.ts
ryanio: ah i found the issue, so `util/src/helpers.ts` is not included in the exports as defined in `util/src/index.ts`.

i'm trying to think which file would be best for this function, but there doesn't seem to be a perfect match. maybe just at the bottom of `types.ts`?

Ghorbanian: I ended up importing the helper functions in `util/src/index.ts` to stay consistent.

ryanio: hm, I'm not sure if we want to do that, as the other methods in that file aren't particularly helpful as exported methods for users of the library. i'd rather find a better place for the function, or even create a new file like `conversion.ts` that I could see us adding more methods to over time.

packages/block/test/from-rpc.spec.ts
```diff
@@ -25,4 +28,31 @@ tape('[fromRPC]: block #2924874', function (t) {
     st.ok(block.validateUnclesHash())
     st.end()
   })
+
+  t.test(
+    'should create a block with json data that includes a difficulty parameter of type hex string',
+    function (st) {
+      const blockDifficultyAsHex = blockFromRpc(blockDataDifficultyAsHex)
+      st.ok(blockDifficultyAsHex)
```
ryanio, line 36: I think in these three tests you added you should check that the parameter is actually the result you expected (e.g. `st.ok(blockDifficultyAsHex.header.difficulty.eqn(1))`

packages/tx/test/base.spec.ts
```diff
@@ -300,4 +300,55 @@ tape('[BaseTransaction]', function (t) {
...
+      gasLimit: '',
+      gasPrice: '0x01',
+      to: '',
+      value: '',
+      data: '',
+      v: '',
+      r: '',
+      s: '',
+    })
+    st.ok(tx)
```
ryanio, line 316: same here, you should test something specific, not just that it returned

packages/tx/src/baseTransaction.ts
```diff
@@ -53,9 +54,22 @@ export abstract class BaseTransaction<TransactionObject> {
     const sB = toBuffer(s === '' ? '0x' : s)
 
     this.nonce = new BN(toBuffer(nonce === '' ? '0x' : nonce))
+
+    if (typeof(gasPrice) === "string") {
```
ryanio, line 58: @Ghorbanian what is your reasoning for adding this logic to the base class in Transaction, but for Block (difficulty) you only added it to `header-from-rpc.ts`? i guess we should be consistent. integer strings would only be incoming from rpc, so i feel like we could also handle this there.

thinking about this in a new way, i almost feel like we should just handle base 10 integer strings in `toBuffer` then all this work would automatically be covered everywhere. @holgerd77 would you have any hesitations against this suggestion?

holgerd77: Yes, I think this makes a lot of sense to integrate into `toBuffer` directly, this PR here is getting far too extensive for what is targeted to be achieved. 

Since we have `PrefixedHexString` now more strict in util, it should also be possible to cleanly differentiate between the two version of strings (so "0x10" vs "10").

holgerd77: (but maybe we should also sleep over it one night and give it some thought if this might have side effects we are atm not anticipating)

Ghorbanian: I second that adding the logic to handle integer strings in `toBuffer` would make more sense, but we would be effectively expanding the inputs supported by the `toBuffer` function:
```
Inputs supported: `Buffer`, `String` (hex-prefixed), `Number`, null/undefined, `BN` and other objects
```

The above becomes:

```
Inputs supported: `Buffer`, `String` (hex-prefixed and integer), `Number`, null/undefined, `BN` and other objects
```

The change would involve something like this:

```
if (typeof v === 'string') {
    if (!isHexString(v)) {
      throw new Error(
        `Cannot convert string to buffer. toBuffer only supports 0x-prefixed hex strings and this string was given: ${v}`
      )
    }
    return Buffer.from(padToEven(stripHexPrefix(v)), 'hex')
  }
```

The above becomes:

```
if (typeof v === 'string') {
    if (!isHexString(v)) {
      try {
        return '0x' + parseInt(input, 10).toString(16)
      } catch {
        const msg = `Cannot convert string to buffer. toBuffer only supports integer or 0x-prefixed hex strings but the given string was: ${v}`
        throw new Error(msg)
      }
    }
    return Buffer.from(padToEven(stripHexPrefix(v)), 'hex')
  }
```

So the question is, is there a case where supporting integer string inputs to the `toBuffer` array may cause unexpected issues? I am having difficulty thinking of any simple cases.

ryanio: thanks for digging into this, i will ask the team for feedback on this idea

ryanio:
hey @Ghorbanian, i'm sorry to lead you astray but I learned that `toBuffer` used to accept strings like `"123"` a few majors ago but was changed because they were interpreted as utf8 and would end up as the byte arrays representing that.

These things used to be true:

* `toBuffer("A").equals(Buffer.from([41])`
* `toBuffer("1").equals(Buffer.from([31])`

So I think it was overreaching to work on `toBuffer`, sorry again about that.

If you'd like to refocus scope back on just solving this at the `from-rpc` layer for the few specific fields that sometimes receive string integer that would be best to solve the issue at hand. I think you actually already had that code before, just not the tests to support it which you now do now and looks great.

Ghorbanian:
I made the changes in the `block` and `util` packages and placed relevant tests in `from-rpc.spec.ts`. Let me know what you think @ryanio.

ryanio:
great, thanks, I will try to review today! if you can rebase on master then the test suite should run (otherwise it's waiting for an update to align with the master branch before running the ci)

packages/block/src/from-rpc.ts
```diff
@@ -10,6 +10,10 @@ function normalizeTxParams(_txParams: any) {
   txParams.gasLimit = txParams.gasLimit === undefined ? txParams.gas : txParams.gasLimit
   txParams.data = txParams.data === undefined ? txParams.input : txParams.data
 
+  // check and convert gasPrice and value params
+  txParams.gasPrice = numberToHex(txParams.gasPrice)
+  txParams.value = numberToHex(txParams.value)
```
ryanio, line 15: great!

packages/block/test/from-rpc.spec.ts
```diff
@@ -25,4 +26,49 @@ tape('[fromRPC]: block #2924874', function (t) {
...
+  t.test(
+    'Should create a block with json data that includes a transaction with defaults with gasPrice parameter as integer string',
+    function (st) {
+      const gasPriceAsIntegerString = '1'
+      const blockDataTransactionGasPriceAsInteger = blockData
+      blockDataTransactionGasPriceAsInteger.transactions[0].gasPrice = gasPriceAsIntegerString
+      const blockFromTransactionGasPriceAsInteger = blockFromRpc(
+        blockDataTransactionGasPriceAsInteger
+      )
+      st.ok(
```
ryanio, line 54: you could use st.equal()

packages/util/src/index.ts
```diff
@@ -44,6 +44,11 @@ export * from './externals'
  */
 export * from './types'
 
+/**
+ * Helper functions
+ */
+export * from './helpers'
```
ryanio, line 50: i'm still not sure if we want to add this file to the exported methods, although it would be a nice catch-all for future helper methods. @holgerd77 any thoughts?

ryanio:
looks like ci is failing with:

```
test/from-rpc.spec.ts(55,63): error TS2339: Property 'gasPrice' does not exist on type 'TypedTransaction'.
Property 'gasPrice' does not exist on type 'FeeMarketEIP1559Transaction'.
```

so since you know the transaction is a legacy one, you could cast it as e.g. `(blockFromTransactionGasPriceAsInteger.transactions[0] as Transaction).gasPrice.toString()`

come to think of it, a good follow-up PR might be to add 1559 support to `from-rpc` if you are up for it, otherwise it could be something I could do next week after we merge this PR :)

Ghorbanian:
Any idea why I would be getting the following error every time I run tests locally from root with `npm run test`:

```
# should colorize key=value pairs
not ok 384 key=value pairs should be colorized
  ---
    operator: equal
    expected: 'test \x1b[32mkey\x1b[39m=value '
    actual:   'test key=value '
    at: Test.<anonymous> (/home/indigoomega021/code/projects/ethereumjs/ethereumjs-monorepo/packages/client/test/logging.spec.ts:31:8)
    stack: |-
      Error: key=value pairs should be colorized
          at Test.assert [as _assert] (/home/indigoomega021/code/projects/ethereumjs/ethereumjs-monorepo/node_modules/tape/lib/test.js:228:54)
          at Test.bound [as _assert] (/home/indigoomega021/code/projects/ethereumjs/ethereumjs-monorepo/node_modules/tape/lib/test.js:80:32)
          at Test.equal (/home/indigoomega021/code/projects/ethereumjs/ethereumjs-monorepo/node_modules/tape/lib/test.js:389:10)
          at Test.bound (/home/indigoomega021/code/projects/ethereumjs/ethereumjs-monorepo/node_modules/tape/lib/test.js:80:32)
          at Test.<anonymous> (/home/indigoomega021/code/projects/ethereumjs/ethereumjs-monorepo/packages/client/test/logging.spec.ts:31:8)
          at Test.bound [as _cb] (/home/indigoomega021/code/projects/ethereumjs/ethereumjs-monorepo/node_modules/tape/lib/test.js:80:32)
          at Test.exports.Test.run (/home/indigoomega021/code/projects/ethereumjs/ethereumjs-monorepo/node_modules/tape-catch/index.js:26:10)
          at Test.bound [as run] (/home/indigoomega021/code/projects/ethereumjs/ethereumjs-monorepo/node_modules/tape/lib/test.js:80:32)
          at Test._end (/home/indigoomega021/code/projects/ethereumjs/ethereumjs-monorepo/node_modules/tape/lib/test.js:165:11)
          at Test.exports.Test._end (/home/indigoomega021/code/projects/ethereumjs/ethereumjs-monorepo/node_modules/tape-catch/index.js:16:8)
  ...
```

I did some digging and I think it only fails locally. I just force push with `--no-verify` and CI/CD tests don't catch it at any point.

ryanio:
hm no need to worry about that, maybe you are using a terminal without color support

packages/util/src/index.ts
holgerd77: Also given this some thought.

I think we should definitely not just open up a whole new helper category here on the sideline, we've Util relatively well structured now and to introduce some "fall-back" category for things we are not sure doesn't seem like a very good idea to me.

Generally with this implementation from above we run into this problem we've discussed in the chat 1-2 weeks ago (@Ghorbanian sorry, this was in our internal chat, so you just couldn't follow-up on this) that with allowing both hex-prefixed strings (being seen as hex-byte-representations) and non-hex-strings (being assumed to be numbers) there is no chance to differentiate if people *want* non-hex strings to be integers or (this actually happens a lot in practice) just have forgotten the `0x` hex prefix but the input is actually a hex-byte-representation string.

This is really dangerous. With the regex implementation from above one is already catching a lot of stuff but just not all, and this would "break" respectively get ambiguous on strings like "1102" or the like.

I think we should *at least* not allow these kind of things officially in the util library when we just did the iteration of getting a lot stricter on type checking along with the v7 releases https://github.com/ethereumjs/ethereumjs-util/releases/tag/v7.0.0 for safety reasons.

Two things I can imagine here.

1. I think the functionality would fit a lot more naturally in the Util method suite as a `intToBuffer` method as a counter part to `bufferToInt` (in the `bytes` package) we already have. But this method should then really only allow integer input of various forms (`integer` itself, `integer` string, something else?). If we want to make an additional exception for the RPC use case we should add this as custom code to the `from-rpc` methods in `Block` itself (`if (hasSomeHexPrefix) use toBuffer(), else use intToBuffer)`). This might be an exception worth to make in this use case since RPC responses should be relatively well defined I guess. On the Util integration side I am still not 100% sure if this aligns well with e.g. the more general `toBuffer` method, we likely don't want to end up with 7 different flavors of these Buffer conversion methods.
2. We generally hang this lower and just add this to an internal helper function in a new `util.ts` file in the `Block` library which we should then exclude from documentation in the `typedoc.js` file.

Atm I have a slight preference for 2., but not totally sure.

WDYGT? 🤔

Ghorbanian: Option 2 sounds like it would come with much less ambiguity. The function can keep the name `integerToHex`. We can think about making the helper function externally available if the need arises in the future. Any thoughts, @ryanio?

PR #1259, opened 2021-05-19
ci: add redundancy check for temporaryClientFix

ryanio:
This PR adds a check to skip the bcrypto node-gyp rebuild if the `build` directory already exists.


holgerd77 (approved):
LGTM

PR #1261, opened 2021-05-20
tests: additional coverage

ryanio:
This PR adds coverage for missed lines of code, mostly gathered from codecov check warnings from #1148.

### DIFF
--- a/packages/client/test/rpc/eth/sendRawTransaction.spec.ts
+++ b/packages/client/test/rpc/eth/sendRawTransaction.spec.ts
@@ -1,5 +1,8 @@
 import tape from 'tape'
-import { baseSetup, params, baseRequest } from '../helpers'
+import { FeeMarketEIP1559Transaction } from '@ethereumjs/tx'
+import Common from '@ethereumjs/common'
+import { toBuffer } from 'ethereumjs-util'
+import { baseSetup, params, baseRequest, createClient, createManager, startRPC } from '../helpers'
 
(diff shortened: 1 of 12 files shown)


holgerd77:
Ah, when seeing this `cliqueBeneficiary` test: this is one of the last open questions at least I have if it would make sense to generalize this option and not limit to PoA, see the following comment: https://github.com/ethereumjs/ethereumjs-monorepo/pull/1148/files#r628869203

Would be a pity if we miss this opportunity and introduce half-thought-through new options which are hard to revert or change later on.

@ryanio maybe you have also got an opinion on this, or @jochem-brouwer might also want to express some final thoughts here since there were still some open questions before.

packages/client/test/rpc/eth/sendRawTransaction.spec.ts
```diff
@@ -41,3 +44,50 @@ tape(`${method}: call with invalid tx (wrong chain ID)`, (t) => {
...
+      t.pass(msg)
+    } else {
+      throw new Error(msg)
+    }
+  }
+  baseRequest(t, server, req, 200, expectRes)
+})
+
+tape(`${method}: call with no peers`, (t) => {
+  const client = createClient({ noPeers: true })
```
holgerd77, line 76: Ah, I actually didn't know how to test this. 😄

holgerd77 (approved):
Nice, looks good, really nice test additions! 😀

PR #1266, opened 2021-05-22
feat: extend eth_getBlockByNumber JSON-RPC method

### LINKED ISSUE #1175: On RPC method block parameters, allow block identifiers as "latest", "earliest" or "pending"
https://github.com/ethereumjs/ethereumjs-monorepo/blob/53150c159b559b31b6b7c321d784dbb11b082fb9/packages/client/lib/rpc/modules/eth.ts#L259-L264

This helps on compatibility with other tooling. Getting the latest block is a common type of request.

gabrocheleau:
This PR resolves #1175, implementing the `"latest"` and `"earliest"` cases of the `eth_getBlockByNumber` JSON-RPC method. The "pending" case is not been implemented but an error message is now provided.

The PR also changes every method of the `validators` object in `packages\client\lib\rpc\validation.ts` to getter methods. The reason for this is to bind `this` to that the validator methods can be cross-referenced within the object. Previously, every instance of `this.[validationMethod]` used within the `validators` object would fail and return `{"jsonrpc":"2.0","id":1,"error":{"code":-32603,"message":"Internal error"}}`

There are other ways to fix that `validators` issue, see [this article that talks about other options](https://ponyfoo.com/articles/binding-methods-to-class-instance-objects ).


packages/client/lib/rpc/modules/eth.ts
```diff
@@ -423,7 +437,7 @@ export class Eth {
     if (latest < blockNumberBN) {
       return {
         code: INVALID_PARAMS,
-        message: `specified block greater than current height`,
+        message: `Specified block greater than current height`,
```
ryanio, line 440: why capitalize? the other error messages seem to start lowercase

ryanio:
thanks for the PR, looks great!

hm i don't love the complexity of the binds, this doesn't seem so elegant to me:
```typescript
this.getTransactionCount = middleware(this.getTransactionCount.bind(this), 2, [
  [validators.address.bind],
  [validators.blockOption],
])
```

maybe we can try to simplify the setup or usage. i wouldn't consider it a blocker, but something we can improve on.

packages/client/lib/rpc/modules/eth.ts
gabrocheleau: Ah actually I thought the others were capitalized, e.g. all the: "message: `Currently only "latest" block supported`", it seems like it's not consistent though so yeah no need to capitalize.

gabrocheleau:
> thanks for the PR, looks great!
> 
> hm i don't love the complexity of the binds, this doesn't seem so elegant to me:
> 
> ```ts
> this.getTransactionCount = middleware(this.getTransactionCount.bind(this), 2, [
>   [validators.address.bind],
>   [validators.blockOption],
> ])
> ```
> 
> maybe we can try to simplify the setup or usage. i wouldn't consider it a blocker, but something we can improve on.

`the validators.address.bind` was actually unnecessary so removed that. I'm not sure what the middleware binding is actually necessary for, would need to investigate that.

packages/client/lib/rpc/modules/eth.ts
ryanio: oops yeah we were looking at different ones, for new messages it's okay but in general it wouldn't be good practice to change existing messages since it would disrupt anyone possibly doing exact string matches (not that they should, but better safe than sorry)

ryanio:
>the validators.address.bind was actually unnecessary so removed that

oh great, that's what i was more confused about, i thought it was because you were using internal references as you explained in opening post.

> I'm not sure what the middleware binding is actually necessary for, would need to investigate that.

it's so the function code can use the variables bound to `this` on its class.

packages/client/lib/rpc/modules/eth.ts
```diff
@@ -273,15 +274,28 @@ export class Eth {
...
+      return {
+        code: INVALID_PARAMS,
+        message: `"pending" is not yet supported`,
+      }
+    } else if (blockOpt === 'earliest') {
+      block = await this._chain.getBlock(new BN(0))
+    } else {
+      const blockNumberBN = new BN(stripHexPrefix(blockOpt), 16)
+      block = await this._chain.getBlock(blockNumberBN)
+    }
```
evertonfraga, line 297:
```suggestion
    } else if (blockOpt.substr(0, 2) === '0x') {
      const blockNumberBN = new BN(stripHexPrefix(blockOpt), 16)
      block = await this._chain.getBlock(blockNumberBN)
    } else {
      return {
        message: `invalid argument ${index}: argument must be a hex string`,
        code: INVALID_PARAMS,
      }
    }
```

evertonfraga (dismissed):
Covering one more corner case. There's a related test failing, I believe you just need to double check on the error message being returned.

packages/client/lib/rpc/modules/eth.ts
gabrocheleau: Thanks!

Shouldn't that case be always covered by the [`blockOption` validation here](https://github.com/gabrocheleau/ethereumjs-monorepo/blob/feat/extend-getBlockByNumber/packages/client/lib/rpc/validation.ts#L126-L154)

gabrocheleau:
This is ready to review! :)

holgerd77 (approved):
This looks great, nice test cases with sufficient coverage, thanks a lot, will merge. 😄

PR #1272, opened 2021-05-29
1559: add more constraints to tx validation

ryanio:
See https://github.com/ethereum/EIPs/pull/3594

### DIFF
--- a/packages/tx/src/eip1559Transaction.ts
+++ b/packages/tx/src/eip1559Transaction.ts
@@ -158,10 +158,19 @@ export default class FeeMarketEIP1559Transaction extends BaseTransaction<FeeMark
       toBuffer(maxPriorityFeePerGas === '' ? '0x' : maxPriorityFeePerGas)
     )
 
-    this._validateCannotExceedMaxInteger({
-      maxFeePerGas: this.maxFeePerGas,
-      maxPriorityFeePerGas: this.maxPriorityFeePerGas,
-    })
+    this._validateCannotExceedMaxInteger(
+      {
+        maxFeePerGas: this.maxFeePerGas,
+        maxPriorityFeePerGas: this.maxPriorityFeePerGas,
+      },
+      256
+    )
+
+    if (this.maxFeePerGas.lt(this.maxPriorityFeePerGas)) {
+      throw new Error(
+        'maxFeePerGas cannot be less than maxPriorityFeePerGas (The total must be the larger of the two)'
+      )
+    }
 
     if (!this.chainId.eq(this.common.chainIdBN())) {
       throw new Error('The chain ID does not match the chain ID of Common')
@@ -196,12 +205,9 @@ export default class FeeMarketEIP1559Transaction extends BaseTransaction<FeeMark
    * The up front amount that an account must have for this transaction to be valid
    * @param baseFee The base fee of the block (will be set to 0 if not provided)
    */
-  getUpfrontCost(baseFee?: BN): BN {
-    if (!baseFee) {
-      baseFee = new BN(0)
-    }
-    const inclusionFeePerGas = BN.min(this.maxPriorityFeePerGas, this.maxFeePerGas.sub(baseFee!))
-    const gasPrice = inclusionFeePerGas.add(baseFee!)
+  getUpfrontCost(baseFee: BN = new BN(0)): BN {
+    const inclusionFeePerGas = BN.min(this.maxPriorityFeePerGas, this.maxFeePerGas.sub(baseFee))
+    const gasPrice = inclusionFeePerGas.add(baseFee)
     return this.gasLimit.mul(gasPrice).add(this.value)
   }
 
--- a/packages/vm/lib/runTx.ts
+++ b/packages/vm/lib/runTx.ts
@@ -243,6 +243,17 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
         `sender doesn't have enough funds to send tx. The upfront cost is: ${cost} and the sender's account only has: ${balance}`
       )
     }
+    if (tx.transactionType === 2) {
+      // EIP-1559 spec:
+      // The signer must be able to afford the transaction
+      // `assert balance >= gas_limit * max_fee_per_gas`
+      const cost = tx.gasLimit.mul((<FeeMarketEIP1559Transaction>tx).maxFeePerGas)
+      if (balance.lt(cost)) {
+        throw new Error(
+          `sender doesn't have enough funds to send tx. The max cost is: ${cost} and the sender's account only has: ${balance}`
+        )
+      }
+    }
   } else if (!opts.skipNonce) {
     if (!nonce.eq(tx.nonce)) {
       throw new Error(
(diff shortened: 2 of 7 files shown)


packages/tx/src/eip1559Transaction.ts
```diff
@@ -196,12 +205,9 @@ export default class FeeMarketEIP1559Transaction extends BaseTransaction<FeeMark
    * The up front amount that an account must have for this transaction to be valid
    * @param baseFee The base fee of the block (will be set to 0 if not provided)
    */
-  getUpfrontCost(baseFee?: BN): BN {
-    if (!baseFee) {
-      baseFee = new BN(0)
-    }
-    const inclusionFeePerGas = BN.min(this.maxPriorityFeePerGas, this.maxFeePerGas.sub(baseFee!))
-    const gasPrice = inclusionFeePerGas.add(baseFee!)
+  getUpfrontCost(baseFee: BN = new BN(0)): BN {
```
ryanio, line 208: there are no changes here, just inlining the default baseFee in the method argument (https://github.com/ethereumjs/ethereumjs-monorepo/pull/1148#discussion_r637437606)

packages/vm/tests/api/runTx.spec.ts
```diff
@@ -180,8 +180,9 @@ tape('runTx() -> API parameter usage/data errors', (t) => {
 
     try {
       await vm.runTx({ tx })
+      // TODO uncomment:
+      // t.fail('should throw error')
```
jochem-brouwer, line 184: TODO

packages/vm/tests/api/runTx.spec.ts
```diff
@@ -228,10 +230,32 @@ tape('runTx() -> API parameter usage/data errors', (t) => {
       const tx = getTransaction(vm._common, txType.type, true)
       try {
         await vm.runTx({ tx })
+        // TODO uncomment:
+        // t.fail('should throw error')
```
jochem-brouwer, line 234: TODO

packages/vm/tests/api/runTx.spec.ts
holgerd77: Was also irritated here at first but now had a closer look and ran locally. On second look it seems that this was just not working before and Ryan has just added an additional comment with a TODO that this `t.fail()` call should not be reached (which it is). This was not working before the PR though as well and the issue is unrelated, so it's just a net improvement here and not some left-over comment.

I guess this tests generally has not yet found its place respectively I remember that I also already stumbled upon it. We just not have this coherently in place how to deal with these kind of situations with two different `Common` instances. 

We likely want to first give these checks from here https://github.com/ethereumjs/ethereumjs-monorepo/issues/969 on Common consistency some more thought before we give this test more attention. This is rather just a test atm without the necessary implementation. 😋

P.S.: have removed the other TODO, this was actually not necessary (any more?).

holgerd77 (approved):
LGTM, cool, nice test additions! 😄

ryanio:
yes sorry I should have clarified with the unrelated TODO comments. thanks for some explanation, makes sense.

i found them false positives worth addressing. to fix the TODO, the goal would be to keep the line `t.fail('should throw error')`. i was able to add them in other places in the file that passed, but the two commented out ones were having some troubles, so that was my todo note for them.

thanks for review + merge!

packages/vm/tests/api/runTx.spec.ts
jochem-brouwer: I just checked but this `t.fail` should be called, right? The VM is on istanbul and the tx is on Berlin, so running access list txs on Istanbul should be invalid. Now, the test passes if the catch clause is not reached.

packages/vm/tests/api/runTx.spec.ts
```diff
@@ -230,8 +230,6 @@ tape('runTx() -> API parameter usage/data errors', (t) => {
       const tx = getTransaction(vm._common, txType.type, true)
       try {
         await vm.runTx({ tx })
-        // TODO uncomment:
-        // t.fail('should throw error')
```
jochem-brouwer: This should be in the test, right? If this `fail` is not reached then the test could pass if the `runTx` does not throw.

packages/vm/tests/api/runTx.spec.ts
holgerd77: Yes - again - therefore the "TODO" from Ryan here. Please read my longer comment from above (if you haven't already) with a suggestion on how to proceed here.

packages/vm/tests/api/runTx.spec.ts
holgerd77: Yes, it is in the tests added some commits later. If you comment on commits please always make sure to have a look at the final PR state, otherwise one is often falling in the trap (like here) of commenting on outdated code parts.

packages/vm/lib/runTx.ts
```diff
@@ -243,6 +243,17 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
...
+      // EIP-1559 spec:
+      // The signer must be able to afford the transaction
+      // `assert balance >= gas_limit * max_fee_per_gas`
+      const cost = tx.gasLimit.mul(tx.maxFeePerGas)
+      if (balance.lt(cost)) {
+        throw new Error(
+          `sender doesn't have enough funds to send tx. The max cost is: ${cost} and the sender's account only has: ${balance}`
+        )
+      }
+    }
```
holgerd77, line 256: @ryanio Hmm, when having a second look here, I guess we should rather adopt the 1559 tx [getUpfrontCost()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/eip1559Transaction.ts#L239) implementation instead of doing this just-one-tx-type check here in *addition* to the `getUpfrontCost()` check (above in the code). 🤔

This is what the `getUpfrontCost()` docs say it should do: "The up front amount that an account must have for this transaction to be valid"

holgerd77: (or are these two different things?)

ryanio: @holgerd77 under its current implementation it is a different check, since `getUpfrontCost` uses the `inclusionFeePerGas ` and also adds the tx's `value`:

```typescript
getUpfrontCost(baseFee: BN = new BN(0)): BN {
    const inclusionFeePerGas = BN.min(this.maxPriorityFeePerGas, this.maxFeePerGas.sub(baseFee))
    const gasPrice = inclusionFeePerGas.add(baseFee)
    return this.gasLimit.mul(gasPrice).add(this.value)
  }
```

compared to the 1559 validation check which simply does:

```typescript
const cost = tx.gasLimit.mul(tx.maxFeePerGas)
```

Example tx with:
  `gasLimit=21000, maxFeePerGas=100, maxPriorityFeePerGas=10, value=100, block.baseFeePerGas=7`

getUpfrontCost: `21000*(10+7)+100=357100`
1559 validation max cost check: `21000*100=2100000`

that said, we could change `getUpfrontCost` to the new validation logic, but it seems like it could be nice to have both, @jochem-brouwer wdyt?

holgerd77: Yeah, might be good to just keep both, thanks for these concrete examples.

PR #1276, opened 2021-06-01
VM: add check to runTx that tx.maxFeePerGas >= block.baseFeePerGas

ryanio:
Adding this check from the [1559](https://eips.ethereum.org/EIPS/eip-1559) spec:

```python
# ensure that the user was willing to at least pay the base fee
assert transaction.max_fee_per_gas >= block.base_fee_per_gas
```

Thanks to @fvictorio for the tip.


ryanio:
@jochem-brouwer noted that we already have this check in [`block.validateTransactions()`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/e31687a492e2d96abd4781101de75ff8c0a4d4f0/packages/block/src/block.ts#L238) which is applied in `runBlock` when `skipBlockValidation=false` [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/e31687a492e2d96abd4781101de75ff8c0a4d4f0/packages/vm/lib/runBlock.ts#L212), so this extra check may be unnecessary, I will leave this PR open for feedback whether it would be valuable to have or not.

holgerd77:
Yeah, I think this check is worth to have here as well. Will merge here once tests pass.

holgerd77 (approved):
LGTM

PR #1277, opened 2021-06-02
ethash: add test case for verifyPOW for uncles

acolytec3:
Adds an additional test case to ethash package to validate proof of work for blocks with uncles


holgerd77 (approved):
Looks good, thanks, nice additional test case! 😄

acolytec3:
@holgerd77 Thanks!  Just trying to start wrapping my mind around ethash.

Issue #1243, opened 2021-05-07
Static method to know if a chainId is included in `@etheruemjs/common`


alcuadrado:
I'm opening this issue to propose adding a method to figure out if a chainId can be used to construct an instance of `Common`.

We found ourselves doing something like this pretty often:

```ts
try {
  return new Common({ chain: chainId });
} catch (e) {
  return undefined;
}
```

ryanio:
that's interesting, I figured setting the chainId from user input would be infrequent. maybe something like `Common.isKnownChainId(chainId)`? it wouldn't support custom chains though, like if you were to let the user pass in a `Common.forCustomChain()`

holgerd77:
@ryanio I guess this is explicitly *not* for custom chains but to detect the chains "baked in"? 😄 

Some API specifics:

1. I would suggest we limit this to `BN` input directly to non run into these large-chain ID issues?
2. As a name I would do the slightly modified suggestion `isSupportedChainId`

So API suggestion would be:

```typescript
Common.isSupportedChainId(id: BN): boolean
```

PR #1281, opened 2021-06-03
Common: added static method for checking if chainId is supported

emersonmacro:
Addresses #1243. Feedback welcome!


packages/common/tests/supportedChainId.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+  t.test('Should return correct response', function (st: tape.Test) {
+    st.equal(Common.isSupportedChainId(new BN(1)), true, 'should return true for supported chainId')
+    st.equal(
+      Common.isSupportedChainId(new BN(1234)),
+      false,
+      'should return false for unsupported chainId'
+    )
+    st.end()
+  })
+})
```
holgerd77, line 15: Thanks for the PR! 😄 

Two things here:

1. We currently do not organize the tests in the libraries on that granularity level of having one file per method and so this test file would jump somewhat out of the line, can you please add to one of the existing files? I guess `chains.spec.ts` should be a good candidate. Not that we can't reorg at some PIT if we see the need, but these kind of things should generally be discussed before.
2. Can you also add a positive test case here for a chain ID existing?

emersonmacro: @holgerd77 Thanks for the review!

Re: 1, 👍. I wasn't sure what existing file it should go in so I made a new one, but your point makes sense. I'll move the test into `chains.spec.ts`.

Re: 2, I think I have that already here (https://github.com/ethereumjs/ethereumjs-monorepo/pull/1281/files#diff-00e881872be66da72fe4e5df60785751bb9b0fce068df279a7fd63b9dcdd0423R7), unless I'm not understanding? Actually now that I look at it, it is a little awkward having the top-level test be `should return the correct response`. I can break them out into separate tests.

emersonmacro: Also, I had a question unrelated to my code changes, just something I noticed. There's a file in `packages/common/tests` named `customChains.ts`. Should the file name be `customChains.spec.ts`?

holgerd77: @emersonmacro oh yes, good catch. 😄  We have a collecting issue open here where we collect some of the smaller fixes to then be worked on in batch (or along some other PRs) https://github.com/ethereumjs/ethereumjs-monorepo/issues/905, have added there so that it won't get lost.

holgerd77 (approved):
LGTM, thanks! 😀

PR #1293, opened 2021-06-10
util: rename bnToRlp to bnToUnpaddedBuffer

ryanio:
This PR renames `bnToRlp` to `bnToUnpaddedBuffer` to better describe its function (closes https://github.com/ethereumjs/ethereumjs-util/issues/284)

It also converts some uses of `unpadBuffer(toBuffer(bn))` to the function, and adds a test for it in the util package. `bnToRlp` is kept as a deprecated alias that can be removed in a future major version.

### DIFF
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -659,7 +659,7 @@ export class BlockHeader {
     ]
 
     if (this._common.isActivatedEIP(1559)) {
-      rawItems.push(bnToUnpaddedBuffer(this.baseFeePerGas))
+      rawItems.push(bnToUnpaddedBuffer(this.baseFeePerGas!))
     }
 
     return rawItems
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -3,6 +3,7 @@ import {
   Address,
   BN,
   bnToHex,
+  bnToUnpaddedBuffer,
   ecrecover,
   ecsign,
   intToBuffer,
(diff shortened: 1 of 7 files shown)


holgerd77 (approved):
Thanks, nice, yes this is actually a lot more readable. 😄 

Please always use `@deprectated`  for the code docs on such deprecations, otherwise e.g. Visual Studio is not picking this up by striking through. Have added on a review-commit.

Unrelated to PR: I think these two/three methods `bnToHex()`, `bnToUnpaddedBuffer()` (and `bnToRlp()`) are very much misplaced in the `types` module and should rather go to `bytes`. Have added to the v6 notes (let me know if you disagree).

ryanio:
@holgerd77 sounds good, ah thanks for the `@deprecated` tag I forgot to add that.

I think we can go ahead and move those methods to the `bytes` file as a "tiny fixes" issue rather than  breaking change, since they are exported the same on the top level `ethereumjs-util` module (i.e. `import { bnToHex } from 'ethereumjs-util'`) so it wouldn't be breaking unless someone was importing from the file directly, but we don't suggest or encourage that anywhere.

holgerd77:
@ryanio hmm, I would prefer to rather be cautious here and only do on next breaking release, also absolutely no urgency here. In the README these modules [are stated very explicitly](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/util#modules) and it is as valid to import from the respective modules (can be even seen as a bit more clean) than from the top level. So I guess this can't really be foreseen if people are doing this or not.

PR #1297, opened 2021-06-12
Migrate config to monorepo

ryanio:
This PR migrates `ethereumjs-config` to `./config`


ryanio:
seeing some lint errors on ci that I didn't encounter on my local, will investigate

ryanio:
ok this is ready for review now!

holgerd77:
Hi Ryan, great work here, really cool!! 🎉

I was getting a bit sceptical if we are on the right track again when I realized that we actually have *two* repositories outside the scope of the monorepo which we still actively develop (maintain), which are both the [rlp](https://github.com/ethereumjs/rlp) and the [ethereumjs-wallet](https://github.com/ethereumjs/ethereumjs-wallet) library, and at least the wallet library we'll likely don't want to take over I would say.

Had to let this sink in a bit, but my current thinking is now: realistically we are really not doing much on both of these libraries, and it will very likely be enough if we update on config there once every quarter or so (I would assume that in practice this number will turn out to be even lower, but we'll see). On these occasions we can very well just copy over the whole `config` folder from the monorepo (why not?). This is only a one-way street - I would very much assume - so no syncing issues or the like, and this seems to me very much the most pragmatic approach, really not introducing any substantial side effects.

And at the end having the config locally in the monorepo (as opposed to be served as a package) - and not having the config somewhat artifically integrated into this whole bootstrap process of the monorepo - will likely very much outweight with its advantages (simplicity, no bootstrap side effects, no package maintenance) the "burden" of these occasionally to do side tasks.

So will continue with a review here and eventually directly merge. 😄

holgerd77 (approved):
Cool, thanks for this careful execution e.g. with adopting the typedoc stuff preserving the VM specifics and the like. Some housekeeping stuff mentioned, also: we might want to also have a look at the developer docs here, eventually also some need to adopt, currently don't have the overview. Will also directly deprecate on the `ethereumjs-config` side (eventually it will make sense to align rlp and wallet with this `config` folder copy-over in some short time period some weeks or so after this integration after we - likely - have done some further optimizations and another round of evolution on things like the linting rules).

Will approve and merge.

holgerd77:
Short note: running `npm run bootstrap` from the root folder is still failing for me (also tested with an `npm run clean` before), erroring on the client build with:

```shell
@ethereumjs/client: Module not found: Error: Can't resolve 'node-fetch' in '/EthereumJS/ethereumjs-monorepo/node_modules/native-fetch/src'
```

ryanio:
> Short note: running `npm run bootstrap` from the root folder is still failing for me (also tested with an `npm run clean` before), erroring on the client build with:
> 
> ```shell
> @ethereumjs/client: Module not found: Error: Can't resolve 'node-fetch' in '/EthereumJS/ethereumjs-monorepo/node_modules/native-fetch/src'
> ```

hmm I haven't seen that error before, bootstrap is working for me on local and ci seems to be fine. any more details?

i did accidentally remove `ts-loader` devDep from the client mistakenly thinking it wasn't being used but it is in the webpack config. I added it back in a commit to #1300

holgerd77:
@ryanio also don't want to dig deeper right now, also not so urgent. Maybe this solves itself on the sidelines or someone has an easy take once this gets new exposure.

holgerd77:
> > Short note: running `npm run bootstrap` from the root folder is still failing for me (also tested with an `npm run clean` before), erroring on the client build with:
> > ```shell
> > @ethereumjs/client: Module not found: Error: Can't resolve 'node-fetch' in '/EthereumJS/ethereumjs-monorepo/node_modules/native-fetch/src'
> > ```
> 
> hmm I haven't seen that error before, bootstrap is working for me on local and ci seems to be fine. any more details?

Just discovered that this is actually also the reason why the Node version tests [fail](https://github.com/ethereumjs/ethereumjs-monorepo/runs/2834873711?check_suite_focus=true). 😜

holgerd77:
This seems to result in the `multiaddr` package from the `client` if one is following up the stack here:

image

So an update of the `multiaddr` package to [v9.0.1](https://github.com/multiformats/js-multiaddr/releases) might do it, there are (likely easy to handle) breaking changes on the v9 version switch though.

Do you might want to give this a try (in case you find useful), you are likely more familiar with the `multiaddr` package since you did all related code changes/introductions?

ryanio:
that's interesting, why would ci for PRs pass and these ones are getting stuck?

I also can't recreate on local so I'm not sure if I can open a PR to fix, if the PR would pass without it?

I do see on my local the file `node_modules/native-fetch/src/index.js`, what do you see? `native-fetch/dist/src` doesn't contain a `index.js` though, just `index.d.ts`, maybe that could be part of the issue.

I checked `multiaddrs` and it doesn't seem upgrading to the next major version would address this as `native-fetch` is on its latest version (v3.0.0) via the `dns-over-http-resolver` package.

holgerd77:
The problem is `node-fetch` not being available, not `native-fetch`, this seems to be triggered from within the `native-fetch`  package though. There is some package [node-fetch-npm](https://www.npmjs.com/package/node-fetch-npm) which is a fork of `node-fetch` and which is deprecated and there seems to be no releases available on npm any more, so might be that these have be depublished or something.

On my local installation I have only this `node-fetch-npm` package installed in the root `node_modules` folder, while `native-fetch` (oh my 😜 ) has a `peerDependency` entry for the `node-fetch` package.

Not sure, I guess a without-digging-too-deep fix for now should be to just add the `native-fetch` dependency to the root `package.json`? Will try.

PR #1301, opened 2021-06-16
Monorepo: Add src files to package build / VM lib -> src renaming

### LINKED ISSUE #1287: Include original .ts files in the NPM packages to allow proper debugging
The published package in NPM only contains the `/dist/*` files but when debugging with i.e. the Chrome DevTools and stepping into the library, an error is thrown because the original `/src/*.ts` files cannot be found and therefore the sourcemaps can't be used to display the proper source lines being executed.

Please update the NPM package to also include the `/src` directory.

Other packages distributing transpiled TypeScript files have the same issue. See https://github.com/MetaMask/eth-sig-util/issues/142.

holgerd77:
Fixes #1287 by adding `src` folders to the `package.json` file of all packages except the client which still generally has got a somewhat differing build structure.

This PR also does the `lib` -> `src` renaming for the VM package to align with the other packages. This has gotten a bit extensive 😋 , did my best to catch every occurrence.


ryanio (approved):
lgtm, nice!

packages/vm/typedoc.js
```diff
@@ -1,10 +1,9 @@
...
   extends: '../../config/typedoc.js',
-  entryPoints: ['lib'],
   exclude: [
     'tests/**/*.ts',
-    'lib/bloom/*.ts',
-    'lib/evm/**',
-    'lib/state/cache.ts',
+    'src/bloom/*.ts',
+    'src/evm/**',
+    'src/state/cache.ts',
```
holgerd77, line 7: Ah, important catch, I really thought I had it. Thanks for having another closer look here.

PR #1305, opened 2021-06-17
Fixing node-versions GitHub actions run

holgerd77:
Analyzing the node-versions GitHub actions failure. First run here is to check if this "properly fails" without any fix. 😋


holgerd77:
See comment here https://github.com/ethereumjs/ethereumjs-monorepo/pull/1297#issuecomment-862321573 for related discussion.

holgerd77:
Adding to the root `package.json` doesn't work, this errors out with:

```shell
> @ethereumjs/client@0.0.6 tape /home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/packages/client
> tape -r ts-node/register "test/!(integration)/**/*.spec.ts" "test/*.spec.ts"


/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/node_modules/multibase/src/util.js:3
const textDecoder = new TextDecoder()
                    ^
ReferenceError: TextDecoder is not defined
    at Object.<anonymous> (/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/node_modules/multibase/src/util.js:3:21)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/node_modules/multibase/src/base.js:3:24)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
```

This is actually a different error, don't have the motivation to follow-up on this though. Will now try to add to the client `package.json` as a dev dependency and see how this goes.

holgerd77:
Nope, also not working with `node-fetch` as a dev dependency in client `package.json`, same error as above ("TextDecoder is not defined").

Stumbled upon a thing when I tried to do `npm ls multibase` from the root folder, I get (on Node 14):
"This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!"

Likely someone with a newer Node version has generated this last `package-lock.json` file from root.

Not sure if this might be related, but I'll try a run with a regenerated lockfile with lockfileVersion@1 (not that I was aware before that these lockfile versions actually exist 😋 ).

holgerd77:
Hmm, the error (now again the "can't resolve node-fetch" error) also appears for me with a deleted `package-lock.json`, so I guess the test from above is useless (try to regenerate `package-lock.json` with an older Node version), will skip.

Nevertheless: I generally have the impression that this error is very likely triggered by the Node version the installation/bootstrap process is running from (this would make very much sense also considering that we are in this "Node versions test" context here).

holgerd77:
Will do another test run adding `node-fetch` to the client dependencies and not the dev dependencies, since error is triggered in the `multiaddr` package which is a main dependency as well.

holgerd77:
Side note: another thing to consider here: this is *only* triggered in the browser build of the client. 

(likely because of the following code from `/ethereumjs-monorepo/node_modules/native-fetch/src/index.js` where the bug is triggered and where there is `fetch` in the global context for a browser related build? Not sure.)

```javascript
'use strict'

if (globalThis.fetch && globalThis.Headers && globalThis.Request && globalThis.Response) {
  module.exports = {
    default: globalThis.fetch,
    Headers: globalThis.Headers,
    Request: globalThis.Request,
    Response: globalThis.Response
  }
} else {
  module.exports = {
    default: require('node-fetch').default,
    Headers: require('node-fetch').Headers,
    Request: require('node-fetch').Request,
    Response: require('node-fetch').Response
  }
}
```

holgerd77:
Ok, adding to dependencies goes back to the "TextDecoder is not defined" issue, just realized that the other package tests ("Run block",...) do run though.

I will now follow some hints from this issue description https://github.com/ipfs/js-ipfs/issues/3620 and also add the `web-encoding` package as a fixed dependency (now back in dev dependencies since it seems that this is sufficient).

holgerd77:
Not working.

Ok, I will now follow the path and for now just comment out the client node tests and add a TODO note. I guess it's safe to say that the client is very much the least important library to have Node version compatibility since we haven't even have a (recent) npm release out (side note: actually we really *should* do a release here at some point relatively soon, just occurring to me that there *are* the old version releases on npm and it's not getting any better if people install just these very old versions. Then rather release something still being incomplete in parts but much more up-to-date on the latest development state. Anyhow: another topic. 😄 ).

On the other side these Node tests are super-important for the other library releases, since it is critical that the VM or Tx libraries e.g. are not breaking on some node versions.

I will do another run with the other workflows deactivated and generally not cleaned up yet here though to see if the Node tests *are* actually passing right now (I would give this a 50% chance that they are not).

holgerd77:
Error is also triggered in the devp2p package, will deactivate as well.

holgerd77:
Ok, this is finally letting the node-versions tests pass:

image

I will clean up (also remove the CI run for node-versions since this is only run nightly due to resource constraints) and open up for review. 😄

packages/client/package.json
```diff
@@ -104,6 +105,9 @@
     "webpack": "^4.43.0",
     "webpack-cli": "^3.3.11"
   },
+  "devDependenciesComments": {
+    "node-fetch": "Hotfix for client browser build error in older Node versions (12/14), #1305, 2021-06-18"
```
ryanio, line 109: oh nice idea to add some context like this, i may borrow this idea in the future :)

.github/workflows/node-versions.yml
```diff
@@ -52,17 +52,26 @@ jobs:
...
+      # Temporarily disabled due to browser build error along node-versions CI run
+      # "TextDecoder is not defined" triggered in older Noder versions (Node 12/14)
+      # https://github.com/ethereumjs/ethereumjs-monorepo/pull/1305
+      # 2021-06-18
+      #
+      # TODO: give this another test run in times and reactivate (note that this
+      # might also be an issue which gets fixed downstream at some point in time)
+      # - name: Test Client
+      #   run: npm run test
+      #   working-directory: packages/client
```
ryanio, line 64: instead of disabling these tests completely, i will try adding `if: ${{ matrix.node >= 15 }}`

ryanio: awesome, looks like the skip worked:

Screen Shot 2021-06-21 at 10 28 17 AM

---

Screen Shot 2021-06-21 at 10 26 42 AM

---

Screen Shot 2021-06-21 at 10 27 28 AM

ryanio (approved):
lgtm!

packages/client/package.json
holgerd77: Yeah, I just thought that this would be just prone to get lonst again over time and I googled a bit how to get around this annoying comment restriction for `json` files and that's what I found. 😄 

I would be relatively confident that this has no side effects, we might want to keep an eye on this though for some time (if there are for example resulting build problems or the like if one adds unknown keys to `package.json` files? 🤔 But again: likely not I would assume).

.github/workflows/node-versions.yml
holgerd77: Ah, great, yeah, that's a lot better, thanks for doing this! 😄 👍

PR #1308, opened 2021-06-20
vm: Add first tests for bls12-g2multiexp precompile

acolytec3:
Adds tests for error conditions for the g2 multiexponent EIP-2537 precompile.  This is just scratching the surface but I noticed our test coverage was low in this area and I want to learn more about the EIP-2537 precompiles anyway so this should be a good start.

I got the test vector and return result from here - https://github.com/matter-labs/eip1962/blob/master/src/test/test_vectors/eip2537/g2_multiexp.csv

Will see if I can add additional test coverage on the other EIP-2537 precompiles from this test data in future PRs


ryanio:
Looks like karma got [stuck](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1308/checks?check_run_id=2866980443#step:10:1988) in the ci on the test:

```
Chrome Headless 91.0.4472.101 (Linux x86_64) LOG: '# Precompiles: BLS12-G2MultiExp'

Chrome Headless 91.0.4472.101 (Linux x86_64) LOG: '# BLS12-G2MultiExp'

20 06 2021 01:26:34.723:WARN [Chrome Headless 91.0.4472.101 (Linux x86_64)]: Disconnected (0 times) , because no message in 100000 ms.
Chrome Headless 91.0.4472.101 (Linux x86_64) ERROR
  Disconnected , because no message in 100000 ms.
```

I'll give it a try on local and see if I can debug what the problem might be.

acolytec3:
I'll have to look too. The EIP-2537 stuff isn't available in browser so
maybe that's causing the problem? I didn't think about that when running
the tests since I didn't think Karma would try to run the tests in browser.

ryanio:
haha oh yeah that would be it then, i forgot about that.

you can use
`import { isRunningInKarma } from '../util'`
and
```typescript
    if (isRunningInKarma()) {
      st.skip('BLS does not work in karma')
      return st.end()
    }
```

ryanio:
i'll add it as commit 👍

ryanio:
You know we already have this file, I wonder if we should consolidate them or do you think it's okay to have them in two separate places: https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/tests/api/EIPs/eip-2537-BLS.spec.ts

acolytec3:
Oh, hah, totally missed that. I could go either way on this one. We have an
existing pattern of one set of tests per precompile but this group from
EIP2537 is also all grouped together. I think I'll migrate these tests over
to the EIP test file and add the others there. They should all be similar
and go quickly

packages/vm/tests/api/EIPs/eip-2537-BLS.spec.ts
```diff
@@ -75,4 +76,31 @@ tape('EIP-2537 BLS tests', (t) => {
...
+        'hex'
+      ),
+      gasLimit: new BN(5000000),
+      _common: common,
+      _VM: vm,
+    })
+
+    const testVectorResult =
+      '0x00000000000000000000000000000000083ad744b34f6393bc983222b004657494232c5d9fbc978d76e2377a28a34c4528da5d91cbc0977dc953397a6d21eca20000000000000000000000000000000015aec6526e151cf5b8403353517dfb9a162087a698b71f32b266d3c5c936a83975d5567c25b3a5994042ec1379c8e526000000000000000000000000000000000e3647185d1a20efad19f975729908840dc33909a583600f7915025f906aef9c022fd34e618170b11178aaa824ae36b300000000000000000000000000000000159576d1d53f6cd12c39d651697e11798321f17cd287118d7ebeabf68281bc03109ee103ee8ef2ef93c71dd1dcbaf1e0'
```
ryanio, line 97: nice, could you add a little explanation of what this is actually testing and what the expected result represents?

packages/vm/tests/api/EIPs/eip-2537-BLS.spec.ts
```diff
@@ -75,4 +76,31 @@ tape('EIP-2537 BLS tests', (t) => {
     st.pass('BLS precompiles throw correctly on empty inputs')
     st.end()
   })
+
+  t.test('BLS12-G2MultiExp', async (st) => {
+    const common = new Common({ chain: 'mainnet', hardfork: 'berlin', eips: [2537] })
```
ryanio, line 81: you'll need here...

```suggestion
    if (isRunningInKarma()) {
      st.skip('BLS does not work in karma')
      return st.end()
    }
    const common = new Common({ chain: 'mainnet', hardfork: 'berlin', eips: [2537] })
```

packages/vm/tests/api/EIPs/eip-2537-BLS.spec.ts
acolytec3: Will do my best. I'd be lying if I said I really understood the details of the math but I can try. At this point, I basically just understand that this is one of the operations used in verifying BLS signatures and the input and output strings are just test data verified to work by matter labs.

ryanio: haha ok makes sense, yeah even just some info like that and a link to the file in OP would be good, I'm just thinking if someone wants to debug an issue there in the future it'd be helpful to at least have some context of what it is.

acolytec3:
@ryanio I added some basic comments to the test where it's been added to the EIP 2537 test cases with a link to my source test data. I'm thinking I'll add additional test cases for the other precompiles here as I have time and the same explanation will apply since the matter labs repo has valid test data for all of the precompiles as far as I can tell. What do you think?

ryanio:
nice the description looks great thanks for adding that, sure sounds good

acolytec3:
@ryanio Cool. Do y'all want to go ahead and merge this one or us it worth adding tests for all the other precompiles before proceeding?

ryanio (approved):
lgtm!

PR #1312, opened 2021-06-21
Add EIP-2098 support to work with compact 64-byte signatures

k06a:
https://eips.ethereum.org/EIPS/eip-2098

Moved from here: https://github.com/ethereumjs/ethereumjs-util/pull/295


ryanio (approved):
this LGTM! thanks @k06a for following up and opening this on the monorepo.

i like that you matched our code style and even improved the internal fromRpcSig flow so that the return values are simplified, i am finding it easy to follow.

the only other thing I could think to add would be the [test vectors](https://eips.ethereum.org/EIPS/eip-2098#test-vectors) from the EIP itself, but i liked that you adapted the existing test setup to test the compact representation.

(a note for us on the team to remember to regenerate the docs and add a line for this in the changelog for the next util release.)

PR #1316, opened 2021-06-23
block: check and convert difficulty, gasPrice, and value parameters if string integers in from-rpc

Ghorbanian:
Fixes issue #726. Also see PR #1258 for discussion. Internal helper function `numberToHex` added to `block` package.


holgerd77:
Thanks for the newly submitted PR! 😄 

Since we stick with option 2) now:

> We generally hang this lower and just add this to an internal helper function in a new util.ts file in the Block library which we should then exclude from documentation in the typedoc.js file.

Can you also include the helper directory to the `typedoc.js` `exclude` directive, so that we won't exclude this function on doc generation? That would be great, thanks!

Ghorbanian:
> Thanks for the newly submitted PR! smile
> 
> Since we stick with option 2) now:
> 
> > We generally hang this lower and just add this to an internal helper function in a new util.ts file in the Block library which we should then exclude from documentation in the typedoc.js file.
> 
> Can you also include the helper directory to the `typedoc.js` `exclude` directive, so that we won't exclude this function on doc generation? That would be great, thanks!

Done!

ryanio (approved):
looks great, thanks @Ghorbanian for sticking through with this!

Issue #1290, opened 2021-06-08
Tx/Common: Arbitrum One Support


holgerd77:
Part of #1288 

[Arbitrum One](https://arbitrum.io/) building on the Arbitrum Optimistic Rollup technology is a L2 solution by [Offchain Labs](https://offchainlabs.com/) which just recently launched in mainnet beta.

We should see if we get a connection going - eventually by adopting the Common library and a new network - and if it's possible to send a tx with our tx library.

There is a [public testnet](https://developer.offchainlabs.com/docs/public_testnet) which might be a good starting point for trying to send a test tx, token bridge (from Kovan, interestingly) can be found [here](https://bridge.arbitrum.io/) and explorer [here](https://explorer5.arbitrum.io/).

holgerd77:
I've now tried to compose together an Arbitrum testnet tx with the following script:

```typescript
/**
 * Usage:
 * ts-node create-arbitrum-tx.ts PRIVATE_KEY
 */
import { Transaction } from '../ethereumjs-monorepo/packages/tx/src'
import Common from '../ethereumjs-monorepo/packages/common/dist'

const from = '0xaaec86394441f915bce3e6ab399977e9906f3b69'

const PRIV_KEY = process.argv[2]
const to = '0xe13ece23b514caa5b53395c01d0d53d1843258ad'

const common = Common.forCustomChain(
  'mainnet',
  {
    name: 'arbitrum-testnet-v5',
    networkId: 144545313136048,
    chainId: 144545313136048,
  },
  'berlin',
)

const txData = {
  from,
  nonce: 0,
  gasPrice: 1000000000,
  gasLimit: 100000,
  to,
  value: 100,
}

const tx = Transaction.fromTxData(txData, { common })
const signedTx = tx.sign(Buffer.from(PRIV_KEY, 'hex'))

console.log(signedTx.serialize().toString('hex'))
```

This results in the following serialized tx format: `f86b80843b9aca00830186a094e13ece23b514caa5b53395c01d0d53d1843258ad6480870106ed28163b84a038234a7282181eac9cb3e735357b60f92db0c250d868622308b7b4c7e9ec9ab0a00f6a9412b5061a439105fe46e6879f346eddb39cfb446a8a1ff299379c901ca5`

holgerd77:
There is a public [RPC endpoint](https://kovan5.arbitrum.io/rpc) for the testnet.

I've now sent the tx with:

```shell
[...]

Issue #1289, opened 2021-06-08
Tx/Common: Polygon PoS Chain Support


holgerd77:
Part of #1288 

[Polygon](https://polygon.technology/) is a protocol and framework for building Ethereum compatible networks which gained a lot of traction lately. Note that the dedicated chains are named differently from the protocol, there is e.g. the [Polygon PoS Chain](https://docs.matic.network/docs/develop/ethereum-matic/pos/getting-started/) (this is the main one I guess?).

We should test if we can get a connection going with the tx library respectively the underlying common library and ultimately send a transaction and doing an according PR on the libraries. If setting up a connection is not possible with our current library structure we should alternatively mention in the tx README.

ryanio:
I will give this a try later today :)

ryanio:
Cool news, so Matic has a Mumbai testnet which extends goerli, so with a custom common I was able to send `eth_sendRawTransaction` to the rpc endpoint and it got confirmed within seconds.

```typescript
const customChainParams = { name: 'matic-mumbai', chainId: 80001, networkId: 80001 }
const common = Common.forCustomChain('goerli', customChainParams)
```

I used our test dev account: [`0xaaec86394441f915bce3e6ab399977e9906f3b69`](https://polygon-explorer-mumbai.chainstacklabs.com/address/0xaAeC86394441f915bcE3E6AB399977e9906f3B69/transactions), funded it with 0.1 MATIC (native chain currency) from the [faucet](https://faucet.matic.network/), then sent out some txs with this test script I was using a few weeks ago to bombard baikal 😄

```typescript

const http = require('http')
import { Transaction } from '../ethereumjs-monorepo/packages/tx/src'
import Common from '../ethereumjs-monorepo/packages/common/dist'

const FROM = '0xaaec86394441f915bce3e6ab399977e9906f3b69'
const to = '0xe13ece23b514caa5b53395c01d0d53d1843258ad'

const customChainParams = { name: 'matic-mumbai', chainId: 80001, networkId: 80001 }
const common = Common.forCustomChain('goerli', customChainParams) as any

const txData = {
    nonce: 3,
    gasPrice: 20000000000,
    gasLimit: 210000,
    to,
    value: 100,
    data: '0x',
}

const start = async () => {
    for (let i = 0; i < 1; i++) {
        const tx = Transaction.fromTxData({ ...txData, nonce: txData.nonce + i }, { common })
[...]

Issue #1033, opened 2021-01-07
Using xDAI chain not working


Aeroxander:
Using the xDAI chain gives a `Error HH604: Error running JSON-RPC server: Chain with ID 100 not supported` error in Hardhat

Original hardhat issue: https://github.com/nomiclabs/hardhat/issues/1008

cgewecke:
@Aeroxander Could you provide some steps to reproduce this error (or a link to a public repo)?

Using HH's Greeter sample project and running the node as shown in the linked issue I see
```
$ npx hardhat node --fork https://rpc.xdaichain.com
Started HTTP and WebSocket JSON-RPC server at http://127.0.0.1:8545/

Accounts
========
etc...
```

When I run the tests, the node errors but doesn't mention common...
```
web3_clientVersion (2)
eth_chainId
eth_accounts
eth_chainId
eth_estimateGas
  Error: Invalid RpcBlockWithTransactions or null
      at codec.decode.fold (/Users/cgewecke/code/typechain-test/node_modules/hardhat/src/internal/hardhat-network/jsonrpc/types.ts:8:11)
      at Left.fold (/Users/cgewecke/code/typechain-test/node_modules/io-ts/node_modules/fp-ts/lib/Either.js:113:16)
      at Object.decode (/Users/cgewecke/code/typechain-test/node_modules/hardhat/src/internal/hardhat-network/jsonrpc/types.ts:6:30)
      at JsonRpcClient._perform (/Users/cgewecke/code/typechain-test/node_modules/hardhat/src/internal/hardhat-network/jsonrpc/client.ts:230:27)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

alcuadrado:
I don't think this is caused by ethereumjs, but rather some of our validation logic being too strict.

holgerd77:
I've just opened a PR to improve the custom chain support situation on `Common` at #1034, just to mention here.

I've not very much insights on the xDAI chain, is this a "normal" chain with hardforks and everything? Does this chain has got something like a `genesis.json` file or does someone have some informational source here to link to? Would be very interested to test the `Common` library against an xDAI configuration.

cgewecke:
(Cross-linking to a recent related Hardhat issue: [hardhat 1177][1].) 

[1]: https://github.com/nomiclabs/hardhat/issues/1177

holgerd77:
Will close in favor of #1323

PR #1317, opened 2021-06-24
Common: Reworked Custom Chain Instantiation, Polygon/Arbitrum Supported Chains Integration

holgerd77:
Ok, this is a suggestion for an API consolidation on `Common.forCustomChain()` together with an integration of the the basic supported chain integration as discussed with @ryanio along the Arbitrum #1290 and Polygon #1289 work. I am actually pretty happy with this, gave this some longer thought how a nice API could look like. 😄 

`Common.forCustomChain()` was very much outdated on the API side, since it didn't catch up with e.g. the transition of `Common` parameters to be served as an options dict (12 months or so ago) and therefore didn't support things like setting EIPs. At the same time the API was generally not so optimal and user friendly, leading to initialization "monsters" like:

```javascript
const common  = Common.forCustomChain(
    'mainnet',
    {
      name: 'custom-chain',
      networkId: 1234,
      chainId: 1234,
    },
    'muirGlacier'
  )
```

I've now deprectated the `Common.forCustomChain()` method and created a new `Common.custom()` method with a simplified API: `static custom(chainParamsOrName: Partial<Chain> | CustomChains, opts: CustomCommonOpts = {}): Common` (actually this doesn't look simplified when looking at the API itself, but you'll see it in the usage context 😋).

The method has now a dedicated options dict which now also takes in the options like `eips` like the main constructor. The `baseChain` - being a separate param before - has now been integrated there as well and defaults to `mainnet` if not provided. Chain name - if not provided - will be set to `custom-chain` (instead taking the base chain name), which should suffice in many cases.

So in the very most cases a custom chain can now be initialized as simply as:

```javascript
const common = Common.custom({ chainId: 1234 })
```

😄 

(for many use cases it should actually not be needed to provide a dedicated network ID, since the very most applications (e.g. the tx library for signing) nowadays take the chain ID as "the important thing", we also don't use this further within the `Common` library itself).

Along this refactoring work the `custom()` method now alternatively also takes a `string` as a first input and therfore integrates the supported custom chains as discussed in the above-mentioned L2 integration issues.

A basic `Common` for a Polygon chain can e.g. be instantiated like this:

```javascript
const common = Common.custom(CustomChains.PolygonMainnet)
```

PR is still a bit WIP, need to settle this down a bit and have another look and also still need to adopt on the `forCustomChain()` usage within the monorepo libraries. Still pushing though for an early look and eventually review.


holgerd77:

I might also do some more experimentation on xDai to eventually integrate - see also #1033 - but in doubt this can be merged indepedently, can also do some subsequent dedicated PR on this.

holgerd77:
Ok, now also added xDai after the successfull test in #1323. A bit hard to name, I tested `XDaiChain` to stick to the capital letter as a first letter normal convention but I found this too ugly and switched to `xDaiChain`, let me know if you go along here. 😄

ryanio:
looks great! just adding a commit to update some of the link resolutions, i'm really becoming a fan of these 😄

ryanio (approved):

PR #1329, opened 2021-06-30
common: fix hardforkBlockBN() to correctly return null for unscheduled hardforks

### LINKED ISSUE #1328: Common: hardforkBlockBN() (hardforkIsActiveOnBlock()) Bug for HFs with null value
`hardforkBlockBN()` is currently returning an empty `BN` instance for HFs with a null value, see e.g.:

grafik

This is likely not what we want - this should rather return `null`. This also leads to at least one internal bug with `hardforkIsActiveOnBlock()` falsely returning `true` for a `null` HF check (also see screenshot above).

Issue is somewhat urgent since we should include in #1327, fix should be simple though.

ryanio:
Fixes #1328


holgerd77:
Thanks for taking on this so quickly! 😄 I am really getting somewhat unsure again though here and wonder if this has too much the character of a breaking change will all these necessary `this._common.hardforkBlockBN('dao')!` (and the like) additions.

I therefore have a tendency not to rush this into tomorrow's release, otherwise the risk of introducing something annoying and TypeScript code breaking is too high I guess. And with the proper additions of all the `london` blocks this issue get's at least a bit less pressing.

So let's please give this some couple of days of reflection, I've added the "blocked" label for now.

holgerd77:
To start a bit on the discussion:

The current situation is that we have "two types" of hardforks added to the chain json files having `null` values. The ones are hardforks which have already passed but which never happened on a specific chain (e.g. the `dao` HF for Rinkeby), the others are intended future HFs which just doesn't have a block number yet ('london').

Regarding this first type, here is e.g. the corresponding entry on Rinkeby:

```json
{
  "name": "dao",
  "block": null,
 "forkHash": null
}
```

I've added these myself some years ago when I wrote the initial version of the library. I am very much convinced for some time now that this just was a mistake and these not-activated HFs just shouldn't appear *at all* in a respective chain file (Rinkeby just *doesn't know* about this HF, no matter if it happened somewhere else or not). I think we can just remove these entries without any side effects, this seems very much like a bug fix to me. Then the problem space is already reduced and we only have to deal with type 2.

So Type 2 has totally different characteristics. This is a HF which is a) known to occur and is b) known to have a higher block number than all the existing HFs. Not sure if this is a good idea, but one way to handle this would also be to just add a artificially high dummy block number on first round, like `9999999999` or whatever. This would at least provide a way to better ensure larger parts of the library logic (e.g. this `hardforkIsActiveOnBlock()` method) to work more aligned with what would be expected.

Thoughts or ideas? 🤔 (also feedback on the "let's remove type 1 entries would be nice)

acolytec3:
@holgerd77 For "type 1" scenarios, makes total sense to me to just remove them as that shouldn't impact the methods like `activeHardorks` Wouldn't setting an artificially high block number for "type 2" return odd results on `hardforkBlockBN` and related methods?

holgerd77:
> Wouldn't setting an artificially high block number for "type 2" return odd results on hardforkBlockBN and related methods?

Yes, actually it would (it would just the artificially high block number 😋). It would just be more accurate (on the existing code base) for things like HF comparison.

On some second thought: *once* we have removed all these type 1 `null` occurrences (`null` for past HFs not applied) we can just switch to give `null` *always* the semantics "stands for a HF that is not yet applied and will occur some time in the future) and make sure to do our implementations of the methods after this definition (so e.g. to make sure that a HF comparison of an existing/applied HF against a `null` HF returns the `null` HF being regarded as greater, while the comparison result against a non-existing (`undefined`) HF might be different.

I guess that's very much the cleaner solution together with communicating in the release notes that block numbers for some specific HF are just not finalized yet (what we do already).

ryanio:
thanks @emersonmacro for opening #1344, just merged it in, ok what should be the next step here for this PR then?

holgerd77:
@ryanio I am really getting a bit paranoid with our latest Hardhat breakages and I am getting pretty cautious on changes like this one. While from a functionality perspective this is a bugfix (an empty `BN` for `null` just doesn't make that much sense) at least on the `TypeScript` level this is a breaking change adding an additional `null` return type.

It definitely also doesn't make sense to wait for the next release on this, but we should minimally wait for the Hardhat E2E tests #1348 to be working to that we test against at least one production library.

Not sure, I guess we don't have some "official policy" or even some common sense agreement on TypeScript vs JavaScript breaking (or: potentially breaking) changes. I also haven't given this too much thought yet. I guess the main difference is that TypeScript breaking changes (like this one) are a lot more likely to already occur on third-party-integrators CI runs (if CI is using the TypeScript code) and this is therefore in many cases not as grave as doing breaking changes which also trickle down to the JavaScript distribution version (which then eventually break directly on consumer installation).

Is this a fair approach or is this thought too short? More thoughts and/or opinions on this?

acolytec3:
@ryanio @holgerd77 This should be ready for review now that the hardhat tests are integrated. I merged in latest master and fixed a couple of small CI issues on the tests. I also added Shanghai as a new hardfork to the enum with no scheduled block since London is all set now so the tests weren't passing there.

ryanio:
@holgerd77 yes I agree it's important to keep in mind types breaking changes as they can be disruptive. As you point out in its current state this PR would require downstream users of the method to update to the non null assertion.

I think this PR is a good bug fix nonetheless, and if we want to be sensitive about the types change we can maintain the original signature as a forced override over the fix, and then change the type in the next major release as a todo item. But then the newest users of the library would be using the outdated types, so I'm really not sure if it's better to just rip the bandaid off 😄 I'll let you decide.

thanks so much @emersonmacro for removing the extra dao hardfork entries and @acolytec3 for working on the hardhat e2e tests!

PR looks great. i'd approve it but it's my own.

holgerd77:
I have some tendency to merge here but don't dare yet. 😛

Will let this sink in for another day.

One pro-merge-as-is argument is also that the `hardforkBlockBN()` method is pretty new and likely not many people have switched to this method yet.

ryanio:
@holgerd77 yes, although it does change the non-BN method as well. We could just save this for v6 breaking change release as it seems like we'll be ready for that tomorrow as soon as mainnet forks London successfully.

holgerd77:
Ok, I will merge this in once the tests pass. At the end this was wrong and this is a bug fix and we can't wait for breaking releases to do these kind of fixes, even if they might be mildly annoying/disruptive on CI or the like for implementers.

ryanio:
force pushing to restart stuck ci jobs

ryanio:
i'll go ahead and merge in since you gave the go ahead

PR #1324, opened 2021-06-29
devp2p/client: add les v3/v4

ryanio:
This PR fixes `devp2p/examples/peer-communication-les.ts` by upgrading to les v4.

Sample run:

```
➜  examples git:(les-v4) ✗ ts-node peer-communication-les.ts
Add peer: 52.169.42.101:30303 Geth/v1.10.2-unstable-aab35600-20210322/linux-amd64/go1.16.2 (les4) (total: 1)
Add peer: 159.89.28.211:30303 Geth/v1.10.5-unstable-10eb654f-20210623/linux-amd64/go1.16.5 (les4) (total: 2)
----------------------------------------------------------------------------------------------------------
block 8845079 received: 0a3040af08dc247ad9e4290285360b9af0d58b8363c8cd1f0e0832bb2f98630a (from 52.169.42.101:30303)
----------------------------------------------------------------------------------------------------------
----------------------------------------------------------------------------------------------------------
block 8845079 received: 0a3040af08dc247ad9e4290285360b9af0d58b8363c8cd1f0e0832bb2f98630a (from 159.89.28.211:30303)
----------------------------------------------------------------------------------------------------------
```

WIP, now working on adding to the client to start light sync from genesis block.

Current challenge is finding LES peers that don't disconnect with `TOO_MANY_PEERS` (happening for me on ropsten/rinkeby/goerli). When trying the rinkeby bootstrap nodes (which the devp2p example above successfully works), getting a consistent but weird error from both peers `CLIENT_QUITTING`:

```
  devp2p:les Send GET_BLOCK_HEADERS message to 159.89.28.211:30303: c601c401328080 +1s
  devp2p:les Send GET_BLOCK_HEADERS message to 52.169.42.101:30303: c602c433328080 +1ms
  devp2p:rlpx:peer Received header 52.169.42.101:30303 +327ms
  devp2p:rlpx:peer Received body 52.169.42.101:30303 01c108 +0ms
  devp2p:rlpx:peer Received DISCONNECT (message code: 1 - 0 = 1) 52.169.42.101:30303 +0ms
  devp2p:rlpx:peer DISCONNECT reason: CLIENT_QUITTING 52.169.42.101:30303 +0ms
  devp2p:rlpx Restart connection refill .. with selector 0 peers: 1, queue size: 0, open slots: 24 +1ms
  devp2p:rlpx 52.169.42.101:30303 disconnect, reason: CLIENT_QUITTING +0ms
[...]

PR #1331, opened 2021-07-02
devp2p/client: upgrade to eth/66

ryanio:
This PR upgrades `devp2p` and `client` to communicate `eth/66` since geth plans to deprecate `eth/65` after london is activated on mainnet.

I updated devp2p's `peer-communication.ts` which was working for me (would appreciate someone else to run it to verify everything looks good), along with client's full sync with `npm run client:start`.

`devp2p` has no breaking changes as the payload logic is handled in the `client`, but the `client` will no longer support versions prior to `eth/66` since structural changes needed to be made to support the encoding and decoding of messages to handle the new `reqId` field.


packages/client/lib/sync/fetcher/blockfetcher.ts
```diff
@@ -24,13 +24,13 @@ export class BlockFetcher extends BlockFetcherBase<Block[], Block> {
...
-    })
-    const bodies: BlockBodyBuffer[] = <BlockBodyBuffer[]>(
-      await peer!.eth!.getBlockBodies(headers.map((h) => h.hash()))
-    )
+    const headers = (
+      await (peer!.eth as EthProtocolMethods).getBlockHeaders({
+        block: first,
+        max: count,
+      })
+    )[1]
```
acolytec3, line 32: if (!headers) {
      this.config.logger.error(`peer ${peer?.id} returned no headers for block ${first}`)
      return
    }

acolytec3 (changes requested):
I was able to successfully run `peer-communication.ts` so that looks good.  As far as syncing the chain, here are my observations:

- mainnet, ropsten, and goerli all start syncing just fine
- rinkeby doesn't seem to find any suitable peers so not sure if the bootnodes there just haven't adopted `eth/66` or not
- kovan won't start syncing though that doesn't appear to be an issue related to this PR as it throws the same error when trying to sync from the `master` branch

packages/client/lib/sync/fetcher/blockfetcher.ts
```diff
@@ -24,13 +24,13 @@ export class BlockFetcher extends BlockFetcherBase<Block[], Block> {
...
+    const bodies = (await peer!.eth!.getBlockBodies({ hashes: headers.map((h) => h.hash()) }))[1]
```
acolytec3, line 33:
```suggestion
    if (!headers) {
      this.config.logger.error(`peer ${peer?.id} returned no headers for block ${first}`)
      return
    }
    const bodies = (await peer!.eth!.getBlockBodies({ hashes: headers.map((h) => h.hash()) }))[1]
```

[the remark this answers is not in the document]
ryanio: thanks for the in-depth test runs, yeah this is tricky and at first I thought it was because of [this change](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1324/files#diff-5cdab2534be43ab3aaced39ef3b1124d0329d2c8c897b9343ff914c02b227b36L187-R187) but that turns out to be in a separate PR 😄 thanks for the code suggestion that seems reasonable I'll give it a run today and if it handles the issue well I will merge it in :)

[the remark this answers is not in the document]
acolytec3: If I'm reading the logs correctly, it looks like the peer just isn't returning anything, so maybe they're missing a block for some reason?  It wasn't obvious to me from the logs but it didn't seem to break anything.

acolytec3:
Ironically, it looks like my suggested change is causing the CI to fail though I can say with confidence that this is not always falsy.   Is this a case of `//@ts-ignore` or whatever the equivalent linter ignore command is?
```
error: Unnecessary conditional, value is always falsy (@typescript-eslint/no-unnecessary-condition) at lib/sync/fetcher/blockfetcher.ts:33:10:
  31 |       })
  32 |     )[1]
> 33 |     if (!headers) {
```

packages/client/lib/sync/fetcher/blockfetcher.ts
```diff
@@ -24,13 +24,17 @@ export class BlockFetcher extends BlockFetcherBase<Block[], Block> {
...
     const { task, peer } = job
     const { first, count } = task
-    const headers = await (peer!.eth as EthProtocolMethods).getBlockHeaders({
-      block: first,
-      max: count,
-    })
-    const bodies: BlockBodyBuffer[] = <BlockBodyBuffer[]>(
-      await peer!.eth!.getBlockBodies(headers.map((h) => h.hash()))
-    )
+    const headers = (
```
acolytec3, line 27:
```suggestion
       const headersResponse = await (peer!.eth as EthProtocolMethods).getBlockHeaders({
      block: first,
      max: count,
    })
    /* eslint-disable @typescript-eslint/no-unnecessary-condition*/
    // Conditional catches occasional null responses from peers who do not return block headers from peer.eth request
    if (!headersResponse) {
      this.config.logger.warn(
        `peer ${peer?.id} returned no headers for segment starting with block ${first}`
      )
      return []
    }
    /* eslint-enable  @typescript-eslint/no-unnecessary-condition*/
    const headers = headersResponse[1]
```

acolytec3: This works.  I didn't realize the `getBlockHeaders` actually returned an array so the error was trying to access the headers from that array when a peer returns a null response.  This produces the intended result of an early return and logging a warning saying what happened (though maybe this should be a debug thing since the client will just fetch them from elsewhere if it doesn't get anything)?

ryanio: ah thanks, I think warn is fine for this

holgerd77:
Cool! 🎉

PR #1327, opened 2021-06-30
New Library Releases (tx.supports() integration, Common custom chains + enum Chain/HF, Util v7.1.0)

holgerd77:
Releases with finalized `london` support and other substantial updates.

PR #1326, opened 2021-06-29
blockbuilder: allow customizable baseFeePerGas

ryanio:
This PR allows the blockbuilder to set a customizable `baseFeePerGas` if set in the `headerData`.

This was requested by hardhat to let users manipulate the base fee to test e.g. varying edge cases.

PR #1295, opened 2021-06-10
New Releases VM v5.4.1, Tx v3.2.1, Common v2.3.1

holgerd77:
Ok, a new round of releases integrating the latest EIP-1559 spec changes and updating on tx library usability and generally integrating user feedback. ~Still WIP, VM is still missing.~

I very much anticipate a relatively narrow cadence of follow-up releases in the next weeks on approaching the `london` HF (so at least every two weeks, eventually also more on a weekly basis), so no worry if some feature or fix hasn't made this yet in this release.

We generally should release here soon since there are features/fixes both HardHat and Web3.js are waiting for to integrate.

PR #1333, opened 2021-07-06
VM: New Release v5.4.2

holgerd77:
This is for a time critical in-between VM release requested by Hardhat (@alcuadrado @@fvictorio) since #1327 is taking more time than anticipated. Single change is the "allow customizable baseFeePerGas" change in the block builder from 487b472 (#1326).

PR is not for merge. I branched of 26de933 (merge commit from the latest releases PR #1295) and re-added this one small change manually and added release notes.

Instead this is just to check if CI passes. If it does I will do the release locally, then close here and adopt the CHANGELOG from the upcoming releases in #1327 accordingly.

Replaces #1332

### DIFF
--- a/packages/blockchain/src/db/manager.ts
+++ b/packages/blockchain/src/db/manager.ts
@@ -8,12 +8,12 @@ import {
   BlockBodyBuffer,
 } from '@ethereumjs/block'
 import Common from '@ethereumjs/common'
+import { CliqueLatestSignerStates, CliqueLatestVotes, CliqueLatestBlockSigners } from '../clique'
 import Cache from './cache'
 import { DatabaseKey, DBOp, DBTarget, DBOpData } from './operation'
 
+// eslint-disable-next-line implicit-dependencies/no-implicit
 import type { LevelUp } from 'levelup'
-import { CliqueLatestSignerStates, CliqueLatestVotes, CliqueLatestBlockSigners } from '../clique'
-
 const level = require('level-mem')
 
 /**
--- a/packages/blockchain/src/index.ts
+++ b/packages/blockchain/src/index.ts
@@ -20,6 +20,7 @@ import {
 
 const debug = createDebugLogger('blockchain:clique')
 
+// eslint-disable-next-line implicit-dependencies/no-implicit
 import type { LevelUp } from 'levelup'
 const level = require('level-mem')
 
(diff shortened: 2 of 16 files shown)


holgerd77:
Ah, these things are really somewhat trickier than first thought. When opening this towards `master` this creates some merge conflicts (which makes sense). So I now instead branched off another branch `vm-release-v542-base` from the same commit but not doing the respective changes and will open this PR towards this new branch as a base branch (main goal of everything: at least once run the CI).

holgerd77:
Oh, CI is failing.

Impressive.

I have to say I am not sure if I find the ressources to dig deeper here right now. We'll see if this in-between release can be done or if we will just switch back to the former plan of releasing together with #1327.

acolytec3:
@holgerd77 It looks like its the linter complaining about packages that are missing from package.json in a variety of packages because we have the [`no-implicit-dependencies`](https://palantir.github.io/tslint/rules/no-implicit-dependencies/) set in TSLint.  Seems like this shouldn't be a new issue though since a lot of the imports have been around for a while.

holgerd77:
Hmm, I branched off from an *old* state here actually though. Maybe this is this old problem (I guess we had?) where the linter was not working any more and @ryanio updated the version at some point? 🤔

Maybe he can say something on this accordingly, we shouldn't spent too much time here though (since this is just an eventual extra release).

ryanio:
hm, this is odd it is only complaining about `no-implicit-dependencies`. the base branch goes back to before the monorepo `./config` updates, but it should be using the prior `@ethereumjs/eslint-config-defaults` config that was set up and working properly so i'm a bit scratching my head why it would be complaining about this now. afaik there were no changes in between, especially to this rule.

i thought it could be a confused package-lock, but the one in the base branch seems to be fine.

i'll keep looking further and will try to run the lint locally to see if the same issue happens.

ryanio:
ahh it looks like `eslint-plugin-implicit-dependencies` made a new release `v1.1.0` [4 days ago](https://github.com/lennym/eslint-plugin-implicit-dependencies/commit/7056d3f43cf4208ac87d463648c8f9c7ac3b97e7) that added support for `import` statements.

holgerd77:
Seems we've got a mainnet block now, so please let's not move this further. Then we can directly do the full releases and Hardhat can directly integrate with the finalized London blocks and all the other improvements.

holgerd77:
Maybe I'll do the release anyhow tomorrow morning since it might still make sense to wait 1-2 more days until releasing with the settled blocks.

Anyhow @ryanio thanks for fixing! 😀

ryanio:
yes no problem, sounds good, let me know if you'd like me to apply that commit to the other release PR then (or you can go ahead and do that 👍)

holgerd77:
Just published as v5.4.2 on npm, will close here.

PR #1334, opened 2021-07-06
block: Add fix for kovan nonce

acolytec3:
The Kovan block nonce is a hex string of 65 0s and our `block` package expects a nonce of 8 bytes in length so trying to sync Kovan with the `client` results in an error.  
```
INFO [07-06|11:00:47] Connecting to network: kovan
ERROR [07-06|11:00:47] Error: nonce must be 8 bytes, received 65 bytes
    at BlockHeader._validateHeaderFields (/home/jim/development/ethereumjs-monorepo/packages/block/src/header.ts:323:13)
    at new BlockHeader (/home/jim/development/ethereumjs-monorepo/packages/block/src/header.ts:267:10)
    at Function.fromHeaderData (/home/jim/development/ethereumjs-monorepo/packages/block/src/header.ts:77:12)
    at Function.fromBlockData (/home/jim/development/ethereumjs-monorepo/packages/block/src/block.ts:36:32)
    at Function.genesis (/home/jim/development/ethereumjs-monorepo/packages/block/src/block.ts:131:18)
    at Blockchain._init (/home/jim/development/ethereumjs-monorepo/packages/blockchain/src/index.ts:327:28)
    at Blockchain.initAndLock (/home/jim/development/ethereumjs-monorepo/packages/blockchain/src/index.ts:413:5)
    at Blockchain.getLatestHeader (/home/jim/development/ethereumjs-monorepo/packages/blockchain/src/index.ts:776:12)
    at Chain.update (/home/jim/development/ethereumjs-monorepo/packages/client/lib/blockchain/chain.ts:222:22)
    at Chain.open (/home/jim/development/ethereumjs-monorepo/packages/client/lib/blockchain/chain.ts:186:5)
```
Adds hack to `_validateHeaderData` for Kovan genesis nonce so that client can try to start syncing Kovan testnet.  

What does the team think?  Do we need to support Kovan in the client?

### DIFF
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -301,6 +301,7 @@ export class BlockHeader {
    */
   _validateHeaderFields() {
     const { parentHash, stateRoot, transactionsTrie, receiptTrie, mixHash, nonce } = this
+
     if (parentHash.length !== 32) {
       throw new Error(`parentHash must be 32 bytes, received ${parentHash.length} bytes`)
     }
@@ -320,7 +321,14 @@ export class BlockHeader {
     }
 
     if (nonce.length !== 8) {
-      throw new Error(`nonce must be 8 bytes, received ${nonce.length} bytes`)
+      // Hack to check for Kovan due to non-standard nonce length (65 bytes)
+      if (this._common.networkIdBN().eqn(42)) {
+        if (nonce.length !== 65) {
+          throw new Error(`nonce must be 65 bytes on kovan, received ${nonce.length} bytes`)
+        }
+      } else {
+        throw new Error(`nonce must be 8 bytes, received ${nonce.length} bytes`)
+      }
     }
   }
 


ryanio:
Nice, thanks for opening this PR. It looks like [all blocks](https://kovan.etherscan.io/block/25951140) (not just genesis) have this 65-byte length nonce so it may make sense to check if common is on chain `kovan` and pass if it is exactly 65 bytes instead of checking the stateRoot to be the genesis block.

acolytec3:
> Nice, thanks for opening this PR. It looks like [all blocks](https://kovan.etherscan.io/block/25951140) (not just genesis) have this 65-byte length nonce so it may make sense to check if common is on chain `kovan` and pass if it is exactly 65 bytes instead of checking the stateRoot to be the genesis block.

Cool, I've updated to check for networkId instead of my previous hack.  It still starts up but looks like none of the bootnodes we have in the genesis config are accepting incoming connections so I'm not able to get any blocks.  But, I think it's probably worth merging this if everyone is okay with it so if someone else is running a kovan node and wants to test with us, they have access to it.

packages/block/src/header.ts
```diff
@@ -320,7 +321,10 @@ export class BlockHeader {
     }
 
     if (nonce.length !== 8) {
-      throw new Error(`nonce must be 8 bytes, received ${nonce.length} bytes`)
+      // Hack to check for Kovan due to non-standard nonce length (65 bytes)
+      if (this._common.networkIdBN().toNumber() === 42 && nonce.length !== 65) {
+        throw new Error(`nonce must be 8 bytes, received ${nonce.length} bytes`)
```
ryanio, line 326:
```suggestion
      if (this._common.networkIdBN().eqn(42)) {
        if (nonce.length !== 65) {
          throw new Error(`nonce must be 65 bytes on kovan, received ${nonce.length} bytes`)
        }
      } else {
        throw new Error(`nonce must be 8 bytes, received ${nonce.length} bytes`)
      }
```

acolytec3:
@ryanio I've fixed the typo and looks like everything is passing now.  Anything else you'd add at this point?

ryanio (approved):
lgtm!

holgerd77:
I am not sure if I am a fan of this change. Some background: Kovan uses the [Aura](https://openethereum.github.io/Pluggable-Consensus#Aura) (this is generally not so well documented or specified) PoA protocol developed by Parity Technologies and others. This is not compatible with the PoA protocol from geth (Clique).

We currently just don't support Aura, neither in `Common` nor in the upper layer libraries (e.g. block validator selection/voting in blockchain which we implemented for Clique e.g.) up to the Client.

If we now start to hotfix things for Kovan we just start tricking people into thinking that Kovan is supported (we should rather state more clearly that *it is not* in the documentation) while this is just not supported and will likely break on various occasions when Kovan specific mechanisms are triggered.

If we would want to support that would be a major dedicated effort, just some context for Andrew: we implemented Clique earlier this year and this was 3-4 weeks of work for 1-2 people working on the implementation. For Kovan we very likely not want to do this, the network is not very heavily used and will likely get either obsolete or at least very much less relevant after the PoS merge.

So: also no super strong opinion on this change here. From my side we can also leave but it might also make sense to just revert.

PR #1336, opened 2021-07-08
client: README updates - CLI parameters

acolytec3:
Updates readme to start adding in CLI parameters reference


packages/common/README.md
```diff
@@ -93,7 +93,7 @@ Supported chains:
 - `mainnet`
 - `ropsten`
 - `rinkeby`
-- `kovan`
+- `kovan` -- IMPORTANT NOTE: Kovan's PoA consensus mechanism is not implemented so attempts to sync the client on this network will likely fail
```
holgerd77, line 96: The `Common` library has nothing to do with the client (in the sense that it is a completely separate library) and we shouldn't add any upstream library references here, I guess that would count for all the other libraries as well. So such a note should rather be placed on the client README itself.

Here in `Common` should only be mentioned what is relevant in the `Common` library context (so I guess we don't need to mention here anything at all, first I thought that the `kovan` HF blocks wouldn't be added but had another look and even this is present.

The client [README](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/client) is actually the *one* README of our repos which is not really up to date and at some point we should substantially overhaul, I think all other README files are at least somewhat properly updated (which doesn't mean that there wouldn't be room for improvements as well).

For this particular case I would suggest that you add a new dedicated **Supported Chains** section relatively at the beginning of the client README and list all the chains we support and then you can list `kovan` as well together with a somewhat similar new note like the above.

holgerd77 (changes requested):
See comment

packages/common/README.md
acolytec3: Makes sense.  I've removed the comments from `common` and rebooted this as an update to the client readme with a focus on CLI parameters.

holgerd77 (approved):
LGTM

ryanio:
nice improvements 🎉

PR #1337, opened 2021-07-08
block: add london from-rpc test

ryanio:
This PR adds a test block from goerli in london to add coverage for the newly introduced param in #1330


holgerd77 (approved):
Nice, LGTM!

packages/block/test/from-rpc.spec.ts
```diff
@@ -73,4 +77,15 @@ tape('[fromRPC]: block #2924874', function (t) {
       st.end()
     }
   )
+
+  t.test('should create a block from london hardfork', function (st) {
+    const common = new Common({ chain: 'goerli', hardfork: 'london' })
+    const block = blockFromRpc(testDataFromRpcGoerliLondon, [], { common })
```
holgerd77, line 83: Never occurred to me before, but now when thinking about it I find it rather pretty obvious:
we very likey just want to include these `fromRpc()` methods (maybe in a somewhat reworked form) as additional static constructors to the main class? 🤔

Then they are not lying around here so much on the sideline and can also be better aligned and updated with our overall constructor format. And we'll likely want to do for tx as well I guess?

PR #1344, opened 2021-07-10
common: remove dao hardfork from chains when not applied

emersonmacro:
Per the comment here https://github.com/ethereumjs/ethereumjs-monorepo/pull/1329#issuecomment-872002349, this PR cleans up the "type 1" `null` HFs (i.e. the `dao` HF) when not applied to a particular chain.

Please note the change in `hardforks.spec.ts` still provides a valid test as long as `london` remains unimplemented for `kovan`. If and when `kovan` gets the London HF, that test will fail.


ryanio (approved):
lgtm!

PR #1345, opened 2021-07-12
tests: run hardhat e2e tests 

### LINKED ISSUE #1342: Monorepo: Run Hardhat E2E Tests
Hardhat now broke along two releases we have done, the latest on the `@ethereumjs/tx`[v3.3.0](https://github.com/ethereumjs/ethereumjs-monorepo/releases/tag/%40ethereumjs%2Ftx%403.3.0) release where there needed to be a hotfix which then was applied here https://github.com/nomiclabs/hardhat/pull/1641.

Both breakages would likely be able to have been prevented with running Hardhat E2E tests, so we should prioritize an integration. E2E tests have been first introduced by @cgewecke in #1134 on a selected third party target (https://github.com/NoahZinsmeister/as-you-permit) (note that the intial description of the PR is misleading since we first had a somewhat differing goal (run/test old EthereumJS library versions in the VM) and then switched over to the goal of testing third-party upstream libraries (like Hardhat) within the context of our libraries as they currently are on `master` respectively a new PR towards `master`).

This initial E2E target needed to be disabled (see #1284) since it wasn't reliable. We nevertheless can take this first work as a blueprint respectively the whole integration as a basis to integrate Hardhat E2E tests.

Depending on the CI run times added these kind of tests might to be run along a PR or alternatively also as nightly tests (like the Node tests, see [node-versions.yml](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/.github/workflows/node-versions.yml) as reference for an integration).

After having integrated Hardhat (or parts of the Hardhat tests) it might also make sense to expand to other libraries (e.g. Web3.js), just as some side note.

//cc @alcuadrado @fvictorio

acolytec3:
Will fix #1342 once it's done.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,41 @@
+name: E2E Hardhat Tests
+on:
+  workflow_dispatch
+env:
+  cwd: ${{github.workspace}}
+
+jobs:
+  virtual-npm-publish:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [15]
+    steps:
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
+      # Installs all dependencies, bootstrapping everything.
+      - run: npm install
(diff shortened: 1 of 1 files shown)


acolytec3:
Opened on wrong branch.  Will re-open with correct branch.

PR #1347, opened 2021-07-13
Update tests to 9.0.3

jochem-brouwer:
This PR updates tests to 9.0.3 and imports the `BerlinToLondonAt5` network upgrade, along with some changes which are necessary, since there is a consensus bug.


packages/block/src/block.ts
```diff
@@ -49,6 +49,7 @@ export class Block {
     // parse uncle headers
     const uncleHeaders = []
     for (const uhData of uhsData ?? []) {
+      header._common.setHardforkByBlockNumber(uhData.number!)
```
ryanio, line 52: would it be possible to use the block option `hardforkByBlockNumber` in the `BlockHeader.fromHeaderData` below instead?

jochem-brouwer: Oh that's a great suggestion! However, we should then explicitly revert back into `block.ts` to the old common, since it edits the common directly (and does not clone it).

jochem-brouwer: Hmm, I guess I should then cache the current fork before creating the uncle header, right? Otherwise it is always assumed that we use `hardforkByBlockNumber`?

holgerd77: Can we also have one or two tests in the block library itself on this change?

jochem-brouwer: Good point, I will write some tests, but at this point I think the implementation is not really consistent yet. I think it would make sense to check if block options has set `hardforkByBlockNumber` set explicitly to `false`, if that's the case then also don't set the fork on the uncles, because that's a bit inconsistent at this point.

jochem-brouwer:
Have added a rather big test, hope the explanation is clear. I think this is the way to go, it is non-breaking and helps with uncles where the uncle is pre-fork, but is included in a block after the fork.

packages/block/src/block.ts
ryanio: looks great!

packages/block/src/block.ts
```diff
@@ -49,15 +49,18 @@ export class Block {
     // parse uncle headers
     const uncleHeaders = []
     for (const uhData of uhsData ?? []) {
+      const cachedFork = header._common.hardfork()
       const uh = BlockHeader.fromHeaderData(uhData, {
-        ...opts,
+        hardforkByBlockNumber: true,
+        ...opts, // This potentially overwrites hardforkByBlocknumber
         // Use header common in case of hardforkByBlockNumber being activated
         common: header._common,
```
ryanio, line 57: if we do `common: header._common.copy()`, then could we skip needing to reset back to the cachedFork?

packages/block/src/block.ts
holgerd77: @jochem-brouwer ok, thanks for the additional context.

packages/block/src/block.ts
holgerd77: Hmm, `Common` is now copied on the structurally static blockchain objects (so: block and tx) anyhow on instantiation, see [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/header.ts#L203), so I guess this shouldn't make a difference? 🤔

jochem-brouwer: Oh that's funny, I did not notice it. I'll try removing it, to see if it works.

jochem-brouwer: Yep, this indeed works. Ready for re-review.

ryanio (approved):
looks great, and really nice test and diagram

Issue #702, opened 2019-12-09
Use Enums to define chain names, forks and other parameter values


evertonfraga:
I believe that there's some benefit to use Enums to define chain names and forks. I made some mistakes while writing `goerli` in the past, and I am probably not the first to make typos around chain and fork names.

GIven that `goerli` in German is originally `görlitzer`, that should be a valid source of confusion.

[![image](https://user-images.githubusercontent.com/47108/70463663-50682400-1a8b-11ea-9918-634d742d0365.png)](https://trends.google.com/trends/explore?geo=US&q=gorli,goerli)

Using [Enums](https://www.typescriptlang.org/docs/handbook/enums.html), developers also take advantage of code completion.

Rough example:

```js
const goerli = new common(Chains.Goerli)
```

holgerd77:
I still find this desirable and already thought a couple of times about this, @rumkin recently mentioned this again along a review on #937, which triggered these additional comments.

I have the following open questions here:

### A. Parameter Choice  

What parameters should get enum types? Candidates I would see:

1. chains (e.g. string 'mainnet')
2. hardforks (e.g. string 'istanbul')
3. eips (e.g. number `2537`) (yes/no? I have some tendency to do as well)
4. Consensus types (e.g. string 'pow')
5. Consensus algorithm (e.g. string 'ethash')

### B. Library Extendability

My biggest concern with enums e.g. for chains is that we might have the side effect that this restricts the extendability of the library, e.g. if someone wants to fork the library and would want to add an own `SOME_CHAIN.json` file or something, another use case worth to mention might be our own (living a bit on the sideline) `forCustomChain()` functionality.

If we use these enum types for parameter typing - so e.g. do in `CommonOpts` `hardfork?: Hardforks` (or whatever the naming is) we would prevent people from being able to use their custom hardfork names, do I get this right? Has anyone a suggestion for a pattern for this to circumvent? Should we just leave the typing "as is" and do the enum types as an optional convenience which we then use ourselves? 🤔

### C. API

What is a good API and/or integration choice for that so that import is easy and usage is simple and maybe don't produce to long function calls?

Some suggestion from my side, but rather just a first shot on this:

```typescript
class Common() {
  public static enum ConsensusType = {
    PoW = 'pow',
    PoA = 'poa'
  }
}
```

And the usage would be:

```typescript
Common.ConsensusType.PoW
```

[...]

PR #1363, opened 2021-07-18
Chain & HF enum usage

emersonmacro:
Second stage of work for #702. Updates usage of `common` package throughout monorepo to use new `Chain` and `Hardfork` enums.

My methodology was to use the global search term `new Common` and go package by package, so it's possible I may have missed some spots, but I tried to be as comprehensive as possible.

### DIFF
--- a/packages/block/README.md
+++ b/packages/block/README.md
@@ -63,8 +63,8 @@ To instantiate an EIP-1559 block the hardfork parameter on the `Common` instance
 
 ```typescript
 import { Block } from 'ethereumjs-block'
-import Common from '@ethereumjs/common'
-const common = new Common({ chain: 'mainnet', hardfork: 'london' })
+import Common, { Chain, Hardfork } from '@ethereumjs/common'
+const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.London })
 const block = Block.fromBlockData({
   header: {
     //...,
@@ -85,8 +85,8 @@ An Ethash/PoW block can be instantiated as follows:
 
 ```typescript
 import { Block } from '@ethereumjs/block'
-import Common from '@ethereumjs/common'
-const common = new Common({ chain: 'mainnet' })
+import Common, { Chain } from '@ethereumjs/common'
+const common = new Common({ chain: Chain.Mainnet })
 console.log(common.consensusType()) // 'pow'
 console.log(common.consensusAlgorithm()) // 'ethash'
 const block = Block.fromBlockData({}, { common })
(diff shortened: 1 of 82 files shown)


acolytec3:
This looks good.  I spotted 3 places where the `chain: 'chainName'` pattern is used in the tests and updated the obvious ones.  A question though:
If the chain is something other than a supported network as [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/bfd53752d8d589a38fc3fb41bb67f5f5acb74f9d/packages/client/test/sync/execution/vmexecution.spec.ts#L58) or [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/bfd53752d8d589a38fc3fb41bb67f5f5acb74f9d/packages/client/lib/util/debug.ts#L33), are we just continuing to support `string` as an acceptable type for the `chain` and `hardfork` parameters?

acolytec3 (commented):
This looks great though I'm a little concerned about the namespace collision between the new `Chain` enum in `@ethereumjs/common` and then the `Chain` class exported by `@ethereumjs/client`.  It's easy enough to hack around by just aliasing one of them as [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/84270a512ea3e7b689f4832b52be7de3deec97ca/packages/client/test/sync/execution/vmexecution.spec.ts#L10) but I wonder if it wouldn't be better to rename the enum to something like `ChainName` or something?  I'm not how often sure users would be instantiating a new `Common` instance with the client since we don't have a client package actually live on NPM yet or if client even exports `chain` but I could see this tripping people up down the line.  Thoughts?

ryanio:
> This looks great though I'm a little concerned about the namespace collision between the new `Chain` enum in `@ethereumjs/common` and then the `Chain` class exported by `@ethereumjs/client`. It's easy enough to hack around by just aliasing one of them as [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/84270a512ea3e7b689f4832b52be7de3deec97ca/packages/client/test/sync/execution/vmexecution.spec.ts#L10) but I wonder if it wouldn't be better to rename the enum to something like `ChainName` or something? I'm not how often sure users would be instantiating a new `Common` instance with the client since we don't have a client package actually live on NPM yet or if client even exports `chain` but I could see this tripping people up down the line. Thoughts?

thanks for contributing to the PR!

good point about the namespace collision, but you also make a good point that not many people will likely be developing in the client and be instantiating commons so I don't think it requires renaming the whole enum. I think having the simpler enum name of `Chain` is more useful to have. If importing both, one could also alias the client's `Chain` as `ClientChain` or something. Thinking more about it, I think the client's `Chain` class itself could maybe have a better name, since it almost seems like a nickname and doesn't help to clarify what it exactly is (my understanding is it's the client's blockchain helper class that interfaces with the underlying `@ethereumjs/blockchain`).

ryanio (approved):
thanks @emersonmacro, this looks great! 🎉

emersonmacro:
@ryanio thanks for updating and merging 🎉 

@acolytec3 thanks for taking a look as well, and definitely valid points. i kept `string` as an acceptable type based on the discussion here https://github.com/ethereumjs/ethereumjs-monorepo/issues/702#issuecomment-869589882, and i think it provides some nice flexibility. the namespace collision is unfortunate, but the more time i spent looking at usage in the codebase, the more i started leaning toward what Ryan said above. the `Chain` enum in common works pretty well, and i do think the `Chain` class in client could be a candidate for a better name (just not sure what yet).

Issue #1215, opened 2021-04-21
VM: Support EIP-3155 EVM trace specification


holgerd77:
There is a unified trace specification in the works with [EIP-3155](https://eips.ethereum.org/EIPS/eip-3155) which we should support.

holgerd77:
Once we have the debug logger from #1269 back in place we could actually relatively easily support this by adding a dedicated `trace` (or so, naming scheme has TBD) or so debug logger.

This would at least be some start and it could e.g. be used for our internal use cases but also by others. Not sure if this is sufficient and this should also be added as some kind of a) new event or b) as part of the [step](https://github.com/ethereumjs/ethereumjs-monorepo/blob/a459a0ebd8ae7f23a8d52c0fedef1111f5417ba8/packages/vm/lib/evm/interpreter.ts#L212) event or c) have the step event modified in a way that it get's compatible. 😄 

Thoughts?

ryanio:
@jochem-brouwer mentioned in https://github.com/ethereumjs/ethereumjs-monorepo/issues/1169#issuecomment-808744216 that he started thinking about how to implement this EIP, would be great to hear more info/details, I asked in a more recent comment further down on the issue (https://github.com/ethereumjs/ethereumjs-monorepo/issues/1169#issuecomment-859935662)

jochem-brouwer:
@ryanio here are some edge cases which need further discussion, as they are either not in the EIP, or from my point of view they are inconsistent; [magicians post.](https://ethereum-magicians.org/t/eip-3155-create-evm-trace-specification/5007/15?u=jochem-brouwer)

jochem-brouwer:
#1553

jochem-brouwer:
Since Geth has recently updated their tracing specs such that (as far as I can see) they now match the spec, this is something which we can start on. It indeed has no urgency. However, it would be very nice if we have matching traces - if all clients would match a certain trace specification and implement tools to re-execute bad blocks and get their traces, then it would be incredibly helpful to quickly figure out where the clients diverge and thus fix this quickly.

jochem-brouwer:
https://eips.ethereum.org/EIPS/eip-7756 will supersede this EIP

PR #1364, opened 2021-07-18
VM: report dynamic gas values in `fee` field of `step` event

### LINKED ISSUE #593: Dynamic gas value for special opcodes
At present, on event 'step', data for the opcode is getting fetched from the file statically [here](https://github.com/ethereumjs/ethereumjs-vm/blob/master/lib/evm/interpreter.ts#L136). but for some opcodes like `SSTORE`, gas calculation is done dynamically.

This would be great if opcode data can return dynamically calculated fee for such opcodes.

CC: @yann300

### LINKED ISSUE #1169: Dynamic gas costs are not exposed by the VM, but rather consumed during execution
(Moving a discussion here from discord)

While implementing `debug_traceTransaction` we realized that ethereumjs-vm handle gas in a very different way form Geth, so it's not really possible to implement this RPC method as Geth does it without reimplementing significant parts of gas-cost-related logic.

Each opcode in Geth has functions that return how much gas they cost, one for the constant part of the gas, and another one for the dynamic part. The dynamic one takes the entire state of the EVM, and returns the max amount of gas that the execution of that instruction can consume. With that, and extra function that returns the amount of memory that the instruction will use, you can use/consume the max gas that this step (execution + memory expansion) can use before executing.

A key thing to note is that only *CALL instructions don't use the max amount of gas, and that is handled in this way: the opcode handler is in charge of initializing a new child vm, with the max gas that it can spent, and run it. Once it returns, any leftover gas is returned to the parent vm.

Instead of doing this, ethereumjs-vm now consumes the gas while executing, so in order to know before execution how much an instruction can spend, we have to reimplementing its gas counting logic.

PS: We've already handled this by duplicating parts of the gas cost calculations that geth does, but ideally this should be provided by the vm.

jochem-brouwer:
This PR closes #1169, closes #593

Related: #1215

This PR internally refactors the EVM. Opcodes are now split into their behavior (stack edits, getting block hashes, etc.) and their gas cost. However, due to the complex behavior of gas cost they still edit some of the state of the VM, for instance, for memory expansions this is done into the gas calculation part.

The result is that the `fee` result of the `step` event now represents the actual cost of an operation, not just the base fee.

This PR is a breaking change; it changes the `fee` field of the `step` event from `number` to `BN`.

The `_runStepHook()` function signature is also edited, but since this is an underscore, I assume this was meant to be private, so that should be non-breaking.


jochem-brouwer:
This should pass at least Istanbul, MuirGlacier, Berlin and London. Running all tests to see if I missed something on earlier forks.

There are some TODOs left, but this is ready for early-review @ryanio @acolytec3

jochem-brouwer:
Just a note, this is not yet EIP3155-ready, this will happen in another PR.

jochem-brouwer:
@alcuadrado could you maybe check if you see some inconsistencies in the fee output? There are some cases not yet covered; one of them is reporting the call gas limit for any CALL opcode, and the invalid opcode now should report 0 fee, this is not consistent with Geth's output.

ryanio:
just getting started reviewing, looks really great jochem, feels very organized and logical to follow.

added a commit with some small improvements as I dig in further.

i'm taking a look at some of these missed coverage lines and wondering if these are already covered by some `ethereum/tests` that are just not run in our coverage suite, or if they're really not being tested at all then I would be interested to write some of our own tests for them, at least for the delicate parts like `opcodes/EIP1283.ts` with `netSstoreInitGas`, `netSstoreCleanGas`, and `netSstoreDirtyGas`

ryanio:
this comment is not relevant for this particular PR, but do you know the history of this todo? `TODO: Replace getContractStorage with EEI method`

I wouldn't mind fixing it along with typing the returned `found` object (which seems to be `value: Buffer | { current: Buffer, original: Buffer }`)

ryanio:
Rebased on master and force pushed

ryanio:
If we want to merge this in without worrying about the breaking change on the fee field of the `step` event, can we at the last moment in the `InterpreterStep` do `fee: fee.toNumber()`? Then we can change the type to BN as a breaking change item in #1024

packages/vm/src/evm/interpreter.ts
```diff
@@ -140,13 +141,29 @@ export default class Interpreter {
...
+    const gasLimitClone = this._eei.getGasLeft()
+
+    if (opInfo.dynamicGas) {
+      const dynamicGasHandler = dynamicGasHandlers.get(this._runState.opCode)!
+      // This function updates the gas BN in-place using `i*` methods
+      // It needs the base fee, for correct gas limit calculation for the CALL opcodes
+      await dynamicGasHandler(this._runState, gas)
+    }
+
+    // TODO: figure out if we should try/catch this (in case step event throws)
```
ryanio, line 157: do you have an example of when or how the step event would throw? it seems innocent enough with some debug and `emit`.

if it throws, in some extreme case, we would probably want the error to bubble up to the user, although we could wrap it more nicely.

jochem-brouwer: If we use a `number` instead of `BN` (so we convert the BN to a number) then it might throw if the BN does not fit in a number. What I actually meant with this TODO is to figure out what happens if any of the listeners throws. The previous behavior is that the tx then stops executing, so I guess we should keep that.

ryanio: > What I actually meant with this TODO is to figure out what happens if any of the listeners throws. The previous behavior is that the tx then stops executing, so I guess we should keep that.

ooh I see what you mean, if the consumer of the step event has a listener that throws? i thought you were maybe referring to it throwing within our own code, but it externally throwing also makes sense.

> If we use a number instead of BN (so we convert the BN to a number) then it might throw if the BN does not fit in a number.

yeah, in what kind of situation would the fee be `> 2^53 - 1` though? seems like a pretty extreme case. i only mention it because then we can merge this in without a breaking change and then update the number to BN in a future planned breaking changes release (which we can get started on as soon as london goes live on mainnet)

ryanio:
sorry for the incoming img dump, but continuing from https://github.com/ethereumjs/ethereumjs-monorepo/pull/1364#issuecomment-883587657, do you think we should add coverage for these missed lines? 

they seem like pretty delicate parts of the code. are these probably otherwise tested in `ethereum/tests` that our coverage suite isn't running?

Screen Shot 2021-07-21 at 12 08 40 PM

Screen Shot 2021-07-21 at 12 08 43 PM

Screen Shot 2021-07-21 at 12 08 46 PM

Screen Shot 2021-07-21 at 12 08 49 PM

Screen Shot 2021-07-21 at 12 08 59 PM

Screen Shot 2021-07-21 at 12 09 09 PM

Screen Shot 2021-07-21 at 12 09 11 PM

Screen Shot 2021-07-21 at 12 09 26 PM

Screen Shot 2021-07-21 at 12 09 42 PM

Screen Shot 2021-07-21 at 12 09 52 PM

Screen Shot 2021-07-21 at 12 10 09 PM

packages/vm/src/evm/interpreter.ts
jochem-brouwer: I have ran into the number being too large, I am assuming this was some ethereum/tests test. It could theoretically also happen if we implement the trace specification; then if we normally would forward "all" gas, then Geth would report this amount of gas which would get forwarded, so if we have blocks with extremely large gas limits (i.e. in ethereum/tests) then it would report a number which is too large.

jochem-brouwer:
These are probably not ran by coverage, since these are hardfork specific rules, and IIRC we only run a specific fork in the coverage?

jochem-brouwer:
I will write some tests for the uncovered lines as reported by codecov.

jochem-brouwer:
That's weird, we don't have API tests for EIP 1283. And for 2929 this is probably not triggered by codecov since that runs in Berlin, and codecov runs Istanbul (should we update the default fork to either Berlin or London?).

jochem-brouwer:
I will add a few more tests to increase coverage of currently uncovered lines.

jochem-brouwer:
I found some redundant code after some long research, here a quick explanation. 

Code in question is: (context is [CALLCODE](https://github.com/ethereumjs/ethereumjs-monorepo/blob/15f0a2d2c153e280a1487b5e0608143a2782021d/packages/vm/src/evm/opcodes/gas.ts#L358))

```
      if (gasLimit.gt(runState.eei.getGasLeft().isub(gas))) {
        trap(ERROR.OUT_OF_GAS)
      }
      if (gas.gt(runState.eei.getGasLeft())) {
        trap(ERROR.OUT_OF_GAS)
      }
```

The second check is redundant. Note that `gasLimit` is a stack value, so this is 0 or larger. If we have the case that `gas > eei.getGasLeft()`, then this means that `eei.getGasLeft().isub(gas)` is thus a negative value, so the first check will always return true, and we can thus remove the second check. (I will link the commit where I will remove this)

packages/vm/src/evm/opcodes/util.ts
```diff
@@ -231,30 +234,25 @@ export function writeCallOutput(runState: RunState, outOffset: BN, outLength: BN
...
-    runState.eei.useGas(
-      new BN(adjustSstoreGasEIP2929(runState, keyBuf, sstoreResetCost, 'reset', common)),
-      'updateSstoreGas'
-    )
+    const gas = new BN(common.param('gasPrices', 'sstoreReset'))
     runState.eei.refundGas(new BN(common.param('gasPrices', 'sstoreRefund')), 'updateSstoreGas')
-  } else if (value.length !== 0 && !currentStorage.length) {
-    runState.eei.useGas(new BN(common.param('gasPrices', 'sstoreSet')), 'updateSstoreGas')
+    return gas
+  } /*(value.length !== 0 && !currentStorage.length)*/ else {
```
jochem-brouwer, line 255: I commented this; it was an `else if` before. The situation reduces to either the value = 0, storage = 0 or value > 0, storage > 0 (first if statement), or value = 0, storage > 0, or (thus) value > 0, storage = 0. Thus the final statement can be reduced to an `else` statement, as all the other conditions are not true, thus this statement must be true.

ryanio: great, how about we add your reasoning above as a comment under the else statement so it's easy to follow?

jochem-brouwer:
Rebasing this one.

jochem-brouwer:
This was an extremely lengthy rebase process - lets hope tests pass.

ryanio:
looks great, I've been reviewing and contributing as you've been working on this but will do a final review and approval today :)

packages/vm/src/evm/interpreter.ts
ryanio: circling back to this TODO, do you have any fresh thoughts on this?

`// TODO: figure out if we should try/catch this (in case step event throws)`

i figure it would be harmless to add a try/catch as an extra precaution, can even add a test for it where  a listener on the `step` event throws and expect that it doesn't stop vm execution.

jochem-brouwer:
As mentioned on the call, I can make this non-breaking by not changing the fee type; in case that BN cannot be converted to fee, report a bogus value like `-1`. This cannot happen in practice due to block gas limit constraints.

ryanio:
just did a review, looks great to me! really nice organization and i find everything easy to follow. just adding a few commits with some tidying up, and will merge in master via commit as well (rebase here is starting to get a little scary :)

ryanio:
now that we are starting our new breaking changes on master, i don't mind if we merge this in with the field as `BN` if we feel like it is the better future-proof option.

also, one last TODO in the PR to address jochem if you want I can add this (https://github.com/ethereumjs/ethereumjs-monorepo/pull/1364#discussion_r683651720)

jochem-brouwer:
I will rebase here, the hardhat test is failing, this is due to the change of the emitted `InterpreterStep` where now gas is a BN and not a number.

I also checked the TODO: on master currently if the step event throws, then the entire call will also throw. This is the same behavior as is currently, so this behavior does not change.

After rebase I will change the `gas` event from number to BN and re-trigger the CI, I am pretty sure that there are some tests which now fail due to very large gas numbers, if that is not the case we can also consider using a number for now.

jochem-brouwer:
Rebased, but test:API:browser (VM) fails in CI but does pass locally :thinking:

jochem-brouwer:
Ok, this is now ready for merge, have also checked the rebase which for some reason went wrong.

holgerd77:
Ok, pretty cool! 😀 Please do not merge here yet, since we might want to do another VM bug fix release (and generally no pressure to merge early here).

jochem-brouwer:
Rebased this on master, `develop` branch now points to current master chain tip (866f865e792a613e9d417f67eecec36f06dc47a0). Please merge this PR (if CI passes) into develop so we can use this branch for breaking changes.

packages/vm/src/evm/interpreter.ts
holgerd77: Not sure if this has been addressed. Will nevertheless merge this in here, in doubt we can add separtely along a follow-up PR.

holgerd77 (approved):
Looks really great, again, really remarkable work! Totally love this now having all the gas logic together in one file and so transparently together, I would assume this will open up more use cases beyond the concrete (important) one addressed here.

Will merge to `develop`.

Issue #731, opened 2020-04-27
stateManager: add modifyAccountFields method


s1na:
Discussing the API with @jochem-brouwer we thought it might be easier from a developer PoV to have a `modifyAccountFields` that internally fetches an account from the trie, modify the given fields and then do a `putAccount`. This is to avoid users having to to the `getAccount` everytime they want to change an account. The method should probably throw if account doesn't exist.

holgerd77:
This is still open (I would also say still useful?) and a potentially good first issue for some trie-releated `StateManager`/`VM` work.

Please assign yourself if you want to take and also add a test case testing the new method.

emersonmacro:
I can try to tackle this

Issue #1226, opened 2021-04-26
VM: test VM.runBlock() ommer reward functionality


holgerd77:
Currently in the VM the ommer (uncle) reward functionality within `assignBlockRewards()` and `calculateOmmerReward()` is completely [untested](https://codecov.io/gh/ethereumjs/ethereumjs-monorepo/src/master/packages/vm/lib/runBlock.ts). This should be changed by adding one or more tests in `tests/api/runBlock.spec.ts`.

Some inspiration on how a block with uncles can be generated might eventually be taken from the `Block` library [tests](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/block/test). Also note that the accounts receiving the ommer rewards need to be added with a custom `StateManager` before running `VM.runBlock()` for this to work properly.

jochem-brouwer:
Why do the ommer accounts need to get added to `StateManager `manually?

holgerd77:
Because they are otherwise not present in the state and can't receive a reward? Or am I missing something here?

jochem-brouwer:
If you `getAccount` on a non-existing account, `StateManager` reports an empty account, which you can then edit :) So this is not a problem.

acolytec3:
I'm experimenting on this one.

PR #1369, opened 2021-07-21
VM: stateManager -> add modifyAccountFields method

emersonmacro:
Addresses #731.

This was my first time working with `stateManager` and I was learning as I went, so any feedback, suggestions, or critique are welcomed.

There was one part of the original issue that I wasn't able to figure out: `The method should probably throw if account doesn't exist.`. I attempted that but couldn't get it to work, and after some digging, it looks like if the account doesn't exist, a new one will be created, so an account always exists when `getAccount` is called. Is there something I'm not understanding?


acolytec3:
Per [previous comments](https://github.com/ethereumjs/ethereumjs-monorepo/issues/1226#issuecomment-826785596), `getAccount` creates a new account if it doesn't already exist in the `stateManager` as you noted so I think the question here is, should `getAccount` create a new account or not?  The docs currently state `Returns an empty account if the account does not exist` so that seems pretty clear to me that it shouldn't throw unless we're going to change the design, which would be a breaking change here since it would change the behavior of the function.  @ryanio @jochem-brouwer thoughts?

acolytec3 (dismissed):
This looks great.  I added one additional test to prove out the fact that modifying a non-existent account will first create an empty one and then update its values accordingly.

jochem-brouwer:
I don't know the reasoning behind throwing if the account does not exist. I'd say it makes much sense, and is consistent, to just take the empty account and then modify these fields.

jochem-brouwer (dismissed):
This looks good to me!

ryanio:
looks great. is there anywhere we can use this new method within our own internal VM then? or is it just a convenience helper for external use?

emersonmacro:
@ryanio Ah, yes I think there are some spots internally where this new method could be used. Should I include that as part of this PR or tackle separately?

ryanio:
I say let's include it as part of this PR, so that the changes are contained together nicely.

emersonmacro:
PR updated with some usage changes

ryanio (approved):
looks really great!

PR #1372, opened 2021-07-21
VM: fix EIP1559 bug to include tx value in balance check, fix nonce check

jochem-brouwer:
This PR fixes an [EIP1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md) bug, where we checked:

```
			# the signer must be able to afford the transaction
			assert signer.balance >= transaction.gas_limit * transaction.max_fee_per_gas
```

However, in the EIP, the call value is first deducted from the accounts' balance, and we should thus add this to the `transaction.gas_limit * transaction.max_fee_per_gas`.

This PR also fixes a bug where nonce values are not checked at runtime.

### DIFF
--- a/packages/vm/src/runTx.ts
+++ b/packages/vm/src/runTx.ts
@@ -284,14 +284,15 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
       // EIP-1559 spec:
       // The signer must be able to afford the transaction
       // `assert balance >= gas_limit * max_fee_per_gas`
-      const cost = tx.gasLimit.mul((tx as FeeMarketEIP1559Transaction).maxFeePerGas)
+      const cost = tx.gasLimit.mul((tx as FeeMarketEIP1559Transaction).maxFeePerGas).iadd(tx.value)
       if (balance.lt(cost)) {
         throw new Error(
           `sender doesn't have enough funds to send tx. The max cost is: ${cost} and the sender's account only has: ${balance}`
         )
       }
     }
-  } else if (!opts.skipNonce) {
+  }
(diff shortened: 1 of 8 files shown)


packages/vm/src/runTx.ts
```diff
@@ -284,14 +284,15 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
       // EIP-1559 spec:
       // The signer must be able to afford the transaction
       // `assert balance >= gas_limit * max_fee_per_gas`
-      const cost = tx.gasLimit.mul((tx as FeeMarketEIP1559Transaction).maxFeePerGas)
+      const cost = tx.gasLimit.mul((tx as FeeMarketEIP1559Transaction).maxFeePerGas).iadd(tx.value)
```
ryanio, line 287: can we just use `add` here instead of the in-place `iadd`? I want to make sure we don't mutate the original `tx.gasLimit` BN

jochem-brouwer: Sure! But this should return a copied BN though, since we `mul` and not `imul`.

ryanio: Ok sounds good I'll add a commit, yeah I figured but I think it's a better exercise of safety in case the line gets changed in the future. I'm sure it'd show up on some consensus tests though if it did get mutated 😄

ryanio: updated in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1372/commits/03dc6d304ba59ca8c698860eb54a73b0e2b06b3d

ryanio (approved):
lgtm!

Issue #1055, opened 2019-02-28
Trie: verifyProof() error on nested embeded nodes / batch() function UX issues


Lion17:
When you try to call the verifyProof function for a node that is embedded in another node, which in turn is also embedded, an error occurs.

To reproduce:

```
var Trie = require('merkle-patricia-tree')

trie = new Trie()
var ops = [
   { type: 'put', key: 'dog', value: 'puppy' }
 , { type: 'put', key: 'doge', value: 'coin' }
]

trie.batch(ops,
 () => {
  Trie.prove(trie, 'doge', function (err, prove) {
   if (err) console.log(err)
   else
    Trie.verifyProof(trie.root, 'doge', prove, function (err, value) {
     if (err) console.log(err)
     else console.log(value.toString())
    })
  })
 }
)
```

Output:

> node proof.js
> Error: Unexpected end of proof
>     at Object.exports.verifyProof (/home/local/tmp/node_modules/merkle-patricia-tree/proof.js:122:6)
>     at Function.verifyProof (/home/local/tmp/node_modules/merkle-patricia-tree/index.js:56:32)
>     at /home/local/tmp/proof.js:14:10
>     at /home/local/tmp/node_modules/merkle-patricia-tree/proof.js:33:5
>     at /home/local/tmp/node_modules/merkle-patricia-tree/baseTrie.js:520:18
>     at /home/local/tmp/node_modules/async/dist/async.js:473:16
>     at iteratorCallback (/home/local/tmp/node_modules/async/dist/async.js:1064:13)
>     at /home/local/tmp/node_modules/async/dist/async.js:969:16
>     at Object._return [as return] (/home/local/tmp/node_modules/merkle-patricia-tree/baseTrie.js:545:13)
>     at processNode (/home/local/tmp/node_modules/merkle-patricia-tree/baseTrie.js:325:34)

The reason is that only the variant with one level of embedded nodes is considered. This can be seen in file proof.js:82, where the key length is compared with 1, while with multiple nesting of nodes the key will be longer. In our case, it is equal to 2.

holgerd77:
@ryanio do you have got some insight if this might still be an issue? 🤔

ryanio:
I am not familiar with this issue yet but I will give the example scripts a try to see what could be going on and will report back.

[...]

PR #1373, opened 2021-07-21
trie: Better error checking for invalid proofs

### LINKED ISSUE #1368: verifyProof does not throw an error on invalid proofs
We are using the `BaseTrie` class to check the validity of Ethereum storage proofs and proof of non-existence.

According to the Readme, `verifyProof` [should be throwing an error](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/trie/docs/classes/basetrie.md#verifyproof) if `proof` is found to be invalid.

However:
- We are setting invalid values to the `proof` array but no error is thrown
- It is thrown is the root is the invalid value (expected)

By always receiving `null` as a response, we have no way to tell if it is because (1) the proof is valid and there is no such value or because (2) the proof is invalid and the value doesn't match anyway.

Steps to reproduce:
- Adding the following test case to `trie/test/index.spec.ts`

```typescript

tape('shall throw an error when a proof is invalid', async (t) => {
  const trie = new BaseTrie()
  await trie.put(Buffer.from('a'), Buffer.from('value1'))
  await trie.put(Buffer.from('aa'), Buffer.from('value2'))
  await trie.put(Buffer.from('aaa'), Buffer.from('value3'))

  const proof = await BaseTrie.createProof(trie, Buffer.from("a"))

  // Normal verification
  await BaseTrie.verifyProof(trie.root, Buffer.from('a'), proof)

  // Corrupting the proof
  proof[0].reverse()

  try {
    // @throws — If proof is found to be invalid.
    await BaseTrie.verifyProof(trie.root, Buffer.from('a'), proof)
    t.fail('should have thrown an error, but didn\'t')
  } catch (err) { }

  t.end()
})
```

Some testing for invalid roots should also be added. 
Thank you!

(thanks @ed255 for the discovery)

acolytec3:
Fixes #1368.

Adds a new check to the `baseTrie.verifyProof` function to verify that the proof trie for the provided key actually contains the value associated with that key and throws if the key doesn't exist in the trie (as described in the function's behavior).


acolytec3:
@ryanio I'm not sure if I'm handling this exactly right or not but the way that the documentation for `verifyProof` is written, it should [throw an error](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/trie/docs/classes/basetrie.md#verifyproof) if an invalid proof is provided for a given key.  Given my general reading on merkle proofs, is a proof considered invalid for a given trie and key if the proof Trie generated by the `verifyProof` function doesn't contain the key?  That's what our code is doing at present is just returning a null if the value isn't found and there isn't any additional verification done.

ryanio:
thanks for taking this on!

it looks like the library was designed to return null for these keys, check out these tests which are now failing with the new error (line 26):

https://github.com/ethereumjs/ethereumjs-monorepo/blob/7a2301c2fc00f3350634bcc7f4a50c41d1445094/packages/trie/test/proof.spec.ts#L22-L46

the very last equals is interesting, you would think as least a completely invalid proof would throw as that what the typedoc you linked to states, but that's also checked for null (line 45 above)

it is a bit perplexing, i'll see if I can get any thoughts or history/context from @s1na :)

s1na:
> is a proof considered invalid for a given trie and key if the proof Trie generated by the verifyProof function doesn't contain the key?

This would be a valid exclusion proof, i.e. a proof that a leaf in the tree is empty which has its own use-cases. 

I understand how it'd be useful to distinguish between a valid exclusion proof and an invalid proof and returning `null` for both is not ideal. But I thought `get()` would throw an error when it tried to read a key that doesn't exist in leveldb :-? if that doesnt happen then `Trie.fromProof` might need to be re-written to hash up the proof and check against the root.

acolytec3:
>  But I thought `get()` would throw an error when it tried to read a key that doesn't exist in leveldb :-? if that doesnt happen then `Trie.fromProof` might need to be re-written to hash up the proof and check against the root.

Now that I've soaked in this a little more, I think I'm starting to understand the logic a little better but I'm sure I'm still missing things.  Is this an accurate description of how `verifyProof` is expected to work.
1) Create a new proofTrie from just the roothash provided.  If the parameter `roothash` is invalid in the sense that it is a malformed buffer from which a trie can't be constructed, the `trie` constructor would throw so this is where we get the idea of an invalid proof in the documentation.
2) Pass that proofTrie and the proof to `fromProof` to get a new trie where any values in the original trie are updated based on the proof hash through a series of `puts`.  If this throws, we have invalid proof nodes though I'm not sure I see anywhere in the logic of `put` that it would ever throw.

Here's where I'm just not that deep on merkle proofs/tries but should you be able to construct a trie from any random set of bytes?  If so, what constitutes an invalid proof?  If we never generate a natural exception when constructing the trie from a proof, doesn't that mean that a prooftrie that doesn't contain the provided key just show evidence of non-existence of that key within the trie and nothing more?  What defines it as an invalid proof?

ed255:
> Here's where I'm just not that deep on merkle proofs/tries but should you be able to construct a trie from any random set of bytes? If so, what constitutes an invalid proof? If we never generate a natural exception when constructing the trie from a proof, doesn't that mean that a prooftrie that doesn't contain the provided key just show evidence of non-existence of that key within the trie and nothing more? What defines it as an invalid proof?

I think that `verifyProof` should be able to throw errors even if `fromProof` succeeds.  To me the definition of a valid proof for a given key, is a set of nodes which can be traversed either:
- following the complete key reaching a leaf (in which case the leaf value is returned)
- following a prefix of the key (which is shorter than the key) reaching a termination (in which case the null value is returned)

An invalid proof for a key would be a set of nodes which you can't traverse with the key or a prefix of the key reaching to a termination (because there are missing nodes).  In this case I think the `verifyProof` should throw an error.  This error can't be cought in `fromProof`, because you could have a valid proof for key `A`, and then verify it for key `B` (where `B` can't be proved with `proofOf(A)`); with the current implementation, if you try to verify (`B`, `proofOf(A)`) you'd get null, but you can't either prove if `B` belongs to the tree or not.

This is the behaviour that is implemented in go-ethereum: https://github.com/ethereum/go-ethereum/blob/a1f16bc74c7efb593db2982c92222d1e4a201c25/trie/proof.go#L103

ryanio:
thanks all for the discussion, here is the old (pre-refactored) [verifyProof](https://github.com/ethereumjs/merkle-patricia-tree/blob/v2.3.2/proof.js#L39) which does seem to have additional checks and errors

acolytec3:
@ryanio That looks like a straight port of geth's code if I'm reading it correctly. Maybe the best course of action is just to bring that back in?

s1na:
Thanks @ryanio for digging up the older version. That's essentially the usual algorithm for verifying a proof, i.e.:

- Start from the leaf we're interested in and its sibling. Hash them together
- Hash the result of that with the sibling one level higher
- Go till the root
- Compare root against expected root

Now this implemented version does this in a hacky way. It inserts all the proof nodes in the database, and then tries to `get` the leaf with its key and the expected root. This essentially verifies the proof because you do sth like this:

- Get the expected root via its hash from db. so at this point we know the correct hash of its children
- Traverse down via the key (one nibble at a time). If the node with the correct hash doesn't exist in db, it means it wasn't inserted as part of `Trie.fromProof`. So the proof is either missing something or has an invalid hash for one of the levels.
- If we can go all the way to the leaf (be it empty or full) and return that it means all of the proof nodes on the way have been correct

It might be possible to distinguish between invalid and valid exclusion proofs via a minor tweak where `get` returns an error instead of nil when asked for a node that doesnt exist. Not to say I have anything against bringing back the old version. Just wanted to provide some context.

acolytec3:
> Thanks @ryanio for digging up the older version. That's essentially the usual algorithm for verifying a proof, i.e.:
> 
> * Start from the leaf we're interested in and its sibling. Hash them together
> * Hash the result of that with the sibling one level higher
> * Go till the root
> * Compare root against expected root
> 
> Now this implemented version does this in a hacky way. It inserts all the proof nodes in the database, and then tries to `get` the leaf with its key and the expected root. This essentially verifies the proof because you do sth like this:
> 
> * Get the expected root via its hash from db. so at this point we know the correct hash of its children
> * Traverse down via the key (one nibble at a time). If the node with the correct hash doesn't exist in db, it means it wasn't inserted as part of `Trie.fromProof`. So the proof is either missing something or has an invalid hash for one of the levels.
> * If we can go all the way to the leaf (be it empty or full) and return that it means all of the proof nodes on the way have been correct
> 
> It might be possible to distinguish between invalid and valid exclusion proofs via a minor tweak where `get` returns an error instead of nil when asked for a node that doesnt exist. Not to say I have anything against bringing back the old version. Just wanted to provide some context.

Thanks @s1na.  I think I have an idea how to resolve leveraging your current implementation.  

I think if we insert a `throw` [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/008a4c3a044908614d2f0da22583537b7b383289/packages/trie/src/baseTrie.ts#L212) and maybe [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/008a4c3a044908614d2f0da22583537b7b383289/packages/trie/src/baseTrie.ts#L187) we will have either hit a branch node where the key/proof don't match or else reached an extension node where the key doesn't match but haven't found a leaf.  Either of these scenarios feels like we've hit an invalid proof.  Agree/disagree?

ryanio:
> I think if we insert a `throw` [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/008a4c3a044908614d2f0da22583537b7b383289/packages/trie/src/baseTrie.ts#L212) and maybe [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/008a4c3a044908614d2f0da22583537b7b383289/packages/trie/src/baseTrie.ts#L187) we will have either hit a branch node where the key/proof don't match. Wouldn't that land us at the end goal of identifying a proof that shows a value's non-existence in the tree?

that seems like it might also help solve https://github.com/ethereumjs/ethereumjs-monorepo/issues/1055#issuecomment-870818271

acolytec3:
At the end of the day, if we're going to continue to use the existing logic and not revert back to what @s1na noted as the conventional way for identifying an invalid proof, we have to put some thought into [onFound](https://github.com/ethereumjs/ethereumjs-monorepo/blob/a88e0643bfa7d85a9791dc0ea224d4db56620945/packages/trie/src/baseTrie.ts#L188-L188) that's buried inside the `asyng get` that is returning null to `verifyProof`.  There are several different conditions here and I'm still not clear on when not finding a node in `onFound` indicates a proof of non-existence versus an invalid proof.

The way I'm reading the code, a valid exclusion proof would result in a final state where the `keyRemainder.length` is 0 and the `node` value is null.  If at any point in traversing the trie we get to node with a `null` value but we still have a `keyRemainder`, that should constitute an invalid proof, right?

zmitton:
The use case for validating a proof, is when someone else sends you the proof, and you do not trust them. This software is meant to be run locally, to verify a key-value pair without trusting the source. This software does not have any method of creating an invalid proof, maybe such a function would be useful for understanding the tests. I wrote the tests above, and I agree they are very unclear, because I found that making an invalid proof (and testing many different cases) was somewhat difficult to do.

First I will say that I think the best/easiest solution might be to just implement the fix I described [here](https://github.com/ethereumjs/ethereumjs-monorepo/issues/1055#issuecomment-870818271).

Now let me try to describe how inserting the proof nodes into a tree and then `get`ing them back out, verifies the proof. The verification actually takes place in this `put` step [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/008a4c3a044908614d2f0da22583537b7b383289/packages/trie/src/baseTrie.ts#L637). The user's computer is calculating the keccak (as key) of every given proofnode, and placing them in a db. The traversal will start at the root (which the user already supplies). The root is looked up in the db, and the value found will have to be correct (its hash == the root), because all db keys were generated as described above. Here there will likely be a branch node, the branch is picked based on the key, and again this key is looked up in the db, where it must exist (for the proof to be valid) and whose data has already been verified during insert. If erroneous data is supplied if will be given a different key. Therefore, when the Trie `get` operation is performed, such nodes will be ignored. If a legitimate null proof is supplied, the tree will read and return the null value [from one of these cases](https://github.com/ethereumjs/ethereumjs-monorepo/blob/008a4c3a044908614d2f0da22583537b7b383289/packages/trie/src/baseTrie.ts#L186-L218). But if the trie tries to lookup the next db node's key, and that node does not exist in the db, that means that either the corresponding proofnode was not supplied (by the proover), or that the data for that node was tampered with, which would have caused it to have a different key then expected. Either way, the proof is invalid.

Hope this helps

acolytec3:
> The use case for validating a proof, is when someone else sends you the proof, and you do not trust them. This software is meant to be run locally, to verify a key-value pair without trusting the source. This software does not have any method of creating an invalid proof, maybe such a function would be useful for understanding the tests. I wrote the tests above, and I agree they are very unclear, because I found that making an invalid proof (and testing many different cases) was somewhat difficult to do.
> 
> First I will say that I think the best/easiest solution might be to just implement the fix I described [here](https://github.com/ethereumjs/ethereumjs-monorepo/issues/1055#issuecomment-870818271).
> 
> Now let me try to describe how inserting the proof nodes into a tree and then `get`ing them back out, verifies the proof. The verification actually takes place in this `put` step [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/008a4c3a044908614d2f0da22583537b7b383289/packages/trie/src/baseTrie.ts#L637). The user's computer is calculating the keccak (as key) of every given proofnode, and placing them in a db. The traversal will start at the root (which the user already supplies). The root is looked up in the db, and the value found will have to be correct (its hash == the root), because all db keys were generated as described above. Here there will likely be a branch node, the branch is picked based on the key, and again this key is looked up in the db, where it must exist (for the proof to be valid) and whose data has already been verified during insert. If erroneous data is supplied if will be given a different key. Therefore, when the Trie `get` operation is performed, such nodes will be ignored. If a legitimate null proof is supplied, the tree will read and return the null value [from one of these cases](https://github.com/ethereumjs/ethereumjs-monorepo/blob/008a4c3a044908614d2f0da22583537b7b383289/packages/trie/src/baseTrie.ts#L186-L218). But if the trie tries to lookup the next db node's key, and that node does not exist in the db, that means that either the corresponding proofnode was not supplied (by the proover), or that the data for that node was tampered with, which would have caused it to have a different key then expected. Either way, the proof is invalid.
> 
> Hope this helps

I think I follow your logic but I still have two challenges with it:

1) There doesn't seem to be a way to distinguish between a truly invalid proof and a proof of non-existence in the object returned by `onFound` since a couple of the valid proof cases return a `null` value for the `node` along with what I believe your'e saying the invalid proof case is.  If your logic is sound in determining between valid and invalid proofs, should we throw rather than [resolve](https://github.com/ethereumjs/ethereumjs-monorepo/blob/a88e0643bfa7d85a9791dc0ea224d4db56620945/packages/trie/src/baseTrie.ts#L226-L226) so as to distinguish between a valid proof of non-existence and an invalid proof?
2) I think I'm still missing why we're going this route to look for invalid proofs rather than just do what we did before (which is what geth does and would seem to be the accepted way for identifying an invalid proof and also distinguishing that from a proof of non-existence since a proof of non-existence would traverse the set of hashes in the proof and just end with a null value in the node at the end (if I'm following things correctly).  

I don't have a strong opinion one way or the other on this and I'm not an expert on merkle proofs (as demonstrated above) so just looking to make sure we account for this situation correctly.

brickpop:
I opened the issue that may have landed this PR. 

We need to verify a Storage Proof submitted by a user, and the reason we picked the library is because the `verifyProof` method says the following on the Readme:

https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/trie/docs/classes/basetrie.md#verifyproof

```
Verifies a proof
Throws If proof is found to be invalid
```

Why should `verifyProof` not fail if the given proof is wrong?

ryanio:
> Why should `verifyProof` not fail if the given proof is wrong?
> What's the point of a Storage Proof in general, if fake values cannot be validated?

@brickpop that was the existing behavior in mpt [v2.3.2](https://github.com/ethereumjs/merkle-patricia-tree/blob/v2.3.2/proof.js#L39), there was some refactoring that happened in v3+ that seems to have introduced this which we are trying to fix with this PR.

There seems to be two routes we can go here: one is to revert back to the v2.3.2 logic which matches geth's [implementation](https://github.com/ethereum/go-ethereum/blob/a1f16bc74c7efb593db2982c92222d1e4a201c25/trie/proof.go#L103), or resolve with the current design.

ryanio:
ok I'm working on this a bit late but I think I got the gist and introduced a new `get` param `throwWhenNotFound` to allow for more strict checking when verifying the proof (https://github.com/ethereumjs/ethereumjs-monorepo/pull/1373/commits/15e34bd11f868548b68be7a4e824b67389dfecc9). also rebased on master and added an entry to changelog. would appreciate some review.

acolytec3:
To make sure I track the logic, adding this new parameter to `findPath` allows us to specify when we should throw, and we should always throw in the `get` if no node is found by `findPath`.  All the other branching logic allows us to handle the valid proofs of non-existence as it were (per what @zmitton said above.  In other usages of `findPath`, we may not want to worry about it since we're not using it for proofs as much as deleting nodes if they exist, etc.

acolytec3:
@ryanio I updated the typedoc for `verifyProof` and changed the new parameter name to the one that makes more sense to me (but happy to be persuaded otherwise).  LGTM!

zmitton:
> There doesn't seem to be a way to distinguish between a truly invalid proof and a proof of non-existence in the object returned by onFound since a couple of the valid proof cases return a null value for the node along with what I believe your'e saying the invalid proof case is. 

There should not be any valid case where the *DB* returns null. The DB is only implementation detail. All nodes requested  from the DB should exists (any empty key in the tree will return null from [the following points](https://github.com/ethereumjs/ethereumjs-monorepo/blob/008a4c3a044908614d2f0da22583537b7b383289/packages/trie/src/baseTrie.ts#L186-L218) specifically lines 189, 195, 208, or 214).

The problem with @ryanio's method, is that it proabely cant differentiate between an valid/invalid proof of exclusion (null). 

Im not sure how else to succinctly describe why this works other than asking yourself to take a deep dive into [how](https://eth.wiki/en/fundamentals/patricia-tree#example-trie) the merkle-patricia-trie stores its data.

ryanio:
@zmitton we do have a proof of exclusion test and it seems to be passing fine with my suggested changes:

[proof.spec.ts:L33-37](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/trie/test/proof.spec.ts#L33-L37):

```typescript
    myKey = Buffer.from('anothergarbagekey') // should generate a valid proof of null
    proof = await CheckpointTrie.createProof(trie, myKey)
    proof.push(Buffer.from('123456')) // extra nodes are just ignored
    val = await CheckpointTrie.verifyProof(trie.root, myKey, proof)
    t.equal(val, null, 'Expected value to be null')
```

s1na:
Sorry I haven't read the whole conversation and might be repeating someone else, but I think #1368 is not an issue. I.e. it is possible to distinguish between a valid exclusion proof and an invalid proof. You only need to check the proof value against the value returned by `verifyProof`. e.g.:

- if leaf value in proof is `0xabac` and verify returns `null` then invalid
- leaf value is `null` (exclusion) and verify returns `null` then valid

zmitton:
Right. How to get an invalid exclusion proof

ryanio:
would this be a valid test case to test an invalid exclusion proof?

```typescript
    // test an invalid exclusion proof by creating
    // a valid exclusion proof then making it non-null
    myKey = Buffer.from('anyrandomkey')
    proof = await CheckpointTrie.createProof(trie, myKey)
    val = await CheckpointTrie.verifyProof(trie.root, myKey, proof)
    t.equal(val, null, 'Expected value to be null')
    // now make the key non-null so the exclusion proof becomes invalid
    await trie.put(myKey, Buffer.from('thisisavalue'))
    try {
      await CheckpointTrie.verifyProof(trie.root, myKey, proof)
      t.fail('should have thrown an error')
    } catch (err) {
      if (err.message.includes('Path not found')) {
        t.pass('threw correct error on invalid proof')
      } else {
        t.fail('did not throw correct error')
      }
    }
```

brickpop:
In case it might help, this is the test case that I suggested on the original issue:

```ts
tape('shall throw an error when a proof is invalid', async (t) => {
  const trie = new BaseTrie()
  await trie.put(Buffer.from('a'), Buffer.from('value1'))
  await trie.put(Buffer.from('aa'), Buffer.from('value2'))
  await trie.put(Buffer.from('aaa'), Buffer.from('value3'))

  const proof = await BaseTrie.createProof(trie, Buffer.from("a"))

  // Normal verification
  await BaseTrie.verifyProof(trie.root, Buffer.from('a'), proof)

  // Corrupting the proof
  proof[0].reverse()

  try {
    // @throws — If proof is found to be invalid.
    await BaseTrie.verifyProof(trie.root, Buffer.from('a'), proof)
    t.fail('should have thrown an error, but didn\'t')
  } catch (err) { }

  t.end()
})
```

ryanio:
@brickpop yes thanks so much, have added it [here](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1373/files#diff-31717aedba209c646988550ff9a91ee518d515d3693d568651dd28f7681c0683R57)

zmitton:
> would this be a valid test case to test an invalid exclusion proof?

Yes that looks like the right idea to me (It may not be sufficient to test all edge cases).

ryanio:
@zmitton great to hear, thanks. would love to come up with some more ideas for cases to add.

zmitton:
From the example ethereum wiki key-values of 
```
<64 6f> : 'verb'
<64 6f 67> : 'puppy'
<64 6f 67 65> : 'coin'
<68 6f 72 73 65> : 'stallion'
``` 
will produce the following DB nodes:
```
rootHash: [ <16>, hashA ]
hashA:    [ <>, <>, <>, <>, hashB, <>, <>, <>, [ <20 6f 72 73 65>, 'stallion' ], <>, <>, <>, <>, <>, <>, <>, <> ]
hashB:    [ <00 6f>, hashD ]
hashD:    [ <>, <>, <>, <>, <>, <>, hashE, <>, <>, <>, <>, <>, <>, <>, <>, <>, 'verb' ]
hashE:    [ <17>, [ <>, <>, <>, <>, <>, <>, [ <35>, 'coin' ], <>, <>, <>, <>, <>, <>, <>, <>, <>, 'puppy' ] ]
```
A legitimate `null` value exists at any key unmentioned above. You can just make them up. For instance the value at key `<63>` can be found by looking up the `rootHash`,  which matches the `6`, and points to `hashA`. In hashA the 3rd value of the branch is `<>` which indicates `null` (or any keys that even begin with `<63>` for that matter). The exclusion proof for this key would consist of those 2 nodes that we had to traverse to find that out.

But there are also other ways to arrive at `null`. For instance trying to lookup `<68 6f 00>` would take you to the extension node `[ <20 6f 72 73 65>, 'stallion' ]`, and since the extension does not correctly match, this would indicate that the only key beginning with `<68 6f>` is followed by `<72 73 65>` so `<68 6f 00>` is `null`. Its proof would consist of the same 2 nodes. There are other cases for null, like having a leaf node of `null` ...etc.

The method of proof verification by puting the proofnodes into a trie and then looking them up is the best way to do it because it does not require you to rebuild all of this logic. Just throw the error [when a node is missing](https://github.com/ethereumjs/ethereumjs-monorepo/issues/1055#issuecomment-870818271)

This probably does not need as many tests because the trie logic is very well used at this point. If you try to rebuild the logic, then I would recommend more tests by building very specific shaped trie's for each case.

zmitton:
Diving deeper I found that my code was merged in b8f612b0c206b15b2ec282d0034209153c69d392 and then the relevant code (`BaseTrie._lookupNode`) was broken by 2e4c40e1851fe038229c39e74d367168e02db002. Im not really sure why or how come it didn't break any tests

acolytec3:
@ryanio Anything outstanding on this one?  I'm waiting on the tests to re-run from merging in master but seems like this one might be ready to go or is it considered a breaking change since we added that new optional parameter to the `get` method?

ryanio (dismissed):
@acolytec3 thanks, LGTM, really love the readme updates for explaining proof of non-existence.

I don't believe it's breaking because it's an optional parameter that defaults to the prior behavior. It is still a little strange to me to add to the `Trie.get` param though, but it was the only way I could think of. I would be open to other ideas.

I wonder if we should consider a bug fix release with this if invalid proofs are indeed totally broken?

ryanio:
Let's also wait for an approval or any thoughts from Holger before merging in.

holgerd77:
Might take me some more days until I worked through this here and can do some qualified comment.

gabrocheleau (dismissed):
Looks good, left a few minor comments

packages/trie/src/baseTrie.ts
```diff
@@ -102,8 +102,16 @@ export class Trie {
    * Checks if a given root exists.
    */
   async checkRoot(root: Buffer): Promise<boolean> {
-    const value = await this._lookupNode(root)
-    return !!value
+    try {
+      const value = await this._lookupNode(root)
+      return !!value
```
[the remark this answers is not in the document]
holgerd77: I am bit fan of corrections like this one. We had various cases in tx where we ran into substantial bugs with a similar pattern used.

packages/trie/package.json
```diff
@@ -29,6 +29,7 @@
     "lint": "../../config/cli/lint.sh",
     "lint:fix": "../../config/cli/lint-fix.sh",
     "tsc": "../../config/cli/ts-compile.sh",
+    "tape": "tape -r ts-node/register",
```
holgerd77, line 32: Already asked this in #1399, where is this new `tape` command coming from and why is this necessary?

packages/trie/src/baseTrie.ts
```diff
@@ -270,9 +288,10 @@ export class Trie {
     let value = null
     let foundNode = null
     value = await this.db.get(node as Buffer)
-
     if (value) {
       foundNode = decodeNode(value)
+    } else {
+      throw new Error('Missing node in DB')
```
holgerd77, line 294: Maybe we want to add a short dev comment here that this is a critical respectively sensitive error message where the text string is needed to remain stable for other core functionality to work.

holgerd77 (approved):
LGTM, nice work, thanks everyone for this extensive and clarifying discussion!

Will merge and relatively soon prepare a bugfix release on this.

holgerd77:
Changes from this PR have been published along `merkle-patricia-tree` [v4.2.1](https://github.com/ethereumjs/ethereumjs-monorepo/releases/tag/merkle-patricia-tree%404.2.1).

PR #1374, opened 2021-07-22
VM: add tests for wrong transactions

jochem-brouwer:
We are not running all checks which are expected in `ethereum/tests`; if there is a `transactionSequence` field, try to execute the transactions on top of the last header block. This would have caught our nonce check problem in the transactions (see #1372).

### DIFF
--- a/packages/vm/tests/BlockchainTestsRunner.ts
+++ b/packages/vm/tests/BlockchainTestsRunner.ts
@@ -126,10 +126,10 @@ export default async function runBlockchainTest(options: any, testData: any, t: 
           blockOpts: { calcDifficultyFromHeader: parentBlock.header, freeze: false },
         })
         for (const txData of raw.transactionSequence) {
-          const txRLP = Buffer.from(txData.rawBytes.slice(2), 'hex')
-          const tx = TransactionFactory.fromSerializedData(txRLP)
           const shouldFail = txData.valid == 'false'
           try {
+            const txRLP = Buffer.from(txData.rawBytes.slice(2), 'hex')
+            const tx = TransactionFactory.fromSerializedData(txRLP)
             await blockBuilder.addTransaction(tx)
             if (shouldFail) {
               t.fail('tx should fail, but did not fail')
@@ -142,7 +142,12 @@ export default async function runBlockchainTest(options: any, testData: any, t: 
             }
           }
         }
-        await vm.stateManager.revert()
+        // Have to try/catch this revert, in some cases trie is not checkpointed and we should thus not revert changes
+        try {
+          await vm.stateManager.revert()
+          // eslint-disable-next-line no-empty
+        } finally {
+        }
       }
 
       const blockRlp = Buffer.from(raw.rlp.slice(2), 'hex')
--- a/packages/vm/tests/BlockchainTestsRunner.ts
+++ b/packages/vm/tests/BlockchainTestsRunner.ts
@@ -122,8 +122,7 @@ export default async function runBlockchainTest(options: any, testData: any, t: 
         const parentBlock = await vm.blockchain.getIteratorHead()
         const blockBuilder = await vm.buildBlock({
           parentBlock,
-          headerData: { coinbase: '0x96dc73c8b5969608c77375f085949744b5177660' },
-          blockOpts: { calcDifficultyFromHeader: parentBlock.header, freeze: false },
+          blockOpts: { calcDifficultyFromHeader: parentBlock.header },
         })
         for (const txData of raw.transactionSequence) {
           const shouldFail = txData.valid == 'false'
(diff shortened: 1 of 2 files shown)


packages/vm/tests/BlockchainTestsRunner.ts
```diff
@@ -5,6 +5,7 @@ import { Block } from '@ethereumjs/block'
 import Blockchain from '@ethereumjs/blockchain'
 import { setupPreConditions, verifyPostConditions } from './util'
 import Common from '@ethereumjs/common'
+import { TransactionFactory } from '../../tx/dist'
```
ryanio, line 8: small nit, I'm sure VS Code automatically did this import hehe
```suggestion
import { TransactionFactory } from '@ethereumjs/tx'
```

packages/vm/tests/BlockchainTestsRunner.ts
```diff
@@ -117,6 +118,33 @@ export default async function runBlockchainTest(options: any, testData: any, t:
       // Update common HF
       common.setHardforkByBlockNumber(currentBlock.toNumber())
 
+      if (raw.transactionSequence) {
+        const parentBlock = await vm.blockchain.getIteratorHead()
+        const blockBuilder = await vm.buildBlock({
```
jochem-brouwer, line 123: What do we think of this? Ethereum/tests should test our VM, so we'd actually would like to use runBlock, but for these cases we use buildBlock... I don't have any idea how to run it with runBlock though, since we can't produce invalid blocks.

ryanio: seems reasonable, `buildBlock` uses `runTx` under the hood

packages/vm/tests/BlockchainTestsRunner.ts
```diff
@@ -119,14 +119,18 @@ export default async function runBlockchainTest(options: any, testData: any, t:
       common.setHardforkByBlockNumber(currentBlock.toNumber())
 
       if (raw.transactionSequence) {
+        const parentBlock = await vm.blockchain.getIteratorHead()
+        const blockBuilder = await vm.buildBlock({
+          parentBlock,
+          headerData: { coinbase: '0x96dc73c8b5969608c77375f085949744b5177660' },
```
ryanio, line 125: i don't think we need `{ headerData: coinbase }` or `{ blockOpts: freeze }` so we can simplify by removing those opts

packages/vm/tests/BlockchainTestsRunner.ts
```diff
@@ -119,14 +119,18 @@ export default async function runBlockchainTest(options: any, testData: any, t:
...
+          blockOpts: { calcDifficultyFromHeader: parentBlock.header, freeze: false },
+        })
         for (const txData of raw.transactionSequence) {
           const txRLP = Buffer.from(txData.rawBytes.slice(2), 'hex')
           const tx = TransactionFactory.fromSerializedData(txRLP)
           const shouldFail = txData.valid == 'false'
```
ryanio, line 131: looks like the `blockchain:london` tests are having some `expectException: TR_TipGtFeeCap` failures here, so I am wondering if `shouldFail` should also take `expectException` into account

ryanio: e.g. [here](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1374/checks?check_run_id=3137135710#step:7:57872):
Screen Shot 2021-07-22 at 1 15 38 PM
```
expectException:
        ">=London": TR_TipGtFeeCap
```

packages/vm/tests/BlockchainTestsRunner.ts
jochem-brouwer: Removed it, I indeed copied it from a buildBlock test! :smile:

packages/vm/tests/BlockchainTestsRunner.ts
jochem-brouwer: That should be a new PR. We indeed have some tests which pass, but they only pass because they throw on something else, and not for the expected reason. However this means we have to inspect each error string and figure out what that means. But on the other hand, we want to ensure that the tests fail for the right reason. This error is fixed now.

jochem-brouwer:
Tests should now pass, also integrated your comments @ryanio. Ready for review.

packages/vm/tests/BlockchainTestsRunner.ts
ryanio: hm i'm still seeing it

ryanio: i'll remove it

ryanio:
this was still [failing](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1374/checks?check_run_id=3138699571#step:7:31263):

```
# file: tipTooHigh test: tipTooHigh_d0g0v0_London
ok 23644 correct genesis RLP
ok 23645 correct pre stateRoot
ok 23646 tx succesfully failed
Error: trying to revert when not checkpointed
```

I tested locally using blockBuilder's revert since it will only run if it is checkpointed, and that seemed to fix. I'll add the commit now.

ryanio (approved):
lgtm!

PR #1376, opened 2021-07-26
trie: remove use of deprecated setRoot

acolytec3:
Small change to remove internal use of deprecated `setRoot` method in `trie` package

### DIFF
--- a/packages/trie/src/baseTrie.ts
+++ b/packages/trie/src/baseTrie.ts
@@ -65,13 +65,18 @@ export class Trie {
     this._deleteFromDB = deleteFromDB
 
     if (root) {
-      this.setRoot(root)
+      this.root = root
     }
   }
 
   /** Sets the current root of the `trie` */
   set root(value: Buffer) {
-    this.setRoot(value)
+    /* eslint-disable-next-line @typescript-eslint/no-unnecessary-condition */
+    if (!value) {
+      value = this.EMPTY_TRIE_ROOT
+    }
+    assert(value.length === 32, 'Invalid root length. Roots are 32 bytes')
+    this._root = value
   }
 
   /** Gets the current root of the `trie` */
--- a/packages/trie/src/baseTrie.ts
+++ b/packages/trie/src/baseTrie.ts
@@ -69,13 +69,21 @@ export class Trie {
     }
   }
 
-  /** Sets the current root of the `trie` */
+  /**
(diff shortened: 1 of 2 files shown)


acolytec3:
While working in the `baseTrie` file, I noticed we were still using a deprecated `setRoot` method in some of our internal logic.  I've updated the `root` setter to use the correct logic from `setRoot` and adjusted our internal usage accordingly so we're no longer using a deprecated method.

packages/trie/src/baseTrie.ts
```diff
@@ -65,13 +65,18 @@ export class Trie {
...
     if (root) {
-      this.setRoot(root)
+      this.root = root
     }
   }
 
   /** Sets the current root of the `trie` */
   set root(value: Buffer) {
-    this.setRoot(value)
+    /* eslint-disable-next-line @typescript-eslint/no-unnecessary-condition */
```
ryanio, line 74: hm can you make `value` param optional with `value?: Buffer` so that it's not `no-unnecessary-condition`? (not sure if that's possible with the setter). another option would be to do `set root(value = this.empty_TRIE_ROOT)` and then no need for the `if (!value)` (I think I'd prefer this).

overall I like that you brought the `setRoot` logic here though, to dedupe the code then can we clear the code in `setRoot` and simply do `this.root = value`? I tend to like to use the new methods in the deprecated methods so they are easy to follow (and remove) in the future, and also if we update the set code we only have to do it in one place.

acolytec3: > hm can you make `value` param optional with `value?: Buffer` so that it's not `no-unnecessary-condition`? (not sure if that's possible with the setter). another option would be to do `set root(value = this.empty_TRIE_ROOT)` and then no need for the `if (!value)` (I think I'd prefer this).
Typescript doesn't allow optional parameters in setters so I'm changing it to a union type with `undefined` to get around that and will update it my next commit.  Does that work?

ryanio: ah ok thanks, yeah let me add a follow up commit and I think it will be great. I will also add an entry to changelog.

ryanio: you know what, I changed my mind, I think your proposed solution was ultimately best, I'll change it back to ignoring the lint rule.

ryanio (approved):
lgtm!

PR #1377, opened 2021-07-26
Client: Small types clean-up

acolytec3:
More clean-up items I ran across as I've been digging into the guts of the fetching process in the client.

### DIFF
--- a/packages/client/lib/sync/fetcher/blockfetcher.ts
+++ b/packages/client/lib/sync/fetcher/blockfetcher.ts
@@ -106,7 +106,7 @@ export class BlockFetcher extends BlockFetcherBase<Block[], Block> {
    * @return {Peer}
    */
   // TODO: find out what _job is supposed to be doing here...
-  peer(): Peer {
-    return this.pool.idle((p: any) => p.eth)
+  peer(): Peer | undefined {
+    return this.pool.idle((p: Peer) => 'eth' in p)
   }
 }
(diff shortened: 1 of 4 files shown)


packages/client/lib/sync/fetcher/blockfetcher.ts
```diff
@@ -106,7 +106,7 @@ export class BlockFetcher extends BlockFetcherBase<Block[], Block> {
    * @return {Peer}
    */
   // TODO: find out what _job is supposed to be doing here...
-  peer(): Peer {
-    return this.pool.idle((p: any) => p.eth)
```
ryanio, line 110: hm is this perhaps checking something different with `p.eth` checking if it has the `eth` protocol methods bound to the peer? curious what your reasoning was to update to `instanceof Peer`

ryanio:
nice, love it :) 🎉

packages/client/lib/sync/fetcher/blockfetcher.ts
acolytec3: Just not me thinking it through.  I've changed this to a `hasOwnProperty` check since the filtering function wants a boolean return value which is what was tripping Typescript up.

ryanio: Ooh we can do one of Holger's favorites, `'eth' in peer`

acolytec3: Done.  We'll see if that makes Typescript happy.  I noticed that the CI linter seems to be applying rules that aren't applied on local as I didn't see that error about `no-use-builtins` or whatever it was.

ryanio (approved):
lgtm!

PR #1378, opened 2021-07-26
eslint: rule clean up and consolidation

ryanio:
This PR does some eslint rule cleanup, bringing common rule configurations up to the root config for more consistent behavior across packages.

This PR:
* Configures `no-unused-vars` in the root eslint
* Clarifies `chainParams `todo comment
* Turns off `no-unnecessary-condition` in root config
* Turns off `no-dupe-class-members` in root config
* Cleans up unneeded disables

### DIFF
--- a/config/eslint.js
+++ b/config/eslint.js
@@ -32,6 +32,7 @@ module.exports = {
     'no-unused-vars': 'off',
     '@typescript-eslint/no-unused-vars': ["error", { argsIgnorePattern: "^_", varsIgnorePattern: "^_" }],
     '@typescript-eslint/no-unnecessary-condition': 'off',
+    'no-dupe-class-members': 'off',
     'prettier/prettier': 'error',
   },
   parserOptions: {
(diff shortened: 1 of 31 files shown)


ryanio:
This is soo nice and easy to do now that we have our root config in a local folder as opposed to a separate npm package :)

acolytec3 (approved):
Looks good overall.  There's one place in client [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/bfd53752d8d589a38fc3fb41bb67f5f5acb74f9d/packages/client/lib/types.ts#L56-L56) where I'm not sure how to make Typescript happy.  This is a generic declaration so maybe we just leave it and move on?:

ryanio:
> Looks good overall.  There's one place in client [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/bfd53752d8d589a38fc3fb41bb67f5f5acb74f9d/packages/client/lib/types.ts#L56-L56) where I'm not sure how to make Typescript happy.  This is a generic declaration so maybe we just leave it and move on?:  

Perfect thanks, yeah I think we can just keep that one since it's a bit of a special case.

PR #1380, opened 2021-07-29
Common, Blockchain: add browser tests

### LINKED ISSUE #1338: Common/Blockchain: add browser tests
There are actually no Common browser tests. Karma is not integrated and there are no respective npm script commands like `npm run test:browser` and the like.

This should be added following the structure of the other libraries like e.g. the [browser test integration](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/block) in the block library (just as some example which makes some consistent impression on first look).

emersonmacro:
Closes #1338


emersonmacro:
@ryanio I took a stab at #1338. Build processes are definitely not my specialty - I got a setup working in `common`, but a similar setup is failing for `blockchain`. The files are there and it's entering the first test, but it gets stuck at this line and times out https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/blockchain/test/clique.spec.ts#L13. Wondering if either A) you might have an idea of why it's not working for the `blockchain` tests or B) maybe we just add browser tests for `common` now and handle `blockchain` separately. Or I guess C) I keep digging on why it's not working for `blockchain`. Any thoughts?

Also, I got a linting error on code I didn't touch - it was complaining that `Partial` was not defined. Have you seen that before? I added an ignore comment just so I could get this WIP PR up, but wouldn't want to merge that if there's a real fix available. https://github.com/ethereumjs/ethereumjs-monorepo/pull/1380/files#diff-970bb4d849f8275a7e4a4d17f71211e3a99fac7e27a9f3c0b826746287135ddbR169

ryanio:
@emersonmacro thanks for getting started on this!

i am going to switch from `browserify` to `karma-typescript` since that is a more modern setup that doesn't require a build step. i will also convert `block` and `tx` from `browserify` to `karma-typescript`.

seems to all be passing for me locally, let's see how ci does.

ryanio:
> Also, I got a linting error on code I didn't touch - it was complaining that `Partial` was not defined. Have you seen that before? I added an ignore comment just so I could get this WIP PR up, but wouldn't want to merge that if there's a real fix available.

haven't seen that before, and didn't get any errors locally when i removed the newly added ignores for `Partial`, i will see if it catches on ci though.

ryanio (approved):
lgtm!

emersonmacro:
@ryanio thanks!

PR #1382, opened 2021-07-29
tx: fix getMessageToSign return type for EIP1559Transaction

### LINKED ISSUE #1381: FeeMarketEIP1559Transaction#getMessageToSign has incorrect signature
[FeeMarketEIP1559Transaction#getMessageToSign](https://github.com/ethereumjs/ethereumjs-monorepo/blob/8d140c3ba11a5f9824c7a0dd5952eea8e80f9a75/packages/tx/src/eip1559Transaction.ts#L309) is marked as returning Buffer | Buffer[]. However, it only returns a Buffer.

If hashMessage == true, it returns keccak256(Buffer), which is a buffer.
if hashMessage == false, it returns a Buffer.

ryanio:
Fixes #1381


emersonmacro (approved):
Looks good to me

ryanio:
thanks!

PR #1385, opened 2021-07-31
return value of type never|BNLike is assignable to a variable of type BNLike

Leon-Africa:
The set of never is an empty set.

The empty set is a subset of every set .viz Given a set U, the union of U and the empty set is just U.

Given that no values are assignable to a variable of type never saying:

````
 gasPrice?: never 
````

Can result in bugs/miscellaneous limitations/errors.  Example: [gasprice](https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-core/src/internal/core/providers/accounts.ts#:~:text=gasPrice%3A%20undefined%2C) 

Hence suggesting
````
 gasPrice?: never | BNLike
````

Which gives gasPrice assignability to a variable of type BNLike

### DIFF
--- a/packages/tx/src/types.ts
+++ b/packages/tx/src/types.ts
@@ -184,7 +184,7 @@ export interface FeeMarketEIP1559TxData extends AccessListEIP2930TxData {
   /**
    * The transaction's gas price.
    */
-  gasPrice?: never
+  gasPrice?: never | BNLike
   /**
    * The maximum inclusion fee per gas (this fee is given to the miner)
    */


acolytec3:
Thanks for this PR! The issue here is that ``EIP1559` transactions don't use this `gasPrice` field so should never actually have a value and therefore is correctly typed as `never` in our current implementation so we won't be merging this PR.  This `gasPrice` field is inherited from the `txData` type for legacy transactions but is replaced by [maxFeePerGas](https://github.com/ethereumjs/ethereumjs-monorepo/blob/49877f4d6b0ffb8438d4b4101299a2e10d576f43/packages/tx/src/types.ts#L195) for EIP1559 transactions.  We will update our documentation to clarify this.

Leon-Africa:
@acolytec3 Thanks for the clarity.

Anyone else coming here can have a read @:
[eip-1559.md](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md)
[eip-1559-faq](https://notes.ethereum.org/@vbuterin/eip-1559-faq)

PR #1391, opened 2021-08-04
ci: fix hardhat e2e resolutions

ryanio:
The hardhat e2e setup is not picking up the new ethereumjs deps and I found that the resolutions need to be added to the [root package.json](https://github.com/yarnpkg/yarn/issues/5039#issuecomment-880166818).

### DIFF
--- a/.github/workflows/e2e-hardhat.yml
+++ b/.github/workflows/e2e-hardhat.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [15]
+        node-version: [16]
     steps:
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v1
@@ -55,10 +55,9 @@ jobs:
 
       # Clone hardhat repo, inject proxied dependencies, and run tests
       - name: Build hardhat and test with new EthereumJS dependencies
+        run: bash ./scripts/e2e-hardhat.sh
         env:
           FORCE_COLOR: 3
-          INFURA_URL: ${{ secrets.URL_MAINNET }}
-        run: bash ./scripts/e2e-hardhat.sh
 
       - name: Remove hardhat directory to clean working state (for saving dependency cache)
         run: rm -rf hardhat
--- a/config/README.md
+++ b/config/README.md
@@ -4,8 +4,8 @@ This folder contains documentation on monorepo setup and configuration as well a
 configuration settings and scripts for common modules (like a shared `tsconfig.json` TypeScript
 configuration file or `cli/coverage.sh` script).
 
-- [Monorepo](./monorepo.md) (separate docs)
-- [E2E Testing](./testing.md) (separate docs)
+- [Monorepo](./MONOREPO.md) (separate docs)
+- [E2E Testing](./E2E_TESTING.md) (separate docs)
 - [Linting](#linting)
 - [Coverage](#coverage)
 - [TypeScript](#typescript)
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -459,7 +459,8 @@ export class BlockHeader {
     let parentGasLimit = parentBlockHeader.gasLimit
     // EIP-1559: assume double the parent gas limit on fork block
     // to adopt to the new gas target centered logic
-    if (this.number.eq(this._common.hardforkBlockBN('london'))) {
+    const londonHardforkBlock = this._common.hardforkBlockBN('london')
+    if (londonHardforkBlock && this.number.eq(londonHardforkBlock)) {
       const elasticity = new BN(this._common.param('gasConfig', 'elasticityMultiplier'))
       parentGasLimit = parentGasLimit.mul(elasticity)
     }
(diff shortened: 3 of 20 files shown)


acolytec3:
Ah, I think I remember that from my attempts to get it working. I guess I missed that in your final iteration on the PR.

ryanio:
> Ah, I think I remember that from my attempts to get it working. I guess I missed that in your final iteration on the PR.

yeah oops sorry about that, I think while I was trying to fix/refactor I updated it to hardhat-core's package.json when I saw the deps listed in there.

acolytec3:
So maybe this job is working as intended now. I'm seeing compile errors specific to our libraries around the `stateManager` now. That's promising.

ryanio:
> So maybe this job is working as intended now. I'm seeing compile errors specific to our libraries around the `stateManager` now. That's promising.

yes although it's a different error, I'm having trouble because if we put `ethereumjs-util` in the `resolutions` then it will overwrite the older `eth-sig-util` use of `ethereumjs-util v5` and causes these errors we've seen before ([here](https://github.com/ethereumjs/ethereumjs-monorepo/runs/3237299405#step:9:3356)):

```
1) Local accounts provider
       eth_signTypedData_v4
         Should be compatible with EIP-712 example:
     Error: This method only supports Buffer but input was: EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)
      at Object.exports.assertIsBuffer (/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/hardhat/node_modules/ethereumjs-util/src/helpers.ts:21:11)
      at Object.exports.keccak (/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/hardhat/node_modules/ethereumjs-util/src/hash.ts:13:3)
      at Object.hashType (/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/hardhat/node_modules/eth-sig-util/src/index.ts:267:20)
```

so I tried skipping `ethereumjs-util` as a test (although would be nice to have it) but then getting a weird type mismatch from it picking up `node_modules/ethereumjs-util` as well as `node_modules/@ethereumjs/block/node_modules/ethereumjs-util` (although these files should be the same as they haven't changed, the differing locations seems to be triggering it?) [here](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1391/checks?check_run_id=3243356722#step:9:343):

> Type 'Promise<import("/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/hardhat/packages/hardhat-core/node_modules/ethereumjs-util/dist/account").Account>' is not assignable to type 'Promise<import("/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/hardhat/node_modules/**@ethereumjs/block/node_modules/ethereumjs-util**/dist/account").Account>'.

> `├─ @ethereumjs/block@3.5.0`
> `│  └─ ethereumjs-util@7.2.0`

which patricio has pointed out would be solved by using peerDependencies, although that's a breaking change itself and peerDependencies typically don't install by default so it adds a step to the install process.

I'm still thinking/working on a solution for this ci, will probably test some ideas out.

ryanio:
I think using `resolutions` is overreaching since it is exceeding max major version ranges specified, e.g.:

```
yarn list v1.22.11
warning Resolution field "ethereumjs-util@7.2.0" is incompatible with requested version "ethereumjs-util@^5.1.1"
warning Resolution field "ethereumjs-util@7.2.0" is incompatible with requested version "ethereumjs-util@^6.0.0"
warning Resolution field "merkle-patricia-tree@4.3.0" is incompatible with requested version "merkle-patricia-tree@3.0.0"
warning Resolution field "ethereumjs-util@7.2.0" is incompatible with requested version "ethereumjs-util@^6.0.0"
warning Resolution field "ethereumjs-util@7.2.0" is incompatible with requested version "ethereumjs-util@^6.0.0"
warning Resolution field "ethereumjs-util@7.2.0" is incompatible with requested version "ethereumjs-util@^5.0.0"
warning Resolution field "merkle-patricia-tree@4.3.0" is incompatible with requested version "merkle-patricia-tree@^2.1.2"
```

Ideally we would do something like `"ethereumjs-util@^7.0.0": "7.2.0"` but that doesn't seem to be supported yet (https://github.com/yarnpkg/yarn/issues/5794).

I'm surprised more things didn't break! The only thing that seems to be tripping up is the ethereumjs-util change to only accept Buffers over strings.

My next thought is instead of using resolutions what if we just upgrade the direct ethereumjs deps in `hardhat-core/package.json`, that seems like a more reasonable approach now knowing that the resolutions behavior is to overwrite all sub-deps regardless of major version range. So this way only the direct deps used in the hardhat code should be upgraded to latest. I'll give it a try.

ryanio:
great! I have this working now. The new deps are correctly being proxied and I tested this by adding a commit that adds a new method to the statemanger interface and that failed during hardhat's e2e build with a type mismatch (```Property 'someNewlyAddedBreakingMethod' is missing in type 'ForkStateManager' but required in type 'StateManager'.```). this would have caught the v5.5.1 bug :|

acolytec3:
Pending a clean pass from these CI jobs. This looks good! Great work!

ryanio:
> Pending a clean pass from these CI jobs. This looks good! Great work!

thanks! I'll merge in

PR #1393, opened 2021-08-05
block, common: set constants block headers fields for The Merge

gabrocheleau:
This PR

- Adds EIP-3675 to the supported EIPs
- Sets the relevant block header fields as constants if EIP-3675 is activated
- Adds some basic tests for merge blocks

### DIFF
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -228,9 +228,9 @@ export class BlockHeader {
     }
 
     if (this._common.isActivatedEIP(3675)) {
-      uncleHash = keccak256(rlp.encode([]))
+      uncleHash = KECCAK256_RLP_ARRAY
       difficulty = new BN(0)
-      extraData = rlp.encode("b''")
+      extraData = Buffer.from('80', 'hex')
       mixHash = zeros(32)
       nonce = zeros(8)
     }
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -13,7 +13,6 @@ import {
   rlphash,
   toBuffer,
   zeros,
-  keccak256,
 } from 'ethereumjs-util'
 import { HeaderData, JsonHeader, BlockHeaderBuffer, Blockchain, BlockOptions } from './types'
 import {
@@ -212,7 +211,7 @@ export class BlockHeader {
       }
     }
 
-    //NOTE: The Merge fork block number won't be known prior to the actual fork (since it's decided by total difficulty)
+    // NOTE: The Merge fork block number won't be known prior to the actual fork, since it's determined by total difficulty
     if (options.hardforkByBlockNumber) {
       this._common.setHardforkByBlockNumber(number.toNumber())
     }
@@ -227,6 +226,7 @@ export class BlockHeader {
       }
     }
 
+    // Set these values as constants after the Merge
     if (this._common.isActivatedEIP(3675)) {
       uncleHash = KECCAK256_RLP_ARRAY
       difficulty = new BN(0)
@@ -243,13 +243,13 @@ export class BlockHeader {
       if (timestamp.isZero()) {
         timestamp = new BN(toBuffer(this._common.genesis().timestamp))
       }
-      if (difficulty.isZero() && this._common.isActivatedEIP(3675)) {
+      if (difficulty.isZero() && !this._common.isActivatedEIP(3675)) {
         difficulty = new BN(toBuffer(this._common.genesis().difficulty))
       }
-      if (extraData.length === 0 && this._common.isActivatedEIP(3675)) {
+      if (extraData.length === 0 && !this._common.isActivatedEIP(3675)) {
         extraData = toBuffer(this._common.genesis().extraData)
       }
-      if (nonce.equals(zeros(8)) && this._common.isActivatedEIP(3675)) {
+      if (nonce.equals(zeros(8)) && !this._common.isActivatedEIP(3675)) {
         nonce = toBuffer(this._common.genesis().nonce)
       }
       if (stateRoot.equals(zeros(32))) {
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -226,13 +226,36 @@ export class BlockHeader {
       }
     }
 
-    // Set these values as constants after the Merge
-    if (this._common.isActivatedEIP(3675)) {
-      uncleHash = KECCAK256_RLP_ARRAY
-      difficulty = new BN(0)
-      extraData = Buffer.from('80', 'hex')
-      mixHash = zeros(32)
-      nonce = zeros(8)
+    // Check for constant values for PoS blocks
+    if (this._common.consensusType() === 'pos') {
+      let error = false
+      let errorMsg = ''
+
+      if (!uncleHash.equals(KECCAK256_RLP_ARRAY)) {
+        errorMsg += `, uncleHash: ${uncleHash.toString(
+          'hex'
+        )} (expected: ${KECCAK256_RLP_ARRAY.toString('hex')})`
+        error = true
+      }
+      if (!difficulty.eq(new BN(0))) {
+        errorMsg += `, difficulty: ${difficulty} (expected: 0)`
+        error = true
+      }
+      if (!extraData.equals(Buffer.from([]))) {
+        errorMsg += `, extraData: ${extraData.toString('hex')} (expected: '')`
+        error = true
+      }
+      if (!mixHash.equals(zeros(32))) {
+        errorMsg += `, mixHash: ${mixHash.toString('hex')} (expected: ${zeros(32).toString('hex')})`
+        error = true
+      }
+      if (!nonce.equals(zeros(8))) {
+        errorMsg += `, nonce: ${nonce.toString('hex')} (expected: ${zeros(8).toString('hex')})`
+        error = true
+      }
+      if (error) {
+        throw new Error('Invalid PoS block' + errorMsg)
+      }
     }
 
     if (options.initWithGenesisHeader) {
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -225,38 +225,6 @@ export class BlockHeader {
       }
     }
 
-    // Check for constant values for PoS blocks
-    if (this._common.consensusType() === 'pos') {
-      let error = false
-      let errorMsg = ''
-
-      if (!uncleHash.equals(KECCAK256_RLP_ARRAY)) {
-        errorMsg += `, uncleHash: ${uncleHash.toString(
-          'hex'
-        )} (expected: ${KECCAK256_RLP_ARRAY.toString('hex')})`
-        error = true
-      }
-      if (!difficulty.eq(new BN(0))) {
-        errorMsg += `, difficulty: ${difficulty} (expected: 0)`
-        error = true
-      }
-      if (!extraData.equals(Buffer.from([]))) {
-        errorMsg += `, extraData: ${extraData.toString('hex')} (expected: '')`
-        error = true
-      }
-      if (!mixHash.equals(zeros(32))) {
-        errorMsg += `, mixHash: ${mixHash.toString('hex')} (expected: ${zeros(32).toString('hex')})`
-        error = true
-      }
-      if (!nonce.equals(zeros(8))) {
-        errorMsg += `, nonce: ${nonce.toString('hex')} (expected: ${zeros(8).toString('hex')})`
-        error = true
-      }
-      if (error) {
-        throw new Error('Invalid PoS block' + errorMsg)
-      }
-    }
-
     if (options.initWithGenesisHeader) {
       number = new BN(0)
       if (gasLimit.eq(DEFAULT_GAS_LIMIT)) {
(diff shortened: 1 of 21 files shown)


packages/block/src/header.ts
```diff
@@ -225,6 +227,17 @@ export class BlockHeader {
       }
     }
 
+    if (this._common.hardfork === Hardfork.Merge) {
```
jochem-brouwer, line 230: This works until we have a new hardfork after the Merge (i.e. Shanghai). It would be nice to have something on Common to figure out if the Merge is active or not.

gabrocheleau: Good point. Will think of something better, thanks.

ryanio: We could use common.isSupportedEIP(3675)

packages/block/src/header.ts
```diff
@@ -233,13 +243,13 @@ export class BlockHeader {
       if (timestamp.isZero()) {
         timestamp = new BN(toBuffer(this._common.genesis().timestamp))
       }
-      if (difficulty.isZero()) {
+      if (difficulty.isZero() && !this._common.isActivatedEIP(3675)) {
```
ryanio, line 246: i think if you were to put `// Set these values as constants after the Merge` after this `if (options.initWithGenesisHeader)` block then you wouldn't need to add these additional  `&& !this._common.isActivatedEIP(3675)` since the fields will simply be overwritten.

packages/block/src/header.ts
```diff
@@ -225,6 +226,15 @@ export class BlockHeader {
       }
     }
 
+    // Set these values as constants after the Merge
+    if (this._common.isActivatedEIP(3675)) {
+      uncleHash = KECCAK256_RLP_ARRAY
+      difficulty = new BN(0)
+      extraData = Buffer.from('80', 'hex')
```
ryanio, line 233: i know the spec is a little confusing here but for our library we would just set this extraData to an empty buffer: `Buffer.from([])`

packages/block/src/header.ts
```diff
@@ -211,6 +211,7 @@ export class BlockHeader {
       }
     }
 
+    // NOTE: The Merge fork block number won't be known prior to the actual fork, since it's determined by total difficulty
```
ryanio, line 214: I like the note but not sure if it is relevant in this location

packages/block/test/util.ts
```diff
@@ -14,8 +14,12 @@ function createBlock(
   uncles?: BlockHeader[],
   common?: Common
 ): Block {
-  uncles = uncles ?? []
   common = common ?? new Common({ chain: Chain.Mainnet })
+  if (common.isActivatedEIP(3675)) {
+    uncles = []
```
ryanio, line 19: nice, i'm thinking we can also do this in `block.ts` under each `// parse uncle headers` section, after `const uncleHeaders = []` wrap the `for (const uncleHeaderData of uhsData` loop in a `if (!common.isActivatedEIP(3675)) {` so if activated it will skip populating the array.

packages/block/test/mergeBlock.spec.ts
```diff
@@ -40,6 +40,11 @@ tape('The Merge tests', function (t) {
     st.end()
   })
 
+  t.test('should return that EIP 3675 is activated', function (st) {
+    st.ok(common.isActivatedEIP(3675))
```
holgerd77, line 43: I guess this test is unnecessary, this is such common functionality and tested elsewhere. Will remove.

holgerd77:
Ok, I've rebased here and squashed several commits.

I already removed the CHANGELOG entries again as per my comment on Discord. It is also generally far too early for release notes on this I guess since we are far far from the situation that we have even a somewhat clear picture on where this will lead too. 😋

Sorry that I am cutting so harshly through the woods here and also don't worry. It's actually super couragious that you gave this some start, great that we even already have some initial test cases! 😀

I've some more comments, will do separately a bit later.

holgerd77:
This EIP-3675 config in Common definitely needs to be added as some base selector. And I think we can (and should) also directly add "The Merge" as a hardfork. As some note on the switch discussion if you want to do this "by hardfork" you can use the `gte*` functionality from Common (instead of `===`), so this is e.g. from the VM `if (this._common.gteHardfork('byzantium')`.

What I had in mind as an additional switch is the ability to switch the consensus engine (type) setting in Common (see e.g. [goerli.json](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/chains/goerli.json) also on a per-hardfork basis and adopt the associated Common methods `consensusType()`, `consensusAlgorithm()` and `consensusConfig()` accordingly.

The advantage of this is that this frees us from this process of always needing to go through this Merge HF for getting into a PoS state and it would allow e.g. also for things like setting up a pure PoS test network which just runs on PoS from the beginning.

So the differentiation would be: for everything which is just pure PoS and is not just triggered by this Merge transition we should use a `common.consensusType() === 'pos' switch (and at some point it would really be nice to also switch to enums there, especially with this close 'poa', 'pos',... strings). For everthing which just one-time happens *if* there is an Eth1 chain available which need to be transitioned we should use checks based on the EIP or hardfork (I would actually prefer the hardfork checks since in this case we will (likely?) only have one EIP anyhow and this would improve readability I guess).

So in the case of these validation checks and header field adjustments these would clearly be `consensusType` checks since this is just about a block being instantiated as a PoS block.

Does this make sense?

I would directly integrate this in Common now and push here later on.

holgerd77:
(please read the last comment on site, many later-on updates and corrections)

packages/block/src/header.ts
holgerd77: I think we generally don't want to set this here so harsh that we always overwrite the initially provided values but rather either expect uses to provide the correct values if they want to provide themselves (and throw if incorrect) or otherwise set as the default value if not provided. Otherwise we let through bogus blocks if e.g. passed in as serialized data.

Will update along with Ryan's comment.

packages/block/src/header.ts
holgerd77: Update: This also goes for the adopted logic of throwing on wrong values. Apart from the question if/when there will be such thing like a PoS genesis header: if such a header *is defined including* the respective constant values, these will for sure match the anticipated values. If these values are left out, they will be filled up by our library with the correct values.

Will remove the checks.
---

Side note: I actually got a bit unsure again along the way when doing b9e9625 but this is confirmed by the spec:

> Add verification of the fields noted in block structure as their specified constant value, including the enforcement of empty ommers list.

packages/block/src/header.ts
holgerd77: Will remove.

packages/block/test/util.ts
holgerd77: Yes, will also remove this for now (tests are actually also passing without this).

In the block constructor if uncleHeaders are provided we should rather also throw.

holgerd77:
@gabrocheleau Ok, handinging it over again. 😄 

I know I've edited an insane amount of the initial code, but really nothing to worry about. Getting a start on these specs is pretty pretty hard, especially when it comes to such unchartered territory like the Merge and it's really great that we have gotten a start on really substantially modfiying and updating our base libraries.

holgerd77:
Ok, I just checked from this list from the specification:


    Remove verification of the block’s difficulty value with respect to the difficulty formula.
    Remove verification of the block’s nonce and mixHash values with respect to the Ethash function.
    Remove all validation rules that are evaluated over the list of ommers and each member of this list, except for enforcing this list to be empty.
    Add verification of the fields noted in block structure as their specified constant value, including the enforcement of empty ommers list.
    Add verification of block validity with respect to the PoS consensus rules, i.e. assert that there MUST be a corresponding event for the block to consider it valid. Such events are limited to the following:
        A POS_CONSENSUS_VALIDATED event considering the block or any of its descendants as valid.
        A POS_CHAINHEAD_SET event, including the POS_BLOCK_FINALIZED form of this event, designating the block or any of its descendants as the head of the canonical chain.

I guess apart from this event based check in the last rule - which is likely happening on another level together with client communication or something - we already have the relevant checks covered (some are just implicitly covered, e.g. these uncle checks we have in our library now always pass since uncles are guaranteed to be null.

So eventually if things are otherwise accepted on review we might already be able to merge this in.

packages/common/src/index.ts
```diff
@@ -959,8 +989,24 @@ export default class Common extends EventEmitter {
...
+   * casper: -
+   *
+   * Note: This value can update along a hardfork.
    */
   consensusConfig(): any {
+    const hardfork = this.hardfork()
+
+    let value
+    for (const hfChanges of HARDFORK_CHANGES) {
+      if (hfChanges[1].hasOwnProperty('consensus')) { // eslint-disable-line
```
ryanio, line 1001: @holgerd77 I thought you were a fan of the `in` operator :) these should be equivalent right?
```suggestion
      if ('consensus' in hfChanges[1]) {
```

packages/common/src/index.ts
```diff
@@ -59,6 +59,7 @@ export enum Hardfork {
   Berlin = 'berlin',
   London = 'london',
   Shanghai = 'shanghai',
+  TheMerge = 'theMerge',
```
ryanio, line 62: @holgerd77 what do you think about just calling it `merge` for short?

packages/common/src/index.ts
```diff
@@ -933,19 +934,48 @@ export default class Common extends EventEmitter {
...
   }
 
   /**
    * Returns the concrete consensus implementation
    * algorithm or protocol for the network
-   * e.g. "ethash" for "pow" consensus type or
-   * "clique" for "poa" consensus type
+   * e.g. "ethash" for "pow" consensus type,
+   * "clique" for "poa" consensus type or
+   * "casper" for "pos" consensus type.
```
ryanio, line 962: after using enums for `chain` and `hardfork` I wish we also had them for these 🤔

packages/common/src/index.ts
```diff
@@ -933,19 +934,48 @@ export default class Common extends EventEmitter {
...
+   *
+   * Note: This value can update along a hardfork.
    */
   consensusAlgorithm(): string {
+    const hardfork = this.hardfork()
+
+    let value
+    for (const hfChanges of HARDFORK_CHANGES) {
+      if (hfChanges[1].hasOwnProperty('consensus')) { // eslint-disable-line
```
ryanio, line 971: https://github.com/ethereumjs/ethereumjs-monorepo/pull/1393/files#r684523143
```suggestion
      if ('consensus' in hfChanges[1]) {
```

packages/block/test/block.spec.ts
```diff
@@ -738,4 +738,8 @@ tape('[Block]: block functions', function (t) {
       st.end()
     }
   )
+
+  /*t.test('should set constant block fields if on the merge hardfork', function (st) {
+    //TODO
```
ryanio, line 743: @acolytec3 mentioned we already have `mergeBlock.spec.ts` added in this PR, @holgerd77 is there something more you were expecting to test here?

ryanio: oh I don't see the code that sets these fields anymore, just validation, maybe you mean adding that back. we can probably add these tests to `mergeBlock.spec.ts` then.

ryanio: nevermind the default values are already set to the merge defaults if not provided e.g.

```
difficulty ? new BN(toBuffer(difficulty)) : new BN(0),
...
extraData ? toBuffer(extraData) : Buffer.from([]),
mixHash ? toBuffer(mixHash) : zeros(32),
nonce ? toBuffer(nonce) : zeros(8),
```
etc.

packages/common/src/index.ts
acolytec3: either `merge` or `theMerge(tm)`  Them `(tm)` is important!

packages/block/test/block.spec.ts
ryanio: will remove for now so we can merge this in, let me know if i should do a follow-up PR if i misunderstood the todo intention here and if it's not covered by `mergeBlock.spec.ts:32`

ryanio (approved):
wohoo, looks great, nice work everyone @gabrocheleau @holgerd77 @acolytec3!

packages/common/src/index.ts
holgerd77: Ok, let's do `merge` 😛 😅 (feel free everyone to update on occasion).

packages/common/src/index.ts
holgerd77: Yeah, for sure. Note sure, maybe we keep the strings in our json files, but expose these enums for users and also use these internally for comparisons in the code base?

packages/common/src/index.ts
holgerd77: 👍

packages/block/test/block.spec.ts
holgerd77: Not sure, Gabriel added this and I commented it out since it was breaking hte tests cause it had to `end()` call. I guess we are ok (for now) on the testing side I would say?

packages/common/src/index.ts
ryanio: That sounds nice to me, we took the same approach with the chain/hardfork enums by just introducing them as additional input options first.

PR #1395, opened 2021-08-09
monorepo: replace lerna with npm workspaces

ryanio:
This PR updates the monorepo from using `lerna` to [`npm workspaces`](https://docs.npmjs.com/cli/v7/using-npm/workspaces), simplifying dev ux by removing the need for bootstrapping.

Typescript [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) are used for linking packages during build time.

Run scripts in specific packages at root with e.g. `npm run docs:build --workspace=@ethereumjs/vm` or in all packages with `--workspaces`.


holgerd77:
Great, that sounds really promising on first sight! 😄 

Several questions here:

1. If I understand correctly `npm` is at least somewhat attached to a specific Node version (or is this not correct, always thought so?) So do I then always have to run some latest Node version here? And does this more broadly affect the ability to test with different Node versions?
2. Are users of the packages - apart from devs directly working on the monorepo - in any way enforced to use npm 7?
3. `npm run docs:build --workspace=@ethereumjs/vm`: so is it not possible any more to run `npm run docs:build` from the package folder? I would find this pretty inconvenient especially for things like testing.

holgerd77:
Let's please wait until the next EthereumJS call until we merge this in.

holgerd77:
(doesn't mean that this can't be reviewed yet of course)

acolytec3 (commented):
Once I installed the correct version of NPM and completely blew away the repo and started from scratch, everything seems to work fine and I like not having lerna around anymore.  :-)

A couple of general comments/questions before approving:
1) I think we need to somehow specify in `MONOREPO.md` that you need to be on NPM version 7 to make contributions to the project.  `npm i` threw errors when I tried to run it without being on npm7 but there weren't any obvious error messages telling me what was wrong.
2) What do you think about adding back the global lint/test scripts in the monorepo package.json?  I ran both and (admittedly test is still running) they seem to work so might be nice to have those global scripts available.
3) I noticed that the `packages/client/dist/bin/cli.js` file isn't being ignored after running `npm run build --workspaces` from the root.  Is this a necessary file for CI purposes?  I guess it's probably not a big deal but it could be a pain to deal with if we have that file sneaking into every new commit to the client package.

ryanio:
Thanks guys for taking an initial look! I'm excited for these improvements.

> 1. If I understand correctly `npm` is at least somewhat attached to a specific Node version (or is this not correct, always thought so?) So do I then always have to run some latest Node version here? And does this more broadly affect the ability to test with different Node versions?

Yes usually a Node version has a specific npm version installed with it (see [this table](https://nodejs.org/en/download/releases/)).

It's easy to upgrade to use npm v7 though regardless of the node version you are on, simply `npm i -g npm@7`, which I am doing in the ci.

> 2. Are users of the packages - apart from devs directly working on the monorepo - in any way enforced to use npm 7?

I don't think so, only developers of the monorepo.

> 3. `npm run docs:build --workspace=@ethereumjs/vm`: so is it not possible any more to run `npm run docs:build` from the package folder? I would find this pretty inconvenient especially for things like testing.

Yes you still can run it in the project directory, this is just a convenience when on root level to specify which workspace(s) the cmd should run.

---

> Once I installed the correct version of NPM and completely blew away the repo and started from scratch, everything seems to work fine and I like not having lerna around anymore. :-)

wohoo


> 1. I think we need to somehow specify in `MONOREPO.md` that you need to be on NPM version 7 to make contributions to the project.  `npm i` threw errors when I tried to run it without being on npm7 but there weren't any obvious error messages telling me what was wrong.

Yes this is actually what I was trying to do with the `engines` -> `npm` field but it just outputs a small warning instead of halting. Good point though, I will add extra documentation and I'm also thinking of adding a `preinstall` script that will grep `npm -v` and exit with a friendly error suggesting to do `npm i -g npm@7` if <v7 for workspaces support.

> 2. What do you think about adding back the global lint/test scripts in the monorepo package.json?  I ran both and (admittedly test is still running) they seem to work so might be nice to have those global scripts available.

I can add them back, it's a bit redundant though, because you can just run the script you want e.g. `npm run lint:fix --workspaces` to run the cmd (I updated the documentation to reflect this). I thought it was a bit redundant to add it to the root package.json and also it screws up if you want to run e.g. `npm run lint:fix --workspace=@ethereumjs/vm` then it would call the root script to run in all packages.

> 3. I noticed that the `packages/client/dist/bin/cli.js` file isn't being ignored after running `npm run build --workspaces` from the root.  Is this a necessary file for CI purposes?  I guess it's probably not a big deal but it could be a pain to deal with if we have that file sneaking into every new commit to the client package.

I was having a bit of trouble with this one, without that file present `npm install` was trying to `chmod` the file since it is defined as a bin path in `client/package.json` - and it was erroring out on 404 since the file is not present before building - I'm going to fiddle around with this one this morning and see if I can get a better/cleaner solution that does  not need to define that file as a placeholder field since as you mention it doesn't work when the file gets updated with code then tries to check into git.

ryanio:
ok, I wrote a nice little preinstall script to assist with npm 7 env:

```bash
➜  ethereumjs-monorepo git:(npm-workspaces) ✗ npm -v
6.14.5

➜  ethereumjs-monorepo git:(npm-workspaces) ✗ npm i

> root@ preinstall /Users/ryanghods/dev/ethereumjs-monorepo
> test $(npm -v | grep -o [0-9]*) -ge 7 && echo 'npm >=v7 satisfied' || (echo 'npm version 7 or greater is required for workspaces support. Please update with `npm i -g npm@7`' && exit 1)

npm version 7 or greater is required for workspaces support. Please update with `npm i -g npm@7`
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! root@ preinstall: `test $(npm -v | grep -o [0-9]*) -ge 7 && echo 'npm >=v7 satisfied' || (echo 'npm version 7 or greater is required for workspaces support. Please update with `npm npm i -g npm@7`' && exit 1)`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the root@ preinstall script.

➜  ethereumjs-monorepo git:(npm-workspaces) ✗ nvm use v12
Now using node v12.16.3 (npm v7.13.0)

➜  ethereumjs-monorepo git:(npm-workspaces) ✗ npm i

> preinstall
> test $(npm -v | grep -o [0-9]*) -ge 7 && echo 'npm >=v7 satisfied' || (echo 'npm version 7 or greater is required for workspaces support. Please update with `npm i -g npm@7`' && exit 1)

npm >=v7 satisfied
...
```

will push up a commit along with some other updates this afternoon.

.github/workflows/block-build.yml
```diff
@@ -34,17 +34,20 @@ jobs:
         uses: actions/cache@v2
         id: cache
         with:
-          key: Block-${{ runner.os }}-${{ matrix.node-version }}-${{ hashFiles('**/package-lock.json') }}
+          key: cache-${{ runner.os }}-${{ matrix.node-version }}-${{ hashFiles('**/package-lock.json') }}
```
alcuadrado, line 37: I believe the latest version of actions/setup-node does this automatically for you. I haven't tried that yet, but looks promising.

https://github.com/actions/setup-node#caching-packages-dependencies

acolytec3:
LGTM!

acolytec3 (approved):

.github/workflows/block-build.yml
ryanio: oh, nice, thanks! that does look a lot simpler. let me give it a try.

package.json
```diff
@@ -1,6 +1,12 @@
 {
   "name": "root",
   "private": true,
+  "workspaces": [
+    "packages/*"
```
holgerd77, line 5: There is the `ethereumjs-testing` submodule implicitly included here, might this have any side effects?

holgerd77:
Also gave this a brief review, looks really great! 😄 🎉 Will be brave and directly merge.

package.json
ryanio: that's a great point, it does seem to be ignoring it so far as a workspace.

when running commands at root with `--workspaces` the command is not executed in `ethereumjs-testing`.

PR #1399, opened 2021-08-11
devp2p: Add snappy compression for devp2p - EIP-706

### LINKED ISSUE #1004: devp2p: support for RLPx version 5
Have no deep insight on this yet, just dropping here so that it gots not forgotten.

There seems to be a newer version 5 of the `RLPx` protocol we are currently on version 4 (see [BASE_PROTOCOL_VERSION](https://github.com/ethereumjs/ethereumjs-devp2p/blob/master/src/rlpx/peer.ts#L15) setting.

Connections do still work, there needs to be an update nevertheless at some point I guess, also not really aware of the implications.

Some references (feel free to add here):

- Go Ethereum: https://github.com/ethereum/go-ethereum/pull/1821
- pydevp2p: "[WIP] Changes for RLPx version 5
" https://github.com/ethereum/pydevp2p/pull/7 (from 2015, so quite old)

Note: have labelled this "good first issue" since scope should be relatively clear once one is finding some valid specs here 😛, rather a PRO good first issue for someone with networking experience though.

acolytec3:
Fixes #1004 

Works for `eth` and `les`.  Client starts up and begins syncing successfully.


ryanio:
great start, does it make sense to add a few simple spec tests to the devp2p package to test the correct result outputs when snappy is on and off?

i will also help look into the failing les integration test and see where it may be going wrong as you mentioned on the call.

acolytec3:
@ryanio I was looking into tests but just haven't had a change to write them yet.  Will definitely try to add in one or two to verify that it correctly applies compression only when the remote node advertises devp2p v5.

The devp2p CI is still failing because the browser build process can't find snappyjs and this isn't something I'm 100% clear on how to resolve.  The new snappy library doesn't have any external dependencies and it definitely works in browser.  Do we have to tell typescript to specifically import it `tsconfig.browser.json` maybe?  [This](https://github.com/zhipeng-jia/snappyjs#browser) is all the library says about use.

packages/devp2p/src/rlpx/peer.ts
```diff
@@ -499,7 +501,14 @@ export class Peer extends EventEmitter {
     )
 
     try {
-      obj.protocol._handleMessage(msgCode, body.slice(1))
+      let payload = body.slice(1)
+
+      // Use snappy uncompression if peer supports DevP2P >=v5
+      if ((this._hello?.protocolVersion ?? 0) >= 5) {
```
holgerd77, line 507: What is this fallback to `0` here respectively what are the conditions? Is this when a peer didn't announce its RLPx version? Is this the still a valid connection which needs to be handled?

packages/devp2p/test/integration/eth-simulator.ts
```diff
@@ -208,3 +208,9 @@ test('ETH: invalid status send', async (t) => {
   }
   util.twoPeerMsgExchange(t, opts, capabilities)
 })
+
+test('RLPX: verify that snappy compression does not work with an RLPX peer that only supports devp2p 4', (t) => {
```
holgerd77, line 212: Sorry, a bit picky here, but can you reword "does not work" -> "is not used", if find this otherwise a bit misleading on a first read.

packages/devp2p/test/integration/util.ts
```diff
@@ -161,3 +161,48 @@ export function destroyRLPXs(rlpxs: any) {
...
+      t.equal(
+        err.message,
+        'Invalid Snappy bitstream',
+        'unable to process snappy compressed message'
+      )
+      t.end()
+      destroyRLPXs(rlpxs)
+    })
+  })
+}
```
holgerd77, line 208: Hmm, this is somewhat a code doubling of a utility method meant to be somewhat generic more or less for one test. Not sure, can we have this a bit more generic, eventually also integrated in the first version of the function? Otherwise it's a bit misplaced here.

Didn't have "the solution" [tm] though as well on first look, just to admit.

holgerd77 (changes requested):
Hi there,
pretty cool 😄, some remarks and one structurally substantial change request.

packages/devp2p/src/rlpx/peer.ts
acolytec3: I'm not knowledgeable enough on all the edge cases to know if this is truly a valid scenario or not, and was driven because Typescript is concerned that `_hello.protocolVersion` could be undefined.  Now that you've called it out, I'm assuming we would never try to send a message to a peer where we didn't have the `_hello` fields set?  If the peer didn't announce a Devp2p protocol version, we certainly would not want to run the snappy compression step but maybe a better question is would we want to send a message at all if we don't have the `protocolVersion` identified.  We certainly need to check if `protocolVersion >= 5`, but we may not need to specifically handle for the case of `undefined` at this point in the code.

packages/devp2p/package.json
```diff
@@ -37,6 +37,7 @@
     "tsc": "../../config/cli/ts-compile.sh",
     "lint": "../../config/cli/lint.sh",
     "lint:fix": "../../config/cli/lint-fix.sh",
+    "tape": "tape -r ts-node/register",
```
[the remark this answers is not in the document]
acolytec3: I think I first encountered the `tape` script in the `client` package a while back and found it very useful.  I really like it as a shorthand for being able to run an individual test file for those instances when I'm either triaging an existing test that is broken because of code changes I've made or else where I'm constructing a new one and I don't want to run the entire test suite for a package every time I need to re-run a test.  It's just a convenience thing for me as typing `npm run tape -- 'test/whatever_test_script_I_need_to_run.spec.ts'`  is easier than `npx tape -r ts-node/register 'test/...`  It's not original to me and existed at least in the client package  so I've just been adding it to packages that I work on when I discover that it's not there.  `npm run test` is a different use case as it always runs the complete test suite for the the package where it's executed from.

All that to say, I don't have a strong opinion on this if you feel strongly otherwise and can just easily set up a local snippet in my own dev environment.

packages/devp2p/src/eth/index.ts
```diff
@@ -230,7 +231,15 @@ export class ETH extends EventEmitter {
...
-    this._send(ETH.MESSAGE_CODES.STATUS, rlp.encode(this._status as any))
+
+    let payload = rlp.encode(this._status as any)
+
+    // Use snappy compression if peer supports DevP2P >=v5
+    if ((this._peer._hello?.protocolVersion ?? 0) >= 5) {
+      payload = snappy.compress(payload)
+    }
+
+    this._send(ETH.MESSAGE_CODES.STATUS, payload)
```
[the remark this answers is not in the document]
acolytec3: I think fundamentally you are correct in that Snappy is an RLPx protocol level change and not within subprotocols.  The challenge I had when initially coding this was that when I added the snappy compression check in `Peer._sendMessage`, I couldn't find a clear way to identify when the message being sent was part of the handshake (i.e. `HELLO/DISCONNECT/PING/PONG`) versus a subprotocol specific message since the `code` passed to `_sendMessage` can indicate either an RLPx handshake message or any of the first 4 messages sent by a subprotocol and `_sendMessage` can be called directly from the subprotocol handler.  That said, maybe I'm missing something here and this is more straight forward than I'm thinking.

packages/devp2p/test/integration/util.ts
acolytec3: I don't know if it's possible to merge the tests from this second method into the `twoPeerMsgExchange` message since that isn't truly a utility method given there are specific test conditions that are checked within `twoPeerMsgExchange`.  We might be able to but seems to me it would make more sense to completely change how this set of tests is structures so that the specific test conditions are passed into the utility method (if that's even feasible).  Just my $0.02 though.

packages/devp2p/src/eth/index.ts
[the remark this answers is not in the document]
ryanio: hmm, that is an interesting problem! one idea could be to add a param to  Peer._sendMessage() `fromSubprotocol = false` to get better granularity on `code`.

packages/devp2p/src/rlpx/peer.ts
ryanio: yes sorry this was a way for me to shortern the logic for `protocolVersion` being typed as `| undefined` but it looks kind of confusing.

@acolytec3's original code:
```typescript
if (this._peer._hello?.protocolVersion && this._peer._hello?.protocolVersion >= 5) {
```

Another way could be to check it beforehand like so:

```typescript
    if (!this._peer._hello?.protocolVersion) {
      return // or throw?
    }
    // Use snappy compression if peer supports DevP2P >=v5
    if (this._peer._hello.protocolVersion >= 5) {
```

We could go more deep and try to remove its optionality before it gets there as well.

packages/devp2p/package.json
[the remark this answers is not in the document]
ryanio: yes I find it helpful to have as well, we should update the test script to use it so it doesn't go unused and dedupes code (so for devp2p `"test": "npm run tape --  ./test/index.ts"`)

@holgerd77 it was originally introduced here https://github.com/ethereumjs/ethereumjs-client/pull/169 - I've found it really helpful in running individual test files during development and have added it in a few other packages since.

packages/devp2p/test/integration/eth-simulator.ts
ryanio:
```suggestion
test('RLPX: verify that snappy compression is not used with an RLPX peer that only supports devp2p 4', (t) => {
```

packages/devp2p/package.json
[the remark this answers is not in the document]
holgerd77: Ok, can go along, have answered in the chat as well on this.

packages/devp2p/src/rlpx/peer.ts
holgerd77: @ryanio Yeah, I guess this is rather a bit confusing and leads to unnecessary questions, will convert back to the old comparison version, this is more clear to read.

(will directly push here)

packages/devp2p/test/integration/util.ts
holgerd77: Yeah, that's also a good point, just wanted to mention. The whole test structure of the devp2p tests is not so optimal yet. Would be happy if we at some point take the time to get to some substantial structural improvements here with both a better test setup - eventually also with some help of networking mocking test tools and/or libraries - as well as some more fine-grained unit tests on the different communication scenarios.

Well well. So much to do. 😋

packages/devp2p/src/eth/index.ts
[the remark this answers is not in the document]
holgerd77: Ok, I wasn't aware of the differentiation here. So I guess it's also ok to keep as an alternative to add these kind of new flags or something, this is not bloating the structure or anything and a very local addition.

holgerd77 (approved):
LGTM, cool 😄, will merge.

packages/devp2p/test/integration/util.ts
ryanio: > Would be happy if we at some point take the time to get to some substantial structural improvements here with both a better test setup - eventually also with some help of networking mocking test tools and/or libraries - as well as some more fine-grained unit tests on the different communication scenarios.

yeah sounds great, maybe I can help with some of that, I would like to get more familiar with the devp2p internals and refactoring test code is always a good way to do that :)

PR #1402, opened 2021-08-13
Add peerDependencies to packages

### LINKED ISSUE #1200: Incompatible Library versions (TypeScript errors) / Peer dependency switch
I'm getting compilation errors when i upgrade some of the libraries.

For example, if I have `@ethereumjs/vm`, `@ethereumjs/block`, and `@ethereumjs/blockchain`, and upgrade the first two, yarn may decided not to upgrade `block` inside of `@ethereumjs/blockchain`. 

This leads to my project having multiple, sometimes incompatible, versions of the same libs. Even if the changes were backwards compatible, typescript may decided that two types are incompatible.

This image shows an example of this:

image

This error comes from incorrectly declaring some things as `dependencies` when they should be `peerDependencies`. Failing to do this properly leads to a very fragile system that breaks in this unexpected way, and can be easily bloated with multiple versions of the same library. 

I think I mentioned this problem before, and also pointed to [this excellent article](https://lexi-lambda.github.io/blog/2016/08/24/understanding-the-npm-dependency-model/) explaining what peer dependencies are and when should be used.

As an incomplete workaround, you can delete all of your project's `node_modules` and `yarn.lock`/`package-lock.json`. This would only help you compile your project, but it may be just transferring the problem to your users.

ryanio:
Fixes #1200

I would appreciate a double check of every listed peerDep, that it is used in the public interface of the package.

I didn't add any for `client` since it is an executable that you pass in command line args, I don't think it has any peerDeps then, is this reasoning correct?

npm 7 [installs peer dependencies by default](https://github.blog/2021-02-02-npm-7-is-now-generally-available/#peer-dependencies) so this goes great with our new developer setup, and should ease end user library incompatibilities as explained in the linked issue.


acolytec3 (commented):
Here's the additional ones I think we need:
`blockchain` <-- `block`, `@types/levelup`
`ethereum-js/util` <-- Do we need `rlp` since the whole module is exported as part of `util/externals`?
`vm` <-- `merkle-patricia-tree` as it's the type for one of the `VMOpts` properties

General question - do we need `@types/bn.js` as a peerDependency wherever we have `bn.js`?

holgerd77:
Hmm, can we please finish the discussion on this before opening PRs here?

I have got several issues here:

1. This would re-introduce on the sideline the need for breaking releases on all libraries, I thought we had some at least preliminary agreement that we only want to do VM, Blockchain(, Block)?
2. If I get this right this would require npm 7 for having peer dependencies automatically installed. Also thought that we had some somewhat agreeing discussion here: we can do some requirement like this in our own dev setup, but put this requirement on the library users is a totally other (and very likely not acceptable) thing (have no stats, but likely only a very tiny minority will have already upgraded to npm v7)
3. Last but not least, will just cite from my thread on #strategy: "Apart from what Patricio mentioned in the initial issue there had been not much complain about these kind of inconsistencies from other users, the UX is eventually questionable with the need to manually install dependencies (do I get this right?) and this is generally some strong intervention in our current installation process." -> so this is a very heavy change for a not yet very strongly proven benefit

I've switched the PR label here to "needs discussion" and "blocked".

Some general note here for reviewers:  apart from the discussion outcome this is a super-delicate PR. Please only review and do not merge until we have a very clear agreement here.

ryanio:
my apologies for getting ahead of the discussion

> we can do some requirement like this in our own dev setup, but put this requirement on the library users is a totally other (and very likely not acceptable) thing (have no stats, but likely only a very tiny minority will have already upgraded to npm v7)

yeah, I forgot that's really important. I'll close this PR for now and we can reopen if we'd like to continue in the future.

holgerd77:
No problem. 😄 

This will likely be valuable for some future integration, have added an additional note to the breaking release issue on the PR. For now rather too early, but we might want to re-check on the subsequent round of breaking releases next year or so.

PR #1406, opened 2021-08-16
Add fullsync integration test with merge fork set

acolytec3:
This is a proof of concept to create a full integration test of syncing a blockchain with the merge fork set to identify areas of potential change needed to support `the Merge(tm)`. Impacts at this point are primarily limited to the `block` and `blockchain` packages with a focus on the `_putBlockOrHeader` method to account for what logic changes are needed when a PoS block is to be inserted into the chain.

### DIFF
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -342,7 +342,7 @@ export class BlockHeader {
     }
 
     // Check for constant values for PoS blocks
-    if (this._common.consensusType() === ConsensusType.ProofOfStake) {
+    if (this._common.consensusType() === ConsensusType.ProofOfStake && this.number >= this._common.hardforks().filter((hf: any) => hf.name === 'merge').block) {
       let error = false
       let errorMsg = ''
 
--- a/packages/blockchain/src/index.ts
+++ b/packages/blockchain/src/index.ts
@@ -778,7 +778,7 @@ export default class Blockchain implements BlockchainInterface {
       if (!this._headHeaderHash) {
         throw new Error('No head header set')
       }
-
+      console.log('getting latest header hash', this._headHeaderHash.toString('hex'))
       const block = await this._getBlock(this._headHeaderHash)
       return block.header
     })
@@ -810,8 +810,11 @@ export default class Blockchain implements BlockchainInterface {
   async putBlocks(blocks: Block[]) {
     await this.initPromise
     for (let i = 0; i < blocks.length; i++) {
+      console.log('putting block ', blocks[i].header.number.toNumber(), blocks[i]._common.consensusType())
       await this.putBlock(blocks[i])
     }
+
+    console.log('finished putting blocks')
   }
 
   /**
@@ -880,6 +883,7 @@ export default class Blockchain implements BlockchainInterface {
       const isGenesis = block.isGenesis()
       const isHeader = item instanceof BlockHeader
 
+      console.log('got new block ready', block.header.number.toNumber(), block.header.hash().toString('hex'))
       // we cannot overwrite the Genesis block after initializing the Blockchain
       if (isGenesis) {
         throw new Error('Cannot put a genesis block: create a new Blockchain')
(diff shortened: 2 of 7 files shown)


packages/block/src/header.ts
```diff
@@ -342,7 +342,7 @@ export class BlockHeader {
     }
 
     // Check for constant values for PoS blocks
-    if (this._common.consensusType() === ConsensusType.ProofOfStake) {
+    if (this._common.consensusType() === ConsensusType.ProofOfStake && this.number >= this._common.hardforks().filter((hf: any) => hf.name === 'merge').block) {
```
ryanio, line 345: consensusType() changes to ProofOfStake after arriving at the merge hardfork block, so the added condition is redundant

packages/blockchain/src/index.ts
```diff
@@ -939,30 +943,48 @@ export default class Blockchain implements BlockchainInterface {
         }
       }
 
-      // set total difficulty in the current context scope
-      if (this._headHeaderHash) {
-        currentTd.header = await this.getTotalDifficulty(this._headHeaderHash)
+      if (block._common.consensusType() === ConsensusType.ProofOfStake) {
+        if (block._common.hardfork() !== 'merge') {
```
ryanio, line 947: what is this if statement trying to check? as noted in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1406/files#r689925035, consensusType() changes to ProofOfStake after arriving at the merge hardfork block, so this added condition is redundant

packages/client/test/integration/mocks/mockchain.ts
```diff
@@ -1,4 +1,5 @@
 import { Block } from '@ethereumjs/block'
+import { BN } from '../../../../util/dist'
```
ryanio, line 2:
```suggestion
import { BN } from 'ethereumjs-util'
```

packages/client/test/integration/mocks/mockchain.ts
```diff
@@ -23,16 +24,22 @@ export default class MockChain extends Chain {
 
   async build() {
     const blocks: Block[] = []
+    const mergeBlockNumber = this.config.chainCommon.hardforkBlockBN('merge')?.toNumber();
     for (let number = 0; number < this.height; number++) {
       const block = Block.fromBlockData({
         header: {
           number: number + 1,
-          difficulty: 1,
+          difficulty: number > 10 ? 0 : new BN(1),
```
ryanio, line 32: i would imagine this would break the other tests?

acolytec3: Yeah, I need to change that hardcoded 10 to be the block number for the merge hardfork.

packages/blockchain/src/index.ts
```diff
@@ -939,30 +943,48 @@ export default class Blockchain implements BlockchainInterface {
...
+          throw new Error('cannot insert PoS block when merge hardfork not set')
+        }
+        // TODO: Check if POS block is descendant of valid block
```
ryanio, line 950: could you explain more what you mean by this?

packages/client/test/integration/mocks/mockchain.ts
```diff
@@ -23,16 +24,22 @@ export default class MockChain extends Chain {
...
         },
       })
       blocks.push(block)
     }
-    await this.putBlocks(blocks)
+    try {
+      await this.putBlocks(blocks)
+    }
+    catch (err) {
+      console.log(err);
```
ryanio, line 42: wouldn't this end up printing the error to console anyway even without the added code if it throws?

packages/block/src/header.ts
acolytec3: Okay, let me try backing this change out and see if so. I think I was seeing it throw here with some PoS issue but that may be cause of me passing the wrong common instance in the test.

packages/blockchain/src/index.ts
acolytec3: From the EIP

Add verification of block validity with respect to the PoS consensus rules, i.e. assert that there MUST be a corresponding event for the block to consider it valid. Such events are limited to the following:
A POS_CONSENSUS_VALIDATED event considering the block or **any of its ancestors as valid.**
A POS_CHAINHEAD_SET event, including the POS_BLOCK_FINALIZED form of this event, designating the block as the head of the canonical chain.

I'm not sure what exactly this would look like but some sort of check the block hash is either matched by that `POS_CONSENSUS_VALIDATED` event or else an ancestor block.

packages/client/test/integration/mocks/mockchain.ts
acolytec3: Yes, just bad console logging on my part trying to figure out that EIP1559 error I kept seeing.

packages/block/src/header.ts
holgerd77: Yes, this check can (and should) definitely be removed.

packages/blockchain/src/index.ts
holgerd77: Yes, also here: the PoS check is enough. We likely also (a bit like the Block library) want to have the Blockchain library as "Merge" HF independent as possible, so if things are on `ProofOfStake`, the blockchain should behave like a PoS blockchain, no matter where this is coming from (from a Merge HF, from another HF switching to PoS, from Genesis).

The `Hardfork.Merge` HF should *only* be used if things *only* happen right along this switch (so e.g. if there would be something in the spec like `when the Merge HF occur put "We made it!" in the `extraData` field of the block 😋).

packages/blockchain/src/index.ts
holgerd77: I would rather assume that these things will implicitly happen in the client later on, e.g. once the client has received a `POS_CONSENSUS_VALIDATED` call by RPC or something it then calls the `put...` function on the blockchain class.

For now I guess it's ok if we skip this event related part completely on a first round, this will likely become more clear once we start testing with another (Eth 2.0) client.

acolytec3:
Closing for now.  Will reboot under a `blockchain` test instead.

PR #1408, opened 2021-08-17
PoS/The Merge(tm) changes for `Blockchain` package

acolytec3:
Introduces changes in the `blockchain` package to support the new `ProofOfStake` consensus algorithm/`merge` hardfork introduced in `block` in #1393 and #1398.


holgerd77:
Short nit: please only label the packages where you actually changed code (you also labeled `Block`) initially, this will otherwise confuse me on releases. 😛

packages/blockchain/package.json
```diff
@@ -26,8 +26,9 @@
     "tsc": "../../config/cli/ts-compile.sh",
     "lint": "../../config/cli/lint.sh",
     "lint:fix": "../../config/cli/lint-fix.sh",
+    "tape": "tape -r ts-node/register",
     "test": "npm run test:node && npm run test:browser",
-    "test:node": "tape -r ts-node/register ./test/*.spec.ts",
```
holgerd77, line 30: I never really got when this `ts-node/register` is necessary and when it is not, can someone explain this to me?

packages/blockchain/src/index.ts
```diff
@@ -939,30 +938,45 @@ export default class Blockchain implements BlockchainInterface {
...
       }
 
-      // set total difficulty in the current context scope
-      if (this._headHeaderHash) {
-        currentTd.header = await this.getTotalDifficulty(this._headHeaderHash)
-      }
-      if (this._headBlockHash) {
-        currentTd.block = await this.getTotalDifficulty(this._headBlockHash)
+      if (block._common.consensusType() === ConsensusType.ProofOfStake) {
+        if (block._common.hardfork() !== 'merge') {
```
holgerd77, line 942: As stated in the old PR, this check should not be there, see [comment](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1406/files#r689925951) for reference.

packages/blockchain/test/pos.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+  const londonBlockNumber = common.hardforkBlockBN('london')!.toNumber()
+  const genesis = Block.genesis({}, { common })
+
+  const initialBaseFee = new BN(common.param('gasConfig', 'initialBaseFee'))
+  blocks.push(genesis)
+
+  for (let number = 1; number < height; number++) {
+    console.log(number)
+    let baseFeePerGas: BN = new BN(0);
+    if (number === londonBlockNumber) baseFeePerGas = new BN(10000)
```
holgerd77, line 19: I wonder if we can make this more comfortable in the `Block` library, but I guess for now it's best to just manually set a hard-coded (so: not read with `Common.param()`) initial base fee to be set on the block, since the `Common` HF switch trigger hasn't occurred yet here.

I wonder though where you are getting this number from (`10000`), this would be `1000000000` (reading from the Common EIP definition.

Then the 1559 error goes away and this switches to a more descent error (`difficulty calculation is only supported on PoW chains`), which likely points to something not adopted in the Block or Blockchain library yet.

holgerd77: This new error is then triggered by the `calcDifficultyByHeader` option for the `Block` package. I guess we want this option to just have no effect on non-PoW chains (of after PoW dismissing HFs 😋), I will directly push here with this change.

holgerd77: Ok, did another push, tests should now pass. 😄

packages/blockchain/package.json
acolytec3: Is [this](https://github.com/substack/tape#preloading-modules) what you mean?  If I'm reading this correctly, we're telling `tape` to `require('ts-node/register') which then ensures that `ts-node` will compile all `.ts` files and then pass them to tape to run the tests using source maps so that stack traces will be legible when looking at test output.

I guess in this case, it was just telling tape to run all `.ts` files through ts-node before running the test scripts.

packages/blockchain/src/index.ts
acolytec3: Yep, I missed this when I cherrypicked the commit from the old PR.  Will be removed in my next commit.

packages/blockchain/test/pos.spec.ts
acolytec3: Concur, thanks!

packages/blockchain/src/index.ts
acolytec3: Yep, I missed this when I cherrypicked the commit from the old PR. Will be removed in my next commit.

packages/blockchain/package.json
holgerd77: Ok, and why we don't need this any more here?

acolytec3: I replaced `tape -r ts-node/register` with `npm run tape --` so the `test:node` script uses the `tape` script.

packages/blockchain/test/pos.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const buildChain = async (blockchain: Blockchain, common: Common, height: number) => {
+  const blocks: Block[] = []
+  const londonBlockNumber = common.hardforkBlockBN('london')!.toNumber()
+  const genesis = Block.genesis({}, { common })
+  blocks.push(genesis)
+
+  for (let number = 1; number <= height; number++) {
+    let baseFeePerGas: BN = new BN(0)
+    if (number === londonBlockNumber) baseFeePerGas = new BN(1000000000)
```
ryanio, line 16: For readability I would prefer if we always used curly braces with if statements (I should enable [eslint/curly](https://eslint.org/docs/rules/curly) hehe), it's just a personal preference but we stick to it pretty well in the monorepo. The only place where it doesn't really bother me is with `return` i.e. `if (statement) return`.

packages/blockchain/test/pos.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+    else if (number > londonBlockNumber) baseFeePerGas = blocks[number - 1].header.calcNextBaseFee()
```
ryanio, line 17: For readability I would prefer if we always used curly braces with if statements (I should enable [eslint/curly](https://eslint.org/docs/rules/curly) hehe), it's just a personal preference but we stick to it pretty well in the monorepo. The only place where it doesn't really bother me is with `return` i.e. `if (statement) return`.

ryanio (approved):
looks great! love the test with the custom common and impressed with how your original PR evolved

packages/blockchain/package.json
holgerd77: Ah, sorry for this likely confusing re-questioning. I actually misread the whole thing. 😜

packages/blockchain/test/pos.spec.ts
holgerd77: Yes, can strongly second that.

holgerd77:
Ah, are you guys really sure that we have already everything necessary covered in `Blockchain`? I expected this PR to be at least a couple of days more open and have some more discussion. 🤔

Ok, we'll see, will also give this another look, we likely can also always add things later if we discover.

acolytec3:
I think the biggest thing left is to really handle the switch from PoW to PoS via the terminal difficulty per the EIP rather than a specific block number provided in the chain info file. Maybe we can handle that under your `totalDifficulty` issue.

holgerd77:
@acolytec3 have you seen #1412? This is exactly on this question.

PR #1409, opened 2021-08-18
Common: allow Common with custom genesis state (for a custom chain)

### LINKED ISSUE #553: How to initialize the VM for a custom network with a custom genesis block

holgerd77:
Fixes #553

Base work for #1132 or follow-up PR

PR description see below


holgerd77:
Ok, some background on this:

I initially implemented (~3 years ago) the genesis state access functionality in Common [separately](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/common#genesis-states) from the main Common class with the main argument that the large genesis account state from `mainnet` shouldn't go into the main Common class for performance reasons.

I now did some more in-depth tests and also did some memory usage checks with this [roughSizeOfObject()](https://stackoverflow.com/questions/1248302/how-to-get-the-size-of-a-javascript-object) function from StackOverflow and it turned out that this was actually a naive assumption and the genesis state files are actually not read at all when - as being done in this PR - the `require` calls are encapsulated in the respective function call. So the `mainnet.json` file is ~300kb large, calling `roughSizeOfObject()` on a newly instantiated Common with the changes from this PR gives a very mild 4k (I actually did some more tests I don't all want describe here to be really really sure).

So the base assumption which lead to this separation is actually wrong and we can safely integrate into Common, which is generally a lot more natural way of accessing the genesis state.

When having a closer look and doing some tests on the VM integration in [StateManager.generateCanonicalGenesis()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/state/stateManager.ts#L556) I realized that this new version actually *improves* the status quo since in the old version the whole (mainnet) genesis account state array actually *was* created on every run of the VM along the import of this `genesisStateByName` function from `Common` together with this global array in [genesisStates/index.js](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/genesisStates/index.ts#L3) in Common adding all genesis states (I manually tested and confirmed this). So until now every VM initialization was actually unnecessarily creating the mainnet dict with ~ 50.000 entries. This is now not the case any more with this new implementation.

Since the old way of accessing genesis state was so artificial and was created on false assumptions I also deprecated the two old functions along the way, think this should be a no-brainer.

This whole integration then prepared the ground (this was actually the main motivation 😋) to now also allow for adding genesis states along custom chain initializations which solves one open issue we have (see above) and is also preparatory work for the client.

Generally open for review, eventually one small question: I now decided to just return an empty array if `Common.genesisState()` is called on a custom chain with no genesis state provided. I could imagine that this might be more convenient for some use cases but we could alternatively also throw. No strong opinion here.

holgerd77:
(`test-client-cli (goerli, full)` failure can be ignored)

acolytec3 (approved):
LGTM!  I was able to fire up a script, import multiple custom chains (including one with no genesis state), switch between them, and access genesis state items.

acolytec3:
Do y'all think we should add a `pattern 3` to the docs on the where someone is importing multiple custom chains with a mix of having and not having genesis states or is that too much of an edge case?  

You have to code it like `const customChains: [IChain, GenesisState][] = [[testnet, genesisStateOne], [testnet2, genesisStateTwo], [testnet3, {}]]` to get it to work

holgerd77:
> Do y'all think we should add a `pattern 3` to the docs on the where someone is importing multiple custom chains with a mix of having and not having genesis states or is that too much of an edge case?
 
Too much of an edge case I would say

PR #1410, opened 2021-08-18
Remove tsbuild caches

acolytec3:
Adds a step in `npm run clean` in the directory root to remove all the `tsbuild` caches so Typescript doesn't throw build errors due to not being able to find `ethereumjs` packages that haven't been built yet when running `npm run clean` and then `npm i`.

### DIFF
--- a/packages/block/package.json
+++ b/packages/block/package.json
@@ -19,7 +19,7 @@
   "scripts": {
     "build": "../../config/cli/ts-build.sh",
     "prepublishOnly": "npm run clean && npm run build && npm run test",
-    "clean": "rm -Rf ./dist && rm -Rf ./dist.browser",
+    "clean": "../../config/cli/clean-package.sh",
     "coverage": "../../config/cli/coverage.sh",
     "docs:build": "typedoc --options typedoc.js",
     "format": "ethereumjs-config-format",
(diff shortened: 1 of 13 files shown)


holgerd77:
Pretty cool find! 😄

Can you add this to the local `clean` command in all package folders as well (that's what I was using), so that the local .tsbuildinfo files (`tsconfig.prod.tsbuildinfo` and `tsconfig.browser.tsbuildinfo`) are deleted? Then we have this complete.

ryanio (approved):

packages/block/package.json
```diff
@@ -19,7 +19,7 @@
   "scripts": {
     "build": "../../config/cli/ts-build.sh",
     "prepublishOnly": "npm run clean && npm run build && npm run test",
-    "clean": "rm -Rf ./dist && rm -Rf ./dist.browser",
+    "clean": "../../config/cli/clean-package.sh",
```
holgerd77, line 22: Ah, no, this is actually not serving my needs, I would want to keep this local and not always delete builds from all packages, since then I always have to recreate my local environment after publishing a package.

I would switch this back to an extended local version along my next PR, if you had some strong reasoning for this change in semantics on this script let me otherwise know.

holgerd77: Update: ah, sorry, seems that this works actually only locally when invoked from the package folder (just tested)? Ok, then everything is fine! 😄

Issue #943, opened 2020-11-06
Ethash: check possible invalid RLP input data in tests


jochem-brouwer:
`validBlockRlp` from `ethash_block_rlp_tests.json` has a lot of `extraData` (>32 bytes) if you try to create the block. The block number is 1 so it is not a Genesis block. This RLP should be invalid. What is the purpose of this extra data? Why is there an invalid block used as test in ethash?

holgerd77:
These kind of smaller issues should be validated and/or tackled at some point so that they don't lay around forever. 😛

PR #1411, opened 2021-08-18
ethash: Switch out valid block RLP test data

acolytec3:
This addresses the concern raised by @jochem-brouwer in #943.  I've replaced the `validBlockRlp` with an rlp-encoded block from mainnet that conforms to actual block validity standards where `extradata` is not longer than 32 bytes.

I think the naming of the test might be a little obscure though since it's that the RLP encoding for the invalid block is invalid, but rather the block parameters don't generate a valid proof of work.


ryanio (approved):
nice, lgtm, thanks!

PR #1414, opened 2021-08-19
common: create custom chain `common` from geth genesis file

acolytec3:
Adds a new static constructor to `common` that allows the creation of a `common` instance from genesis file that follows the [geth/puppeth format](https://geth.ethereum.org/docs/interface/private-network)


packages/common/scripts/gethGenesisConverter.ts
```diff
@@ -0,0 +1,162 @@
+import { SecureTrie as Trie } from 'merkle-patricia-tree'
+import { BlockHeader } from '../../block/src/index'
```
ryanio, line 2:
```suggestion
import { BlockHeader } from '@ethereumjs/block'
```

packages/common/src/index.ts
```diff
@@ -303,6 +303,42 @@ export default class Common extends EventEmitter {
   }
 
   /**
+   *
+   * Creates a {@link Common} object for a custom chain, based off a complete {@link common.Chain} dictionary
+   * and optional {@link common.GenesisState}.
+   *
+   * These can be read in from a JSON file and passed to the constructor or the entire dictionary passed in directly.
+   * There is a helper script `scripts/gethGenesisConverter.ts` that takes the genesis file [layout]9https://geth.ethereum.org/docs/interface/private-network)
```
ryanio, line 311:
```suggestion
   * There is a helper script `scripts/gethGenesisConverter.ts` that takes the genesis file [layout](https://geth.ethereum.org/docs/interface/private-network)
```

packages/common/src/index.ts
```diff
@@ -303,6 +303,42 @@ export default class Common extends EventEmitter {
...
+   * `@ethereumjs/block` and `merkle-patricia-trie` packages.
+   *
+   * Creating custom genesis/genesisState objects from Geth
+   *
+   * ```bash
+   * geth dumpgenesis > genesis.json
+   * npx ts-node gethGenesisConverter.ts genesis.json
+   * ```
+   *
+   * Using with `Common.fromCustomGenesis`
```
ryanio, line 323:
```suggestion
   * Using with {@link Common.fromCustomGenesis}
```

packages/common/scripts/gethGenesisConverter.ts
holgerd77, line 2: wrong import

packages/common/src/index.ts
```diff
@@ -303,6 +303,42 @@ export default class Common extends EventEmitter {
...
+   * const common = Common.fromCustomGenesis(genesis, genesisState)
+   * ```
+   *
+   * @param genesisParams Chain options conforming to {@link Chain}
+   * @param genesisState Genesis state conforming to {@link common.GenesisState}
+   * @returns Promise<Common>
+   */
+  public static fromCustomGenesis(genesisParams: IChain, genesisState?: any): Common {
+    return new Common({ chain: genesisParams.name, customChains: [[genesisParams, genesisState]] })
+  }
```
holgerd77, line 339: Ah, no, I am not going along with this. That makes not so much sense actually to more or less replicate the Common custom chain format constructor. The main logic of parsing and everything should be in this constructor, otherwise there is not much gained.

So this should take in the geth genesis.json file.

holgerd77: Or am I missing something, maybe I am also a bit confused. Sorry, need to go to bed with my kids, please don't merge this PR yet. Would want to do another review.

holgerd77:
Rebased this.

packages/client/bin/cli.ts
```diff
@@ -184,12 +188,31 @@ async function run() {
...
-  // TODO: map chainParams (and lib/util.parseParams) to new Common format
-  // and pass into common constructor below
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  const chainParams = args.params ? await parseParams(args.params) : args.network
+  // configure common based on args given
+  let common: Common = {} as Common
+  // Use custom chain parameters file if specified
+  if (args.params) {
+    try {
+      const params = JSON.parse(fs.readFileSync(args.params, 'utf-8'))
```
holgerd77, line 196: This `params` naming is obviously just some relic from the past. But we can keep for this PR, will update when I do the Common chain file format integration.

Some other note: the whole naming of methods in `util/parse.ts` is not so optimal (not introduced by this PR but rather taken from the previous work). Some methods are labelled `parse*` and are doing something completely different. Geth-specific functionality is sometimes labelled accordingly, sometimes not. This should be updated at some point (especially the two exported methods `parseParams` and `parseGenesisState` should get some Geth prefix (`parseGethParams`).

holgerd77 (approved):
LGTM, cool, thanks! 😄

PR #1415, opened 2021-08-20
client: move missing headers/bodies message to correct log level

acolytec3:
As mentioned [here](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1176#issuecomment-900218077), the log messages regarding peers sending useless responses to `eth` protocol messages are moved from `warn` to `debug` to reflect that these are not items that our client is responsible for.


ryanio (approved):
great

PR #1419, opened 2021-08-24
Monorepo: separate build commands for node and browser

### LINKED ISSUE #1339: Monorepo: Separate Node and Browser Build commands
One CI improvement which is relatively high on my personal wishlist would be to have separate build commands for Node and the browser, since in 90%+ of the cases I am only interested in the Node builds and always doing the browser builds takes an unnecessary amount of extra time (especially for the client, there it is rather insane).

My suggestion here would be to follow the pattern of the `package.json` `test` commands and separate into a `build:node` and `build:browser` + an encompassing `build` (calling into both the other commands) command. I would then also suggest that we retire (respectively: delete) the current [config/cli/ts-build.sh](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/config/cli/ts-build.sh) script. I have the impression that this is not adding a lot of value and we can just simply call into `tsc -p ./tsconfig.browser.json"` via the local `package.json` file commands. There is nothing lost which was centralized before and the config is then still inherited from within the local `tsconfig.browser.json` files from the central `config` directory.

emersonmacro:
Closes #1339


holgerd77 (approved):
Looks great, tested this locally and everything works, already love it! 😄 ❤️

Thanks Emmett! Will merge.

PR #1423, opened 2021-08-24
client: Fix connection issues with geth when using custom genesis file

### LINKED ISSUE #1418: client computing wrong genesis `stateRoot` from geth genesis block
In recently merged #1409, a new `--gethGenesis` CLI parameter was added that aims to support starting up the client with genesis parameters imported from a running instance of geth via `geth dumpgenesis > genesis.json`.

When this `genesis.json` is provided as input to the client and then the client run, the client reports a `genesis block mismatch` with the paired geth node running on the same genesis parameters like so:

image

acolytec3:
Fixes #1418.

When using the `--gethGenesis` CLI parameter, the parameter parsing does not currently calculate the forkhash for each hardfork block in common so the client throws an `unknown fork` and refuses to connect with geth [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/892c71c473119e7f610841f569f2c03a4db0eac9/packages/devp2p/src/eth/index.ts#L133-L133) because `hardforkForForkHash` is only looking in `common._chainParams.hardforks` for the fork hash rather than computing it.  

This PR resolves the issue by duplicating the `_calcForkHash` helper method from `common` in `client/lib/util/parse.ts` and adding the fork hashes to each `hardfork` object in the `common` instance.

This currently works when using a mainnet/rinkeby/goerli genes file generated from `geth dumpgenesis > genesis.json`.

## To Do
 - [ ] Resolve forkhash calculation issue where hardforks with null block value occur 
 - [ ] Add better tests for custom genesis parameters
 
Note: This still breaks with `unknown fork` when using a geth `genesis` file where there is a hardfork with a `null` value for the block (e.g. on a chain where the `daoFork` doesn't occur).

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -2,7 +2,7 @@
 
 import { Server as RPCServer } from 'jayson/promise'
 import Common, { Hardfork } from '@ethereumjs/common'
-import { parseParams, parseMultiaddrs, parseGenesisState } from '../lib/util'
+import { parseMultiaddrs, parseGenesisState, parseCustomParams } from '../lib/util'
 import EthereumClient from '../lib/client'
 import { Config } from '../lib/config'
 import { Logger } from '../lib/logging'
@@ -93,8 +93,12 @@ const args = require('yargs')
       number: true,
       default: Config.MAXPEERS_DEFAULT,
     },
-    params: {
-      describe: 'Path to chain parameters json file',
+    customGenesis: {
+      describe: 'Path to custom genesis parameters json file',
+      coerce: path.resolve,
+    },
+    customGenesisState: {
+      describe: 'Path to custom genesis state json file',
       coerce: path.resolve,
     },
     dnsAddr: {
@@ -191,17 +195,21 @@ async function run() {
   // configure common based on args given
   let common: Common = {} as Common
   // Use custom chain parameters file if specified
-  if (args.params) {
+  if (args.customGenesis) {
     try {
-      const params = JSON.parse(fs.readFileSync(args.params, 'utf-8'))
-      common = new Common({ chain: params })
+      const genesisParams = JSON.parse(fs.readFileSync(args.customGenesis, 'utf-8'))
+      const genesisState = args.readFileSync(args.customGenesisState) ?? {}
+      common = new Common({
+        chain: args.customGenesis.split('.')[0],
+        customChains: [[genesisParams, genesisState]],
+      })
     } catch (err) {
-      throw new Error('invalid chain parameters')
+      throw new Error(`invalid chain parameters: ${err.message}`)
     }
     // Use geth genesis parameters file if specified
   } else if (args.gethGenesis) {
     const genesisFile = JSON.parse(fs.readFileSync(args.gethGenesis, 'utf-8'))
-    const genesisParams = await parseParams(genesisFile, 'custom-chain')
+    const genesisParams = await parseCustomParams(genesisFile, args.gethGenesis.split('.')[0])
     const genesisState = genesisFile.alloc ? await parseGenesisState(genesisFile) : {}
     common = new Common({
       chain: genesisParams.name,
@@ -213,6 +221,7 @@ async function run() {
     common = new Common({ chain: chain, hardfork: Hardfork.Chainstart })
   }
 
+  console.log(common.hardforks())
   const datadir = args.datadir ?? Config.DATADIR_DEFAULT
   const configDirectory = `${datadir}/${common.chainName()}/config`
   fs.ensureDirSync(configDirectory)
--- a/packages/client/lib/util/parse.ts
+++ b/packages/client/lib/util/parse.ts
@@ -11,9 +11,11 @@ import {
   unpadBuffer,
   isHexPrefixed,
   stripHexPrefix,
+  intToBuffer,
 } from 'ethereumjs-util'
 import { MultiaddrLike } from '../types'
-import { GenesisState } from '@ethereumjs/common/dist/types'
+import { GenesisState, Hardfork } from '@ethereumjs/common/dist/types'
+import { buf as crc32Buffer } from 'crc-32'
 
 /**
  * Parses multiaddrs and bootnodes to multiaddr format.
@@ -208,6 +210,10 @@ async function parseGethParams(json: any) {
     name: name,
     block: name === 'chainstart' ? 0 : config[forkMap[name]] ?? null,
   }))
+
+  for (const hf of params.hardforks) {
+    hf.forkHash = calcForkHash(hf.name, hash, params.hardforks)
+  }
   return params
 }
 /**
@@ -238,10 +244,10 @@ function formatNonce(nonce: string): string {
  * @returns
  */
 
-export async function parseParams(json: any, name?: string) {
+export async function parseCustomParams(json: any, name?: string) {
   try {
     if (json.config && json.difficulty && json.gasLimit && json.alloc) {
-      json.name = json.name || name
+      json.name = name
       json.nonce = formatNonce(json.nonce)
       return parseGethParams(json)
     } else {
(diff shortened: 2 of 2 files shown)


packages/client/bin/cli.ts
```diff
@@ -191,17 +195,21 @@ async function run() {
...
   // Use custom chain parameters file if specified
-  if (args.params) {
+  if (args.customGenesis) {
     try {
-      const params = JSON.parse(fs.readFileSync(args.params, 'utf-8'))
-      common = new Common({ chain: params })
+      const genesisParams = JSON.parse(fs.readFileSync(args.customGenesis, 'utf-8'))
+      const genesisState = args.readFileSync(args.customGenesisState) ?? {}
+      common = new Common({
+        chain: args.customGenesis.split('.')[0],
```
holgerd77, line 203: Maybe I am missing something, what is this `.split('.')[0]` doing? Tought this should very much take in our file format 1:1?

packages/client/bin/cli.ts
```diff
@@ -93,8 +93,12 @@ const args = require('yargs')
       number: true,
       default: Config.MAXPEERS_DEFAULT,
     },
-    params: {
-      describe: 'Path to chain parameters json file',
+    customGenesis: {
+      describe: 'Path to custom genesis parameters json file',
```
holgerd77, line 97: Can you add "from Common" here and below? Otherwise this is lacking context for users and people do not know what type of json file to insert here.

acolytec3: Done

packages/client/bin/cli.ts
acolytec3: You're right.  I corrected this in a follow-on commit where the chain name is pulled from the `name` property of the `Common` object.

acolytec3:
Quick write-up on how to test this:

1) Download this geth genesis file (created using `puppeth`) and rename to `macro.json`
[macro.json.txt](https://github.com/ethereumjs/ethereumjs-monorepo/files/7042750/macro.json.txt)
2) `geth init --datadir data macro.json`
3) `geth --datadir data --nat extip:[your local ip address here] --networkid 50434  --mine --miner.threads=1 --miner.etherbase=0x0000000000000000000000000000000000000000 --networkid 45375 --verbosity 6 --nodiscover`
4) Open another console and then run `geth attach data/geth.ipc`
5) Run `admin.nodeInfo` in the geth javascript console and then copy the `enode...` address
6) Open another console and navigate to the ethereumjs-monorepo root
7) `DEBUG=devp2p:* npm run client:start -w @ethereumjs/client -- --gethGenesis=path/to/macro.json --bootnodes=[enode address of your geth node] --port=30305
8) Copy the enode address the client prints to the logs
9) Go back to the geth dev console and run: admin.addPeer("[ethereumjs client enode address]")
10) Go back to the console where the client is running and you should shortly see it connect to geth and start receiving blocks.  

The issue where the `unknown forkhash` error appears seems to happen between block 60 and 80 (maybe when Geth switches to the byzantium hardfork -- but not yet confirmed)

holgerd77:
Hi Andrew, hanging a bit at step 3., can you recheck?

Several questions:

1. So this local IP is really my externally visible IP, is this correct? Why is this necessary?
2. There is a redundant `--networkid` option, can you update if one if falsy?
3. I am now using `geth --datadir data --nat extip:[MY_IP] --mine --miner.threads=1 --miner.etherbase=0x0000000000000000000000000000000000000000 --networkid 45375 --verbosity 6 --nodiscover` but I am getting a "Failed to start mining: etherbase missing: etherbase must be explicitly specified" error, do you know why?
4. Follow-up question: is setting `--miner.etherbase` generally necessary since the provided value seems to be the default anyhow?

holgerd77:
Ok, got it (I guess), it seems that etherbase can't be set to the zero address, see https://github.com/ethereum/go-ethereum/issues/19547.

Will update the initial guide along testing with what I assume is correct.

acolytec3:
> Several questions:
> 
> 1. So this local IP is really my externally visible IP, is this correct? Why is this necessary?

No, this is your local wifi IP address (so something like 192.168.0.XXX)
> 2. There is a redundant `--networkid` option, can you update if one if falsy?

It's not clear to me yet but if using custom genesis parameters, Geth doesn't broadcast the chainId included in the genesis.json.  You have to explicitly tell it via this `--networkid` parameter.  Otherwise, it broadcasts a network ID of 1 and then our client will disconnect with a network ID mismatch error
> 3. I am now using `geth --datadir data --nat extip:[MY_IP] --mine --miner.threads=1 --miner.etherbase=0x0000000000000000000000000000000000000000 --networkid 45375 --verbosity 6 --nodiscover` but I am getting a "Failed to start mining: etherbase missing: etherbase must be explicitly specified" error, do you know why?

Not sure if it's required or not (I just duplicated that from the instructions on the geth repo related to setting up a private network with a miner.  Further experimentation on my end indicates it is not necessary.
> 4. Follow-up question: is setting `--miner.etherbase` generally necessary since the provided value seems to be the default anyhow?

Based on above, probably not.

acolytec3:
One additional

> Ok, got it (I guess), it seems that etherbase can't be set to the zero address, see [ethereum/go-ethereum#19547](https://github.com/ethereum/go-ethereum/issues/19547).
> 
> Will update the initial guide along testing with what I assume is correct.

Curious that this didn't trigger an error for me.  Maybe Geth isn't as bullet proof as I've always assumed.  :grimacing:

holgerd77:
This DAG generation is taking a lot of time (and I also don't have enough disc space for this left, lol 😛) and generally I guess it would be even better fitting for our testing if we had this as a private PoA network. Will try to adopt the example a bit to get this working.

holgerd77:
Ok, I'll take the following chain configuration for a Clique private network:

```json
{
    "config": {
      "chainId": 45375,
      "homesteadBlock": 0,
      "eip150Block": 20,
      "eip150Hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
      "eip155Block": 40,
      "eip158Block": 60,
      "byzantiumBlock": 80,
      "constantinopleBlock": 100,
      "petersburgBlock": 120,
      "istanbulBlock": 140,
      "berlinBlock": 160,
      "londonBlock": 180,
      "clique": {
        "period": 5,
        "epoch": 30000
      }
    },
    "difficulty": "1",
    "gasLimit": "8000000",
    "extradata": "0x00000000000000000000000000000000000000000000000000000000000000007df9a875a174b3bc565e6424a0050ebc1b2d1d820000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
    "alloc": {
      "aaec86394441f915bce3e6ab399977e9906f3b69": { "balance": "300000" },
      "e13ece23b514caa5b53395c01d0d53d1843258ad": { "balance": "400000" }
    }
  }
  ```

These two accounts are our two development accounts.

**Comment by acolytec3** [comment]
> Ok, I'll take the following chain configuration for a Clique private network:

>     "alloc": {
>       "aaec86394441f915bce3e6ab399977e9906f3b69": { "balance": "300000" },
>       "e13ece23b514caa5b53395c01d0d53d1843258ad": { "balance": "400000" }

> These two accounts are our two development accounts.

Be careful with that one.  Those balances need to be hex strings.  I think that documentation is out of date because I used that exact setup and couldn't get geth to connect with the client with a `genesis block mismatch`.  Once I changed those balances to hex strings, I had no more issues.  Geth also outputs balances in `alloc` as hex strings if you do `geth dumpgenesis`.

**Comment by acolytec3** [comment]
And I think I know where the forkhash issue is coming from.  Geth has these forks defined in their genesis parameters:
```
    "homesteadBlock": 0,
    "eip150Block": 20,
    "eip150Hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "eip155Block": 40,
    "eip158Block": 60,
    "byzantiumBlock": 80,
    "constantinopleBlock": 100,
    "petersburgBlock": 120,
    "istanbulBlock": 140,
    "berlinBlock": 160,
    "londonBlock": 180,
```

What is `EIP158` equivalent to? Is there a hardfork in our `common` instance that corresponds to?

holgerd77:
Yes, that's the Spurious Dragon HF (these legacy names are really insane and pretty annoying, also causing confusion along the ethereum/tests integration).

PR #1425, opened 2021-08-25
Devp2p: RLPx send message code and debug improvements

holgerd77:
Ok, I hope that this will bring substantially more transparency to this central part of the RLPx code base. This whole "when is it called into which `sendMessage()` function from what context" topic was one of the main sources of confusion before. I actually already figured this out 3-4 times repeatedly and for this PR had to do another 30 minutes of research. 😜

Apart from several documentation additions one core of this PR is the improved "Received ..." debug message. This was before *always* inserting `undefined` for all subprotocol (e.g. `ETH`) messages, where people stumbled upon on debugging several times. What makes it worse is that people *then* interpret respectively associate the message code wrongly. So for `undefined` cases one is intuitively looking at the RLPx message code while the respective code is in this case referring to the subprotocol (e.g. `ETH` message codes) having - partly - the same number but a totally different semantics.

It is now explicitly logged if an RLPx or a subprotocol message is send, RLPx method:

grafik

Subprotocol method:

grafik

Hope this substantially reduces confusion on this front in the future. 😄

@acolytec3 maybe you can directly merge this in (if everything is ok) and do a short rebase of your PR on top, should be seamless to integrate.

### DIFF
--- a/packages/devp2p/src/rlpx/peer.ts
+++ b/packages/devp2p/src/rlpx/peer.ts
@@ -52,7 +52,7 @@ export type HelloMsg = {
 }
 
 export interface ProtocolDescriptor {
-  protocol?: any
+  protocol: any
   offset: number
   length?: number
 }
@@ -99,6 +99,11 @@ export class Peer extends EventEmitter {
   _disconnectReason?: DISCONNECT_REASONS
   _disconnectWe: any
   _pingTimeout: number
+
+  /**
+   * Subprotocols (e.g. `ETH`) derived from the exchange on
+   * capabilities
+   */
   _protocols: ProtocolDescriptor[]
 
   constructor(options: any) {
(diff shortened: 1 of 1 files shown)


holgerd77:
(several updates to the comment, please read on site)

holgerd77:
Ah, maybe also worth to mention in this context: there is at least one case (and where we also already stumbled upon) where an Ethereum client (Nethermind with a new [addCapability](https://github.com/NethermindEth/nethermind/blob/1d2f21ba04d4fd14cada95479d57b8d368b70af7/src/Nethermind/Nethermind.Network/P2P/P2PMessageCode.cs#L19) method adds to the existing RLPx messages (and where we for the moment can't do anything with these messages).

This case should now also be better distinguishable than before (the message name should remain emtpy, but the message code still present and *now* really pointing to a (missing) RLPx message.

packages/devp2p/src/rlpx/peer.ts
```diff
@@ -52,7 +52,7 @@ export type HelloMsg = {
 }
 
 export interface ProtocolDescriptor {
-  protocol?: any
+  protocol: any
```
holgerd77, line 55: I guess this is ok, there is always a protocol present (otherwise the whole thing makes no sense).

acolytec3 (approved):
LGTM!

PR #1437, opened 2021-08-28
client: parse gasLimit as number for type safety

acolytec3:
Geth genesis parameters store the `gasLimit` as a hex string while our `Common` instance calls for a `number`.  It's only currently used in the [header constructor](https://github.com/ethereumjs/ethereumjs-monorepo/blob/3d9774defa21cc9ca3af9c188d4be60d9de8c246/packages/block/src/header.ts#L231-L231) which is incidentally passing the value directly to a `new BN` constructor which currently takes a string as a possible input type so no errors are occuring.  

This small change parses the gasLimit to a `number` to ensure we have type safety for any future use and also keeps Typescript from complaining if you happen to be importing geth genesis parameters that have been converted to our file format as the gas limit will now correctly be converted to an integer first.

### DIFF
--- a/packages/client/lib/util/parse.ts
+++ b/packages/client/lib/util/parse.ts
@@ -166,7 +166,7 @@ async function parseGethParams(json: any) {
     genesis: {
       hash,
       timestamp,
-      gasLimit,
+      gasLimit: parseInt(gasLimit), // Geth stores gasLimit as a hex string while our gasLimit is a `number`
       difficulty,
       nonce,
       extraData,


packages/client/lib/util/parse.ts
```diff
@@ -166,7 +166,7 @@ async function parseGethParams(json: any) {
     genesis: {
       hash,
       timestamp,
-      gasLimit,
+      gasLimit: parseInt(gasLimit), // Geth stores gasLimit as a hex string while our gasLimit is a `number`
```
ryanio, line 169: since it's a hex string you'll want to specify radix of 16
```suggestion
      gasLimit: parseInt(gasLimit, 16), // Geth stores gasLimit as a hex string while our gasLimit is a `number`
```

acolytec3: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt), if it encounters 0x at the start of the string, it treats it as hexadecimal. From what I observe, geth always attaches the 0x at the front so this should not be an issue.  But, no strong opinion.

ryanio (approved):
great!

PR #1438, opened 2021-08-29
VM: Make CodeHashes have a DB prefix in StateManager

### LINKED ISSUE #1271: VM: Fix Undefined Opcode Bug
Test runs along #1269 have discovered a bug on calling a contract with an undefined opcode as first byte of the code added as a test in `ethereum/tests` here https://github.com/ethereum/tests/pull/836.

Test run can be reproduced with:

```shell
ts-node ./tests/tester --state --fork=Berlin --test='undefinedOpcodeFirstByte'
```

As some debug trace shows exception is thrown within an `SSTORE` operation on a `const found = await getContractStorage(runState, runState.eei.getAddress(), keyBuf)` call (see code [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/lib/evm/opcodes/functions.ts#L766)):

image

This respectively the root cause of the bug should be examined and the test readded to the tests to be run on CI.

jochem-brouwer:
Closes #1271.

In #1271, a contract is deployed with code 0x80, this 0x80 is the same as `RLP([])`, which is the empty trie. Thus, if we hash this (to get the codeHash) then we get the hash of the empty trie. This confuses StateManager, as now suddenly empty state roots point to `0x80`, which is an invalid node which we cannot decode. Therefore, there are two options:

(1) Hardcode that the empty trie hash is indeed an empty trie, if a code is requested then instead return 0x80

(2) prefix codeHashes

(2) is better (after a discussion with Geth). We currently only have one case (the code 0x80) where this runs into problems. However, Geth has reported that they also run into problems when using fast sync. If we at some point start to assume that if a BranchNode/ExtensionNode has an underlying DB value, then we could assume that we have the entire trie below. However, if we thus put the raw code of this BranchNode/ExtensionCode (so not the hash), then we could suddenly assume that we have certain tries, while we only have the raw trie node (or the code), instead of the entire trie. This could lead to problems if we start to use fast sync in a BFS approach (we might think we have certain tries, while we don't have them). We circumvent this problem by prefixing code hash `Buffer.from('c')`. (This is also taken from [Geth](https://github.com/ethereum/go-ethereum/blob/ffae2043f0482e81ac9f5459bf9fceafd74f52e9/core/rawdb/schema.go#L91))

This is a breaking change as it corrupts all existing MPT DBs and therefore points to `develop`.

This PR also ensures that BlockchainTests/StateTests use StateManager methods directly, and don't raw-write to Trie.

### DIFF
--- a/packages/vm/src/state/stateManager.ts
+++ b/packages/vm/src/state/stateManager.ts
@@ -20,6 +20,14 @@ import { AccessList, AccessListItem } from '@ethereumjs/tx'
 
 const debug = createDebugLogger('vm:state')
 
+/**
+ * Prefix to distinguish between a contract deployed with code `0x80`
+ * and `RLP([])` (also having the value `0x80`).
+ *
+ * Otherwise the creation of the code hash for the `0x80` contract
+ * will be the same as the hash of the empty trie which leads to
+ * misbehaviour in the underyling trie library.
+ */
 const CODEHASH_PREFIX = Buffer.from('c')
 
 type AddressHex = string
--- a/packages/vm/src/state/stateManager.ts
+++ b/packages/vm/src/state/stateManager.ts
@@ -20,6 +20,8 @@ import { AccessList, AccessListItem } from '@ethereumjs/tx'
 
 const debug = createDebugLogger('vm:state')
 
+const CODEHASH_PREFIX = Buffer.from('c')
+
 type AddressHex = string
 
 /**
@@ -174,7 +176,8 @@ export default class DefaultStateManager implements StateManager {
       return
     }
 
-    await this._trie.db.put(codeHash, value)
+    const key = Buffer.concat([CODEHASH_PREFIX, codeHash])
+    await this._trie.db.put(key, value)
 
     const account = await this.getAccount(address)
     if (this.DEBUG) {
@@ -195,7 +198,8 @@ export default class DefaultStateManager implements StateManager {
     if (!account.isContract()) {
       return Buffer.alloc(0)
     }
-    const code = await this._trie.db.get(account.codeHash)
+    const key = Buffer.concat([CODEHASH_PREFIX, account.codeHash])
+    const code = await this._trie.db.get(key)
     return code ?? Buffer.alloc(0)
   }
 
--- a/packages/vm/tests/BlockchainTestsRunner.ts
+++ b/packages/vm/tests/BlockchainTestsRunner.ts
@@ -79,7 +79,7 @@ export default async function runBlockchainTest(options: any, testData: any, t: 
   await blockchain.initPromise
 
   // set up pre-state
-  await setupPreConditions(vm.stateManager._trie, testData)
+  await setupPreConditions(vm.stateManager, testData)
 
   t.ok(vm.stateManager._trie.root.equals(genesisBlock.header.stateRoot), 'correct pre stateRoot')
 
--- a/packages/vm/tests/BlockchainTestsRunner.ts
+++ b/packages/vm/tests/BlockchainTestsRunner.ts
@@ -179,6 +179,9 @@ export default async function runBlockchainTest(options: any, testData: any, t: 
         return
       }
     } catch (error) {
+      if (options.debug) {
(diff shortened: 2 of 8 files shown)


holgerd77:
Hi @jochem-brouwer, thanks a lot for this fantastic write-up and the associated PR. 🤩

I think I am getting both the problem and the solution 😋, just to make sure that I get all the implications, could you give a short confirmation on the following assumptions:

1. This is a breaking change for the VM. Old state DBs could not be read any more with this new version (wondering if this is too harsh - not sure how many use cases are out there where the VM is used in some stateful context (in the sense of: keeping the state and doing something with it a couple of days later) and if we should add a backwards compatibility option (a flag)? 🤔 Totally not sure here or even suggesting, just throwing this in the ring for discussion. Worth it or too much overhead and technical debt?)
2. This is not breaking for `Trie` in any way and als has no side effects on Trie usage (from all logical angles (no Trie code edited, trie being a dependency in the VM) it shouldn't, nevertheless writing it down here to reassure.
3. For our client this would mean that it breaks all existing sync folder states (which I guess is something we can live with).

jochem-brouwer:
Hi @holgerd77, yeah some clarifcations on your points:

1. Yes, this will indeed break any DB which has been written to disk. I have thought about some sort of flag to keep using the old behavior, but due to this 0x80 bug I don't think this is what we want. What we could do, however, is to provide an upgrade script. What we'd then do is take the old DB, traverse the entire trie (`walkTrie`), and then for each account we just write the values of each `codeHash` to key `Buffer.concat([Buffer.from('c'), codeHash])`, essentially upgrading the database. I don't think we should provide this in this PR, as I honestly doubt anyone is writing state to disk, neither is anyone using our client for anything useful. But if we get an issue like this, it should not be super hard to write this database upgrade.
2. Yup, exactly! The first approach as outlined in the OP of this PR would edit Trie.
3. Yes, this is true, but this is also fixable as described in 1. However, as stated there, I doubt anyone is using our client for anything serious as it is not very useful (for practical purposes) at this moment.

jochem-brouwer:
I still have to figure out why tests are breaking though, some specific ones seem to fail, but I have not yet figured out why.

jochem-brouwer:
(Small note here, we might want to integrate some central place for these prefixes at some point, because we also have these prefixes in the [blockchain library](https://github.com/ethereumjs/ethereumjs-monorepo/blob/866f865e792a613e9d417f67eecec36f06dc47a0/packages/blockchain/src/db/constants.ts), and now thus also in the state trie. Maybe a good place for this would be in Common? Then we have something similar as what [Geth](https://github.com/ethereum/go-ethereum/blob/ffae2043f0482e81ac9f5459bf9fceafd74f52e9/core/rawdb/schema.go#L91) has, which is cleaner)

jochem-brouwer:
I'm touching a bit in the dark here because these tests make no sense. The pre-stateroots are right, but then when we run blocks then they suddenly fail. It almost seems that we are writing directly to the DB somewhere in our code without using StateManager :thinking: I just ran a test, checked the JSON stack traces of the opcodes, code address, gas left, and stack and they both report the same :thinking:

jochem-brouwer:
Ok great, finally found the issue :sweat_smile: 

Tests should now pass, ready for review.

holgerd77:
Ok, I have added an additional longer comment to `CODEHASH_PREFIX`, I guess this will otherwise be one of these things where future developers will scratching their head. 😋 Otherwise this looks good, thanks Jochem!

Have rebased the `develop` branch (and made a backup), rebase this here against the rebased `develop` and will merge in (without deleting the PR branch) when tests will pass.

holgerd77 (approved):
LGTM

davidmurdoch:
Hi, Ganache Team here. We write state to disk.

holgerd77:
> Hi, Ganache Team here. We write state to disk.

I'm honestly totally not getting the context of this comment here. 🙂

davidmurdoch:
Haha. Sorry, I was referring to the comment above that says "I honestly doubt anyone is writing state to disk". I'm in Discord chatting with Jochem about this issue now.

PR #1441, opened 2021-08-30
Client: New Pooled Tx Hashes Service Functionality (Re-broadcasting/Tracking) for TxPool

holgerd77:
This implements the service functionality around the rebroadcasting of new pooled tx hashes received.

[WIP]

### DIFF
--- a/packages/client/lib/net/protocol/ethprotocol.ts
+++ b/packages/client/lib/net/protocol/ethprotocol.ts
@@ -77,6 +77,12 @@ export class EthProtocol extends Protocol {
         }
         return serializedTxs
       },
+      decode: ([txs]: [Buffer[]]) => {
+        // TODO: add proper Common instance (problem: service not accesible)
+        //const common = this.config.chainCommon.copy()
+        //common.setHardforkByBlockNumber(this.service.synchronizer.syncTargetHeight)
+        return txs.map((txData) => TransactionFactory.fromBlockBodyData(txData))
+      },
     },
     {
       name: 'GetBlockHeaders',
@@ -166,6 +172,9 @@ export class EthProtocol extends Protocol {
       },
       decode: ([reqId, txs]: [Buffer, any[]]) => [
         new BN(reqId),
+        // TODO: add proper Common instance (problem: service not accesible)
+        //const common = this.config.chainCommon.copy()
+        //common.setHardforkByBlockNumber(this.service.synchronizer.syncTargetHeight)
         txs.map((txData) => TransactionFactory.fromBlockBodyData(txData)),
       ],
     },
--- a/packages/client/lib/net/protocol/ethprotocol.ts
+++ b/packages/client/lib/net/protocol/ethprotocol.ts
@@ -3,7 +3,7 @@ import { BlockHeader, BlockHeaderBuffer } from '@ethereumjs/block'
 import { Chain } from './../../blockchain'
 import { Message, Protocol, ProtocolOptions } from './protocol'
 import { BlockBodyBuffer } from '@ethereumjs/block'
-import { TypedTransaction } from '../../../../tx/dist'
+import { TransactionFactory, TypedTransaction } from '@ethereumjs/tx'
 
 interface EthProtocolOptions extends ProtocolOptions {
   /* Blockchain */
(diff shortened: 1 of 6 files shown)


packages/client/test/sync/txpool.spec.ts
```diff
@@ -9,6 +9,8 @@ tape('[TxPool]', async (t) => {
   const common = new Common({ chain: 'mainnet', hardfork: 'london' })
   const config = new Config({ transports: [], loglevel: 'error' })
 
+  const { PeerPool } = await import('../../lib/net/peerpool')
```
ryanio, line 12: this can be imported normally at the top of the file (doesn't need `await import` unless modifying with testdouble)

holgerd77: Ah, lost oversight, thanks. Will update.

holgerd77:
Ok, I am not getting an integration test to run (have left a TODO comment), but otherwise this is ready for review. 😄

holgerd77:
(and from my side it can be merged)

packages/client/lib/net/protocol/ethprotocol.ts
```diff
@@ -65,6 +66,17 @@ export class EthProtocol extends Protocol {
     {
       name: 'Transactions',
       code: 0x02,
+      encode: (txs: TypedTransaction[]) => {
+        const serializedTxs = []
+        for (const tx of txs) {
```
ryanio, line 71: PR looks good! I'm just taking an initial look and I'm noticing coverage missing for these newly added encode/decode functions. Do you think it would be worth the time to add tests for them? We ran into the same thing with #1443. I think it's important to keep these lines covered to reduce the likelihood of breakage in the future. We should ensure we have a nice test suite that can be expanded easily as more functions are added in these files (ethprotocol and fullethereumservice). I believe there should be existing tests for some of the prior methods that we can just replicate.

packages/client/lib/sync/txpool.ts
```diff
@@ -12,17 +12,23 @@ export interface TxPoolOptions {
 
 type TxPoolObject = {
   tx: TypedTransaction
-  hash: string // plain strings without hex prefix
+  hash: UnprefixedHash
```
ryanio, line 15: nice I like the increased clarity with the naming :)

holgerd77:
Ok, added the respective integration tests (and as always: fixed an additional bug along the way).

Now also added the functionality for retrieving Transactions messages and properly rebroadcast (this is split by sending full tx messages to a fraction of peers and hash announcement messages to the rest, see devp2p [Transaction Exchange](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#transaction-exchange) section).

Ready for review. 😄

packages/client/lib/sync/txpool.ts
```diff
@@ -12,17 +12,23 @@ export interface TxPoolOptions {
...
+  address: UnprefixedAddress
+  added: number
+}
+
+type SentObject = {
+  hash: UnprefixedHash
   added: number
 }
 
 type UnprefixedAddress = string
```
ryanio, line 29: Just taking a look at how this is used (`UnprefixedAddress`), actually when you do `getSenderAddress().toString()` that will return the 0x-prefixed address (if you look at `Address` class `toString` method).

Two ways to get the unprefixed hash would be:
`getSenderAddress().toString().slice(2)`
or
`getSenderAddress().buf.toString('hex')`

acolytec3: Couldn't we also just use `stripPrefix` from `ethereumjs-util`?  I know it's a simple convenience function but is that fruitful in this instance?

acolytec3:
This looks great!  With some slight hacks, I'm seeing it receive and rebroadcast transactions on `goerli`!

```sh
  devp2p:eth Received NEW_POOLED_TRANSACTION_HASHES message from 163.172.109.219:30303: e1a08b4092bbeb597631369630fab684716705e003f87ea4898df5e897dd... +990ms
DEBUG [09-03|14:50:15] TxPool: received new pooled hashes number=1 
DEBUG [09-03|14:50:15] TxPool: requesting txs number=1 pending=0 
  devp2p:eth Send GET_POOLED_TRANSACTIONS message to 163.172.109.219:30303: e5820109e1a08b4092bbeb597631369630fab684716705e003f87ea4898d... +2ms
  devp2p:eth Received POOLED_TRANSACTIONS message from 163.172.109.219:30303: f877820109f872f870028504a817c80082a42494ce43d3d6f0829481d970... +262ms
DEBUG [09-03|14:50:16] TxPool: received requested txs number=1 
```

acolytec3 (approved):
I merged in the latest from master and fixed the one spot where my new `NEW_BLOCK` test broke here.  LGTM!

holgerd77:
@acolytec3 ok, cool, thanks for the review! 😄

Issue #1416, opened 2021-08-23
Client: Invalid remainder Sync Error


holgerd77:
We currently have a lot of the following errors occuring directly at the beginning of sync:

```shell
ERROR [08-23|10:12:39] Error: invalid remainder
    at Object.decode (/ethereumjs-monorepo/node_modules/rlp/src/index.ts:75:11)
    at Peer._handleMessage (/ethereumjs-monorepo/packages/devp2p/src/rlpx/peer.ts:430:25)
    at Peer._handleBody (/ethereumjs-monorepo/packages/devp2p/src/rlpx/peer.ts:511:20)
    at Peer._onSocketData (/ethereumjs-monorepo/packages/devp2p/src/rlpx/peer.ts:540:18)
    at Socket.emit (events.js:315:20)
    at Socket.EventEmitter.emit (domain.js:467:12)
    at addChunk (internal/streams/readable.js:309:12)
    at readableAddChunk (internal/streams/readable.js:284:9)
    at Socket.Readable.push (internal/streams/readable.js:223:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:188:23)
```

Tested this with `npm run client:start --network=rinkeby --minPeers=1` (`minPeers` was for other reasons not relevant here) and it directly occurs along the beginning of the sync process, often several times (side note: so this is just confirmed for `Rinkeby` yet, I didn't test on other networks). The sync process than normally continues.

I first suspected the changes from the tip of the chain PR #1132 to be responsible for this, but I tested by checking out a previous commit and the error is still there, so this is therefore not the case.

Won't dig deeper here for now but just opening as a tracking issue. If you've got new findings around this please add here so that we can keep track.

acolytec3:
I've seen it too and I wonder if it isn't related to #1399  somehow.  I've seen it on mainnet as well so will research some today.

holgerd77:
@acolytec3 oh yeah, I guess this would be a very fitting candidate for this kind of bug behavior. So it might be good that we give this another more close review and/or writing some more tests.

(as some general note: we might have gotten a bit loose on reviews lately (including myself) and waving things through a bit too light heartedly. Eventually we should generally lift up here again, otherwise we'll reintroduce hard-to-track bugs in our libraries and especially the client).

As some first step it should be easy to test if things work "normal" again if one is manually downgrading to RLPx v4 in `devp2p` and therefore deactivating the snappy compression.

holgerd77:
[...]

PR #1442, opened 2021-08-30
devp2p: add snappy compression to ping/pong

### LINKED ISSUE #1421: Client: Connection to sync peer gets lost quickly
Just observed when trying to reproduce on #1416 that the sync connection to a peer is lost pretty quickly (within 1-2 minutes I would say). Sync is then rebooted and continued, soon after rebooting and reconnecting the connection is lost again and so on.

I tested syncing with `goerli`: `DEBUG=devp2p:* npm run client:start -- --loglevel=debug --network=goerli`

What I noticed: immediately before the disconnection we are sending another `PING` method to the already connected peer. I could capture this in the following screenshot:

grafik

So here we are already connected to this peer with the IP `145.239.255.212` and sync is processing. Then we are sending another `PING` and we disconnect afterwards with a `USELESS_PEER` disconnect reason.

Two possible explanations here (I can think of on first thought):

1. These `PING` re-sending is part of normal protocol behavior, might be requested to re-send a `PING` after some amount of time to reassure the connection is still active (for now I would assume this is rather not the case from my rough protocol knowledge, but might also be taken into consideration as a possibility). Might be that we then not analyze properly the message flow and inappropriately disconnect.
2. We re-stumble upon this peer because it is again send to us by peer discovery. We likely do not check here if we already have an ongoing connection with such peer, then try to reconnect which interferes with the already ongoing sync process with named peer.

I would assume 2. is rather the more likely possibility. Nevertheless wrote down both cause I am not 100% sure.

We should analyze this (with some priority), the current behavior is not really sustainable for a robust sync process.

acolytec3:
Fixes #1421

When we did the original devp2p v5 in #1399 to add snappy compression to inbound/outbound devp2p messages, we didn't add snappy compression to the `ping`/`pong` messages since I wrongly interpreted them as not having a body so therefore not needing snappy.  This proved incorrect as we are still RLP encoding an empty array for the body and this also needs snappy compression.  Geth will now maintain connection through the `ping`/`pong` process.

### DIFF
--- a/packages/devp2p/src/rlpx/peer.ts
+++ b/packages/devp2p/src/rlpx/peer.ts
@@ -282,7 +282,7 @@ export class Peer extends EventEmitter {
    */
   _sendPong() {
     debug(`Send PONG to ${this._socket.remoteAddress}:${this._socket.remotePort}`)
-    let data = rlp.encode([])    
+    let data = rlp.encode([])
 
     if (this._hello?.protocolVersion && this._hello.protocolVersion >= 5) {
       data = snappy.compress(data)
--- a/packages/devp2p/src/rlpx/peer.ts
+++ b/packages/devp2p/src/rlpx/peer.ts
@@ -202,6 +202,7 @@ export class Peer extends EventEmitter {
    */
   _sendMessage(code: number, data: Buffer) {
     if (this._closed) return false
+
     const msg = Buffer.concat([rlp.encode(code), data])
     const header = this._eciesSession.createHeader(msg.length)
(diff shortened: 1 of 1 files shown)


acolytec3:
This may also fix #1416 but I don't have any way to prove that.  I think if we don't see that `invalid remainder` error, #1416 can probably be closed as well.

ryanio (approved):
nice, lgtm!

holgerd77:
Great find, thanks for the deep analysis here! 😀

PR #1443, opened 2021-08-30
client: add `NEW_BLOCK` devp2p message handler

acolytec3:
Adds a handler in the `ETH` handler for the `NEW_BLOCK` message for when the client receives a newly minted block from a peer.  Currently passes this block over to the `NEW_BLOCK_HASHES` handler if the new block's hash doesn't match the current tip of the chain.


packages/client/lib/service/fullethereumservice.ts
```diff
@@ -94,6 +94,8 @@ export class FullEthereumService extends EthereumService {
       peer.eth!.send('BlockBodies', { reqId, bodies })
     } else if (message.name === 'NewBlockHashes') {
       this.synchronizer.handleNewBlockHashes(message.data)
+    } else if (message.name === 'NewBlock') {
+      this.synchronizer.handleNewBlockHashes(message.data[0])
```
ryanio, line 98: should this be the new `handleNewBlock`?

packages/client/test/integration/fullsync.spec.ts
```diff
@@ -1,7 +1,8 @@
 import tape from 'tape'
 import { Event } from '../../lib/types'
 import { wait, setup, destroy } from './util'
-
+import { Block } from '@ethereumjs/block'
+import { BN } from '../../../util/dist'
```
ryanio, line 5: from `ethereumjs-util`

packages/client/test/integration/fullsync.spec.ts
```diff
@@ -40,17 +41,35 @@ tape('[Integration:FullSync]', async (t) => {
...
+    const peer = await localServer.discover('remotePeer1', '127.0.0.2')
     await localServer.discover('remotePeer2', '127.0.0.3')
     localService.config.events.on(Event.SYNC_SYNCHRONIZED, async () => {
-      if (localService.chain.blocks.height.toNumber() === 10) {
-        t.pass('synced with best peer')
-        await destroy(localServer, localService)
-        await destroy(remoteServer1, remoteService1)
-        await destroy(remoteServer2, remoteService2)
-        t.end()
+      switch (localService.chain.blocks.height.toNumber()) {
```
ryanio, line 47: i'm kind of confused what you changed here

packages/client/lib/service/fullethereumservice.ts
```diff
@@ -94,6 +95,9 @@ export class FullEthereumService extends EthereumService {
       peer.eth!.send('BlockBodies', { reqId, bodies })
     } else if (message.name === 'NewBlockHashes') {
       this.synchronizer.handleNewBlockHashes(message.data)
+    } else if (message.name === 'NewBlock') {
+      const block = Block.fromValuesArray(message.data[0], { common: this.config.execCommon })
```
holgerd77, line 99: The exec common is generally (or maybe: in the most cases) not a good approximation for instantiating chain objects, since the exec common state is often completely unpredictable since block execution often lags behind chain updates and the HF might be behind.

In this case chain common is also not a fitting choice though since the chain sync can be behind the tip of the chain. I had a similar problem when along the recent [sendRawTransaction](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/lib/rpc/modules/eth.ts#L469) RPC method updates, I guess you can take the pattern from there with a copied common and the HF derived from the latest `syncTargetHeight`:

```typescript
const common = this.client.config.chainCommon.copy()
common.setHardforkByBlockNumber(this.service.synchronizer.syncTargetHeight)
```

packages/client/lib/sync/sync.ts
```diff
@@ -248,8 +249,21 @@ export abstract class Synchronizer {
...
+  async handleNewBlock(block: Block) {
+    const chainTip = (await this.chain.getLatestHeader()).hash()
+    // If block parent is current chain tip, insert block into chain
+    if (chainTip.toString('hex') === block.header.parentHash.toString('hex')) {
+      await this.chain.putBlocks([block])
+    } else {
+      // If block is beyond current tip, handle as `NEW_BLOCK_HASHES`
+      this.handleNewBlockHashes([[block.header.hash(), block.header.number]])
+    }
+  }
```
holgerd77, line 265: At least the `await this.chain.putBlocks([block])` part will only work for full sync and needs to be different, so I guess we need to add this into `FullSynchronizer` and add a slighly different version for the `LightSynchronizer`.

One additional big thing is the rebroadcasting functionality for the `NEW_BLOCK` message, see [Block Propagation](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#block-propagation) section from the devp2p spec:

> Newly-mined blocks must be relayed to all nodes. This happens through block propagation, which is a two step process. When a NewBlock announcement message is received from a peer, the client first verifies the basic header validity of the block, checking whether the proof-of-work value is valid. It then sends the block to a small fraction of connected peers (usually the square root of the total number of peers) using the NewBlock message.
> 
> After the header validity check, the client imports the block into its local chain by executing all transactions contained in the block, computing the block's 'post state'. The block's state-root hash must match the computed post state root. Once the block is fully processed, and considered valid, the client sends a NewBlockHashes message about the block to all peers which it didn't notify earlier. Those peers may request the full block later if they fail to receive it via NewBlock from anyone else.

Maybe you also already give this some thought here and eventually also start to add here to the PR. My first idea is that we do this in a generalized method `handleNewBlock()` in the `Sync` class and the two sub class methods are then both calling in.

I actually had to solve very similar problems with this peer-tracking data structures for the "PooledTxHashes rebroadcasting" ETH service PR #1441 I am just working on, see e.g. the data structure I added in 7e2bb4f7a636d4863beec0feec1bb5bb1bcb3e13 and refined on subsequent commits to map peers (by `PeerId`) to txs send out, and later on clean this data structure on periodic intervals.

Maybe you can directly take some larger - somewhat adopted - parts from this over to not reinvent the wheel here. I guess a lot should be fitting here.

holgerd77:
I know it is challenging, but we should really try to get the test coverage up to decent levels on PRs - especially - in the client. I guess the main mental challenge will be on this to just accept that writing these tests will take significantly more time than writing the functionality itself. 😛 But we will otherwise again and again get in unchartered waters since especially in the client overall behavior is so interconnected that we will continue to re-break things which will in worst case remain undercover for some significant amount of time and are then even harder to track down.

Again here: I have got the same problem/situation on my PR #1441 and I am also challenged by this. Will try to write some additional (integration) tests directly after writing this. Eventually you can also have a look here and take over some structures (If I am successful, lol 😜).

But we should get back to a situation where we are also good on the `codecov/patch` part from CI (maybe we should just activate as mandatory in CI to foster some more self discipline, even if this gets a bit hard sometimes).

holgerd77 (changes requested):
This needs some more iterations, great start though! 😄 Closing in on the last devp2p handlers to address! 👍

holgerd77:
(but we can also very well do the re-broadcasting part as some subsequent work/PR, whatever fit's best into the workflow. We should add additional tests though)

packages/client/test/integration/fullsync.spec.ts
acolytec3: My intent was just to have a complete integration test of the full sync process where it syncs up to the tip of the chain and then inserts a new block after it's announced by a peer.  It felt like a logical extension of this particular test though if you think that should be in a separate integration test, I'm good with that too.  I'll work on adding some more unit tests as well.

packages/client/lib/service/fullethereumservice.ts
acolytec3: Ah, thanks for the insight.  I've adjusted to an equivalent instance for this use case.

packages/client/lib/sync/sync.ts
acolytec3: >At least the await this.chain.putBlocks([block]) part will only work for full sync and needs to be different, so I guess we need to add this into FullSynchronizer and add a slighly different version for the LightSynchronizer.

On the point about the `LightSynchronizer`, I don't see anything in the LES spec about a `NEW_BLOCK` message.  So, assuming that's the case, should the `handleNewBlock` method just move entirely to `FullSynchronizer`?

I'll take a look at your work on rebroadcasting the message.  I hadn't looked that closely at the spec so didn't realize peers had any obligations around re-broadcasting `NEW_BLOCK` in addition to `NEW_BLOCK_HASHES`.

packages/client/lib/service/fullethereumservice.ts
```diff
@@ -94,6 +95,11 @@ export class FullEthereumService extends EthereumService {
       peer.eth!.send('BlockBodies', { reqId, bodies })
     } else if (message.name === 'NewBlockHashes') {
       this.synchronizer.handleNewBlockHashes(message.data)
+    } else if (message.name === 'NewBlock') {
+      const common = this.config.chainCommon.copy()
+      common.setHardforkByBlockNumber(this.chain.blocks.height)
+      const block = Block.fromValuesArray(message.data[0], { common })
+      await this.synchronizer.handleNewBlock(block)
```
holgerd77, line 102: Ah, two things here.

1. This is double decoding. The decoding and block creation is already done in the `decode()` method from `ethprotocol.ts` - you can test this by do a `console.log()` output on the `message` object. So you are actually creating the block twice. So the block creation from here should be removed and limited to the `decode()` function.
2. Setting the HF with `this.chain.blocks.height` doesn't do the trick respectively leads to wrong results, since our own chain might be behind the tip of the chain. Then you will set the wrong HF. Therefore the suggestion to use the `syncTargetHeight`. On a second I realized though that this is a) - while being a bit better, lol - also not hitting it precisely since there can be a HF switch after `syncTargetHeight` which also isn't caught with this but luckily also b) that I didn't see the forest for all the trees and now remembered that you can just use the `hardforkByBlockNumber` option which should be exactly what you want. 😛

packages/client/lib/net/protocol/ethprotocol.ts
```diff
@@ -121,6 +121,12 @@ export class EthProtocol extends Protocol {
       ],
       decode: ([reqId, bodies]: [Buffer, BlockBodyBuffer[]]) => [new BN(reqId), bodies],
     },
+    {
+      name: 'NewBlock',
+      code: 0x07,
+      encode: ([block, td]: [Block, BN]) => [block.serialize(), td.toBuffer()],
+      decode: ([block, td]: [BlockBuffer, Buffer]) => [Block.fromValuesArray(block), new BN(td)],
```
holgerd77, line 128: I was actually able to add an integration test for the first time here https://github.com/ethereumjs/ethereumjs-monorepo/pull/1441/commits/5a152ec2a9afb53552cccf9c57d3bb626a7fdd80 for the `Transactions` method respectively I digged into the test setup code base for the first time and roughly followed upon what is done there.

You can add an analogue test for the `NewBlock` method.

```typescript
peer.eth!.send('Transactions', [tx])
t.pass('handled Transactions')
```

This doesn't seem like much but it is actually already a huge win since it at least ensures that this code is running through and doesn't throw with an exception on decoding, which is the most likely thing which can happen here I guess (some room for other additional bugs though as well here, but at least some good start).

holgerd77 (changes requested):
Some change requests.

packages/client/lib/net/protocol/ethprotocol.ts
acolytec3: Okay, I'll do you one better and not just trigger the event but re-write the test to pass once the service received the decoded event :-)

packages/client/lib/service/fullethereumservice.ts
acolytec3: Ah, I see this now though odd that it wasn't throwing before.  Using just the  `hardforkByBlockNumber` parameter without providing a common instance in the `decode` function doesn't seem to work (at least on my private network where I'm way passed the London hardfork and it's throwing the error about base fee only can be set when 1559 is activated).  I need to pass a `common` instance in that aligns with the current state of the chain and I don't think the `common` that's available by default in the `ethProtocol` is the same as the chain.

acolytec3:
Okay, it looks like code coverage is happy now and everything excepting those client cli tests seem to be passing.  Ready for review again I think.  Planning to do an additional PR to handle the lower level rebroadcast logic.

holgerd77 (approved):
Nice, LGTM 👍, will merge this in now, thanks @acolytec3 for the work here! 😄

I am still not 100% sure if this covers all sync combinations with new block announcements received by `NewBlockHashes` and `NewBlock` in combination correctly, but I guess this will be easier to find out in practice along several tip-of-the-chain runs - either towards public testnets or otherwise our own local testnets set up for development.

And having this in will likely be beneficial for some other tasks like developing and testing the block builder.

PR #1444, opened 2021-08-31
client: add simple miner

ryanio:
This PR adds a `--mine` option to the client for custom private network mining.

An account should be passed into `--unlock` which the client will then prompt you for the private key (for extra caution please only use developer keys with no mainnet funds). In the future keystore functionality can be added so only the keystore password would need to be entered during runtime.

Sample run:
<details>
  <summary>Click to expand!</summary>
  
```shell
➜  client git:(client-miner) ✗ npm run client:start -- --mine --gethGenesis=poa.json --unlock=0xaaec86394441f915bce3e6ab399977e9906f3b69 --rpc

> @ethereumjs/client@0.0.6 client:start
> ts-node bin/cli.ts "--mine" "--gethGenesis=poa.json" "--unlock=0xaaec86394441f915bce3e6ab399977e9906f3b69" "--rpc"

Please enter the 0x-prefixed private key to unlock 0xaaec86394441f915bce3e6ab399977e9906f3b69:
[redacted]
INFO [08-30|20:38:15] Data directory: /Users/ryanghods/Library/Ethereum/ethereumjs
INFO [08-30|20:38:15] Initializing Ethereumjs client...
INFO [08-30|20:38:15] Full sync mode
INFO [08-30|20:38:15] Connecting to network: poa
INFO [08-30|20:38:15] Initializing VM execution hardfork=berlin
INFO [08-30|20:38:15] Latest local block: number=0 td=1 hash=352e215a... hardfork=berlin
INFO [08-30|20:38:15] TxPool started.
INFO [08-30|20:38:15] Synchronizing blockchain...
INFO [08-30|20:38:15] Started eth service.
INFO [08-30|20:38:15] Miner started. Assembling next block in 15s
INFO [08-30|20:38:15] Started rlpx server.
INFO [08-30|20:38:15] Started libp2p server.
INFO [08-30|20:38:15] Listener up transport=libp2p url=/ip4/127.0.0.1/tcp/50580/ws/p2p/12D3KooWKVqx4kbLuvX9qY3gihjvBEto8wdWzrMo5cfGukC7B7nt
INFO [08-30|20:38:15] Server listening: libp2p - /ip4/127.0.0.1/tcp/50580/ws/p2p/12D3KooWKVqx4kbLuvX9qY3gihjvBEto8wdWzrMo5cfGukC7B7nt
INFO [08-30|20:38:15] Listener up transport=rlpx url=enode://46c9a66980c45e140f317c5b269ee7159f2f4a13c3f5081b544b1e5720a97738f5f2f5f8551f90494c589327e6605d23eeb4649a0e38ad9d98fc4e76fd47ebdd@[::]:30303
INFO [08-30|20:38:15] Server listening: rlpx - enode://46c9a66980c45e140f317c5b269ee7159f2f4a13c3f5081b544b1e5720a97738f5f2f5f8551f90494c589327e6605d23eeb4649a0e38ad9d98fc4e76fd47ebdd@[::]:30303
INFO [08-30|20:38:15] RPC HTTP endpoint opened: http://localhost:8545
INFO [08-30|20:38:25] TxPool Statistics transactions=0 senders=0
INFO [08-30|20:38:31] Miner: Assembling block from 0 eligible txs
INFO [08-30|20:38:31] Miner: Sealed block with 0 txs
INFO [08-30|20:38:31] Executed blocks count=1 first=0 hash=352e215a... hardfork=berlin last=1 hash=f2c4b427... with txs=0
INFO [08-30|20:38:35] TxPool Statistics transactions=0 senders=0
INFO [08-30|20:38:45] TxPool Statistics transactions=2 senders=1
INFO [08-30|20:38:46] Miner: Assembling block from 3 eligible txs
INFO [08-30|20:38:46] Miner: Sealed block with 3 txs
INFO [08-30|20:38:46] Executed blocks count=1 first=1 hash=f2c4b427... hardfork=berlin last=2 hash=a766a0b0... with txs=3
INFO [08-30|20:38:55] TxPool Statistics transactions=4 senders=1
INFO [08-30|20:39:01] Miner: Assembling block from 7 eligible txs
INFO [08-30|20:39:01] Miner: Sealed block with 7 txs
INFO [08-30|20:39:01] Executed blocks count=1 first=2 hash=a766a0b0... hardfork=berlin last=3 hash=3dcca111... with txs=7
INFO [08-30|20:39:05] TxPool Statistics transactions=0 senders=0
INFO [08-30|20:39:15] TxPool Statistics transactions=0 senders=0
INFO [08-30|20:39:16] Miner: Assembling block from 0 eligible txs
INFO [08-30|20:39:16] Miner: Sealed block with 0 txs
INFO [08-30|20:39:16] Executed blocks count=1 first=3 hash=3dcca111... hardfork=berlin last=4 hash=8cb9d647... with txs=0
```
</details>

The `poa.json` geth genesis file I passed in was modified from `packages/client/testdata/poa.json`. I replaced my signer address in the extraData as an authorized signer as well as the last item on the `alloc` list so the address has some balance.

I then executed my external tx script that sent 10 txs to the `sendSignedTransaction` rpc endpoint which you can see above got added to the `TxPool` then got mined and subsequently removed from the `TxPool`! wohoo!


holgerd77:
@ryanio hi there, phew, wasn't aware yet how extensive this PR is, pretty impressive! 😄 Is this still "WIP" respectively what is the "WIP" part of this? Or did you just want to keep this open a bit since the code changes are so vast?

packages/client/lib/miner/miner.ts
```diff
@@ -0,0 +1,206 @@
...
+  /**
+   * Create miner
+   * @param options constructor parameters
+   */
+  constructor(options: MinerOptions) {
+    this.config = options.config
+    this.synchronizer = options.synchronizer
+    this.running = false
+    this.assembling = false
+    this.period = 15000 // default: 15s period defined in ms
```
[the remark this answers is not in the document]
holgerd77: Update: ah, ok, I've discovered the code from `start()`. Maybe rather directly put this into the constructor for visibility?

packages/client/lib/miner/miner.ts
```diff
@@ -0,0 +1,206 @@
...
+    // Use a copy of the vm to not modify the existing state.
+    // The state will be updated when the newly assembled block
+    // is inserted into the canonical chain.
+    const vmCopy = this.synchronizer.execution.vm.copy()
+
+    // If we are on the genesis block the canonical
+    // genesis state has not been initialiezd yet.
+    if (parentBlock.header.number.isZero()) {
+      await vmCopy.stateManager.generateCanonicalGenesis()
+    }
```
holgerd77, line 122: Already mentioned this somewhere on the side line, I guess this is pretty dangerous to get into some totally weird and undefined behavior when the chain is actually "existing" but we assume a non-existing chain here because sync has not yet started or something.

Geth has a dedicated `init` command for this to ensure that a new chain is not started just-by-accident:

grafik

Do we want to add something similar here, so maybe an option `--initGenesis` or something and just do the above if the option is provided and otherwise throw?

packages/client/lib/miner/miner.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    let difficulty
+    if (this.config.chainCommon.consensusType() === ConsensusType.ProofOfAuthority) {
+      // Determine if signer is INTURN (2) or NOTURN (1)
+      const signers = vmCopy.blockchain.cliqueActiveSigners()
+      const signerIndex = signers.findIndex((address) => address.equals(signerAddress))
+      const inTurn = parentBlock.header.number.addn(1).modn(signers.length) === signerIndex
+      difficulty = inTurn ? 2 : 1
+    }
```
holgerd77, line 131: I guess this whole process - 1. getting the tip of the chain block, 2. getting the active signers and 3. then determining if a signer is in turn or not would be a good candidate for a utility method on the blockchain package, something like `Blockchain.cliqueSignerInTurn(signer)` (or similar).

packages/client/lib/miner/miner.ts
```diff
@@ -0,0 +1,206 @@
...
+        difficulty,
+        gasLimit,
+      },
+      blockOpts: {
+        cliqueSigner: signerPrivKey,
+      },
+    })
+
+    let baseFee
+    if (this.config.chainCommon.isActivatedEIP(1559)) {
```
holgerd77, line 145: One nit/half-nit: this is actually one off on the transition block, right? 🤔

Not sure if it's dramatically important in this context.

packages/client/lib/miner/miner.ts
```diff
@@ -0,0 +1,206 @@
...
+                .toNumber()})`
+            )
+          }
+        } else {
+          // If there is an error adding a tx, it will be skipped
+          const hash = '0x' + txs[index].hash().toString('hex')
+          this.config.logger.error(
+            `Skipping tx ${hash}, error encountered when trying to add tx:\n${error}`
+          )
+        }
```
holgerd77, line 177: This is so so great seeing this all coming together here!!! 🎉 🎉 🎉

packages/client/lib/miner/miner.ts
```diff
@@ -0,0 +1,206 @@
...
+      index++
+    }
+    if (interrupt) return
+    // Build block, sealing it
+    const block = await blockBuilder.build()
+    this.config.logger.info(`Miner: Sealed block with ${block.transactions.length} txs`)
+    this.assembling = false
+    if (interrupt) return
+    // Put block in blockchain and remove included txs from TxPool
+    await (this.synchronizer as any).chain.putBlocks([block]) // when #1443 is merged replace this line with `await this.synchronizer.handleNewBlock(block)`
```
holgerd77, line 188: Just as a note: #1443 is now merged. 😄

holgerd77 (commented):
Hi @ryanio, just went through this for a first round, some comments, generally this looks really great! 😄 👍

holgerd77:
Update on my "is this ready?" question: I guess this is at least missing some tests to be opened up for review, right?

packages/client/bin/cli.ts
```diff
@@ -127,6 +129,16 @@ const args = require('yargs')
...
       boolean: true,
     },
+    mine: {
+      describe: 'Enable private custom network mining (beta, PoA only)',
+      boolean: true,
+      default: false,
+    },
+    unlock: {
+      describe:
+        'Comma separated list of accounts to unlock - currently only the first account is used for sealing mined blocks. Beta, you will be promped for a 0x-prefixed private key until keystore functionality is added',
```
holgerd77, line 139: Do we also want to add the warning to this prompt you wrote somewhere in the PR description, maybe even bold: "(DO NOT USE WITH ACCOUNTS HOLDING SUBSTANTIAL AMOUNTS OF ETH)"?

packages/client/bin/cli.ts
```diff
@@ -222,12 +234,56 @@ async function run() {
...
-  // Use default common configuration for specified `chain` if no custom parameters specified
-  else {
+  } else {
+    // Use default common configuration for specified `chain` if no custom parameters specified
     common = new Common({ chain: chain, hardfork: Hardfork.Chainstart })
   }
 
+  if (args.mine) {
+    if (!args.customChain && !args.gethGenesis) {
+      throw new Error('Currently mining is only supported for custom chains')
```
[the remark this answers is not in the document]
ryanio: the reason for this is that we wouldn't have a valid signer so we can't actually seal a block on an existing PoA chain, it would have to be a custom genesis block with a signer that is configured to the one passed into `--unlock`

packages/client/lib/miner/miner.ts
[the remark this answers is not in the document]
ryanio: ah good idea! not sure why I decided to set it later in `start()`, it's not like it can change during init.

I will update to this:

```suggestion
    this.period = (this.config.chainCommon.consensusConfig().period ?? 15) * 1000 // default: 15s period defined in ms
```

packages/client/lib/miner/miner.ts
```diff
@@ -0,0 +1,206 @@
...
+   */
+  private async latestBlock() {
+    return this.synchronizer.execution.vm.blockchain.getLatestBlock()
+  }
+
+  /**
+   * Sets the timeout for the next block assembly
+   */
+  private queueNextAssembly(timeout?: number) {
+    this._nextAssemblyTimeoutId && clearTimeout(this._nextAssemblyTimeoutId)
```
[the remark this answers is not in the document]
ryanio: yes it is a short-hand way of doing it: [Logical AND (&&) - Short-circuit evaluation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND#short-circuit_evaluation)

I was going back and forth between the two options and agree that it is simpler and more understandable to just stick to the `if` statement. Will update.

packages/client/lib/miner/miner.ts
ryanio: great suggestion. In this case it should be safe because we're only doing it on the VM copy to assemble a first valid block after genesis to ensure a matching end stateRoot.

This is mostly an edge case of our own VM execution setup where when starting up the blockchain hasn't been initialized properly yet - on the insertion of the first block it would run [`generateCanonicalGenesis()`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/c008e8eb76f520df83eb47c769e3a006bc24124f/packages/client/lib/sync/execution/vmexecution.ts#L100) - but we need this to happen before starting the BlockBuilder otherwise we will be working on an invalid state and conclude with an invalid stateRoot.

Alternatively, I might be able to change the line here to just run the iterator instead, will investigate!

packages/client/lib/miner/miner.ts
ryanio: yeah, sure, making a util method of this on `blockchain` sounds like a great idea.

I already had this code from `BlockHeader.validateCliqueDifficulty(blockchain)` so I can also dedupe it there when added 👍

packages/client/lib/miner/miner.ts
ryanio: great, thanks! will integrate :)

ryanio:
> Update on my "is this ready?" question: I guess this is at least missing some tests to be opened up for review, right?

tests already added in `miner.spec.ts`! check them out :)

I tried to be comprehensive with the initial set of tests, and I think I did well based on coverage so far, but I may add some more as I start testing some multi-client mining setups (`Client<>Client`, `Client<>Geth`, `Client<>Geth<>Client`, etc.) if there are any tricky scenarios encountered in my trials. I will do some of that today and tomorrow, and if all looks good, this PR will be ready for review and merge :)

packages/client/bin/cli.ts
ryanio: yeah, good suggestion, will do

packages/client/lib/miner/miner.ts
ryanio: hm, well we can't run `calcNextBaseFee()` unless the common already has 1559 activated or it will [throw](https://github.com/ethereumjs/ethereumjs-monorepo/blob/5b91ccde820dae0b4eb969eb53d88227c9582931/packages/block/src/header.ts#L665).

packages/client/bin/cli.ts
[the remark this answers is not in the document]
holgerd77: Ah, that can't be the reason for this restriction, the decision if someone has a signer on a chain is happening on some other level. Let's say Erigon has no signer on Rinkeby, that wouldn't mean that they would restrict their client in a similar way and say "we don't support Rinkeby".

At the end the list of signers in Clique is dynamic and it can theoretically be that one of our signers will be voted into the list of active signers for Clique in the future. If we are brave, we can also join the next ephemeral testnet as a signer. 😄

So this is very much a limitation happing at the wrong place and putting unnecessary limitations to the client.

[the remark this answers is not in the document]
holgerd77: (or am I missing the point here?)

packages/client/lib/@types/qheap/index.d.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * QHeap types.
+ * @types/qheap does not exist, so we define it here.
+ * https://www.npmjs.com/package/qheap
+ */
+declare module 'qheap' {
+  type QHeapOptions = {
+    comparBefore?(a: any, b: any): boolean
+    compar?(a: any, b: any): number
```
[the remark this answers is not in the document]
holgerd77: Strange. Who the hell named this? 🙂

packages/client/lib/miner/miner.ts
holgerd77: Ah, thanks for the clarification.

Yeah, I would find it more intuitive if this "progressing" is done by the iterator but - not sure - if it's not fitting our design it might not be worth it atm.

packages/client/lib/miner/miner.ts
holgerd77: Ah, I am getting what you mean, but I think we still have got some inconsistencies around this whole base fee intitialization thing and HF switching on the block builder, need to give this some short thought.

So the one thing is, that I guess that you are missing the `initialBaseFee` case here and - with our current design - you might need to replicate somewhat the code from [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/header.ts#L643) in the Block header if I get this right?

I do have to say that I find this whole design a bit unfortunate though. With this introduction of the [BlockHeader.calcNextBaseFee()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/header.ts#L663) function we have introduced some functionality which is referring to the HF state of the next block which often leads to such a "one-off" factor.

Might it improve the situation if we add a function `getNextCommon()` (or `getNextHeaderCommon()`)  to the `Block` library? This function would do a copy of the internal `Common` instance, set the HF with the `hardforkByBlockNumber` option to the HF of `block.number + 1` and return the Common instance.

We could use this on several fronts:

1. Internally in `calcNextBaseFee()` and then we could return the initial base fee for block with a block number one below the 1559 HF (this should only be called if 1559 is not already activated since otherwise we would restrict on non-blocknumber-based HF setting use cases).
2. You can use this here instead of adding the initial base fee logic manually.

Not 100% sure on this suggestion and still on the 80% level or so, might need some sleeping over.

A bit independent from the suggestion above. When I see this correctly I think the block builder is also (and generally) one off with the HF when creating a new block. this takes `this._vm._common` as a [Common input for the next block](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/buildBlock.ts#L67) and on a HF transition this will be one-off potentially leading to undefined behavior.

This might rather be a case for the `hardforkByBlockNumber` though and we should generally - as well - be careful here to not restrict on block-number-independent use cases (as Hardhat is eventually (likely?) using this for. So `hardforkByBlockNumber` should (I guess) rather be an additional block builder option passed through to the block and initially set to `false`.

Does all/most of the above make sense? 😛

holgerd77:
> I will do some of that today and tomorrow, and if all looks good, this PR will be ready for review and merge :)

Nice. 🙂

packages/blockchain/src/index.ts
```diff
@@ -1564,4 +1564,18 @@ export default class Blockchain implements BlockchainInterface {
       return false
     }
   }
+
+  /**
+   * Helper to deterimine if a signer is in or out of turn.
+   * @param signer
+   * @param blockNumber
+   */
+  cliqueSignerInTurn(signer: Address, blockNumber: BN): boolean {
```
holgerd77, line 1573: Ah, I think it would make sense if we use the head of the chain as the block number here?

holgerd77: (and actually for all other cases it generally makes no sense since `cliqueActiveSigners()` is only providing the active signer state of the latest block, just seeing on a second look)

packages/client/lib/miner/miner.ts
```diff
@@ -0,0 +1,217 @@
...
+      index++
+    }
+    if (interrupt) return
+    // Build block, sealing it
+    const block = await blockBuilder.build()
+    this.config.logger.info(`Miner: Sealed block with ${block.transactions.length} txs`)
+    this.assembling = false
+    if (interrupt) return
+    // Put block in blockchain and remove included txs from TxPool
+    await this.synchronizer.handleNewBlock(block)
```
acolytec3, line 197: Does making this call versus just `putBlock` assume that `handleNewBlock` will appropriately announce the new block to peers?  I think it makes sense that it would but just want to confirm as I'm starting to think about my next PR handling the rebroadcast mechanism for the `NEW_BLOCK` message.

packages/blockchain/src/index.ts
ryanio: ah you are right.

I think I will make blockNumber optional and fall back default to head of the chain since the miner actually needs head+1, so it would be nice to pass in that exact number for my intended usage.

I will add a big warning to the param typedoc though that this is only based on the blockchain's current state of active signers. That's definitely important to understand when using the fn.

packages/client/lib/miner/miner.ts
ryanio: that makes sense to me!

packages/client/bin/cli.ts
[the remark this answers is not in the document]
ryanio: > Let's say Erigon has no signer on Rinkeby, that wouldn't mean that they would restrict their client in a similar way and say "we don't support Rinkeby".

Makes sense, so in `assembleBlock()` instead I could check the unlocked signer account against the `activeSignersList` and return with an error if not present, since otherwise an error will be thrown when putting the block in our own chain since it won't match signer validation.

packages/client/lib/miner/miner.ts
ryanio: Yeah I think this is an improvement we can make to the design of vmexecution, to better initialize the vm to a more "ready" state when just starting with the genesis block, rather than needing a sync to start. This is uncharted territory since before we could only sync and now we can create our own chain :) so it's understandable to need some adjustments to these newer parts.

packages/client/lib/miner/miner.ts
ryanio: > So the one thing is, that I guess that you are missing the `initialBaseFee` case here and - with our current design - you might need to replicate somewhat the code from [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/header.ts#L643) in the Block header if I get this right?

yes good point, will make sure that is handled.

> I do have to say that I find this whole design a bit unfortunate though (around `calcNextBaseFee` and the checks and the like). With this introduction of the [BlockHeader.calcNextBaseFee()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/header.ts#L663) function we have introduced some functionality which is referring to the HF state of the next block which often leads to such a "one-off" factor.
> 
> Might it improve the situation if we add a function `getNextCommon()` (or `getNextHeaderCommon()`) to the `Block` library? This function would do a copy of the internal `Common` instance, set the HF with the `hardforkByBlockNumber` option to the HF of `block.number + 1` and return the Common instance.
> 
> We could use this on several fronts:
> 
> 1. Internally in `calcNextBaseFee()` and then we could return the initial base fee for block with a block number one below the 1559 HF (this should only be called if 1559 is not already activated since otherwise we would restrict on non-blocknumber-based HF setting use cases).
> 2. You can use this here instead of adding the initial base fee logic manually.
> 
> Not 100% sure on this suggestion and still on the 80% level or so, might need some sleeping over.

yeah this is a little complex, seems like a pretty good suggestion though so I'll keep thinking about it and possibly try an implementation in this PR if it helps simplify/improve things.

> A bit independent from the suggestion above. When I see this correctly I think the block builder is also (and generally) one off with the HF when creating a new block. this takes `this._vm._common` as a [Common input for the next block](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/buildBlock.ts#L67) and on a HF transition this will be one-off potentially leading to undefined behavior.
> 
> This might rather be a case for the `hardforkByBlockNumber` though and we should generally - as well - be careful here to not restrict on block-number-independent use cases (as Hardhat is eventually (likely?) using this for. So `hardforkByBlockNumber` should (I guess) rather be an additional block builder option passed through to the block and initially set to `false`.

This can already be passed in to `blockOpts` and defaults to `false`, so that should already be the existing behavior. For the miner I should probably pass this option as `true` to satisfy some of these edge cases.

packages/client/bin/cli.ts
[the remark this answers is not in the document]
holgerd77: Yes, I guess that makes more sense. 🙂 👍

packages/blockchain/src/index.ts
holgerd77: Hmm, I have the impression that I can't yet follow parts of the argumentation.

So, the miner use case is the same as the "normal" tip of the chain use case (so: latest header + 1), or am I missing something?

So for this to work we could just completely drop the `blockNumber` input, do an async function and use the `getLatestHeader()` call as you suggested?

I am generally not a fan of doing this large warning, this is just not consistent. So I think we should rather do this limited version suggested above or rather go the whole round and read the active signers directly from `_cliqueLatestSignerStates` respectively iterate through that and match with the block number provided (would still be a fan of an additional option to leave `blockNumber` out). Then we should maybe also make `CLIQUE_SIGNER_HISTORY_BLOCK_LIMIT` public and refer to this constant in the code docs and point out that the function can only be called back by this block number range.

I guess this second version would be the most flexible implementation and rather covering the broadest range of eventual use cases - we should craft a bit properly here since we'll add this to the official library API.

Ok with both versions though, we can also start with the very simple one since the extended version can also be added later without being a breaking change.

packages/client/lib/miner/miner.ts
```diff
@@ -126,47 +154,52 @@ export class Miner {
...
-        parentBlock.header.number.addn(1)
-      )
+      const inTurn = vmCopy.blockchain.cliqueSignerInTurn(signerAddress, number)
       difficulty = inTurn ? 2 : 1
     }
 
+    let baseFeePerGas
+    const londonHardforkBlock = this.config.chainCommon.hardforkBlockBN('london')
+    const isInitialEIP1559Block = londonHardforkBlock && number.eq(londonHardforkBlock)
+    if (isInitialEIP1559Block) {
```
holgerd77, line 170: Ok, we can also just stay on the status quo for now since it's not making anything structurally worse. 🙂

Will likely be worth to keep an eye on this though.

ryanio:
I will wait for #1458 to merge in first to rebase this one on since it touches some of the same code.

packages/blockchain/src/index.ts
ryanio: ok, thanks for explaining your viewpoint a bit more :) makes a lot of sense, I will just do the simpler version for now and once we have the need for looking back in history at different block numbers we can add that functionality. updated in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1444/commits/dbc215e228466bdc8a6786695aaf0fed26e857cf

holgerd77: Thanks for your flexibility here. 😛

ryanio:
great I will get this merge ready.

ryanio:
ok I have my commits ready but I also updated to the tape v5 expected use since that was leftover in my node_modules, so I will help get that PR merged in tomorrow and then update this one to go in after!

packages/vm/src/index.ts
```diff
@@ -386,8 +386,8 @@ export default class VM extends AsyncEventEmitter {
   copy(): VM {
     return new VM({
       stateManager: this.stateManager.copy(),
-      blockchain: this.blockchain,
-      common: this._common,
+      blockchain: this.blockchain.copy(),
+      common: this._common.copy(),
```
holgerd77, line 390: 👍

packages/client/lib/miner/miner.ts
```diff
@@ -0,0 +1,258 @@
...
+                .sub(blockBuilder.gasUsed)
+                .toNumber()})`
+            )
+          }
+        } else {
+          // If there is an error adding a tx, it will be skipped
+          const hash = '0x' + txs[index].hash().toString('hex')
+          this.config.logger.error(
+            `Skipping tx ${hash}, error encountered when trying to add tx:\n${error}`
+          )
```
holgerd77, line 220: Also no blocker, but are these likely errors in our code (then this would be ok) or errors with malformed txs submitted to the tx pool (then we should rather do `logger.debug()`?

packages/client/lib/miner/miner.ts
```diff
@@ -0,0 +1,258 @@
...
+    if (interrupt) return
+    // Put block in blockchain
+    await this.synchronizer.handleNewBlock(block)
+    // Remove included txs from TxPool
+    this.synchronizer.txPool.removeNewBlockTxs([block])
+    // Inform connected peers of new block
+    for (const peer of (this.synchronizer as any).pool.peers) {
+      const { td } = (this.synchronizer as any).chain.headers
+      peer.eth.send('NewBlock', [block, td])
+    }
```
holgerd77, line 239: This peer informing is also done in `synchronizer.handleNewBlock()`, so it is redundant if I am not overlooking or misreading something.

holgerd77 (changes requested):
Will just submitting a first batch of comments, this overall looks super great 🤩, especially all this code in the miner module itself.

One rather major change request until now.

Will continue in 20-30 minutes.

packages/client/lib/sync/fullsync.ts
```diff
@@ -81,6 +80,10 @@ export class FullSynchronizer extends Synchronizer {
         hash
       )} hardfork=${this.config.chainCommon.hardfork()}`
     )
+    if (this.config.mine) {
+      // Start the TxPool immediately if mining
+      this.txPool.start()
```
holgerd77, line 85: I am a bit confused by this distinction. Shouldn't the miner as well only start once sync of an eventual existing chain is completed?

packages/client/lib/sync/txpool.ts
```diff
@@ -157,14 +165,14 @@ export class TxPool {
    * @param tx Transaction
    */
   add(tx: TypedTransaction) {
-    const sender: UnprefixedAddress = tx.getSenderAddress().toString()
+    const sender: UnprefixedAddress = tx.getSenderAddress().toString().slice(2)
```
holgerd77, line 168: Thanks for updating.

packages/client/lib/sync/txpool.ts
```diff
@@ -419,6 +427,92 @@ export class TxPool {
...
 
+  /**
+   * Helper to return a normalized gas price across different transaction types.
+   * @param tx The tx
+   * @param priority If a 1559 tx and true, returns the maxPriorityFeePerGas. If false, returns the maxFeePerGas. (default: true)
+   */
+  private txGasPrice(tx: TypedTransaction, priority = true) {
+    if (tx.supports(Capability.EIP1559FeeMarket)) {
+      if (priority) {
+        return (tx as FeeMarketEIP1559Transaction).maxPriorityFeePerGas
```
holgerd77, line 438: Just re-asking here since I can't really imagine a situation where it would make sense to compare the priority fee with the classic gas price of a non-1559 tx: this is correct that it is not "max fee + max priority fee"?

packages/client/lib/sync/txpool.ts
```diff
@@ -419,6 +427,92 @@ export class TxPool {
...
+   * @param baseFee Provide a baseFee to exclude txs with a lower gasPrice
+   */
+  async txsByPriceAndNonce(stateManager: StateManager, baseFee?: BN) {
+    const txs: TypedTransaction[] = []
+    // Separate the transactions by account and sort by nonce
+    const byNonce = new Map<string, TypedTransaction[]>()
+    for (const [address, poolObjects] of this.pool) {
+      let txsSortedByNonce = poolObjects
+        .map((obj) => obj.tx)
+        .sort((a, b) => a.nonce.sub(b.nonce).toNumber())
```
holgerd77, line 471: I guess we might want to directly have this in the pool data structure itself that the txs are sorted by nonce when added? 🤔 (I somehow assumed that this would "naturally" be the case by the order they are added but - you are right - this is likely not always the case).

packages/client/test/miner/miner.spec.ts
```diff
@@ -0,0 +1,375 @@
...
+    t.notOk((miner as any).assembling, 'miner should have stopped assembling')
+    miner.stop()
+    txPool.stop()
+  })
+
+  t.test('should reset td', (t) => {
+    td.reset()
+    t.end()
+  })
+})
```
holgerd77, line 375: Fantastic test setup. Cool. 👍 😎 😀

holgerd77 (commented):
Ok, some things to address but I think I am generally through with everthing. 😄

packages/vm/src/index.ts
ryanio: yes this was much needed :)

packages/client/lib/miner/miner.ts
```diff
@@ -0,0 +1,258 @@
...
+  /**
+   * Create miner
+   * @param options constructor parameters
+   */
+  constructor(options: MinerOptions) {
+    this.config = options.config
+    this.synchronizer = options.synchronizer
+    this.running = false
+    this.assembling = false
+    this.period = (this.config.chainCommon.consensusConfig().period ?? 15) * 1000 // default: 15s period defined in ms
```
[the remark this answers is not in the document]
ryanio: great point, will add `DEFAULT_PERIOD` as the fallback for increased clarity.

packages/client/lib/miner/miner.ts
ryanio: good point i'll update to `debug`, yes these would be errors from malformed txs in the pool, ideally we would regularly check txs in the pool and validate/invalidate based on the current state as to not take up precious time during block assembly, however i figure we don't have to worry about mainnet load or speed quite yet and can optimize later as needed :)

packages/client/lib/miner/miner.ts
ryanio: yes thanks, will remove

packages/client/lib/sync/fullsync.ts
ryanio: Well if this is a private dev network scenario where you are the only miner and connect to no peers, a sync will never "start" to "complete" to then start the tx pool. It will just keep searching for potential peers. That was the scenario I was facing which is why I added this here.

packages/client/lib/sync/txpool.ts
ryanio: I think it might be more work than needed right now to maintain the order on insertion/deletion, as it might be fragile since txs can be replaced/repriced. The sort here should be relatively quick and saves headaches in other places.

packages/client/lib/sync/txpool.ts
ryanio: yes thanks you made me realize I need the `baseFee` here to properly return the normalized priority tip. Omitting it will return the max total fee, used for excluding txs below that amount for assembling the current block. You can see it's used for exactly those two uses in `txsByPriceAndNonce()`. I'll try to improve the clarity on the typedoc too, let me know if it's more understandable.

packages/client/lib/miner/miner.ts
```diff
@@ -0,0 +1,258 @@
...
+    const parentBlockHeader = this.latestBlockHeader()
+    const number = parentBlockHeader.number.addn(1)
+    const { gasLimit } = parentBlockHeader
+    const [signerAddress, signerPrivKey] = this.config.accounts[0]
+
+    // Abort if we have too recently signed
+    if (this.config.chainCommon.consensusType() === ConsensusType.ProofOfAuthority) {
+      const header = BlockHeader.fromHeaderData(
+        { number },
+        { common: this.config.chainCommon, cliqueSigner: signerPrivKey }
```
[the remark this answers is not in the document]
ryanio: well, not really needed here because this is just constructing a header to use the function `blockchain.cliqueCheckRecentlySigned()` and I doubt any future hf will have PoA changes that affect the logic in deriving that answer.  It wouldn't hurt to add it though, so I will if you think it would be more future proof.

re: using `hardforkByTD` I'm hesitant to add merge logic to the miner yet since it's still a bit undecided, once everything is more formalized I can separately work on a PR for it. I'm sure there will be lots of questions for a miner in a post-merge world - but I guess its utility would be to provide the execution environment to return a valid block for an eth2 validator? Would PoA continue incognizant of the merge hardfork? Interesting times ahead!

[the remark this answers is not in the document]
holgerd77: 🙂

packages/client/lib/miner/miner.ts
```diff
@@ -0,0 +1,254 @@
...
+    this.running = false
+    this.assembling = false
+    this.period = (this.config.chainCommon.consensusConfig().period ?? this.DEFAULT_PERIOD) * 1000 // defined in ms for setTimeout use
+  }
+
+  /**
+   * Convenience alias to return the latest block in the blockchain
+   */
+  private latestBlockHeader(): BlockHeader {
+    return (this.synchronizer as any).chain.headers.latest
```
acolytec3, line 51: Would this ever not be the `fullSynchronizer`?  Just wondering why it's casted as `any`?

ryanio: `chain` in `FullSynchronizer` is declared as private so just trying to get access to it :/

holgerd77 (approved):
Nice, looks good to me, from my side feel free to merge! 😄

Issue #775, opened 2020-06-14
Performance optimization low-hanging fruit


alcuadrado:
I'm creating this issue to list multiple easy refactors that could be implemented to improve the performance of the VM. Some of them come from [EthWork's involvement](https://www.notion.so/nomiclabs/EthWorks-results-summary-e60e2cdd500e4afe9db573fe22ee44c2). Others are mine.

  - [ ] `Common#param` should be an O(1) operation. It's called an incredible amount of times.
  
    - Analysis in https://github.com/ethereumjs/ethereumjs-monorepo/issues/775#issuecomment-860545856
  
  - [x] The VM's `Memory` class should be implemented with a Buffer

  - [x] `jumpIsValid` is a linear operation. It can be constant. *Update:* After talking with Jochem and reading evmone's code, I'm not sure if this is possible nor beneficial.

  - [x] There's a huge number of unnecessary calls to `toString()`. Sometimes it is used instead of `Buffer#equals`. Sometimes it's called to initialize a `BN` from a `Buffer`, instead of just passing the `Buffer`. There obvious instances of this in these functions:
    - `Block#validateTransactionsTrie`
    - `Block#validateUnclesHash`
    - `Block#validateUncles`
    - `Account#isContract`
    - `Account#setCode`
    - `Account#isEmpty`
    - `FakeTransaction#hash`
    - `Transaction#toCreationAddress`
    - `EEI#getExternalBalance`
    - `PUSH`'s handler: creats a string to create a BN instead of just creating the BN
    - `VM#runBlock`: multiple comparisons of Buffers by string
    - `VM#runCall`
    - `VM#runTx`: calls toString on a Buffer instead of just checking its size
    - `DefaultStateManager#accountIsEmpty`

  - [x] The opcodes `handlers` map is indexed by string, it should be done by number. 

  - [x] The opcodes `handlers` map and the `opcodes` `OpcodeList`'s one, should probably be `Map`s. The semantic of `Map` is way simpler than an accessing an object with `object[key]`. Done in #852 

  - [x] `Transaction#getDataFee` is called very frequently and super slow:

    - [x] It should be cached

    - [x] Calls `Common#param` (which is really slow) within a loop, could be done outside of it.

    - [x] Uses `BN`s in a loop while it could use normal numbers and then transform it to `BN`. -- `BN` operations allow new objects

  - [x] `Transaction#hash` should be cached

  - [x] `BlockHeader#hash` should be cached

[...]

PR #1445, opened 2021-09-01
Block, Tx: Caching for hash method

emersonmacro:
Addresses a couple items from #775, specifically "`Transaction#hash` should be cached" and "`BlockHeader#hash` should be cached".


packages/tx/src/legacyTransaction.ts
```diff
@@ -232,6 +235,9 @@ export default class Transaction extends BaseTransaction<Transaction> {
    * Use {@link Transaction.getMessageToSign} to get a tx hash for the purpose of signing.
    */
   hash(): Buffer {
```
emersonmacro, line 237: @holgerd77 One thing I noticed while working on this... The comment for the `hash` method here says it should throw if not signed (https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/legacyTransaction.ts#L231)... but it's not actually checking?

Contrast with the other transaction types, which both have an explicit check:
https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/eip2930Transaction.ts#L299
https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/eip1559Transaction.ts#L329

It's definitely possible I'm not understanding fully, but wanted to surface the question in case it is missing the `isSigned` check here.

packages/block/src/header.ts
```diff
@@ -725,6 +728,10 @@ export class BlockHeader {
    * Returns the hash of the block header.
    */
   hash(): Buffer {
+    if (Object.isFrozen(this)) {
+      return this._cachedHash as Buffer
```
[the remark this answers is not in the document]
emersonmacro: I had that originally but typescript was complaining because the type is technically `Buffer | undefined` since it's optional (since it's only used when the object is frozen in this initial implementation). I try to avoid `as` normally, definitely open to other approaches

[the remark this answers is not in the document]
emersonmacro: oh i misread your suggestion. that's a good idea, will try it

packages/tx/src/legacyTransaction.ts
ryanio: good catch, seems like it should be added to legacy.

packages/block/src/header.ts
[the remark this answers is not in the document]
ryanio: yeah it will gate it as defined with that being the if statement, just a nice simplification :)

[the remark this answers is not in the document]
ryanio: haha sorry to be picky, but then you actually don't need `Object.isFrozen(this)` then, because cachedHash will only be set if that is true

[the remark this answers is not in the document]
emersonmacro: ah yes, good catch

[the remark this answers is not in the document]
emersonmacro: updated. thanks for the quick review btw 🙂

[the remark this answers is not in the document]
ryanio: woot nice looks much better! yeah np late nights :)

packages/tx/src/eip1559Transaction.ts
```diff
@@ -228,6 +230,9 @@ export default class FeeMarketEIP1559Transaction extends BaseTransaction<FeeMark
 
     const freeze = opts?.freeze ?? true
     if (freeze) {
+      if (this.isSigned()) {
+        this._cachedHash = this.hash()
+      }
```
holgerd77, line 235: Hi there, thanks for the PR! 😄 

The cache creation for the hash shouldn't be done in the constructor but in the `hash()` function itself on the first execution (same for `Block`). Otherwise we are actually adding *overhead* for use cases where the`hash()` function is not needed.

If your problem is that the object actually will get frozen after the constructor is called:

I did some experiments on this and I guess it should work to wrap the `_cachedHash` property into some object, which will *then* still be modifiable. See the following results:

```shell
> const inner = { hash: '123' }
'use strict'
> const outer = { inner, test: 5 }
undefined
> outer.test = 4
4
> outer
{ inner: { hash: '123' }, test: 4 }
> Object.freeze(outer)
{ inner: { hash: '123' }, test: 4 }
> outer.test = 3
3
> outer
{ inner: { hash: '123' }, test: 4 }
> outer.inner.hash = '345'
'345'
> outer
{ inner: { hash: '345' }, test: 4 }
```

So as one can see the `inner` object `hash` property can still be modified on a frozen object (maybe someone else should nevertheless have an additional look here to see if this has no side effects whatsoever. Can't think of any thow atm).

Maybe choose some structure like:

```typescript
private cache = { hash: undefined }
```

Then we can eventually expand this with other stuff. 😄

packages/block/src/header.ts
```diff
@@ -295,6 +297,7 @@ export class BlockHeader {
 
     const freeze = options?.freeze ?? true
     if (freeze) {
+      this._cachedHash = this.hash()
```
holgerd77, line 300: -> to `hash()` function (see tx comment).

holgerd77 (changes requested):
Hi there, one bigger structural change request.

Additionally: can you please also add a comment to the `freeze` option comment for tx and blocks on the caching? Something like "...parameters. It also enables tx hash caching when the `hash()` function is called several times." (feel free to improve on this). 😄

packages/tx/src/eip1559Transaction.ts
emersonmacro: Sounds good - will give it a try

emersonmacro: @holgerd77 I have an implementation in `packages/block/src/header.ts` based on your suggestion - tests are passing and I think it makes sense. Wanted to get your feedback on the approach, and if it looks good, I'll add it in `tx` and add the documentation comments. Thanks!

holgerd77: Yes, that looks good, ready to proceed! 😄

packages/block/src/types.ts
```diff
@@ -101,6 +101,10 @@ export interface BlockData {
   uncleHeaders?: Array<HeaderData>
 }
 
+export interface HeaderCache {
+  hash: Buffer | undefined
+}
```
holgerd77, line 106: One small nit: please don't export this here but only define interally in the class itself (should be possible, right?). Otherwise this is getting part of the public part of the API which is not what we want.

emersonmacro: Ah ok, I don't think I had made that connection before. PR updated. So to confirm my understanding, `types.ts` in the various packages is only for things that are part of the public-facing API of that package?

holgerd77: Yes, that's correct. People use these types for typing their applications and the more "garbage" we add there, the more confusing it will be (to be fair: we are generally still half-disciplined on this and this is often not as cleaned up as it could (likely: should) be).

packages/tx/src/baseTransaction.ts
```diff
@@ -43,6 +47,8 @@ export abstract class BaseTransaction<TransactionObject> {
 
   public readonly common!: Common
 
+  public cache: TransactionCache
```
holgerd77, line 50: Same thing here with the API: this can't be `public` since it is then exposed to the library users (which shouldn't be the case). I guess `protected` should be ok here though.

Small nit, since you need to do a small update anyhow: maybe do the initialization directly here as you have done for the block library? (unless there is some reason to here in tx do it in the constructor that I am not seeing right now).

packages/tx/src/legacyTransaction.ts
```diff
@@ -232,6 +232,17 @@ export default class Transaction extends BaseTransaction<Transaction> {
...
+    if (!this.isSigned()) {
+      throw new Error('Cannot call hash method if transaction is not signed')
+    }
```
holgerd77, line 237: Maybe I am getting paranoid, but I am getting a bit shaky here and wondering if we *did* have some reason to *not* throw here yet and rather just add to the new typed txs for some weird backwards compatibility reason.

CC-ing @jochem-brouwer in here, was just seeing that you wrote a lot around the whole topic in your typed tx implementation https://github.com/ethereumjs/ethereumjs-monorepo/pull/1048. Can you eventually remember something on this? (if not we should really do here and give this a final go)

holgerd77 (changes requested):
Hi there, one small change needed.

Can you also make sure that the test coverage is above the threshold? I guess this mainly means that - if we decide to keep this new throw-clause for the legacy tx, you would need to add one test within the `legacy.spec.ts` file (there is an existing test on this in the `typedTxsAndEIP2930.spec.ts` test file for reference).

packages/tx/src/legacyTransaction.ts
jochem-brouwer: I do remember that I wanted to implement that logic first, but that broke things for some reason which is why we did not edit it. (It is a breaking change also, I guess?). I'll take a look this afternoon to figure out what it was.

emersonmacro: The thing that prompted me to add it to this PR was this comment - https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/legacyTransaction.ts#L231. I noticed the typed txs had an explicit check but the legacy tx didn't, so I thought it might just have been missed. But maybe the comment was copied over and it wasn't meant to be implemented. Happy to take out that change if that's the case.

packages/tx/src/baseTransaction.ts
emersonmacro: Updated. Not sure why I didn't follow the same approach as the block library, but fixed now

emersonmacro:
> Hi there, one small change needed.
> 
> Can you also make sure that the test coverage is above the threshold? I guess this mainly means that - if we decide to keep this new throw-clause for the legacy tx, you would need to add one test within the `legacy.spec.ts` file (there is an existing test on this in the `typedTxsAndEIP2930.spec.ts` test file for reference).

Added a test case to cover the throw clause, but it's still saying the test coverage is below threshold? Did I miss something? I tried to check the Codecov report but was getting 403 API rate limit messages.

packages/tx/src/legacyTransaction.ts
holgerd77: @jochem-brouwer Ok, thanks for the clarification. I guess this is to risky then to add on the side line here, I've reverted the change respectively replaced with a longer comment so that no future devs will stumble upon this again.

Will also add a note to the breaking changes TODO so that we will not forget on next breaking tx release.

holgerd77 (approved):
LGTM, great work Emmett! 😄

PR #1447, opened 2021-09-01
Monorepo: Update tape from v4 to v5

holgerd77:
This is a test to see what will happen to CI when updating the `tape` dependency from v4 to v5.

Please do not merge!


holgerd77:
Not so bad actually. 😄

holgerd77:
(up for everyone to be continued of course...)

holgerd77:
@emersonmacro what's the status of this? Also could you give a short summary what (mainly) needed to be adopted so that everyone gets a bit onboard what's necessary to use the new v5 version? Thanks! 😄

emersonmacro:
@holgerd77 @ryanio I thought I was on the right track for a fix on this, but now I'm not sure. When I run `npm run test:unit` in `client` locally, I get a successful run even though I see `not ok` on some of the individual tests in the logs. And I'm having some trouble with `npm run test:integration`. I can keep working on it, but also feel like I'm spinning my wheels a bit at the moment - wondering if it would make sense to have a fresh set of eyes on it, from someone who is more familiar with how these client tests are set up.

From what I can tell, the main difference is in how the callback to the `test` function is handled. In tape v4 (https://github.com/substack/tape/tree/4.x#testname-opts-cb), the docs have this:

```
If you forget to t.plan() out how many assertions you are going to run and you don't call t.end() explicitly, your test will hang.
```

so we use `t.end()` a lot in the tests. In v5 (https://github.com/substack/tape#testname-opts-cb), the docs have this:

```
If you forget to t.plan() out how many assertions you are going to run and you don't call t.end() explicitly, or return a Promise that eventually settles, your test will hang.

If cb returns a Promise, it will be implicitly awaited. If that promise rejects, the test will be failed; if it fulfills, the test will end. Explicitly calling t.end() while also returning a Promise that fulfills is an error.
```

So the callback is handled differently depending on whether it's `async` or not. My reading is that if the callback is non-async `(t) => {}`, we want to continue calling `t.end()` explicitly to end the test. If the callback is async `async (t) => {}`, we do not want to `t.end()` explicitly and instead just want to let the promise resolve.

holgerd77 (commented):
I love you all. 🥰

(PR state hasn't been changed but I am not seeing any reason why this wouldn't be ready? So I would (in this case: admin) merge here (technically not my PR any more so should be ok).

PR #1452, opened 2021-09-03
vm: debug failing nightly browser test

### LINKED ISSUE #1448: Monorepo, VM: API browser tests failing on `master` (nightly CI run)
VM API browser tests are currently failing on `master` along the [nightly CI run](https://github.com/ethereumjs/ethereumjs-monorepo/runs/3490843724?check_suite_focus=true). Here is an extract of the log:

```shell
Run npm run test:API:browser

> @ethereumjs/vm@5.5.2 test:API:browser
> karma start karma.conf.js

02 09 2021 01:09:00.706:INFO [compiler.karma-typescript]: Compiling project using Typescript 4.4.2
Error: 02 09 2021 01:09:06.462:ERROR [compiler.karma-typescript]: src/evm/evm.ts(261,11): error TS2322: Type 'unknown' is not assignable to type 'VmError | undefined'.
  Type 'unknown' is not assignable to type 'VmError'.

Error: 02 09 2021 01:09:06.464:ERROR [compiler.karma-typescript]: src/evm/evm.ts(361,11): error TS2322: Type 'unknown' is not assignable to type 'VmError | undefined'.
  Type 'unknown' is not assignable to type 'VmError'.

Error: 02 09 2021 01:09:06.465:ERROR [compiler.karma-typescript]: src/evm/interpreter.ts(116,30): error TS2571: Object is of type 'unknown'.

Error: 02 09 2021 01:09:06.466:ERROR [compiler.karma-typescript]: src/evm/interpreter.ts(116,35): error TS2571: Object is of type 'unknown'.

Error: 02 09 2021 01:09:06.467:ERROR [compiler.karma-typescript]: src/evm/interpreter.ts(120,13): error TS2571: Object is of type 'unknown'.
```

At the same time PR CI runs still pass, so it seems likely (?) that something is rather wrong with the CI setup in [vm-nightly-test.yml](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/.github/workflows/vm-nightly-test.yml), this would need some investigation and further analysis though.

Overall this should be fixed relatively soon so that nightly tests will pass again and we are not risking that other bugs remain hidden.

ryanio:
Fixes #1448 

**Note:** This PR upgrades TypeScript to latest (`typescript: v4.4.2`, `ts-node: v10.2.1`)


ryanio:
okay the problem here is that karma was using the top level `node_modules/typescript` which was on `v4` to satisfy the root typedoc plugin dependency requirement:

```json
"peerDependencies": {
        "typescript": "4.0.x || 4.1.x || 4.2.x || 4.3.x"
      }
```

I believe with npm v7 it installed these dependencies automatically (that's a bit tricky, and something to watch out for in the future). Our packages were still using TS v3.9 though because otherwise we would have run into the noImplicitAny error types solved in the following commit.

In talking with Andrew about this, we have been patiently waiting to upgrade our TypeSciprt for access to named tuples to improve the client's global event bus typings. TypeScript doesn't follow semver so 3.9 -> 4.0 actually doesn't mean anything, so I think we are safe to upgrade across our monorepo.

Just as a sanity check, after building fresh packages with the new typescript I ran ev's really nice interface diff checker script for every package that checks the local interface compared to latest on npm (`scripts/ts-interface-diff.sh`) and that didn't seem to show any issues.

holgerd77:
@ryanio thanks for the analysis, cool find! 😄

Since you guys have talked this through I would go along with the switch to TypeScript 4.

What would be the final state here for this to be ready to be reviewed and merged? Would you remove the nightly test run triggers again and otherwise the PR stays the same?

packages/block/src/header.ts
```diff
@@ -913,7 +913,7 @@ export class BlockHeader {
     try {
       const header = (await blockchain.getBlock(hash)).header
       return header
-    } catch (error) {
+    } catch (error: any) {
```
holgerd77, line 916: Where is this coming from? Is this a new linter rule? Or is it a TypeScript version related change?

ryanio: yes it was changed in v4, from my understanding it is a more strict enforcement of `noImplicitAny` (https://github.com/microsoft/TypeScript/issues/36775#issuecomment-651890339)

ryanio:
> @ryanio thanks for the analysis, cool find! 😄
> 
> Since you guys have talked this through I would go along with the switch to TypeScript 4.
> 
> What would be the final state here for this to be ready to be reviewed and merged? Would you remove the nightly test run triggers again and otherwise the PR stays the same?

thanks! great, yes just removed it.

packages/block/src/header.ts
holgerd77: Seems to have been promoted by the Ethereum community. Really strange how small the world sometimes is! 😛

holgerd77: (but makes sense, I like that)

holgerd77 (approved):
LGTM

PR #1453, opened 2021-09-03
ci: add develop branch

ryanio:
This PR adds the new `develop` branch to the ci triggers.

The only one left would be the nightly test, do we also want to run this once a day on `develop`? Since `on.schedule` doesn't allow you to specify the branch (https://stackoverflow.com/a/58800550) I am thinking of adding a matrix of `branch: [master, develop]` to the nightly jobs if we do want to add it. Or we could do it less frequently like once every 3 days, or on every push to `develop`, but I think then I'll have to create a new workflow file to customize.

If we merge this commit into master then I think we would also have to do it on `develop` so it has the updated files. I can probably re-open the same PR to merge into `develop` afterwards? Not sure if that would confuse rebasing master onto develop in the future, advice welcome.

### DIFF
--- a/.github/workflows/block-build.yml
+++ b/.github/workflows/block-build.yml
@@ -1,10 +1,8 @@
 name: Block
 on:
   push:
-    branches:
-      - master
-    tags:
-      - '*'
+    branches: [master, develop]
+    tags: ['*']
   pull_request:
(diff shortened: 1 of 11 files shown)


holgerd77 (approved):
Nice, LGTM! 😄

PR #1456, opened 2021-09-04
client: fix reqId overflow issue in ETH handler

acolytec3:
Fixes issue where geth sends a `reqId` that is too big for the JS number type and the `ETH` handler in the synchronizer errors out, causing us to lose sync with a peer.

I've run this PR on goerli for 80000 blocks syncing from a fully synced geth clients and experienced no recurrences of it.


ryanio (approved):
great, nice!

PR #1457, opened 2021-09-07
client: change event params to named tuples

acolytec3:
This changes the event parameters from mere specified types (e.g. `SOME_PEER_EVENT = [any, Peer]`) to a named tuples which are supported in Typescript 4 (so `SOME_PEER_EVENT = [msg: any, sendingPeer: Peer]`) and gives more context to the parameters associated with each event.


ryanio (approved):
finally! this looks awesome :)

holgerd77 (approved):
Oh, yes, that's actually pretty nice 😄, LGTM as well!

PR #1458, opened 2021-09-07
client -> FullSynchronizer: handle NEW_BLOCK message and rebroadcast to peers

acolytec3:
Rebroadcasts `NEW_BLOCK` and `NEW_BLOCK_HASHES` when `NEW_BLOCK` is received from a peer following the [devp2p spec](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#block-propagation)


packages/client/lib/sync/fullsync.ts
```diff
@@ -205,10 +213,62 @@ export class FullSynchronizer extends Synchronizer {
    * @param blockData `NEW_BLOCK` received from peer
    */
   async handleNewBlock(block: Block) {
-    const chainTip = (await this.chain.getLatestHeader()).hash()
+    const chainTip = await this.chain.getLatestBlock()
+
     // If block parent is current chain tip, insert block into chain
-    if (chainTip.toString('hex') === block.header.parentHash.toString('hex')) {
+    if (chainTip.header.hash().toString('hex') === block.header.parentHash.toString('hex')) {
```
ryanio, line 219: just FYI for the future, we replaced a lot of `buffer1.toString('hex') === buffer2.toString('hex')` in the monorepo previously since `toString` is a rather slow operation and comparison, the preferred faster way is to use `buffer1.equals(buffer2)`
```suggestion
    if (chainTip.header.hash().equals(block.header.parentHash)) {
```

packages/client/lib/sync/fullsync.ts
```diff
@@ -205,10 +213,62 @@ export class FullSynchronizer extends Synchronizer {
...
+        const currentPeer = this.pool.peers[x]
+        if (!this.blocksKnownByPeer.has(currentPeer.id)) {
+          // Create new map of blocks known by peer if none exists
+          this.blocksKnownByPeer.set(currentPeer.id, [])
+        }
+
+        const knownBlocks = this.blocksKnownByPeer.get(currentPeer.id) ?? []
+
+        if (
+          knownBlocks.filter((sentBlock) => sentBlock.hash === block.hash().toString('hex'))
```
ryanio, line 237: can use `buffer.equals` here as well:
```suggestion
          knownBlocks.filter((sentBlock) => sentBlock.hash.equals(block.hash()))
```
but I think it would be faster to use `knownBlocks.find()` rather than `filter().length > 0`

packages/client/lib/sync/fullsync.ts
```diff
@@ -8,6 +8,11 @@ import { VMExecution } from './execution/vmexecution'
 import { TxPool } from './txpool'
 import { Event } from '../types'
 
+interface sentBlock {
+  hash: string
```
ryanio, line 12: maybe store this as a `Buffer`? since that is our typical internal value for `hash`

packages/client/lib/sync/fullsync.ts
```diff
@@ -17,9 +22,12 @@ export class FullSynchronizer extends Synchronizer {
 
   public txPool: TxPool
 
+  private blocksKnownByPeer: Map<string, sentBlock[]>
```
ryanio, line 25: now that we are on the latest TS version, is it possible to annotate `string`? Just so it's easier to read/understand/use. If not maybe we can do here `type PeerId = string` so it can be written as `Map<PeerId, sentBlock[]>`, that seems a lot more friendly to me.

ryanio: might this also be more clear if named `newBlocksKnownByPeer` ?

packages/client/lib/sync/fullsync.ts
holgerd77: `sentBlock` -> `SentBlock`

holgerd77: @ryanio: just as some note: I had taken the [same string-based datastructure](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1441/files#diff-257377ef4396527158bfb57c5d9b985465a25b021e4e0083de5af8460ee446abR24) on the tx pool since there it was easier to chain on requests  since the hash is used also as string as a key for other data structure.

Not sure if we want to deviate here, seems rather getting a bit inconsistent to me. That said: I am far from sure that I've introduced the optimal data structures along the tx pool PR. So if we generally want to improve here, we can very well do.

packages/client/lib/sync/fullsync.ts
holgerd77: > now that we are on the latest TS version, is it possible to annotate string

A bit - a tiny bit 😋 - similar direction here: this is very analogue to problems/solutions in the tx pool implementation. My solution there were these types like `UnprefixedHash` and the like. We should just generally stay consistent here, no matter what we do.

packages/client/lib/sync/fullsync.ts
```diff
@@ -8,6 +8,11 @@ import { VMExecution } from './execution/vmexecution'
 import { TxPool } from './txpool'
 import { Event } from '../types'
 
+interface sentBlock {
+  hash: string
+  timeSent: number
```
holgerd77, line 13: Also here: please remain consistent on naming with the tx pool naming (there: `added`). This can very well mean that you rename the tx pool name depending on what you find more appropriate, we just shouldn't name differently for the same things.

packages/client/lib/sync/fullsync.ts
acolytec3: Looks like we can't add an additional type annotation to the type (at least not based on my small experimentation) so I went with the type alias suggested above (`peerId = string`).

packages/client/lib/sync/fullsync.ts
acolytec3: I've updated this to `handledObject` to match the tx pool.

packages/client/lib/sync/fullsync.ts
acolytec3: I switched these to the `buffer.equals...` approach suggested by @ryanio Happy to have it either way but I'm all for any performance optimizations we can do so would tend to go with `buffer.equals` over converting everything to strings if that's more performant.

acolytec3:
Pending successful CI runs, this is ready for review (again).

packages/client/lib/sync/fullsync.ts
```diff
@@ -8,6 +8,12 @@ import { VMExecution } from './execution/vmexecution'
 import { TxPool } from './txpool'
 import { Event } from '../types'
 
+interface handledObject {
```
holgerd77, line 11: Please keep interface and type names capitalized as per convention (see e.g. [styleguide here](https://github.com/basarat/typescript-book/blob/master/docs/styleguide/styleguide.md#interface)).

Will do these last updates on this locally and then push for review.

packages/client/lib/sync/fullsync.ts
```diff
@@ -204,11 +213,57 @@ export class FullSynchronizer extends Synchronizer {
...
+          x++
+          continue
+        }
+
+        knownBlocks.push({ hash: block.hash(), added: Date.now() })
+        this.newBlocksKnownByPeer.set(currentPeer.id, knownBlocks)
+        currentPeer.eth?.send('NewBlock', [block, this.chain.blocks.td])
+
+        x++
+      }
```
holgerd77, line 253: Same here, this can be extracted as an [addToKnownByPeer()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/lib/sync/txpool.ts#L226) function analogue to the tx pool.

These extractions have the advantage that this outer `handleNewBlock()` function becomes better readable and shorter and it becomes furthermore also possible to do more granular testing.

(sorry, don't want per-se impose my introduced structures on you, just think in this case it does just make sense to align since the functionality is so similar and the client code will get easier to read if we align on things like this).

packages/client/lib/sync/fullsync.ts
```diff
@@ -8,6 +8,12 @@ import { VMExecution } from './execution/vmexecution'
 import { TxPool } from './txpool'
 import { Event } from '../types'
 
+interface HandledObject {
+  hash: Buffer
+  added: number
+}
```
holgerd77, line 14: And a final "do it like in the tx pool" note 😛: this data structure also needs a `cleanup()` call since otherwise it is growing indefinitely, see tx pool [cleanup()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/lib/sync/txpool.ts#L396) for inspiration.

My thoughts on why to clean up by time (and not e.g. on the peers in the pool) was that this should be more robust if we e.g. loose connection to a peer and then re-connect a bit later on. Then this data structure is still assured to be intact and the known blocks for a peer still there.

holgerd77 (changes requested):
Ok, I have got several structural change suggestions, mainly targeted to further break down the functionality and better align with the tx pool.

Please make sure to do a first `git pull` before starting to work on this since I pushed some changes.

holgerd77:
Ok, picked this up and did some further improvements and clean up work. I might have also fixed this inconsistency bug along syncing with efe1b8f35e70874b99966bfd407ec82fce1b5fec, not confirmed though.

Will work on the tests here tomorrow and try to bring the test coverage above the threshold.

acolytec3:
I ran my private network again with the latest commits  unfortunately it looks like the tasks enqueued by `NEW_BLOCK_HASHES` still aren't getting processed.  If you put a console log [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/a3984a3b80bccbfb376918c51556a845afbecad9/packages/client/lib/sync/fetcher/fetcher.ts#L214-L214) in the `fetcher` and then watch the logs, you'll see the same job get reprocessed over and over again.  Whenever the `NEW_BLOCK_HASHES` message is received and processed, the `enqueueTask` call is inserting the new block into the fetcher queue but it evidently is getting reset somehow and its always the last task that was created prior to the chain initially getting in sync that is processed and as a result, the new block that was announced is never getting inserted into the blockchain.  Will continue to research as time allows.

ryanio:
This seems like a great candidate for a new integration test, to test the nuanced behaviors when arriving to and continuing to follow the tip of the chain. It gets rather delicate and I think it would give us increased confidence in these newer abilities/scenarios of the client.

Thanks @holgerd77 for continuing the work, I will catch up on your commits and see if I can contribute as well. I am going to start with some work in my miner PR this morning and will come back to this for the second half of my day.

holgerd77:
Will continue to work on this this morning.

I will try to bring the test coverage (mostly with non-integration tests I guess for now) to a mergeable level and then otherwise merge this in. I guess this can be a lot better debugged along the miner work from Ryan (or: with having this work in to create local test scenarios). And then this is not further holding up the miner PR.

holgerd77 (approved):
LGTM 😄, see previous comment for approval comment.

acolytec3:
> Ok, this should do it, I fixed another bug in [78515a8](https://github.com/ethereumjs/ethereumjs-monorepo/commit/78515a83d9caf701e42b3de537f72c33c218e04a) where existing block hashes in addToKnownByPeer() were not filtered correctly and added additional test coverage so that the major code paths are now covered. This might have solved the the outstanding re-broadcasting bug (have not tested). At least this should be some solid ground to work on and eventually do continued refinements.

Awesome, I'll fire up my private network a little later and see how it goes.

Issue #1412, opened 2021-08-19
Blockchain, Common: The Merge transition detection


holgerd77:
I was thinking a bit how we can integrate the transition detection and switch from [here](https://eips.ethereum.org/EIPS/eip-3675#definitions) with the terminal total difficulty and the terminal PoW block into our library set.

My current state of thinking is that we can add the `TERMINAL_TOTAL_DIFFICULTY` difficulty as a parameter to the hardfork file in`Common` as a PoW consensus config paramter (e.g. in [mainnet.json](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/chains/mainnet.json)). I was first thinking about adding this to the HF file but one thing is that this a) has to available actually before Common switches to the Merge and b) the value of this will be different on a per-chain basis. And at the end this is a PoW parameter (this is the moment where the current PoW usage will end).

The `Blockchain` library can then use this parameter to track the total difficulty value. Once the `pow_block.total_difficulty >= TERMINAL_TOTAL_DIFFICULTY` condition is satisfied for a block added to the chain, the `Blockchain` library emits and event `POW_TERMINAL_TOTAL_DIFFICULTY_REACHED` or the like which can be caught by the client and the client can react accordingly.

(this would mean that we would need to extend the `Blockchain` class as an EventEmitter - I remember from when we switched on this for the `Common` library for a new (and first) `hardforkChanged` event this was more or less side effects free - but just to note)

Does this make sense?

I guess we can actually already start integrating this adding this to blockchain and adding 2-3 tests for this but for now excluding (most) of the client part, maybe giving this a first stub to get a first feeling where this could be integrated and give this some first thought what the client would do next (things like not listening to `NEW_BLOCK_HASHES` any more I guess).

holgerd77:
(please read on site, several post-submit changes)

acolytec3:
[...]

PR #1473, opened 2021-09-15
Common: allow setting and retrieving HF by total difficulty

holgerd77:
Ok, this builds upon the ideas I outlined in #1412 on how to integrate the total diffculty HF setting in Common. I refined the original idea of adding a separate `getHardforkByTD()` function and instead added an optional `td` (couldn't decide if to use the long or short form here, also possible to switch to `totalDifficulty`) parameter to the `getHardforkByBlockNumber()` and `setHardforkByBlockNumber()` functions. `Common` is then deciding on its own depending on the chain HFs defined which HF to choose. This will likely be easier to use upstream since we can just add the TD in addition to the block number - where relevant - and Common should still (hopefully 😋) do the right thing respectively make the correct HF choice.

I will keep this as WIP for now, I would also like to do the upstream integrations (e.g. in Block and Blockchain), so that we can see if this plays out as intended.

If we want we can switch on the names of these functions to `getHardforkByBlockNumberOrTD()` wrapper functions and deprecate the old function names. My personal feeling would be that the current names are fine and it would be a bit of an overkill. No very strong opinion on this though.


holgerd77:
(lot's of post edits, please read on site)

holgerd77:
Ok, ready for review. 😀

packages/block/src/block.ts
```diff
@@ -46,16 +46,19 @@ export class Block {
 
     // parse uncle headers
     const uncleHeaders = []
+    const uncleOpts: any = {
```
ryanio, line 49: did it complain if typed as `BlockOptions` instead of `any`?

packages/block/src/block.ts
```diff
@@ -46,16 +46,19 @@ export class Block {
...
+      hardforkByBlockNumber: true,
+      ...opts, // This potentially overwrites hardforkByBlocknumber
+      // Use header common in case of hardforkByBlockNumber being activated
+      common: header._common,
+      // Disable this option here (all other options carried over), since this overwrites the provided Difficulty to an incorrect value
+      calcDifficultyFromHeader: undefined,
+    }
+    if (uncleOpts.hardforkByTD) {
+      delete uncleOpts.hardforkByBlockNumber
```
gabrocheleau, line 58: I'm a little confused by:

- The initial setting of `hardforkByBlockNumber` as true; followed by
- The potential overwriting by `...options`; followed by
- The potential deletion when `uncleOpts.hardforkByTD` is `truthy`

Do we always want `uncleOpts.hardforkByTD` to take precedence over `hardforkByBlockNumber`?

packages/vm/src/runBlock.ts
```diff
@@ -114,8 +114,8 @@ export default async function runBlock(this: VM, opts: RunBlockOpts): Promise<Ru
    */
   await this._emit('beforeBlock', block)
 
-  if (this._hardforkByBlockNumber) {
-    this._common.setHardforkByBlockNumber(block.header.number.toNumber())
+  if (this._hardforkByBlockNumber || this._hardforkByTD) {
+    this._common.setHardforkByBlockNumber(block.header.number.toNumber(), this._hardforkByTD)
```
ryanio, line 118: we don't need `number.toNumber()` here anymore since we expanded the input params for `setHardforkByBlockNumber` it accepts BN now.

```suggestion
    this._common.setHardforkByBlockNumber(block.header.number, this._hardforkByTD)
```

packages/block/src/header.ts
```diff
@@ -219,8 +219,15 @@ export class BlockHeader {
...
-      this._common.setHardforkByBlockNumber(number.toNumber())
+    if (options.hardforkByBlockNumber !== undefined && options.hardforkByTD !== undefined) {
+      throw new Error(
+        `The hardforkByBlockNumber and hardforkByTD options can't be used in conjunction`
+      )
+    }
+
+    const hardforkByBlockNumber = options.hardforkByBlockNumber ?? false
+    if (hardforkByBlockNumber || options.hardforkByTD) {
+      this._common.setHardforkByBlockNumber(number.toNumber(), options.hardforkByTD)
```
ryanio, line 230: we don't need `number.toNumber()` here anymore since we expanded the input params for `setHardforkByBlockNumber` it accepts BN now.

```suggestion
      this._common.setHardforkByBlockNumber(number, options.hardforkByTD)
```

packages/block/src/block.ts
```diff
@@ -107,17 +110,19 @@ export class Block {
 
     // parse uncle headers
     const uncleHeaders = []
+    const uncleOpts: any = {
```
ryanio, line 113: (same question here if you tried to type this as `BlockOptions`, could give us some better security when working on the opts in the future)

packages/block/src/types.ts
```diff
@@ -28,6 +28,17 @@ export interface BlockOptions {
...
   hardforkByBlockNumber?: boolean
+  /**
+   * Determine the HF by total difficulty (Merge HF)
+   *
+   * This option is a superset of `hardforkByBlockNumber` (so only use one of both options)
+   * and determines the HF by both the block number and the TD.
+   *
+   * Since the TD is only a threshold the block number will in doubt take precedence (imagine
+   * e.g. both Merge and Shanghai HF blocks set and the block number from the block provided
+   * pointing to a Shanghai block: this will lead to set the HF as Shanghai and not the Merge).
```
ryanio, line 39:
```suggestion
   * pointing to a Shanghai block: this will set the HF as Shanghai and not the Merge).
```

packages/block/test/block.spec.ts
```diff
@@ -5,6 +5,7 @@ import { Block, BlockBuffer, BlockHeader } from '../src'
 import blockFromRpc from '../src/from-rpc'
 import { Mockchain } from './mockchain'
 import { createBlock } from './util'
+import testnetMerge from './testdata/testnetMerge.json'
```
ryanio, line 8: oh nice i guess we don't need the `import * as` syntax as used for the other json files, would be nice to update the lines below so they all match :)

ryanio (approved):
a few small comments but otherwise lgtm! i like how rather unintrusively this was accomplished :)

packages/block/test/block.spec.ts
gabrocheleau: I _think_ that these are two different things, `import * as foo from 'foo'` is when you import all modules with an alias (so then you can access modules with `foo.module1; foo.module2;` etc.) and `import foo from 'foo'` is when `foo` provides a default export

gabrocheleau: Actually I take that back, it looks like both work, at least for JSON files!

packages/block/test/block.spec.ts
```diff
@@ -61,6 +62,61 @@ tape('[Block]: block functions', function (t) {
...
+    )
+    st.equal(block._common.hardfork(), 'berlin', 'should use hardforkByBlockNumber option')
+
+    block = Block.fromBlockData(
+      {
+        header: {
+          number: 20, // Future block
+        },
+      },
+      { common, hardforkByTD: 5001 }
```
gabrocheleau, line 85: Might it also be worth adding a test case for `TD` equal to the threshold (so 5000 in this case)?

packages/blockchain/src/db/manager.ts
```diff
@@ -166,8 +160,12 @@ export class DBManager {
...
       }
     }
-    const blockData = [header, ...body] as BlockBuffer
-    const opts = { common: this._common, hardforkByBlockNumber: true }
+    const blockData = [header.raw(), ...body] as BlockBuffer
+    let parentTd
+    if (!number.eqn(0)) {
+      parentTd = await this.getTotalDifficulty(header.parentHash, number.subn(1))
+    }
+    const opts = { common: this._common, hardforkByTD: parentTd }
```
gabrocheleau, line 168: Shouldn't `hardforkByTD: parentTd` instead be `hardforkByTD: parentTd.add(header.difficulty)`?

My reasoning is that the current totalDifficulty of the block (and the one that's used for the fork choice) is the parent total difficulty + the current block difficulty.

packages/blockchain/src/db/manager.ts
```diff
@@ -184,7 +182,12 @@ export class DBManager {
    */
   async getHeader(blockHash: Buffer, blockNumber: BN) {
     const encodedHeader = await this.get(DBTarget.Header, { blockHash, blockNumber })
-    const opts = { common: this._common, hardforkByBlockNumber: true }
+    let parentTd
+    if (!blockNumber.eqn(0)) {
+      const parentHash = await this.numberToHash(blockNumber.subn(1))
+      parentTd = await this.getTotalDifficulty(parentHash, blockNumber.subn(1))
+    }
+    const opts = { common: this._common, hardforkByTD: parentTd }
```
gabrocheleau, line 190: Same comment as above concerning `hardforkByTD`

gabrocheleau (commented):
Added a few comments :)

holgerd77:
I'll merge this in since I can't work on this today since my younger one is a bit ill. Will address your points though now or later.

packages/blockchain/src/db/manager.ts
holgerd77: Then we wouldn't need to get the parent TD in the first place and could directly take the one from the block. 😀

But I was confused by this as well and actually first had this differently. Still a bit unsure but I guess it doesn't make sense to interpret in a way that the - theoretical - POW difficulty of a block should still be counted in to then determine the switch of the very same block to POS and then rather set this difficulty to 0.

So I guess this has to be the parent TD by definition. Let me know if I'm missing something.

packages/block/test/block.spec.ts
holgerd77: Test cases can and should generally be substantially expanded on the continued road to POS, especially in the blockchain library.

packages/block/src/block.ts
holgerd77: This code improvement works but is very messy and likely suboptimal, I was already pretty sleepy. 🙂

Would be great if someone could improve this at done point.

gabrocheleau: I'm working on a smallish Type-improvements PR, I'll see if I can easily fit that in!

packages/block/src/block.ts
holgerd77: hardforkByTD is a superset of the old option, so that will always imply the old behavior (and add some new one).

That said: I didn't really understand this code in the first place, so my modest goal was just to preserve the existing functionality by doing it the most native and unfancy way. 😜

packages/block/src/block.ts
holgerd77: Cool. 🙂

gabrocheleau: Implemented in #1480

holgerd77:
Generally our Blockchain library is in a very bad shape and we should give this some dedicated (but at the end rather small) internal refactoring rounds (not necessarily breaking but rather cleaning things up and making stuff more consistent).

Will try to give this some write-up with a more concrete Todo list at some point.

packages/blockchain/src/db/manager.ts
gabrocheleau: Oh, I'm probably missing something basic then... Is this always going to be called in the context of PoS? Your answer seems to imply that but I can't see why that would be the case in the code.

holgerd77: Hmm, what do you mean with "context of PoS"? This is just trying to determine the correct HF for a block or header. The `hardforkByTD` option (respectively the total difficulty part of it) is generally - practically - only relevant for now for the Merge HF switch, is this what you mean with PoS context?

And the total difficulty provided for this option should just be the correct total difficulty which is used (per spec) for the decision if the block we be seen as a PoS block. I am arguing that this must be logically the parent TD, but I might also be missing something here.

PR #1475, opened 2021-09-15
Ethash: add CPU miner

jochem-brouwer:
This PR adds a simple CPU miner to Ethash. This could be used for testing purposes, for instance, in the client.

Still WIP, I want to add some things, specifically also an interface which also seals a block and does not just return the nonce/mixHash (so it should return a block or block header). I also have to tests these timers, since the only async part of the miner is the epoch loading. I am not sure how JavaScript VM handles deterministic loops. Does it keep looping or sometimes jump out and check if there are other floating promises? No idea.

What is not very nice is that this cannot run in parallel; Ethash first has to load an epoch and then `verifyPOW` further logic expects that this epoch is still cached and has not changed. (I also wonder - does that mean we might have a race condition somewhere, especially if we start dumping large amounts of blocks in the blockchain - if that is done in parallel, for instance because we just got two different chains in from the client, then the POW check might suddenly change the epoch while it is still validating?)


jochem-brouwer:
This is ready for review. Once #1444 is merged, will open a new PR to also integrate PoW mining in the client.

packages/ethash/src/index.ts
```diff
@@ -11,9 +11,127 @@ import {
...
+  private stopMining: boolean
+
+  /**
+   * Create a Miner object
+   * @param mineObject - The object to mine on, either a `BlockHeader` or a `Block` object
+   * @param ethash - Ethash object to use for mining
+   */
+
+  constructor(mineObject: BlockHeader | Block, ethash: Ethash) {
+    if (mineObject.constructor === BlockHeader) {
```
ryanio, line 40: just curious, is there a reason you chose to compare it this way over using `instanceof`?

packages/ethash/src/index.ts
```diff
@@ -11,9 +11,127 @@ import {
...
+   */
+  async mine(iterations: number = 0): Promise<undefined | BlockHeader | Block> {
+    const solution = await this.iterate(iterations)
+
+    if (solution) {
+      if (this.block) {
+        const data = this.block.toJSON()
+        data.header!.mixHash = '0x' + solution.mixHash.toString('hex')
+        data.header!.nonce = '0x' + solution.nonce.toString('hex')
+        return Block.fromBlockData(data)
```
ryanio, line 74: You should be able to just pass these in as Buffers, and I think you should specify the `common` so it is extended from the previous block.

```suggestion
        data.header!.mixHash = solution.mixHash
        data.header!.nonce = solution.nonce
        return Block.fromBlockData(data, { common: (this.block as any)._common })
```

jochem-brouwer:
Addressed these suggestions @ryanio :smile:

packages/ethash/package.json
```diff
@@ -38,6 +38,7 @@
   },
   "devDependencies": {
     "@ethereumjs/block": "^3.4.0",
+    "@ethereumjs/common": "^2.4.0",
```
ryanio, line 41: hmm do common and block need to be regular dependencies since they're used in the source now? this might be a good use for peerDependencies.

jochem-brouwer: Oh right, it is also in the source, so it should not be a devDependency. What exactly is a peer dependency? Have never encountered this.

holgerd77: `Common` is not in the source - unless I am overlooking something - there is just one place where it is taken from a `block` instance.

`Block` is in the source, but it actually seems that `Block` types had been in the source before? 🤔 So this would rather be a bug fix I guess?

Not sure, I would tend to not start experimenting with peer dependencies now here on the side line but rather do in the "classic" way it is currently done in all other libraries and then execute on a common strategy (if we decide to) somewhat later.

jochem-brouwer:
I have created a dirty fuzzer and ran 1000 random inputs against Infura and checked if the VM reported the same outcome.

I have tested rounds up to `2^16-1`, if I use more rounds (one could theoretically do `2^32-1` rounds) I run against out of gas at Infura.

packages/ethash/package.json
ryanio: @holgerd77 ah right thanks for thinking it through. sounds good.

I believe `Block` was fine as a devDep when it was just used for its types because it would have only been needed at build time for creating the `.d.ts` files.

So here I think we can move `@ethereumjs/block` to a normal dep, and common can remain a devDep.

> What exactly is a peer dependency? Have never encountered this.

@jochem-brouwer check out #1200 for a deep dive on the issue. [This article](https://lexi-lambda.github.io/blog/2016/08/24/understanding-the-npm-dependency-model/) sums it up well, which is basically: if the dep is used in the exported public API then when specified as a peerDep it will be more compatible with varying versions in node_modules from other installed libraries since it will opt to use those first before installing its own, so it reduces finicky type mismatches from very small differences in types.

ryanio:
Thanks @jochem-brouwer for working on this, looks really clean and excited to get this in the client and getting a PoW compatible miner running :) Then we should be in a really good position to keep working on merge transition PoW->PoS scenarios and tests.

packages/ethash/package.json
holgerd77: Ok, I updated this according to the discussion. Also re-checked our [dependency relationship diagram](https://github.com/ethereumjs/ethereumjs-monorepo#package-dependency-relationship) to make sure that we are not introducing a circular dependency here.

packages/ethash/src/index.ts
```diff
@@ -11,9 +11,127 @@ import {
...
+  nonce: Buffer
+}
+
+class Miner {
+  private blockHeader: BlockHeader
+  private block?: Block
+  private ethash: Ethash
+
+  private currentNonce: BN
+  private solution?: Solution
```
holgerd77, line 28: @jochem-brouwer I wouldn't want to unnecessarily expose all the class members and therefore make it part of the public API and rather keep all the above as `private`, have updated.

Let me know if there are good reasons to keep any of these as `public`.

holgerd77 (approved):
LGTM, pretty cool, thanks Jochem! 😀

packages/ethash/src/index.ts
jochem-brouwer: I'd say that the `solution` would be handy to keep `public`, then if a solution is found it can be taken from the Miner object directly. (One can also just re-call the miner function and just set the iterations to 0, it will immediately return the cache).

holgerd77: Ok, I am preparing a release PR for the various libraries right now, will add a small fix along, guess that should be ok.

PR #1477, opened 2021-09-16
util: return undefined or null from toType

### LINKED ISSUE #1472: Util: toType() behaviour for `undefined` and `null` values
Just stumbled upon this along some `Common` library Merge work and want not to side-introduce, so I'll work around that in my own PR and post here instead so that this can be tackled independently.

So `toType()` is currently returning `<BN: 0>` values for `undefined` and `null`, see:

```shell
> toType(3, TypeOutput.BN)
<BN: 3>
> toType(undefined, TypeOutput.BN)
<BN: 0>
> toType(null, TypeOutput.BN)
<BN: 0>
```

(this should also be checked for conversion to other types if existant)

I would assume that this is error prone and shouldn't be the case (as we have seen along some `null`, 0, undefined problems within the tx library along setting the v, r, s values e.g.).

So this should very likely return:

- `undefined` -> `undefined`
- `null` -> `null`

Anything else?

ryanio:
Fixes #1472 

Returns `null` or `undefined` if passed in to `toType`, and adds function overloads to specify those as the output types if passed in as the input.


holgerd77 (approved):
Nice, this looks good, thanks for addressing so quickly! 😄 There is a slight risk of breaking on the TypeScript level I guess (by now having `null` and `undefined` as possible return types of `toType`, woul be happy on a short correction if I misinterpret the typings though). But I guess this is bearable since I would very much assume that `toType` is not that wildly used and at the end this is somewhat of a necessary bug fix here which we need to make at some point.

holgerd77:
The Hardhat tests are [failing](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1477/checks?check_run_id=3649607080) on this, will nevertheless merge since unrelated.

If I interpret the test run correctly: it seems that our own tests are *all* (run) along publishing to the virtual registry? Is this correct? I guess this is not really necessary (or is it?) since we are running the monorepo tests already on the PR separately and we could save a lot of time and ressources if we skip? 🤔

PR #1479, opened 2021-09-16
client: Bug fix in logic for adding known blocks to peers

acolytec3:
Fixes bug in the logic for determining if a `NEW_BLOCK` is already known to a given peer

### DIFF
--- a/packages/client/lib/sync/fullsync.ts
+++ b/packages/client/lib/sync/fullsync.ts
@@ -220,7 +220,7 @@ export class FullSynchronizer extends Synchronizer {
     }
 
     const knownBlocks = this.newBlocksKnownByPeer.get(peer.id) ?? []
-    if (knownBlocks?.filter((knownBlock) => knownBlock.hash.equals(blockHash))) {
+    if (knownBlocks?.filter((knownBlock) => knownBlock.hash.equals(blockHash)).length > 0) {
       return true
     }
     knownBlocks.push({ hash: blockHash, added: Date.now() })
@@ -253,7 +253,6 @@ export class FullSynchronizer extends Synchronizer {
       // Don't send NEW_BLOCK announcement to peer that sent original new block message
       this.addToKnownByPeer(block.hash(), peer)
     }
-
     try {
       await block.header.validate(this.chain.blockchain)
     } catch (err) {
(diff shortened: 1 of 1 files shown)


packages/client/lib/sync/fullsync.ts
```diff
@@ -220,7 +220,7 @@ export class FullSynchronizer extends Synchronizer {
     }
 
     const knownBlocks = this.newBlocksKnownByPeer.get(peer.id) ?? []
-    if (knownBlocks?.filter((knownBlock) => knownBlock.hash.equals(blockHash))) {
+    if (knownBlocks?.filter((knownBlock) => knownBlock.hash.equals(blockHash)).length > 0) {
```
ryanio, line 223: This whole method seems a bit repetitive in the code, I would rewrite it like this:

```typescript
const knownBlocks = this.newBlocksKnownByPeer.get(peer.id) ?? []

if (knownBlocks.find((known) => known.hash.equals(blockHash))) {
  return true
}

knownBlocks.push({ hash: blockHash, added: Date.now() })
this.newBlocksKnownByPeer.set(peer.id, knownBlocks)
return false
```

ryanio (approved):
thanks, lgtm!

packages/client/lib/sync/fullsync.ts
```diff
@@ -214,13 +214,8 @@ export class FullSynchronizer extends Synchronizer {
...
    */
   private addToKnownByPeer(blockHash: Buffer, peer: Peer): boolean {
-    if (!this.newBlocksKnownByPeer.has(peer.id)) {
-      this.newBlocksKnownByPeer.set(peer.id, [{ hash: blockHash, added: Date.now() }])
-      return false
-    }
-
     const knownBlocks = this.newBlocksKnownByPeer.get(peer.id) ?? []
-    if (knownBlocks?.filter((knownBlock) => knownBlock.hash.equals(blockHash)).length > 0) {
+    if (knownBlocks.find((knownBlock) => knownBlock.hash.equals(blockHash))) {
```
holgerd77, line 218: Post merge note: sorry for that, I accidentally introduced this bug in the original code version along doing some refinement work on #1458 since I misread the code.

acolytec3:
No worries! Also, these updates still haven't resolved the issue where a
client who finishes syncing and then begins receiving `NEW_BLOCK_HASHES`
doesn't restart the fetcher. I will continue to research and hopefully find
a fix this week.

holgerd77:
I guess the the intention along the current implementation is rather that the fetcher should just continue to work (by taking in the new tasks from the incoming `NEW_BLOCK_HASHES` messages) and not that it should be completely restarted? 🤔 This alson doesnt' work yet?

acolytec3:
I'm not sure exactly what is going on but from my observations, once the client has synced with some peer, when a `NEW_BLOCK_HASHES` message is received, it updates the sync target with the block number from the message, enqueues the new task to retrieve the blocks, and then somehow the last task from before the the initial sync gets enqueued by the fetcher and it never actually retrieves the new blocks. I'm going to try stepping through it in the debugger and figure out what's going on.

acolytec3:
I mispoke.  I pulled down the latest master and fired up geth mining blocks on a private network, had our client sync to it, then added a second client syncing to our first client and it's now correctly getting `NEW_BLOCK_HASHES` messages and then requesting the corresponding block headers/bodies and added them to the chain, so I think we can finally call this issue closed.  :-)

holgerd77:
🎉

Ok, thanks for the clarification and additional testing! 🙂

PR #1480, opened 2021-09-17
common: improve TS types for hardfork, chain and genesis parameters

gabrocheleau:
This PR improves a few types from the common package declared in the `index` and `types` files, and implements the necessary fixes throughout the other packages.

### DIFF
--- a/packages/client/lib/config.ts
+++ b/packages/client/lib/config.ts
@@ -288,7 +288,7 @@ export class Config {
       // Otherwise parse transports from transports option
       this.servers = parseTransports(this.transports).map((t) => {
         if (t.name === 'rlpx') {
-          // TODO: Commented out this.chainCommon.bootstrapNodes() as its type is incompatible with the format required byRlpxServer
+          // TODO: Commented out this.chainCommon.bootstrapNodes() as its type is incompatible with the format required by RlpxServer
           const bootnodes = this.bootnodes ?? [] // this.chainCommon.bootstrapNodes()
           const dnsNetworks = options.dnsNetworks ?? this.chainCommon.dnsNetworks()
           return new RlpxServer({ config: this, bootnodes, dnsNetworks })
--- a/packages/client/lib/config.ts
+++ b/packages/client/lib/config.ts
@@ -288,7 +288,8 @@ export class Config {
       // Otherwise parse transports from transports option
       this.servers = parseTransports(this.transports).map((t) => {
         if (t.name === 'rlpx') {
-          const bootnodes = this.bootnodes ?? this.chainCommon.bootstrapNodes()
+          // TODO: Commented out this.chainCommon.bootstrapNodes() as its type is incompatible with the format required byRlpxServer
+          const bootnodes = this.bootnodes ?? [] // this.chainCommon.bootstrapNodes()
           const dnsNetworks = options.dnsNetworks ?? this.chainCommon.dnsNetworks()
           return new RlpxServer({ config: this, bootnodes, dnsNetworks })
         } else if (t.name === 'libp2p') {
--- a/packages/client/lib/config.ts
+++ b/packages/client/lib/config.ts
@@ -288,8 +288,7 @@ export class Config {
       // Otherwise parse transports from transports option
       this.servers = parseTransports(this.transports).map((t) => {
         if (t.name === 'rlpx') {
-          // TODO: Commented out this.chainCommon.bootstrapNodes() as its type is incompatible with the format required by RlpxServer
-          const bootnodes = this.bootnodes ?? [] // this.chainCommon.bootstrapNodes()
+          const bootnodes = this.bootnodes ?? (this.chainCommon.bootstrapNodes() as any)
           const dnsNetworks = options.dnsNetworks ?? this.chainCommon.dnsNetworks()
           return new RlpxServer({ config: this, bootnodes, dnsNetworks })
         } else if (t.name === 'libp2p') {
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -860,10 +860,10 @@ export default class Common extends EventEmitter {
     // Logic: if accumulator is still null and on the first occurence of
     // a block greater than the current hfBlock set the accumulator,
     // pass on the accumulator as the final result from this time on
-    const nextHfBlock = this.hardforks().reduce((acc: BN, hf: HardforkParams) => {
-      const block = hf.block ? new BN(hf.block) : null
-      return block && block.gt(hfBlock) && acc === null ? block : acc
-    }, new BN(0))
+    const nextHfBlock = this.hardforks().reduce((acc: BN | null, hf: any) => {
+      const block = new BN(hf.block)
+      return block.gt(hfBlock) && acc === null ? block : acc
+    }, null)
     return nextHfBlock
   }
 
@@ -877,6 +877,7 @@ export default class Common extends EventEmitter {
     blockNumber = toType(blockNumber, TypeOutput.BN)
     hardfork = this._chooseHardfork(hardfork, false)
     const nextHardforkBlock = this.nextHardforkBlockBN(hardfork)
+
     return nextHardforkBlock === null ? false : nextHardforkBlock.eq(blockNumber)
   }
 
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -4,7 +4,13 @@ import { BN, BNLike, toType, TypeOutput, intToBuffer } from 'ethereumjs-util'
 import { _getInitializedChains } from './chains'
 import { hardforks as HARDFORK_CHANGES } from './hardforks'
 import { EIPs } from './eips'
-import { BootstrapNode, Chain as IChain, GenesisBlock, GenesisState, HardforkParams } from './types'
+import {
+  BootstrapNode,
+  Chain as IChain,
+  GenesisBlock,
+  GenesisState,
+  Hardfork as HardforkParams,
+} from './types'
 
 export enum CustomChain {
   /**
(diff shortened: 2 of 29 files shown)


packages/common/src/index.ts
```diff
@@ -1076,7 +1078,7 @@ export default class Common extends EventEmitter {
    *
    * Note: This value can update along a hardfork.
    */
-  consensusType(): string | ConsensusType {
+  consensusType(): string | ConsensusType | undefined {
```
holgerd77, line 1082: This should always have some decent return value, so adding `undefined` is somewhat a step in the wrong direction (and would also be somewhat of a breaking change, at least on the TypeScript level).

I added this in the `Chain` interface as optional for some backwards compatibility reasons, see also the associated TODO "// TODO: make mandatory in next breaking release".

We should nevertheless just use the `!` operator here (below) instead of `?`. It's not 100% clean but we have consensus all defined in our own chain definitions and it would very much be the greater annoyance if we introduce breaking changes here which somewhat only theoretically apply.

packages/common/src/index.ts
```diff
@@ -754,8 +754,8 @@ export default class Common extends EventEmitter {
    * @param opts Hardfork options (onlyActive unused)
    * @return Array with hardfork arrays
    */
-  activeHardforks(blockNumber?: BNLike | null, opts: hardforkOptions = {}): Array<any> {
-    const activeHardforks = []
+  activeHardforks(blockNumber?: BNLike | null, opts: hardforkOptions = {}): HardforkParams[] {
```
holgerd77, line 757: Just up for discussion, don't have a clear view myself: are/can be seen `any` -> `ConcreteType` typings as a breaking change (you just can do more nasty things with a return value typed as `any` and eventually (likely ?) people have done? 😀)?

packages/common/src/index.ts
```diff
@@ -860,7 +860,7 @@ export default class Common extends EventEmitter {
     // Logic: if accumulator is still null and on the first occurence of
     // a block greater than the current hfBlock set the accumulator,
     // pass on the accumulator as the final result from this time on
-    const nextHfBlock = this.hardforks().reduce((acc: BN, hf: any) => {
+    const nextHfBlock = this.hardforks().reduce((acc: BN | null, hf: any) => {
```
holgerd77, line 863: What is the effect of this change?

packages/common/src/index.ts
```diff
@@ -1145,7 +1149,11 @@ export default class Common extends EventEmitter {
     if (value) {
       return value
     }
-    return (<any>this._chainParams)['consensus'][this.consensusAlgorithm()]
+    const consensusAlgorithm = this.consensusAlgorithm() as
+      | ConsensusAlgorithm.Ethash
+      | ConsensusAlgorithm.Clique
+      | ConsensusAlgorithm.Casper
```
holgerd77, line 1155: Hmm, I find it a bit unfortunate to hardcode the consensus algorithms here for redundancy reasons. Chances are high that this will be forgotten to be updated once we add a new consensus algorithm (not sure if this is around the corner, but who knows?). Can this be achieved in another way?

packages/common/tests/chains.spec.ts
```diff
@@ -159,7 +159,7 @@ tape('[Common/Chains]: Initialization / Chain params', function (t: tape.Test) {
       const c = new Common({ chain: network })
       const dnsNetworks = c.dnsNetworks()
       st.ok(Array.isArray(dnsNetworks), 'is an array')
-      st.equal(typeof dnsNetworks[0], 'string', 'returns the DNS ENR url as a string')
+      st.equal(typeof dnsNetworks?.[0], 'string', 'returns the DNS ENR url as a string')
```
holgerd77, line 162: Ah, this answers a bit my question on the "any -> something" renamings. If this makes changes like this necessary we should likely not do it in this round.

packages/vm/tsconfig.json
```diff
@@ -1,4 +1,5 @@
 {
   "extends": "../../config/tsconfig.json",
   "include": ["src/**/*.ts", "src/**/*.json", "tests/**/*.ts", ],
+  "compilerOptions": {"skipLibCheck": true}
```
holgerd77, line 4: What is this option doing?

holgerd77 (changes requested):
Ok, done, don't worry about the rather long list on change requests! 🙂

The thing is, this library has got some technical debt collected over the years, especially on the side of the JSON file definitions and their accuracy, and I guess we just can't fix this all on the side line.

So we should rather be a bit more conservative and cautious now - and not risk break things for anyone unnecessarily - and then do a broader clean-up and conceptional re-building on the next breaking release.

Not sure, If we anyhow substantially postpone the current breaking releases due to Merge and other client work we might also reconsider doing breaking releases on all libraries and then give the release round substantially more time (early next year or so).

Thinking loud and just an idea, we should rather discuss together if this is desired.

packages/vm/tsconfig.json
gabrocheleau: It skips node_modules during `tsc` so that errors from installed libraries don't show up.

packages/common/src/index.ts
gabrocheleau: I think it depends on the context. In this case, since we're just really being more precise concerning the value that was already being returned, I feel like this is non-breaking. If we actually changed the return value, then I would agree it's breaking.

packages/common/src/index.ts
gabrocheleau: Makes sense, changing for `!`

packages/common/tests/chains.spec.ts
gabrocheleau: Fixed by adding `!` to 
```
  dnsNetworks(): string[] {
    return this._chainParams['dnsNetworks']!
  }
``` 
And therefore removing the `undefined` return type

packages/common/src/index.ts
gabrocheleau: This is being more accurate with regards to the accumulator type and doesn't change the actual logic (in line 865 we were already testing for `acc === null`, which implies that the ). This became necessary when the `this.hardforks()` return type changed from `any[]` to `HardforkParams[]`

packages/common/src/index.ts
gabrocheleau: Good point. I removed this and simplified the typecasting on line 1156. The typecasting is necessary because the TS requires the indexing to be done with the `ConsensusAlgorithm` enum type instead of a string.

gabrocheleau:
Thanks @holgerd77 for the review. This should be good for another round :)

packages/client/lib/config.ts
```diff
@@ -288,7 +288,8 @@ export class Config {
       // Otherwise parse transports from transports option
       this.servers = parseTransports(this.transports).map((t) => {
         if (t.name === 'rlpx') {
-          const bootnodes = this.bootnodes ?? this.chainCommon.bootstrapNodes()
+          // TODO: Commented out this.chainCommon.bootstrapNodes() as its type is incompatible with the format required by RlpxServer
+          const bootnodes = this.bootnodes ?? [] // this.chainCommon.bootstrapNodes()
```
holgerd77, line 292: One can't just comment out default values only because the types doesn't work any more 😜, this is obviously leading the client to not work anymore if the default bootstrap nodes are not set any more??? 😀 😀 😀

(can confirm, just tested)

packages/common/src/index.ts
```diff
@@ -992,24 +1001,24 @@ export default class Common extends EventEmitter {
...
   }
 
   /**
    * Returns DNS networks for the current chain
    * @returns {String[]} Array of DNS ENR urls
    */
-  dnsNetworks(): any {
-    return (<any>this._chainParams)['dnsNetworks']
+  dnsNetworks(): string[] {
+    return this._chainParams['dnsNetworks']!
```
holgerd77, line 1021: 👍

packages/vm/src/state/stateManager.ts
```diff
@@ -544,7 +544,7 @@ export default class DefaultStateManager implements StateManager {
    */
   async hasGenesisState(): Promise<boolean> {
     const root = this._common.genesis().stateRoot
-    return await this._trie.checkRoot(root)
+    return await this._trie.checkRoot(toBuffer(root))
```
holgerd77, line 547: Just so that this is not slipping through: can you give a short comment on the nature respectively potential effect of this change? Is this just some more-concrete-typing adoption? Or might this also have fixed up-till-now hiddedn bugs?

packages/common/tests/hardforks.spec.ts
```diff
@@ -390,7 +390,7 @@ tape('[Common]: Hardfork logic', function (t: tape.Test) {
 
     let msg = 'should return the correct HF array for a matching forkHash'
     const res = c.hardforkForForkHash('0x3edd5b10')
-    st.equal(res.name, 'spuriousDragon', msg)
+    st.equal(res.name, Hardfork.SpuriousDragon, msg)
 
     msg = 'should return null for a forkHash not matching any HF'
     st.equal(c.hardforkForForkHash('0x12345'), null, msg)
```
holgerd77, line 396: Nice. 👍 😄

holgerd77 (changes requested):
Thanks a lot for the quick updates, generally looks great! 😄 Still feeling a bit shaky on this changeset in general, since I guess there is some rest of a chance that we introduce something breaking for people, e.g. if people are just not providing all the necessary stuff in their custom JSON files (if there are may who do this) and up-till-now have gotten through with that.

I guess though that it's better to take that risk at some point and do this totally overdue step to better types here than staying in the blurry forever and hope that everyone muddles through correctly with these under-typed JSON files and associated functions. 😋

Ok, one super biggy 😛, but I can address directly myself. Otherwise looks good and will merge.

holgerd77 (approved):
LGTM

packages/client/lib/config.ts
holgerd77: Yeah, these kind of conclusions just can't be drawn at all I guess. There is just so much of an open design space how to work around not matching types (at the end that was the "standard" way of during things in vanilla JS 😅), being it just using a subset of keys in a dict or even using an `any` cast in the consuming function.

In this case I was curious and followed it up: indeed just an [any cast](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/lib/util/parse.ts#L42) in the consuming function together with non-matching types on the function signature. 😜 Definitely not optimal on the implementation side with (large) room for improving things. Nevertheless as some illustration how things just *are* in practice and need to be considered. 😬

PR #1484, opened 2021-09-21
devp2p: fix `RLPx` snappy uncompress issue for `DISCONNECT` msgs along handshake

### LINKED ISSUE #1416: Client: Invalid remainder Sync Error

holgerd77:
Fixes #1416 

Ok, I was able to track this down. This has indeed a connection to the introduced snappy compression. Snappy should only be applied after the protocol handshake is finished and for `DISCONNECT` msgs it is often hard to decide if they were send along or after the initial handshake.

I've now applied a hotfix here which fixes the problem (in both directions) by retrying to RLP decode (this is where execution breaks) with the respective non-uncompressed/compressed message (only for `DISCONNECT`) and only then re-throw the error respectively give up. 😋

I am not 100% sure if I am not seeing the forest for all the trees and there *does exist* some clean way to distinguish here. So if someone has a clearer view here feel free to improve the PR.

I would otherwise recommend that we merge in here for now. This is definitely a net improvement and shouldn't blow up the code since the fix is applied very locally and I added a long explanation what is happening.

*If* there is a cleaner way to do it this might be very well something which "solves itself" during a major refactoring of this whole `Peer` class code base (which we should do at some point).

### DIFF
--- a/packages/devp2p/src/rlpx/peer.ts
+++ b/packages/devp2p/src/rlpx/peer.ts
@@ -452,14 +452,12 @@ export class Peer extends EventEmitter {
    * @param msg
    */
   _handleMessage(code: PREFIXES, msg: Buffer) {
-    const payload = rlp.decode(msg)
-
     switch (code) {
       case PREFIXES.HELLO:
-        this._handleHello(payload)
+        this._handleHello(msg)
         break
       case PREFIXES.DISCONNECT:
-        this._handleDisconnect(payload)
+        this._handleDisconnect(msg)
         break
       case PREFIXES.PING:
         this._handlePing()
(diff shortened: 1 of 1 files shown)


acolytec3 (approved):
I agree that there's probably a "better" way to do this but this is just for disconnects which aren't really an impactful message to LGTM!  And, I'm getting clean disconnects between clients on my devnet and no more invalid remainder errors.

PR #1486, opened 2021-09-21
VM: Fix Blake2B with messages with a length >= 5

### LINKED ISSUE #1482: VM: Blake2 precompiled contract gives incorrect result on some inputs
Version: release v4.1.3 (commit feef85f7d51c4e09da3105ebe473e86dcff45eb1)
Node: v10.24.1
Steps to reproduce:
1. go to `tests/api/istanbul/eip-152.js` and add following testcase to `const testCases = [...`
```
  {
    input: '0000000c48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b61626364650000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000500000000000000000000000000000001',
    expected: 'f3e89a60ec4b0b1854744984e421d22b82f181bd4601fb9b1726b2662da61c29dff09e75814acb2639fd79e56616e55fc135f8476f0302b3dc8d44e082eb83a8',
    name: 'vector 9'
  }
```
2. run `npm run test:API`

One can verify that expected value is correct by querying mainnet node
```
curl https://mainnet.infura.io/v3/<infura token> \
    -X POST \
    -d '{
    "jsonrpc": "2.0",
    "method": "eth_call",
    "params": [
        {
            "from": "0x0000000000000000000000000000000000000000",
            "to": "0x0000000000000000000000000000000000000009",
            "data": "0x0000000c48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b61626364650000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000500000000000000000000000000000001"
        },
        "latest"
    ],
    "id": 0
}' \
    -H "Content-Type: application/json"
{"jsonrpc":"2.0","id":0,"result":"0xf3e89a60ec4b0b1854744984e421d22b82f181bd4601fb9b1726b2662da61c29dff09e75814acb2639fd79e56616e55fc135f8476f0302b3dc8d44e082eb83a8"}
```

Hashing string `abcde` using some online hash calculator confirms this result (https://www.toolkitbay.com/tkb/tool/BLAKE2b_512)

One can also write similar test in go-ethereum repo (which I initially did), but this requires building go-ethereum.

**It seems that all strings of length less or equal 4 are hashed correctly** (thus all mentioned test vectors work), problem occurs whenever length of string is *at least 5* (and it so happened all current test vectors are hashing string `abc`)

I was not able to figure out what's causing this issue, js implementation seems quite different from golang in go-ethereum

jochem-brouwer:
Closes #1482 

I need to verify that these changes are correct. I noticed that in the original spec (https://github.com/dcposch/blakejs/blob/master/blake2b.js) the B2B_GET32 receives an Uint8Array, where we pass an Uint32Array. The Uint32Array we are using is [already](https://github.com/ethereumjs/ethereumjs-monorepo/blob/69266a231fb47351e114cf5ba43c0bb98bc42cac/packages/vm/src/evm/precompiles/09-blake2f.ts#L210) little endian.

I have locally fuzzed 1000 random messages (rounds <= `2^16-1`) against Infura and verified these were the same as which the VM reports after these changes.


jochem-brouwer:
I noticed that Uint32 arrays endianness is based upon the OS endianness. So we might want to test this also on big endian OS - then this might fail.

holgerd77:
@jochem-brouwer pretty cool that you got something on this that quickly! 😄

This is what I got from Sina, asked him per DM if he has got some insight:

> oof that's a tricky one 🙂 I copied the blake2 implementation from a library and added the wrapping which makes it a precompile. From the issue description I gather the problem should be somehow with the m variable (message being hashed). But having looked at it for some time nothing jumps out in the precompile-part of the code and I don't really understand the blake2 part enough...

jochem-brouwer:
I have created a dirty fuzzer and ran 1000 random inputs against Infura and checked if the VM reported the same outcome.

I have tested rounds up to `2^16-1`, if I use more rounds (one could theoretically do `2^32-1` rounds) I run against out of gas at Infura.

jochem-brouwer:
To clear up any confusion about endianness; we actually read the message in little endian as can be seen [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/69266a231fb47351e114cf5ba43c0bb98bc42cac/packages/vm/src/evm/precompiles/09-blake2f.ts#L210).

jochem-brouwer:
I am pretty convinced this is now fixed.

axic:
Is it worth adding an upstream test case for this in https://github.com/ethereum/tests?

jochem-brouwer:
Hi @axic, I have contacted @winsvega from ethereum/tests, they will add this case to ethereum/tests :smile:

s1na (approved):
LGTM

holgerd77 (approved):
Very cool, Ori has even already added some new test cases for this (see https://github.com/ethereum/tests/pull/948).

Will merge, planning monorepo releases for Monday or Tuesday anyhow, so this will get into a release soon.

PR #1487, opened 2021-09-21
client: Filter out forks with no block set in config

### LINKED ISSUE #1466: Client throws `outdated fork status` error when syncing to client on different hardfork
When running multiple instances of the client, if a new instance of the client is started and tries to connect to the first client and the first client is on a later hardfork, the first client throws the below error when it receives the STATUS message from the second client and won't connect:
```js
devp2p:eth Send STATUS message to 127.0.0.1:37270 (eth66): [V:66, NID:15470, TD:3699, BestH:130a6fd, GenH:2eed290, ForkHash: 0x14f6d8f0, ForkNext: NaN] +881ms
  devp2p:eth Received STATUS message from 127.0.0.1:37270: [V:66, NID:15470, TD:1, BestH:2eed290, GenH:2eed290, ForkHash: 0x5616a08c, ForkNext: 20] +5ms
  devp2p:eth Outdated fork status, remote needs software update +0ms
ERROR [09-09|15:36:16] AssertionError [ERR_ASSERTION]: Outdated fork status, remote needs software update
```

If the second client connects to a geth peer on the same network, geth will connect and start exchanging messages with the second client:
```
 devp2p:eth Send STATUS message to 127.0.0.1:37792 (eth66): [V:66, NID:15470, TD:1, BestH:2eed290, GenH:2eed290, ForkHash: 0x5616a08c, ForkNext: 20] +1m
  devp2p:eth Received STATUS message from 127.0.0.1:37792: [V:66, NID:15470, TD:3707, BestH:5d6a712, GenH:2eed290, ForkHash: 0x14f6d8f0, ForkNext: NaN] +1ms
  devp2p:eth Send GET_BLOCK_HEADERS message to 127.0.0.1:37792: e601e4a05d6a712507a2c24c9f0251e6498e25499ac93e9c6bc319508a53... +4s
```

To recreate, set up a private network using the attached geth genesis file.
[poa.json.txt](https://github.com/ethereumjs/ethereumjs-monorepo/files/7139348/poa.json.txt)

1) `cat poa.json.txt > poa.json`
2) Create a new account for mining and update the `extradata` field in `poa.json` with this new account.
3) `geth --datadir data --networkid 15470 --log.debug --mine --unlock 0xMyMinerAccount -nodiscover` and unlock your account for mining 
4) Run `DEBUG=devp2p:eth npm run client:start -- --gethGenesis=path/to/poa.json --port=30305` in a new terminal window
5) Run  `geth attach data/geth.ipc` in another terminal window
6) `admin.addPeer("client enode address")`
7) Go back to the terminal window with the client instance running and wait until it passes the first hardfork block (block 20)
8) Open a new terminal window and run `DEBUG=devp2p:eth npm run client:start -- --datadir=data2 --gethGenesis=../../../../poa.json --port=30307 --bootnodes=enode://mysyncedEthJSClientEnodeAddress@0.0.0.0:30305 --transports=rlpx`
9) Go back to the terminal window with the first client and you should see an error message similar to the top of this post when the two clients exchange `STATUS` messages

acolytec3:
Fixes #1466.  

The source of the issue is that when a geth genesis file is provided to the client, the `parseGethParams` method is inserting hardforks into the chain parameters with null block numbers where the hardfork is not present in the geth file.  When a client tries to connect to another client using the same geth genesis file, if one client isn't fully synced and sends a forkhash that is a subset of the other, the issue causes an error in `_validateForkId` at [this point](https://github.com/ethereumjs/ethereumjs-monorepo/blob/6e6dfb8f4c813c8be4be23b3f5a627e3aa545835/packages/devp2p/src/eth/index.ts#L157-L157) if the hardfork with a null block number is follow a hardfork with a block number since the forkhash is computed for all hardforks that occur on a given block is [the same](https://eips.ethereum.org/EIPS/eip-2124#specification).  If a hardfork exists in the `chainParams` that has a null block value, it is treated as having the same forkhash as the hardfork immediately before it and so `_validateForkId` was finding that the `peerNextFork` was a hardfork with no block number instead of the fork immediately before it in sequence.

This resolves the issue by simply filtering out any hardforks that weren't included in the geth genesis file as these shouldn't exist in the `Common` for that custom network.

### DIFF
--- a/packages/client/lib/util/parse.ts
+++ b/packages/client/lib/util/parse.ts
@@ -217,10 +217,12 @@ async function parseGethParams(json: any) {
     berlin: 'berlinBlock',
     london: 'londonBlock',
   }
-  params.hardforks = hardforks.map((name) => ({
-    name: name,
-    block: name === 'chainstart' ? 0 : config[forkMap[name]] ?? null,
-  }))
+  params.hardforks = hardforks
+    .map((name) => ({
+      name: name,
+      block: name === 'chainstart' ? 0 : config[forkMap[name]] ?? null,
+    }))
+    .filter((fork) => fork.block !== null)
   return params
 }
 /**


ryanio (approved):
looks good, thanks!

PR #1489, opened 2021-09-22
client: Fix gas calculations on EIP1559 forkblock in miner

acolytec3:
When the miner is assembling a new block and the new block happens to be the EIP1559 block, the `assembleBlock` was failing since the `baseFeePerGas` param wasn't active since EIP1559 isn't technically active in the `common` instance of the miner at the point in time where it's assembling the EIP1559 fork block (i.e. `Common` thinks it's EIP1559 fork block - 1).  This looks up the `initialBaseFee` from `common.paramByEIP` instead of `common.param` to retrieve that initial value when executing on the EIP1559 fork block.  It also adjusts the gas limit in line with the logic [validateGasLimit](https://github.com/ethereumjs/ethereumjs-monorepo/blob/90fea3df7a07dd1b1af248a19e76ad56a5432b94/packages/block/src/header.ts#L520-L520).

### DIFF
--- a/packages/client/lib/miner/miner.ts
+++ b/packages/client/lib/miner/miner.ts
@@ -196,7 +196,6 @@ export class Miner {
       vmCopy.stateManager,
       baseFeePerGas
     )
-    console.log(baseFeePerGas)
     this.config.logger.info(
       `Miner: Assembling block from ${txs.length} eligible txs ${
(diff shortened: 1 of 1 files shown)


ryanio (approved):
lgtm thanks, I will separately open an integration test in the miner for this london hardfork switchover block. it will be a good template to also use for the merge.

PR #1490, opened 2021-09-23
client: add miner london hf switch test

ryanio:
Following up on #1489, this PR adds a test to cover the scenario of the miner upgrading through the london hardfork, ensuring no errors are encountered and the gasLimit and baseFeePerGas values are correct.


packages/client/test/miner/miner.spec.ts
```diff
@@ -368,6 +369,79 @@ tape('[Miner]', async (t) => {
     txPool.stop()
   })
 
+  t.test('should handle mining over the london hardfork block', async (t) => {
+    const customChainParams = {
+      hardforks: [
+        { name: 'chainstart', block: 0 },
+        { name: 'berlin', block: 2 },
+        { name: 'london', block: 3 },
+      ],
```
holgerd77, line 378: Ah, funny, I guess that's the first time (that I see) that we are totally detaching from the history-given HF sequence. 😀

holgerd77 (approved):
Really nice test case addition, looks good! 😄

PR #1492, opened 2021-09-24
client: add dev param to start devnet with miner

acolytec3:
Adds a new `cli` param `--dev` that starts the client with a miner with a prefunded account


acolytec3:
Here's an initial version of the `--dev` CLI command.  At the moment, it just starts up the miner with a generated account and preconfigured hardforks.  Do y'all think it's worth adding a flow to allow you to specify the block numbers for the hardforks?

packages/client/bin/cli.ts
```diff
@@ -199,8 +206,68 @@ async function run() {
...
+    }
+
+    try {
+      for (const addressString of args.unlock) {
+        const address = Address.fromString(addressString)
+        const inputKey = await question(
+          `Please enter the 0x-prefixed private key to unlock ${address}:\n`
+        )
+        const privKey = toBuffer(inputKey)
+        const derivedAddress = new Address(privateToAddress(privKey))
```
ryanio, line 229: I know this part is my code but I wanted to update this :)

```suggestion
        const derivedAddress = Address.fromPrivateKey(privKey)
```

packages/client/bin/cli.ts
```diff
@@ -199,8 +206,68 @@ async function run() {
...
+    rl.close()
+  }
+
   let common: Common = {} as Common
+  if (args.dev) {
+    fs.rmdirSync('./dev', { recursive: true })
+    if (accounts.length === 0) {
+      // Create new account for devnet if not provided
+      const privKey = crypto.randomBytes(32)
+      const account = new Address(privateToAddress(privKey))
```
ryanio, line 250:
```suggestion
      const account = Address.fromPrivateKey(privKey)
```

packages/client/bin/cli.ts
```diff
@@ -199,8 +206,68 @@ async function run() {
...
+      accounts.push([account, privKey])
```
ryanio, line 251:
```suggestion
      const address = Address.fromPrivateKey(privKey)
      accounts.push([address, privKey])
```

packages/client/bin/devnet.json
```diff
@@ -0,0 +1,800 @@
...
+    "eip150Block": 0,
+    "eip150Hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+    "eip155Block": 0,
+    "eip158Block": 0,
+    "byzantiumBlock": 60,
+    "constantinopleBlock": 80,
+    "petersburgBlock": 100,
+    "istanbulBlock": 120,
+    "berlinBlock": 140,
+    "londonBlock": 160,
```
ryanio, line 14: hm we might want to reconsider what to set these defaults since it might trip people up when hardforks suddenly change

packages/client/bin/devnet.json
```diff
@@ -0,0 +1,800 @@
+{
```
ryanio, line 1: I don't love this file in this folder, but maybe for clarity we can rename it to `devnet-chain.json`

packages/client/bin/cli.ts
```diff
@@ -199,8 +206,68 @@ async function run() {
...
+      console.log(`Account generated for mining blocks: ${account.toString()}`)
+      console.log(`Private key associated with mining account: 0x${privKey.toString('hex')}`)
+      console.log('WARNING: Do not use this account for mainnet funds')
```
ryanio, line 254:
```suggestion
      const address = Address.fromPrivateKey(privKey)
      accounts.push([address, privKey])
      console.log(
      `===========================================\n
      Account generated for mining blocks: \n
        Address: ${account.toString()}\n
        Private key: 0x${privKey.toString('hex')}\n
      WARNING: Do not use this account for mainnet funds\n
      ===========================================
      `)
```

packages/client/bin/cli.ts
```diff
@@ -307,7 +341,7 @@ async function run() {
     debugCode: args.debugCode,
     discDns: args.discDns,
     discV4: args.discV4,
-    mine: args.mine,
+    mine: args.dev ?? args.mine,
```
ryanio, line 344: hm with the nullish coalescing operator if `args.dev` is false it would return `false`:

> The nullish coalescing operator (??) is a logical operator that returns its right-hand side operand when its left-hand side operand is `null` or `undefined`, and otherwise returns its left-hand side operand.

```shell
➜  ~ ts-node
> false ?? true
false
```

packages/client/bin/cli.ts
```diff
@@ -199,8 +206,68 @@ async function run() {
   // give network id precedence over network name
   const chain = args.networkId ?? args.network ?? 'mainnet'
 
-  // configure common based on args given
+  // configure accounts for mining and prefunding in a local devnet
+  const accounts: [Address, Buffer][] = [] // [address, privateKey]
```
ryanio, line 210: (my code again) we can inline this comment now that we are on latest TS :)

```suggestion
  const accounts: [address: Address, privateKey: Buffer][] = []
```

packages/client/bin/cli.ts
```diff
@@ -199,8 +206,68 @@ async function run() {
...
+      }
+    } catch (error) {
+      console.error(`Encountered error unlocking account:\n${error}`)
+    }
+    rl.close()
+  }
+
   let common: Common = {} as Common
+  if (args.dev) {
+    fs.rmdirSync('./dev', { recursive: true })
```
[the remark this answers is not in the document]
acolytec3: That could make sense.  If we want to allow that option (as opposed to a truly ephemeral devnet), I'll set it up so it only deletes the folder if there's no mining account/private key provided.  Otherwise, the client won't start because the genesis state will be different when a new mining account gets generated.

packages/client/bin/devnet.json
acolytec3: None, I was just using the block numbers from one of my private network files.  I'll update.

packages/client/bin/cli.ts
acolytec3: Ah, thanks for the clarification.  I actually need something slightly different `mine: args.mine || args.dev` to make this work.  We want to start the miner if either `args.mine===true` or if `args.dev===true`

packages/client/bin/devnet.json
acolytec3: I need to look at the `parse` functions again.  I may be able to slim down the genesis parameters and just construct an object  instead of the extra file.  I just needed something for generating the genesis block stateRoot

acolytec3: I've replaced the external json file with a custom params object.  Let me know if you have an alternative you'd prefer.

ryanio (approved):
looks great, thanks for adding this param, so easy! i will be using this a lot.

holgerd77:
Hi there,
that's pretty cool, so great! 🎉

I've got some questions, but I'll leave it for now. Please let's not do *any* further client work until the end of the week apart from bug fixing and making things more robust and resilient - so doing some kind of feature freeze here - so that we get well through the releases. 😄

PR #1493, opened 2021-09-27
client: merge related work

ryanio:
This PR continues implementation of [EIP-3675](https://eips.ethereum.org/EIPS/eip-3675) in the client:
* ignores`NewBlockHashes` and `NewBlock` messages received after the merge and drops peers who gossip them
* adds `td` to `setHardforkByBlockNumber` to handle merge hardfork transition (#1473)

I will start work on implementing the [Engine API](https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.1/src/engine/interop/specification.md) which includes preparing/executing blocks and reacting to `consensusValidated` and `forkchoiceUpdated` events.


ryanio:
Have added ethash cpu PoW mining with https://github.com/ethereumjs/ethereumjs-monorepo/pull/1493/commits/f5d4c0830f856d1027991feb35dbc10827a17430 - thanks to @jochem-brouwer and #1475 

@acolytec3 Please take a look and let me know if you have any suggestions or changes, I updated the `--dev` parameter to a string that defaults to `poa` but can be set to `pow` (so plain `--dev` will still work as before)

holgerd77:
> Have added ethash cpu PoW mining with [f5d4c08](https://github.com/ethereumjs/ethereumjs-monorepo/commit/f5d4c0830f856d1027991feb35dbc10827a17430) - thanks to @jochem-brouwer and #1475
> 
> @acolytec3 Please take a look and let me know if you have any suggestions or changes, I updated the `--dev` parameter to a string that defaults to `poa` but can be set to `pow` (so plain `--dev` will still work as before)

I wonder if we also want the `--dev` option to react to the `--mine` option (and only mine if set), otherwise we can't really set up two-client networks (or more) - one active and one passive - since one can't recreate the genesis used for the mining client in a simple way and then pass on to the other client atm. Does this make sense or am I missing something? 🤔

acolytec3:
> > Have added ethash cpu PoW mining with [f5d4c08](https://github.com/ethereumjs/ethereumjs-monorepo/commit/f5d4c0830f856d1027991feb35dbc10827a17430) - thanks to @jochem-brouwer and #1475
> > @acolytec3 Please take a look and let me know if you have any suggestions or changes, I updated the `--dev` parameter to a string that defaults to `poa` but can be set to `pow` (so plain `--dev` will still work as before)
> 
> I wonder if we also want the `--dev` option to react to the `--mine` option (and only mine if set), otherwise we can't really set up two-client networks (or more) - one active and one passive - since one can't recreate the genesis used for the mining client in a simple way and then pass on to the other client atm. Does this make sense or am I missing something? thinking

It's on my mental to-do list to add an additional step here where the cli dumps the common to a json file if using the `--dev` param that can then be picked up by additional instances of the client.  Would just be tricky since we have to specify a custom `datadir` as well so we don't get database lock issues.  I'll do some research on it.

packages/client/lib/miner/miner.ts
```diff
@@ -97,8 +143,8 @@ export class Miner {
     }
     this.running = true
     this.config.events.on(Event.CHAIN_UPDATED, this.chainUpdated.bind(this))
-    void this.queueNextAssembly() // void operator satisfies eslint rule for no-floating-promises
     this.config.logger.info(`Miner started. Assembling next block in ${this.period / 1000}s`)
```
acolytec3, line 146: This is cosmetic but this should say something different if the consensus algorithm is PoW.  Maybe something like.  You'll have to adjust the formatting because Prettier and ESLint can't agree..
```suggestion
this.config.logger.info(
      `Miner started. ${this.config.execCommon.consensusType() === ConsensusType.ProofOfAuthority
        ? `Assembling next block in ${this.period / 1000}s`
        : `Finding next PoW solution 🔨`
      }`
    )
```

packages/client/lib/miner/miner.ts
```diff
@@ -229,8 +296,12 @@ export class Miner {
     }
     if (interrupt) return
     // Build block, sealing it
-    const block = await blockBuilder.build()
-    this.config.logger.info(`Miner: Sealed block with ${block.transactions.length} txs`)
+    const block = await blockBuilder.build(this.nextSolution)
+    this.config.logger.info(
+      `Miner: Sealed block with ${block.transactions.length} txs${
```
acolytec3, line 301: Same as above, cosmetic but aligns the logs with the consensus algorithm
```suggestion
his.config.logger.info(
      `Miner: ${this.config.chainCommon.consensusType === ConsensusType.ProofOfAuthority
        ? `Sealed`
        : `Mined`
      } block with ${block.transactions.length} txs${
        this.nextSolution ? ` (difficulty: ${block.header.difficulty})` : ''
      }`
    )
```

ryanio:
maybe with the `--dev` flag we can create the common files in `${datadir}/devnet` and a `--dev2` flag can look for them there and set its own datadir to `/devnet2`

acolytec3:
> Have added ethash cpu PoW mining with [f5d4c08](https://github.com/ethereumjs/ethereumjs-monorepo/commit/f5d4c0830f856d1027991feb35dbc10827a17430) - thanks to @jochem-brouwer and #1475
> 
> @acolytec3 Please take a look and let me know if you have any suggestions or changes, I updated the `--dev` parameter to a string that defaults to `poa` but can be set to `pow` (so plain `--dev` will still work as before)

What do you think about switching the consensus type to be passed to `--mine`? Then, we could use the `--dev` flag to specify multiple instances of the client, something like `--dev=2` where specifying a number starts up an instance of the client using the `common` parameters dumped from the miner client?

ryanio:
> > Have added ethash cpu PoW mining with [f5d4c08](https://github.com/ethereumjs/ethereumjs-monorepo/commit/f5d4c0830f856d1027991feb35dbc10827a17430) - thanks to @jochem-brouwer and #1475
> 
> > 
> 
> > @acolytec3 Please take a look and let me know if you have any suggestions or changes, I updated the `--dev` parameter to a string that defaults to `poa` but can be set to `pow` (so plain `--dev` will still work as before)
> 
> 
> 
> What do you think about switching the consensus type to be passed to `--mine`? Then, we could use the `--dev` flag to specify multiple instances of the client, something like `--dev=2` where specifying a number starts up an instance of the client using the `common` parameters dumped from the miner client?

I like that idea!

holgerd77:
Rebased this.

holgerd77 (approved):
LGTM! 😄

ryanio:
Thanks! I will open a new PR to continue working on the Engine API.

Issue #252, opened 2018-01-26
Support Promises


dmihal:
I would love if the API supported returning Promises as an alternative to callbacks. This would be implemented by returning Promises if no callback is provided.

I'm happy to add the feature and submit a PR if you're open to it!

jwasinger:
I like this idea.  It is backwards compatible as well!

Pending any reservations by @holgerd77 @axic , I would say go ahead :)

holgerd77:
Would also very much support this, great that you want to implement it! 😄 

Could you post some code snippet with 10LOC or so on how you think/plan this would look like when being called with promises from the outside?

dmihal:
With promises, the code can be executed in 3 ways:

The existing way with callbacks
```javascript
vm.runCode({
  code: Buffer.from(code, 'hex'), // code needs to be a Buffer
  gasLimit: Buffer.from('ffffffff', 'hex')
}, function(err, results){
  console.log('returned: ' + results.return.toString('hex'));
})
```

Using promises
```javascript
vm.runCode({
  code: Buffer.from(code, 'hex'), // code needs to be a Buffer
  gasLimit: Buffer.from('ffffffff', 'hex')
})
.then(function(results){
  console.log('returned: ' + results.return.toString('hex'));
})
.catch(function(err){
  console.error(err);
})
```

Or using async/await

```javascript
try {
  var results = await vm.runCode({
    code: Buffer.from(code, 'hex'), // code needs to be a Buffer
    gasLimit: Buffer.from('ffffffff', 'hex')
  });
  console.log('returned: ' + results.return.toString('hex'));
} catch (err) {
  console.error(err);
}
```

dmihal:
Based on the feedback, I made a cleaner implementation in #273

Silur:
@jwasinger actually I made a Promised (no async dep) version of all etherumjs repos with @wanderer in around last year may

holgerd77:
@Silur Tell us more, is this code locked on an USB-Stick in a secret cupboard in a forgotten Lisbon library?? 😄 🔑

holgerd77:
Addressed in #546, will close.

PR #1494, opened 2021-09-27
util: upgrade deps

talentlessguy:
this PR upgrades ethereumjs-util dependencies which add a few fixes:

## bn.js `5.1.2 => 5.2.0`:

- fix: Buffer not using global in browser (#260)
- Add support for defined but not implemented Symbol.for (#252)

## rlp `2.2.4 => 2.2.6`

- Add bigint support
-  Fix long string decode


ryanio:
hey @talentlessguy, these should already be included as we periodically refresh the lock file and they're specified with carats.

also for deps like `bn.js` that span across the whole monorepo we want to ensure we use the same version and not fragment - so if we upgrade a version of a dep we want to do it in all package.jsons that contain the same dependency (sorry I know that's tedious, but again since it's covered by the carat up to the next major version I don't think this PR is necessary).

talentlessguy:
ok, closing then

talentlessguy:
and please check my pr to rlp :)

ryanio:
will do! thanks for sticking through on that one

PR #1500, opened 2021-09-29
Tx/Util: ensure only integers are allowed as input

### LINKED ISSUE #1497: Passing decimal values to FeeMarketEIP1559Transaction generates invalid output silently
In the snippet attached below, if you pass decimal numbers to the FeeMarketEIP1559Transaction constructor it doesn't throw any errors, but alters the values silently, this can be very dangerous as the final payload when ran in a browser environment converts `33974229950.550003` -> `35624562130324685` which is a much larger number than what was originally passed, increasing the gas price.

```
const ethereumJsTx = require('@ethereumjs/tx') //  "@ethereumjs/tx": "^3.3.0",

const decimal_gas_fee = 33974229950.550003
const base_fee = 5 * 10 ** 9

const rawTx = {
  maxPriorityFeePerGas: decimal_gas_fee,
  maxFeePerGas: decimal_gas_fee + base_fee,
  chainId: 1,
  nonce: 0,
  gasPrice: null,
  gasLimit: 144000,
  to: '0x0000000000000000000000000000000000000000',
  value: 0,
  data: []
}

const tx = ethereumJsTx.FeeMarketEIP1559Transaction.fromTxData(rawTx)

// Browser environment
// maxFeePerGas: 45061746130324685, maxPriorityFeePerGas: 35624562130324685

// Node.js environment
// maxFeePerGas: 2123389402, maxPriorityFeePerGas: 2123389402

console.log(tx.maxFeePerGas.toString(), tx.maxPriorityFeePerGas.toString())
```

jochem-brouwer:
This PR closes #1497 

`intToHex` now explicitly checks that the input value can be used. This also means that consumers who use JS (not TypeScript) are ensured to get correct values.

 `@ethereumjs/util` also does not export the `intToHex` and `intToBuffer` methods from `ethjs-util` anymore.


packages/tx/test/eip1559.spec.ts
```diff
@@ -15,6 +15,28 @@ const validSlot = Buffer.from('01'.repeat(32), 'hex')
...
+      'nonce',
+      'gasLimit',
+      'value',
+      'v',
+      'r',
+      's',
+    ]
+    for (const value of values) {
+      const txData: any = {}
+      txData[value] = 10.1
```
ryanio, line 32: can we also try string decimal? `'10.1'`

packages/util/src/bytes.ts
```diff
@@ -1,8 +1,33 @@
...
+/**
+ * Converts a `Number` into a hex `String`
+ * @param {Number} i
+ * @return {String}
+ */
+export const intToHex = function (i: number) {
+  if (typeof i !== 'number' || isNaN(i) || i === Infinity || i < 0 || i % 1 !== 0) {
+    throw new Error(`Received an invalid integer type: ${i}`)
+  }
+  var hex = i.toString(16); // eslint-disable-line
```
ryanio, line 15: should we just do

```
return `0x${i.toString(16)}`
```

(no need to initialize a const?)

or also would this satisfy eslint:

```suggestion
  const hex = i.toString(16)
```

packages/util/src/bytes.ts
```diff
@@ -1,8 +1,33 @@
...
 import { PrefixedHexString, TransformableToArray, TransformableToBuffer } from './types'
 import { assertIsBuffer, assertIsArray, assertIsHexString } from './helpers'
 
+/**
+ * Converts a `Number` into a hex `String`
+ * @param {Number} i
+ * @return {String}
+ */
+export const intToHex = function (i: number) {
+  if (typeof i !== 'number' || isNaN(i) || i === Infinity || i < 0 || i % 1 !== 0) {
```
ryanio, line 12: should we add `|| !Number.isSafeInteger(i)`?

gabrocheleau, line 12: That last check `i % 1 !== 0` will evaluate to false when `i === 0`, which I don't think is what we want, since we want to allow 0.
What do you think of the following:
 ```suggestion
  if (typeof i !== 'number' || isNaN(i) || i === Infinity || i < 0 || (i !== 0 && i % 1 !== 0)) {
```

File: packages/tx/test/eip1559.spec.ts (hunk 1 above)
Comment (jochem-brouwer) [reply packages/tx/test/eip1559.spec.ts:32]: I have not checked anything related with strings - good point because this will probably also produce garbage output.

File: packages/util/src/bytes.ts (hunk 3 above)
Comment (jochem-brouwer) [reply packages/util/src/bytes.ts:12]: Yep good point!

Comment (gabrocheleau) [reply packages/util/src/bytes.ts:12]: Actually I take that back, I thought that `0 % 1` returned NaN for some reason.

Comment (jochem-brouwer) [reply packages/util/src/bytes.ts:12]: It seems like I can directly use this method instead of checking all these conditions :smile:

Comment (gabrocheleau) [reply packages/util/src/bytes.ts]: Cool, didn't know about that method!

I tested the conditions and they account for all of them except `i < 0`

Comment (jochem-brouwer) [reply packages/util/src/bytes.ts:12]: 0 % 1 = 0?

**Review by gabrocheleau — APPROVED** [review]

File: packages/util/src/bytes.ts (hunk 3 above)
Comment (jochem-brouwer) [reply packages/util/src/bytes.ts]: Yup, just pushed the changes that accounts for it :)

File: packages/tx/test/eip1559.spec.ts (hunk 1 above)
Comment (ryanio) [reply packages/tx/test/eip1559.spec.ts:34]: awesome, trying to think of some other last cases we can add here: `-1`, `-1.1`, `'-1.1'`

File: packages/util/src/bytes.ts [hunk 4]
```diff
@@ -1,8 +1,31 @@
...
+}
+
+/**
+ * Converts an `Number` to a `Buffer`
+ * @param {Number} i
+ * @return {Buffer}
+ */
+export const intToBuffer = function (i: number) {
+  const hex = intToHex(i)
+
```
ryanio, line 25: harmless nit
```suggestion
```

packages/tx/test/eip1559.spec.ts
jochem-brouwer: I'll add all cases which I also used in util :)

jochem-brouwer:
Note: Setting `chainId` to `false` or `NaN` for EIP1559/AccessList still creates a valid tx.

packages/util/src/bytes.ts
```diff
@@ -1,8 +1,30 @@
...
+import { stripHexPrefix, padToEven, isHexString, isHexPrefixed } from 'ethjs-util'
 import { PrefixedHexString, TransformableToArray, TransformableToBuffer } from './types'
 import { assertIsBuffer, assertIsArray, assertIsHexString } from './helpers'
 
+/**
+ * Converts a `Number` into a hex `String`
+ * @param {Number} i
+ * @return {String}
+ */
+export const intToHex = function (i: number) {
```
gabrocheleau, line 11: Not a big deal, but wouldn't it be simpler to just:
`export function intToHex(i: number) { ... ` ?

packages/util/src/bytes.ts
```diff
@@ -1,8 +1,30 @@
...
+  }
+  return `0x${i.toString(16)}`
+}
+
+/**
+ * Converts an `Number` to a `Buffer`
+ * @param {Number} i
+ * @return {Buffer}
+ */
+export const intToBuffer = function (i: number) {
```
gabrocheleau, line 23: Same as above

packages/tx/test/typedTxsAndEIP2930.spec.ts
```diff
@@ -88,6 +88,37 @@ tape(
...
+        10.1,
+        '10.1',
+        '0xaa.1',
+        -10.1,
+        '-10.1',
+        '-0xaa',
+        Infinity,
+        -Infinity,
+        NaN,
+        {},
```
gabrocheleau, line 103: Perhaps also add `null` and `undefined`

jochem-brouwer: `undefined` won't work since that's the default field of an object if an attribute does not exist (tx will then choose a default value)

packages/util/src/bytes.ts
jochem-brouwer: Maybe, but this file exports all methods like this :smile:

packages/tx/test/typedTxsAndEIP2930.spec.ts
jochem-brouwer: `null` also fails tests

gabrocheleau: Ok thanks for clarifying 🙂

jochem-brouwer:
Okay this is super weird, browser tests fail on `MAX_SAFE_INTEGER+1` ??

jochem-brouwer:
Oh lol, never mind, I always immediately quit when browser build starts. Fixed upon rebuilding.

gabrocheleau:
> Okay this is super weird, browser tests fail on `MAX_SAFE_INTEGER+1` ??

Wow weird. Just guessing here... but might it be that `MAX_SAFE_INTEGER+1` results in a rounding error that makes it equal to  `MAX_SAFE_INTEGER` (which passes the Number.isSafeInteger check) ?

jochem-brouwer:
Ok I have addressed everything noted so far, this should be OK :)

packages/util/test/bytes.spec.ts
```diff
@@ -285,3 +287,43 @@ tape('baToJSON', function (t) {
...
+  st.throws(() => intToBuffer(<any>-1), 'throws on negative numbers')
+  st.throws(() => intToBuffer(<any>1.05), 'throws on decimal numbers')
+  st.throws(() => intToBuffer(<any>{}), 'throws on objects')
+  st.throws(() => intToBuffer(<any>true), 'throws on true')
+  st.throws(() => intToBuffer(<any>false), 'throws on false')
+  st.throws(() => intToBuffer(<any>[]), 'throws on arrays')
+  st.throws(() => intToBuffer(<any>(() => {})), 'throws on arrays')
+  st.throws(() => intToBuffer(Number.MAX_SAFE_INTEGER + 1), 'throws on unsafe integers')
+  st.ok(intToBuffer(0).equals(Buffer.from('00', 'hex')), 'correctly convers 0 to a buffer')
+  st.ok(intToBuffer(1).equals(Buffer.from('01', 'hex')), 'correctly convers 1 to a buffer')
```
gabrocheleau, line 307:
```suggestion
  st.ok(intToBuffer(0).equals(Buffer.from('00', 'hex')), 'correctly converts 0 to a buffer')
  st.ok(intToBuffer(1).equals(Buffer.from('01', 'hex')), 'correctly converts 1 to a buffer')
```

packages/util/test/bytes.spec.ts
```diff
@@ -285,3 +287,43 @@ tape('baToJSON', function (t) {
...
+  st.throws(() => intToHex(<any>-1), 'throws on negative numbers')
+  st.throws(() => intToHex(<any>1.05), 'throws on decimal numbers')
+  st.throws(() => intToHex(<any>{}), 'throws on objects')
+  st.throws(() => intToHex(<any>true), 'throws on true')
+  st.throws(() => intToHex(<any>false), 'throws on false')
+  st.throws(() => intToHex(<any>[]), 'throws on arrays')
+  st.throws(() => intToHex(<any>(() => {})), 'throws on arrays')
+  st.throws(() => intToHex(Number.MAX_SAFE_INTEGER + 1), 'throws on unsafe integers')
+  st.ok(intToHex(0) == '0x0', 'correctly convers 0 to a hex string')
+  st.ok(intToHex(1) == '0x1', 'correctly convers 1 to a hex string')
```
gabrocheleau, line 327:
```suggestion
  st.ok(intToHex(0) == '0x0', 'correctly converts 0 to a hex string')
  st.ok(intToHex(1) == '0x1', 'correctly converts 1 to a hex string')
```

gabrocheleau (changes requested):
typos

gabrocheleau (approved):
Looks good to me.

jochem-brouwer:
Let's wait for one more review before merging.

ryanio (approved):
lgtm!

holgerd77 (approved):
This is actually such an insanely great and clean work - especially regarding the test cases - I honestly don't know what to say.

Thanks a lot Jochem! 👍 🙂

I'll directly prepare a release on Util and Tx.

holgerd77:
Ah, and I've tested this locally, also in direct comparison with the `master` branch version.

packages/util/test/bytes.spec.ts
```diff
@@ -285,3 +287,43 @@ tape('baToJSON', function (t) {
...
+  st.throws(() => intToBuffer(<any>NaN), 'throws on NaN')
+  st.throws(() => intToBuffer(<any>undefined), 'throws on undefined')
+  st.throws(() => intToBuffer(<any>null), 'throws on null')
+  st.throws(() => intToBuffer(<any>-1), 'throws on negative numbers')
+  st.throws(() => intToBuffer(<any>1.05), 'throws on decimal numbers')
+  st.throws(() => intToBuffer(<any>{}), 'throws on objects')
+  st.throws(() => intToBuffer(<any>true), 'throws on true')
+  st.throws(() => intToBuffer(<any>false), 'throws on false')
+  st.throws(() => intToBuffer(<any>[]), 'throws on arrays')
+  st.throws(() => intToBuffer(<any>(() => {})), 'throws on arrays')
```
jochem-brouwer, line 304: nit: function

packages/util/test/bytes.spec.ts
```diff
@@ -285,3 +287,43 @@ tape('baToJSON', function (t) {
...
+  st.throws(() => intToHex(<any>NaN), 'throws on NaN')
+  st.throws(() => intToHex(<any>undefined), 'throws on undefined')
+  st.throws(() => intToHex(<any>null), 'throws on null')
+  st.throws(() => intToHex(<any>-1), 'throws on negative numbers')
+  st.throws(() => intToHex(<any>1.05), 'throws on decimal numbers')
+  st.throws(() => intToHex(<any>{}), 'throws on objects')
+  st.throws(() => intToHex(<any>true), 'throws on true')
+  st.throws(() => intToHex(<any>false), 'throws on false')
+  st.throws(() => intToHex(<any>[]), 'throws on arrays')
+  st.throws(() => intToHex(<any>(() => {})), 'throws on arrays')
```
jochem-brouwer, line 324: nit: function

Comandante318 (commented):
?

PR #1508, opened 2021-10-01
client: add websocket support

### LINKED ISSUE #1503: Client: Support Websocket endpoint in addition to HTTP
Merge relevant, see [Engine API](https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.1/src/engine/interop/specification.md), also take note of the priority and urgency label on the sideline (may change over time).

Feedback from Ryan required before starting on an implementation: better, but not necessary

---

We currently only support serving [JSON-RPC](https://eth.wiki/json-rpc/API) via HTTP. This should be extended to also support serving via WebSocket.

Fortunately enough WebSocket support has been added to the [Jayson](https://github.com/tedeh/jayson) server we are using (see `bin/cli.ts`) in the latest version [v3.6.4](https://github.com/tedeh/jayson#changelog-only-notable-milestoneschanges) - which even is in our current version range.

So this might be relatviely easy to add.

acolytec3:
Fixes #1503 

Adds a WSS endpoint to the RPC server (defaults to port 8544).

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -80,10 +80,15 @@ const args = require('yargs')
       boolean: true,
       default: Config.RPC_DEFAULT,
     },
-    rpcport: {
+    rpcHttpPort: {
       describe: 'HTTP-RPC server listening port',
       number: true,
-      default: Config.RPCPORT_DEFAULT,
+      default: Config.RPCHTTPPORT_DEFAULT,
+    },
+    rpcWssPort: {
+      describe: 'WSS-RPC server listening port',
+      number: true,
+      default: Config.RPCWSSPORT_DEFAULT,
     },
     rpcaddr: {
       describe: 'HTTP-RPC server listening interface',
@@ -194,11 +199,13 @@ async function runNode(config: Config) {
 }
 
 function runRpcServer(client: EthereumClient, config: Config) {
-  const { rpcport, rpcaddr } = config
+  const { rpcHttpPort, rpcWssPort, rpcaddr } = config
   const manager = new RPCManager(client, config)
   const server = new RPCServer(manager.getMethods())
-  config.logger.info(`RPC HTTP endpoint opened: http://${rpcaddr}:${rpcport}`)
-  server.http().listen(rpcport)
+  config.logger.info(`RPC HTTP endpoint opened: http://${rpcaddr}:${rpcHttpPort}`)
+  config.logger.info(`RPC WSS endpoint opened: http://${rpcaddr}:${rpcWssPort}`)
+  server.http().listen(rpcHttpPort)
+  server.websocket({ port: rpcWssPort })
 
   return server
 }
@@ -393,7 +400,8 @@ async function run() {
     port: args.port,
     multiaddrs: args.multiaddrs ? parseMultiaddrs(args.multiaddrs) : undefined,
     rpc: args.rpc,
-    rpcport: args.rpcport,
+    rpcHttpPort: args.rpcHttpPort,
+    rpcWssPort: args.rpcWssPort,
     rpcaddr: args.rpcaddr,
     loglevel: args.loglevel,
     maxPerRequest: args.maxPerRequest,
--- a/packages/client/lib/config.ts
+++ b/packages/client/lib/config.ts
@@ -99,10 +99,16 @@ export interface ConfigOptions {
    *
    * Default: 8545
    */
-  rpcport?: number
+  rpcHttpPort?: number
 
   /**
-   * HTTP-RPC server listening interface
+   * WSS-RPC server listening port
+   *
+   * Default: 8544
+   */
+  rpcWssPort?: number
+  /**
+   * RPC server listening interface
    */
   rpcaddr?: string
 
@@ -214,7 +220,8 @@ export class Config {
   public static readonly TRANSPORTS_DEFAULT = ['rlpx', 'libp2p']
   public static readonly PORT_DEFAULT = 30303
   public static readonly RPC_DEFAULT = false
-  public static readonly RPCPORT_DEFAULT = 8545
+  public static readonly RPCHTTPPORT_DEFAULT = 8545
+  public static readonly RPCWSSPORT_DEFAULT = 8544
   public static readonly RPCADDR_DEFAULT = 'localhost'
   public static readonly LOGLEVEL_DEFAULT = 'info'
   public static readonly MAXPERREQUEST_DEFAULT = 50
@@ -234,7 +241,8 @@ export class Config {
   public readonly port?: number
   public readonly multiaddrs?: Multiaddr[]
   public readonly rpc: boolean
-  public readonly rpcport: number
+  public readonly rpcHttpPort: number
+  public readonly rpcWssPort: number
   public readonly rpcaddr: string
   public readonly loglevel: string
   public readonly maxPerRequest: number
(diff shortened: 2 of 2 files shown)


acolytec3:
Can be tested with the [`wscat`](https://www.npmjs.com/package/wscat) package.  Sample output below:
```shell
jim:~/development/ethereumjs-monorepo/packages/client$ wscat -c ws://0.0.0.0:8544
Connected (press CTRL+C to quit)
> {"jsonrpc":"2.0","method":"web3_clientVersion","params":[],"id":53}
< {"jsonrpc":"2.0","id":53,"result":"EthereumJS/0.1.0/linux/node14.16.1"}
> {"jsonrpc":"2.0","method":"net_version","params":[],"id":53}
< {"jsonrpc":"2.0","id":53,"result":"5"}
> {"jsonrpc":"2.0","method":"net_peerCount","params":[],"id":53}
< {"jsonrpc":"2.0","id":53,"result":"0x3"}
> {"jsonrpc":"2.0","method":"eth_syncing","params":[],"id":53}
< {"jsonrpc":"2.0","id":53,"result":{"startingBlock":"0x2a7156","currentBlock":"0x2a7224","highestBlock":"0x5560e1"}}
> {"jsonrpc":"2.0","method":"admin_nodeInfo","params":[],"id":53}
< {"jsonrpc":"2.0","id":53,"result":{"name":"EthereumJS/0.1.0/linux/node14.16.1","enode":"enode://cd8e0cbb6354547f0d3906854198f765ff01a2147d614a2fa60b2e06972c207c32e26c95b85cbd970d2d96a6793160df20b338dcc4c3d2a4b0b817146e027f05@[::]:30303","id":"cd8e0cbb6354547f0d3906854198f765ff01a2147d614a2fa60b2e06972c207c32e26c95b85cbd970d2d96a6793160df20b338dcc4c3d2a4b0b817146e027f05","ip":"::","listenAddr":"[::]:30303","ports":{"discovery":30303,"listener":30303},"protocols":{"eth":{"difficulty":"1","genesis":"0xbf7e331f7f7c1dd2e05159666b3bf8bc7a8a3a9eb1d518969eab529dd9b88c1a","head":"0x0000000000000000000000000000000000000000000000000000000000000000","network":"5"}}}}
> {"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":53}
< {"jsonrpc":"2.0","id":53,"result":"0x2a72e6"}
```

packages/client/bin/cli.ts
```diff
@@ -80,10 +80,15 @@ const args = require('yargs')
...
+    rpcHttpPort: {
       describe: 'HTTP-RPC server listening port',
       number: true,
-      default: Config.RPCPORT_DEFAULT,
+      default: Config.RPCHTTPPORT_DEFAULT,
+    },
+    rpcWssPort: {
+      describe: 'WSS-RPC server listening port',
+      number: true,
+      default: Config.RPCWSSPORT_DEFAULT,
```
holgerd77, line 91: I know this is still WIP, but just to mention: we likely also want to have the option to activate/deactivate separtely here? So with `--rpcWss` I guess?

Side question: is this `Wss` the `Wss` from `Web Socket Secure` as mentioned in Wikipedia? 🤔

> The WebSocket protocol specification defines ws (WebSocket) and wss (WebSocket Secure)

acolytec3: > I know this is still WIP, but just to mention: we likely also want to have the option to activate/deactivate separtely here? So with `--rpcWss` I guess?
> 
Ah, good catch.  I'll update that in my next commit.
> Side question: is this `Wss` the `Wss` from `Web Socket Secure` as mentioned in Wikipedia? thinking
> 
> > The WebSocket protocol specification defines ws (WebSocket) and wss (WebSocket Secure)

I guess I was thinking WebSocket secure but this truly will be `ws://...`.  Will update and have better names args.

acolytec3:
@ryanio I added a couple of tests to the cli tests to check connectivity to the HTTP and WS RPC servers.  Are you okay with them being in the same test suit as your block import tests or do you want them split out?

ryanio:
> @ryanio I added a couple of tests to the cli tests to check connectivity to the HTTP and WS RPC servers. Are you okay with them being in the same test suit as your block import tests or do you want them split out?

hm, thanks for asking, I'd prefer if we made a tiny new spec file to test these, since it kind of clutters and is non relevant to the cli sync test. I think even if it is just a few lines in its own file it will be nicely separated - nice idea for the tests!

ryanio:
on a side note, it's pretty confusing to have both camelCase and lowercased rpc args, will probably get even more confusing over time to remember each variation, should we just bite the bullet and camelCase everything? The other option is to use periods like geth which does nicely separate the config for the same categories but we don't have a lot of the more narrow values yet so it might be overkill, but does leave more room to grow.

packages/client/bin/cli.ts
ryanio: for default-on values it would be cool to also work with them being passed as `false` to turn them off. I know WS probably won't be on by default, but if it was, then e.g. to disable them with `--rpcWss=false` - it would be nice to support this across the other default-on values in general so there is an expected default way to disable things. Just a thought!

holgerd77:
Yeah, I would also pledge for just camelcase everything.

acolytec3:
Sounds good. I can address all of those as part of my next round of fixes.

acolytec3:
Here are the latest changes:
-- removed `--rpc` CLI parameter and replaced the logic for enabling the rpc to look at the contents of the two `--rpcWsPort` and `--rpcHttpPort` params.  If they're numbers, enable those RPC endpoints
-- Moved the rpc cli tests to their own script

I haven't done the camelCase switch (or the `rpc.http...` approach of geth) for all of the RPC parameters but happy to pick that up in this PR if we want to finish it out.  Otherwise, it should be ready for review again.

ryanio:
looking good, thanks, can you revert the changes to `cli.spec.ts`?

I am thinking we can rename the files for increased clarity to: `cli-sync.spec.ts` and `cli-rpc.spec.ts`

acolytec3:
> looking good, thanks, can you revert the changes to `cli.spec.ts`?
> 
> I am thinking we can rename the files for increased clarity to: `cli-sync.spec.ts` and `cli-rpc.spec.ts`

Done!

packages/client/bin/cli.ts
```diff
@@ -199,12 +197,16 @@ async function runNode(config: Config) {
...
-  config.logger.info(`RPC HTTP endpoint opened: http://${rpcaddr}:${rpcport}`)
-  server.http().listen(rpcport)
-
+  typeof rpcHttpPort === 'number' &&
+    config.logger.info(`RPC HTTP endpoint opened: http://${rpcaddr}:${rpcHttpPort}`)
+  typeof rpcHttpPort === 'number' && server.http().listen(rpcHttpPort)
+  rpcWsPort !== false && typeof rpcWsPort === 'number' && server.websocket({ port: rpcWsPort })
+  rpcWsPort !== false &&
+    typeof rpcWsPort === 'number' &&
+    config.logger.info(`RPC WS endpoint opened: ws://${rpcaddr}:${rpcWsPort}`)
```
ryanio, line 209: would prefer if we could rewrite this to use if() statements rather than inlined short-circuiting behavior of the `&&` operator (kind of hard to read/parse)

```suggestion
  if (typeof rpcHttpPort === 'number') {
    server.http().listen(rpcHttpPort)
    config.logger.info(`RPC HTTP endpoint opened: http://${rpcaddr}:${rpcHttpPort}`)
  }
  if (typeof rpcWsPort === 'number') {
    server.websocket({ port: rpcWsPort })
    config.logger.info(`RPC WS endpoint opened: ws://${rpcaddr}:${rpcWsPort}`)
  }
```

packages/client/test/cli/cli-rpc.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    child.on('close', (code) => {
+      if (code > 0) {
+        t.fail(`child process exited with code ${code}`)
+        end(child, hasEnded)
+        t.end()
+      }
+    })
+  })
+})
```
holgerd77, line 94: Ah, that's really a nice new test setup! 🙂 👍

Issue #1468, opened 2021-09-10
Monorepo: Distribute modules with esm as well


kumavis:
since the packages are authored as typescript esm it should be easy to distribute as esm in addition to cjs

kumavis:
eg `@ethereumjs/tx` only provides a "main" https://npmfs.com/package/@ethereumjs/tx/3.3.0/package.json#L16

holgerd77:
Some context on this, [here](https://nodejs.org/api/esm.html#esm_modules_ecmascript_modules) is the Node.js ESM documentation.

Some articles on possible approaches:

- [How to Create a Hybrid NPM Module for ESM and CommonJS](https://www.sensedeep.com/blog/posts/2021/how-to-create-single-source-npm-module.html), Feb 12 2021 (I like that one on a very first read)
- [Hybrid npm packages (ESM and CommonJS)](https://2ality.com/2019/10/hybrid-npm-packages.html), 2019-10-22 (a bit older)

Related issue: #978

What do you guys think, might be a good time to prioritize a bit on this? (@kumavis thanks for the suggestion)

paulmillr:
This should be delayed until node.js drops support for v14, which is 2023+. The constraint list:

It is not possible to use dynamic requires same code to be used in node.js and browser (NPM etc). I'm talking about this pattern:

```ts
export const crypto: { node?: any; web?: any } = (() => {
  const webCrypto = typeof self === 'object' && 'crypto' in self ? self.crypto : undefined;
  const nodeRequire = typeof module !== 'undefined' && typeof require === 'function';
  return {
    node: nodeRequire && !webCrypto ? require('crypto') : undefined,
    web: webCrypto,
  };
})();
```

PR #1514, opened 2021-10-03
feat(vm): use promisify from node.js

talentlessguy:
util.promisify is supported since Node.js 8, there's no point in using a polyfill.

That polyfill also breaks in ESM environments so removing it would be helpful.

### DIFF
--- a/packages/vm/package.json
+++ b/packages/vm/package.json
@@ -59,8 +59,7 @@
     "ethereumjs-util": "^7.1.2",
     "functional-red-black-tree": "^1.0.1",
     "mcl-wasm": "^0.7.1",
-    "rustbn.js": "~0.2.0",
-    "util.promisify": "^1.0.1"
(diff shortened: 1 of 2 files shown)


acolytec3 (approved):
LGTM!

ryanio:
@talentlessguy cool, thanks so much!

ryanio:
@talentlessguy we recently had a suggestion to additionally export ESM modules (https://github.com/ethereumjs/ethereumjs-monorepo/issues/1468) and it sounds like that would be pretty helpful for you too so it's something we can try to tackle relatively soon!

talentlessguy:
@ryanio alright I can try pull requesting ESM output

and also please publish a patch version for the VM package

talentlessguy:
and please take a look at my ethereumjs/rlp PR please •-•

PR #1516, opened 2021-10-05
vm: fix accountexists bug in pre-SD hardforks

### LINKED ISSUE #1502: `RunBlock` calculates incorrect gas usage for contract interaction
When we run the client on mainnet, it stops syncing at block [1230833](https://etherscan.io/block/1230833) with the below error:
```
  vm:block ---------------------------------------------------------------------------------------------------- +0ms
  vm:block Running block hash=cae0bfedc3e72c255f89762e440ae7bdba387b18f55974728b078adc1e878394 number=1230833 hardfork=homestead +0ms
  vm:block Set provided state root a7af0b7d460c5b6a87214ed76bddcca518c417be68439f80a074494366126350 +0ms
  vm:block block checkpoint +1ms
  vm:block Apply transactions +1ms
  vm:block ---------------------------------------------------------------------------------------------------- +57ms
  vm:block Add tx gas used (49146) to total block gas usage (-> 49146) +0ms
  vm:block ---------------------------------------------------------------------------------------------------- +3ms
  vm:block Add tx gas used (21000) to total block gas usage (-> 70146) +0ms
  vm:block Assign block rewards +1ms
  vm:block Add uncle reward 1250000000000000000 to account 0xf8b483dba2c3b7176a3da549ad41a48bb3121069 (-> 6601910985267258093980) +1ms
  vm:block Add miner reward 5156250000000000000 to account 0x4bb96091ee9d802ed039c4d1a5f6216f90f81b01 (-> 141091438505783008181) +0ms
  vm:block Received block results gasUsed=70146 bloom=00000000000000000000000000000000000000000000000000... (256 bytes) receiptRoot=2bd009c38528a66c3dfa59ab0634fcea3769e78740ff4f64ad988b8fdb561115 receipts=2 txResults=2 +0ms
  vm:block block checkpoint committed +3ms
  vm:block Invalid receiptTrie received=2bd009c38528a66c3dfa59ab0634fcea3769e78740ff4f64ad988b8fdb561115 expected=76aae47d1d953d3188ffc948d496f95a0cf44877351b47c0e8f068671adc0256 +0ms
WARN [09-30|17:30:38] Execution of block number=1230833 hash=cae0bfed... hardfork=homestead failed:
```

From looking at the transactions in that block, it appears that we are incorrectly calculating the gas usage associated with [`0x26ea8719eeca5737f8ca872bca1ac53cea9bf6e11462dd83317c2e66a4e43d7b`](https://etherscan.io/tx/0x26ea8719eeca5737f8ca872bca1ac53cea9bf6e11462dd83317c2e66a4e43d7b).  From the logs above, we are calculating the gas used as 49146 while Etherscan reports a gas usage of 74146 which is resulting in us calculating a `receiptTrie` that doesn't match the one provided with the block and resulting in the error.

acolytec3:
Fixes #1502 based on test case and suggestion from @jochem-brouwer.  I've added the test case to the VM tests though not sure if we want to restructure it along the lines of `ethereum-tests` somehow.  Seems like there ought to be a good way to do that and I can explore how to open that PR if we like this approach.

I have confirmed that this resolves the specific issue when syncing the client on mainnet and it is able to progress beyond this block but I've encountered another block `1233073` with an invalid `ReceiptsTrie` so will need some additional review to see if my code change caused this new issue or if it's something else now.


packages/vm/src/state/stateManager.ts
```diff
@@ -603,7 +603,14 @@ export default class DefaultStateManager implements StateManager {
   async accountExists(address: Address): Promise<boolean> {
     const account = this._cache.lookup(address)
     if (account && !this._cache.keyIsDeleted(address)) {
-      return true
+      if (account.balance.isZero() && !this._common.gteHardfork('spuriousDragon')) {
```
jochem-brouwer, line 606: A more complete check here is to use `account.isEmpty()` (this would return `true` for contracts which have 0 balance :) )

packages/vm/tests/accountexists.spec.ts
```diff
@@ -0,0 +1,45 @@
+import Common, { Chain, Hardfork } from '@ethereumjs/common'
+import { Address, BN } from 'ethereumjs-util'
+import tape from 'tape'
+import VM from '../src'
+
+tape('correctly apply new account gas fee on pre-Spurious Dragon hardforks', async (t) => {
+  // setup the accounts for this test
```
jochem-brouwer, line 7: For completeness of this test and some background info I think it would be nice if we can add the txhash of this specific test.

packages/vm/src/state/stateManager.ts
acolytec3: Thanks!  Updated in the latest commit.

packages/vm/tests/accountexists.spec.ts
acolytec3: Review the comments I added and let me know if this is clear enough.

packages/vm/tests/accountexists.spec.ts
```diff
@@ -4,6 +4,10 @@ import tape from 'tape'
 import VM from '../src'
 
 tape('correctly apply new account gas fee on pre-Spurious Dragon hardforks', async (t) => {
+  // This transaction https://etherscan.io/tx/0x26ea8719eeca5737f8ca872bca1ac53cea9bf6e11462dd83317c2e66a4e43d7b produced an error
+  // in our VM where we were not applying the new account gas fee (25k wei) for account creation in hardforks before Spurious Dragon
```
jochem-brouwer, line 8:
```suggestion
  // in our VM where we were not applying the new account gas fee (25k) for account creation in hardforks before Spurious Dragon
```

packages/vm/tests/accountexists.spec.ts
acolytec3: Done!

ryanio:
smallest nit ever, but let's move this file and rename to camelCase: `packages/vm/test/api/state/accountExists.spec.ts`

jochem-brouwer (approved):
LGTM

jochem-brouwer:
I don't think we should run all hardforks in VM PRs, but we should if it clearly hits a hardfork which is not covered by tests (mostly rather old forks).

PR #1517, opened 2021-10-05
feat(util): remove ethjs-util dependency

### LINKED ISSUE #1474: Replace ethjs-util dependency with something else
The ethjs-util package is on version 0.1.6. The API for that package could change and would end up breaking this package. 

Additionally, ethjs-util is including two 1 liner packages (also made by the same author):

https://github.com/SilentCicero/strip-hex-prefix/blob/master/src/index.js
`return isHexPrefixed(str) ? str.slice(2) : str;`

https://github.com/SilentCicero/is-hex-prefixed/blob/master/src/index.js
`return str.slice(0, 2) === '0x';`

This just creates extra lines in our lock files and extra requests on install. Personally I think we should discourage these 1 line packages in the NodeJS community.

talentlessguy:
fixes #1474 

this PR replaces `ethjs-util` dependency with an `internal.ts` file which re-exports all the used functions.

this has a list of benefits:
- lesser maintenance burden (lesser dependencies to care about)
- better typings and avoiding usage of deprecated APIs (e.g. `new Buffer`)
- smaller bundle/install size


ryanio:
@talentlessguy thanks for putting this together!

packages/util/src/internal.ts
```diff
@@ -0,0 +1,224 @@
...
+ * @param str the string input value
+ * @return a boolean if it is or is not hex prefixed
+ * @throws if the str input is not a string
+ */
+export function isHexPrefixed(str: string): boolean {
+  if (typeof str !== 'string') {
+    throw new Error(
+      "[is-hex-prefixed] value must be type 'string', is currently type " +
+        typeof str +
+        ', while checking isHexPrefixed.'
```
ryanio, line 36:
```suggestion
      `[isHexPrefixed] input must be type 'string', received type ${typeof str}`
```

packages/util/src/internal.ts
```diff
@@ -0,0 +1,224 @@
...
+  return str.slice(0, 2) === '0x'
+}
+
+/**
+ * Removes '0x' from a given `String` if present
+ * @param str the string value
+ * @return  a string by pass if necessary
+ */
+export const stripHexPrefix = (str: string): string => {
+  if (typeof str !== 'string') return str
```
ryanio, line 49:
```suggestion
  if (typeof str !== 'string') {
    `[stripHexPrefix] input must be type 'string', received ${typeof str}`
  }
```

packages/util/src/internal.ts
```diff
@@ -0,0 +1,224 @@
...
+ *
+ */
+export function arrayContainsArray(
+  superset: unknown[],
+  subset: unknown[],
+  some?: boolean
+): boolean {
+  if (Array.isArray(superset) !== true) {
+    throw new Error(
+      `[ethjs-util] method arrayContainsArray requires input 'superset' to be an array got type '${typeof superset}'`
```
ryanio, line 103:
```suggestion
      `[arrayContainsArray] method requires input 'superset' to be an array, got type '${typeof superset}'`
```

packages/util/src/internal.ts
```diff
@@ -0,0 +1,224 @@
...
+    )
+  }
+  if (Array.isArray(subset) !== true) {
+    throw new Error(
+      `[ethjs-util] method arrayContainsArray requires input 'subset' to be an array got type '${typeof subset}'`
```
ryanio, line 108:
```suggestion
      `[arrayContainsArray] method requires input 'subset' to be an array, got type '${typeof subset}'`
```

packages/util/src/internal.ts
```diff
@@ -0,0 +1,224 @@
...
+    )
+  }
+
+  return subset[(Boolean(some) && 'some') || 'every']((value) => superset.indexOf(value) >= 0)
```
ryanio, line 112: hard for me to read/parse this if it can be improved/simplified at all

packages/util/src/internal.ts
```diff
@@ -0,0 +1,224 @@
...
+ *
+ * @param  params
+ * @param  key
+ * @param  allowEmpty
+ * @returns output just a simple array of output keys
+ */
+export function getKeys(params: any[], key: any, allowEmpty?: boolean) {
+  if (!Array.isArray(params)) {
+    throw new Error(
+      `[ethjs-util] method getKeys expecting type Array as 'params' input, got '${typeof params}'`
```
ryanio, line 190:
```suggestion
      `[getKeys] method expects input 'params' to be an array, got ${typeof params}`
```

packages/util/src/internal.ts
```diff
@@ -0,0 +1,224 @@
...
+    )
+  }
+  if (typeof key !== 'string') {
+    throw new Error(
+      `[ethjs-util] method getKeys expecting type String for input 'key' got '${typeof key}'.`
```
ryanio, line 195:
```suggestion
      `[getKeys] method expects input 'key' to be type 'string', got ${typeof params}`
```

ryanio:
wondering if we should copy over tets for each method to a `internal.spec.ts` file, coverage doesn't seem affected though.

left some review to make the code easier to read / unify the error messages. probably will need a lint:fix after. 

thx again for working on this!

packages/util/src/internal.ts
talentlessguy: @ryanio this was in ethjs-util source

ryanio: that's fine, was just making a comment about it

kingjerod:
Nice work @talentlessguy

packages/util/src/internal.ts
talentlessguy: i changed it to throw error

packages/util/src/internal.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  if (a.length % 2) a = `0${a}`
+
+  return a
+}
+
+/**
+ * Get the binary size of a string
+ * @param str
+ * @return
```
ryanio, line 71:
```suggestion
 * @returns the number of bytes contained within the string
```

packages/util/src/internal.ts
```diff
@@ -0,0 +1,218 @@
...
+    const code = stringValue.charCodeAt(i)
+    const n = code.toString(16)
+    hex += n.length < 2 ? `0${n}` : n
+  }
+
+  return `0x${hex}`
+}
+
+/**
+ * getKeys([{a: 1, b: 2}, {a: 3, b: 4}], 'a') => [1, 3]
```
ryanio, line 173:
```suggestion
 * Returns the keys from an array of objects.
 * Example: getKeys([{a: 1, b: 2}, {a: 3, b: 4}], 'a') => [1, 3]
```

holgerd77:
@talentlessguy thanks a lot, that's so great that we have this in now! 😄 

Small note as I just discussed with Ryan: over time we likely want to have our own implementations of these functions, so I guess this is rather an inbetween step, but one which totally suffices for now.

We can then on occasion and one by one take a method over respectively rewrite it within our own code base and remove from this `internal.ts` file.

talentlessguy:
any ETA on releasing this in a patch version?

holgerd77:
We can do relatively soon, can't absolutely promise tomorrow, but latest within the next 2-3 days should be possible.

holgerd77:
@talentlessguy just to let you know respectively keep you up to date: there is another `ethereumjs-util` related bugfix PR open and discussed right now at #1522 (feel free to engage) which I would eventually want to take into a release, so release might be another 1-3 days off until we have settled there.

PR #1522, opened 2021-10-11
util: fix toUtf8 when there is a zero at the end of a line

vserpokryl:
Problem:

```js
import { toUtf8 } from 'ethereumjs-util';

const hex = Buffer.from('bip').toString('hex'); // '626970'
toUtf8(hex); // '\x06&�'
```

This [line](https://github.com/ethereumjs/ethereumjs-monorepo/blob/986eb64dde1fa3761b19749d3d6aab8b06b6cbc7/packages/util/src/internal.ts#L115) removes zeros at the beginning and end plus add a zero at the beginning in `padToEven` (same problem in `ethjs-util` dependency, which was removed in #1517)

```js
const bufferValue = Buffer.from(padToEven(stripHexPrefix(hex).replace(/^0+|0+$/g, '')), 'hex');

// in padToEven(stripHexPrefix(hex).replace(/^0+|0+$/g, '')
// from '626970' we get '062697'
```

Solution: remove only the grouped zeros at the beginning and end

`/^0+|0+$/g` -> `/^(00)+|(00)+$/g`

### DIFF
--- a/packages/util/src/internal.ts
+++ b/packages/util/src/internal.ts
@@ -112,7 +112,8 @@ export function arrayContainsArray(
  * @returns ascii string representation of hex value
  */
 export function toUtf8(hex: string) {
-  const bufferValue = Buffer.from(padToEven(stripHexPrefix(hex).replace(/^0+|0+$/g, '')), 'hex')
+  const zerosRegexp = /^(00)+|(00)+$/g
+  const bufferValue = Buffer.from(padToEven(stripHexPrefix(hex).replace(zerosRegexp, '')), 'hex')
 
   return bufferValue.toString('utf8')
 }
--- a/packages/util/test/internal.spec.ts
+++ b/packages/util/test/internal.spec.ts
@@ -42,6 +42,7 @@ tape('internal', (t) => {
   })
   t.test('toUtf8', (st) => {
     st.equal(toUtf8(buf.toString('hex')), 'hello')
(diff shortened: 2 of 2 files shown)


holgerd77:
Hi there, thanks for the PR! 😄 

I am new to this function since we just integrated and also actually do not use at all in our own code base, so I would need some bigger turn to get a bit of a feeling what is going on.

One side question you might also want to answer: in what context is such a function actually used at all? Cannot really well imagine something? 🤔 Just so that I get a bit of a better grip on how this should behave.

More concrete question: why are 0-values stripped at all? At least the stripping of the 0-values at the end (if in big endian notation) would actually change the value of the hex value? So if you switch e.g. `665400` -> `6654`? 🤔

vserpokryl:
@holgerd77
0x00 - unnecessary NULL character (zero byte)
For example:
```js
Buffer.from('bitcoin').toString('hex'); // 626974636f696e
Buffer.from('00626974636f696e00', 'hex').toString(); // '\x00bitcoin\x00'
```
We use this to decode arguments from ethereum contract functions

holgerd77:
(so this is not necessary but rather an additional safety measure on input value validation to fail early if things are wrong)

vserpokryl:
@holgerd77 good idea 👍

holgerd77:
Ok, just published, see #1525. 😄

PR #1526, opened 2021-10-14
client: implement eth_getLogs RPC endpoint

### LINKED ISSUE #1520: Client: Implement eth_getLogs RPC endpoint
During the Merge interop event it turned out that Eth2 clients need the [eth_getLogs](https://eth.wiki/json-rpc/API#eth_getlogs) JSON RPC endpoint, so we should implement (as some hack when doing work on #1512 we just mocked the call and gave a static response).

For these kind of historical data we would need some database to store along client sync, latest state after some discussion with Ryan was that we likely wouldn't want to store in the configuration database (due to being misplaced) but also do not want to add a new database for every new type of data stored, so we settled a bit on doing one additional database for all now and future historical data additions (e.g. receipts as well if needed) and use some prefixing here for access (`log_*` or similar). If there are drawbacks which we might have overlooked let us know.

Update: from looking at the API from the RPC description linked above with the different possible filter parameters (block numbers, address, topic), I wonder if we would rather need a small relational database here? 🤔

Logs are produced along VM execution runs in [runTx.ts](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/runTx.ts) respectively [runBlock.ts](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/runBlock.ts).

gabrocheleau:
This PR resolves #1520, implementing the `eth_getLogs` RPC method.

### DIFF
--- a/packages/client/lib/rpc/modules/eth.ts
+++ b/packages/client/lib/rpc/modules/eth.ts
@@ -453,8 +453,8 @@ export class Eth {
    *   1. address of the account to check for balance
    *   2. fromBlock - integer block number, or the string "latest", "earliest" or "pending"
    *   3. toBlock - integer block number, or the string "latest", "earliest" or "pending"
-   *   4. array of 32 Bytes DATA topics. Topics are order-dependent
-   *   5. blockHash to restrict the logs to a single block.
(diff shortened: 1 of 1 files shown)


ryanio:
continued in #1556

PR #1528, opened 2021-10-15
vm: Import rlp from ethereumjs-util

kevinji:
rlp is directly used in state/stateManager.js, so it should be a dependency.

This fixes an error that yarn v2 produces:
```
Error: @ethereumjs/vm tried to access rlp, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

Required package: rlp
Required by: @ethereumjs/vm@npm:5.5.3 (via ...)
```

### DIFF
--- a/packages/client/lib/rpc/modules/eth.ts
+++ b/packages/client/lib/rpc/modules/eth.ts
@@ -8,11 +8,11 @@ import {
   bufferToHex,
   bnToHex,
   intToHex,
+  rlp,
   toBuffer,
   stripHexPrefix,
   setLengthLeft,
 } from 'ethereumjs-util'
-import { decode } from 'rlp'
 import { middleware, validators } from '../validation'
 import { INTERNAL_ERROR, INVALID_PARAMS, PARSE_ERROR } from '../error-code'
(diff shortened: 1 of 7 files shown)


kevinji:
Updated the PR to reflect the suggested changes.

ryanio:
looks great, thanks!

one last thing, we can just use `rlp.encode()` and `rlp.decode()` rather than the inserted `const { encode, decode } = rlp` which is out of place in the import statements. also i think `rlp.encode()` is more clear in the code than just `encode()` :)

I did a cursory search for more `rlp` deps in our package.jsons and it looks like there's quite a few more (blockchain, client, devp2p, trie) would appreciate if you don't mind updating those too!

ryanio:
thanks, looking good!

so basically you can remove `rlp` from all package.jsons

and update the imports from e.g.:

```typescript
import * as rlp from 'rlp'
```

to

```typescript
import { rlp } from 'ethereumjs-util'
```

or adding it to the existing of `ethereumjs-util` imports if already included, e.g.:

```typescript
import { BN, rlp } from 'ethereumjs-util'
```

this will be a nice cleanup!

PR #1530, opened 2021-10-16
client: merge work (pithos testnet)

ryanio:
This PR continues merge implementation work, focusing on compatibility with the ephemeral pithos testnet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,6 @@ Temporary Items
 .apdisk
 
 # End of https://www.toptal.com/developers/gitignore/api/node
+
+# ethereumjs 
+datadir*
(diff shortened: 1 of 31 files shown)


ryanio:
nice to be back home with the big monitor :)

here's a short story, syncing pithos from genesis and then building a block on top with the Engine API (now on its own dedicated port)

Screen Shot 2021-10-16 at 5 22 23 PM

holgerd77:
Ah, that looks really great 🥳 - didn't expect there to be so many changes already summed up - really looking to get a bit deeper here and give this some review to follow up on what's changed. 🙂 🌻

holgerd77:
Hardhat E2E test failures seem to be related with the latest Ethers v5.5 release - see [comment](https://github.com/ethers-io/ethers.js/issues/2190#issuecomment-947410205) (seems we are actually doing cross-E2E-testing for @ricmoo with our setup, how cool is that! 😄 😋 (unfortunately always only after a release)).

But anyhow: will ignore here.

PR #1546, opened 2021-10-27
vm benchmarks: skip blocks reporting false positives for alert threshold

ryanio:
After reviewing the vm github action benchmark alerts in this repository over the past week, out of the 10 blocks tested in the benchmark the alert mostly happens with blocks 9422909,9422912,9422913,9422914 and the ratio is usually only between 2-2.5 (2.52 was the highest I saw) so this PR should help alleviate some of the false positives until maybe one of us can further dig into why these blocks may have wider variance. It could be differing ci specs between runs or blocks with certain txs that have wider variance of execution time (although not sure what this could be, not sure if e.g. state or memory access could be more variable, or perhaps slow precompile usage). We could also ultimately decide that these benchmarks need to be reworked and could be disabled in the meantime.

I also noticed my editor not respecting our prettier config when working on the workflow yml files so I added a `prettier.conf.js` to root and and gave it a run for posterity :)


holgerd77:
I guess I would have a preference to rather exlude these blocks from execution for now (together with adding a TODO along explaining the decision). Then we would have rather some stable ground for *something*. For me this feels more sustainable than raising the threshold higher and higher (even as a short-term solution).

WDYGT? 🤔

acolytec3:
> I guess I would have a preference to rather exlude these blocks from execution for now (together with adding a TODO along explaining the decision). Then we would have rather some stable ground for _something_. For me this feels more sustainable than raising the threshold higher and higher (even as a short-term solution).
> 
> WDYGT? 🤔

If there's a way to just exclude blocks on that range, I think that makes sense as a short term measure.  That's better than me adding a change in the README that triggers a general slow down in the VM at 269% of baseline. 😁

ryanio:
ok have excluded the problematic blocks that I kept seeing alerts for, if we continue to get them for the other blocks then we can just disable the entire vm benchmarking and prioritize a more stable implementation.

We still have the trie benchmarks which are running quite consistent https://ethereumjs.github.io/ethereumjs-monorepo/dev/bench/trie/

holgerd77:
@ryanio we should really add these benchmark URLs to the READMEs of trie and VM to have these more present and accessable. Have added to the tiny-fixes issue.

holgerd77 (approved):
LGTM

Issue #1537, opened 2021-10-21
Monorepo: Fix npm audit security vulnerabilities (nyc / verdaccio / tap-spec / npm-auth-to-token)


holgerd77:
Doing an `npm i` from the root directory on `master` currently gives the following result:

![grafik](https://user-images.githubusercontent.com/931137/138247091-9b497c29-6ea1-41fb-a41f-1a402ad8e66a.png)

Many of these vulnerabilities are linked to outdated versions of `nyc` (see also #832) and `verdaccio`, so we should give updates of these two libraries some priority in some first round.

While both of the libraries are dev dependencies, generally having a large number of vulnerabilities shown on `npm audit` also comes with some greater risk of overlooking some more severe vulnerabilities.

Mid-term we should optimally come to a situation where we just regularly update and get to a somewhat 0-vulnerabilities situation on all levels.

holgerd77:
Ok, looking into the remaining vulnerabilities here. One other high severe vulnerability is coming from [tap-spec](https://github.com/scottcorgan/tap-spec), which is a dev dependency in the VM [package.json](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/package.json) and is used for the [formatter](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/DEVELOPER.md#running-tests-with-a-reporterformatter) dev command for VM testing which routes to the [formatTest.js](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/scripts/formatTest.js) script in the `vm/scripts` folder, where `tap-spec` is used as the default formatter.

My first thought was that we might want to remove this formatter script all together since I guess this is rarely (not at all?) used. But on another thought it should already be enough to remove `tap-spec` as a default and make this script more generic. If one then wants to use this with `tap-spec` it is always possible to install globally e.g..

(so, some other context on `tap-spec`: package seems no to be maintained any more, last release is from over 3 years ago, so not possible to update (already on latest version) and therefore would be generally desirable to get rid of this package).

holgerd77:
Second high severe vulnerability is coming from [npm-auth-to-token](https://www.npmjs.com/package/npm-auth-to-token), an extremely obscure dev dependency which we should definitely remove. This has been published four years ago and there is neither a README nor a matching code/GitHub site available.

[...]

PR #1547, opened 2021-10-28
monorepo: update nyc and verdaccio

### LINKED ISSUE #832: Upgrade nyc to v15

emersonmacro:
First round for #1537, closes #832


ryanio (dismissed):
lgtm, thanks!

package.json
```diff
@@ -18,7 +18,7 @@
     "husky": "^4.2.5",
     "npm-auth-to-token": "1.0.0",
     "typedoc-plugin-markdown": "^3.10.4",
-    "verdaccio": "4.4.4"
+    "verdaccio": "4.12.2"
```
holgerd77, line 21: Asking independently from this PR: is it necessary to have these fixed dependencies here? 🤔 (so also on `npm-auth-to-token`). In doubt we will always get in new vulnerabilties in like this.

If no one knows anything about it we should rather just test to add a `^`, can't really lead to anything grave here.

holgerd77: Update: I've just directly added here and will see how it goes.

holgerd77 (approved):
Re-approving after having dismissed the last approval by doing an additional commit.

Will merge.

Issue #1533, opened 2021-10-20
Client: Verkle Trie Integration + Proof Consumption - Tracking Issue


holgerd77:
There are recent efforts to replace the current MPT state trie storage datastructure with a new trie format called Verkle Trie which is described in this [draft EIP](https://notes.ethereum.org/@vbuterin/verkle_tree_eip). 

As discussed with @gballet from the go-ethereum team we would like to prototype a verkle trie integration in the EthereumJS client with the main focus on proof consumption and block execution. Guillaume is working on Verkle Trie support for go-ethereum here https://github.com/ethereum/go-ethereum/pull/23630, and outline how a broader network setup might look like is written down [here](https://notes.ethereum.org/@gballet/rJn5JaBSK).

Effort on this is rather broad and this is a tracking issue for high level discussion and TODO checks. Dedicated tasks should rather become it's own place to discuss.

References:

- [Verkle Trie Talk](https://www.youtube.com/watch?v=xJiABLuD06g&t=18351s), Guillaume Ballet, ETHOnline, Oct 15, 2021

Some first TODOs here:

- [ ] `rust-verkle` implementation `WASM` compilation
  - [ ] Analyze if a `WASM` compilation is possible, structural needs, see https://github.com/crate-crypto/rust-verkle/issues/24
  - [ ] Compilation setup and execution from the EthereumJS side
  - [ ] First `VM` `StateManager` experiments and simple tests to do a dedicated Verkle Trie backed state manager (eventually needs some confirmation if this is the correct entry point)
- [ ] Client Proof Consumption: TODO
- [ ] Blockchain, Block, Common: understanding of proof-including block format TODO

holgerd77:
Some more references:

- Fantastic [slides](https://docs.google.com/presentation/d/13bmmLbkKte5QGiJjqpIZLgvKxsYdravIgtytQMPBQ-M/edit#slide=id.g6518b43596c43d74_0) from Dankrad for ACD summary (ACD Call [#401](https://github.com/ethereum/pm/issues/401), Oct 29 2021)
- Also from Dankrad: [Verkle Trie EIP Calculator](https://github.com/ethereum/pm/issues/401)

acolytec3:
[This repo](https://github.com/crate-crypto/rust-verkle-wasm) compiles the verkle verifier to WASM for use in our integration.

holgerd77:
[...]

PR #1548, opened 2021-10-28
VM: Add BaseStateManager

holgerd77:
Part of #1533 

This PR is adding a `BaseStateManager` class as outlined in the diagram I posted on the Discord #stateless channel extracting out non-backend related functionality from the `DefaultStateManager`. This will make it easier for us to exchange the state manager backend e.g. with a verkle trie and will generally make it easier to do custom state manager implementations (my side hope is that we can then really easily implement an RPC state manger which we might also want to directly add to the code base).

WIP


holgerd77:
Ok, rebased this and did some cleanup. This is now open for an early review. 😄 

Please do not merge yet since we (likely) don't want to have this yet in the imminent ArrowGlacier HF (and other stuff) releases, makes more sense to work more with this along the Verkle Trie work and see if/how this fits.

One bigger last change I made: I wondered if I could get the `generateGenesis()` function to the `BaseStateManager` as well - see bcbf341 and subsequent commit - but when I first changed this to direct `putAccount()` calls interestingly enough the following [eth_getBalance](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/test/rpc/eth/getBalance.spec.ts#L40) test in the client started to fail.

I digged a bit deeper. This was due to the `putAccount()` method only writing to the StateManager cache. The VM in the [eth_getBalance implementation](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/lib/rpc/modules/eth.ts#L346) is copied over, but along this copy the `Cache` from the `StateManager` is not taken over with the StateManager (see also comment on the code docs of the [StateManager.copy()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/state/stateManager.ts#L107) function). So only state up to the latest commit is taken over.

I now solved this by doing a `cache.flush()` call at the end of the `generateGenesis()` function, I would say this is a clean solution.

As some side thought we can generally question if the cache shouldn't also be copied over on the `StateManager.copy()` function? 🤔 Maybe this is generally causing some problems here and there? I would assume that no one is using this with this limitation in mind (neither we ourselves nor - even more unlikely - community users).

holgerd77:
Will also circle in @fvictoro, this might generally be interesting for you guys to ease your custom StateManager implementation.

holgerd77:
Will also circle in @fvictorio, this might generally be interesting for you guys to ease your custom StateManager implementation.

ryanio:
> As some side thought we can generally question if the cache shouldn't also be copied over on the `StateManager.copy()` function? 🤔 Maybe this is generally causing some problems here and there? I would assume that no one is using this with this limitation in mind (neither we ourselves nor - even more unlikely - community users).

Ah nice observation, haven't thought about this.

When I have been using `vm.copy()` (e.g. in the client) I am generally using it from a clean state point of view or using `setStateRoot` to set the state to specifically where I want it.

I'm not sure the use case where I'd also want the existing cache to be available, but if we find a use for it we could introduce it as a boolean option (with default to current behavior). I'll keep it in mind if/when I start to work on a "pending block" feature in the client if it could be useful there.

ryanio:
reviewed the code and lgtm! I'll hold off on an approval until we hear back from hardhat if there is any further feedback or requested changes.

holgerd77:
> I'm not sure the use case where I'd also want the existing cache to be available, but if we find a use for it we could introduce it as a boolean option (with default to current behavior). I'll keep it in mind if/when I start to work on a "pending block" feature in the client if it could be useful there.

I would rather think this doesn't even need a flag and can eventually be seen as some wider bug/behavior fix, since this doesn't take anything away from the existing behavior - correct me if I am wrong - but only adds on being more flexible (respectively also: less error prone) on a `copy()` operation.

packages/vm/src/state/baseStateManager.ts
```diff
@@ -0,0 +1,502 @@
...
+   * Saves an account into state under the provided `address`.
+   * @param address - Address under which to store `account`
+   * @param account - The account to store
+   */
+  async putAccount(address: Address, account: Account): Promise<void> {
+    if (this.DEBUG) {
+      this._debug(
+        `Save account address=${address} nonce=${account.nonce} balance=${
+          account.balance
+        } contract=${account.isContract() ? 'yes' : 'no'} empty=${account.isEmpty() ? 'yes' : 'no'}`
```
acolytec3, line 104: This is nitpicky but why are we not just using `true/false` here as returned by the `isContract` and `isEmpty` methods instead of "yes/no"?  This is purely stylistic but true/false just feels like a more helpful reporting of the state of this account address than saying "yes/no".

acolytec3 (approved):
Looks good to me (with one nitpicky comment about debug messaging that you can take or leave). For a sanity check,  I ran this branch for a few thousand blocks on mainnet client and everything seemed fine.

packages/vm/src/state/baseStateManager.ts
holgerd77: I found it a little bit more readable to have `yes/no` here (but brain needs to do a tiny bit less of internal mapping in this occasion 😛), but I have absolutely "no" opinion on things like that, happy with the one way or the other.

Will leave this now to not dismiss the review.

holgerd77:
Ok, I will now merge this in so that we can build upon this.

PR #1550, opened 2021-10-29
tx: remove event listener and add datafee caching with hardfork

emersonmacro:
Partially addresses #775 and fixes an issue found in PR #1532 by removing the `hardforkChanged` event listener and instead caching `hardfork` along with `value` for the `dataFee` cache.


holgerd77 (approved):
Ok, thanks for the persistence 🙂, I guess this should do it!

Will merge.

PR #1553, opened 2021-11-04
VM: Adopted Backwards-compatible Dynamic Gas Cost Refactoring

holgerd77:
It looks like our breaking releases are still some time away, since we just have too much other stuff to do and finish atm. I was wondering what to do with the extensive gas cost refactoring PR from @jochem-brouwer which is hanging in the `devlop` branch for some time, since this was a large amount of work and it would be nice if users could already benefit.

When having another look it just occurred to me that we can make this backwards-compatible by just remaining the old `fee` opcode parameter on the old behavior and the `number` type and instead add an additional parameter `dynamicFee` with the `BN` type and exposing the newly calculated dynamic gas costs. (On the next breaking release we can still rename if we want.)

Let me know if I am missing something here, otherwise it would be great if we can directly integrate this in the upcoming releases, this is already hanging around for far too long. 😄


holgerd77:
And another question for @fvictorio (or eventually @alcuadrado): on this PR the Hardhat tests are currently [failing](https://github.com/ethereumjs/ethereumjs-monorepo/runs/4104292955?check_suite_focus=true), it seems that you are exact-match-comparing the `opcode` object from the VM `step` event ( see related changes in `interpreter.ts`).

Since we are adding a new (optional) parameter `dynamicFee` here, these tests fail.

Is this rather a problem in your test setup and can you relax here? 🤔 Or should we guard this dynamic gas output under a flag (also a practical possibility)?

jochem-brouwer (commented):
In general this looks OK to me (in fact in retrospect we should have done it this way anyways). However, the hardhat tests are failing, we might want to look into that first?

fvictorio:
They seem to be failing because of something else, don't they?

image

Those tests check that the result of our `debug_traceTransaction` matches some traces that "we know" are correct (which IIRC means we obtained them from geth).

Is there a chance that there's a regression here, since the `memory` entry seems different?

fvictorio:
With respect to the `dynamicFee` change, I think something I could do is to link this branch locally, change our trace implementation to use this `dynamicFee` value, and check if our tests still pass.

holgerd77:
> They seem to be failing because of something else, don't they?
> 
> image
> 
> Those tests check that the result of our `debug_traceTransaction` matches some traces that "we know" are correct (which IIRC means we obtained them from geth).
> 
> Is there a chance that there's a regression here, since the `memory` entry seems different?

Ah, interesting, actually not sure how I could misread this so strongly. Anyhow. 🙂 Might be something for @jochem-brouwer to have a look into since he did the initial implementation on this and this might be something implementation specific? 🤔

jochem-brouwer:
This is nasty and it also does not make a lot of sense. I just checked, this is indeed the right trace as reported by Geth.

What confuses me is this. The situation here is PUSH 80 PUSH 40 MSTORE. At the MSTORE the stack is 0x40 0x80 (so this is the situation of the stack /before/ using MSTORE), but for some reason the memory fields are now already filled with zeros (so it does expand the memory first, and then updates the memory after the MSTORE). I find this very inconsistent. A way to fix this is to expand the memory when calling MSTORE(8) but this should thus be done in the dynamic gas handler. Which looks very confusing, as the dynamic gas handler was intended to only calculate the gas costs and not modify anything.

jochem-brouwer:
What does not make sense is that without dynamic gas factoring this test apparently passes :thinking: We first emit the trace log and then actually execute the operation, so it changes nothing to the state before emitting the step event.

holgerd77:
> What does not make sense is that without dynamic gas factoring this test apparently passes 🤔 We first emit the trace log and then actually execute the operation, so it changes nothing to the state before emitting the step event.

Thanks for having a deeper look here. Yeah, I would also guess that at the core here of some analysis is to see why these things passed without the gas changes, since the tests are even still operating on the non-dynamic values. Might be a little introduced inconsistency or some side effect from the gas cost changes itself in some way.

jochem-brouwer:
I just checked on our master branch, it also reports empty memory, so there's something very weird going on over here..? It should thus also fail on branches related to master? Are there certain conditions which have to be met to run these tests @fvictorio - maybe this PR introduces these conditions which is why the test runs (and fails) here and not on master?

jochem-brouwer:
Ah yeah right by now extending the memory inside the gas cost (so before we check if we go out-of-gas) we run in these funky test situations where we want to mstore something at a high location and thus either request massive amount of memory or it is a number which we cannot safely store in javascript ints.

I think I will have to ask geth.

holgerd77:
Any chance that we get this settled so that we might be able to do another VM release with this PR included until the end of the year (respectively: (substantially) before Christmas 😋 ?

holgerd77:
As discussed with @jochem-brouwer we would now like to continue on this to be able to release since other projects are waiting for this feature.

Current state is that the memory trace output after this PR change differs to the one from Geth. This has been reported in the following issue: https://github.com/ethereum/go-ethereum/issues/24109 There is confirmation from the go-ethereum side though that the Geth output is likely wrong and this will be fixed at some point.

Our HardHat tests are failing with this atm. I will remove the tests from being mandatory for CI runs for now. Side question: maybe it's possible to just deactivate the single failing test case from our side? 🤔

I've now rebased the PR, this is now ready for review.

I would target broader VM + other monorepo libs minor release versions for - roughly - next week.

holgerd77:
Also removed this PR from the current `develop` branch and rebased `develop`.

ryanio:
sounds great, I can review today, I can also maybe include a line in e2e-hardhat to disable or remove this one failing test instead of disabling the check completely? not sure how easy it may be so I'll give it a look.

ryanio (approved):
lgtm!

holgerd77:
Ok, great, then we should have everything together for doing the minor releases, right? 🙂

PR #1561, opened 2021-11-10
VM: Memory Allocation Performance Optimizations

holgerd77:
TODO


packages/vm/src/evm/memory.ts
```diff
@@ -76,4 +76,8 @@ export default class Memory {
 
     return returnBuffer
   }
+
+  sharedRead(offset: number, size: number): Buffer {
+    return this._store.slice(offset, offset + size)
```
holiman, line 81: https://nodejs.org/api/buffer.html#blobslicestart-end-type:

> Creates and returns a new Blob containing a subset of this Blob objects data. The original Blob is not altered.

So this does cause a copy. Why not just do 

```
     return Buffer.from(this._store, offset, size)
```
?

holgerd77: No, it does not (otherwise we would have to rewrite the whole code base 😋), `slice` is just returning a memory view, see [docs](https://nodejs.org/api/buffer.html#bufslicestart-end) on the `slice()` method:

> Returns a new Buffer that references the same memory as the original, but offset and cropped by the start and end indices.

(let me know if I misread this sentence)

Issue #1551, opened 2021-11-03
Client: Kintsugi 🍵 Tracking Issue


ryanio:
*Kintsugi is the unique and traditional Japanese art of repairing broken, chipped and cracked pottery using lacquer and gold.*

---

### Overview
This issue details the milestones for the Kintsugi 🍵 testnets in November 2021.

Kintsugi Milestones: https://notes.ethereum.org/@djrtwo/kintsugi-milestones
Kintsugi Spec: https://hackmd.io/@n0ble/kintsugi-spec

### Execution Layer

The scope of work on updating EL client software to Kintsugi specification is as follows:

* [x] Implement the new version of EIP-3675
* [x] Implement EIP-4399
* [x] Implement the new version of Engine API server

### [Milestones](https://notes.ethereum.org/rki-W2RvRf-TkVn3ZIvMFA?both#Milestones)

* [x] **M0:** Implementation started
* [x] **M1:** Spec Implemented
* [x] **M2:** 1-to-1 Interop – One EL and one CL team pair off for basic interop testing
* [x] **M3:** Many-to-Many Interop – 2 or more EL and CL teams perform interop testing
* [x] **M5:** All-to-All persistent testnet – All ready EL and CL teams create a long-lived devnet first week of December
* [x] **M6:** Kintsugi🍵 code released – All clients release Kintsugi🍵 code for public consumption

Work-in-progress PR: #1565

ryanio:
this PR is currently at the kintsugi spec v2 changeset, just rebased on master, and will implement the v3 changeset along with updated instructions for joining the latest devnet-3

PR #1565, opened 2021-11-11
merge: kintsugi updates

ryanio:
Kintsugi tracking issue: #1551 

---

This PR:
  1. Adds [EIP-4399](https://eips.ethereum.org/EIPS/eip-4399): `Supplant DIFFICULTY opcode with RANDOM`
     * [Kintsugi Spec notes](https://hackmd.io/@n0ble/kintsugi-spec#EIP-4399)
  2. [EIP-3675](https://eips.ethereum.org/EIPS/eip-3675) updates: validates extraData length cannot exceed 32 bytes
     * [Closed PRs `EIP-3675` search](https://github.com/ethereum/EIPs/pulls?q=is%3Apr+is%3Aclosed+EIP-3675)
  3. [Engine API](https://github.com/ethereum/execution-apis/blob/89070c4d6f2fbe125a906f0cca1fedda84b8a2a6/src/engine/specification.md) updates:
     * `engine_executePayloadV1`
     * `engine_forkchoiceUpdatedV1`
     *  `engine_getPayloadV1`


packages/block/src/header.ts
```diff
@@ -65,6 +65,16 @@ export class BlockHeader {
     return this.logsBloom
   }
 
+  /**
+   * EIP-4399: Post-PoS merge, `mixHash` supplanted as `random`
+   */
+  get random() {
+    if (!this._common.isActivatedEIP(4399)) {
+      throw new Error('random can only be accessed when EIP-4399 is activated')
```
holgerd77, line 73: Can you prepare the error msg with the new `_errorMsg()` function (like: `const msg = this._errorMsg(`random can only be accessed when EIP-4399 is activated${errorMsg}`).

Also, this reads a bit ambiguous and hard to grasp what random refers to, maybe `The random parameter can...`.

packages/block/src/header.ts
```diff
@@ -389,8 +399,8 @@ export class BlockHeader {
       }
     }
 
-    // Check for constant values for PoS blocks
-    if (this._common.consensusType() === ConsensusType.ProofOfStake) {
+    // Validation for PoS blocks (EIP-3675)
+    if (this._common.isActivatedEIP(3675)) {
```
holgerd77, line 403: I would like to keep things like this on the `ConsensusType.ProofOfStake` comparison whereever possible. These kind of validations are just Proof-of-Stake dependent, whereever this is coming from (from a consensus change by EIP-3675 or whatever else). This will alow us more flexibility to create e.g. pure PoS blocks from the beginning of a chain.

Logically checking for EIP-3675 is just a narrowing of things, since this by-design also switches to PoS. So directly checking for PoS is the more flexible solution which will allow for more use cases.

I guess the best mental separation is the following: for things which are directly related to the transition itself because there is some kind of trigger on something should get a 3675 check. Things which just define PoS stuff with or without a transition should get a PoS check.

Does this make sense? Or do you have other reasons not directly apparent for this change?

packages/block/src/header.ts
ryanio: yep for sure, will add, thanks

packages/block/src/header.ts
ryanio: yes for sure, didn't think too much about this change, makes sense to keep it as a check on consensusType 👍 will update

holgerd77:
Note: did some link annotations to the innitial post.

ryanio:
this PR is currently at the kintsugi spec v2 changeset, just rebased on master, and will implement the v3 changeset along with updated instructions for joining the latest devnet-3

packages/common/src/index.ts
```diff
@@ -936,6 +936,10 @@ export default class Common extends EventEmitter {
     let hfBuffer = Buffer.alloc(0)
     let prevBlock = 0
     for (const hf of this.hardforks()) {
+      if (hf.name === 'merge' && !hf.block && (hf as any).mergeForkBlock) {
+        hf.block = (hf as any).mergeForkBlock
```
holgerd77, line 940: Hmm, why don't we set the merge block as `block` in the first place? 🤔

holgerd77: Ah, ok, this is only for _calcForkHash.

packages/devp2p/src/eth/index.ts
```diff
@@ -263,6 +263,17 @@ export class ETH extends EventEmitter {
...
       }
       const forkHashB = Buffer.from(this._forkHash.substr(2), 'hex')
+
+      // EIP-3675: Set merge FORK_NEXT to mergeForkBlock
+      if (this._peer._common.hardfork() === 'london' && this._nextForkBlock.eqn(0)) {
+        const mergeForkBlock = (
+          this._peer._common.hardforks().filter((h) => h.name === 'merge') as any
+        )[0]?.mergeForkBlock
+        if (mergeForkBlock) {
+          this._nextForkBlock = new BN(mergeForkBlock)
```
holgerd77, line 273: I am not sure if we might be able to circumvent all this extra code by defining an extra hardfork `preMerge` or similar in `Common` and then this `preMerge` HF gets the block number and the `Merge` HF itself gets the difficulty switch.

Then all the devp2p switches should work automatically I guess and on first thought this shouldn't have any side effects (checks like `if HF >= london do x`) should still work, also all Merge-related stuff should still be on `if HF >= Merge`.

ryanio: that's a great idea, I was struggling on how to conceptualize this, defining a "preMerge" hardfork for this purpose sounds like it would work well, I will give it a try.

ryanio: done in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1565/commits/e0d436eddeea47c7ac8aeee29ff83d11f10582a7, synced to terminal block within a few mins! with very quick block execution

Screen Shot 2022-02-01 at 10 22 32 PM

ryanio:
ok this is now ready for review and merge, since we are quickly syncing to the terminal block, anyone can give it a try with:

```
npm run client:start -- --datadir kintsugi/datadir --gethGenesis kintsugi/config/genesis.json --saveReceipts --rpc --ws --rpcEngine --rpcEnginePort=8545 --rpcDebug --loglevel=debug --bootnodes=161.35.67.219:30303
```

It should provide a great base to continue work on merge spec updates and also implement the newly introduced [Optimistic Sync spec](https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md).

I will start syncing a lodestar node and see if I can continue to sync to tip of kintsugi with this branch. If I find any needed adjustments I will add, but can also do in a new PR if this gets merged.

acolytec3:
> (edit: hit a snag at the first PoS block, number 7052, with "invalid block stateRoot", will investigate further)

Looks like 7052 is the merge block according to the logs.  I made ti this far and then it shut down (as expected).  Need to fire up lodestar and see if we can proceed further.

acolytec3:
> > (edit: hit a snag at the first PoS block, number 7052, with "invalid block stateRoot", will investigate further)
> 
> Looks like 7052 is the merge block according to the logs. I made ti this far and then it shut down (as expected). Need to fire up lodestar and see if we can proceed further.

Nevermind, got to the same point.  7052 is the first PoS block and getting the same invalid block stateroot error you noted.

holgerd77:
> ok this is now ready for review and merge

Cool!!! 🎉 🚀 😀

Won't make it today any more, but I'll try to give this a thorough review tomorrow morning.

ryanio:
wohoo, we're syncing PoS with lodestar now :)

Screen Shot 2022-02-03 at 3 21 58 PM

holgerd77:
Won't make it to do a review on this today, eventually Monday. Otherwise open to everyone else as well of course.

acolytec3:
> Won't make it to do a review on this today, eventually Monday. Otherwise open to everyone else as well of course.

I've been through the code and run it and it looks good though I'll go through it once more with a finer toothed comb today as life allows.

Is the fact that we get stuck at block 13323 with an `invalidReceiptsTrie` a concern for merging this?  We're able to connect with Lodestar, cross the merge block, and sync up to that point.

holgerd77:
> I've been through the code and run it and it looks good though I'll go through it once more with a finer toothed comb today as life allows.

My main viewpoint on such PRs is always the "how does it affect the existing API" (mainly of the non-client packages) Point-of-View.

Just had a short look, if I haven't missed anything this seems mainly all the random functionality associated with `EIP-4399`, and this looks solid to me, e.g. the `random()` function in the Block library, naming seems good and this is also throwing if EIP is not activated.

So from my side this can be merged if all other code has gone through a thorough review as well.

Can't judge on the `invalidReceiptsTrie` breakage (sounds like it should generally be investigated though).

packages/client/kintsugi/Dockerfile
```diff
@@ -1,14 +1,13 @@
-FROM node:14-alpine as build
+FROM node:16-alpine as build
 WORKDIR /usr/app
-RUN apk update && apk add --no-cache g++ make python bash git && rm -rf /var/cache/apk/*
+RUN apk update && apk add --no-cache bash git && rm -rf /var/cache/apk/*
 
 WORKDIR /usr/app
```
acolytec3, line 5: Isn't this duplicative?  I'm still a novice at Docker but doesn't this create a new layer in the image unnecessarily?

packages/client/lib/miner/pendingBlock.ts
```diff
@@ -0,0 +1,171 @@
...
+  stop(payloadId: Buffer) {
+    const payload = this.pendingPayloads.find((p) => p[0].equals(payloadId))
+    if (!payload) return
+    void payload[1].revert()
+  }
+
+  /**
+   * Returns the completed block
+   */
+  async build(payloadId: Buffer): Promise<Block | void> {
```
acolytec3, line 117: What is the difference in usage between this and the `start` method?  They both appear to be async functions that construct a block with the main difference being that `build` returns the block and `start` returns the payload id.

ryanio:
Well not "moot" since we have some kind of spec mismatch due to the invalid receiptTrie, but I will be looking into that today and can definitely address in a follow-up PR if this one has been merged :)

packages/vm/tests/api/EIPs/eip-4399-supplant-difficulty-opcode-with-random.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+      {
+        header: {
+          ...header,
+          mixHash: random,
+        },
+      },
+      { common }
+    )
+    await vm.runCode({ ...runCodeArgs, block })
+    st.ok(stack[0].eq(new BN(random)), '0x44 returns RANDOM (Merge)')
```
acolytec3, line 52: This is excellent!  Learned a lot just from reading through this test.

acolytec3 (changes requested):
What do you think about writing additional tests for the positive case in `pendingBlock`?  That's what's causing the code coverage to fail.  I'd be happy to take a stab at if but not sure how much this API might change going forward to know if it's worth it at this stage.

packages/client/kintsugi/Dockerfile
ryanio: isn't what duplicative? here i just simplified some of the modules that were being installed (didn't need g++, make, python)

packages/client/lib/miner/pendingBlock.ts
ryanio: this is from the engine API. `start` is called when the block should start being prepared, and returns an identifier to CL. Then when CL is ready for the block it will call `build` with the identifier.

ryanio:
> What do you think about writing additional tests for the positive case in pendingBlock? That's what's causing the code coverage to fail.

good idea, thanks for noticing, I will do that now

packages/client/kintsugi/Dockerfile
acolytec3: Oh, sorry, it commented on the wrong line.  I meant line 5
`WORKDIR /usr/app`

Isn't this duplicative of line 2?

ryanio: ah hmm, `WORKDIR` is just like `cd`, so yeah looks duplicative but probably harmless, i can try without it when i give the docker image a spin today and can remove if unnecessary

packages/client/lib/miner/pendingBlock.ts
acolytec3: Got it.  And the section of code below in the `build` method just adds additional transactions to the already constructed block indicated by `payloadId` if it's not already full?

packages/client/lib/miner/pendingBlock.ts
```diff
@@ -0,0 +1,171 @@
...
+    const payload = this.pendingPayloads.find((p) => p[0].equals(payloadId))
+    if (!payload) {
+      return
+    }
+    const builder = payload[1]
+
+    // Add new txs that the pool received
+    const txs = (
```
acolytec3, line 125: Would it make sense to add a check like this from below
`if (builder.gasUsed.gt((builder as any).headerData.gasLimit.subn(21000)))` before you start looking in the transaction pool to add additional transactions?  That way, if the block is already full from the earlier call to `start`, you don't waste the time rummaging through the transaction pool.

packages/client/lib/miner/pendingBlock.ts
ryanio: that's just our own hacky implementation here, see the opening comment in the file:

```js
/**
 * In the future this class will build a pending block by keeping the
 * transaction set up-to-date with the state of local mempool until called.
 *
 * For now this simple implementation just adds txs from the pool when
 * started and called.
 */
```

packages/client/lib/miner/pendingBlock.ts
ryanio: i think it's fine, this code is mostly unoptimized scaffolding to get some kind of basic functionality. this error would be thrown a few lines later that is handled appropriately.

packages/client/lib/miner/pendingBlock.ts
ryanio: This implementation is really suboptimal because we should be shuffling txs to maximize MEV, not just adding new txs to the end of a non-full block, but since we are not targeting to be run as a tip of chain client it's less important to prioritize.

acolytec3 (commented):
At this point, I think everything looks good.  Will wait on the new tests for `pendingBlock` (or let me know if you want me to take a whack at them) before approving.

ryanio:
thanks! I will be adding a file for `test/miner/pendingBlock.spec.ts`, seems reasonable enough to have and should also get us coverage for all the lines :)

holgerd77:
Ah, hold on with merging for now, I guess we want to do the bugfix release before we are merging in something as big as this PR.

Maybe you guys want to directly do? I won't have time any more and could only do Monday. So this would need a VM + Common bugfix release, and you should also add a short note on the 1-2 other changes (I guess in VM?) till the minor releases in the CHANGELOG with the release PR.

acolytec3 (dismissed):
LGTM!

ryanio:
as Holger mentioned let's wait for #1691 release before merging this.
also gives me more time to explore the receiptTrie issue at block num 13323 :)

ryanio:
@acolytec3 i just gave it a try in Docker, worked well! removed the extra WORKDIR, thanks for that

Screen Shot 2022-02-04 at 5 23 00 PM

ryanio:
please note: when this PR is merged, please do NOT delete the branch for a few weeks, since the kintsugi Dockerfile still references the branch by name.

ryanio:
made it to tip of chain :)

Screen Shot 2022-02-09 at 12 32 16 AM

acolytec3 (approved):
I rebuilt everything and was able to sync the PoS chain past our previous blocking point so LGTM!  @holgerd77 what do you think about admin merging this one since @ryanio is getting ready to start the work on Kiln so we should be able to get the test coverage up in the two areas where it's lacking (`pendingBlock` and `engine`) with that PR work since these two areas will likely change a fair bit with the kiln work anyway.

acolytec3:
>  what do you think about admin merging this one 

Actually, looks like we don't need to.  I thought we had the code coverage requirement on to merge to master.

PR #1566, opened 2021-11-15
blockchain: Change toBuffer to toArrayLike for BNs

acolytec3:
Fixes #1552.  

This addresses an issue where our code currently uses the bn.js `toBuffer()` function in several places in the `Blockchain` library which then breaks in the browser context.  Changes this usage to `toArrayLike()` per the [bn.js docs](https://github.com/indutny/bn.js/#utilities) to fix compatibility issues with browserify and similar bundlers.

PR #1567, opened 2021-11-15
blockchain: v5.5.1 release

ryanio:
This patch release contains a bug fix for using the blockchain package in a browser context with tools like browserify, see PR [#1566](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1566).

### DIFF
--- a/packages/client/test/cli/cli-rpc.spec.ts
+++ b/packages/client/test/cli/cli-rpc.spec.ts
@@ -4,17 +4,17 @@ import { Client } from 'jayson/promise'
 
 const cliArgs = ['--rpc', '--ws', '--dev', '--transports=rlpx']
 
-const end = (child: ChildProcessWithoutNullStreams, hasEnded: boolean, t: tape.Test) => {
+const end = (child: ChildProcessWithoutNullStreams, hasEnded: boolean, st: tape.Test) => {
   if (hasEnded) return
   hasEnded = true
   child.stdout.removeAllListeners()
   child.stderr.removeAllListeners()
   child.kill('SIGINT')
-  t.end()
+  st.end()
 }
 
 tape('[CLI] rpc', (t) => {
-  t.test('should return valid responses from http and ws endpoints', (t) => {
+  t.test('should return valid responses from http and ws endpoints', (st) => {
     const file = require.resolve('../../dist/bin/cli.js')
     const child = spawn(process.execPath, [file, ...cliArgs])
     const hasEnded = false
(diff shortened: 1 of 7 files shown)


packages/client/test/cli/cli-rpc.spec.ts
```diff
@@ -4,17 +4,17 @@ import { Client } from 'jayson/promise'
...
 
-const end = (child: ChildProcessWithoutNullStreams, hasEnded: boolean, t: tape.Test) => {
+const end = (child: ChildProcessWithoutNullStreams, hasEnded: boolean, st: tape.Test) => {
   if (hasEnded) return
   hasEnded = true
   child.stdout.removeAllListeners()
   child.stderr.removeAllListeners()
   child.kill('SIGINT')
-  t.end()
+  st.end()
```
holgerd77, line 13: I am sure you know what you are doing here but just for my understanding: what/how is this changing things? Wouldn't `t` be a local variable and therefore take precedence over the outer `t`? 🤔

(I am especially interested since this seems to be very much a reocurring thing and it's likely important to grasp for future fixes)

holgerd77 (approved):
LGTM, will merge and do a release on this.

holgerd77:
Just published this as [v5.5.1](https://github.com/ethereumjs/ethereumjs-monorepo/releases/tag/%40ethereumjs%2Fblockchain%405.5.1) on npm.

packages/client/test/cli/cli-rpc.spec.ts
ryanio: I made the change in response to [this cli failure](https://github.com/ethereumjs/ethereumjs-monorepo/runs/4214652340?check_suite_focus=true) on the first commit when opening this PR:

```
# [CLI] rpc
# should return valid responses from http and ws endpoints
ok 1 read from WS RPC
# http and ws endpoints should not start when cli args omitted
not ok 2 .end() already called: read from HTTP RPC
```

I assumed there was some overlap on `t` being overwritten with each sub-test suite so I made the change to make the inner test `st`, I think it's a better practice to not overwrite this variable, especially when passing it around to functions to call `end` as we do in that file.

ryanio:
thanks!

packages/client/test/cli/cli-rpc.spec.ts
holgerd77: Ah, that makes sense, thanks!

PR #1568, opened 2021-11-15
Update ethereum-tests to v10.2

acolytec3:
- Updates Ethereum-Tests to v10.2


holgerd77:
You should really do this step-by-step (so in a first iteration only point the difficulty tests to our existing state of `ethereum-tests` without anything else), this will otherwise close to impossible to be achieved. There are so many changes in the `ethereum-tests` repo lately that will will very much need some dedicated focus on its own.

acolytec3:
> You should really do this step-by-step (so in a first iteration only point the difficulty tests to our existing state of `ethereum-tests` without anything else), this will otherwise close to impossible to be achieved. There are so many changes in the `ethereum-tests` repo lately that will will very much need some dedicated focus on its own.

I've moved the block difficulty tests section of it to a dedicated PR #1571 which is now ready for review.

packages/tx/src/legacyTransaction.ts
```diff
@@ -76,6 +76,14 @@ export default class Transaction extends BaseTransaction<Transaction> {
 
     const [nonce, gasPrice, gasLimit, to, value, data, v, r, s] = values
 
+    const integerProps = [nonce, gasPrice, gasLimit, value, v, r, s]
+    for (const x of integerProps) {
+      if (toBuffer(x).length > 0 && x[0] === 0x00) {
```
ryanio, line 81: these are already Buffers so they don't need to be routed through toBuffer again

packages/tx/src/baseTransaction.ts
```diff
@@ -91,8 +91,20 @@ export abstract class BaseTransaction<TransactionObject> {
     const rB = toBuffer(r === '' ? '0x' : r)
     const sB = toBuffer(s === '' ? '0x' : s)
 
-    this.nonce = new BN(toBuffer(nonce === '' ? '0x' : nonce))
-    this.gasLimit = new BN(toBuffer(gasLimit === '' ? '0x' : gasLimit))
+    const bufferedNonce = toBuffer(nonce === '' ? '0x' : nonce)
```
ryanio, line 94: if you want to use the same naming as above this can be `nonceB`

packages/tx/src/baseTransaction.ts
```diff
@@ -91,8 +91,20 @@ export abstract class BaseTransaction<TransactionObject> {
...
+    if (bufferedNonce.length > 8) {
+      // Implements EIP-2681 which limits nonce to 2^64-1
+      // https://eips.ethereum.org/EIPS/eip-2681
+      throw new Error('nonce cannot be more than 2^64-1')
+    }
+    this.nonce = new BN(bufferedNonce)
+
+    const bufferedGasLimit = toBuffer(nonce === '' ? '0x' : gasLimit)
+    if (bufferedGasLimit.length > 8) {
+      throw new Error('gasLimit can not be more than 2^64-1')
```
ryanio, line 104: is this specified somewhere?

packages/tx/src/legacyTransaction.ts
acolytec3: Hmm, in some of the API tests, this was failing with a type error related to x.length being undefined but this resolved it since it accounts for blank values.  Is this rather a matter for fixing another test you think?

packages/tx/src/eip2930Transaction.ts
```diff
@@ -188,6 +188,10 @@ export default class AccessListEIP2930Transaction extends BaseTransaction<Access
 
     this.gasPrice = new BN(toBuffer(gasPrice === '' ? '0x' : gasPrice))
 
+    if (this.gasPrice.mul(this.gasLimit).toArray().length > 32) {
+      throw new Error('gas limit * price overflow')
```
ryanio, line 192: is this specified somewhere?

packages/tx/src/baseTransaction.ts
acolytec3: I guess that should have been obvious to me.  Thanks!

packages/tx/src/legacyTransaction.ts
ryanio: hm, i don't think we should expect a non-Buffer value to be passed to fromValuesArray, it should be formatted beforehand to respect the input param type (`Buffer[]`). if `length` is undefined then a Buffer is probably not being passed in?

acolytec3: Okay, let me change it back and look at the test that was failing.

packages/tx/src/eip2930Transaction.ts
acolytec3: Here again, not specified anywhere but if the length of the array version of the BN is greater than 32, that would imply it's greater than the greatest possible integer that Ethereum handles, right?  [This test](https://github.com/ethereum/tests/blob/develop/TransactionTests/ttGasLimit/TransactionWithGasLimitxPriceOverflow.json) expects this to be an overflow error and was failing before this change.

packages/tx/src/baseTransaction.ts
acolytec3: Here's [the test](https://github.com/ethereum/tests/blob/develop/TransactionTests/ttGasLimit/TransactionWithGasLimitOverflow64.json) that fails if you don't put this change in.

packages/tx/src/legacyTransaction.ts
acolytec3: It's failing in the [base.spec.ts](https://github.com/ethereumjs/ethereumjs-monorepo/blob/eb5240bec432b5c37f38fd5483483f805f24f234/packages/tx/test/base.spec.ts#L128) because an unsigned transaction was being passed in.  I added a second check for v, r, and s only if they exist in the `txValues` passed in.

acolytec3:
@ryanio @holgerd77 @jochem-brouwer Okay, this one is ready for review.  All tests are passing.  Conceivably, we could do more validation with the new fields in the tests (e.g. `expectException` in the `post` section of any test that is expected to have an error associated with it) but that feels beyond the scope of this PR unless you tell me otherwise.

packages/tx/src/baseTransaction.ts
ryanio:
```suggestion
      throw new Error('gasLimit cannot be more than 2^64-1')
```
to match the above `nonceB` cannot

packages/vm/src/runTx.ts
```diff
@@ -300,6 +300,11 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
   let fromAccount = await state.getAccount(caller)
   const { nonce, balance } = fromAccount
 
+  // EIP-3607: Reject transactions from senders with deployed code
+  if (fromAccount.codeHash.toString('hex') !== KECCAK256_NULL_S) {
```
ryanio, line 304: last nit, can compare the buffers directly here
`if (!fromAccount.codeHash.equals(KECCAK256_NULL)) {`

packages/tx/src/legacyTransaction.ts
```diff
@@ -76,6 +77,20 @@ export default class Transaction extends BaseTransaction<Transaction> {
...
+      // If transaction is signed, add v, r, and s to values checked for leading zeroes
+      if (x?.length) integerProps.push(x)
+    }
+
+    for (const x of integerProps) {
+      if (x.length > 0 && x[0] === 0x00) {
+        // Throws if integer is encoded with leading zeroes
+        throw new Error('Integer values cannot have leading zeroes')
+      }
+    }
```
holgerd77, line 92: Some comments here: 😋

1. Again, it would be good if the reasoning is stated here shortly as a one line comment
2. Is there any reason that this is only being done in the legacy transaction?
3. Let's please differentiate the error message with the concrete parameter leading to the error message (even if the code gets slighly more complicated (maybe a small internal function or something).

packages/tx/src/legacyTransaction.ts
```diff
@@ -76,6 +77,20 @@ export default class Transaction extends BaseTransaction<Transaction> {
 
     const [nonce, gasPrice, gasLimit, to, value, data, v, r, s] = values
 
+    const integerProps = [nonce, gasPrice, gasLimit, value]
+
+    for (const x of [v, r, s]) {
+      // If transaction is signed, add v, r, and s to values checked for leading zeroes
+      if (x?.length) integerProps.push(x)
+    }
```
holgerd77, line 85: I guess I would feel better here if we rather do a `values.length === 9` check for a signed tx (as being done above) and *then* just add all values for a check. Feels more consistent to me than this one by one selection.

packages/tx/src/legacyTransaction.ts
holgerd77: This needs some test cases, minimally:
- Signed tx with focus on the v,r,s values (v is always minimally a good candidate)
- Same for unsigned tx
- Nonce being 0
- Nonce being non-0

packages/tx/src/legacyTransaction.ts
```diff
@@ -106,6 +121,9 @@ export default class Transaction extends BaseTransaction<Transaction> {
 
     this.gasPrice = new BN(toBuffer(txData.gasPrice === '' ? '0x' : txData.gasPrice))
 
+    if (this.gasPrice.mul(this.gasLimit).gt(MAX_INTEGER)) {
+      throw new Error('gas limit * price overflow')
+    }
```
holgerd77, line 126: Just to note that we are doing an extra multiplication with this just for the check which will cost some performance. Is this fixing a tx test or rather some VM test? If it is a VM test it might be better to do on the location where this multiplication is done anyhow? 🤔

packages/vm/src/runTx.ts
```diff
@@ -300,6 +300,11 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
   let fromAccount = await state.getAccount(caller)
   const { nonce, balance } = fromAccount
 
+  // EIP-3607: Reject transactions from senders with deployed code
```
holgerd77, line 303: Also got some confirmation on this from the EIP:

> This EIP can be implemented as a soft fork because the new validity rules are a strict superset of the previous validity rules.

packages/vm/tests/tester/runners/GeneralStateTestsRunner.ts
```diff
@@ -116,20 +117,10 @@ async function runTestCase(options: any, testData: any, t: tape.Test) {
...
       } catch (e: any) {
-        // If tx is invalid and coinbase is empty, the test harness
-        // expects the coinbase account to be deleted from state.
-        // Without this ecmul_0-3_5616_28000_96 would fail.
-        const account = await vm.stateManager.getAccount(block.header.coinbase)
-        if (account.balance.isZero()) {
-          await vm.stateManager.putAccount(block.header.coinbase, new Account())
-          await vm.stateManager.cleanupTouchedAccounts()
-          await vm.stateManager._cache.flush()
-        }
```
holgerd77, line 132: Why is this not being cleared by a `revert()` in `VM.runTx()` when the tx fails? Should be the case I guess? 🤔

holgerd77 (changes requested):
Nice, some various comments. This is a really delicate PR doing substantial Tx library changes where it is extremely easy to have side effects (role of 0-values not properly evaluated e.g.), so we should handle with some extra care.

holgerd77:
Ah, and waht actually puzzles me on this PR: I would have expected the test update to be more problematic, since on release [v10.1](https://github.com/ethereum/tests/releases/tag/v10.1) the legacy tests moved into an own repository and are now linked by submodule.

Did this absolutely play no role here and everything still just continue to work out of the box? 🤔

acolytec3:
> Ah, and waht actually puzzles me on this PR: I would have expected the test update to be more problematic, since on release [v10.1](https://github.com/ethereum/tests/releases/tag/v10.1) the legacy tests moved into an own repository and are now linked by submodule.
> 
> Did this absolutely play no role here and everything still just continue to work out of the box? 🤔

It seems to be the case but I have seen weirdness before when working with the tests repo. Once I'm up and running a little later this morning, I'll do a fresh install of the repo and make sure everything is on the correct commit and make sure the tests still pass.

packages/tx/src/baseTransaction.ts
```diff
@@ -91,8 +91,20 @@ export abstract class BaseTransaction<TransactionObject> {
...
 
-    this.nonce = new BN(toBuffer(nonce === '' ? '0x' : nonce))
-    this.gasLimit = new BN(toBuffer(gasLimit === '' ? '0x' : gasLimit))
+    const nonceB = toBuffer(nonce === '' ? '0x' : nonce)
+    if (nonceB.length > 8) {
+      // Implements EIP-2681 which limits nonce to 2^64-1
+      // https://eips.ethereum.org/EIPS/eip-2681
+      throw new Error('nonce cannot be more than 2^64-1')
+    }
+    this.nonce = new BN(nonceB)
```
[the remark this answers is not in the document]
acolytec3: I'm not 100% sure what you mean by routing through Common but Geth [has implemented this](https://github.com/ethereum/go-ethereum/commit/f32feeb260054d7c90c1efd57dfec6b717717e18).

packages/tx/src/baseTransaction.ts
```diff
@@ -91,8 +91,20 @@ export abstract class BaseTransaction<TransactionObject> {
...
+
+    const gasLimitB = toBuffer(nonce === '' ? '0x' : gasLimit)
+    if (gasLimitB.length > 8) {
+      throw new Error('gasLimit cannot be more than 2^64-1')
+    }
+    this.gasLimit = new BN(gasLimitB)
```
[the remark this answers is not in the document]
acolytec3: This is all Dmitry could tell me:
>I think since the block gas limit is capped by eip so the transaction gas limit

[These tests](https://github.com/ethereum/tests/blob/develop/TransactionTests/ttGasLimit/TransactionWithGasLimitOverflow64.json) fail if this limit is not in place.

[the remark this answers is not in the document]
acolytec3: That said, Geth limits gasLimit to a [Uint64](https://github.com/ethereum/go-ethereum/blob/33f28138095faada5cbf027849993808a9dba785/core/genesis.go#L54) so this check would apply the same since we don't have an exact Uint64 in Javascript.

packages/tx/src/eip2930Transaction.ts
```diff
@@ -188,6 +189,10 @@ export default class AccessListEIP2930Transaction extends BaseTransaction<Access
 
     this.gasPrice = new BN(toBuffer(gasPrice === '' ? '0x' : gasPrice))
 
+    if (this.gasPrice.mul(this.gasLimit).gt(MAX_INTEGER)) {
+      throw new Error('gas limit * price overflow')
+    }
```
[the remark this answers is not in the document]
acolytec3: [This test covers this scenario in the ethereum-tests repo])https://github.com/ethereum/tests/blob/develop/TransactionTests/ttGasLimit/TransactionWithGasLimitxPriceOverflow.json).  Do you want us to have one in our own test suite as well?

packages/tx/src/baseTransaction.ts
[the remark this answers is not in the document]
holgerd77: "Routing through Common" would mean that we would add this as a dedicated EIP (in Common) and do a switch like we always do on the other EIPs (`if (Common.isActivatedEIP(2681) { // do this check }`).

packages/tx/src/baseTransaction.ts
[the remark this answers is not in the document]
holgerd77: Ok, thanks, maybe just add "Geth limits gasLimit to a Uint64" as a one-line comment above here.

packages/tx/src/eip2930Transaction.ts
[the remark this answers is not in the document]
holgerd77: Ah ok, not sure, maybe that's ok then.

packages/tx/src/legacyTransaction.ts
acolytec3: > Some comments here: yum
> 
> 1. Again, it would be good if the reasoning is stated here shortly as a one line comment

Okay, will add.  It comes down to the [RLP rules](https://eth.wiki/en/fundamentals/rlp) that state that deserialized positive integers with leading zeroes are invalid.

> 2. Is there any reason that this is only being done in the legacy transaction?
User error.  I'll move the check for all the numeric fields in the base transaction to `baseTransaction` and then check transaction type specific fields (e.g. `gasPrice` for legacy transactions, `baseFeePerGas` for EIP1559) to their specific transaction constructors.

> 3. Let's please differentiate the error message with the concrete parameter leading to the error message (even if the code gets slighly more complicated (maybe a small internal function or something).

Will do.

packages/tx/src/legacyTransaction.ts
acolytec3: This was causing [this test](https://github.com/ethereum/tests/blob/develop/TransactionTests/ttGasLimit/TransactionWithGasLimitxPriceOverflow.json) to fail.

packages/vm/tests/tester/runners/GeneralStateTestsRunner.ts
acolytec3: It is now.  @ryanio  and I looked at this and this looks to be some workaround for presumably [this test[(https://github.com/ethereum/tests/blob/33c3174e88fa2315a316602007393ee5180bf851/src/GeneralStateTestsFiller/stZeroKnowledge2/ecmul_0-3_5616_28000_96Filler.json) based on the comments but the code itself is quite old and it was mutating the state after the transaction failed and was causing [this new set of tests for EIP-3607](https://github.com/ethereum/tests/tree/develop/GeneralStateTests/stEIP3607) to fail.  After removing it, I've re-run the tests the workaround was for and they are passing so I think this just became duplicative code that wasn't failing any tests.

packages/tx/src/legacyTransaction.ts
acolytec3: > This needs some test cases, minimally:
> 
> * Signed tx with focus on the v,r,s values (v is always minimally a good candidate)

Should be covered by these tests from the Ethereum-Tests repo:
ValidChainID1InvalidV00
TransactionWithSvaluePrefixed00BigInt
TransactionWithRvaluePrefixed00BigInt

> * Same for unsigned tx

Not sure about this one.  Are you just looking for a test where there is no v, r, or s?  I can write one if you don't feel like the existing test suite covers it though I removed the specific logic that only checked an unsigned transaction and replaced with a check that iterates over all of the transaction fields that could be integers (regardless of signed/unsigned status) and checks them for leading zeroes.  Given this reworking of the logic, it seems like the existing tests in the Ethereum/Tests transaction test suite should cover this since they already have scenarios with leading zeroes for all of the integer fields.

> * Nonce being 0

Maybe covered by [this one](https://github.com/ethereum/tests/blob/develop/TransactionTests/ttNonce/TransactionWithEmptyBigInt.json)
> * Nonce being non-0

Should be covered by [this test](https://github.com/ethereum/tests/blob/develop/TransactionTests/ttNonce/TransactionWithLeadingZerosNonce.json)

packages/tx/src/eip2930Transaction.ts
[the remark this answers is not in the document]
ryanio: i was also thinking about this yesterday, in general i think it is always nice to add coverage in our own test suite with any new lines added to source so we can have more local coverage, otherwise our overall repo coverage will continue to decline over time.

we could also consider adding the ethereum-tests transactionRunner to the coverage step if that doesn't add too much time and does also improve the coverage with these more isolated errors/validation.

packages/vm/tests/tester/runners/GeneralStateTestsRunner.ts
holgerd77: Nice 👍

packages/tx/src/baseTransaction.ts
```diff
@@ -90,23 +91,41 @@ export abstract class BaseTransaction<TransactionObject> {
...
+    const gasLimitB = toBuffer(gasLimit === '' ? '0x' : gasLimit)
+    const valueB = toBuffer(value === '' ? '0x' : value)
+
+    const integerProps = {
+      nonce: nonceB,
+      gasLimit: gasLimitB,
+      value: valueB,
+      v: vB,
+      r: rB,
+      s: sB,
```
holgerd77, line 104: I guess this is not working here in the constructor, since this "no leading zeros" rule really only applies for Txs which have been RLP encoded before. This is only be the case for Txs initialized with the `fromSerializedTx()` (and, alias: `fromRLPSerializedTx()`) and then the `fromValuesArray()` constructors (for the latter one I am only 95% sure that this is the case, might be good if someone confirms. If so we should take this function as a point for adding the checks since all others call into `fromValuesArray()`.

Let's take the situation here and pick out e.g. the `nonce`. As a final value nonce can very well be `0` (actually that's the base/first possible value), same for `v` values for non-legacy txs.

The `nonce` parameter is of type `BNLike` and can be e.g. a `number` and a `Buffer`.

An input value of `nonce = 0` is valid here, but also an empty Buffer. Both translate to a BN value of `0` at the end, see the following for illustration:

```shell
> import { BN, toBuffer } from 'ethereumjs-util'
{}
> let nonceN = 0
undefined
> let nonceB = Buffer.from([])
undefined
> toBuffer(nonceN)
<Buffer 00>
> new BN(toBuffer(nonceN))
<BN: 0>
> toBuffer(nonceB)
<Buffer >
> new BN(toBuffer(nonceN))
<BN: 0>
```

So these different `nonce` inputs lead to different intermediary Buffer values (`<Buffer 00>` and `<Buffer >`). If we are doing the checks from above here, we are unrightfully disallowing the case with the empty Buffer.

So we need to really dedicatedly go into the different `fromValuesArray()` constructors. Maybe we can add a helper function in `util.ts` on this to not do too much code duplication.

packages/tx/src/eip1559Transaction.ts
```diff
@@ -197,11 +197,24 @@ export default class FeeMarketEIP1559Transaction extends BaseTransaction<FeeMark
...
 
+    if (maxFeePerGasB.length > 0 && maxFeePerGasB[0] === 0x00) {
+      // RLP encoded integer values with leading zeroes are invalid
+      throw new Error('maxFeePerGas cannot have leading zeroes')
+    }
+
+    if (maxPriorityFeePerGasB.length > 0 && maxPriorityFeePerGasB[0] === 0x00) {
+      // RLP encoded integer values with leading zeroes are invalid
+      throw new Error('maxPriorityFeePerGas cannot have leading zeroes')
+    }
```
holgerd77, line 213: Would be good if good if these checks on the extra fields can only go into this helper function in `util`, maybe this can be called with something like this.

```javscript
util.leadingZeroCheck([
  [nonceB, 'nonce' ], // Second string is for the error message
  [ maxFeePerGasB, 'maxFeePerGas' ],
])
```

Something like this.

packages/tx/src/legacyTransaction.ts
holgerd77: Thought about this a bit longer.

Yes, I guess we also need these tests ourselves. The thing with the external tests is: these kind of tests are very much a black box. One has no real oversight respectively is loosing track what is actually being tested.

But more importantly: these tests at the end only test the general functionality requirements on a tx, for this they are good and useful. They are *not* testing our API though. This is especially crucial for input values. We have so many static constructors. The tx tests only test one of them. If we forget adding a certain type of functionality to one of the constructors not being called, this would already not be caught. Same goes for different input type (`number`, `BN`, `Buffer`,...) e.g.. So we need these additional local tests for this other-perspective testing. Especially on the tx library which is so sensitive.

packages/vm/src/runTx.ts
```diff
@@ -301,7 +301,7 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
   const { nonce, balance } = fromAccount
 
   // EIP-3607: Reject transactions from senders with deployed code
-  if (fromAccount.codeHash !== KECCAK256_NULL) {
+  if (!fromAccount.codeHash.equals(KECCAK256_NULL)) {
```
holgerd77, line 304: Ah, have done a commit-by-commit review, just seeing the update, so forget about the first comment. 😄

holgerd77 (changes requested):
Thanks for the updates 🙂, some change requests.

packages/tx/src/legacyTransaction.ts
acolytec3: Makese sense.  I'm reworking how I check for leading zeroes and will add appropriate API tests along the way.

packages/tx/src/baseTransaction.ts
acolytec3: Indeed, I wish I had read this last night :grimacing: I independently came to this same realization this morning after much exploring rabbit trails.

acolytec3:
@ryanio @holgerd77 
I've completely reworked the leading zeroes checks along the lines Holger recommended, added a new util function for checking leading zeroes, and then added some tests to the `tx` package to validate the leading zero checks in all three transaction types.  Provided the CI tests pass okay, I think this might be ready for review again

packages/tx/src/baseTransaction.ts
```diff
@@ -90,11 +91,26 @@ export abstract class BaseTransaction<TransactionObject> {
...
+      throw new Error('gasLimit cannot be more than 2^64-1')
+    }
+
+    this.nonce = new BN(nonceB)
+
+    if (this.nonce.gte(MAX_UINT64)) {
+      // Implements EIP-2681 which limits nonce to 2^64-1
+      // https://eips.ethereum.org/EIPS/eip-2681
+      throw new Error('nonce cannot be more than 2^64-1')
+    }
```
holgerd77, line 109: Sorry for being a bit picky here - just live-thinking myself. I am just wondering: since the `BN.js` library is somewhat on its last iterations and we will at some point replace with `BigInt` *and* we have this other check possibility already in here (which also makes the code a bit inconsistent):

Maybe (or - getting a a stronger opinion while writing - likely 😋) we just want to skip this `MAX_UINT64` and also remove from `Util` and not introduce a new `BN.js` related constant for this one single first use case and rather do a byte length check here as well as being done for `gasLimitB`.

packages/tx/src/eip1559Transaction.ts
```diff
@@ -152,6 +154,23 @@ export default class FeeMarketEIP1559Transaction extends BaseTransaction<FeeMark
...
+      v: v,
+      r: r,
+      s: s,
+    }
+
+    const res = hasLeadingZeroes(integerValues)
+
+    if (res) {
+      throw new Error(`${res} cannot have leading zeroes`)
+    }
```
holgerd77, line 172: We could also throw directly in `hasLeadingZeroes()` directly I guess and safe a bit of code or has this side effects? Totally no strong opinion though.

packages/util/src/types.ts
```diff
@@ -141,3 +141,10 @@ export function toType<T extends TypeOutput>(
     return `0x${output.toString('hex')}` as TypeOutputReturnType[T]
   }
 }
+
+/**
+ * An object containing pairs of keys and the RLP encoded value for that key
+ */
+export type RlpValues = {
+  [value: string]: Buffer | undefined
+}
```
holgerd77, line 150: Just as some first read comment: this is a *very*  central place for such a new type which will become part of the overall API/functionality set if we introduce here. Not against it or anything and also haven't made up my mind, @ryanio what do you think (others as well of course), introduce here or not?

Alternative would be to do a bit more locally e.g. also in the `util.ts` file from the tx library.

packages/tx/src/eip2930Transaction.ts
```diff
@@ -186,7 +205,17 @@ export default class AccessListEIP2930Transaction extends BaseTransaction<Access
     // Verify the access list format.
     AccessLists.verifyAccessList(this.accessList)
 
-    this.gasPrice = new BN(toBuffer(gasPrice === '' ? '0x' : gasPrice))
+    const gasPriceB = toBuffer(gasPrice === '' ? '0x' : gasPrice)
+    if (gasPriceB.length > 1 && gasPriceB[0] === 0x00) {
+      // RLP encoded integer values with leading zeroes are invalid
+      throw new Error('gasPrice cannot have leading zeroes')
```
holgerd77, line 211: I am a bit confused, why this single extra check for this one parameter here?

packages/util/src/bytes.ts
```diff
@@ -276,3 +276,15 @@ export const baToJSON = function (ba: any): any {
...
+ *
+ * @param values An object containing a set keys and RLP encoded values
+ * @returns a string with the value that was encoded with leading zeroes or undefined if no values have leading zeroes
+ */
+export const hasLeadingZeroes = function (values: RlpValues): string | undefined {
+  const res = Object.entries(values).find(
+    (entry) => entry[1] && entry[1]?.length > 0 && entry[1][0] === 0x00
+  )
+  return res ? res[0] : undefined
+}
```
holgerd77, line 290: Also here, let's please give this a round of discussion if we want expose or not (or rather hang this lower for now and place in `util.ts` tx.

This question is also connected with my following concern: I really wonder if we want to do this check in the Transaction library for the `0x00` value, for example for the nonce. While technically correct I wonder how much hazzle we will produce with this on consuming libraries. It's rather likely that people do something like this in their code bases:

```typescript
import { toBuffer } from 'ethereumjs-util'
const nonce = (SOME_INPUT_NUMBER) // e.g. 0, 1, 2,...

const tx = Transaction.fromValuesArray([ ..., toBuffer(nonce)) ]

```

So this *will* lead to this `0x00` input *even* with our own `toBuffer()` utility method. And at the end this forms a totally valid respectively *the same*  Transaction object as if initialized with `0x`.

So I wonder if this rather has the nature of some "bug fix" or are we rather (eventually unnecessarily) restricting a somewhat more loose user input with more harm than good? 🤔

To circle back, @acolytec3 some base question again: are these new tests actually also triggering on this specific `0x00` case or are they just targeting something like `0x0003`?

So my current tendency is to not go so strict here, but also unsure. Will also circle in the Hardhat guys here 😋 @fvictorio @alcuadrado since they often have a very practical view on these kind of things.

holgerd77 (changes requested):
Hi Andrew, another review round, don't feel irritated by these lots of discussions, this is wonderful work you are doing and all the updates are great, these things in particular just touch a lot of things by nature and it's just necessary to address some things here before we can continue.

packages/tx/src/eip2930Transaction.ts
acolytec3: I think I just forgot to remove it when cleaning up after my last round of changes.  Will remove in the next commit.

packages/util/src/bytes.ts
acolytec3: >To circle back, @acolytec3 some base question again: are these new tests actually also triggering on this specific 0x00 case or are they just targeting something like 0x0003?

[This test](is an example of one where the nonce passed in is just 0x00 (at least that's what it shows up as when decoded with the `rlp` library.  The set of decoded bytes looks like this:
```
[
  <Buffer 00>,
  <Buffer 01>,
  <Buffer 52 08>,
  <Buffer 09 5e 7b ae a6 a6 c7 c4 c2 df eb 97 7e fa c3 26 af 55 2d 87>,
  <Buffer 0b>,
  <Buffer >,
  <Buffer 1c>,
  <Buffer 48 b5 5b fa 91 5a c7 95 c4 31 97 8d 8a 6a 99 2b 62 8d 55 7d a5 ff 75 9b 30 7d 49 5a 36 64 93 53>,
  <Buffer 10 00 2c ef 53 8b c0 c8 e2 1c 46 08 06 34 a9 3f 4d 75 2b c9 fe 4b 54 6b 60 ac 05 5e 84 2d 34 2b>
]
```

holgerd77: Ok, thanks, well, that's the 0 case. 🙂 

Nevertheless let's keep this open for some more time (maybe we need to halt here for 1-2 days).

Side reference, did the following assurance tests on this:

```shell
> rlp.encode('0')
<Buffer 30>
> rlp.encode(Buffer.from('', 'hex'))
<Buffer 80>
> rlp.encode(Buffer.from('00', 'hex'))
<Buffer 00>
> rlp.decode(rlp.encode(Buffer.from('', 'hex')))
<Buffer >
> rlp.decode(rlp.encode(Buffer.from('00', 'hex')))
<Buffer 00>
```

packages/tx/src/eip2930Transaction.ts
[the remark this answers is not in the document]
acolytec3: Okay, I've added some tests to cover the scenarios Holger outlined so will see where we land once we get final feedback on the new leadingZeroes function.

packages/tx/src/baseTransaction.ts
acolytec3: Okay, I'm good with it either way.  I like the idea of just checking the bytes myself but there was concern in earlier reviews about converting BNs to buffers causing additional performance hits (right @ryanio?) or something to that effect.

packages/tx/src/eip1559Transaction.ts
acolytec3: Hmm, I guess that's a fair point and I hadn't thought about it.  Either is fine with me since the main goal was to check all fields with potentially leading zeroes in one place and then just throw an error with a specific indication of which field had the leading zeroes in it.

packages/util/src/types.ts
acolytec3: Now that you mention it, it probably does make sense to keep it inside `tx`.  I can't envision a case where a helper like this is going to be used outside of the specific use case we're looking at.  I'll move it.

packages/tx/src/eip1559Transaction.ts
ryanio: Yeah that's a good idea I say throw in the function, you can highlight exactly what threw there too.

packages/tx/src/baseTransaction.ts
ryanio: Yeah sure just check the nonceB like the others to match 👍

packages/tx/src/eip1559Transaction.ts
ryanio: I would also suggest moving this to ethereumjs-util under I think bytes.ts would be good and you can name it something like validateNoLeadingZeroes

packages/util/src/bytes.ts
fvictorio: @holgerd77 Hmm, I'm not sure I understand this. Can you elaborate on what are the options here and what each one means for consumers?

packages/tx/src/baseTransaction.ts
acolytec3: So...doing the bytelength comparison actually doesn't work because we have to check not only 2^64 -1 but 2^64-2 per EIP 2681 - "The reason a transaction with nonce 2^64-1 is invalid, because otherwise after inclusion the sender account’s nonce would exceed 2^64-1."  So, I'm going to have to use the new BN either way.

packages/util/src/bytes.ts
```diff
@@ -276,3 +276,17 @@ export const baToJSON = function (ba: any): any {
     return array
   }
 }
+
+/**
+ *
+ * @param values An object containing a set keys and RLP encoded values
+ * @returns a string with the value that was encoded with leading zeroes or undefined if no values have leading zeroes
```
holgerd77, line 283: This return comment is still wrong.

packages/tx/src/baseTransaction.ts
holgerd77: Ok, then just keep the new BN constant in Util, just realized that there are already two BN constants in so we have to deal with this at some point anyhow and really makes no difference if there are 2 or 3 I guess.

packages/tx/src/eip1559Transaction.ts
holgerd77: Ok, as mentioned in the chat, I am ok with that change as well.

packages/util/src/bytes.ts
holgerd77: @fvictorio Thanks for chiming in. 🙂 

The options are:

1. The `Tx.fromValuesArray()` function takes as 0-value in the form of a Buffer - for all values which represent integers like e.g. the nonce - both an empty and a zero-Buffer (so both `<Buffer >` and `<Buffer 00>`, as it is now). Theoretically (or: by RLP specification) only the emtpy Buffer would be strictly valid here.
2. We narrow the input here to only the empty Buffer `<Buffer >`

And the question here would be: would 2. be problematic for you guys?

packages/block/test/testdata/testdata.json
```diff
@@ -25,7 +25,7 @@
           "data": "0x",
           "gasLimit": "0x55f0",
           "gasPrice": "0x0a",
-          "nonce": "0x00",
+          "nonce": "0x",
```
ryanio, line 28: would it be better to use `0x0` to represent the number zero? i would think `0x` represents null. maybe update a few of them so we have both cases covered?

packages/tx/src/eip2930Transaction.ts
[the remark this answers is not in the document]
ryanio: i'm still seeing a check warning "Added line #L207 was not covered by tests"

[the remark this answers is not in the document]
ryanio: ah looks like the one below in baseTransaction is covered but this eip2930Transaction isn't, that's probably fine

packages/util/src/types.ts
ryanio: have inlined it into the function method in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1568/commits/be2bb7c68f647dfe9964ad6eb547d502441d0aa4 so the type is no longer needed

packages/util/test/bytes.spec.ts
```diff
@@ -351,3 +352,38 @@ tape('intToHex', function (st) {
...
+
+tape('validateNoLeadingZeroes', function (st) {
+  const noLeadingZeroes = {
+    a: toBuffer('0x123'),
+    b: undefined,
+    c: toBuffer('0x'),
+  }
+  const leadingZeroes = {
+    a: toBuffer('0x001'),
+    b: toBuffer('0x'),
```
holgerd77, line 364: The `b` case is wrong here I guess?

holgerd77:
Really happy with this progress here on this and how this evolve, cool! Have not done a broader review yet but just skimmed over, but this looks really nice and clean and thought-through now! 😄

packages/util/src/bytes.ts
fvictorio: Oh, I see. Indeed this is one of those "breaking change of bug fix?" scenarios.

I _think_ we are not doing something like your `[..., toBuffer(nonce)]` example anywhere. The places where we create txs use data that comes from deserializing already valid RLP encoded txs, so this doesn't seem like a problem. And I'm fairly confident that if this causes an issue and you run our test suite with this change, it will be caught.

Also, I personally prefer option 2 if it's more compliant.

holgerd77: Ok, thanks' for having a look and the additional dedicated opinion expressed here, that's really valuable. 😄

packages/util/test/bytes.spec.ts
acolytec3: I don't think so.  the `toBuffer` method [strips the 0x from hex strings](https://github.com/ethereumjs/ethereumjs-monorepo/blob/fbd2a320dbeed872420272e688b871fd359e1498/packages/util/src/bytes.ts#L167)) which leaves a null buffer.  Just to verify, I commented out the `a` value in the test data and wrote the value of `b` to the console and it produced 

```
b <Buffer >
not ok 86 should throw
```

I believe this is correct since the `toBufer` considers '0x' to be an empty byte array.

holgerd77: Ah, had a closer look, I guess I have a problem with your test setup here.

So it's just unfortunate to throw this all in one dict (and - to add some confusion - the `leadingZeros` -> `b` naming is also not correct). With this setup it's not possible to test the single values respectively see which one goes wrong. Can you please improve here and do separate dicts for each value?

acolytec3:
Pending any new failing tests, this one should be ready for any final review and merging.  I'll open a new issue for the half of EIP-2861 we didn't complete here since it will need to be addressed whenever we next update the `ethereum/tests`.

packages/tx/src/baseTransaction.ts
```diff
@@ -377,21 +379,55 @@ export abstract class BaseTransaction<TransactionObject> {
...
+            }
+          }
+          break
+        case 256:
+          if (cannotEqual) {
+            if (value?.gte(MAX_INTEGER)) {
+              const msg = this._errorMsg(
+                `${key} cannot equal or exceed MAX_INTEGER (2^256-1), given ${value}`
+              )
+              throw new Error(msg)
```
jochem-brouwer, line 416: As mentioned by codecov the 256 case values are not tested. If you create a test for a fee market transaction then that would be resolved.

packages/tx/src/eip1559Transaction.ts
```diff
@@ -152,6 +153,8 @@ export default class FeeMarketEIP1559Transaction extends BaseTransaction<FeeMark
       s,
     ] = values
 
+    validateNoLeadingZeroes({ nonce, maxPriorityFeePerGas, maxFeePerGas, gasLimit, value, v, r, s })
```
jochem-brouwer, line 156: Hmm interesting great catch! Was this found by the new  ethereum/tests?

packages/tx/src/eip1559Transaction.ts
```diff
@@ -202,13 +205,10 @@ export default class FeeMarketEIP1559Transaction extends BaseTransaction<FeeMark
...
-      {
-        maxFeePerGas: this.maxFeePerGas,
-        maxPriorityFeePerGas: this.maxPriorityFeePerGas,
-      },
-      256
-    )
+    this._validateCannotExceedMaxInteger({
+      maxFeePerGas: this.maxFeePerGas,
+      maxPriorityFeePerGas: this.maxPriorityFeePerGas,
+    })
```
jochem-brouwer, line 211: Should the fee market tx also not have a check to ensure that `gasLimit * maxFeePerGas` does not overflow the `MAX_UINT`?

packages/tx/src/eip1559Transaction.ts
holgerd77: Yes.

packages/tx/src/baseTransaction.ts
acolytec3: I'm not aware of any edge cases where a value would need to be less than MAX_INTEGER (since as I understand it, this is always testing for integer overflow and not obscure EIP edge cases) so I don't think this can ever actually happen in reality but I'll add a test case just to be sure.

packages/tx/src/eip1559Transaction.ts
acolytec3: Good catch.  I'll add that here.

acolytec3: So on this one, if I add a check in the `_validateCannotExceedMaxInteger` function call in the 1559 tx constructor, [this test](https://github.com/ethereumjs/ethereumjs-monorepo/blob/531e4b3a1843833b3015ab185fc610f4f2da5d67/packages/tx/test/eip1559.spec.ts#L169) now fails.  From looking at it, I think this test is actually invalid since maxFeePerGas of `TWO_POW256.subn(1)` (aka `MAX_INTEGER`) x gasLimit of 100 would produce a value greater than `MAX_INTEGER`, right?

acolytec3: That's what the test is indicating so I've made the associated changes and pushed but just want to confirm I'm not imaging things.

ryanio: I think using the function with "gasLimitXmaxFeePerGas" is a little confusing, can you just do the math yourself in a one liner like the legacy gasPrice * gasLimit?

acolytec3:
@holgerd77 @ryanio @jochem-brouwer Any additional feedback on this?  I believe we've addressed all the existing questions and I've got the test coverage up to a good level. Also, I think we're agreed that the `CREATE`/`CREATE2` stuff from EIP2681 will be resolved in a separate PR so would love to get this merged if we're good with everything.

jochem-brouwer (approved):
Looks great!

ryanio (approved):
lgtm!

holgerd77:
Great that this is merged now, great work @acolytec3! 😄 👍 

Short improvement suggestion for merging these kind of PRs: it would be good if we get accustomed to always do a short reflection-in-between-step on the sqash-commit-title (and I guess it also makes sense to adopt the PR title as well for reference reasons).

So now this has found its way into `master` as **"Update ethereum-tests to v10.2"**, this just very much hides all the additional work that has been gone into this PR and the substantial changes which are encompassed.

So just to re-bring this into conscious again 😋: these one liner commit messages are really important now to keep a transparent and accessible history on what has being done.

holgerd77:
Update: ah, and actually, on-top suggestion: from a practical PoV I've found that it makes sense respectively is more comfortable to *first* think about the PR title and eventually adopt, then one does not need to think about that part any more on the PR merge itself.

packages/util/test/bytes.spec.ts
```diff
@@ -351,3 +352,46 @@ tape('intToHex', function (st) {
...
+  }
+  const leadingZeroes = {
+    a: toBuffer('0x001'),
+  }
+  const onlyZeroes = {
+    a: toBuffer('0x0'),
+  }
+  const emptyBuffer = {
+    a: toBuffer('0x'),
+  }
```
holgerd77, line 368: Sorry, I am getting unsure here again, so just for a final confirmation so that this is not slipping through:

What is with a case like `0x01` (so the full byte just filled)? Manually tested on the current implementation of `validateNoLeadingZeroes` and this *passes* (since `b[0] === 1`).

Is this confirmed to be the correct behavior?

This would actually also be a good test case, just if someone has the occasion to add.

packages/tx/src/legacyTransaction.ts
```diff
@@ -106,6 +110,9 @@ export default class Transaction extends BaseTransaction<Transaction> {
 
     this.gasPrice = new BN(toBuffer(txData.gasPrice === '' ? '0x' : txData.gasPrice))
 
+    if (this.gasPrice.mul(this.gasLimit).gt(MAX_INTEGER)) {
+      throw new Error('gas limit * price overflow')
```
holgerd77, line 114: Ah, here is still the "old" un-appended + non expressive error msg.

@acolytec3 maybe a short clean up PR together with the small additional test case (and eventually 1-2 other things I might additionally find) might be good.

packages/util/src/bytes.ts
```diff
@@ -276,3 +276,15 @@ export const baToJSON = function (ba: any): any {
     return array
   }
 }
+
+/**
+ * Checks provided Buffers for leading zeroes and throws if found.
```
holgerd77, line 281: It would be good if we add some example values here in the code docs, something like:

Valid: `0x1`, `0x`, `0x01` (again, see question on on the `0x1`, `0x01` case)
Invalid: `0x0`, `0x00`, `0x0001`

(not 100% sure I got all, also there might be more expressive example values out there 🙂)

packages/vm/src/runTx.ts
```diff
@@ -300,6 +300,11 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
   let fromAccount = await state.getAccount(caller)
   const { nonce, balance } = fromAccount
 
+  // EIP-3607: Reject transactions from senders with deployed code
+  if (!fromAccount.codeHash.equals(KECCAK256_NULL)) {
+    const msg = _errorMsg('invalid sender address, address is not EOA', this, block, tx)
```
holgerd77, line 305: Maybe you can directly reference the EIP here to give people a chance to read up on this who are stumbling upon the error, so 'invalid sender address, address is not EOA (EIP-3607)'?

Otherwise this will get extremely hard to interpret.

holgerd77:
Ok, also had a final look on this, some nits and one dedicated question on cases like `0x1`, `0x01` for tx integers (both accepted right now). Otherwise looks good, really great work @acolytec3!! 🥳

packages/tx/src/legacyTransaction.ts
acolytec3: Will do!

PR #1570, opened 2021-11-16
VM: Performance Optimizations

holgerd77:
This PR does some performance optimizations on the VM:

- BNADD/BNMUL precompile input size restrictions before calling WASM module

TODO

### DIFF
--- a/packages/vm/src/evm/precompiles/07-ecmul.ts
+++ b/packages/vm/src/evm/precompiles/07-ecmul.ts
@@ -7,7 +7,7 @@ const bn128 = require('rustbn.js')
 export default function (opts: PrecompileInput): ExecResult {
   assert(opts.data)
 
-  const inputData = opts.data
+  const inputData = opts.data.slice(0, 128)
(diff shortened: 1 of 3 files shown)


packages/vm/src/evm/precompiles/07-ecmul.ts
```diff
@@ -7,7 +7,7 @@ const bn128 = require('rustbn.js')
 export default function (opts: PrecompileInput): ExecResult {
   assert(opts.data)
 
-  const inputData = opts.data
```
jochem-brouwer, line 10: Does the rustbn package throw if we input a buffer which is not 128 bytes?

holgerd77: No, it does not.

The thing here is that only the first 128 bytes are used anyhow for computation. So this is a plain performance optimization. The size of the input value here is not restricted so this gets unnecessarily slow with some larger input data.

holgerd77:
Closing in favor of #2568 (replacement PR)

PR #1572, opened 2021-11-16
VM: Cache Improvements

holgerd77:
This PR does some improvements on the VM StateManager cache.

On this first iteration it only clears the caches and updates the state root in `StateManager.setStateRoot()` if the new state root actually differs from the old one.

This targets the situation where we are calling the `setStateRoot()` function from within the client before running a new block when - in the normal occurence of things - the state root should already have been updated in the trie by the state updates which happened during tx execution and the resulting state root should therefore match the block state root anyhow (the alternative would have been to remove the `setStateRoot()` call in the client but I guess this is the more robust version where more use cases can benefit).

I tested this in the client with the following command `cd ../vm && npm run build:node && cd ../client/ && npm run client:start --executeBlocks=205049-205449` and with counting the cache + the trie reads.

Before: Trie: 1840 Cache: 6475
After: Trie: 817 Cache: 7455

So this *does* save a signifcant amount of Trie reads. Execution time did not change along this test (on my (old) laptop always at 9 seconds for the block range from above). I guess there might be a more signifcant effect later on when having a larger trie, couldn't test this since my local mainnet sync is not that much further advanced in the chain.

With this change we would need to care a bit more about cache size (this might be a good idea in general), will give this some thought here along.

### DIFF
--- a/packages/vm/package.json
+++ b/packages/vm/package.json
@@ -64,6 +64,7 @@
   "devDependencies": {
     "@types/benchmark": "^1.0.33",
     "@types/core-js": "^2.5.0",
+    "@types/functional-red-black-tree": "^1.0.1",
     "@types/lru-cache": "^5.1.0",
     "@types/minimist": "^1.2.2",
     "@types/node": "^11.13.4",
--- a/packages/vm/src/state/cache.ts
+++ b/packages/vm/src/state/cache.ts
@@ -1,5 +1,5 @@
 import { Account, Address } from 'ethereumjs-util'
-const Tree = require('functional-red-black-tree')
+import createTree from 'functional-red-black-tree'
 
 export type getCb = (address: Address) => Promise<Account | undefined>
 export type putCb = (keyBuf: Buffer, accountRlp: Buffer) => Promise<void>
@@ -15,7 +15,7 @@ export interface CacheOpts {
  * @ignore
  */
 export default class Cache {
-  _cache: any
+  _cache: createTree.Tree<string, any>
   _checkpoints: any[]
 
   _getCb: getCb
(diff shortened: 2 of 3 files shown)


packages/vm/package.json
```diff
@@ -64,6 +64,7 @@
   "devDependencies": {
     "@types/benchmark": "^1.0.33",
     "@types/core-js": "^2.5.0",
+    "@types/functional-red-black-tree": "^1.0.1",
```
holgerd77, line 67: Couldn't 100% decide if to put in dependencies or dev dependencies, comments appreciated!

acolytec3: I thought types always went in devDependencies.  That's where I'd lean.

holgerd77: It's a bit more differentiated. If you are exporting the type in some way, e.g. if it is used for a parameter as an input type or if a return value on the API is returning a value with the respective type, the types package should be in the main dependencies, see first answer from [here](https://stackoverflow.com/questions/45176661/how-do-i-decide-whether-types-goes-into-dependencies-or-devdependencies) for a good summary.

In this specific case I wondered if people who fuddle around with their own StateManager implementation (and e.g. eventually also accessing the cache in some way) would benefit from a direct dependency on this.

Not sure, but eventually overblow. So for now here I decided to put into dev dependencies.

holgerd77:
To be fully honest: after this long period of time and with only a short look I cannot completely recall if the logic applied in this PR actually makes sense or not respectively is still relevant. 🙂 🤔 If it is change would be significant and path eventually worth to follow respectively to complete here. Will leave this open for now. Additional comments welcome!

PR #1573, opened 2021-11-16
vm: dedupe ci step

ryanio:
I noticed we were running `npm run test:API` twice in our ci since it is already run during `coverage:test`, so this PR dedupes it.


holgerd77 (approved):
LGTM

PR #1579, opened 2021-11-20
monorepo, vm: remove npm vulnerabilities

emersonmacro:
Part of the work for #1537.

VM: `tap-spec` hasn't been updated in 3 years and npm reports it as having a high vulnerability. I found a fork of `tap-spec` that was last updated 6 months ago and does not seem to have the same vulnerability. https://www.npmjs.com/package/@randomgoods/tap-spec

Monorepo: I found a workaround for `npm login` in the `e2e-publish` script, so `npm-auth-to-token` can also be removed now.

### DIFF
--- a/scripts/e2e-publish.sh
+++ b/scripts/e2e-publish.sh
@@ -23,30 +23,10 @@ verdaccio --config verdaccio.yml &
 npx wait-port 4873
 
 # `npm add user`
-# curl -XPUT \
-#    -H "Content-type: application/json" \
-#    -d '{ "name": "test", "password": "test" }' \
-#    'http://localhost:4873/-/user/org.couchdb.user:test'
 TOKEN=$(curl -XPUT \
   -H "Content-type: application/json" \
   -d '{ "name": "test", "password": "test" }' \
   'http://localhost:4873/-/user/org.couchdb.user:test')
-
-# `npm login`
-# export NPM_USER=test
-# export NPM_PASS=test
-# export NPM_EMAIL=test@test.com
-# npx npm-login-cmd --registry=http://localhost:4873
-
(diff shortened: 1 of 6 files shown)


holgerd77:
Hi Emmett,
thanks for tackling this! 😄 

Let's please not introduce dependencies on forks with such low community establishment and usage per week for security reasons - even if it is "only" for dev dependencies. Especially for this `tap-spec` use case this is not worth it.

Can you please remove and go with the version I suggested, so to just not have any default here installed at all?

emersonmacro:
@holgerd77 Good call, PR updated 👍

scripts/e2e-publish.sh
```diff
@@ -23,17 +23,13 @@ verdaccio --config verdaccio.yml &
...
-  -p test \
-  -e test@test.com \
-  -r http://localhost:4873
+TOKEN=$(curl -XPUT \
+  -H "Content-type: application/json" \
+  -d '{ "name": "test", "password": "test" }' \
+  'http://localhost:4873/-/user/org.couchdb.user:test')
+
+npm set registry "http://localhost:4873"
+npm set //localhost:4873/:_authToken $TOKEN
```
holgerd77, line 32: This direct replacement is really cool! 😄 💯

holgerd77 (approved):
Thanks for the update, nice, important update, feels a lot better to have these things fixed at least on this level. This should make it substantially more unlikely that a severe vulnerability on the main dependencies slips through at some point.

We should really keep up here in the future and have a look and some updates in - somewhat - regular intervals.

(and it of course also wouldn't hurt to fix the moderate vulnerabilities as well. Important first step here though)

PR #1580, opened 2021-11-20
Add MIT License File

michael-spengler:
This PR adds the [corresponding](https://github.com/ethereumjs/ethereumjs-monorepo#license) license file. This shall simplify the license check / license filter - e.g. via npmjs.com ...

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 ethereumjs
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


holgerd77:
Hi there,
this `MIT` note on the main README file is actually highly misleading respectively wrong, sorry sorry, I guess this sneaked in when we created the monorepo and I guess the person doing this README write-up has just been adding this occasionally without too much further thinking about this.

Close to all the monorepo libraries are `MPL-v2.0` licensed, see the respective package folders. Hope this doesn't cause any trouble or impose unexpected restrictions. While we would also prefer MIT ourselves, we are restricted to that due to historic reasons. `MPL-v2.0` is nevetheless also not a very restrictive license and should be unproblematic for most use cases.

Have opened #1582, will close this PR here in return.

Issue #1558, opened 2021-11-08
Add Sepolia PoW chain support


jochem-brouwer:
One of the questions at the ArrowGlacier PR was (see #1527): why is there no difficulty bomb delay planned for Ropsten? Well, this is because the testnet will be deprecated and replaced by Sepolia. See https://github.com/ethereum/go-ethereum/pull/23730

It seems that this will replace Ropsten after the merge, so it has no high priority, but it is anyways nice to implement this at some point and should not be a lot of work.

CC @holgerd77

holgerd77:
Ah, oh, that's nice, have a new on-the-beginning lightweight PoW testnet at hand soon. 🙂

holgerd77:
Ah, just remember that this was also discussed along the Merge even, lol (oh oh, the memories). 😜

emersonmacro:
@holgerd77 @jochem-brouwer I took a stab at this here #1581. I haven't set up a new chain before so there might be something obvious that I missed - any feedback welcome. Specifically in `chains/sepolia.json`, I couldn't find anything for `url` or `dnsNetworks`, and not sure if I got the hardforks right.

PR #1581, opened 2021-11-22
common: add support for new Sepolia chain

emersonmacro:
Initial attempt at tackling #1558. Information for the new chain came primarily from the geth issue & PR here:
https://github.com/ethereum/go-ethereum/pull/23730#issuecomment-942646618
https://github.com/ethereum/go-ethereum/pull/23730/files#diff-c13a1b5c84048a832e07e3503cc35f66614816cb39fe1e2fdd992672a9e1d883R46


packages/common/src/chains/sepolia.json
```diff
@@ -0,0 +1,111 @@
...
+  "name": "sepolia",
+  "chainId": 11155111,
+  "networkId": 11155111,
+  "defaultHardfork": "istanbul",
+  "consensus": {
+    "type": "pow",
+    "algorithm": "ethash",
+    "ethash": {}
+  },
+  "comment": "PoW test network",
```
holgerd77, line 11: Something more specific would be nice here "PoW test network to replace Ropsten" e.g..

packages/common/src/chains/sepolia.json
```diff
@@ -0,0 +1,111 @@
...
+  "url": "https://github.com/ethereum/ropsten",
```
holgerd77, line 12: Please update this URL with something fitting, a robust PR URL like https://github.com/ethereum/go-ethereum/pull/23730 would also be ok (this info is not read "by machine" anywhere, just some helper for someone digging into the chain files).

packages/common/src/chains/sepolia.json
```diff
@@ -0,0 +1,111 @@
...
+  "genesis": {
+    "hash": "0x25a5cc106eea7138acab33231d7160d69cb777ee0c2c553fcddf5138993e6dd9",
+    "timestamp": "0x6159af19",
+    "gasLimit": 30000000,
+    "difficulty": 131072,
+    "nonce": "0x000000000000000",
+    "extraData": "0x5365706f6c69612c20417468656e732c204174746963612c2047726565636521",
+    "stateRoot": "0x5eb6e371a698b8d68f665192350ffcecbbbf322916f4b51bd79bb6887da3f494"
```
holgerd77, line 20: Did you took over `hash` and `stateRoot` here from somewhere or did you calculate yourself?

Can you confirm the correctness of both?

holgerd77 (commented):
Review is not yet complete (to be continued by me or someone else).

Is this Sepolia network already running? If so: have you (@emersonmacro) tested this setup in the client?

packages/common/src/chains/sepolia.json
acolytec3: I'll need to compare in more detail but here's what Besu is using for Genesis config.https://github.com/hyperledger/besu/blob/main/config/src/main/resources/sepolia.json

emersonmacro: I got them from this comment on the original geth PR - https://github.com/ethereum/go-ethereum/pull/23730#issuecomment-942646618. I'll see if I can confirm them with our client

holgerd77: Ah, ok, but that's also already a solid source. 👍

packages/common/src/chains/sepolia.json
```diff
@@ -0,0 +1,111 @@
...
+    }, 
+    {
+      "name": "london",
+      "block": 0, 
+      "forkHash": "0x30c7ddbc"
+    },
+    {
+      "name": "arrowGlacier",
+      "block": 0,
+      "forkHash": "0x30c7ddbc"
```
holgerd77, line 81: If I am not missing something, the `arrowGlacier` HF here should not be included respectively is not applied on `sepolia` yet. I guess `arrowGlacier` will be applied at some point (?) but maybe for now remove the entry completely.

packages/common/src/chains/sepolia.json
```diff
@@ -0,0 +1,111 @@
...
+    {
+      "name": "tangerineWhistle",
+      "block": 0,
+      "forkHash": "0x30c7ddbc"
+    },
+    {
+      "name": "spuriousDragon",
+      "block": 0,
+      "forkHash": "0x30c7ddbc"
+    },
```
holgerd77, line 42: The `arrowGlacier` removal now also changes the fork hash (on all places), this is now `0xfe3366e7`.

packages/common/src/chains/sepolia.json
```diff
@@ -0,0 +1,111 @@
...
+    "ethash": {}
+  },
+  "comment": "PoW test network to replace Ropsten",
+  "url": "https://github.com/ethereum/go-ethereum/pull/23730",
+  "genesis": {
+    "hash": "0x25a5cc106eea7138acab33231d7160d69cb777ee0c2c553fcddf5138993e6dd9",
+    "timestamp": "0x6159af19",
+    "gasLimit": 30000000,
+    "difficulty": 131072,
+    "nonce": "0x000000000000000",
```
holgerd77, line 18: The nonce is off by one 0.

holgerd77 (changes requested):
Some change requests, I will directly push these after the review.

The initial connection to the bootnode is now working with this. 

We get the following follow-up error though:

grafik

Guess that has something to do with the genesis block which might not be written properly to the chain. This will need some debugging. A good starting point here is likely the blockchain library and there the related genesis block initialization code parts.

holgerd77:
What's the status of this?

emersonmacro:
> What's the status of this?

I was able to get the same initial connection to the bootnode and subsequent error that you posted. I'm doing some debugging trying to figure out what's causing the error.

ryanio:
sync working! smooth so far

Screen Shot 2021-12-30 at 10 14 55 AM

ryanio:
i successfully sync'd to head but not sure why it didn't continue following the tip of the chain

Screen Shot 2021-12-30 at 11 17 02 AM

This error:
> `(node:24407) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added to [BlockFetcher]. Use emitter.setMaxListeners() to increase limit`

is concerning, might be related to `lib/sync/fetcher/fetcher.ts:L348` -> `this.on('close', () => {`

jochem-brouwer:
Maybe no one is mining Sepolia so there are no new blocks? Check the timestamp of the latest block?

holgerd77:
But this Sepolia network is a great chance to really really test our client in production, and all the related tip of the chain behavior with it! 😀

emersonmacro:
I got it fully synced yesterday and did not see the same error:
image

Just restarted the client and it synced back up very quickly, at a slightly higher block height:
image

ryanio (approved):
this LGTM!

we can address following the tip of chain behavior in a separate PR, and this should be a nice testnet to do that with :)

ryanio (approved):

PR #1582, opened 2021-11-22
Monorepo: Correct License Note

holgerd77:
At some point a general MIT license note sneaked in to the overall [monorepo README](https://github.com/ethereumjs/ethereumjs-monorepo). This is misleading and actually wrong since close to all libraries are `MPL-2.0` licensed (for historical reasons). This PR corrects that.

See also: #1580


acolytec3 (approved):
LGTM

Issue #1583, opened 2021-11-22
Error in usage demo on main page


samlaf:
The [usage demo script](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/util#usage) on the main page has a problem:

`assert.equal(unpadBuffer(Buffer.from('000000006600', 'hex')), Buffer.from('6600', 'hex'))`
raises a weird error

> internal/process/esm_loader.js:74
>   internalBinding('errors').triggerUncaughtException(
>                               ^
> 
> AssertionError [ERR_ASSERTION]: Buffer(2) [Uint8Array] [
>   102,
>   0
> ] == Buffer(2) [Uint8Array] [
>   102,
>   0
> ]
>     at file:///home/samlaf/dev/ethereum/ecdsa/tx_demo.mjs:7:8
>     at ModuleJob.run (internal/modules/esm/module_job.js:170:25)
>     at async Loader.import (internal/modules/esm/loader.js:178:24)
>     at async Object.loadESM (internal/process/esm_loader.js:68:5) {
>   generatedMessage: true,
>   code: 'ERR_ASSERTION',
>   actual: Buffer(2) [Uint8Array] [ 102, 0 ],
>   expected: Buffer(2) [Uint8Array] [ 102, 0 ],
>   operator: '=='
> }

this is because equality doesn't work for buffers (for some reason that I don't understand):
```
>>> Buffer.from('11', 'hex') == Buffer.from('11', 'hex')
    false
>>> Buffer.from('11', 'hex').equals(Buffer.from('11', 'hex'))
    true
```

So instead the usage demo should be changed to
`assert.ok(unpadBuffer(Buffer.from('000000006600', 'hex')).equals(Buffer.from('6600', 'hex')))`

acolytec3:
Thanks for this.  I've checked and verified this issue and have a PR to resolve it.

samlaf:
Also just realized none of the [module links](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/util#modules) below work.

ryanio:
> Also just realized none of the [module links](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/util#modules) below work.

Thanks, we should remove those links, typedoc likes to play funny games and change permalinks often so it's fruitless to keep trying to match it. It might help if we also move the link to /docs to the top of the section.

holgerd77:
[...]

PR #1585, opened 2021-11-23
util: update readme

ryanio:
Fixes https://github.com/ethereumjs/ethereumjs-monorepo/issues/1583#issuecomment-975862761 by correcting the module links, and some other small adjustments.


ryanio:
I noticed trie, util, blockchain, and ethash coverage reports have been missing from codecov.

looks like I've fixed them with this PR, although still debugging why ethash isn't showing up in codecov's files tab under packages (it is receiving the uploaded report as i can download it from codecov interface)

ryanio:
alright, well we can still merge this in for now and continue to debug the codecov situation, ethash is still missing and the two top level blockchain ts files as well (index.ts and clique.ts, `db` folder is showing up) which is so weird because the packages have identical setups and configs so I'm not sure why these packages are particularly different. again, the reports seems to be properly uploading to codecov.

packages/blockchain/.nycrc
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "../../config/nyc.json",
   "include": [
-    "src/*.ts"
+    "src/**/*.ts"
```
holgerd77, line 4: Nice catches. 👍

.github/workflows/block-build.yml
```diff
@@ -38,7 +38,7 @@ jobs:
 
       - run: npm run coverage
 
-      - uses: codecov/codecov-action@v1
+      - uses: codecov/codecov-action@v2
```
holgerd77, line 41: What is the practical effect of this change?

config/nyc.json
```diff
@@ -1,5 +1,6 @@
 {
   "reporter": ["lcov", "text"],
   "extension": [".ts"],
-  "all": true
+  "all": true,
+  "cache": false
```
holgerd77, line 5: Did this achieve what it should? Just to reassure we are not unnecessarily sacrifice on performance here.

holgerd77 (approved):
LGTM

.github/workflows/block-build.yml
ryanio: nothing really just staying up to latest, from the changelog it says that it just uses their new uploader code, i hoped it would help with the missing coverage files but didn't seem to

config/nyc.json
ryanio: didn't seem to make any difference, but i think it's a good change for us, because from nyc's [caching](https://github.com/istanbuljs/nyc#caching) information it says that it stores the cache in `<project-root>/node_modules/.cache/nyc` and github actions cache stores the whole node_modules folder, so this should help keep differing ci runs to not use the same cache in case the files have changed

PR #1587, opened 2021-11-26
Update dockerfile

Ghorbanian:
Docker fails to build image because docker file is installing decommissioned version (2) of python. Updated to install Python 3. Refer to https://stackoverflow.com/questions/62169568/docker-alpine-linux-python-missing.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM node:14-alpine as build
 WORKDIR /usr/app
-RUN apk update && apk add --no-cache g++ make python bash git && rm -rf /var/cache/apk/*
+RUN apk update && apk add --no-cache g++ make py-pip bash git && rm -rf /var/cache/apk/*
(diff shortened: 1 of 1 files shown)


ryanio:
Thanks @Ghorbanian, actually I already ran into this and found we only need to install "bash" and "git" dependencies - have updated the Dockerfile in https://github.com/ethereumjs/ethereumjs-monorepo/blob/merge-kintsugi/Dockerfile - could you give it a try and let me know if it works for you?

Ghorbanian:
> Thanks @Ghorbanian, actually I already ran into this and found we only need to install "bash" and "git" dependencies - have updated the Dockerfile in https://github.com/ethereumjs/ethereumjs-monorepo/blob/merge-kintsugi/Dockerfile - could you give it a try and let me know if it works for you?

Works like a charm!

PR #1588, opened 2021-11-27
client: fix libp2p sync and browser example

ryanio:
This PR:
 * Fixes `Buffer is not defined` from webpack by specifying the polyfill
 * Fixes libp2p sync by removing a premature `PEER_CONNECTED` in `peer:connected`
 * Fixes browser IndexedDB cross-session data reloading based on chain name
 * Fixes the `browser_sync.png` image location and updates to the latest expected output

To help test this PR, please try running through the steps in `examples/light-browser-sync.md`


packages/client/browser/index.ts
```diff
@@ -46,34 +47,41 @@ export * from './logging'
...
-  logger.info(`Connecting to network: ${common.chainName()}`)
+  const { logger, chainCommon } = client.config
+  logger.info(`Initializing Ethereumjs client network=${chainCommon.chainName()}`)
   client.config.events.on(Event.SERVER_ERROR, (err) => logger.error(err))
   client.config.events.on(Event.SERVER_LISTENING, (details) => {
     logger.info(`Listener up transport=${details.transport} url=${details.url}`)
   })
   client.config.events.on(Event.SYNC_SYNCHRONIZED, (height) => {
-    logger.info(`Synchronized blockchain at height ${height.toNumber}`)
+    logger.info(`Synchronized blockchain at height ${height}`)
```
holgerd77, line 82: Some side note: I would argue that (minimally) all log messages which are not initialization (in the sense of CLI setup and parsing) specific but rather react to some control flow events further down the line (so: e.g. the two events from above) are misplaced both here and in `bin/cli.ts` and just produce code redundancy and should rather be moved in to the core code base under `lib`.

Does that make sense or am I missing some aspect?

packages/client/browser/libp2pnode.ts
```diff
@@ -3,16 +3,14 @@
...
  */
 
-import LibP2p from 'libp2p'
 import { NOISE } from '@chainsafe/libp2p-noise'
+import LibP2p from 'libp2p'
 import PeerId from 'peer-id'
 import { Multiaddr } from 'multiaddr'
-// types currently unavailable for below libp2p deps,
-// tracking issue: https://github.com/libp2p/js-libp2p/issues/659
-const LibP2pWebsockets = require('libp2p-websockets')
```
holgerd77, line 12: Just out of interest (without further digging in): the referenced issue here https://github.com/libp2p/js-libp2p/issues/659 is still open, could this nevertheless be resolved?

holgerd77 (commented):
Hi there,
nice that you are tackling this 🙂, have left some first comments.

There is still 1 client test failing here, not sure if you have seen this.

I would be interested to do a production test on this along the browser example lines, not sure if I'll be getting to that though, just as some note.

packages/client/webpack.config.js
```diff
@@ -1,8 +1,17 @@
...
   mode: 'production',
   entry: './dist.browser/browser/index.js',
+  plugins: [
+    new ProvidePlugin({
+      Buffer: ['buffer', 'Buffer'],
+    }),
+    new ProvidePlugin({
+      process: 'process/browser',
+    }),
+  ],
```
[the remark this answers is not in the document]
ryanio: thanks, and I added `process/browser` because one node_module in particular was using `process.env.SOME_VARIABLE` and adding this seemed to be the best way to harden the bundle from breaking from any other module checking `process.env` in the future.

packages/client/browser/index.ts
ryanio: yes that's a good idea :) will add in a commit

packages/client/browser/libp2pnode.ts
ryanio: I was able to convert a few of the modules to import, the rest are still on require, I think it's obvious at this point that any modules using require are due to missing types, so I thought I'd clean up/remove the comment.

ryanio:
ok this should be ready for final review/merge unless there are any other suggested changes

packages/client/webpack.config.js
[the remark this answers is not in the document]
acolytec3: I assume there aren't any other gotcha node types you can think of that would break in the browser version of the client thanks to webpack v5 not polyfilling in obscure edge cases, right?

acolytec3 (commented):
This generally looks great to me.  Only thing I see is the code coverage missing on some lines.  I'll see about adding some tests on the sync pieces to test the new logger warnings.

packages/client/webpack.config.js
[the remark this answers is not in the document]
ryanio: so far i believe i've added everything we need, webpack 5 is quite vocal if it is missing a module it needs as part of the bundle, so in case something gets added in the future we should see the error and can add it appropriately. 

although it would be nice if we had some kind of ci run that actually used the webpack bundle, it would have likely discovered this "Buffer is undefined" error before it was reported to us by a user. lately i am thinking this would be much more useful (or at least, in addition to) the karma-typescript setup we have for the client specifically.

packages/client/bin/cli.ts
```diff
@@ -33,11 +37,6 @@ const args = yargs(hideBin(process.argv))
     choices: networks.map((n) => parseInt(n[0])),
     default: undefined,
   })
-  .option('network-id', {
```
ryanio, line 38: `'network-id'` option was defined twice here

acolytec3:
@ryanio after I spent the time writing them, I'm not so certain the new tests I added for the `blockfetcher` actually test anything useful.  They do help with code coverage but methinks they're kind of pointless since they mock the part of the code that actually triggers the error or not.  Thoughts?

ryanio:
I think that's fine, thanks for adding them! definitely helped with the coverage :)

acolytec3:
And I think we finally hit the code coverage requirements too.  I've spent the last while trying to figure out how to grab the debugger output but can't for the life of me figure out how to redirect it from stderr to something I can actually check.  I'm gonna add one more small set of tests around the fetcher since we don't have much coverage there anyway and will then be done.

acolytec3 (approved):
LGTM

packages/client/bin/cli.ts
```diff
@@ -608,36 +632,38 @@ async function run() {
...
   })
   config.events.setMaxListeners(50)
 
-  const client = await runNode(config)
-  const servers = args.rpc || args.rpcEngine ? runRpcServers(client, config, args) : []
+  const client = await startClient(config)
+  const servers = args.rpc || args.rpcEngine ? startRPCServers(client) : []
 
   process.on('SIGINT', async () => {
     config.logger.info('Caught interrupt signal. Shutting down...')
```
holgerd77, line 669: Whew.

That's a really great and much needed clean-up on the CLI script!!! 😄 👍 

Nice.

🎉

packages/client/test/sync/fetcher/blockfetcher.spec.ts
```diff
@@ -101,6 +102,36 @@ tape('[BlockFetcher]', async (t) => {
...
+      st.ok(err.message === 'err0', 'store() threw on invalid block')
+    }
+    td.reset()
+    chain.putBlocks = td.func<any>()
+    td.when(chain.putBlocks(td.matchers.anything())).thenResolve(1)
+    config.events.on(Event.SYNC_FETCHER_FETCHED, () =>
+      st.pass('store() emitted SYNC_FETCHER_FETCHED event on putting blocks')
+    )
+    await fetcher.store([])
+  })
```
holgerd77, line 134: Nice, all these new test cases are really great as well! 😄 

Especially the Fetcher is a beasty thing, so that's nice if we get to better coverage there, also for reproducing things and having certain functionality parts more directly called.

PR #1593, opened 2021-12-02
Ethtests 10.2 cleanup items

acolytec3:
Clean-up items from #1568 
 - Add explanatory notes to `validateNoLeadingZeroes` method in `util`
 - Clarify error message in VM related to EIP3607
 - Clean up error message in `legacyTransaction` constructor related to integer overflow
 - Add test to demonstrate `0x01` is valid RLP encoding (since leading zeroes only occur if there are leading zero bytes (e.g. `0x001` which translates to `Buffer<00 01>`

### DIFF
--- a/packages/tx/src/legacyTransaction.ts
+++ b/packages/tx/src/legacyTransaction.ts
@@ -111,7 +111,8 @@ export default class Transaction extends BaseTransaction<Transaction> {
     this.gasPrice = new BN(toBuffer(txData.gasPrice === '' ? '0x' : txData.gasPrice))
 
     if (this.gasPrice.mul(this.gasLimit).gt(MAX_INTEGER)) {
-      throw new Error('gas limit * price overflow')
+      const msg = this._errorMsg('gas limit * gasPrice cannot exceed MAX_INTEGER(2^256-1)')
+      throw new Error(msg)
     }
     this._validateCannotExceedMaxInteger({ gasPrice: this.gasPrice })
 
--- a/packages/tx/src/legacyTransaction.ts
+++ b/packages/tx/src/legacyTransaction.ts
@@ -111,7 +111,7 @@ export default class Transaction extends BaseTransaction<Transaction> {
     this.gasPrice = new BN(toBuffer(txData.gasPrice === '' ? '0x' : txData.gasPrice))
 
     if (this.gasPrice.mul(this.gasLimit).gt(MAX_INTEGER)) {
-      const msg = this._errorMsg('gas limit * gasPrice cannot exceed MAX_INTEGER(2^256-1)')
+      const msg = this._errorMsg('gas limit * gasPrice cannot exceed MAX_INTEGER (2^256-1)')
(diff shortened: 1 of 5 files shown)


packages/util/src/bytes.ts
```diff
@@ -279,7 +279,16 @@ export const baToJSON = function (ba: any): any {
...
+ *
+ * Examples:
+ *
+ * Valid values: 0x1, 0x, 0x01, 0x1234
+ * Invalid values: 0x0, 0x00, 0x001, 0x0001
+ *
+ * Note: This method is useful for validating that RLP encoded integers comply with the rule that all
+ * integer values encoded to RLP must be in the most compact form and contain no leading zeroes
  * @param values An object containing string keys and Buffer values
+ * @throws if any provided value is found to have leading zeroes
```
jochem-brouwer, line 291: This comment is not entirely correct, as `0x01` is valid; should be something like "throws if value has any (partial) leading zero bytes"

acolytec3: Good call.  I'll update once I validate this on the R&D discord.

packages/tx/src/legacyTransaction.ts
```diff
@@ -111,7 +111,8 @@ export default class Transaction extends BaseTransaction<Transaction> {
     this.gasPrice = new BN(toBuffer(txData.gasPrice === '' ? '0x' : txData.gasPrice))
 
     if (this.gasPrice.mul(this.gasLimit).gt(MAX_INTEGER)) {
-      throw new Error('gas limit * price overflow')
+      const msg = this._errorMsg('gas limit * gasPrice cannot exceed MAX_INTEGER(2^256-1)')
```
ryanio, line 114:
```suggestion
      const msg = this._errorMsg('gas limit * gasPrice cannot exceed MAX_INTEGER (2^256-1)')
```

the eip1559Transaction error said `2^256` as opposed `2^256-1` so probably want to align them (also the space after MAX_INTEGER is more understandable imo)

packages/util/src/bytes.ts
```diff
@@ -279,7 +279,16 @@ export const baToJSON = function (ba: any): any {
...
 /**
  * Checks provided Buffers for leading zeroes and throws if found.
+ *
+ * Examples:
+ *
+ * Valid values: 0x1, 0x, 0x01, 0x1234
+ * Invalid values: 0x0, 0x00, 0x001, 0x0001
+ *
+ * Note: This method is useful for validating that RLP encoded integers comply with the rule that all
+ * integer values encoded to RLP must be in the most compact form and contain no leading zeroes
```
jochem-brouwer, line 289:
```suggestion
 * integer values encoded to RLP must be in the most compact form and contain no leading zero bytes
```

jochem-brouwer (approved):
LGTM!

ryanio (approved):
nice, lgtm, thanks for addressing these!

holgerd77:
Nice. 🙂

PR #1598, opened 2021-12-08
client: allow past block numbers in RPC queries

ryanio:
This PR enables the RPC layer to use past block numbers in `call`, `getBalance`, `getCode`, `getStorageAt`, `getTransactionCount`, and `estimateGas`.

This is possible by setting the vm's stateManager's stateRoot to the block's stateRoot. Thanks @jochem-brouwer for the tip :) The logic is fully tested in `getBalance.spec.ts` by comparing a past block to the latest.

The second commit (unrelated) simplifies some of the tape result comparisons.


jochem-brouwer:
Have given this a quick glance, will give it a more thorough review later, but this looks good - and also much more simple than I expected 😄

jochem-brouwer:
Could we wait merging this until #1590 is merged? Then we can also include these changes for the getProof RPC call :)

packages/client/test/rpc/eth/call.spec.ts
```diff
@@ -10,48 +11,65 @@ import { checkError } from '../util'
 const method = 'eth_call'
 
 tape(`${method}: call with valid arguments`, async (t) => {
-  const blockchain = await Blockchain.create()
+  const blockchain = await Blockchain.create({ validateBlocks: false, validateConsensus: false })
```
jochem-brouwer, line 14: How come this passed before? It should throw on invalid PoW blocks right?

packages/client/test/rpc/eth/getBalance.spec.ts
```diff
@@ -54,27 +27,54 @@ tape(`${method}: ensure balance deducts after a tx`, async (t) => {
...
+  req = params(method, [address.toString(), '0x0'])
+  expectRes = (res: any) => {
+    const msg = 'should return the correct balance with a past block number'
+    t.equal(res.body.result, bnToHex(genesisBalance), msg)
+  }
+  await baseRequest(t, server, req, 200, expectRes, false)
+
+  // call with height that exceeds chain height
+  req = params(method, [address.toString(), '0x1'])
+  expectRes = checkError(t, INVALID_PARAMS, 'specified block greater than current height')
```
jochem-brouwer, line 77: This test in general looks great, but it does not explicitly check that when we pass a specific block, it actually uses that block. Probably just create 2 blocks (dont have to be valid) so we have blocks 0 (earliest), 1 and 2 (latest). Now you can check explicitly for earliest/latest (these are two different blocks), explicit 0x0, explicit 0x2, explicit 0x3 (throws) and explicit 0x1. Just check the balance of coinbase of all blocks should be fine.

Also 2 questions: (1) what happens if we pass a negative number over RPC? and (2) in general what happens if the RPC throws? Does it then still return a request with an error code, or does it fail to reply to the request?

jochem-brouwer (commented):
This PR looks very great! I love also the improvements in the tests so that these are more compact.

I have 1 request to expand the tests and some not super-related questions.

jochem-brouwer:
BTW, this PR also shows that we should really start creating a standalone package for StateManager soon. It does not make a lot of sense to use the VM interface when we have "simple" state lookups such as getCode/getBalance and alike. It only makes sense when we actually simulate txs, such as `call` or `estimateGas`.

packages/client/test/rpc/eth/call.spec.ts
ryanio: before we were just using `vm.runTx`, but now are creating blocks and adding them to the blockchain (so our source code can setStateRoot properly), so needed these additional overrides

packages/client/test/rpc/eth/getBalance.spec.ts
ryanio: good idea, will add another block, thanks.

1. numbers are only passed as hex encoded over the RPC layer, so something like `-0x10`?

Trying with `toBuffer` which we mostly use internally to format, I get:
> Error: Cannot convert string to buffer. toBuffer only supports 0x-prefixed hex strings and this string was given: -0x10

2. the RPC needs to throw an error of shape `{ code, message }` to properly transmit as an JSON-RPC formatted error object. otherwise if something throws deeper, sometimes it may return as `Internal Error` with no error object/msg displayed, which is annoying and challenging to debug :) so you can see in some delicate places in the engine API I do:

```typescript
} catch (error: any) {
      throw {
        code: INTERNAL_ERROR,
        message: error.message.toString(),
      }
}
```

jochem-brouwer: Ah great thanks.

For (2) I think we should log the entire error message in the console? In Geth, for instance, if you do an RPC call and there is some internal error (e.g. trying to mine without etherbase/coinbase set) then it puts `[WARN] <error message>` in the console.

jochem-brouwer: (This is not in the scope of this PR though)

acolytec3 (approved):
Looks good, especially the test clean up.  I experimented with some rpc calls to a couple of the `eth` methods and got good results.

holgerd77:
Will merge here.

packages/client/test/rpc/eth/getBalance.spec.ts
ryanio: yeah I agree with you, haven't found a straightforward way to forward all errors like this through `jayson` yet, definitely worth more investigation to see if it is possible (i assume it should be, we have to place a catch at a higher level and throw with the right obj shape)

PR #1600, opened 2021-12-09
devp2p: ETH-LES class refactor

### LINKED ISSUE #1450: devp2p: Add superclass to ETH and LES protocol implementations
There is actually a lot of redundancy between the ETH and LES protocol implementation. This should be combined into a superclass sitting between e.g. `ETH` and `EventEmitter` at some point. This will likely lead to a substantially cleaner implementation of the subclasses, increase readability as well as ease adding new features and avoid introducing consistency gaps along the way.

(this new superclass should likely be not exposed in any way to the users, on a first run it would likely also be good to keep the API unchanged and eventually rather wrap calls if a method signature is rather realized in a more optimal way in the superclass)

Also totally possible to start with a subset of superclass integrations and then expand along the way "on occasion".

This is a good first issue for getting a bit more familiar with the `devp2p` library.

Ghorbanian:
Creates a new super-class to consolidate commonalities between the ETH and LES protocols. Fixes #1450.


Ghorbanian:
@ryanio @holgerd77 Anything I'm missing?

ryanio:
will try to review tomorrow! 🙏

packages/devp2p/src/ExchangeProtocol.ts
```diff
@@ -0,0 +1,23 @@
...
+    this._peer = peer
+    this._statusTimeoutId = setTimeout(() => {
+      this._peer.disconnect(DISCONNECT_REASONS.TIMEOUT)
+    }, ms('5s'))
+  }
+
+  getVersion() {
+    return this._version
+  }
+}
```
holgerd77, line 23: Don't change this yet until there is some more confirmation from other team members, but I am not so satifsfied with the overall (already existing) directory structure and naming here. So to have this file on the top level is somewhat suboptimal respectively too exposed. And these dedicated directories for `eth`, `les` are also somewhat overblown.

I guess we might be good here to rename since this is all re-exported by the main `index.ts` file? 🤔

My suggestion here would be to follow a bit the naming from similar client directory structures and do:

- `src/eth/index.ts` -> `src/protocol/eth.ts`
- `src/les/index.ts` -> `src/protocol/les.ts`
- `src/ExchangeProtocol.ts` -> `src/protocol/protocol.ts`

And also for the class name: `ExchangeProtocol` -> `Protocol`

What do others think?

holgerd77: Also, just noting here before I forget, should likely only be done after a renaming though: this new protocol class should be added to the `exclude` directive from the main `typedoc.ts` file so that it doesn't get included into the code documentation on generation.

Maybe a code doc that this class is only internal and shouldn't be directly used would also not hurt, so that this doesn't sneek into the public API.

holgerd77 (changes requested):
Great start, cool, looking forward to have this implemented! 😄

The naming thing is somewhat important, we should give this some thought.

Also have some suggestions for additional unifications, will address in a separate comment.

Generally also really nice to review, thanks for splitting this up in such easy to follow steps, this helps a lot! 🙂

holgerd77:
Ok, so the `initMsgDebuggers()` function is actually already a good example for why this PR/work is really useful. This can actually be unified as well, code is slightly different in `ETH` and `LES` but for `LES` this is just because some functionality has been forgotten to be added. 😋 So you can just take the `ETH` version of the code and unify (I guess with keeping subclass functions passing over the message code enums).

Same goes for `_addFirstPeerDebugger()` and `debug()`.

Regarding the global `debug` object, I guess you can also switch over to pass the `DEBUG_BASE_NAME` in the constructor to the super class and initialize a private (protected) `_debug` class member instead of the global `debug`  variable. Then the rest of the functions should (probably ?) work.

Member functions which had been `private` before need to be made `protected` with this move.

packages/devp2p/src/ExchangeProtocol.ts
Ghorbanian: @holgerd77 Just unsure about where to document that `Protocol` shouldn't be directly used. Should that be done inside `README.md` in the `devp2p` package?

ryanio:
thanks @Ghorbanian! this looks really great. I'll give this a more in-depth review now.

@holgerd77 do you have any further comments?

@Ghorbanian do you have anything else you'd like to add? should we move from WIP to "needs review"?

packages/devp2p/typedoc.js
```diff
@@ -2,5 +2,5 @@ module.exports = {
   extends: '../../config/typedoc.js',
   entryPoints: ['src'],
   out: 'docs',
-  exclude: ["test/**/*.ts", "examples/*.ts"],
+  exclude: ["test/**/*.ts", "examples/*.ts", "src/protocol/protocol.ts"],
```
ryanio, line 5: I think it would be okay to keep the generated documentation for the parent class, did you have a reason to exclude it?

packages/devp2p/src/protocol/protocol.ts
```diff
@@ -0,0 +1,83 @@
+import ms from 'ms'
+import { debug as createDebugLogger } from 'debug'
+import { EventEmitter } from 'events'
+import { Peer, DISCONNECT_REASONS } from '../rlpx/peer'
+
+export class Protocol extends EventEmitter {
+  _peer: Peer
+  _statusTimeoutId: NodeJS.Timeout
+  _message_codes: any
```
ryanio, line 9: let's stick to camelCase and name this `_messageCodes`

packages/devp2p/typedoc.js
Ghorbanian: @holgerd77 Mentioned in one of his review comments that it might be a good idea to keep people from using it directly.

packages/devp2p/src/protocol/protocol.ts
Ghorbanian: Good idea. I've been jumping between different style standards.

Ghorbanian:
> thanks @Ghorbanian! this looks really great. I'll give this a more in-depth review now.
> 
> @holgerd77 do you have any further comments?
> 
> @Ghorbanian do you have anything else you'd like to add? should we move from WIP to "needs review"?

Should be good to move to "needs review."

packages/devp2p/src/protocol/protocol.ts
```diff
@@ -0,0 +1,83 @@
...
+import { debug as createDebugLogger } from 'debug'
+import { EventEmitter } from 'events'
+import { Peer, DISCONNECT_REASONS } from '../rlpx/peer'
+
+export class Protocol extends EventEmitter {
+  _peer: Peer
+  _statusTimeoutId: NodeJS.Timeout
+  _messageCodes: any
+  _debug: any
+  _verbose: any
```
ryanio, line 11: PR looks great! I'm about ready to approve and merge, but one last nit here maybe we can type these for added clarity?

My first pass writing it here `_messageCodes` would be something like `{ [key: string]: number }`, `_debug` you can try `createDebugLogger` or `ReturnType<createDebugLogger>`, and `_verbose` would be `boolean`?

holgerd77:
@Ghorbanian Yeah, can agree, PR looks great! 🥳

Maybe you can do the final typing suggestions from Ryan, then I would also say that this is ready to be merged.

Ah, and maybe one last quality assurance thing: did you do a final "live" test run on this within the client (make sure to build before) (see [here](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/client#using-debug-loggers) for guidance), and optimally doing 3-4 runs using these special debug loggers (first peer, IP, so basically the things which can be activated from [here](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/devp2p#introduction))?

All this debugging stuff is not at all covered by tests, and it is otherwise not really detectable if things are or are not running as expected.

But otherwise: again, great work, thanks for keeping up on this for such a long time. 🙂

Ghorbanian:
> @Ghorbanian Yeah, can agree, PR looks great! 🥳
> 
> Maybe you can do the final typing suggestions from Ryan, then I would also say that this is ready to be merged.
> 
> Ah, and maybe one last quality assurance thing: did you do a final "live" test run on this within the client (make sure to build before) (see [here](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/client#using-debug-loggers) for guidance), and optimally doing 3-4 runs using these special debug loggers (first peer, IP, so basically the things which can be activated from [here](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/devp2p#introduction))?
> 
> All this debugging stuff is not at all covered by tests, and it is otherwise not really detectable if things are or are not running as expected.
> 
> But otherwise: again, great work, thanks for keeping up on this for such a long time. 🙂

Appreciate the support along the way! 

Bear with me as I look into QA, I have a busy week ahead of me.

packages/devp2p/src/protocol/protocol.ts
```diff
@@ -6,9 +6,9 @@ import { Peer, DISCONNECT_REASONS } from '../rlpx/peer'
 export class Protocol extends EventEmitter {
   _peer: Peer
   _statusTimeoutId: NodeJS.Timeout
-  _messageCodes: any
-  _debug: any
-  _verbose: any
+  _messageCodes: { [key: string : number }
```
ryanio, line 9: i think you are missing a closing bracket here `{ [key: string]: number }`

ryanio:
awesome @Ghorbanian, thanks for updating! there is one merge conflict on `packages/devp2p/src/eth/index.ts` before the tests can run, but then it LGTM!

holgerd77:
Ah, there is another PR which happened in parallel here #1643 doing somewhat heavy changes on very much the same code parts - a bit unfortunate.

Think it might be a bit too much to ask to do a rebase/merge resolve here, will try to do myself locally and then re-push a PR, authorship should hopefully be preserved.

holgerd77:
Or maybe: @ScottyPoi, would you dare to do this? 🙂 This would also make sense, since you already did the work on #1643 and this work here goes in a bit similar direction.

So this is basically a refactoring Amir did here with combining logic from the `ETH` and `LES` classes into one super-class. Since there was also some greater consolidation opportunity on parts of the debug code, this is touching a lot of debug code as well as in your PR.

So it would be great if you can fetch the branch from Amir and try to carefully do a `git rebase master` on this. This will likely get a bit hairy and one need to be careful. 😛 If you don't feel safe enough on rebases feel free to push the ball back. 🙂 There should likely already be enough back assurance though if one gets the tests to pass.

And finally: Amir @Ghorbanian, this is not to take this out of your hands or something, but just an offer to share the work load since you also seem busy with other stuff. If you would like to finish yourself here also feel free to speak out loud and clearly. 😀 Again, just an offer where I thought that it might make sense.

ryanio (approved):
LGTM! thanks @Ghorbanian!

ryanio:
@holgerd77 do you want to give a quick review, ensure everything still works as expected with the debug loggers, etc. and if all looks good, you can merge in?

holgerd77 (approved):
I've the impression that not all DEBUG loggers are working, but I counter-tested on `master`, and there some loggers are working neither, so this is very likely not related to this PR (so e.g. `DEBUG=devp2p:eth:GET_BLOCK_HEADERS npm run client:start -- --network=sepolia` shows no output).

So I will merge here, we should address this at some point though (not super-critical though).

Thanks Ghorbanian! 🙂

ryanio:
I found the logging issue, will open a separate PR to fix

PR #1602, opened 2021-12-10
Pin typescript dependency to 4.4.4 for VM nightly tests

acolytec3:
- Bumps `libp2p-bootstrap` to 0.14.0 to resolve missing dependency issue within that dep
- Pins Typescript to v4.4.4 as workaround for [this `karma` issue](https://github.com/monounity/karma-typescript/issues/499) that causes `vm` browser tests on nightly to fail.

### DIFF
--- a/.github/workflows/vm-nightly-test.yml
+++ b/.github/workflows/vm-nightly-test.yml
@@ -35,6 +35,8 @@ jobs:
       - run: npm i
         working-directory: ${{github.workspace}}
 
+      - run: cat package-lock.json
+
       - run: npm run test:API
       - run: npm run test:API:browser
 
--- a/.github/workflows/vm-nightly-test.yml
+++ b/.github/workflows/vm-nightly-test.yml
@@ -2,6 +2,9 @@ name: VM Nightly
 on:
   schedule:
     - cron: 0 0 * * *
+  push:
+    branches:
+      - vm-nightly-typescript-version
 
 env:
   cwd: ${{github.workspace}}/packages/vm
--- a/.github/workflows/vm-nightly-test.yml
+++ b/.github/workflows/vm-nightly-test.yml
@@ -5,7 +5,7 @@ on:
   push:
     branches:
       - vm-nightly-typescript-version
-
+      
 env:
   cwd: ${{github.workspace}}/packages/vm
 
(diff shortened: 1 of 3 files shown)


ryanio:
I will spend some time tomorrow and see if I can fix this on the karma-typescript side, then we won't need to pin the TS version and we can use my temporary fork until merged upstream.

.github/workflows/vm-nightly-test.yml
```diff
@@ -2,7 +2,10 @@ name: VM Nightly
 on:
   schedule:
     - cron: 0 0 * * *
-
+  push:
```
jochem-brouwer, line 5: This does not disable the nightly tests right? It just adds it to this specific branch?

acolytec3: Right, I'll remove it from our final PR if we end up going with this solution. This was just to allow me to test without running the entire ci suite every time I committed.

ryanio:
PR in flight to fix https://github.com/monounity/karma-typescript/pull/502

acolytec3:
Your fix was merged into karma so we just need to keep an eye out for the next release so we can bump our dependency. I'll try building locally off whatever their latest commit from GitHub is and verify that it resolves the issue for us

ryanio:
I went down the rabbit hole in ethereumjs/rlp to see if this could be patched by using an updated karma-typescript location, but since it's a monorepo it needs some transformation with https://gitpkg.now.sh and then some custom scripts for a build step.

This monstrosity is what I came up with, that works [here](https://github.com/ethereumjs/rlp/pull/109/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R63):

`
"karma-typescript": "https://gitpkg.now.sh/monounity/karma-typescript/packages/karma-typescript?8c0664d864a00c8d6a3bf6d04d26cc2255600435&scripts.preinstall=npm%20config%20set%20puppeteer_skip_chromium_download%20true%20%26%26%20%%26%26%20npm%20i%20typescript%404%20--ignore-scripts%20%26%26%20npm%20i%20--only%3Ddev%20--ignore-scripts%20%26%26%20npx%20json%20-I%20-f%20tsconfig.json%20-e%20%22this.compilerOptions.noImplicitAny%3Dfalse%3Bthis.compilerOptions.skipLibCheck%3Dtrue%22%20&scripts.postinstall=%26%26%20npm%20run%20build"
`

It does these things:

1. Sets `puppeteer_skip_chromium_download` to `true` to skip that step
2. Installs typescript@4, otherwise the package-lock had an old typescript 3.9.9 that was being installed and used in karma-typescript which conflicted with our version used in the library.
3. Installs devDeps, otherwise build step has missing types
4. Updates tsconfig noImplicityAny: false and skipLibCheck: true to avoid more build step errors
5. Finally, runs build so dist can be made available

acolytec3:
Wow. I guess that works. Nice work!

PR #1603, opened 2021-12-10
VM: address consensus issue: tx goes OOG but refunds get applied anyways

### LINKED ISSUE #1601: Maybe an error with not reverting gas refunds when the contract code storage failed
Hi. I'm using ethereumjs-vm to re-execute the transactions from ethereum mainnet. Specifically this one: https://ethtx.info/mainnet/0xfd544642ce821bda8a5b90aa0d8583e52853ad7082c84d650ac0009e7773131b/
After that I check that the gas consumed is correct (consistent with the mainnet). And I'm getting an error.
Ethereumjs-vm applies the refund from a failed code.
```
  vm:evm Received message results gasUsed=128892 execResult: [ gasUsed=128892 exceptionError=[object Object] returnValue= gasRefund=undefined ] +2ms
  vm:evm message checkpoint reverted +1ms
  vm:tx ---------------------------------------------------------------------------------------------------- +0ms
  vm:tx Received tx results gasUsed=128892 execResult: [ gasUsed=128892 exceptionError=out of gas returnValue= gasRefund=15000 ] +0ms
  vm:tx Generated tx bloom with logs=0 +0ms
  vm:tx:gas tx add baseFee 271108 to gasUsed (-> 400000) +0ms
  vm:tx Subtract tx gasRefund (15000) from gasUsed (-> 385000) +0m
```

I see that there is a mechanism on `runInterpreter` to avoid that: https://github.com/ethereumjs/ethereumjs-monorepo/blob/501be8da11b07d138bb4497f8a18bdf0fa2d11d7/packages/vm/src/evm/evm.ts#L498
I think that there needs to be a similar one on `executeMessage`. I've added it locally here: https://github.com/ethereumjs/ethereumjs-monorepo/blob/501be8da11b07d138bb4497f8a18bdf0fa2d11d7/packages/vm/src/evm/evm.ts#L199 but I'm not sure if this is the best place. There is also a TODO above that seems related: https://github.com/ethereumjs/ethereumjs-monorepo/blob/501be8da11b07d138bb4497f8a18bdf0fa2d11d7/packages/vm/src/evm/evm.ts#L191

### LINKED ISSUE #1604: Check VM edge case: if contract creation goes out of gas on code deposit, check if logs/selfdestruct reset correctly
#1603 fixes a problem with gas refunds in case that contract creation execute code finishes, but there is not enough gas left to deposit the code.

There could be two more cases when this exact situation leads to a problem: it could not clear the selfdestruct list, and it could also not clear the emitted logs. [Context code](https://github.com/ethereumjs/ethereumjs-monorepo/blob/bffbc0385d0aef62097fd1bffeb9abdecdaa344d/packages/vm/src/evm/evm.ts#L492)

To setup these situations:

For selfdestruct, have a contract somewhere which selfdestructs itself. Now create a contract which calls into this selfdestruct contract upon creation, and then deposit some code. Make sure that the contract execution code finishes but that there is not enough gas left to deposit the code. In the end, the selfdestructed contract should still exist and not get selfdestructed.

For logs, create a contract which does emit some event in the constructor. Deploy the code and also make sure that execution finishes but that there is not enough gas left to deposit the code. Afterwards, check if logs are emitted: no logs should be emitted since the create frame reverted.

Possibly related to #1601

jochem-brouwer:
Closes #1601, closes #1604


jochem-brouwer:
This doesn't work, will stop here for now. The problem seems to be that we revert to old refund [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/d518e9832f6374050cff7a92e5a38146485f77a3/packages/vm/src/evm/evm.ts#L498), but this does not catch all errors which are not caught in this scope; if anything happens which makes a message scope revert then the entire refund should be forfeited.

jochem-brouwer:
CC @winsvega we have encountered an consensus bug in this PR which is not covered in `ethereum/tests`. I will write up a more detailed explanation later on, if I forget please remind me.

winsvega:
Hm, should be covered, lets check

jochem-brouwer:
Here's the problem.

We check if there's an execution error in `runInterpreter` and if thats the case we reset the refund counter to the previous refund counter [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/bffbc0385d0aef62097fd1bffeb9abdecdaa344d/packages/vm/src/evm/evm.ts#L498).

This is all fine if we don't do additional error checks afterwards. For call frames, we dont do this, however, for create frames we do this and here lies the problem. In case that we throw an [error](https://github.com/ethereumjs/ethereumjs-monorepo/blob/bffbc0385d0aef62097fd1bffeb9abdecdaa344d/packages/vm/src/evm/evm.ts#L396) over there, then we don't reset the refunds. Note that any error down the `runInterpreter` should be caught.

NOTE: there might be an uncovered edge case as well, in run interpreter in case of an error we also reset;

```
      // Clear the result on error
      result = {
        ...result,
        logs: [],
        selfdestruct: {},
      }
```

We might have to check what happens in case that in case we dont have enough gas to pay for code deposit, if we also revert the selfdestruct / logs correctly.

packages/vm/tests/api/runTx.spec.ts
```diff
@@ -518,3 +518,43 @@ tape('runTx() -> API return values', async (t) => {
...
+  nonce: 2,
+  r: '0x79d4e717d1249512572b90ca87fe545dde24815e33eb74064bdd7336463f6d8',
+  s: '0x4a16b7d119cdc34e454fa2cc0a152904f7deb23e2a5f2966f70981361c853874',
+  v: '0x26',
+  value: '0x0',
+}
+
+tape('consensus bugs', async (t) => {
+  t.test('validate out-of-gas does not refund any refunds', async (t) => {
```
acolytec3, line 543:
```suggestion
  t.test('validate out-of-gas does not give any refunds', async (t) => {
```

acolytec3 (approved):
Looks good to me.  One nigly change in the wording for the test to clarify what's happening that you can take or leave but awesome work getting a fix so quickly!

gabrocheleau (dismissed):
Looks good to me.

jochem-brouwer:
https://github.com/ethereum/tests/pull/1003 has found that the possible selfdestruct bug as described in #1605 is also present, will fix here. We probably also have the issue in #1605 with logs.

acolytec3:
@jochem-brouwer So do we need to add some additional tests for these other cases regard logs/self-destruct?

jochem-brouwer:
I could add tests, but since these test cases which I want are covered in the `ethereum-tests` PR linked above, I'd say that I run that branch locally, verify that the relevant tests pass, and then once there's a new `ethereum-tests` release we integrate it, which then implicitly means that the relevant tests which were failing before this PR then pass. The relevant selfdestruct test passes.

jochem-brouwer:
Okay, https://github.com/ethereum/tests/pull/1003 now has tests for both selfdestructs and logs into the contract creation context for cases where (0) contract gets deployed succesfully, (1) contract goes OOG due to invalid opcode, and (2) contract goes OOG because execution finishes but there is not enough gas left to deploy the code. 

Tests now pass after https://github.com/ethereumjs/ethereumjs-monorepo/pull/1603/commits/2f33d13faeddf8186a070c23c7851af0dde7fc42. We had the selfdestruct vulnerability, we did not have the log vulnerability.

ryanio (approved):
amazing speed and execution on this, thank you everyone! lgtm

PR #1606, opened 2021-12-12
tx/vm/util: throw when provided negative BN

### LINKED ISSUE #1513: Negative BNs causing weird situations as txs with negative value
Our VM is OK with running message calls and transactions where the call value is a negative BN. This should throw on both Tx and on VM. This creates ETH out of thin air. I wonder if there is any case in our stack where we would actually need negative BNs - if that's nowhere, then we should probably block negative BNs at the BN module level.

What's also weird is that we can `serialize` txs with negative call value, and then recover the transaction (where now the BN is the positive variant of the number). This should throw on RLP level, since I don't think RLP supports negative numbers either?

gabrocheleau:
This PR resolves #1513 with the following changes:

- [x] tx package: Throw when constructing a Transaction with a negative BigNumber value (in toBuffer)
- [x] vm package: Throw when passed a negative call value (throws in the Message constructor)

As mentioned in a comment of the above issue, the `toBuffer` method also has the side effect of converting negative BN to positive BN. I decided to throw an error in the case of negative BN in that method, since it seemed to me like that behavior isn't intended. 

I was not sure about disabling negative BNs at the module level, so omitted that for now.


packages/tx/src/baseTransaction.ts
```diff
@@ -83,7 +83,17 @@ export abstract class BaseTransaction<TransactionObject> {
   protected DEFAULT_HARDFORK: string | Hardfork = Hardfork.Istanbul
 
   constructor(txData: TxData | AccessListEIP2930TxData | FeeMarketEIP1559TxData) {
-    const { nonce, gasLimit, to, value, data, v, r, s, type } = txData
+    const { data, nonce, gasLimit, gasPrice, r, s, to, type, v, value } = txData
+    this._validateCannotHaveNegativeBN({
```
ryanio, line 87: since we run all these values through toBuffer do we need this `_validateCannotHaveNegativeBN` function since you already added the negative check to toBuffer?

gabrocheleau: Indeed! removed.

packages/vm/tests/api/runCall.spec.ts
```diff
@@ -262,3 +262,23 @@ tape('Ensure that IDENTITY precompile copies the memory', async (t) => {
 
   t.end()
 })
+
+tape('Throws on negative call value', async (t) => {
```
jochem-brouwer, line 266: Since the check is in `message.ts` (which I think is the right place), I think we should also test on `runCode` and `runTx`.

gabrocheleau: Sure, I'll add a test for each of those before EOD.

packages/tx/src/baseTransaction.ts
```diff
@@ -83,7 +83,7 @@ export abstract class BaseTransaction<TransactionObject> {
   protected DEFAULT_HARDFORK: string | Hardfork = Hardfork.Istanbul
 
   constructor(txData: TxData | AccessListEIP2930TxData | FeeMarketEIP1559TxData) {
-    const { nonce, gasLimit, to, value, data, v, r, s, type } = txData
+    const { data, nonce, gasLimit, r, s, to, type, v, value } = txData
```
ryanio, line 86: hm why reorder these? better to just leave them as is?

ryanio:
i'm seeing uses of both "can not" and "cannot" in this PR, let's just use "cannot"

> Both cannot and can not are perfectly fine, but cannot is far more common and is therefore recommended, especially in any kind of formal writing.
https://www.merriam-webster.com/words-at-play/cannot-vs-can-not-is-there-a-difference

packages/tx/src/baseTransaction.ts
jochem-brouwer: My guess is that this is the linter 😄

gabrocheleau: Ah yeah I have a reordering thing setup, can undo

packages/vm/tests/api/runCall.spec.ts
gabrocheleau: Added for runCode, but wasn't able to add one for runTx since I can't figure out how to build a negative value tx at the moment (left a message in the discord about that)

ryanio:
lgtm! will wait if you want to try to add a negative value tx to runTx (using value=0 and `tx.value.isubn(1)` as jochem suggested)

packages/vm/tests/api/runTx.spec.ts
```diff
@@ -559,3 +560,29 @@ tape('runTx() -> consensus bugs', async (t) => {
     t.end()
   })
 })
+
+tape('runTx: RunTxOptions', (t) => {
```
ryanio, line 564: nit to match above naming

```suggestion
tape('runTx() -> RunTxOptions', (t) => {
```

packages/vm/tests/api/runTx.spec.ts
```diff
@@ -559,3 +560,29 @@ tape('runTx() -> consensus bugs', async (t) => {
...
+      tx.value.isubn(1)
+
+      try {
+        await vm.runTx({
+          tx,
+          skipBalance: true,
+        })
+        t.notOk('should not accept a negative call value')
+      } catch (err: any) {
+        console.log(err)
```
ryanio, line 579:
```suggestion
```

packages/vm/tests/api/runTx.spec.ts
```diff
@@ -559,3 +560,28 @@ tape('runTx() -> consensus bugs', async (t) => {
     t.end()
   })
 })
+
+tape('runTx() -> RunTxOptions', (t) => {
+  t.test('should throw on negative value args', async (t) => {
+    for (const txType of TRANSACTION_TYPES) {
+      const vm = new VM({ common })
```
ryanio, line 567: another tiny nit but you can move this VM init above the for loop scope to make it a bit faster? probably tiny difference but adds up in the suite over time

packages/vm/tests/api/runCall.spec.ts
```diff
@@ -217,7 +217,6 @@ tape('Ensure that Istanbul sstoreCleanRefundEIP2200 gas is applied correctly', a
   }
 
   const result = await vm.runCall(runCallArgs)
-  console.log(result.gasUsed, result.execResult.gasRefund)
   t.equal(result.gasUsed.toNumber(), 5812, 'gas used incorrect')
   t.equal(result.execResult.gasRefund!.toNumber(), 4200, 'gas refund incorrect')
```
ryanio, line 221:
```suggestion
  t.ok(result.gasUsed.eqn(5812), 'gas used correct')
  t.ok(result.execResult.gasRefund!.eqn(4200), 'gas refund correct')
```

I've seen this fixed in multiple PRs already (one of mine, one of holger's) but neither have been merged in yet lol so let's finally fix it here today for merging

packages/vm/tests/api/runTx.spec.ts
gabrocheleau: Yeah good point

packages/vm/tests/api/runCode.spec.ts
```diff
@@ -95,3 +95,22 @@ tape('VM.runCode: interpreter', (t) => {
...
+  t.test('should throw on negative value args', async (st) => {
+    const vm = new VM()
+
+    const runCodeArgs = {
+      value: new BN('-10'),
+    }
+
+    try {
+      await vm.runCode(runCodeArgs)
+      st.notOk('should not accept a negative call value')
```
ryanio, line 109: notOk is a boolean input, should use fail

```suggestion
      st.fail('should not accept a negative call value')
```

packages/vm/tests/api/runTx.spec.ts
```diff
@@ -559,3 +560,27 @@ tape('runTx() -> consensus bugs', async (t) => {
...
+    for (const txType of TRANSACTION_TYPES) {
+      const tx = getTransaction(vm._common, txType.type, true)
+      tx.value.isubn(1)
+
+      try {
+        await vm.runTx({
+          tx,
+          skipBalance: true,
+        })
+        t.notOk('should not accept a negative call value')
```
ryanio, line 576:
```suggestion
        t.fail('should not accept a negative call value')
```

packages/vm/tests/api/runCall.spec.ts
```diff
@@ -318,3 +317,23 @@ tape('Ensure that IDENTITY precompile copies the memory', async (t) => {
...
+  const vm = new VM({ common: common })
+
+  // setup the call arguments
+  const runCallArgs = {
+    value: new BN(-10),
+  }
+
+  try {
+    await vm.runCall(runCallArgs)
+    t.notOk('should not accept a negative call value')
```
ryanio, line 333:
```suggestion
    t.fail('should not accept a negative call value')
```

ryanio (approved):
looks great! can merge after ci passes

PR #1607, opened 2021-12-13
Add a method wrapper to directly extract the recovery bit value from ecsign.

### LINKED ISSUE #1597: Add support for EIP-1559
For EIP-1559, the signature v value is now a simple parity bit ("signature Y parity"), which is either 0 or 1, depending on which point on the elliptic curve should be used.
The `ecsign` does not support EIP-1559 because the it never returns the corresponding v value.

jochem-brouwer:
Closes #1597

In `ecsign` we manually add 27 to the recovery bit. `ecsignRaw` returns the recovery bit and not this semi-artificially created "recovery bit" where we add 27 to the bit.

This also neatly cleanups some of the 'hacky' comments in EIP2930/EIP1559 txs and adds a neat flag to base transaction. (Really cool feature these `protected abstract` boolean flags :smile: )

### DIFF
--- a/packages/util/src/signature.ts
+++ b/packages/util/src/signature.ts
@@ -44,6 +44,11 @@ export function ecsign(msgHash: Buffer, privateKey: Buffer, chainId: any): any {
   return { r, s, v }
 }
 
+export function ecsignRaw(msgHash: Buffer, privateKey: Buffer): ECDSASignature {
+  const { r, s, v } = ecsign(msgHash, privateKey)
+  return { r, s, v: v - 27 }
+}
+
 function calculateSigRecovery(v: BNLike, chainId?: BNLike): BN {
   const vBN = toType(v, TypeOutput.BN)
   if (!chainId) {
--- a/packages/tx/src/baseTransaction.ts
+++ b/packages/tx/src/baseTransaction.ts
@@ -10,6 +10,7 @@ import {
   publicToAddress,
   BNLike,
   bufferToHex,
+  ecsignRaw,
 } from 'ethereumjs-util'
 import {
   TxData,
@@ -51,6 +52,10 @@ export abstract class BaseTransaction<TransactionObject> {
   public readonly s?: BN
 
   public readonly common!: Common
+
+  // This flag switches between using ecsignRaw (v is either 0 or 1) in case it is true,
+  // or ecsign (v is either 27 or 28) in case it is false.
+  protected abstract readonly useRawSign: boolean
 
   protected cache: TransactionCache = {
     hash: undefined,
@@ -307,7 +312,9 @@ export abstract class BaseTransaction<TransactionObject> {
     }
 
     const msgHash = this.getMessageToSign(true)
-    const { v, r, s } = ecsign(msgHash, privateKey)
+    const { v, r, s } = this.useRawSign
+      ? ecsignRaw(msgHash, privateKey)
+      : ecsign(msgHash, privateKey)
     const tx = this._processSignature(v, r, s)
 
     // Hack part 2
--- a/packages/tx/src/eip1559Transaction.ts
+++ b/packages/tx/src/eip1559Transaction.ts
@@ -40,6 +40,8 @@ export default class FeeMarketEIP1559Transaction extends BaseTransaction<FeeMark
 
   public readonly common: Common
 
+  protected useRawSign: boolean = true
+
   /**
    * The default HF if the tx type is active on that HF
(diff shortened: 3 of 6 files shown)


packages/util/src/signature.ts
```diff
@@ -44,6 +44,11 @@ export function ecsign(msgHash: Buffer, privateKey: Buffer, chainId: any): any {
   return { r, s, v }
 }
 
+export function ecsignRaw(msgHash: Buffer, privateKey: Buffer): ECDSASignature {
+  const { r, s, v } = ecsign(msgHash, privateKey)
```
jochem-brouwer, line 48: Small question here, would it be cleaner to make `ecsign` use `ecsignRaw` and then add 27 to it, instead of that we call into `ecsign` and then subtract 27 here? :thinking:

holgerd77: I am very much on shaky ground here, but on first look: wouldn't it be cleanest to directly add a `raw` parameter to the `ecsign` method, and then just keep the 27 out completely? 🤔

I know the function signature doesn't get any better with this with this optional `chainId` parameter. But this might generally need some clean-up in the next breaking release, e.g. by using a dict for options like chainId or raw or generally rethink the overall API and naming a bit.

Also, naming: is `raw` the best name for this? Just openly throwing in the ring.

jochem-brouwer: I have created a new method here for the reason you note here; otherwise you have to call `ecsign(msg, key, undefined, true)`. But, what we could do, is to change the chainId field to a different name and then allow the values `undefined`, `number`, `BNLike` and `boolean`; in case its a boolean, check if the boolean is true and if that's the case then return the raw signed message.

Have no strong opinion here, waiting for other opinions.

holgerd77: Hmm, my fear is a bit that this won't get any better to understand how to use with another method on this.

Let's please not merge this too hastily and keep a bit in discussion how to structure and name things.

jochem-brouwer: Idea from @holgerd77: change the `chainId` optional value to an optional dictionary where we can pass in either a `chainId` field or a `raw` boolean.

holgerd77:
I have thought about this a bit more. This method signature of the current `ecsign` method is indeed so complicated to adopt that it might be the better way to introduce a new method. Or - to be more precise - this "use an additional dict" for the chain ID parameter *would*  work. But: the usage of this function would be suboptimal - to say the least - so 1) not use the chain ID parameter and 2) set a `raw` or whatever named parameter to `true`. And given the facts that a) signature scheme with typed tx signing has generally rather become *super-simple* and b) the type tx scheme is now the one people will mainly use it seems counter-user-friendly if people have to deal with this complicated API when the very most use case is not actually *not* use both of these parameters.

So a separate method is likely a lot better, with some clear code docs that the new method is the one that people likely want to use for typed tx-like usage contexts and also adding additional code docs for the original method describing the two main use cases there (no chain ID: pre-EIP-155, chain ID. EIP-155 legacy txs). 

One additional remark:

1. On the implementation side I would actually find it cleanest if we just do not call into the original `ecsign` method *at all* any more where we would expose us to all the - unnecessary - complexity from the old signature schemes. So I would rather suggest that we simply take over the necessary three lines of code and call into the `ethereum-cryptopgraphy` `ecdsaSign()` method directly.

On the naming side I was first still struggling with the `raw` name but maybe that's really the best choice and would work together with these enhanced code docs on the methods itself. I guess we might then want to rename the original `ecsign` method to something less generic on the next breaking release (eventually really: `ecsignLegacy()` (?)) and then keep (to avoid turbulences on renaming/not renaming) the `raw` name for the other method.

Phew. What a juggle. 😋

We might actually already want to a) deprecate the `ecsign` method *now* and b) introduce a wrapper `ecsignLegacy`.

Does this all make sense?

joa2040:
Guys any update about this PR?

ryanio:
@joa2040 would this new `ecsignRaw` method be useful for you?

We may be removing some of our intermediary crypto functions in util and replace directly from functions from [ethereum-cryptography](https://www.npmjs.com/package/ethereum-cryptography), see [this comment](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1714#issuecomment-1036491485). Could you see if there is something in that package you could use directly, rather than util, and if not, what yo would need? would be great to know.

jochem-brouwer:
I will close this, not sure what the initial issue indicated :thinking:

PR #1613, opened 2021-12-15
eslint: disable no-extra-semi, remove overrides

ryanio:
This PR disables eslint's `no-extra-semi` rule since it gets in the way when a semicolon is needed, like when the line starts with an opening parenthesis. I found that prettier rules already exhaustively cover semi formatting.


gabrocheleau (approved):

ryanio:
thanks!

PR #1616, opened 2021-12-23
common/vm: add EIP-3855, PUSH0 instruction

jochem-brouwer:
https://eips.ethereum.org/EIPS/eip-3855

### DIFF
--- a/packages/vm/src/runCode.ts
+++ b/packages/vm/src/runCode.ts
@@ -83,7 +83,7 @@ export default function runCode(this: VM, opts: RunCodeOpts): Promise<ExecResult
     new Message({
       code: opts.code,
       data: opts.data,
-      gasLimit: opts.gasLimit ?? new BN(0),
+      gasLimit: opts.gasLimit,
       to: opts.address ?? Address.zero(),
       caller: opts.caller,
       value: opts.value,
--- a/packages/vm/src/runCode.ts
+++ b/packages/vm/src/runCode.ts
@@ -83,7 +83,7 @@ export default function runCode(this: VM, opts: RunCodeOpts): Promise<ExecResult
     new Message({
       code: opts.code,
       data: opts.data,
-      gasLimit: opts.gasLimit,
+      gasLimit: opts.gasLimit ?? new BN(0),
       to: opts.address ?? Address.zero(),
       caller: opts.caller,
       value: opts.value,
--- a/packages/common/src/eips/3855.json
+++ b/packages/common/src/eips/3855.json
@@ -1,0 +1,18 @@
+{
+    "name": "EIP-3855",
+    "number": 3855,
+    "comment": "PUSH0 instruction",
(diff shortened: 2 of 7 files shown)


packages/vm/src/runCode.ts
```diff
@@ -83,7 +83,7 @@ export default function runCode(this: VM, opts: RunCodeOpts): Promise<ExecResult
     new Message({
       code: opts.code,
       data: opts.data,
-      gasLimit: opts.gasLimit,
+      gasLimit: opts.gasLimit ?? new BN(0),
```
jochem-brouwer, line 86: Might look a bit out of place in this PR, but I noticed that `gasLimit` is optional, but if no `gasLimit` is provided it is thus `undefined` in the context, which causes the EVM to throw internally [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/0c7bdfcdf5866fc31df2b87508156f3294bbae45/packages/vm/src/evm/evm.ts#L483). (cannot `clone()` the `undefined` `gasLimit`).

ryanio: Would it always go OOG if provided with gasLimit of 0?

If so maybe we should throw with an error, or make it default to e.g. the last block's gasLimit?

Should we consider making it non-optional in a next breaking change release? (We can add to our list)

jochem-brouwer: Ah right I did not give this a lot of thought. I think an error here is cleaner (non-breaking) but I think we should make `gasLimit` mandatory in the breaking release.

There are a few edge cases where it does not go OOG: either if you run no code, or if you run a code which uses 0 gas (`STOP`). All other opcodes which have 0 gas cannot be used AFAIK since these need stack items and you would thus need to push items on the stack, which costs gas.

jochem-brouwer: Actually thinking about this, if we make it error in case that we do not provide gasLimit, then that's breaking as well right?

ryanio: That's true, well if the behavior is correct then we can leave it as be and consider the error for a breaking change release.

ryanio: It didn't work without providing the gasLimit anyway.

We can check if the typedoc for Message has information about providing a gasLimit otherwise default: 0

jochem-brouwer: I have removed this change; if we default it to 0 then the behavior is slightly different than if we do not provide a gasLimit; in the case that we just send eth OR we run the `00` code, the code will run and finish if we default it to 0, while if we don't do this it would throw, so this is somewhat a breaking change. (I also should not have added this whole thing since this is out of scope of this PR). Will add this thing to the tiny fixes / breaking issue.

ryanio (approved):
lgtm!

jochem-brouwer:
Thanks @ryanio :smile:

PR #1617, opened 2021-12-28
vm/tests: add support for test networks with activated EIPs

jochem-brouwer:
This PR adds support for `ethereum/tests` forks where there are added EIPs; these are usually experimental and it is not clear if they will ever be in an actual fork, but they have tests anyways.

To test;

```cd packages/ethereum-tests
git checkout push0
npm run test:blockchain -- --fork=London+3855
```

Results in 12 passing tests, so looks like we pass the (draft) tests for EIP3855 as well :smile:

### DIFF
--- a/packages/vm/tests/tester/config.ts
+++ b/packages/vm/tests/tester/config.ts
@@ -258,7 +258,12 @@ export function getTestDirs(network: string, testType: string) {
  * @param {String} network - the network field of a test
  * @returns {Common} the Common which should be used
  */
-export function getCommon(network: string) {
+export function getCommon(targetNetwork: string) {
+  let network = targetNetwork
+  if (network.includes('+')) {
+    const index = network.indexOf('+')
+    network = network.slice(0, index)
+  }
   const networkLowercase = network.toLowerCase()
   if (normalHardforks.map((str) => str.toLowerCase()).includes(networkLowercase)) {
     // normal hard fork, return the common with this hard fork
@@ -289,13 +294,20 @@ export function getCommon(network: string) {
         })
       }
     }
-    return Common.forCustomChain(
+    console.log(network)
+    console.log(hfName)
+    const common = Common.forCustomChain(
       'mainnet',
       {
         hardforks: testHardforks,
       },
       hfName
     )
+    const eips = targetNetwork.match(/(?<=\+)(.\d+)/g)
+    if (eips) {
+      common.setEIPs(eips.map((e: string) => parseInt(e)))
+    }
+    return common
   } else {
     // this is not a "default fork" network, but it is a "transition" network. we will test the VM if it transitions the right way
     const transitionForks =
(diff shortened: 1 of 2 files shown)


ryanio:
There are a few leftover console.logs

jochem-brouwer:
Don't know why I did not notice that when linting :thinking: Removed them.

ryanio:
Thanks! I think we have the no-console line overridden to be ignored in tester.ts since we log some other important output.

acolytec3 (dismissed):
I made one change to use `common.custom` instead of the deprecated `common.forCustomChain` but looks great otherwise!  I tested with a couple of different hardfork combinations and it parsed them all correctly (and threw an error if trying to set an unsupported hardfork - as expected).

packages/vm/tests/tester/index.ts
```diff
@@ -91,6 +91,25 @@ async function runTests() {
   runnerArgs.value = argv.value // GeneralStateTests
   runnerArgs.debug = argv.debug // BlockchainTests
 
+  /**
+   * Edit the forkConfig string to ensure it works with RegEx (escape + characters)
+   */
+
```
ryanio, line 97: extra newline
```suggestion
```

packages/vm/tests/tester/config.ts
```diff
@@ -258,7 +258,12 @@ export function getTestDirs(network: string, testType: string) {
  * @param {String} network - the network field of a test
  * @returns {Common} the Common which should be used
  */
-export function getCommon(network: string) {
+export function getCommon(targetNetwork: string) {
+  let network = targetNetwork
+  if (network.includes('+')) {
```
ryanio, line 263: could you include some explanation comment here what the `+` symbol represents? (or maybe can include in the typedoc)

packages/vm/tests/tester/index.ts
```diff
@@ -91,6 +91,25 @@ async function runTests() {
   runnerArgs.value = argv.value // GeneralStateTests
   runnerArgs.debug = argv.debug // BlockchainTests
 
+  /**
+   * Edit the forkConfig string to ensure it works with RegEx (escape + characters)
```
ryanio, line 95: suggestion for clarity
```suggestion
   * Modify the forkConfig string to ensure it works with RegEx (escape `+` characters)
```

holgerd77:
Can we also add 1-2 lines on how to use this in the developer README somewhere [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/DEVELOPER.md#running-different-test-types)?

Also ok of course to somewhat restructure along if the current structure of the section is generally not so optimal or something.

holgerd77:
Can this be updated at some point?

jochem-brouwer:
I will update this before the end of this month.

jochem-brouwer:
Updated docs, thanks for the reviews :smile:

holgerd77 (approved):
LGTM

PR #1619, opened 2021-12-29
VM, Tx, Common: Implement EIP3860 "limit and meter initcode"

jochem-brouwer:
Implements https://eips.ethereum.org/EIPS/eip-3860

Requires #1617

Note: EIP is in review but it is likely that some details get changed.

To test:

```
npm run build --workspaces
cd ./packages/ethereum-tests
git remote add ewasm git@github.com:ewasm/tests.git
git checkout ewasm/eip-3860
cd ../vm
npm run test:blockchain -- --fork=London+3860
```

Another note: the current tests do not match the EIP spec so at this point we cannot test it because the tests are invalid.


jochem-brouwer:
```
Errors thrown in file: GeneralStateTests/stCodeSizeLimit/create2InitCodeSizeLimit.json test: create2InitCodeSizeLimit_d0g0v0_London+3860:
	Error: invalid receiptTrie (vm hf=london -> block number=1 hash=0xbfbd043495ac98f0a3a03edcce1cafab3a3d2c752cdff090465f772f549fd102 hf=london baseFeePerGas=10 txs=1 uncles=0)
	correct last header block
Errors thrown in file: GeneralStateTests/stCodeSizeLimit/create2InitCodeSizeLimit.json test: create2InitCodeSizeLimit_d1g0v0_London+3860:
	Error: invalid receiptTrie (vm hf=london -> block number=1 hash=0x5bb4b3bf60d0dc6ea703640941bdbcc231028c05256edfe836655007821165b9 hf=london baseFeePerGas=10 txs=1 uncles=0)
	correct last header block
Errors thrown in file: GeneralStateTests/stCodeSizeLimit/createInitCodeSizeLimit.json test: createInitCodeSizeLimit_d0g0v0_London+3860:
	Error: invalid receiptTrie (vm hf=london -> block number=1 hash=0xe5437b836f3d7ace37d5716bac8f811e7ab55d4e6851e2e28f12fa4dd1460f9a hf=london baseFeePerGas=10 txs=1 uncles=0)
	correct last header block
Errors thrown in file: GeneralStateTests/stCodeSizeLimit/createInitCodeSizeLimit.json test: createInitCodeSizeLimit_d1g0v0_London+3860:
	Error: invalid receiptTrie (vm hf=london -> block number=1 hash=0x94ab4d95df530d9fad183d554aab9628a4314ba0511761ec3d7071127c229fba hf=london baseFeePerGas=10 txs=1 uncles=0)
	correct last header block

1..22
# tests 22
# pass  14
# fail  8
```

holgerd77:
Some reference: Transaction tests for EIP-3860 PR https://github.com/ethereum/tests/pull/1012

jochem-brouwer:
Ah those are new, thanks @holgerd77 :smile: Seems like we can test this now!

jochem-brouwer:
We should also require #1701 here due to the added TransactionTests.

Will continue here to see if I can figure out what's causing the tests to fail.

jochem-brouwer:
Updated to latest version, tests pass on `London+3860`, on this tests commit: https://github.com/ethereum/tests/commit/9a19e31ef2cd40411c02ab0da6503daf296334ae

holgerd77:
So would you regard this as ready for review?

jochem-brouwer:
We could do a preliminary review, the problem is that the EIP will most likely get changed (in particular gas costs), so I don't think we should merge it already.

holgerd77:
@jochem-brouwer ah ok, sorry, wasn't aware of that, then let's wait.

Just had a look: Go Ethereum PR is also still in a `DRAFT` state: https://github.com/ethereum/go-ethereum/pull/23847

axic:
> the problem is that the EIP will most likely get changed (in particular gas costs)

@jochem-brouwer which part do you mean? Though in your implementation it seems everything relevant is in the config file (limits and gas costs), so it shouldn't be a big deal.

jochem-brouwer:
@axic This comment: https://github.com/ethereum/tests/pull/990#issuecomment-1033665320

The EIP spec (text) would thus get updated, but it seems that the tests already match what "the probable direction EIP spec will go", so in that case nothing would change in this implementation.

holgerd77:
@jochem-brouwer I guess since @axic is one of the EIP authors we can then relatively safely merge here?

holgerd77:
(or: review and merge)

jochem-brouwer:
I just re-read the EIP and noticed that there is a check which I have not yet added (check that a contract creation tx has a calldata payload `<= maxInitCodeSize`. These are invalid in the same way as underpricing txs (sending eth to someone and having `gasLimit` less than 21000). I should also add TransactionTests (this one is also not yet merged in `ethereum/tests`) in the transaction tests and ensure they pass as well.

axic:
I think the EIP and tests should be in sync now, and the test suite will only be merged once geth merges the code?

I do appreciate your work here quite a bit, as any second implementation can unearth questions, just like it did for you @jochem-brouwer.

jochem-brouwer:
Hi @axic, the ethereum/tests PR is currently *not* in sync with the EIP spec.

This line from the EIP: 

>  In these three cases the entire execution is exceptionally aborted.

In those cases, all gas in the frame is consumed, but in ethereum/tests only the gas cost of `CREATE`/`CREATE2` will be consumed and 0 will be put on the stack.

I also agree with you on the second implementation thing - in fact, I think this should be considered for all EIPs which change execution logic. Second implementations will be a "check" to see if the EIP spec is not ambigous and if you can thus use the self-contained EIP to implement it. I will also start implementing other EIPs which are considered for Shanghai. It is both very fun to implement those, and it is also healthy for the ecosystem to ensure that the EIPs are correct 😄

axic:
Pinging @gumb0 about this 😬

gumb0:
> This line from the EIP:
> 
> > In these three cases the entire execution is exceptionally aborted.
> 
> In those cases, all gas in the frame is consumed, but in ethereum/tests only the gas cost of `CREATE`/`CREATE2` will be consumed and 0 will be put on the stack.

Wait, that is the line from Rationale, it refers to already existing (non-creation-specific) failure cases, and says new failures from this EIP _don't_ do this.

jochem-brouwer:
Ah right, upon re-reading I agree that it indeed matches the current EIP spec. I do think that the addition of these "3 checks" is a bit confusing though (this lead me to believe I should spend all gas). For clarity it might be better to add a line like "in case the initcode limit is violated when trying to execute CREATE/CREATE2, then only the upfront gas is consumed: this is thus the base cost + memory expansion cost + initcode cost (+ hashing cost (CREATE2)), and 0 is put on the stack". I think the addition of mentioning the cases where all gas is consumed is not necessary and would only lead to confusion.

But it might also be my poor reading ability to lead to implement the "wrong" version 😅 

@axic @gumb0

jochem-brouwer:
To test TransactionTests:

``` 
git remote add ipsilon git@github.com:ipsilon/tests.git
git fetch ipsilon
git checkout ipsilon/eip-3860-transactions
cd ./packages/tx
npm run test -- -t
```

(They pass)

packages/tx/src/legacyTransaction.ts
```diff
@@ -135,6 +135,16 @@ export default class Transaction extends BaseTransaction<Transaction> {
       }
     }
 
+    if (this.common.isActivatedEIP(3860)) {
```
jochem-brouwer, line 138: This block is in all three transaction types, the reason is that we do not have access to common in baseTransaction. Adding it in baseTransaction is somewhat hard as especially legacyTransaction has some extra logic to extract chainId.

ryanio: hm. i don't like how we have to keep duplicating code across the txs classes like this. but thanks for the reasoning why.

would be much better if we always write the code just once. what if we move it to a method in`util.ts` called e.g. `checkMaxInitCodeSize()`, then we can just keep the if isActivatedEIP(3860) and call the method inside.

ryanio: other than that, PR looks great! I'm not sure why the errors don't have any coverage on them though (I'm seeing codecov warnings), i assume the added TransactionTests should cover that. if not maybe we should add a few explicit tests for coverage.

jochem-brouwer: Yep, great suggestion moving it to `util`! I also did not like it.

jochem-brouwer: Addressed this

packages/vm/src/evm/evm.ts
```diff
@@ -298,6 +298,20 @@ export default class EVM {
...
 
+    if (this._vm._common.isActivatedEIP(3860)) {
+      if (message.data.length > this._vm._common.param('vm', 'maxInitCodeSize')) {
+        return {
+          gasUsed: message.gasLimit,
+          createdAddress: message.to,
+          execResult: {
+            returnValue: Buffer.alloc(0),
+            exceptionError: new VmError(ERROR.INITCODE_SIZE_VIOLATION),
+            gasUsed: message.gasLimit,
```
acolytec3, line 309: When I read the EIP, it says that if a "create transaction exceeds `MAX_INITCODE_SIZE`, transaction is invalid."  Does that mean that we do an exceptional abort and consume all gas or that the transaction is just rejected and not included in the block?

acolytec3: I went ahead and added a new EIP-3860 test file to the `api/EIPs` section and added a real basic test to trigger this error so we can bump test coverage.

jochem-brouwer: It is the latter, I remember having the same question and it is answered somewhere, but I cannot find it again 😅 So, the block is invalid if you include such transaction and nodes will reject that broadcasted block.

holgerd77 (approved):
Latest comments from @ryanio have been addressed, will approve and merge.

Thanks Jochem! 🙂 Cool! 👍

holgerd77:
Hmm, codecov is not reporting. 🤔 Will nevertheless merge here.

PR #1620, opened 2021-12-30
tx: input value testing

### LINKED ISSUE #1610: Tx: More systematic input value and constructor tests
The different possible input values (e.g. if a [BNLike](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/src/types.ts#L9) type is used) on object instantiation within our libraries are often not so systematically tested, which is potentially dangerous. A more structured approach would be particularly important in the context of the tx library.

I have looked over the test code a bit and it is a bit tricky. Often test data is coming from some pre-given sources (JSON files or similar) and it wouldn't be that much of a natural fit to integrate on the sideline when other things are tested (e.g. in files like `base.spec.ts`). Existing test files for the different tx type are also already very much bloated and packed (side note: these could need some clean-up and better/more consistent structuring as well) and adding more input value tests would bloat these files even more.

The following approach seems the most pragmatic to me atm:

1. We add a dedicated new test file `inputValues.spec.ts` (or similarly named)

2. We add some appropriate and dedicated input value data structures with explicit values for a) the `base` values and then b) the additional legacy, eip1559,... values

3. For a dedicated parameter all possible value types (BN, number, string) should be encompassed.

A fitting data structure could e.g. roughly look like this:

```typescript
const baseValues = {
  nonce: [ new BN(5), 5, '0x5',... ],
  from: [ Buffer.from('12345...', 'hex'), '0x12345', ... ]
}

const legacyValues = { // only the missing values here
  gasPrice: [ new BN(100000), 100000, ... ]
}
```

And so on. Test execution should then iterate over the data structure from above (respectively: base + legacy, base + eip1559,...), do instantiations with all (most/some) value combinations and e.g. compare the resulting hashes of instantiated txs or similar (note that the data structure from above should be constructed towards the goal that it actually get's easy to do these kind of iterations, not sure if this is already fitting, array instead of dict might be easier e.g., not sure).

Eventually similar separate data structures are needed for the different constructors. At the end these tests should *also* run on the all various direct static constructors and also on the factory constructors.

On a second iteration the data structures from above might be reused for some edge case value testing (zero for nonce e.g.), this shouldn't be too extensive and should be taken a bit pragmatically, but would be nice if some/most significant cases are touched here.

And on a third iteration there might also be some reusage to test on invalid values.

Also some note: eventually (likely) some existing tests from the existing files fitting within this scope should also be moved into this file and - if it makes sense - integrated into these expanded test strucure.

Does this all makes sense? 🤔

gabrocheleau:
Resolves #1610


packages/tx/test/inputValue.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+function mulberry32(seed: number) {
+  let t = (seed += 0x6d2b79f5)
+  t = Math.imul(t ^ (t >>> 15), t | 1)
+  t ^= t + Math.imul(t ^ (t >>> 7), t | 61)
+  return ((t ^ (t >>> 14)) >>> 0) / 4294967296
+}
+
+function getRandomSubarray(array: any[], size: number) {
+  const shuffled = array.slice(0),
+    seed = 1559
```
acolytec3, line 62: This is probably just me not being a Javascript guru but this syntax... 
```
 const shuffled = array.slice(0),
    seed = 1
```
threw me for a loop. Is the comma operator here just applying `const` to the `seed = 1` expression?  Reading on mozilla dev docs here didn't make it any more obvious to me - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comma_Operator

packages/tx/test/inputValue.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+  let index: number,
+    i = array.length,
+    temp: any[]
+  while (i--) {
```
acolytec3, line 66: Here again, more "I'm not a javascript guru" questions but presumably `while (i--)` is just shorthand for `while (i === 0) { i--} `?  I dropped some console logs in the test and that seems to be what's happening but just wanted to confirm.

gabrocheleau: It's a shorthand for 
```
while (i > 0) {
 i = i - 1;
}
```

packages/tx/test/inputValue.spec.ts
gabrocheleau: Yeah, it's a shorthand way of declaring multiple variables, but I agree it's not too readable and commonly used, I'll revert to an explicit declaration per line.

packages/tx/test/inputValue.spec.ts
ryanio: oh nice didn't know this one. maybe you can also do one declaration per line for the above let variables as well. i wonder if there's a eslint rule to prefer one line per declaration 

found it: ^_^ https://eslint.org/docs/rules/one-var-declaration-per-line

holgerd77: I think we should avoid such abbreviations, we have already two people stumbling upon this here respectively not knowing how this would behave (with me it would be three 😋) and this just makes code less approachable if 50% of developers do not know how to read the code.

(sometimes a bit difficult to draw the lines but I would say in this case it's relatively clear)

gabrocheleau: Good point, will update! :)

holgerd77:
What's the status of this? Can we have this fixed?

gabrocheleau:
> What's the status of this? Can we have this fixed?

Yes, I have some time tomorrow evening/Saturday and will get that PR finished then.

acolytec3 (approved):
Looks great!  I dumped some of the `txData` values to the console for both legacy and EIP1559 transactions and am seeing all the different data types represented and no issues testing.  I also ran it several times just to get a variety of different combos and its working great.  Nice work!

packages/tx/test/inputValue.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+    const options = { ...baseTxValues, ...legacyTxValues, type: '0' }
+    const legacyTxData = generateCombinations({
+      options,
+    })
+    const expectedHash = Transaction.fromTxData(legacyTxData[0]).hash()
+    const randomSample = getRandomSubarray(legacyTxData, 100)
+    for (const txData of randomSample) {
+      const tx = Transaction.fromTxData(txData, { common })
+
+      st.deepEqual(tx.hash(), expectedHash), 'correct tx hash'
```
holgerd77, line 113: This is a fun one (analog below): the test output string is actually outside the function call and is not taken but text output is "should be deeply equivalent". 😅 

Totally love this, that even the syntax here seems to work (what kind of syntax is that actually?) Did some tests: `'test', 'toast'` actually throws with "Left side of comma operator is unused and has no side effects." but `let t = function() {}` followed by `t(), 'test'` actually works and outputs the string. Lol.

Anyhow: maybe you can update! 🙂  Is there a (simple) way to get some of the test data into the output message too? Or is this too complicated respectively get's too verbose with the current code structure? Would be nice to identify single test cases, but also no absolute must.

Otherwise PR looks really great! 👍

holgerd77 (changes requested):
Some nit, otherwise looks good, see other comment.

gabrocheleau:
> Some nit, otherwise looks good, see other comment.

Sure thanks for the review! Will check that out and address later.

packages/tx/test/inputValue.spec.ts
gabrocheleau: Ah! Nice catch, that syntax is really weird indeed, now fixed!

I've added the `tx.hash()` to the output message, this at least allows us to differentiate between transaction objects (e.g. the legacy case vs. the eip1559 case). I think it would be too verbose to display the tx args themselves.

holgerd77: Yeah, that's a good compromise, thanks!

holgerd77 (approved):
Nice, cool, thanks! 😄

Will merge.

PR #1621, opened 2022-01-04
Add fallback for node constants

acolytec3:
Fix broken dependency in webpack where node's `constants` package isn't being polyfilled by webpack 5.


ryanio (approved):
thanks!

PR #1624, opened 2022-01-10
ci: upgrade node to v16

ryanio:
This PR upgrades the ci to run default on Node v16 since v12 will be EOL in April.

Node v16 includes npm v7 by default so we can remove the extra install step.

We still test Node v12 on a nightly basis thanks to @evertonfraga's `node-versions.yml` workflow that pulls and runs all active and LTS versions.

### DIFF
--- a/.github/workflows/block-build.yml
+++ b/.github/workflows/block-build.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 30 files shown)


ryanio:
It seems that the expected tests on the PR checks is tied to the name with 12, so I will remember to update the settings after this PR is merged

jochem-brouwer (approved):
LGTM!

PR #1625, opened 2022-01-10
Update ethereum/tests to v10.3

acolytec3:
Updates ethereum/tests to v10.3

### DIFF
--- a/.github/workflows/block-build.yml
+++ b/.github/workflows/block-build.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [12]
+        node-version: [16]
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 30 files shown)


acolytec3:
@jochem-brouwer @holgerd77 @ryanio I updated the tests to the latest release and it looks like everything is passing without needing to make any changes to any libraries (yay - given how much we work went into the bump to v10.2).

I looked through the release notes and the main thing that jumped out to me about the changes is they added a `sender` (address) field to the transaction section of the State tests to allow the public key to be recovered from the private key.  Is there any value in updating something in the state test runner to check this?

jochem-brouwer:
There isn't, if we would report the wrong sender we would have a gigantic bug in `tx` and we would also fail all tests which have transactions originating from said sender which have either gasPrice nonzero or a value nonzero.

jochem-brouwer (approved):
LGTM, great that these pass.

holgerd77:
Oh, that was an easy one, nice! 🙂

holgerd77:
Short note: did a manual lookup on one of the CI runs here and checked for a new test file name (`jumpToPush.json`) to be present and the tests being run, can confirm that this is running on the new tests.

(I was a bit suspicious seeing everything pass since there is really a substantial amount of new test cases, some had been explicitly adressed already in work before though).

Anyhow: cool. 😄 👍

PR #1626, opened 2022-01-10
rlp: add to monorepo

ryanio:
This PR adds [ethereumjs/rlp](https://github.com/ethereumjs/rlp) to the monorepo.

I ran a modified version of ev's [monorepify.sh](https://github.com/evertonfraga/monorepify-ethereumjs/blob/master/monorepify.sh) to preserve git history and tags (thanks ev!)

I also converted rlp's tests from mocha to tape so we can be consistent across the whole monorepo.


packages/rlp/test/official.spec.ts
```diff
@@ -148,9 +148,6 @@ const gethCases = [
 
   { input: 'C3808005', value: ['', '', '05'] },
   { input: 'C50183040404', value: ['01', '040404'] },
-  {
-    input: 'c330f9c030f93030ce3030303030303030bd303030303030',
```
jochem-brouwer: This one is gone?

packages/rlp/CHANGELOG.md
```diff
@@ -55,7 +55,7 @@ PRs included in this release:
 
 Source files from the `src` folder are now included in the distribution build, see PR [#97](https://github.com/ethereumjs/rlp/pull/97). This allows for a better debugging experience in debug tools like Chrome DevTools by having working source map references to the original sources available for inspection.
 
-[v2.2.7]: https://github.com/ethereumjs/rlp/compare/v2.2.6...v2.2.7
+[v2.2.7]: https://github.com/ethereumjs/ethereumjs-monorepo/compare/rlp@2.2.7...rlp@2.2.6
```
jochem-brouwer, line 58: These URLs are not accessible - I think for legacy releases (so the ones not in the monorepo) they should link back to the (soon-to-be) archived repo of RLP.

packages/rlp/test/official.spec.ts
ryanio: Its value was missing, I went looking for it but it seems to have always have been missing.

I got an error trying to decode it:

```
❯ node bin/rlp decode 0xc330f9c030f93030ce3030303030303030bd303030303030
Error decoding RLP: invalid RLP (safeSlice): end slice of Uint8Array out-of-bounds
```

packages/rlp/CHANGELOG.md
ryanio: I haven't pushed the tags yet, I wasn't sure if I should do that now or after merging

ryanio: this one is in the wrong order though :)

jochem-brouwer (commented):
I checked from https://github.com/ethereumjs/ethereumjs-monorepo/pull/1626/commits/e4b371dcfdf19c505fd23e53bbc2a3d73ea6f7fe, this is the current commit which we have at master at https://github.com/ethereumjs/rlp currently. I checked the diffs since the visualization does not work.

Have two comments. A general comments outside the scope of this PR

- [x] Use codecov
- [x] Add codecov stickers
- [x] Add RLP issue label

EDIT: And right as general comment: do not squash-and-merge but instead import all commits as normal, since we want to preserve commits history in this case (of the old repo).

acolytec3 (approved):
Looks good.  Tests are all passing and the CLI tool is working great locally.

holgerd77:
🎉

Great stuff, this really feels like the right decision (and actually makes me more excited than I anticipated 😋). RLP is (still) such a central part of the Ethereum protocol stack and our library version could really need a bit more love and I am confident that it will get that now somewhat more sustainably with this new placement.

Small nit: main [README](https://github.com/ethereumjs/ethereumjs-monorepo) still lacks an integration/listing.

Issue #1575, opened 2021-11-17
VM: Optimize (Replace) Jump Analysis Algorithm


holgerd77:
The EVM uses the `JUMPDEST` and `BEGINSUB` opcodes to mark jump or subroutine destinations. Control flow is then directed to these potential destinations with the `JUMP`, `JUMPI` and `JUMPSUB` opcodes which are respectively setting the program counter, see the respective [opcode implementations in functions.ts](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/evm/opcodes/functions.ts).

Our VM determines which jump destinations are valid in the [Interpreter._getValidJumpDests()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/evm/interpreter.ts#L240) function. Martin from the geth team pointed out that our implementation is generally suboptimal from a performance perspective and would become really slow with a lot of jump destinations, since we put all jump dests into a large array.

Go-Ethereum has its own - more sophisticated - algorithm for a jump analysis which Martin described. Will repost here:

---

Look at https://github.com/ethereum/go-ethereum/blob/master/core/vm/analysis.go#L81. It has several tricks

- The output is a bitmap, each bit telling if the op is part of `data` or part of `code`. 

Thus, when checking if a jump is valid, there are two parts: is it code and is the code actuall `JUMPDEST`. So the opcode check is deferred until the jump is made. 

By doing the bitmap this way, we improve the speed. Let's consider your algo. There are two things which make it slow. 
1. The speed of the read 
2. The speed of the write

For speed of read -- there is a worst-case on "everything except PUSH". For any non-push, it will only progress one byte at a time. 
For speed of write -- there is a worst-case on JUMPDEST and BEGINSUB. For those two, it will push something on every byte. 

So the data-filled-with-jumpdest neatly hits both worst-cases. Now, if you consider my algo instead

- The speed of read is also worst on non-push. 
- The speed of write is worst on push -- because then we need to write 1 into the bitmaps. However, e.g. on PUSH32, we can write 32 bits at the same time. 

So with my algo, there is no singular worst-case which combines the two. Most likely PUSH1 is worst, since it "only" progresses 2 byte at a time, and fills every second bit with `1`. 

Also, your algo is slow because  you alloc -- every jumpdest forces a vector to be extended. In my algo, we pre-allocate a bitmap of 1/8th the size of the code, and then just flips the bits.

[...]

PR #1629, opened 2022-01-11
VM: Jump dest analysis enhancements

acolytec3:
- Switch valid Jumps storage from multiple `number[]` to a single Uint8Array of predefined length
- Only do jump destination analysis if `JUMP`/`JUMPI`/`JUMPSUB` opcode encountered

Should finish off #1575


ryanio:
just had one comment but otherwise this lgtm! thanks for taking it on so quickly

let's wait for one more approval from @jochem-brouwer or @holgerd77 since this is critical code (I think we should expect for vm internals)

acolytec3:
> just had one comment but otherwise this lgtm! thanks for taking it on so quickly
> 
What's the comment?  :-)

ryanio (changes requested):

packages/vm/src/evm/interpreter.ts
```diff
@@ -109,6 +98,15 @@ export default class Interpreter {
     // Iterate through the given ops until something breaks or we hit STOP
     while (this._runState.programCounter < this._runState.code.length) {
       const opCode = this._runState.code[this._runState.programCounter]
+
+      if (
+        (opCode === 0x56 || opCode === 0x57 || opCode === 0x5e) &&
+        this._runState.validJumps.length === 0
```
jochem-brouwer, line 104: I'm not sure in what order JS evaluates an `A && B` statement, but if A is first evaluated and then B then I'd say swap A and B in this case, or probably even better cache a boolean which saves if we have ran JUMPDEST analysis.

packages/vm/src/evm/interpreter.ts
```diff
@@ -261,27 +259,27 @@ export default class Interpreter {
   }
 
   // Returns all valid jump and jumpsub destinations.
-  _getValidJumpDests(code: Buffer): JumpDests {
-    const jumps = []
-    const jumpSubs = []
+  _getValidJumpDests(code: Buffer) {
+    const jumps = new Uint8Array(code.length).fill(0)
```
jochem-brouwer, line 263: This creates a byte array while Geth uses a bitmap (so 1/8 length of the array is used there). I'm not sure if this is faster or practical in JS but worth mentinoing here.

packages/vm/src/evm/interpreter.ts
```diff
@@ -261,27 +259,27 @@ export default class Interpreter {
...
 
     for (let i = 0; i < code.length; i++) {
       const curOpCode = this.lookupOpInfo(code[i]).name
 
-      // no destinations into the middle of PUSH
+      // skip over PUSH0-32 since no jump destinations in the middle of a push block
       if (curOpCode === 'PUSH') {
```
jochem-brouwer, line 269: This still does string comparison which per #1575 should be removed. 
```
    for (let i = 0; i < code.length; i++) {
      const curOpCode = code[i]

      // no destinations into the middle of PUSH
      if (curOpCode >= 0x60 && curOpCode <= 0x7f){
        i += curOpCode - 0x5f
      }
```

packages/vm/src/evm/interpreter.ts
```diff
@@ -261,27 +259,27 @@ export default class Interpreter {
...
         i += code[i] - 0x5f
       }
 
+      // Define a JUMPDEST as a 1 in the valid jumps array
       if (curOpCode === 'JUMPDEST') {
```
jochem-brouwer, line 274: String comparison

packages/vm/src/evm/interpreter.ts
```diff
@@ -261,27 +259,27 @@ export default class Interpreter {
...
-        jumps.push(i)
+        jumps[i] = 1
       }
 
+      // Define a BEGINSUB as a 2 in the valid jumps array
       if (curOpCode === 'BEGINSUB') {
```
jochem-brouwer, line 279: String comparison

packages/vm/src/evm/interpreter.ts
jochem-brouwer: In fact I think we should use a cache, there is an edge case where there is a JUMP but there are no JUMPDESTs, in that case `validJumps.length` remains 0.

packages/vm/src/evm/interpreter.ts
```diff
@@ -21,8 +21,7 @@ export interface RunState {
   stack: Stack
   returnStack: Stack
   code: Buffer
-  validJumps: number[]
-  validJumpSubs: number[]
+  validJumps: Uint8Array
```
jochem-brouwer, line 24: Can we add a comment here what this Uint8Array means? It should describe that opcode at pc `index` has value 0 (default), 1 (jumpdest), 2 (beginsub)

packages/vm/src/evm/interpreter.ts
acolytec3: I don't think the bitmaps thing is relevant to us.  Geth is flipping bits for every push (or flipping a bunch at once) where as we don't do any writes at all for `PUSH` but just skip over however many `PUSH`s there are.

packages/vm/src/evm/interpreter.ts
acolytec3: Makes sense.  I've added a new property to `runState` to track this.

packages/vm/src/evm/interpreter.ts
acolytec3: Good call, fixed.

packages/vm/src/evm/interpreter.ts
acolytec3: Does the comment I added align with your thoughts?

packages/vm/src/evm/interpreter.ts
```diff
@@ -98,14 +100,13 @@ export default class Interpreter {
     // Iterate through the given ops until something breaks or we hit STOP
     while (this._runState.programCounter < this._runState.code.length) {
       const opCode = this._runState.code[this._runState.programCounter]
-
       if (
-        (opCode === 0x56 || opCode === 0x57 || opCode === 0x5e) &&
-        this._runState.validJumps.length === 0
+        !this._runState.jumpAnalysisDone &&
```
jochem-brouwer, line 104: Super nit but maybe we want to name this flag `shouldDoJumpAnalysis` which we initialize to `true`, this saves a NOT operation for each opcode.

packages/vm/src/evm/interpreter.ts
jochem-brouwer: Agreed

packages/vm/src/evm/interpreter.ts
jochem-brouwer: Yep 👍 😄

jochem-brouwer (approved):
LGTM!

ryanio (approved):

jochem-brouwer (approved):

jochem-brouwer:
🎉

PR #1630, opened 2022-01-11
rlp: update monorepo usage to v3

ryanio:
This PR integrates rlp v3 into the monorepo to test the new version.

WIP, will continue on this tomorrow.

Please note if you see anything that can be improved or simplified.

For now I did not spend any time on refactoring the actual internals to fewer buffers but instead just wrapped RLP.encode/decode with Buffers so they behave like before. as a next step we could use the Uint8Array values more natively, in some cases not needing to convert to buffer at all, but that can be saved for work in future PRs.

If all goes well, after this PR is passing, I can release rlp v3.0.0 on npm.


ryanio:
Note: this PR would require a ethereumjs-util v8 release with the rlp v2 to v3 bump. This would also be a good time to move to the updated scoped naming of @ethereumjs/util (we can do in a separate util release PR)

I also noticed there is a script to migrate npm releases in ev's monorepify repository so we can use that to add the prior versions to the new scoped name. Then we can deprecate the latest ethereumjs-util release to point people to the new scoped name for new version updates (as we've done with the past repos e.g. ethereumjs-block)

ryanio:
I also ran into this npm error locally trying to bump ethereumjs-util to v8 and it not picking up the local changes, instead erroring out when looking on npm: https://github.com/npm/cli/issues/3637

So I'm going to have to figure out a workaround for us, locally to fix I just globally installed npm 7.20.2 and that seemed to have fixed it, so that could work, maybe as a preinstall script as a temporary workaround.

holgerd77:
Ah, we shouldn't merge this into `master` respectively not do a new breaking Util release for this one change. `Util` is extremely laggy in adoption and we should only do major version bumps there for a very good reason.

So I would suggest that we rather do this work against `develop` and then take the occasion to continue work on a new Util version which we can then finally release together with new major versions of the other libraries (realistically still some months away).

This Util version should then e.g. also include the new noble crypto versions I guess (respectively the new audited `ethereum-cryptography` version), a more consistent BigInt support, and eventually other things. Do not have the complete picture here yet, this would need some more and broader release planning and some overall strategy we develop from the thoughts around that.

holgerd77:
Respectively - another angle (independent from the aspect from above): is there any value to actually still *do* this reexport of `RLP` in Util or should we generally deprecate this and remove in the next Util version and use `RLP` directly? Does this have any valuable unification effect?

ryanio:
@holgerd77 that's a good idea, I think it would be fine to move away from the re-export, so I will update this PR to do that and we can evaluate from there. It will solve the util rlp breaking change problem as well (well, the breaking change would then be that rlp is removed as a re-export entirely)

paulmillr:
What's the plan here? What are the blockers for v3.0 release? Can't wait to upgrade to it.

holgerd77:
> What's the plan here? What are the blockers for v3.0 release? Can't wait to upgrade to it.

We want to have our CI passing here at least once, that's the minimum goal. 🙂 

Ryan, do you want to reopen or retarget this PR towards `develop`, then we can directly merge in there and then calmly integrate this into the upstream libraries and eventually `Util` along the way to the breaking releases? 🤔

holgerd77:
(@paulmillr so couple of days more patience please 😋 , but release is really not that far away any more now.)

ryanio:
yes @paulmillr thanks for your patience, juggling a few things right now, but we should be very close on this, trying to finish today.

i'd like to:
- get CI passing (just a few last things i have to look into, almost everything is passing now)
  - update this PR to target `develop` for the monorepo integration
- open a new PR to update the rlp package changes from this PR on master
  - we would then be ready for rlp v3 release from master branch

then we can merge the rlp changes from `develop` package-by-package as we do major version releases for our packages. the reason we have to do it like this is to include rlp v3 we include bigint support so it conflicts with our current node es6 and browser es5 outputs, so we'd have to simultaneously update to es2020 output.

ryanio:
ok awesome, looks like everything is passing now!

I'll open a PR later today to extract the rlp changes so it can be updated on master, and this PR can be targeted to be merged into develop.

Then we can release rlp v3 on npm and gh releases (wohoo!)

I'll include some pointers in the release notes how to update from Buffers to Uint8Arrays when dealing with the updated RLP.encode/decode, from my learnings with this integration.

@holgerd77 do you think I should add the two new util method helpers `bufArrToArr` and `arrToBufArr` for the Uin8Array<>Buffer conversions on master, or keep it on develop? They are non-breaking on util so I figure we can add them now to help as we update each package individually, and also others can use them in updating from rlp v2 to v3.

holgerd77:
@ryanio great work, wohoo!!! 🎉

I guess it makes sense to integrate these two new functions into the existing v2 version (so: to `master`). Is it possible to type these and also add 1-2 test cases?

ryanio:
will open a new PR here to target the `develop` branch, changes from `master` are getting cluttered

ryanio:
Continued in #1662

PR #1634, opened 2022-01-14
VM: Add comparison testing between branchs for State Test Runner

acolytec3:
This adds a new script `diffTester.sh` that can be used to execute the same State Test against two different branches (e.g. testing a change in jump destination analysis versus `master`).  I'm not guaranteeing it's full proof but it should stash any current changes, switch to each of the appropriate branches, execute the specified test, and then restore your current working branch and any uncommitted changes from stash.

It also adds a new parameter to the State Test Runner `--reps=[number of test repetitions` which will run the specified test the number of times specified by `reps` and then output the average run time across the test repetitions as below:
```sh
TAP version 13
# GeneralStateTests
# file: ../../../performance_state_tests/jumpdest_analysis/jd_attack_london_10MGas_JUMPDEST_250M.json test: jd_attack_london_10MGas_JUMPDEST_250M
ok 1 [ 3.389 secs ] the state roots should match (successful tx run)
ok 2 [ 1.017 secs ] the state roots should match (successful tx run)
ok 3 [ 0.964 secs ] the state roots should match (successful tx run)
# Average test run: 1.79 s
```

The easiest way to test this would be to create a new branch off this existing branch and then run the script from the `packages/vm` directory as below:
`scripts/diffTester.sh -a [your branch here] -b performance-testing -r 3 -t ""path/to/your/favorite/state/test.json"


packages/vm/scripts/diffTester.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
```
holgerd77, line 1: For me this works, but I guess there is often some demand (and generall better) to define this as a general shell script and not a bash script, see e.g. https://stackoverflow.com/questions/5725296/difference-between-sh-and-bash ?

packages/vm/scripts/diffTester.sh
```diff
@@ -0,0 +1,35 @@
...
+
+usage() {
+    echo 'Usage: ./diffTester.sh [-a branch1] [-b branch2] [-t "path/to/my/test.json"] [-r 5] [-f London]'
```
holgerd77, line 4: Not sure, might this be overengineered to allow to pass in two branches here and might it be enough to just always compare against the current branch, guess that's the main use case (and then call the option maybe just `-b` for branch)?

But totally no strong opinion here and can go either way.

holgerd77:
Nice, I just gave this a test run, works generally great!

Two comments:

1. Can you please add a note or small sub section on this along the [Benchmarks](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/DEVELOPER.md#benchmarks) section in the DEVELOPER.md developer documentation? Single scripts laying around have otherwise a tendency to get forgotten.


2. When done with repetition the first result is always very much off the charts (so: first run is a lot slower than all the others). Haven't looked deeper in the code, but is the start timer maybe placed too early in the code flow so that some stuff of the setup stuff is also measured along which distort the results? 🤔 No total blocker though, one can get used to that.

holgerd77 (commented):
Cool, made some comments along first testing! 🙂 

Please switch to "Ready for Review" and remove the DRAFT state once you are ready again with this.

packages/vm/scripts/diffTester.sh
acolytec3: I'll verify this works (i.e. just using `#!/bin/sh` and update if so).  I don't think there's anything fancy in this script but will confirm.

packages/vm/scripts/diffTester.sh
acolytec3: Fair point.  Will update accordingly.

acolytec3:
> 2. When done with repetition the first result is always very much off the charts (so: first run is a lot slower than all the others). Haven't looked deeper in the code, but is the start timer maybe placed too early in the code flow so that some stuff of the setup stuff is also measured along which distort the results? thinking No total blocker though, one can get used to that.

On this one, I moved the "Date.now()` method after the `VM = require` statement and that seems to significantly cut down on the variance between the first run and following ones but I can't it down to where the first test is within the exact same range as the others.  I think Node is doing some dependency caching that was throwing it off on the first run but I can't define exactly where it is.  One option would be to just drop the first run and not calculate the average based on that.

Otherwise, this is more or less ready for re-review again.  I've incorporated your other suggestions and added some additional logic to make the `git stash` logic a little more intelligent.

holgerd77 (approved):
Cool, thanks for the updates, looks great! 😄 

Will directly merge.

PR #1637, opened 2022-01-16
Include polyfill for Node `assert` API to ensure compatibility with Webpack 5

steveluscher:
`object.ts` makes use of the `assert` module. This is a Node API that's no longer polyfilled by Webpack 5.

Anyone who depends upon `ethereumjs-util/object` won't be able to upgrade their project to Webpack 5. Webpack will fail to build their project, citing that `assert` is missing. This also implies that anyone trying to use `ethereumjs-util` with Create React App 5 won't be able to either, since CRA5 uses Webpack 5.

Here's an example of folks in the Web3 community having trouble with this: https://github.com/solana-labs/wallet-adapter/issues/241

### DIFF
--- a/packages/util/package.json
+++ b/packages/util/package.json
@@ -36,6 +36,7 @@
     "tsc": "../../config/cli/ts-compile.sh"
   },
   "dependencies": {
+    "assert": "^2.0.0",
     "@types/bn.js": "^5.1.0",
     "bn.js": "^5.1.2",
     "create-hash": "^1.1.2",


steveluscher:
This probably isn't what we ultimately want to do here, because I'm pretty sure that this will cause the `assert` polyfill to be used in both the `node` and `browser` builds. You probably want to configure the build to replace the path in the browser build, maybe using something like `tsconfig-replace-paths`.

ryanio:
@steveluscher thanks for opening a PR and explaining your reasoning!

We also experienced the lack of polyfills when upgrading our client to Webpack 5. The proper configuration is to use the resolve.fallback property to specify the core modules needed. See the various answers here https://stackoverflow.com/questions/64557638/how-to-polyfill-node-core-modules-in-webpack-5

I understand your pain here though, maybe we can provide some instructions in the util readme for tips on configuring webpack 5 with assert? Or if it's downstream maybe something like the suggested node-polyfill-webpack-plugin could help resolve all upstream issues simultaneously (haven't tried that plugin so can't say how good it is, but sounds promising as a catch all solution)

steveluscher:
Well, here's the thing. Folks that are having this sort of trouble don't have access to the Webpack config. They're typically using Create React App 5, and they typically don't even know that they're depending on `ethereumjs-util`.

The particular instance that was brought to my attention was because people couldn't use the Solana wallet adapter with CRA5. This turned out to be because there was a dependency on `ethereumjs-util`:

```
@solana/wallet-adapter -> @solana/wallet-adapter-torus-wallet -> @ethererumjs-util
```

The only way that such a person can get access to the Webpack config to do what you've suggested is to:

* Eject from Create React App (bad)
* Monkeypatch the Webpack config with something like https://github.com/timarney/react-app-rewired (added complexity)

steveluscher:
More context: https://github.com/facebook/create-react-app/issues/11756#issuecomment-996464456

steveluscher:
I wonder if maybe this is an _opportunity_ for `ethereumjs`, after all. The `assert` polyfill that build systems like Webpack used to include is really overkill for the sorts of things that we need it for here.

Most uses of `assert` in `ethereumjs` look like this:

https://github.com/ethereumjs/ethereumjs-monorepo/blob/ade4233ddffffdd146b386de701762196a8c941c/packages/util/src/account.ts#L217-L218

We could eliminate `import assert from 'assert'` entirely and replace those callsites very simply with:

```
if (from.length !== 20) {
  throw new Error('Expected `from` to be of length `20`.');
}
if (salt.length !== 32) {
  throw new Error('Expected `salt` to be of length `32`.');
}
```

Eliminating our dependency on the monolithic `assert` polyfill could save ~9KB in the bundle. The only way that this could go wrong is if a consumer of `ethereumjs` is doing something like this today:

```
try {
  generateAddress2(/* ... */);
} catch (e) {
  if (e instanceof assert.AssertionError) {
    /* ... */
  }
}
```

Not throwing an instance of `AssertionError` would break this code.

ryanio:
I do like the idea of removing assert, it feels antiquated as part of the actual source.

As you mention we probably have to save it for a breaking change/major version release to be safe with the changing error messages, but definitely seems like something we could accomplish for util v8!

steveluscher:
> …but definitely seems like something we could accomplish for util v8!

For argument's sake then… https://github.com/ethereumjs/ethereumjs-monorepo/pull/1638!

holgerd77:
Also a big fan of removing `assert`, we should really go through the whole monorepo stack and remove everywhere (found occurrences in VM and devp2p with a search). 😄 

Will directly open a broader issue on that.

@ryanio Maybe we can also add a lint rule here to avoid usage? 🤔

I would also agree that we - unfortunately - can't do in the current Util version due to potential compatibility issues. Did a quick search and we even use this `e instanceof assert.AssertionError` pattern in some (few) occurrences in our own code.

@steveluscher could you point your work from #1638 towards our `develop` branch? That's where we are preparing our next round of breaking releases right now and then we can directly merge this in here.

Hmm, is there anything we can do for the Util v7 (and maybe also: at least the VM users, devp2p not so critical)? 🤔  So - yes - simply adding the polyfill doesn't seem like a clean solution to me. Other options?

steveluscher:
> …could you point your work from #1638 towards our develop branch?

Done!

> I would also agree that we - unfortunately - can't do in the current Util version due to potential compatibility issues. Did a quick search and we even use this e instanceof assert.AssertionError pattern in some (few) occurrences in our own code.

The only thing I would have done differently in #1638 would be to have made a (lightweight) custom `Error` class that lives in `@ethereumjs/common` and throw that. That way people can still make `instanceof` assertions like this, but they have to switch to importing the new `Error` subclass.

ryanio:
closing since #1638 was merged into `develop`, thanks @steveluscher!

PR #1638, opened 2022-01-16
Util: Remove `assert` in favor of simply throwing `Errors` to ensure compatibility with Webpack 5

steveluscher:
# Motivation

`object.ts` makes use of the `assert` module. This is a Node API that's no longer polyfilled by Webpack 5.

Anyone who depends upon `ethereumjs-util/object` won't be able to upgrade their project to Webpack 5. Webpack will fail to build their project, citing that `assert` is missing. This also implies that anyone trying to use `ethereumjs-util` with Create React App 5 won't be able to either, since CRA5 uses Webpack 5.

Here's an example of folks in the Web3 community having trouble with this: https://github.com/solana-labs/wallet-adapter/issues/241

# Solution

In this PR, rather than to include a full polyfill of `assert` for browsers (eg. https://github.com/ethereumjs/ethereumjs-monorepo/pull/1637) we boil the assertions down to regular `if` statements and thrown `Errors`.

# Test plan

```
cd packages/util
yarn test
```

# Caveat

Eliminating our dependency on the monolithic `assert` polyfill could save ~9KB in the bundle. The only way that this could go wrong is if a consumer of `ethereumjs-util` is doing something like this today:

```
try {
  generateAddress2(/* ... */);
} catch (e) {
  if (e instanceof assert.AssertionError) {
    /* ... */
  }
}
```

Not throwing an instance of `AssertionError` would break this code.

### DIFF
--- a/packages/util/src/account.ts
+++ b/packages/util/src/account.ts
@@ -1,4 +1,3 @@
-import assert from 'assert'
 import { BN, rlp } from './externals'
 import {
   privateKeyVerify,
@@ -214,8 +213,12 @@ export const generateAddress2 = function (from: Buffer, salt: Buffer, initCode: 
   assertIsBuffer(salt)
   assertIsBuffer(initCode)
 
-  assert(from.length === 20)
-  assert(salt.length === 32)
+  if (from.length !== 20) {
+    throw new Error('Expected from to be of length 20')
+  }
+  if (salt.length !== 32) {
+    throw new Error('Expected salt to be of length 32')
+  }
 
   const address = keccak256(
     Buffer.concat([Buffer.from('ff', 'hex'), from, salt, keccak256(initCode)])
@@ -262,7 +265,9 @@ export const pubToAddress = function (pubKey: Buffer, sanitize: boolean = false)
   if (sanitize && pubKey.length !== 64) {
     pubKey = Buffer.from(publicKeyConvert(pubKey, false).slice(1))
   }
-  assert(pubKey.length === 64)
+  if (pubKey.length !== 64) {
+    throw new Error('Expected pubKey to be of length 64')
+  }
   // Only take the lower 160bits of the hash
   return keccak(pubKey).slice(-20)
 }
--- a/packages/util/package.json
+++ b/packages/util/package.json
@@ -43,7 +43,6 @@
     "rlp": "^2.2.4"
   },
   "devDependencies": {
-    "@types/assert": "^1.5.4",
     "@types/node": "^16.11.7",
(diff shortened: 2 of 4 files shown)


steveluscher:
One thing I could have done differently here would be to have made a (lightweight) custom `Error` class that lives in `@ethereumjs/common` and throw that. That way people can still make `instanceof` assertions against that, but they would have to switch to importing the new `Error` subclass.

holgerd77:
@steveluscher Thanks a lot! 😄 

Don't think it's necessary to add a custom Error class. The occurrences/usages of `assert` were really not that special, and I think things get rather more inconsistent if there is one special class on this for some selected errors.

packages/util/src/account.ts
```diff
@@ -214,8 +213,12 @@ export const generateAddress2 = function (from: Buffer, salt: Buffer, initCode:
   assertIsBuffer(salt)
   assertIsBuffer(initCode)
 
-  assert(from.length === 20)
-  assert(salt.length === 32)
+  if (from.length !== 20) {
+    throw new Error('Expected `from` to be of length `20`.')
```
holgerd77, line 217: Hi there, sorry, I know it's - viewed in isolation - an improvement, but we just have no usages of marking key words with "\`" in any other error messages in the library (and the others as well), and things would get inconsistent if we add just for these few new ones. Could you please remove?

Also please do error messages with no dots.

So, example: 'Expected `from` to be of length `20`.' -> 'Expected from to be of length 20'

For the codecov failure I guess that's ok in this case, since semantically this is not really adding functionality, it's just an - somewhat accidental - line expansion here. 🙂

ryanio: how about `Expected 'from' to be of length 20`, otherwise in some cases it might be confusing what the property name is and blends too much into the message, so i think some punctuation does help

even better if we want to add more context for helpful debugging downstream:

`Expected 'from' to be of length 20, received length ${from.length}`

holgerd77: I find all of this better than before, I am just not a fan of introducing selected formatting changes, this makes things - somewhat - inconsistent. Same argument would be true for - I would guess - rather hundreds than dozens of error messages in the monorepo.

So we should rather decide to update them all or not but not starting to switch off from the common pattern on 4-5 selected ones.

steveluscher:
Done, @holgerd77.

holgerd77 (approved):
Thanks a lot @steveluscher, that looks great! 😄 

Will merge into `develop`.

Please note again that the release on this is still some months away, so this will need some time to unfold.

Issue #1636, opened 2022-01-15
Client: Release Consistency/Functionality Checks


holgerd77:
## Introduction

We are now getting into a mode where we are doing regular releases on the client, see #1635 for an example on the next iteration here. The client is a bit of a different beast than the other libraries and running tests is not enough to ensure everything is working properly in a live environment.

It would be good to somewhat agree and define a set of tests we would want to run before a client release which both remains practicable but also ensures that the basic functionality of the client works.

Maybe some (all?) items from this list can also be (semi-)automated in some way, e.g. with a script where one manually checks the log output afterwards (just a first idea, open for alternative ones).

## Possible Tasks

Here is a **wishlist** what would be nice, feel free to add (if you have the permissions directly change in here) or update:

- Run the client on all networks provided and see if a connection is possible
- Start the RPC client and see if a request is possible
- See if the shutdown works correctly
- Let the client run for some substantial amount of time (> 1 hour) and see if the connection remains stable + sync continues
- Short test if mining blocks works
- Test if a (local) libp2p connection works

Guess that would already be most of the essential/core stuff. We can also start/limit on the easier things (network connections) and leave out the more difficult ones (long-running test, local interoperability tests).

acolytec3:
>Start the RPC client and see if a request is possible

What kind of request are you thinking here, just any RPC request to verify that the RPC is up and running or something specific like `eth_getBlockByHash`?

acolytec3:
I think a lot of these can be automated (except for maybe the long run one).

ryanio:
Yeah sounds good, we already have a lot of these.

> Run the client on all networks provided and see if a connection is possible

The only network I've found reliable enough to run on ci is goerli, the others (mainnet, ropsten, rinkeby) can take 5-20 minutes to find an available peer for a block request (sync).

I have a github actions matrix in `workflows/client-build.yml` that uses `test/rpc/cli-sync.spec.ts` to sync and execute a first block. I think adding sepolia would be interesting to try.

We could have a separate test here that runs on label and/or release that gives ample time for the more important set of networks. Ones that have dns network routing should be more reliable to find available peers.

> Start the RPC client and see if a request is possible

Yes @acolytec3 has added `test/cli/cli-rpc.spec.ts` 👍

> See if the shutdown works correctly
> Let the client run for some substantial amount of time (> 1 hour) and see if the connection remains stable + sync continues
👍

> Short test if mining blocks works

[...]

PR #1641, opened 2022-01-17
Client: Add tests to verify shutdown

acolytec3:
Add two tests to detect client shutdown message and that `SIGINT` signal successfully shuts down client process.

Addresses 
>See if the shutdown works correctly
>Test if a (local) libp2p connection works

from #1636


packages/client/test/cli/cli-rpc.spec.ts
```diff
@@ -7,9 +7,8 @@ const cliArgs = ['--rpc', '--ws', '--dev', '--transports=rlpx']
 const end = (child: ChildProcessWithoutNullStreams, hasEnded: boolean, st: tape.Test) => {
   if (hasEnded) return
   hasEnded = true
-  child.stdout.removeAllListeners()
-  child.stderr.removeAllListeners()
-  child.kill('SIGINT')
+  const res = child.kill('SIGINT')
+  st.ok(res === true, 'client shut down successfully')
```
ryanio, line 11:
```suggestion
  st.ok(res, 'client shut down successfully')
```

packages/client/test/cli/cli-rpc.spec.ts
```diff
@@ -7,9 +7,8 @@ const cliArgs = ['--rpc', '--ws', '--dev', '--transports=rlpx']
 const end = (child: ChildProcessWithoutNullStreams, hasEnded: boolean, st: tape.Test) => {
   if (hasEnded) return
   hasEnded = true
-  child.stdout.removeAllListeners()
-  child.stderr.removeAllListeners()
```
ryanio, line 11: is it possible to keep these here in the end() function, but just after the new st.ok check?

acolytec3: I just gave up on checking for the exit message.  If you remove them in function after the kill signal is sent, it removes them before the client prints the "Exiting..." message to the console so the test never registers.  But the `res` value of `true` should be just as good because that `SIGINT` is what we listen for in the client to shut it down and if the result we get back is something screwy, that return value should be `false` in our test.

acolytec3:
I added a new test to connect from one libp2p peer to another so this PR technically addresses 2 of @holgerd77 's points from #1636 now.  :grin:

PR #1642, opened 2022-01-17
rlp: add to root readme

ryanio:
This PR adds rlp to the root readme on master.

It also adds the carryforward option for the rlp flag since the coverage is not run on every pr or build (https://docs.codecov.com/docs/flags#carryforward-flags)


holgerd77 (approved):
LGTM

PR #1643, opened 2022-01-17
Devp2p, Client: Fix duplicated debug messages

ScottyPoi:
Updates debug package to 4.3.3

Also fixes some logic devp2p code that caused multiple debug messages.

### DIFF
--- a/packages/devp2p/src/dpt/server.ts
+++ b/packages/devp2p/src/dpt/server.ts
@@ -255,9 +255,8 @@ export class Server extends EventEmitter {
    * @param msg Message text to debug
    */
   private debug(messageName: string, msg: string) {
-    debug(msg)
     if (this.msgDebuggers[messageName]) {
       this.msgDebuggers[messageName](msg)
-    }
+    } else debug(msg)
   }
 }
--- a/packages/devp2p/src/dpt/server.ts
+++ b/packages/devp2p/src/dpt/server.ts
@@ -4,12 +4,12 @@ import ms from 'ms'
 import { debug as createDebugLogger } from 'debug'
 import LRUCache = require('lru-cache')
 import { encode, decode } from './message'
-import { keccak256, pk2id, createDeferred, formatLogId } from '../util'
+import { keccak256, pk2id, createDeferred, formatLogId, devp2pDebug } from '../util'
 import { DPT, PeerInfo } from './dpt'
 import { Socket as DgramSocket, RemoteInfo } from 'dgram'
 
-const DEBUG_BASE_NAME = 'devp2p:dpt:server'
-const debug = createDebugLogger(DEBUG_BASE_NAME)
+const DEBUG_BASE_NAME = 'dpt:server'
+const debug = devp2pDebug.extend(DEBUG_BASE_NAME)
 const verbose = createDebugLogger('verbose').enabled
 
 const VERSION = 0x04
@@ -48,7 +48,7 @@ export class Server extends EventEmitter {
   _socket: DgramSocket | null
 
   // Message debuggers (e.g. { 'findneighbours': [debug Object], ...})
-  private msgDebuggers: { [key: string]: (debug: string) => void } = {}
+  // private msgDebuggers: { [key: string]: (debug: string) => void } = {}
 
   constructor(dpt: DPT, privateKey: Buffer, options: DPTServerOptions) {
     super()
@@ -62,7 +62,7 @@ export class Server extends EventEmitter {
     this._parityRequestMap = new Map()
     this._requestsCache = new LRUCache({ max: 1000, maxAge: ms('1s'), stale: false })
 
-    this.initMsgDebuggers()
+    // this.initMsgDebuggers()
 
     const createSocket = options.createSocket ?? dgram.createSocket.bind(null, { type: 'udp4' })
     this._socket = createSocket()
@@ -241,12 +241,12 @@ export class Server extends EventEmitter {
     }
   }
 
-  private initMsgDebuggers() {
-    const MESSAGE_NAMES = ['ping', 'pong', 'findneighbours', 'neighbours']
-    for (const name of MESSAGE_NAMES) {
-      this.msgDebuggers[name] = createDebugLogger(`${DEBUG_BASE_NAME}:${name}`)
-    }
-  }
+  // private initMsgDebuggers() {
+  //   const MESSAGE_NAMES = ['ping', 'pong', 'findneighbours', 'neighbours']
+  //   for (const name of MESSAGE_NAMES) {
+  //     this.msgDebuggers[name] = createDebugLogger(`${DEBUG_BASE_NAME}:${name}`)
+  //   }
+  // }
 
   /**
    * Debug message both on the generic as well as the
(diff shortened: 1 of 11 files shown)


packages/devp2p/src/rlpx/peer.ts
```diff
@@ -699,18 +699,15 @@ export class Peer extends EventEmitter {
...
     const ip = this._socket.remoteAddress
-    if (ip && this.msgDebuggers[ip]) {
-      this.msgDebuggers[ip](msg)
-    }
     if (disconnectReason && messageName === 'DISCONNECT') {
       if (this.msgDebuggers[disconnectReason]) {
         this.msgDebuggers[disconnectReason](msg)
       }
-    }
+    } else if (this.msgDebuggers[messageName]) {
```
acolytec3, line 707: @holgerd77 Which messages did we want to dedup on this one?  With this PR's changes, it looks like this:
image

And on current `master`, it looks like this.
image

Should we be removing the IP based messages?  I believe with these changes we will lose the IP based messages entirely since I don't know if we'll ever encounter a situation where there is an IP based message debugger that doesn't also have a `messageName` debugger (which is a higher priority in the conditional logic here.  Maybe we should change it to something like below so users could specify if they want to see log messages from a specific IP address:
```
if (ip && process.env.DEBUG === `devp2p:${ip}`) {
      this.msgDebuggers[ip](msg)
    } else if (disconnectReason && messageName === 'DISCONNECT') {
      if (this.msgDebuggers[disconnectReason]) {
        this.msgDebuggers[disconnectReason](msg)
      }...
```

It produces output that looks something like this:
image

We'd have to apply this same check everywhere we have the IP based messaging and then add some updates to the docs to reflect this usage but it seems to work well.

acolytec3 (dismissed):
Great work @ScottyPoi and yay for the first PR!  

I'm concerned about the IP based debug messages and would like others to weigh in.  We need to decide how to handle them since we lose them with these changes (as far as I can tell based on testing it out) without adding an additional way to access those messages.

packages/devp2p/src/rlpx/peer.ts
holgerd77: These `devp2p:127.0.0.1` or `devp2p:rlpx:peer:HELLO` prefixes on the log messages are actually *only* the names of the loggers activated and not a "feature" to display the IP adress on some loggers. 😋

So on your usage this is the correct dedublication, the main point of this whole thing is the we do not have situations where the same message (in the screenshot from above e.g. "Received HELLO message...") is sent multiple times.

And even if only `devp2p:rlpx:peer:HELLO` is logged, the remote peer IP is visible in the debug message itself.

That said:

I also tested a local connection along [these](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/examples/local-debugging.md#local-connections--ethereumjs---ethereumjs) instructions with the following command:

```shell
cd ../devp2p && npm run build && cd ../client && DEBUG=devp2p:127.0.0.1 npm run client:start:dev2 -- --bootnodes=enode://90e0424dff457a1b677b45aa0bf861e71284c45514e944060f8fb3cca65ddc1b241e686e7a9fb8ac37e6090376a4faf379368cc8ad59ae29194ee3126c3e9c88@127.0.0.1:30303
```

And with the current conditional order the IP logger wouldn't display in this case.

I guess it's enough if we just switch the meesage logger and the IP logger order and keep the rest in place, so this would work for me locally:

```typescript
const ip = this._socket.remoteAddress
if (disconnectReason && messageName === 'DISCONNECT') {
  if (this.msgDebuggers[disconnectReason]) {
    this.msgDebuggers[disconnectReason](msg)
  }
} else if (ip && this.msgDebuggers[ip]) {
  this.msgDebuggers[ip](msg)
} else if (this.msgDebuggers[messageName]) {
  this.msgDebuggers[messageName](msg)
} else {
  debug(msg)
}
```

Short nit for @ScottyPoi (actually all of this comment is mainly for Scotty of course 😛): it would be good if you also add brackets to one-liner code (so this simple `debug(msg)`. We had some bugs lately caused by someone adding something below such a line and not noticing this is not executed in this conditional context. So this would be a bit more on the safe side.

Note: this order change from above would also need to be adjusted in the `eth/index.ts` file.

packages/devp2p/src/rlpx/rlpx.ts
```diff
@@ -233,15 +233,15 @@ export class RLPx extends EventEmitter {
...
-      }
+      // if (disconnectWe) {
+      //   debug(
+      //     `disconnect from ${socket.remoteAddress}:${socket.remotePort}, reason: ${DISCONNECT_REASONS[reason]}`
+      //   )
+      // } else {
+      //   debug(
+      //     `${socket.remoteAddress}:${socket.remotePort} disconnect, reason: ${DISCONNECT_REASONS[reason]}`
+      //   )
+      // }
```
holgerd77, line 244: What's this? 😀

holgerd77 (commented):
Thanks, great start 🎉, some comments. 🙂

holgerd77:
Ah, just realizing that this special `FIRST_PEER` logger (should take the first peer respectively its IP and then only log the messages from this peer) is also not working yet, also not with the code change I suggested.

I used:

```shell
cd ../devp2p && npm run build && cd ../client && DEBUG=devp2p:FIRST_PEER npm run client:start:dev2 -- --bootnodes=enode://90e0424dff457a1b677b45aa0bf861e71284c45514e944060f8fb3cca65ddc1b241e686e7a9fb8ac37e6090376a4faf379368cc8ad59ae29194ee3126c3e9c88@127.0.0.1:30303
```

Note that this might also not have been working before this PR, didn't test.

packages/devp2p/src/dpt/server.ts
```diff
@@ -255,9 +255,8 @@ export class Server extends EventEmitter {
    * @param msg Message text to debug
    */
   private debug(messageName: string, msg: string) {
-    debug(msg)
     if (this.msgDebuggers[messageName]) {
       this.msgDebuggers[messageName](msg)
-    }
+    } else debug(msg)
```
ryanio, line 260:
```suggestion
    } else {
      debug(msg)
```

packages/devp2p/src/eth/index.ts
```diff
@@ -377,14 +377,12 @@ export class ETH extends EventEmitter {
...
+    const ip = this._peer._socket.remoteAddress
     if (this.msgDebuggers[messageName]) {
       this.msgDebuggers[messageName](msg)
-    }
-    const ip = this._peer._socket.remoteAddress
-    if (ip && this.msgDebuggers[ip]) {
+    } else if (ip && this.msgDebuggers[ip]) {
       this.msgDebuggers[ip](msg)
-    }
+    } else debug(msg)
```
ryanio, line 385:
```suggestion
    } else {
      debug(msg)
```

packages/devp2p/src/dpt/server.ts
ryanio: ah sorry nvm already saw holger mentioned this in his comment below https://github.com/ethereumjs/ethereumjs-monorepo/pull/1643/files#r786632639

packages/devp2p/src/rlpx/rlpx.ts
ScottyPoi: This seemed entirely reduntant to debug messages coming from peer.ts.  I left it here commented out in case I was mistaken

ScottyPoi:
devp2p debug

ScottyPoi:
Changed to using the debug.extend() method to create the extended message names

ScottyPoi:
Screenshot from 2022-01-18 23-34-59

holgerd77 (commented):
Ah, you found the way how to actually *properly* do this, that's so great!!! 🙏 🙏 🙏

That's a lot better, thanks for digging deeper into this.

One comment on the disconnect debug removal, also please remove all commented out code, I guess otherwise this should be ready for merge.

Great work! 🙂

holgerd77 (approved):
LGTM, thanks a lot Scotty! 🙂

PR #1648, opened 2022-01-18
rlp: v3 updates from integration

ryanio:
This PR picks updates from #1630 to merge onto master to prep the rlp v3 release.

Then #1630 can be rebased and retargeted to the `develop` branch to be included in the other packages over time as breaking releases add es2020 support for BigInt.

Two new methods (with tests) added to util to help with `Buffer`<>`UInt8Array` conversions: `arrToBufArr` and `bufArrToArr`, with usage examples added to the rlp changelog.

### DIFF
--- a/packages/rlp/CHANGELOG.md
+++ b/packages/rlp/CHANGELOG.md
@@ -27,7 +27,26 @@ RLP.encode(1)
 
 ### Uint8Array
 
-Buffers were replaced in favor of using Uint8Arrays for greater compatibility with browsers.
+Buffers were replaced in favor of using Uint8Arrays for improved performance and greater compatibility with browsers.
+
+When upgrading from rlp v2 to v3, you must convert your Buffers to Uint8Arrays before passing in. To help, two new utility methods were added to `ethereumjs-util v7.1.4`: `arrToBufArr` and `bufArrToArr`. These will recursively step through your arrays to replace Buffers with Uint8Arrays, or vise versa.
+
+Example:
+
+```typescript
+// Old, rlp v2
+import * as rlp from 'rlp'
+const bufArr = [Buffer.from('123', 'hex'), Buffer.from('456', 'hex')]
+const encoded = rlp.encode(bufArr)
+const decoded = rlp.decode(encoded)
+
+// New, rlp v3
+import RLP from 'rlp'
+const encoded: Uint8Array = RLP.encode(bufArrToArr(bufArr))
+const encodedAsBuffer = Buffer.from(encoded)
+const decoded: Uint8Array[] = RLP.decode(encoded)
+const decodedAsBuffers = arrToBufArr(decoded)
+```
 
 ### Invalid RLPs
 
--- a/packages/rlp/src/index.ts
+++ b/packages/rlp/src/index.ts
@@ -1,15 +1,11 @@
-export type Input = string | number | bigint | Uint8Array | List | null | undefined
+export type Input = string | number | bigint | Uint8Array | Array<Input> | null | undefined
 
-// Use interface extension instead of type alias to
-// make circular declaration possible.
-export interface List extends Array<Input> {}
+export type NestedUint8Array = Array<Uint8Array | NestedUint8Array>
 
 export interface Decoded {
   data: Uint8Array | NestedUint8Array
   remainder: Uint8Array
 }
-
-export interface NestedUint8Array extends Array<Uint8Array | NestedUint8Array> {}
 
 /**
  * RLP Encoding based on https://eth.wiki/en/fundamentals/rlp
@@ -51,13 +47,11 @@ function safeSlice(input: Uint8Array, start: number, end: number) {
 /**
  * Parse integers. Check if there is no leading zeros
  * @param v The value to parse
- * @param base The base to parse the integer into
  */
 function decodeLength(v: Uint8Array): number {
-  if (v[0] === 0 && v[1] === 0) {
+  if (v[0] === 0) {
     throw new Error('invalid RLP: extra zeros')
   }
-
   return parseHexByte(bytesToHex(v))
 }
 
@@ -147,12 +141,12 @@ function _decode(input: Uint8Array): Decoded {
       remainder: input.slice(length + llength),
     }
   } else if (firstByte <= 0xf7) {
-    // a list between  0-55 bytes long
+    // a list between 0-55 bytes long
     length = firstByte - 0xbf
     innerRemainder = safeSlice(input, 1, length)
     while (innerRemainder.length) {
       d = _decode(innerRemainder)
-      decoded.push(d.data as Uint8Array)
+      decoded.push(d.data)
       innerRemainder = d.remainder
     }
 
@@ -161,7 +155,7 @@ function _decode(input: Uint8Array): Decoded {
       remainder: input.slice(length),
     }
   } else {
-    // a list  over 55 bytes long
+    // a list over 55 bytes long
     llength = firstByte - 0xf6
     length = decodeLength(safeSlice(input, 1, llength))
     if (length < 56) {
@@ -176,7 +170,7 @@ function _decode(input: Uint8Array): Decoded {
 
     while (innerRemainder.length) {
       d = _decode(innerRemainder)
-      decoded.push(d.data as Uint8Array)
+      decoded.push(d.data)
       innerRemainder = d.remainder
     }
 
@@ -198,7 +192,6 @@ function bytesToHex(uint8a: Uint8Array): string {
 }
 
 function parseHexByte(hexByte: string): number {
-  if (hexByte.length !== 2) throw new Error('Invalid byte sequence')
   const byte = Number.parseInt(hexByte, 16)
   if (Number.isNaN(byte)) throw new Error('Invalid byte sequence')
   return byte
--- a/packages/util/README.md
+++ b/packages/util/README.md
@@ -80,6 +80,8 @@ import { stripHexPrefix } from 'ethereumjs-util'
 
 ### Re-Exports
 
+(Please note these are deprecated and planned to be removed in the next major release, util v8)
+
 `ethereumjs-util` re-exports the following commonly-used libraries:
 
 - [BN.js](https://github.com/indutny/bn.js) (version `5.x`)
--- a/packages/util/README.md
+++ b/packages/util/README.md
@@ -80,8 +80,6 @@ import { stripHexPrefix } from 'ethereumjs-util'
 
 ### Re-Exports
 
-(Please note these are deprecated and planned to be removed in the next major release, util v8)
-
 `ethereumjs-util` re-exports the following commonly-used libraries:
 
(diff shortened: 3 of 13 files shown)


packages/rlp/src/index.ts
```diff
@@ -1,16 +1,12 @@
...
-export interface List extends Array<Input> {}
+export type NestedUint8Array = Array<Uint8Array | NestedUint8Array>
 
 export interface Decoded {
   data: Uint8Array | NestedUint8Array
   remainder: Uint8Array
 }
 
-export interface NestedUint8Array extends Array<Uint8Array | NestedUint8Array> {}
-
```
ryanio, line 13: I didn't see any type issues changing to circular declaration here, so perhaps TypeScript added support for this from when the comment was originally written.

packages/util/src/bytes.ts
```diff
@@ -300,3 +306,29 @@ export const validateNoLeadingZeroes = function (values: { [key: string]: Buffer
...
   }
 }
+
+/**
+ * Converts a {@link Uint8Array} or {@link NestedUint8Array} to {@link Buffer} or {@link NestedBufferArray}
+ */
+export function arrToBufArr(arr: Uint8Array): Buffer
+export function arrToBufArr(arr: NestedUint8Array): NestedBufferArray
+export function arrToBufArr(arr: Uint8Array | NestedUint8Array): Buffer | NestedBufferArray
+export function arrToBufArr(arr: Uint8Array | NestedUint8Array): Buffer | NestedBufferArray {
```
ryanio, line 316: I'm not quite sure why I needed to do this final line twice, I thought I was getting pretty good at understanding these overloads too hehe, maybe a TypeScript wiz can help inform me, but I believe I set it up correctly.

packages/util/README.md
```diff
@@ -80,6 +80,8 @@ import { stripHexPrefix } from 'ethereumjs-util'
 
 ### Re-Exports
 
+(Please note these are deprecated and planned to be removed in the next major release, util v8)
```
holgerd77, line 83: I am not sure if we can say this so confidentaly on the `BN.js` export, I guess this only holds if we really manage to remove `BN.js` completely from our monorepo?

So I would leave this open for now.

In case we do not remove we likely need this re-export to not fall into this trap again of eventually having different BN.js major version BNs being passed around between the libraries and not being compatible with each other? 🤔 Eventually this is not relevant though if we are really determined to have this current BN version as the last version we support/have integrated. Not sure though if we can say this with confidence right now.

packages/rlp/CHANGELOG.md
```diff
@@ -27,7 +27,26 @@ RLP.encode(1)
...
+const encoded = rlp.encode(bufArr)
+const decoded = rlp.decode(encoded)
+
+// New, rlp v3
+import RLP from 'rlp'
+const encoded: Uint8Array = RLP.encode(bufArrToArr(bufArr))
+const encodedAsBuffer = Buffer.from(encoded)
+const decoded: Uint8Array[] = RLP.decode(encoded)
+const decodedAsBuffers = arrToBufArr(decoded)
+```
```
holgerd77, line 49: Just a note: when we release v3 we should add this to the README to for greater visibility (alternative we can also already add now with a "(not yet released)" addition after "v3" and then remove on v3 release.

packages/rlp/src/index.ts
```diff
@@ -1,16 +1,12 @@
-export type Input = string | number | bigint | Uint8Array | List | null | undefined
+export type Input = string | number | bigint | Uint8Array | Array<Input> | null | undefined
 
-// Use interface extension instead of type alias to
-// make circular declaration possible.
-export interface List extends Array<Input> {}
```
holgerd77, line 5: Not sure if this removal is worth the risk of breaking something on this release, in case someone has imported and used? Maybe we just keep and remove along v3?

packages/rlp/src/index.ts
```diff
@@ -1,16 +1,12 @@
...
+export type NestedUint8Array = Array<Uint8Array | NestedUint8Array>
 
 export interface Decoded {
   data: Uint8Array | NestedUint8Array
   remainder: Uint8Array
 }
 
-export interface NestedUint8Array extends Array<Uint8Array | NestedUint8Array> {}
```
holgerd77, line 12: Not sure, might this `interface` -> `type` switch cause compatibility problems as well?

holgerd77 (changes requested):
Hmm, can you please test this new typing of the conversion methods on the `v3` PR on #1630? I just injected the typings in there and get a lot of "No overload matches this call." errors already in the `Util` library when trying to compile.

Maybe these all-catching conversion methods aren't so practical when it comes to typing and one should rather consider 1-to-1 methods with a single input and output? 🤔

Otherwise one would need to always re-cast when these methods get applied.

I guess in most (all?) cases the structure of the input is clear, e.g. if something is a simple Buffer or a Buffer array?

packages/rlp/src/index.ts
ryanio: Well, v3 is a breaking release so the types are different and not backward compatible

packages/rlp/src/index.ts
ryanio: This is a new type that didn't exist in v2

packages/rlp/CHANGELOG.md
ryanio: sounds good, I already updated the README but can add a section for Buffer compatibility since the topic isn't directly discussed there.

packages/util/README.md
ryanio: sounds good, will remove the deprecation notices

packages/rlp/CHANGELOG.md
ryanio: > Hmm, can you please test this new typing of the conversion methods on the v3 PR on #1630? I just injected the typings in there and get a lot of "No overload matches this call." errors already in the Util library when trying to compile.

Just tried it out, this was the one place that error'd out: (this is because the version on that branch uses `any` so it basically skips type checks)

> Type 'Decoded' is missing the following properties from type '(Uint8Array | NestedUint8Array)[]': length, pop, push, concat, and 32 more.
>  `const values = arrToBufArr(RLP.decode(Uint8Array.from(serialized)))`

This is because we don't know what type `RLP.decode` outputs, so this can be fixed with:

`const values = arrToBufArr(RLP.decode(Uint8Array.from(serialized)) as Uint8Array[])`

then we have to cast `as Buffer[]` for `fromValuesArray(values: Buffer[])` because `NestedBufferArray` !== `Buffer[]`

which I think is okay. So:

`const values = arrToBufArr(RLP.decode(Uint8Array.from(serialized)) as Uint8Array[]) as Buffer[]`

a little ugly but better than using `any`?

My overload typings above just help if you pass in a plain non-array Uint8Array or Buffer you get the non-array version back, so there should be less need for manual overrides in those cases to cast.

ryanio: have added a Buffer compatibility section to the readme here: https://github.com/ethereumjs/ethereumjs-monorepo/pull/1648/files#diff-68eff8537817899e4ff13ddb30d65da860ae9de825c4426c5551b3bf5ede53a6R35 let me know if it looks good

packages/util/src/bytes.ts
ryanio: Actually, instead of doing this, it would be much cooler if the return type was the same shape array as the input type but Buffers replaced with Uint8Arrays. That would probably require much less casting afterward. I'm not sure if this can be accomplished with generics, but I'll read the documentation and try to see.

holgerd77 (commented):
Ok, I am confused here on various stuff, e.g. there is also an `ES2020` target in here.

So to recap: My current assumption is that this PR is for an RLP `v2.x` release which does some preparation for an upcoming v3 release. Is this correct or am I mixing things up?

holgerd77:
Update: Ah, now I'll slowly get my misconception.

We *do* have RLP v3 already on `master`.

Wow.

That was a long one. 🙄 😛

Hmm, I find this a pretty much unlucky decision though with the scope of BigInt integration not really decided. Cannot really recall why we had been so confident on this? 🤔

I would really very much suggest that we cherry-pick the RLP v3 commits out to a separate branch and then do the releases from there and revert on `master` for now?

paulmillr:
Any updates @ryanio ?

holgerd77 (approved):
LGTM

holgerd77:
Sorry, don't want to delay here that much further, but maybe two questions:

1. @paulmillr a real browser build - as stated as some goal in your initial PR https://github.com/ethereumjs/rlp/pull/90 with "This would allow to support browsers out-of-box." is not yet included in this release. So this would still need bunding/packaging or the like. Does this still serve your (and others?) needs or should we add a more direct-to-use dedicated browser build? 🤔
2. @ryanio @paulmillr Since #1654 (ESM build) has progressed that far I wonder if we should integrate here (and: eventually in v2) as well?

paulmillr:
@holgerd77 this is good enough for browsers. The main thing was "0 deps", and "no buffers".

As for ESM, could I suggest to not add it to v2? People will upgrade to v3 sooner in this case 😄

holgerd77:
> @holgerd77 this is good enough for browsers. The main thing was "0 deps", and "no buffers".

Ok, I'll trust you on this one, but just for me to get a better picture (really do not have that much browser-need feeling TBH, would be nice if we expand a bit on this over time 😬): what then is the main browser advantage here, if you integrate this into a build chain (Webpack, whatever) wouldn't this be polyfilled anyhow?

Is there any usability advantage here? Or are you expecting foremost some performance gains? 

> As for ESM, could I suggest to not add it to v2? People will upgrade to v3 sooner in this case 😄

Hehe.

I am not sure if you are underestimating build targets though. I would assume there are various lib users out there who are just not ready yet for a BigInt switch, so I assume that the v2 release will still be out there for some time (and that would be an argument to also bring generic improvements). Metamask e.g. has [Chrome 66](https://github.com/MetaMask/metamask-extension/blob/develop/babel.config.js#L9) in it's minimal release target, which is still one off from BigInt introduction (not to speak about other ES2020 features. So I wonder if we should really at least disallow some of the newer language constructs/methods per ESLint? 🤔).

paulmillr:
> what then is the main browser advantage here, if you integrate this into a build chain (Webpack, whatever) wouldn't this be polyfilled anyhow?

That doesn't always happen, not all bundlers auto-shim buffers. My main goal, as usual, is supply chain auditability - lack of bn.js makes the code much easier to reason about. 

As for MM not supporting bigints - well, they'll need to bump their build target from the version which has less than 0.7% of global usage. Otherwise, they won't be able to use our new js-ethereum-cryptography.

Old appe and old build targets can continue to use rlp v2. No one shuts them down, after all. Just want to make transition to bigints faster, because I know how hard it can be to ask folks to upgrade.

holgerd77:
Ok, I've now published the new version on npm: https://www.npmjs.com/package/rlp

We'll keep the release a bit under radar until things are a bit tested and there is some first feedback.

PR #1649, opened 2022-01-19
Common, VM: Remove unused HF options from Common, move supportedHardforks to VM

### LINKED ISSUE #1274: Deprecate supportedHardforks/onlySupported/onlyActive HF options
The mentioned HF [options](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/index.ts#L50) in Common are very much unused, misplaced and (in the case of `onlyActive`) also semantically confusing and should be deprecated, a removal is noted for v6 #1024.

See this (shortened) comment from the chat:

> The supportedHardforks, onlySupported and onlyActive  options are overengineered (I have to know, I once introduced them :stuck_out_tongue_winking_eye: ) and will likely be removed on the next round of breaking releases, already added this todo some time ago to our next breaking release planning list (https://github.com/ethereumjs/ethereumjs-monorepo/issues/1024) (not discussed with the team yet, but I would very much assume that everyone should be ok with that).
>
> The initial idea was to make sure e.g. for the VM that function calls will only work for the supported HFs provided to the option if the option is used (otherwise they will always work and there is no limitation on the supported HFs, therefore the true default mentioned by Ryan). So this option was for limiting the VM to the supported HFs e.g..
>
> It became clear over time though that this option is really at the wrong place (at the end the VM should care for the supported HFs itself, actually the current implementation also does this) and it is just increasing code complexity and bloats the API. And furthermore the option is just not very much used, internally we don't use at all.

The current `onlySupported` and `onlyActive` options are currently combined in a TypeScript type `hardforkOptions` and taken as an input for various functions. Often this is highly confusing - e.g. for `activeHardfork(blockNumber?: BNLike | null, opts: hardforkOptions = {})` (only use the active HFs for determining if a HF is active, häh??? 😝 ). These usages should likely also get a deprecation note. There should nevertheless be a closer look at the implications of a removal respectively one should develop an idea how a consistent removal will look like (it might very well be that the current state is just *not* very consistent or it might not make very much sense to use the option, on the other hands I had some thoughts on why I introduced these options several years ago, so there *does likely exist* some reasoning which should be evaluated on validity).

These thoughts should then go into the deprecation notes for the method parameter usage.

emersonmacro:
Closes #1274

Removing `onlySupported` and `onlyActive` was pretty straightforward, but it took me a bit longer to work out how to move `supportedHardforks` to the `vm` package. When doing so, I realized we might not need to store `opts.supportedHardforks` in the VM instance (e.g. in `this._supportedHardforks`) at all, and instead it just becomes a check at time of instantiation. But I'm not sure if I'm missing something, so I left some potential changes commented out in `packages/vm/src/index.ts`. I can update it either way but wanted to get an initial review to see if I'm on the right track.

I also updated the Common and VM READMEs to reflect the respective changes.

### DIFF
--- a/packages/common/README.md
+++ b/packages/common/README.md
@@ -48,18 +48,6 @@ c.genesis().hash // 0x41941023680923e0fe4d74a34bdac8141f2540e3ae90623718e47d66d1
 c.bootstrapNodes() // Array with current nodes
 ```
 
-If the initializing library only supports a certain range of `hardforks` you can use the `supportedHardforks` option to restrict hardfork access on the `Common` instance:
-
-```typescript
-const c = new Common({
-  chain: 'ropsten',
-  supportedHardforks: ['byzantium', 'constantinople', 'petersburg'],
-})
-```
-
-This will e.g. throw an error when a param is requested for an unsupported hardfork and
-like this prevents unpredicted behaviour.
-
 For an improved developer experience, there are `Chain` and `Hardfork` enums available:
 
 ```typescript
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -103,10 +103,6 @@ interface BaseOpts {
    */
   hardfork?: string | Hardfork
   /**
-   * Limit parameter returns to the given hardforks
-   */
-  // supportedHardforks?: Array<string | Hardfork>
-  /**
    * Selected EIPs which can be activated, please use an array for instantiation
    * (e.g. `eips: [ 2537, ]`)
    *
@@ -162,13 +158,6 @@ export interface CustomCommonOpts extends BaseOpts {
   baseChain?: string | number | Chain | BN
 }
 
-// interface hardforkOptions {
-//   /** optional, only allow supported HFs (default: false) */
-//   onlySupported?: boolean
-//   /** optional, only active HFs (default: false) */
-//   onlyActive?: boolean
-// }
-
 /**
  * Common class to access chain and hardfork parameters and to provide
  * a unified and shared view on the network and hardfork state.
@@ -182,7 +171,6 @@ export default class Common extends EventEmitter {
 
   private _chainParams: IChain
   private _hardfork: string | Hardfork
-  private _supportedHardforks: Array<string | Hardfork> = []
   private _eips: number[] = []
   private _customChains: IChain[] | [IChain, GenesisState][]
 
@@ -293,13 +281,11 @@ export default class Common extends EventEmitter {
    * chain params on.
    * @param customChainParams The custom parameters of the chain.
    * @param hardfork String identifier ('byzantium') for hardfork (optional)
-   * @param supportedHardforks Limit parameter returns to the given hardforks (optional)
    */
   static forCustomChain(
     baseChain: string | number | Chain,
     customChainParams: Partial<IChain>,
     hardfork?: string | Hardfork
-    // supportedHardforks?: Array<string | Hardfork>
   ): Common {
     const standardChainParams = Common._getChainParams(baseChain)
 
@@ -309,7 +295,6 @@ export default class Common extends EventEmitter {
         ...customChainParams,
       },
       hardfork: hardfork,
-      // supportedHardforks: supportedHardforks,
     })
   }
 
@@ -361,9 +346,6 @@ export default class Common extends EventEmitter {
       }
     }
     this._hardfork = this.DEFAULT_HARDFORK
-    // if (opts.supportedHardforks) {
-    //   this._supportedHardforks = opts.supportedHardforks
-    // }
     if (opts.hardfork) {
       this.setHardfork(opts.hardfork)
     }
@@ -416,9 +398,6 @@ export default class Common extends EventEmitter {
    * @param hardfork String identifier (e.g. 'byzantium') or {@link Hardfork} enum
    */
   setHardfork(hardfork: string | Hardfork): void {
-    // if (!this._isSupportedHardfork(hardfork)) {
-    //   throw new Error(`Hardfork ${hardfork} not set as supported in supportedHardforks`)
-    // }
     let existing = false
     for (const hfChanges of HARDFORK_CHANGES) {
       if (hfChanges[0] === hardfork) {
@@ -520,10 +499,6 @@ export default class Common extends EventEmitter {
    * @returns Hardfork chosen to be used
    */
   _chooseHardfork(hardfork?: string | Hardfork | null): string {
-    // if (!hardfork) {
-    //   hardfork = this._hardfork
-    // }
-    // return hardfork
     return hardfork ?? this._hardfork
   }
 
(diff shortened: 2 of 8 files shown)


packages/common/README.md
```diff
@@ -48,18 +48,6 @@ c.genesis().hash // 0x41941023680923e0fe4d74a34bdac8141f2540e3ae90623718e47d66d1
...
-```typescript
-const c = new Common({
-  chain: 'ropsten',
-  supportedHardforks: ['byzantium', 'constantinople', 'petersburg'],
-})
-```
-
-This will e.g. throw an error when a param is requested for an unsupported hardfork and
-like this prevents unpredicted behaviour.
-
```
holgerd77, line 62: Ah, nice that you caught that as well. 🙂

packages/common/src/index.ts
```diff
@@ -519,13 +498,8 @@ export default class Common extends EventEmitter {
...
    */
-  _chooseHardfork(hardfork?: string | Hardfork | null, onlySupported: boolean = true): string {
-    if (!hardfork) {
-      hardfork = this._hardfork
-    } else if (onlySupported && !this._isSupportedHardfork(hardfork)) {
-      throw new Error(`Hardfork ${hardfork} not set as supported in supportedHardforks`)
-    }
-    return hardfork
+  _chooseHardfork(hardfork?: string | Hardfork | null): string {
+    return hardfork ?? this._hardfork
```
holgerd77, line 502: I guess I would be a fan of now completely remove this internal function and everywhere replace with this `hardfork ?? this._hardfork` switch (like we do with options on other places), seems more natural to me.

packages/vm/src/index.ts
```diff
@@ -117,6 +117,10 @@ export interface VMOpts {
    * pointing to a Shanghai block: this will lead to set the HF as Shanghai and not the Merge).
    */
   hardforkByTD?: BNLike
+  /**
+   * Limit parameter returns to the given hardforks
+   */
+  supportedHardforks?: Array<string | Hardfork>
```
holgerd77, line 123: Ah no, we do not want to "recreate" this as a feature here, so you can remove this again.

The only reason we have this in the VM is that `Common` sometimes temporarily supports more HFs than implemented in the VM (e.g. if we already add the Merge HF).

So you can keep the `const supportedHardforks = [ ... ]` array from below and then the only thing is you need to do:

1. Wait for `Common` to be initialized inside the VM (also: below)
2. Check if the HF from Common (which might be passed in) is present in the `supportedHardforks` array and otherwise throw

Some background on this: this whole `supportedHardforks` thing had more reason to be there in the times where we didn't support the full range of HFs yet with the VM (so didn't yet backport the old HFs). Now it would even be an option to remove this check completely, but I guess for these temporary edge cases with new not-applied HFs it should still make some sense.

holgerd77 (commented):
Ok, provided some clarification on the VM stuff, otherwise great, that is such a relief to see the Common library simplify, this really has gotten some bloat over the years.

If you stumble upon other stuff in Common where you scratch your head and think "does this really need to be this way/does this really need to be that complicated?" feel free to point out or open a new issue on that.

packages/common/src/index.ts
emersonmacro: Yeah I think that's the right call. PR updated

packages/vm/src/index.ts
emersonmacro: Ah ok, that makes sense. PR updated per your feedback. I updated the test case and the VM README as well.

One question: when I used the `supportedHardforks` array that was already in the code, about 20 VM API tests failed because `London` was not included in the `supportedHardforks` array. I added `London` to the array and all the tests passed. Which made me think, I noticed there was another item in the breaking release issue to update the default hardfork to `London`. Should I include that change in this PR? Or I could tackle it separately in a different PR (or someone else could).

packages/vm/src/index.ts
```diff
@@ -210,32 +206,34 @@ export default class VM extends AsyncEventEmitter {
...
+      'spuriousDragon',
+      'byzantium',
+      'constantinople',
+      'petersburg',
+      'istanbul',
+      'muirGlacier',
+      'berlin',
+      'london',
+      'arrowGlacier',
+    ]
```
holgerd77, line 230: I am not seeing any value in making this a member variable. Can we maybe stick with keeping this local? Or am I overlooking something?

packages/vm/README.md
```diff
@@ -140,18 +140,13 @@ const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Berlin })
...
+Please note that the `VM` constructor will throw an error if you pass in a `Common` instance that is set to a hardfork that the `VM` does not currently support. For example, the following would throw because the `VM` does not yet support the Merge hardfork:
 
 ```typescript
-const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Byzantium })
-const vm = new VM({
-  common,
-  supportedHardforks: [Hardfork.Byzantium, Hardfork.Constantinople],
-})
+const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Merge })
+const vm = new VM({ common })
```
holgerd77, line 147: Actually I would remove this whole doc section. We have already the list with supported HFs above, it doesn't provide much value to describe that things won't work when a non-supported HF is used.

This section is also coming very much from the old not-all-HFs-backported days. 🙂

holgerd77 (changes requested):
Thanks for the updates, two small additional change requests.

packages/vm/src/index.ts
ryanio: just a thought, would it be possible to use the `Hardfork` enum for the values here? The result should still be the same string values.

emersonmacro: @holgerd77 tbh I added it last minute, didn't really have a good reason. Reverted it to local variable

@ryanio Good call! Updated

packages/vm/src/index.ts
```diff
@@ -205,30 +205,34 @@ export default class VM extends AsyncEventEmitter {
...
+      Hardfork.Petersburg,
+      Hardfork.Istanbul,
+      Hardfork.MuirGlacier,
+      Hardfork.Berlin,
+      Hardfork.London,
+      Hardfork.ArrowGlacier,
+    ]
+    if (!supportedHardforks.includes(this._common.hardfork() as Hardfork)) {
+      throw new Error(
+        `Hardfork ${this._common.hardfork()} not set as supported in supportedHardforks`
```
holgerd77, line 232: Small nit, won't make this a blocker though: I guess you can write error messages "one level higher", so end users do not really have a notion of this internal "supportedHardforks" variable and the like, so something like "The provided hardfork {hardfork} is not supported by the VM" is likely better readable/less confusing.

If someone thinks about updating this along some other PR would be nice, but totally no blocker as well.

holgerd77 (approved):
This is such a great PR, thanks a lot Emmett, love seeing all these removed lines 😋, especially on things like the hardfork comparison.

Will merge.

packages/vm/src/index.ts
emersonmacro: Yeah I think that makes sense. I was thinking of trying to tackle the London-as-default-hardfork issue, so I could add it that PR

Issue #1651, opened 2022-01-19
Monorepo -> BN.js to BigInt: Performance Analysis


holgerd77:
Part of #1024

We are currently considering transitioning either core parts of our library code or the libraries as a whole from using [BN.js](https://github.com/indutny/bn.js) to using the new (ES2020) JS [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) data type.

One advantage we suspect from this is getting to a better library performance, especially for the VM, which e.g. uses BN.js instances for the VM [stack](https://github.com/ethereumjs/ethereumjs-monorepo/blob/develop/packages/vm/src/evm/stack.ts) which in turn provides the data basis for most of the ocode [operations](https://github.com/ethereumjs/ethereumjs-monorepo/blob/develop/packages/vm/src/evm/opcodes/functions.ts).

This is a meta issue for collecting some references and experiments along BigInt performance to do some deeper analysis here and somewhat harden the evidence that performance gains play out as expected. 🙂

holgerd77:
This is a really valuable source I found today with someone having written a master thesis trying to performance-optimize a crypto library, one of his trials was actually a BN.js to BigInt switch (actually in the analyzed case having the biggest impact on performance from all measures): https://www.epfl.ch/labs/dedis/wp-content/uploads/2020/01/report-2019-2-Julien-von-Felten-KyberJS-performance.pdf

holgerd77:
Did the following very simple performance tests on arithmetics (relevant for the VM), promising results:

```shell
> console.time('p'); for (let i=1; i <= 100000; i++) { let res = new BN(500).mul(new BN(23)) }; console.timeEnd('p');
// I would manually re-run the above statement for each result below
p: 9.702ms
p: 9.578ms
p: 8.957ms
p: 9.114ms

> console.time('p'); for (let i=1; i <= 100000; i++) { let res = 500n*23n }; console.timeEnd('p');
p: 6.033ms
p: 5.498ms
p: 6.225ms
p: 6.503ms
p: 5.579ms

> console.time('p'); for (let i=1; i <= 100000; i++) { let res = new BN(500).add(new BN(23)) }; console.timeEnd('p');
p: 12.963ms
p: 14.301ms
p: 13.72ms
p: 14.609ms
p: 11.483ms

> console.time('p'); for (let i=1; i <= 100000; i++) { let res = 500n+23n }; console.timeEnd('p');
[...]

PR #1652, opened 2022-01-19
VM: Stack BigInt Transition

holgerd77:
Part of #1651

I gave this a first try today but I got stuck very quickly on the VM arithmetics conversion. Good thing is that I think I've got a much clearer picture now on the challenges ahead. 😋

Thing is: the opcode [functions](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/evm/opcodes/functions.ts) in the VM make pretty heavy usage of bit arithmetics in various ways, I hadn't the extend of this completely on the radar TBH.

So the various opcode implementations are calling into various functions which are already present in [BN.js](https://github.com/indutny/bn.js/), like `toTwos()` (two complement with associated bit width), `fromTwos()`, `notn(width)`, `toRed()`. All this is missing for the `BigInt` data type.

I went a bit deeper, let's have a look at the [SUB](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/evm/opcodes/functions.ts#L60) opcode, using the `toTwos()` function. [Here](https://github.com/indutny/bn.js/blob/master/lib/bn.js#L752) is how this is implemented in `BN.js`.

I also had a look into the noble libraries to see how @paulmillr (will directly tag here, Paul, maybe you have also some insight here 🙂), and Paul has actually integrated various helper functions on this directly in the libraries.

I am not a bit operator pro here and therefore I can't completely judge on the amount of work here needed. We would at least need a somewhat substantial set of robust helper functions which assist here and somewhat provide selected functionality similar to what BN.js is providing.

As a side note, I haven't found a suited library on this.


packages/vm/src/evm/stack.ts
```diff
@@ -80,12 +81,16 @@ export default class Stack {
    * Pushes a copy of an item in the stack.
    * @param position - Index of item to be copied (1-indexed)
    */
-  dup(position: number) {
+  // I would say that we do not need this method any more
+  // since you can't copy a primitive data type
+  // Nevertheless not sure if we "loose" something here?
+  // Will keep commented out for now
```
ryanio, line 87: right BigInt is immutable so it can't be changed, but wouldn't we still want to push it onto the stack?

packages/vm/src/evm/stack.ts
```diff
@@ -17,14 +16,16 @@ export default class Stack {
...
+  push(value: bigint) {
+    if (typeof value !== 'bigint') {
       throw new VmError(ERROR.INTERNAL_ERROR)
     }
 
-    if (value.gt(MAX_INTEGER)) {
-      throw new VmError(ERROR.OUT_OF_RANGE)
-    }
+    // Why do we have this MAX_INTEGER check here on the BN.js values?
+    // Is this still needed for BigInt? (in doubt I would leave to not chance behavior)
```
ryanio, line 25: It's a security/validation check that the evm (or really ethereum overall) should only handle numbers less than MAX_INTEGER

had to check regarding bigint:
> It seems like there is no maximum limit to a BigInt as per spec, which makes sense considering BigInts are supposed to be arbitrary-precision integers, whose "digits of precision are limited only by the available memory of the host system".
https://stackoverflow.com/a/53336729

so it would be wise to keep these checks, but maybe they can be located somewhere where it's checked less frequently so it isn't as taxing on the system (if any extra checks are redundant, not sure)

paulmillr:
You should not use % operator, because it's not a modulo. It's a remainder. To see how to do modulo, check noble-secp256k1, it's one-line function.

Also keep in mind that toString(16) does not pad the output at start. So, "010000" would never appear instead of "10000".

packages/vm/src/evm/opcodes/functions.ts
```diff
@@ -70,7 +71,7 @@ export const handlers: Map<number, OpHandler> = new Map([
     0x03,
     function (runState) {
       const [a, b] = runState.stack.popN(2)
-      const r = a.sub(b).toTwos(256)
+      const r = (a - b) & TWO_POW256_BIGINT
```
axic, line 74: This uses and while others use mod.

acolytec3: Yep, I was following PyEVM here - https://github.com/ethereum/py-evm/blob/1af151ab218b905f4fdf7a285cbe14ebf094a7c4/eth/vm/logic/arithmetic.py#L85 since we don't have a native twosComplement function with bigints.  Just experimenting for now but will see what results it provides.

paulmillr:
@acolytec3 I suggest to do benchmarks. It's only 3.8x faster -- not 20x/100x. Is this really a bottleneck that suggests adding a native dependency?

The pure-js approach does 2.5M operations per second on my machine. Is this "slow"?

```
bigint(bytesToHex(arr)) x 2,583,979 ops/sec @ 387ns/op
bigint-buffer#toBigIntBE x 9,615,384 ops/sec @ 104ns/op
```

```js
const bench = require('micro-bmark');
const {run, mark} = bench;
const {toBigIntBE, toBigIntLE, toBufferBE, toBufferLE} = require('bigint-buffer');

const hexes = Array.from({ length: 256 }, (v, i) => i.toString(16).padStart(2, '0'));
function bytesToHex(uint8a) {
    // pre-caching improves the speed 6x
    let hex = '';
    for (let i = 0; i < uint8a.length; i++) {
        hex += hexes[uint8a[i]];
    }
    return hex;
}

run(async () => {
  const samples = 20_000_000;
  const big = 2n**256n-123n;
  const buf = Buffer.from(big.toString(16), 'hex');
  const ui8a = Uint8Array.from(buf);
  mark('bigint(bytesToHex(arr))', samples,  () => {
    BigInt('0x' + bytesToHex(ui8a));
  });

  mark('bigint-buffer#toBigIntBE', samples, () => {
    toBigIntBE(ui8a);
  });
});
```

holgerd77:
@paulmillr the thing is: these things add up very quickly, especially in the VM where there are thousand and thousand of operations repeatedly along opcode execution, and one goal of this kind of work is to make things faster for consumers like Hardhat, so that they have faster VM execution in test runs e.g..

That said, I am also skeptical on directly adding new dependencies here, especially on work where we want to rip another dependency out. 😋

@acolytec3 maybe we stick with native here for now to see how this goes, we can then still decide later on if an additional dependency would be worth it? I would also like this more from the approach/process side, then we have a clearer picture on the "pure" effect of the transition, this dependency injection that early on feels a lot like some sort of distortion to me which somewhat hinders to get some transparent results on the BigInt integration itself.

acolytec3:
Sure, happy to pull it back out. I saw that the chainsafe implementation of discv5 used it and there usually a good point of reference.

paulmillr:
@acolytec3 modExp takes 6 lines of code to implement. These folks somehow made 30? Lines of code. 

Feel free to copy-paste my code from bls12-381 impl:

https://github.com/paulmillr/noble-bls12-381/blob/c558ffbfc9359ae4a2205059baebe8eab5a4a8da/math.ts#L79

holgerd77:
Oh, *that's* some work here. Should have named this "Switch VM to BigInt". 😬 🙂 🙏 👍

holgerd77:
To everyone: please do not merge any `BigInt` related changes into `develop` right now since scope of the next breaking release round is not yet clear/decided, especially regarding the question if `BigInt` should be introduced and `ES2020` along accepted as a higher transpilation/compilation target.

holgerd77:
Note: I am just in the process of recreating the `develop` branch with a clean commit history now always using rebases to get on top of the `master` branch, see post in the internal "#BigInt Conversion" channel.

I will close this PR here, have done a branch copy locally as a backup just in case.

I will then delete and repush the `develop` branch, then under a fitting backup name, and repush a "new" `develop` branch for continued development with the existing squashed merge PR commits cherry-picked.

Please reopen this PR - or an adopted version of this branch - towards this new `develop` branch (not yet created, will drop a note).

Thanks! 🙂

holgerd77:
Ok, have recreated the `develop` branch, feel free to reopen now.

PR #1654, opened 2022-01-20
monorepo: esm support

ryanio:
This PR adds `./dist.esm` builds for each package and fixes errors encountered.

WIP, some last few issues to fix, but this should be the majority of it.


holgerd77:
That's really great! 👍 

Some note: *if* we want to integrate in the current minor releases, we should give this some really extensive testing - eventually even consider doing a round of `rc` releases (though I am not to keen on that one) - otherwise this is likely too risky.

So really do some test installations in various environments and see how things go.

ryanio:
going to pick this back up and get it updated to the latest master, so we can ship esm support

holgerd77:
> going to pick this back up and get it updated to the latest master, so we can ship esm support

Ok, great, I guess it makes sense to then include this in the coming - originally - EIP-1153 focused release round mid next week or so?

Or can't you completely foresee how long this will take (or does this no make sense for other reasons?)?

paulmillr:
It would be great to understand if that's needed. If yes, we could make a round of js-eth-crypto releases as well

ryanio:
continued in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1936

Issue #1574, opened 2021-11-17
VM, Tx: Run Code Examples in CI


holgerd77:
We have code examples included in various packages. These are somewhat neglected respectively forgotten right now and there is a somewhat higher chance that these break over time with nobody noticing.

This should be changed by adding these examples to CI being run with GitHub actions, optimally in a (somewhat) standardized/unified way (but also: likely no need to introduce overblown centralized structures like a script to route everything through or something), but also without too much custom code on the examples itself (we'll see how it goes).

Following packages contain relevant examples (client + devp2p examples should likely be omitted, too complex, too diverse):

- VM: [examples](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/vm/examples) folder
- Tx: [examples](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/tx/examples) folder

Just to mention here: these examples could generally need some love ❤️ and it would make sense to modernize and also think about new relevant (and also: a bit more exciting!) use cases, e.g. creating a L2 tx with the Transaction library, I am sure there will be others. Other libraries like `Block` or `Blockchain` would likely also benefit from a - small - examples collection.

cbrzn:
I was thinking of tackling this the following way:

With `npm workspaces` we can create a script (which can be `examples` for example) and in the root package.json the script: `"examples": "npm run examples --workspaces --if-present"` will allow us to add the `examples` script in the packages and simply run it if it exists in the package.

The way we would add the examples in packages would be the following:
- Create a folder called `examples` (As is currently done)
- In the package.json, add the script `examples: ts-node examples/*/.ts`
- Create a file for specific examples (i.e: In `transactions`, create one that executes different transaction types and hard forks, or in `block`, a file where a block for PoW and PoS are created) - PD: Just thinking out of loud here, lmk if this doesn't make sense :P
-  Create a README.md in every example folder, to explain every example that lives in this folder

Then adding the command `npm run examples` in the CI will make sure every example works as expected 

At the same time, we can create a new package where we can have reusable functions (probably we can use the `common` library? Not sure, I was thinking of utils package which won't be ever published, to npm, it's just for the sake of the playground), making it easier to create more complex examples and allowing new developers to dive in a more comfortable way into the code by learning from these 

The main points I see with this proposal:
- Reusing code will help to keep consistency between the different packages examples, hence, making it easier to maintain
- By having just one readme per example folder, it's easier to see all the examples in a single source of truth
[...]

PR #1655, opened 2022-01-20
VM, Tx: Examples improvements

cbrzn:
As mentioned in https://github.com/ethereumjs/ethereumjs-monorepo/issues/1574#issuecomment-1015464119 - This PR aims to add the examples to the CI and give them a better structure. I am probably missing a lot of things, please enlight me if you think I should follow some path 🙏


scripts/examples-runner.ts
```diff
@@ -0,0 +1,23 @@
...
+    return console.log("Error loading examples directory: ", err)
+  }
+  const importedFiles = files.map(fileName => {
+    if (extname(fileName) === ".ts") {
+      return import(examplesPath + fileName);
+    }
+  }).filter(file => file)
+
+  await Promise.all(importedFiles)
+})
```
holgerd77, line 23: Do I get this right that this is only for CI and one can still run the examples "stand-alone"?

packages/vm/package.json
```diff
@@ -43,13 +43,15 @@
...
-    "tsc": "../../config/cli/ts-compile.sh"
+    "tsc": "../../config/cli/ts-compile.sh",
+    "examples": "ts-node ../../scripts/examples-runner.ts -- vm"
   },
   "dependencies": {
     "@ethereumjs/block": "^3.6.0",
     "@ethereumjs/blockchain": "^5.5.1",
     "@ethereumjs/common": "^2.6.0",
     "@ethereumjs/tx": "^3.4.0",
+    "@ethereumjs/test-cases": "0.0.1",
```
holgerd77, line 54: We can't have such an addition to the main dependencies.

(or do you want to really deploy these examples into production for people to run them when having the package installed? Then we would need to discuss, that's a big change, no opinion on that yet but sceptical)

holgerd77 (changes requested):
Hi Cesar @cbrzn,
this has various cool ideas in it, great! 😄

This is tackling too much at once though, especially this `test-cases` folder introduction would be a big change (I am somewhat sceptical on the current implementation but open for discussion so that we get to something everyone can agree with).

Can you please resubmit and just extract the CI integration part on the existing examples together with the small tweaks you made in the examples and in some other places of the code?

That would be very much enough for a single PR, especially since this is logically not interdependent with the `test-cases` suggestion. We try to keep PRs modular and as small as possible for security and review reasons, otherwise things easily overload, also discussions get more twisted since one is often discussing several things at once. 🙂

scripts/examples-runner.ts
cbrzn: That's correct! Actually, this is a workaround because I thought we could just use `ts-node examples/*.ts` and it would run all the ts files, but it doesn't :(

PR #1658, opened 2022-01-20
Monorepo (& Tx, VM): Examples scripts in CI

cbrzn:
This PR aims to run the example scripts in the CI to guarantee that they are always working with the latest changes :-)

Currently, it is only working with the `Tx` package because of the structure of the examples in different packages (I.e: In the `VM` package they are organized as folders with `package.json` & `index.ts` in each, and in the `Tx` folder there is only `ts` files). At the moment, the script only supports a folder with all the `ts` files.

I would like to somehow, standardize the structure of these `examples` folders, and IMO I think that we should create only `.ts` files in `/examples` folders. 

My main thought of doing it this way is that we don't have a dependency file per example (if needed) - Instead, we can have these minimal re-utilizable functions somewhere (_probably in a utils file in the packages folder where the example lives(?), and in the `devDependencies` of that package add the external ones needed to runs the example? - just thinking out loud here_). This will also make things easier when it comes to running the script in the CI because we wont need to run `npm i` per example

I added the `monorepo` tag because I am adding a script in the root - Feel free to remove it if you think it doesn't make sense 🙏 😛

### DIFF
--- a/.github/workflows/examples-works.yml
+++ b/.github/workflows/examples-works.yml
@@ -1,0 +1,20 @@
+name: VM
+on:
+  push:
+    branches: [examples-in-ci]
+
+jobs:
+  test-examples:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 16
+          cache: 'npm'
+
+      - run: npm i
+      - run: npm run examples:ci
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "preinstall": "npm run checkNpmVersion",
     "postinstall": "npm run build --workspaces",
     "checkNpmVersion": "./scripts/check-npm-version.sh",
-    "examples:ci": "CI=true npm run examples --workspaces --if-present",
+    "examples": "npm run examples --workspaces --if-present",
     "clean": "./config/cli/clean-root.sh",
     "e2e:publish": "./scripts/e2e-publish.sh",
     "e2e:resolutions": "node ./scripts/e2e-resolutions.js",
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "preinstall": "npm run checkNpmVersion",
     "postinstall": "npm run build --workspaces",
     "checkNpmVersion": "./scripts/check-npm-version.sh",
+    "examples:ci": "CI=true npm run examples --workspaces --if-present",
     "clean": "./config/cli/clean-root.sh",
     "e2e:publish": "./scripts/e2e-publish.sh",
     "e2e:resolutions": "node ./scripts/e2e-resolutions.js",
--- a/packages/tx/examples/transactions.ts
+++ b/packages/tx/examples/transactions.ts
@@ -43,18 +43,18 @@ console.log('--------------------')
 // You can decode it with the rlp module.
 // After that you should have something like:
 const rawTx = [
-  '0x00',
+  '0x',
   '0x09184e72a000',
   '0x2710',
   '0x0000000000000000000000000000000000000000',
-  '0x00',
+  '0x',
   '0x7f7465737432000000000000000000000000000000000000000000000000000000600057',
   '0x1c',
   '0x5e1d3a76fbf824220eafc8c79ad578ad2b67d01b0c2425eb1f1347e8f50882ab',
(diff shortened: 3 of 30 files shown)


packages/tx/examples/transactions.ts
```diff
@@ -43,18 +43,18 @@ console.log('--------------------')
 // You can decode it with the rlp module.
 // After that you should have something like:
 const rawTx = [
-  '0x00',
+  '0x',
```
cbrzn, line 46: This was done because the method `fromValuesArray` changed

.github/workflows/examples-works.yml
```diff
@@ -0,0 +1,20 @@
...
+        with:
+          submodules: recursive
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 16
+          cache: 'npm'
+
+      - run: npm i
+      - run: npm run examples:ci
```
ryanio, line 20: I know you're probably just testing with this file, but when the time is right I can help add the "npm run examples:ci" line to vm-pr.yml and vm-build.yml

Can we also do a test if the example fails or errors out that it causes the ci step to fail?

scripts/examples-runner.ts
```diff
@@ -0,0 +1,23 @@
+import {extname, join} from "path";
```
ryanio, line 1: can you run prettier on this file according to our .prettierrc rules at root

so here would be no semicolons and spaces between the curly braces

thanks, looks good otherwise!

.github/workflows/examples-works.yml
holgerd77: @ryanio I would assume that it is rather meant to be set up in this way by Cesar? 🤔

So, we have the options here to run the examples through this dedicated extra file or integrate in the existing CI files per package as Ryan mentioned. I have actually no strong opinion here, seems to work well in both ways.

holgerd77: Short nit: in case we keep here, can we remove to just `examples.yml`?

packages/vm/package.json
```diff
@@ -43,7 +43,8 @@
     "lint": "../../config/cli/lint.sh",
     "lint:fix": "../../config/cli/lint-fix.sh",
     "formatTest": "node ./scripts/formatTest",
-    "tsc": "../../config/cli/ts-compile.sh"
+    "tsc": "../../config/cli/ts-compile.sh",
+    "examples": "ts-node ../../scripts/examples-runner.ts -- vm"
```
holgerd77, line 47: I guess this can be removed?

holgerd77 (approved):
Hi Cesar,
this looks already pretty great 👍, some comments but nothing too major.

Can you do at least one test on CI pushing a non-working example and see if the CI really breaks (maybe do a short screenshot)? That would be a good test, not that this always passes no matter what happens. 😋

.github/workflows/examples-works.yml
cbrzn: Indeed this was just for testing purposes - I was not sure where to add the script example

> add the "npm run examples:ci" line to vm-pr.yml and vm-build.yml

At first, I thought of doing so, but with this script we aim to run all existing examples in the repository, aiming at different packages

Personally, I would go with this dedicated script to run all the examples but open to suggestions :-D

scripts/examples-runner.ts
cbrzn: Sure will do! Also will configure the formatted of my editor to use the monorepo config

package.json
```diff
@@ -29,6 +29,7 @@
     "preinstall": "npm run checkNpmVersion",
     "postinstall": "npm run build --workspaces",
     "checkNpmVersion": "./scripts/check-npm-version.sh",
+    "examples:ci": "CI=true npm run examples --workspaces --if-present",
```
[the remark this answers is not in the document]
cbrzn: Negative, I just added it before so we do **not** show `console.log`'s when running the example script from the root - So we remove unnecessary logs on CI

Will remove it and add it in upcoming PRs

.github/workflows/examples-works.yml
cbrzn: Hey guys after thinking a bit more about this I think I will move forward with creating the `examples.yml` file in the workflow folder - Let me know if we should proceed otherwise

cbrzn:
Btw here's an action failing if it throws an error https://github.com/ethereumjs/ethereumjs-monorepo/runs/4909269302?check_suite_focus=true#step:5:23 @ryanio @holgerd77

holgerd77 (approved):
LGTM, thanks a lot Cesar! 😄 

Will merge now.

PR #1660, opened 2022-01-22
VM, Client: Remove ProofStateManager interface

holgerd77:
Part of #1533
(to be done towards `master` since generic and relevant for minor releases)

I had some doubts on the ProofStateManager for some time since this would add some additional complexity on the StateManager interface collection. Now I directly stumbled upon this on the start of trying to add a VerkleTrie StateManager.

So I wouldn't want (couldn't) implement this functionality in such a state manager. But then in these RPC code parts of the client I would need some somewhat ugly `any` casts (and then do this functionality check anyhow) to get this working.

So I think this is a somewhat more flexible and practical solution - to add these as optional methods to the generic interface. Then TypeScript knows that this functionality might be there (people who use this would as well) and one is generally incentiviced to put some guard around that.

An alternative would be to generally keep this a bit more under radar and not expose at all atm (since this is very much an "extra" functionality) and for now in our single place own internal usage for now go with the `any` cast + method check solution.

Then this would cause no friction/confusion for anyone at all. I am somewhat more inclined to this solution here, but not a totally strong opinion.

### DIFF
--- a/packages/client/lib/rpc/modules/eth.ts
+++ b/packages/client/lib/rpc/modules/eth.ts
@@ -31,7 +31,7 @@ import type {
   TxReceipt,
 } from '@ethereumjs/vm/dist/types'
 import type { Log } from '@ethereumjs/vm/dist/evm/types'
-import type { Proof, ProofStateManager } from '@ethereumjs/vm/dist/state'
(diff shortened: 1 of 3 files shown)


ryanio (approved):
ok, sounds good!

PR #1596, opened 2021-12-06
vm: verkle trees demo

gabrocheleau:
This PR implements a basic `VerkleStateManager` that makes use of the the [verkle-trees rust library](https://github.com/crate-crypto/rust-verkle-wasm). 

Related to #1578

## Tasks:

- [x] Add the verkle-trees library compiled WASM to the vm package
- [x] Scaffold a VerkleStateManager in the vm package (copied from stateManager, no implementation, only renaming)
- [ ] Implement the VerkleStateManager manager methods
- [ ] Test!

Issue #1578, opened 2021-11-18
VM: Implement Verkle Trie StateManager


holgerd77:
Part of #1533
Towards branch: `#verkle-stateless-demo`

To be able to verify a state root from a "verkle" block we need to be able compute a new (verkle) state root by running all txs from a block in the VM (as we do now) and feeding the state changes to a verkle trie (instead of a merkle trie) which then updates along and at the end (of a block) - hopefully gets to a new state root which matches the verkle state root of the block.

For this to work we need a special `VerkleStateManager` which feeds account and contract changes to the (a) verkle trie. Base work on this has been done in #1548 where a `BaseStateManager` has been implemented taking out all non-backend (MPT) related functionality from both the `DefaultStateManager` and the backend `Cache`. So it should now be possible to only adopt the remaining functions in the `DefaultStateManager` to get this working.

To reduce on complexity this work should be done without considering any proof related functionality. So this is mainly to connect an appropriate form of `WASM` compilation of the [rust-verkle](https://github.com/crate-crypto/rust-verkle) library and call into specific functions like [insert()](https://github.com/crate-crypto/rust-verkle/blob/master/verkle-trie/src/trie.rs#L113) or [get()](https://github.com/crate-crypto/rust-verkle/blob/master/verkle-trie/src/trie.rs#L453) from the `VerkleStateManager()`.

If the data backend (leveldb) is a problem on `WASM` compilation it is also possible to take a very simple backend as a data store (e.g. just some Rust dictionary or something which compiles easily to WASM) as some simplification. This doesn't need to store a lot of data for the demo, just some selected pre-state of a block and then the additions from the block transactions.

And, as a third simplification and a suggestion where/how to start: on a first round we can also ignore the state root itself (if we haven't e.g. a block with a verkle state root and a corresponding state test data setup available) and try to get e.g. mainnet addition of the genesis block and execution of block 1 running, ignoring the state root output (also ok to e.g. comment out the state root checks on a test run). This would already mean to have a lot going with the VM tx execution putting data into the verkle trie, being able to complete on a tx execution and - optimally - finishing up on the whole block. 🙂 

Ok, so far as a first TODO write-up on the Verkle Trie demo network.

Ah, and added a new branch `verkle-stateless-demo` (with appropriate checks), work can be done towards this branch.

//cc @kevaundray

holgerd77:
[...]

PR #1661, opened 2022-01-22
VM: add StatelessVerkleStateManager

holgerd77:
Part of #1533 
Implements #1578
Target branch: #verkle-stateless-demo

This continues #1596, so to create a dedicated stateless Verkle StateManager which takes in a pre-state from a block, then executes in a stateless way upon this pre-state and at the end provides the computed state changes to `js_verify_update` function from the [WASM build](https://github.com/crate-crypto/rust-verkle-wasm) of the rust verkle library.

WIP


holgerd77:
Ok, I think I've gotten pretty far here.

I've implemented most methods of a stateless verkle state manager, all based on a generic pre-state structure containing accounts, code and storage. This should hopefully be easily adoptable to a specific proof format later on.

I have added various test cases, these pass for a simple transfer tx as well as a contract tx for the stateless cases. Other methods are also already somewhat sufficiently tested.

Finally I integrated this new alternative state manager into the state test runner which can be run with the `--stateless` option. This then generates a pre-state from the test data and runs the VM with a `StatelessVerkleStateManager`. Note that the state root is not compared on these runs. Nevertheless this minimally shows if things pass under a variety of execution conditions.

This works astonishingly well already. I've integrated this in CI here - see the [vm-stateless](https://github.com/ethereumjs/ethereumjs-monorepo/runs/4908183475?check_suite_focus=true) and a lot of tests pass (execute) here already.

So far. This should already some pretty solid basis for some continued work. 🙂

holgerd77:
Closing this in favor of a PR directly targeted against `master`. I have already done such a broad range of work here that we can use this branch as the base branch I guess and it's a bit overkill to always update an underlying `verkle-stateless-demo` branch (will delete, no commits yet).

Eventual subsequent work on this base work might still be done in dedicated PRs towards `vm-add-stateless-verkle-statemanager`.

PR #1663, opened 2022-01-25
client: move execution from full sync

g11tech:
As discussed with @ryanio over discord, Execution would be better placed inside Ethereum client than inside Full sync. This PR aims to attain the same without modifying any call flow.
Next step would be do not pass execution inside the full sync and let client bind the event to run execution once the sync fetches some blocks from peer.


packages/client/lib/config.ts
```diff
@@ -17,6 +17,11 @@ export enum DataDirectory {
   Meta = 'meta',
 }
 
+export enum SyncMode {
+  full = 'full',
+  light = 'light',
```
ryanio, line 22: I think we should use PascalCase for enum keys here as we do in other parts of the monorepo (lowercase or camelCase for the values is fine)

```suggestion
  Full = 'full',
  Light = 'light',
```

Also, what do you think about moving this enum definition to `lib/types.ts`?

packages/client/lib/client.ts
```diff
@@ -54,8 +55,11 @@ export interface EthereumClientOptions {
  */
 export default class EthereumClient {
   public config: Config
+  public chain: Chain
 
-  public services: (FullEthereumService | LightEthereumService)[]
+  public node:
```
ryanio, line 60: We've previously used the word "node" before (actually it was what EthereumClient was called before it was renamed, if I remember correctly) so I'd prefer if we can find a more descriptive word here, since "node" is a bit meaningless without more context.

syncmode should already be available in config, so I'm not sure if we need it again. Could you explain your reasoning for combining these properties?  Should we just add "execution" as an optional property on the top level of EthereumClient? Then "services" can also remain as it is.

packages/client/lib/config.ts
g11tech: makes sense :slightly_smiling_face: :+1:

packages/client/lib/client.ts
g11tech: ummm we can use any word other from node as well thats not an issue, basically idea was to use typescript to couple execution with FullEthereumService, syncmode can be dropped. Also a Q, do we need services to be an array?

ryanio: what if we moved execution to FullEthereumService? (i think i remember you suggesting that at first?) because there would never be Light+Execution, so that only leaves Full+Execution

services should be an array to support a full client syncing light peers, unless light sync doesn't go through there, then we could think about it.

packages/client/lib/config.ts
ryanio: actually we can keep it in this file, makes sense here too 👍

packages/client/lib/client.ts
g11tech: I think in that case, we can just move execution and services out, as services could be a mix (of light or full) as per your original design , I still would like to keep execution here in client because extracting execution out of services (for e.g. in rpc module) feels a bit weird.(and FullEthereumService still enforces execution in its constructor)

ryanio: nice, cool, just took a look - looks clean!

packages/client/lib/sync/fullsync.ts
```diff
@@ -223,6 +176,43 @@ export class FullSynchronizer extends Synchronizer {
...
+    this.config.logger.info(
+      `Imported blocks count=${
+        blocks.length
+      } number=${first} hash=${hash} ${baseFeeAdd}hardfork=${this.config.chainCommon.hardfork()} peers=${
+        this.pool.size
+      }`
+    )
+    this.txPool.removeNewBlockTxs(blocks)
+
+    if (this.running || this.config.chainCommon.gteHardfork(Hardfork.Merge)) {
```
ryanio, line 210: I think we can remove this here now `this.config.chainCommon.gteHardfork(Hardfork.Merge)` since in engine.ts I manually do `execution.run()`. it was an artifact of having execution as part of sync, but now that we don't, it would no longer be needed

g11tech: I guess, since beacon sync/optimistic will have its own synchronizer, it makes sense :+1:

ryanio:
@g11tech thank you for this wonderful PR! it looks great. i just had a few small nits from reviewing I'll push a commit. I tested syncing on all major networks (mainnet, ropsten, rinkeby, goerli, sepolia) and was able to successfully sync and execute a few thousand blocks from each.

what a big improvement to the architecture of the client. super excited for this. thanks again!

ryanio (approved):

PR #1665, opened 2022-01-26
VM, devp2p: Remove Node.js specific assert usage

ScottyPoi:
Removed all uses of assert in favor of Error to match changes in   #1638

### DIFF
--- a/packages/vm/src/evm/precompiles/01-ecrecover.ts
+++ b/packages/vm/src/evm/precompiles/01-ecrecover.ts
@@ -1,10 +1,11 @@
 import { setLengthLeft, setLengthRight, ecrecover, publicToAddress, BN } from 'ethereumjs-util'
 import { PrecompileInput } from './types'
 import { OOGResult, ExecResult } from '../evm'
-const assert = require('assert')
 
 export default function (opts: PrecompileInput): ExecResult {
-  assert(opts.data)
+  if (!opts.data) {
+    throw new Error('opts.data is undefined')
+  }
 
   const gasUsed = new BN(opts._common.param('gasPrices', 'ecRecover'))
 
--- a/packages/devp2p/src/dns/dns.ts
+++ b/packages/devp2p/src/dns/dns.ts
@@ -1,4 +1,3 @@
-import assert from 'assert'
 import { PeerInfo } from '../dpt'
 import { ENR } from './enr'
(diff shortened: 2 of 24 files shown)


packages/vm/src/evm/precompiles/01-ecrecover.ts
```diff
@@ -1,10 +1,11 @@
 import { setLengthLeft, setLengthRight, ecrecover, publicToAddress, BN } from 'ethereumjs-util'
 import { PrecompileInput } from './types'
 import { OOGResult, ExecResult } from '../evm'
-const assert = require('assert')
 
 export default function (opts: PrecompileInput): ExecResult {
-  assert(opts.data)
+  if (!opts.data) {
+    throw new Error('opts.data is undefined')
```
holgerd77, line 7: I've written a long thread on our current error management which also touches the current problem field of changing error msgs, see here: https://github.com/ethereumjs/ethereumjs-monorepo/pull/1469#issuecomment-1023016428 (where we likely want to improve on this round of breaking releases).

But since these error messages haven't been before anyhow: can we directly take something a bit more high-level and expressive here? Something like `No data provided for the ECRECOVER precompile` (feel free to improve).

holgerd77: (same for the other precompiles of course)

holgerd77 (changes requested):
This looks great, thanks Scotty! 😄 

I have had a close look on all the negation of all the conditional checks, since this would likely be the most probable place for overlooking some stuff, all looks good though.

One small comment, there is also an API test in the VM which needs to be fixed (still an AssertError comparison if I read correctly).

Otherwise we can merge here I would say.

holgerd77:
Note: I am just in the process of recreating the `develop` branch with a clean commit history now always using rebases to get on top of the `master` branch, see post in the internal "#BigInt Conversion" channel.

I will close this PR here, have done a branch copy locally as a backup just in case.

I will then delete and repush the `develop` branch, then under a fitting backup name, and repush a "new" `develop` branch for continued development with the existing squashed merge PR commits cherry-picked.

Please reopen this PR towards this branch.

Thanks! 🙂

holgerd77:
Ok, have recreated the `develop` branch, feel free to reopen now.

PR #1667, opened 2022-01-26
ci: add concurrency

ryanio:
This PR adds a new feature recently added to GitHub Action workflows, concurrency: https://docs.github.com/en/actions/using-jobs/using-concurrency

This allows us to automatically cancel previous runs for the same PR, rather than manually doing it or waiting for the outdated runs to finish.

To test, I will push a commit right after this to ensure that the behavior works as expected.

### DIFF
--- a/.github/workflows/block-build.yml
+++ b/.github/workflows/block-build.yml
@@ -13,6 +13,10 @@ defaults:
   run:
     working-directory: packages/block
 
+concurrency:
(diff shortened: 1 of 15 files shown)


ryanio:
looks like it worked! after pushing a new commit the previous runs were cancelled


Screen Shot 2022-01-26 at 11 12 47 AM

holgerd77 (approved):
Great stuff, can't see a reason why this shouldn't be merged, so will do. 🙂

PR #1670, opened 2022-01-27
VM: Examples simplified

cbrzn:
This PR aims to improve the examples in the VM package - This is not finished yet but it's the first iteration so I can get some feedback and make sure this is what we want. I have removed the folders in favor of creating files per example, this will allow us to run the examples in the CI in an easier way - Also, I have moved the dependencies to `devDependencies` 

I think that we can add examples of different transaction types (the three that exists) and eventually we can use this as a playground to test transactions with Kintsugi fork (It's what I am trying to achieve locally after successfully running the nodes) - Open to suggestions on what I should add here

Another thing I noticed is that in the `Message` [constructor](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/evm/message.ts#L19) we have any type, allowing a user to **not** pass the `caller`, hence, making the [`executeCall`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/aa7d323a1bddc5e1250c6861402c733037f32fa2/packages/vm/src/evm/evm.ts#L232) method break on runtime. Imo this can be fixed in another PR but just saying this in case I am missing something


holgerd77:
Hi Cesar, cool, maybe you can break down your work in a couple of more commits next time - then it will get a bit easier to review - but otherwise this looks pretty good already! 😄 

I guess I'm ok with placing these some dependencies in the `devDependencies` on first thought, if no one else has reservations I would be ok if this gets merged on a continued review regarding this specific aspect.

Not sure, do we want these examples to be deployed on npm releases as well (first thought: why not, not much code and would be nice to get a start)? Then we would need to add the `examples` folders to the `package.json` file (let's wait for some more opinions here). This would also go along with some notes how to run in the main `README.md` files (both on a cloned repo as well as an installed version (need hint on `devDependencies` to be installed, so no `--production` flag). I can also add this section along the including release though, then this wouldn't confuse anyone.

Haven't looked more closely at the helper functions, but I like that this is - for now - located locally in the VM folder.

cbrzn:
Hey @holgerd77 
> maybe you can break down your work in a couple of more commits next time

Will do! Sorry for that

> do we want these examples to be deployed on npm releases as well

I think we should not add these examples in the production build, since this is just to play with it, if a user wants to get more familiar with it he can download the repository locally instead of doing it through an exported function of the package (If I understand your point correctly)

On another point, the `helpers` folder has been created to add the shared functions that are being used only by the examples on this package. I think that every example folder can have its own custom helpers folder, making it somehow cleaner (IMO)

holgerd77:
> I think we should not add these examples in the production build, since this is just to play with it, if a user wants to get more familiar with it he can download the repository locally instead of doing it through an exported function of the package (If I understand your point correctly)

Ah, no, definitely not expose on the API level but rather only "include", so that people would be able to run the script when installing the package.

Update: ah, on thinking about it, that's not so straight-forward anyhow with the path being in `node_modules`, right?

So I agree with you, let's just not follow this path further and only have these prepared for people cloning the repo.

cbrzn:
Hey @holgerd77 !

Sure, will update the `package-lock`. Also, I added the WIP and needs review labels because I wanted you to check it before moving forward, but now that I am re-thinking this I feel that it's in a state where it can be merged and when tackling https://github.com/ethereumjs/ethereumjs-monorepo/issues/1675 I can add some more examples (I am sure that issue will open my mind and will allow me to think in even more complex examples :-D)

holgerd77:
Thanks, yes, I would also say we should merge this here after review and then tackle more work in another PR.

holgerd77:
This `run-tx-complete` example: did you remove and if so what is the reasoning?

holgerd77 (approved):
Nice, LGTM, thanks Cesar! 🙂

PR #1671, opened 2022-01-27
VM: Replace BN.js with bigints, Util helpers

acolytec3:
Replaces BN.js with bigints across the VM and updated compilation targets for VM and util to ES2020
Implementation details:
- Replaces BN.js with bigints in all calculations across all VM internals (`evm`/`StateManager`/etc.).  
- All VM exposed APIs now require `bigint` inputs instead of `BN.js` and numeric outputs in `execResult` and similar VM outputs are passed as `bigint`s instead of `BN`s
- Adds new helpers in `util` to convert BNs to bigints, buffers, and vice-versa.
- Upgrades libraries to target ES2020 to support `bigint` usage
- All BNs are currently converted to `bigints` when passed from other packages into VM API (e.g. from `tx` and `block` objects in the State Manager)


packages/util/src/constants.ts
```diff
@@ -14,14 +14,27 @@ export const MAX_INTEGER = new BN(
...
 export const TWO_POW256 = new BN(
   '10000000000000000000000000000000000000000000000000000000000000000',
   16
 )
 
+/**
+ * 2^256
+ */
+export const TWO_POW256_BIGINT =
+  BigInt(0x10000000000000000000000000000000000000000000000000000000000000000)
```
ryanio, line 36: cool, could we just do `2n ** 256n` here?

packages/util/src/constants.ts
```diff
@@ -14,14 +14,27 @@ export const MAX_INTEGER = new BN(
...
 
+/**
+ * The max integer that the evm can handle (2^256-1) as a bigint
+ */
+export const MAX_INTEGER_BIGINT = 2n ** 256n - 1n
+
 /**
  * 2^256
+ *
+ * @deprecated
```
ryanio, line 25: maybe along with a note
```suggestion
 * @deprecated bn.js constants are deprecated, please use the newly introduced bigint constants
```

packages/vm/src/buildBlock.ts
```diff
@@ -132,11 +132,11 @@ export class BlockBuilder {
    * Calculates and returns the receiptTrie for the block.
    */
   private async receiptTrie() {
-    const gasUsed = new BN(0)
+    let _gasUsed = 0n
```
ryanio, line 135: nice, i'd say we don't need the underscore?

acolytec3: Do you even use gasUsed there anymore?  It looks like it's an unused variable.

ryanio: ah hm good point, you're right it doesn't look like it's being used! ok it can be removed then :)

acolytec3:
One thing that is definitely broken is the EXP function handler.  It seems to work in most instances but in some cases it's producing values out of range.  Since I don't really understand the math that bn.js uses to do exponentiation, I can't comment on how the bigint version is different or why it is producing incorrect results for some state tests

packages/util/src/constants.ts
holgerd77: Hmm, is this wise if this always have to be recalculated? 🤔

ryanio: that's a good question, if the calculation is run every time the constant is imported or if it is calculated once on init of the package.

To test:

constants.js
```js
const TWO_POW256_BIGINT_HEX =
  BigInt(0x10000000000000000000000000000000000000000000000000000000000000000);

const TWO_POW256_BIGINT_EXP = 2n ** 256n;

exports = [TWO_POW256_BIGINT_HEX, TWO_POW256_BIGINT_EXP];
```

index.js
```js
const Benchmark = require("benchmark");
const BN = require("bn.js");

const suite = new Benchmark.Suite();

suite
  .add("bigint init with hex", function () {
    const { TWO_POW256_BIGINT_HEX } = require("./constants");
    if (100n > TWO_POW256_BIGINT_HEX) return true;
  })
  .add("bigint init with exp", function () {
    const { TWO_POW256_BIGINT_EXP } = require("./constants");
    if (100n > TWO_POW256_BIGINT_EXP) return true;
  })
  .on("cycle", function (event) {
    console.log(String(event.target));
  })
  .on("complete", function () {
    console.log("Fastest is " + this.filter("fastest").map("name"));
  })
  .run({ async: true });
```

On my Macbook Air (M1, 2020) my results are always about even:

```
bigint init with hex x 6,561,811 ops/sec ±0.37% (97 runs sampled)
bigint init with exp x 6,573,137 ops/sec ±0.10% (99 runs sampled)
Fastest is bigint init with exp,bigint init with hex
```

I'm not quite sure if this exactly replicates our util setup, but thought I'd try it.

Let's compare this to inline definitions and not importing (snippet):

```js
const TWO_POW256_BIGINT_HEX =
  BigInt(0x10000000000000000000000000000000000000000000000000000000000000000);
const TWO_POW256_BIGINT_EXP = 2n ** 256n;

suite
  .add("bigint init with hex", function () {
    if (100n > TWO_POW256_BIGINT_HEX) return true;
  })
  .add("bigint init with exp", function () {
    if (100n > TWO_POW256_BIGINT_EXP) return true;
  })
```

Much faster 😄

```
bigint init with hex x 110,702,702 ops/sec ±0.12% (101 runs sampled)
bigint init with exp x 110,732,705 ops/sec ±0.12% (94 runs sampled)
Fastest is bigint init with exp,bigint init with hex
```

One last comparison, with requiring but only once in the file header (snippet):

```js
const { TWO_POW256_BIGINT_HEX, TWO_POW256_BIGINT_EXP } = require("./constants");

suite
  .add("bigint init with hex", function () {
    if (100n > TWO_POW256_BIGINT_HEX) return true;
  })
  .add("bigint init with exp", function () {
    if (100n > TWO_POW256_BIGINT_EXP) return true;
  })
```

Results:
```
bigint init with hex x 85,243,817 ops/sec ±0.41% (90 runs sampled)
bigint init with exp x 86,028,773 ops/sec ±0.19% (97 runs sampled)
Fastest is bigint init with exp
```

ryanio:
I was curious about BigInt conversion performance with the varying ways to do it, so I put together a small benchmarking setup, my results in the first comment: https://gist.github.com/ryanio/89cc842a7573dd077cbb7cfa4469f8ec

jochem-brouwer:
Some small stuff I noticed, which we should look at (will edit this if I find more)

- [ ] `npm i --workspaces` fails on build
- [ ] Browser build are failing on some packages
- [ ] Coding style should be consistent, for instance `gas = gas + 2n` and ` gas += 2n` are used mixed

packages/vm/src/evm/opcodes/util.ts
```diff
@@ -258,6 +263,50 @@ export function updateSstoreGas(
...
+}
+
+export function abs(a: bigint) {
+  if (a > 0) {
+    return a
+  }
+  return a * -1n
+}
+
+export function bufferToBigInt(buf: Buffer) {
```
ryanio, line 293: would this be better placed in ethereumjs-util?

acolytec3: I'm not opposed to it.  I was just trying to limit the amount of change outside the VM package when I wrote that.

acolytec3:
Okay, I think this bigint PR is now ready for review at long last.  All the tests are passing (except Hardhat which isn't surprising).  Question for the group, do you want me to figure out which values Hardhat is failing on (looks like there are VM bigints that it's trying to reference) and put a BN wrapper around them or do we want to review as is, potentially merge, and then continue on with a bigint conversion across the whole monorepo.

ryanio:
it's okay if hardhat isn't passing on `develop` since it's the breaking changes, which is expected

packages/vm/src/evm/opcodes/util.ts
ryanio: i suggest we add a few functions to ethereumjs-util that you do repeatedly across this PR. then let's use them in all the code, so for example if we find a faster way to do the function, then we automatically get the updated benefits across all packages.

to start, I am thinking `bufferToBigInt` and `bigIntToBuffer`

I would suggest also adding `bnToBigInt` and `bigIntToBN` but if we plan to remove bn.js completely then they would also become unsupported. For now bn -> BigInt can just do in one style so it's easier to search/replace later, maybe stick to `BigInt(bn.toString(10))` for consistency.

packages/vm/src/buildBlock.ts
```diff
@@ -228,7 +227,7 @@ export class BlockBuilder {
     const transactionsTrie = await this.transactionsTrie()
     const receiptTrie = await this.receiptTrie()
     const logsBloom = this.logsBloom()
-    const gasUsed = this.gasUsed
+    const gasUsed = new BN(this.gasUsed.toString(10), 10)
```
ryanio, line 230: would suggest leaving the default values out when unneeded, i.e.
```suggestion
    const gasUsed = new BN(this.gasUsed.toString())
```

packages/vm/tests/api/EIPs/eip-2929.spec.ts
```diff
@@ -51,20 +51,20 @@ tape('EIP 2929: gas cost tests', (t) => {
...
     })
 
     const tx = unsignedTx.sign(senderKey)
 
     const result = await vm.runTx({ tx })
 
-    const totalGasUsed = initialGas.sub(currentGas)
-    st.equal(true, totalGasUsed.eq(new BN(test.totalGasUsed).addn(21000))) // Add tx upfront cost.
+    const totalGasUsed = initialGas - currentGas
+    st.equal(true, totalGasUsed === BigInt(test.totalGasUsed) + 21000n) // Add tx upfront cost.
```
ryanio, line 63:
```suggestion
    st.ok(totalGasUsed === BigInt(test.totalGasUsed) + 21000n) // Add tx upfront cost.
```

packages/vm/src/evm/opcodes/util.ts
acolytec3: For the `bigIntToBuffer` route, does it make more sense to just add to `toBuffer`?

ryanio:
awesome, thanks for doing that!

let's move the three functions you added to `types.ts` over to `bytes.ts`, all of them there sounds good to me

acolytec3:
> edit: I'm still seeing a handful of places where these new functions aren't used (big pr smile) do you want me add a commit to convert more of them over, or point them out in review comments, or you got them?

I'll take another skim through it.  I was just searching with VS Code so probably missed a couple of spots.  Will just go line by line.

acolytec3:
> let's move the three functions you added to `types.ts` over to `bytes.ts`, all of them there sounds good to me
> 
I'm not sure I follow the logic on this one.  These three are type conversion utilities between non `Buffer` types and we already have similar type converters for `intToHex`/`bnToHex` etc in the `types` section.  Happy to move but it seems like it's more fitting for them to stay in `types` unless we're also going to move all the other type converters to `bytes` too (and I could make a case for that I think.

ryanio:
good point, sounds good!

holgerd77:
> I'm not sure I follow the logic on this one. These three are type conversion utilities between non Buffer types and we already have similar type converters for intToHex/bnToHex etc in the types section. Happy to move but it seems like it's more fitting for them to stay in types unless we're also going to move all the other type converters to bytes too (and I could make a case for that I think.

Yes, these conversion function like `bnToHex()` in `types.ts` are totally misplaced and we should move, wonder if I have at some point already opened an issue on that (or at least mentioned somewhere some time ago)?

If you do please make sure to clearly document somewhere, otherwise this will likely be forgotten on the release notes at some point.

ryanio:
@acolytec3 mentioned on our stand up this is ready for final review and merging into develop, so he doesn't have to continually rebase it to keep it up to date. i will give it a final review today, @holgerd77 feel free to take a look as you can as well

holgerd77:
I'll try to also have a look into it.

Let's please wait another 7 days or so until we have had a minimal discussion on the topic with MetaMask. Sorry for the inconveniences, we'll likely be able to merge soon then and this we can already start to give this some thorough review in the meantime (so we should really give this some emphasis, this is a *huge* PR).

Let's please also give this dedicated question what to do with the other ES2020 features, that is still a bit a hard one: so if we enforce people to stay on ES2020, should we not use - and eventually remove - all other ES2020 features?

I asked Richard on that and he is actually doing this and I guess that would make a lot of sense (be needed) on our side as well:

grafik

So this would mean e.g. that we would need to remove all nullish coalescent operator usages.

paulmillr:
`**` has been supported [for a long time](https://caniuse.com/mdn-javascript_operators_exponentiation), since ES2016. What's the deal with it?

I've took a glance at all features es2015-2021: there are very few that could be useful. `??` operator, and maybe `a.?b` property chaining. It's easy to not use them. I think there needs to be a specific browser target for the libraries, then it's easy to make decisions.

holgerd77:
Ok, seems Ryan has some positive news from Metamask on BigInt support, so we should be ready to go here throughout the week! 🎉

Let's please give this a thorough review over some several days, so that we earliest merge on Thursday or so if this gets some continued attention until then.

@acolytec3 Some basic things:

1. I guess we then really do not want to use the `*n` notation, can you then switch this out here? Maybe you can also use some constants like `const _0n = BigInt(0)` as Paul does in his libraries? (not sure, would this be overkill to add some of these to Util)?
2. Are these tsconfig and the like adoptions somewhat consistent and final? //cc @ryanio 
3. Can you please provide a list in the initial comment with all the relevant things this PR does and/or changes? Thanks!

Yeah, cool, but totally exciting to see this find it's way in pretty likely.

Ah, maybe to finally asked (and maybe adding @jochem-brouwer in //cc on this): are the core-VM-related integrations/adoptions on this really settled or are there still some shaky things (on arithmetics or whatever)?

acolytec3:
> 1. I guess we then really do not want to use the `*n` notation, can you then switch this out here? Maybe you can also use some constants like `const _0n = BigInt(0)` as Paul does in his libraries? (not sure, would this be overkill to add some of these to Util)?

acolytec3:
> 1. I guess we then really do not want to use the `*n` notation, can you then switch this out here? Maybe you can also use some constants like `const _0n = BigInt(0)` as Paul does in his libraries? (not sure, would this be overkill to add some of these to Util)?
>
What's driving this decision?  I'm not going to fight it but I have really grown to like the `0n` notation over `BigInt(0)`.  The latter just feels really clunky to me when I've had to use it.

> 2. Are these tsconfig and the like adoptions somewhat consistent and final? //cc @ryanio

I think so, though would be happy to have the experts sign off on my tweaks.

> 3. Can you please provide a list in the initial comment with all the relevant things this PR does and/or changes? Thanks!

Will do.

paulmillr:
Benchmark comparison against bn:

- `develop` bn.js branch:

```
Block 9422905 x 55,277 ops/sec ±4.84% (82 runs sampled)
Block 9422906 x 52,745 ops/sec ±1.09% (91 runs sampled)
Block 9422907 x 44,853 ops/sec ±8.28% (79 runs sampled)
Block 9422908 x 40,065 ops/sec ±12.29% (73 runs sampled)
Block 9422910 x 27,322 ops/sec ±40.03% (61 runs sampled)
```

- `bigints` native branch:

```
Block 9422905 x 76,584 ops/sec ±3.21% (78 runs sampled)
Block 9422906 x 71,360 ops/sec ±1.06% (94 runs sampled)
Block 9422907 x 65,537 ops/sec ±6.04% (88 runs sampled)
Block 9422908 x 55,722 ops/sec ±7.63% (77 runs sampled)
Block 9422910 x 59,979 ops/sec ±1.69% (85 runs sampled)
```

Note that variance became reasonable after the move. `±40.03%` was very bad.

holgerd77:
| I'm not going to fight it but I have really grown to like the 0n notation over BigInt(0)

@acolytec3 yeah, sorry for that. We had a longer discussion around that along some call (I guess one you weren't attending or had to leave or something, guess we might want to communicate/protocol our call results a bit better? 🤔) and we agreed upon that, Richard made a strong pledge here basically for the same reasons Paul stated as well.

I also a fan of the *n notation, so let's switch as soon as circumstances allow. 🙂 But for now I guess backwards-compatibility does count more than our own convenience. Switching later on can be easily done with some regEx search-and-replace.

Did some search on your code base, you have ~180 occurrences of BigInt values and 110 or so of them are `0n` and 30 or so `1n`, so a lot can already be caught with this helper mentioned above.

| As for "moving _0n = BigInt(0) into util" - meh.

Hehe. 😛 Yes, so let's very well leave, already felt somewhat stronger on the overkill side along suggesting.

holgerd77:
> Benchmarked comparison against bn. Seems **1.5-2x**
> 
>     * `develop` bn.js branch:
> 
> 
> ```
> Block 9422905 x 55,277 ops/sec ±4.84% (82 runs sampled)
> Block 9422906 x 52,745 ops/sec ±1.09% (91 runs sampled)
> Block 9422907 x 44,853 ops/sec ±8.28% (79 runs sampled)
> Block 9422908 x 40,065 ops/sec ±12.29% (73 runs sampled)
> Block 9422910 x 27,322 ops/sec ±40.03% (61 runs sampled)
> ```
> 
>     * `bigints` native branch:
> 
> 
> ```
> Block 9422905 x 76,584 ops/sec ±3.21% (78 runs sampled)
> Block 9422906 x 71,360 ops/sec ±1.06% (94 runs sampled)
> Block 9422907 x 65,537 ops/sec ±6.04% (88 runs sampled)
> Block 9422908 x 55,722 ops/sec ±7.63% (77 runs sampled)
> Block 9422910 x 59,979 ops/sec ±1.69% (85 runs sampled)
> ```
> 
> Note that variance became reasonable after the move. `±40.03%` was very bad.

Ah, that's interesting, again - as mentioned - our benchmarks are not super reliable. But locally run they *do* have some notion of repeatability. Let's keep an eye on this, if these results hold that would be exciting.

packages/vm/src/evm/precompiles/02-sha256.ts
```diff
@@ -8,12 +8,11 @@ export default function (opts: PrecompileInput): ExecResult {
 
   const data = opts.data
 
-  const gasUsed = new BN(opts._common.param('gasPrices', 'sha256'))
-  gasUsed.iadd(
-    new BN(opts._common.param('gasPrices', 'sha256Word')).imuln(Math.ceil(data.length / 32))
-  )
+  let gasUsed = BigInt(opts._common.param('gasPrices', 'sha256'))
```
holgerd77, line 11: Not for this PR and a bit of a side note: but I guess these continued Common parameter reads (+ re-conversions) to BigInt all over again will really cost us a lot of performance (think this was in discussion before in some way? Also mentioned (in a somewhat different context) in Patricios performance low-hanging-fruits issue #775.

I guess a complete refactor (of the Common JSON files) will likely overload us (and others who are already using this format), but what we could (and should likely) really do is caching these values - as BigInt - in Common on first read and then always return these values from the cache (and: generally return BigInt values on `param()` calls and similar.

Does this make sense?

Maybe also something for @emersonmacro being the caching expert. 😋

ryanio: nice idea, just thinking off the top of my head, maybe we add an optional parameter to `param()` with `asBigInt = false`, then the signature could be something like:

```typescript
param(topic: string, name: string, asBigInt: true): bigint
param(topic: string, name: string, asBigInt = false): number
```

another option would be to add a new method e.g. `paramAsBigInt(topic: string, name: string): bigint`

would be great to see if an internal cache would help with speedup too

acolytec3:
@holgerd77 @ryanio @jochem-brouwer This PR has been rebased on current `develop` with all `0n` and similar usage switched to `BigInt(0)` and the like.  I've also removed a few extraneous commits that were hanging around that got folded in from previous PRs on develop that shouldn't have been there so I think this is ready for final review at this point.

packages/vm/src/runTx.ts
```diff
@@ -505,7 +513,9 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
   state.clearOriginalStorageCache()
 
   // Generate the tx receipt
-  const cumulativeGasUsed = (opts.blockGasUsed ?? block.header.gasUsed).add(results.gasUsed)
+  const gasUsed = opts.blockGasUsed ?? bnToBigInt(block.header.gasUsed)
```
paulmillr, line 516: `??` should not be used as per #1729

packages/vm/src/evm/precompiles/05-modexp.ts
```diff
@@ -1,74 +1,77 @@
...
-function expmod(B: BN, E: BN, M: BN): BN {
-  if (E.isZero()) return new BN(1).mod(M)
-  // Red asserts M > 1
-  if (M.lten(1)) return new BN(0)
-  const red = BN.red(M)
-  const redB = B.toRed(red)
-  const res = redB.redPow(E)
-  return res.fromRed()
+export function expmod(a: bigint, power: bigint, modulo: bigint) {
+  if (power === BigInt(0)) {
```
paulmillr, line 65: This is performance-concious function, I think BigInt(0) and BigInt(1) can be assigned to _0n, _1n to not do the type casts over and over.

packages/vm/src/evm/opcodes/util.ts
```diff
@@ -258,6 +263,42 @@ export function updateSstoreGas(
...
+
+export function abs(a: bigint) {
+  if (a > 0) {
+    return a
+  }
+  return a * BigInt(-1)
+}
+
+const N = BigInt(115792089237316195423570985008687907853269984665640564039457584007913129639936)
+export function exponentation(bas: bigint, exp: bigint) {
```
paulmillr, line 294: Is this `modExp` again? Why two duplicate methods? Also I think it makes sense to use `require('ethereum-cryptography').CURVE.n` if vm has access to it somehow.

paulmillr (approved):
Reviewed all 100 files - looks good other than the comments.

holgerd77:
Ok, since this got another review from Paul I will merge this in now, then the subsequent BigInt (and other) work can be done on top (please: do by rebase 🙂).

Thanks again @acolytec3 for this great work here and @jochem-brouwer and @paulmillr for the assistance and guidance on some things.

Really cool! 🎉
Let's go BigInt. 🚀

holgerd77 (approved):
LGTM.

packages/vm/src/evm/opcodes/util.ts
acolytec3: It's not intended to be.  I followed the 2^k-ary method [here](https://www.geeksforgeeks.org/exponential-squaring-fast-modulo-multiplication/) for doing exponentiation.  And the `vm` package doesn't currently have `ethereum-cryptography` as a dependency so we don't currently have access to the constant you referenced.  We could add it for develop but I'm not sure if it makes sense to add a new dependency just for that.  If/when we pull the `ethereum-cryptography` re-exports out of `util` and reference them directly, I imagine we will pull that dependency into the VM and could then reference it there.

packages/vm/src/evm/precompiles/05-modexp.ts
acolytec3: Just to clarify my understanding, are you suggesting we cache the `BigInt(0)`/`BigInt(1)` just for this function to maximize performance or do it across the VM?  I thought we had agreed to not bother with the `_0n`/`_1n` constants and stick with the BigInt constructor everywhere.  I'm happy to go back and reverse the previous thought for `_0n` if there is enough of a performance differential between caching it and referencing vs just `BigInt(0)` but I wasn't aware there was any material difference there.

packages/vm/src/runTx.ts
holgerd77: I guess we might find a better solution here, see comment from @ryanio on #1729

packages/vm/src/evm/precompiles/05-modexp.ts
holgerd77: Wanted to do the merge now so follow-up work can be done, maybe this can be done along a follow-up PR since not so critical.

holgerd77: @acolytec3 ah, totally forget, yes I guess the *n -> BigInt(*) conversion still needs to be done. 😬 (or might there be some other solution? The TypeScript compiler will likely not do this? 🤔

Just tested on performance, there is indeed some substantial difference (but in a very low range), so might be worth to do on occasions where there is a potentially great round of repetitions:

```shell
> console.time('1');for (let i=0; i<= 1000000; i++) { let num = BigInt(0) }; console.timeEnd('1');
1: 18.209ms
undefined
> console.time('1');const _0n = BigInt(0); for (let i=0; i<= 1000000; i++) { let num = _0n }; console.timeEnd('1');
1: 1.449ms
undefined
```

holgerd77: Update: ah, sorry sorry sorry, just seeing that you already did. Thanks a lot!

packages/vm/src/evm/precompiles/02-sha256.ts
holgerd77: Do you think this is necessary to split this up? 🤔

Don't we need most of this as BigInt anyhow and should therefore go full-bigint with the function? Not sure, might need some confirmation first.

Another note: I guess I gave the wrong context for the caching suggestion from Patricio. If I recall correctly he rather meant caching in the VM context (e.g., but also mainly) to avoid this repeated calling into the `Common` functions.

Nevertheless: a number -> BigInt cache in Common itself might independently also worth to investigate.

ryanio: yes we could just return bigint, i think in some cases a number is fine so didn't want to do extra work and then have to additionally call `toNumber()`, but maybe we can do a rough count and see how often bigint vs number is used, and if number is only a few non-essential cases we can go for bigint only.

holgerd77: Have done some rough search on this (all monorepo, not only VM), this is extremely strongly on the BigInt/BN usage side, so I would roughly say something like 10:1.

(I would be calling occurences with these *n() functions from BN in - so like `addn()` - that makes sense, right?)

ryanio: > (I would be calling occurences with these *n() functions from BN in - so like `addn()` - that makes sense, right?)

not really sure what you mean. the search would be for `common.param` wrapped with `new BN()`

holgerd77: I am also calling occurrences in where a value is taken for an input for e.g. the `addn()` method, since after the conversion to BigInt this kind of arithmetic would be implemented using two BigInt values. So then it would also make sense to have a BigInt return value for these kind of situations.

Issue #1446, opened 2021-09-01
Difference between gasUsed and execResult.gasUsed


yann300:
the `afterTx` event (link below) used to contain 2 different values for the gas consumption (tx cost and exec cost): `{ gasUsed }` and `{ execResult: { gasUsed } }` . The latter seems now to also return the tx cost.

https://github.com/ethereumjs/ethereumjs-monorepo/blob/587ee70ae9b21cfe9c0eb438d946248382468c31/packages/vm/src/runTx.ts#L473

ryanio:
Oops, I remember this TODO actually, I should have more carefully looked into it rather than hastily changing: https://github.com/ethereumjs/ethereumjs-monorepo/pull/1375/files#diff-b48b85860e898fbfe1fec61ec4e174d679faf07ec106c31a847c8dad7f7d5adeL386-L387

Sorry about that. I need to dig in to understand this better. I can also do a patch release to fix, and I can add a test too to ensure it doesn't happen again. I guess none of our existing tests checks this value?

I also noticed that the debug output is outputting the same variable instead of execResult.gasUsed: https://github.com/ethereumjs/ethereumjs-monorepo/blob/587ee70ae9b21cfe9c0eb438d946248382468c31/packages/vm/src/runTx.ts#L361-L362

holgerd77:
Just for the record: the mentioned PR #1375 has been merged on July 25, so there are the two VM releases `v5.5.1` and `v5.5.2` affected by this.

holgerd77:
Update: oh, just stricing the above comment through again, I was writing based on the un-edited text I received via mail.

@ryanio so, just for the (updated) record 😄: you came to the conclusion after further analysis that #1375 didn't introduce this as a new bug?

ryanio:
> @ryanio so, just for the (updated) record 😄: you came to the conclusion after further analysis that #1375 didn't introduce this as a new bug?

I thought it might have, but the line I was considering ([here](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1375/files#diff-b48b85860e898fbfe1fec61ec4e174d679faf07ec106c31a847c8dad7f7d5adeL386-L387)) was actually `gasRefund` not `gasUsed`.

It could still be introduced by that PR, but it's not immediately obvious to me yet.

One common cause is that It could be a mutated BN we need to clone that's getting accidentally modified.

holgerd77:
[...]

PR #1677, opened 2022-02-01
vm: improve `runTx` and `evm` debug logs

ryanio:
This PR improves the runTx debug logs as investigated in #1446 and https://github.com/ethereumjs/ethereumjs-monorepo/issues/1446#issuecomment-1026463341


acolytec3 (approved):
LGTM!  Not sure why the code coverage is complaining but this shouldn't need new test coverage since the changes are limited to debug messages.  @holgerd77 thoughts on whether we could admin merge this given above?

acolytec3:
image

Was able to find a test where the RETURN opcode is used and it populates that `execResult.returnValue` field.  It's all 0s but at least we can see it being populated in the revised logger message.  LGTM!

acolytec3 (dismissed):
LGTM!

packages/vm/src/runTx.ts
```diff
@@ -395,22 +395,19 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
...
     debug(
-      `Received tx results gasUsed=${results.gasUsed} execResult: [ gasUsed=${
-        results.gasUsed
-      } exceptionError=${
-        results.execResult.exceptionError ? results.execResult.exceptionError.error : ''
-      } returnValue=${short(results.execResult.returnValue)} gasRefund=${
-        results.execResult.gasRefund
-      } ]`
+      `Received tx execResult: [ gasUsed=${gasUsed} exceptionError=${
+        exceptionError ? `'${exceptionError.error}'` : 'none'
```
acolytec3, line 409: What are the `'...'` for in the exceptionError?  Is it just to clearly demarcate the error message?

acolytec3 (approved):

packages/vm/src/runTx.ts
ryanio: What do you mean? I don't see that. Maybe you are confusing with the short() formatting on returnValue that after a certain length adds ...

ryanio: Ohh the single quotes? To contain the error message so it looks like `exceptionError='out of gas'`

ryanio:
thanks!!

packages/vm/src/runTx.ts
acolytec3: Yup, sorry, that comment wasn't clear but thanks for the clarification!

PR #1681, opened 2022-02-01
vm: Check vm.DEBUG for emitting step event

acolytec3:
Fix gap left in #1676 to check for `vm.DEBUG` before skipping `stepHook` event


acolytec3:
I've validated locally that this additional check resolves the issue created in #1676 where the `vm:ops:*` logs wouldn't show even though DEBUG was set when running the vm.

packages/vm/src/evm/interpreter.ts
```diff
@@ -150,9 +150,11 @@ export default class Interpreter {
       await dynamicGasHandler(this._runState, gas, this._vm._common)
     }
 
-    if (this._vm.listenerCount('step') > 0) {
+    if (this._vm.listenerCount('step') > 0 || this._vm.DEBUG) {
       // Only run this stepHook function if there is an event listener (e.g. test runner)
-      // since its sole purpose is to emit the step event which is not used anywhere else
+      // or if vm is running in debug mode to display op code debug logs since its sole
+      // purpose is to create debug logs for the step event
+
```
ryanio, line 157: looks great, thanks for addressing so quickly.

one suggestion to write this a bit more succinctly. also i'd say we don't need the newline after the comment to keep it more compact, should still be easy to read on the eyes

```suggestion
      // or if the vm is running in debug mode (to display opcode debug logs)
```

holgerd77:
Short note: have updated the branch, hope that was ok. Assume this is ready for review.

acolytec3:
> Short note: have updated the branch, hope that was ok. Assume this is ready for review.

Yep!  I committed @ryanio 's edit to the comment and there shouldn't be any other changes to be made (that I'm aware of since this is just fixing what I broke before)

holgerd77 (approved):
LGTM

PR #1640, opened 2022-01-17
New Minor Releases (VM Dynamic GasCosts, EIP-2681, Sepolia, various other)

holgerd77:
Fixes #1635 

Still WIP but libraries tackled already open for review.

Will likely remain open (and updated) for a couple of days (or maximally 1-2 weeks) for some final discussions and eventual inclusions (big thing: ESM support still addressable for release or not?).

PR #1683, opened 2022-02-02
Block: New Release v3.6.1 (Minor Release Series Follow-Up Release)

### LINKED ISSUE #1680: Client: Sepolia network connection in v0.3 release not working
The `Sepolia` network connection is not working when syncing the chain with an installed client `v0.3.0` version, which is particularly annoying, since I relatively prominently pointed this out as a good client testing opportunity in the release announcements.

Here is the output, deleted the `sepolia` folder before and ran `ethereumjs --network=sepolia`:

grafik

This is actually *not* happening when running the client from within the monorepo.

Someone should give this some prioritization throughout the day so that we can quickly do a bugfix release.

Not sure, one thesis on this: maybe we should have/would need to also to `Block` and `Blockchain` releases, to make sure that these operate on the latest Common as well? (even if these libraries have no other changes included) 🤔

Would need some validation though.

holgerd77:
Fixes #1680, follow-up on #1640

### DIFF
--- a/packages/block/CHANGELOG.md
+++ b/packages/block/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 (modification: no type change headlines) and this project adheres to
 [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## 3.6.1 - 2022-02-02
+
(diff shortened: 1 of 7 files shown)


ryanio (approved):
lgtm, thanks!

holgerd77:
Did a fresh `@ethereumjs/client` install and can confirm that this works (so: properly syncs `Sepolia`). 😄 🥳

Issue #1397, opened 2021-08-10
Util: Deprecate ethereumjs-units and integrate into Util


holgerd77:
As discussed with @ryanio the [ethereumjs-units](https://github.com/ethereumjs/ethereumjs-units/) package should be deprecated (somewhat separate from this issue) and the functionality integrated into the `Util` package.

This would be useful especially for denominations when using the tx library.

Currently that look like this which is not very user friendly and somewhat error prone:

```typescript
const txData = {
  from,
  nonce: 0,
  gasPrice: 1000000000,
  gasLimit: 21000,
  to,
  value: 1,
}

const tx = Transaction.fromTxData(txData, { common })
```

In `Util` this should likely be a separate module `units`. Otherwise we should give the API some additional thought before we integrate (I wonder e.g. if we have some use case for this `lazyConvert` method/API?).

ryanio:
I can work on this, it would be nice to have a simple internal units util.

`bn.js` doesn't have decimal support though, so we'll have to do something about that, it looks like `ethereumjs-units` uses `bignumber.js`.

For our own monorepo use, I think it might be more helpful to return a `BN` based in wei, something like this:

```typescript
const txData = {
  value: Units.ether(1),
  maxFeePerGas: Units.gwei(10),
  maxPriorityFeePerGas: Units.gwei(10)
  ...
}
```

although this could be the same:

`Units.convert(value: number | string | BN, from: Unit, to: Unit): string`

where you could do

`Units.convert(10, Unit.Gwei, Unit.Wei)` => `10000000000`

But we do need to return strings unfortunately for decimal support:

`Units.convert(1, Unit.Wei, Unit.Ether)` => `0.000000000000000001`

I also don't really see the need for `lazyConvert`, since we are mostly developer tools I'm not sure when we need to input or output formatted strings like that.

holgerd77:
I actually think we should still execute on this (3y old) issue but simply add a super-simple `units.ts` file to `@ethereumjs/util` and more or less only from-scratch "implement" (add) the API Ryan suggested:

```ts
const txData = {
  value: Units.ether(1),
  maxFeePerGas: Units.gwei(10),
  maxPriorityFeePerGas: Units.gwei(10)
  ...
}
```

So, these *three* methods (returning a BigInt), exactly named like this, add a few tests and *NOT* having a look at all any more at this endless discussion from the tries of the closed PR and **ALSO FULLY IGNORE THE ethereumjs-units package**.

[...]

PR #1689, opened 2022-02-03
VM, Block:  Execute block on Layer 2 (Polygon Network)

cbrzn:
_Disclaimer: This is my understanding from what I have been doing so far, if something does not make sense please feel free to correct me :-D_

After playing a bit with this, I have achieved to run an empty block on polygon (with the `generate` flag on `true`). Which was the first step - After trying to execute `runBlock` with a ["stub" block ](https://github.com/ethereumjs/ethereumjs-monorepo/blob/feat/l2-vm-block-execution-support/packages/vm/examples/polygon-example.ts#L69-L82) I realized that we are not supporting these functionalities for these [Custom chains](https://github.com/ethereumjs/ethereumjs-monorepo/blob/feat/l2-vm-block-execution-support/packages/common/src/index.ts#L15). Note that this block just have one transaction which sends ether, nothing too deep atm.

This [comment](https://github.com/ethereumjs/ethereumjs-monorepo/blob/feat/l2-vm-block-execution-support/packages/common/src/index.ts#L194-L196) means that actually this is supposed to be that way and currently there is no way to have full support for L2 chain.

Modifying the [custom method](https://github.com/ethereumjs/ethereumjs-monorepo/blob/develop/packages/common/src/index.ts#L201-L204) to be smart enough to create the context for these L2 chains would be the first step to achieve this. It will allow initiating these chains with mocked data & with the functionalities we want, by activating different EIPs.

So basically my idea of how we should tackle **VM (Block) Execution Support**:

- Update [`custom` method from `Common` class](https://github.com/ethereumjs/ethereumjs-monorepo/blob/develop/packages/common/src/index.ts#L201-L204) to support more options for Layer 2 chains. This would take some refactor on that method, as we would need to change those multiple `if` to be smarter - If this is action is approved I would like to get creative with it and try to come up with something cool :-)
- Support these Layer 2 chains in the [`_getInitializedChains`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/develop/packages/common/src/chains/index.ts#L12) & add the JSON data of chain config (i.e: https://github.com/ethereumjs/ethereumjs-monorepo/blob/develop/packages/common/src/chains/mainnet.json) - This should not take too much time, it's just a matter of getting the data from these L2s and add them to the map object

This will allow us to create the VM with a state good enough to allow the creation of a new block when running `runBlock` method. Also when thinking in the future, like adding custom opcodes, this custom function can be the handler of all these metadata. Just thinking out loud here :-)


holgerd77:
Hi Cesar,
thanks for this head-start, great! 😄 👍 

Some first aspects here:

1. Not sure if I get you right here, but `Common.custom()` vs chain files (like `rinkeby.json`) is not: main/"real" chains vs. L2 chains, but the delimiter is really just: limited functionality (`Common.custom()`) vs full functionality (`rinkeby.json`). So if we get the monorepo to somewhat "fully" support Polygon, it is also possible that we add a dedicated `polygon.json` file to the realm of chain files. For the time being (or also: for the first iterations on development), it is also possible to load a full custom chain file in with the main `Common` constructor, see [here](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/common#initialize-using-customchains-array). This will then (on the `Common` side, not necessarily on upstream libraries) have the same set of functionality then every other chain defined.

1b. Building on 1.: if you think you need to have the `Common.custom()` static constructor to have expanded, please specifically state these things, eventually in dedicated issues or something.

2. When writing up the original Layer 2 issue I was just thinking on an approach the other way around: getting a *real* block from the Polygon network (either you have some ideas yourself or I can alternatively assist, others eventually as well) and *then* trying to read this in with our `Block` library and - if this would work (50% chance I guess 😋) feeding this block into the `VM.runBlock()` method. With this let's call it "constructed approach" from here (creating a Block within the realm of our libraries) you just have no guarantuee whatsoever that this would actually be a block accepted by the Polygon network.

2b. So getting a block from the Polygon network could likely be done by using some exposed Polygon RPC endpoint or the like. If you find something on this (Polygon developer docs or the like) please document here. 🙂 I will also send you the keys to our EthereumJS dev account via DM which you can use for testing.

3. Some very high-level meta comment: please be prepared that the work started here will lead to a myriad of very dedicated sub-tasks, likely done in a lot of different PRs, maybe some towards `master`, maybe some towards `develop`. This is as much about identifying suboptimalities in our libraries on the way and fix/improve on them as it is about this final goal of having L2 support improved. Does this make sense? 🤓

3b. And already one sub task I have: you are always doing these `const oneEthInGwei = new BN(1).pow(new BN(9))` conversions for feeding amounts to our libraries, which in fact is already a lot better than the existing habit of just throwing `900000000....` or the like in. 😅 We have an old really not integrated library [ethereumjs-units](https://github.com/ethereumjs/ethereumjs-units/) in the `ethereumjs` organization together with an issue #1397 to deprecate this library and instead include this functionality in the `Util` package. This would be a nice sub task for you to include along the way (to be done towards `master`).

Ok. This should be enough for the next week. Lol. 😅

packages/common/src/chains/index.ts
```diff
@@ -6,40 +6,39 @@ import kovan from './kovan.json'
...
+  return chainsMapping
 }
 
 /**
- * @deprecated this constant will be internalized (removed)
- * on next major version update
- */
+ - * @deprecated this constant will be internalized (removed)
+ - * on next major version update
+ - */
```
holgerd77, line 43: Not sure if you opened up already for review 🙂, nevertheless some first notes: there really shouldn't be any need for any `Common` library changes along this work.

As mentioned you should be able to directly use the [Common initialization with a customChains Array](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/common#initialize-using-customchains-array) as described in the README.

So these changes here should be reverted. If you stumble upon something you want to improve - e.g. how typing is done in Common - this should be rather done in a separate PR.

packages/vm/examples/l2-playground/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "scripts": {
+    "fetch": "ts-node fetch-polygon.ts"
+  },
+  "devDependencies": {
+    "node-fetch": "3.2.0"
+  }
+}
```
holgerd77, line 8: Why are you now introducing a `package.json` file here? I thought you would have just removed these and put eventual dependencies in the main `devDependencies`? 🤔

packages/vm/examples/l2-playground/fetch-polygon.ts
```diff
@@ -0,0 +1,127 @@
+const fetch = require('node-fetch')
+import VM from '../../src'
+import Common, { CustomChain, Hardfork } from '@ethereumjs/common'
+import { Transaction } from '@ethereumjs/tx'
+import { Account, Address, toBuffer } from 'ethereumjs-util'
+import { BN } from 'ethereumjs-util/dist/externals'
+import { Block } from '@ethereumjs/block'
+import { mumbaiData } from './polygon-mumbai-data'
+
+const url = 'https://polygon-mumbai.g.alchemy.com/v2/EomkJjyUZ_sLzKWuaExlMwcFQZ-hDpjZ'
```
holgerd77, line 10: Not sure if this is only meant to be temporary anyhow, just that we start some early exchange here: we for sure can't expose some private Alchemy API key here which is then used in ways (quantities) we can't control (maybe you should generally not expose in a PR, might have unwanted side effects already).

I wonder if we want to change the `ethers/abi` dev dependency we have already in the VM to a generic `ethers` dependency. I guess this will likely be useful for future examples as well and then people can use their own providers and/or API keys, eventually with the need to adopt an example a bit.

WDYT?

packages/vm/examples/l2-playground/polygon-mumbai-data.ts
```diff
@@ -0,0 +1,69 @@
...
+  comment: '',
+  url: '',
+  genesis,
+  consensus: {
+    type: 'pos',
+    algorithm: '',
+  },
+  hardforks: [hardfork],
+  bootstrapNodes: [node],
+}
```
holgerd77, line 69: Please do not recreate the whole `Common` chain data structure here but just use a plain json file in the form of e.g. [goerli.json](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/chains/goerli.json).

holgerd77: (again, if you think there is some need for improvements on the `Common` library please open up a dedicated issue instead)

holgerd77 (commented):
Cool, great 🥳, already had some first look.

Any first results already? Does this run a Polygon block without error? Or does it stop somewhere?

cbrzn:
Hey @holgerd77 - Thanks for your feedback :-) Indeed this is super WIP and not ready for review

> Any first results already? Does this run a Polygon block without error? Or does it stop somewhere?

At this point, I am getting this error: 
> Fetch polygon has failed because of:  Error: invalid receiptTrie (vm hf=london -> block number=1 hash=0x5aa81807d278e68f8a46a5d9a9b7d2a671584e9ea8dfa2f45c24f1882e079748 hf=london baseFeePerGas=none txs=1 uncles=0)

It is coming from [this line](https://github.com/ethereumjs/ethereumjs-monorepo/blob/feat/l2-vm-block-execution-support/packages/vm/src/runBlock.ts#L197) and I am planning to play more with it today so I can understand where this error comes from (If you can enlight me here I would be more than grateful😄 ). I thought it might be because the polygon chain uses PoS and we currently don't support their algorithm? This probably is a non-sense but it's what came to my mind first lol.

cbrzn:
Also @holgerd77 I think that you forced pushed and now we have more diff - I can delete the branch and re-create it with the old state and afterward we can merge the main dev if you'd like. Let me know if I can help somehow :pray:

packages/vm/examples/l2-playground/polygon-mumbai-data.ts
cbrzn: Nothing to improve here imo :-) Was just playing with it :P

packages/vm/examples/l2-playground/package.json
cbrzn: Because the develop branch does not have the latest changes from master, which is this new structure you're talking about. I thought that in order to keep consistency in the code, for now, to use the same structure we currently have, and then migrate to the latest one when the time of merge comes. Also, this helped me to have the script in the `package.json` and run it easier with the debugger from WebStorm 😛

packages/vm/examples/l2-playground/fetch-polygon.ts
cbrzn: Si, thanks for the catch, indeed it is temporary and I did not mean to push this to GH

> I wonder if we want to change the ethers/abi dev dependency we have already in the VM to a generic ethers dependency. I guess this will likely be useful for future examples as well and then people can use their own providers and/or API keys, eventually with the need to adopt an example a bit.

Yah I like this, let's do it - Also, for the time being, I will add these variables (like API key from alchemy or an account private key) through environment variables

holgerd77:
> Also @holgerd77 I think that you forced pushed and now we have more diff - I can delete the branch and re-create it with the old state and afterward we can merge the main dev if you'd like. Let me know if I can help somehow 🙏

Yes, please do regular rebases of your work towards `develop`, thanks!

holgerd77:
> It is coming from [this line](https://github.com/ethereumjs/ethereumjs-monorepo/blob/feat/l2-vm-block-execution-support/packages/vm/src/runBlock.ts#L197) and I am planning to play more with it today so I can understand where this error comes from (If you can enlight me here I would be more than grateful😄 ). I thought it might be because the polygon chain uses PoS and we currently don't support their algorithm? This probably is a non-sense but it's what came to my mind first lol.

This should not be related to PoS, the consensus algorithm normally doesn't change the result of the VM state transition or the VM run or something (apart from PoW actually updating the miner account, this is happening on block level though).

An `invalid receiptTrie` error is often a - very - general indication that there is something wrong with the VM execution. So there is e.g. `gasUsed` and `logs` in a tx receipt. If this doesn't match up to what would be expected the `receiptTrie` - summing up on these receits - will not match.

Early warning: this is often a pretty hard debug scenario, since this is a very very general error and *everything* [TM] can go wrong. Lol. 😋

So, ways to approach here:

1. Start with the easiest possible case (so first: a block with no tx, does this work? second: a block with one 1 transfer tx, will this work?) and then slowly go up in complexity
2. Check on some basic framing things which could be wrong. Since `gas` is involved in the tx receipt it often happens that there is something off with the gas costs. Two things I can think of: is this really running on the correct HF expected by Polygon? Maybe do some `console.log` or so tests on the VM `common` in the code. Other thing: are there any gas deviations mentioned in the Polygon docs?

acolytec3:
> 2. Check on some basic framing things which could be wrong. Since `gas` is involved in the tx receipt it often happens that there is something off with the gas costs. Two things I can think of: is this really running on the correct HF expected by Polygon? 

I do know that Polygon just implemented the London hardfork recently so I wonder if that could be a possibility?

PR #1693, opened 2022-02-06
Common: method deprecations

ScottyPoi:
1. VM / Client / Devp2p / Block / Common: Rename hardforkBlockBN() to hardforkBlock()
2. Common: Remove deprecated hardforkBlock()
3. Common: Update build:docs
4. Rename hardforkBlock in tests 
5. Fix tests by using .eqn to test BN's against numbers.

### DIFF
--- a/packages/common/tests/hardforks.spec.ts
+++ b/packages/common/tests/hardforks.spec.ts
@@ -1,5 +1,4 @@
 import tape from 'tape'
-import { BN } from 'ethereumjs-util'
 import Common, { Chain, ConsensusAlgorithm, ConsensusType, Hardfork } from '../src'
 
 tape('[Common]: Hardfork logic', function (t: tape.Test) {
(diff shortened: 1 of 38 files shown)


ScottyPoi:
I'm not sure why the 'hardhat' test fails

acolytec3:
It's not your code. Develop is behind master and the package numbers are out of date.  We need to rebase develop and they should solve this problem.

packages/common/tests/hardforks.spec.ts
```diff
@@ -225,14 +224,13 @@ tape('[Common]: Hardfork logic', function (t: tape.Test) {
...
-    st.equal(c.hardforkBlock(Hardfork.Berlin), 12244000, msg)
-    st.ok(c.hardforkBlockBN(Hardfork.Berlin)!.eq(new BN(12244000)), msg)
+    st.ok(c.hardforkBlock(Hardfork.Berlin)!.eqn(12244000), msg)
 
     msg = 'should return null for unscheduled hardfork'
     // developer note: when Shanghai is set,
     // update this test to next unscheduled hardfork.
     st.equal(c.hardforkBlock(Hardfork.Shanghai), null, msg)
-    st.equal(c.hardforkBlockBN(Hardfork.Shanghai), null, msg)
+    st.equal(c.hardforkBlock(Hardfork.Shanghai), null, msg)
```
acolytec3, line 233: I think you can safely remove this line as it duplicates the one directly above it.

acolytec3 (commented):
Looking good so far!  I left one comment to clean up a duplicate line.

Also, once we rebase develop on current master, the hardhat test should start passing again unless they use the `hardforkBlockBN` method in their code.

@holgerd77 In the next breaking release, should we entirely remove `hardforkBlockBN` or just deprecate it and internally rewrite it to use the now rewritten `hardforkBlock`?

holgerd77:
@acolytec3 The old `hardforkBlockBN` names will be removed entirely.

(also, side note: it will likely be the case that this methods will also be transitioned from the `BN.js` to `BigInt` usage. This step is just for separation of concerns.)

I will close this PR here, have mailed Scotty on this, still some things to address.

PR #1694, opened 2022-02-07
Rebase develop on current master

acolytec3:


holgerd77:
Ah, please do not open any PRs on this, as I said I will do frequent rebases (roughly every week) and directly backup `develop` and then force-push on `develop` (if we are do this by PR we are back on the merge-front 😋). Also please always update your own branches (e.g. the `BigInt` branch) with going to the branch and the do `git rebase master` (on an updated master + always do a backup of your branch before) and never merge anything in there.

Have pushed an updated `develop`, will close here.

Will also do a rebase on Scotty's PR.

PR #1697, opened 2022-02-07
Utils: Amount formatter

cbrzn:
This PR aims to implement the logic from the package [`ethereum-units`](https://github.com/ethereumjs/ethereumjs-units/blob/master/index.js).

I went ahead and tried to do it with `BigInt`, when doing it I noticed that it does not support decimals, so it got a bit tricky. I do not think this is bulletproof but at least it's the first start to iterate over it. It is still a work in progress but I wanted to know your thoughts on how it is so far 😛 .

I saw @ryanio [comment](https://github.com/ethereumjs/ethereumjs-monorepo/issues/1397#issuecomment-898135030) and tried to come up with an approach - If you feel it's hacky I am open to discussing how can we achieve cool utils for amounts 😎 

Edit: Notice that the only test cases that are passing are the ones uncommented on the test file


holgerd77:
Dropping this here from a post in the chat so that it doesn't get lost:

---

I am generally not sure on the API yet, we do not necessarily need to recreate the everything-to-everything API from https://github.com/ethereumjs/ethereumjs-units (example: `Units.convert('1', 'eth', 'wei')`).

Richard is also dividing this into two separate functions - in contrast to the `ethereumjs-units` functions, `parseUnits` and `formatUnits` - see https://docs.ethers.io/v5/api/utils/display-logic/#display-logic - we might also take some inspiration from this, this then allows for an easier input -> output API (so `parseUnits`: string -> BigNumber (special Ethers object), `formatUnits`: BigNumber -> string).

I guess what we mainly want for now is the conversion from a higher unit to the plain data input, to allow for easier feeding values on object instantiation.

So I would somewhat be a fan of the following API, bringing the unit into the API function name and eventually also just limiting to the two functions (being the most important units (?)) for now:

- `units.fromEth()`: string -> (native) BigInt, example `units.fromEth('1.1') // 1100000000000000000n`
- `units.fromGwei()`: string -> (native) BigInt, example `units.fromGwei('5') // 5000000000n`

Ah, and maybe it *does* make sense to then also provide the other way around:

- `units.toEth()`: BigInt -> string, example `units.toEth(1100000000000000000n) // '1.1'`
- `units.toGwei()`: BigInt -> string, example `units.toGwei(5000000000n) // '5'`

(I also wouldn't do this "lazy" thing (see `ethereum-units` API) with appending "wei" or the like to the strings)

(and it would also be no drama if we would expand this API over time to also include the other units)

But likely wait for some other voices here, that's just my take on this.

packages/util/src/units.ts
```diff
@@ -0,0 +1,72 @@
+// Source of constants: https://www.languagesandnumbers.com/articles/en/ethereum-ether-units/
+export const unitsConstants = {
```
ryanio, line 2: Can you format this as a typescript enum `Unit`? I just tested and the value can be a negative number so that should work. Key should be PascalCase to match other enums in the monorepo. Then you shouldn't have to use any string values in the actual code, i.e. UnitName.Ether should === 'ether'

packages/util/src/units.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (stringValue.length < decimalsToConvert) {
+      decimalsToConvert -= stringValue.length
+      const hasTrailingZeros = stringValue.endsWith('0')
+
+      // We need to remove the zeros at the end of the final number
+      // We have the zeros counted variable to flag that it's only taking
+      // in account zeros from the end
+      // For example: 200 Gwei -> 0.000000200 Eth
+      let zerosCounted = false
+      const removeTrailingZeros = (acc: number, current: string) => {
```
ryanio, line 51: can you maybe format this `removeTrailingZeros` as a helper method on the class instead, to help better organize the code? you can do that with the other small formatting things you are doing as well, so then this static method can read a lot cleaner/simpler (including the math like multiplication or division)

packages/util/test/unit.spec.ts
```diff
@@ -0,0 +1,30 @@
+import tape from 'tape'
+import { Units } from '../src'
+
+tape('unit', function (t) {
+  t.test('should work for big unit to small unit', function (st) {
+    st.equal(Units.convert('1', 'ether', 'wei'), '1000000000000000000')
```
ryanio, line 6: once you have the enum, then these can be rewritten e.g.

```suggestion
    st.equal(Units.convert('1', Unit.Ether, Unit.Wei), '1000000000000000000')
```

ryanio: can you also test with the other input value types other than string to ensure the resulting values are the same, so number and bigint (whichever types are accepted in the method)

holgerd77:
Ok, I am a bit more sold (but not yet completely) on keeping the base API as is with now also seeing the code examples with the enums like `Units.convert('1', Unit.Ether, Unit.Wei)`. It also adds that I just discovered that one can [initialize](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) a `BigInt` with a string, which would fit well here.

This still leads to pretty long code fragments on common use cases like:

```typescript
const tx = Transaction.fromTxData({
  gasPrice: BigInt(Units.convert('45', Unit.Gwei, Unit.Wei)),
  gasLimit: 10000000,
  value: 42,
})
```

One way to settle here that we agree on this base API and then add the functions I suggested on top as some semantic-sugar helpers for the 3-4 most common use cases (so, likely/eventually: ETH/Gwei -> WEI and back?).

So this could simplify the above to (and would also save on the enum import):

```typescript
const tx = Transaction.fromTxData({
  gasPrice: Units.fromGwei('45'),
  gasLimit: 10000000,
  value: 42,
})
```

We can discuss on our call a bit. Also no need for a rushed decision, we can very well give this a couple of more days.

Find this important to get right/practical, this might get a relatively widely used (at least within our own context) API.

cbrzn:
Okay, so I have implemented enum for `Unit` and moved the `removeTrailingZeros` function as a private method of `Units` class as recommended by @ryanio. I also created two methods: `toEth` and `toGwei` as suggested by @holgerd77.

I updated the tests to use these two methods to check how I thought the API could be. If this is approved, I think we should have the methods: `toWei`, `toGwei`, `toEth`, `fromWei`, `fromGwei`, `fromEth`, which will use under the hood the `convert` method. If someone wants a more customized use case (i.e., working with a different unit than wei, gwei, or eth`), they can also use this `convert` function. 

Note that these two methods currently implemented have dynamic `from`, but the default is the most common one. I think that all these methods can have this pattern?

I agree with you, @ryanio. This should be aimed to develop. I will update this branch accordingly and change the base branch. I was wondering if I should wait that the PR that adds BigInt into the VM is merged rebasing before doing so?

packages/util/src/units.ts
```diff
@@ -0,0 +1,118 @@
...
+export enum UnitValue {
+  'wei' = -18,
+  'kwei' = -15,
+  'mwei' = -12,
+  'gwei' = -9,
+  'micro' = -6,
+  'milli' = -3,
+  'ether' = 1,
+  'kether' = 3,
+  'megaether' = 6,
```
axic, line 24: Who is using these units? And why `megather` and not `Mether`? Why not have `nanoether`/`nanoeth` which is used in some EIPs (like 1559)? How about the other old/odd units (`finney`/`shannon`/etc)?

The library this tries to replace has a lot more units: https://github.com/ethereumjs/ethereumjs-units/blob/master/units.json

I am not sure it is a good idea to support all these units, but I'd probably suggest to have at least two functions, a strict and a loose one. The strict should only support `wei`/`gwei`/`eth` and maybe `nanoeth`. The loose one can support all the legacy ones.

holgerd77: Thanks for the additional context, we need to bring all this together with the overall API design, so we should give this some thought.

ATM (here) we are on a path where we have convenience functions for the most common unit conversions (so e.g. `fromGwei()`, which are wrapping the generic `convert()` function.

I would be a fan of instead of splitting this up in strict and loose but rather throwing obscure ones out and eventually take one or two missed in.

So that would mean:

- kether, mether, gether, tether out (would be in strong support, totally obscure and not suited given the value of ETH)
- For the lower ETH ones I guess we just might want to keep everything in for completeness + who knows what denominations become useful due to price evolution in some areas (gas for L2s, whatever)
- nanoeth == gwei, right? So we could add an alias here for the convenience functions, not really sure if this is necessary though (but also no strong opinion)

Generally: my "unit usage" feeling is limited, so other voices and additions here very welcome.

cbrzn: > Who is using these units? And why megather and not Mether? Why not have nanoether/nanoeth which is used in some EIPs (like 1559)? How about the other old/odd units (finney/shannon/etc)? 

To be honest, I searched for Ethereum units on the internet and got these, I linked the URL in the code as a comment. I will update to the ones you suggested.

As Holger said:
> ATM (here) we are on a path where we have convenience functions for the most common unit conversions (so e.g. fromGwei(), which are wrapping the generic convert() function.

I was thinking that probably creating this `to${Unit}` methods it's a way to do these strict functions. I thought `Units.toGwei(10)` can be a good interface to make code easier to read

In advance, I want to say thank you for taking the time to review this :-)

packages/util/src/units.ts
```diff
@@ -0,0 +1,118 @@
...
+  'gether' = 9,
+  'tether' = 12,
+}
+export type UnitsNames = keyof typeof UnitValue
```
ryanio, line 28: would suggest singular here

```suggestion
export type UnitNames = keyof typeof UnitValue
```

holgerd77:
Ok, tried some review here, but I have to say:

I am actually still a lot more sold on my initial proposal from above, so basically doing:


- `units.fromEth()`: string -> (native) BigInt, example `units.fromEth('1.1') // 1100000000000000000n`
- `units.fromGwei()`: string -> (native) BigInt, example `units.fromGwei('5') // 5000000000n`

Ah, and maybe it *does* make sense to then also provide the other way around:

- `units.toEth()`: BigInt -> string, example `units.toEth(1100000000000000000n) // '1.1'`
- `units.toGwei()`: BigInt -> string, example `units.toGwei(5000000000n) // '5'`

And that's it (that would actually have been 2 hours of work, lol 😜).

There is just not a single use case which came up during this discussion and exchange which would ask for the need of more functionality.

But the more important thing is: the above is dramatically easier and simpler to implement. With this everything-to-everything approach we are just running into such wild territory from a security perspective with this taking in of `BigInt`, `number` AND `string` and *then* going both up and down in conversion.

There are so many things arising from this like:

- Accepted floating point decimals as input
- Max integer input for `number`
- Floating points being injected or not injected in the string output depending on the up or down conversion
- Guarding against wrong string input
- ...

This test extract shows some of the possible variety of possibilities here, with some tests not even working yet.

```typescript
st.equal(Units.convert('20', 'gwei', 'wei'), '20000000000')
// st.equal(Units.convert('20.05', 'gwei', 'wei'), '20050000000')
// st.equal(Units.convert('20.005', 'kwei', 'wei'), '20005')
// st.equal(Units.convert('20.0005', 'kwei', 'wei'), '20000')
st.equal(Units.toGwei('1'), '1000000000')
st.equal(Units.convert('1', 'tether', 'ether'), '1000000000000')
st.equal(Units.convert('1', 'tether', 'wei'), '1000000000000000000000000000000')
```

So I would strongly suggest that we either go to the simple version or rather cancel this whole thing.

There is just one simple mistake to be made here from us in this PR and we are potentially in dramatic charter waters when only one edge case goes wrong here (like we had in this 20 million fee bug).

This seems very much not worth the risk here for some additional convenience method(s).

holgerd77:
Ah, we have gone through so many iterations here and ended up with a somewhat little amount of code changes. Can you please resubmit just the final changes once #1771 has been merged on top of a clean `develop` and then close this one here? 🙂 

That would be great, then we have a manageable change set and things also get easier along follow-up rebases.

PR #1702, opened 2022-02-08
Add latest hardforks to tx tests

jochem-brouwer:
I noticed we are not using all hardforks on the tx ethereum/tests tests, so added them here.


ryanio (approved):
nice, thanks!

PR #1705, opened 2022-02-09
VM: Add option to add custom opcodes

jochem-brouwer:
This PR adds a new field to `VMOpts` to allow users to add/change/delete opcodes in the VM.

Please note: in `gas.ts` the linter has inserted whitespace in almost every line, the only change there is to also allow for synchronous gas functions.


packages/vm/src/evm/opcodes/codes.ts
```diff
@@ -322,5 +325,36 @@ export function getOpcodesForHF(common: Common): OpcodeList {
...
+  if (customOpcodes) {
+    for (const _code of customOpcodes) {
+      const code = <any>_code
+      if (code.logicFunction === undefined) {
+        delete opcodeBuilder[code.opcode]
+        continue
+      }
+
+      // Sanity checks
+      if (code.opcodeName === undefined || code.baseFee === undefined) {
```
jochem-brouwer, line 337: Do we want extra checks here? What if I try to delete an opcode which is not activated, should it throw? What if I try to add opcodes at the same opcode number twice?

jochem-brouwer:
Oof I just realized we have a problem here.

I am writing to the constants of `gas.ts` and `function.ts` if I add custom opcodes. Thus that means that whatever I have changed (lets say I once deleted `KECCAK` in a new VM) then that also means I have deleted it in further VMs. 

How should I approach solving this? It would imply that we now need to deep-copy `gas.ts` and `function.ts` exported constants..?

holgerd77:
This looks already pretty cool, thanks a lot for addressing so quickly Jochem! 🙂

Let's keep this a bit in discussion for a couple of days - independently from if we find answers to the existing questions. This is such a profound change, I guess it would be good if we let this new API sink in a bit and shovel this back and forth to think about use cases and implications. 😋

packages/vm/tests/api/runCode.spec.ts
```diff
@@ -96,6 +98,71 @@ tape('VM.runCode: interpreter', (t) => {
   })
 })
 
+tape('VM: custom opcodes', (t) => {
```
ryanio, line 101: not sure if this new test suite belongs in runCode, i suggest we make a new file `tests/api/evm/customOpcodes.spec.ts`

packages/vm/src/index.ts
```diff
@@ -143,6 +166,7 @@ export default class VM extends AsyncEventEmitter {
   protected _opcodes: OpcodeList
   protected readonly _hardforkByBlockNumber: boolean
   protected readonly _hardforkByTD?: BNLike
+  protected readonly _customOpcodes?: CustomOpcode[]
```
holgerd77, line 169: Totally a side topic here but just want to mention on the occasion since we are already so much in these StateManager design questions: my dream ("I have a dream", lol) for the breaking releases is still that we also manage to separate the EVM itself (so basically: everything in the `evm` folder) and all this outer execution stuff (so roughly: everything - at least - from `VM.runTx()` upwards).

This mangling together of the pure EVM execution logic (opcodes in, result out, basically 😋) and all this outer environment stuff with running txs, creating receipts, running the block, assiging block rewards in this one big `VM` object is one of the biggest design flaws we have. One can see this over and over again, e.g. this `customOpcodes` option would - if we would have that - also rather only be a thing to be passed to a clean (inner) `EVM` interface (in the sense of: API) and nothing to be babbled to this monster object with all the additional logic.

I guess this outer thing is something which rather would belong to the client then to the VM (we for sure doesn't want to integrate though) and historically likely was bapped on to the VM because there just wasn't a wrapping client or something and people needed this functionality.

Maybe we can come up with a separate package or so - `VMExecutionEnv` (I have totally no idea on a fitting name yet) - and then create a stable interface (again: API) to the EVM package.

This would lead to a lot of new use cases. The EVM package could be swapped out, extended by others, used in a standalone-context, etc.. I am also pretty sure that our overall VM architecture would greatly benefit by refactoring questions which come up along the way.

This would need some serious design thinking though, not sure if we would make it with this into this round of breaking releases. Maybe you also want to give this some thinking already. 🙂

P.S.: I know, comment is totally misplaced here. 😛

packages/vm/src/index.ts
```diff
@@ -117,6 +118,28 @@ export interface VMOpts {
...
+   *    opcodeName: string
+   *    // The base fee of the opcode
+   *    baseFee: number
+   *    // If the opcode charges dynamic gas, add this here. To charge the gas, use the `i` methods of the BN, to update the charged gas
+   *    gasFunction?: function(runState: RunState, gas: BN, common: Common)
+   *    // The logic of the opcode which holds the logic of changing the current state
+   *    logicFunction: function(runState: RunState)
+   * }
+   * Note: gasFunction and logicFunction can both be async or synchronous functions
+   */
```
holgerd77, line 140: This is already a great summary - love this detailed API description. 🥰

Maybe you can give an additional note here that this functionality is not "hardfork aware", so opcodes added here will be included in a static way, no matter what HF is set or reset.

packages/vm/src/evm/opcodes/codes.ts
holgerd77: I guess it provides more freedom if it just deletes an opcode if the opcode is there and otherwise does nothing rather than throwing. Let's imagine the case where poeple just haven't included opcode x in their sidechain at all and in a normal mainnet context the opcode gets activated along a HF. Then this would throw here when used in a generic context and the VM gets set before this HF. I guess that's not what we want and rather generally not necessary to include this limitation.

Hmm, I guess the "opcode twice" case might also not need extra handling. People should just intuitively grasping that they shouldn't really be doing this. And if this happens for whatever reasons then simply the second implementation is taken. For the "opcode twice" case I guess it wouldn't hurt either though to add a simple check if easily doable.

packages/vm/src/evm/opcodes/codes.ts
```diff
@@ -322,5 +325,36 @@ export function getOpcodesForHF(common: Common): OpcodeList {
...
+        throw new Error(
+          'Custom opcode does not have the required values: opcodeName and baseFee are required'
```
holgerd77, line 339: Maybe you can add the opcode number here for reference.

packages/vm/src/evm/opcodes/codes.ts
```diff
@@ -322,5 +325,36 @@ export function getOpcodesForHF(common: Common): OpcodeList {
...
+        [code.opcode]: {
+          name: code.opcodeName,
+          isAsync: true,
+          dynamicGas: code.gasFunction !== undefined,
+          fee: code.baseFee,
+        },
+      }
+      opcodeBuilder = { ...opcodeBuilder, ...entry }
+      if (code.gasFunction) {
+        dynamicGasHandlers.set(code.opcode, code.gasFunction)
```
holgerd77, line 352: I love this so much to see how these things play so nicely together close to everytime we do a profound refactoring which was initially meant to solve something completely different! 🥳

Thanks again for this great work! ❤️

packages/vm/src/evm/types.ts
```diff
@@ -1,4 +1,21 @@
...
+  opcode: number
+}
+
+export type AddOpcode = {
+  opcode: number
+  opcodeName: string
+  baseFee: number
+  gasFunction?: AsyncDynamicGasHandler | SyncDynamicGasHandler
+  logicFunction: OpHandler
+}
```
holgerd77, line 19: Not a totally strong opinion here, but I wonder if it's enough if we publicly only expose the `CustomOpcode` type here? 🤔

Tendentially always a fan of keeping the exposed interface/API lean if possible.

packages/vm/tests/api/runCode.spec.ts
```diff
@@ -96,6 +98,71 @@ tape('VM.runCode: interpreter', (t) => {
...
+    const gas = 123456
+    const res = await vm.runCode({
+      code: Buffer.from('20', 'hex'),
+      gasLimit: new BN(gas),
+    })
+    st.ok(res.gasUsed.eqn(totalFee), 'succesfully charged correct gas')
+    st.ok(res.runState!.stack._store[0].eqn(stackPush), 'succesfully ran opcode logic')
+  })
+})
+
```
holgerd77, line 165: Love these tests, great! 🙂

holgerd77 (commented):
This looks already pretty great, cool, thanks for tackling this so quickly! 🚀

Some things for discussion and comments with suggestions for some additions/modifications.

holgerd77:
@jochem-brouwer do you have some time to pick this up this week? Would be great if we can move this forward (merge). 🙂

jochem-brouwer:
I'm gonna need to give this a little bit more thought. The problem is that the handlers (gas handler + logic handler) are used everywhere from using the exported values from their respective files (e.g. here: https://github.com/ethereumjs/ethereumjs-monorepo/blob/c842a973bc8058df71cda140fe0720292e1aafe1/packages/vm/src/evm/interpreter.ts#L182). We somehow need to integrate the "copies"  of these (only necessary when using custom opcodes to prevent editing the original) in (probably) EVM. So instead of getting the opcodes from their opcode files, we would get them from the respective Map which resides inside EVM. So, `getOpcodesForHF` here https://github.com/ethereumjs/ethereumjs-monorepo/blob/c842a973bc8058df71cda140fe0720292e1aafe1/packages/vm/src/evm/opcodes/codes.ts#L302 should also get the EVM object and edit the gas+logic maps in there. 

Does that make sense? So we (1) need to create copies of the gas/logic maps (https://github.com/ethereumjs/ethereumjs-monorepo/blob/c842a973bc8058df71cda140fe0720292e1aafe1/packages/vm/src/evm/opcodes/gas.ts, https://github.com/ethereumjs/ethereumjs-monorepo/blob/c842a973bc8058df71cda140fe0720292e1aafe1/packages/vm/src/evm/opcodes/functions.ts) and (2) we need to somehow integrate it in the VM such that these values are read internally (probably EVM).

However, maybe EVM is also not the right place because each Tx would create new EVM https://github.com/ethereumjs/ethereumjs-monorepo/blob/c842a973bc8058df71cda140fe0720292e1aafe1/packages/vm/src/runTx.ts#L392. I'm not sure how expensive copying these maps are (I don't think super expensive). However, on the other hand, these copies would only work if we have custom opcodes so that would not slow down anything related to most practical implementations where we need speed (mainnet). If we want to prevent these copies then the right place would probably be VM itself.

Thoughts?

packages/vm/src/index.ts
jochem-brouwer: I think the general point you are making here is that we should take a deep look at our internals in the VM to see if we should refactor there. I'm also getting the feeling that some things are "all over the place" and maybe a bit misplaced in some cases as well. We might do an internal refactor soon?

jochem-brouwer:
I think at this point EVM.ts would be the right place to put these maps.

holgerd77:
I think I am not getting the problem yet.

So lets take e.g. the `dynamicGasHandlers`. These are only imported once in `interpreter.ts` and then also only accessed in one single location in `Interpreter`, so [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/evm/interpreter.ts#L147).

Istn't this just a one-liner to expand `dynamicGasHandlers` at this specific point and add/modfiy/remove the custom opcode dynamic gas handlers and then just work on this modified Map? 🤔

holgerd77:
Update: ah, I think I am slowly getting what you mean.

I have the impression you are thinking the wrong way around. `getOpcodesForHF()` doesn't need to be passed anything additionally in if I am not mistaken. Instead the outer `VM` already has this created dynamic (custom) opcode list in the `_opcodes` property. The outer `VM` is also passed to the inner `EVM` in the constructor `constructor(vm: any, txContext: TxContext, block: Block)`, so you can access the opcodes from there with `this._vm._opcodes` and use the dynamically created list in places like you mentioned (https://github.com/ethereumjs/ethereumjs-monorepo/blob/c842a973bc8058df71cda140fe0720292e1aafe1/packages/vm/src/evm/interpreter.ts#L182) instead of taking the static maps from the import (and use this in other places as well).

Does this make sense?

holgerd77:
Update 2: ah, just seeing that this `VM._opcodes` property only contains the opcode information, and not the function and gas logic itself. So I would just add similar properties to the EVM - so `EVM._handlers` and `EVM.dynamicGasHandlers` and assign these properties along the `getOpcodesForHF` run and then do all VM executions from there and replace all static usages of `ophandlers`?

Wouldn't think that this would draw in some performance penalty.

packages/vm/src/index.ts
holgerd77: I would like to do one along the breaking releases, with the optimal outcome that we can separate EVM and VM and have a clean (and somewhat refactored) API between them (what also comes into play here is this EVMC interface https://github.com/ethereum/evmc eventually).

Not sure if we'll make it though to full extend in this round, this is particularly hard.

jochem-brouwer:
For some reason I cannot reply to this; https://github.com/ethereumjs/ethereumjs-monorepo/pull/1705#discussion_r806732462

There is no check if the opcode is present; it does not throw if it is not present and user wants to delete (`logicFunction` set to `undefined`). There are also no checks for "opcode twice", you are right, users are responsible for their own inputs we have no reason to validate their input.

packages/vm/src/evm/types.ts
jochem-brouwer: Ah wait this is actually the case. I have exported them specifically so you can get this extra typescript info. Otherwise it is not super clear how to use `DeleteOpcode`, if we export it you can use it without reading the manual :smile: 

But can still remove it if that is better, have no strong opinion

jochem-brouwer:
I have updated, tests should pass, this should be ready for review now.

Note: I added the maps to VM, not to EVM. The reason is that EVM is instantiated many times and we do not want to copy the opcode map each time. (Only once and only if it is really necessary, if opcodes have changed or upon initializing). Otherwise, each time on `runTx` we copy these maps, which is not what we want. (Not sure if it is deep or shallow copy, but it is unnecessary anyways)

jochem-brouwer:
Addressed the review comments as well.

packages/vm/src/evm/types.ts
holgerd77: Ok, let's keep, also no super strong opinion here

packages/vm/src/index.ts
```diff
@@ -424,7 +429,18 @@ export default class VM extends AsyncEventEmitter {
...
+
+  /**
+   * Update the opcode data
+   * TODO: make this EIP-friendly
+   */
+  updateActiveOpcodes() {
+    const data = getOpcodesForHF(this._common, this._customOpcodes)
+    this._opcodes = data.opcodes
+    this._dynamicGasHandlers = data.dynamicGasHandlers
+    this._handlers = data.handlers
```
holgerd77, line 443: This method shouldn't be exposed to the users I guess but rather be kept internally (and therefore made `private` and renamed to `_updateActiveOpcodes()` I guess (due to our current naming scheme within the VM)).

The opcodes + handlers just should be up-to-date when changes occur, and I guess this is also happening in the current state of the code, e.g. by being called on a HF-changed event.

While eventually being a bit less clean I guess I would even be a fan to "internalize" this functionality into `getActiveOpcodes`, so basically doing all of this functionality inside that function. Practically it comes out to the same result and then we do not have another `_updateActiveOpcodes()` function exposed to JS users (where this is not hidden but just visible e.g. on auto-completion in the Node CLI).

packages/vm/src/index.ts
```diff
@@ -163,7 +165,11 @@ export default class VM extends AsyncEventEmitter {
   protected readonly _opts: VMOpts
   protected _isInitialized: boolean = false
   protected readonly _allowUnlimitedContractSize: boolean
-  protected _opcodes: OpcodeList
+  // This opcode data is always set since `updateActiveOpcodes()` is called in the constructor
+  protected _opcodes!: OpcodeList
+  protected _handlers!: Map<number, OpHandler>
+  protected _dynamicGasHandlers!: Map<number, AsyncDynamicGasHandler | SyncDynamicGasHandler>
```
holgerd77, line 171: If `updateActiveOpcodes()` is called in the constructor, are these assigments then not detected by TypeScript (and the `!` additions therefore unnecessary)? 🤔

holgerd77 (changes requested):
One API thing and one small question, otherwise looks good. 😄

packages/vm/src/index.ts
jochem-brouwer: You would assume that TypeScript can detect this, especially because (1) the method is called in the constructor and (2) it does not have any branching (no if statements). However, it does not detect this, I had to add these `!`s. I can re-check though but I'm rather sure it threw if I removed them.

packages/vm/src/index.ts
holgerd77: Ah, yes, `getOpcodesForHF` likely makes even more sense for this.

(and then also maybe leave the existing return type for this function (this should then be possible, right?)? I am a bit afraid that someone might use this function directly for whatever purposes)

holgerd77:
Short note: have just merged #1757, so the PR here would need another update to be ready for review.

jochem-brouwer:
I took a look at this today but the merge conflicts are very weird :thinking: Need to give this a new look later because I messed up my local merge :disappointed:

jochem-brouwer:
Ok, this should work, ready for re-review.

packages/vm/src/index.ts
jochem-brouwer: This should be addressed now!

packages/vm/src/index.ts
```diff
@@ -126,6 +128,29 @@ export interface VMOpts {
...
+   * If this opcode is then used in the VM, the `INVALID` opcode would instead be used.
+   * To add an opcode, add an entry of the following format:
+   * {
+   *    // The opcode number which will invoke the custom opcode logic
+   *    opcode: number
+   *    // The name of the opcode (as seen in the `step` event)
+   *    opcodeName: string
+   *    // The base fee of the opcode
+   *    baseFee: number
+   *    // If the opcode charges dynamic gas, add this here. To charge the gas, use the `i` methods of the BN, to update the charged gas
```
acolytec3, line 145: Just a note but this will need to be updated when we rebase this change on develop to reference BigInts and not BN since the concept of `i` methods doesn't exist and can easily trip people up since bigints aren't objects that can be modified in place.

acolytec3 (approved):
Generally looks good.  A couple of comments on documentation that we could take or leave.  

I also verified the delete/add opcode logic by inspecting the opcode lists directly in the tests so can vouch that code works great.

packages/vm/src/evm/opcodes/codes.ts
```diff
@@ -322,5 +334,40 @@ export function getOpcodesForHF(common: Common): OpcodeList {
...
+      // Sanity checks
+      if (code.opcodeName === undefined || code.baseFee === undefined) {
+        throw new Error(
+          `Custom opcode ${code.opcode} does not have the required values: opcodeName and baseFee are required`
+        )
+      }
+      const entry = {
+        [code.opcode]: {
+          name: code.opcodeName,
+          isAsync: true,
```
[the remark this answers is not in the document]
jochem-brouwer: I have set this to `true` to always `await` when invoking this opcode. However, if desired, can also do a check here to see if it is an async method (is this possible?) and set the `isAsync` var as desired.

packages/vm/src/evm/opcodes/codes.ts
```diff
@@ -293,15 +296,24 @@ function createOpcodes(opcodes: OpcodeEntryFee): OpcodeList {
...
+  dynamicGasHandlers: Map<number, AsyncDynamicGasHandler | SyncDynamicGasHandler>
+  handlers: Map<number, OpHandler>
+  opcodes: OpcodeList
+}
+
 /**
  * Get suitable opcodes for the required hardfork.
  *
  * @param common {Common} Ethereumjs Common metadata object.
  * @returns {OpcodeList} Opcodes dictionary object.
```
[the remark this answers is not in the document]
jochem-brouwer: Yep good point, will do

packages/vm/src/index.ts
jochem-brouwer: Good point we have to keep this in mind when we rebase develop! :)

holgerd77 (approved):

PR #1714, opened 2022-02-11
Util: Upgrade to ethereum-cryptography 1.0

### LINKED ISSUE #1668: Upgrade ethereum-cryptography to 1.0
Hey folks, we've finally released ethereum-cryptography 1.0: https://github.com/ethereum/js-ethereum-cryptography

I think this could be a tracking issue for the monorepo upgrade process?

Note that we've made it extremely easy to upgrade. It contains exactly three incompatibilities:

1. We return `Uint8Array` from all methods that worked with `Buffer` before
2. We target runtimes with `bigint` support, which is Chrome 67+, Edge 79+, Firefox 68+, Safari 14+, node.js 10+
3. Users of `ethereum-cryptography/secp256k1` should rename import to `ethereum-cryptography/secp256k1-compat`. There is no immediate urge in switching to `noble-secp256k1` abstractions, we provide a compatibility layer for old code and do integration testing

paulmillr:
Two caveats:

1. Added @noble/secp256k1 directly because we forgot to export `recoverPublicKey` to e-c. I suggest to merge like this, and i'll do another PR after we release 1.0.1
2. isValidPrivate tests are failing: noble does not throw on any invalid private, `isValidPrivateKey` is always either true or false. What should I do?

Closes #1668

### DIFF
--- a/packages/util/src/account.ts
+++ b/packages/util/src/account.ts
@@ -1,10 +1,5 @@
 import { BN, rlp } from './externals'
-import {
-  privateKeyVerify,
-  publicKeyCreate,
-  publicKeyVerify,
-  publicKeyConvert,
-} from 'ethereum-cryptography/secp256k1'
+import { Point, utils } from 'ethereum-cryptography/secp256k1'
 import { stripHexPrefix } from './internal'
 import { KECCAK256_RLP, KECCAK256_NULL } from './constants'
 import { zeros, bufferToHex, toBuffer } from './bytes'
@@ -231,7 +226,7 @@ export const generateAddress2 = function (from: Buffer, salt: Buffer, initCode: 
  * Checks if the private key satisfies the rules of the curve secp256k1.
  */
 export const isValidPrivate = function (privateKey: Buffer): boolean {
-  return privateKeyVerify(privateKey)
+  return utils.isValidPrivateKey(privateKey)
 }
 
 /**
@@ -244,14 +239,25 @@ export const isValidPublic = function (publicKey: Buffer, sanitize: boolean = fa
   assertIsBuffer(publicKey)
   if (publicKey.length === 64) {
     // Convert to SEC1 for secp256k1
-    return publicKeyVerify(Buffer.concat([Buffer.from([4]), publicKey]))
+    // Automatically checks whether point is on curve
+    try {
+      Point.fromHex(Buffer.concat([Buffer.from([4]), publicKey]))
+      return true
+    } catch (e) {
+      return false
+    }
   }
 
   if (!sanitize) {
     return false
   }
 
-  return publicKeyVerify(publicKey)
+  try {
+    Point.fromHex(publicKey)
+    return true
+  } catch (e) {
+    return false
+  }
 }
 
 /**
@@ -263,7 +269,7 @@ export const isValidPublic = function (publicKey: Buffer, sanitize: boolean = fa
 export const pubToAddress = function (pubKey: Buffer, sanitize: boolean = false): Buffer {
   assertIsBuffer(pubKey)
   if (sanitize && pubKey.length !== 64) {
-    pubKey = Buffer.from(publicKeyConvert(pubKey, false).slice(1))
+    pubKey = Buffer.from(Point.fromHex(pubKey).toRawBytes(false).slice(1))
   }
   if (pubKey.length !== 64) {
     throw new Error('Expected pubKey to be of length 64')
@@ -280,7 +286,7 @@ export const publicToAddress = pubToAddress
 export const privateToPublic = function (privateKey: Buffer): Buffer {
   assertIsBuffer(privateKey)
   // skip the type flag and use the X, Y points
-  return Buffer.from(publicKeyCreate(privateKey, false)).slice(1)
+  return Buffer.from(Point.fromPrivateKey(privateKey).toRawBytes(false).slice(1))
 }
 
 /**
@@ -297,7 +303,7 @@ export const privateToAddress = function (privateKey: Buffer): Buffer {
 export const importPublic = function (publicKey: Buffer): Buffer {
   assertIsBuffer(publicKey)
   if (publicKey.length !== 64) {
-    publicKey = Buffer.from(publicKeyConvert(publicKey, false).slice(1))
+    publicKey = Buffer.from(Point.fromHex(publicKey).toRawBytes(false).slice(1))
   }
   return publicKey
 }
--- a/packages/util/package.json
+++ b/packages/util/package.json
@@ -39,8 +39,7 @@
   "dependencies": {
     "@types/bn.js": "^5.1.0",
     "bn.js": "^5.1.2",
-    "create-hash": "^1.1.2",
-    "ethereum-cryptography": "^0.1.3",
+    "ethereum-cryptography": "^1.0.3",
     "rlp": "^2.2.4"
   },
   "devDependencies": {
--- a/packages/util/src/hash.ts
+++ b/packages/util/src/hash.ts
@@ -1,16 +1,11 @@
 import { keccak224, keccak384, keccak256 as k256, keccak512 } from 'ethereum-cryptography/keccak'
-const createHash = require('create-hash')
+import { ripemd160 as rmd160 } from 'ethereum-cryptography/ripemd160'
+import { sha256 as s256 } from 'ethereum-cryptography/sha256'
 import { rlp } from './externals'
 import { toBuffer, setLengthLeft } from './bytes'
 import { assertIsString, assertIsBuffer, assertIsArray, assertIsHexString } from './helpers'
 
-/**
- * Creates Keccak hash of a Buffer input
- * @param a The input data (Buffer)
- * @param bits (number = 256) The Keccak width
- */
-export const keccak = function (a: Buffer, bits: number = 256): Buffer {
-  assertIsBuffer(a)
+const _keccak = function (a: Buffer, bits: number): Uint8Array {
   switch (bits) {
     case 224: {
       return keccak224(a)
@@ -28,6 +23,15 @@ export const keccak = function (a: Buffer, bits: number = 256): Buffer {
       throw new Error(`Invald algorithm: keccak${bits}`)
     }
   }
+}
+/**
+ * Creates Keccak hash of a Buffer input
+ * @param a The input data (Buffer)
+ * @param bits (number = 256) The Keccak width
+ */
+export const keccak = function (a: Buffer, bits: number = 256): Buffer {
+  assertIsBuffer(a)
+  return toBuffer(_keccak(a, bits))
 }
 
 /**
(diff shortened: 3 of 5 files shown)


acolytec3:
> 2. isValidPrivate tests are failing: noble does not throw on any invalid private, `isValidPrivateKey` is always either true or false. What should I do?
>

@paulmillr Just to confirm my understanding, the `isValidPrivateKey` method in `@noble/secp256k1` never throws.  From looking at the code, it always returns false if the key isn't valid, regardless of the reason.  It's just a simple true/false.

I think on this question, we can just update our tests so it just follows the behavior in `@noble/secp256k1`.  @ryanio @holgerd77 is there a reason we need to have both `false` and `throws` as possible outcomes in our `isValidPrivate` method?  To my way of thinking, if you provide an invalid private key, it should just return false, period.  No need to throw an error for some cases and not others.

acolytec3:
Provided there's a yes vote from @ryanio and/or @holgerd77, I've got a commit ready that fixes our tests.  We can safely disregard the Hardhat test failure on this PR.

paulmillr:
@acolytec3 correct

ryanio:
returning false rather than throwing for isPrivateKey seems fine, that would be what I would expect, but we should probably note it in the changelog as part of breaking changes just in case downstream users would want to update how they use the function (e.g. removing a try/catch clause, to just react on the Boolean)

holgerd77:
Meta-Comment: I was thinking for some time if it would make sense to do a broader overhaul here on the Util side for the next breaking release and throw out all the `hash` and `signature` related functionality out of `Util` (so basically: delete the modules) as well as the re-exports and go to `ethereum-cryptography` directly. What do others think here?

(for the tx signature stuff, as far as I've looked (but roughly: only a first look) this [code](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/src/signature.ts#L25) is only complex due to this legacy v +27/28 stuff. This is mostly only used for legacy tx and we might be able to transfer this (encapsulate this) along the legacy tx code directly.)

All this came to my mind after reading this MetaMask crypto chain analysis here https://blog.cryptographyengineering.com/2022/01/14/an-extremely-casual-code-review-of-metamasks-crypto/ from this security professor where we are also in and I realized that all this is really long and now our `ethereumjs-util` step might be really unnecessary with `ethereum-cryptography` providing a stable ground.

WDYGT? 🤔

Then this would need some broader reflection before we can proceed and eventually adopt here, the above is for sure a substantially deeper-reaching change (also still needs some additional confirmation, especially regarding the Util signature code).

paulmillr:
@holgerd77 sounds great for a future PR, that was exactly my thinking - too many boilerplate methods

acolytec3:
Sounds good to me, though I second @paulmillr in that it should be a future PR.  I vote we merge this in her and then do a systematic review of the whole monorepo and switch out the methods then.

holgerd77:
> returning false rather than throwing for isPrivateKey seems fine, that would be what I would expect, but we should probably note it in the changelog as part of breaking changes just in case downstream users would want to update how they use the function (e.g. removing a try/catch clause, to just react on the Boolean)

Yes, can second that (so all said here).

I guess I'll take a whole week off for the changelogs on this release round anyhow. 😅

packages/util/src/account.ts
```diff
@@ -244,14 +239,16 @@ export const isValidPublic = function (publicKey: Buffer, sanitize: boolean = fa
   assertIsBuffer(publicKey)
   if (publicKey.length === 64) {
     // Convert to SEC1 for secp256k1
-    return publicKeyVerify(Buffer.concat([Buffer.from([4]), publicKey]))
+    // Automatically checks whether point is on curve
+    Point.fromHex(Buffer.concat([Buffer.from([4]), publicKey]));
+    return true;
```
ryanio, line 244: will this throw if invalid? should we think about returning false instead, to match isValidPrivate now only returning boolean?

holgerd77: Yes, having the same question on review. 🙂

holgerd77:
@acolytec3 Please really do not merge in any updates from the `develop` branch into the `develop` feature branches, the only-rebase policy applies for both the `develop` branch itself as well as the `develop` feature branches. So updates on the feature branch needs to be done via rebase.

So please don't use the plain "Update branch" button for `develop` work. In this case the merge commit is an empty commit [9f9ebb2](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1714/commits/9f9ebb2c722952d405a54a66ee8fd4b97cd0e25d), so it might not do any harm. Already feel a bit uneasy with this though TBH (after sleeping one or two more days on this I might request to redo the PR, not sure if I am comfortable to remain "as is").

The GitHub branch update button for some week now also allows for branch updates via rebases:

grafik

We can experiment a bit with that instead and see how it goes.

paulmillr:
I've removed noble-secp since we've released ec 1.0.2. I think it's ready to be merged now.

packages/util/src/account.ts
```diff
@@ -240,7 +240,12 @@ export const isValidPublic = function (publicKey: Buffer, sanitize: boolean = fa
...
     // Convert to SEC1 for secp256k1
     // Automatically checks whether point is on curve
-    Point.fromHex(Buffer.concat([Buffer.from([4]), publicKey]))
+    try {
+      Point.fromHex(Buffer.concat([Buffer.from([4]), publicKey]))
+      return true
+    } catch (e) {
+      return false
+    }
     return true
```
ryanio, line 249: this line can be removed to satisfy the linter

packages/util/src/account.ts
```diff
@@ -244,14 +239,25 @@ export const isValidPublic = function (publicKey: Buffer, sanitize: boolean = fa
...
   if (!sanitize) {
     return false
   }
 
-  return publicKeyVerify(publicKey)
+  try {
+    Point.fromHex(publicKey)
+    return true
+  } catch (e) {
+    return false
```
ryanio, line 259: this also seems to be a last uncovered line, if it's possible to add an invalid non-64 byte public key length, then diff coverage should be 100% :)

ryanio:
wohoo!
Screen Shot 2022-02-18 at 10 03 29 AM

ryanio:
Looks like we have a failing test

```
# file: stTransactionTest/PointAtInfinityECRecover.json test: PointAtInfinityECRecover
not ok 11782 [ 0.018 secs ] the state roots should match (successful tx run)
```

in `packages/vm` run `npm run test:state -- --fork=Berlin --test=PointAtInfinityECRecover`

not sure what's up, looks like the `ecrecover` usages in tx package are all wrapped with try/catch 🤔

jochem-brouwer:
Weird how that test only fails on Berlin and not on London.

paulmillr:
It runs fine on my machine.


```
vm master ❯ npm run test:state -- --fork=Berlin --test=PointAtInfinityECRecover
> @ethereumjs/vm@5.7.1 test:state
> npm run tester -- --state "--fork=Berlin" "--test=PointAtInfinityECRecover"
> @ethereumjs/vm@5.7.1 tester
> ts-node ./tests/tester --stack-size=1500 "--state" "--fork=Berlin" "--test=PointAtInfinityECRecover"

TAP version 13
# GeneralStateTests
# file: stTransactionTest/PointAtInfinityECRecover.json test: PointAtInfinityECRecover
ok 1 [ 0.031 secs ] the state roots should match (successful tx run)

1..1
# tests 1
# pass  1

# ok
```

paulmillr:
Have been debugging this for 4 hours.

1. Seems like i've found a bug in both current secp256k1 and noble-secp256k1. The bug is not really related to the error, but needs fixing. 
2. Found that the old impl returned short keys, noble-secp returns long pubkeys. Adjusted this.

Continuing...

paulmillr:
https://github.com/paulmillr/noble-secp256k1/releases/tag/1.5.4
https://github.com/ethereum/js-ethereum-cryptography/pull/36

This could fix it.

packages/util/src/account.ts
```diff
@@ -231,7 +226,7 @@ export const generateAddress2 = function (from: Buffer, salt: Buffer, initCode:
  * Checks if the private key satisfies the rules of the curve secp256k1.
  */
 export const isValidPrivate = function (privateKey: Buffer): boolean {
-  return privateKeyVerify(privateKey)
+  return utils.isValidPrivateKey(privateKey)
```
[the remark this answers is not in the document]
paulmillr: The tests are already using `st.notOk`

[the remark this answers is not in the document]
ryanio: There are both

paulmillr:
Ready now @holgerd77

ryanio (approved):
LGTM!

holgerd77:
@paulmillr very cool, thanks a lot! 😄 👍 

Small thing: can you do a `rebase -i develop` and then delete this `develop-into-develop` merge commit and then do another force-push? I would want to make sure that this has not side effects on our `develop` branch work.

(if you think this is nonsense could you do nevertheless 😋 - sorry - I am just extra cautious since we are working on a dedicated `develop` branch for the first time with such a large group of people together and for now I wouldn't want to get anything unnecessary in there for the fear of side effects on rebases).

paulmillr:
@holgerd77 i've did the rebase, which screwed @acolytec3 commit (it became signed by my pgp key), then i've decided to just merge all of it into one.

paulmillr:
These failing `vm-benchmarks` is bullshit. Probably result of CI load. Here are results on my M1 mac:

- Old
```
Block 9422905 x 55,277 ops/sec ±4.84% (82 runs sampled)
Block 9422906 x 52,745 ops/sec ±1.09% (91 runs sampled)
Block 9422907 x 44,853 ops/sec ±8.28% (79 runs sampled)
Block 9422908 x 40,065 ops/sec ±12.29% (73 runs sampled)
Block 9422910 x 27,322 ops/sec ±40.03% (61 runs sampled)
```

- Upgraded
```
Block 9422905 x 54,547 ops/sec ±5.16% (76 runs sampled)
Block 9422906 x 53,623 ops/sec ±0.74% (90 runs sampled)
Block 9422907 x 45,192 ops/sec ±9.76% (81 runs sampled)
Block 9422908 x 39,797 ops/sec ±12.60% (71 runs sampled)
Block 9422910 x 35,623 ops/sec ±26.35% (66 runs sampled)
```

holgerd77 (approved):
Review done by @ryanio, had a short confirmational look after rebase (thanks Paul!). 

Will merge.

holgerd77:
> These failing `vm-benchmarks` is bullshit. Probably result of CI load. Here are results on my M1 mac:
> 
>     * Old
> 
> 
> ```
> Block 9422905 x 55,277 ops/sec ±4.84% (82 runs sampled)
> Block 9422906 x 52,745 ops/sec ±1.09% (91 runs sampled)
> Block 9422907 x 44,853 ops/sec ±8.28% (79 runs sampled)
> Block 9422908 x 40,065 ops/sec ±12.29% (73 runs sampled)
> Block 9422910 x 27,322 ops/sec ±40.03% (61 runs sampled)
> ```
> 
>     * Upgraded
> 
> 
> ```
> Block 9422905 x 54,547 ops/sec ±5.16% (76 runs sampled)
> Block 9422906 x 53,623 ops/sec ±0.74% (90 runs sampled)
> Block 9422907 x 45,192 ops/sec ±9.76% (81 runs sampled)
> Block 9422908 x 39,797 ops/sec ±12.60% (71 runs sampled)
> Block 9422910 x 35,623 ops/sec ±26.35% (66 runs sampled)
> ```

Yes, our benchmarks for the VM are not reliable.

PR #1719, opened 2022-02-16
VM, Common: implement eip-3540 (EOF1)

### LINKED ISSUE #1704: VM: Implement EVM Object Format (EOF) (EIP-3540)
[EIP-3540](https://eips.ethereum.org/EIPS/eip-3540) is long in the making and moved to CFI (Considered-for-Inclusion) for the `Shanghai` HF with strong developer backing in [ACD 131](https://github.com/ethereum/pm/issues/459) (February 4, 2022), so this should be safe to implement on our side as well.

The EIP introduces a container format for EVM bytecode which allows for validation of this code at the time of deployment (pros and cons of deployment vs. runtime validation in the EIP). This will allow to more easily add new features to the EVM and deprecate existing feature and/or bytecodes.

For the VM we need to implement the switches from the [Code Validation](https://eips.ethereum.org/EIPS/eip-3540#code-validation) section from the EIP for contract creation txs and `CREATE` or `CREATE2` instructions, likely to happen somewhere in/around the [evm._executeCreate()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/evm/evm.ts#L296) method.

For EOF(1) identified bytcode the [validation rules](https://eips.ethereum.org/EIPS/eip-3540#validation-rules) need to be applied. There are also some changes to the VM [execution semantics](https://eips.ethereum.org/EIPS/eip-3540#changes-to-execution-semantics) which needs to be considered, e.g. things need to run on `code` (e.g. "JUMPDEST analysis is only run on the code.") and not the whole container thing.

Test for this EIP are in the works here: https://github.com/ethereum/tests/pull/847

Two additional implementation notes:

1. Work should very likely be done towards `master`, do not expect anything breaking here
2. Please do not introduce a new HF `shanghai` here yet (or assign this functionality) - too early in the process - but limit to an EIP which can be activated or not

acolytec3:
Fixes #1704 

This adds support for EIP-3540.  This EIP is not part of any hardfork and can only be activated in concert with 3541.  At present, this implements EVM Object Format V1(EOF1) checks as a single utility function in `src/evm/opcodes/util.ts`.  There was some significant discussion around keeping all of the specific parameters governing the EOF1 checks as parameters in the EIP specification in `Common` but after working through it, it feels like it makes more sense (at least based on the current status of the EIP) that we just incorporate all of those parameters as hard coded values in a utility function that validates the code passed in based on the EOF1 rules (rather than try to generalize to a generic EOF validation function).  If/when future EOF versions come up for review/implementation, we can revisit this.

There are a couple of other changes made to the VM internals, notably:
- The interpreter's `runState.code` is now set based on whether or not the bytecode passed with the message is EOF1 compliant or not (code section of EOF1 container if EOF1 bytecode or entire `message.code` for legacy code)
- This allows for a change in the `PUSH` opcode handler that checks to see if it goes out of bounds and reverts if so (this doesn't seem to break API tests and gives a guard in EOF1 compliant bytecode to going out of bounds)
- LIkewise Jump Destination analysis is now limited to the `runState.code` which will be set as the code section of the EOF1 container if the bytecode is EOF1 compliant.

The [ethereum/tests PR containing tests for this EIP](https://github.com/ethereum/tests/pull/847) is still WIP as well so we should probably wait to merge that PR gets merged or else feels stable.


packages/vm/src/evm/opcodes/util.ts
```diff
@@ -261,3 +261,35 @@ export function updateSstoreGas(
...
+      code.length > 10 &&
+      code[3] === secCode &&
+      code[6] === secData &&
+      code[9] === secTerminator
+    ) {
+      codeSize = 10 + ((code[4] << 8) | code[5]) + ((code[7] << 8) | code[8])
+      sectionSizes.code = (code[4] << 8) | code[5]
+      sectionSizes.data = (code[7] << 8) | code[8]
+    }
+    if (code.length !== codeSize) {
```
jochem-brouwer, line 288: There is no check for the data size, what if this underflows or overflows?

packages/vm/src/evm/opcodes/util.ts
```diff
@@ -261,3 +261,35 @@ export function updateSstoreGas(
...
+    code: 0,
+    data: 0,
+  }
+  if (code[1] === 0x00 && code[2] === 0x01) {
+    if (code.length > 7 && code[3] === secCode && code[6] === secTerminator) {
+      codeSize = 7 + ((code[4] << 8) | code[5])
+      sectionSizes.code = (code[4] << 8) | code[5]
+    } else if (
+      code.length > 10 &&
+      code[3] === secCode &&
```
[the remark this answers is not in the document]
acolytec3: >The EIP does not specify an order of the sections, so `code[3]` could also be `secData`

I don't believe this is correct.  The EIP validation rules (excerpted below) require an order.  Also, I'm following the reference implementation here.
"
Exactly one code section MUST be present.
The code section MUST be the first section.
A single data section MAY follow the code section.
"

I can't link to the section directly but if you look under the validation rules, it details the specifics about EOFv1 rules.

jochem-brouwer:
@axic I have raised some questions about this EIP on https://ethereum-magicians.org/t/evm-object-format-eof/5727/28 😄

packages/vm/src/evm/evm.ts
```diff
@@ -408,7 +408,16 @@ export default class EVM {
         this._vm._common.isActivatedEIP(3541) &&
         result.returnValue.slice(0, 1).equals(Buffer.from('EF', 'hex'))
       ) {
-        result = { ...result, ...INVALID_BYTECODE_RESULT(message.gasLimit) }
+        if (!this._vm._common.isActivatedEIP(3540)) {
+          result = { ...result, ...INVALID_BYTECODE_RESULT(message.gasLimit) }
+        }
+        // EIP-3540 EOF1 checks
+        if (!eof1CodeAnalysis(result.returnValue)) {
```
jochem-brouwer, line 415: I don't think I like this function signature, so this code path will run if the `eof1CodeAnalysis` returns `undefined`, but I'm also not super sure what would make this more clean... 🤔

acolytec3:
>I would suggest a loop over these sections, this is also forward-compatible if new sections get introduced.

I want to discuss this question more generally than in the specific code comment.  I'm open to something like this (and my original implementation was more along this line) but I decided after going through it that it made more sense (at least to me) that we implement separate functions for each EOF version to check code validity.  That way, we don't have to be parsing common parameters each time we do this code check.  Given we don't know what the future EOFs may look like, there may be additional handling that is not currently envisaged that we will need to incorporate and it feels like unnecessary work at the present time to try and predict what form that will take and generalize our code in some other way.  My thought is if/when a new EOF v2 is added to the protocol, we can just look at that version byte in the header and call the appropriate code validation helper.

packages/vm/src/evm/opcodes/util.ts
```diff
@@ -261,3 +261,35 @@ export function updateSstoreGas(
...
+  const secCode = 0x01
+  const secData = 0x02
+  const secTerminator = 0x00
+  let codeSize = 0
+  const sectionSizes = {
+    code: 0,
+    data: 0,
+  }
+  if (code[1] === 0x00 && code[2] === 0x01) {
+    if (code.length > 7 && code[3] === secCode && code[6] === secTerminator) {
```
jochem-brouwer, line 275: What if `code.length` is 7 or less? (Then this EOF format is invalid, right? - it does not throw here?)

Is 7 correct?

Format, magic, version, 0x01, uint16 -> 6 bytes. It is invalid because code is now 0, but it would also not throw here

acolytec3: >What if code.length is 7 or less? (Then this EOF format is invalid, right? - it does not throw here?)

You also need the `secTerminator` which is 00 so a compliant header would look like `ef 00 01 01 00 01 00` which would indicate an EOF1 with a one byte code section expected after the section terminator.

packages/vm/src/evm/opcodes/util.ts
acolytec3: Ah, sorry, the naming here is imprecise. `codeSize` is the computed length of the entire bytestring based on the size of the header + the size of the code section and the data section.  I'll clean up the naming to make things a little more clear.

packages/vm/src/evm/interpreter.ts
```diff
@@ -87,9 +93,29 @@ export default class Interpreter {
...
+    if (
+      this._vm._common.isActivatedEIP(3540) &&
+      code.slice(0, 3).equals(Buffer.from('ef0001', 'hex'))
+    ) {
+      // Code is EOF1 format
+      const codeSections = eof1CodeAnalysis(code)
+      if (!codeSections) {
+        return {
+          runState: this._runState,
+          exceptionError: new VmError(ERROR.INVALID_EOF_FORMAT),
```
jochem-brouwer, line 105: Is this in the right place? These checks should be done at contract creation, right? Because this would OOG the `EF` contracts on mainnet (the 3 stray contracts)

jochem-brouwer: I just realized that the EIP also states that you can have EOF in `initCode`

packages/vm/src/evm/opcodes/util.ts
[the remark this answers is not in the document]
jochem-brouwer: As pointed out by @acolytec3 the first section should be the code section so in version 1 of EOF then the second section can only be the data section.

acolytec3:
> * `section_size` MUST NOT be 0 -> not present?

Ah, thanks for the catch.  Addressed in my latest commit.

> * Version check is not implemented

This is implemented [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/7e8aac6b0b44d07f470eded0a16a330c137871f2/packages/vm/src/evm/opcodes/util.ts#L276),  I fixed the names so it would be more obvious

> * Program going out-of-bounds (into data section or at end of code section) not present

I guess I thought this would be taken care of by the PC counter checks in the `interpreter.run` method but will think a little more on this.  I'll have to think on this some more.

> * Not entirely sure if `PC` opcode now returns the right value (but I think the implementation with this new `code` field is the way to go on the `runState`)

jochem-brouwer:
This PR was merged into this one: #1743

packages/vm/src/evm/opcodes/util.ts
```diff
@@ -358,12 +358,12 @@ export const eof1ValidOpcodes = (code: Buffer) => {
...
       x += opcode - 0x5f
       if (x > code.length - 1) {
-        // Push blocks mmust not exceed end of code section
+        // Push blocks must not exceed end of code section
         return false
       }
     }
   }
-  const terminatingOpcodes = new Set([0x00, 0xd3, 0xfd, 0xfe, 0xff])
+  const terminatingOpcodes = new Set([0x00, 0xf3, 0xfd, 0xfe, 0xff])
```
jochem-brouwer, line 366: Oh wow, missed this 👍

acolytec3: Yeah, me too. But writing those new tests you suggested helped me catch it 🎆

acolytec3:
Hmm, it seems that I somehow mixed up `develop` and `master` in these branches.  :thinking: Will try and clean it up with an interactive rebase.

acolytec3:
Okay, this is now targeted towards master and I've purged out all the develop commits that accidentally got pulled in and we're back to current state.

jochem-brouwer:
Hi @acolytec3 did you see this? https://github.com/ethereum/tests/pull/847

acolytec3:
> Hi @acolytec3 did you see this? [ethereum/tests#847](https://github.com/ethereum/tests/pull/847)

Yes, I haven't run them since the latest update but about 75% of the tests were passing last I checked (I think some of it was invalid test data but I'm sure some is edge cases in my code). Will run them again and see what fails.

acolytec3:
I'm going to tentatively call this PR ready for review.  I know I haven't addressed all of the feedback @jochem-brouwer gave under the other PR, notably around adding some more tests.  I'm reproducing the list below with some comments.  Lines with strikethrough should be covered by my current tests.  I would also point out that the code in this PR currently passes all of the draft tests [here](https://github.com/ethereum/tests/pull/847) :
- ~~Add a test for the uncovered codecov line (code section does not end with a terminating instruction)~~
- ~~Add a test where the code section ends with 0xFE (INVALID) (this is a sanity check)~~
- ~~Add a check for each terminating instruction (00, F3, FD, FF) (also a sanity check)~~
- Tests for a "create transaction" (so a transaction from an EOA which to field is empty) which tries to deploy a contract using invalid code (so it does not stop with a trunacting opcode, truncated PUSHn instruction, or it has a non-assigned instrucition besides INVALID/0xFE) and verifies that this consumes all gas
- - On this one, the above interpretation isn't actually correct behavior based on [this comment](https://github.com/ethereum/tests/pull/847#issuecomment-1056764081) from @gumb0 
- Same as above but now the code is contained in CREATE/CREATE2
- - I would like some assistance here as I'm not experienced writing bytecode and don't know the ins and outs of the CREATE/CREATE2 opcodes.  I'm going to experiment further with crafting some bytecode to cover these scenarios but will need some time to learn how these particular opcodes work (that is different than the standard `create contract` transaction.

holgerd77:
@acolytec3 I guess this should be ok, seems the mentioned test cases have been added in the `ethereum/tests` PR.

Congrats, great work here! 😄

Just thinking out loud, for a next round of releases I guess it fits well if we somewhat synchronize around the upcoming Kiln testnet and do a release round once our client implementation is prepared for this. @ryanio @g11tech any context on a timeline here and some insight in our status? Are we likely compatible with the latest work from #1750? Or is this likely/eventually not yet the last stage before the Kiln launch (when is this roughly planned)? 🤔

@jochem-brouwer @acolytec3 would you guys say we can already publish this work here delclared as `stable` or should we be a bit cautious and still lable `experimental` (that's always in the sense of: you can use, but still expect some changes) (downside: another round of iteration and some overhead, so if we are confident there is also some good argument to directly declare stable).

acolytec3:
As far as the PR being stable, I don't get the impression that the associated EIPs are likely to change at this point so once any reviews are complete, I'd say this one is ready.

g11tech:
@holgerd77 the kiln V2 branch should be ready , however will run against test vectors and sim tests today and report. Will also start an instance for the kiln. V2 testnet, it should hit merge tomorrow as TTD has been set quite high this time around. 🙂

jochem-brouwer:
I'd say that we should mark it as stable, at this point we match the EIP. But, we might mention that the EIP itself might change. Right now we match the spec, so that'd be stable.

jochem-brouwer:
There are some steps left to do for this PR which are essentially listed in this https://github.com/ethereumjs/ethereumjs-monorepo/pull/1719#issuecomment-1058760853 comment. I can help creating that CREATE/CREATE2 bytecode. We should also update the case where the "intrinsic gas" is consumed (data fee + upfront fee) in case we try to de a create transaction where the EOF format is invalid. I have pinged an author to clarify/update the EIP regarding what "abort" means in terms of gas.

(Also seems that there is a linting problem)

Side note: once the EIP-specific tests get merged in Ethereum/tests we should also add those forks to CI in case we run "test all hardforks".

jochem-brouwer:
Have added invalid eof initcode tests for tx/create/create2.

jochem-brouwer (approved):
As far as I am concerned this is ready to merge, let's wait for one more review.

holgerd77:
Ah, ok, interesting (and: cool!). 👍 😄 🎉

I would then have a strong tendency to integrate this in the current round of releases, seems to be a very good occasion to directly get this out.

Will do a final review on this and merge if everything is ok and add to #1774.

I would then try to do the releases itself tomorrow, Tuesday (CET) morning - also depending on the outcome of #1781 and a final "go!" from @g11tech and @ryanio, and then do a first Twitter thread directly that day solely focusing on the Kiln aspects of the release round.

This would be followed up by a bugfix release round either at the end of the week or early next week, fixing some things from Kiln + adding the merge blocks and stuff and on *that* release occasion I would do another Twitter thread (more or less) solely focusing on all other release features (EIP-3860, Extended VM Genesis, EIP-3540,...).

So far on the current plan [TM] 😎.

packages/vm/src/evm/opcodes/util.ts
```diff
@@ -261,3 +262,70 @@ export function updateSstoreGas(
...
+      // Code & Data sizes cannot be 0
+      if (sectionSizes.code < 1 || sectionSizes.data < 1) return
+    }
+    if (container.length !== computedContainerSize) {
+      // Scanned code does not match length of contract byte code
+      return
+    }
+    return sectionSizes
+  }
+}
```
holgerd77, line 301: This is checking on so many conditions in such a condensed form that it is extremely hard to review and generally follow. Can we have this dramatically [TM] more commented, so basically to a comment with a condition on more or less every line of code? Guess that would make things easier to grasp.

Also: one technique @ryanio often applies and I guess which would help here as well is to instead of checking for the positive condition (`if (container[1] === magic && container[2] === version) {`) and then nesting further down check for the negative (`if (container[1] !== magic || container[2] !== version) {`) and the directly do the `return`. This prevents further nesting and could be minimally applied on this example I made but I guess also on other code parts.

holgerd77:
Oh, that's a really great update! Thanks a lot! 👍 😍

holgerd77:
I can't find/identify all parts of spec integrations in this huge EIP TBH, so I'll go with some basic questions for @acolytec3 just for re-assurance for a second review:

So first round would be: are you confident that you have "caught" all execution semantics changes from this list:


1. JUMPDEST analysis is only run on the code.
2. Execution starts at the first byte of the code, and PC is set to 0.
3. If PC goes outside of the code section bounds, execution aborts with failure.
4. PC returns the current position within the code.
5. JUMP/JUMPI uses an absolute offset within the code.
6. CODECOPY/CODESIZE/EXTCODECOPY/EXTCODESIZE/EXTCODEHASH keeps operating on the entire container.
7. The input to CREATE/CREATE2 is still the entire container.

Or is there one where one should have another eye upon and do some verification?

packages/vm/src/evm/evm.ts
```diff
@@ -422,7 +430,36 @@ export default class EVM {
...
+        } else if (this._vm._common.isActivatedEIP(3670)) {
+          // EIP-3670 EOF1 opcode check
+          const codeStart = eof1CodeAnalysisResults.data > 0 ? 10 : 7
+          // The start of the code section of an EOF1 compliant contract will either be
+          // index 7 (if no data section is present) or index 10 (if a data section is present)
+          // in the bytecode of the contract
+          if (
+            !eof1ValidOpcodes(
+              result.returnValue.slice(codeStart, codeStart + eof1CodeAnalysisResults.code)
+            )
```
holgerd77, line 453: This is from the `EIP-3670` spec:

> At contract creation time both initcode and code are iterated instruction-by-instruction

Might it be that you are doing validation only on the code and not the initcode? Or am I missing something?

acolytec3:
Thanks for the feedback.  I'll address these as best I can but happy to dig into things further(and @jochem-brouwer keep me honest on this).

> 1. JUMPDEST analysis is only run on the code.
> 2. Execution starts at the first byte of the code, and PC is set to 0.
> 3. If PC goes outside of the code section bounds, execution aborts with failure.
> 4. PC returns the current position within the code.
> 5. JUMP/JUMPI uses an absolute offset within the code.

I believe these 5  are all implicitly addressed by [this line of code]((https://github.com/ethereumjs/ethereumjs-monorepo/blob/80933d2eca3da4c650a3b293237abb3ed517c95b/packages/vm/src/evm/interpreter.ts#L127)

If the code passed to the interpreter contains an EOF1 header, we define the `runState.code` to be interpreted as just the code section of the container so JUMPDEST analysis, code execution starting point, PC position, and JUMP/JUMPI offset are all conducted on this code only.

For JUMP/JUMPI offsets, I think you might have a point that we missed a step in checking [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/38a31411edd9bbd580211e147eabef1a59e9a726/packages/vm/src/evm/opcodes/functions.ts#L725) since check against `runState.eei.getCodeSize()` is looking at the entire code and not the code section.  I think this can be fixed by just checking against `runState.code.length`.  That said, this check feels redundant since it would throw on an invalid jump destination anyway the next line down.  @jochem-brouwer  - what do you think on this?

> 6. CODECOPY/CODESIZE/EXTCODECOPY/EXTCODESIZE/EXTCODEHASH keeps operating on the entire container.

I looked at all 4 of these function handlers again and they are all checking against `runState.eei.getCodeSize`/`runState.eei.getCode` or equivalents so they are operating on the container.  The EOF1 code section is found in `runState.code` so we should be good here.

> 7. The input to CREATE/CREATE2 is still the entire container.

From looking at the opcode handlers, the inputs are entirely dependent on what is passed to them as [here for CREATE](https://github.com/ethereumjs/ethereumjs-monorepo/blob/38a31411edd9bbd580211e147eabef1a59e9a726/packages/vm/src/evm/opcodes/functions.ts#L882).  Whatever offset/length are popped from the stack determines what code is pulled in but once that code is passed in, it is treated as an entire container and analyzed for EOF1 validity in the same way as a basic create transaction.  So, the long answer to your short question is "yes" as far as "the entire container" corresponds to whatever code is passed to `_executeCreate` as determined by the inputs to the CREATE/CREATE2 opcode handler.

I'd also like @jochem-brouwer to comment on this one.  He wrote the tests that validate the `CREATE`/`CREATE2` behavior.  

Let me know if this addresses your questions.

packages/vm/src/evm/evm.ts
acolytec3: This is checking the code returned by the create contract to verify its valid .  The initcode is checked when it is run through the `runInterpreter` function call.

holgerd77:
> Let me know if this addresses your questions.

Totally, thanks a lot, explanations couldn't be better!

Ok, I guess I am confident enough here that I would do the following, since timing of this whole release round is really a bit challenging with all different requirements: I will merge this in now and - directly - release with the Kiln updates.

In case that we stumble upon some additional things post-release we can still add this to the very imminent bugfix releases (also necessary for some Kiln updates), to be released directly on Friday or early next week.

I guess that should catch all release requirements best, so we have the Kiln release out in time and also do not miss this release opportunity on the work here but still have the time for a short-termed correction if necessary (which should be very much ok given that we release as `experimental` anyhow + we are already somewhat confident on the implementation).

//cc @g11tech @ryanio @jochem-brouwer

holgerd77 (approved):
LGTM

PR #1720, opened 2022-02-16
Client: Geth genesis parser minor fix

cbrzn:
We had a logic where if the london block was zero and the `baseFeePerGas` undefined, we would set the `baseFeePerGas` to `1000000000`, otherwise, it would be undefined, this would not allow us to set the `baseFeePerGas` on the genesis initialization when parsing it with geth format. Now this logic has been moved to the block package (It was done by Ryan a few weeks ago)

### DIFF
--- a/packages/client/lib/util/parse.ts
+++ b/packages/client/lib/util/parse.ts
@@ -166,15 +166,12 @@ async function createGethGenesisBlockHeader(json: any) {
  * @returns genesis parameters in a `CommonOpts` compliant object
  */
 async function parseGethParams(json: any) {
-  const { name, config, difficulty, nonce, mixHash, coinbase } = json
+  const { name, config, difficulty, nonce, mixHash, coinbase, baseFeePerGas } = json
 
-  let { gasLimit, extraData, baseFeePerGas, timestamp } = json
+  let { gasLimit, extraData, timestamp } = json
 
   // geth stores gasLimit as a hex string while our gasLimit is a `number`
   json['gasLimit'] = gasLimit = parseInt(gasLimit)
-  // geth assumes an initial base fee value on londonBlock=0
-  json['baseFeePerGas'] = baseFeePerGas =
-    baseFeePerGas === undefined && config.londonBlock === 0 ? 1000000000 : undefined
   // geth is not strictly putting in empty fields with a 0x prefix
   json['extraData'] = extraData = extraData === '' ? '0x' : extraData
   // geth may use number for timestamp
--- a/packages/client/test/util/parse.spec.ts
+++ b/packages/client/test/util/parse.spec.ts
@@ -102,4 +102,16 @@ tape('[Util/Parse]', (t) => {
       'consensus config matches'
     )
   })
+
+  t.test(
+    'should generate expected hash with london block zero and base fee per gas defined',
+    async (t) => {
+      const json = require('../testdata/post-merge.json')
+      const params = await parseCustomParams(json, 'post-merge')
+      t.equals(
+        params.genesis.hash,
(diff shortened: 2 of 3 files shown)


packages/client/test/util/parse.spec.ts
```diff
@@ -102,4 +102,16 @@ tape('[Util/Parse]', (t) => {
...
   })
+
+  t.test(
+    'should generate expected hash with london block zero and base fee per gas defined',
+    async (t) => {
+      const json = require('../testdata/post-merge.json')
+      const params = await parseCustomParams(json, 'post-merge')
+      t.equals(
+        params.genesis.hash,
+        '0x3b8fb240d288781d4aac94d3fd16809ee413bc99294a085798a589dae51ddd4a'
```
jochem-brouwer, line 113: Cool! I think for clarity, it would be great if you add an extra check that the base fee of this block is correct as defined in `post-merge.json`. (This is of course part of the hash, and is thus implicitly included in the test, but this would improve clarity)

jochem-brouwer (commented):
Great PR, one comment.

packages/client/lib/util/parse.ts
```diff
@@ -166,15 +166,12 @@ async function createGethGenesisBlockHeader(json: any) {
...
+  const { name, config, difficulty, nonce, mixHash, coinbase, baseFeePerGas } = json
 
-  let { gasLimit, extraData, baseFeePerGas, timestamp } = json
+  let { gasLimit, extraData, timestamp } = json
 
   // geth stores gasLimit as a hex string while our gasLimit is a `number`
   json['gasLimit'] = gasLimit = parseInt(gasLimit)
-  // geth assumes an initial base fee value on londonBlock=0
-  json['baseFeePerGas'] = baseFeePerGas =
-    baseFeePerGas === undefined && config.londonBlock === 0 ? 1000000000 : undefined
```
jochem-brouwer, line 177: Hmm upon thought, do we want to keep this default value in case it is not present in the json? @ryanio

packages/client/test/util/parse.spec.ts
cbrzn: Sure!

packages/client/lib/util/parse.ts
cbrzn: If I understand this correctly, Ryan already handled it by added that logic here https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/header.ts#L267-L272, being the default value defined here https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/eips/1559.json#L19 (which is the same as the previous code we had here) - Probably I am wrong tho :thinking:

jochem-brouwer: Nope, good point, you are right 👍 😄

holgerd77:
Please let Ryan have a look on this as well before merging, I guess he has got some insight here on the original implementation.

packages/client/test/util/parse.spec.ts
```diff
@@ -106,12 +106,14 @@ tape('[Util/Parse]', (t) => {
...
     'should generate expected hash with london block zero and base fee per gas defined',
     async (t) => {
+      t.plan(2)
       const json = require('../testdata/post-merge.json')
       const params = await parseCustomParams(json, 'post-merge')
       t.equals(
         params.genesis.hash,
         '0x3b8fb240d288781d4aac94d3fd16809ee413bc99294a085798a589dae51ddd4a'
       )
+      t.equals(params.genesis.baseFeePerGas, json.baseFeePerGas)
```
jochem-brouwer, line 116: 👍

ryanio (approved):
Thanks, looks perfect!

PR #1721, opened 2022-02-17
VM: add StatelessVerkleStateManager (with map-based commits) (old)

holgerd77:
Part of #1533 
Implements #1578
Base branch for stateless verkle demo work

This replaces #1661 and continues #1596, so to create a dedicated stateless Verkle StateManager which takes in a pre-state from a block, then executes in a stateless way upon this pre-state and at the end provides the computed state changes to `verify_update` function from the [WASM build](https://github.com/crate-crypto/rust-verkle-wasm) of the rust verkle library.

Demo data taken from: https://github.com/gballet/verkle-block-sample

WIP


gabrocheleau:
Closing this in favour of #2328

PR #1723, opened 2022-02-19
client: Jwt token based auth for engine api

g11tech:
Kiln spec v2 has jwt based auth for the CL requests for engine methods with following validations
- jwt token auth
- iat (issued at) +- 5 seconds from current time
- `--jwt-secret` param for file

This PR intends to implement the same.

TODOS:
- [x] cleanup (console logs)
- [x] read  from file provided by --jwt-secret and validate its length as per spec requirement
- [x] if no secret provided, dump a secret in a file which can be grabbed and supplied to CL 
- [x] Unit tests


packages/client/bin/startRpc.ts
```diff
@@ -0,0 +1,193 @@
...
+  rpcport,
+  server,
+  withEngineMiddleware,
+}: {
+  rpcport: number
+  server: RPCServer
+  withEngineMiddleware?: { jwtEngineSecret: string }
+}): Application {
+  const app = Express()
+  app.use(jsonParser())
```
cbrzn, line 36: Nit: I think that here you can do `app.use(Express.json())` since it's already in the express library (https://expressjs.com/en/4x/api.html) - This way we don't need to add the `body-parser` dependency in our package json

g11tech:
> I know this is probably not fair to ask but is there any way we can avoid adding adding `express` and its associated dependencies just to verify JWTs? Would it be possible to use [jose](https://github.com/panva/jose) instead and just import the JWT dependency, rather than adding a whole new server (i.e. `express`) and the associated dependencies needed to do this step? I'm sure its a bit more code up front and I haven't worked with JWTs in a while (if ever) so not 100% clear on how difficulty what I'm asking is but I'd feel better about this if we weren't adding express along with the already existing jayson and web socket servers we have in the client just for the sake of one middleware validation step. Thoughts?

ummm... reason to add express was to primarily intercept the requests as jayson seems to not provide middleware-ing capability.
how about using `connect` to middleware the validation?

acolytec3:
Could work, or as simple as just using the built in [router](https://jayson.tedeh.net/3.6.5/#method-routing) capability to check the jwt on receipt using your preferred (dependency minimized jwt library) and then calling the right message handler

g11tech:
> Could work, or as simple as just using the built in [router](https://jayson.tedeh.net/3.6.5/#method-routing) capability to check the jwt on receipt using your preferred (dependency minimized jwt library) and then calling the right message handler

method routing doesn't seem to provide handle on req or headers for jwt token to be extracted or validated.

acolytec3:
Does [this](https://jayson.tedeh.net/3.6.5/#server-context) `useContext` provide what you need? I can dig into it tomorrow but looks like this might pass the headers in where the JWT would live, right?

g11tech:
> Does [this](https://jayson.tedeh.net/3.6.5/#server-context) `useContext` provide what you need? I can dig into it tomorrow but looks like this might pass the headers in where the JWT would live, right?

unfortunately, the libraries' examples of *building* that context itself requires using express/connect to plug in a json body parser.
https://github.com/tedeh/jayson/blob/master/examples/context/server.js 

If we totally want to avoid any extra dependency, we can catch inbuild's http req via http.createServer((req,res)=>{...}), collect data chunks, parse them (and validate the headers) and then call jayson's server.call to pass on the request for evaluation. Thats exactly what jayson is trying to do under the hood (minus the auth part)

acolytec3:
Hmm, okay. I still think we should consider minimizing the number of new dependencies we add in and lean on standard nodejs inbuilts where feasible, even if it is `express` and surrounds that are the new dependencies in question. @holgerd77 , @ryanio , any thoughts?

ryanio:
@acolytec3 appreciate you taking a look, jayson already uses express under the hood so it was already a transitive dependency, we just need to use it in this way to get deeper in the stack as the jayson readme suggests. I'll be reviewing this Pr today and see if we can simplify at all based on your suggestion, but it might be needed in this case.

acolytec3:
Fair. I didn't look down the dependency tree so not to worry if it's already there.

ryanio:
ok this looks pretty awesome! just went over all the code and had just a few things i included in some commits.

i did try to connect to lodestar but i don't think i have the right command handy, i pulled and built this branch https://github.com/ChainSafe/lodestar/pull/3777 - how should i run lodestar on kiln? our typical kintsugi run looked like this: `./lodestar beacon --rootDir kintsugi/temp --network=kintsugi --eth1.providerUrls=http://localhost:8545 --execution.urls=http://localhost:8545` i guess adding `--jwt-secret=/Users/me/Library/Ethereum/ethereumjs/jwtsecret`

i did have one question if it's possible to also do this auth with the websockets endpoint? i know some CL clients were communicating over ws

---

regarding not using express, I'm sympathetic as a way to reduce our dep count, so @g11tech if you want to see if it's possible without writing too much boilerplate code we should definitely consider it. perhaps give your suggested route a try and see if it's doable within a few hours and report back. if it gets needlessly complicated and we find express provides enough value in exchange for the added dep bloat, then we can take that route.

g11tech:
> ok this looks pretty awesome! just went over all the code and had just a few things i included in some commits.

Thank you :heart: 
> i did try to connect to lodestar but i don't think i have the right command handy, i pulled and built this branch [ChainSafe/lodestar#3777](https://github.com/ChainSafe/lodestar/pull/3777) - how should i run lodestar on kiln? our typical kintsugi run looked like this: `./lodestar beacon --rootDir kintsugi/temp --network=kintsugi --eth1.providerUrls=http://localhost:8545 --execution.urls=http://localhost:8545` i guess adding `--jwt-secret=/Users/me/Library/Ethereum/ethereumjs/jwtsecret`

yes this command would work, but lodestar master (and the above branch) is now incompatible with kintsugi because of kiln updates. However your merge-kiln PR can be reabased on top of this for interop with lodestar

> 
> i did have one question if it's possible to also do this auth with the websockets endpoint? i know some CL clients were communicating over ws
> 
let me dig into it
> regarding not using express, I'm sympathetic as a way to reduce our dep count, so @g11tech if you want to see if it's possible without writing too much boilerplate code we should definitely consider it. perhaps give your suggested route a try and see if it's doable within a few hours and report back. if it gets needlessly complicated and we find express provides enough value in exchange for the added dep bloat, then we can take that route.

Yes I agree, swapped express with `connect` middleware to auth /patch requests to jayson :slightly_smiling_face:

packages/client/test/rpc/rpc.spec.ts
```diff
@@ -1,23 +1,119 @@
...
+    .set('Authorization', `Bearer ${token}`)
+    .send(req)
+    .expect(401)
+    .end((err: any) => {
+      closeRPC(server)
+      t.end(err)
+    })
+})
+
+tape('call JSON-RPC auth protected server with an valid token', (t) => {
```
ryanio, line 69:
```suggestion
tape('call JSON-RPC auth protected server with a valid token', (t) => {
```

packages/client/test/rpc/rpc.spec.ts
```diff
@@ -1,23 +1,119 @@
...
+    .set('Authorization', `Bearer ${token}`)
+    .send(req)
     .expect(415)
     .end((err: any) => {
       closeRPC(server)
       t.end(err)
     })
 })
 
+tape('call JSON-RPC auth protected server with an valid but stale token', (t) => {
```
ryanio, line 86:
```suggestion
tape('call JSON-RPC auth protected server with a valid but stale token', (t) => {
```

ryanio (approved):
looks great! let's wait a few days before merging until we get either a new spec release or when your lodestar PR is merged, so it doesn't break current engine connections for devnet-4

ryanio:
@g11tech sorry i don't seem to be able to reopen this PR, but once i have merge-kiln-v2 branch up you can target that as base branch

PR #1727, opened 2022-02-21
Client: Engine API tests implementation

cbrzn:
This PR aims to add tests to the Engine API (https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/lib/rpc/modules/engine.ts#L168). Aiming mainly at the three methods implemented in it which are `newPayload`, `getPayload` and `forkchoiceUpdated`

Edit: This job has been inspired in [Lodestar tests](https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/test/sim/merge-interop.test.ts#L136-L232) & [this guide](https://notes.ethereum.org/rmVErCfCRPKGqGkUe89-Kg?view)

### DIFF
--- a/packages/client/lib/util/parse.ts
+++ b/packages/client/lib/util/parse.ts
@@ -165,7 +165,6 @@ async function parseGethParams(json: any) {
   const { name, config, difficulty, nonce, mixHash, gasLimit, coinbase, baseFeePerGas } = json
   let { extraData, timestamp } = json
   const { chainId } = config
-
   // geth is not strictly putting empty fields with a 0x prefix
   if (extraData === '') {
     extraData = '0x'
--- a/packages/client/lib/util/parse.ts
+++ b/packages/client/lib/util/parse.ts
@@ -174,7 +174,6 @@ async function parseGethParams(json: any) {
   if (!isHexPrefixed(timestamp)) {
     timestamp = intToHex(parseInt(timestamp))
   }
-
   // EIP155 and EIP158 are both part of Spurious Dragon hardfork and must occur at the same time
   // but have different configuration parameters in geth genesis parameters
   if (config.eip155Block !== config.eip158Block) {
--- a/packages/client/lib/util/parse.ts
+++ b/packages/client/lib/util/parse.ts
@@ -165,6 +165,7 @@ async function parseGethParams(json: any) {
   const { name, config, difficulty, nonce, mixHash, gasLimit, coinbase, baseFeePerGas } = json
   let { extraData, timestamp } = json
   const { chainId } = config
+
   // geth is not strictly putting empty fields with a 0x prefix
   if (extraData === '') {
     extraData = '0x'
@@ -173,6 +174,7 @@ async function parseGethParams(json: any) {
   if (!isHexPrefixed(timestamp)) {
     timestamp = intToHex(parseInt(timestamp))
   }
+
   // EIP155 and EIP158 are both part of Spurious Dragon hardfork and must occur at the same time
   // but have different configuration parameters in geth genesis parameters
   if (config.eip155Block !== config.eip158Block) {
--- a/packages/client/test/rpc/engine/newPayloadV1.spec.ts
+++ b/packages/client/test/rpc/engine/newPayloadV1.spec.ts
@@ -3,8 +3,8 @@ import { INVALID_PARAMS } from '../../../lib/rpc/error-code'
 import { params, baseRequest, baseSetup, setupChain } from '../helpers'
 import { checkError } from '../util'
 import genesisJSON from '../../testdata/geth-genesis/post-merge.json'
-import { FeeMarketEIP1559Transaction } from '@ethereumjs/tx'
-import { Address } from 'ethereumjs-util'
+import { FeeMarketEIP1559Transaction, Transaction } from '@ethereumjs/tx'
+import { Address, TWO_POW256 } from 'ethereumjs-util'
 
 const method = 'engine_newPayloadV1'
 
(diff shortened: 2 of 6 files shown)


packages/client/test/rpc/engine/newPayloadV1.spec.ts
```diff
@@ -40,16 +45,24 @@ tape(`${method}: call with invalid block hash without 0x`, async (t) => {
...
 })
+
+// tape(`${method}: call with valid data`, async (t) => {
+//   const { server } = await setupChain(genesisJSON, 'post-merge', { engine: true })
+//
+//   const req = params(method, [blockData])
+//
+//   const expectRes = (res: any) => {
+//     console.log(res)
+//     // t.equal(res.body.result, 'SYNCING')
```
cbrzn, line 65: When running this method, I am getting this error 
> Error verifying block while running: Error: BlockHeader.cliqueIsEpochTransition() call only supported for clique PoA networks (block header number=1 hash=0x3559e851470f6e7bbed1db474980683e8c315bfce99b2a6ef47c057c04de7858 hf=merge baseFeePerGas=7)'


It's being triggered when [`putBlock`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/test/engine-api/packages/client/lib/rpc/modules/engine.ts#L366) method is executed. After debugging for some time, I noticed that there's a mismatch between `this` and `header` variables 
```ts
this._common.consensusAlgorithm() -> 'clique'
header._common.consensusAlgorithm() -> 'casper'
```


https://github.com/ethereumjs/ethereumjs-monorepo/blob/c9d5fe97e2e4383bdd1d06a61b8c0aa80ced07ab/packages/blockchain/src/index.ts#L951
I think this is related (again) to the way we are instantiating the client

Will try to keep debugging later, but if someone has any idea I am all ears :ear:

ryanio: hm well for post-merge consensusAlgorithm() shouldn't return clique, since TTD is 0 it should always return casper. i checked and setupChain does call common.setHardforkByBlockNumber(0) so that seems right. which this._common.consensusAlgorithm() is returning clique? I'll have to check out the code and see what's going on.

cbrzn: After doing that, I noticed that the common `hardfork` variable is still `london`, not `merge`, and that's the reason why we are getting this missmatch

cbrzn: I found the problem! :-D Indeed we needed to pass the `TD` param from the Geth Config. I went deep and tried to understand why do we need this, and if I understood correctly, I can say that the reason of this mismatch behavior between the integration tests with lodestar and w/this unit tests, is that, since we are mocking the client, we need to explicitly set the the `hardfork` based on the block number; but when running the process from the CLI, it does this when the `Synchronizer` opens, which is this line https://github.com/ethereumjs/ethereumjs-monorepo/blob/merge-kiln/packages/client/lib/sync/fullsync.ts#L61

Would that make sense to you? @ryanio

ryanio: correct, nice find!

packages/client/lib/util/parse.ts
```diff
@@ -246,6 +246,7 @@ async function parseGethParams(json: any) {
       block: null,
     })
   }
+  params.td = terminalTotalDifficulty
```
ryanio, line 249: you don't need to add this because it's just a few lines up that pushes `td` into the hardfork array for `merge`

packages/client/test/rpc/helpers.ts
```diff
@@ -166,11 +166,12 @@ export async function baseRequest(
 export async function setupChain(genesisFile: any, chainName = 'dev', clientOpts: any = {}) {
   const genesisParams = await parseCustomParams(genesisFile, chainName)
   const genesisState = genesisFile.alloc ? await parseGenesisState(genesisFile) : {}
+
   const common = new Common({
     chain: chainName,
     customChains: [[genesisParams, genesisState]],
   })
-  common.setHardforkByBlockNumber(0)
+  common.setHardforkByBlockNumber(0, genesisParams.td)
```
ryanio, line 174: this would be the total terminal difficulty for the chain, not the difficulty for the genesis block, so you only need to do `genesisParams.difficulty`. please watch for #1741, as this will change to a number to better match up our libs, and then you can wrap with `new BN()` if needed. i wll rebase my kiln branch as well once merged.

ryanio: ok, have merged and updated merge-kiln branch

packages/client/lib/util/parse.ts
cbrzn: Interesting, you mean this line https://github.com/ethereumjs/ethereumjs-monorepo/blob/merge-kiln/packages/client/lib/util/parse.ts#L240-L246 ?

At first, I thought that we don't need it because of that line, but since the block is being defined as `null` and we were not passing the `totalTerminalDifficulty`; which was the reason for the hard fork being set to `london` and not to `merge`

Probably I am missing something? :thinking:

packages/client/test/rpc/helpers.ts
cbrzn: Correct, that's the `totalTerminalDifficulty` - Got inspired from [this line](https://github.com/ethereumjs/ethereumjs-monorepo/blob/merge-kiln/packages/client/lib/sync/fullsync.ts#L61) that I mentioned before; if I got this right, we should pass this variable instead of `difficulty`. Note that for `post-merge` mock data the block difficulty is `0x400000000` and not `0`

That's why I thought that we needed to pass this variable, just for the sake of setting the hard fork

Please let me know if this does not make sense

ryanio (approved):
looks amazing, thanks for your hard work and sticking through with this!

holgerd77:
Cool! 😄 🎉

PR #1730, opened 2022-02-21
Convert bn to bigint (old)

### LINKED ISSUE #1728: Monorepo: BigInt Input Value API Transition Suggestion (BNLike -> BigIntLike)
For our v6 releases #1717 we are planning a wider `BigInt` transition and so to both use `BigInt` values internally throughout our monorepo stack as well as for input values taken in the form of `BNLike` right now (we have more or less no BN-only inputs, just did some search).

This is a suggestion on how to do the latter one and what design decisions we need to take there.

So the [BNLike](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/src/types.ts#L9) type consists of the following four different input types which are allowed for library input:

- `BN`
- `PrefixedHexString`
- `number`
- `Buffer`

For an API transition we can start with replacing this type by a new type `BigIntLike`. I would suggest that we - regardless of any design decisions - on an implementation start this type as a superset of `BNLike` (so, this would be: everything from `BNLike` + `BigInt`) and then gradually go down on unwanted types. Then we can concentrate on a first round on the internal transition which would still be covered by all the tests.

Some suggestions/thoughts on the different types itself:

**PrefixedHexString/Buffer**: I have the thesis right now that we *do* want to keep both these types since the usage of the affected functionality is too diverse and the introduction of these two forms had some solid reasons and use cases (hex strings from JSON, Buffer from the wire/networking).

**number**: I am a bit unsure here. On the one hand this would remove some inconsistency with always being this hidden threshold baked in (JS MAX INTEGER) where things would start breaking. On the other hand this might take some handyness for several developer context/usage stuff where there are a lot of use cases where it is just clear that things are only used in a below-max-interger form (chainID the most obvious, developing with fake block numbers (1), ...). Have *some* weak tendency to remove but would love to hear some more opinions here.

**BN.js**: Also unsure if we want to do this harsh cut already and directly remove or keep this release for soften the transition. Good thing is: we can make this decision very much last-minute, once we have all the internal BigInt things going this will be an easy eventual remove.

paulmillr:
Closes #1728. Will continue after #1671 is merged.


packages/util/src/signature.ts
```diff
@@ -175,14 +188,15 @@ export const isValidSignature = function (
     return false
   }
 
-  const rBN = new BN(r)
-  const sBN = new BN(s)
+  const rBN = bufferToBigint(r)
+  const sBN = bufferToBigint(s)
+  const _0n = BigInt(0)
 
-  if (rBN.isZero() || rBN.gt(SECP256K1_N) || sBN.isZero() || sBN.gt(SECP256K1_N)) {
```
paulmillr, line 181: I think there was an error in the logic. 

`s.gt(n)` is `s > n`. It must be `s >= n`.

ryanio: oh, nice find, do you think you could add some test cases to `signature.spec.ts` that covers this discrepancy?

I'm a little surprised it wasn't in [`ethereum/tests`](https://github.com/ethereum/tests), might be a good suggestion for over there too, boundaries can be problematic and always good to be well tested.

ryanio: (also, @paulmillr if you confirm this is a bug, we should also plan to fix on master and include as part of a util v7.x patch release)

packages/common/src/index.ts
```diff
@@ -127,7 +127,7 @@ export interface CommonOpts extends BaseOpts {
    * either from a chain directly supported or a custom chain
    * passed in via {@link CommonOpts.customChains}.
    */
-  chain: string | number | Chain | BN | object
+  chain: string | number | Chain | bigint | object
```
holgerd77, line 130: Just for the record (nothing to be changed here imminently): these kind of API parts are very much like the `BNLike` ones and at the end we want to remain consistent here. So we either remove BN - as it is done here - or also keep like in `BigIntLike` right now.

But I guess the overall API needs some final consistency looks anyhow once the major work is done.

(side note: my own tendency towards removing `BN.js` completely is actually growing now seeing all this fresh air coming in. 😋 Why wait another year and *then* put the burden on the users to make this transition when we can very well do it now. And since many return values are now getting switched to `BigInt` anyhow, the argument "maybe someone wants to stay on BN.js" is also very much de-valued. Just not really possible anyhow."

packages/common/src/index.ts
```diff
@@ -780,9 +780,9 @@ export default class Common extends EventEmitter {
     // Logic: if accumulator is still null and on the first occurrence of
     // a block greater than the current hfBlock set the accumulator,
     // pass on the accumulator as the final result from this time on
-    const nextHfBlock = this.hardforks().reduce((acc: BN | null, hf: HardforkParams) => {
-      const block = new BN(hf.block!)
-      return block.gt(hfBlock) && acc === null ? block : acc
+    const nextHfBlock = this.hardforks().reduce((acc: bigint | null, hf: HardforkParams) => {
+      const block = BigInt(hf.block === null ? 0 : hf.block)
+      return block > hfBlock && acc === null ? block : acc
```
holgerd77, line 785: I have a bit my problems to read this code TBH since I am not so familiar with the `reduce()` functionality. Just to mention here: so the return value for a HF where block is `null` (this means: HF in the future, e.g. Merge/Shanghai) should not be BigInt(0) but `null`. Is this the case here (should be covered by tests though)? 🤔

packages/common/tests/customChains.spec.ts
```diff
@@ -123,19 +122,19 @@ tape('[Common]: Custom chains', function (t: tape.Test) {
       customChains: [testnet],
     })
     st.equal(c.chainName(), 'mainnet', 'customChains, chain set to supported chain')
-    st.ok(c.hardforkBlock()!.eqn(4370000), 'customChains, chain set to supported chain')
+    st.ok(c.hardforkBlock()! === BigInt(4370000), 'customChains, chain set to supported chain')
 
     c.setChain('testnet')
     st.equal(c.chainName(), 'testnet', 'customChains, chain switched to custom chain')
-    st.ok(c.hardforkBlock()!.eqn(4), 'customChains, chain switched to custom chain')
+    st.ok(c.hardforkBlock()! === BigInt(4), 'customChains, chain switched to custom chain')
```
holgerd77, line 129: Just to mention here, also for @acolytec3: we *can* very well use the *n notation in the tests, the limitation to the BigInt() constructor is just for the production code (correct me if I am missing something).

holgerd77: (of course also no need to do or even change here - again - just as some note)

packages/tx/src/baseTransaction.ts
```diff
@@ -40,15 +41,15 @@ interface TransactionCache {
...
+  public readonly gasLimit: bigint
   public readonly to?: Address
-  public readonly value: BN
+  public readonly value: bigint
   public readonly data: Buffer
 
-  public readonly v?: BN
-  public readonly r?: BN
-  public readonly s?: BN
+  public readonly v?: bigint
```
holgerd77, line 50: A bit of a side topic, but since we change the function signature on this here wanted to mention: thinking for some time now if we want to switch the type of the `v` value to `number` to better differentiate towards `r` and `s` values? No strong opinion here though (but some slight tendency towards doing).

packages/tx/src/baseTransaction.ts
```diff
@@ -84,22 +85,22 @@ export abstract class BaseTransaction<TransactionObject> {
...
+    this._type = Number(bufferToBigInt(toBuffer(type)))
 
     const toB = toBuffer(to === '' ? '0x' : to)
     const vB = toBuffer(v === '' ? '0x' : v)
     const rB = toBuffer(r === '' ? '0x' : r)
     const sB = toBuffer(s === '' ? '0x' : s)
 
-    this.nonce = new BN(toBuffer(nonce === '' ? '0x' : nonce))
-    this.gasLimit = new BN(toBuffer(gasLimit === '' ? '0x' : gasLimit))
+    this.nonce = bufferToBigInt(toBuffer(nonce === '' ? '0x' : nonce))
```
holgerd77, line 95: Also just to mention here: as @ryanio noted in some chat as well, on these kind of initializations the `0` - or sometimes empty - input values (also: in it's different type flavors) are really critical and we already stumbled upon that (`v` being initially empty on initialization ended up with `BN(0)` and therefore let `isSigned()` show `true` 😬).

So there should be an additional look if tests like `inputValue.spec.ts` really cover these cases and are adopted adequately. In doubt it also can't hurt to write 2-3 additional test cases.

packages/tx/src/baseTransaction.ts
```diff
@@ -169,21 +170,43 @@ export abstract class BaseTransaction<TransactionObject> {
...
+   */
+  protected _validateHighS() {
+    const { s } = this
+    if (this.common.gteHardfork('homestead') && s !== undefined && s > SECP256K1_ORDER_DIV_2) {
+      const msg = this._errorMsg(
+        'Invalid Signature: s-values greater than secp256k1n/2 are considered invalid'
+      )
+      throw new Error(msg)
+    }
+  }
```
holgerd77, line 193: To clarify: are these checks which initiallly have been missing or is this some kind of refactor?

packages/tx/src/baseTransaction.ts
```diff
@@ -237,18 +260,10 @@ export abstract class BaseTransaction<TransactionObject> {
...
-      }
+    if (v === undefined || r === undefined || s === undefined) {
+      return false
     } else {
-      if (v === undefined || !r || !s) {
-        return false
-      } else {
-        return true
-      }
+      return true
```
holgerd77, line 266: Is this really totally safe to say that this code can be simplified like this with v, r, s now being represented as bigint? This code part is really critical.

(again, would be good to have a look on occasion if tests are covering all cases and in doubt another test case would also not hurt)

packages/tx/src/baseTransaction.ts
```diff
@@ -383,37 +398,37 @@ export abstract class BaseTransaction<TransactionObject> {
...
+    values: { [key: string]: bigint | undefined },
     bits = 256,
     cannotEqual = false
   ) {
     for (const [key, value] of Object.entries(values)) {
       switch (bits) {
         case 64:
           if (cannotEqual) {
-            if (value?.gte(MAX_UINT64)) {
+            if (value !== undefined && value >= MAX_UINT64) {
```
holgerd77, line 409: This is just rewriting, semantics stay the same, am I am correct with this?

packages/tx/src/baseTransaction.ts
holgerd77: Update: Ah, just seeing that this is taken out of the inherited classes.

So `_validateHighS()` is ok to move out I guess but `_validateYParity()` should not be taken out since this doesn't go for the legacy tx and in base tx we only want to have code which is really common for all three existing tx types.

Side note: this whole inheritance structure for txs with tx types is not optimal. We have tried to come up with some - more modular - replacement but weren't satisfied yet on the results. If you want you can give this some initial thought. That would be a somewhat big refactor though, not sure if the scope on this would be too large to get into the coming breaking release round.

packages/util/src/types.ts
```diff
@@ -81,14 +83,15 @@ export function bnToRlp(value: BN): Buffer {
...
+  BigInt,
   Buffer,
   PrefixedHexString,
 }
 
 export type TypeOutputReturnType = {
   [TypeOutput.Number]: number
-  [TypeOutput.BN]: BN
+  [TypeOutput.BigInt]: bigint
+  // [TypeOutput.BN]: BN
```
holgerd77, line 94: Please don't remove this type conversion yet.

holgerd77 (commented):
This looks really, really great, also some nice targeted refactoring work! 😄 

Some comments for a first round here.

packages/tx/src/baseTransaction.ts
paulmillr: What's your suggesting regarding yParity validation? It's not great to have it duplicated in 3 files. The error names are the same, as is the logic. Having one different behavior somewhere is error-prone, we can forget to update all 3 files. 

As for class inheritance, check out micro-eth-signer, which implements all txs in 200 locs:

https://github.com/paulmillr/micro-eth-signer/blob/727d822030cbb5ba7e542a3a4ec6d40504069f82/index.ts#L314

holgerd77: Let's please not try to solve this on the sideline here, in doubt open a small issue on this.

holgerd77: > As for class inheritance, check out micro-eth-signer, which implements all txs in 200 locs:
> 
> https://github.com/paulmillr/micro-eth-signer/blob/727d822030cbb5ba7e542a3a4ec6d40504069f82/index.ts#L314

We can talk about the relationship of libraries like `micro-eth-signer` and `@ethereumjs/tx`, but this will likely take some time. 🙂

holgerd77:
Just force-pushed a rebased `develop`.

Sorry, I wonder how much hazzle this will cause, might be best on this round to create a new branch from the new `develop` head and then cherry-pick commits on top one-by-one. We might want to switch to only do these base branch rebases on times where all `develop` work has been merged, this should make things easier.

packages/tx/src/baseTransaction.ts
ryanio: yes this should be better code by checking undefined.

i remember we had a bug that led to this code in the first place so i had to look it up 😄 https://github.com/ethereumjs/ethereumjs-monorepo/commit/3f8f382e1c1cff74ca61077b7ad9e56fa148272a where EIP2930 supports v's with value 0 (empty Buffer)

it looks like we have a good amount of `isSigned()` tests in `test/legacy.spec.ts:462`, but only for legacy tx. this would be a good candidate to move to `base.spec.ts` and iterate through txTypes array. i'll try to do soon.

holgerd77:
Will close this in favor of #1744

PR #1732, opened 2022-02-22
README: Use GitHub built-in mermaid parser.

paulmillr:


holgerd77:
Oh, that's super cool, just read about this Mermaid integration a couple of days ago, thanks for the PR! 👍 

Currently the diagram in preview looks like this (for me):

grafik

Do you think this is only a preview thing or is something not working yet? 🙂

Also a basic note: I guess this naming is generally misleading in our current work setup: so we *do* do all patch/minor release work (so: the everyday work) on the libraries towards `master`. `develop` is only our development branch for the next round of breaking releases, we rebase this regularly (optimally weekly) with the changes from `master`.

(maybe it would be better to rename `develop`, eventually just the release series version (we take that from the VM), so `v6-develop` or something, need to think about that)

Anyhow: so this kind of work should be done towards `master`.

Issue #1729, opened 2022-02-21
Monorepo: Limit ES2020 Support to BigInt


holgerd77:
For our v6 releases #1717 we are planning a wider BigInt support for our libraries and are therefore lifiting up the TypeScript [compilation target](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/config/tsconfig.json) to [ES2020](https://en.wikipedia.org/wiki/ECMAScript#11th_Edition_%E2%80%93_ECMAScript_2020) which is a huge step especially for our browser target coming from [ES5](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/config/tsconfig.browser.json).

If I get this right users are *not* able to further compile this down to lower ECMAScript versions. So we need to self-restrict on minimally other ES2020 and likely also lower-version-features and not use in our libraries respectively remove, since we likely do not want to limit our user base for some side-syntactic-sugar features, which didn't make it into all browsers yet. Also to note here is that BigInt has been adopted by browsers relatively easy.

So I guess we need to go down feature-by-feature to at least ES2018/2019 and see what we would need to remove from our code base (I guess in doubt: everything else).

We also need to see how we enforce this not to be re-introduced, e.g. for the nullish coalescent operator (this is a bit of a pity, no drama though as well) there is an ESLint rule [es/no-nullish-coalescing-operators](https://eslint-plugin-es.mysticatea.dev/rules/no-nullish-coalescing-operators.html).

There might be alternative ways to go here, for TypeScript I haven't found a way for a more selective compilation/transpilation (might have overlooked something though), we might also want to think if we want to add a more complex build pipeline (not sure, would this be able to solve the problem? 🤔)

ryanio:
I guess I'm confused, if we minimally support bigint for our next major releases then es2020 support is required so what is the benefit of removing use of the other es2020 features?

holgerd77:
Browsers do not switch over and support "ES2020" by one switch-over but they start adding these features one-by-one. So BigInt was implemented relatively early on and is supported in Chrome 67 onwards (see [caniuse](https://caniuse.com/?search=bigint) page) (which would likely be a good target browser to settle on given that we have the ok for Metamask).

For other ES2020 features (and eventually also older ECMAScript features) the situation is very different, for e.g. nullish coalescent operator is only [supported in Chrome 80](https://caniuse.com/?search=bigint) (!) an upwards.

[...]

PR #1734, opened 2022-02-22
Target ES2018 + misc cleanup

acolytec3:
- Downgrades ES target to ES2018 with specific library targets for `es2019.object` and `es2020.bigint` to address https://github.com/ethereumjs/ethereumjs-monorepo/issues/1729. 
- Fixes a couple of missed native bigint notation numbers left over from https://github.com/ethereumjs/ethereumjs-monorepo/pull/1671
- Dedups `expmod`/`exponentiation` (thanks @paulmillr !!!)

### DIFF
--- a/config/eslint.js
+++ b/config/eslint.js
@@ -16,6 +16,7 @@ module.exports = {
   ],
   extends: ['typestrict', 'eslint:recommended'],
   rules: {
+    'no-undef': 'off',
     'no-console': 'warn',
     'no-debugger': 'error',
     'prefer-const': 'error',
--- a/config/tsconfig.browser.json
+++ b/config/tsconfig.browser.json
@@ -1,6 +1,6 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
-    "lib": ["dom", "ES2020"]
+    "lib": ["dom"]
   }
 }
--- a/config/tsconfig.json
+++ b/config/tsconfig.json
@@ -11,7 +11,7 @@
     "resolveJsonModule": true,
     "downlevelIteration": true,
     "strict": true,
-    "target": "es2020",
-    "lib": ["ES2020"]
+    "target": "es2019",
+    "lib": ["ES2020.BigInt"]
   }
 }
--- a/config/tsconfig.json
+++ b/config/tsconfig.json
@@ -11,7 +11,7 @@
     "resolveJsonModule": true,
     "downlevelIteration": true,
     "strict": true,
-    "target": "es2019",
-    "lib": ["ES2020.BigInt"]
+    "target": "es2018",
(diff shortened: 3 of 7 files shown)


paulmillr (dismissed):

paulmillr (dismissed):

holgerd77:
I guess we need to have a closer look if we do can support all from ES2019 as well or if we want to go lower (this should be possible, right?).

Just read [here](https://selvaganesh93.medium.com/javascript-whats-new-in-ecmascript-2019-es2019-es10-35210c6e7f4b) e.g. that many ES2019 features were introduced in Chrome 72. So if we e.g. want to take Chrome 67 (BigInt) as a base line, we would need to compile down further.

Guess this needs some more dedicated and differentiated look, likely feature by feature.

acolytec3:
From looking at the `lib` targets available for 2019, it's generically noted as `Array`, `Object`, `String`, and `Symbol` so we should look at it from that lens if we're focused on particular features of ES2019 to support.

paulmillr:
@holgerd77 

- `Object.entries` we are already using
- `flatMap` is Chrome 69, Safari 12
- `trimStart`, `trimEnd` is Chrome 69, Safari 12
- `Object.fromEntries` is **Chrome 73, Safari 12.1**
- Optional catch binding is Chrome 66, Safari 11
- JavaScript is a superset of JSON is Chrome 66, Safari 12
- built-in Symbol description is Chrome 70, Safari 12.1
- Strings are escaped to well-formed UTF-8 is Chrome 72, Safari 12.1
- Array: sort: Stable sorting is Chrome 70, Safari 10.1
- ES2019 toString: Implements Function.prototype.toString revision is not supported anywhere

holgerd77:
So what would be a good settlement here?

Do we want to keep the **Node.js build target on ES2020** (as @ryanio mentioned Node.js 12 is getting OOL soon (April) and 14 has full ES2020 support) and go down with the **browser build target to ES2018**?

(haven't given this too much thought yet, so very much open for suggestions. Might make sense e.g. to go a bit further down with the browser target. And would it make sense to go down with the Node target a bit as well to not unnecessarily throw Node 12 out of the ring?

holgerd77:
(I guess in doubt we can take the suggested above as a working state and eventually also still adjust later on)

paulmillr:
It depends on MM and others' targets. If we want no hassle, i'd get es2018 + es2019 object.entries + es2020 bigint.

acolytec3:
I updated the tsconfig targets to 2018 plus the specific library targets for es2019.object and 2020.bigint and everything seems to be passing. @holgerd77 @paulmillr any further discussion here? Seems like we should be good to go with this one otherwise.

paulmillr:
LGTM

holgerd77 (approved):
LGTM

PR #1740, opened 2022-02-23
ci: fix client integration failure

### LINKED ISSUE #1736: Monorepo, Client: Node Version CI Tests failing (Node 16)
Currently nightly Node version CI tests are failing, see e.g. [this run](https://github.com/ethereumjs/ethereumjs-monorepo/runs/5297284441?check_suite_focus=true).

There is some concurrency problem in the client, giving an error message as the following:

```shell
# [Integration:LightEthereumService]
# should handle LES requests
not ok 17 Error: Already running a server at 127.0.0.1
  ---
    operator: error
    stack: |-
      Error: Already running a server at 127.0.0.1
```

Might be due some side effects along the latest Merge work, not sure?

Anyhow: this should be investigated. 😄

ryanio:
Closes #1736 

This was a weird race condition that sometimes passed and sometimes didn't on my local without any code changes.

By moving the localServer up in the destroy sequence it should execute it more reliably before the next test starts. I also replaced the setTimeout in `await new Promise` with `this.wait`

### DIFF
--- a/.github/workflows/node-versions.yml
+++ b/.github/workflows/node-versions.yml
@@ -5,6 +5,8 @@ name: Node versions
 on:
   schedule:
     - cron: 0 0 * * *
+  pull_request:
+    types: [opened, reopened, synchronize]
 
 jobs:
   get-node-versions:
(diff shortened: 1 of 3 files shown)


ryanio:
looks like it's working:
Screen Shot 2022-02-23 at 1 37 33 PM

ryanio:
will merge this in since it's rather harmless ci fix

PR #1743, opened 2022-02-24
VM, Common: Add EIP-3670 (EOF - Code Validation)

acolytec3:
Implements EIP-3670 on top of EIP-3540 PR

https://eips.ethereum.org/EIPS/eip-3670

### DIFF
--- a/packages/common/src/eips/3670.json
+++ b/packages/common/src/eips/3670.json
@@ -6,8 +6,7 @@
     "status": "Review",
     "minimumHardfork": "london",
     "requiredEIPs": [
-        3540,
-        3541
+        3540
     ],
     "gasConfig": {},
     "gasPrices": {},
--- a/packages/common/src/eips/3670.json
+++ b/packages/common/src/eips/3670.json
@@ -1,0 +1,16 @@
+{
+    "name": "EIP-3670",
+    "number": 3670,
+    "comment": "EOF - Code Validation",
+    "url": "https://eips.ethereum.org/EIPS/eip-3670",
+    "status": "Review",
+    "minimumHardfork": "london",
+    "requiredEIPs": [
+        3540,
+        3541
+    ],
+    "gasConfig": {},
+    "gasPrices": {},
+    "vm": {},
+    "pow": {}
+}
--- a/packages/vm/src/evm/evm.ts
+++ b/packages/vm/src/evm/evm.ts
@@ -415,13 +415,12 @@ export default class EVM {
             ...result,
             ...INVALID_BYTECODE_RESULT(message.gasLimit),
           }
-        } else {
+        } else if (this._vm._common.isActivatedEIP(3670)) {
           // EIP-3670 EOF1 code check
-          const codeStart = eof1CodeAnalysisResults?.data > 0 ? 10 : 7
+          const codeStart = eof1CodeAnalysisResults.data > 0 ? 10 : 7
           if (
             !eof1ValidOpcodes(
-              result.returnValue.slice(codeStart, codeStart + eof1CodeAnalysisResults!.code),
-              eof1CodeAnalysisResults!.code
+              result.returnValue.slice(codeStart, codeStart + eof1CodeAnalysisResults.code)
             )
           ) {
             result = {
--- a/packages/vm/src/evm/evm.ts
+++ b/packages/vm/src/evm/evm.ts
@@ -418,6 +418,9 @@ export default class EVM {
         } else if (this._vm._common.isActivatedEIP(3670)) {
           // EIP-3670 EOF1 code check
           const codeStart = eof1CodeAnalysisResults.data > 0 ? 10 : 7
+          // The start of the code section of an EOF1 compliant contract will either be
+          // index 7 (if no data section is present) or index 10 (if a data section is present)
+          // in the bytecode of the contract
           if (
             !eof1ValidOpcodes(
               result.returnValue.slice(codeStart, codeStart + eof1CodeAnalysisResults.code)
--- a/packages/vm/src/evm/evm.ts
+++ b/packages/vm/src/evm/evm.ts
@@ -16,7 +16,7 @@ import TxContext from './txContext'
 import Message from './message'
 import EEI from './eei'
 // eslint-disable-next-line
-import { eof1CodeAnalysis, short } from './opcodes/util'
+import { eof1CodeAnalysis, eof1ValidOpcodes, short } from './opcodes/util'
 import { Log } from './types'
 import { default as Interpreter, InterpreterOpts, RunState } from './interpreter'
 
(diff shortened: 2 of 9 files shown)


packages/common/src/eips/3670.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "EIP-3670",
+    "number": 3670,
+    "comment": "EOF - Code Validation",
+    "url": "https://eips.ethereum.org/EIPS/eip-3670",
+    "status": "Review",
+    "minimumHardfork": "london",
+    "requiredEIPs": [
+        3540,
+        3541
```
holgerd77, line 10: Please do not add `3541` here, this should be only in `3540`, otherwise this will be redundant.

packages/vm/src/evm/evm.ts
```diff
@@ -407,12 +408,26 @@ export default class EVM {
...
+        // EIP-3540 EOF1 header check
+        const eof1CodeAnalysisResults = eof1CodeAnalysis(result.returnValue)
+        if (!eof1CodeAnalysisResults?.code) {
           result = {
             ...result,
             ...INVALID_BYTECODE_RESULT(message.gasLimit),
           }
+        } else if (this._vm._common.isActivatedEIP(3670)) {
+          // EIP-3670 EOF1 code check
+          const codeStart = eof1CodeAnalysisResults.data > 0 ? 10 : 7
```
holgerd77, line 420: Can you add an additional comment on the meaning of this 10, 7?

packages/vm/src/evm/opcodes/util.ts
```diff
@@ -334,9 +335,38 @@ export const eof1CodeAnalysis = (container: Buffer) => {
...
+    if (opcode >= 0x60 && opcode <= 0x7f) {
+      // Skip data block following push
+      x += opcode - 0x5f
+      if (x > code.length - 1) {
+        // Push blocks mmust not exceed end of code section
+        return false
+      }
+    }
+  }
+  const terminatingOpcodes = new Set([0x00, 0xd3, 0xfd, 0xfe, 0xff])
```
holgerd77, line 366: I guess a dedicated list why each of the opcodes here is regarded as "terminating" would be good as well.

packages/common/src/eips/3670.json
acolytec3: Understood.  What about 3860?  I just noticed that EIP3670 says its requires that one as well which would notionally make this PR dependent on #1619.

packages/vm/src/evm/opcodes/util.ts
acolytec3: I expanded the comment to clearly identify the terminating opcodes (STOP, RETURN, REVERT, INVALID, and SELFDESTRUCT) as identified in the EIP.  Not sure there's much more to the logic of why each of these specifically other than these are the only opcodes that logically terminate a sequence of instructions.  Is this what you meant?

packages/vm/src/evm/evm.ts
acolytec3: Done.

acolytec3:
Any thoughts on what will determine the definition of done for this PR and letting it safely be merged into the EIP-3540 PR?  I've got tests written that reflect the test scenarios identified in the EIP (excluding ones already testing EIP3540). The 
`ethereum/tests` for these EIPs are still unfinished and seems like they may require an upgrade to our test runner to work correctly anyway so not sure that my code passing the current tests should be the right decision point.

acolytec3:
> Any thoughts on what will determine the definition of done for this PR and letting it safely be merged into the EIP-3540 PR? I've got tests written that reflect the test scenarios identified in the EIP (excluding ones already testing EIP3540). The `ethereum/tests` for these EIPs are still unfinished and seems like they may require an upgrade to our test runner to work correctly anyway so not sure that my code passing the current tests should be the right decision point.

I was able to get some of the new tests to pass after finding a bug in the `PUSH` handler that has now been fixed.  Given that geth seems to be implementing the two EIPs (3540 and 3670) together as well as the tests for these EIPs being worked together for `ethereum/tests`, I'm inclined to merge this PR into the EIP-3540 one.  Any objections at this point?

packages/vm/src/evm/opcodes/util.ts
holgerd77: Ah, yes, that works respectively gives enough context I guess, thanks! 😄

packages/common/src/eips/3670.json
holgerd77: Ah, yes, then it makes sense to add here as well.

jochem-brouwer: I will finish 3860 this week, once that is merged then we should also add 3860 there (and rebase/merge `master` into #1719)

packages/vm/src/evm/evm.ts
```diff
@@ -407,12 +408,29 @@ export default class EVM {
...
+          // The start of the code section of an EOF1 compliant contract will either be
+          // index 7 (if no data section is present) or index 10 (if a data section is present)
+          // in the bytecode of the contract
+          if (
+            !eof1ValidOpcodes(
+              result.returnValue.slice(codeStart, codeStart + eof1CodeAnalysisResults.code)
+            )
+          ) {
+            result = {
+              ...result,
```
jochem-brouwer, line 430: This branch is not covered by tests

packages/vm/src/evm/opcodes/util.ts
```diff
@@ -334,9 +335,38 @@ export const eof1CodeAnalysis = (container: Buffer) => {
...
+    x++
+    if (!opcodes.has(opcode)) {
+      // No invalid/undefined opcodes
+      return false
+    }
+    if (opcode >= 0x60 && opcode <= 0x7f) {
+      // Skip data block following push
+      x += opcode - 0x5f
+      if (x > code.length - 1) {
+        // Push blocks mmust not exceed end of code section
```
jochem-brouwer, line 361: Nit: spelling mistake `mmust` -> `must`

jochem-brouwer (commented):
LGTM, but I think tests need some love here:

- Add a test for the uncovered codecov line (code section does not end with a terminating instruction)
- Add a test where the code section ends with `0xFE` (`INVALID`) (this is a sanity check)
- Add a check for each *terminating instruction* (`00`, `F3`, `FD`, `FF`) (also a sanity check)
- Tests for a "create transaction" (so a transaction from an EOA which `to` field is empty) which tries to deploy a contract using invalid code (so it does not stop with a trunacting opcode, truncated `PUSHn` instruction, or it has a non-assigned instrucition besides `INVALID`/`0xFE`) and verifies that this consumes all gas
- Same as above but now the code is contained in `CREATE`/`CREATE2` 

Also, the "test cases" section of the EIP has a few test cases not mentioned here and also not covered in tests yet. (All of them seem like sanity checks to me)

acolytec3:
Thanks for the review!  Will work on adding the additional tests in the other PR

Issue #1703, opened 2022-02-09
Client: Kiln🔥🧱 Tracking Issue


ryanio:
*Kiln🔥🧱 – the last Merge sprint before launching pre-production testnets. Kiln aims to move client implementations into the production readiness stage.*

*Please note: EthereumJS does not aim to contend as a production mainnet client, but rather participating to ensure compatibility for downstream users of our libraries.*

---

### Overview
This issue details the milestones for the Kiln🔥🧱 testnets starting in February 2022.

Kiln Spec: https://hackmd.io/@n0ble/kiln-spec

### Execution Layer

Changes to Execution layer spec introduced in Kiln:

* [x] Extend semantics of executePayload and forkchoiceUpdated methods [#165](https://github.com/ethereum/execution-apis/pull/165)
* [x] Replace terminal block error with INVALID_TERMINAL_BLOCK status [#170](https://github.com/ethereum/execution-apis/pull/170)
* [x] Refine message ordering. Take 2 [#148](https://github.com/ethereum/execution-apis/pull/148)
* [x] Authentication for engine api [#167](https://github.com/ethereum/execution-apis/pull/167), PR #1723 
  * (pending change to the Engine API and will be added in a subsequent version of the Kiln specification)
* [ ] Add support for [Optimistic Sync](https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md)
  * (for reference, geth's implementation: [beaconsync.go](https://github.com/MariusVanDerWijden/go-ethereum/blob/merge-kiln/eth/downloader/beaconsync.go)) 
#### Milestones

* [x] **M0:** Implementation started
* [x] **M1:** Spec Implemented
* [x] **M2:** 1-to-1 Interop – One EL and one CL team pair off for basic interop testing
* [ ] **M3:** Many-to-Many Interop – 2 or more EL and CL teams perform interop testing
* [ ] **M5:** All-to-All persistent testnet – All ready EL and CL teams create a long-lived devnet
* [ ] **M6:** Kiln🔥🧱 code released – All clients release Kiln🔥🧱 code for public consumption

PRs:
 - v1: #1712
 - v2: #1750

PR #1750, opened 2022-02-25
merge: kiln v2 updates

ryanio:
Continues #1703 with [kiln v2 change set](https://hackmd.io/@n0ble/kiln-spec#v2-change-set)


packages/block/src/header.ts
```diff
@@ -66,12 +66,12 @@ export class BlockHeader {
   }
 
   /**
-   * EIP-4399: Post-PoS merge, `mixHash` supplanted as `random`
+   * EIP-4399: After merge to PoS, `mixHash` supplanted as `prevRandao`
    */
-  get random() {
+  get prevRandao() {
```
holgerd77, line 71: What's this for a strange name? 🤔

packages/client/lib/rpc/modules/engine.ts
```diff
@@ -522,4 +533,23 @@ export class Engine {
       }
     }
   }
+
+  /**
+   * Sets and returns the transition configuration parameters.
+   *
+   * @param params An array of one parameter:
```
g11tech, line 540: @ryanio this api doesn't sets the configuration, it just matches and logs (on a mismatch) and nothing else: https://github.com/ethereum/execution-apis/blob/f6fdc57ccb0ae3dd82812993a84e9fb12ebff798/src/engine/specification.md#engine_exchangetransitionconfigurationv1 
keyword being *surface*, also responds with its own user set params so that CL can also log mismatch if any.

packages/block/src/header.ts
ryanio: Some explanation here: https://twitter.com/TimBeiko/status/1494768704143298560

packages/client/lib/rpc/modules/engine.ts
ryanio: ooh that makes sense, thanks :) updated in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1750/commits/851378cc0f2e980f542cf08172e4351ec7843ecf

g11tech (dismissed):
LGTM! left running for the night on devnet5, works like a charm :rocket:

ryanio:
thanks @g11tech!! just shaped up the readme a bit and will merge after ci completes.

PR #1753, opened 2022-02-27
Common, Block: remove and rename ambiguous active methods

### LINKED ISSUE #1674: Common: Rename ambiguous active* methods
There currently is some ambiguity around methods named with "active*" in the Common library, see [index.ts](https://github.com/ethereumjs/ethereumjs-monorepo/blob/develop/packages/common/src/index.ts) implementation (best take the version from the `develop` branch, since there has already been some additional bloat along "active HFs" being removed by #1649 so this is easier to digest).

So there is basically the meaning of "active" in meaning that a HF respectively its rules are activated (e.g. on a certain block number). I guess this is the intuitive and expected meaning respectively semantics. So the associated method names should be kept:

- `hardforkIsActiveOnBlock()`
- `activeOnBlock()`

The second used version for "active" in method names is rather meaning: "is generally included (but not necessarily: active 😋) on the chain.

These should be renamed for clarity, this ambiguity has already let to wrong usage patterns in the past.

Following suggestings:

- `hardforkIsActiveOnChain` -> `isIncludedHardfork`
- `activeHardforks` -> `includedHardforks`
- `activeHardfork` -> `lastestIncludedHardfork`

Another thing I noticed together with an alternative solution:

All these three methods are extremely little used. For our monorepo:

1. `hardforkIsActiveOnChain` -> One usage in [block header](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/header.ts#L1012)
2. `activeHardforks` -> No external usage, several Common internal usages
3. `activeHardfork` -> One usage in [block header](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/header.ts#L988)

Both usages are actually suboptimal API usages and can be replaced by other Common functions. For 1. all the four lines or so of various checks can be replaced with a single `hardforkIsActiveOnBlock()` call if I am not mistaken. For 3. `getHardforkByBlockNumber()` and - furthermore - this whole `_getHardfork()` function in block is generally misplaced and should be removed entirely (not sure if this would lead to some necessary tweaks), since normally the hardfork should directly be retrieved by Common, this is some sort of strange functionality "doubling".

So I wonder if we should rather - at least for 1. and 3. - remove these functions at all. 🤔

I think this whole semantics is somewhat confusing, regardless of the name.

For 2. we might consider to rather make an internal private function (since this is a helpful utility function inside of Common) or maybe we keep this single one (but then also do the renaming).

Thoughts?

emersonmacro:
closes #1674 

Summary of changes:

Removed methods:
- `hardforkIsActiveOnChain`
- `activeHardforks`
- `activeHardfork`

Modified methods:
- `paramByBlock`
  - Original signature: `paramByBlock(topic: string, name: string, blockNumber: BNLike)`
  - New signature: `paramByBlock(topic: string, name: string, blockNumber: BNLike, td?: BNLike)`
- `_getHardfork` (internal helper function)
  - Returns `null` instead of throwing if HF does not exist on chain

### DIFF
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -1002,7 +1002,7 @@ export class BlockHeader {
    * activation block (see: https://blog.slock.it/hard-fork-specification-24b889e70703)
    */
   private _validateDAOExtraData() {
-    if (!this._common.hardforkIsActiveOnChain(Hardfork.Dao)) {
+    if (!this._common.isIncludedHardfork(Hardfork.Dao)) {
       return
     }
     const DAOActivationBlock = this._common.hardforkBlock(Hardfork.Dao)
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -998,7 +998,7 @@ export class BlockHeader {
    * activation block (see: https://blog.slock.it/hard-fork-specification-24b889e70703)
    */
   private _validateDAOExtraData() {
-    if (!this._common.isIncludedHardfork(Hardfork.Dao)) {
+    if (!this._common.hardforkIsActiveOnBlock(Hardfork.Dao, this.number)) {
       return
     }
     const DAOActivationBlock = this._common.hardforkBlock(Hardfork.Dao)
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -444,7 +444,7 @@ export class BlockHeader {
       )
       throw new Error(msg)
     }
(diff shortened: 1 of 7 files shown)


holgerd77 (approved):
Thanks a lot, look great! 😄 👍

holgerd77:
Need to admin-merge again due to client tests still failing (should be temporary as noted before).

PR #1755, opened 2022-02-28
Common: Genesis state logic improved

cbrzn:
This is a very WIP PR that aims to improve the genesis state management in the common library. Open to any feedback just wanted to see what you think 

- State root and hash of genesis block can be passed as GenesisState in attribute customChains
- Now we show genesis block hash and state root in client process logs
- Genesis state is now storing storage, nonce, code and balance attributes to work with complex genesis files

Ideas on TODO:
- Handle geth genesis file format in common
- Add some examples
- Study if we can remove the static files in genesisState() method and try to retrieve it from configuration given, this way we can remove a lot of code from Common package


holgerd77:
This is tackling too many things at once (as written per DM), will close this. Please re-submit separately (first: tests for existing functionality, second: geth genesis inclusion).

Issue #1708, opened 2022-02-09
Common: add support for genesis to add code/storage/nonces to genesis accounts


jochem-brouwer:
Some chains, like MATIC/Polygon, initialize certain contracts on the Genesis block.  This is not supported currently in Common, we can only add (?) balance to accounts. To import these chains, genesis should support setting code/storage/nonces of accounts as well.

jochem-brouwer:
Related: #1696

holgerd77:
I will expand this issue a bit with adding "Support Geth Config (in Common)" to the title. This is very related since in the Client we already support this functionality along the [Geth genesis.json import/parse logic](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/lib/util/parse.ts#L109) and it would be good to not double the code base here but optimally unify in Common.

Furthermore it would generally be good if Common would be able to itself read in Geth config files since this would expand on the use cases, e.g. to have this kind of configuration files as a basis for VM runs.

I can vaguely remember that this was in discussion already when we were implementing, and then we had some blockers with prevented this from being implementable in an easy way. Also a vague remembrance: I think it had something to do that - atm we have the `stateRoot` and the `hash` in the Common chain file genesis definition - see e.g. [mainnet.json](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/chains/mainnet.json#L20), which are both not present in the Geth `genesis.json` files and which would need to be recomputed which was not easily possible. I wonder if this would be easier (to remove both of those values) along the upcoming breaking releases? 🤔

Not sure, can't recall the whole chain of complexity here any more, maybe @acolytec3 Andrew can shine some light here since I think he did the implementation on this and stumbled upon these kind of things.

holgerd77:
> I'd have to do some exploratory coding but it should be feasible to compute the stateroot and hash if not provided in a genesis file. That's what we do in the client to parse the geth genesis file. The stateroot is calculated [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/125e210cdcc412f901420ef038f230a6d060cb51/packages/client/lib/util/parse.ts#L109). As I recall, the issue with putting this in common was the concern over adding the trie dependency to common.

Ah yes, thanks, remember a bit more clearly. I guess we need to have a closer look where `stateRoot` and `hash` are actually used directly and if/how this can be easily replaced.

[...]

PR #1757, opened 2022-03-01
VM, Common: Complex Genesis State tests

cbrzn:
Changes on this PR:

- Genesis state with contract account stored are now being tested
- Type of `AccountState` has been created, making it part of `GenesisState`
- Documentation improved on `customChain` attribute in `Common`

Note that the `GenesisState` was previously being used only as string (not the complex object which is now `AccountState`) - Hence, not introducing breaking changes, since the complex object was not being used at all

Now, what I aim to do next, is point 3 from this comment https://github.com/ethereumjs/ethereumjs-monorepo/issues/1708#issuecomment-1039049282 and execute `runTx` first with a simple transfer tx and then a contract interaction. 

If my understanding is correct, I think that I will need to first get the hash and state root through the client process (since in the `Common` package it still does not take into account the genesis state given to calculate these fields) and create a custom test data JSON, this way we instantiate a new `Common` object with this new JSON (+ custom state object) - Then, the execution should work, let's see how it goes :-)


packages/common/src/types.ts
```diff
@@ -40,8 +40,11 @@ export interface Chain {
   }
 }
 
+export type StoragePair = [key: string, value: string]
```
ryanio, line 43: maybe we use `PrefixedHexString` for increased clarity? can also use in AccountState

holgerd77: +1 for `PrefixedHexString` (this is in Util `types.ts`)

It is generally good if we try to add new exports as cautiously as possible, since every new export becomes part of the API and this becomes difficult to change later on (since many changes then become "breaking" changes, e.g. if we would want to rename `StoragePair` and someone uses it this is then breaking people's code.

Just as some context, concretely here: do we need to expose `StoragePair` here? I guess this is rather taken implicitly in in `AccountState` and external libraries/users won't need this for typing? 🤔

packages/common/src/types.ts
```diff
@@ -40,8 +40,11 @@ export interface Chain {
...
+export type AccountState = [balance: string, code: string, storage: Array<StoragePair>] // [balance, code, [[storageKey, storageValue]]]
```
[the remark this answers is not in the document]
holgerd77: Ah, yes, this is really beautifully to read. 🙂 👍

packages/common/src/index.ts
```diff
@@ -141,13 +141,25 @@ export interface CommonOpts extends BaseOpts {
...
+   * A complex genesis state with Contract and EoA states would have the following format:
+   *
+   * ```javascript
+   * const complexState = {
+   *   '0x0...01': '0x100', // For EoA
+   *   '0x0...02': ['0x1', '0xBYTECODE', [[ keyOne, valueOne ], [ keyTwo, valueTwo ]]] // For contracts
+   * }
+   * ```
+   * import myCustomChain1 from '[PATH_TO_MY_CHAINS]/myCustomChain1.json'
+   * const common = new Common({ chain: 'myCustomChain1', customChains: [ [ myCustomChain1, complexState ] ]})
```
holgerd77, line 162: 👍

holgerd77 (commented):
Thanks, looks great already and super clean test setup. 🙂

Description of the next TODO steps also makes sense.

holgerd77:
> If my understanding is correct, I think that I will need to first get the hash and state root through the client process

Update: I am actually not sure if this is really needed for this. Just had a look again, the state root is only used for [genesis block creation](https://github.com/ethereumjs/ethereumjs-monorepo/blob/c842a973bc8058df71cda140fe0720292e1aafe1/packages/block/src/header.ts#L299) in the `Block` library. So you might want to try to run without in a first iteration.

Update 2: Uh uh uh. I am just realizing that we really just don't have this functionality or mechanism that the genesis state is initialized in the VM and a tx or so can then be executed with this state context. Ah. Interesting. 🙂

I guess this is something which can be added in a very similar way like the [activatePrecompiles](https://github.com/ethereumjs/ethereumjs-monorepo/blob/c842a973bc8058df71cda140fe0720292e1aafe1/packages/vm/src/index.ts#L285) option we have in the VM already, so `activateGenesisState` I guess, which would then (also - similar to `activatePrecompiles`, see code, when no state manager is given) initialize the state acounts + code.

Would value a confirmation from @jochem-brouwer on this, guess he might have some clear opinion on this.

I think such an option would generally be handy for various scenarios (e.g. also for our internal testing).

holgerd77:
(but Cesar @cbrzn I guess you can continue with this also without an additional confirmation from Jochem or someone else, would be nice though)

holgerd77:
Update 3 😋: ah, and you can then internally call the `StateManager.generateGenesis()` function for this - no need to reimplement this state addition logic - this should - theoretically - already be prepared for the initialization of the extended genesis state with code + storage.

jochem-brouwer:
I'm not entirely sure what the question is here. Common cannot calculate state root. So it should be generated in VM? Isn't this functionality already around by importing the genesis block and then just reading the state root and verifying this is correct? Or is the problem that you only have a genesis *state* file and no genesis *block*  file (so you cannot read state root?).

cbrzn:
> I'm not entirely sure what the question is here. Common cannot calculate state root. So it should be generated in VM? Isn't this functionality already around by importing the genesis block and then just reading the state root and verifying this is correct? Or is the problem that you only have a genesis state file and no genesis block file (so you cannot read state root?).

@jochem-brouwer 

The problem we were having is that, when trying to run a transaction in the VM, we were not being able to somehow inject the `genesisState` attribute from `common` in the state context inside of the VM. In order to achieve this, we have added a flag called [`activateGenesisState`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/common/genesis-state-tests/packages/vm/src/index.ts#L95-L102) in the `VMOpts` interface, which if true, will call the method  [`generateCanonicalGenesis`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/common/genesis-state-tests/packages/vm/src/index.ts#L295), hence, adding the genesis state in `common` in the `stateManager` of the VM.

What I am currently trying in this PR, more specifically in this [file](https://github.com/ethereumjs/ethereumjs-monorepo/blob/common/genesis-state-tests/packages/vm/tests/api/customChain.spec.ts) is to create a `common` instance with a given `genesisState`, create a new instance `VM` and try to execute transactions based on this data added in the genesis state.

I have a question tho, which I will dig through it later, but for now, if someone can give me insight would be amazing :-)

With this [test](https://github.com/ethereumjs/ethereumjs-monorepo/blob/common/genesis-state-tests/packages/vm/tests/api/customChain.spec.ts#L85-L103), I am trying to execute a transaction to a contract previously added in the genesis state... In this contract [(in solidity)](https://github.com/ethereumjs/ethereumjs-monorepo/blob/common/genesis-state-tests/packages/vm/tests/api/customChain.spec.ts#L23-L31), a variable is defined, which to my understanding, is stored in the first slot of the `storage` attribute (which is `0x00..00`). I am trying to execute the method that is defined in the contract, that (again, if I understand this correctly),  returns the value of this stored variable. But in the object [`execResult`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/common/genesis-state-tests/packages/vm/tests/api/customChain.spec.ts#L98), I am getting the `bytecode` of the contract as the `returnValue` , which I find weird.

It is worth mentioning that I am trying to trigger this [`retrieve method`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/common/genesis-state-tests/packages/vm/tests/api/customChain.spec.ts#L27-L29) by encoding it into hex and passing it as the [`data` argument](https://github.com/ethereumjs/ethereumjs-monorepo/blob/common/genesis-state-tests/packages/vm/tests/api/customChain.spec.ts#L87-L93). I am almost sure that I am doing something wrong because of my lack of holistic knowledge but I feel I am very close to achieve this :-)

jochem-brouwer:
So we add an extra flag to signal the VM that there is a custom genesis state? Why can't we check if the provided common has a custom genesis state and then do this by default? People who would initialize common with a custom state would expect that VM also uses that state, and would take some time to figure out there is a flag.

Regarding your other questions: here is the gotcha.

You are right that the this `4` number is indeed stored at slot 0. You are right that the actual runtime code is returned. Why is this? This is because you have initiated the contract with the `initcode`. The `initcode` is the data you would send to your "create contract"  transaction. What this does is (1) setup the storage fields to the desired values (e.g. set slot 0 to 4) and (2) return the runtime code. Since you put the initcode at the contract address, it thus returns the runtime bytecode, so your result is expected. Note that the code which you returned is not the same as the code which is at the contract (compare both).

What you should do is initiate the contract with code `0x6080604052600460005534801561001557600080fd5b5060b6806100246000396000f3fe6080604052348015600f57600080fd5b506004361060285760003560e01c80632e64cec114602d575b600080fd5b60336047565b604051603e9190605d565b60405180910390f35b60008054905090565b6057816076565b82525050565b6000602082019050607060008301846050565b92915050565b600081905091905056fea2646970667358221220338001095242a334ada78025237955fa36b6f2f895ea7f297b69af72f8bc7fd164736f6c63430008070033` (i.e. the returned code which you got earlier). Now what happens if you call `retrieve()`? Would you get 4? 

No, you would not, as the contract has all storage slots set to 0 in the genesis. So, `retrieve` would return 0, not 4. To solve this, also put in genesis that at the contract at slot 0 you want value 4. 

(Side note: it might be possible that if you the solidity optimizer it might realize that the value in the slot is impossible to change (it is a constant). In that case, solidity might hardcode the method to return 4. However, I checked, this is not the case, it actually reads from storage).

I also checked that your calldata is correct. Let me know if something is unclear or you want some more info :smile:

holgerd77:
> So we add an extra flag to signal the VM that there is a custom genesis state? Why can't we check if the provided common has a custom genesis state and then do this by default? People who would initialize common with a custom state would expect that VM also uses that state, and would take some time to figure out there is a flag.

Ah, that's an interesting idea, haven't thought about that from that perspective. Some aspects to that:

1. In case an external `StateManager` is provided, you definitely *don't* want to have that (so in the same way this is excluded for `activatePrecompiles` if a `StateManager` [is provided](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/index.ts#L285)).

2. If people are initializing a VM, and then building up some state around this - maybe by directly accessing/using the internal `StateManager` from the VM - they will some different total set of state entries (and therefore also a different state root) if we just activate this by default. In our tests we are actually using the internal VM `StateManager` for this all the time, e.g. [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/tests/api/runBlock.spec.ts#L32) (just the first one I picked, not comparing the state root though).

So I would definitely stick with a flag here. The question is if we might want to activate by default - and rather let people opt out. I would have a tendency to not do so though, I find this a bit risky not really knowing what people do around scenario 2., this comes to close to a breaking change for my taste.

holgerd77:
(lot's of post-submission-edits in the last post, please read "on site" and not via email 😄)

holgerd77:
Otherwise: love this conversation here! 😀 ❤️

holgerd77:
@jochem-brouwer update: also, some room for a compromise: if you think this would really be highly convenient for users and would therefore push for having this activated, we can also add the flag being *deactivated* for now and switch over to activate along the breaking releases in `develop` and doing a note on the VM TODO.

(but also just a first thought, not completely sure if we should activate at all)

packages/vm/tests/api/customChain.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    // Returned value should be 4, because we are trying to trigger the method `retrieve`
+    // in the contract, which returns the variable stored in slot 0x00..01
+    t.equal(
+      callResult.execResult.returnValue.toString('hex'),
+      genesisState[contractAddress][2][0][1].slice(2)
+    )
+
+    t.end()
+  })
```
holgerd77, line 103: Ugh. Cesar, I am so impressed honestly. 🙏 🙂 🤓

cbrzn:
Thanks for the help @jochem-brouwer now the test is working as expected 😄 

This is now ready for review @holgerd77

packages/common/src/index.ts
ryanio, line 162: I'm seeing this a few times around the docs, looks like it was preexisting, can you search for `'myCustomChain1'` and replace with `myCustomChain1` (no single quotes)

```suggestion
   * const common = new Common({ chain: myCustomChain1, customChains: [ [ myCustomChain1, complexState ] ]})
```

packages/vm/src/index.ts
```diff
@@ -92,6 +92,14 @@ export interface VMOpts {
    * Default: `false`
    */
   activatePrecompiles?: boolean
+  /**
+   * If true, the state of the VM will add the genesis state given by {@link Common} to a new
+   * created state manager instance. Note that if stateManager option is also passed as argument
+   * this flag wont have any effect.
+   *
+   * Default: `undefined`
```
ryanio, line 100:
```suggestion
   * this flag won't have any effect.
   *
   * Default: `false`
```

packages/common/src/index.ts
```diff
@@ -141,13 +141,25 @@ export interface CommonOpts extends BaseOpts {
    * const common = new Common({ chain: 'myCustomChain1', customChains: [ myCustomChain1 ]})
    * ```
    *
-   * Pattern 2 (with genesis state, see {@link GenesisState} for format):
+   * Pattern 2 (with genesis state see {@link GenesisState} for format). Note that in {@link AccountState} there are two
+   * accepted types. This allows to easily insert accounts in the genesis state
```
ryanio, line 145: I think instead of this additional note here, you can remove and add some more context below with a `Pattern 3 (with complex genesis state, containing contract accounts and storage):` for the next example below

```suggestion
   * Pattern 2 (with genesis state see {@link GenesisState} for format):
```

ryanio:
just a few review/nit comments, but otherwise great PR!! 👏

packages/common/src/index.ts
holgerd77: Ah, no, that's not correct. This is not for passing in the chain itself but the chain *selection* via string (normally our enums), so this should indeed be `'myCustomChain1'` which is the string name of the `myCustomChain` configuration. 🙂

holgerd77 (changes requested):
Can second that, this looks really great! 😀

There is an `ethereum-tests` folder update which sneaked in here, this needs to be removed. Otherwise I would say this is ready to go! 🎉

packages/common/src/index.ts
ryanio: oo ok

packages/vm/README.md
```diff
@@ -7,7 +7,7 @@
 [![Discord][discord-badge]][discord-link]
 
 | TypeScript implementation of the Ethereum VM. |
-| --- |
+| --------------------------------------------- |
```
cbrzn, line 10: Prettier linter did this, should I remove it?

packages/vm/README.md
```diff
@@ -218,17 +236,17 @@ If you want to understand your VM runs we have added a hierarchically structured
...
-| `vm:tx` | Transaction operations (account updates, checkpointing,...) |
-| `vm:tx:gas` | Transaction gas logger |
-| `vm:evm` | EVM control flow, CALL or CREATE message execution |
-| `vm:evm:gas` | EVM gas logger |
-| `vm:eei:gas` | EEI gas logger |
-| `vm:state`| StateManager logger |
-| `vm:ops` | Opcode traces |
-| `vm:ops:[Lower-case opcode name]` | Traces on a specific opcode |
+| Logger                            | Description                                                        |
+| --------------------------------- | ------------------------------------------------------------------ |
```
cbrzn, line 240: Same as the comment above, should I remove these non-related changes?

packages/common/src/index.ts
cbrzn: Just updated it! Please let me know if you have any other feedback and I can implement it - Once everything is cool you can resolve this @ryanio :-D

packages/vm/README.md
ryanio: It's fine you can leave it, as it will be fixed moving forward 👍

packages/vm/README.md
ryanio: This one is kind of annoying, I like the prettier format in the code sections especially, but could do without these table formatting. I wonder if we could turn off table formatting specifically in our prettier root config file.

holgerd77 (approved):
LGTM, thanks a lot Cesar! 🙂

Issue #1745, opened 2022-02-25
Util: Deprecation Tasks


holgerd77:
Part of #1717 

Following deprecations on Util are uncritical and it can be executed upon:

- Remove the whole `object.ts` module (so the complete file)
- Remove deprecated `bnToRlp()` method

Other deprecations are around signature functionality or BN/BigInt and should rather be done along the respective dedicated task executions.

acolytec3:
Addressed by #1809

PR #1767, opened 2022-03-07
util: deprecation tasks (object.ts, bnToRlp)

ScottyPoi:
#1745 

Executes on 2 Util package deprecations.

1.  Deletes object.ts
2.  Deletes object.spec.ts
3. Removes deprecated bnToRlp() method from util/src/types.ts


ryanio:
> I don't think the client test fail has anything to do with these changes.

it doesn't, i have a pending PR to fix on develop

holgerd77:
Ok, thanks, let's wait with merging this as well until #1771 is merged into `develop`.

holgerd77:
Continued in #1809 will close.

PR #1768, opened 2022-03-07
VM, Common: Implement EIP 1153, Transient storage opcodes

tynes:
This is a draft PR that implements [EIP 1153](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1153.md). I'm opening it to get feedback on its structure, to make sure its going in the correct direction and to see if its something that is desired to be upstreamed.

Its only slightly tested and there is a problem where the opcodes for simple subroutines clash with the opcodes used in our forks of [geth and solc](https://github.com/tynes/eip1153). This must be fixed before merging is considered

Ideally if this was merged and [1] is implemented in hardhat, then users will be able to very easily develop EIP 1153 based smart contracts in hardhat

[1] - https://github.com/NomicFoundation/hardhat/issues/2463

If it is something desired, I will clean up the commits and add more tests


ryanio:
pretty cool, thanks for taking the initiative! looks like a pretty clean implementation so far on first look :)

packages/common/src/eips/1153.json
```diff
@@ -0,0 +1,22 @@
...
+  "minimumHardfork": "chainstart",
+  "requiredEIPs": [],
+  "gasConfig": {},
+  "gasPrices": {
+    "tstore": {
+      "v": 200,
+      "d": "Base fee of the TSTORE opcode"
+    },
+    "tload": {
+      "v": 200,
```
jochem-brouwer, line 16: Same here, should be 100 gas per EIP spec? (unless I'm misreading)

packages/vm/src/evm/evm.ts
```diff
@@ -153,6 +153,7 @@ export default class EVM {
 
     const oldRefund = this._refund.clone()
 
+    // TODO: checkpoint needs to take into account transient storage
```
jochem-brouwer, line 156: Yep, good point, this should probably be added in `StateManager`, we cannot directly use the storage logic since this is saved into a `Map` and not to disk / temporary to memory.

packages/vm/src/evm/opcodes/codes.ts
```diff
@@ -249,6 +249,16 @@ const hardforkOpcodes: { hardforkName: string; opcodes: OpcodeEntry }[] = [
 ]
 
 const eipOpcodes: { eip: number; opcodes: OpcodeEntry }[] = [
+  {
+    // These numbers differ than what is currently implemented
+    // in geth due to other eips being included here that are
+    // not in geth
```
jochem-brouwer, line 255: Right, the EIP also does not yet state whatever opcodes numbers we should use.

packages/vm/src/state/transientStorage.ts
```diff
@@ -0,0 +1,85 @@
...
+      if (map) {
+        for (const [key, value] of storage.entries()) {
+          map.set(key, value)
+        }
+      }
+    }
+  }
+
+  checkpoint(): void {
+    const copy = copyStorage(this._storage)
```
jochem-brouwer, line 71: This is OK for now, but for production we probably want to update this. The logic would be to keep a list of `Map` items where each time if we `checkpoint` we add an empty map to the list. When you want to get a value at `index` (so `TLOAD`) then traverse the list from the back: for each map attempt to get the key. If no key is returned then the value is thus `0`. On `TSTORE` put the key in the last item of the map. On reverting, pop the last list item. On `commit`, copy all items from the last map to the previous map and pop that map from the list.

If we copy the map each time then I think it can blow up, first fill the map with X items and then call own contract until all gas is consumed. Then if we call to depth `L` we have `X*L` 32-byte items in total in memory. This might be a DoS vector which makes the implementation very slow.

packages/vm/src/evm/evm.ts
jochem-brouwer: I think the current logic in `transientStorage` would make this work, right?

jochem-brouwer (commented):
Great start so far! Some comments. The state manager tests are great, but we should also add tests for the VM (so running the `TLOAD` / `TSTORE` opcodes).

tynes:
> Great start so far! Some comments. The state manager tests are great, but we should also add tests for the VM (so running the `TLOAD` / `TSTORE` opcodes).

Thank you for the quick review! I'll address the comments when I have some time over the next couple of days

packages/common/src/eips/1153.json
tynes: You are correct, I need to update the values. There still isn't complete consensus on what they should be. They cannot be so cheap that a tx that uses a full block's worth of gas to store as many key/value pairs as possible can oom a node on consumer hardware

packages/common/src/eips/1153.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "EIP-1153",
```
tynes, line 2: This file needs 4 spaces in the json, my linter likes 2

packages/common/src/hardforks/chainstart.json
```diff
@@ -23,15 +23,7 @@
...
-      "v": [
-        0,
-        2,
-        3,
-        5,
-        8,
-        10,
-        20
-      ],
+      "v": [0, 2, 3, 5, 8, 10, 20],
```
tynes, line 26: Sorry for this unnecessary diff, will remove, my linter did it automatically

packages/vm/src/evm/evm.ts
tynes: You are correct, this is a dead comment now

tynes:
This pr also needs to take into account the debug logging that shows what values are being stored/read still

packages/vm/src/state/transientStorage.ts
jochem-brouwer: @ryanio do you think this approach would be a "correct" one to optimize this?

packages/common/src/hardforks/chainstart.json
jochem-brouwer: Are you using `npm run lint:fix` on the package?

packages/vm/src/state/transientStorage.ts
ryanio: that sounds good to me, agree that copyStorage on every checkpoint would have a detrimental impact on performance

packages/common/src/hardforks/chainstart.json
ryanio: should be fine, as long as it's following our monorepo's prettier rules and not your own local config

holgerd77:
Some context: [Tweet from Tim Beiko on ACD from January 2022](https://twitter.com/TimBeiko/status/1479486058291253248)

ryanio:
looks like coverage is down a lot, let's make sure we have all new lines covered with this PR as to not affect overall VM code coverage. probably just need a few more test cases to test the uncovered lines.

packages/vm/src/evm/opcodes/codes.ts
tynes: We now have some numbers that we can use: https://github.com/ethereum/EIPs/commit/b6b3857fbcec36048e527ebb5e0eb8b582f6c96c

tynes:
> looks like coverage is down a lot, let's make sure we have all new lines covered with this PR as to not affect overall VM code coverage. probably just need a few more test cases to test the uncovered lines.

Working on adding more coverage still, will squash my commits once I get some more coverage

packages/vm/src/state/transientStorage.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+    const storage = this._checkpoints[this._checkpoints.length - 1].get(addr)
+    storage!.set(key.toString('hex'), value)
+  }
+
+  revert() {
+    this._checkpoints.pop()
+  }
+
+  commit(): void {
```
tynes, line 49: `commit` needs to handle when there is only 1 item in `_checkpoints`

packages/vm/src/state/transientStorage.ts
```diff
@@ -0,0 +1,80 @@
+import { Address } from 'ethereumjs-util'
+
+export interface TransientStorageOps {
+  storage?: Map<Address, Map<string, Buffer>>
+  checkpoints?: any[]
```
tynes, line 5: Maybe `checkpoints` should't be exposed in the constructor

tynes:
> looks like coverage is down a lot, let's make sure we have all new lines covered with this PR as to not affect overall VM code coverage. probably just need a few more test cases to test the uncovered lines.

Based on https://app.codecov.io/gh/ethereumjs/ethereumjs-monorepo/compare/1768/overview it looks like codecov is still working off of older code. Is there any way to refresh codecov? I've just been running the coverage locally to be able to add more coverage

tynes:
code-cov

There are a couple branches inside the transient storage class that aren't covered, they are when [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) is used to make typescript remove the "object may be undefined" error in cases where it is known that the object will not be undefined. 100% of lines are covered

packages/vm/src/state/transientStorage.ts
tynes: I've updated the algorithm to be more similar to how geth does it, let me know what you think. It feels a bit more simple to me doing it this way

jochem-brouwer:
We have to manually approve the CI to run, just did that for you, codecov will be updated.

As for this optional chaining, in case you are 100% sure that it is not undefined then you can use the `!` operator, right? Would that also impact coverage?

packages/vm/src/state/interface.ts
```diff
@@ -21,6 +21,8 @@ export interface StateManager {
   getOriginalContractStorage(address: Address, key: Buffer): Promise<Buffer>
   putContractStorage(address: Address, key: Buffer, value: Buffer): Promise<void>
   clearContractStorage(address: Address): Promise<void>
+  getContractTransientStorage?(address: Address, key: Buffer): Buffer
+  putContractTransientStorage?(address: Address, key: Buffer, value: Buffer): void
   checkpoint(): Promise<void>
```
tynes, line 26: The `StateManager` interface cannot be changed without breaking hardhat, the fork state manager implements this interface. I made the transient storage functions optional to get around this, but that results in some `ts-ignores` throughout this codebase to prevent a "cannot call possibly undefined function" error

https://github.com/NomicFoundation/hardhat/blob/f682724fa16cddfb0a5977b841f274fb743c4657/packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts#L45

packages/vm/src/evm/eei.ts
```diff
@@ -380,6 +380,25 @@ export default class EEI {
     }
   }
 
+  /**
+   * Store 256-bit a value in memory to transient storage.
+   * @param key - Storage key
+   * @param value - Storage value
+   */
+  transientStorageStore(key: Buffer, value: Buffer): void {
+    // @ts-ignore
```
tynes, line 389: Instead of using `ts-ignore`, I could also do something like:

```js
if (typeof this._state.putContractTransientStorage === 'function') {
  // call it
}
```

packages/vm/src/state/interface.ts
tynes: Another option would be to not expose high level setters/getters for transient storage

packages/vm/src/evm/eei.ts
ryanio: sure, or another way we use frequently is the `in` operator: `if ('putContractTransientStorage' in this._state) {`

but here instead maybe throw an error if it is not available? it would add more safety for it being accidentally called

tynes: I've updated the code to follow this style

packages/vm/src/state/interface.ts
holgerd77: Yes, please definitely not expose something like this in the official interface, we wouldn't want to bloat here for others implementing the interface along doing such experimental integrations like the PR here.

In doubt just do any-casts when using, even if it get's a bit ugly.

tynes: Will remove from the interface

tynes: I pushed a fix to remove it from the interface and also added a bit more test coverage

tynes:
This is ready for review again

packages/vm/src/state/transientStorage.ts
```diff
@@ -0,0 +1,109 @@
+import { Address } from 'ethereumjs-util'
+
+export interface TransientStorageOps {
+  storage?: Map<Address, Map<string, Buffer>>
+  changesets?: Changeset[][]
+}
+
+export type TStorage = Map<Address, Map<string, Buffer>>
```
moodysalem, line 8: the map should not be keyed by address, but rather string, two addresses that have the same string representation are not equal in the hashmap

moodysalem:
In my testing it seems like the VM in this PR is not clearing transient storage after each transaction. I think the transient storage should be moved from the state manager to the `RunState`, since it's specific to each transaction

tynes:
> In my testing it seems like the VM in this PR is not clearing transient storage after each transaction. I think the transient storage should be moved from the state manager to the `RunState`, since it's specific to each transaction

Good catch, I believe that I have pushed a fix for this. Will add additional test coverage for this case

tynes:
We pushed a new approach and added some additional test coverage, would love to hear what you think of the new approach. Now there are no changes to the state manager and the changes are encapsulated in the eei

packages/vm/src/evm/opcodes/functions.ts
```diff
@@ -873,7 +873,43 @@ export const handlers: Map<number, OpHandler> = new Map([
       runState.eei.log(mem, topicsCount, topicsBuf)
     },
   ],
+  // 0xb3: TLOAD
+  [
+    0xb3,
+    function (runState, common) {
+      if (!common.isActivatedEIP(1153)) {
```
jochem-brouwer, line 880: This check should not be necessary, this opcode should be unreachable if  the EIP is not activated

packages/vm/src/evm/opcodes/functions.ts
```diff
@@ -873,7 +873,43 @@ export const handlers: Map<number, OpHandler> = new Map([
...
+      const value = runState.eei.transientStorageLoad(keyBuf)
+      const valueBN = value.length ? new BN(value) : new BN(0)
+      runState.stack.push(valueBN)
+    },
+  ],
+  // 0xb4: TSTORE
+  [
+    0xb4,
+    function (runState, common) {
+      if (!common.isActivatedEIP(1153)) {
```
jochem-brouwer, line 895: Same here with this check

packages/vm/src/state/transientStorage.ts
```diff
@@ -0,0 +1,131 @@
...
+      throw new Error('Transient storage value cannot be longer than 32 bytes')
+    }
+
+    if (!this._storage.has(addr.toString())) {
+      this._storage.set(addr.toString(), new Map())
+    }
+    const map = this._storage.get(addr.toString())!
+
+    const str = key.toString('hex')
+    const prevValue = map.get(str) ?? Buffer.alloc(32, 0x00)
```
jochem-brouwer, line 62: Nit: if you use `Buffer.alloc` with only 1 argument then it defaults to fill it with zeros.

packages/vm/src/state/transientStorage.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    map.set(str, value)
+  }
+
+  public revert() {
+    const changeset = this._changesets.pop()
+    if (!changeset) {
+      throw new Error('cannot revert without a changeset')
+    }
+    const reversed = changeset.slice().reverse()
```
jochem-brouwer, line 78: The `slice` copies the array, right? And the reversal seems like it can be optimized, for each key we would only need the first entry in the `changeset`? As that is the original value which we should revert to?

tynes: According to the docs, `slice` creates a shallow copy
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice
I can refactor this to use a simple for loop that decrements

You need to iterate through the changeset and apply each diff. Perhaps some better naming could help here.

As things are added to transient storage, a `TransientStorageModification` is pushed to the latest `changeset` array. On a `commit`, the latest `changeset` is discarded and on a `revert` the latest `changeset` is used to set each value that was changed to its previous value and then discarded
```js
export type Changeset = TransientStorageModification[]
```

You only need to touch the latest `changeset` and all items in that changeset.

packages/vm/src/state/transientStorage.ts
tynes: Will remove the second arg, was being explicit here

packages/vm/src/state/transientStorage.ts
moodysalem: the array could be optimized to be a map, since we only care about the first previous value

moodysalem: i addressed this in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1768/commits/aa79dcd44a562655398d6180576227b85e338b0f

tynes: Oh I see what you mean, thanks for the commit @moodysalem

packages/vm/src/index.ts
```diff
@@ -207,7 +208,7 @@ export default class VM extends AsyncEventEmitter {
    * performance reasons to avoid string literal evaluation
    * @hidden
    */
-  protected readonly DEBUG: boolean = false
+  readonly DEBUG: boolean = false
```
ryanio, line 211: same here, why remove `protected`?

packages/vm/src/index.ts
```diff
@@ -176,7 +177,7 @@ export default class VM extends AsyncEventEmitter {
 
   protected readonly _opts: VMOpts
   protected _isInitialized: boolean = false
-  protected readonly _allowUnlimitedContractSize: boolean
+  public readonly _allowUnlimitedContractSize: boolean
```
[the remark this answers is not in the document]
tynes: This is changed to `public` because it is accessed here:

https://github.com/ethereumjs/ethereumjs-monorepo/blob/7330c4ace495903748c606a47a8b993812504db8/packages/vm/src/evm/evm.ts#L428

`DEBUG` is accessed here:

https://github.com/ethereumjs/ethereumjs-monorepo/blob/7330c4ace495903748c606a47a8b993812504db8/packages/vm/src/evm/evm.ts#L409

This is because in the constructor to the `EVM` the first arg is updated from `any` to `VM` here.
https://github.com/ethereumjs/ethereumjs-monorepo/pull/1768/files#diff-44f5139e1722034ba835a86c3333b7452e621e5ecb324b0b12e9c7798f678acfR145

Happy to revert that change if you prefer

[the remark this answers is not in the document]
tynes: I feel like its unrelated changes for this PR but probably better to have the argument be type but happy to do whatever you think is best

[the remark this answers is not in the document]
ryanio: Oh I see, yes much better to type it as VM and get all the type goodness :) this is fine then, thanks!

holgerd77 (approved):
Ok, thanks for the CI update! I will now merge this in then. 🙂

PR #1769, opened 2022-03-07
client: increase json parser size

g11tech:
the kilnv2 payload sizes were throwing error on engine api, updated the size limit in the body-parse as per the bellatrix specs of 10MiB of gossip size.


acolytec3 (approved):
Just because I don't know, is there any concern we should think about from specifying a size limit of 10+mb? Seems like a non-issue to me since we handle block bodies bigger than that but just curious if you can think of any. Otherwise, LGTM!

ryanio (approved):
thanks!

Issue #1599, opened 2021-12-08
Create standalone StateManager package


jochem-brouwer:
In #1590 we implement stuff which is not directly related to the VM. It therefore makes sense to create a standalone package for StateManager - in some cases, we don't need VM support but only use StateManager as an interface to the DB.

jochem-brouwer:
@g11tech 

I think this should be done in the following steps (these steps are different PRs)

- Create the interface for StateManager. What should be in it and what not? Anything VM-related (warm addresses, touched accounts) should be removed from StateManager and should go into VM. (`evm.ts` or `eei.ts`)
- Second PR should actually remove the StateManager into a seperate package. 

We should also figure out exactly what the semantics of StateManager should be. This is essentially a bridge between VM and Trie. The specific VM-related tasks such as warm addresses should be kept in the VM, StateManager should not have to deal with these (this also makes implementing custom state managers much easier as you do not have to "copy" this code).

g11tech:
closing because of successful merge of https://github.com/ethereumjs/ethereumjs-monorepo/pull/1817 :slightly_smiling_face:

PR #1770, opened 2022-03-07
VM: add feature to add/disable (custom) precompiles

### LINKED ISSUE #1706: VM: add option to insert/override (custom) precompiles
It would be nice, especially for custom chains, to insert or override precompiles in the VM. This should be done in the constructor of the VM.

jochem-brouwer:
Closes #1706 

This will refactor `precompiles`; this will start using a Map. It will remove all exported functions and now add a function to `getPrecompiles`, this returns a (deep-copied) map. `VM` will have a pointer to this map. Now to get the precompile we can just use `get`. The end result will be similar to the approach used in #1705


jochem-brouwer:
Removing `precompiles` export is breaking so will mark this targeted to `develop` for now.

holgerd77:
Do you really pushed the full code you have written here? 🤔

holgerd77:
> Removing precompiles export is breaking so will mark this targeted to develop for now.

Ah sorry, in particular the above comment sounded as if you were further down the implementation road. Plainly asked since I was a bit confused. 🙂

jochem-brouwer:
Have to address the issue in `StateManager` because `StateManager` does not have access to VM and we want to keep it like that. (`generateAccessList` should move to VM, #1599). Also have to figure out why tests fail if I remove the arguments in the custom precompiles test file.

This PR should also wait for #1705 to merge since we can then incorporate this "update precompile logic" in the same way as the opcodes are updated. The access lists thing is a bit nasty we might want to wait until we have #1599. 

This PR moves the precompiles to a Map. This is very nice because now we can easily internally deal with the case that we have 2 EIPs which try to put a precompile at the same address. By `getActivePrecompiles` we can then, based upon the `Common`, figure out what precompile that is, this was not really doable before.

This is somewhat ready for a first round of reviews.

packages/vm/src/state/baseStateManager.ts
```diff
@@ -497,12 +497,13 @@ export abstract class BaseStateManager {
...
       const address = Address.fromString(`0x${addressStr}`)
-      const check1 = getActivePrecompiles(this._common).find((a) => a.equals(address))
+      // TODO FIXME
+      //const check1 = getActivePrecompiles(this._common).find((a) => a.equals(address))
       const check2 = addressesRemoved.find((a) => a.equals(address))
       const check3 =
         addressesOnlyStorage.find((a) => a.equals(address)) !== undefined && slots.size === 0
 
-      if (!check1 && !check2 && !check3) {
+      if (/*!check1 && */ !check2 && !check3) {
```
ryanio, line 506: todo still here

packages/vm/tests/api/customPrecompiles.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+      ],
+    })
+    const result = await VMOverride.runCall({
+      to: shaAddress,
+      gasLimit: BigInt(30000),
+      data: Buffer.from(''),
+      caller: sender,
+    })
+    st.ok(result.execResult.returnValue.equals(expectedReturn), 'return value is correct')
+    st.ok(result.execResult.gasUsed === expectedGas, 'gas used is correct')
```
ryanio, line 39: just a note, i'll add a commit to change these, but we can use `st.equal` with bigint 👍 and provides better error messages (expected, result) on error

```suggestion
    st.equal(result.execResult.gasUsed, expectedGas, 'gas used is correct')
```

packages/vm/tests/api/customPrecompiles.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+const sender = new Address(Buffer.from('44'.repeat(20), 'hex'))
+const newPrecompile = new Address(Buffer.from('ff'.repeat(20), 'hex'))
+const shaAddress = new Address(Buffer.from('0000000000000000000000000000000000000002', 'hex'))
+const expectedReturn = Buffer.from('1337', 'hex')
+const expectedGas = BigInt(10)
+
+// TODO if `opts` is removed (so function has no arguments) then tests fail!? Why?
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+function customPrecompile(opts: PrecompileInput): ExecResult {
```
ryanio, line 15: just looked into this, I believe because of PrecompileFunc's definition

```typescript
export interface PrecompileFunc {
  (opts: PrecompileInput): Promise<ExecResult> | ExecResult
}
```

if the number of params don't match when the function is called (here: one, `opts`) then it would error when calling the method.

packages/vm/src/state/baseStateManager.ts
ryanio: I figure the issue here is, you don't have the customPrecompiles that are passed and stored in the vm in the stateManager here. 

you could add it as an optional additional constructor opt and property on the baseStateManager, so they are available if defined

ryanio: have addressed in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1770/commits/b4342f83c06de41e4ef7ef328659c8de283d414c

ryanio (approved):
ok, nice, this lgtm! great work :) very easy to follow work as always. i made a few nits and finished two last todos, if my work looks all good to you @jochem-brouwer feel free to merge, or if @holgerd77 would like to take a look as well

packages/vm/src/state/stateManager.ts
```diff
@@ -57,6 +58,10 @@ export interface DefaultStateManagerOpts {
    * A {@link SecureTrie} instance
    */
   trie?: Trie
+  /**
+   * Optional custom precompiles (to exclude from access list generation)
+   */
+  customPrecompiles?: CustomPrecompile[]
```
holgerd77, line 64: Hmm, I find this dependency introduction very unfortunate and rather displaced, also in regard to the StateManager extraction.

Or can we assume that `generateAccessList()` will move out of the StateManager along refactoring anyhow and that this will therefore only be a temporary addition anyhow? 🤔

holgerd77:
I find it still very much a pity that we target *this one* towards `develop` with all these other great customizability additions (genesis, custom opcodes) which already went inot `master`. 😄 

Without having looked too deep, would it be a way to rather deprecated instead of remove the exports which are now the reason for the develop target?

If possible I would find it great if we might even want to target to bring this in the current release round, since it looks that we can do with the opcodes PR as well. That would make everything very "round"! 🙂

holgerd77:
Ug. Sorry. I am in a confused state today.

Wanted to merge the opcode PR. Argh.

Will push a new precompile branch and delete the new commit from `develop`.

holgerd77:
Have also force-pushed on develop and deleted the last commit resulting from this accidental merge.

PR #1776, opened 2022-03-10
devp2p: fix per-message debug logging

ryanio:
Quick fix following up on https://github.com/ethereumjs/ethereumjs-monorepo/pull/1600#pullrequestreview-905465290, the main issue here was that the per-message debug logs were not prefixed with `devp2p` (so they were being output e.g. `eth:STATUS` rather than `devp2p:eth:STATUS`)

edit: I have also further consolidated the classes, moving _version and _send to the parent class


holgerd77 (approved):
LGTM

PR #1777, opened 2022-03-10
common: update status fields for EIPs and hardforks

### LINKED ISSUE #1773: Common: Update EIP status fields, VM: move EIP-2315 to EXPERIMENTAL state
The `status` field in the [EIP files](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/common/src/eips) in `Common` has gotten a bit forgotten. We should go through each EIP file and update to the latest status from the different EIP specs, eventually (likely) with some additional other clean-up (switch to finalized EIP URLs (so: this format https://eips.ethereum.org/EIPS/eip-3675)), update descriptions and the like.

For the VM we have concluded that we should move [EIP-2315](https://eips.ethereum.org/EIPS/eip-2315) support back to `EXPERIMENTAL` (so: I guess basically adopt EIP comment block in VM `index.ts`), being still in `DRAFT` status and an eventual HF inclusion still somewhat far away.

ryanio:
Closes #1773

Have only updated the "status" fields, the rest are prettier formats.

(I wonder if we should open a separate PR to format all of our JSON and markdown files in the monorepo at once, the command from root would be: `npx prettier "**/*.{json,md}" --write`)

### DIFF
--- a/packages/common/src/eips/2315.json
+++ b/packages/common/src/eips/2315.json
@@ -3,7 +3,7 @@
   "number": 2315,
   "comment": "Simple subroutines for the EVM",
   "url": "https://eips.ethereum.org/EIPS/eip-2315",
-  "status": "Draft",
+  "status": "EXPERIMENTAL",
   "minimumHardfork": "istanbul",
   "gasConfig": {},
   "gasPrices": {
--- a/packages/common/src/eips/2315.json
+++ b/packages/common/src/eips/2315.json
@@ -3,7 +3,7 @@
   "number": 2315,
   "comment": "Simple subroutines for the EVM",
   "url": "https://eips.ethereum.org/EIPS/eip-2315",
-  "status": "EXPERIMENTAL",
+  "status": "Draft",
   "minimumHardfork": "istanbul",
   "gasConfig": {},
   "gasPrices": {
--- a/packages/common/src/eips/1559.json
+++ b/packages/common/src/eips/1559.json
@@ -1,26 +1,26 @@
 {
-    "name": "EIP-1559",
-    "number": 1559,
-    "comment": "Fee market change for ETH 1.0 chain",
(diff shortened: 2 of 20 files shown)


packages/common/src/eips/2315.json
```diff
@@ -3,7 +3,7 @@
   "number": 2315,
   "comment": "Simple subroutines for the EVM",
   "url": "https://eips.ethereum.org/EIPS/eip-2315",
-  "status": "Draft",
+  "status": "EXPERIMENTAL",
```
holgerd77, line 6: Ah, no, *this* status here is really the EIP status. This `EXPERIMENTAL` is only [used](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/index.ts#L46) in the VM to *then* indicate the status of the implementation.

I guess if we update here to `EXPERIMENTAL` and mix this with the EIP states it get's chaotic. That said, this `status` field is also only used internally, I added this when setting up the `Common` class with the above semantics in mind. Not sure if it's very helpful and we generally want to keep or not.

We might want to discuss separately. In doubt we just repeat redundant information here, which will likely always get outdated again and again. I guess if we decide we could even "just remove" without the fear of any breaking-change consequences.

Nevertheless: for now will update here back to `Draft` to keep things consistent.

holgerd77: But, to add some own opinion on that: I am currently also undecided (to remove or not), this might also be some useful helper on the status of EIPs we have implemented (otherwise one would likely not always look this up manually).

holgerd77, line 6:
```suggestion
  "status": "Draft",
```

holgerd77 (approved):
LGTM

jochem-brouwer:
Great, thanks! But moving 2315 to experimental in VM has not been addressed here (have reopened the issue).

deauna:
cheers

PR #1781, opened 2022-03-12
client: Sync stall investigation improvements

### LINKED ISSUE #1780: The syncing stalls pre-merge on Kiln
Left ethereumjs running overnight, but when checked back in morning, it was lagging behind, on a restart the fetcher queued jobs 
```
[03-12|06:41:28] DEBUG Syncing with peer: id=8cca8af03d89d026c9214607b443751c70ab2c6349e5cb502d3d1ad0a2c451f44fec8e7af2451c9a4e0751f6ef0888d4316d8f29b7e0b4fa961544c52a24387c address=10.244.0.164:30303 transport=rlpx protocols=eth inbound=false height=29590 
  client:fetcher Block fetcher instantiated interval=8000 first=26149 count=3442 destroyWhenDone=false +0ms
  client:fetcher Setup writer pipe. +1ms
  client:fetcher Created new tasks num=69 first=29549 count=42 +1ms
  client:fetcher Enqueued num=69 tasks +0ms
  client:fetcher No idle peer available, skip next job execution. +2ms
  client:fetcher Requested blocks=26149-26199 from id=8cca8af0 address=10.244.0.164:30303 (received: 50 headers / 50 bodies) +66ms
  client:fetcher Error storing fetcher results in blockchain (blocks num=50): Error: could not find parent header (block header number=26149 hash=0xe68f98465a56fd75f9790ebe5abae37154961ff5f8b29b7435fa2d65f82bd534 hf=preMerge baseFeePerGas=7) +45ms
[03-12|06:41:28] WARN Error storing received block or header result: Error: could not find parent header (block header number=26149 hash=0xe68f98465a56fd75f9790ebe5abae37154961ff5f8b29b7435fa2d65f82bd534 hf=preMerge baseFeePerGas=7) 
  client:fetcher Requested blocks=26199-26249 from id=8cca8af0 address=10.244.0.164:30303 (received: 50 headers / 50 bodies) +20ms
  client:fetcher Requested blocks=26249-26299 from id=8cca8af0 address=10.244.0.164:30303 (received: 50 headers / 50 bodies) +75ms
  client:fetcher Requested blocks=26299-26349 from id=8cca8af0 address=10.244.0.164:30303 (received: 50 headers / 50 bodies) +15ms
  client:fetcher Requested blocks=26349-26399 from id=8cca8af0 address=10.244.0.164:30303 (received: 50 headers / 50 bodies) +58ms
  client:fetcher Requested blocks=26399-26449 from id=8cca8af0 address=10.244.0.164:30303 (received: 50 headers / 50 bodies) +10ms
  client:fetcher Readable state length exceeds max queue size, skip next job execution. +3ms
[03-12|06:41:33] DEBUG eth_blockNumber called with params:
```
but resulted into an error when it fetched the first job as seen in log above:
```
  client:fetcher Error storing fetcher results in blockchain (blocks num=50): Error: could not find parent header (block header number=26149 hash=0xe68f98465a56fd75f9790ebe5abae37154961ff5f8b29b7435fa2d65f82bd534 hf=preMerge baseFeePerGas=7) +45ms
```
Post this all other enqueued jobs obviously fail as well.
This goes away only if db is clean wiped and client restarted and linearity of the chain is established.

cc @ryanio

g11tech:
While running on kiln, some more issues of sync stalling were observed: https://github.com/ethereumjs/ethereumjs-monorepo/issues/1780

The trigger of the sync stall is when the fetcher adds block to blockchain and the blocks fetched are not linearly connected, it leads to issue. 
~~One of the issue was writer was being destroyed.  This PR removes that condition of the writer being destroyed on error  as well as improves the fetcher logging a bit for better debugging.~~
UPDATE: Figured out that the writer's call back `cb(error)` switches off the writer. In case such a thing happened, its better to return an error in `await fetch(..)` and let the sync loop try with a different peer if required.
This also fixes the behavior that other peers were never tried because `fetcher.fetch` was never failing and `syncWithPeer` was never returning/erroring if there was any writer error.
 
- [x] fixes the fetcher getting stalled because if now  `syncWithPeer` errors, and then the new `sync` loop is started with the creating of a new fetcher.
- [x] ~~However there is still an issue that the process dies after too many errors (modified the `store` fn to error every 5 times its called for debugging purposes), so futher resolution will be required, but that would be independent of this `syncWithPeer` fix.~~ RESOLVED the above crash by removing fetcher restart when errored already
- [x] ~~UNRESOLVED: seems like running peer isn't being updated of its best ttd when new sync target is updated in in NewBlockHashes, or NewBlock handling of `async handleEth(message: any, peer: Peer): Promise<void> {`~~ needed this update when the fetcher failed and best peer need to be found again to syncWithPeer with new fetcher (even though current peer was best peer), disconnect/reconnet lead to update of its status (though temporary ban would have been better but it wasn't reconnecting to the peer fast) and restart of the sync
- [x] Handle `could not find parent header` reorg possibility 

Closes #1780

### DIFF
--- a/packages/client/lib/config.ts
+++ b/packages/client/lib/config.ts
@@ -198,6 +198,12 @@ export interface ConfigOptions {
    * If not provided, defaults to the primary account.
    */
   minerCoinbase?: Address
+
+  /**
+   * If there is supposedly a re-org, this is a safe distance from which
+   * to refetch and try feeding the blocks
+   */
+  safeReorgDistance?: number
 }
 
 export class Config {
@@ -218,6 +224,7 @@ export class Config {
   public static readonly MAXPEERS_DEFAULT = 25
   public static readonly DNSADDR_DEFAULT = '8.8.8.8'
   public static readonly DEBUGCODE_DEFAULT = false
+  public static readonly SAFE_REORG_DISTANCE = 100
 
   public readonly logger: Logger
   public readonly syncmode: SyncMode
@@ -242,6 +249,7 @@ export class Config {
   public readonly mine: boolean
   public readonly accounts: [address: Address, privKey: Buffer][]
   public readonly minerCoinbase?: Address
+  public readonly safeReorgDistance: number
 
   public synchronized: boolean
   public lastSyncDate: number
@@ -274,6 +282,7 @@ export class Config {
     this.mine = options.mine ?? false
     this.accounts = options.accounts ?? []
     this.minerCoinbase = options.minerCoinbase
+    this.safeReorgDistance = options.safeReorgDistance ?? Config.SAFE_REORG_DISTANCE
 
     this.synchronized = false
     this.lastSyncDate = 0
--- a/packages/client/lib/config.ts
+++ b/packages/client/lib/config.ts
@@ -200,8 +200,8 @@ export interface ConfigOptions {
   minerCoinbase?: Address
 
   /**
-   * If there is supposedly a re-org, this is a safe distance from which
-   * to refetch and try feeding the blocks
+   * If there is a reorg, this is a safe distance from which
+   * to try to refetch and refeed the blocks.
    */
   safeReorgDistance?: number
 }
--- a/packages/client/lib/net/peerpool.ts
+++ b/packages/client/lib/net/peerpool.ts
@@ -25,6 +25,7 @@ export class PeerPool {
   public running: boolean
 
   private _statusCheckInterval: NodeJS.Timeout | undefined /* global NodeJS */
+  private _reconnectTimeout: NodeJS.Timeout | undefined
 
   /**
    * Create new peer pool
@@ -88,6 +89,7 @@ export class PeerPool {
       await this.close()
     }
     clearInterval(this._statusCheckInterval as NodeJS.Timeout)
+    clearTimeout(this._reconnectTimeout as NodeJS.Timeout)
     this.running = false
     return true
   }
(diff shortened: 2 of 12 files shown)


ryanio:
thanks for opening! just as a stylistic note for the future, we usually don't put commas in between the key=value outputs

ryanio:
this is looking really promising :) wasn't able to test today but tomorrow morning will definitely spin up and test these changes.

ryanio (dismissed):
LGTM!

g11tech:
> LGTM!

@ryanio fixed the out of order processing of the fetcher results, seems to be working good now. :t-rex:

holgerd77:
@g11tech can I update the branch here and eventually merge?

holgerd77:
(I would do a merge of #1790 before and wait for the CI to finish, so please do not update the branch yourself yet (at least not via UI here on GitHub))

holgerd77 (approved):
LGTM

holgerd77:
Now sync looks like this, is this correct? 🤔

grafik

Is this all due to re-orgs - these re-downloads of block ranges (so: is it so chaotic at the tip of the chain 😬 😋)? Or is this still due to some bugs?

holgerd77:
(if you have a quick answer - great - but generally take your time. If we do not release today, so be it. Was somewhat ambitious anyhow, in doubt we should rather release something (more) stable 1-2 days later.

But we'll see.)

g11tech:
> (if you have a quick answer - great - but generally take your time. If we do not release today, so be it. Was somewhat ambitious anyhow, in doubt we should rather release something (more) stable 1-2 days later.
> 
> But we'll see.)

hi @holgerd77 :wave: 
yes this looks good, https://github.com/ethereumjs/ethereumjs-monorepo/pull/1792 will improve it a bit further, so we can include that as well :slightly_smiling_face:

PR #1783, opened 2022-03-13
client: expand getLogs to accept array of addresses

ryanio:
Thanks to "Sea Monkey" in our discord, teku was calling eth_getLogs with an array of addresses (rather than a single address) and this is valid according to the spec, so this PR fixes our endpoint.

### DIFF
--- a/packages/client/lib/execution/receipt.ts
+++ b/packages/client/lib/execution/receipt.ts
@@ -159,7 +159,7 @@ export class ReceiptsManager extends MetaDBManager {
   async getLogs(
     from: Block,
     to: Block,
-    address?: Buffer,
+    addresses?: Buffer[],
     topics: (Buffer | Buffer[] | null)[] = []
   ): Promise<GetLogsReturn> {
     const returnedLogs: GetLogsReturn = []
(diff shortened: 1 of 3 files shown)


deauna (dismissed):
left  behind

packages/client/lib/execution/receipt.ts
```diff
@@ -159,7 +159,7 @@ export class ReceiptsManager extends MetaDBManager {
   async getLogs(
     from: Block,
     to: Block,
-    address?: Buffer,
+    addresses?: Buffer[],
```
holgerd77, line 162: Not daring a review atm, not "into" client dev and context sufficiently enough atm. But first read question: isn't this change now preventing the original use case just passing in a single address? 🤔

holgerd77: Yes, obviously, but the question was rather if it should nevertheless still allow passing in as a non-array address?

The [spec](https://eth.wiki/json-rpc/API#eth_getlogs) says:

> address: DATA|Array, 20 Bytes - (optional) Contract address or a list of addresses from which logs should originate

So if I read correctly I guess this allows both and typing should therefore be rather: `Buffer|Buffer[]` ? 🤔

holgerd77: Ah ok, thanks!

holgerd77 (approved):
LGTM

PR #1774, opened 2022-03-10
New Releases: Kiln v2 Testnet, EIP-3860, Extended VM Genesis (Code/Storage)

holgerd77:
This PR prepares on releases for the Client, VM, Block, Trie, Tx, Common libraries (main uncertainty point: `devp2p`, maybe someone can confirm that no release is necessary), mainly centering around client support for the upcoming Kiln `v2` Merge testnet, but also coming with other substantial features like EIP-3860 support and extended support for genesis state initialization in the VM.

Note that this release round will likely *not yet* include the EOF work from #1719, since this needs some more time and at the same time Kiln v2 launch is relatively imminent and we would want to have a client version out of the door to allow people participating with our client.

Release is currently planned for Saturday or Monday. Another subsequent bugfix release will likely be following at the end of next week and e.g. containing the final Merge block after the Merge transition happening sometime next week.

WIP but already open for some first-read-review. 🙂

PR #1785, opened 2022-03-13
vm: delete extra console.log

tynes:
This was showing up when linting https://github.com/ethereumjs/ethereumjs-monorepo/pull/1768, so this PR fixes the linting error


ryanio:
Thanks, actually we already have this in #1774 and we should be merging that in the next day or two so I'll close here. Thanks again.

Issue #1717, opened 2022-02-16
v6 Breaking Release Planning


holgerd77:
Follow-up on #907
Work branch: [`develop`](https://github.com/ethereumjs/ethereumjs-monorepo/tree/develop)
Replaces #1024 

Update of `develop` branch: no merges / only rebase (~weekly by @holgerd77)
Update of feature branches: no merges / only rebase

**Planned Release Date** (Timeframe): Latest somewhat before Merge HF (~Summer 2022), eventually earlier

#### Error Management

- [ ] Rework of error management with more targeted error messages and error-string-independent handling, work is being done on this by @gabrocheleau in: #1469

#### BigInt Support

- [x] Switch from `BN.js` to native JS BigInts for external API usage and/or inner-functional computations and data handling, so on the entire stack (also switch along API on BN.js using functions (e.g. `Common.nextHardforkBlock()`)).
  - Note: This is highly connected to a decision on #1668 - so an upgrade to `ethereum-cryptography` `v1.0`, since the underlying noble libraries use `BigInt` anyhow which would be an implicit decision to integrate native BigInts throughout the stack.

##### TODOs

- [x] PoC PR on 2. on some selected functions proposing a coherent strategy on how to replace `BNLike`, what values to allow as input and the like, #1744 
- [x] Coherent `ES2020` strategy: `ES2018` + `ES2020.BigInt`, `ES2019.Object`, #1729 

#### Consensus Encapsulation / Separation

- [x] Separate consensus functionality in Blockchain/Block (and eventually other libraries) and give this a unified API. This will allow for an easier consensus switch or replacement. #1044 

#### ESM Support

- [ ] Add ESM Support as outlined in #1654 

#### [ In Discussion / Needs Decision ]

- [x] Make all constructors from libraries with async constructor methods (VM, Blockchain) `protected` #1811
- [ ] Consider a hash/stateRoot genesis handling refactor along Blockchain/VM/Block/Common dependency chain, see https://github.com/ethereumjs/ethereumjs-monorepo/pull/1833#issuecomment-1088445250 and https://github.com/ethereumjs/ethereumjs-monorepo/pull/1833#issuecomment-1091181809 for discussion and TODO summary

#### Standalone StateManager

- [x] Create a standalone `StateManager` package together with a state manager functionality rework (leave all VM related stuff in the VM) and an API overhaul (all interface functions necessary? Genesis functionality?) Issue: #1599 

##### Related TODOs / Decisions

[...]

Issue #1044, opened 2021-01-15
Block: Encapsulate Consensus Mechanism


holgerd77:
Now with PoA clique added to the `Block` class in #1032 it becomes apparent that the consensus functionality in the `Block` class is too tightly coupled with the structural and further generic functionality and - especially on Clique - new functions like `cliqueIsEpochTransition()` or similar somewhat pollute the namespace of the main API.

It would be therefore desirable to better encapsulate the consensus logic, tow possibilities - subclassing and a dedicated consensus module - have been discussed in the chat, here is an extract:

---

From @holgerd77 (somewhat summarizing the previous posts):

"@Jochem-brouwer @ryanio yeah, I was thinking about this too a bit and wanted to raise this question about the clique* methods along the PR code. I decided to keep it simple but a bit unelegant at the end for backwards-compatibility reasons.

To add on...

>to not crowd the block class with PoA concepts

Actually the class was "crowded" with ethash concepts already before. Now it just becomes more apparent with the two concepts included.

During my work I briefly thought about both concepts - so the subclassing with PoABlock (or something similar) as well as the consensus capsulation - didn't dig much further with my thoughts though here.

Subclassing I mentally discarded relatively quickly because I wasn't a fan of having code like:

```typescript
if (PoA) {
  new PoABlock()
} else {
  new PoWBlock()
}
```

throughout the codebase. Not sure if this is a valid argument though.

Regarding a consensus class I was on first thought somewhat sceptical as well if this can be encapsulated (easily enough) since there is a lot of code - like e.g. the extraData  checks - really deeply in the generic functions code base. On second thought though I am slowly moving over that this might just be a question of giving this some more design thinking.

I am still sceptical if we completely manage to capture a generic API with an abstract class or an interface since e.g. on clique there is very specific logic in these extra functions with getEpochTransitionBlock() or whatever. We might nevertheless be able to get to good common denominator which already encapsulates a lot.
Not the end of the world if there is still some type-specific functionality in there at the end like - to imagine something:

```typescript
block.consensus.isEpochTransition()
```

or something. These kind of calls should be encapsulated well enough in the other libraries by the Common switches `(if (common.consensusType() === 'poa') { ... }` .

[...]

PR #1786, opened 2022-03-14
blockchain: fix updateCache

samlior:
When the database operation is `del`, `this.baseDBOp.value` is undefined instead of `Buffer`.
In the old code, when deleting a value, the value is deleted in the database but not in the cache, causing the next read value still exists (because `DBManager` always tries to load data from cache first)

BTW, I tried to add a test case for this fix, but to find out that neither `DBManager` nor `DBOp` has a test case, maybe this should be improved?

### DIFF
--- a/packages/blockchain/src/db/operation.ts
+++ b/packages/blockchain/src/db/operation.ts
@@ -144,9 +144,10 @@ export class DBOp {
   }
 
   public updateCache(cacheMap: CacheMap) {
-    if (this.cacheString && cacheMap[this.cacheString] && Buffer.isBuffer(this.baseDBOp.value)) {
+    if (this.cacheString && cacheMap[this.cacheString]) {
       if (this.baseDBOp.type == 'put') {
-        cacheMap[this.cacheString].set(this.baseDBOp.key, this.baseDBOp.value)
+        Buffer.isBuffer(this.baseDBOp.value) &&
+          cacheMap[this.cacheString].set(this.baseDBOp.key, this.baseDBOp.value)
       } else if (this.baseDBOp.type == 'del') {
         cacheMap[this.cacheString].del(this.baseDBOp.key)
       } else {


packages/blockchain/src/db/operation.ts
```diff
@@ -144,9 +144,10 @@ export class DBOp {
   }
 
   public updateCache(cacheMap: CacheMap) {
-    if (this.cacheString && cacheMap[this.cacheString] && Buffer.isBuffer(this.baseDBOp.value)) {
+    if (this.cacheString && cacheMap[this.cacheString]) {
       if (this.baseDBOp.type == 'put') {
-        cacheMap[this.cacheString].set(this.baseDBOp.key, this.baseDBOp.value)
+        Buffer.isBuffer(this.baseDBOp.value) &&
```
ryanio, line 149: thanks for the PR!

i find this type of conditional chaining kind of awkward, can you instead add to the if statement above?

samlior:
```ts
public updateCache(cacheMap: CacheMap) {
  if (this.cacheString && cacheMap[this.cacheString]) {
    if (this.baseDBOp.type == 'put' && Buffer.isBuffer(this.baseDBOp.value)) {
      cacheMap[this.cacheString].set(this.baseDBOp.key, this.baseDBOp.value)
    } else if (this.baseDBOp.type == 'del') {
      cacheMap[this.cacheString].del(this.baseDBOp.key)
    } else {
      throw new Error('unsupported db operation on cache')
    }
  }
}
```

It will throw an exception if type is `put` and value is not `Buffer`

holgerd77: Ah ok, thanks for the explanation! 👍

holgerd77 (approved):
Thanks a lot, this is a really important fix, since this can cause all sorts of underlying weird behavior, which is extremely hard to debug, especially if triggered e.g. in the outer client context! 😄

Will directly add this to the upcoming round of releases here #1774.

Yes, additional tests might be a good idea. We will likely do some substantial refactoring around the Blockchain class during the work towards the next major releases #1717 - e.g. with the encapsulation of all consensus functionality. Will eventually make sense to wait for that before going too deep into test writing. But also not 100% sure, might not affect the tests (test structure) as much as I am thinking on this first round.

holgerd77:
(small nit: in case you are opening more PRs towards the monorepo (highly welcome! 😀) it would be good if you can open on your side from a dedicated feature branch and not from (your) fork `master` branch. That makes local testing and the like more robust and less confusing).

packages/blockchain/src/db/operation.ts
ryanio: gotcha, thanks. i think i still would've preferred if we formatted this as an additional inside if statement

```
if (this.baseDBOp.type == 'put') {
  if (Buffer.isBuffer(this.baseDBOp.value)) {
    cacheMap[this.cacheString].set(this.baseDBOp.key, this.baseDBOp.value)
  }
}
```

but that's fine for now. thanks again for the pr :)

good point about missing tests, would be great to have some for these

holgerd77: Ok, guess we can maybe update "on occasion" if someone stumbles upon further down the road (e.g. @emersonmacro along his work on #1044).

Issue #1738, opened 2022-02-23
Block: Rework fromHeaderData() and main BlockHeader constructor


holgerd77:
Part of #1717 

The static constructors for the Block and Tx libraries (like `Block.fromBlockData()`, Block.fromValuesArray(),...) have been introduced along the last breaking release round in November 2020 (e.g. [Block v3.0](https://github.com/ethereumjs/ethereumjs-monorepo/releases/tag/%40ethereumjs%2Fblock%403.0.0).

While we aligned the main constructor and the `from*Data()` constructor method in the tx library in an additional iteration (see e.g. for the [legacy tx](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/legacyTransaction.ts#L35)) (so basically: we got this right) by having the constructor take in the same data dictionary as the `from*Data()` constructor (which is then simply an alias), we missed this opportunity in the `Block` library for the `BlockHeader`.

So a `BlockHeader` object should take in the same `headerData` dictionary for the main constructor as being taken in `fromHeaderData()` and the latter should be made just an alias for the main constructor. Along the default value setting should be moved over to the main constructor (this needs to be aligned with the `fromValuesArray()` instantiation.

gabrocheleau:
Should we close this since https://github.com/ethereumjs/ethereumjs-monorepo/pull/1787 was merged in?

PR #1787, opened 2022-03-14
block: unify blockHeader constructor input shapes

gabrocheleau:
This PR resolves https://github.com/ethereumjs/ethereumjs-monorepo/issues/1738

### DIFF
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -78,46 +78,7 @@ export class BlockHeader {
    * @param opts
    */
   public static fromHeaderData(headerData: HeaderData = {}, opts: BlockOptions = {}) {
-    const {
-      parentHash,
-      uncleHash,
-      coinbase,
-      stateRoot,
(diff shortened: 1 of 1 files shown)


packages/block/src/header.ts
```diff
@@ -243,6 +186,46 @@ export class BlockHeader {
...
+      difficulty: BigInt(0),
+      number: BigInt(0),
+      gasLimit: DEFAULT_GAS_LIMIT,
+      gasUsed: BigInt(0),
+      timestamp: BigInt(0),
+      extraData: Buffer.from([]),
+      mixHash: zeros(32),
+      nonce: zeros(8),
+      baseFeePerGas: undefined,
+    }
```
holgerd77, line 206: Ah, that's nice. We might want to do this in other libraries as well, that gives a really good overview! 😄

holgerd77 (approved):
LGTM, thanks! 😄

PR #1788, opened 2022-03-14
Implement EIP3074: AUTHCALL

jochem-brouwer:
Implements https://eips.ethereum.org/EIPS/eip-3074

### DIFF
--- a/packages/common/src/eips/3074.json
+++ b/packages/common/src/eips/3074.json
@@ -1,0 +1,25 @@
+{
+  "name": "EIP-3074",
+  "number": 3074,
+  "comment": "AUTH and AUTHCALL opcodes",
+  "url": "https://eips.ethereum.org/EIPS/eip-3074",
+  "status": "Review",
+  "minimumHardfork": "london",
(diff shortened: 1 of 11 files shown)


ryanio:
if there is no urgency to get this in a vm v5 release, then i wonder if we should target develop (vm v6) so we don't have to refactor the code from BN to bigint again

jochem-brouwer:
There is no urgency, will target `develop`

jochem-brouwer:
Right, before I continue should first rebase on develop.

jochem-brouwer:
Closed in favor of #1789

holgerd77:
> if there is no urgency to get this in a vm v5 release, then i wonder if we should target develop (vm v6) so we don't have to refactor the code from BN to bigint again

Yes, that makes a dramatic amount of sense 😜, hadn't thought about that either yet but we can very well start to do so. This will take some time to implement anyhow on these new bigger tasks, and v6 is finally not *that* far away any more (2-3 months I would say).

PR #1789, opened 2022-03-14
Implement EIP3074: Authcall

jochem-brouwer:
Implements https://eips.ethereum.org/EIPS/eip-3074

### DIFF
--- a/packages/vm/src/evm/opcodes/EIP2929.ts
+++ b/packages/vm/src/evm/opcodes/EIP2929.ts
@@ -18,7 +18,7 @@ export function accessAddressEIP2929(
   address: Address,
   common: Common,
   chargeGas = true,
-  isSelfdestruct = false
+  isSelfdestructOrAuthcall = false
 ): bigint {
   if (!common.isActivatedEIP(2929)) return BigInt(0)
 
@@ -35,7 +35,7 @@ export function accessAddressEIP2929(
       return BigInt(common.param('gasPrices', 'coldaccountaccess'))
     }
     // Warm: (selfdestruct beneficiary address reads are not charged when warm)
-  } else if (chargeGas && !isSelfdestruct) {
+  } else if (chargeGas && !isSelfdestructOrAuthcall) {
     return BigInt(common.param('gasPrices', 'warmstorageread'))
   }
   return BigInt(0)
--- a/packages/vm/src/evm/opcodes/functions.ts
+++ b/packages/vm/src/evm/opcodes/functions.ts
@@ -5,9 +5,12 @@ import {
   KECCAK256_NULL,
   TWO_POW256,
   MAX_INTEGER_BIGINT,
-  setLengthLeft,
   bufferToBigInt,
   bigIntToBuffer,
+  setLengthLeft,
+  ecrecover,
+  publicToAddress,
+  SECP256K1_ORDER_DIV_2,
 } from 'ethereumjs-util'
 import {
   addressToBuffer,
@@ -24,6 +27,8 @@ import {
 import { ERROR } from '../../exceptions'
 import { RunState } from './../interpreter'
 import { exponentation } from '.'
+
+const EIP3074MAGIC = Buffer.from('03', 'hex')
 
 export interface SyncOpHandler {
   (runState: RunState, common: Common): void
@@ -982,7 +987,71 @@ export const handlers: Map<number, OpHandler> = new Map([
       runState.stack.push(ret)
     },
   ],
-  // 0x06: STATICCALL
+  // 0xf6: AUTH
+  [
+    0xf6,
+    async function (runState) {
+      const [commitUnpadded, yParity, r, s] = runState.stack.popN(4)
+      if (s > SECP256K1_ORDER_DIV_2) {
+        trap(ERROR.AUTH_INVALID_S)
+      }
+
+      const commit = setLengthLeft(bigIntToBuffer(commitUnpadded), 32)
+      const paddedInvokerAddress = setLengthLeft(runState.eei._env.address.buf, 32)
+      const chainId = setLengthLeft(bigIntToBuffer(runState.eei.getChainId()), 32)
+      const message = Buffer.concat([EIP3074MAGIC, chainId, paddedInvokerAddress, commit])
+      const msgHash = keccak256(message)
+
+      let recover
+      try {
+        recover = ecrecover(msgHash, Number(yParity) + 27, bigIntToBuffer(r), bigIntToBuffer(s))
+      } catch (e) {
+        // Malformed signature, push 0 on stack, clear auth variable and return
+        runState.stack.push(BigInt(0))
+        runState.eei._env.auth = undefined
+        return
+      }
+
+      const addressBuffer = publicToAddress(recover)
+      const address = new Address(addressBuffer)
+      runState.eei._env.auth = address
+
+      const addressBigInt = bufferToBigInt(addressBuffer)
+      runState.stack.push(addressBigInt)
+    },
+  ],
+  // 0xf7: AUTHCALL
+  [
+    0xf7,
+    async function (runState) {
+      const [
+        _currentGasLimit,
+        addr,
+        value,
+        _valueExt,
+        argsOffset,
+        argsLength,
+        retOffset,
+        retLength,
+      ] = runState.stack.popN(8)
+
+      const toAddress = new Address(addressToBuffer(addr))
+
+      const gasLimit = runState.messageGasLimit!
+      runState.messageGasLimit = undefined
+
+      let data = Buffer.alloc(0)
+      if (argsLength !== BigInt(0)) {
+        data = runState.memory.read(Number(argsOffset), Number(argsLength))
+      }
+
+      const ret = await runState.eei.authcall(gasLimit, toAddress, value, data)
+      // Write return data to memory
+      writeCallOutput(runState, retOffset, retLength)
+      runState.stack.push(ret)
+    },
+  ],
+  // 0xfa: STATICCALL
   [
     0xfa,
     async function (runState) {
--- a/packages/vm/src/evm/opcodes/gas.ts
+++ b/packages/vm/src/evm/opcodes/gas.ts
@@ -441,6 +441,63 @@ export const dynamicGasHandlers: Map<number, AsyncDynamicGasHandler | SyncDynami
       },
     ],
     [
+      /* AUTHCALL */
+      0xf7,
+      async function (runState, gas, common): Promise<bigint> {
+        if (runState.eei._env.auth === undefined) {
+          trap(ERROR.AUTHCALL_UNSET)
+        }
+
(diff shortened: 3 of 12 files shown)


packages/vm/tests/api/EIPs/eip-3074-authcall.spec.ts
```diff
@@ -0,0 +1,597 @@
...
+      const expected = preGas - preGas / 64n - 2n
+      st.equals(gasBigInt, expected, 'forwarded max call gas')
+
+      const expectedBalance = PREBALANCE - result.amountSpent - value
+      const account = await vm.stateManager.getAccount(callerAddress)
+
+      st.equals(account.balance, expectedBalance, 'caller balance ok')
+
+      const contractAccount = await vm.stateManager.getAccount(contractAddress)
+      st.equals(contractAccount.balance, 2n, 'contract balance ok')
```
jochem-brouwer, line 448: Extremely weird, this test fails :thinking:

packages/vm/tests/api/EIPs/eip-3074-authcall.spec.ts
```diff
@@ -0,0 +1,597 @@
...
+      gasLimit: 1000000,
+      gasPrice: 10,
+    }).sign(callerPrivateKey)
+
+    const result = await vm.runTx({ tx, block })
+    st.ok(
+      result.execResult.exceptionError?.error === ERROR.AUTHCALL_UNSET,
+      'threw with right error'
+    )
+    st.ok(result.amountSpent === tx.gasPrice * tx.gasLimit, 'spent all gas')
```
ryanio, line 475: can use st.equal here
```suggestion
    st.equal(result.amountSpent, tx.gasPrice * tx.gasLimit, 'spent all gas')
```

packages/vm/src/evm/opcodes/functions.ts
```diff
@@ -976,7 +981,71 @@ export const handlers: Map<number, OpHandler> = new Map([
...
+        retLength,
+      ] = runState.stack.popN(8)
+
+      const toAddress = new Address(addressToBuffer(addr))
+
+      const gasLimit = runState.messageGasLimit!
+      runState.messageGasLimit = undefined
+
+      let data = Buffer.alloc(0)
+      if (!(argsLength === BigInt(0))) {
```
ryanio, line 1038: nit
```suggestion
      if (argsLength !== BigInt(0)) {
```

packages/vm/tests/api/EIPs/eip-3074-authcall.spec.ts
ryanio: there are few more places like this, can help point them out or i can fix if you'd like.

another nit if you could use `st.equal` rather than `st.equals` we've normalized the repo on the former :)

packages/vm/src/evm/opcodes/gas.ts
```diff
@@ -428,6 +428,58 @@ export const dynamicGasHandlers: Map<number, AsyncDynamicGasHandler> = new Map<
...
+        }
+      }
+
+      let gasLimit = maxCallGas(
+        runState.eei.getGasLeft() - gas,
+        runState.eei.getGasLeft() - gas,
+        runState,
+        common
+      )
+      if (!(currentGasLimit === BigInt(0))) {
```
ryanio, line 472: nit
```suggestion
      if (currentGasLimit !== BigInt(0)) {
```

packages/vm/tests/api/EIPs/eip-3074-authcall.spec.ts
```diff
@@ -0,0 +1,597 @@
...
+      to: contractAddress,
+      gasLimit: 1000000,
+      gasPrice: 10,
+    }).sign(callerPrivateKey)
+
+    const result = await vm.runTx({ tx, block })
+    st.ok(
+      result.execResult.exceptionError?.error === ERROR.AUTHCALL_UNSET,
+      'threw with right error'
+    )
```
ryanio, line 509: can also use st.equal with string error msg comparison

```suggestion
    st.equal(result.execResult.exceptionError?.error, ERROR.AUTHCALL_UNSET, 'threw with right error')
```

ryanio:
super cool, clean implementation as always!

jochem-brouwer:
Cool, I have written the tests which I wanted, so this is now ready for review. Please let me know if you need help reviewing those bytecode tests.

packages/vm/src/evm/opcodes/EIP2929.ts
```diff
@@ -18,7 +18,7 @@ export function accessAddressEIP2929(
   address: Address,
   common: Common,
   chargeGas = true,
-  isSelfdestruct = false
+  isSelfdestructOrAuthCall = false
```
ryanio, line 21: should this be `isSelfdestructOrAuthcall`? we should ensure we are using the same case style for every use.

packages/vm/tests/api/EIPs/eip-3074-authcall.spec.ts
```diff
@@ -0,0 +1,680 @@
...
+  const message = Buffer.concat([Buffer.from('03', 'hex'), chainId, paddedInvokerAddress, commit])
+  const msgHash = keccak256(message)
+  return ecsign(msgHash, privateKey, 0)
+}
+
+/**
+ * This method returns the bytecode in order to set AUTH
+ * @param commitUnpadded - The commit
+ * @param signature - The signature as obtained by `signMessage`
+ * @returns
```
ryanio, line 75: empty `@returns` here (and L110), you can just remove and typedoc knows it returns a Buffer, unless you want to add more description (same with the params). i do know this is in a test file though so just a nit :)

packages/vm/tests/api/EIPs/eip-3074-authcall.spec.ts
```diff
@@ -0,0 +1,680 @@
...
+// Bytecode to store CALLER in slot 0 and GAS in slot 1 and the first 32 bytes of the input in slot 2
+// Returns the entire input as output
+const STORECALLER = Buffer.from('5A60015533600055600035600255366000600037366000F3', 'hex')
+
+/**
+ * This signs a message to be used for AUTH opcodes
+ * @param commitUnpadded - The commit which we want to sign
+ * @param address - The contract address we are using AUTH on
+ * @param privateKey - The private key of the account to sign
+ * @returns - Returns the signed message
```
ryanio, line 60: nit, no need for extra dash (same with all the params, but it doesn't seem VS Code adds an extra `-` like in this case)

> Screen Shot 2022-03-24 at 12 19 26 PM

```suggestion
 * @returns The signed message
```

packages/vm/tests/api/EIPs/eip-3074-authcall.spec.ts
```diff
@@ -0,0 +1,680 @@
...
+})
+
+// Setups the environment for the VM, puts `code` at contractAddress and also puts the STORECALLER bytecode at the contractStorageAddress
+async function setupVM(code: Buffer) {
+  const vm = new VM({ common })
+  await vm.stateManager.putContractCode(contractAddress, code)
+  await vm.stateManager.putContractCode(contractStorageAddress, STORECALLER)
+  const account = await vm.stateManager.getAccount(callerAddress)
+  account.balance = PREBALANCE
+  await vm.stateManager.putAccount(callerAddress, account)
```
ryanio, line 286: instead of this pattern you can use modifyAccountFields now:

```suggestion
  await vm.stateManager.modifyAccountFields(callerAddress, { balance: PREBALANCE })
```

jochem-brouwer:
Addressed your items @ryanio :smile: 

Seems like I have a lot of merge conflicts now. How do I fix this? Should I cherry-pick all commits on top of the new develop? Otherwise I think I am gonna do that rebase work twice (this should already be in `develop` ?)

ryanio:
🙏 yes cherry picking on new develop seems to be easiest way, i just did it on my branch and had no conflicts

ryanio (approved):
LGTM! implementation, tests were all really easy to follow and comprehensive. re-checked the spec and everything seems covered. super exciting!

holgerd77:
Rebased this via UI (did a test rebase locally).

holgerd77:
Ok, given the existing approval I would now merge this in. Also had a very very rough look - without going into *any* implementation details, can agree, this looks extremely clean, with a substantial amount of tests added.

Also think that this PR shouldn't cause too much problems from a rebase perspective since functionality is added extremely locally/grouped.

So, thanks a lot @jochem-brouwer, great work on this! 😄 🎉

cyrusadkisson:
Hi. I'm an app/solidity dev but have an interest in 3074 implementations due to a technical quirk of a contract I deployed in 2015. 

I'm keen to see the following in 3074 evm client implementations:
https://ethereum-magicians.org/t/eip-3074-auth-and-authcall-opcodes/4880/103

TL;DR - I think it makes the most sense, in an AUTHCALL, for tx.origin to be set to the address of the account being remote-controlled, not the account initializing the transaction (i.e. paying the gas) as it is, in spirit, the remote-controlled address performing the action.

I'm way out of my depth with your code, so a question: how does your implementation handle tx.origin?

jochem-brouwer:
The implementation will follow the spec as it is currently written in the EIP, i.e. `tx.origin` is the account which has instantiated the original transaction, i.e. also the account paying for gas.

cyrusadkisson:
Ok so I need to get it changed in the EIP first. Thanks.

Do you have an opinion either way? Seems to me that if tx.origin were meant to strictly be the gas payer, it would have been called "tx.gaspayer" instead.

jochem-brouwer:
My initial opinion would be that we should not change the semantics of the origin opcode, I think it is generally agreed that this is the transaction origin thus the gaspayer. If you want the actual remote-controlled address (i.e. the AUTH address) then in the initial CALL layer you can use the `CALLER` opcode. If you also want this in deeper layers then you should probably add another opcode. This however begs the question, what should this opcode do in the situation if you first AUTHCALL from account A and then in this layer you AUTHCALL from account B. What should the opcode return, and why?

I think this is a rather complex situation which I don't have a strong opinion on yet. What motivates you to introduce these rules?

cyrusadkisson:
Well this is the issue: Before 3074 became a thing, the transaction originator and gaspayer were one and the same. 3074 seeks to separate these identities. Some of the contracts I wrote in 2015 used tx.origin under the assumption of the former (transaction originator... it is, after all called "tx.origin" amirite?) and depend on this logic. If 3074 enables tx.origin/gaspayer to be someone other than the account originating the action, that breaks my contracts. 

I'm open to two solutions:
1. Make tx.origin the account from which the action originates. To me, this is logical anyway. (maybe a new tx.gaspayer opcode is created to cement the distinction)
2. If I'm the only party affected by this and everyone wants tx.origin to be the gaspayer, then we could do an irregular state change of /s/tx.origin/msg.sender on 2 of my 2015 contracts.

PR #1790, opened 2022-03-15
Client: Unlock miner with file pk

cbrzn:
This PR aims to add support to unlock the miner with a PK in a file in the CLI script, the reason for adding this change is that, in the Hive tests, we need to somehow unlock miner so we can produce new blocks; this is achieved by receiving the private key through the Hive environment variables


holgerd77 (approved):
LGTM

PR #1791, opened 2022-03-15
Client: Mix hash added to JsonRpcBlock

cbrzn:
This PR aims to add the `mixHash` attribute on the `JsonRpcBlock` type, the reason is that when running the tests against the hive, they query the latest block with `eth_getBlockByNumber` and then they parse that data with Geth lib, which expects mixHash to be in the object, otherwise the hash wont match, making the tests go red


ryanio:
Thanks so much! Sorry to nit, but can you put it right before nonce?

ryanio (approved):
Lgtm, thanks!

PR #1792, opened 2022-03-15
client: Fetcher reorg handling

g11tech:
This PR is follow up of #1781 and adds the followinf refinements
- [x] fetcher reorg handling
- [x] reverting the processed count on enqueue of a dequeued job which was previously not increasing leading to those many tasks not being pickedup
- [x] ~~:construction:  Still a WIP :construction:~~  , fixed a few corner cases regarding adopting tasks

### DIFF
--- a/packages/client/lib/sync/fetcher/fetcher.ts
+++ b/packages/client/lib/sync/fetcher/fetcher.ts
@@ -373,7 +373,7 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable 
             new BN(this.config.safeReorgDistance)
           ).toNumber()
           ;(jobItems[0].task as any).first = (jobItems[0].task as any).first.subn(stepBack)
-          ;(jobItems[0].task as any).count = (jobItems[0].task as any).count + stepBack
+          ;(jobItems[0].task as any).count = ((jobItems[0].task as any).count as number) + stepBack
           this.clear()
           this.failure(job, error, false)
           cb()
--- a/packages/client/lib/sync/fetcher/fetcher.ts
+++ b/packages/client/lib/sync/fetcher/fetcher.ts
@@ -8,6 +8,7 @@ import { Peer } from '../../net/peer'
 import { Config } from '../../config'
 import { Event } from '../../types'
 import { Job } from './types'
+import { JobTask as BlockFetcherJobTask } from './blockfetcherbase'
 
 export interface FetcherOptions {
   /* Common chain config*/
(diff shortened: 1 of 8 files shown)


holgerd77:
Hi Gajinder, can you please always add tags, minimally on the libraries you are touching with your work? That would be nice, thanks!

g11tech:
> Hi Gajinder, can you please always add tags, minimally on the libraries you are touching with your work? That would be nice, thanks!

Sounds good 👍. I think I would need some access as can't seem to link tags. 🙂

packages/client/lib/sync/fetcher/fetcher.ts
```diff
@@ -348,8 +368,8 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable
       encoding: string | null,
       cb: Function
     ) => {
+      const jobItems = job instanceof Array ? job : [job]
```
cbrzn, line 371: Would it makes sense to do `Array.isArray(job)` instead of `instanceof` ?

ryanio: i think they are functionally equivalent, any reason for suggestion Array.isArray over instanceof?

cbrzn: No strong reason, I just felt it's easier to read

ryanio:
going to add some coverage for the stepBack in fetcher.ts, then I think this should be ready to go :) i will approve after i add the commit(s)

ryanio:
ok I have added an extra test that should test the uncovered code path, i will check once codecov is updated.

in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1792/commits/a6ece5150743f5368ab0fa88e6ecdc5cf615e86a, i extracted JobTask since it was giving us issues trying to access job.task without casting `as any` in fetcher. Since JobTask was always the same across our different fetchers (`{ first, count }`) it made sense to me to inline it as an expected type of `{first, count}`. this also simplifies the type setup for Job. I don't expect new fetchers would have a different JobTask, and if one is introduced that does, maybe we can reconsider or re-architect then.

ryanio (approved):
lgtm! thanks for another spectacular PR :)

ryanio (changes requested):
hmm actually i'm going to put JobTask back, i had a hard time imagining a different one but i can now imagine something like a statefetcher not using a sequence of blocks like block and header fetcher. i  can try to cast as the JobTask to get over the `any` fixes, and we should think about how to keep deduping logic between header/block fetcher, like using the baseblockfetcher class

ryanio:
ok, looking a lot cleaner! i was able to accomplish my goal of improving type access to `job.task` with using a type predicate `task is BlockFetcherJobTask`

ryanio (approved):

g11tech:
> ok, looking a lot cleaner! i was able to accomplish my goal of improving type access to `job.task` with using a type predicate `task is BlockFetcherJobTask`

thanks for taking it home @ryanio (as usual :slightly_smiling_face: ) !!! :heart: :rocket:

PR #1795, opened 2022-03-16
VM: EIP-3450, EIP-3670 Improvements and Test Clean-Ups

holgerd77:
This is some follow-up work on #1719 with some improvements and clean-ups on the tests, also meant as some additional implicit review by dealing with the code.

I removed some various redundand code parts in the test files and and did some (very few) test additions and/or modifications. One major change is that I moved the whole EOF logic into its own file, it very much feels to me that this is justified since this is some major new functionality, likely also with additions and evolutions in the future. Let me know if you are ok with how this is integrated.

Otherwise: again, work looks extremely solid, thanks Andrew - also Jochem - for these great PRs! 😄

### DIFF
--- a/packages/vm/src/evm/evm.ts
+++ b/packages/vm/src/evm/evm.ts
@@ -16,7 +16,8 @@ import TxContext from './txContext'
 import Message from './message'
 import EEI from './eei'
 // eslint-disable-next-line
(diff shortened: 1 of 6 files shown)


acolytec3 (approved):
LGTM! 

I like the cleaned up version.  Way easier to read than my gibberish :-)

PR #1796, opened 2022-03-16
Client: Improve stalled Kiln startup

holgerd77:
This will redo a new bootstrap in the client along status checks to allow for connection re-pickup if the initial connection fails. Behavior is now also activated in non-discovery scenarios (as in Kiln where a single bootnode is used for connection).

This was requested in the chat since otherwise auto-starting a client is very much unreliable.


holgerd77:
I moved back to a 20 sec status check interval since this made my local integration tests unstable, so better be cautious here (might be something for a more general investigation at some PIT - not time thought right now - but this generally shouldn't be the case).

Should nevertheless help to improve the overall situation, especially in an automated context.

g11tech (approved):
LGTM! :rocket:

PR #1798, opened 2022-03-17
Client: Bootnodes fix

holgerd77:
This PR is adding additional Kiln bootnodes from [here](https://config.kiln.themerge.dev/el/bootnodes). It also fixes a bug where comma-separated bootnodes passed on CLI have not been parsed correctly.

The whole input situation - so, what get's into this `parseMultiaddrs()` in what format from where - is generally a bit chaotic (and might benefit from some small refactoring at some point), so I stayed conservative and rather a bit over-cautious and very much limited the added condition on this very specific case, since I couldn't completely oversee eventual side effects.

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -68,7 +68,7 @@ const args = yargs(hideBin(process.argv))
     array: true,
   })
   .option('bootnodes', {
-    describe: 'Network bootnodes',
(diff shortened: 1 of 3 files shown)


acolytec3 (approved):
LGTM.  Devp2p tried to connect to all 3 bootnodes at start so the parsing seems to be working correctly. 
image

PR #1799, opened 2022-03-17
Common: Fixed Event Listener Take-Over on Common.copy()

holgerd77:
On the Common `copy()` method the event listeners from the original object persist also on the copied object. This has  unwanted side effects. I discovered this in the client where the Block instantiation with the chain common [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/lib/net/protocol/ethprotocol.ts#L138) causes the `hardforkUpdated` event to still get consumed by the original listeners added to `chainCommon` when fired on the copied Common in the new Block object when the [hardfork is set](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/header.ts#L260) in the constructor.

### DIFF
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -1218,6 +1218,8 @@ export default class Common extends EventEmitter {
    * Returns a deep copy of this {@link Common} instance.
    */
   copy(): Common {
-    return Object.assign(Object.create(Object.getPrototypeOf(this)), this)
+    const copy = Object.assign(Object.create(Object.getPrototypeOf(this)), this)
+    copy.removeAllListeners()
+    return copy
   }
 }
--- a/packages/common/tests/chains.spec.ts
+++ b/packages/common/tests/chains.spec.ts
@@ -231,3 +231,18 @@ tape('[Common]: isSupportedChainId static method', function (t: tape.Test) {
     st.end()
   })
 })
+
+tape('[Common]: copy() listener tests', (t) => {
+  const common = new Common({ chain: 'mainnet' })
+  common.on('hardforkChanged', () => {
+    'its a new hardfork!'
+  })
+  common.on('hardforkChanged', () => {
+    'goodbye old hardfork'
+  })
+  const listeners = common.listenerCount('hardforkChanged')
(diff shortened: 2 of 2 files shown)


acolytec3 (approved):
LGTM!  I'm not sure precisely how to validate this but I do remember seeing the error when running the client in the past.   I tried to create a test like below:
```

tape('[Common]: listener cleanup on copy()', (t) => {
  const common = new Common({ chain: 'mainnet' })
  const listeners = common.listenerCount('hardforkChanged')
  const fakeCommon = common.copy()
  const fakeCommonListeners = fakeCommon.listenerCount('hardforkChanged')
  t.equal(listeners, fakeCommonListeners)
  t.end()
})
``` 
but it diidn't seem to detect different amounts of listeners if I removed that "removeAllListeners" call.

acolytec3:
> @acolytec3 in that code snippet, are there 0 hardforkChanged listeners? think you should add one or two pre-copy, then ensure it is zero post-copy?
> 
> would be a nice test to additionally have, thanks for writing :)

Ah, ok, that would make sense wouldn't it.  :facepalm:   

Doing it like this makes it work as expected.  
```
tape('[Common]: listener cleanup on copy()', (t) => {
  const common = new Common({ chain: 'mainnet' })
  common.on('hardforkChanged', () => {
    'its a new hardfork!'
  })
  common.on('hardforkChanged', () => {
    'goodbye old hardfork'
  })
  const listeners = common.listenerCount('hardforkChanged')
  const fakeCommon = common.copy()
  const fakeCommonListeners = fakeCommon.listenerCount('hardforkChanged')
  t.notEqual(listeners, fakeCommonListeners)
  t.end()
})
```

acolytec3:
@holgerd77 @ryanio I added the new test I wrote.  Feel free to remove it (or move to a better location within the tests folder) if it feels useless (since we're basically just verifying that basic javascript functionality works) but it will at least satisfy the code coverage complaining.

ryanio:
nice, thanks! i'll add a nit commit real quick then approve :)

PR #1800, opened 2022-03-18
Client: Merge/HF Logger Improvements

holgerd77:
Working on some client logger improvements around the Merge and HF announcements in general.


holgerd77:
First outcome is an attention API addition to the logger which allows to temporarily set some attention note. This can e.g. be used to pre-announce on a hardfork and give some more prominent notice on the process, so that we hopefully do not have to stare on the logs like the snake sitting in front of the rabbit waiting for some move 😋:

grafik

Notice sets in - very roughly - one minute before the "event" (can be modified though and we could also set lower or higher).

Have also added some dedicated Merge attention note (with TD countdown), can't test it though atm since my client gets stuck before (see chat #Merge channel note).

holgerd77:
The next logger is a consensus client connection logging as well as a clear status indicator on the connection.

This looks like this when a connection is established:

grafik

And like this if a connection is lost:

grafik

This should help to get this more explicit - also for normal users who are not running the full RPC logging.

Of course this can be improved over time, e.g. with additional information from the new `exchangeTransitionConfigurationV1` endpoint.

holgerd77:
And the last logger is some logging on the payload and forkchoice updates from the consensus client in case a connection is established, looks like this:

grafik

This should help to get a basic sense about what the Consensus Client is actually doing (or trying to do 😋) by following on the rough direction of block numbers or enabling to look up finalized block hashes in a block exporer.

holgerd77:
Have to stop for today unfortunately, would love to continue. 🙂

Seems that the client tests are not yet working, maybe/likely the intervals need to be cleared on close? If someone wants to pick this up and bring this over the finish line feel free, guess this might be useful for the current debugging and bug fixing rounds.

packages/client/lib/sync/fullsync.ts
```diff
@@ -208,12 +208,32 @@ export class FullSynchronizer extends Synchronizer {
...
+        const nextHF = this.config.chainCommon.getHardforkByBlockNumber(nextHFBlockNum)
+        attentionHF = `${nextHF} HF in ${remaining} blocks`
+      }
+    } else {
+      if (this.config.chainCommon.hardfork() === Hardfork.PreMerge) {
+        const mergeTD = this.config.chainCommon.hardforkTD(Hardfork.Merge)
+        const td = this.chain.blocks.td
+        const remaining = mergeTD!.sub(td)
+        if (remaining.lten(50000000000)) {
+          attentionHF = `Merge HF in ${remaining} TD (diff)`
```
acolytec3, line 226: If I read this code correctly, it looks like we should see this attention banner within 5MM total difficulty of TTD and I'm not seeing it yet and we're within that 5MM range (kiln having a TTD of 20,000,000)

image

packages/client/lib/blockchain/chain.ts
```diff
@@ -186,6 +186,18 @@ export class Chain {
...
+
+    this.config.chainCommon.on('hardforkChanged', async (hardfork: string) => {
+      if (hardfork !== Hardfork.Merge) {
+        const block = this.config.chainCommon.hardforkBlockBN()
+        this.config.logger.info(`New hardfork reached 🪢 ! hardfork=${hardfork} block=${block}`)
+      } else {
+        const block = await this.getLatestBlock()
+        const num = block.header.number
+        const td = await this.blockchain.getTotalDifficulty(block.hash(), num)
+        this.config.logger.info(`Merge hardfork reached 🐼 👉 👈 🐼 ! block=${num} td=${td}`)
```
jochem-brouwer, line 198: Love this message :)

packages/client/lib/sync/fullsync.ts
ryanio: wow that's weird, I tried this as a test:

```typescript
console.log(remaining.lten(20000000000000), remaining.lte(new BN(20000000000000)))
```

output: `false true`

how could that be?

ryanio: Screen Shot 2022-03-18 at 2 32 46 PM

ok cool it works now, but i think we should set it back quite a bit more

currently it is `50000000000`, ttd is `20000000000000`. should we say when it's 10% away? so once td reaches `18000000000000`. in `kiln` that would be block number `53623` (merge reached at `55127`)

ryanio (dismissed):
lgtm, awesome!!

ryanio (approved):

holgerd77:
Great improvements, thanks a lot!

PR #1801, opened 2022-03-18
Client: eth_getBlockByNumber method returns array of tx object when asked

cbrzn:
This PR aims to make the `eth_getBlockByNumber` return an array of transactions when `true` is passed as the second parameter; also, it returns `input` instead of `data` and `gas` instead of `gasLimit` as this is what is expected based on the [JSON-RPC Specs](https://github.com/ethereum/execution-apis/blob/main/src/schemas/transaction.json#L141-L176)

### DIFF
--- a/packages/client/lib/rpc/modules/eth.ts
+++ b/packages/client/lib/rpc/modules/eth.ts
@@ -72,7 +72,7 @@ type JsonRpcBlock = {
   gasLimit: string // the maximum gas allowed in this block.
   gasUsed: string // the total used gas by all transactions in this block.
   timestamp: string // the unix timestamp for when the block was collated.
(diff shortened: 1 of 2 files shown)


ryanio (approved):
nice, great, thanks!

holgerd77:
Hi Cesar, oh, I just did a rebase (with GitHub UI "Update branch" button) on this PR, didn't see yet that you did another push more or less right now.

Is this now finalized and could be merged or are you still working on the PR?

Hope I didn't cause you any inconveniences.

holgerd77 (approved):
Update: ah, just saw that you also only rebased. 🙂

Will merge, thanks a lot!

PR #1802, opened 2022-03-19
client: Handle partial results in multi-peer safe manner

g11tech:
The way we were adopting/handling jobs wasn't multi-peer safe when the results were provided less than asked for, as any shifting of jobs were not being reflected in the next job being fetched  by some other peer for e.g.
This PR fixes that, and enqueues the job again (instead of shifting/adopting all other jobs) to fetch the remaining data:
Example log:
```
client:fetcher Job index greater than processed + max queue size, skip next job execution. +0ms
  client:fetcher Requested blocks=24034-24051 from id=e9631196 address=10.244.0.164:30303 (received: 17 headers / 7 bodies) +23ms
  client:fetcher Adopt start block/header number from remaining jobs (provided=40 expected=50) +6ms
  client:fetcher Re-enqueuing job index=480 first=24041 count=10 partialResults=40 from peer id=e9631196 (reply contains unexpected data). +0ms
```

### DIFF
--- a/packages/client/lib/sync/fetcher/blockfetcher.ts
+++ b/packages/client/lib/sync/fetcher/blockfetcher.ts
@@ -26,7 +26,7 @@ export class BlockFetcher extends BlockFetcherBase<Block[], Block> {
     let { first, count } = task
     if (partialResult) {
       first = first.addn(partialResult.length)
-      count = count - partialResult.length
+      count -= partialResult.length
     }
     const blocksRange = `${first}-${first.addn(count)}`
     const peerInfo = `id=${peer?.id.slice(0, 8)} address=${peer?.address}`
--- a/packages/client/lib/sync/fetcher/blockfetcher.ts
+++ b/packages/client/lib/sync/fetcher/blockfetcher.ts
@@ -1,5 +1,5 @@
 import { Block, BlockBuffer } from '@ethereumjs/block'
-import { KECCAK256_RLP, KECCAK256_RLP_ARRAY, BN } from 'ethereumjs-util'
+import { KECCAK256_RLP, KECCAK256_RLP_ARRAY } from 'ethereumjs-util'
 import { Peer } from '../../net/peer'
 import { Job } from './types'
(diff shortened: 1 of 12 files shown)


acolytec3 (dismissed):
I'd be lying if I said I stepped through all of the deeper aspects of the code since I haven't been very involved in fetcher stuff in the past while but I can confirm this has solved the issues I observed in syncing with multiple peers.

packages/client/lib/sync/fetcher/blockfetcher.ts
```diff
@@ -22,9 +22,12 @@ export class BlockFetcher extends BlockFetcherBase<Block[], Block> {
    * @param job
    */
   async request(job: Job<JobTask, Block[], Block>): Promise<Block[]> {
-    const { task, peer } = job
-    const { first, count } = task
-
+    const { task, peer, partialResult } = job
+    let { first, count } = task
+    if (partialResult) {
```
ryanio, line 27: i think we should add some test coverage here, since it's a delicate part, and codecov is reporting missing

ryanio:
very very nice PR as always. let's add a test to cover the blockfetcher partialResult request (I also can) and get this in. I'm syncing kiln with 4 bootnodes and importing at breakneck speed 🏎🏎

ryanio (dismissed):
LGTM!

packages/client/lib/sync/fetcher/blockfetcher.ts
ryanio: thanks for adding!!

ryanio (approved):
just tested a complete sync and it went well, including over the transition, so lgtm!

g11tech:
> just tested a complete sync and it went well, including over the transition, so lgtm!

just tested, transition went  :zap: !!!

holgerd77:
Guys, if you add last-minute new VM options then please at least update on the PR (naming, labels, textual mention), this is otherwise too implicit.

Issue #528, opened 2019-05-28
VM: Message and TxContext as interfaces


s1na:
The `Message` and `TxContext` class were created before the typescript migration to define a specific set of fields that the evm interpreter should operate on. However, they are more or less only data structures and have no functionality. This can be expressed now quite well as typescript interfaces.

Something that needs to be determined is how to handle optional fields and default values.

holgerd77:
Can you do some labelling on opening issues?

s1na:
Added, thanks for the reminder

holgerd77:
Great, thanks

alcuadrado:
> Something that needs to be determined is how to handle optional fields and default values.

I think this may enough to justify upgrading an interface into a class, but it's mostly a personal preference.

On the other side, the current implementation of those classes doesn't offer much value. In particular, `Message`'s constructor receives an object that it's almost exactly a `Message`.

holgerd77:
Maybe this is worth a more detailed look as well.

holgerd77:
(eventually this might also have some performance implications?)

ryanio:
nice I wasn't aware of this issue, I like this idea.

TxContext doesn't have optional or default values so that should be able to be converted without a problem.

For Message, that's a good question. I searched a bit and one solution could be to define a defaultMessage obj with all the fallback values then use the spread operator when creating a message to fill in the missing values, e.g.: `const msg = { ...defaultMessage, msgValues }`

At the end of the day, I'm not sure if this is better or worse than doing `new Message(msgValues)` and have the class handle the fallback values. I almost prefer the simplicity of having the class do it in this case, but that's really the only thing the class is doing as OP suggests 😄

gabrocheleau:
Like @ryanio mentioned in the last comment I'd also be inclined to keep the class for usage simplicity, but I think the constructor types+default values handling could easily be improved. 

TxContext could definitely be refactored to an interface since its constructor just returns the input values.

gabrocheleau:
Resolved by #1892

PR #1810, opened 2022-03-23
vm: refactor Message and TxContext 

gabrocheleau:
Resolves https://github.com/ethereumjs/ethereumjs-monorepo/issues/528


gabrocheleau:
An issue I ran into while attempting to improve the types is the following: https://github.com/ethereumjs/ethereumjs-monorepo/blob/vm/refactor-message-tx-context/packages/vm/src/evm/evm.ts#L481-L483 
```
    if (message.selfdestruct) {
      eei._result.selfdestruct = message.selfdestruct
    }
```
`message.selfdestruct ` used to be typed as `any`, but it only ever gets assigned with the following type: `selfdestruct?: { [k: string]: boolean }` (e.g. in `runCall`), but `eei._result.selfdestruct` has type `selfdestruct: { [k: string]: Buffer }`, where the Buffer is the Address where the selfdestructed address funds were sent. 

Should be assigning an Address to the `[k:string]` keys whose value is `true`?

ryanio:
@gabrocheleau it looks like it sets it in eei _selfDestruct: https://github.com/ethereumjs/ethereumjs-monorepo/blob/019604cc95e377eaeb42c24710197088799d377b/packages/vm/src/evm/eei.ts#L415-L420

packages/vm/src/runCall.ts
```diff
@@ -35,24 +35,24 @@ export interface RunCallOpts {
...
+  const txContext: TxContext = {
+    gasPrice: opts.gasPrice ?? BigInt(0),
+    origin: opts.origin ?? opts.caller ?? Address.zero(),
+  }
 
   const message = new Message({
     caller: opts.caller,
-    gasLimit: opts.gasLimit ?? 0xffffffn,
+    gasLimit: opts.gasLimit ?? BigInt(0xffffff),
     to: opts.to ?? undefined,
```
ryanio, line 46: Don't need undefined

packages/vm/src/runCall.ts
```diff
@@ -12,7 +12,7 @@ export interface RunCallOpts {
   block?: Block
   gasPrice?: bigint
   origin?: Address
-  caller?: Address
```
ryanio, line 15: This should still be optional, can just default to zero address in `new Message`

holgerd77:
This will need some cherry-picked resubmission.

packages/vm/src/runCall.ts
gabrocheleau: Fixed!

packages/vm/src/runCall.ts
gabrocheleau: Fixed!

gabrocheleau:
Closing in favor or this clean-up PR: https://github.com/ethereumjs/ethereumjs-monorepo/pull/1819

PR #1806, opened 2022-03-22
 Merged VM: add feature to add/disable (custom) precompiles

holgerd77:
Continues #1770 (accidentally merged)

PR #1813, opened 2022-03-24
Custom precompiles

jochem-brouwer:
Cherry-picked from #1806


holgerd77:
Can this get a short analysis if a fix can be applied quickly or not (and then optimally also an update on the fix 😋, but first task would also already bring this further, slowly starting to think about the next small release round, mainly to bring EIP-1153 in on request)?

holgerd77:
Have moved this back to `WIP` (see thread above), would be great if this can be tackled soon-ish, so that this can optimally get included into eventual next-week releases.

jochem-brouwer:
This should fix it, but let's see. I am also suddenly getting local lint errors but that might be because I'm on a different version (the `develop` version?)

This can definitely get into the next release, will fix today.

packages/vm/src/evm/evm.ts
```diff
@@ -573,7 +573,7 @@ export default class EVM {
    * if no such precompile exists.
    */
   getPrecompile(address: Address): PrecompileFunc {
-    return (this._vm._precompiles as any).get(address.buf.toString('hex'))
+    return ((this._vm as any)._precompiles as any).get(address.buf.toString('hex'))
```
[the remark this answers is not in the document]
jochem-brouwer: Original error @ https://github.com/ethereumjs/ethereumjs-monorepo/pull/1813/commits/c168dd79bd7e322baef91d024f6d39cf2647f8e2

image

jochem-brouwer:
OK, this fixes the hardhat e2e tests, but in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1813/commits/c168dd79bd7e322baef91d024f6d39cf2647f8e2 it created an issue on CI on client which we cannot reproduce locally. Thoughts?

packages/vm/src/evm/evm.ts
```diff
@@ -572,7 +572,7 @@ export default class EVM {
    * Returns code for precompile at the given address, or undefined
    * if no such precompile exists.
    */
-  getPrecompile(address: Address): PrecompileFunc {
+  getPrecompile(address: Address): PrecompileFunc | undefined {
```
jochem-brouwer, line 575: I have changed this function sig, because it is incorrect anyways. Our internal logic calls this method, if `undefined` is returned we run a normal call, otherwise we run the precompile. (Also per the comment above)

jochem-brouwer:
Tests pass, ready for re-review

packages/vm/src/index.ts
```diff
@@ -276,10 +292,13 @@ export default class VM extends AsyncEventEmitter {
     }
     this._common.on('hardforkChanged', () => {
       this.getActiveOpcodes()
+      this._precompiles = getActivePrecompiles(this._common, this._customPrecompiles)
     })
 
-    // Initialize the opcode data
+    // Set list of opcodes based on HF
+    // TODO: make this EIP-friendly
```
acolytec3, line 299: What does this mean?  Is this something that needs to be resolved before this can be merged?

acolytec3 (changes requested):
Looks good except you have an open TODO so wanted to understand if we need to wrap that up before merging?

packages/vm/src/index.ts
jochem-brouwer: I accidentally copied this from `master` when I started this, it is also gone in there, so I removed it!

packages/vm/src/state/baseStateManager.ts
```diff
@@ -480,7 +485,9 @@ export abstract class BaseStateManager {
     const accessList: AccessList = []
     folded.forEach((slots, addressStr) => {
       const address = Address.fromString(`0x${addressStr}`)
-      const check1 = getActivePrecompiles(this._common).find((a) => a.equals(address))
+      const check1 = getActivePrecompiles(this._common, this._customPrecompiles).has(
+        address.buf.toString('hex')
+      )
```
holgerd77, line 490: This very much feels like the wrong abstraction to me to bake the precompiles even more into `StateManager`, I guess this was suboptimal from the beginning to have this `getActivePrecompiles` call in here.

This becomes even more apparent regarding the upcoming `StateManager` extraction.

Would it be a way to generally remove this `check1` and instead add the precompiles (e.g.) to the passed in `addressesRemoved` array? 🤔

holgerd77 (changes requested):

packages/vm/src/state/baseStateManager.ts
jochem-brouwer: Ah right, I don't know why I added this, it should not be there. I think I had in the original PR also the remark that we should probably wait for the state manager to be extracted.

jochem-brouwer:
I have blocked the PR, I wanted to say let's wait for the extracted state manager but we can't do that since we are pointed to `master` while that PR #1817 is pointed to develop. 

The reason precompiles are added is because they are used to generate an access list. You do not need to add precompiles since these are in the access list by default (they are always warm addresses). We could disable this on master but then it would break things. The alternative is to move this `getAccessList` from StateManager but then the rebase on develop would probably get a bit nasty... Thoughts? 🤔

holgerd77:
Can't we just pass in the precompiles from the outside to the method as I suggested above? 🤔

jochem-brouwer:
@holgerd77 I somehow misread that, you are right, that'll definitely work and is really clean. Will do :smile:

packages/vm/src/state/baseStateManager.ts
```diff
@@ -75,6 +76,10 @@ export abstract class BaseStateManager {
 
     this._checkpointCount = 0
 
+    if (opts.customPrecompiles) {
+      this._customPrecompiles = opts.customPrecompiles
+    }
```
holgerd77, line 81: These custom precompiles properties and options would need to be removed as well.

packages/vm/src/state/stateManager.ts
```diff
@@ -48,6 +49,10 @@ export interface DefaultStateManagerOpts {
    * A {@link SecureTrie} instance
    */
   trie?: Trie
+  /**
+   * Optional custom precompiles (to exclude from access list generation)
+   */
+  customPrecompiles?: CustomPrecompile[]
```
holgerd77, line 55: Same here.

holgerd77 (changes requested):

packages/vm/src/runTx.ts
```diff
@@ -211,6 +210,12 @@ export default async function runTx(this: VM, opts: RunTxOpts): Promise<RunTxRes
       const { tx } = opts
       // Do not include sender address in access list
       const removed = [tx.getSenderAddress()]
+      // Add the active precompiles as well
+      // Note: `_precompiles` is always updated if the hardfork of `common` changes
+      const activePrecompiles = this._precompiles
```
ryanio, line 215: can we use the public accessor you added without the underscore here? and also on L267

jochem-brouwer: Done!

PR #1815, opened 2022-03-24
vm, client: continuing v6 breaking changes

ryanio:
This PR accomplishes VM breaking change tasks listed in #1717:

- [x] Remove deprecated `state` (Trie) constructor option
- [x] Remove deprecated `generateTxReceipt` function in `runBlock` (preferring newer version located in `runTx`), remove `Receipt` re-exports
- [x] Remove `StateManager.getStateRoot()` `force=true` parameter (also in interface)
- [x] Removed deprecated `EIP2930Receipt` and `EIP1559Receipt` types
- [x] Do `gasUsed` deduplication https://github.com/ethereumjs/ethereumjs-monorepo/issues/1446#issuecomment-1013359621
  - Moved EVMResult.gasUsed to RunTxResult where it is added to there, and execResult.gasUsed remains the same
- [x] Update BaseTxReceipt gasUsed from Buffer to BigInt (suggestion added by @ryanio)
- [x] `runCode` throws if `gasLimit` is `undefined`: the `gasLimit` parameter is mandatory
- [x] Move `gasRefund` to a tx-level result object instead of `ExecResult` [todo link](https://github.com/ethereumjs/ethereumjs-monorepo/blob/0126581c6e3795d249dfa7522f3cf86e0f0f30b4/packages/vm/src/evm/evm.ts#L191-L193)
  - Moved one level up to EVMResult, since it does not seem appropriate (or used) in execResult

### DIFF
--- a/packages/client/lib/net/protocol/ethprotocol.ts
+++ b/packages/client/lib/net/protocol/ethprotocol.ts
@@ -233,7 +233,7 @@ export class EthProtocol extends Protocol {
           let encodedReceipt = rlp.encode([
             (receipt as PreByzantiumTxReceipt).stateRoot ??
               (receipt as PostByzantiumTxReceipt).status,
-            receipt.gasUsed,
+            bigIntToBuffer(receipt.gasUsed),
             receipt.bitvector,
             receipt.logs,
           ])
@@ -252,7 +252,11 @@ export class EthProtocol extends Protocol {
           // Legacy receipt if r[0] >= 0xc0, otherwise typed receipt with first byte as TransactionType
           const decoded = rlp.decode(r[0] >= 0xc0 ? r : r.slice(1)) as any
           const [stateRootOrStatus, cumulativeGasUsed, logsBloom, logs] = decoded
-          const receipt = { gasUsed: cumulativeGasUsed, bitvector: logsBloom, logs } as TxReceipt
+          const receipt = {
+            gasUsed: bufferToBigInt(cumulativeGasUsed),
+            bitvector: logsBloom,
+            logs,
+          } as TxReceipt
           if (stateRootOrStatus.length === 32) {
             ;(receipt as PreByzantiumTxReceipt).stateRoot = stateRootOrStatus
           } else {
--- a/packages/vm/src/index.ts
+++ b/packages/vm/src/index.ts
@@ -75,11 +75,6 @@ export interface VMOpts {
    */
   stateManager?: StateManager
   /**
-   * A {@link SecureTrie} instance for the state tree (ignored if stateManager is passed)
-   * @deprecated will be removed in next major version release
-   */
-  state?: Trie
-  /**
    * A {@link Blockchain} object for storing/retrieving blocks
    */
   blockchain?: Blockchain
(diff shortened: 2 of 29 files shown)


holgerd77:
Force-pushed on this with the cherry-picked relevant commits.

holgerd77:
Not sure why CI is stuck here, otherwise I would give this a go!

ryanio:
> Not sure why CI is stuck here, otherwise I would give this a go!

thanks, looks good now!

holgerd77 (dismissed):
LGTM

ryanio:
ok, noticed some last few things going over the diff again, this should be ready to merge now, thanks @holgerd77 !!!

packages/client/lib/net/protocol/ethprotocol.ts
```diff
@@ -6,7 +6,8 @@ import {
   BlockBodyBuffer,
 } from '@ethereumjs/block'
 import { TransactionFactory, TypedTransaction } from '@ethereumjs/tx'
-import { bigIntToBuffer, bufferToBigInt, bufferToInt, intToBuffer, rlp } from 'ethereumjs-util'
+import { encodeReceipt } from '@ethereumjs/vm/dist/runBlock'
```
acolytec3, line 9: Is there an easy way to move `encodeReceipt` to a package level export since we're using this outside of the `vm` package and this import looks kinda hideous, maybe in a new `util.ts` since this looks like a pure function?  This could be a follow-up as part of a more general overhaul of the vm package getting away from default exports and definitely shouldn't hold up merging this.

acolytec3 (approved):
Looks great, and probably a long slog.  I left one comment about an export that isn't directly related to the overall project of breaking changes but one we should consider.

packages/client/lib/net/protocol/ethprotocol.ts
ryanio: yeah, good eye, this isn't as scary as it looks (we access vm/dist in a few other places) but definitely a good point to ease access a bit, i think this is likely what holger was aiming with this breaking releases task:

> - [ ] Restructure/sort VM exports (directly expose Bloom, other?, unified type location, type/interface exposure, index.ts -> vm.ts+ index.ts only for exports

holgerd77:
Rebased this and fixed some small conflicts along.

packages/vm/src/index.ts
```diff
@@ -288,7 +283,7 @@ export default class VM extends AsyncEventEmitter {
     if (opts.stateManager) {
       this.stateManager = opts.stateManager
     } else {
-      const trie = opts.state ?? new Trie()
+      const trie = new Trie()
```
holgerd77, line 286: Side note, not directly related: I guess it would rather make sense, to make `trie` optional for the `StateManager` options and then do this trie instantiation there if not provided.

holgerd77 (approved):
Looks great, again, thanks a lot for all the code clean-ups! 😄

packages/vm/src/index.ts
ryanio: that makes sense, i should have checked, we already do that :) so i can just remove this, i will make a note for myself to do in a follow up develop PR

Issue #1429, opened 2021-08-26
Devp2p: DPT discovers nodes even if open slots = 0 and eats up memory


andkom:
```
Total nodes in DPT: 139, open slots: 0, queue: 3758 / 3757
```

How to stop DPT from node discovering when there are no open slots?
`_peersQueue` is constantly growing and eating up all memory

holgerd77:
Still relevant and might be a good issue if someone wants to dig a bit into the `devp2p` code. While touching core logic this is very local and not getting too deep into specific networking mechanisms.

holgerd77:
Closed by #1816

PR #1816, opened 2022-03-26
devp2p: solve "DPT discovers nodes when open_slots = 0"

ScottyPoi:
Fix for #1429 

 issue: _peersQueue is constantly growing and eating up all memory
`Total nodes in DPT: 139, open slots: 0, queue: 3758 / 3757`

fix:  
- Limit the peersQueue size to  _maxPeers * 2.  Hardcoding 50 as a limit also worked, however this is more flexible.
- Helper function _getOpenQueueSlots() that checks if the _peersQueue has reached the limit.
- Pushes to peersQueue now check for open queue slots instead of blindly adding peers


ryanio:
should the queue be an additional cap after open slots are filled? so maybe 25 slots, then 25 queue, so 50 peers max we're handling?

packages/devp2p/src/rlpx/rlpx.ts
```diff
@@ -82,8 +82,10 @@ export class RLPx extends EventEmitter {
         if (this._peersLRU.has(peer.id!.toString('hex'))) return
         this._peersLRU.set(peer.id!.toString('hex'), true)
 
-        if (this._getOpenSlots() > 0) return this._connectToPeer(peer)
-        this._peersQueue.push({ peer: peer, ts: 0 }) // save to queue
+        if (this._getOpenSlots() > 0) {
+          return this._connectToPeer(peer)
+        }
+        this._getOpenSlots() > 0 && this._peersQueue.push({ peer: peer, ts: 0 }) // save to queue
```
ryanio, line 88: i'm a little confused by this code section, so if open slots > 0, we connect to the peer, but then this line would never get executed if it's guarded with the same condition? i assume this is then meant we have no open slots then we should save them to the queue?

packages/devp2p/src/rlpx/rlpx.ts
```diff
@@ -244,14 +246,16 @@ export class RLPx extends EventEmitter {
 
       if (!disconnectWe && reason === DISCONNECT_REASONS.TOO_MANY_PEERS) {
         // hack
```
ryanio, line 248: i wonder why this says "hack"?

packages/devp2p/src/rlpx/rlpx.ts
ryanio: would also prefer if you used the same `if` style, rather than if block above then && on the next line

ScottyPoi: It doesn't make sense to me either -- but changing that line does change the outcome.  This is kind of a bandaid.  I'll dig into it more.

packages/devp2p/src/rlpx/rlpx.ts
ScottyPoi: Me too!  Wasn't my comment

ScottyPoi:
> 

After tweaking it a bit, now it seems like it stops adding peers to the queue while there are 0 open slots, but whenever there is 1 open slot, it adds a LOT of peers the queue before it gets back to 0.  So the queue still builds up to thousands pretty quickly.

If we were to cap the size of the queue at 50, would we want to stop adding new peers if the queue is at 50, or do we want to boot old peers from the queue in favor of new ones?

ScottyPoi:
Simple solution -- only add to queue is queue.length < 50

This seems to work fine without any obvious performance issues.  Now I wonder if the original issue here (not adding to the queue if open slots = 0) is still relevant?  The solution to that was just adding a missing `else` clause that probably should have been there to begin with, so I don't see any reason to undo that.

ryanio:
Well the else doesn't really do anything because the if has a return statement.

I think you can revert everything, and only push to the peerQueue if the length is less than 50, that sounds like a nice solution. Admittedly I'm not too deep in devp2p though so we can see what Holger thinks.

packages/devp2p/src/rlpx/rlpx.ts
```diff
@@ -244,14 +247,16 @@ export class RLPx extends EventEmitter {
...
-        this._peersQueue.push({
-          peer: {
-            id: peer.getId()!,
-            address: peer._socket.remoteAddress,
-            tcpPort: peer._socket.remotePort,
-          },
-          ts: (Date.now() + ms('5m')) as number,
-        })
+        if (this._getOpenSlots() > 0 && this._peersQueue.length < 50) {
```
ryanio, line 250: This shouldn't check open slots,only the peerQueue length is important?

ScottyPoi:
> Well the else doesn't really do anything because the if has a return statement, so everything after is an "else", so it isn't needed.

I agree with you, logically.  However in practice adding the else made a difference.  I don't know how to make sense of that.

ryanio:
i was thinking, instead of hardcoding 50 for the queue, can we do 2 * maxPeers? If it's helpful to add a helper function, complementary to `_getOpenSlots`, can do something like `_getOpenQueueSlots`

packages/devp2p/src/rlpx/rlpx.ts
```diff
@@ -82,8 +82,11 @@ export class RLPx extends EventEmitter {
         if (this._peersLRU.has(peer.id!.toString('hex'))) return
         this._peersLRU.set(peer.id!.toString('hex'), true)
 
-        if (this._getOpenSlots() > 0) return this._connectToPeer(peer)
-        this._peersQueue.push({ peer: peer, ts: 0 }) // save to queue
+        if (this._getOpenSlots() > 0) {
+          return this._connectToPeer(peer)
+        } else if (this._peersQueue.length < 50) {
+          this._peersQueue.push({ peer: peer, ts: 0 }) // save to queue
```
ryanio, line 88: can just be `peer` rather than `peer: peer`
```suggestion
          this._peersQueue.push({ peer, ts: 0 }) // save to queue
```

packages/devp2p/src/rlpx/rlpx.ts
holgerd77: If I recall correctly this was me adding this some time ago, I guess this was from a time where client/devp2p connections were still more tricky then today (where we fixed several bugs along the way making devp2p more stable) and I think I added this for situations where there were just no other peers available any more and so I readded existing but disconnecting peers to the queue to at least have a chance to at some point get a connection (which felt a bit hacky to me, therefore the comment).

holgerd77:
> i was thinking, instead of hardcoding 50 for the queue, can we do 2 * maxPeers

Seems like a good idea to me on first thought. Many of these limits here in the `devp2p` class are still very very basic and only roughly chosen without very deep analysis or something.

Otherwise don't wait too much on my comments on this, you guys (so: particularly Scotty and Andrew) have a lot more feeling for this networking stuff then I do through all the work you are doing on the portal network stuff.

Generally the devp2p connection management still has a lot of room for improvements I guess, e.g. a better kind of reputation building/peer selection based on former experiences with peers - so that not all "hopeless" peers are ending up in the queue again and again - might be something worth to tackle at some point, stuff like that.

If you have ideas just give them a short write-up.

ScottyPoi:
> i was thinking, instead of hardcoding 50 for the queue, can we do 2 * maxPeers? If it's helpful to add a helper function, complementary to `_getOpenSlots`, can do something like `_getOpenQueueSlots`

I have incorporated both of these suggestions :)

ScottyPoi:
> Generally the devp2p connection management still has a lot of room for improvements I guess, e.g. a better kind of reputation building/peer selection based on former experiences with peers - so that not all "hopeless" peers are ending up in the queue again and again - might be something worth to tackle at some point, stuff like that.

@holgerd77 I will do some research into this

holgerd77 (approved):
Nice, LGTM, will merge! 😄

PR #1819, opened 2022-03-27
vm: refactor Message and TxContext

gabrocheleau:
Resolves https://github.com/ethereumjs/ethereumjs-monorepo/issues/528


ryanio:
nice, looking good!

holgerd77:
Not very deep into this yet, feel free to approve and merge if you are confident.

gabrocheleau:
> Not very deep into this yet, feel free to approve and merge if you are confident.

There is one small improvement that I would still like to do concerning the selfDestruct property (see discussion in #strategy). Will get that finished at the latest this coming weekend.

packages/vm/src/evm/eei.ts
```diff
@@ -534,7 +534,10 @@ export default class EEI {
 
   async _baseCall(msg: Message): Promise<bigint> {
     const selfdestruct = { ...this._result.selfdestruct }
-    msg.selfdestruct = selfdestruct
+    msg.selfdestruct = Object.fromEntries(
+      Object.entries(selfdestruct).map(([key, value]) => [key, !!value])
+    )
```
ryanio, line 539: i don't love this `selfdestructTo` refactoring yet, it doesn't make it any easier to understand what is going on.

packages/vm/src/evm/types.ts
```diff
@@ -19,3 +20,7 @@ export type AddOpcode = {
 }
 
 export type CustomOpcode = AddOpcode | DeleteOpcode
+export interface TxContext {
```
ryanio, line 23: i think best to add a newline above so it isn't thought to be related to CustomOpcode above it. even better to have a typedoc for it.

```suggestion

/**
 * TxContext for vm execution
 */
export interface TxContext {
```

packages/vm/src/evm/eei.ts
ryanio: from my understanding we set the `selfdestruct` address to `true` as a placeholder, then overwrite it with the `to` address, is that correct? if that's the case then I think `boolean | Buffer` is fine. is it ever set to `false`? Can it be typed as `true | Buffer`? We could use `null` as a special placeholder that the buffer will overwrite but that would probably be confusing where you might think `null` means `false`.

another idea, can we have an `executed: boolean` property on the message, and when message.executed is false then selfdestruct has type boolean, and when it's true, it has type Buffer? not sure if typescript would be able to accommodate that.

ryanio: from re-reading the existing code, it's challenging to understand when `message.selfdestruct` and `_result.selfdestruct` are used. so in this `_baseCall`, we copy the selfdestruct from the `_result`, set it to the `message` which gets updated from boolean to buffer, then re-assign it back to the `_result`. that's hard to grasp/follow, and perhaps part of the problem in trying to make it more understandable.

packages/vm/src/runCode.ts
```diff
@@ -74,23 +74,24 @@ export default function runCode(this: VM, opts: RunCodeOpts): Promise<ExecResult
   const block = opts.block ?? Block.fromBlockData({}, { common: this._common })
 
   // Backwards compatibility
```
ryanio, line 76: Since it doesn't look like we pass `opts.txContext` into `runCode` anywhere (in our source or tests), I say we can remove the opt and this note

ryanio: nvm I will handle this in my own develop breaking changes PR

packages/vm/src/evm/eei.ts
gabrocheleau: Thanks for taking the time to think this through! Definitely agree with you, I don't like the solution as-is either. I've gone back and forth on this but now I think I'd just keep, as you mention, a single `selfdestruct` property with type `true | Buffer`. I don't think it's ever set to false, it's either undefined, true, or a Buffer. If we add an executed boolean prop, then I'm pretty sure that some typescript magic could help us out (basically Message would have a generic TExecuted and then the selfdestruct type would be `TExecuted ? Buffer | true` or something like that.

Will play around, hopefully we find a solution that makes the whole thing significantly easier to follow & understand.

gabrocheleau: Question concerning execution. So my base understanding was that the selfdestruct property is a Buffer when the message has been executed, and a boolean when it's not. However, it looks like in the `create` method from the `eei`, we are assigning it with the Buffer type before `executeMessage` runs, which I think would make the above `TExecuted ? Buffer | true` tricky to implement.

See [line 585](https://github.com/ethereumjs/ethereumjs-monorepo/blob/refactor/message-tx-context-clean/packages/vm/src/evm/eei.ts#L585) where `_result.selfdestruct` is a Buffer type that gets assigned to the message, and then the message is executed on [line 624](https://github.com/ethereumjs/ethereumjs-monorepo/blob/refactor/message-tx-context-clean/packages/vm/src/evm/eei.ts#L624)

ryanio: Then maybe a SelfdestructAssigned boolean?

gabrocheleau:
Closing in favor of #1862

PR #1820, opened 2022-03-28
Client: Fcu handles if headBlockHash is an old block

cbrzn:
With this PR, I aim to tackle: `Latest block after reorg` from the [Hive tests](https://hackmd.io/lKw1UArUSWu1_lOdbt3heQ?view#Eth-RPC-Status-on-ForkchoiceUpdated-Events1). Which makes sure that `forkchoiceUpdated` method is able to reorg if a previous block was sent as `headBlockHash`


ryanio (approved):
looks great, thanks!

PR #1784, opened 2022-03-13
Blockchain, VM, Client: Head function renamings

ScottyPoi:
Fixes #1754 
Execute on deprecation notes in Blockchain.

Part of #1717 

Removed get meta()

Removed getHead()
Removed setHead()

Renamed getLatestHeader() to getCanonicalHeadHeader()
Renamed getLatestBlock() to getCanonicalHeadBlock()

PR #1822, opened 2022-03-28
Blockchain, Client, VM: Head function rename rebase

acolytec3:
Rebases and replaces #1784 on current `develop`


acolytec3:
@ryanio, @holgerd77  I think this is what @ScottyPoi was talking about where expected number of tests doesn't match what's actually executed.  Any thoughts here?  I don't think the `ethereum/tests` changed since the PR was started.  Is it possible that some whole subset of tests got shortcircuited?

ryanio:
Well, there are vm tests removed? So it makes sense they would be lower. Instead of removing, is it possible to replace them with equivalent checks?

acolytec3:
I'm going to look at adding a couple of tests in the client to bump the coverage on the changed lines but this is ready for review.  @ScottyPoi , can you be sure and validate I didn't miss anything from your changes?

holgerd77:
Hi @acolytec3, sorry for that, but could you maybe give this another rebase and solve the current conflicts?

I guess then we can merge this in, thanks!

packages/blockchain/test/reorg.spec.ts
```diff
@@ -64,23 +58,6 @@ tape('reorg tests', (t) => {
...
-
-      t.ok(
-        head_lowTD.hash().equals(lowTDBlock.hash()),
-        'head on the low TD chain should equal the low TD block'
-      )
-      t.ok(
-        head_highTD.hash().equals(highTDBlock.hash()),
-        'head on the high TD chain should equal the high TD block'
-      )
-
```
holgerd77, line 83: Hi @ScottyPoi, what is the reason for this test cases to be removed? Couldn't this be replaced by the newly named function calls?

Won't make this a blocker here so that we do not have another iteration, but in case that you would think this could be kept it would be nice if you submit a small follow-up PR. Thanks! 🙂

packages/blockchain/test/reorg.spec.ts
```diff
@@ -170,19 +147,8 @@ tape('reorg tests', (t) => {
...
-      const head_high = await blockchain.getHead()
-
-      t.ok(
-        head_low.hash().equals(block2_low.hash()),
-        'head on the low chain should equal the low block'
-      )
-      t.ok(
-        head_high.hash().equals(block3_high.hash()),
-        'head on the high chain should equal the high block'
-      )
```
holgerd77, line 185: Same question here.

(or let me know if this has been settled in some chat or something and I missed)

packages/vm/tests/api/runBlockchain.spec.ts
```diff
@@ -70,11 +68,7 @@ tape('runBlockchain', (t) => {
 
     const vm = await setupVM({ common, genesisBlock })
 
-    st.equal(vm.blockchain.meta.genesis?.toString('hex'), testData.genesisBlockHeader.hash.slice(2))
-
     await vm.blockchain.putBlock(block)
-    const head = await vm.blockchain.getHead()
-    st.equal(head.hash().toString('hex'), testData.blocks[0].blockHeader.hash.slice(2))
```
holgerd77, line 77: Same here I guess?

packages/vm/tests/api/runBlockchain.spec.ts
```diff
@@ -101,11 +95,7 @@ tape('runBlockchain', (t) => {
...
```
holgerd77, line 108: Same here.

packages/vm/tests/api/runBlockchain.spec.ts
```diff
@@ -148,18 +138,12 @@ tape('runBlockchain', (t) => {
     await blockchain.putBlock(b2)
     await blockchain.putBlock(b3)
 
-    let head = await blockchain.getHead()
-    st.deepEqual(head.hash(), b3.hash(), 'block3 should be the current head')
```
holgerd77, line 152: Same here.

packages/vm/tests/api/runBlockchain.spec.ts
```diff
@@ -148,18 +138,12 @@ tape('runBlockchain', (t) => {
...
-
     try {
       await vm.runBlockchain()
       st.fail('should have returned error')
     } catch (e: any) {
       st.equal(e.message, 'test')
 
-      head = await blockchain.getHead()
-      st.deepEqual(head.hash(), b2.hash(), 'should have removed invalid block from head')
```
holgerd77, line 161: Same here.

holgerd77 (approved):
Some questions about eventual missing test cases, otherwise looks good, thanks Scotty, great work! 😄 👍 

(and thanks Andrew for another rebase on this)

Will merge this in, @emersonmacro just to let you know that this will likely affect your work on #1823 and will need some dedicated attention along a rebase (no need to imminently do so though).

packages/blockchain/test/reorg.spec.ts
ScottyPoi: I'm going back to understand why i would have done that.
May have been a blunder

holgerd77: 👍

packages/vm/tests/api/runBlockchain.spec.ts
ScottyPoi: @holgerd77 
What happened here is that one of the changes from getHead() to getIteratorHead() involved this:
```
Note that the function deviates from the old behavior and returns the genesis hash instead of the current head block if an iterator has not been run
```

While other tests could be adapted, this test is trying test whether an invalid block has been removed by checking the head before and after vm.runBlockchain().
However with the change in behavior, this will always return the genesis block before vm.runBlockchain() has been called.

so the same test passes if changed to:
```
    let head = await blockchain.getIteratorHead()
    st.deepEqual(head.hash(), genesisBlock.hash(), "...")
```

unless there is another way to `run the iterator` ?

packages/vm/tests/api/runBlockchain.spec.ts
ScottyPoi: i was able to fix this by moving the test to after vm.runBlockchain() has been called.  Otherwise function returns genesis block

packages/vm/tests/api/runBlockchain.spec.ts
ScottyPoi: This one does pass with the new name

packages/vm/tests/api/runBlockchain.spec.ts
ScottyPoi: while getHead() would have returned this block.  getIteratorHead() returns the genesis block if the iterator has not been run (with vm.runBlockchain())

Also does not work to move it after the runBlockchain() call because this is a test where runBlockchain() throws an error

packages/blockchain/test/reorg.spec.ts
ScottyPoi: @holgerd77 

both of the tests in `reorg.spec.ts` test getHead() in a way that can't directly apply to getIteratorHead().

As explained in comments below, getIteratorHead() will actually just return the genesis block if the `iterator has not been run`

holgerd77:
@ScottyPoi thanks for the explanations here, sounds all reasonable, just do a new feature PR towards `develop` which the changes you can re-integrate (or, if you are working on some other `develop` stuff, guess these very small changes can alternatively also added to a commit on some other work).

Issue #1756, opened 2022-03-01
Blockchain: Encapsulate Consensus Mechanism


holgerd77:
Ok, I thought a bit longer about this topic today and I think I now have got the pieces together, so that we can remove all consensus logic out of the blockchain class and encapsulate so that this can be easily overridden (and leads to a much cleaner core-functionality code base). 😄 

Here is the rough proposal:

We add a new folder `consensus`, following files:

```shell
consensus/ethash.ts
consensus/clique.ts
```

All - especially clique, but also ethash - functionality from `index.ts` will move into these files. These files contain a dedicated `CliqueConsensus` respectively `EthashConsensus`  class adhering to a generic `Consensus` interface which should look roughly as follows:

```typescript
export interface Consensus {
  genesisInit(genesisBlock: Block): Promise<void> 
  setup(): Promise<void>
  validate(block: Block): Promise<void>
  newBlock(bock: Block, ancientHeaders): Promise<void>
}
```

+ additionally, the constructor should take in a DB object.

There might be some tweaks necessary (changed return type or something), but this should be generally it and it should be possible to encompass both the ethash and clique (and hopefully other consensus mechanisms) into this simple interface. If no functional executions on one specific method for a consensus mechanism is needed (e.g. for `genesisInit` for `ethash`) the function body just remains empty (or we make a respective function optional? might also be a possibility).

Within `Blockchain` we can then default-instantiate one of the two consensus classes we implement ourself along a `this._common.consensusAlgorithm() === ConsensusAlgorithm.Ethash` (or: clique) switch in the constructor and can otherwise eliminate all other such switches in the main class.

On top we add a `consensus` option to the blockchain options dict (which also needs to adhere to the interface from above). If this is passed in, the consensus from this object is taken.

Some more concrete notes on the respective replacements, since this is a bit tricky and needs some close looks into the code base.

**setup(): Promise<void>**

Called after `genesisInit()`.

- `Ethash`: `new Ethash(this.db)` initialization from [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/develop-old-save-2022-01-27/packages/blockchain/src/index.ts#L275) (setup() should be called later though to unify with clique call), guess this (the `Ethash` object) can always be initialized (no overhead, just looked) and not only when `_validateConsensus` is activated, that should make things easier
[...]

PR #1823, opened 2022-03-29
Blockchain: encapsulate consensus mechanism (old)

emersonmacro:
Initial work towards #1756. Moves Ethash- and Clique-related code from the main `blockchain` file to standalone consensus classes.


emersonmacro:
@holgerd77 @ryanio This could use an initial review to see if I'm moving in the right direction. I need to fix a couple of the `blockchain` tests and update a usage in `block`, and the failed CI tests surfaced what looks like another usage in `client` that will need to be updated.

packages/blockchain/src/consensus/ethash.ts
```diff
@@ -0,0 +1,27 @@
...
+  constructor(db: LevelUp) {
+    super(db)
+  }
+
+  async setup(): Promise<void> {
+    this._ethash = new Ethash(this.db)
+  }
+
+  async validate(block: Block): Promise<void> {
+    const valid = await this._ethash!.verifyPOW(block)
```
cbrzn, line 22: Would it make sense to check if `ethash` is defined and explicitly return an error that you need to trigger setup method first?

cbrzn:
Note: Feel free to ignore my comments, I was just being curious here :P

packages/blockchain/src/consensus/ethash.ts
```diff
@@ -0,0 +1,27 @@
+// eslint-disable-next-line implicit-dependencies/no-implicit
+import type { LevelUp } from 'levelup'
+import { Block } from '@ethereumjs/block'
+import Ethash from '@ethereumjs/ethash'
+import { ConsensusBase } from './base'
+
+/**
+ * This class encapsulates Ethash-related consensus functionality when used with the Blockchain class.
+ */
+export class EthashConsensus extends ConsensusBase {
```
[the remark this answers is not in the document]
ryanio: yeah, i could perhaps see this getting confused with the Ethash class exported from our ethash package though, so maybe this is better. will have to see it in action to see.

packages/blockchain/src/consensus/ethash.ts
ryanio: yeah, well even better ethash should not be optionally undefined in this class, ideally it would be strictly defined, can you do the ethash assignment currently in `setup()` in the constructor in this case? then it should let you removed the `| undefined`

packages/blockchain/test/clique.spec.ts
```diff
@@ -348,7 +352,7 @@ tape('Clique: Initialization', (t) => {
     await addNextBlock(blockchain, blocks, A, [C, false])
     await addNextBlock(blockchain, blocks, B, [C, false])
 
-    st.deepEqual(blockchain.cliqueActiveSigners(), [A.address, B.address])
+    st.deepEqual((blockchain as any).consensus.cliqueActiveSigners(), [A.address, B.address])
```
ryanio, line 355: should we make `blockchain.consensus` public? lots of `any` casts in these tests.  and i do think we need to use some of these methods outside of blockchain, like in block and vm. we could keep the `_` prefixed methods like `_cliqueLatestSigners` private if they are not used externally (and if they are, maybe we should think about public accessors for them)

emersonmacro:
> Note: Feel free to ignore my comments, I was just being curious here :P

No worries 😄 - the more eyes on it, the better

packages/blockchain/src/consensus/base.ts
```diff
@@ -0,0 +1,72 @@
...
+  newBlock?(block: Block, commonAncestor: BlockHeader, ancientHeaders: BlockHeader[]): Promise<void>
+
+  // eslint-disable-next-line
+  new(db: LevelUp, dbManager?: DBManager, _common?: Common, _validateConsensus?: boolean): ConsensusBase
+}
+
+/**
+ * Base class that a consensus class should extend.
+ */
+export class ConsensusBase implements ConsensusBase {
```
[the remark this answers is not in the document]
emersonmacro: I tried that at first, but problem is it breaks this case here in `index.ts` https://github.com/ethereumjs/ethereumjs-monorepo/pull/1823/files#diff-db407020ca647b02014b648ca3a09fa88c5278fae3055e19ffec073c33e96678R217. After going down a stackoverflow rabbit hole, the only technique I could find that typescript would accept was this way of merging a same-named interface and class, but the class couldn't be abstract for the technique to work. That was my understanding anyway - definitely not an expert on the finer points of typescript.

packages/blockchain/src/consensus/ethash.ts
emersonmacro: Just read through Holger's comments on the issue more closely, and looks like there's no issue with moving the initialization to the constructor. Will update 👍

packages/blockchain/test/clique.spec.ts
emersonmacro: @ryanio I think this is where my understanding of typescript falls a little short. `blockchain.consensus` isn't private at the moment. But when I remove the `any` cast, typescript gives me this error in `clique.spec.ts`:

```
Property 'cliqueActiveSigners' does not exist on type 'ConsensusBase | CliqueConsensus | EthashConsensus'.
  Property 'cliqueActiveSigners' does not exist on type 'ConsensusBase'.ts(2339)
```

`cliqueActiveSigners` should exist when `blockchain.consensus` is set to `CliqueConsensus` (which should be true in this file) but not when it's set to one of the other two, and I couldn't figure out how to get it to work as I wanted. Unless I add `cliqueActiveSigners` as an optional property on the base class, but that didn't feel like a clean solution. I was googling a bit about type guards, but couldn't make sense of it in this context.

If you have any guidance or suggestions, I'm definitely open to it. I went with the `any` cast since it was already being used in the `reorg` tests (https://github.com/ethereumjs/ethereumjs-monorepo/blob/develop/packages/blockchain/test/reorg.spec.ts#L187), but must admit I did have to add a bunch of them in the `clique` tests files and it's not a very clean solution either.

packages/blockchain/src/consensus/base.ts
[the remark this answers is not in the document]
ryanio: your link isn't working for me. is the problem that they are the same name? you can just call the interface `Consensus`

packages/blockchain/test/clique.spec.ts
ryanio: in that case you can cast `(blockchain.consensus as CliqueConsensus)`, but it looks like the clique methods are still private in the class. you can keep the underscore-prefixed ones private, and make the functions public.

if you want to use type guarding, something like this should work:

```typescript
function isClique(consensus: CliqueConsensus | EthashConsensus | ConsensusBase): consensus is CliqueConsensus {
  return this._common.consensusType() === ConsensusType.ProofOfAuthority;
}

if (isClique(blockchain.consensus)) {
  // blockchain.consensus should be cast as `CliqueConsensus`, 
  // and you should be able to access the public methods now
  blockchain.consensus.cliqueActiveSigners()
}
```

I'm not sure if someone knows how to do this on the class init so `blockchain.consensus` is automatically cast as `CliqueConsensus` or `EthashConsensus` or `ConsensusBase` based on `_common.consensusType()`, that would be really cool/nice.

Another question, are we going to have a consensus class for Proof of Stake?

packages/blockchain/src/index.ts
```diff
@@ -308,6 +272,13 @@ export default class Blockchain implements BlockchainInterface {
     }
   }
 
+  /**
+   * Returns the consensus class.
+   */
+  get consensusClass() {
+    return this.consensus
+  }
```
ryanio, line 280: is this used anywhere?

holgerd77: 👍

packages/blockchain/src/consensus/base.ts
```diff
@@ -0,0 +1,72 @@
+// eslint-disable-next-line implicit-dependencies/no-implicit
+import type { LevelUp } from 'levelup'
+import { Block, BlockHeader } from '@ethereumjs/block'
+import Common from '@ethereumjs/common'
+import { DBManager } from '../db/manager'
+
+/**
+ * Base interface that a consensus class should implement.
+ */
+export interface ConsensusBase {
```
holgerd77, line 10: This should just be called `Consensus`.

packages/blockchain/src/consensus/base.ts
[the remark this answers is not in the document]
holgerd77: There shouldn't be any type of base class here, consensus implementations will just be too different, so no need to unify any code here, this will not lead to much.

Instead just do the interface and implement the interface for Clique and Ethash.

packages/blockchain/src/index.ts
```diff
@@ -267,12 +213,30 @@ export default class Blockchain implements BlockchainInterface {
...
     this.dbManager = new DBManager(this.db, this._common)
 
+    if (opts.consensus) {
+      this.consensus = new opts.consensus(
+        this.db,
+        this.dbManager,
+        this._common,
+        this._validateConsensus
+      )
+    } else if (this._common.consensusAlgorithm() === ConsensusAlgorithm.Clique) {
```
holgerd77, line 223: To bring this together with Ryans question on POS consensus: yes there should also be a consensus class `POS`, this is our third used consensus type, if I am not forgetting anything the initialization should throw in all other cases.

This would then used here instead of initializing on this "base" consensus, implementation should be pretty simple, in most cases there will just be no or very little functionality within the function calls (which is ok).

holgerd77 (changes requested):
Some very first note, looks really great in general, total relief to see all this code being removed from the `Blockchain` class! 😄 

I guess the main thing here is getting the tests back up and running as quickly as possible to not get lost on all the changes.

packages/blockchain/src/consensus/base.ts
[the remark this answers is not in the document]
emersonmacro: Removed the class

packages/blockchain/src/index.ts
emersonmacro: Removed

packages/blockchain/src/consensus/base.ts
emersonmacro: Renamed

packages/blockchain/src/index.ts
emersonmacro: I added a new class called `CasperConsensus` since the other two (`EthashConsensus` and `CliqueConsensus`) corresponded with the consensus algorithm, but let me know if you think it should be called `POSConsensus`

emersonmacro:
Made some progress but still have a couple things to sort out. The reorg tests in `blockchain` are still failing, but will probably be pretty straightforward to resolve.

The deeper issue might be what's happening in `client` (specifically in the miner). We're calling a clique method on a copy of a VM instance (https://github.com/ethereumjs/ethereumjs-monorepo/blob/develop/packages/client/lib/miner/miner.ts#L219), but it's saying `vmCopy.blockchain.consensus` is undefined, so I'm guessing I need to incorporate the new `consensus` property when `blockchain.copy()` gets called. Not sure on the exact issue yet, but will look into it more.

packages/blockchain/src/index.ts
```diff
@@ -109,17 +110,17 @@ export interface BlockchainOptions {
   genesisBlock?: Block
 
   /**
-   * Optional consensus class. If not provided, this will default to the consensus
-   * class determined by the consensus algorithm set in the {@link Common} instance.
+   * Optional consensus algorithm. If not provided, this will default to the
+   * consensus algorithm set in the {@link Common} instance.
    */
-  consensus?: ConsensusBase
+  consensus?: ConsensusAlgorithm
```
holgerd77, line 116: If you don't have a really strong reason I am not seeing right now there shouldn't be such an extra parameter and `consensus` should solely taken from the `Common` instance. If there is the need (for whatever reasons) to set on another consensus this should be done via `Common` as well - even if it might be slightly more complicated (by instantiating a default one and then set a new consensus).

packages/blockchain/src/index.ts
```diff
@@ -109,17 +110,17 @@ export interface BlockchainOptions {
...
 }
 
 /**
  * This class stores and interacts with blocks.
  */
 export default class Blockchain implements BlockchainInterface {
-  consensus: ConsensusBase | CliqueConsensus | EthashConsensus
+  consensus: Consensus | CasperConsensus | CliqueConsensus | EthashConsensus
```
holgerd77, line 123: This here should only be typed with the main `Consensus` interface, even if there are some additional methods. This is one of the goals of this refactoring to keep this as unified as possible. If there *remains* some need to access e.g. `CasperConsensus` methods, there should be some dedicated cast on the call location itself.

packages/blockchain/src/index.ts
```diff
@@ -213,21 +214,26 @@ export default class Blockchain implements BlockchainInterface {
...
         this._validateConsensus
       )
-    } else if (this._common.consensusAlgorithm() === ConsensusAlgorithm.Ethash) {
+    } else if (
+      opts.consensus === ConsensusAlgorithm.Ethash ||
+      (!opts.consensus && this._common.consensusAlgorithm() === ConsensusAlgorithm.Ethash)
+    ) {
       this.consensus = new EthashConsensus(this.db)
     } else {
       throw new Error('Unable to initialize consensus class')
```
holgerd77, line 239: Can you add a more explicit error here, "...consensus algorithm ${algorithm} not supported"?

packages/blockchain/src/consensus/clique.ts
```diff
@@ -31,7 +32,13 @@ export const CLIQUE_NONCE_DROP = Buffer.alloc(8)
 /**
  * This class encapsulates Clique-related consensus functionality when used with the Blockchain class.
  */
-export class CliqueConsensus extends ConsensusBase {
+export class CliqueConsensus implements Consensus {
+  blockchain: Blockchain
+  db: LevelUp
+  dbManager?: DBManager
+  _common?: Common
+  _validateConsensus?: boolean
```
holgerd77, line 40: This `_validateConsensus` should not be passed over here but just "count" for the outer `Blockchain` class, so this guard should only take place from `Blockchain`.

holgerd77: We should generally see that we also align the constructors of the different consensus classes as much as we can, optimally completely. `blockchain`, `db` and `common` can be passed in on all I guess (even if not used on some). Maybe rather also transfer to taking in an options dict as we do on other libraries (together with a respective `ConsensusOptions` interface), then we can easier expand here later on.

The `dbManager` shouldn't passed in here at all but the functionality should just be extracted (to some minor extend: replicated, but in a simpler form), see this passage from the task description:

Special note on the Clique DB functionality

> There is some clique specific DB functionality in db/manager.ts. This let me scratch my head a bit at first but on a second look: this can just be taken out from there and included into the new consensus module. The DB manager is generally a bit over-engineered and this is not much functionality related specifically to Clique (side note: and this is also not relevant regarding the DB cache, not much data).

Let me know if you need some help here, you can generally just simplify a lot here (on this complicated key <-> constant mapping with things like `CLIQUE_SIGNER_STATES_KEY` or the like, no need to keep) and then move the few remaining lines to an internal function or something.

holgerd77 (commented):
Some further review comments, looks really great in general, all the relevant structure already very much in place! 😄

holgerd77:
Continued in #1841, will close.

packages/blockchain/src/index.ts
emersonmacro: Removed

packages/blockchain/src/index.ts
emersonmacro: Updated in new PR

packages/blockchain/src/index.ts
emersonmacro: Updated in new PR

packages/blockchain/src/consensus/clique.ts
emersonmacro: I switched the various consensus classes to take in just the `blockchain` instance in an options dict. That way, each consensus class has full access to the `blockchain` instance and all its various components (`db`, `common`, etc.). Seems pretty clean.

I had removed the clique-related code from `db/manager.ts`, but didn't realize there was clique-related stuff in `operations` and `constants`. I removed that in the new PR and simplified the db approach in `consensus/clique`, so it's not using `dbManager` anymore.

PR #1824, opened 2022-03-29
Client: NewPayload handle already existing payload

cbrzn:
This PR aims to fix the [re-execute payload test](https://hackmd.io/lKw1UArUSWu1_lOdbt3heQ?view#Payload-Execution1) from the hive. Also, a new file has been created for mock block data, so it can be shared between the different files of engine methods. I thought this would make the tests cleaner :-)


ryanio (approved):
looks great, thank you!

PR #1825, opened 2022-03-29
util: move misplaced functions

acolytec3:
Addresses the below item from #1717

- [x] Move misplaced bigIntToHex() and bigIntToUnpaddedBuffer() methods (and tests) from types to the bytes module
- [ ] Unify const stripZeros = function (a: any): Buffer | number[] | string and bigIntToUnpaddedBuffer() to const stripZeros = function (a: any): BufferLike (which also encompassed BN)? (from @holgerd77, thought about this while reading over https://github.com/ethereumjs/ethereumjs-util/issues/141 ) (Update @holgerd77: removed from TODO list)

On the second item, in looking at the code, I'm not really sure it makes sense to merge `stripZeroes` and `bigIntToUnpaddedBuffer` as they do different things.  To my way of thinking, it makes much more sense to merge:
- `stripZeros`
- `unpadBuffer`
- `unpadArray`
- `unpadHexString`

@holgerd77 Any thoughts here since it looks like the second item was your recommendation?

### DIFF
--- a/packages/util/src/account.ts
+++ b/packages/util/src/account.ts
@@ -7,7 +7,7 @@ import { keccak, keccak256, keccakFromString, rlphash } from './hash'
 import { assertIsString, assertIsHexString, assertIsBuffer } from './helpers'
 import { BigIntLike, BufferLike } from './types'
 
-const _0tttttttttttttn = BigInt(0)
+const _0n = BigInt(0)
 
 export interface AccountData {
   nonce?: BigIntLike
--- a/packages/util/src/account.ts
+++ b/packages/util/src/account.ts
@@ -2,12 +2,12 @@ import { rlp } from './externals'
 import { Point, utils } from 'ethereum-cryptography/secp256k1'
 import { stripHexPrefix } from './internal'
 import { KECCAK256_RLP, KECCAK256_NULL } from './constants'
-import { zeros, bufferToHex, toBuffer, bufferToBigInt } from './bytes'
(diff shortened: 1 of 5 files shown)


acolytec3:
One note, I also modified `unpadHexString` to actually return a hex prefixed string.  I guess this has been otherwise forever but it doesn't make sense to me that we would not return a hex prefixed string if the sole point of the function is to strip leading zeros.  This can be reverted if the team disagrees.

holgerd77:
> One note, I also modified `unpadHexString` to actually return a hex prefixed string. I guess this has been otherwise forever but it doesn't make sense to me that we would not return a hex prefixed string if the sole point of the function is to strip leading zeros. This can be reverted if the team disagrees.

Yes, I think that makes very much sense, actually thought so too some time ago.

First TODO item implementation also looks good.

holgerd77:
> On the second item, in looking at the code, I'm not really sure it makes sense to merge stripZeroes and bigIntToUnpaddedBuffer as they do different things. To my way of thinking, it makes much more sense to merge

TBH, I can't really follow up which arguments I had here, especially since `stripZeroes` is actually an *internal*  function, hmm. 🤔

I wonder if we want to do any further consolidation here, will cautiously remove this TODO item from the v6 list. We might give things like that some additional thought once we do a final clean-up on the the Util library (which is still needed with all the BigInt changes and stuff, we need to have some look here with a broader perspective and see if things are still consistent).

holgerd77:
Have rebased this, also added a note on the second TODO item on the initial comment. Would merge in this current state if tests pass.

holgerd77 (approved):
LGTM

PR #1827, opened 2022-03-30
client/engine: speed and resiliency improvements

ryanio:
This PR:

1. client/engine: add ability to add blocks to blockchain without setting the head
   - this allows us to do `runBlock` work in `newPayload` once, and not again in `fcU`
   - also removes the need for `vm.copy()`
   - result: much faster engine calls! and we also don't log on every individual block executed
1. chain: safer closing to not cause db corruption during shutdown while handling engine requests
   - chain had the behavior that it would open if called when closed, and i believe this was leading to our chain db race conditions on shutdown.
1. CLConnectionManager: starts on first `updateStatus` in case `preMerge` isn't reached before `merge`
   - this happened on goerli shadow fork. hive will be adding a test for `merge` happening before `mergeForkBlock` to ensure all clients can handle properly. There should be no expectation that `preMerge` will happen before `mergeForkBlock`, it should only be used for `forkid` change. In this PR I've renamed`PreMerge` to `MergeForkBlock` to reduce confusion if it comes after.
1. parse: some improvements and reorganization

---

EthereumJS on top, Lodestar on bottom: 

Screen Shot 2022-03-29 at 6 11 25 PM


ryanio:
Simplified and aligned log output with most recent commit (https://github.com/ethereumjs/ethereumjs-monorepo/pull/1827/commits/43a143b6dbc570ffbd6b31d3c2b23e4ae2b08ee2): 

Screen Shot 2022-03-29 at 6 47 17 PM

packages/client/lib/rpc/modules/engine.ts
```diff
@@ -363,11 +330,7 @@ export class Engine {
       } catch (error) {
         const validationError = `Invalid tx at index ${index}: ${error}`
         this.config.logger.error(validationError)
-        const latestValidHash = await validHash(
-          toBuffer(payloadData.parentHash),
-          this.validBlocks,
-          this.chain
-        )
+        const latestValidHash = await validHash(toBuffer(payloadData.parentHash), this.chain)
```
acolytec3, line 333: Is `payloadData.parentHash` ever anything but a hex string?  If not, I wonder if there's an opportunity to add an explicit `hexToBuffer` utility method in `ethereumjs-util` that we could leverage here?  I'm remembering past conversations we've had about when to use `toBuffer` versus a more explicit utility method.

packages/client/lib/rpc/modules/engine.ts
```diff
@@ -397,38 +360,24 @@ export class Engine {
...
     const vmHead = this.chain.headers.latest!
     let blocks: Block[]
     try {
-      blocks = await recursivelyFindParents(
-        vmHead.hash(),
-        block.header.parentHash,
-        this.validBlocks,
-        this.chain
-      )
+      blocks = await recursivelyFindParents(vmHead.hash(), block.header.parentHash, this.chain)
```
acolytec3, line 380: This is just a nit but can we set the type of `block` as `let block : Block` up above at line 350 so Typescript knows what it is?

packages/client/lib/rpc/util/CLConnectionManager.ts
```diff
@@ -113,44 +126,54 @@ export class CLConnectionManager {
...
+
   private timeDiffStr(block: Block) {
     const timeDiff = new Date().getTime() / 1000 - block.header.timestamp.toNumber()
     const min = 60
     const hour = min * 60
     const day = hour * 24
     let timeDiffStr = ''
     if (timeDiff > day) {
-      timeDiffStr = `${Math.floor(timeDiff / day)} days`
+      timeDiffStr = `${Math.floor(timeDiff / day)} day`
```
acolytec3, line 167: Who needs `moment` when we can do it ourselves?

packages/client/lib/util/parse.ts
```diff
@@ -234,54 +253,39 @@ async function parseGethParams(json: any) {
     [Hardfork.MuirGlacier]: 'muirGlacierBlock',
     [Hardfork.Berlin]: 'berlinBlock',
     [Hardfork.London]: 'londonBlock',
-    [Hardfork.PreMerge]: 'mergeForkBlock',
+    [Hardfork.MergeForkBlock]: 'mergeForkBlock',
   }
   params.hardforks = Object.values(Hardfork)
     .map((name) => ({
```
acolytec3, line 259: Can we define the type here as `name: Hardfork` to make Typescript happy?

packages/client/test/miner/pendingBlock.spec.ts
```diff
@@ -140,13 +146,16 @@ tape('[PendingBlock]', async (t) => {
     const parentBlock = await vm.blockchain.getLatestBlock()
     const payloadId = await pendingBlock.start(vm, parentBlock)
     t.equal(pendingBlock.pendingPayloads.length, 1, 'should set the pending payload')
-    const block = await pendingBlock.build(payloadId)
-    t.ok(block?.header.number.eqn(1), 'should have built block number 1')
+    const built = await pendingBlock.build(payloadId)
+    if (!built) return t.fail('pendingBlock did not return')
+    const [block, receipts] = built
+    t.ok(block.header.number.eqn(1), 'should have built block number 1')
```
acolytec3, line 152: Can we change this to `t.equals(block.header.number.toNumber(), 1, 'should have built block number 1')` so it'll be easier to map over to develop?

acolytec3 (changes requested):
Love the logger changes!  A few niggly things I'd like your feedback on before we merge though.

packages/client/lib/rpc/modules/engine.ts
ryanio: I think using toBuffer is fine for hex string to buffer conversion, don't really see the benefit of using something specifically for 0x-prefixed hex string to buffer conversion when toBuffer is meant to cover that case.

packages/client/lib/util/parse.ts
ryanio: What do you mean to make typescript happy?

packages/client/test/miner/pendingBlock.spec.ts
ryanio: This line is existing so it's already mapped and changed on develop, if I change it here then there will be a merge error for the line on develop.

packages/client/lib/rpc/modules/engine.ts
acolytec3: At least in VS Code, intellisense was calling `block` an `any`, though not obvious to me why.

ryanio: have added the type

packages/client/lib/util/parse.ts
ryanio: this is fine on my side, not seeing any issues or errors

acolytec3 (approved):
LGTM!

ryanio:
Thanks @acolytec3 for the review and additional coverage! Will merge in for now so those testing the merge can get the immediate benefits. If anyone has any follow up comments please leave them here or message me and we can address in a new PR.

packages/client/lib/rpc/util/CLConnectionManager.ts
holgerd77: I hated this but I didn't want to add an extra dependency. 🙂 (or should we do in this case? This *would* likely be somewhat handy for some several use cases.)

ryanio:
`validBlocks` map was storing blocks marked valid in newPayload to be later processed by `fcU`. Now, inspired by geth, we have these separate `runBlockWithoutSetHead` and `setHead` methods, if the block is valid in `newPayload` we store it in the blockchain and then update the head in the `fcU` call.

packages/client/lib/rpc/util/CLConnectionManager.ts
ryanio: I kind of like it and its simplicity! if we want to use it elsewhere, we can move it to a util file. much better than a whole new dependency for these ~15 lines :)

PR #1828, opened 2022-03-30
Client: remove leading zeros on cumulative gas in eth_getTransactionsReceipt method

cbrzn:
This PR aims to fix the test case [Suggested fee recipient in payload creation](https://hackmd.io/lKw1UArUSWu1_lOdbt3heQ?both#Suggested-Fee-Recipient-in-Payload-creation1). 

We were getting a weird error when the method `eth_getTransactionReceipt` was being called, what was happening is that in one of the transactions, `cumulativeGas` value was being returned as `0x014820` and geth parser was throwing an error of leading zeros in this value.

When trying to debug this, I found out that indeed the leading zeros were not being removed from the `cumulativeGas` attribute; so my question was: How can this happen?

I went ahead and tried to reproduce what the test does, which basically inserts 20 transactions with 21000 gas (eth transfer tx). When it came to the 4th transaction, the `cumulativeGas` is `84000` (represented as BN) - Based on this [line](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/runTx.ts#L566), If we try to do:
```typescript
const num = new BN(84000)
const bufferNum = num.toArrayLike(Buffer)
```
`bufferNum` will be equal to `<Buffer 01 48 20>` and then, when trying to fetch all the receipts and return them, the [function `bufferToHex`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/src/bytes.ts#L212-L215) is being called [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/lib/rpc/modules/eth.ts#L253) - This will do `bufferNum.toString('hex')`; returning `014820`. 

I tried to do `bufferToHex(stripZeros(cumulativeGas))` but it is not working. I went deeper and noticed that, in the [first line](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/src/bytes.ts#L97) of `stripZeros` function, we are checking if the first element is zero, but when doing `bufferNum[0]`, one was being received instead of zero; so I came with this solution. Please let me know if you think this is not the correct approach!

### DIFF
--- a/packages/client/lib/rpc/modules/eth.ts
+++ b/packages/client/lib/rpc/modules/eth.ts
@@ -18,7 +18,7 @@ import {
   rlp,
   toBuffer,
   setLengthLeft,
-  stripZeros,
+  unpadHexString,
 } from 'ethereumjs-util'
 import { middleware, validators } from '../validation'
 import { INTERNAL_ERROR, INVALID_PARAMS, PARSE_ERROR } from '../error-code'
@@ -245,10 +245,7 @@ const jsonRpcReceipt = async (
   logIndex: number,
   contractAddress?: Address
 ): Promise<JsonRpcReceipt> => {
-  let cumulativeGas = receipt.gasUsed
-  if (cumulativeGas.toString('hex').startsWith('0')) {
-    cumulativeGas = Buffer.from(stripZeros(receipt.gasUsed.toString('hex')))
-  }
+  const cumulativeGas = '0x'.concat(receipt.gasUsed.toString('hex'))
   return {
     transactionHash: bufferToHex(tx.hash()),
     transactionIndex: intToHex(txIndex),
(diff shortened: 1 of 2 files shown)


acolytec3:
One question here is could you try running this test against the develop branch where we're using `bigints`?  Part of me wonders if this leading zero problem wouldn't go away were we using bigints instead of BNs in this particular line?

acolytec3:
> > One question here is could you try running this test against the develop branch where we're using `bigints`? Part of me wonders if this leading zero problem wouldn't go away were we using bigints instead of BNs in this particular line?
> 
> This make total sense
> 
> edit/update: @acolytec3 same error is happening in develop :-(

Ugh, my bad, I mislead you on that one.  That `receipt.gasUsed` is already a buffer.  I was thinking it was a `BN`.  

I think what you should do here is just go ahead and convert the `receipt.gasUsed` to a hex string.  Then, in the return statement, use `cumulativeGasUsed: '0x' + unpadHexString(cumulativeGas)`.

You're doing too many buffer to string conversions in all those checks.

ryanio:
have added here, thanks @cbrzn! https://github.com/ethereumjs/ethereumjs-monorepo/pull/1827/commits/ded23ff75993665c0dd28848b871e0def6971dae

PR #1830, opened 2022-04-01
client: fix eth_call, add remoteBlocks to engine

ryanio:
This PR:

1. Fixes `eth_call` error when prysm calls deposit contract with `data: '0x621fd130'` (`get_deposit_count()`). The simulated tx was running into some `gasPrice` issues with `baseFeePerGas`, so I updated the code to use `vm.runCall` which should relax the tx requirements.
2. engine: Adds `remoteBlocks` to handle storing blocks with unknown parent in case `fcU` sets head to it when parent is later given to `newPayload`. Also adds a test for local coverage of this scenario.
3. More strictly follow spec when returning `SYNCING` ensure we return `latestValidHash: null`

### DIFF
--- a/packages/client/lib/rpc/modules/engine.ts
+++ b/packages/client/lib/rpc/modules/engine.ts
@@ -349,6 +349,13 @@ export class Engine {
     const [payload] = params
     const { parentHash, blockHash } = payload
 
+    const block = await assembleBlock(payload, this.chain)
+    if (!(block instanceof Block)) {
+      // Return error response
+      this.connectionManager.lastNewPayload({ payload, response: block })
+      return block
+    }
+
     const blockExists = await validHash(toBuffer(blockHash), this.chain)
     if (blockExists) {
       const response = {
@@ -358,13 +365,6 @@ export class Engine {
       }
       this.connectionManager.lastNewPayload({ payload: params[0], response })
       return response
-    }
-
-    const block = await assembleBlock(payload, this.chain)
-    if (!(block instanceof Block)) {
-      // Return error response
-      this.connectionManager.lastNewPayload({ payload, response: block })
-      return block
     }
 
     try {
--- a/packages/client/lib/rpc/modules/engine.ts
+++ b/packages/client/lib/rpc/modules/engine.ts
@@ -173,7 +173,7 @@ const validateTerminalBlock = async (block: Block, chain: Chain): Promise<boolea
 const assembleBlock = async (
   payload: ExecutionPayloadV1,
   chain: Chain
-): Promise<Block | PayloadStatusV1> => {
+): Promise<{ block?: Block; error?: PayloadStatusV1 }> => {
   const {
     blockNumber: number,
     receiptsRoot: receiptTrie,
@@ -193,7 +193,8 @@ const assembleBlock = async (
       const validationError = `Invalid tx at index ${index}: ${error}`
       config.logger.error(validationError)
       const latestValidHash = await validHash(toBuffer(payload.parentHash), chain)
-      return { status: Status.INVALID, latestValidHash, validationError }
+      const response = { status: Status.INVALID, latestValidHash, validationError }
+      return { error: response }
     }
   }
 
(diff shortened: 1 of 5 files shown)


packages/client/lib/rpc/modules/engine.ts
```diff
@@ -277,17 +346,15 @@ export class Engine {
...
-    } = payloadData
-    const { chainCommon: common } = this.config
+    const [payload] = params
+    const { parentHash, blockHash } = payload
+
+    const block = await assembleBlock(payload, this.chain)
+    if (!(block instanceof Block)) {
+      // Return error response
+      this.connectionManager.lastNewPayload({ payload, response: block })
+      return block
```
cbrzn, line 356: What do you think about making the assemble block to return
```typescript
{
  block: Block,
  error: PayloadStatusV1
}
```
and then we can do something like
```typescript
const { block, error: assembleBlockError } = await assembleBlock(payload, this.chain)
if (assembleBlockError) {
  // Return error response
  ...
}
```
The reason for this is that since we are not returning a block but instead, an error object; (IMO) this would make it easier to read?

cbrzn (approved):
This looks amazing ser :-)

packages/client/lib/rpc/modules/eth.ts
```diff
@@ -459,23 +461,25 @@ export class Eth {
     const vm = this._vm.copy()
```
jochem-brouwer, line 461: If we copy the VM we still save the `runCall` stuff to disk right? (We discussed this yesterday on the call?). If VM originally wrote things to disk then it would also do this time right?

packages/client/lib/rpc/modules/engine.ts
ryanio: Ah good idea, I will update that :)

packages/client/lib/rpc/modules/eth.ts
ryanio: It copies the StateManager, which copies the trie, so it shouldn't write it to the original trie.

packages/client/lib/rpc/modules/engine.ts
ryanio: how's this https://github.com/ethereumjs/ethereumjs-monorepo/pull/1830/commits/6355cc72fe90e5d9b3438b09c3cf72e624b6a77d

holgerd77:
Updated this on site by rebase.

packages/client/lib/rpc/modules/engine.ts
```diff
@@ -277,17 +349,21 @@ export class Engine {
...
-      blockHash,
-    } = payloadData
-    const { chainCommon: common } = this.config
+    const [payload] = params
+    const { parentHash, blockHash } = payload
+
+    const { block, error } = await assembleBlock(payload, this.chain)
+    if (!block || error) {
+      let response = error
+      if (!response) {
```
cbrzn, line 358: If I understand this correctly, the function `assembleBlock` will return **or** a block **or** an error - Probably I am wrong but I feel that L359-L362 is unreachable?

I feel that I am most likely wrong here tho

holgerd77: "will return or a block or an error"

Is there a word or something missing between "return" and "or"? 🙂

cbrzn: @holgerd77 negative! what I meant is that it wont return both, but only one of them - does that clarify things?

holgerd77: Ah, that would be "either a block or an error".

Yes, thanks, that clarifies. 😄

ryanio: I am returning an object with optional properties so I needed this additional !block check otherwise it thinks it might be undefined.

Let me know if there is another way to do it, the OR case was my previous implementation but you said it was confusing to assign the error result to "block"

cbrzn (approved):

PR #1818, opened 2022-03-27
vm: refactor Message and TxContext

gabrocheleau:
Resolves https://github.com/ethereumjs/ethereumjs-monorepo/issues/528

PR #1831, opened 2022-04-01
EVM/VM Refactor

holgerd77:
This is the first round on an EVM/VM refactoring with the goal to get a better separtion between both (so: the pure inner bytecode processing engine (EVM) and the outer execution context). Optimally this should lead to extract the inner EVM into its own package - with all the heavy dependencies like Blockchain/Block/Tx remain in the outer package - and therefore prepare the ground in the future for things like:

- Custom L2 EVM implementations respectively easier EVM customization in general
- Swapping out the EVM for more performant solutions (EVMONE e.g.)
- Own WASM experiments on the inner core
- ...

Atm it is necessary to add some code redundancy to achieve a clean transition and not to break too much test runs, there are so many twisted interdependencies which are extremely difficult to tear appart in a clean way. I nevertheless hope that it already gets visible on the surface that this should lead to a much cleaner code base with a better separation of concerns.

There needs to be some further clean-up on the core `EVM` `executeMessage`, `executeCall`,... functions - I do not have a complete picture on this TBH (maybe @jochem-brouwer can help here?). Otherwise the next step would be to extract the EEI to the outside of the package (EVM), so that everything related to blocks, blockchain, tx can cleanly be passed into the `EVM` object and environment access is encapsulated by this EEI class. The inner `EVM` should provide the interface for the EEI so that this reliably works with - eventually - different EEI implementations. This should hopefully allow to remove all the remaining block, blockchain, tx (context) passing to the inner EVM.

Still WIP. Some tests are failing, but still on a manageable level.


holgerd77:
Procedural side note: I will likely work this through towards the state of `develop` I started with and not even try to rebase again to not get in conflict with PRs like #1811. At the end I would re-apply this commit-by-commit (and also eventually doing the test fixes along by looking at the respective commit changes).

So this is more of an applyable proof-of-concept.

Also to note, mainly for @gabrocheleau: this PR is highly intercepting with #1818. Do we maybe want to work together here? I could also use some more direct feedback on this refactor - maybe you have also some different-perspective ideas and views how things fit together (this is really structurally complex stuff on some fronts), eventually you can also do the intial PR on top of this one, so that we can bring the ideas together and make this consistent? 🤔

gabrocheleau:
> Procedural side note: I will likely work this through towards the state of `develop` I started with and not even try to rebase again to not get in conflict with PRs like #1811. At the end I would re-apply this commit-by-commit (and also eventually doing the test fixes along by looking at the respective commit changes).
> 
> So this is more of an applyable proof-of-concept.
> 
> Also to note, mainly for @gabrocheleau: this PR is highly intercepting with #1818. Do we maybe want to work together here? I could also use some more direct feedback on this refactor - maybe you have also some different-perspective ideas and views how things fit together (this is really structurally complex stuff on some fronts), eventually you can also do the intial PR on top of this one, so that we can bring the ideas together and make this consistent? thinking

Yes, agreed, let's coordinate on this. I will look at this PR this week and report back! Also makes sense to eventually rebase #1818 onto this one.

jochem-brouwer:
Gave this a look commit-per-commit, this is a great refactor and definitely makes these files less coupled (which is good!). You want me to take a look at the internals of `EVM.ts` and see if we can cleanup there as well @holgerd77 ?

holgerd77:
> Gave this a look commit-per-commit, this is a great refactor and definitely makes these files less coupled (which is good!). You want me to take a look at the internals of `EVM.ts` and see if we can cleanup there as well @holgerd77 ?

Thanks! 😄 

I would want to continue on this tomorrow morning, let me do some additional work (also on the EVM class), then I would "re-open" this for review and it would be great if you can give some additional assessment. 🙂

holgerd77:
@jochem-brouwer @gabrocheleau ok, I am really stuck a bit and would rather like to discuss with you guys.

So I would like to unify on `executeMessage(message: Message): Promise<EVMResult>` and `async runCall(opts: RunCallOpts): Promise<EVMResult>`. Both are now very much redundant and there should be only one resulting interface (in the sense of API method) which will be exposed to `EVM` users (and - once the `EEI` is passed in from the outside, also used by the `EEI`).

(this goes analogue for `async runInterpreter(message: Message, opts: InterpreterOpts = {}): Promise<ExecResult>` and `async runCode(opts: RunCodeOpts): Promise<ExecResult>`, but I would discuss this first round along the first example).

(also, just as some context: I guess Sinas initial intend along his refactoring 2-3 years ago was to just drop `runCall` and `runCode`).

So, actually the design space is very much around the question if we want to keep this `Message` class instantiation to be needed for running a message call with the respective data. `Message` highly intersects (is redundant) with the `RunCallOpts` and is also *always* routed through the `executeMessage` function. Its more or less solely function is to set default values on the various properties.

### Possibility 1

So, we could rather keep a `executeMessage(message: Message): Promise<EVMResult>` style API (I would generally be a fan of rename slighly to `runMessageCall(message: Message)` to unify on the naming a bit, also find this more clear.

The main implication fo this would be that we "burden" users with always explicitly instantiate a `Message` object before calling the method. So for e.g. this code from `customChain.spec.ts`:

```typescript
const callResult = await vm.evm.runCall({
  to: Address.fromString(contractAddress),
  data: Buffer.from(sigHash.slice(2), 'hex'),
  caller: Address.fromPrivateKey(privateKey),
})
```

would become

```typescript
const msg = new Message({
  to: Address.fromString(contractAddress),
  data: Buffer.from(sigHash.slice(2), 'hex'),
  caller: Address.fromPrivateKey(privateKey),
})

const callResult = await vm.evm.runMessageCall(msg)
```

### Possibility 2

The other way would be to drop the whole `Message` class entirely, do the default value instantiation in `executeMessage(message: Message): Promise<EVMResult>` and switch to a method rather roughly with this interface: `runMessageCall(msgCallOpts: CallOpts): Promise<EVMResult>`.

This would leave the outer API more "intact" (so all current `VM.runCall()` usages could remain the same) and also more simple to use. I initially tried on refactoring along this line, this opens up some - somewhat - challenging value setting questions along the message-call call-chain in EEI (so there, all internatl `:Message` typed function calls would also get this new `CallOpts` typing.

WDYGT? 🤔

In case we go route 2 it might be best if rather @jochem-brouwer would do that, I think I might not have enough insight into all the differentiated value selections here.

Also some additional side note: if we go route 2, a better API would rather be: `runMessageCall(msgCallData: CallData, opts: CallOpts): Promise<EVMResult>` - being the more differentiated version, with `msgCallData` 1:1 resembling everything in current `Message` and - I guess only (?) - `block` being an option from `opts`.

holgerd77:
So, as one can see, this is highly intercepting (and partially touching the very same questions) as Gabriels @gabrocheleau work on #1819.

Also, eventually @s1na still has some very long distance-opinion on this now with some extremely laid-back look. 😋

holgerd77:
Did several comment updates, please read on site. 🙂

holgerd77:
Ah, another argument here which I just thought about the first time during a walk and which might be the final direction painter: performance.

Creating a new object on every internal call (+ every code execution (!)) likely has some substantial additional costs which we can relatively easily avoid here.

jochem-brouwer:
Would creating this `MessageOpts` also not be an object which is essentially the same as the `Message` object? I am not sure if this would differ much performance wise. 

I think I am leaning towards route 2 too. The second option feels more natural. I would be somewhat okay with renaming, but we can't do something with "call" in the name, since either a call is done (to a contract or account) or a contract is being created. I think I agree with essentially "renaming" the `runInterpreter` and the `executeMessage` to `runCall` and `runCode`, that seems rather natural.  And it also does not have this burden on `Common`, it is somewhat refactoring and renaming the methods.

gabrocheleau:
I've given this some thought. I definitely agree that we should unify `executeMessage` & `runCall`. 

So taking a step back, all `runCall` currently does is assign the corresponding opts properties to `this._block` and `this._tx` (with some hardcoded fallbacks), and then create a new `Message` instance that's passed to `executeMessage`. So the first thing that comes to mind for me is: 

Why not do away with Message input entirely, and pass `RunCallOpts` to the new unified method (let's call it `runMessageCall` for this example, I don't have a clear view on naming in this context, see @jochem-brouwer's comment above). That new method logic would just be the same as `executeMessage`, with the addition of:
```
    this._block = block

    const txContext = new TxContext(
      opts.gasPrice ?? BigInt(0),
      opts.origin ?? opts.caller ?? Address.zero()
    )
    this._tx = txContext
```
And actually, we can also get rid of the TxContext constructor (as I did in #1819 by replacing it with an interface, see [this commit](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1819/commits/a6e05ee16c4f8507e2b92f1f926aca00773b46da)). So then the only other thing we need to worry about inside of `executeMessageCall` is whether or not it makes sense to still instantiate a `Message`. I would be inclined to say "yes" because of the following:
- We can move the default/fallback values logic inside the constructor, instead of hardcoding it in the method (like I also did in #1819).
- The instantiated `message` is currently passed down to many different methods, some of which mutate the `message` object, which means that just passing the options directly would require a significant change in the way we handle the 'message' downstream.

So with this solution, the API wouldn't change significantly (the same options would be passed, just to a differently named method e.g. `executeMessageCall`), but we would still get rid of a method (`runCall`) and therefore simplify the overall structure of the evm. 

Thoughts?

gabrocheleau:
Hi @holgerd77 . Would you like me to rebase my work in #1819 on top of this PR?

holgerd77:
> Hi @holgerd77 . Would you like me to rebase my work in #1819 on top of this PR?

That would generally be great. 😄 The work *here* would also need a rebase though, respectively I thought about cherry-picking the single commits one by one, best with doing test runs after each single commit and fixing the tests along as well as adopt along the lines we outlined above.

Would you also want to take this over? Then we would already have had two pair of eyes having a look at the initial code base and we have things in a clean state.

Otherwise you would likely need to wait, I won't make it to do any work on this any any more before or during Devconnect.

Third option is also that we are doing ping-pong work 😋, so you can also just start with the first commit(s) on top of latest `develop` and then do as much as you can (better slow and correct on the tests than quick and doing a lot).

holgerd77:
> Sure! I'm not sure how much time I'll be able to put in before Devconnect but we can definitely ping pong if I can't finish until then! :)

Great! 👍 😄

gabrocheleau:
Should this be closed?

Issue #1673, opened 2022-01-29
Monorepo: Analyze Bundle Sizes


holgerd77:
This is just in it's very very early stages and was inspired by [this](https://twitter.com/timdaub/status/1480880763654619139) tweet: while we have decided some time ago that we do not completely optimize for bundle sizes on releases (and e.g. also include the source along with the distribution): when looking of what our bundle sizes compose of e.g. with bundlephobia [for the VM](https://bundlephobia.com/package/@ethereumjs/vm@5.6.0) one is seeing that this has some serious tilt with just three - somewhat - side libraries very much dominating the whole package:

<img width="889" alt="grafik" src="https://user-images.githubusercontent.com/931137/151656036-49b1dcd3-be95-4624-aeca-84065485e2d1.png">

Again, very early stages and we shouldn't draw overly hastily any conclusions from this.

But we might want to at least have a look and take this into consideration among other aspects for the next round of breaking releases.

As a side note: so for Common e.g. this mainly comes (I'm quite sure) from the large mainnet genesis state we are shipping with it. Not sure if we want or can do something about this and if there are structural alternatives here.

holgerd77:
Bundle size will be dramatically reduced by #1916 (MB -> KB), will close here.

holgerd77:
Actually: will reopen and just take the above comment as an update. While the bundle size reduction in `Common` is a big step we might want to have a continued look here e.g. regarding the VM bundle size also influenced strongly by other packages as can be seen in the diagram above.

(mid-term task for after the releases as well 😋)

holgerd77:
So much bundle work lately, will close here.

PR #1833, opened 2022-04-04
Common: Support custom hash & state root on genesis

cbrzn:
This PR aims to support custom hash and state root in `Common` initialization. Currently, we are updating the [`genesis` object](https://github.com/ethereumjs/ethereumjs-monorepo/blob/common/geth-file-support/packages/common/src/index.ts#L411-L415) with the desired hash and state root, but the state given (which is the `json` attribute) is not being used at all, the reason for this is that the geth state parser is in `Client` and not in `Common` package.

I think that we should be able to support also the genesis state object (Since we are passing a state root hash, we need a state that can represent that hash - _If I understand this correctly_). After thinking a bit though this, I think that the easiest way to achieve this is to just make this `json` param to be [`GenesisState` type](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/types.ts#L51)

I am seeing this from the point of view of initiating for example the polygon chain, we can set the state with proper hashes, being able to fully execute a block

edit: I marked this as draft because it's not ready yet but would be awesome to get a review just to make sure I am going in the correct path

### DIFF
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -1032,11 +1032,9 @@ export default class Common extends EventEmitter {
    */
   genesisState(): GenesisState {
     // Custom chains with genesis state provided
-    if (this._customChains?.length && Array.isArray(this._customChains[0])) {
-      for (const chainArrayWithGenesis of this._customChains as [IChain, GenesisState][]) {
-        if (chainArrayWithGenesis[0].name === this.chainName()) {
-          return chainArrayWithGenesis[1]
-        }
+    for (const customChain of this._customChains) {
+      if (Array.isArray(customChain) && customChain[0].name === this.chainName()) {
+        return customChain[1] as GenesisState
       }
     }
 
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -9,6 +9,7 @@ import {
   Chain as IChain,
   GenesisBlock,
   GenesisState,
+  GethGenesisState,
   Hardfork as HardforkParams,
 } from './types'
 
@@ -132,7 +133,7 @@ export interface CommonOpts extends BaseOpts {
    * Initialize (in addition to the supported chains) with the selected
    * custom chains
    *
-   * Usage (directly with the respective chain intialization via the {@link CommonOpts.chain} option):
+   * Usage (directly with the respective chain initialization via the {@link CommonOpts.chain} option):
    *
    * Pattern 1 (without genesis state):
    *
@@ -167,7 +168,7 @@ export interface CommonOpts extends BaseOpts {
    * const common = new Common({ chain: 'myCustomChain1', customChains: [ [ myCustomChain1, complexState ] ]})
    * ```
    */
-  customChains?: IChain[] | [IChain, GenesisState][]
+  customChains?: IChain[] | [IChain, GenesisState | GethGenesisState][]
 }
 
 /**
@@ -203,7 +204,7 @@ export default class Common extends EventEmitter {
   private _hardfork: string | Hardfork
   private _supportedHardforks: Array<string | Hardfork> = []
   private _eips: number[] = []
-  private _customChains: IChain[] | [IChain, GenesisState][]
+  private _customChains: IChain[] | [IChain, GenesisState | GethGenesisState][]
 
   /**
    * Creates a {@link Common} object for a custom chain, based on a standard one.
@@ -401,18 +402,29 @@ export default class Common extends EventEmitter {
     if (typeof chain === 'number' || typeof chain === 'string' || BN.isBN(chain)) {
       // Filter out genesis states if passed in to customChains
       let plainCustomChains: IChain[]
-      if (
-        this._customChains &&
-        this._customChains.length > 0 &&
-        Array.isArray(this._customChains[0])
-      ) {
-        plainCustomChains = (this._customChains as [IChain, GenesisState][]).map((e) => e[0])
+      if (this._customChains?.length && Array.isArray(this._customChains[0])) {
+        plainCustomChains = (this._customChains as [IChain, GenesisState | GethGenesisState][]).map(
+          ([chain, state]) => {
+            if ('hash' && 'stateRoot' in state) {
+              return {
+                ...chain,
+                genesis: {
+                  ...chain.genesis,
+                  hash: state.hash,
+                  stateRoot: state.stateRoot,
+                },
+              } as IChain
+            }
+
+            return chain
+          }
+        )
       } else {
         plainCustomChains = this._customChains as IChain[]
       }
       this._chainParams = Common._getChainParams(chain, plainCustomChains)
     } else if (typeof chain === 'object') {
-      if (this._customChains.length > 0) {
+      if (this._customChains.length) {
         throw new Error(
           'Chain must be a string, number, or BN when initialized with customChains passed in'
         )
@@ -1019,6 +1031,15 @@ export default class Common extends EventEmitter {
    * all values are provided as hex-prefixed strings.
    */
   genesisState(): GenesisState {
+    // Custom chains with genesis state provided
+    if (this._customChains?.length && Array.isArray(this._customChains[0])) {
+      for (const chainArrayWithGenesis of this._customChains as [IChain, GenesisState][]) {
+        if (chainArrayWithGenesis[0].name === this.chainName()) {
+          return chainArrayWithGenesis[1]
+        }
+      }
+    }
+
     // Use require statements here in favor of import statements
     // to load json files on demand
     // (high memory usage by large mainnet.json genesis state file)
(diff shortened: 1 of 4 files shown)


packages/common/src/index.ts
```diff
@@ -401,18 +402,29 @@ export default class Common extends EventEmitter {
...
-      if (
-        this._customChains &&
-        this._customChains.length > 0 &&
-        Array.isArray(this._customChains[0])
-      ) {
-        plainCustomChains = (this._customChains as [IChain, GenesisState][]).map((e) => e[0])
+      if (this._customChains?.length && Array.isArray(this._customChains[0])) {
+        plainCustomChains = (this._customChains as [IChain, GenesisState | GethGenesisState][]).map(
+          ([chain, state]) => {
+            if ('hash' && 'stateRoot' in state) {
```
ryanio, line 408: i believe this needs to be
```suggestion
            if ('hash' in state && 'stateRoot' in state) {
```

packages/common/tests/customChains.spec.ts
```diff
@@ -246,4 +245,19 @@ tape('[Common]: Custom chains', function (t: tape.Test) {
 
     st.end()
   })
+
+  t.test('custom hash and state root sent in genesis state', function (st: tape.Test) {
+    const genesisState = {
+      stateRoot: 'cool-state-root',
+      hash: 'cool-hash',
```
ryanio, line 252: can you use valid values? maybe we should strictly validate them

ryanio:
wouldn't it be safer to derive the block hash and state root ourselves to ensure they equal the submitted params, rather than using them as an "override"? that would ensure everything in our stack is working to derive them correctly

packages/common/src/index.ts
ryanio: a trick i use when the list starts getting long (from parse.ts):

```typescript
if (['config', 'difficulty', 'gasLimit', 'alloc'].some((field) => !(field in json))) {
  throw new Error('Invalid format, expected geth genesis fields missing')
}
```

cbrzn:
> wouldn't it be safer to derive the block hash and state root ourselves to ensure they equal the submitted params, rather than using them as an "override"? that would ensure everything in our stack is working to derive them correctly

I agree with you, but the problem is that, if we want to generate these hashes we would need the `Trie` library, which we don't want to add to `Common` package (discussion [here](https://github.com/ethereumjs/ethereumjs-monorepo/issues/1708#issuecomment-1048776659) - The important point is Holger's answer)

> it doesn't make sense to me that common should have any "geth" logic, rather i like that the client (or somewhere, could be a common util) sets everything up for a proper common instance

That's a super important point and thanks for bringing that up I think is aligned with my thought, which is making the `json` param a `GenesisState` object. What do you think about removing the `GethGenesisState` type, and extending the `GenesisState` to support the genesis `hash` and `state root`

ryanio:
sounds good, thanks for explaining. i guess we can use this as a "trusted source" then. it might be a good idea to check in the client that these hash and stateRoot match on bootup and error and abort if not, i imagine it would help to know in advance rather than getting different values in different places.

`GenesisState` should probably stay the `alloc` portion, can we add an optional properties object to the tuple that contains these `block` fields? maybe call it GenesisBlockData. then we can also add to it in the future.

holgerd77:
@cbrzn thanks for a start on this, great! 🙂 

@ryanio the main use case for this is that people can easily (easier) use a Geth config file within our lower level libraries (so everything below the client, e.g. - guess main use case - use this for running a tx in the VM or generally creating a tx - if one only has a Geth config file at hand).

The other alternative would be to remove - along the breaking work - `stateRoot` and `hash` from our own `Common` config setup and therefore get "en match" with what Geth is providing in a chain/genesis context. This *might* even be the cleaner solution, this would have a lot of refactoring implications though if you start going along the usage chain.

So `hash()` might be the lesser problem (and thinking about it: we *might* actually want to treat this differently and see one-by-one what would be the removal-implications). `stateRoot` is used for the first time in the `Block` library for the `initWithGenesisHeader` option which is encapsuled by the `Block.genesis()` constructor which is used for the first time in the `Blockchain` library.

So a chain of refactoring would look roughly as follows:

- Remove `stateRoot` and `hash` from `Common`
- Remove `genesis` functionality from Block, move to Blockchain (which might generally a more fitting place)
- Remove genesis functionality from StateManager, move to Blockchain as well
- Remove `VM.runBlockchain()`, remove Blockchain dependency

Have no 100% proof that this would be it, but I am thinking about this topic for several weeks now and have followed this whole interdependency chain a couple of times and I think this would roughly be it which might lead us to generally leave some non-logical twists behind, also regarding e.g. some overload/misplacement on StateManager and inconsistent library placements/dependencies regarding the Client/Blockchain/VM correlation.

This is pretty heavy stuff though. 😬 😋

We could tackle nevertheless.

The solution here would be the - very - local one just extending a bit (but not too dramatically) on our own sub-optimalities.

I have no totally strong bias in the one or the other direction. We should just know what we are doing. 🙂

ryanio:
thanks for the in-depth walk through, super interesting! and a great time to consider some broader changes while working on develop. i will let this sink in before i give an opinion on a route, but i do like your bullet points, they seem like improvements in the right direction to me.

holgerd77:
As a second thing, this kind of deeper-reaching refactoring would also allow to solve on #1673 - so the heavy bundle sizes due to the bundling of the genesis state within Common - by moving genesis states to `Blockchain`, where it more naturally belong. There it would be more feasible respectively ok to have this in the bundle, while the heavy burden on all libraries below (VM, Block, Tx - so everything which is most likely to be bundled) would be lifted.

At the same time (if I am not mistaken) genesis state could still be provided on an API level for us internally on all places where it is necessary. Common would then solely take genesis in as a constructor option.

jochem-brouwer:
Very good points @holgerd77!

`Remove VM.runBlockchain(), remove Blockchain dependency`

Right, so we move `runBlockchain` into the blockchain package, and instantiate `blockchain` with VM? Then in Client we would also need to change each instance of VM by instantiating a blockchain, right?

ryanio:
Still have my sights on this, will try to pick up and continue this week. I'll open a new PR targeting develop with the changes outlined in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1833#issuecomment-1088445250

holgerd77:
> Still have my sights on this, will try to pick up and continue this week. I'll open a new PR targeting develop with the changes outlined in [#1833 (comment)](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1833#issuecomment-1088445250)

Oh, interesting (and great), totally forgot about this TBH.

holgerd77:
@ryanio you can also consider https://github.com/ethereumjs/ethereumjs-monorepo/pull/1833#issuecomment-1091181809 along if you want.

holgerd77:
Also have added an additional "In Discussion/TODO" note on the v6 Release Notes so that we do not forget about this task.

ryanio:
I've gotten started on this but have run into a few things so far:

1. common uses the genesis hash in `_calcForkHash()`, if we remove it how should we replace it there
2. we use `runBlockchain()` in the `BlockchainTestsRunner`, if we remove it how should we run those tests
    1. we can't remove the blockchain dependency from vm since the vm always has a blockchain  
4. what do you mean by removing genesis functionality from statemanager (now vmState) and moving to blockchain? vmState needs to initialize accounts in the trie, and the blockchain doesn't have access to the putAccount methods (the blockchain is a dependency of the vm). i can move `generateCanonicalGenesis()` one level up from vmState to the vm in order to access the new `blockchain.genesisState()` if we move it from `common.genesisState()`

holgerd77:
@ryanio great that you had a first look! 🙂 

All of these are first-shots, so we might want to deepen discussion on the call or something:

> 1. common uses the genesis hash in _calcForkHash(), if we remove it how should we replace it there

We could make the `forkHash` entry in the HF JSOn files mandatory and then omit calculating the fork hash in the `C-ommon` constructor if not present. for the `forkHash()` function itself we would demand the genesis hash as an input. All of this should be viable on first thought.

> 2. we use runBlockchain() in the BlockchainTestsRunner, if we remove it how should we run those tests
   - we can't remove the blockchain dependency from vm since the vm always has a blockchain

Yeah, maybe we wait on this for the VM/EVM refactor? I guess the `Blockchain` dependency remains somewhat better placed there if there is this single extracted `VMExecutionContext` package, there it does make somewhat more sense again I guess?

> what do you mean by removing genesis functionality from statemanager (now vmState) and moving to blockchain? vmState needs to initialize accounts in the trie, and the blockchain doesn't have access to the putAccount methods (the blockchain is a dependency of the vm). i can move generateCanonicalGenesis() one level up from vmState to the vm in order to access the new blockchain.genesisState() if we move it from common.genesisState()

If the genesis functionality is now in `vmState` then it might also be an option to then keep it there and then bundle all the genesis files with this outer package (so `VMExecutionContext`) instead of `Blockchain`? This is already a lot higher in the dependency hierachy and would also provide the benefit of creating a lot less hazzle for the lower level libraries (Tx, Block,...) and it generally might also be a fitting place to be.

ryanio:
Have continued in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1916 so I will close here

Issue #1718, opened 2022-02-16
Future Release Planning


holgerd77:
Replaces #1024 

Planning on library releases where there is no intended release date and that shouldn't be integrated into the next round of releases.

---

### Trie

Tasks moved to #1717 (decided to handle in current breaking release cycle)

holgerd77:
Replacing would be a massive undertaking since Buffer is used throughout our whole stack. I would doubt that it will make it into this round of breaking releases, we can talk about it for future releases though.

Generally we would need some more evidence and some deepened analysis on the benefits here, this can be in some form of continued write-up or discussion, some PoC implementations and the like.

Some questions here:

- Is this *generally* really wanted by some broader part of the community?
- Does this really improve browser compatbility or is this rather theoretical (e.g. because several other dependencies make a cut on this anyhow)?
- How is the effect on performance?
- Where does this take place/is this relevant?
- How extensive are we using Buffer features and API and what would we loose there?
- How would debuggability be affected?

For a first round. 😋

holgerd77:
(wouldn't 100% include this from this release round, that is rather some stronger tendency I have right now that I think this would overload us if we also take on the BigInt transition)

(+ again, some more evidence on the benefits is still strongly needed)

holgerd77:
This doesn't really has much content and has already trickled down a lot in the issue list.

Will eventually recreate if needed again and if we have some clearer picture on scope and time horizon on upcoming breaking release rounds. Will close for now.

PR #1834, opened 2022-04-05
develop: ci fix, some last backwards compatibility removals

ryanio:
This PR fixes an issue in develop where a karma workaround was erroring in the hardhat e2e installation for our packages. I moved the fix to the blockchain karma config.

I also executed some final breaking change items in VM and Trie.


ryanio:
I noticed a series of my commits missing from my prior PR #1815:

Screen Shot 2022-04-04 at 9 25 38 PM

Highlighted here:

Screen Shot 2022-04-04 at 9 26 15 PM

So I have cherry-picked them and added them back to this PR.

holgerd77:
Rebased this via UI with a local test rebase before.

packages/vm/src/index.ts
```diff
@@ -284,9 +283,7 @@ export default class VM extends AsyncEventEmitter {
     if (opts.stateManager) {
       this.stateManager = opts.stateManager
     } else {
-      const trie = new Trie()
```
holgerd77, line 287: 👍

packages/trie/src/baseTrie.ts
```diff
@@ -104,7 +104,7 @@ export class Trie {
    */
   async checkRoot(root: Buffer): Promise<boolean> {
     try {
-      const value = await this._lookupNode(root)
+      const value = await this.lookupNode(root)
```
holgerd77, line 107: We didn't plan to do a breaking Trie release initially, but my previously somewhat stronger opinion on this is wayning, so we might just do along (and then also finally do the renaming).

Just wanted to mention here. We then eventually want to have one additional thought round or broader look on other breaking changes.

holgerd77: So, some additional context: currently Trie is still listed in the "Future Release Planning" issue: https://github.com/ethereumjs/ethereumjs-monorepo/issues/1718. Should I move over? 🤔

packages/client/lib/service/fullethereumservice.ts
```diff
@@ -190,9 +191,18 @@ export class FullEthereumService extends EthereumService {
...
+          receipts.map((r) =>
+            rlp.encode([
+              (r as PreByzantiumTxReceipt).stateRoot ?? (r as PostByzantiumTxReceipt).status,
+              bigIntToBuffer(r.gasUsed),
+              r.bitvector,
+              r.logs,
+            ])
+          )
+        )
+        receiptsSize += Buffer.byteLength(receiptsBuffer)
```
holgerd77, line 195: Could you drop a short note what's happening here?

Does this fix some concrete issues? Can't oversee this on a first look.

holgerd77 (approved):
Nice PR 🙂, looks good! 👍

packages/trie/src/baseTrie.ts
ryanio: ah sorry didn't notice that.

yeah sure let's include these refactorings, i think now is the time to do it while we are preparing develop. not sure if it makes sense to wait any longer. we can assign some team members to the tasks and i can also dedicate some time this month to tackle them.

packages/client/lib/service/fullethereumservice.ts
ryanio: ah yeah, actually you commented on an old version of this code, if you look at the final version in this PR i made it much more simple by using our existing encodeReceipt function. this is simply keeping track of the size of the GetReceipts response, so we can respect capping at 10mib as the spec suggests.

PR #1835, opened 2022-04-05
Monorepo: Develop Branch Rebase Test 2022-04-05

holgerd77:
Rebase test of the `develop` branch, do not merge.


packages/client/lib/rpc/modules/engine.ts
```diff
@@ -513,10 +513,16 @@ export class Engine {
       await this.execution.setHead(headBlock)
       this.txPool.removeNewBlockTxs(blocks)
 
-      const timeDiff = new Date().getTime() / 1000 - headBlock.header.timestamp.toNumber()
+      // TODO: removed along rebase, 2022-04-05, @holgerd77
+      // Analyze if code should be there for some reason
+      /*for (const block of blocks) {
+        this.validBlocks.delete(block.hash().toString('hex'))
+      }*/
```
holgerd77, line 520: I wasn't 100% sure if this code part should be there or not, guess @ryanio or @g11tech will likely know.

ryanio: That can be removed, I removed validBlocks (we store them in the blockchain directly now)

g11tech: I think this is needed, because @ryanio has been maintaining the validBlocks set of the blocks which have valid execution, but not yet part of the chain, since here fcU has arrived for those blocks, they are being moved from validBlock to the main chain via putBlocks.

acolytec3:
> @acolytec3 thanks for all the fixes, fixes look good. slightly_smiling_face
> 
> For now I would assume that you are still working on this? Can you then also remove the commented out code from above?
> 
> Will have a short look into the failing test messages and drop a subsequent note if I notice what is to be done there.

Sure, there a couple of broken tests in the new EIP-1153 test suite so working to address them right now.  Will also remove that commented code.

holgerd77:
Short note: I merged in an additional PR on `master`. It is *not* strictly necessary to update (`master` will evolve anyhow), if you want you can do so by rebase towards the latest master and the fix the conflict along.

holgerd77:
Not 100% sure, but I might have accidentally reintroduced the following two test cases in `chain.spec.ts` from client and I would assume these can be just removed:

```typescript
await chain.getCanonicalHeadHeader()
t.ok(chain.opened, 'chain should open if getCanonicalHeadHeader() called')
await chain.close()
await chain.getCanonicalHeadBlock()
t.ok(chain.opened, 'chain should open if getCanonicalHeadBlock() called')
```

acolytec3:
> Not 100% sure, but I might have accidentally reintroduced the following two test cases in `chain.spec.ts` from client and I would assume these can be just removed:
> 
> ```ts
> await chain.getCanonicalHeadHeader()
> t.ok(chain.opened, 'chain should open if getCanonicalHeadHeader() called')
> await chain.close()
> await chain.getCanonicalHeadBlock()
> t.ok(chain.opened, 'chain should open if getCanonicalHeadBlock() called')
> ```

I think these are valid.  Looks like they were part of PR #1835 (the rebase of @ScottyPoi 's PR to rename those `getLatestHead`/`getLatestHeader` methods.

holgerd77:
> > Not 100% sure, but I might have accidentally reintroduced the following two test cases in `chain.spec.ts` from client and I would assume these can be just removed:
> > ```ts
> > await chain.getCanonicalHeadHeader()
> > t.ok(chain.opened, 'chain should open if getCanonicalHeadHeader() called')
> > await chain.close()
> > await chain.getCanonicalHeadBlock()
> > t.ok(chain.opened, 'chain should open if getCanonicalHeadBlock() called')
> > ```
> 
> I think these are valid. Looks like they were part of PR #1835 (the rebase of @ScottyPoi 's PR to rename those `getLatestHead`/`getLatestHeader` methods.

Hmm, not 100% sure, Ryans PR on the Blockchain unsettled Promise fix also seems to play a role here https://github.com/ethereumjs/ethereumjs-monorepo/pull/1811, where `Blockchain` is not auto-opened in the client `chain.ts` file any more. Maybe @ryanio can shed some light? 🤔

acolytec3:
> Correct I did change the chain behavior as Holger mentioned to be safer, please use the latest code as of #1827 (second list item in opening post)

This should be resolved now.

holgerd77:
Gave this another rebase and fixed 1-2 things along.

I also squashed all the fix commits into one.

@acolytec3 @ryanio short note: saw your discussion on this ("holger suggested to just directly commit"  or something). TBH I also didn't envision this out fully, original idea/imagination was, that we would likely have somewhat between 1-2 fixing commits - which would be ok. If it tends that we have substantially more it would be a pity if we bloat our otherwise really clean `develop` commit history with this.

So, not really clear. If things can be fixed easily by just doing a single commit, still be ok to directly push. But also ok to do a PR on top which we can squash-merge in. All not totally a game changer on the procedure, I can still do this squashing here at the end, just for some context.

Ok, I would admin-merge this in now in case of tests passing - guess this has been implicitly reviewed by the work on top. If you want to have a last look now it's the time.

holgerd77:
(ah, and after admin-merge then force-push on the original `develop` branch)

holgerd77:
Update: haha, of course not admin-merge - sorry if I have caused some gasps here 😋 - so just: using the branch as `develop`.

PR #1840, opened 2022-04-06
Set caller to zero address if not provided

acolytec3:
The `caller` parameter in `runCallOpts` is optional but causes the EVM to throw if not provided.  The upcoming docs on the `runCallOpts` that are in the `develop` branch indicate the message caller should default to the zero address if not provided in the message so this fix accounts for that design.


ryanio (approved):
Lgtm, thanks!

PR #1817, opened 2022-03-27
vm: extract statemanager to own package

g11tech:
Refactoring vm specific logic out from statemanager

~~PS: very very preliminary stage~~

PR #1841, opened 2022-04-07
 Blockchain: encapsulate consensus mechanism (new)

holgerd77:
Cherry-picked resubmission of #1823 after latest `develop` rebase, see #1835.


emersonmacro:
Continuing the discussion from the previous PR here https://github.com/ethereumjs/ethereumjs-monorepo/pull/1823

@holgerd77 I feel like I'm most of the way there, but I'm really struggling with the two failing `blockchain` tests (`clique.spec.ts` and `reorg.spec.ts`). I get the sense I may be missing a small piece of clique functionality somewhere and it causes the tests to fail "downstream" with different errors, but I may be wrong on that. I can try to keep digging into it, but wondering if something obvious would jump out if you (or maybe @ryanio?) have time to take a look?

I'm also not sure what's going on with failed client test (https://github.com/ethereumjs/ethereumjs-monorepo/runs/5966438065?check_suite_focus=true) and the failed codecov/patch test. Codecov seems to be saying there's new code in `devp2p` that doesn't have coverage (https://app.codecov.io/gh/ethereumjs/ethereumjs-monorepo/compare/1841/changes), but I didn't make any changes in `devp2p`...

packages/block/src/header.ts
```diff
@@ -518,13 +518,13 @@ export class BlockHeader {
       )
       throw new Error(msg)
     }
-    if ('cliqueActiveSigners' in blockchain === false) {
+    if ('cliqueActiveSigners' in (blockchain as any).consensus === false) {
       const msg = this._errorMsg(
         'PoA blockchain requires method blockchain.cliqueActiveSigners() to validate clique difficulty'
```
ryanio, line 523:
```suggestion
        'PoA blockchain requires method blockchain.consensus.cliqueActiveSigners() to validate clique difficulty'
```

packages/blockchain/src/consensus/clique.ts
```diff
@@ -0,0 +1,557 @@
...
+import { Block, BlockHeader } from '@ethereumjs/block'
+import { ConsensusAlgorithm } from '@ethereumjs/common'
+import { Address, rlp, bigIntToBuffer, bufferToBigInt } from 'ethereumjs-util'
+import Blockchain from '..'
+import { Consensus, ConsensusOptions } from './interface'
+
+const debug = createDebugLogger('blockchain:clique')
+
+// Clique Signer State: [blockNumber, signers]
+type CliqueSignerState = [bigint, Address[]]
```
ryanio, line 11: our TS version has named tuples now, so you can inline this comment and the ones below like this

```suggestion
// Clique Signer State
type CliqueSignerState = [blockNumber: bigint, signers: Address[]]
```

packages/blockchain/src/consensus/clique.ts
```diff
@@ -0,0 +1,557 @@
...
+    await this.blockchain.db.put('CliqueBlockSignersSnapshot', rlp.encode(formatted), DB_OPTS)
+  }
+
+  /**
+   * Fetches clique signers.
+   * @hidden
+   */
+  private async getCliqueLatestSignerStates(): Promise<CliqueLatestSignerStates> {
+    try {
+      const signerStates = await this.blockchain.db.get('CliqueSigners', DB_OPTS)
```
ryanio, line 451: you should use the constant DB keys here (`CLIQUE_SIGNERS_KEY`, `CLIQUE_VOTES_KEY`, `CLIQUE_BLOCK_SIGNERS_KEY`) rather than string values

packages/blockchain/src/consensus/clique.ts
```diff
@@ -0,0 +1,557 @@
...
+    ])
+    await this.blockchain.db.put('CliqueVotes', rlp.encode(formatted), DB_OPTS)
+  }
+
+  /**
+   * Returns a list with the current block signers
+   * (only clique PoA, throws otherwise)
+   */
+  cliqueActiveSigners(): Address[] {
+    this._requireClique()
```
ryanio, line 343: I'm not sure if we need this `_requireClique` check anymore since this class can only be initialized when consensus is clique

packages/blockchain/src/index.ts
```diff
@@ -271,12 +213,20 @@ export default class Blockchain implements BlockchainInterface {
     this.db = opts.db ? opts.db : level()
     this.dbManager = new DBManager(this.db, this._common)
 
+    if (this._common.consensusAlgorithm() === ConsensusAlgorithm.Casper) {
```
ryanio, line 216: maybe we can do a switch instead of if statements here? should be cleaner since we're just checking consensusAlgorithm()

packages/vm/examples/run-blockchain.ts
```diff
@@ -31,7 +31,7 @@ async function main() {
   // Note that this optimization is a bit hacky and might
   // not be working in the future though. :-)
   if (validatePow) {
-    blockchain._ethash!.cacheDB = level('./.cachedb')
+    (blockchain.consensus as any)._ethash!.cacheDB = level('./.cachedb')
```
ryanio, line 34: just a nit, can you update this example and BlockchainTestsRunner.ts to do this `as EthashConsensus`? and ethash doesn't need the `!` anymore since it is no longer optionally undefined.

packages/blockchain/src/consensus/clique.ts
```diff
@@ -0,0 +1,557 @@
...
+      }
+      throw error
+    }
+  }
+
+  /**
+   * Gets a header by number. Header must be in the canonical chain
+   * @hidden
+   */
+  private async _getCanonicalHeader(number: bigint) {
```
ryanio, line 514: why redefine this method, can just use from the blockchain class?

packages/blockchain/src/consensus/clique.ts
```diff
@@ -0,0 +1,557 @@
...
+   *   1 -> 1, 2 -> 2, 3 -> 2, 4 -> 2, 5 -> 3, ...
+   * @hidden
+   */
+  private cliqueSignerLimit() {
+    return Math.floor(this.cliqueActiveSigners().length / 2) + 1
+  }
+
+  /**
+   * Checks if signer was recently signed.
+   * Returns true if signed too recently: more than once per {@link Blockchain.cliqueSignerLimit} consecutive blocks.
```
ryanio, line 370:
```suggestion
   * Returns true if signed too recently: more than once per {@link CliqueConsensus.cliqueSignerLimit} consecutive blocks.
```

packages/block/src/header.ts
emersonmacro: Good catch - updated

packages/blockchain/src/consensus/clique.ts
emersonmacro: Ah nice, I didn't know that. Updated

packages/blockchain/src/consensus/clique.ts
emersonmacro: Holger was advocating for not using dbManager and instead just simplifying the db interactions as much as possible in the clique class, so that's why it's not currently using the DBTarget pattern. But also, agreed, it's better to define the strings once as constants. Updated

packages/blockchain/src/consensus/clique.ts
emersonmacro: Yeah good call - it's only used in this one spot. Removed

packages/blockchain/src/index.ts
```diff
@@ -1,28 +1,18 @@
...
-  CliqueLatestVotes,
-  CliqueBlockSigner,
-  CliqueLatestBlockSigners,
-  CLIQUE_NONCE_AUTH,
-  CLIQUE_NONCE_DROP,
-} from './clique'
+import { CasperConsensus } from './consensus/casper'
+import { CliqueConsensus } from './consensus/clique'
+import { Consensus } from './consensus/interface'
+import { EthashConsensus } from './consensus/ethash'
```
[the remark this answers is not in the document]
emersonmacro: good call - updated

packages/blockchain/src/index.ts
emersonmacro: updated

packages/blockchain/src/consensus/clique.ts
emersonmacro: Good catch - this was also an early change. I changed it in the blockchain class to a public method `getCanonicalHeader` and removed the duplicate methods in `consensus/clique`

emersonmacro:
> @emersonmacro I found the issue which fixes the failing tests. If you look prior to this PR, `ancientHeaderNumber` is being set earlier in the code, but in yours it is set later, which ends up returning the same header being inserted, so the auth/drop vote logic wasn't being run on the new block.
> 
> also, what was the reason for adding `ancestorHeaders`? I don't see it being used anywhere.

@ryanio Ahhh, good call. I moved it back to the earlier position and it fixed most of the tests. Thanks for the review, btw, much appreciated.

Re: `ancestorHeaders`, that came from my understanding of Holger's suggested approach in the original issue:

> This might be the most tricky one and the one which needs the most explanation. So following thoughts: This _findAncient(newHeader) function is - in the current code base - only called in a Clique context. However this is a very generic functionality to find a common ancestor of a new block and the latest canonical (?) head. So I would say that we should keep this functionality in the main Blockchain class, rename to findCommonAncestor() and along not only returning the single header found at the end but a whole array with the all the headers found on the way (so: this is a super-set of the previous result). This can then be fed to the generic newBlock(bock: Block, ancientHeaders): Promise<void> function. My guts say that this should likely be generally the important information needed in this context also for other consensus mechanisms.

But I didn't fully understand how when calling `newBlock(block: Block, ancientHeaders)`, how `newBlock` would know which item in the `ancientHeaders` array is the one common ancestor that we're currently using. So for flexibility, I changed `findCommonAncestor` return both the one identified common ancestor _and_ the full array. But yeah, the full array is not currently being used anywhere. Not sure if we should just strip it out or not.

One last question... There are still 3 individual failing tests across `clique.spec.ts` and `reorg.spec.ts`. I've looked into the first one, happening here https://github.com/ethereumjs/ethereumjs-monorepo/blob/develop/packages/blockchain/test/clique.spec.ts#L196, and it looks like the block has a normal header hash, but the parent hash is set to all zeroes, and so the test fails with a "not found in db" error. I haven't yet been able to figure out why the parent hash is zeroes when the block number is being set to 30,000 here https://github.com/ethereumjs/ethereumjs-monorepo/blob/develop/packages/blockchain/test/clique.spec.ts#L188.

ryanio:
> One last question... There are still 3 individual failing tests across `clique.spec.ts` and `reorg.spec.ts`. I've looked into the first one, happening here https://github.com/ethereumjs/ethereumjs-monorepo/blob/develop/packages/blockchain/test/clique.spec.ts#L196, and it looks like the block has a normal header hash, but the parent hash is set to all zeroes, and so the test fails with a "not found in db" error. I haven't yet been able to figure out why the parent hash is zeroes when the block number is being set to 30,000 here https://github.com/ethereumjs/ethereumjs-monorepo/blob/develop/packages/blockchain/test/clique.spec.ts#L188.

well, so that test is checking on an epoch checkpoint (every 30k blocks) that if a signer that is unauthorized is on the list it should throw. if you init a block without specifying a parentHash it will be set to all zeroes, but maybe you could set it to the genesis block hash to get over the db error.

emersonmacro:
@ryanio Thanks - I went down quite a rabbit hole but finally tracked it down. I was really confused why that test used to work but was now failing, but once again, it was an instance of the code reorg changing the timing of when things were happening. For that particular test, changing `blockchain._validateConsensus` to `true` and signing the new block fixed it, so that the appropriate error got triggered.

@holgerd77 I think this is finally ready for consideration.

holgerd77:
Did a rebase on this after huge changeset merge along #1817 (was less problematic than I thought though).

Have a backup copy locally of the original branch just in case.

holgerd77:
@emersonmacro Ok, this did break some tests, I guess it will make sense if you have got a look on this, since you will likely easier spot the context.

I have a bit the impression that this rebase in particular did *not* do all the automatic things right (so these were rather not my manual changes breaking things). So it is likely that something has not been updated in a correct way.

If possible would be great if you find the time to fix today, it would be nice if we can finish this up before the weekend since this is also a structurally heavy change and it would be good to get this into `develop` soon. 🙂

emersonmacro:
@holgerd77 Some of the recent changes disappeared, but I was able to add them back pretty easily. I think the PR is in a good state.

packages/block/src/header.ts
ryanio: I think this was reverted, if you want to re-apply

ryanio: also for the line above it, is it possible to cast as CliqueConsensus instead of any? if not no worries

packages/block/src/header.ts
```diff
@@ -518,13 +518,13 @@ export class BlockHeader {
...
       )
       throw new Error(msg)
     }
-    const signers = (blockchain as any).cliqueActiveSigners()
+    const signers = (blockchain as any).consensus.cliqueActiveSigners()
```
ryanio, line 527: can TS infer that cliqueActiveSigners is available after the `in` check above? if not, at least cast as CliqueConsensus

packages/blockchain/src/consensus/ethash.ts
```diff
@@ -0,0 +1,28 @@
...
+  blockchain: Blockchain
+  _ethash: Ethash
+
+  constructor({ blockchain }: ConsensusOptions) {
+    this.blockchain = blockchain
+    this._ethash = new Ethash(this.blockchain.db)
+  }
+
+  async validate(block: Block): Promise<void> {
+    const valid = await this._ethash.verifyPOW(block)
```
ryanio, line 19: could we maybe add a test to add coverage for this line? bonus would be one more test to test invalid :)

packages/client/lib/miner/miner.ts
```diff
@@ -216,7 +217,13 @@ export class Miner {
       const [signerAddress, signerPrivKey] = this.config.accounts[0]
       cliqueSigner = signerPrivKey
       // Determine if signer is INTURN (2) or NOTURN (1)
-      inTurn = await vmCopy.blockchain.cliqueSignerInTurn(signerAddress)
+      // console.log(vmCopy.blockchain)
+      // inTurn = await (vmCopy.blockchain.consensus as CliqueConsensus).cliqueSignerInTurn(
+      //   signerAddress
+      // )
```
ryanio, line 223:
```suggestion
```

emersonmacro: Good catch. Removed this yesterday but missed it in today's reconciliation

ryanio:
a few last comments from me, but otherwise looks great! thanks for your persistence getting this through :)

packages/block/src/header.ts
emersonmacro: Ryan and I tried out a couple ideas but couldn't get it to work without introducing a circular dependency between `block` and `blockchain`. Decided to leave it as is for now.

holgerd77 (commented):
❤️❤️❤️

That's purely wonderful and something to be really proud of!!! 🙂💯

Can second Ryan, thanks for holding on with this.

Will merge.

PR #1849, opened 2022-04-07
vm: Improve `skipBalance` logic

### LINKED ISSUE #1839: Negative sender balance throws error in `_executeCall`
On `develop`, if `_executeCall` is executed during `executeMessage` and the sender's balance becomes negative, instead of correctly reverting the call, an error as shown below is thrown because the `bigIntToBuffer` doesn't correctly account for negative values.

```sh
Error: Cannot convert string to buffer. toBuffer only supports 0x-prefixed hex strings and this string was given: 0x-1
  at toBuffer (/home/jim/development/ethereumjs-monorepo/packages/util/src/bytes.ts:168:13)
  at bigIntToBuffer (/home/jim/development/ethereumjs-monorepo/packages/util/src/bytes.ts:224:18)
  at bigIntToUnpaddedBuffer (/home/jim/development/ethereumjs-monorepo/packages/util/src/bytes.ts:370:22)
  at Account.raw (/home/jim/development/ethereumjs-monorepo/packages/util/src/account.ts:86:29)
  at Account.serialize (/home/jim/development/ethereumjs-monorepo/packages/util/src/account.ts:96:28)
  at Cache._update (/home/jim/development/ethereumjs-monorepo/packages/vm/src/state/cache.ts:189:23)
```

Theoretically, this can only happen in `runTx` if the `skipBalance` parameter is passed to the function call but since we don't currently handle negative `bigint` to `buffer` conversions in our code, it throws instead of proceeding with function execution.

One possible solution would be to add add an arbitrarily high balance to the `from` account if `skipBalance` is set and then subtract out the added balance before returning the execution result (or set to 0 if it would be negative).  We also need to handle the same situation in `runCall` as a similar error will occur if the balance ever goes negative.

acolytec3:
Fixes #1839 

This addresses the fact that `skipBalance` in `runTx` doesn't handle skipping balance checks correctly after the initial transaction options are parsed.  Now, if `skipBalance` is set to true and the calling account balance is less than the transaction cost (i.e. intrinsic gas + value), it sets the balance equal to the transaction cost so the transaction will not fail due to an account having an insufficient balance.

It adds similar logic in `runCall` that checks to verify that the account has sufficient balance for any value transferred in the call.

### DIFF
--- a/packages/vm/src/evm/evm.ts
+++ b/packages/vm/src/evm/evm.ts
@@ -638,7 +638,7 @@ export default class EVM {
   async _reduceSenderBalance(account: Account, message: Message): Promise<void> {
     account.balance -= message.value
     if (account.balance < BigInt(0)) {
-      throw new Error('sender account has insufficient funds to execute message')
+      throw new VmError(ERROR.INSUFFICIENT_BALANCE)
     }
     const result = this._state.putAccount(message.authcallOrigin ?? message.caller, account)
     if (this._vm.DEBUG) {
--- a/packages/vm/src/evm/evm.ts
+++ b/packages/vm/src/evm/evm.ts
@@ -244,14 +244,18 @@ export default class EVM {
 
   async _executeCall(message: Message): Promise<EVMResult> {
     const account = await this._state.getAccount(message.authcallOrigin ?? message.caller)
+    let errorMessage
     // Reduce tx value from sender
     if (!message.delegatecall) {
-      await this._reduceSenderBalance(account, message)
+      try {
+        await this._reduceSenderBalance(account, message)
+      } catch (e) {
+        errorMessage = e
+      }
     }
     // Load `to` account
     const toAccount = await this._state.getAccount(message.to)
     // Add tx value to the `to` account
-    let errorMessage
     if (!message.delegatecall) {
       try {
         await this._addToBalance(toAccount, message)
@@ -633,6 +637,9 @@ export default class EVM {
 
   async _reduceSenderBalance(account: Account, message: Message): Promise<void> {
     account.balance -= message.value
+    if (account.balance < BigInt(0)) {
+      throw new Error('sender account has insufficient funds to execute message')
+    }
     const result = this._state.putAccount(message.authcallOrigin ?? message.caller, account)
     if (this._vm.DEBUG) {
       debug(`Reduced sender (${message.caller}) balance (-> ${account.balance})`)
--- a/packages/vm/src/runCall.ts
+++ b/packages/vm/src/runCall.ts
@@ -66,7 +66,7 @@ export interface RunCallOpts {
    */
   selfdestruct?: { [k: string]: boolean }
   /**
-   * Skip balance checks if true.
+   * Skip balance checks if true and adds transaction value to balance to ensure message execution doesn't fail on balance checks
    */
   skipBalance?: boolean
   /**
--- a/packages/vm/src/runCall.ts
+++ b/packages/vm/src/runCall.ts
@@ -66,6 +66,10 @@ export interface RunCallOpts {
    */
   selfdestruct?: { [k: string]: boolean }
   /**
+   * Skip balance checks if true.
+   */
+  skipBalance?: boolean
+  /**
    * If the call is a DELEGATECALL. Defaults to false.
    */
   delegatecall?: boolean
@@ -74,7 +78,7 @@ export interface RunCallOpts {
 /**
  * @ignore
  */
-export default function runCall(this: VM, opts: RunCallOpts): Promise<EVMResult> {
+export default async function runCall(this: VM, opts: RunCallOpts): Promise<EVMResult> {
   const block = opts.block ?? Block.fromBlockData({}, { common: this._common })
 
   const txContext = new TxContext(
@@ -82,6 +86,15 @@ export default function runCall(this: VM, opts: RunCallOpts): Promise<EVMResult>
     opts.origin ?? opts.caller ?? Address.zero()
   )
 
+  if (opts.skipBalance) {
+    const caller = opts.caller ?? Address.zero()
+    const callerAccount = await this.stateManager.getAccount(caller)
+    const balance = callerAccount.balance
+    if (opts.value && balance < opts.value) {
+      callerAccount.balance += opts.value - balance
+      await this.stateManager.putAccount(caller, callerAccount)
+    }
+  }
   const message = new Message({
     caller: opts.caller ?? Address.zero(),
     gasLimit: opts.gasLimit ?? BigInt(0xffffff),
--- a/packages/vm/src/runCall.ts
+++ b/packages/vm/src/runCall.ts
@@ -99,7 +99,7 @@ export default async function runCall(this: VM, opts: RunCallOpts): Promise<EVMR
     caller,
     gasLimit: opts.gasLimit ?? BigInt(0xffffff),
     to: opts.to ?? undefined,
-    value: opts.value,
+    value,
     data: opts.data,
     code: opts.code,
     depth: opts.depth ?? 0,
--- a/packages/vm/src/runCall.ts
+++ b/packages/vm/src/runCall.ts
@@ -93,7 +93,7 @@ export default async function runCall(this: VM, opts: RunCallOpts): Promise<EVMR
     const balance = callerAccount.balance
 
     if (opts.value && balance < opts.value) {
-      callerAccount.balance += opts.value - balance
+      callerAccount.balance += opts.value
       await this.stateManager.putAccount(caller, callerAccount)
     }
   }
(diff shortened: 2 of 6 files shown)


acolytec3:
I'm not sure if this is the best approach or not but this solves the immediate issue of a call to reduceSenderBalance where the cost is greater than the sender's balance.  It does so for `skipBalance` by ensuring the sender's balance equals the cost of the transaction whatever balance was set so nets out that the sender's balance gets reduced to 0 in runTx and then checks the sender's balance against the `message.value` in `runCall` if skipBalance is set to true.  

Then, in `reduceSenderBalance` we throw an error if the balance is negative and this gets handled with an exceptionError message in `executeMessage`.

packages/vm/src/runTx.ts
```diff
@@ -339,6 +340,9 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
         throw new Error(msg)
       }
     }
+  } else {
+    fromAccount.balance += cost - balance
```
jochem-brouwer, line 344: What if `cost < balance`? Then this reduces the balance..?

packages/vm/tests/api/runTx.spec.ts
```diff
@@ -640,3 +640,37 @@ tape('runTx() -> RunTxOptions', (t) => {
...
+  const tx = Transaction.fromTxData({
+    gasLimit: BigInt(21000 + 9000),
+    to: address,
+    value: BigInt(1),
+  }).sign(senderKey)
+
+  const res = await vm.runTx({ tx, skipBalance: true })
+  t.pass('runTx should not throw with no balance and skipBalance = true')
+
+  const callerBalance = (await vm.stateManager.getAccount(new Address(privateToAddress(senderKey))))
```
ryanio, line 668: you can use `Address.fromPrivateKey` instead of `privateToAddress`

packages/vm/src/evm/evm.ts
```diff
@@ -632,6 +636,9 @@ export default class EVM {
 
   async _reduceSenderBalance(account: Account, message: Message): Promise<void> {
     account.balance -= message.value
+    if (account.balance < BigInt(0)) {
+      throw new Error('sender account has insufficient funds to execute message')
```
ryanio, line 640: can you add a test to verify this message is thrown? it is currently uncovered

packages/vm/src/runCall.ts
```diff
@@ -74,14 +78,23 @@ export interface RunCallOpts {
...
-export default function runCall(this: VM, opts: RunCallOpts): Promise<EVMResult> {
+export default async function runCall(this: VM, opts: RunCallOpts): Promise<EVMResult> {
   const block = opts.block ?? Block.fromBlockData({}, { common: this._common })
 
   const txContext = new TxContext(
     opts.gasPrice ?? BigInt(0),
     opts.origin ?? opts.caller ?? Address.zero()
   )
 
+  if (opts.skipBalance) {
```
ryanio, line 89: Can you add a comment explaining, like...

```suggestion
  // if skipBalance, ensure the caller has enough balance for `value` transfer
  if (opts.skipBalance) {
```

packages/vm/src/runTx.ts
ryanio, line 344: same here, comment with an explanation

```suggestion
    // when skipBalance, ensure the caller has enough balance for the cost
    fromAccount.balance += cost - balance
```

packages/vm/src/runCall.ts
```diff
@@ -74,14 +78,23 @@ export interface RunCallOpts {
...
+    const caller = opts.caller ?? Address.zero()
```
ryanio, line 90: i think instead of defining caller twice, once here and once in `new Message`, you can define above the if statement `const caller` then in `new Message` just do `{ caller }`. then we are using one "source of truth"

packages/vm/src/runTx.ts
ryanio: it's also kind of hard to follow this else condition, i suggest we change the if without the negation and do these two lines first, then else do the checks, that might be easier to parse mentally

jochem-brouwer: Wait, does this mean that in the end the balance is always 0 of the account..?

ryanio: i think we could just add the cost to the balance? and we should note this in the param typedoc so the user isn't confused what is happening on the back end

jochem-brouwer: I think it should be consistent, in one scenario (`evm.ts`) if cost > balance then in the end balance is 0. But in here, currently it seems balance always zeroes out, and if we add balance then it does something different than in `evm.ts`, I think add the same check here, if cost > balance then zero out the account (do the same as in `evm.ts`)

ryanio: i think we'd rather error on balance going negative inside the VM than set it to zero. that seems safer to me, rather than design around this hacky skipBalance flag.

so skipBalance behavior would be: the tx cost is added to the sender prior to the tx as to not affect the sender balance (warning: this ether will be created out of thin air)

acolytec3: >I think it should be consistent, in one scenario (evm.ts) if cost > balance then in the end balance is 0. But in here, currently it seems balance always zeroes out, and if we add balance then it does something different than in evm.ts, I think add the same check here, if cost > balance then zero out the account (do the same as in evm.ts)

I'm not 100% sure I follow this.  In both `runTx` and `runCall`, if the balance is less than the transaction cost (or the value transferred) and skipBalance is set to true, we add enough ether to the caller's balance to let the account hit zero during the execution of the transaction/message.  The change in EVM.ts is to handle the case in runCall where the balance would go negative during message execution but the `skipBalance`  flag isn't set to true (or other edge cases I'm not currently thinking of where a balance could go negative).  While it kind of makes sense to have `runCall` return this error directly, it would require a deeper reworking of `runCall` since it doesn't natively return an EVM `execResult`.  Would it be okay if I just construct one with the appropriate error and return it instead of the `evm.executeMessage` call if `skipBalance === false` and the balance would end up negative?

packages/vm/src/runCall.ts
```diff
@@ -65,6 +65,10 @@ export interface RunCallOpts {
    * Addresses to selfdestruct. Defaults to none.
    */
   selfdestruct?: { [k: string]: boolean }
+  /**
+   * Skip balance checks if true and adds transaction value to balance to ensure message execution doesn't fail on balance checks
```
ryanio, line 69:
```suggestion
   * Skip balance checks if true. Adds transaction value to balance to ensure execution doesn't fail.
```

packages/vm/src/runTx.ts
```diff
@@ -47,7 +47,7 @@ export interface RunTxOpts {
    */
   skipNonce?: boolean
   /**
-   * If true, skips the balance check
+   * Skip balance checks if true and adds transaction value to balance to ensure transaction execution doesn't fail on balance checks
```
ryanio, line 50:
```suggestion
   * Skip balance checks if true. Adds transaction cost to balance to ensure execution doesn't fail.
```

acolytec3:
Okay, I think all of the current feedback has been addressed so I'm going to call this ready for review.

packages/vm/src/runTx.ts
```diff
@@ -47,7 +47,7 @@ export interface RunTxOpts {
    */
   skipNonce?: boolean
   /**
-   * If true, skips the balance check
+   * Skip balance checks if true. Adds transaction value to balance to ensure execution doesn't fail.
```
ryanio, line 50: nit, this is different than runCall
```suggestion
   * Skip balance checks if true. Adds transaction cost to balance to ensure execution doesn't fail.
```

ryanio:
looks great, thanks for all the requested updates, looks like ci is failing with a nonce off by one error

acolytec3:
> looks great, thanks for all the requested updates, looks like ci is failing with a nonce off by one error

Fixed that last issue.  I forgot to add the `nonce` to the `modifyAccountFields` method.

ryanio (approved):
lgtm! since this is rather delicate let's wait for one more 👀 and approval before merging

holgerd77:
I wouldn't include this any more in the releases for tomorrow, please do not merge.

acolytec3:
> I wouldn't include this any more in the releases for tomorrow, please do not merge.

This targets `develop` so shouldn't be impacted by the new releases, right?  It does need a second set of eyes before we merge in develop though.

ryanio (dismissed):
this still lgtm :)

holgerd77:
@ryanio in doubt maybe just merge if you feel good with it. 🙂

holgerd77:
Rebased this on latest `develop`.

holgerd77 (approved):
Also had a quick look, this looks good to me as well.

Will merge.

PR #1852, opened 2022-04-08
Client/TxPool: add tx validation

### LINKED ISSUE #1850: client: TxPool tx validation
Currently our TxPool implementation is rather simple and doesn't include tx validation checks that other tx pools have to ensure validity with consensus.

These should be added so our users can receive the error first (e.g. when using RPC via `eth_sendRawTransaction`) rather than accepting and other nodes rejecting upon broadcast.

Validation checks to add:

- [x] tx nonce is greater than sender's current nonce
- [x] tx doesn't exceed the current block gas limit
- [x] sender has enough funds to cover the costs (cost == V + GP * GL)
- [x] tx supplies enough intrinsic gas
- [x] replacement tx's gas must be 10% higher to replace (if not, error as underpriced replacement)
- [x] reject txs with data over a defined max size (to prevent DOS attacks, not a consensus validation)
- [x] make sure tx is properly signed
- [x] error if tx is already known in the pool
- [x] error if the tx pool is full and can't accept another remote tx
  - (local txs should be prioritized and accepted even when remote pool is full)

I can work on this, I just wanted to open an issue to start tracking.

For reference, geth: [tx_pool.go](https://github.com/ethereum/go-ethereum/blob/master/core/tx_pool.go), [tx_list.go](https://github.com/ethereum/go-ethereum/blob/master/core/tx_list.go)

jochem-brouwer:
Closes #1850


packages/client/lib/sync/txpool.ts
```diff
@@ -165,16 +177,38 @@ export class TxPool {
    * Adds a tx to the pool.
    *
    * If there is a tx in the pool with the same address and
-   * nonce it will be replaced by the new tx.
+   * nonce it will be replaced by the new tx, if it has a sufficent gas bump.
+   * (TODO) verifies certain constraints, if these are not matches, throws.
    * @param tx Transaction
    */
-  add(tx: TypedTransaction) {
+  async add(tx: TypedTransaction) {
```
jochem-brouwer, line 184: @ryanio do you think this method is a good central place for most checks? We can handle the throws internally here if we have a list of txs (remove them from list since these are invalid), but for RPC we catch those and instead throw the RPC error

packages/client/lib/sync/txpool.ts
```diff
@@ -119,6 +124,11 @@ export class TxPool {
    * Log pool statistics on the given interval
    */
   private LOG_STATISTICS_INTERVAL = 20000 // ms
+  
+  /**
+   * 
+   */
+  private minGasPriceBump = 10 // percentage
```
ryanio, line 131: nice yeah i don't think this needs to be a configurable opt, just a constant, like MIN_GAS_PRICE_BUMP_PERCENT

packages/client/lib/sync/txpool.ts
ryanio: yes, but i'm thinking make a `validate(tx: TypedTransaction)` method in this class, and call into it at the top of add.

packages/client/lib/rpc/modules/eth.ts
```diff
@@ -931,7 +931,8 @@ export class Eth {
 
     // Add the tx to own tx pool
     const { txPool } = this.service.synchronizer as FullSynchronizer
-    txPool.add(tx)
+    // TODO handle any errors here
+    await txPool.add(tx)
```
ryanio, line 935: yep i think this just can be

```suggestion
    try {
      await txPool.add(tx)
    } catch (error: any) {
      throw {
        code: INVALID_PARAMS
        message: error?.message ?? error.toString()
      }
    }
```

packages/client/lib/sync/txpool.ts
```diff
@@ -165,16 +177,38 @@ export class TxPool {
...
+          case 1: 
+            currentGasPrice = (existingTxn.tx as AccessListEIP2930Transaction).gasPrice
+            break
+          case 2:
+            currentGasPrice = (existingTxn.tx as FeeMarketEIP1559Transaction).maxFeePerGas
+            break
+          default: 
+            throw new Error("unknown tx type") // TODO make this error better
+        }
+        let minimumGasPrice = currentGasPrice.add(currentGasPrice.muln(this.minGasPriceBump).divn(100))
```
ryanio, line 206: hm that's a lot of boilerplate code just to get the tx gasPrice. would be nice to have a tx helper for this in the base tx class, can use the tx.supports(Capability) API to return the right response? (basically maxFeePerGas if FeeMarket, else gasPrice)

packages/client/lib/sync/txpool.ts
jochem-brouwer: Yep, makes sense, will do :+1: :smile:

packages/client/lib/sync/txpool.ts
```diff
@@ -325,7 +359,7 @@ export class TxPool {
 
     const newTxHashes = []
     for (const tx of txs) {
-      this.add(tx)
+      await this.add(tx)
```
ryanio, line 362: here (and below on L418) you can add try catch to `await this.add(tx)`, and in the catch just do a ```logger.info(`Error adding tx to TxPool: ${error.message}`)``` and continue, not much else we need to do if we get a bad tx from devp2p. we could do a `logger.error`, but i think that is too verbose, for something that is out of our control.

packages/client/lib/sync/txpool.ts
jochem-brouwer: I had exactly the same idea, I just took a look how we do this in VM, we could refactor [there](https://github.com/ethereumjs/ethereumjs-monorepo/blob/07a6d2e8a672d6bebad57a1e9df80c3e735fbf2e/packages/vm/src/runTx.ts#L356) (this gasPrice propagates via `TxContext` to `EVM` to `EEI`, where it is used to get the gas price in the `GASPRICE` opcode) as well, the only thing is that the helper needs a `baseFee` parameter in case of a 1559 txn.

Something like `getGasPrice(baseFee?: BN)` on `BaseTransaction`, on 1559 we know that we are in a 1559 block which thus has `baseFee` and otherwise in 2930/Legacy txs we can just make it `undefined` since it points to `gasPrice`. (This whole `baseFee` stuff also makes TxPool more interesting since you could include a tx in the pool which could be mined now, but at some later blocks not if `baseFee` rises above the `maxPriorityFeePerGas` ... :thinking: )

ryanio: Yeah that would be great!

The baseFee does add an interesting dynamic, so I don't think we should reject txs if they are below the baseFee, since it could eventually lower, but it would be nice on the rpc layer to know if a tx I am submitting is severely underpriced (meaning, it has very little chance to be mined). To start we can probably just model around how geth is already thinking about it, rather than coming up with our own rational, to best match their tx pool credentials.

jochem-brouwer:
Merged latest `master` changes so also have the TxPool extraction in it. 

Most of the validation logic is now in there, this of course has to be tested. Tests are now also broken due to these stricter validation rules.

jochem-brouwer:
TODOs:

- [ ] Ensure that the added tests check if adding invalid txs to TxPool throws, but also that this throws due to the right reason (check error message)
- [ ] Move the gas price calculation to Tx (however, `FeeMarketEIP1559Transaction` needs both `maxPriorityFeePerGas` and `maxFeePerGas` bumped by 10% in case user wants to override the txn)
- [ ] Fix bug in `@ethereumjs/tx` where tx options do not propagate from unsiged tx to signed tx.

jochem-brouwer:
I added the fee bump checks for all txs. I had to look into Geth how they do this. They have getters for each tx type which returns the "tip" and the "max fee" for each tx (so for Legacy and AccessLists it just returns `gasPrice`). I have implemented that here as well.

See for instance https://github.com/ethereum/go-ethereum/blob/af806168b62b6a0e772b73ce25d31d8e61d26043/core/types/legacy_tx.go#L101

Also added a check to see if tx can pay current basefee.

jochem-brouwer:
I think this is ready for a new round of reviews. However, I somewhat expect that the coverage is not 100% so might want to add more tests (will therefore keep this PR as draft PR)

packages/client/lib/service/txpool.ts
```diff
@@ -398,7 +540,7 @@ export class TxPool {
 
     const newTxHashes = []
     for (const tx of txs) {
-      this.add(tx)
+      await this.add(tx)
```
ryanio, line 543: this needs a try/catch around it in case the tx is invalid, we don't want to abort processing the other txs

packages/client/lib/service/txpool.ts
```diff
@@ -1,18 +1,25 @@
...
 import { Peer } from '../net/peer'
 import type VM from '@ethereumjs/vm'
 import type { FullEthereumService } from './fullethereumservice'
 import type { PeerPool } from '../net/peerpool'
 import type { Block } from '@ethereumjs/block'
 
+const MIN_GAS_PRICE_BUMP_PERCENT = 10
+const TX_MAX_DATA_SIZE = 128 * 1024 // 128KB
+const MAX_POOL_SIZE = 5000
+const MIN_GASPRICE = new BN(1000000000) // 1 GWei
```
ryanio, line 20: nit, the first constant is `MIN_GAS_PRICE_BUMP_PERCENT` so this one should also have a separator between GAS_PRICE so `MIN_GAS_PRICE` (can also bring it up to the top of the list)

packages/client/lib/service/txpool.ts
```diff
@@ -73,6 +80,12 @@ export class TxPool {
    */
   public pool: Map<UnprefixedAddress, TxPoolObject[]>
 
+  /**
+   * The number of txs currently in the pool
+   */
+
```
ryanio, line 86: extra newline
```suggestion
```

packages/client/lib/service/txpool.ts
```diff
@@ -178,27 +192,148 @@ export class TxPool {
...
+    // to mine underpriced txs where miner gets almost no fees
+    const currentGasPrice = this.txGasTipCap(tx)
+    if (!isLocalTransaction) {
+      const txsInPool = this.txsInPool
+      if (txsInPool >= MAX_POOL_SIZE) {
+        throw new Error('Cannot add tx: pool is full')
+      }
+      // Local txs are not checked against MIN_GASPRICE
+      if (currentGasPrice.lt(MIN_GASPRICE)) {
+        throw new Error(`Tx does not pay the minimum gas price of ${MIN_GASPRICE.toString()}`)
```
ryanio, line 266:
```suggestion
        throw new Error(`Tx does not pay the minimum gas price of ${MIN_GASPRICE}`)
```
you don't need to call `toString()` in template literals it will automatically call it for you. it would be great if you can also remove in the other places in this file.

packages/client/lib/service/txpool.ts
```diff
@@ -178,27 +192,148 @@ export class TxPool {
...
+    }
+  }
+
   /**
    * Adds a tx to the pool.
    *
    * If there is a tx in the pool with the same address and
-   * nonce it will be replaced by the new tx.
+   * nonce it will be replaced by the new tx, if it has a sufficent gas bump.
+   * (TODO) verifies certain constraints, if these are not matches, throws.
```
ryanio, line 319: todo comment still here

ryanio:
looking great! just a few last comments from me, and maybe a last look for some additional coverage/test cases, then i'd say this is pretty ready! thanks for working on it, really cool stuff, big step in sophistication for our tx pool :)

packages/client/lib/service/txpool.ts
jochem-brouwer: Ouch, good catch :open_mouth: :+1:

packages/client/lib/service/txpool.ts
```diff
@@ -178,27 +192,148 @@ export class TxPool {
     }
   }
 
+  /**
+   * Returns the gas tip (the max priority fee) for each transaction type
+   * @param tx - tx to check
+   * @returns - max priority fee
```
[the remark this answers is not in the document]
jochem-brouwer: I keep forgetting this, thanks :smile:

jochem-brouwer:
Addressed the review items, thanks! Code coverage should now cover all lines (except the log line but I don't think that is important).

ryanio:
looks excellent!

packages/client/test/miner/miner.spec.ts
```diff
@@ -325,11 +329,25 @@ tape('[Miner]', async (t) => {
...
 
-    await setBalance(vm.stateManager, A.address, new BN('200000000000001'))
+    let pkey = keccak256(Buffer.from(''))
 
     // add many txs to slow assembling
     for (let i = 0; i < 1000; i++) {
-      txPool.add(createTx())
+      // In order not to pollute TxPool with too many txs from the same address
+      // (or txs which are already known), keep generating a new address for each tx
+      const address = new Address(privateToAddress(pkey))
```
ryanio, line 338: btw we have this method
```suggestion
      const address = Address.fromPrivateKey(pkey)
```

ryanio (approved):
lgtm! great work

PR #1854, opened 2022-04-13
Migrate `common.param` (and friends) to return `BigInt`

### LINKED ISSUE #1846: Common: Return BigInt for param() functionality
There are four param-retrieving [functions](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/index.ts#L616) in Common:

- `param()`
- `paramByEIP()`
- `paramByHardfork()`
- `paramByBlock()`

These functions currently return a `number` if available, otherwise `null` (needs confirmation). When doing a search on usages of the functions it turns out that on 80%+ of the occurrences, the value is used in a BigInt context and therefore transformed to BigInt with a `BigInt()` constructor. In many of the remaining 20% cases there is an indirect need of a BigInt result (e.g. at the end of some calculation).

It would therefore make a lot of sense to directly return a BigInt on these function calls.

In addition all these functions have an `any` return type. This is highly suboptimal and the functions should be properly typed along with this transition work.

acolytec3:
Fixes #1846.  

This changes `param`, `paramByEIP`, `paramByHardfork`, and `paramByBlock` to return `bigints`.


acolytec3:
Following on conversation on #1846, I'd be curious for thoughts on whether to convert these functions to throw when a parameter isn't found or continue to return null (or alternatively just return 0).  If we convert to throw when a parameter is not found, it will require some changes in code in other places to try/catch the `common.param` method where the caller assumes a null or `Number` value like [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/2221b79b863910c3ed281bbab295ae081fcbb0f6/packages/tx/test/typedTxsAndEIP2930.spec.ts#L323). The test expects the data fee returned to be zero but two layers down in the call stack, you have [this](https://github.com/ethereumjs/ethereumjs-monorepo/blob/2221b79b863910c3ed281bbab295ae081fcbb0f6/packages/tx/src/util.ts#L99) call that expects a return value or 0 (not an error if the specific parameter doesn't exist).

acolytec3:
I'm working through the instances in the `vm` where the `param` getters are called and discovered that EIP-2537 has a parameter that returns an array of tuples which is going to add some further complexity to our implementation.
[This call](https://github.com/ethereumjs/ethereumjs-monorepo/blob/46278c4b390133fc51c9497af84f244fa737defc/packages/vm/src/evm/precompiles/0c-bls12-g1multiexp.ts#L25) expects an array returned from [this value in the EIP](https://github.com/ethereumjs/ethereumjs-monorepo/blob/46278c4b390133fc51c9497af84f244fa737defc/packages/common/src/eips/2537.json#L43).

I guess the technically "right" solution here is going to require us to check the return type in every place we use `param` or friends, something like: 
```
let value = common.param('someTopic', 'someParam')
if (typeof value === 'bigint') return value 
else value = BigInt(0)
```
And that feels really clunky.  I haven't encountered any other places where a parameter would be anything other than a number (bigint or otherwise) so I just cast that particular return type to `any` for simplicity's sake.  That's probably not the ideal solution so we need to think if maybe we should introduce a special handler for just that EIP or something else if we want to maintain type safety.

Also, I don't love what this is doing to our code.  There a lot of places in `vm` where I now have to do `common('someTopic', 'someParam') ?? BigInt(0)` which also feels really clunky.

acolytec3:
This is ready for review.  A couple of points for discussion:
1. All of the `param` getters now are typed to return either `bigint` or `undefined` 
2. This is a change from previous behavior where it returns `number` or `null` and it causes a knock-on effect where anywhere a param could return `undefined`, we have to do something like `common.param(...) ?? BigInt(0)` to satisfy Typescript.  This is clunky but explicitly spells out what was happening previously where the `null` was being implicitly coerced to the number 0 by Javascript.
3. There is one exception where an EIP parameter is not a string representing a number [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/6b0fb9d889a65edf3964f1e90e922fae7dd6327c/packages/common/src/eips/2537.json#L43) in EIP-2537 where it returns an array instead of a string.  For the moment, I've just left this edge-case undocumented in the function signature and just added special handling in the places where functionality related to this particular parameter in EIp-2537 is referenced since explicitly adding the additional typet will make point 2 even more clunky since we'll have to add additional type checking.

acolytec3:
Note, I will also add some additional tests pending final agreement on the function signature for the `param` getters.

acolytec3:
I really like the idea of just returning `BigInt(0)` if it doesn't exist and just document the function signature as such.  It's way cleaner overall.  That said, I'm still 100% sure what to do with that one special case where the parameter is an array.  I kind of want to create a special function in Common just for EIP-2537 to retrieve those values and leave it like that until we encounter another EIP that requires non-numeric values in its new parameters and then revisit.

acolytec3:
I've migrated it to just return BigInt(0) but still some vm tests broken so I think there's an edge case or two I'm missing.

acolytec3:
Does anyone (maybe @jochem-brouwer?) have insight into why the EIP-2929 tests are failing.  I've stepped through this a bunch of times using both the debugger and console logging and according to the first test, the `EXTCODEHASH` gas price should be 100 since it's checking a precompile but the gas cost for that opcode on Berlin is 700  gas plus the 100 gas for a `warmstorageread`.  I feel like I'm missing something here where the param that sets the price to 700 gas should be zeroed out or something.  This test was passing before I changed the `param` getters in common to always return `BigInt(0)` if the param didn't exist in the current common.

packages/common/src/index.ts
```diff
@@ -538,17 +538,15 @@ export default class Common extends EventEmitter {
...
     // TODO: consider the case that different active EIPs
     // can change the same parameter
-    let value = null
+    let value
     for (const eip of this._eips) {
       value = this.paramByEIP(topic, name, eip)
-      if (value !== null) {
-        return value
-      }
+      if (value !== BigInt(0)) return value
```
jochem-brouwer, line 549: Oops

acolytec3: Ah!  I figured it it was something like that.  Thanks!

jochem-brouwer: How do we solve this? We could hardcode a -1 bigint but I don't think that's clean, shall we fix the function signatures such that these can also return `null`?

acolytec3: Wait..no, that shouldn't be it.  paramByEIP will always return a bigint now (either 0n or the actual value).

acolytec3:
@holgerd77 @jochem-brouwer @ryanio (or any others) I've taken a middle path on this one so would like input from the group on this approach.

I've allows `paramByEIP` to return `undefined` if the parameter isn't found since there are cases where `paramByHardfork` will need to check in multiple places (in an EIP and then in a hardfork update if not found in the EIP) to find a parameter value and we don't want to assume that just because the EIP doesn't have a topic, a hardfork may have a value later on (like EIP2929 where a number of tests fail if we return 0 when a param isn't found in the EIP but are found in later hardfork updates) and `paramByHardfork` returns the wrong value because `paramByEIP` returned a 0 when a value wasn't found for a base fee for an opcode but a later hardfork sets a value for that parameter and that is being missed in the loginc of `paramByHardfork`.

Since paramByEIP is only called within `paramByHardfork` and then the EIP-2537 precompiles (and only because there isn't a hardfork for this EIP yet), it feels like the impact is minimal (with some slight clunkiness in those precompiles since we have to check for undefined when retrieving the param).  One thought is we could also possibly make `paramByEIP` private and then this inconsistency between `paramByEIP` and the other param getters would be removed by default since we'd no longer expose that particular getter at all.

One other change is that I hardcoded the array of gas discounts that was a parameter for `Bls12381MultiExpGasDiscount` in EIP 2537 so we no longer need to return it from the param getter and therefore don't have the one parameter that doesn't return a numeric value.

holgerd77:
I gave this another look and have to say that I would a) want to keept the `paramByEIP` function - this is just generic and well-defined functionality from the API (even if we don't use this that much ourself) and b) do not like the discrepancy between the function signature return values from `paramByHardfork` and `paramByEIP`.

I feel really sorry for this suggestion (or even question), but I re-read this thread a bit and wonder now: why did we deviate from this `number` (respectively `BigInt`) or `null` structure at all???

This has nothing to do with the intended change here and in retrospective this seems like the cleanest solution to me? This implicit `BigInt(0)` returning seems generally more and more dangerous to me the longer I think about it, also with things like a value being > 0 in older HFs then (for whatever reasons) being set to `0` and eventually not being picked up correctly and the like.

Sorry sorry sorry. I know how bad this is tearing this somewhat apart again after such a long period of non-response. 😬 😬 😬

acolytec3:
> I gave this another look and have to say that I would a) want to keept the `paramByEIP` function - this is just generic and well-defined functionality from the API (even if we don't use this that much ourself) and b) do not like the discrepancy between the function signature return values from `paramByHardfork` and `paramByEIP`.
> 
I tend to agree though I think the discrepancy made sense to me in the specific case of `paramByEIP`. 

> I feel really sorry for this suggestion (or even question), but I re-read this thread a bit and wonder now: why did we deviate from this `number` (respectively `BigInt`) or `null` structure at all???
> 
The main reason for the deviation is that when we are turning `number | null` (as the current develop branch), the `null` gets coerced to zero anyway when doing arithmetic functions where the `param` getter returns null but this no longer works so cleanly when you switch to returning `bigint | null`.  You have to explicitly check for `null` every time you call the getter and it ends up looking like `gasUsed = someGasPrice * (common.param('topic','param') ?? BigInt(0))` or something like that.  Here's an example from an earlier commit where I was following this patter 145938f30429a211fb547e8937abc3846fc8abfe

> This has nothing to do with the intended change here and in retrospective this seems like the cleanest solution to me? This implicit `BigInt(0)` returning seems generally more and more dangerous to me the longer I think about it, also with things like a value being > 0 in older HFs then (for whatever reasons) being set to `0` and eventually not being picked up correctly and the like.
> 
I guess I hear what you're saying here and the current structure where `paramByEIP` returns undefined if there is no value accounts for the concern (though with the inconsistent function signature.

When all is said and done, I don't know what the right answer is here.  I think to my way of thinking, returning `bigint | undefined` is probably the most correct solution, though it just isn't elegant and requires some extra verbosity in the code that uses these methods.  Always returning a `bigint` carries the slight risk that we made an error somewhere and there is some instance where it won't catch an earlier HF value but that feels remote at this point since the tests are all currently passing and I think we account for all of it already.

holgerd77:
>  You have to explicitly check for null every time you call the getter and it ends up looking like gasUsed = someGasPrice * (common.param('topic','param') ?? BigInt(0)) 

Ah, thanks a lot, that was my missing part that this behavior occurred already with using `null` as before, I thought the chain-of-developments would have been that 1) I suggested using `undefined` 2) you switched over and 3) this would have caused this explicit-check-needed-behavior.

holgerd77:
>  You have to explicitly check for null every time you call the getter and it ends up looking like gasUsed = someGasPrice * (common.param('topic','param') ?? BigInt(0)) 

Ah, thanks a lot, that was my missing part that this behavior occurred already with using `null` as before, I thought the chain-of-developments would have been that 1) I suggested using `undefined` 2) you switched over and 3) this would have caused this explicit-check-needed-behavior.

holgerd77:
> When all is said and done, I don't know what the right answer is here. I think to my way of thinking, returning bigint | undefined is probably the most correct solution, though it just isn't elegant and requires some extra verbosity in the code that uses these methods. Always returning a bigint carries the slight risk that we made an error somewhere and there is some instance where it won't catch an earlier HF value but that feels remote at this point since the tests are all currently passing and I think we account for all of it already.

Ok, I won't want to die on that hill and I do not have the all-catching-all-perfect-all-beautiful solution either here 😋, so I guess then let's stick with this solution, this is definitely pragmatic and won't let our API explode either. 😅

If someone finds some superior approach along the rest of the road towards the releases we can still improve on this eventually.

holgerd77:
> When all is said and done, I don't know what the right answer is here. I think to my way of thinking, returning bigint | undefined is probably the most correct solution, though it just isn't elegant and requires some extra verbosity in the code that uses these methods. Always returning a bigint carries the slight risk that we made an error somewhere and there is some instance where it won't catch an earlier HF value but that feels remote at this point since the tests are all currently passing and I think we account for all of it already.

Ok, I won't want to die on that hill and I do not have the all-catching-all-perfect-all-beautiful solution either here 😋, so I guess then let's stick with this solution, this is definitely pragmatic and won't let our API explode either. 😅

If someone finds some superior approach along the rest of the road towards the releases we can still improve on this eventually.

holgerd77:
Rebase this on latest `develop`.

holgerd77:
Rebase this on latest `develop`.

holgerd77 (approved):
Just had a bit more laid-back view on this for a final review and I have to say that this looks really, really great with all the `BigInt()` castings disappearing. 🙏 🙂 Also to actually *have* the return parameters finally typed now in the `param*` function highly outweights some eventual small remaining inconsistencies, sorry for being so picky here.

Will approve and merge, thanks a lot for tackling and the patience along the way!

PR #1855, opened 2022-04-13
client: make some `eth_` methods available on engine endpoint as per kiln spec v2.1

g11tech:
For the ease of use of CLs, some methods were deemed to be made available on the authenticated engine endpoint.

This PR enables the same.

Some TODOS:
- [x] complete the list
- [x] add availability tests for the engine endpoint of these methods
- [x] check against a CL


g11tech:
:heavy_check_mark:  Validated its working against lodestar using rpcAuth=false param for eth1 endpoints being consumed, and with `seamonkey` giving his thumbs up against other CLs.

ryanio (approved):
LGTM, thanks @g11tech!

PR #1858, opened 2022-04-18
client: refac services, syncronizers and fetchers for beacon sync

g11tech:
This PR refactors services, syncronizers and fetchers to make things cleaner for beacon sync
- [x] ~~Move txPool and `processBlocks` references out of `syncronizer` to the `fullethereumservice` so that txPool's processing will also be available in beacon sync processing of canonical chain~~ 
  As discussed with @ryanio this is per syncronizer processing, so moved back to the fullsync as beacon sync will have its own. however also moved the event association with it as its better that syncronizer wires up/cleans its own events
- [x] Fetcher now doesn't create tasks all the way to chainHeight, rather it creates tasks now in batch of 100 (configurable) and then creates more when the previous one are done
- [x] Earlier if better peer was available (With better height) the previous fetcher was torn apart and new one was created. With the new approach of creating more tasks when previous ones are done with, and updating height if better height is available with, this is no longer required. This could lead to more leaner processing of the fetching jobs.
- [x] Abstracted the fetch loop logic out of the derived syncronizers to the base syncronizer class which also looks extendable to the potential beacon sync syncronizer (with first cut look at @ryanio 's optimistic sync). Hopefull this leads to cleaner code and more usable/tested components across the syncronizers.
- [x] Some code cleanup like move `clearFetcher`,`stop` to the base syncronizer,
- [x] Update Height on the new hashes annoucement (`enqueueByNumberList`)
- [x] Removed the filtering (out) of postmerge blocks on GetBlocks/GetHeaders request responses as this is how some other node using beacon sync might get data from us.  

Testing
- [x] Test sync with single peer pre-merge
- [x] Test sync with multi peer pre-merge
- [x] Test sync post merge

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -182,6 +182,11 @@ const args = yargs(hideBin(process.argv))
     number: true,
     default: Config.MAXPERREQUEST_DEFAULT,
   })
+  .option('maxFetcherJobs', {
+    describe: 'Max tasks or jobs to be created for a fetcher at a time',
+    number: true,
+    default: Config.MAXFETCHERJOBS_DEFAULT,
+  })
   .option('minPeers', {
     describe: 'Peers needed before syncing',
     number: true,
--- a/packages/client/lib/config.ts
+++ b/packages/client/lib/config.ts
@@ -130,9 +130,9 @@ export interface ConfigOptions {
   maxPerRequest?: number
 
   /**
-   * Max jobs to be enqued in a fetcher at any given time
+   * Max jobs to be enqueued in the fetcher at any given time
    *
-   * Default: `100``
+   * Default: `100`
    */
   maxFetcherJobs?: number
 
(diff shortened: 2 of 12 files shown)


g11tech:
hey @ryanio I think its ready for a round of review :pray: :slightly_smiling_face:  will try adding test for `processBlocks` afterwards to increase the coverage.

holgerd77 (approved):
Ok, I think I am through here relatively cautiously, thanks Gajinder, this looks great, lot of really nice improvements, especially this fetcher job size limiting makes so much sense! 😄 ❤️ 👍

One nit, no blocker though: I tested with `--maxFetcherJobs=0` and was suprised that the fetcher is still working and fetching stuff. 😋

Will approve and merge.

g11tech:
> Ok, I think I am through here relatively cautiously, thanks Gajinder, this looks great, lot of really nice improvements, especially this fetcher job size limiting makes so much sense! smile heart +1
> 
> One nit, no blocker though: I tested with `--maxFetcherJobs=0` and was suprised that the fetcher is still working and fetching stuff. yum
> 
> Will approve and merge.

yea its picking maxFetcherJobs as always 100 (default value) disregarding any input, will figure out and fix! thanks @holgerd77 :slightly_smiling_face: :heart:

PR #1859, opened 2022-04-19
util: replace hash module with ethereum-cryptography

### LINKED ISSUE #1844: Util: Removal of the hash utility module / direct ethereum-cryptography usage
With the maturation of the [ethereum-cryptography](https://github.com/ethereum/js-ethereum-cryptography) library the integrated [hash](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/src/hash.ts) module in the `Util` package just represents an unnecessary chaining step.

The `hash` module should therefore be removed and replaced by direct `ethereum-cryptography` usage in all monorepo libraries.

emersonmacro:
Closes #1844


ryanio:
wohoo, thanks for working on this :) looks nice and straightforward on first glance

packages/vm/src/evm/opcodes/functions.ts
```diff
@@ -379,7 +380,7 @@ export const handlers: Map<number, OpHandler> = new Map([
       if (length !== BigInt(0)) {
         data = runState.memory.read(Number(offset), Number(length))
       }
-      const r = bufferToBigInt(keccak256(data))
+      const r = bufferToBigInt(toBuffer(keccak256(data)))
```
ryanio, line 383: not sure if we need to initialize a buffer here, can just use `bytesToHex` exported from the `ethereum-cryptography` utils. so `BigInt('0x' + bytesToHex(keccak256(data)))`

also, anywhere you do `toBuffer().toString('hex')` can directly use `bytesToHex`

packages/vm/src/evm/opcodes/functions.ts
```diff
@@ -532,7 +533,7 @@ export const handlers: Map<number, OpHandler> = new Map([
         return
       }
 
-      runState.stack.push(bufferToBigInt(keccak256(code)))
+      runState.stack.push(bufferToBigInt(toBuffer(keccak256(code))))
```
ryanio, line 536: same here, `BigInt('0x' + bytesToHex(keccak256(code)))`

packages/vm/src/evm/opcodes/util.ts
```diff
@@ -39,7 +40,7 @@ export function addressToBuffer(address: bigint | Buffer) {
  * Error message helper - generates location string
  */
 export function describeLocation(runState: RunState): string {
-  const hash = keccak256(runState.eei.getCode()).toString('hex')
+  const hash = toBuffer(keccak256(runState.eei.getCode())).toString('hex')
```
ryanio, line 43:
```suggestion
  const hash = bytesToHex(keccak256(runState.eei.getCode()))
```

holgerd77:
Yeah, thanks, really cool! 👍 🎉

packages/vm/src/evm/opcodes/functions.ts
emersonmacro: @ryanio Ah, nice, yeah I was wondering if there was a better way to do that. Updated throughout.

However, using `ethereum-cryptography/utils` seems to have introduced a build issue. Getting a strange typescript error during the `npm i` step of all the CI jobs - e.g. `Error: [Node build] Working... ../../node_modules/ethereum-cryptography/utils.d.ts(7,11): error TS2304: Cannot find name 'Crypto'.` (https://github.com/ethereumjs/ethereumjs-monorepo/runs/6106526984?check_suite_focus=true#step:4:102)

`bytesToHex` is just re-exported by `ethereum-cryptography` from `@nobles/hashes`, but we don't have `@nobles/hashes` listed as a dependency anywhere. Maybe it wouldn't hurt to list it as a dependency and use it directly since it's already being installed as a sub-dependency?

Have you seen that error before? I looked into it a bit but ran out of time before needing to go to bed :(

holgerd77: I would want to avoid to use `@noble/hashes` directly respectively add as an additional depedency, that's what `ethereum-cryptography` is for to give an (thin layer of) abstraction here.

ryanio: That's interesting because it's crypto with a capital C, which I haven't seen before (these types come from `@types/node` usually). I haven't checked out the code yet to test, but I have a feeling it is related to this type in `ethereum-cryptography/src/utils:L43`: `export const crypto: { node?: any; web?: Crypto }`. I don't have a solution to solve yet, but I can try playing around.

holgerd77: Maybe something to forward to Patricio? 🤔

alcuadrado: It may be because of a different typescript version?

`Crypto` is defined in `typescript/lib/lib.dom.d.ts`.

alcuadrado: Adding `"dom"` here should fix it

https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/config/tsconfig.json#L15

emersonmacro: @alcuadrado Thank you - that fixed it :)

@holgerd77 @ryanio Build is working again. Ready for review

ryanio: thanks @alcuadrado!!

ryanio (approved):
really nice, lgtm!

holgerd77:
Nice. 🙂

PR #1860, opened 2022-04-20
VM -> EIP-1153 (Transient Storage) Fix: Improve the time complexity of commit by using a journal instead…

moodysalem:
… of stack of maps

this is necessary because creating a large call stack, then writing a bunch of keys and having all the calls return could potentially DOS the client

also this pr removes unused methods

another implementation considered was a stack of maps with checkpoints, instead of merging the maps, but that is easily exploited since there is no difference in gas cost between writing unique keys and writing the same keys. the overhead of maps may not be worth it


moodysalem:
@tynes

holgerd77 (approved):
Thanks, LGTM

PR #1862, opened 2022-04-21
EVM/VM refactor

gabrocheleau:
Started from #1831 

## TODO

- [x] Rebase on top of current develop
- [x] Fix remaining failing tests
- [x] Implement the changes in #1819
- [x] Unify executeMessage and runCall (see https://github.com/ethereumjs/ethereumjs-monorepo/pull/1831#issuecomment-1093055152)

### For another PR
- [ ] Extract the EEI
- [ ] Extract the refactored parts into separate packages (`VmExecution`, `VmContext`, `ExternalEnvironment`, etc.?)


gabrocheleau:
@holgerd77 , I think this is somewhat close to being finished, but I just ran into a bit of a blocker. The blocker is caused by the conflicting changes brought by the refactoring of the EVM in your PR (where we assign a state manager to the EVM instance) and @g11tech's PR #1817 , which abstracts stateManager out and adds vmState to wrap its context in the vm. 

Since VM/EVM/EEI are still highly intertwined, with the `state` or `vmState` of one being assign to others in a couple places, we now get conflicting types. My understanding is that we need the `stateManager` in the EVM, since we need to be accessing the database and updating the trie, whereas we do not in the VM (which then uses `vmState`). See for example [this line](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1862/files#diff-0ae494f1d2235891dc170ffd0679fdc24b7471fedcf419d747bf3fd59d745cd6R78): which is currently broken since we're trying to assign the evm's `state: StateManager` to the Interpreter's `state: VmState`

Happy to "ping-pong" on this, or just to discuss the desired approach before continuing the work.

g11tech:
> @holgerd77 , I think this is somewhat close to being finished, but I just ran into a bit of a blocker. The blocker is caused by the conflicting changes brought by the refactoring of the EVM in your PR (where we assign a state manager to the EVM instance) and @g11tech's PR #1817 , which abstracts stateManager out and adds vmState to wrap its context in the vm.
> 
> Since VM/EVM/EEI are still highly intertwined, with the `state` or `vmState` of one being assign to others in a couple places, we now get conflicting types. My understanding is that we need the `stateManager` in the EVM, since we need to be accessing the database and updating the trie, whereas we do not in the VM (which then uses `vmState`). See for example [this line](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1862/files#diff-0ae494f1d2235891dc170ffd0679fdc24b7471fedcf419d747bf3fd59d745cd6R78): which is currently broken since we're trying to assign the evm's `state: StateManager` to the Interpreter's `state: VmState`
> 
> Happy to "ping-pong" on this, or just to discuss the desired approach before continuing the work.

assigning  `vmState`  should work, as it proxies calls to the underlying statemanager

gabrocheleau:
> > @holgerd77 , I think this is somewhat close to being finished, but I just ran into a bit of a blocker. The blocker is caused by the conflicting changes brought by the refactoring of the EVM in your PR (where we assign a state manager to the EVM instance) and @g11tech's PR #1817 , which abstracts stateManager out and adds vmState to wrap its context in the vm.
> > Since VM/EVM/EEI are still highly intertwined, with the `state` or `vmState` of one being assign to others in a couple places, we now get conflicting types. My understanding is that we need the `stateManager` in the EVM, since we need to be accessing the database and updating the trie, whereas we do not in the VM (which then uses `vmState`). See for example [this line](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1862/files#diff-0ae494f1d2235891dc170ffd0679fdc24b7471fedcf419d747bf3fd59d745cd6R78): which is currently broken since we're trying to assign the evm's `state: StateManager` to the Interpreter's `state: VmState`
> > Happy to "ping-pong" on this, or just to discuss the desired approach before continuing the work.
> 
> assigning `vmState` should work, as it proxies calls to the underlying statemanager (evm's state should also be vmState i think)

Cool, thanks for your help with this @g11tech. So the issue I was mentioning above seems to be fixed, some of the tests are at least passing now. Next step will be going through the failing tests :)

gabrocheleau:
Still some failing tests, but getting closer. I will not be working on this today/tomorrow, so if anyone feels like fixing a couple tests, feel free! :) Btw majority of the remaining failing vm:API tests are related to `eip-3529.spec.ts`.

acolytec3:
The issue with the EIP-3529 tests is that `vm._evm._refund`  isn't getting cleared at the end of each `runCode`.  I put a hack in the test to verify that it works but I think there's a deeper issue in the refactor where this refund needs to be cleared when the interpreter returns its result or something.  Will look at it more later today.

gabrocheleau:
Fixed some additional tests related to EIP-1153 (transient storage). There are still 3 failing tests in `eip-1153.spec.ts`, all related to the expected stack differing from the actual stack.

jochem-brouwer:
Ok, this fixes it, I should have thorougly read the comments before working on this since @acolytec3 already found the problem.

I don't like this fix since this makes the VM edit rather "inner" variables of the EVM which we want to avoid. EVM should (?) b e a black box and should handle this stuff internally. I think we can actually do that, if a message finishes, we can check the callstack depth, if the depth is 1 then we know that the "entire" message (i.e. a transaction) has finished and we can cleanup internally.

packages/vm/src/evm/evm.ts
```diff
@@ -579,12 +805,166 @@ export default class EVM {
     }
   }
 
+  /**
+   * Executes an EVM message, determining whether it's a call or create
+   * based on the `to` address. It checkpoints the state and reverts changes
+   * if an exception happens during the message execution.
+   */
+
```
ryanio, line 813: extra newline can be removed here

packages/vm/src/evm/evm.ts
```diff
@@ -579,12 +805,166 @@ export default class EVM {
...
+        debug(`message checkpoint committed`)
+      }
+    }
+    await this._emit('afterMessage', result)
+
+    return result
+  }
+
+  /**
+   * @ignore
```
ryanio, line 932: I think we usually use `@hidden`. Could you also add a sentence why it's hidden? e.g. this is bound to the global VM and thus shouldn't be used directly from the evm class

packages/vm/tests/api/index.spec.ts
```diff
@@ -53,16 +53,6 @@ tape('VM -> basic instantiation / boolean switches', (t) => {
     )
     st.end()
   })
-
-  t.test('should instantiate with async constructor', async (st) => {
-    const vm = await VM.create({ activatePrecompiles: true })
```
jochem-brouwer, line 58: Why is this removed?

packages/vm/src/evm/evm.ts
```diff
@@ -579,12 +805,166 @@ export default class EVM {
...
+   */
+  async runCode(opts: RunCodeOpts): Promise<ExecResult> {
+    const block = opts.block ?? Block.fromBlockData({}, { common: this._common })
+    this._block = block
+
+    // Backwards compatibility
+    const txContext: TxContext = opts.txContext ?? {
```
ryanio, line 939: on develop i already removed opts.txContext (the updated typedocs were also lost if you can copy them over)
```suggestion
    const txContext: TxContext = {
```

packages/vm/tests/api/index.spec.ts
gabrocheleau: Duplicate of the test just above

packages/vm/src/evm/evm.ts
gabrocheleau: Removed, thanks!

packages/vm/src/evm/evm.ts
gabrocheleau: Done!

packages/vm/src/evm/evm.ts
gabrocheleau: removed!

holgerd77:
Super cool! 🎉 🙏 😀

If you look at the [commit history](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1862/commits), there is still this one commit 34762b66fafe6a10bc1ff0218e11e0dcd3918617 from Emmett which sneaked in, this would need to be removed.

Also in this commit 57ef8e5fed328f61f877117a231739649fecc823 there went in a lot of changes from other work which shouldn't be there.

I guess this needs one other round of commit-cherry-picking one by one, starting with the post-Emmett-commit and then going forward. 🙂 I guess for the 57ef8e5 commit it is *not* possible to use this commit in any way and you need to re-create the work (copy-and-paste) manually and just taking in the changes you intend to apply on this step. Then you should be able to continue to cherry-pick the additional commits.

This might sound laborious but I guess this should be done in < 1 hour (if nothing goes wrong 😋). Sorry, but we are so far in the `develop` work process, we really need to be picky here since we can't risk damaging the branch in this late state of the work.

I guess then we can merge in really quickly though, really looking forward to get the main substance of this into the working base state! 🚀

gabrocheleau:
Closing in favor of https://github.com/ethereumjs/ethereumjs-monorepo/pull/1892

PR #1865, opened 2022-04-25
Fix broken Readme link to run the VM in a browser

andreterron:
Clicking on the "Running the VM in a browser" link was resulting in a 404 error.
This change adds the `.js` extension to the URL, fixing the issue.


ryanio (approved):
lgtm, thanks!

PR #1868, opened 2022-04-28
vm/docs: mention that stack in step event can change

### LINKED ISSUE #1829: VM: Some members of the eventObject are not immutable
In the VM "step" event, it includes a stack reference (Array<BN>), and value at this reference is changed in further "step"s in same message.

https://github.com/ethereumjs/ethereumjs-monorepo/blob/3ae51b7f5d0ab1e6100df71dbf5a91729eefc49e/packages/vm/src/evm/interpreter.ts#L239

Since deep copies would cause a considerable performance hit (given the huge number of "step" events usually), it makes sense that it's not immutable. Mentioning in [docs](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/vm#usage) may help someone save time. Would it be fine if it's mentioned in the front example?

```ts
// Note that numbers added are hex values, so '20' would be '32' as decimal e.g.
const code = [PUSH1, '03', PUSH1, '05', ADD, STOP]

vm.on('step', function (dataMut) {
  // note that dataMut contains references to internal state, so if needed to be snapshotted, please deep copy it
  console.log(`Opcode: ${dataMut.opcode.name}\tStack: ${dataMut.stack}`)
})
```

Initial discussion: https://discord.com/channels/717003406484701255/717022899072008263/958717667617951765

zemse:
Closes #1829.

### DIFF
--- a/packages/vm/README.md
+++ b/packages/vm/README.md
@@ -31,6 +31,7 @@ const PUSH1 = '60'
 const code = [PUSH1, '03', PUSH1, '05', ADD, STOP]
 
 vm.on('step', function (data) {
+  // Note that data.stack is not immutable, i.e. it is a reference to the vm's internal stack object
   console.log(`Opcode: ${data.opcode.name}\tStack: ${data.stack}`)
 })
 


ryanio (approved):
cool, lgtm, thanks!

PR #1874, opened 2022-04-30
trie: convert constructor to options interface, remove deprecated setRoot

ryanio:
Completes tasks in #1717:

 - [x] Replace constructor options with an options dict (like in other libraries), add an options type
 - [x] Remove deprecated `Trie.setRoot()` function

extra:
- [x] Removes `assert` usage, preferring normal errors

### DIFF
--- a/packages/trie/src/baseTrie.ts
+++ b/packages/trie/src/baseTrie.ts
@@ -18,7 +18,6 @@ import {
 import { verifyRangeProof } from './verifyRangeProof'
 // eslint-disable-next-line implicit-dependencies/no-implicit
 import type { LevelUp } from 'levelup'
-const assert = require('assert')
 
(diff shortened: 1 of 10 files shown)


packages/trie/src/baseTrie.ts
```diff
@@ -735,7 +750,7 @@ export class Trie {
    */
   copy(): Trie {
     const db = this.db.copy()
-    return new Trie(db._leveldb, this.root)
+    return new Trie({ db: db._leveldb, root: this.root, deleteFromDB: this._deleteFromDB })
```
holgerd77, line 741: Ah, that's all a lot nicer, great to see this. 👍 🙂

holgerd77 (approved):
LGTM

PR #1875, opened 2022-05-02
devp2p: remove `async` dependency from integration tests

ryanio:
This PR removes the `async` dep from the devp2p integration tests.

I was getting some errors from the peer liveness check when the server closed too quickly so I added some 500ms delays before closing, not sure why I had this problem.


holgerd77 (approved):
Nice, LGTM! 🙂

Issue #1754, opened 2022-02-28
Blockchain: Deprecation Tasks / Head Function Renamings / Iterator adjustments


holgerd77:
Part of #1717 

The following tasks are to be done along deprecation for the `Blockchain` package:

- The [meta](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/blockchain/src/index.ts#L303) property can be removed
- The deprecated [getHead()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/blockchain/src/index.ts#L787) function can be removed (this is *not* the canonical head but the DB iterator head, see associated code docs of the function itself and the replacement function `getIteratorHead()` (above))
- Same with deprecated [setHead()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/blockchain/src/index.ts#L1334) function

I would be a fan of the following renamings (maybe execute upon if no one objects here during the day or so):

- `getLatestHeader()` -> `getCanonicalHeadHeader()`
- `getLatestBlock()` -> `getCanonicalHeadBlock()`

This is to make this more expressive and semantically clear and better differentiate on the iterator head functions.

holgerd77:
Ah, and another deprecation task, will also put here since not too big: the iterator() function was expanded at some point (for the client) to not return `void` (so: give no feedback) but always return the number of blocks being ran.

The blockchain *implementation*  is already reflecting this. The *interface*  though still has `void` in it. This should be removed. Also the `VM.runBlockchain()` function still forwards (returns) this `void` which can also be removed along. (the client is also using `blockchain.iterator()` along execution but I guess that's fine? Might need some short confirmation though.)

holgerd77:
> Ah, and another deprecation task, will also put here since not too big: the iterator() function was expanded at some point (for the client) to not return `void` (so: give no feedback) but always return the number of blocks being ran.
> 
> The blockchain _implementation_ is already reflecting this. The _interface_ though still has `void` in it. This should be removed. Also the `VM.runBlockchain()` function still forwards (returns) this `void` which can also be removed along. (the client is also using `blockchain.iterator()` along execution but I guess that's fine? Might need some short confirmation though.)

Ah, this additional point hasn't been tackled yet, right? @ScottyPoi do you also want to take this over (it's a really really small one, maybe you could do with the test re-additions), since you have also tackled the other stuff?

[...]

PR #1877, opened 2022-05-04
blockchain: fix iterator return type, reinstate tests

ScottyPoi:
Small task to finish #1754.

Removed 'void' as a return type for blockchain.iterator()

Adapted VM and Client methods that use the blockchain iterator

Also reinstate VM tests for blockchain.getIteratorHead()  that were a part of #1822


holgerd77 (approved):
Thanks a lot Scotty, this looks good! 🙂 👍 

Will merge in.

PR #1884, opened 2022-05-13
Tx: Ensure TxOptions propagate from unsigned to signed

jochem-brouwer:
When creating a transaction, `TxOptions` do not get propagated from the unsigned tx to the signed tx. I'd say that if one creates an unsigned tx with `{freeze: false}` then one expects that the signed tx is then also frozen (?). The alternative is that we explicitly add this as potentially undefined parameter in `sign`.

WDYT?

### DIFF
--- a/packages/client/lib/rpc/modules/eth.ts
+++ b/packages/client/lib/rpc/modules/eth.ts
@@ -73,7 +73,7 @@ type JsonRpcBlock = {
   gasLimit: string // the maximum gas allowed in this block.
   gasUsed: string // the total used gas by all transactions in this block.
   timestamp: string // the unix timestamp for when the block was collated.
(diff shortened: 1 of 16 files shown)


holgerd77:
Yes, I would very much agree here, will likely be also true for additional future options as well (one might need to be careful with creational options though which do something in the constructor like (nonsense example here but just to get the idea) `calcBaseFee` or something like that).

Is this still WIP? I would otherwise be confident to review and merge here I would say.

jochem-brouwer:
I wanted to await some comments on this one but had a discussion with @ryanio on discord and I think he also agreed this is correct (?) so I will mark this as ready for review.

jochem-brouwer:
Thanks for the rebase :smile:

ryanio:
My only comment is that we should not store and use this.txOptions.common since we already have this.common and do not want two sources of truth, since common can be changed with functions like setHardforkByBlock

ryanio (approved):
lgtm!

PR #1892, opened 2022-05-17
EVM/VM refactor

### LINKED ISSUE #528: VM: Message and TxContext as interfaces

gabrocheleau:
Cleanup of #1862

Resolves  #528


packages/vm/src/evm/evm.ts
```diff
@@ -9,25 +16,116 @@ import {
...
+   *    // The base fee of the opcode
+   *    baseFee: number
+   *    // If the opcode charges dynamic gas, add this here. To charge the gas, use the `i` methods of the BN, to update the charged gas
+   *    gasFunction?: function(runState: RunState, gas: BN, common: Common)
+   *    // The logic of the opcode which holds the logic of changing the current state
+   *    logicFunction: function(runState: RunState)
+   * }
+   * Note: gasFunction and logicFunction can both be async or synchronous functions
+   */
+
```
ryanio, line 117: extra newline
```suggestion
```

packages/vm/src/evm/evm.ts
```diff
@@ -579,12 +729,168 @@ export default class EVM {
     }
   }
 
+  /**
+   * Executes an EVM message, determining whether it's a call or create
+   * based on the `to` address. It checkpoints the state and reverts changes
+   * if an exception happens during the message execution.
+   */
+
```
ryanio, line 737: extra newline
```suggestion
```

packages/vm/src/evm/evm.ts
```diff
@@ -579,12 +729,168 @@ export default class EVM {
...
+        selfdestruct: opts.selfdestruct ?? {},
+        delegatecall: opts.delegatecall,
+      })
+    }
+
+    await this._emit('beforeMessage', message)
+
+    if (!message.to && this._common.isActivatedEIP(2929)) {
+      message.code = message.data
+      ;(<any>this._state).addWarmedAddress((await this._generateAddress(message)).buf)
```
ryanio, line 778: do we still need this `any` cast on `this._state`?

packages/vm/src/evm/evm.ts
```diff
@@ -129,120 +227,172 @@ export function VmErrorResult(error: VmError, gasUsed: bigint): ExecResult {
...
-      const { gasUsed, exceptionError, returnValue } = result.execResult
-      debug(
-        `Received message execResult: [ gasUsed=${gasUsed} exceptionError=${
-          exceptionError ? `'${exceptionError.error}'` : 'none'
-        } returnValue=0x${short(returnValue)} gasRefund=${result.gasRefund ?? 0} ]`
-      )
+
+    // Safeguard if "process" is not available (browser)
+    if (process !== undefined && process.env.DEBUG) {
+      this.DEBUG = true
```
ryanio, line 355: seems like we have this `DEBUG` property in three places now, in the vm, in evm.ts, and interperer.ts, is there a way to simplify to just use one source of truth, setting and accessing it in one place? doing it additional times feels redundant

packages/vm/src/evm/evm.ts
```diff
@@ -579,12 +729,168 @@ export default class EVM {
...
+  async runCode(opts: RunCodeOpts): Promise<ExecResult> {
+    const block = opts.block ?? Block.fromBlockData({}, { common: this._common })
+    this._block = block
+
+    // Backwards compatibility
+    const txContext: TxContext = {
+      gasPrice: opts.gasPrice ?? BigInt(0),
+      origin: opts.origin ?? opts.caller ?? Address.zero(),
+    }
+    this._tx = txContext
```
ryanio, line 869: instead of defining the const, just set it directly?

```suggestion
    this._tx = {
      gasPrice: opts.gasPrice ?? BigInt(0),
      origin: opts.origin ?? opts.caller ?? Address.zero(),
    }
```

packages/vm/src/evm/evm.ts
```diff
@@ -579,12 +729,168 @@ export default class EVM {
...
+  async runCall(opts: RunCallOpts): Promise<EVMResult> {
+    let message = opts.message
+    if (!message) {
+      const block = opts.block ?? Block.fromBlockData({}, { common: this._common })
+      this._block = block
+      const txContext: TxContext = {
+        gasPrice: opts.gasPrice ?? BigInt(0),
+        origin: opts.origin ?? opts.caller ?? Address.zero(),
+      }
+      this._tx = txContext
```
ryanio, line 747: same here, just set directly instead of defining a const

packages/vm/src/evm/evm.ts
```diff
@@ -579,12 +729,168 @@ export default class EVM {
...
+
+    // Backwards compatibility
+    const txContext: TxContext = {
+      gasPrice: opts.gasPrice ?? BigInt(0),
+      origin: opts.origin ?? opts.caller ?? Address.zero(),
+    }
+    this._tx = txContext
+
+    const message =
+      opts.message ??
```
ryanio, line 872: do we ever pass in opts.message? if this is a relic of the past, do we want to remove it like we did txContext opt?

packages/vm/tests/api/state/accountExists.spec.ts
```diff
@@ -41,8 +41,8 @@ tape('correctly apply new account gas fee on pre-Spurious Dragon hardforks', asy
     value: BigInt(0),
   }
 
-  const result = await vm.runCall(runCallArgs)
-  t.equal(result.execResult.gasUsed, BigInt(53552), 'vm correctly applies new account gas price')
+  const result = await vm.evm.runCall(runCallArgs)
+  t.ok(result.execResult.gasUsed === BigInt(53552), 'vm correctly applies new account gas price')
```
ryanio, line 45: can you please change back to `t.equal` (same with on L91)

packages/vm/tests/api/runCode.spec.ts
```diff
@@ -21,7 +21,7 @@ const testCases = [
 tape('VM.runCode: initial program counter', async (t) => {
   const vm = await VM.create()
 
-  for (const [i, testData] of testCases.entries()) {
+  for (const [, testData] of testCases.entries()) {
```
ryanio, line 24: if you don't need the index, i think you can just do `for (const testData of testCases) {` since it is already an array. 

in general, for the future, i prefer to use underscore prefix to describe a variable as unused (e.g. here `_i`) rather than blank, i find it easier to follow.

packages/vm/src/evm/evm.ts
holgerd77: I guess we DO need in VM and EVM if we separate the packages, but we should definitely unify in the EVM scope (so: evm.ts and interpreter.ts).

packages/vm/src/evm/evm.ts
```diff
@@ -60,6 +60,10 @@ export interface EVMOpts {
    * - `experimental`: behaviour can change on patch versions
    */
   common?: Common
+  /**
+   * A {@link StateManager} instance to use as the state store
+   */
+  stateManager: StateManager
```
holgerd77: Just a note: this should rather be an in-between-step, after the EEI extraction (where external things like state access is encapsulated), a state manager dependency in the EVM itself should not be necessary any more.

packages/vm/src/evm/evm.ts
```diff
@@ -646,6 +644,68 @@ export default class EVM {
...
+    })
+
+    return this.executeMessage(message)
+  }
+
+  /**
+   * Bound to the global VM and therefore
+   * shouldn't be used directly from the evm class
+   * @hidden
+   */
```
holgerd77, line 859: This method should not be hidden since it will become part of the public API and rather get a proper code doc introduction.

holgerd77 (commented):
Wow, this looks hyper-insanely-great 🎉 🚀 😀 ❤️, thanks so much for sticking through with this Gabriel!

From my side this can be approved and merged once the review comments are addressed.

packages/vm/src/evm/evm.ts
holgerd77: Yes, that describes it well.

packages/vm/src/evm/evm.ts
gabrocheleau: Removed hidden.

packages/vm/src/evm/evm.ts
gabrocheleau: good point, fixed

packages/vm/tests/api/runCode.spec.ts
gabrocheleau: Re-added the i since I reverted to the old error message with the index

packages/vm/tests/api/runCode.spec.ts
```diff
@@ -30,13 +30,9 @@ tape('VM.runCode: initial program counter', async (t) => {
...
     try {
-      const result = await vm.runCode(runCodeArgs)
+      const result = await vm.evm.runCode(runCodeArgs)
       if (testData.resultPC !== undefined) {
-        t.equal(
-          result.runState?.programCounter,
-          testData.resultPC,
-          `should start the execution at the specified pc or 0, testCases[${i}]`
-        )
+        t.equals(result.runState?.programCounter, testData.resultPC, 'runstate.programCounter')
```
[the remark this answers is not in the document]
gabrocheleau: yeah, that's a rebasing mistake, fixed

packages/vm/tests/api/state/accountExists.spec.ts
gabrocheleau: reverted both

packages/vm/src/evm/evm.ts
gabrocheleau: we don't! removed

packages/vm/src/evm/evm.ts
gabrocheleau: Unified interpreter and evm to both use the evm debug prop.

packages/vm/src/evm/message.ts
```diff
@@ -1,48 +1,84 @@
...
-  code: Buffer | PrecompileFunc
-  _codeAddress: Address
+  code?: Buffer | PrecompileFunc
+  _codeAddress?: Address
   isStatic: boolean
   isCompiled: boolean
-  salt: Buffer
-  selfdestruct: any
+  salt?: Buffer
+  selfdestruct?: { [key: string]: boolean } | { [key: string]: Buffer }
```
ryanio, line 43: can you add a typedoc with your findings/understanding of the two values (boolean vs buffer), i.e. set to boolean when marked for selfdestruct and set to buffer for the refund address

packages/vm/src/evm/evm.ts
```diff
@@ -579,12 +728,161 @@ export default class EVM {
...
+  }
+
+  /**
+   * Bound to the global VM and therefore
+   * shouldn't be used directly from the evm class
+   */
+  async runCode(opts: RunCodeOpts): Promise<ExecResult> {
+    const block = opts.block ?? Block.fromBlockData({}, { common: this._common })
+    this._block = block
+
```
ryanio, line 859:
```suggestion
    this._block = opts.block ?? Block.fromBlockData({}, { common: this._common })
```

packages/vm/src/evm/message.ts
gabrocheleau: Added a note! :)

packages/vm/src/evm/evm.ts
gabrocheleau: Done!

ryanio (dismissed):
lgtm! holger can give final approval and merge :) really nice work!

packages/vm/src/evm/message.ts
```diff
@@ -1,48 +1,87 @@
...
+  gasLimit: bigint
+  data?: Buffer
+  depth?: number
+  code?: Buffer | PrecompileFunc
+  codeAddress?: Address
+  isStatic?: boolean
+  isCompiled?: boolean
+  salt?: Buffer
+  /**
+   * If boolean, indicates if marked to selfdestruct. If address, provides addresses to selfdestruct.
```
ryanio, line 27:
```suggestion
   * Map of addresses to selfdestruct. Key is the unprefixed address. Value is a boolean when marked for destruction and replaced with a Buffer containing the address where the remaining funds are sent.
```

gabrocheleau: Good idea, updated!

gabrocheleau:
> lgtm! holger can give final approval and merge :) really nice work!

Hey :) Just tagging @holgerd77 for final approval & merge

holgerd77 (approved):
LGTM! 

Thanks a lot Gabriel! 🙂

PR #1893, opened 2022-05-18
devp2p: Snap devp2p tests

scorbajio:
This PR adds two tests for the SNAP protocol that is being added to the client as well as a utility function for setting up two peers for SNAP communication.


holgerd77:
Hey there, how cool is that??? 🙏 ❤️ 🙂

Snap sync tests falling out of the sky. 🌥 😅

Nevertheless rather up to @jochem-brouwer doing a review here! 😄

holgerd77:
(so would this be already open for review?)

scorbajio:
> (so would this be already open for review?)

Yes, this is open for review.

Issue #1890, opened 2022-05-17
Tx: Throw when hash() is called on unsigned legacy txs


holgerd77:
Part of #1717 

Very small remaining breaking-release task:

Throw when `hash()` is called on unsigned legacy txs (so: align with behavior from typed txs), see also comment added along with #1445.

So if one is looking at e.g. the EIP-1559 Tx `hash()` function, the function throws if tx not signed, for the legacy tx this is not the case.

Will assign @ScottyPoi since he asked for a small EthereumJS task. 🙂

ScottyPoi:
Can this be closed?

PR #1894, opened 2022-05-18
Tx: Throw when hash() is called on unsigned legacy transaction

ScottyPoi:
#1890
Part of #1717

1. Edit hash() in legacyTransaction.ts to throw an error if called on an unsigned transaction  to align with #1445
1. Edit test in legacy.spec.ts to expect tx.hash() to throw
1.  Edit tests in inputValue.spec.ts to expect tx.hash() to throw

### DIFF
--- a/packages/tx/src/legacyTransaction.ts
+++ b/packages/tx/src/legacyTransaction.ts
@@ -258,20 +258,10 @@ export default class Transaction extends BaseTransaction<Transaction> {
    * Use {@link Transaction.getMessageToSign} to get a tx hash for the purpose of signing.
    */
   hash(): Buffer {
-    // In contrast to the tx type transaction implementations the `hash()` function
-    // for the legacy tx does not throw if the tx is not signed.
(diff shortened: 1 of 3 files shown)


ScottyPoi:
Not sure if the changes made to tests inputValue.spec.ts were the right move, or if they should be deleted or changed.  With the changes to tx.hash(), those tests now fail.  I changed them to expect that Error, but that failure is tested in legacy.spec.ts, so doing it again in this test is redundant.  Do we like we're losing something by deleting the two tests in inputValue.spec.ts?

holgerd77 (commented):
Nice, from my side this can get an approval, really clean "catches" of the respective areas to adopt and/or delete the code, I would say the way the tests are adopted are also ok.

Will put this as "Blocked" for now so that we can first merge in #1892 before anything else on `develop`.

jochem-brouwer (approved):
LGTM! @holgerd77 I think we can un-block this now, right?

holgerd77 (approved):
LGTM

Issue #1056, opened 2020-04-09
Trie: strictly require level to be passed in, drop level deps


mikeal:
I’m seeing some nastly browser bundling issues because this package is in my dep tree and depends on modules which have ancient versions of level and there have been several breaking changes since.

These level deps are pretty massive and unnecessary deps since they only get used when a level instance is not passed in. If you moved to require this explicitly you could drop the deps.

In the meantime I’m going to have to drop ethereum support from my default set so that I can avoid this issue.

holgerd77:
Should be solved or at least substantially mitigated by #1917 (to be released along upcoming VM v6 Breaking Release round), will close.

PR #1895, opened 2022-05-18
refactor(trie): abstract database implementation with an interface

faustbrian:
@holgerd77 This is an initial draft PR for discussion https://github.com/ethereumjs/ethereumjs-monorepo/discussions/1886 to see if this structure makes sense or not. A few notes since this is a draft:

- `CheckpointDB` now implements the `DB` interface which makes it fully compliant with the underlying `DB` storage that is used. It already was almost fully compliant because it was only missing the `copy` function.
- The `CheckpointTrie` uses an instance of `CheckpointDB` which in turn uses an instance of `DB` _(which is now an interface)_ - Do we want to copy the `CheckpointDB` instance when we call `copy` or just the underlying `DB` instance? The same question applies to the `SecureTrie` since it extends `CheckpointTrie`
- Do we still want to provide a default database? This might be a good time to make `db` a requirement so that `this.db = opts.db ?? new LevelDB()` would become `this.db = opts.db` and the user will always have to do `new Trie(new TheirDatabase())`. This would remove the need for assumption making and the `LevelDB` implementation would still be exported so the user can always use that if they wish to do so. **The only concern here is that a few places have no access to `opts` and will default to an in-memory database. This could be worked around by explicitly passing a `LevelDB` instance for such cases. See commit https://github.com/ethereumjs/ethereumjs-monorepo/pull/1895/commits/2642f92d252fc8319f5e29e74e8c204142f429ca for this.**

Once the questions are answered I'll work on finishing this and fixing tests.

### DIFF
--- a/packages/block/src/block.ts
+++ b/packages/block/src/block.ts
@@ -1,4 +1,4 @@
-import { BaseTrie as Trie, LevelDB } from 'merkle-patricia-tree'
+import { BaseTrie as Trie } from 'merkle-patricia-tree'
 import { keccak256 } from 'ethereum-cryptography/keccak'
 import { arrToBufArr, bufArrToArr, KECCAK256_RLP, bufferToHex } from 'ethereumjs-util'
 import RLP from 'rlp'
@@ -21,7 +21,7 @@ export class Block {
   public readonly header: BlockHeader
   public readonly transactions: TypedTransaction[] = []
   public readonly uncleHeaders: BlockHeader[] = []
-  public readonly txTrie = new Trie({ db: new LevelDB() })
+  public readonly txTrie = new Trie()
   public readonly _common: Common
 
   /**
--- a/packages/client/lib/rpc/modules/engine.ts
+++ b/packages/client/lib/rpc/modules/engine.ts
@@ -2,7 +2,7 @@ import { Block, HeaderData } from '@ethereumjs/block'
 import { TransactionFactory, TypedTransaction } from '@ethereumjs/tx'
 import { bufferToHex, toBuffer, zeros } from 'ethereumjs-util'
 import RLP from 'rlp'
-import { BaseTrie as Trie, LevelDB } from 'merkle-patricia-tree'
+import { BaseTrie as Trie } from 'merkle-patricia-tree'
 import { Hardfork } from '@ethereumjs/common'
 
 import { middleware, validators } from '../validation'
@@ -129,7 +129,7 @@ const recursivelyFindParents = async (vmHeadHash: Buffer, parentHash: Buffer, ch
  * Returns the txs trie root for the block.
  */
 const txsTrieRoot = async (txs: TypedTransaction[]) => {
-  const trie = new Trie({ db: new LevelDB() })
+  const trie = new Trie()
   for (const [i, tx] of txs.entries()) {
     await trie.put(Buffer.from(RLP.encode(i)), tx.serialize())
   }
--- a/packages/client/lib/util/parse.ts
+++ b/packages/client/lib/util/parse.ts
@@ -3,7 +3,7 @@ import { Multiaddr, multiaddr } from 'multiaddr'
 import { BlockHeader } from '@ethereumjs/block'
 import Common, { Hardfork } from '@ethereumjs/common'
 import { keccak256 } from 'ethereum-cryptography/keccak'
-import { LevelDB, SecureTrie as Trie } from 'merkle-patricia-tree'
+import { SecureTrie as Trie } from 'merkle-patricia-tree'
 import {
   Account,
   addHexPrefix,
@@ -95,7 +95,7 @@ export function parseTransports(transports: string[]) {
  * @returns genesis storage trie
  */
 async function createStorageTrie(storage: any) {
-  const trie = new Trie({ db: new LevelDB() })
+  const trie = new Trie()
   for (const [address, value] of Object.entries(storage) as unknown as [string, string]) {
     const key = isHexPrefixed(address) ? toBuffer(address) : Buffer.from(address, 'hex')
     const val = Buffer.from(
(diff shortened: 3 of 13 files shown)


holgerd77:
Hi there,
just having a first look, dropping in here https://github.com/ethereumjs/ethereumjs-monorepo/issues/1056 for some additional eventual reflection and consideration points, since the discussion is a bit related.

holgerd77:
> Do we still want to provide a default database? This might be a good time to make db a requirement so that this.db = opts.db ?? new LevelDB() would become this.db = opts.db and...

I guess I would be a fan of making DB a requirement, other thoughts?

Would this then also eventually allow to get rid of the `level` (or related) dependency in the package, as requested by the issue above? Not completely seeing through on all requirements on e.g. typing yet.

holgerd77:
> CheckpointDB now implements the DB interface which makes it fully compliant with the underlying DB storage that is used. It already was almost fully compliant because it was only missing the copy function.

👍 

General thought: not sure, but do you eventually want to directly implement an LMDB driver along this work? Don't want to lightheartedly load extra work on you, but this would be helping a lot to verify that this proposed design will work under several (somewhat: arbitrary) DB usage contexts.

(would not need to be part of the PR, not sure, or might this be even part of the package (only if no additional dependencies I needed I guess)?)

holgerd77:
> Do we want to copy the CheckpointDB instance when we call copy or just the underlying DB instance? 

I have a bit a hard time to grasp the structure (which doesn't mean it is necessarily wrong, bad or unlogical 😋), what is the main reasoning for switching from extending DB to implementing DB when then all DB functions need to be called anyhow/nevertheless?

holgerd77:
Ok, first round on this, thanks so much, this looks already fantastic and will be a great benefit for a lot of use cases (users) I am very much sure! 🎉 🙏 ❤️

holgerd77:
@faustbrian Thanks a lot for the continued work here 🙏 ❤️, sorry, I am somewhat side tracked with personal stuff right now, I hope I find some in between time though to give this some continued looks!

holgerd77:
> I'll take a look at the E2E errors later today or tomorrow, not sure what causes those.

Ah, the E2E errors test against an external Hardhat version and these are generally broken for some time on `develop`. There are so many breaking changes already in that this is very much to be expected. So no need to tackle. 🙂

holgerd77:
> I think other than [#1895 (comment)](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1895#issuecomment-1133884947) and [#1895 (comment)](https://github.com/ethereumjs/ethereumjs-monorepo/pull/1895#discussion_r878989723) this should be ready for a review.

Super cool! 🙏 😄

packages/block/src/block.ts
```diff
@@ -21,7 +21,7 @@ export class Block {
   public readonly header: BlockHeader
   public readonly transactions: TypedTransaction[] = []
   public readonly uncleHeaders: BlockHeader[] = []
-  public readonly txTrie = new Trie()
+  public readonly txTrie = new Trie({ db: new LevelDB() })
```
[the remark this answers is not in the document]
holgerd77: Yes, maybe switch those directly?

packages/trie/test/checkpoint.spec.ts
```diff
@@ -1,180 +1,182 @@
...
+  tape('testing checkpoints', function (tester) {
+    const it = tester.test
+
+    let trie: CheckpointTrie
+    let trieCopy: CheckpointTrie
+    let preRoot: String
+    let postRoot: String
+
+    it('setup', async function (t) {
+      trie = new CheckpointTrie({ db: new DB() })
```
holgerd77, line 15: Sorry, it will likely be in the code and easy to read out, but just (also) to have the explicit conversation: what happens when this generic `DB` class is used? Does this default to something in particular? The memory DB?

Then I wonder why we are not defaulting to this in the first place since this would be also some kind of implicit defaulting? 🤔

Or don't I get the whole picture?

holgerd77 (commented):
Two or three comments or questions, but this looks tremendously great in general, wohoo!!! 💯 👍

One a bit hard request: we have a strict requirement to only update feature branches towards `develop` via rebase, since otherwise the whole `devlop` -> `master` rebase can get very hazzly and we want to keep the branches as clean as possible, since this round of breaking changes is just massive.

I wouldn't want to deviate from this policy especially so late in the release process, so could you re-submit a rebased version? 😬 Not sure, maybe it's just possible to remove the merge commit via interactive rebase (definitely make a branch backup before) and then "normally" rebase the branch. Otherwise this would need to get a "cherry-pick one-by-one" process (which is often less of a hazzle - less work - than it sounds).

Sorry for the inconvenience and thanks - hopefully 🙂 - for your understanding!

holgerd77:
(this rebase thing can also be done latest in the process once everything else is settled unequivocally)

packages/trie/test/checkpoint.spec.ts
holgerd77: > DB in this case here is just an alias for the concrete implementation that is being tested

Ah, thanks didn't see the broader context.

> We could default to the MemoryDB if that's what you prefer.

I think I do have some preference on defaulting to MemoryDB after thinking about it twice, sorry for the extra work this undecisiveness might eventually cause. My feeling is that this implicit behavior is just "very innocent" - by not using any DB at all and just stay "in code". It is very unlikely that this behavior would every change (to - e.g. - always start with a full on-disk level DB or something) and for using an on-disk DB one can use the option explicitly. And at the same time things get a bit more innocent if people just want to play around with the Trie library without having the need to always look up this `db` constructor usage.

But let's wait, I might be unnecessarily strong opinionated, and eventually in a wrong direction. I will post this within the team chat and collect some feedback from the team. Will come back here if there is some strong tendency in the one or the other direction.

holgerd77:
> Will take a look at the rebase when it's approved. Used to squashing PRs on other projects so didn't consider that 😅

Thanks! 🙂 

We'll squash here too, but then we would also squash in the changes included in the merge commit, which we do not want since it can cause problems on the rebase of the squashed develop PR commits towards master.

(let me know if you think this is logically flawed in some way, always eager to learn. 🙂 We wouldn't change on that nevertheless, since we are only 3-4 PRs away from the final releases and I wouldn't change *anything* on our established processes so late, my respect is just very high since this branch really became a "monster" 😜)

packages/trie/test/checkpoint.spec.ts
holgerd77: I reached out to the team and have unanimous feedback that they want to keep a default value here to make things simpler for beginners / experimental usages.

Can you readopt? Sorry that this has caused some extra work.

PR #1900, opened 2022-05-23
VM/tests: ensure verifyPostConditions works

jochem-brouwer:
If a blockchain test fails and we have the `--debug` flag on, we expect that we do `verifyPostConditions`: https://github.com/ethereumjs/ethereumjs-monorepo/blob/0b7cc1b15d1fd9d58e4ae7db794674c3af4da5ae/packages/vm/tests/tester/runners/BlockchainTestsRunner.ts#L171

However, this actually does not work. If the block is invalid and produces a different `stateRoot` than whatever we expect, then `runBlockchain` actually throws and the `verifyPostConditions` is never ran. 

This PR also updates the formatter such that it is easier to read what goes wrong. On a VM which I intentially made producing wrong output:

Screenshot from 2022-05-23 23-21-05

(Was done by adding this to `runBlock.ts`:)

Screenshot from 2022-05-23 23-25-17

If the tests contain the postConditions we can now very quickly figure out what goes wrong if a test files :smile:

### DIFF
--- a/packages/vm/tests/util.ts
+++ b/packages/vm/tests/util.ts
@@ -70,7 +70,7 @@ export function dumpState(state: any, cb: Function) {
   })
 }
 
-export function format(a: any, toZero: boolean = false, isHex: boolean = false) {
+export function format(a: any, toZero: boolean = false, isHex: boolean = false): Buffer {
   if (a === '') {
     return Buffer.alloc(0)
   }
@@ -91,6 +91,10 @@ export function format(a: any, toZero: boolean = false, isHex: boolean = false) 
   }
 
   return a
+}
+
+function formatNumberString(input: Buffer): string {
+  return new BN(input).toString()
 }
 
 /**
@@ -174,8 +178,16 @@ export function verifyAccountPostConditions(
 ) {
   return new Promise<void>((resolve) => {
     t.comment('Account: ' + address)
-    t.ok(format(account.balance, true).equals(format(acctData.balance, true)), 'correct balance')
-    t.ok(format(account.nonce, true).equals(format(acctData.nonce, true)), 'correct nonce')
+    t.equals(
+      formatNumberString(format(account.balance, true)),
+      formatNumberString(format(acctData.balance, true)),
+      'correct balance'
+    )
+    t.equals(
+      formatNumberString(format(account.nonce, true)),
+      formatNumberString(format(acctData.nonce, true)),
+      'correct nonce'
+    )
 
     // validate storage
     const origRoot = state.root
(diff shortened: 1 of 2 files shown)


jochem-brouwer:
Updated this one, sample messages:

```
# Account: 0x8888f1f195afa192cfee860698584c030f4c9db1
# Expected balance of 30312500000000021000, but got 30312500000000020999
# Expected nonce of 0, but got 1
# Account: 0x0000000000000000000000000000000000000011
# Account: 0xb94f5374fce5edbc8e2a8697c15331677e6ebf0b
# Account: 0xb94f5374fce5ed0000000097c15331677e6ebf0b
# Account: 0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b
# Expected storage key 0x290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563 to have value 0x, but got 0x01}
```

There are two gotcha's in the diffs, will comment.

packages/vm/tests/util.ts
```diff
@@ -144,14 +144,14 @@ export async function verifyPostConditions(state: any, testData: any, t: tape.Te
         const promise = verifyAccountPostConditions(state, address, account, testData, t)
         queue.push(promise)
       } else {
-        t.fail('invalid account in the trie: ' + <string>key)
+        t.comment('invalid account in the trie: ' + <string>key)
```
jochem-brouwer, line 147: Converted all tape commands to `comment`. The reason is that if you have failing tests, then this adds to the number of tests counter. Let's say we accidentally do not run some tests, then `--verify-amount-alltests` flag could still pass while this is actually not the case.

Also note that if any of these comments get triggered, then the state root of the block is wrong, so the test fails anyways.

packages/vm/tests/util.ts
```diff
@@ -188,38 +195,40 @@ export function verifyAccountPostConditions(
...
+          : acctData.storage['0x']
+        delete acctData.storage['0x']
+      }
 
-        t.equal(val, hashedStorage[key], 'correct storage value')
-        delete hashedStorage[key]
-      })
+      if (val !== hashedStorage[key]) {
+        t.comment(
+          `Expected storage key 0x${data.key.toString('hex')} at address ${address} to have value ${
```
jochem-brouwer, line 214: Have to add address here since it is not clear when these promises are fulfilled.

ryanio (approved):
nice, lgtm!

holgerd77:
Updated this via rebase (via Web UI).

holgerd77 (approved):
LGTM

PR #1901, opened 2022-05-23
CI: ensure all hardforks and transitions run when there is a "test all hardforks" label on PR

### LINKED ISSUE #1899: Test-all-hardforks label: ensure all hardforks are indeed tested
PRs with the `test-all-hardforks` label should test all hardforks (including transitions tests). This is currently not the case.

jochem-brouwer:
Closes #1899 

This PR:

- Adds a new test command to VM: `npm run test:blockchain:transitionTests`, this tests all the (small) transition tests
- Updates the right test amounts for transition tests
- Ensures that these transition tests are also ran when running all hardforks

There were two bugs found:

1. Consensus bug: it is possible to import a London block which does not have a base fee set.
2. Fixes test runner, transition tests failed to run since they were always lowercased which resulted in them not being found.


packages/vm/tests/tester/config.ts
```diff
@@ -367,8 +367,8 @@ const expectedTestsFull: any = {
     Berlin: 33,
     ByzantiumToConstantinopleFixAt5: 3,
     EIP158ToByzantiumAt5: 3,
-    FrontierToHomesteadAt5: 12,
-    HomesteadToDaoAt5: 18,
+    FrontierToHomesteadAt5: 13,
+    HomesteadToDaoAt5: 32,
     HomesteadToEIP150At5: 3,
     BerlinToLondonAt5: 0,
```
jochem-brouwer, line 373: This is not 0 and should also be fixed

.github/workflows/vm-pr.yml
```diff
@@ -163,6 +163,8 @@ jobs:
 
       - run: npm run test:blockchain -- ${{ matrix.args }}
 
+      - run: npm run test:blockchain:transitionForks
```
jochem-brouwer, line 166: How come this doesn't trigger the CI?

ryanio: this would be run on every matrix hardfork run, is that what you want?

> How come this doesn't trigger the CI?

not sure what you mean? if you want just one added job you could add to the matrix.args array

packages/block/src/header.ts
```diff
@@ -184,6 +184,11 @@ export class BlockHeader {
       throw new Error('invalid header. Less values than expected were received')
     }
 
+    if (opts.common?.isActivatedEIP(1559) && !baseFeePerGas) {
```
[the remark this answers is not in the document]
jochem-brouwer: This should fix it

.github/workflows/vm-pr.yml
jochem-brouwer: I don't want to add it to the matrix, since these are super small jobs and they will eat up the concurrent jobs which can run. I have added this new npm rule whichruns these tests (in total it are just 78 tests), but I don't want to spawn 6 micro jobs for those. I'd say that adding this line should add all forks but apparently not? Or does it actually trigger this job multiple times each time after it has run a normal test? So it would first run the London tests and then do the transition forks tests..? (Which is also not clean :( )

packages/common/src/index.ts
```diff
@@ -882,6 +882,23 @@ export default class Common extends EventEmitter {
     return new BN(block)
   }
 
+  /**
+   * Returns the hardfork change block for eip
+   * @param eip EIP number
+   * @returns Block number or null if unscheduled
+   */
+  eipBlockBN(eip: number): BN | null {
```
jochem-brouwer, line 890: How should we handle custom chains here..?

packages/block/src/header.ts
```diff
@@ -184,9 +184,11 @@ export class BlockHeader {
       throw new Error('invalid header. Less values than expected were received')
     }
 
-    if (opts.common?.isActivatedEIP(1559) && !baseFeePerGas) {
-      // TODO fixme
-      // throw new Error('invalid header. baseFeePerGas should be provided')
+    if (opts.common?.isActivatedEIP(1559) && (baseFeePerGas === undefined)) {
+      const eip1559ActivationBlock = opts.common?.eipBlockBN(1559)?.toArrayLike(Buffer)
```
ryanio, line 188: you don't need to add `eipBlockBN` here, you can use hardforkBlockBN like on L275:

https://github.com/ethereumjs/ethereumjs-monorepo/blob/2b58e75bfbb9476fe179ba1bf5b6a5f16dff4bd1/packages/block/src/header.ts#L273-L289

so instead of setting default 7 you want to throw? that would change a lot of internal uses where we simplify to a sensible default on blank init

.github/workflows/vm-pr.yml
ryanio: oh the transition forks are already listed at the end of the allForks command, which is run in the VM nightly. i think the reason we limited the ones run on PR to keep the number of jobs down to the amount we can run concurrently at once.

jochem-brouwer: Keep in mind this is only triggered if you add `test all hardforks` label. That's only done on PRs which change EVM logic, which is not the case for most PRs. So I think it is fine to add this here?

packages/block/src/header.ts
jochem-brouwer: Users have to provide the explicit `initialBaseFee` as `baseFee` in the london transition block in case they use `fromValuesArray` or `fromRLP`. Most users who want to construct a block will use `fromData`. If we would remove this change from the `fromRLP`, then we have this problem that someone can feed us an initial London block which has no baseFee in the RLP, which is invalid, and can then start building upon that wrong chain.

Ah OK regarding the `eipBlockBN`, I see that it is fixed by just getting the London fork. But this is a bit inconsistent, right? If you want a custom chain with 1559 but no other EIPs then `London` fork is at block `null`, it is never activated. I think this is rather inconsistent. It is also weird that we cannot check whenever an EIP is activated in Common, which I think we should add as feature.

.github/workflows/vm-pr.yml
ryanio: yeah sure, I can go with that. So I think what you want to do here is generalize the test run command above this line

```run: npm run test:blockchain -- ${{ matrix.args }}```

to

```run: npm run ${{ matrix.args }}```

then you can add testing the transition forks as one more matrix argument, so it will just spawn one job for it. the code right now would run it on every matrix run.

File: packages/block/src/header.ts (hunk 5 above)
Comment (ryanio) [reply packages/block/src/header.ts:188]: yeah I can go along with that, if baseFee is not supplied in fromValuesArray then we can error. in that case we probably don't need the initial base fee logic, since if we are building the block we would use fromBlockData and if we are getting it from peers we assume it to be fully correct.

eipBlockBN, yeah it just gets complicated to support that feature, I think we need a refactoring where we set everything up front and don't do custom chain iterations in any get() method (to make the operation quicker). didn't you start on some kind of refactoring like that but it got complicated?

File: .github/workflows/vm-pr.yml (hunk 2 above)
Comment (jochem-brouwer) [reply .github/workflows/vm-pr.yml:166]: Oh yeah of course :man_facepalming: Thanks :smile: :+1:

**Review by acolytec3 — APPROVED** [review]
LGTM

File: packages/block/src/header.ts [hunk 6]
```diff
@@ -135,6 +135,15 @@ export class BlockHeader {
...
     }
 
+    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+    if (opts.common?.isActivatedEIP(1559) && baseFeePerGas === undefined) {
+      const eip1559ActivationBlock = bigIntToBuffer(opts.common?.eipBlock(1559)!)
+      // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+      if (eip1559ActivationBlock && eip1559ActivationBlock.equals(number)) {
+        throw new Error('invalid header. baseFeePerGas should be provided')
+      }
+    }
```
holgerd77, line 145: I find this code part extremely hacky (only for this one activation block, check only in the value constructor, endless linting rule exceptions).

Is this really necessary in the current form, or can we at least improve here?

PR #1878, opened 2022-05-05
client: optimistic (beacon) sync

ryanio:
This PR adds [optimistic](https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md) (beacon) sync to the client, where blocks are downloaded in reverse from the announced trusted head.

Thanks to go-ethereum for the skeleton design and test cases.

You can test with the instructions in `packages/client/kiln/README.md` with these new params for lodestar: `--weakSubjectivityServerUrl https://lodestar-kiln.chainsafe.io --weakSubjectivitySyncLatest`

Thanks to @g11tech for following this work and providing help and commits along the way!

PR #1902, opened 2022-05-23
client: small engine updates

ryanio:
This PR includes some small engine fixes and nits, following up after #1878, and some tx pool fixes after #1852.

### DIFF
--- a/packages/client/lib/blockchain/chain.ts
+++ b/packages/client/lib/blockchain/chain.ts
@@ -83,9 +83,6 @@ export class Chain {
   public chainDB: LevelUp
   public blockchain: Blockchain
   public opened: boolean
-  public terminalPoWBlock: Block | undefined
-  public transitionPoSBlock: Block | undefined
-  public lastFinalizedBlockHash: Buffer | undefined
 
   private _headers: ChainHeaders = {
     latest: null,
--- a/packages/client/lib/execution/vmexecution.ts
+++ b/packages/client/lib/execution/vmexecution.ts
@@ -133,9 +133,7 @@ export class VMExecution extends Execution {
    * @returns number of blocks executed
    */
   async run(): Promise<number> {
-    if (this.running) {
-      return 0
-    }
+    if (this.running) return 0
     this.running = true
     let numExecuted: number | undefined
 
@@ -159,11 +157,9 @@ export class VMExecution extends Execution {
       this.vmPromise = blockchain.iterator(
         'vm',
         async (block: Block, reorg: boolean) => {
-          if (errorBlock) {
-            return
-          }
+          if (errorBlock) return
           // determine starting state for block run
-          // if we are just starting or if a chain re-org has happened
+          // if we are just starting or if a chain reorg has happened
           if (!headBlock || reorg) {
             const parentBlock = await blockchain.getBlock(block.header.parentHash)
             parentState = parentBlock.header.stateRoot
--- a/packages/client/lib/service/txpool.ts
+++ b/packages/client/lib/service/txpool.ts
@@ -16,7 +16,7 @@ import type { Block } from '@ethereumjs/block'
 
 // Configuration constants
 const MIN_GAS_PRICE_BUMP_PERCENT = 10
-const MIN_GAS_PRICE = new BN(1000000000) // 1 GWei
+const MIN_GAS_PRICE = new BN(100000000) // .1 GWei
 const TX_MAX_DATA_SIZE = 128 * 1024 // 128KB
 const MAX_POOL_SIZE = 5000
 const MAX_TXS_PER_ACCOUNT = 100
@@ -72,6 +72,7 @@ export class TxPool {
   public running: boolean
 
   /* global NodeJS */
+  private _cleanupInterval: NodeJS.Timeout | undefined
   private _logInterval: NodeJS.Timeout | undefined
 
   /**
@@ -178,6 +179,10 @@ export class TxPool {
     if (this.running) {
       return false
     }
+    this._cleanupInterval = setInterval(
+      this.cleanup.bind(this),
+      this.POOLED_STORAGE_TIME_LIMIT * 1000 * 60
+    )
     this._logInterval = setInterval(this._logPoolStats.bind(this), this.LOG_STATISTICS_INTERVAL)
     this.running = true
     this.config.logger.info('TxPool started.')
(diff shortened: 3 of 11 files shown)


packages/client/lib/blockchain/chain.ts
```diff
@@ -83,9 +83,6 @@ export class Chain {
   public chainDB: LevelUp
   public blockchain: Blockchain
   public opened: boolean
-  public terminalPoWBlock: Block | undefined
-  public transitionPoSBlock: Block | undefined
-  public lastFinalizedBlockHash: Buffer | undefined
```
ryanio, line 88: these were not being used so I cleaned them up for now

packages/client/lib/execution/vmexecution.ts
```diff
@@ -115,6 +115,7 @@ export class VMExecution extends Execution {
    * @param blocks Array of blocks to save pending receipts and set the last block as the head
    */
   async setHead(blocks: Block[]): Promise<void> {
+    await this.chain.blockchain.setIteratorHead('vm', blocks[blocks.length - 1].hash())
```
jochem-brouwer, line 118: Does moving this thing a fine lines before change anything?

packages/client/lib/execution/vmexecution.ts
```diff
@@ -278,7 +273,7 @@ export class VMExecution extends Execution {
       canonicalHead = await this.vm.blockchain.getLatestBlock()
     }
     this.running = false
-    return numExecuted as number
+    return numExecuted ?? 0
```
jochem-brouwer, line 276: Can't we default https://github.com/ethereumjs/ethereumjs-monorepo/blob/31391d176b8573939962046a9c8f45d8b9baf1c6/packages/client/lib/execution/vmexecution.ts#L137 to just set it diretly to 0? And also remove the `numExecuted === undefined` check in the loop?

packages/client/lib/service/txpool.ts
```diff
@@ -225,15 +230,11 @@ export class TxPool {
...
-    const newGasPrice = this.txGasPrice(addedTx)
-    const minTipCap = existingTxGasPrice.tip.add(
-      existingTxGasPrice.tip.muln(MIN_GAS_PRICE_BUMP_PERCENT).divn(100)
-    )
-    const minFeeCap = existingTxGasPrice.maxFee.add(
-      existingTxGasPrice.maxFee.muln(MIN_GAS_PRICE_BUMP_PERCENT).divn(100)
-    )
-    if (newGasPrice.tip.lt(minTipCap) || newGasPrice.maxFee.lt(minFeeCap)) {
+    const { maxFee: existingMaxFee, tip: existingTip } = this.txGasPrice(existingTx)
+    const { maxFee: newMaxFee, tip: newTip } = this.txGasPrice(addedTx)
```
jochem-brouwer, line 234: Ah great yes this is much clearer :+1:

packages/client/lib/service/txpool.ts
```diff
@@ -283,27 +282,27 @@ export class TxPool {
...
     }
-    const block = await this.service.chain.getLatestHeader()
-    if (block.baseFeePerGas) {
-      if (currentGasPrice.maxFee.lt(block.baseFeePerGas)) {
-        throw new Error(
-          `Tx cannot pay basefee of ${block.baseFeePerGas}, have ${currentGasPrice.maxFee}.`
-        )
-      }
+    const latest = this.service.chain.headers.latest!
+    if (latest.baseFeePerGas && maxFee.lt(latest.baseFeePerGas.divn(2)) && !isLocalTransaction) {
```
jochem-brouwer, line 286: I have matched Geth's logic here, Geth rejects txs which cannot pay base fee. I think this is OK though :+1:

packages/client/lib/service/txpool.ts
```diff
@@ -645,11 +635,11 @@ export class TxPool {
       )
       if (!txsSortedByNonce[0].nonce.eq(nonce)) {
         // Account nonce does not match the lowest known tx nonce,
-        // therefore no txs from this address are currently exectuable
+        // therefore no txs from this address are currently executable
         continue
       }
       if (baseFee) {
-        // If any tx has an insiffucient gasPrice,
+        // If any tx has an insufficient gasPrice,
```
jochem-brouwer, line 642: I should definitely install a spell checker :smile:

packages/client/test/integration/fullethereumservice.spec.ts
```diff
@@ -69,8 +69,12 @@ tape('[Integration:FullEthereumService]', async (t) => {
     peer.eth!.send('NewBlock', [block, new BN(1)])
 
     const txData =
-      '0x02f90108018001018402625a0094cccccccccccccccccccccccccccccccccccccccc830186a0b8441a8451e600000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000f85bf859940000000000000000000000000000000000000101f842a00000000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000060a701a0afb6e247b1c490e284053c87ab5f6b59e219d51f743f7a4d83e400782bc7e4b9a0479a268e0e0acd4de3f1e28e4fac2a6b32a4195e8dfa9d19147abe8807aa6f64'
+      '0x02f901100180843b9aca00843b9aca008402625a0094cccccccccccccccccccccccccccccccccccccccc830186a0b8441a8451e600000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000f85bf859940000000000000000000000000000000000000101f842a00000000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000060a701a0afb6e247b1c490e284053c87ab5f6b59e219d51f743f7a4d83e400782bc7e4b9a0479a268e0e0acd4de3f1e28e4fac2a6b32a4195e8dfa9d19147abe8807aa6f64'
```
jochem-brouwer, line 72: It wasn't there before, but maybe it's a good idea to drop a quick comment how this tx was generated?

jochem-brouwer (commented):
Great cleanup, some questions/comments.

packages/client/lib/execution/vmexecution.ts
ryanio: Yeah this is so when beaconsync gets CHAIN_UPDATED event from the chain.putBlocks, the vm execution will see the head matches the chain head and return with no further execution.

This is safe because the `runWithoutSetHead` already inserts the block in the blockchain db, so this method can safely reference that block as the new iterator head.

packages/client/lib/execution/vmexecution.ts
ryanio: well not really because if 0 is returned (0 blocks executed) it should return and stop execution, not re-enter the while loop and there is nothing left to do.

packages/client/lib/service/txpool.ts
ryanio: ah ok, thanks yeah I just wanted some flexiblity, if we are in a temporary time of very high basefee it still makes sense to accept at a lower fee because it can eventually go back down, and strictly validating at the current basefee would probably miss a good amount of txs? perhaps geth does it to avoid a DDOS vector where during high basefee people flood with low priced txs which can drown out the higher fee txs.

packages/client/test/integration/fullethereumservice.spec.ts
ryanio: heh i'm not sure the source tx actually, i just updated it to use a higher fee to pass our increased tx pool validation checks.

packages/client/lib/service/txpool.ts
```diff
@@ -283,27 +282,27 @@ export class TxPool {
...
       }
     }
-    const block = await this.service.chain.getLatestHeader()
-    if (block.baseFeePerGas) {
-      if (currentGasPrice.maxFee.lt(block.baseFeePerGas)) {
-        throw new Error(
-          `Tx cannot pay basefee of ${block.baseFeePerGas}, have ${currentGasPrice.maxFee}.`
-        )
-      }
+    const latest = this.service.chain.headers.latest!
```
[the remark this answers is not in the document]
ryanio: it's a bit difficult to tell which line you are commenting about. are you talking about `this.service.chain.getLatestHeader()` -> `this.service.chain.headers.latest!`? if so, yes this is an optimization to reduce the need for a round trip db call and instead use the cached value from the chain class.

if you are talking about the nested ifs combined to one statement, yeah just a simplification for easier readability and less nesting.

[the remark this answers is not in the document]
jochem-brouwer: I was talking about the first, thanks for clarifying :smile:

packages/client/lib/execution/vmexecution.ts
jochem-brouwer: Yup that makes sense!

jochem-brouwer (approved):
LGTM after @ryanio addressed my questions! :smile:

PR #1905, opened 2022-05-25
Util: Allow `v` to be `0` or `1` for EIP1559 transactions

### LINKED ISSUE #1904: Util: Allow `v` to be `0` or `1` for EIP1559 transactions
## Description

Currently, there are 4 functions using the `calculateSigRecovery` method, but it only allows for values to be `27`, `28` or `chainId * 2 + (35 | 36)`, introducing incompatibility for EIP1559 signed transactions, which is avoidable by just extracting the `yParity` from the `v` value (as stated in https://github.com/ethereumjs/ethereumjs-monorepo/issues/1597).

The so-called functions are:

- ecrecover
- toRpcSig
- toCompactSig
- isValidSignature

However, the hack of extracting the `yParity` out of `v` works in most of the cases except for the `toRpcSign` function, which creates its result from the `v` param, making impossible to just `toRpcSign(v - 27, r, s)`, since subtracting 27 actually changes the result.

Another related problem is that the `fromRpcSig` function might receive any signature but if `v` is `< 27`, it'll add 27 anyway. This is fixable once called since you can just subtract the `27` if you know what you're doing, but I think the docs will need at least to state that the function doesn't have enough information to know if adding `27` or not.

## Proposed solution

- Update the `calculateSigRecovery` function to allow `0` and `1` values
- Update `fromRpcSig` docs to state that is adding `27` for `< 27` values in any case

ernestognw:
Fixes #1904

### DIFF
--- a/packages/tx/src/legacyTransaction.ts
+++ b/packages/tx/src/legacyTransaction.ts
@@ -393,7 +393,6 @@ export default class Transaction extends BaseTransaction<Transaction> {
     // No unsigned tx and EIP-155 activated and chain ID included
     if (
       v !== undefined &&
-      !v.eqn(0) && // TODO fixme -> why overwrite/return default common when `v` is set to 0?
       (!common || common.gteHardfork('spuriousDragon')) &&
       !v.eqn(27) &&
       !v.eqn(28)
--- a/packages/tx/src/legacyTransaction.ts
+++ b/packages/tx/src/legacyTransaction.ts
@@ -136,18 +136,6 @@ export default class Transaction extends BaseTransaction<Transaction> {
       }
     }
 
-    if (!this.supports(Capability.EIP155ReplayProtection)) {
-      // In case of a signed non-EIP155 tx, `ecrecover` does not get a `chainId` provided
-      // Internally, `ecrecover` will calculate the recovery bit by `v = v - 27`
-      // This recovery bit is either 0 or 1. Thus, any other value than `v = 27` or `v = 28` will
-      // throw upon calling `getSenderAddress` here. Thus throw early in the constructor
-      if (this.v !== undefined) {
-        if (!this.v.eqn(27) && !this.v.eqn(28)) {
-          throw new Error(`Non-EIP155 txs need either v = 27 or v = 28, got v = ${this.v}`)
-        }
-      }
-    }
-
     if (this.common.isActivatedEIP(3860)) {
       checkMaxInitCodeSize(this.common, this.data.length)
     }
@@ -390,6 +378,17 @@ export default class Transaction extends BaseTransaction<Transaction> {
    * Validates tx's `v` value
    */
   private _validateTxV(v?: BN, common?: Common): Common {
+    // Check for valid v values in the scope of a signed legacy tx
+    if (v !== undefined) {
+      // v is 1. not matching the EIP-155 chainId included case and...
+      // v is 2. not matching the classic v=27 or v=28 case
+      if (v.ltn(37) && !v.eqn(27) && !v.eqn(28)) {
+        throw new Error(
+          `Legacy txs need either v = 27/28 or v >= 37 (EIP-155 replay protection), got v = ${v}`
+        )
+      }
+    }
+
     let chainIdBN
     // No unsigned tx and EIP-155 activated and chain ID included
     if (
--- a/packages/tx/src/legacyTransaction.ts
+++ b/packages/tx/src/legacyTransaction.ts
@@ -136,6 +136,18 @@ export default class Transaction extends BaseTransaction<Transaction> {
       }
     }
 
+    if (!this.supports(Capability.EIP155ReplayProtection)) {
+      // In case of a signed non-EIP155 tx, `ecrecover` does not get a `chainId` provided
+      // Internally, `ecrecover` will calculate the recovery bit by `v = v - 27`
+      // This recovery bit is either 0 or 1. Thus, any other value than `v = 27` or `v = 28` will
+      // throw upon calling `getSenderAddress` here. Thus throw early in the constructor
+      if (this.v !== undefined) {
+        if (!this.v.eqn(27) && !this.v.eqn(28)) {
+          throw new Error(`Non-EIP155 txs need either v = 27 or v = 28, got v = ${this.v}`)
+        }
+      }
+    }
+
     if (this.common.isActivatedEIP(3860)) {
       checkMaxInitCodeSize(this.common, this.data.length)
     }
@@ -382,7 +394,7 @@ export default class Transaction extends BaseTransaction<Transaction> {
     // No unsigned tx and EIP-155 activated and chain ID included
     if (
       v !== undefined &&
-      !v.eqn(0) &&
+      !v.eqn(0) && // TODO fixme -> why overwrite/return default common when `v` is set to 0?
       (!common || common.gteHardfork('spuriousDragon')) &&
       !v.eqn(27) &&
       !v.eqn(28)
--- a/packages/util/src/signature.ts
+++ b/packages/util/src/signature.ts
@@ -46,6 +46,9 @@ export function ecsign(msgHash: Buffer, privateKey: Buffer, chainId: any): any {
 
 function calculateSigRecovery(v: BNLike, chainId?: BNLike): BN {
   const vBN = toType(v, TypeOutput.BN)
+
+  if (vBN.eqn(0) || vBN.eqn(1)) return toType(v, TypeOutput.BN)
+
   if (!chainId) {
     return vBN.subn(27)
   }
@@ -60,6 +63,7 @@ function isValidSigRecovery(recovery: number | BN): boolean {
 
 /**
  * ECDSA public key recovery from signature.
+ * NOTE: Accepts `v == 0 | v == 1` for EIP1559 transactions
  * @returns Recovered public key
  */
 export const ecrecover = function (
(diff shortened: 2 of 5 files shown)


holgerd77:
Ah, just seeing that you already opened a PR on this.

So for context (sorry, this is likely not obvious, we likely want to better document): we are doing work on the current version of the packages towards `master`, on `develop` we are long-term-preparing the next round of breaking releases.

Let's anyhow continue here with discussing the things I raised in the issue, this would be the first step here.

ernestognw:
> Ah, just seeing that you already opened a PR on this.
> 
> So for context (sorry, this is likely not obvious, we likely want to better document): we are doing work on the current version of the packages towards `master`, on `develop` we are long-term-preparing the next round of breaking releases.
> 
> Let's anyhow continue here with discussing the things I raised in the issue, this would be the first step here.

Thanks for context, I changed the target branch to master and rebased accordingly. I replied about the details on the issue, just [right here](https://github.com/ethereumjs/ethereumjs-monorepo/issues/1904#issuecomment-1137623509)

packages/util/src/signature.ts
```diff
@@ -46,6 +46,9 @@ export function ecsign(msgHash: Buffer, privateKey: Buffer, chainId: any): any {
 
 function calculateSigRecovery(v: BNLike, chainId?: BNLike): BN {
   const vBN = toType(v, TypeOutput.BN)
+
+  if (vBN.eqn(0) || vBN.eqn(1)) return toType(v, TypeOutput.BN)
```
holgerd77, line 50: Ok, this is the central line of this PR (this is just an additional transparency explanation for everyone to follow including the team). So this allows `v` to directly pass through if `0` or `1`. The `chainId` parameter is ignored in this case.

The function here is not exported. Effects are indirect, to all methods below this comment where additional `NOTE:` line comments are added. I've checked all the methods and do not see security or other implications or side effects.

Namely:

- `ecrecover`: This will then allow with this update to directly pass in an EIP-1559 (or any other typed) tx with v being `0` or `1` and will recover the correct PK in these cases as well. No side effects on other (legacy) txs.
- `toRpcSig`: Now allows EIP-1559 txs to be converted in `eth_sign` format, no side effects
- `toCompactSig`: Now allows EIP-1559 txs to be converted in `eth_sign` format, no side effects (was first a bit unsure about the `0` case but this is included as well respectively not affected)
- `isValidSignature`: Now allows EIP-1559 txs to be converted in `eth_sign` format, no side effects

So I would give this a go and merge this and would include in the very imminent last round of `master` releases before we merge in `develop`.

We can then continue on `develop` with do some further signature code cleanup - with some inspiration from the breaking-release-TODOs - in favor of a complete direct removal.

holgerd77 (approved):
LGTM, reasoning and justification see separate comment, will merge.

holgerd77:
Oh, respectively tests are failing, then we have to figure out before. This is likely related to the merge of #1900, @jochem-brouwer do you have an idea what happens? 🤔

jochem-brouwer:
Hi @holgerd77, I don't think it is related to #1900 since ecrecover tests are failing. I'll take a look now :smile:

holgerd77:
> Hi @holgerd77, I don't think it is related to #1900 since ecrecover tests are failing. I'll take a look now 😄

Ah, interesting, I thought tests would have been passed before here, but also not 100% sure.

jochem-brouwer:
There are errors in blockchain tests;

```
Errors thrown in file: GeneralStateTests/stPreCompiledContracts2/CALLCODEEcrecover1.json test: CALLCODEEcrecover1_d0g0v0_London:
	Error: invalid receiptTrie (vm hf=london -> block number=1 hash=0x7d96d00345e783092e7b862aaab79d0b08b041d0d91ce8720546c1214895bc91 hf=london baseFeePerGas=10 txs=1 uncles=0)
	correct last header block
Errors thrown in file: GeneralStateTests/stPreCompiledContracts2/CallEcrecover1.json test: CallEcrecover1_d0g0v0_London:
	Error: invalid receiptTrie (vm hf=london -> block number=1 hash=0xf94a90c5e3158e156ad8218bc49e8a59cb17be564648808466302e01040abe49 hf=london baseFeePerGas=10 txs=1 uncles=0)
	correct last header block
Errors thrown in file: GeneralStateTests/stStaticCall/static_CallEcrecover1.json test: static_CallEcrecover1_d0g0v0_London:
	Error: invalid receiptTrie (vm hf=london -> block number=1 hash=0x92f8983a4d37035279956e921f45a17e8b28146591e1dbf031cfffba11a23c2f hf=london baseFeePerGas=10 txs=1 uncles=0)
	correct last header block
```

These all have to do with the `ecrecover` precompile. (This is also the reason why `vm-api` CI fails)

Take a look at where `ecrecover` is called: https://github.com/ethereumjs/ethereumjs-monorepo/blob/2f42dcfedd18852fc95cc4a51da606dfcbf87387/packages/vm/src/evm/precompiles/01-ecrecover.ts#L24

The input to these tests have `v = 1`. The precompile calls into `util` `ecrecover`. This calls into `calculateSigRecovery` (which is edited in this PR): https://github.com/ethereumjs/ethereumjs-monorepo/blob/2f42dcfedd18852fc95cc4a51da606dfcbf87387/packages/util/src/signature.ts#L73

In this PR, `calculateSigRecovery` immediately returns `v = 1`. However, without this PR, and with `chainId` as argument set as `undefined` (as is the case in the precompile!) it returns `v - 27`. This yields a negative number if `v=1` and is thus invalid, the precompile throws. However, in this PR, because `v = 1` it immediately returns `1`, which is a valid value, and thus the call succeeds (and does not throw).

I don't see a clean way how to fix this, it would need an extra argument to `ecrecover` in `util` to signal that we are using a precompile and we thus force `v = v - 27`. (I did not check but I think that the call to `ethereum-cryptography` expects v of either 0 or 1).

jochem-brouwer:
Actually, we could also do these `v` checks in the precompile? Would that work? I'm a bit hesitant here because this has to do with signatures and could have big (security) implications..?

jochem-brouwer:
Tx tests will fail, we should figure out why this `v.eqn(0)` check is in the `_validateTxV` method.

holgerd77:
I am on it.

packages/tx/src/legacyTransaction.ts
```diff
@@ -136,6 +136,18 @@ export default class Transaction extends BaseTransaction<Transaction> {
...
+      // In case of a signed non-EIP155 tx, `ecrecover` does not get a `chainId` provided
+      // Internally, `ecrecover` will calculate the recovery bit by `v = v - 27`
+      // This recovery bit is either 0 or 1. Thus, any other value than `v = 27` or `v = 28` will
+      // throw upon calling `getSenderAddress` here. Thus throw early in the constructor
+      if (this.v !== undefined) {
+        if (!this.v.eqn(27) && !this.v.eqn(28)) {
+          throw new Error(`Non-EIP155 txs need either v = 27 or v = 28, got v = ${this.v}`)
+        }
+      }
+    }
```
holgerd77, line 149: I wouldn't tie this to "after the _validateTxV() analysis" but rather do before respectively directly in the method, think this will get more robust (less ocassion for the analysis to go wrong).

Will try this and update the tests.

holgerd77:
@jochem-brouwer ok, I would have a tendency to give this a go, but I would wait for your final confirmation.

packages/tx/src/legacyTransaction.ts
```diff
@@ -390,6 +378,17 @@ export default class Transaction extends BaseTransaction<Transaction> {
    * Validates tx's `v` value
    */
   private _validateTxV(v?: BN, common?: Common): Common {
+    // Check for valid v values in the scope of a signed legacy tx
+    if (v !== undefined) {
+      // v is 1. not matching the EIP-155 chainId included case and...
+      // v is 2. not matching the classic v=27 or v=28 case
+      if (v.ltn(37) && !v.eqn(27) && !v.eqn(28)) {
```
jochem-brouwer, line 385: If `v = 36` then we get a chainId = 0, right? Why is this not ok? 

In case of `v = 35`, then `v - 35` (thus `0`, which is odd) means we subtract 36, so we end up with a negative chain id which is obviously wrong. But for 36 this logic seems to work..? Unless chain id 0 is special?

holgerd77: TBH I don't find any reference about chain ID 0 and I'm unsure if this theoretically a valid chain ID. I would rather assume for now that it is not, mainnet starts with 1 and all other chain IDs I came across were some arbitrary higher value. In doubt I would tend to leave "as is", if someone demands we could still add later on.

jochem-brouwer (approved):
LGTM, thanks @ernestognw :smile:

packages/tx/src/legacyTransaction.ts
```diff
@@ -382,7 +394,7 @@ export default class Transaction extends BaseTransaction<Transaction> {
     // No unsigned tx and EIP-155 activated and chain ID included
     if (
       v !== undefined &&
-      !v.eqn(0) &&
+      !v.eqn(0) && // TODO fixme -> why overwrite/return default common when `v` is set to 0?
```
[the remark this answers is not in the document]
jochem-brouwer: Ah yes I forgot I wanted to suggest to remove this check. I introduced it a while back but I found no good reason why I introduced it back then (probably some tests failing or something). So yes this can be removed.

```suggestion
```

ernestognw:
> LGTM, thanks @ernestognw 😄

Thanks to you folks! Many things happened here that I would've not been able to address, I appreciate it.
Please let me know if I can add something from my side.

holgerd77 (approved):
LGTM

PR #1906, opened 2022-05-25
Common type safety

### LINKED ISSUE #1888: Common: More Type Safety / any Usage Removal
Part of #1717 

Some last breaking release tasks, from the latest discussion round in the chat:

"Introduce more type safety to common any responses (chains, hardforks, params) (or otherwise framed: remove remaining any usages from Common, try to keep low-level though and avoid new data/type structures and rather reuse (or adopt a bit) where possible)"

Will assign @acolytec3 since he applied for the task, please de-assign and mention here if you won't tackle.

(this task might eventually remain non-breaking, makes nevertheless sense to do towards `develop` regarding the current release state, we are just on the safer side if we keep this within the breaking-releases realm)

acolytec3:
Fixes #1888 

Adds some small type safety improvements in `common`


packages/common/src/index.ts
```diff
@@ -101,6 +103,11 @@ export enum ConsensusAlgorithm {
   Casper = 'casper',
 }
 
+export type cliqueOpts = {
```
ryanio, line 106: suggestion, I think we can call this `CliqueConfig`

packages/common/src/types.ts
```diff
@@ -8,11 +8,11 @@ export interface genesisStatesType {
   [key: string]: {}
 }
 
+export interface nameType {
```
ryanio, line 11: `nameType` isn't very descriptive, maybe `ChainName`? would you be able to define it as `[chainId: string]: string`

packages/common/src/index.ts
acolytec3: Will do.

packages/common/src/index.ts
```diff
@@ -1023,7 +1030,7 @@ export default class Common extends EventEmitter {
    *
    * Note: This value can update along a hardfork.
    */
-  consensusConfig(): { [key: string]: any } {
+  consensusConfig(): { [key: string]: {} | cliqueOpts } {
```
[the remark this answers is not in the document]
acolytec3: Looking at the existing chain configs like [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/0b7cc1b15d1fd9d58e4ae7db794674c3af4da5ae/packages/common/src/chains/mainnet.json#L9),  it's just an empty object so I wouldn't think so.

packages/common/src/types.ts
```diff
@@ -8,11 +8,11 @@ export interface genesisStatesType {
   [key: string]: {}
 }
 
+export interface ChainName {
+  [chainId: string]: string
+}
 export interface chainsType {
```
holgerd77, line 14: Not sure, seems this was lower capital before, but this should be actually capitalized. Can you do along?

packages/common/src/types.ts
```diff
@@ -31,10 +31,7 @@ export interface Chain {
   consensus?: {
     type: ConsensusType | string
     algorithm: ConsensusAlgorithm | string
-    clique?: {
-      period: number
-      epoch: number
-    }
+    clique?: CliqueConfig
     ethash?: any
     casper?: any
```
holgerd77, line 36: I would be a fan of also adding `EthashConfig` and `CasperConfig` types here as well, even if they might be empty.

packages/common/src/index.ts
```diff
@@ -101,6 +103,11 @@ export enum ConsensusAlgorithm {
   Casper = 'casper',
 }
 
+export type CliqueConfig = {
+  period: number
+  epoch: number
+}
+
```
holgerd77, line 110: Wondered if this should really go here and then I re-realized that we are really mixing a lot of things up regarding type, options dicts and main class location, we should re-structure this to match other libraries like tx or block.

Will open an own dedicated issue though, not something to be done here.

packages/common/src/types.ts
acolytec3: Makes sense.  Will update, check all the tests, etc, and then push a new commit.

packages/common/src/index.ts
acolytec3: I went ahead and moved this config back to `./types.ts` and added the `EthashConfig` and `CasperConfig` empty types as well.  All the tests pass locally so  have pushed a commit.  It also necessitated re-exporting `CliqueConfig` from `index.ts` but I think that gives us a point of departure on the general `Common` reorganization issue.

PR #1912, opened 2022-05-26
EVM refactor

### LINKED ISSUE #1903: EEI/EVM package extraction and refactor tracking issue
This issue tracks extracting the EEI and EVM from the VM package and create new packages for those. 

All is open for discussion so is subject to change.

The goal of EEI (Ethereum External Interface) is to provide an implementation of the EEI interface (this interface is located in EVM) to interact with the external environment, such as contract code, storage, account balances, block hashes, etc. 

The goal of the EVM is to provide an interface and implementation to run bytecode. This calls back into the EEI if it needs external information, such as account balances, storages, blockhashes, etc. 

To refactor the current implementation: (Anything with a (?) is stuff I'm not really sure about at this point)

EEI
----

- [x] Remove EVM dependency
- [x] Remove everything related which has to do with gas. (`useGas`, `refundGas`, `subRefund`, `addStipend`, `getGasLeft`). Move this to EVM.
- [x] Remove the `ENV` from `EEI`, this is EVM related and should be moved there. Also remove all getters to this environment (`getCallData`, `getAddress`, etc.)
- [x] Most of the `call` logic (`delegatecall`, `call`, etc.) should also be moved into EVM. Except getting the code for the context of the new message.
- [x] Import `VmState` into EEI and remove it from VM

EVM
----

- [ ] Remove EEI and StateManager dependency
- [ ] Create a minimal EEI interface and export this
- [ ] Import any gas-related logic (touched accounts, warm address, etc.) from EEI into EVM
- [ ] Only dependencies of `EVM` should be `common`, `util` and anything related to precompiles (+ crypto)
- [ ] `runCall` and `runCode` should be moved into one method, the parameters passed on will determine if it is a call or just the code. (Probably the `Message` and this method are the only interface the EVM should provide, together with a constructor which sets the EEI)

I probably missed a few things.

jochem-brouwer:
Closes #1903 

**NOTE: change target branch to `develop`!!**

Commit after branching off from `develop`; https://github.com/ethereumjs/ethereumjs-monorepo/commit/ae950d29e93cbda65b4225431e7560b9e36d5daf

TODOs related to refactor: (probably incomplete)

- [x] Create interfaces for EEI and EVM inside EVM package
- [x] Ensure VM has option to pass in custom EEI or EVM which honour above interfaces
- [x] Improve naming (`EEI._state` -> `state`, or probably a better name, (`journal`?))
- [x] Remove `VmState` dependency from EVM
- [x] ~~Remove `runCode` from EVM and fit this logic inside `runCall`~~ -> Will not do this, will be part of standard EVM, but will not be part of EVM interface (since this is not used internally in any of our packages)
- [x] ~~Possibly add a submodule to `Interpreter` to cover all the methods moved from `EEI` into `EVM`~~ -> This can be refactored in a later PR and is non-breaking.
- [x] Check if some methods moved into `Interpreter` can be optimized or removed (directly read from `_env`, for instance `getCallValue()` (maybe make `_env` -> `env`)
- [ ] Move EVM into a new package
- [ ] Check commits for any TODOs
- [ ] (?) move touched account logic into EVM
- [x] (?) move transientStorage logic into EVM
- [x] `VM` and `EVM` are extensions of `AsyncEventEmitter`, but the latter does not have a type, therefore `VM` and `EVM` are interpreted as `any` type with certain fields which are sure there, but we want to ensure they are cast as right type.
- [x] Add types for the `step` event (an enum to ensure one does not subscribe to e.g. `VM.step` -> should subscribe to `EVM.step`)
- [x] ~~Remove `StateManager` from `VM` as dependency (All chain through EEI)~~
- [ ] Check input types to `EVM`, their names are sometimes inconsistent
- [ ] Checkt output type of `EVM`, also sometimes inconsistent (two `gasUsed`, one for interpreter and one for tx (which also covers data cost))
- [x] Check if more methods in VM/EVM should be marked as `protected` and/or `readonly`
- [ ] Fix client + client tests
- [ ] Fix VM API tests
- [x] Figure out reason of underscore object properties (`_eei` instead of `eei`) and if we should keep this (and be consistent)
- [x] Make the state of Interpreter/Env/RunState consistent. (For instance `auth` seems to be something which should go into `runState`). Possibly also consider the frame types introduced in `types.ts` (`FrameEnvironment` types)
- [x] Move the `activePrecompiles` logic into EVM (from VM)

Note: warm address logic should be kept into EEI because we need to set this up before calling into EVM (add warm addresses in the access list + own account + target account). TODO this is not true, can set this up as input into the EVM.

Cleanups:

- [x] Move dynamic gas usage of `EXP` from `functions.ts` into `gas.ts` (this also ensures the `step` event produces the right dynamic gas once it is invoked)


jochem-brouwer:
ac3394145268d075d36527fa71f0e6532a97ba6d is temporary, need to approach this in small steps otherwise I exhaust myself with too many tasks to do at the same time (and then fixing consensus bugs is gonna be very hard)

(Also writing this down bit for myself)

Next steps:

- [ ] Inject `TransientStorage` into EEI
- [ ] Remove `CreateEIOptions`
- [ ] EEI factory should probably be removed, EEI is instantiated once on EVM creation

jochem-brouwer:
`ethereum/tests` CI passes for f447394646ee805f9aa37cf4a9a3b3b1ae0a62e7

jochem-brouwer:
EEI and EVM both blend some revert/commit logic. Not all revert/commit logic should move to EVM but things like refund counters can easily be handled in EVM. `Interpreter` is always created also when a new `EEI` is created so this is a fantastic candidate to move things to. Probably want to create a subclass later not to bloat Interpreter too much, but for now this seems fine.

holgerd77 (commented):
Had a first look through all the commit changes (up to the latest state), this looks insanely great 👍, also super clean procedural refactoring, that's really fantastic and super-nice to follow along.

So, I am very much confident that this is going in a good direction. 😄

packages/vm/src/evm/opcodes/codes.ts
```diff
@@ -57,7 +57,7 @@ const opcodes: OpcodeEntry = {
   0x07: { name: 'SMOD', isAsync: false, dynamicGas: false },
   0x08: { name: 'ADDMOD', isAsync: false, dynamicGas: false },
   0x09: { name: 'MULMOD', isAsync: false, dynamicGas: false },
-  0x0a: { name: 'EXP', isAsync: false, dynamicGas: false },
+  0x0a: { name: 'EXP', isAsync: false, dynamicGas: true },
```
holgerd77, line 60: I would strongly suggest to not do *any* kind of such non-related fixes along such type of work (or is there a relationship I am not seeing?). Rather open an issue and do separately later. These kind of things DO increase the complexity (there wasn't anything explicitly "not working" before) and even more increases the risk of drawing in new errors, test failures,... on top of already complex work.

jochem-brouwer: Sure, I can revoke this, then I will add a tracking issue for these inconsistencies. This is definitely a bug in the `step` event since the reported gas used there is wrong.

holgerd77: Ah, not sure if a revoke is needed at this point now, maybe better decide for yourself what you think is better/more practical here for now. Just a general comment (I guess along these general refactoring one can find *a lot* of things 😋).

jochem-brouwer: I think you have raised a good point, I'm trying to cramp too much into one PR, it is cleaner for these semi-unrelated things to put into another PR :smile:

jochem-brouwer:
I'm currently removing some variables from EVM and Interpreter which are class-level variables but these should be bound to `RunState`. 

Current errors; (Chainstart)

```
Errors thrown in file: GeneralStateTests/stAttackTest/CrashingTransaction_d0g0v0.json test: CrashingTransaction_d0g0v0_Frontier:
	Error: invalid receiptTrie (vm hf=chainstart -> block number=1 hash=0xd72ac14fba68d65afbd262053804119332b806a2917fe8666cbb11558636324f hf=chainstart baseFeePerGas=none txs=1 uncles=0)
	correct last header block
Errors thrown in file: GeneralStateTests/stRefundTest/refund_CallA_OOG_d0g0v0.json test: refund_CallA_OOG_d0g0v0_Frontier:
	Error: invalid receiptTrie (vm hf=chainstart -> block number=1 hash=0x9831a11f1020b1ae646c45e623307f4a8113741a5efb9b2e3cfdb4bc90a6f56e hf=chainstart baseFeePerGas=none txs=1 uncles=0)
	correct last header block
Errors thrown in file: GeneralStateTests/stRefundTest/refund_OOG_d0g0v0.json test: refund_OOG_d0g0v0_Frontier:
	Error: invalid receiptTrie (vm hf=chainstart -> block number=1 hash=0x46243acb0b4c07648446f91b627d0bce6b95b2e36840584610373befa035494c hf=chainstart baseFeePerGas=none txs=1 uncles=0)
	correct last header block
Errors thrown in file: GeneralStateTests/stReturnDataTest/returndatasize_after_oog_after_deeper_d0g0v0.json test: returndatasize_after_oog_after_deeper_d0g0v0_Frontier:
	Error: invalid receiptTrie (vm hf=chainstart -> block number=1 hash=0xdefe43f9804ca6250525bf702e4b61df476df76e0c5040a0936fb3b7f7affe1a hf=chainstart baseFeePerGas=none txs=1 uncles=0)
	correct last header block
Errors thrown in file: GeneralStateTests/stTransactionTest/ContractStoreClearsOOG_d0g0v0.json test: ContractStoreClearsOOG_d0g0v0_Frontier:
	Error: invalid receiptTrie (vm hf=chainstart -> block number=1 hash=0xc8b19ef4f7ce23bcddebb3ef512870f5bcd53f7948405c138be31d339dc2a73f hf=chainstart baseFeePerGas=none txs=1 uncles=0)
	correct last header block
Errors thrown in file: ValidBlocks/bcExploitTest/SuicideIssue.json test: SuicideIssue_Frontier:
	Error: invalid receiptTrie (vm hf=chainstart -> block number=1 hash=0x32ff603139699bf940c21b224126eb5596ab82d1a3908c019680eb4c4055d7ff hf=chainstart baseFeePerGas=none txs=1 uncles=0)
	correct last header block
```

packages/vm/src/evm/types.ts
```diff
@@ -292,3 +293,10 @@ type EVMResult = {
 export interface EVMInterface {
   runMessage(environment: EVMEnvironment): Promise<EVMResult>
 }
+
+export type EVMEvents = {
+  newContract: (data: NewContractEvent, resolve?: (result: any) => void) => void
+  beforeMessage: (data: Message, resolve?: (result: any) => void) => void
+  afterTx: (data: EVMResult, resolve?: (result: any) => void) => void
+  step: (data: InterpreterStep, resolve?: (result: any) => void) => void
+}
```
holgerd77, line 294: Ok, besides all the fear I have that too much stuff is getting into this PR I have to say:

This is a *dramatic* improvement on the current status quo!!! 🎉 🙏 💯

jochem-brouwer: Good catch, I copied these function sigs around and forgot to remove `afterTx`

jochem-brouwer: And yes, the types are absolutely fantastic :heart_eyes:

holgerd77:
Ok, I very much think we can close here since work is continued in #1950 and will actually do, feel free to reopen if this was mistaken.

jochem-brouwer:
Yep that's right, but will use this original post to keep track of the remaining points

PR #1915, opened 2022-05-26
Common refactor

### LINKED ISSUE #1913: Common: Align File Layout with other Libraries
Depends on #1906 

The distribution of Common library types, option dicts and the main class code is still somewhat chaotically structured and it is not clear where to put things. We should align this along the `develop` work with other libraries, I guess [block](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/block/src) is the example most getting close.

So we should do the following:

- Move all remaining types, option dicts, interfaces, etc. to `types.ts`
- Move the main class code to a new file `common.ts` (maybe do in two commits: 1. create emtpy common.ts file, 2. move the content, this might preserve history better)
- Reexport the default class export (see this [StackOverflow](https://stackoverflow.com/questions/39999282/re-export-default-in-es-6-modules) question for context) and all the types in the `index.ts` file (again see block example for reference)

Maybe this would also be a good occasion to - after that - also have a short look at the type names and see if things make sense and are somewhat consistent and eventually clean up a bit along the way. But in doubt also leave here.

acolytec3:
Fixes #1913


acolytec3:
Barring something I missed somewhere in an obscure test, this should be ready for a first round of review.  The important bits are the `common` reorganization changes.  There are tons of changes elsewhere but they are just copy/paste updates for the enums more or less.

packages/block/test/block.spec.ts
```diff
@@ -20,7 +20,7 @@ import util from 'util'
 
 tape('[Block]: block functions', function (t) {
   t.test('should test block initialization', function (st) {
-    const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Chainstart })
+    const common = new Common({ chain: ChainId.Mainnet, hardfork: HardforkName.Chainstart })
```
holgerd77, line 23: Ah, sorry, but I find both these renaming very unfortunate, I didn't think that you would change these actually.

These are *all* over the place, we settled on the short names a long time ago (and I honestly also like these a lot) and this is also used by the Community on all their instantiations, so this would cause a lot of unnecessary hazzle.

Can you please name back? If this conflicts with something else, then please rename the "else" part, but this should really stay stable and untouched.

holgerd77: Yes, these are a lot more uncritical, since close to no one will use these directly. So maybe we can do `HardforkConfig` and `ChainConfig` for these?

holgerd77: (not sure if we should then rename to `GenesisBlockConfig` and `BootstrapNodeConfig` as well for consistency? A bit undecided)

acolytec3: I believe I've addressed all the feedback here.  I know @ryanio mentioned combining the types/enums into one file but I kind of like them separated.  Just makes things more readable to my way of thinking but could be persuaded otherwise.

ryanio: nice thanks, just took a look, yeah i think it is kind of nice to have them separated actually :)

holgerd77 (approved):
Ah, much smaller diff, nice, this looks great now! 🙂 

Thanks a lot Andrew!

Will merge.

PR #1916, opened 2022-05-27
common/block/blockchain: genesis refactor

ryanio:
This PR works on the ideas presented in https://github.com/ethereumjs/ethereumjs-monorepo/pull/1833, mainly https://github.com/ethereumjs/ethereumjs-monorepo/pull/1833#issuecomment-1088445250 and https://github.com/ethereumjs/ethereumjs-monorepo/pull/1833#issuecomment-1091181809.

This PR:

- Removes `stateRoot` and `hash` from Common
- Moves genesis functionality from Block to Blockchain
- Moves `genesisStates` from Common to Blockchain
- Removes`VM.runBlockchain()`
    - unused, except for in `BlockchainTestsRunner` where we now directly putBlocks and run the blockchain iterator

I haven't run the test suite yet to fix failing tests as I wanted some confirmation that we want to proceed on this path before working on the final touches.

The only downside so far (that I see) is that when setting up a custom chain you used to only need a very portable common but now you have to additionally pass the genesis states into blockchain, but this was easy enough to adjust to in the client by passing an optional blockchain when custom genesis states (or geth genesis) is provided.

Removing `stateRoot` and `hash` was also tricky in a few places but I think I overcame the challenges.


holgerd77:
Thanks so much. 🙂 

I think I am a fan of continuing on this.

ryanio:
Have to look into two ci issues, vm-api browser run timing out (taking too long?) and client cli connection. The PR can still be reviewed for everything else meanwhile.

ryanio:
This is ready for review :)

holgerd77:
Oh, 1 commit 😬 (thanks a thousand times, otherwise totally great that you got through this!!!).

This will be a bit harsh for review, would have preferred to have this step-by-step on such a large change set. But will try my best. 😋

packages/block/test/block.spec.ts
```diff
@@ -658,41 +658,11 @@ tape('[Block]: block functions', function (t) {
...
-    const genesis = Block.genesis({}, { common })
-    const ropstenStateRoot = Buffer.from(
-      '217b0bbcfb72e2d57e28f33cb361b9983513177755dc3f33ce3e7022ed62b77b',
-      'hex'
-    )
-    st.ok(genesis.header.stateRoot.equals(ropstenStateRoot), 'genesis stateRoot match')
+    const rlp = Buffer.from(testDataGenesis.test.genesis_rlp_hex, 'hex')
+    const hash = Buffer.from(testDataGenesis.test.genesis_hash, 'hex')
+    const block = Block.fromRLPSerializedBlock(rlp, { common })
+    st.ok(block.hash().equals(hash), 'genesis hash match')
```
holgerd77, line 665: Ah, nice replacement. 🙂

packages/blockchain/src/index.ts
```diff
@@ -105,6 +107,32 @@ export interface BlockchainOptions {
...
+   * A complex genesis state with Contract and EoA states would have the following format:
+   *
+   * ```javascript
+   * {
+   *   '0x0...01': '0x100', // For EoA
+   *   '0x0...02': ['0x1', '0xRUNTIME_BYTECODE', [[storageKey1, storageValue1], [storageKey2, storageValue2]]] // For contracts
+   * }
+   * ```
+   */
+  genesisState?: GenesisState
```
holgerd77, line 135: Ok, just followed this along just to get some confirmation: so if one now would want to run a VM with/upon some custom genesis state, one would create a Blockchain object first with this genesis state and passes this into the VM along initialization. Seems a reasonable replacement to me for this eventual use case.

packages/blockchain/src/index.ts
```diff
@@ -115,7 +143,8 @@ export default class Blockchain implements BlockchainInterface {
   db: LevelUp
   dbManager: DBManager
 
-  private _genesis?: Buffer // the genesis hash of this blockchain
+  private _genesisBlock?: Block /** The genesis block of this blockchain */
```
holgerd77, line 146: This is actually also a nice replacement to have the `Block` here and not level down to just the hash (why?).

packages/blockchain/src/index.ts
```diff
@@ -288,9 +318,17 @@ export default class Blockchain implements BlockchainInterface {
...
-      const common = this._common.copy()
-      common.setHardforkByBlockNumber(0)
-      genesisBlock = Block.genesis({}, { common })
+      let stateRoot
+      if (this._common.chainId() === BigInt(1) && this._customGenesisState === undefined) {
+        // For mainnet use the known genesis stateRoot to quicken setup
+        stateRoot = Buffer.from(
+          'd7f8974fb5ac78d9ac099b9ad5018bedc2ce0a72dad1827a1709da30580f0544',
+          'hex'
+        )
```
holgerd77, line 327: 👍 

Great optimization!

packages/blockchain/src/index.ts
```diff
@@ -395,7 +431,7 @@ export default class Blockchain implements BlockchainInterface {
   async getIteratorHead(name = 'vm'): Promise<Block> {
     return await this.runWithLock<Block>(async () => {
       // if the head is not found return the genesis hash
-      const hash = this._heads[name] || this._genesis
+      const hash = this._heads[name] ?? this._genesisBlock?.hash()
```
holgerd77, line 434: Again, this is really a nice side refactor.

holgerd77:
Rebased this via UI after local test rebase.

packages/blockchain/test/index.spec.ts
```diff
@@ -596,8 +599,9 @@ tape('blockchain test', (t) => {
   })
 
   t.test('should put multiple blocks at once', async (st) => {
+    const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Chainstart })
     const blocks: Block[] = []
-    const genesisBlock = Block.genesis({ header: { gasLimit: 8000000 } })
```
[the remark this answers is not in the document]
ryanio: heh yeah since the block number still defaults to zero, we weren't losing much, as there were not very many cases were we needed the actual genesis params.

packages/client/lib/net/protocol/ethprotocol.ts
```diff
@@ -117,7 +118,12 @@ export class EthProtocol extends Protocol {
       response: 0x04,
       encode: ({ reqId, block, max, skip = 0, reverse = false }: GetBlockHeadersOpts) => [
         bigIntToBuffer(reqId ?? ++id),
-        [typeof block === 'bigint' ? bigIntToBuffer(block) : block, max, skip, !reverse ? 0 : 1],
+        [
+          typeof block === 'bigint' ? bigIntToBuffer(block) : block,
+          max === 0 ? Buffer.from([]) : intToBuffer(max),
+          skip === 0 ? Buffer.from([]) : intToBuffer(skip),
+          !reverse ? Buffer.from([]) : Buffer.from([1]),
+        ],
```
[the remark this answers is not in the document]
ryanio: this is related to the RLP v3 changes, not sure why it just popped up in my PR, but when doing `bufArrToArr` we need all fields to be Buffer, otherwise if it is is a plain number it didn't convert correctly to Uint8Array.

packages/client/lib/util/parse.ts
```diff
@@ -89,81 +76,6 @@ export function parseTransports(transports: string[]) {
...
-    // hardfork to London for baseFeePerGas support
-    const hardforks = common
-      .hardforks()
-      .map((h) => (h.name === Hardfork.London ? { ...h, block: 0 } : h))
-    common = Common.custom({ chainId: 1, hardforks })
-  }
-  common.setHardforkByBlockNumber(0)
-  return BlockHeader.fromHeaderData(headerData, { common })
-}
-
```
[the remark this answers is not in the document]
ryanio: yeah! some great simplifications through the process here

packages/blockchain/src/index.ts
```diff
@@ -1135,6 +1166,70 @@ export default class Blockchain implements BlockchainInterface {
...
+      extraData: toBuffer(extraData),
+      nonce: toBuffer(nonce),
+      stateRoot,
+    }
+
+    if (baseFeePerGas !== undefined && common.gteHardfork(Hardfork.London)) {
+      header.baseFeePerGas = BigInt(baseFeePerGas)
+    }
+
+    return Block.fromBlockData({ header }, { common })
```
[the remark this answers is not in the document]
ryanio: ah yeah really good point actually, and we don't need to do any bigint or buffer conversions since fromBlockData will handle that, so to make this more future-proof I can do something like:

```
const header: BlockData['header'] = {
      ...common.genesis(),
      number: 0,
      stateRoot
}
```

I will add in a small follow up PR with a few other small fixes on my develop todo list.

Issue #1914, opened 2022-05-26
v6 "To Infinity and Beyond" 🚀 Breaking Releases Finalization Planning 🙂


holgerd77:
Continuation/concretization of #1717 

The v6 releases are shy around the corner (weeks away), so this document outlines a concrete timeline for the releases to get things finalized.

Following Major Version Releases are planned:

- VMExecution (or similar, Renamed, previously: VM (+EVM)) v6 (make sense to keep VM versions since this from the API level this is somewhat of a continuation on what people are used to)
- StateManager (New) v1
- EVM (New) v1
- Blockchain v6
- Block v4, Tx v4
- Common v3
- Util (Renamed) v6
- Trie (Renamed) v5
- Devp2p v5 (new) (-> RLP, BigInt)
- Ethash v2 (new) (-> RLP, BigInt)

Phew. What a line-up. 😋

I guess for Client (this would be v0.6) we can decouple and wait a bit (few weeks or so) and see if things are stable, no need to do a syncronous release.

So following planned timeline:

| Week        | Status | Final Date           | Task  | Status |
| ------------- |---|-------------|-----|:-----:|
| CW21 |  |  |  |  |
| May 23 - 29 | ✔️ | May 26 | Finalization Planning  | 🟢 |
| CW22 |  |  |  |  |
| May 30 - June 06 | ✔️ | May 30 | Last Round v5 Releases, Release Notes  | 🟠 |
|  | ✔️ | June 01 | v5 Releases  | 🟠 |
|  | ✔️ | June 03 | Branch off `v5-maintenance`, work stop `master`  | 🟢 |
| CW23 |  |  |  |  |
| June 06 - 12 | ✔️ | June 06 | Last `develop` rebase on `master`  | 🟠 |
|  | ✔️ | June 07 | `develop` merge into `master`, PR #1943  | 🟠 |
|  | ✔️ | June 08 - 11 | Util + Trie Library Renaming  | 🟢 |
|  |  | June 08 - 11 | VM/EVM/StateManager Package Creation  | 🟠 |
|  |  | June 12 | `develop` (respectively: v6 `master`) Feature Freeze  | 🔴 |
| CW24 |  |  |  |  |
| June 13 - 19 | [ ] | June 13 -14 | Release Notes v6  | 🔴 |
|  | [ ] | June 15 | v6 Beta Releases  | 🔴 |
| CW25 |  |  |  |  |
[...]

PR #1917, opened 2022-05-27
refactor(trie): abstract database implementation with an interface

faustbrian:
Resubmission of https://github.com/ethereumjs/ethereumjs-monorepo/pull/1895 with squashed history and rebase on `develop`

### DIFF
--- a/packages/trie/benchmarks/index.ts
+++ b/packages/trie/benchmarks/index.ts
@@ -1,45 +1,50 @@
-import Benchmark = require('benchmark')
+import Benchmark from 'benchmark'
 import { runTrie } from './random'
 import { iterTest } from './checkpointing'
+import { MemoryDB, LevelDB } from '../dist'
 
 const suite = new Benchmark.Suite()
 
-// random.ts
-// Test ID is defined as: `pair_count`-`era_size`-`key_size`-`value_type`
-// where value_type = symmetric ? 'mir' : 'ran'
-// The standard secure-trie test is `1k-9-32-ran`
-// https://eth.wiki/en/fundamentals/benchmarks#results-1
-suite
-  .add('1k-3-32-ran', async () => {
-    await runTrie(3, false)
-  })
-  .add('1k-5-32-ran', async () => {
-    await runTrie(5, false)
-  })
-  .add('1k-9-32-ran', async () => {
-    await runTrie(9, false)
-  })
-  .add('1k-1k-32-ran', async () => {
-    await runTrie(1000, false)
-  })
-  .add('1k-1k-32-mir', async () => {
-    await runTrie(1000, true)
-  })
+for (const [name, DB] of Object.entries({ MemoryDB, LevelDB })) {
+  const db = new DB()
 
-// checkpointing.ts
-suite
-  .add('Checkpointing: 100 iterations', async () => {
-    await iterTest(100)
-  })
-  .add('Checkpointing: 500 iterations', async () => {
-    await iterTest(500)
-  })
-  .add('Checkpointing: 1000 iterations', async () => {
-    await iterTest(1000)
-  })
-  .add('Checkpointing: 5000 iterations', async () => {
-    await iterTest(5000)
-  })
+  // random.ts
+  // Test ID is defined as: `pair_count`-`era_size`-`key_size`-`value_type`
+  // where value_type = symmetric ? 'mir' : 'ran'
+  // The standard secure-trie test is `1k-9-32-ran`
+  // https://eth.wiki/en/fundamentals/benchmarks#results-1
+  suite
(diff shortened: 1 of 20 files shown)


holgerd77:
Rebased this via UI after testing locally.

holgerd77 (approved):
LGTM, great great work, thanks so much! 😄 

Will merge.

holgerd77:
Could you outline a potential API usage (instantiation) in 4-5 lines?

holgerd77:
Ok, thanks, will ask within the team! 🙂

holgerd77:
(I guess chances are 50:50)

packages/trie/benchmarks/index.ts
```diff
@@ -1,45 +1,50 @@
...
-  .add('1k-9-32-ran', async () => {
-    await runTrie(9, false)
-  })
-  .add('1k-1k-32-ran', async () => {
-    await runTrie(1000, false)
-  })
-  .add('1k-1k-32-mir', async () => {
-    await runTrie(1000, true)
-  })
+for (const [name, DB] of Object.entries({ MemoryDB, LevelDB })) {
```
ryanio, line 8: hey thanks for the great PR, continued work and interest!

I'm not sure if it's worth doubling the number of benchmarks run here on every PR. What we could do instead is a one time profile difference between the two and record the results in the readme, and save the script so we could add in the future like lmdb comparison.

holgerd77:
> @holgerd77 created a draft of the changes to show what it would actually entail - [faustbrian@da41ce5](https://github.com/faustbrian/ethereumjs-monorepo/commit/da41ce53a0b1b5b124751eec4aa17c5417706d6a)

Thanks a lot for all this pro-active engagement, that's really great!! 👍 

I guess we do not have an agreement yet, opinions are somewhat mixed and reserved, will keep you updated!

packages/trie/benchmarks/index.ts
ryanio: Sure a BENCHMARK.md sounds good too, you can link to it from the readme so people can discover it. There is already a trie-benchmarks GitHub action so my intention was to not need to run the comparison every time.

holgerd77:
> @holgerd77 created a draft of the changes to show what it would actually entail - [faustbrian@da41ce5](https://github.com/faustbrian/ethereumjs-monorepo/commit/da41ce53a0b1b5b124751eec4aa17c5417706d6a)

Hi there, to follow-up: still no clear opinion crystalizing out. Do you see (have yourself) some substantial practical gains from this separation or is this rather some "we should do it since this is how things should be?" thing?

(To be clear: I (and I guess I can say: we) would also very much value the second argument, however in the current situation first argument would give a stronger reasoning to do since we do have to limit ourselves right now and from the "it would be structurally cleaner to do x" perspective we still have a lot on the plate which we would/could/would want to realize at some point (but also can't/won't do right now for very much practical/procedural reasons))

holgerd77:
(and so be totally transparent on this: my current tendency is to rather *not* do since this does feel as overloading to me, this this current release round is just already demanding *a lot* from the team due to it's dramatically large scope it already has)

holgerd77:
> Pretty much "how it should be done" but also practical gains resulting from that because then the trie package can be used in browsers which would resolve #1056. Alternatively you could also upgrade `level/level` to `^8.0` and that way get an easy way to add browser support without removing LevelDB.

Thanks, we have various `level` usages in our package suite - 4-6 or so and all on a similar version range - e.g. also as a block data store in the blockchain package, in the client.

Can you judge on the effort to then bring *all* of this to level 8? This might then also be a way here, would also fit well in the scope of our release process, to do such an upgrade right now.

holgerd77:
> I upgraded it in of my own applications and that only took an hour with testing if everything worked but I can work on a PR tomorrow and submit a draft during the weekend so you can see what it looks like.

Uh. If it is not too much of a burden too you this would be great! 💯

Thanks a lot for all these improvement offers - also for being flexible what to integrate. So when you do a `"level` search with the `packages` scope, you get 8 occurrences of `level` and `level`-related (`level-mem`) packages.

Is this strictly a breaking change, or would old DBs still work, do you know that (we would nevertheless do towards `develop`, just asking)?

Also, in case you do you might want to wait for #1928, this is our very last rebase of `develop` towards the `master` branch, from that on we will have no more `master` changes and `develop` will get merged into `master` soon (days, sometime next week).

holgerd77:
@faustbrian Hi there, we have now merged in `develop` into `master`. So if you find some time you can submit a PR directly towards `master` now 🙂.

According to our timeline from #1914 a feature freeze for the v6 `master` is intended to be applied on June 12. We are - mostly - in time, but this might also get 1-2 more days though.

PR #1919, opened 2022-05-27
client: Subsume engine's INVALID_TERMINAL_BLOCK into INVALID response

g11tech:
The https://github.com/ethereum/execution-apis/pull/217 part of the v1.0.0-alpha.9 subsumes the INVALID_TERMINAL_BLOCK into INVALID response via  {status: INVALID, latestValidHash: 0x0000000000000000000000000000000000000000000000000000000000000000}

This PR implements the same.

### DIFF
--- a/packages/client/lib/rpc/modules/engine.ts
+++ b/packages/client/lib/rpc/modules/engine.ts
@@ -16,7 +16,6 @@ import type { VMExecution } from '../../execution'
 import type { Config } from '../../config'
 import type { FullEthereumService } from '../../service'
 
-export const ZERO_VALID_HASH = '0x0000000000000000000000000000000000000000000000000000000000000000'
 export enum Status {
   ACCEPTED = 'ACCEPTED',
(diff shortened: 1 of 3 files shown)


packages/client/lib/rpc/modules/engine.ts
```diff
@@ -396,9 +395,9 @@ export class Engine {
         const validTerminalBlock = await validateTerminalBlock(parent, this.chain)
         if (!validTerminalBlock) {
           const response = {
-            status: Status.INVALID_TERMINAL_BLOCK,
+            status: Status.INVALID,
             validationError: null,
-            latestValidHash: null,
+            latestValidHash: ZERO_VALID_HASH,
```
ryanio, line 400: thanks for opening! looks great. just thinking, could you instead use `bufferToHex(zeroes(32))`? don't think we need an extra constant really.

ryanio (approved):
Thanks, lgtm!

PR #1920, opened 2022-05-28
Refactor EVM internals

jochem-brouwer:
Part of #1912 (added by @holgerd77)

Extra PR because I am not entirely sure this is the way to go (but getting more convinced that this has extremely nice properties). Also writing this down for myself so I ensure I don't lose track.

The idea is that the bare interface of an `EVM` implementation should be only a single method:

`runCall(env: EVMEnvironment): Promise<EVMResult>`

Here `EVMEnvironment` (see `src/evm/types.ts`) is supposed to cover the bare minimum which we want the VM to receive. This is equal to the input to `runCall` from `src/runTx.ts`. Since we also provide the bare minimum of the output (`EVMResult`) this ensures that our VM can work with the output whatever the EVM gives.

However, in our internal implementation these types will be much larger. In particular, `EVMEnvironment` also holds the state of the Interpreter once we invoke it. Second of all, any local variables written (such as decreasing gas limit, setting auth variables, etc.) is not done on class-level but is done within the environment. 

This has the very nice property that you can take a "snapshot" of the current environment (so we capture current address, code, the current PC, gasLeft, etc.) and you can dump this into `runCall` to re-execute everything from that point. Note that the external environment might have a different state at this point though, but that can be edited via the public `eei` of EVM.

Internally (this is also currently done like this, but we expand here) anything which has to do with calls/creates eventually calls back into EVM's `runCall`. Since Interpreter and EVM do not have any variables which change (all changes are written to current environment which is in a function closure, not dangling in the object somewhere), we can create a singleton instance of Interpreter inside EVM (just like what we do with EEI, this was also instantiated each call/create but not anymore).

This has the extremely nice property that we can now do calls like `evm.interpreter.delegatecall(args)` - i.e. we can directly execute a delegatecall without creating a tx first. I am sure that some test tools have integrated this themselves, but it would be very nice to support. It is also more clean to me to not use class-wide semi-public variables to track state, but instead keep track of it during the chained function calls, which mirrors how the EVM works too.

Steps:

- [ ] Extract minimum types for the `EVMEnvironment` 
- [ ] Expand `EVMEnvironment` with extra variables which we need inside EVM and Interpreter (these are taken from Env, RunState, and the properties of the classes which change during executiong (`_lastReturned`, `_gasLeft`). This deletes the classes/types `Env`, `RunState`, `Message`
- [ ] Ensure that this is an argument passed into `runCall` and ensure that all (?) of these properties are allowed to be undefined, the internal value should have all values defined (if a value has to be undefined such as `auth`, use explit `null`)
- [ ] Ensure that `EVMEnvironment` object gets passed around (somewhat like how we pass `_runState` around)
- [ ] Ensure that any call/create method inside Interpreter creates the correct new environment and calls back into EVM
- [ ] Add warm addresses + touched accounts to MachineState and remove it from EEI

### DIFF
--- a/packages/vm/src/evm/evm.ts
+++ b/packages/vm/src/evm/evm.ts
@@ -26,6 +26,8 @@ import { CustomPrecompile, getActivePrecompiles, PrecompileFunc } from './precom
 import { TransientStorage } from '../state'
 import {
   CustomOpcode,
+  EVMEnvironment,
+  EVMEnvironmentExtended,
   /*ExternalInterface,*/
   /*ExternalInterfaceFactory,*/
   Log,
@@ -34,6 +36,8 @@ import {
   TxContext,
 } from './types'
 import EEI from '../eei/eei'
+import Memory from './memory'
+import Stack from './stack'
 
 const debug = createDebugLogger('vm:evm')
 const debugGas = createDebugLogger('vm:evm:gas')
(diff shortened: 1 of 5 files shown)


jochem-brouwer:
Oh right, just realized: with this approach we can also move the warm addresses and the touched accounts from EEI into the EVM, because we can inject it into the `MachineState`

holgerd77:
Hi there, great, thanks a lot and so much, this in-depth analysis and thinking about the structure is so super valuable!! 🙂 👍 On first read this sounds both logical and very appealing to me, I am not sure though, if I am the right person to judge on all the possible implications and side effects.

I will draw in two additional persons here, namely Franco Victorio @fvictorio from Hardhat and Tair Asim @assimovt from evm.codes, guys, are you eventually able to give some feedback here respectively have an opinion, since you will likely be heavy users of this and might have a clearer view on the implications and eventual possible drawbacks and limitations?

For context: we are in a very heavy VM refactor for the - relatively imminent (weeks, expect a dev release round before) - breaking release round. The main goal is to separate the outer call-(chain-)context VM from the inner bytecode-processing EVM with a clean interface between both, to allow on better modularization/customization on both parts as well as dramatically more lightweight inner EVM.

jochem-brouwer:
I have experimented a bit with these new types today and yesterday but this refactor is blowing up too much. I will try and see to remove some inconsistencies in the EVM/Interpreter, but will do that in the other PR, as this is growing way too large. This refactor would probably also grow so big that we won't be able to review it before our planned release target. I'll go back to the "simpler" tasks on the other PR and see if I can refactor these small inconsistencies there without introducing these new types.

holgerd77:
> I have experimented a bit with these new types today and yesterday but this refactor is blowing up too much. I will try and see to remove some inconsistencies in the EVM/Interpreter, but will do that in the other PR, as this is growing way too large. This refactor would probably also grow so big that we won't be able to review it before our planned release target. I'll go back to the "simpler" tasks on the other PR and see if I can refactor these small inconsistencies there without introducing these new types.
> 
> Also a good choice to open a seperate PR for this one :)

Ok, sounds good. 🙂 Thanks for all these experiements and testing the limits!

jochem-brouwer:
Will close here, this is a mid/long-term goal, #1912 will have to be finished and merged first.

PR #1924, opened 2022-05-31
develop: small fixes

ryanio:
This PR:
* Simplifies setting blockchain genesis parameters from common for future extensibility of additional fields
    * (feedback from https://github.com/ethereumjs/ethereumjs-monorepo/pull/1916#discussion_r885457874)
    * also makes genesis `timestamp` optional instead of `| null` to satisfy BlockHeader type constraints.
* Moves vm's `EOF` file one folder up (couldn't find a reason it needed to be in `vm/opcodes`), export default rather than wildcard import
* Removes last `assert` usages from evm/precompiles, devp2p
    * (no `assert` usage left across the monorepo, woo!)

### DIFF
--- a/packages/blockchain/src/index.ts
+++ b/packages/blockchain/src/index.ts
@@ -214,12 +214,6 @@ export default class Blockchain implements BlockchainInterface {
    * {@link BlockchainOptions}.
    */
   protected constructor(opts: BlockchainOptions = {}) {
-    // Throw on chain or hardfork options removed in latest major release to
-    // prevent implicit chain setup on a wrong chain
-    if ('chain' in opts || 'hardfork' in opts) {
-      throw new Error('Chain/hardfork options are not allowed any more on initialization')
-    }
-
     if (opts.common) {
       this._common = opts.common
     } else {
--- a/packages/blockchain/src/index.ts
+++ b/packages/blockchain/src/index.ts
@@ -424,11 +424,7 @@ export default class Blockchain implements BlockchainInterface {
   async getIteratorHead(name = 'vm'): Promise<Block> {
     return await this.runWithLock<Block>(async () => {
       // if the head is not found return the genesis hash
-      const hash = this._heads[name] ?? this._genesisBlock?.hash()
-      if (!hash) {
-        throw new Error('No head found.')
-      }
-
+      const hash = this._heads[name] ?? this.genesisBlock.hash()
       const block = await this._getBlock(hash)
       return block
     })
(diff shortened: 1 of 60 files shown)


packages/blockchain/src/index.ts
```diff
@@ -215,12 +214,6 @@ export default class Blockchain implements BlockchainInterface {
    * {@link BlockchainOptions}.
    */
   protected constructor(opts: BlockchainOptions = {}) {
-    // Throw on chain or hardfork options removed in latest major release to
-    // prevent implicit chain setup on a wrong chain
-    if ('chain' in opts || 'hardfork' in opts) {
-      throw new Error('Chain/hardfork options are not allowed any more on initialization')
-    }
```
ryanio, line 222: I removed this since it would be two major releases ago for develop, which I think is far enough to remove the legacy check, and TypeScript should guard downstream users against these options being used.

packages/vm/src/evm/evm.ts
```diff
@@ -22,7 +22,7 @@ import EEI from './eei'
 import { ERROR, VmError } from '../exceptions'
 import { default as Interpreter, InterpreterOpts, RunState } from './interpreter'
 import Message, { MessageWithTo } from './message'
-import * as eof from './opcodes/eof'
+import EOF from './eof'
```
holgerd77, line 25: Ah, yes, that's better. I was actually me who introduced the structure if I remember correctly.

packages/devp2p/src/dns/enr.ts
```diff
@@ -73,7 +70,7 @@ export class ENR {
       obj.secp256k1
     )
 
-    assert(isVerified, 'Unable to verify ENR signature')
+    if (!isVerified) throw new Error('Unable to verify ENR signature')
```
holgerd77, line 73: Ah, also nice here.

packages/client/bin/cli.ts
```diff
@@ -371,7 +371,7 @@ async function startClient(config: Config, customGenesisState?: GenesisState) {
       validateBlocks: true,
       validateConsensus,
     })
-    setCommonForkHashes(config.chainCommon, blockchain.genesisBlock().hash())
+    setCommonForkHashes(config.chainCommon, blockchain.genesisBlock.hash())
```
holgerd77, line 374: Yes, also nicer. 🙂

packages/blockchain/src/index.ts
holgerd77: 👍

holgerd77 (approved):
Yes, all fine! 🙂 👍

Issue #1889, opened 2022-05-17
Block: Directly throw in validation functions


holgerd77:
Part of #1717
Builds upon #1879 

Another last-round breaking release task:

Directly throw in validation functions like `BlockHeader.validateGasLimit()`, throw with more appropriate error messages (e.g. EIP-1559 specific), see https://github.com/ethereumjs/ethereumjs-monorepo/pull/1148#issuecomment-834239077.

There are various validation functions (also e.g. validateDifficulty(), validateCliqueDifficulty()) which just operate on a true/false return basis and the question is if we also want to move these to a return-void-or-throw-error scheme like already being done in the broader validate() function.

There should be a really close look in both `BlockHeader` and `Block` if all respective validation functions have been "caught". 😋

Will assign @emersonmacro Emmett here, since he has already taken on a Block-validation-methods refactoring task and I guess this would be a good fit.

PR #1925, opened 2022-06-01
block: low-level validation method refactor

### LINKED ISSUE #1879: Block: Low-Level Validation Method Refactor
This is a (relatively) modest proposal to refactor the Block validation methods and logic as one of the last bigger tasks on the way to the breaking releases. Major outcome would be that we would free ourself from the problematic (and in some sense circular) `Blockchain` dependency and that subclassing of `Block` or `Header` for consensus specific customizations would get easier.

Lots of little tasks which will sum up to - hopefully - a consistent set of improvements:

### BlockHeader

- [ ] `_validateHeaderFields()` called in the constructor should be renamed to `_genericFormatValidation()`
- [ ] `Header.validate()`:
  - [ ] The first upper consensus-specific checks are misplaced here since these are just static format checks as well. These should be moved to a separate `_consensusFormatValidation()` method called from the constructor after `_genericFormatValidation()`. This also allows for easier consensus-specific replacement on subclassing
  - [ ] The rest of the `validate()` functionality (Including the following consensus-specific checks) should be moved to a new method in the `Blockchain` class `validateBlockHeader(header: BlockHeader, height)`
- [ ] The `validateDifficulty()` and `validateCliqueDifficulty()` methods should be unified to `validateDifficulty()` with appropriate consensus checks
  - [ ] Subtask: same for the `Block` class

### Block

- [ ] `validateUncles()`: method should be moved to `Blockchain` as `validateBlockUncles(block: Block)`
  - [ ] Same with private `_validateUncleHeaders()` method called from within the above method
- [ ] `validate()`: method can be removed without replacement
- [ ] Bonus task: The `onlyHeader` param on the `validateData()` method is a bit quirky, wonder if this can be removed along the refactoring (or in general)?
- [ ] TBD - also on th `validateData()` method: this method is a bit the corresponding method to `_genericFormatValidation()` in `BlockHeader` if I am not mistaken. I wonder if we should align here and a) remove this method (respectively rename to private `_genericFormatValidation()`) and then call from the constructor? 🤔

### Generic

- [ ] Remove `@ethereumjs/blockchain` dependency (good check if nothing gots forgotten)

I think @emersonmacro would be a good candidate for this task, since he already dealt with the `Blockchain` refactoring and this is going somewhat along the same line. 🙂 Haven't got a confirmation from him though yet if he is available and think this fits.

emersonmacro:
Closes #1879 and #1889


emersonmacro:
@holgerd77 @ryanio This task seemed fairly straightforward at the start, but I started to struggle with it as things started moving around as part of the refactor. I had branched off of `develop` almost a month ago so it was out of date. Today I pulled down latest `develop`, created a new branch, and applied my changes on top, but Ryan, I'm worried I may have missed some of your recent changes from your genesis refactor PR.

Also I know I had some failing tests in `block` and `blockchain` that I haven't been able to figure out yet (currently marked with `TODO`'s in the test files), but the CI run seemed to uncover a few more just now (in `client`, `ethash`, `test-examples`, `vm`). I can keep trying to get them fixed tomorrow evening, but if either of you have time to take a look and have any suggestions, that would be very helpful. Basically I'm at the point where I could use some more eyes on this to get it back on track.

packages/block/test/block.spec.ts
```diff
@@ -232,9 +184,16 @@ tape('[Block]: block functions', function (t) {
   })
 
   t.test('should test transaction validation with legacy tx in london', async function (st) {
+    // TODO: fix this test
+    // failing to initialize because mixHash is not filled with zeroes. is this a problem with
+    // the testdata, or a problem with the code?
```
acolytec3, line 189: I think this is a problem with the data.  The block used in this test isn't a valid block under clique consensus rules given the mixHash has a non-zero value.

holgerd77:
(so haven't looked too much into the code yet)

emersonmacro:
@holgerd77 Sorry, I screwed this up. I guess I was thinking about it wrong, that one big commit would make it easier to cherry-pick, but didn't factor in conflicting changes or moving large blocks of code between packages. I think I'm going to start over and reapply the changes in smaller chunks based on what's currently in develop, with a close eye on making sure Ryan's recent changes are retained.

holgerd77:
> @holgerd77 Sorry, I screwed this up. I guess I was thinking about it wrong, that one big commit would make it easier to cherry-pick, but didn't factor in conflicting changes or moving large blocks of code between packages. I think I'm going to start over and reapply the changes in smaller chunks based on what's currently in develop, with a close eye on making sure Ryan's recent changes are retained.

No problem, that's just part of the learning process to occasionally screw things up temporarily. 🙂 Nothing which can't be fixed. Yes, these one-commits-with-large-chainset really *only* work if one already has a consistent and working change set (e.g. when we squash together commits after approval), otherwise things can be a lot better controlled (and also: reviewed) if commits are modular and small.

I think the best way is really to start over and reapply, then it will also be best to really push very often, so after 2-3 atomic change commits e.g., and make sure on every run that the CI keeps passing in between. Then it should really be possible to get things under control.

Please wait with this until we (I) have pushed the latest rebased `develop`, so once @g11tech is ready with #1928. Then we are directly on the latest state and can likely also just merge your PR in the submitted form. 🙂 

Thanks for your patience and endurance!

emersonmacro:
This branch is fairly out of sync with current `develop`. Closing this PR and moving to a new branch.

PR #1931, opened 2022-06-02
Blockchain: Clique-recently-signed False Positive Fix

### LINKED ISSUE #1930: Client: Release-blocking Client Bug
We have a too severe bug in the client to continue with the releases.

So basically: if a chain has a partially synced state (or is just synced from the beginning) and then the client is re-started we repeatedly get the following error:

grafik

(here is the initial sync for comparison, going well:)

grafik

I can confirm this for at least Rinkeby and Sepolia, for Mainnet things seem to work (so maybe this is a PoA related thing?).

This bug is *urgent* since it is actively blocking the current release round.

holgerd77:
Fixes #1930 

Triggered in the client in a reorg-like scenario on syncing a Clique consensus-based chain (Sepolia, Rinkeby,...)

### DIFF
--- a/packages/blockchain/src/index.ts
+++ b/packages/blockchain/src/index.ts
@@ -478,6 +478,9 @@ export default class Blockchain implements BlockchainInterface {
     // construct recent block signers list with this block
     let signers = this._cliqueLatestBlockSigners
     signers = signers.slice(signers.length < limit ? 0 : 1)
+    if (signers.length > 0 && !signers[signers.length - 1][0].eq(header.number.subn(1))) {
+      signers = []
+    }
     signers.push([header.number, header.cliqueSigner()])
     const seen = signers.filter((s) => s[1].equals(header.cliqueSigner())).length
     return seen > 1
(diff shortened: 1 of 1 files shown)


g11tech (approved):

packages/blockchain/src/index.ts
```diff
@@ -478,6 +478,9 @@ export default class Blockchain implements BlockchainInterface {
     // construct recent block signers list with this block
     let signers = this._cliqueLatestBlockSigners
     signers = signers.slice(signers.length < limit ? 0 : 1)
+    if (signers.length > 0 && !signers[signers.length - 1][0].eq(header.number.subn(1))) {
+      signers = []
```
ryanio, line 482: Should we just return false here? With an explanatory comment like `// if the last signed block is not one from the current head we do not have a complete picture of the state to verify if too recently signed`

holgerd77: Yes, guess that's the same, right?

holgerd77: (but definitely better)

holgerd77: Can you reapprove here?

ryanio (approved):
lgtm!

Issue #1615, opened 2021-12-16
Unnecessary enforcing `esModuleInterop` for users of the package 


nazarhussain:
Currently the `BN` package imported as default import causing the users of the ethereumjs to always enable `esModuleInterop` in their projects. 

https://github.com/ethereumjs/ethereumjs-monorepo/blob/0c7bdfcdf5866fc31df2b87508156f3294bbae45/packages/util/src/externals.ts#L7

This is not necessary if we fix the import syntax as. 

```
import * as BN from 'bn.js'
```

The error log: 

```
./../node_modules/ethereumjs-util/dist/constants.d.ts:2:8 - error TS1259: Module '"/web3.js/node_modules/@types/bn.js/index"' can only be default-imported using the 'esModuleInterop' flag

2 import BN from 'bn.js';
         ~~

  ../../node_modules/@types/bn.js/index.d.ts:594:1
    594 export = BN;
```

nazarhussain:
For reference: https://github.com/ChainSafe/web3.js/pull/4607/files#r768699324

ryanio:
I'm open to changing this if there aren't any other objections or anticipated downsides.

Thanks @nazarhussain. Would you like to open a PR, or I can?

This is the way bn.js sets its exports ([link](https://github.com/indutny/bn.js/blob/4187ca213e91b41acf72be046072f2dc1f06d0de/lib/bn.js#L42)):

```js
  if (typeof module === 'object') {
    module.exports = BN;
  } else {
    exports.BN = BN;
  }
```

nazarhussain:
Seems like the right change. @ryanio Go ahead and open the PR if you have time.

ryanio:
@nazarhussain I started to work on this PR but when I change the import to your suggestion I get this error:
> Type originates at this import. A namespace-style import cannot be called or constructed, and will cause a failure at runtime. Consider using a default import or import require here instead.

To take its suggestion, an alternative could be to use `import BN = require('bn.js')`?

ryanio:
i will be opening an ESM PR for develop today that will disable esModuleInterop by default to solve this issue for downstream users

PR #1928, opened 2022-06-02
Develop rebased on 3e4e7be

g11tech:
rebased on
```
commit 3e4e7bed681bfa6d1c95abbcc28f6508e828339d (HEAD -> master, origin/master, origin/HEAD)
Author: Holger Drewes <Holger.Drewes@gmail.com>
Date:   Thu Jun 2 22:14:19 2022 +0200

    Blockchain: Clique-recently-signed False Positive Fix (#1931)
    

```

PR #1936, opened 2022-06-05
develop: esm support

ryanio:
This PR adds ES modules support to develop, making it the primary module resolution method by adding `"type": "module"` to the `package.json`s.

There is still some final work needed to complete all the pieces. Currently local tests are passing but karma is having some trouble, more details about karma in the bottom of this post.

This PR:

- Makes `dist` a container folder with folders inside: `esm`, `cjs`, and `types`
- Removes `dist.browser` build
	- We used to export separate browser builds to support ES5, but now that we are using native BigInts the browser builds also output ES2020, so there is no benefit or differentiation for a separate browser build.
- Adds eslint rule `node/file-extension-in-import` to automatically add `.js` file extensions (really nice to have).
- A couple node_modules have hotfix workarounds on postinstall scripts. For the ones I haven't opened issues on the respective repositories, would be nice to eventually do so the hotfixes can be removed in the future.
- Sets esModuleInterop and allowSyntheticDefaultImports back to false (see [here](https://www.semver-ts.org/#module-interop) and #1615)

`package.json` exports now looks like this:

```json
  "main": "./dist/cjs/index.js",
  "types": "./dist/types/index.d.ts",
  "browser": "./dist/esm/index.js",
  "exports": {
    ".": {
      "import": "./dist/esm/index.js",
      "require": "./dist/cjs/index.js",
      "types": "./dist/types/index.d.ts",
      "browser": "./dist/esm/index.js"
    },
    "./*": {
      "import": "./dist/esm/*",
      "require": "./dist/cjs/*",
      "types": "./dist/types/*",
      "browser": "./dist/esm/*"
    }
  },
```

this allows for top-level exports, e.g.:

`import VM from '@ethereumjs/vm`
`import Common, { Chain } from '@ethereumjs/common'`

while also providing deep-access to files inside in case they are not exported at the top level, e.g.:

`import Bloom from '@ethereumjs/vm/bloom/index.js'`
`import { GenesisState } from '@ethereumjs/blockchain/genesisStates/index.js'`

---

I haven't been able to get the commonjs typescript build working yet since it complains about some of the esm changes, so one approach might be to instead build the commonjs version with webpack from the esm code. Additionally single file browser bundles could be added so others can easily use these packages from the browser:

```json
  "exports": {
    ...
    "./esm-browser-bundle": "./dist/bundles/esm.js",
    "./iife-browser-bundle": "./dist/bundles/iife.js",
    "./umd-browser-bundle": "./dist/bundles/umd.js",
  }
```

@MicahZoltu shared his thoughts about bundling in discord:

> I generally recommend against libraries doing bundling.  Bundling should be done (IMO) at the application layer.
> Under this paradigm, you would publish ESM and CJS builds of your library, and you wouldn't bundle in transitive dependencies.
> The caveat here is that if you have any dependencies that don't provide ESM builds then things get complicated.
Ideally, you just don't have any non-ESM dependencies.
> And if there is some dependency that you really need which doesn't have an ESM build you would submit a PR to add an ESM build and make the world a better place for everyone.  😀

I would agree here to keep our builds light and modern for the esm future without the burden of compiling the transitive dependencies. This would be aligned with our current strategy (on master branch) to let application layer do the bundling, since each set of needs could be different.

#### karma 

I haven't been able to get karma fully working yet, getting this error on start: `SyntaxError: import declarations may only appear at top level of a module`. If we export browser bundles with e.g. webpack, then it might make sense to switch from `karma-typescript` to `karma-webpack` for better e2e testing.

(I only suggest webpack because we already use it in the client. If rollup is a better experience, then that can also be used and the client build can also be switched over to it to share the same deps.)


ryanio:
Some more digging into karma, I thought these acorn options might help but no luck so far. Still getting `Uncaught SyntaxError: Cannot use import statement outside a module`. I suspect it's routing it through commonjs bundling instead of trying to use esm.

```
        acornOptions: {
          ecmaVersion: 11,
          sourceType: 'module',
          allowImportExportEverywhere: true,
        },
```

There's some lengthy information about the bundling here https://github.com/monounity/karma-typescript/tree/master/packages/karma-typescript#automatic-bundling and it says
> However, this intentional behavior makes it possible to use karma-typescript for projects that use the Typescript module system, or have the compiler option module set to another value than commonjs, or simply put everything on the global scope.

So this should be supported.

MicahZoltu:
Why are the types separated from the module output?  One issue with this is that the type definitions for `cjs` may not be the same as the type definitions for `esm`, so they generally *shouldn't* share type definition files.  Also, it is a defacto standard to include TS definitions next to the output files and source maps, and following that defacto standard is more likely to result in stuff "just working" without the user needing to do additional work.

config/tsconfig.base.json
```diff
@@ -0,0 +1,16 @@
+{
+  "compilerOptions": {
+    "declaration": true,
+    "downlevelIteration": true,
+    "lib": ["ES2020", "DOM"],
+    "module": "node16",
+    "moduleResolution": "node16",
+    "resolveJsonModule": true,
+    "sourceMap": true,
```
MicahZoltu, line 9: Recommend `declarationMap: true` as well (https://www.typescriptlang.org/tsconfig#declarationMap).

packages/block/src/block.ts
```diff
@@ -11,8 +11,8 @@ import {
   Transaction,
   Capability,
 } from '@ethereumjs/tx'
-import { BlockHeader } from './header'
-import { BlockData, BlockOptions, JsonBlock, BlockBuffer, Blockchain } from './types'
+import { BlockHeader } from './header.js'
+import { BlockData, BlockOptions, JsonBlock, BlockBuffer, Blockchain } from './types.js'
```
MicahZoltu, line 15: Rather than doing this, you could use a compiler plugin to automatically do it.  There is much disagreement on what is "correct" here, but I'm of the opinion that you should leave `./header` and let the compiler add the extension.  The reason for this is because you aren't actually importing `./header.js`, you are importing `./header.ts` and then after transpilation it should turn into `./header.js`.

If you want to have the compiler automatically add the extension, rather than adding them manually, you can use the extension here:
https://github.com/Zoltu/typescript-transformer-append-js-extension/

holgerd77:
Some procedural note: this is still so highly invasive that I am not sure if we can merge this in early respectively "in time", so I would consider to not include this in the first round of beta releases for the breaking releases.

Since other work is waiting on a new `develop` branch basis I would still continue and replace `develop` with the rebased version once #1928 is ready, hope this is not causing too much hazzle regarding this PR.

acolytec3:
I will also say I don't love the idea of adding a webpack build step to every one of our libraries.  If the build step is half as slow as it is for our client build step at present, it will likely greatly increase the time it takes for an `npm run clean && npm i` refresh of the repo (which we have to do somewhat regularly at present).

MicahZoltu:
> @MicahZoltu Could you _very concretely_ outline what this would specifically mean so that we make sure that we are on the same page? Not using the `types` location reference in `package.json`? 🤔

The fix I am proposing would just involve removing the `types` keys from `package.json` and removing `declarationDir` key from `tsconfig.json`.

ryanio:
i've separated out tasks from this PR to #1946, will close this PR for now, @MicahZoltu pls take a look at the issue and feel free to suggest anything else in there.

PR #1940, opened 2022-06-07
Util: Signature Code Cleanup (old)

holgerd77:
Cleanup of the Util signature code as one of the remaining tasks for #1914, WIP.

This will be re-applied once `develop` gets merged.

Already open for first looks.


holgerd77:
I would have a strong tendency to remove all `BigIntLike` usages here in the signature module, let me know if you feel otherwise. Calling into the signing functionality is so targeted (e.g. compared to the broad "How to instantiate a Transaction?" use cases) and I have the impression that this is rather just very much bloating the API (and things like test cases), also considering that one can't get completely consistent anyhow since things are just coming "in and out" (some methods take `v` (respectively a whole signature) in - where one can allow `BigIntLike` - some give it as an output/return value - where one needs to limit to `BigInt` anyhow).

So will proceed with this - feedback nevertheless welcome.

holgerd77:
Closed in favor of #1945

Issue #1934, opened 2022-06-05
devp2p: deduplicate `secp256k1` and `keccak` dependencies


ryanio:
The devp2p package uses its own dependencies for `secp256k1` and `keccak`.

These should ideally be deduplicated and use `ethereum-cryptogrpahy/keccak` and `ethereum-cryptography/secp256k1` like the rest of the monorepo on the `develop` branch.

ScottyPoi:
I have this mostly worked out.
There's one method from `ethereum-cryptography` for which I can't produce the same results as we get from `secp256k1`

This is in `rlpx/ecies.ts`

```
function ecdhX(publicKey: Buffer, privateKey: Buffer) {
  function hashfn(x: Uint8Array, y: Uint8Array) {
    const pubKey = new Uint8Array(33)
    pubKey[0] = (y[31] & 1) === 0 ? 0x02 : 0x03
    pubKey.set(x, 1)
    return pubKey
  }
  // @ts-ignore
  return Buffer.from(ecdh(publicKey, privateKey, { hashfn }, Buffer.alloc(33)).slice(1))
}
```

Replacing this:
```
import {ecdh } from 'secp256k1'
```
With this:
```
import {  ecdh } from 'ethereum-cryptography/secp256k1-compat'
```
Gives a similar method, which works with some tweaking, but causes a few tests to fail like this:
```
    operator: throws
    expected: |-
      undefined
    actual: |-
      [Error: should have valid tag: 42670be82545bddbd98525ba99f3ee36cf343e213844aa901c646f0488ac5840 / e2d5b5340c9e488eefea198576344afbdf66db5f51204a6961a63ce072c8926c]
    at: <anonymous> (/ethereumjs-monorepo/packages/devp2p/test/rlpx-ecies.ts:132:7)
```

The error is coming from that `ecdhX` method

paulmillr:
@ScottyPoi could you please write an exact input that produces different outputs between these two `ecdh` functions, without the need to run tests?

ScottyPoi:
Thanks @paulmillr -- but I managed to work out what was happening with that.  Just had to tweak the way the hash function was working.
```
function ecdhX(publicKey: Buffer, privateKey: Buffer) {
  function hashfn(x: Uint8Array, y: Uint8Array) {
    const pubKey = new Uint8Array(33)
    pubKey[0] = (y[31] & 1) === 0 ? 0x02 : 0x03
    pubKey.set(x, 1)
[...]

PR #1947, opened 2022-06-08
Devp2p-ethereum-cryptography-dependency

ScottyPoi:
Resubmitting on updated master.
Fixes for #1934

Removing `secp256k1` and `keccak` dependencies in favor of `ethereum-cryptography`

- `secp256k1` removal pretty straight forward.  Most of the functions were indentical.  
- `ecdh` method required some changes to the hash function to make `ecdhX()` work
- `keccak` class used in `rlpx/mac.ts` is not exposed by `ethereum-cryptography`  
  -  Talking to PaulM about a fix, possibly on the `ethereum-cryptography` side.
  
- Used `@noble/hashes` import for Keccak in `rlpx/mac.ts`
- Replaced `hi-base32` dependency with `@scure/base` from @paulmillr


packages/devp2p/package.json
```diff
@@ -49,6 +49,7 @@
     "bl": "^1.1.2",
     "debug": "^4.3.3",
     "ethereumjs-util": "^7.1.5",
+    "ethereum-cryptography": "^1.0.3",
     "hi-base32": "^0.5.0",
```
paulmillr, line 53: You should also remove this package, since ethereum-cryptography uses `@scure/base`, which contains audited implementation of base32/bech32.

paulmillr:
```
import { base32 } from '@scure/base';
```

ScottyPoi: Would I need to add `@scure/base` as a new dependency in devp2p or can i access that from  ethereum-cryptography somehow?

paulmillr: We've decided against exposing base32 because ETH apps don't need it. This is the first time I see an eth primitive requiring it. 

We can try exposing base32 in e-c, but i'd rather not to, at least for now. 

You can use the same scure/base version we're using in e-c for now, which is 1.1.0. You can use it as-is, or a range, I recommend ~1.1.0.

paulmillr:
We have just released ethereum-cryptography@1.1.0, which exposes `keccak256.create`, so you can use it directly instead of `@noble/hashes`.

paulmillr:
Note that it doesn't re-export `Hash`, but you can use `type Hash = ReturnType<typeof keccak256.create>;`

packages/devp2p/src/util.ts
```diff
@@ -11,17 +10,17 @@ export const devp2pDebug = createDebugLogger('devp2p')
...
-  const privateKey = randomBytes(32)
-  return privateKeyVerify(privateKey) ? privateKey : genPrivateKey()
+  const privateKey = utils.randomPrivateKey()
+  return utils.isValidPrivateKey(privateKey) ? Buffer.from(privateKey) : genPrivateKey()
 }
 
 export function pk2id(pk: Buffer): Buffer {
   if (pk.length === 33) {
-    pk = Buffer.from(publicKeyConvert(pk, false))
+    pk = Buffer.from(getPublicKey(pk, false))
```
paulmillr, line 23: This method took 33 or 65-byte public key and converted it to 65-byte decompressed form. Now, it would treat `pk` as `private key` and convert it to `public key`. And private keys cannot be 33-byte, so there's an error in any case.

packages/devp2p/src/rlpx/mac.ts
```diff
@@ -1,12 +1,13 @@
 import { createCipheriv } from 'crypto'
-import createKeccakHash from 'keccak'
+import { Keccak, keccak_256 } from '@noble/hashes/sha3'
```
paulmillr, line 2: Can be replaced with `import { keccak256} from 'ethereum-cryptography/keccak';` now.

paulmillr (changes requested):

packages/devp2p/package.json
```diff
@@ -42,24 +42,24 @@
   },
   "dependencies": {
     "@ethereumjs/common": "^2.6.4",
+    "@scure/base": "1.1.0",
```
paulmillr, line 45: 1.1.1 now

packages/devp2p/package.json
```diff
@@ -42,24 +42,24 @@
...
     "@types/bl": "^2.1.0",
     "@types/k-bucket": "^5.0.0",
     "@types/lru-cache": "^5.1.0",
     "base64url": "^3.0.1",
     "bl": "^1.1.2",
     "debug": "^4.3.3",
     "@ethereumjs/util": "^8.0.0",
-    "hi-base32": "^0.5.0",
+    "ethereum-cryptography": "^1.0.3",
```
paulmillr, line 53: `^1.1.0` now, better to enforce new version/changes

packages/devp2p/src/util.ts
paulmillr: There's `publicKeyConvert` method inside secp256k1-compat

paulmillr (approved):

PR #1951, opened 2022-06-10
client: Miscellaneous fixes for beacon sync in reverseblockfetcher, skeleton, merge forkhash 

g11tech:
Miscellaneous fixes for beacon sync in reverseblockfetcher, skeleton, merge forkhash

Discovered while running beacon sync on the kiln.

### DIFF
--- a/packages/client/lib/sync/fetcher/blockfetcherbase.ts
+++ b/packages/client/lib/sync/fetcher/blockfetcherbase.ts
@@ -68,12 +68,13 @@ export abstract class BlockFetcherBase<JobResult, StorageItem> extends Fetcher<
 
     while (count >= BigInt(max) && tasks.length < maxTasks) {
       tasks.push({ first: first, count: max })
-      first += BigInt(max)
+      !this.reverse ? (first += BigInt(max)) : (first -= BigInt(max))
       count -= BigInt(max)
       pushedCount += BigInt(max)
     }
     if (count > BigInt(0) && tasks.length < maxTasks) {
       tasks.push({ first: first, count: Number(count) })
+      !this.reverse ? (first += BigInt(count)) : (first -= BigInt(count))
       pushedCount += count
       count = BigInt(0)
     }
--- a/packages/client/lib/sync/skeleton.ts
+++ b/packages/client/lib/sync/skeleton.ts
@@ -406,8 +406,8 @@ export class Skeleton extends MetaDBManager {
       const block = await this.getBlock(number)
       if (!block) break
       // Insert into chain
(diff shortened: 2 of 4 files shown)


holgerd77:
Hi Gajinder, as a reminder: these things now needs to be double-applied to both `master` and [v5-maintenance](https://github.com/ethereumjs/ethereumjs-monorepo/tree/v5-maintenance) if you (we) want to have a release on this during the next 5-6 weeks (which I guess we want to have some follow-up client releases (?)).

holgerd77:
(so for client changes, you should generally see `v5-maintenance` as your "working branch" and this `master` submission is rather the "extra" thing)

g11tech:
am currently running beacon sync on this branch to see if everything is good, so might hold of merging till i can fully validate (moving PR to draft then)

holgerd77:
Just thinking out loudly: *in case* we get the client up-and-running relatively quickly on `master`, it might actually also be a way to also do Client beta releases, and point Merge users (so: rather the "Pro" Merge testing guys) to these beta releases. Then we would also have an additional "test bed" for the other beta-release libraries, so this might be even beneficial for the release process itself?

WDYGT? 🤔

(but depends a lot on how "hazzle-free" the client can get up and running)

g11tech:
> WDYGT

I think there are still couple of issues pending that i can see in kiln run, but we most likely will be resolved in 4-5 days, we can go for client releases then :+1:

packages/client/lib/sync/fetcher/blockfetcherbase.ts
```diff
@@ -68,12 +68,13 @@ export abstract class BlockFetcherBase<JobResult, StorageItem> extends Fetcher<
 
     while (count >= BigInt(max) && tasks.length < maxTasks) {
       tasks.push({ first: first, count: max })
-      first += BigInt(max)
+      !this.reverse ? (first += BigInt(max)) : (first -= BigInt(max))
```
acolytec3, line 71: LGTM though I guess I'm curious, do the existing reverse fetcher tests not catch this?  What's the difference between this scenario and whatever is happening today?  And if it is different, is it possible to add a test to cover this particular scenario?

g11tech: I will check and add some tests, though some other things seemingly need to be fixed as well.

acolytec3 (dismissed):
LGTM, with one minor nit that you can take or leave.

acolytec3 (approved):

acolytec3 (approved):
LGTM

PR #1954, opened 2022-06-11
Block: low-level validation methods refactor

emersonmacro:
Closes https://github.com/ethereumjs/ethereumjs-monorepo/issues/1879 and https://github.com/ethereumjs/ethereumjs-monorepo/issues/1889

Previous branch in #1925 got stale, so opening a new branch/PR


holgerd77:
Have taken a deeper look here, there are still some basic problems:

1. This should have been targeted towards `master` and not `develop`, not sure if this was some miscommunication? 🤔

2. The separation lines on the commits here are still not optimal, the base line on this would have been that things (tests) should work after every single commit. Theoretically this should be very much possible since all outlined tasks from the initial issue description from #1879 are very much atomic "we move this from a to b" tasks. With commits like https://github.com/ethereumjs/ethereumjs-monorepo/pull/1954/commits/9f1dad097d3d8392033d4088708980c078c2aa4e - removing methods on the one side but not adding to the other - we are again back in the situation that things get into an inconsistent and hard-to-debug in-between state.

I will move this out of the Beta-1 releases, this is getting too short termed.

Emmett likely won't have the time to finish. If someone wants to pick this up we might still be able to integrate in Beta-2 and still integrate in the breaking releases, otherwise it is also no total drama / irretrievable loss if we skip. Would still be nice to integrate though.

holgerd77:
Superseeded by #1959, will close.

PR #1950, opened 2022-06-10
Evm refactor (rebased on master 8a65e19, 2022-06-08)

holgerd77:
This is a cherry-picked version of #1912 rebased on the latest `master` (up to 8a65e19).

This can be seen as a test rebase or taken as a new work state. #1912 has been integrated up to 00e29e4dfdfb95bf9f6744a6e800548965124033, a total of 36 commits.

Rebase went relatively well, roughly 7-9 conflict resolution needs.

PR #1955, opened 2022-06-13
Evm refactor (rebased on master a4c379a, 2022-06-13)

holgerd77:
Rebased version and potential replacement for #1950 

Rebase was a bit more hairy than expected 😋. 17 tests failing (VM -> API), most seem to be easy fixes (some base object not present, some EVM result not available "as is".

Nevertheless did not directly force-pushed on the original branch, but I would assume that it's worth to take this over here, fix the tests and then continue from here.


holgerd77:
Rebased this after a local test rebase, also did a backup of the branch state before the rebase.

Will push soon afterwards with a VM api test fix.

holgerd77:
Will leave some generic TODOs here, absolutely don't need to be answered right now.

- We shouldn't just double on the supported EIPs in EVM and VM and instead have a closer look which EIP belongs to what entity (my guess would be that the very most EIPs just only belong to EVM)
- Blockchain and optimally Block dependencies should be removed from EVM
- I am not really sure if I am getting for 100% why Interpreter is encapsulating so much EEI/state accessing logic (again?) but it looks clean and I guess I might be ok with it 😋

Ok, I will actually merge this in here so that we can proceed.

This still need an in-depth after-merge review, but then we can go on and e.g. do the package separation. Review work can be done in separate PRs.

Uff. No way back anyhow. 😜 😅 🥳

holgerd77 (commented):
Looks good.

Still needs an in-depth post-merge review.

Will merge by admin-merge for procedural reasons.

PR #1556, opened 2021-11-08
client: save receipts and implement getLogs, getTxReceipt, getTxByHash

ryanio:
This PR adds receipt and log saving functionality to the client.

A new leveldb instance is added called `metaDB` which stores this data and can be used for more data/caching/indexes in the future.

The new `ReceiptsManager` handles receipt and log saving and lookup, along with the `txHash -> [blockHash, txIndex]` index that is limited by `--txLookupLimit` (`default = 2350000 blocks = about one year, 0 = entire chain`)

Enable with `--saveReceipts` and use with `--rpc` to enable `eth_getLogs`, `eth_getTransactionReceipt` and `eth_getTransactionByHash` endpoints.

PR #1956, opened 2022-06-13
Fix client integration tests

acolytec3:
This PR re-enables the Client integration tests that were accidentally removed in #1556 and also fixes some broken integration tests related to beacon sync.

### DIFF
--- a/packages/client/test/integration/mocks/mockchain.ts
+++ b/packages/client/test/integration/mocks/mockchain.ts
@@ -29,7 +29,6 @@ export default class MockChain extends Chain {
         {
           header: {
             number: number + 1,
-            difficulty: 1,
             parentHash: number ? blocks[number - 1].hash() : this.genesis.hash(),
           },
         },
--- a/.github/workflows/client-build.yml
+++ b/.github/workflows/client-build.yml
@@ -45,6 +45,8 @@ jobs:
           flags: client
         if: ${{ matrix.node-version == 16 }}
 
+      - run: npm run test:integration
+      
       - run: npm run lint
   test-client-cli:
     runs-on: ubuntu-latest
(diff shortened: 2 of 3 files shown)


packages/client/test/integration/mocks/mockchain.ts
```diff
@@ -29,7 +29,6 @@ export default class MockChain extends Chain {
         {
           header: {
             number: number + 1,
-            difficulty: 1,
```
g11tech, line 32: is this only being used for PoS blocks?

acolytec3: No, we have other integration tests that use the `mockChain`.  We don't have any reason to hard code this value though since the Block constructor will check the hardfork and give us a reasonable default (i.e. 0 if the consensus in `common` is PoS

packages/client/test/integration/mocks/network.ts
```diff
@@ -45,13 +45,16 @@ export function destroyStream(id: string, location: string) {
...
-  if (servers[location]) {
-    for (const id of Object.keys(servers[location].streams)) {
-      destroyStream(id, location)
+export async function destroyServer(location: string) {
+  return new Promise<void>((resolve) => {
+    while (servers[location]) {
+      for (const id of Object.keys(servers[location].streams)) {
+        destroyStream(id, location)
+      }
+      delete servers[location]
```
g11tech, line 54: but by the time we resolve this all servers would be deleted (i think)

acolytec3: You're welcome to experiment with other approaches.  All I can say is this seems to deterministically solve the issue of "server already exists at this location" errors intermittently appearing in the integration test.  It's just a brute force way of forcing the servers to get cleared.  I still think a long term better approach would be to somehow remove that global servers object that's in `./network.ts` or else revise our `destroy` method to just take an array of servers/services and then blow them all away and resolve once the servers object is empty

PR #1959, opened 2022-06-14
Block validation methods, take 3

### LINKED ISSUE #1879: Block: Low-Level Validation Method Refactor

acolytec3:
Fixes #1879 and #1889

- Removes `block.validate` and `blockheader.validate` and moves the format/basic consensus checks into helper methods that are called in the respective constructors.  These helpers throw if invalid fields are found
- Creates two new methods in `blockchain`, `validateBlock` and `validateHeader` that are called when a block is inserted into the chain.  All of the checks from the removed `validate` methods that require a blockchain context are now done in these methods.
- Move `validateUncles` to `blockchain
- Rewrite block tests to remove `blockchain` dependency where possible and moved to `blockchain` where not
- Some test data was modified (notably modifying the extradata fields to be PoA compliant) to make it work with the tests
- fixes qheap import issue related to #1975


acolytec3:
@holgerd77 here's my WIP on the block validation methods.  `block.spec.ts` tests don't pass at present since I need to address the changes made with moving the header validation methods around but I've taken the first step of moving header validation to blockchain and am going to focus on the difficulty checks next.  Blockchain tests are all currently passing so hopefully won't have to adjust them too much more once I finish the difficulty check updates.

acolytec3:
Correction: all the blockchain tests are passing locally.  CI is failing because I haven't addressed the broader monorepo yet and was focused on getting things working between `block` and `blockchain` and would then address other places where build/tests are failing.

packages/block/src/block.ts
```diff
@@ -261,24 +262,6 @@ export class Block {
...
-   * Throws if invalid.
-   *
-   * @param blockchain - validate against an @ethereumjs/blockchain
-   * @param onlyHeader - if should only validate the header (skips validating txTrie and unclesHash) (default: false)
-   */
-  async validate(blockchain: Blockchain, onlyHeader: boolean = false): Promise<void> {
-    await this.header.validate(blockchain)
-    await this.validateUncles(blockchain)
-    await this.validateData(onlyHeader)
-  }
```
holgerd77, line 281: I think I wrote otherwise in the original issue but I guess we really want to keep this combined version (which would also solve a related client-TODO) in a public `Blockchain.validateBlock()` method.

acolytec3: The thought being we just want to be able to validate a new block in it's entirety without having to actually put it in the chain?

holgerd77: (or even better in the cases where this is possible: both doesn't need to be in the chain, so these would be the cases where a method could even be called in a static context)

acolytec3: I've added a public `blockchain.validateBlock` method that wraps up those three validation checks.  I don't think we can make it a static method without breaking it into further pieces since some the checks are heavily dependent on the existence of the chain.

acolytec3:
@holgerd77 This PR is now ready for review.  I've revised the initial description to reflect the changes made.

Thanks @g11tech for helping track down some of the trickier client test failures!

g11tech (approved):
LGTM!

packages/block/src/header.ts
```diff
@@ -525,35 +501,29 @@ export class BlockHeader {
...
-   * @param blockchain - validate against an @ethereumjs/blockchain
-   * @param height - If this is an uncle header, this is the height of the block that is including it
+   * Checks static parameters related to consensus algorithm
+   * @throws if any check fails
    */
-  async validate(blockchain: Blockchain, height?: bigint): Promise<void> {
-    if (this.isGenesis()) {
-      return
-    }
+  _consensusFormatValidation() {
```
holgerd77, line 526: Can you put this method upper in the method order and place it directly under `_genericFormatValidation()`? Looks a bit misplaced here.

packages/block/test/difficulty.spec.ts
```diff
@@ -12,7 +12,6 @@ function runDifficultyTests(
 ) {
   const dif = block.canonicalDifficulty(parentBlock)
```
holgerd77, line 13: Argh, I am not allow to comment on the `canonicalDifficulty()` method itself. 😜

So here: I would suggest that we rename this to `ethashCanonicalDifficulty()` to align with the clique-specific naming and also place this method further down the line directly above the `clique*()` methods to group these consensus-specific functions together.

packages/block/src/block.ts
```diff
@@ -261,24 +262,6 @@ export class Block {
...
-   *
-   * Throws if invalid.
-   *
-   * @param blockchain - validate against an @ethereumjs/blockchain
-   * @param onlyHeader - if should only validate the header (skips validating txTrie and unclesHash) (default: false)
-   */
-  async validate(blockchain: Blockchain, onlyHeader: boolean = false): Promise<void> {
-    await this.header.validate(blockchain)
-    await this.validateUncles(blockchain)
-    await this.validateData(onlyHeader)
```
holgerd77, line 280: Again, can't comment on the method itself: I would have a tendency to at least rename this to `validateBlockData()`, I always stumble upon this name since data is so ambiguous. If you go along maybe just do.

packages/blockchain/src/consensus/casper.ts
```diff
@@ -14,5 +15,11 @@ export class CasperConsensus implements Consensus {
   public async genesisInit(): Promise<void> {}
   public async setup(): Promise<void> {}
   public async validate(): Promise<void> {}
+  public async validateDifficulty(header: BlockHeader): Promise<void> {
```
holgerd77, line 18: Can we have some new lines between all the methods here (actually also the empty ones)? This looks pretty crambled.

packages/blockchain/src/consensus/clique.ts
```diff
@@ -32,6 +32,11 @@ const CLIQUE_SIGNERS_KEY = 'CliqueSigners'
 const CLIQUE_VOTES_KEY = 'CliqueVotes'
 const CLIQUE_BLOCK_SIGNERS_SNAPSHOT_KEY = 'CliqueBlockSignersSnapshot'
 
+// Block difficulty for in-turn signatures
+export const CLIQUE_DIFF_INTURN = BigInt(2)
+// Block difficulty for out-of-turn signatures
+export const CLIQUE_DIFF_NOTURN = BigInt(1)
```
holgerd77, line 38: Can we change the order of constants and types here (or is it just me who feel that constants would belong to the top?) and then also put all *exported* constants grouped together first?

packages/blockchain/src/index.ts
```diff
@@ -641,6 +646,186 @@ export default class Blockchain implements BlockchainInterface {
...
+        const msg = header._errorMsg('invalid timestamp diff (lower than period)')
+        throw new Error(msg)
+      }
+      // Validate clique difficulty
+      await this.consensus.validateDifficulty(header)
+    }
+
+    if (header._common.consensusType() === 'pow') {
+      await this.consensus.validateDifficulty(header)
+    }
```
holgerd77, line 699: I guess we can (and should) remove this PoW check here, to make sure this works in a generic context.

packages/blockchain/src/index.ts
```diff
@@ -641,6 +646,186 @@ export default class Blockchain implements BlockchainInterface {
...
+    if (this._common.consensusAlgorithm() === ConsensusAlgorithm.Clique) {
+      const period = (this._common.consensusConfig() as CliqueConfig).period
+      // Timestamp diff between blocks is lower than PERIOD (clique)
+      if (parentHeader.timestamp + BigInt(period) > header.timestamp) {
+        const msg = header._errorMsg('invalid timestamp diff (lower than period)')
+        throw new Error(msg)
+      }
+      // Validate clique difficulty
+      await this.consensus.validateDifficulty(header)
+    }
```
holgerd77, line 695: Now we have this *one* consensus-related check back in here which is unlucky. Wonder if there is a good place in `consensus` we can safely move.

Maybe in the `clique.validate()` method as well? 🤔 Might need a double check if the semantic result would be somewhat matching/equivalent.

But we should really target to remove here.

holgerd77 (changes requested):
Thanks, looks already pretty great, made some comments and change requests on things like structure, (method) order, naming and visibility of things.

packages/block/src/header.ts
```diff
@@ -1000,12 +873,12 @@ export class BlockHeader {
...
 
   /**
-   * Internal helper function to create an annotated error message
+   * Helper function to create an annotated error message
    *
    * @param msg Base error message
    * @hidden
    */
-  protected _errorMsg(msg: string) {
+  public _errorMsg(msg: string) {
```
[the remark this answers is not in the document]
acolytec3: Nope, fixed.

packages/blockchain/src/consensus/clique.ts
```diff
@@ -131,6 +136,40 @@ export class CliqueConsensus implements Consensus {
     }
   }
 
+  async validateDifficulty(header: BlockHeader): Promise<void> {
+    header._requireClique('validateCliqueDifficulty')
```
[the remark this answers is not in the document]
acolytec3: Done

packages/blockchain/src/consensus/clique.ts
```diff
@@ -131,6 +136,40 @@ export class CliqueConsensus implements Consensus {
...
+    if (header.difficulty !== CLIQUE_DIFF_INTURN && header.difficulty !== CLIQUE_DIFF_NOTURN) {
+      const msg = header._errorMsg(
```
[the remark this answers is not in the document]
acolytec3: Thanks!  Didn't notice that before

packages/block/src/block.ts
acolytec3: Done

packages/blockchain/src/index.ts
acolytec3: This one is super tricky I've discovered.  Because `this.consensus` is the blockchain's current consensus mechanism, this fails on merge transition blocks where the header is now PoS but the blockchain is still PoW (since the blockchain hasn't switched to PoS so `blockchain.consensus` is `ethashConsensus`.  We should skip this check on the merge block because the PoS difficulty check (i.e. PoS blocks must always have difficulty 0) is done in the block constructor via `block._consensusValidation) and it will always fail if you pass in a new header that's a PoS block but the current chain is either still PoA or PoW.  I'll check for PoS instead and skip if so.

acolytec3:
@holgerd77 I think I've addressed all the feedback at this point.

Two open questions for you/others to consider:

- Decide whether `block.validateTransactions` should throw or not (seems like it shouldn't since it can return an array of errors if multiple transactions fail
- Decide if we should add a flag to the block/blockchain constructors to allow the validation checks to be bypassed (like the `validateBlocks` parameter on the blockchain constructor

packages/block/src/header.ts
```diff
@@ -351,210 +337,38 @@ export class BlockHeader {
...
       }
-      dif = parentDif + offset * a
-    } else {
-      // pre-homestead
-      if (parentTs + this._common.paramByHardfork('pow', 'durationLimit', hardfork) > blockTs) {
-        dif = offset + parentDif
-      } else {
-        dif = parentDif - offset
+      const londonHfBlock = this._common.hardforkBlock(Hardfork.London)
+      const isInitialEIP1559Block = londonHfBlock && this.number === londonHfBlock
```
gabrocheleau, line 353: Since we know that this.number is defined, perhaps we could simplify this a bit:
```suggestion
      const isInitialEIP1559Block = this.number === this._common.hardforkBlock(Hardfork.London)
```

packages/blockchain/src/index.ts
```diff
@@ -641,6 +646,174 @@ export default class Blockchain implements BlockchainInterface {
...
+        )
+      }
+    }
+
+    // check blockchain dependent EIP1559 values
+    if (this._common.isActivatedEIP(1559)) {
+      // check if the base fee is correct
+      let expectedBaseFee
+      const londonHfBlock = this._common.hardforkBlock(Hardfork.London)
+      const isInitialEIP1559Block = number === londonHfBlock
```
gabrocheleau, line 710:
```suggestion
      const isInitialEIP1559Block = number === this._common.hardforkBlock(Hardfork.London)
```
same as above

packages/blockchain/src/index.ts
```diff
@@ -641,6 +646,174 @@ export default class Blockchain implements BlockchainInterface {
...
+   * The following rules are checked in this method:
+   * Uncle Header is a valid header.
+   * Uncle Header is an orphan, i.e. it is not one of the headers of the canonical chain.
+   * Uncle Header has a parentHash which points to the canonical chain. This parentHash is within the last 7 blocks.
+   * Uncle Header is not already included as uncle in another block.
+   * @param block - block for which uncles are being validated
+   */
+  private async _validateUncleHeaders(block: Block) {
+    const uncleHeaders = block.uncleHeaders
+    if (uncleHeaders.length == 0) {
```
gabrocheleau, line 744:
```suggestion
    if (uncleHeaders.length === 0) {
```

packages/blockchain/src/index.ts
```diff
@@ -1187,7 +1360,15 @@ export default class Blockchain implements BlockchainInterface {
       number: 0,
       stateRoot,
     }
-
+    if (common.consensusType() === 'poa') {
```
gabrocheleau, line 1363:
```suggestion
    if (common.consensusType() === ConsensusType.ProofOfAuthority) {
```

packages/vm/examples/run-solidity-contract.ts
```diff
@@ -7,12 +7,14 @@ import { Transaction } from '@ethereumjs/tx'
...
 import { buildTransaction, encodeDeployment, encodeFunction } from './helpers/tx-builder'
 import { getAccountNonce, insertAccount } from './helpers/account-utils'
+import { Block } from '@ethereumjs/block'
 const solc = require('solc')
 
 const INITIAL_GREETING = 'Hello, World!'
 const SECOND_GREETING = 'Hola, Mundo!'
 
 const common = new Common({ chain: Chain.Rinkeby, hardfork: Hardfork.Istanbul })
+const block = Block.fromBlockData({ header: { extraData: Buffer.alloc(97) }},{ common })
```
gabrocheleau, line 17:
```suggestion
const block = Block.fromBlockData({ header: { extraData: Buffer.alloc(97) } }, { common })
```

packages/vm/tests/api/runTx.spec.ts
```diff
@@ -37,8 +38,16 @@ tape('runTx() -> successful API parameter usage', async (t) => {
       const caller = tx.getSenderAddress()
       const acc = createAccount()
       await vm.eei.state.putAccount(caller, acc)
+      let block
+      if (vm._common.consensusType() === 'poa') {
```
gabrocheleau, line 42:
```suggestion
      if (vm._common.consensusType() === ConsensusType.ProofOfAuthority) {
```

gabrocheleau (approved):
Left some minor comments, look good to me overall.

acolytec3:
> Left some minor comments, look good to me overall.

Thanks for the detailed review!  Will address all this cleanup.

acolytec3:
Please don't merge this.  I've got to fix the client tests, again.

holgerd77:
@acolytec3 thanks a lot for addressing all the feedback, this looks really great! 👍 

> Decide whether block.validateTransactions 

Yes, I guess we should not throw here.

> Decide if we should add a flag to the block/blockchain constructors to allow the validation checks to be bypassed (like the validateBlocks parameter on the blockchain constructor

Maybe I am missing something but there is a `validateBlocks` parameter for `Blockchain`? 🤔

There is also a `validateConsensus` flag, looking at that part I realized that we should rename the `consensus.validataBlockData()` methods to `consensus.validateConsensus()`, this makes things a lot more clear.

acolytec3:
> 
> Maybe I am missing something but there is a `validateBlocks` parameter for `Blockchain`? thinking
> 
> There is also a `validateConsensus` flag, looking at that part I realized that we should rename the `consensus.validataBlockData()` methods to `consensus.validateConsensus()`, this makes things a lot more clear.

The reason I ask the question is we have format level consensus checks in the blockheader constructor that require things like PoA blocks have to have minimum 97 bytes of `extraData` which in turn means we have to populate these fields when we're instantiating blocks.  Do we want to have a `validateConsensus` flag as part of the `blockOpts` so these checks can be bypassed the way we support it at the `blockchain` level.  At the moment, these checks are enforced unless you stub them out in the test itself (with occasionally frustrating results trying to triage why a test is breaking)

holgerd77:
> The reason I ask the question is we have format level consensus checks in the blockheader constructor that require things like PoA blocks have to have minimum 97 bytes of `extraData` which in turn means we have to populate these fields when we're instantiating blocks. Do we want to have a `validateConsensus` flag as part of the `blockOpts` so these checks can be bypassed the way we support it at the `blockchain` level. At the moment, these checks are enforced unless you stub them out in the test itself (with occasionally frustrating results trying to triage why a test is breaking)

Yes, I guess we want but can you leave it out here and instead open a short issue, since this is not breaking and we can add later (if I am not mistaken).

I will merge in #1974 once CI passes. Can you prioritize on bringing this branch up to date then so that we can merge in here as well? I would want to have the big PRs out of the way.

ugh. 🙂

Thanks (I am getting a bit stressed 😋).

packages/block/src/header.ts
```diff
@@ -1000,12 +860,12 @@ export class BlockHeader {
...
 
   /**
-   * Internal helper function to create an annotated error message
+   * Helper function to create an annotated error message
    *
    * @param msg Base error message
    * @hidden
    */
-  protected _errorMsg(msg: string) {
+  private _errorMsg(msg: string) {
```
holgerd77, line 868: Small nit here, this should remain `protected` to allow for sub-classing.

packages/block/src/block.ts
```diff
@@ -524,7 +389,7 @@ export class Block {
    * @param msg Base error message
    * @hidden
    */
-  protected _errorMsg(msg: string) {
+  private _errorMsg(msg: string) {
```
holgerd77, line 392: Same here.

packages/block/src/header.ts
acolytec3: Got it.  Adding this to #1982

PR #1963, opened 2022-06-15
Disable client browser tests in Node nightly CI jobs

acolytec3:
- Switches Node nightly CI tests to just run client unit tests
- Sets fallbacks for nodejs `async_hooks` dependency in webpack config
- Adds note regarding browser builds being unsupported at present


acolytec3:
@holgerd77 Are we definitely dropping node 12 support? Just want to verify so I can fix this build too

holgerd77:
> @holgerd77 Are we definitely dropping node 12 support? Just want to verify so I can fix this build too

We have already if I am not totally mistaken, this is EOL for some time now.

Guess this is only an unwanted remainder in the Node versions test.

acolytec3:
> We have already if I am not totally mistaken, this is EOL for some time now.

Looks like it hit EOL at the end of April.  I looked at the script we use to get the testable node versions and it looks back 6 months so it's still pulling 12 in for the moment.  So, as a workaround, I just specified all node versions 14 or above in our jobs for now.  It's slightly hideous but will ensure we pick up new testable versions as they come and we can remove it.

acolytec3:
This is ready for review.  I verified the nightlies are passing again here - https://github.com/ethereumjs/ethereumjs-monorepo/actions/runs/2503263165  The main fix is ensuring we don't run tests below NodeJS 14 and also switching client test run to just execute the unit tests and skip browser.

.github/workflows/node-versions.yml
```diff
@@ -1,6 +1,6 @@
 # This special file aims to run node tests for each relevant node version.
 # A relevant node version can be: active, current, maintenance or EOL with some months of tolerance
-# For more details, please check ./scripts/node-versions.js
+# For more details, please check https://github.com/evertonfraga/testable-node-versions
```
holgerd77, line 3: This left me a bit confused, then I followed the path and realized that the `npx testable-node-versions` command within this file is indeed installing the package from the above-changed-location, being a script from Everton (former EthereumJS team member).

I would have a strong tendency to internalize this script (not part of this PR), I am not feeling well with these kind of super-implicit hidden external dependencies.

holgerd77 (approved):
LGTM

.github/workflows/node-versions.yml
acolytec3: I'm good with that.  It was internal originally and then I guess Ev moved it out to make it usable wherever else.  Can put that as a new issue to resolve.

acolytec3:
@holgerd77 I'm assuming these failing Client/VM tests are unrelated to this PR.  I didn't change anything there so maybe we can go ahead and merge?

holgerd77:
Yes, ah, we are already in this admin-merge thing. 🤪

Will do.

holgerd77 (approved):
LGTM, will admin-merge (current `vm-api` tests failing after VM/EVM refactor merge, only temporary).

Issue #1964, opened 2022-06-15
Monorepo: Remove esModuleInterop and allowSyntheticDefaultImports Options


holgerd77:
As stated in the associated meta issue #1946 the [esModuleInterop](https://www.typescriptlang.org/tsconfig#esModuleInterop) and [allowSyntheticDefaultImports](https://www.typescriptlang.org/tsconfig#allowSyntheticDefaultImports) TypeScript compiler options should be set to `false` in our base `tsconfig.json` file, see reasoning linked in the meta issue.

This task should optimally be done before the Beta 1 release round, latest before Beta 2.

The options should be deactivated in succession respectively commit by commit with a successful CI run in between.

acolytec3:
Doesn't this require ESM builds in order to complete?

paulmillr:
It seems like in `ESM` you could still import typescript-compiled files even without `esModuleIntrop`. I've tested this in node.js v18 REPL. So the option seems useless.

holgerd77:
> Doesn't this require ESM builds in order to complete?

Read up on what the options do (these only cover some special import cases being otherwise problematic), also on the reasoning mentioned in the original TODO from the meta issue.

ScottyPoi:
I worked on this yesterday and got most of the tests passing.

These two options work together to compensate for differences in how various modules export themselves.  Modules that do not `export default` their *namespace* cannot be imported in the same way as those that do.  The two compiler options allow us to ignore this and import all *namespaces* using the same syntax.   With them both set to *true*, all of our imports can look like:
 ```
import { BlockHeader } from '@ethereumjs/block'
import PeerId from 'peer-id'
import bootnodes from './bootnodes.json'
 ```
With these two options set to *false* (or **default** by excluding them), certain imports would require different syntax:
 ```
import { BlockHeader } from '@ethereumjs/block'
import * as PeerId from 'peer-id'
import * as bootnodes from './bootnodes.json'
 ```
From [](https://www.typescriptlang.org/tsconfig#esModuleInterop) 

```
By default (with esModuleInterop false or not set) TypeScript treats CommonJS/AMD/UMD modules similar to ES6 modules. In doing this, there are two parts in particular which turned out to be flawed assumptions:

    a namespace import like import * as moment from "moment" acts the same as const moment = require("moment")

    a default import like import moment from "moment" acts the same as const moment = require("moment").default

This mis-match causes these two issues:

[...]

Issue #1946, opened 2022-06-08
Monorepo: Build, Compiler and Export Updates


ryanio:
We decided that tackling esm support for this breaking change release would be too large of an effort while tooling and dependencies continue to finalize support.

There are a few concepts from #1936 that still have value to bring into this next major release:

- [x] Set [esModuleInterop](https://www.typescriptlang.org/tsconfig#esModuleInterop) and [allowSyntheticDefaultImports](https://www.typescriptlang.org/tsconfig#allowSyntheticDefaultImports) to `false` (see reasoning [here](https://www.semver-ts.org/#module-interop)) to not force the option for downstream consumers. (this is a breaking change), #1964 

- [x] Activate [declarationMap](https://www.typescriptlang.org/tsconfig#declarationMap) (`declarationMap=true`) as suggested by @MicahZoltu in this comment https://github.com/ethereumjs/ethereumjs-monorepo/pull/1936#discussion_r889802870, #1965

- [x] Remove `dist.browser` builds/folders - not sure if there is any difference in this build anymore since it has a ES2020 target for bigint support now. in maintenance branch we downleveled to ES5 for legacy browser support, which is no longer really needed. downstream consumers can continue to bundle for browsers themselves depending on their specific needs.

- [x] Export all exports from top level. Micah suggested this is better for browser support since browser deep imports still has its challenges. We already do this for many libs like Common, Tx, Block, etc., but it could be done for VM (and blockchain's new genesisStates).

- When doing the actual upgrade to TS esm, we can use Micah's wonderful [typescript-transformer-append-js-extension](https://github.com/Zoltu/typescript-transformer-append-js-extension) to eliminate the need to add .js file extensions to all import paths.

Note from @holgerd77: I will update this post with additional tasks, links, issue references and the like.

paulmillr:
>Set esModuleInterop and allowSyntheticDefaultImports to false (see reasoning [here](https://www.semver-ts.org/#module-interop)) to not force the option for downstream consumers. (this is a breaking change)

Since this is a breaking change, could we do it with this major release? Delaying it will either a) make minor release breaking b) require major release which is 6-12 months away.

I have updated dependencies in ethereum-cryptography to also remove this option.

holgerd77:
> > Set esModuleInterop and allowSyntheticDefaultImports to false (see reasoning [here](https://www.semver-ts.org/#module-interop)) to not force the option for downstream consumers. (this is a breaking change)
> 
[...]

PR #1973, opened 2022-06-17
Monorepo: Remove esModuleInterop and allowSyntheticDefaultImports Options

ScottyPoi:
TypeScript compiler options:
- `esModuleInterop` 
- `allowSyntheticDefaultImports` 
set to `false` as suggested in #1964

Part of meta issue #1946

### DIFF
--- a/config/tsconfig.json
+++ b/config/tsconfig.json
@@ -6,7 +6,7 @@
     "moduleResolution": "node",
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
-    "esModuleInterop": true,
+    "esModuleInterop": false,
     "allowSyntheticDefaultImports": true,
     "resolveJsonModule": true,
     "downlevelIteration": true,


ScottyPoi:
With `ESModuleInterop` and `AllowSyntheticDefaultImports` set to false, we have to change the way we import some modules.  Modules that do not `export default` cannot be imported like `import foo from 'bar'`.  Most of the time, these can be changed to `import * as foo from 'bar'`.

packages/client/lib/sync/fetcher/blockfetcherbase.ts
```diff
@@ -115,7 +115,7 @@ export abstract class BlockFetcherBase<JobResult, StorageItem> extends Fetcher<
     // Some jobs could have refetch tasks enqueued, so it is better to find
     // the first and last by examining each job.
     while (this.in.length > 0) {
-      const job = this.in.remove()
+      const job = this.in.dequeue()
```
paulmillr, line 118: Is this OK?

packages/client/lib/types.ts
```diff
@@ -1,7 +1,7 @@
 import { EventEmitter } from 'events'
 import type { Multiaddr } from 'multiaddr'
 import type { Block, BlockHeader } from '@ethereumjs/block'
-import type Connection from '../../../node_modules/libp2p-interfaces/dist/src/connection/connection'
+import type Connection = require('../../../node_modules/libp2p-interfaces/dist/src/connection/connection')
```
paulmillr, line 4: Why?

packages/devp2p/src/dpt/dpt.ts
```diff
@@ -1,4 +1,4 @@
-import ms from 'ms'
+import ms = require('ms')
```
paulmillr, line 1: does this work?

paulmillr (dismissed):

packages/devp2p/src/dpt/dpt.ts
ScottyPoi: It does.

Neither import `ms from 'ms'` or `import * as ms from 'ms'` would work.  I don't know what is different about 'ms' but from experimenting this is what worked.

packages/client/lib/types.ts
ScottyPoi: ` import type Connection from '../../../node_modules/libp2p-interfaces/dist/src/connection/connection' `

does not work because there is no "default" export from `connection`, and 
`import * as Connection from`
does not work because it `Connection` in that case is an object not a type.

packages/client/lib/sync/fetcher/blockfetcherbase.ts
ScottyPoi: Yes because i changed the Heap package we're using.

This solved one problem but created another, so this particular change is something I'm still tinkering with.

holgerd77:
> This is blocked and will be put earliest in Beta 2, see discussion on related issue.

Outdated and re-prioritized, will be directly put into Beta 1.

holgerd77:
Have submitted a cherry-picked version here #1975, will close here.

PR #1974, opened 2022-06-18
evm - Extract evm from vm as standalone package

g11tech:
WIP!


holgerd77:
Great start 👍, this looks already pretty good with seeing the two packages take shape! 🙏 

So Blockchain/State tests (so the official `ethereum/tests` suite) will likely stay in the outer package, right?

Is it challenging so far to separate on the other tests? 🤔 In doubt leave difficult to tackle tests in the outer package for now I guess (but would also be good if things which *can* move move over, otherwise there is some risk that this gets forgotten over time).

Ok, cool cool. 😎

g11tech:
> Client integration tests are hanging (> 20 min already), guess this won't finish any more.
> 
> Side note: the branch also still needs an update here before we can merge.

updating and checking, saw something similar in Andrew's PR, let me see

g11tech:
> > > Client integration tests are hanging (> 20 min already), guess this won't finish any more.
> > > Side note: the branch also still needs an update here before we can merge.
> > 
> > 
> > updating and checking, saw something similar in Andrew's PR, let me see
> 
> Is it ok if I merge in here once the CI is done (even if client tests are still hanging) and you eventually fix in a separate PR?
> 
> I would like to merge in these last big PRs as early as possible, so that we can get the conflicting work out of the way.

yes we can go that way! :+1:

holgerd77:
> yes we can go that way! 👍

Ok, thanks, I will then directly merge once CI passes. 👍

g11tech:
> > yes we can go that way! +1
> 
> Ok, thanks, I will then directly merge once CI passes. +1

test:integration which was getting stuck cleared on my local after rebasing on latest master so i guess CI should just pass :crossed_fingers:

holgerd77 (approved):
LGTM, thanks a lot Gajinder!

Will merge.

PR #1980, opened 2022-06-21
Blockchain reorganization

acolytec3:
Reorganizes `blockchain` to only do exports `index.ts`
`blockchain` class and its opts are moved to `blockchain.ts`


holgerd77 (changes requested):
Ah, the `blockchain.ts` file has lost its history, this would need to be re-applied I guess.

So the procedure would be: do a first commit renaming `index.ts` to `blockchain.ts`. And then do a second creating and adding a new `index.ts` just with the exports.

Otherwise we would loose the whole `index.ts` file and change history.

Does this make sense or am I missing something?

acolytec3:
> Ah, the `blockchain.ts` file has lost its history, this would need to be re-applied I guess.
> 
> So the procedure would be: do a first commit renaming `index.ts` to `blockchain.ts`. And then do a second creating and adding a new `index.ts` just with the exports.
> 
> Otherwise we would loose the whole `index.ts` file and change history.
> 
> Does this make sense or am I missing something?

I followed the path you suggested in #1981 and it looks like it did the same thing.  Does it look any different to you?

holgerd77:
Did you move natively with git, so with `git mv`? 🤔

acolytec3:
> Did you move natively with git, so with `git mv`? thinking

Ah, no, I didn't even realize that existed.  Learn something new every day.  Will close this PR and open a new one done under that approach.

PR #1981, opened 2022-06-21
Reorganize blockchain file layout

acolytec3:
Clones #1980 and hopefully keeps file history straight


holgerd77:
And I guess you really need to do these two separate commits (so not creating the new index.ts in the same commit).

PR #1982, opened 2022-06-22
blockchain: Reorganize exports

acolytec3:
Take three on reorganize the `blockchain` exports.

@holgerd77, I used `git mv` and it seems to have produced the same outcome as before where the file history sort of gets lost.  Not sure if there some additional git wizardry I'm missing out on here.


holgerd77:
Sorry if this might force you into a "take four" 😬, I also thought I had written this somewhere but can't find the reference any more so I might not have pressed the "submit" button or something:

You should actually do this (first) step in two commits. 1. Solely move `index.ts` to `blockchain.ts` without commiting anything else. 2. Add a new index.ts with the reduced re-export content.

Otherwise - minimally GitHub - and eventually git as well do not get this sorted out since the same file is overwritten in the same commit step.

(in the current status quo GitHub displays this as a new `blockchain.ts` and shows the diff in the `index.ts` file. This is not what we want here)

Issue #1960, opened 2022-06-15
Gray Glacier HF Support


holgerd77:
In order to get a bit more time for the Merge a difficulty bomb delay has been decided on [ACD 140](https://github.com/ethereum/pm/issues/538) (June 10 2022).

There is an EIP out for that [EIP-5133](https://eips.ethereum.org/EIPS/eip-5133) but I am not 100% sure if this will be taken in 1:1 so this would need some confirmation, there is also work on some associated tests done [here](https://github.com/ethereum/tests/issues/1042).

We should implement this to support this HF, implementation is pretty straight forward and can be followed I guess more or less 1:1 along how the `ArrowGlacier` HF has been implemented in #1527 by @jochem-brouwer.

This should be submitted to both `v5-maintenance` and `master`.

holgerd77:
Additional note: I am also not 100% sure if this affects only `mainnet` or also some other PoW test network.

jochem-brouwer:
FYI, Geth just released this Gray Glacier HF so we can assume that the numbers are ok

holgerd77:
Have assigned Scotty here.

holgerd77:
Closed by #1988 and #1984

PR #1983, opened 2022-06-22
Grayglacier HF Support

ScottyPoi:
Adds support for GrayGlacier HF re: #1960

### DIFF
--- a/packages/block/test/difficulty.spec.ts
+++ b/packages/block/test/difficulty.spec.ts
@@ -37,6 +37,9 @@ const hardforkTestData: TestData = {
   arrowGlacier:
     require('../../ethereum-tests/DifficultyTests/dfArrowGlacier/difficultyArrowGlacier.json')
       .difficultyArrowGlacier.ArrowGlacier,
+  grayGlacier:
+    require('../../ethereum-tests/DifficultyTests/dfGrayGlacier/difficultyGrayGlacier.json')
+      .difficultyGrayGlacier.GrayGlacier,
 }
 
 const chainTestData: TestData = {
--- a/packages/common/src/chains/mainnet.json
+++ b/packages/common/src/chains/mainnet.json
@@ -83,6 +83,11 @@
       "forkHash": "0x20c327fc"
     },
     {
+      "name": "grayGlacier",
+      "block": 15050000,
+      "forkHash": null
+    },
+    {
       "name": "mergeForkIdTransition",
       "block": null,
       "forkHash": null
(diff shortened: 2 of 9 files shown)


ScottyPoi:
The latest `develop` branch for `ethereum-tests` has the .json test files for Gray Glacier.  
I think the block test is failing because it is not looking at the current branch.

I don't know how to address this

packages/common/src/chains/mainnet.json
```diff
@@ -82,6 +82,11 @@
       "block": 13773000,
       "forkHash": "0x20c327fc"
     },
+    {
+      "name": "grayGlacier",
+      "block": 15050000,
+      "forkHash": null
```
holgerd77, line 88: Please calculate an explicit forkHash here, you can do this with the respective method in the `Common` library. If you would get some confirmation that the forkHash is correct - e.g. by looking at other client implementation GrayGlacier HF PRs - that would also be good.

packages/common/src/enums.ts
```diff
@@ -21,6 +21,7 @@ export enum Hardfork {
   Berlin = 'berlin',
   London = 'london',
   ArrowGlacier = 'arrowGlacier',
+  GrayGlacier = 'grayGlacier',
```
holgerd77, line 24: Unrelated, but since you are already working on the code: can you please move `Shanghai` to the last position here? This is some relic from times where people thought Shanghai would happen before the Merge. 😋

(general note: please do all change requests on both PRs, maybe do all in one commit and then cherry-pick over?)

packages/block/test/difficulty.spec.ts
```diff
@@ -37,6 +37,9 @@ const hardforkTestData: TestData = {
   arrowGlacier:
     require('../../ethereum-tests/DifficultyTests/dfArrowGlacier/difficultyArrowGlacier.json')
       .difficultyArrowGlacier.ArrowGlacier,
+  grayGlacier:
+    require('../../ethereum-tests/DifficultyTests/dfGrayGlacier/difficultyGrayGlacier.json')
+      .difficultyGrayGlacier.GrayGlacier,
```
holgerd77, line 42: So to your question with the tests: just copy the whole file over and move to the local Block test data folder and reference through the local path. You can maybe add a TODO here that this should be replaced with the `ethereum-tests` folder again at some point.

holgerd77 (changes requested):
Nice, thanks Scotty, looks good already, some few comments!

PR #1985, opened 2022-06-22
Remove explicit browser builds as monorepo builds already target es2020

g11tech:
Except `client` this PR removes explicit builds for the browser. The client seems to be having some replacements in the webpack config, so separate browser build target retained there for building replacements:
```typescript
replacement(resourcePath) {
            const mapping = {
              [resolve('./dist.browser/lib/logging.js')]: resolve(
                './dist.browser/browser/logging.js'
              ),
              [resolve('./dist.browser/lib/net/peer/libp2pnode.js')]: resolve(
                './dist.browser/browser/libp2pnode.js'
              ),
            }
```

### DIFF
--- a/config/cli/ts-build.sh
+++ b/config/cli/ts-build.sh
@@ -72,13 +72,9 @@ build_browser() {
 
 # Begin build process.
 
-if [ "$1" = "node" ];
-then
-    build_node
-elif [ "$1" = "browser" ];
+if [ "$1" = "browser" ];
 then
     build_browser
 else
     build_node
-    build_browser
 fi
--- a/.gitignore
+++ b/.gitignore
@@ -67,7 +67,6 @@ jspm_packages/
 
 # generate output
 dist
-dist.browser
 
 # Gatsby files
 .cache/
--- a/config/README.md
+++ b/config/README.md
@@ -99,8 +99,7 @@ Exposed CLI commands:
 
 ### Usage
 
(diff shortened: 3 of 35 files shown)


acolytec3 (commented):
Do we even need this browser key at all in package.json if it's pointing to the same place as main? Feels duplicative to me if it doesn't change anything. In looking at the npm docs, seems like it's not needed for us (except for client).
https://docs.npmjs.com/cli/v8/configuring-npm/package-json#main

g11tech:
> Do we even need this browser key at all in package.json if it's pointing to the same place as main? Feels duplicative to me if it doesn't change anything. In looking at the npm docs, seems like it's not needed for us (except for client). https://docs.npmjs.com/cli/v8/configuring-npm/package-json#main

i agree, can remove it if @holgerd77 concurs :+1: .

holgerd77:
> i agree, can remove it if @holgerd77 concurs 👍 .

Attention, I have just rebased here. 🙂 

But otherwise: yeah, agree, we can likely remove.

(so, you can also rebase again locally and then force-push, as you like of course)

holgerd77 (approved):
LGTM, thanks!

paulmillr:
Yay!

config/cli/ts-build.sh
```diff
@@ -72,13 +72,9 @@ build_browser() {
 
 # Begin build process.
 
-if [ "$1" = "node" ];
-then
-    build_node
-elif [ "$1" = "browser" ];
+if [ "$1" = "browser" ];
```
paulmillr, line 75: why was the browser past left?

PR #1991, opened 2022-06-23
VM Folder restructure

### LINKED ISSUE #1990: VM: Restructure Exports / Extract Types
For the VM there is also this last minute task of restructuring the exports and also extract (most of) the types. This can't be as pure as in the other packages since the VM is a bit differently structured - we should also resist to do a last minute refactor on this component structure - nevertheless we should do the 80% part with the main goal to avoid deep imports and have things - somewhat - aligned with the other libraries.

So, things to do:

- A dedicated `vm.ts` file (again: move `index.ts` with `git mv` command, then separate commit for new `index.ts`)

Re-exports:

- Bloom: yes
- EEI: yes
- VMState: I guess no
- BlockBuilder, also runBlock, runTx functions: no

Types moving to `types.ts`:
- main VMopts: yes
- Opts and Result types from BlockBuilder, runBlock, runTx (reasoning: the main exports shouldn't be accessed directly from these files as well - the files are only artificial helpers to extract functionality from VM for development reasons - the API takes place in VM and the functions are called from within the `vm` instance, e.g. by executing `vm.runTx()`. So it doesn't make sense that then the types are the only things which are accessed API-wise through these files

On top thinking: maybe we should even move these three files (runTx.ts, runBlock.ts, buildBlock.ts) to a folder `internal`, to make this a lot more clear that these shouldn't be referenced directly. I would actually be a fan.

Would be really great if someone could pick up this task during the day! 🙂

acolytec3:
Fixes #1990 

- Moves `VM` class to `vm.ts`
- Moves all exported types to `types.ts`
- Re-exports `Bloom` and `EEI`


packages/client/lib/execution/receipt.ts
```diff
@@ -1,6 +1,5 @@
-import { PreByzantiumTxReceipt, PostByzantiumTxReceipt, TxReceipt } from '@ethereumjs/vm/dist/types'
+import { PreByzantiumTxReceipt, PostByzantiumTxReceipt, TxReceipt, Bloom } from '@ethereumjs/vm'
```
holgerd77, line 1: 👍

packages/client/lib/execution/vmexecution.ts
```diff
@@ -15,8 +15,7 @@ import { Event } from '../types'
 import { Execution, ExecutionOptions } from './execution'
 import { ReceiptsManager } from './receipt'
 import type { Block } from '@ethereumjs/block'
-import type { RunBlockOpts } from '@ethereumjs/vm/dist/runBlock'
-import type { TxReceipt } from '@ethereumjs/vm/dist/types'
+import type { RunBlockOpts, TxReceipt } from '@ethereumjs/vm'
```
holgerd77, line 18: Oh, that wasn't apparent to me that these things get so much nicer out of the box and many places going down to one-line imports. Really cool! 😄

holgerd77 (approved):
Nice, looks really good, thanks Andrew!

PR #1992, opened 2022-06-23
Add declaration map

### LINKED ISSUE #1965: Monorepo: Activate declarationMap Compiler Option
There was a suggestion from @MicahZoltu to activate the [declarationMap](https://www.typescriptlang.org/tsconfig#declarationMap) (`declarationMap=true`) TypeScript compiler option which seems to make a lot of sense, see also the associated meta issue #1946 

This should be added to our base TypeScript configuration file (`config/tsconfig.json`) and also actively tested (maybe with screenshot in VSCode) if properly working (before case would also make sense to post).

This task can be done before the Beta 1 or Beta 2 releases.

g11tech:
Scenario: goto definition for `TypedTransaction`
image

### Before map
goes to `packages/tx/dist/.../types.d.ts`
image

### After map
Goes to `packages/tx/src/types.ts`
image

Closes #1965

### DIFF
--- a/config/tsconfig.json
+++ b/config/tsconfig.json
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "sourceMap": true,
     "declaration": true,
+    "declarationMap": true,
     "module": "commonjs",
     "moduleResolution": "node",
     "emitDecoratorMetadata": true,


holgerd77:
Rebased this via UI.

holgerd77 (approved):
Nice, thanks for the illustrative screen shots, looks good, happy for that feature actually! 😀

Issue #1993, opened 2022-06-23
Blockchain: Add consensus Option


holgerd77:
We have forgotten to allow the `consensus` class being passed in as an option to the Blockchain class, just realizing this while writing the Blockchain release notes where I already wanted to write this as a sentence because I thought: "sure, that's possible now". 🤪 😅

We should add, maybe a small test for now is enough where one of the existing consensus classes is passed in from the outside.

Would eventually be nice to do before Beta 1 but no total drama if done afterwards.

holgerd77:
Update: maybe let's *really* do before Beta 1, I will already write this in the release notes, this is actually a very small update.

ScottyPoi:
Was this meant only for `master` branch?

holgerd77:
Yes.

PR #1994, opened 2022-06-24
Blockchain--Add-consensus-Option (master)

ScottyPoi:
re: #1993

Add consensus as an option for Blockchain.create()

Implemented two ways:

`consensus` is an optional string input that functions much like common.consensusAlgorithm().   You can pass in `casper`, `clique`, or `ethash`, and the blockchain constructor will create a `CasperConsensus`, `CliqueConsensus`, or `EthashConsensus`.

`consensusAlgorithm` is an optional constructor parameter.  A user could pass in a constructor for a custom `consensusAlgorithm` as long as it implements the base `Consensus` interface.

Also added tests for both approaches.

### DIFF
--- a/packages/blockchain/src/types.ts
+++ b/packages/blockchain/src/types.ts
@@ -1,6 +1,8 @@
 import { Block } from '@ethereumjs/block'
 import Common from '@ethereumjs/common'
 import { AbstractLevel } from 'abstract-level'
+import { Consensus } from '.'
+import { Blockchain } from './blockchain'
 import { GenesisState } from './genesisStates'
 
 export type OnBlock = (block: Block, reorg: boolean) => Promise<void> | void
@@ -81,6 +83,14 @@ export interface BlockchainOptions {
    * Default: `true`.
    */
   validateConsensus?: boolean
+
+  /**
+   * Options to pass in string to define consensus
+   * Or to pass in custom consensusAlgorithm that correctly implements `Consensus`
+   */
+
+  consensus?: 'casper' | 'clique' | 'ethash'
+  consensusAlgorithm?: (blockchain: { blockchain: Blockchain }) => Consensus
 
   /**
    * This flag indicates if protocol-given consistency checks on
--- a/packages/blockchain/test/index.spec.ts
+++ b/packages/blockchain/test/index.spec.ts
@@ -1,7 +1,7 @@
 import Common, { Chain, Hardfork } from '@ethereumjs/common'
 import { Block, BlockHeader, BlockOptions } from '@ethereumjs/block'
 import * as tape from 'tape'
-import Blockchain from '../src'
+import Blockchain, { CasperConsensus, Consensus, EthashConsensus } from '../src'
 import { generateBlockchain, generateBlocks, isConsecutive, createTestDB } from './util'
 import * as testDataPreLondon from './testdata/testdata_pre-london.json'
 import * as blocksData from './testdata/blocks_mainnet.json'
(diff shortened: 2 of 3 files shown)


ScottyPoi:
```
> @ethereumjs/vm@5.9.1 build
> ../../config/cli/ts-build.sh
[Node build] Using tsconfig.prod.json
> tsc --build ./tsconfig.prod.json
[Node build] Working... DONE
Error: tsconfig.benchmarks.json(6,5): error TS5069: Option 'declarationMap' cannot be specified without specifying option 'declaration' or option 'composite'.
npm ERR! Lifecycle script `build:benchmarks` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: @ethereumjs/vm@5.9.1 
npm ERR!   at location: /home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/packages/vm 
Error: Process completed with exit code 1.
```

Does this vm-benchmarks build error have anything to do with this PR?

acolytec3:
> Does this vm-benchmarks build error have anything to do with this PR?

Shouldn't. That's an error that was carried over from #1992. Will add it as a separate issue to fix

packages/blockchain/src/types.ts
```diff
@@ -82,6 +84,14 @@ export interface BlockchainOptions {
    */
   validateConsensus?: boolean
 
+  /**
+   * Options to pass in string to define consensus
+   * Or to pass in custom consensusAlgorithm that correctly implements `Consensus`
+   */
+
+  consensus?: 'casper' | 'clique' | 'ethash'
```
acolytec3, line 92:
```suggestion
  consensus?: ConsensusAlgorithm
```

Let's use the `ConsensusAlgorithm` enum from `common` here

packages/blockchain/src/types.ts
```diff
@@ -82,6 +84,14 @@ export interface BlockchainOptions {
...
+  consensusAlgorithm?: (blockchain: { blockchain: Blockchain }) => Consensus
```
acolytec3, line 93: What does this do?  I see the comment but I'm not following what it actually does since it doesn't appear to attach the consensus class to the `blockchain` object in the constructor.

packages/blockchain/test/index.spec.ts
```diff
@@ -38,6 +38,64 @@ tape('blockchain test', (t) => {
     st.end()
   })
 
+  t.test('should initialize correctly with consensus option passed in', async (st) => {
+    const common = new Common({ chain: Chain.Goerli, hardfork: Hardfork.Chainstart })
+    let blockchain = await Blockchain.create({ common: common, consensus: 'casper' })
+    let iteratorHead = await blockchain.getIteratorHead()
+    st.ok(
```
acolytec3, line 45: Can we revise these tests to confirm that the instantiated `blockchain` object contains the expected consensus algorithm?  Also, I see you're using `goerli` as the chain.  It would be good to validate that `common.consensusAlgorithm` matches the `consensus` class instantiated with the blockchain.  Since Goerli is PoA, we wouldn't want to add an ethash consensus class to the blockchain.

acolytec3 (changes requested):
General question, shouldn't we be checking that whatever consensus type we specify in this constructor matches the `Common` instance we provide?  We could easily get logical inconsistencies if someone passes in a `common` instance that specifies the Mainnet chain configuration but then calls for `PoA` consensus on the blockchain.  The validation checks in blockchain will be confused at a minimum.

More generally, should we really be allowing a user to specify the consensus type in the blockchain class?  It seems to me we should keep this functionality in `Common` and continue to use `Common` to determine what consensus class to instantiate in the blockchain object.  @holgerd77 thoughts on this?

holgerd77:
@acolytec3 will continue here since this is - somewhat - time critical.

holgerd77:
(respectively likely do a new PR)

Issue #1961, opened 2022-06-15
Util: Signature ecsign / ecrecover functionality


holgerd77:
Some last minute thinking about the signature functionality for the breaking releases we now have in [signature.ts](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/src/signature.ts) in `Util`, especially regarding the role of the `v` and `recovery` value.

This was a bit triggered by stumbling upon #1597 which mentions the `ecsign()` function, which we haven't touched yet (regarding the aspect from above) during the up-till-now refactoring work.

So the thing is `v` is still returning the calculated values (so e.g. "+27" for legacy, "+ chainId + additional stuff" for typed tx), and people might or might not need rather the `recovery` value.

I wondered in a first round if it would make sense to just add plain `recovery` to the `ECDSASignature` interface and just return the plain `recovery` value in addition to what we return as `v` at the moment. Then people can choose what to use depending on their needs?

I have to admit though that I am still not 100% saddle-proof on the semantics of names, so on the very correct usage of `v` (that one in particular), `recovery` and `yParity`.

My current assumptions are:

- `v`: Legacy -> the "recovery + 27" stuff, Typed: the "recovery + chainId + stuff" stuff (is/can this still called v?)
- `recovery` == `yParity` (is this also 100% correct)

Will circle in @jonathansmirnoff (from the issue above) here, everyone else welcome to join as well.

My current assumption is still that the suggestion from above (adding `recovery` to the interface) would help and would be very easy to be added (still breaking, so worth to do now).

paulmillr:
`recovery` in secp256k1 is always 0 or 1, so I guess it can be `== yParity`. As a side note, `@noble/secp256k1` signatures don't have `recovery` bit in them — instead, there's `Point.fromSignature(msgHash: Hex, signature: Sig, recovery: number)` method that creates ECDSA points (public keys) from signatures and their message hash.

ScottyPoi:
We have a function `ecrecover`.  I can refactor this function to calculate recovery if a chainId is not passed in, however the current tests are set up to expect this to fail when certain `v` values are used.  If I let this method calculate `recovery` from `v` when no `chainId` is passed, than it no longer throws in those two test cases.

'should fail on an invalid signature (v = 29)'
and
'should fail on an invalid signature (v = 21)'

holgerd77:
[...]

PR #1995, opened 2022-06-24
Util--Signature-ecsign-ecrecover-functionality

ScottyPoi:
Re: #1961 

Util:

- Adds `recovery` to `ECDSASignature` interface.
- Creates a new method `calculateSigRecoveryFromV(v: bigint)`
  - This includes a calculation for recovery based on the `recovery = chainId * 2 + 35` formula.
- Refactors `fromRpcSig()` to return `ECDSASignature` with `recovery`.
- Updates relevant  `signature.spec.ts` tests for new interface, methods.

TX:
For legacy transactions, `Transaction._validateTxV()` was updated to filter for hardforks that include EIP-150

Client:
- Small changes were made regarding `miner` and `pendingBlock` in response to failing tests.

### DIFF
--- a/packages/client/lib/miner/miner.ts
+++ b/packages/client/lib/miner/miner.ts
@@ -260,6 +260,7 @@ export class Miner {
         common: this.config.chainCommon,
         calcDifficultyFromHeader,
         putBlockIntoBlockchain: false,
+        hardforkByBlockNumber: true,
       },
     })
 
--- a/packages/client/lib/miner/miner.ts
+++ b/packages/client/lib/miner/miner.ts
@@ -257,7 +257,7 @@ export class Miner {
       },
       blockOpts: {
         cliqueSigner,
-        hardforkByBlockNumber: true,
+        common: this.config.chainCommon,
         calcDifficultyFromHeader,
         putBlockIntoBlockchain: false,
       },
--- a/packages/tx/src/legacyTransaction.ts
+++ b/packages/tx/src/legacyTransaction.ts
@@ -296,13 +296,7 @@ export default class Transaction extends BaseTransaction<Transaction> {
     this._validateHighS()
 
     try {
-      return ecrecover(
-        msgHash,
-        v!,
-        bigIntToUnpaddedBuffer(r!),
-        bigIntToUnpaddedBuffer(s!),
-        this.supports(Capability.EIP155ReplayProtection) ? this.common.chainId() : undefined
-      )
+      return ecrecover(msgHash, v!, bigIntToUnpaddedBuffer(r!), bigIntToUnpaddedBuffer(s!))
     } catch (e: any) {
       const msg = this._errorMsg('Invalid Signature')
       throw new Error(msg)
--- a/packages/tx/src/legacyTransaction.ts
+++ b/packages/tx/src/legacyTransaction.ts
@@ -361,6 +361,15 @@ export default class Transaction extends BaseTransaction<Transaction> {
           `Legacy txs need either v = 27/28 or v >= 37 (EIP-155 replay protection), got v = ${v}`
         )
       }
+      if (
+        common?.hardfork() === 'chainstart' ||
+        common?.hardfork() === 'homestead' ||
+        common?.hardfork() === 'tangerineWhistle'
+      ) {
+        if (v > 28) {
+          throw new Error('Legacy txs need either v = 27/28 or v >= 37 (EIP-155 replay protection)')
+        }
+      }
     }
 
     // No unsigned tx and EIP-155 activated and chain ID included
--- a/packages/util/src/signature.ts
+++ b/packages/util/src/signature.ts
@@ -18,17 +18,16 @@ export interface ECDSASignature {
  * accordingly, otherwise return a "static" `v` just derived from the `recovery` bit
  */
 export function ecsign(msgHash: Buffer, privateKey: Buffer, chainId?: bigint): ECDSASignature {
-  const [signature, recovery] = signSync(msgHash, privateKey, { recovered: true, der: false })
+  const [signature, _recovery] = signSync(msgHash, privateKey, { recovered: true, der: false })
 
   const r = Buffer.from(signature.slice(0, 32))
   const s = Buffer.from(signature.slice(32, 64))
-
   const v =
     chainId === undefined
-      ? BigInt(recovery + 27)
-      : BigInt(recovery + 35) + BigInt(chainId) * BigInt(2)
-
-  return { r, s, v }
+      ? BigInt(_recovery + 27)
+      : BigInt(_recovery + 35) + BigInt(chainId) * BigInt(2)
+  const recovery = BigInt(_recovery)
+  return { r, s, v, recovery }
 }
 
 function calculateSigRecovery(v: bigint, chainId?: bigint): bigint {
--- a/packages/util/src/signature.ts
+++ b/packages/util/src/signature.ts
@@ -39,6 +39,21 @@ function calculateSigRecovery(v: bigint, chainId?: bigint): bigint {
   return v - (chainId * BigInt(2) + BigInt(35))
 }
 
+export function calculateSigRecoveryFromV(v: bigint) {
+  if (v === BigInt(0) || v === BigInt(1)) {
+    return v
+  } else {
+    const _v = Number(v)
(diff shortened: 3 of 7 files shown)


packages/util/src/signature.ts
```diff
@@ -39,6 +39,21 @@ function calculateSigRecovery(v: bigint, chainId?: bigint): bigint {
...
+  if (v === BigInt(0) || v === BigInt(1)) {
+    return v
+  } else {
+    const _v = Number(v)
+    if (Number.isInteger((0 - 35 - _v) / -2)) {
+      return BigInt(0)
+    } else if (Number.isInteger((1 - 35 - _v) / -2)) {
+      return BigInt(1)
+    } else {
+      throw new Error('Invalid v value')
```
holgerd77, line 52: Can we here remain in the BigInt context? `v` *can* theoretically (actually: this also already happened practically) get in the > MAX_SAFE_INTEGER range, when there is a very large chain ID (we once had this for a testnet), so arithmetics here should take place in BigInt.

Just tested, there is the modulo operator `%` also working in `BigInt`, so: `BigInt(5) % BigInt(2)` gave me `1n`, so it should be possible with something like this to get this working in plain `BigInt` as well.

packages/util/src/signature.ts
```diff
@@ -39,6 +39,21 @@ function calculateSigRecovery(v: bigint, chainId?: bigint): bigint {
   return v - (chainId * BigInt(2) + BigInt(35))
 }
 
+export function calculateSigRecoveryFromV(v: bigint) {
```
holgerd77, line 42: Totally a fan of having this exported in this new form, seems pretty useful to me, but can we then (also generally makes sense) some explicit test cases for this method itself, atm all tests are implicit (another method using this function is tested)?

So some edge cases like 1 (I guess we at some point decided/agreed that 0 is not a practical/used chain ID which we can ignore), a mid range even and odd value, and (related to the other comment) a very high v exceeding MAX_SAFE_INTEGER.

holgerd77 (changes requested):
Hi Scotty,
this looks totally great, thanks for tackling and love the cleanliness in this sensitve parts of our code base.

Two comments two address, also there are some `vm-api` CI tests still failing directly related to the new `recovery` property in the signature. This would need fixes as well.

Side note: branch update / rebase also necessary.

packages/util/src/signature.ts
```diff
@@ -8,6 +8,7 @@ export interface ECDSASignature {
   v: bigint
   r: Buffer
   s: Buffer
+  recovery: bigint
 }
```
holgerd77, line 12: Another short review round 🙂: can you please add some code docs to this interface?

packages/util/src/signature.ts
holgerd77, line 42: Also here, it would be great if you can add a TSDocs description.

packages/util/src/signature.ts
```diff
@@ -39,6 +39,21 @@ function calculateSigRecovery(v: bigint, chainId?: bigint): bigint {
   return v - (chainId * BigInt(2) + BigInt(35))
 }
```
holgerd77, line 40: I guess we can (and also: should) delete this method since this is redundant now.

packages/util/src/signature.ts
```diff
@@ -128,6 +143,7 @@ export const fromRpcSig = function (sig: string): ECDSASignature {
     throw new Error('Invalid signature length')
   }
 
+  const recovery: bigint = calculateSigRecoveryFromV(v)
```
holgerd77, line 146: Small nit: you do not need to explicity type this, this is inferred by the method call.

holgerd77 (changes requested):
Short second round. 🙂

packages/util/src/signature.ts
ScottyPoi: By deleting this method, I had to shift `ecrecover` to use the new `calculateSigRecoveryFromV()` instead of the deleted `calculateSigRecovery()`

`calculateSigRecovery` had a clause that if the `chainId` parameter was undefined, the method would throw.  Without chainId as a parameter at all, `calculateSigRecoveryFromV()` will always return.

Several tests in util and tx relied on this method throwing an error.

holgerd77:
This is getting too short-termed, so we'll rather add in Beta 2, no problem though. 🙂

holgerd77:
Giving this a "Blocked" label right now, can be removed again after the Beta 1 releases.

packages/client/lib/miner/miner.ts
```diff
@@ -257,7 +257,7 @@ export class Miner {
       },
       blockOpts: {
         cliqueSigner,
-        hardforkByBlockNumber: true,
```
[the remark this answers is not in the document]
ScottyPoi: Do you know why it was on?

This caused a test to fail that involved building a `Fake Chain`, and arbitrarily assigning hardforks to block numbers.
```
    const customChainParams = {
      hardforks: [
        { name: 'chainstart', block: 0 },
        { name: 'berlin', block: 2 },
        { name: 'london', block: 3 },
      ],
    }
```


https://github.com/ethereumjs/ethereumjs-monorepo/blob/26e1890d800851bd9907a1978ff83e92594840a1/packages/client/test/miner/miner.spec.ts#L373

with `hardforkByBlockNumber` on, it would always revert back to `chainstart`, even with the   `common` option set to  `this.config.chainCommon`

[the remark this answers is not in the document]
ScottyPoi: I think Ryan designed the test.

If there's a strong reason for having `hardforkByBlockNumber` on, I could instead look into changing `VM.buildBlock()` to ignore it if the `common` option is set?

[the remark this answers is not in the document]
g11tech: its on i think to make sure the correct block is produced according to hardfork at that current block number (could be a side chain/non-canonical chain and loading hardfork at that block number makes sense)

but how should your PR which doesn't look relevant to this, affect this logic, thats the first thing to understand why this spec test is failing.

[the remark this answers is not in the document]
g11tech: most likely the mockchain might not be build properly because of some validation failing, that the miner is building block from `0` where it seems like it should build post `3`. I will try this PR out tomorrow locally to confirm this hypo.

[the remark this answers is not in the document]
ScottyPoi: Thanks I spent more time with it today and didn't get any further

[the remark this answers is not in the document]
g11tech: i think I have resolved the spec test, so turning this back on :+1:

packages/util/src/signature.ts
```diff
@@ -43,10 +43,9 @@ export function calculateSigRecoveryFromV(v: bigint) {
   if (v === BigInt(0) || v === BigInt(1)) {
     return v
   } else {
-    const _v = Number(v)
-    if (Number.isInteger((0 - 35 - _v) / -2)) {
+    if ((0n - 35n - v) % 2n === 0n) {
       return BigInt(0)
-    } else if (Number.isInteger((1 - 35 - _v) / -2)) {
+    } else if ((1n - 35n - v) % 2n === 0n) {
```
holgerd77: Please do not use `n` notation for BigInt for now, we decided at some point to keep with the `BigInt()` constructor for tool compatibility reasons.

packages/tx/src/legacyTransaction.ts
```diff
@@ -296,13 +296,7 @@ export default class Transaction extends BaseTransaction<Transaction> {
     this._validateHighS()
 
     try {
-      return ecrecover(
-        msgHash,
-        v!,
-        bigIntToUnpaddedBuffer(r!),
-        bigIntToUnpaddedBuffer(s!),
-        this.supports(Capability.EIP155ReplayProtection) ? this.common.chainId() : undefined
```
holgerd77, line 304: Please do not do any changes to the transaction library in this PR, this is out of the scope of the task and by far too dangerous.

packages/util/src/signature.ts
holgerd77: If this causes any problems than in  doubt *keep* the method. This was just a quick shot from me assuming this internal method is not needed any more.

holgerd77 (changes requested):

holgerd77:
Replaced by #2012, will close.

PR #1997, opened 2022-06-24
chore: remove untrue jsdoc comment on `clearWarmedAccounts`

MicaiahReid:
I was using the StateManager's `clearWarmedAccounts` function and noticed that the jsdoc comment says that the function takes a `boolean`, but found this to be untrue. This PR removes that jsdoc comment.


acolytec3 (approved):
Thanks for the PR!  LGTM

PR #2000, opened 2022-06-27
EVM: Add minimal EEI implementation

jochem-brouwer:
This adds a minimal EEI implementation. This features almost all logic of the EEI, including commit/revert/checkpointing logic. 

~~All features which the EEI does not support (these are NOT necesary to use the EVM, these are all things the EVM does not know about (state roots), or does not care about (cleanup touched accounts)) throw when called.~~

End goal: run state tests within EVM


jochem-brouwer:
Follow up #2649

PR #2001, opened 2022-06-28
EVM Types Cleanup

holgerd77:
Types clean-up for the EVM package with the main goal to reduce type exposure since every type we have exported gets part of the API and makes it harder to change things later. One main thing is that I removed the exports for all the redundant state types and only exported the resulting `EVMStateAccess` type (also renamed some types).

For the types file I deviated a bit from the order-by-type-dependency sorting with the goal of getting this file more readable, since otherwise all the important types gets totally buried at the end of the file (even the main `EVMInterface` type!).

Also did some other clean-up along the way (removed the :API appendix for the EVM tests + removed `api` subfolder e.g.).

Open for review.


g11tech (approved):

acolytec3 (commented):
Looks good to me in general.  Added one specific comment on `EVMResult` `ExecResult` that I think we should consider.

holgerd77:
> Looks good to me in general. Added one specific comment on `EVMResult` `ExecResult` that I think we should consider.

Thanks, I am not seeing the comment TBH, did you submit?

Have now merged, but we can still address post-merge as separate PR or PR addition.

PR #2002, opened 2022-06-28
Blockchain: optional consensus

### LINKED ISSUE #1993: Blockchain: Add consensus Option

acolytec3:
Fixes #1993 

Adds a new optional consensus property to the `blockchain` constructor and a new `algorithm` property to the `Consensus` interface that provides either a `Common.ConsensusAlgorithm` or `string` that identifies the consensus type.


packages/blockchain/test/customConsensus.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+    _block: Block,
+    _commonAncestor?: BlockHeader | undefined,
+    _ancientHeaders?: BlockHeader[] | undefined
+  ): Promise<void> {
+    return new Promise<void>((resolve) => resolve())
+  }
+
+  consensusAlgorithm() {
+    return 'fakeConsensus'
+  }
```
holgerd77: This should then also be part of the `Consensus` interface.

holgerd77: So, and I would then also have a tendency to check this in the Blockchain constructor against the consensus algorithm set in Common and throw if this doesn't match (otherwise this doesn't make things any more consistent and we should in doubt leave completely for now I guess?).

Tests can be adopted in a way that you are just setting this to the `ethash` value and in the interface we can mention in the TSDocs for the `consensusAlgorithm()` method that this will need to match the `Common` one.

Again, I am also still open to leave this completely to simplify things for now. 🙂

holgerd77 (commented):
Nice, great, this looks already pretty good, nice test setup! 👍

packages/blockchain/test/customConsensus.spec.ts
acolytec3: > This should then also be part of the `Consensus` interface.

Should be addressed in [a41816b](https://github.com/ethereumjs/ethereumjs-monorepo/pull/2002/commits/a41816bd37bc24f7f1e1eaabf08d88b2ec539200)

>So, and I would then also have a tendency to check this in the Blockchain constructor against the consensus algorithm set in Common and throw if this doesn't match (otherwise this doesn't make things any more consistent and we should in doubt leave completely for now I guess?).

If we do this, don't we also have to add in the custom consensus algorithm to `common`?

acolytec3:
@holgerd77 @jochem-brouwer This should now be ready for review.

holgerd77:
Will adding a "not merge" label here and put this on hold for the Beta 1 releases, since this is getting too short-termed. You can of course continue to work on this, just please do not merge until Beta 1 is released.

packages/blockchain/src/consensus/clique.ts
```diff
@@ -46,7 +46,8 @@ type CliqueLatestBlockSigners = CliqueBlockSigner[]
  * This class encapsulates Clique-related consensus functionality when used with the Blockchain class.
  */
 export class CliqueConsensus implements Consensus {
-  blockchain: Blockchain
+  blockchain: Blockchain | undefined
```
jochem-brouwer, line 49: Why can this be `undefined`? Would it not be cleaner to mark `setup` as static, and then pass `blockchain` in as an argument in the (`private`) `constructor`? It seems that `blockchain` is a mandatory field?

packages/blockchain/src/consensus/ethash.ts
```diff
@@ -7,15 +8,18 @@ import { Consensus, ConsensusOptions } from './interface'
  * This class encapsulates Ethash-related consensus functionality when used with the Blockchain class.
  */
 export class EthashConsensus implements Consensus {
-  blockchain: Blockchain
-  _ethash: Ethash
+  blockchain: Blockchain | undefined
```
jochem-brouwer, line 11: Same here regarding the `undefined`

jochem-brouwer (commented):
Some questions regarding the structure, but looks great in general, and great tests :+1: :smile:

packages/blockchain/src/consensus/clique.ts
acolytec3: The goal was to not require a `blockchain` when instantiating the consensus object. Otherwise, we get a circular dependency where we want to allow users to define a custom consensus algorithm and pass that into the blockchain constructor. This was just a hack to get us there while also keeping the blockchain as an internal property of the consensus object. The alternative was to remove Blockchain from the internal properties and then require it be passed in as a parameter in every consensus method that needs the blockchain context.  I think either works but the alternative here was clunkier in practice when I implemented it

packages/blockchain/src/consensus/ethash.ts
acolytec3: See my earlier comment as the same applies. In practice, this Blockchain property should never be undefined because the `setup` method is called right after the blockchain is instantiated.

packages/blockchain/src/consensus/clique.ts
jochem-brouwer: Ah right I get your point. I need to give this some thought :thinking:

acolytec3:
Update here.

I've revised the custom consensus tests to have some more substantial checks to validate that custom consensus rules are actually applied. 

Also, I've made some tweaks to the `consensus` interface and `blockchain._putBlockOrHeader` method to remove methods that are specific to PoA and account for PoA specifics.

packages/blockchain/src/blockchain.ts
```diff
@@ -1173,17 +1177,17 @@ export class Blockchain implements BlockchainInterface {
     switch (this._common.consensusAlgorithm()) {
```
g11tech, line 1177: @acolytec3 do you want to skip `checkAndTransitionHardForkByNumber` if the opts.consensus if present in constructor? (might need a flag) or one can also provide this fn as a constructor arg.

acolytec3: Actually, nm, it already exists.  [hardforkByHeadBlockNumber](https://github.com/ethereumjs/ethereumjs-monorepo/blob/f4cb2524210d1b7af55d4cd2c0004aee1966b01f/packages/blockchain/src/types.ts#L62)

This is only set to true in the `client` as far as I can tell. If we want the blockchain to transition hardforks for custom consensus types, we should probably just remove the `default` behaviour in the switch statement to throw if an unsupported consensus algorithm is provided.  Thoughts?

acolytec3:
> The thing is: a custom consensus implementation might need `Blockchain` as well. So now with this test consensus things work, but with another custom consensus needing `Blockchain` we have the same problems again.

`CliqueConsensus` already requires a `blockchain` to actually run the chain as it stands and it seems to work just fine with my current approach. Now that I've coded it up, I honestly feel like this is probably the best approach (where we don't require `blockchain` in the constructor` but then do want it in `setup`)  without a radical redesign of the whole consensus mechanism.  But, I'm happy to overruled if you don't think it's ready.

This feels like a good first step to my way of thinking and should allow for some interesting experimentation.

holgerd77:
> CliqueConsensus already requires a blockchain to actually run the chain as it stands and it seems to work just fine with my current approach. Now that I've coded it up, I honestly feel like this is probably the best approach (where we don't require blockchain in the constructorbut then do want it insetup`) without a radical redesign of the whole consensus mechanism. But, I'm happy to overruled if you don't think it's ready.
> 
> This feels like a good first step to my way of thinking and should allow for some interesting experimentation.

I am not 100% seeing through, would your approach still work if we make `Blockchain` just mandatory in `setup` (even for the implementations which do not need it)? I really would like to have the interface as equal as possible for all consensus implementations and not making specific assumptions for any specific consensus algorithm (even if it comes at the "cost" as having an unnecessary parameter set or a method unnecessarily called).

acolytec3:
> I am not 100% seeing through, would your approach still work if we make `Blockchain` just mandatory in `setup` (even for the implementations which do not need it)? I really would like to have the interface as equal as possible for all consensus implementations and not making specific assumptions for any specific consensus algorithm (even if it comes at the "cost" as having an unnecessary parameter set or a method unnecessarily called).

I think so but will check and see :-)

packages/blockchain/src/blockchain.ts
g11tech: yes we can ignore default behavior if custom consensus is provided :+1:

acolytec3:
Okay, I've made all the requested changes and set guards against transitioning to the merge if custom consensus is implemented so should be ready for another round (maybe final) of review. @holgerd77  @g11tech

g11tech (approved):
LGTM! :rocket:

holgerd77 (approved):
Yeah, looks really great, thanks a lot for working this through! 👍🙂

Issue #1976, opened 2022-06-21
EVM V7 tracking issue


jochem-brouwer:
To be removed:

- Touched accounts in EEI
- Warm addresses in EEI

holgerd77:
Closed in favor of #2428 (unifies TODOs + better visibility in issue list)

PR #2003, opened 2022-06-29
EVM: Cleanup EEI

jochem-brouwer:
This PR cleanups the EEI interface.

In particular, it provides a minimal interface for the EEI. The EEI encapsulates everything with which the EVM tries to interact externally with, such as getting/setting contract code, storage, updating accounts, or retrieving blockhashes (BLOCKHASH opcode). It can be noted that the EEI is kind-of an extension of StateManager - in here, we have additional logic for touched accounts and warm storage.

(In an ideal world, i.e. V7 of VM see #1976, the touched accounts and warm account logic all moves into EVM, then we essentially only need the StateManager and we need access to the blockchain to retrieve the blockhashes).

This cleanup cleans up some of the EEI (interface) methods which are not necessary or only used internally. Some of them, i.e. `accountExists` are also removed since this logic can easily be done inside the EVM (and in fact, in some cases, was either done there already or it called into this "helper method" of EEI)

WIP. Needs another round of cleanup.

### DIFF
--- a/packages/client/lib/execution/vmexecution.ts
+++ b/packages/client/lib/execution/vmexecution.ts
@@ -74,7 +74,7 @@ export class VMExecution extends Execution {
     this.hardfork = this.config.execCommon.hardfork()
     this.config.logger.info(`Initializing VM execution hardfork=${this.hardfork}`)
     if (number === BigInt(0)) {
-      await this.vm.eei.state.generateCanonicalGenesis(this.vm.blockchain.genesisState())
+      await this.vm.eei.generateCanonicalGenesis(this.vm.blockchain.genesisState())
     }
   }
 
--- a/packages/client/lib/miner/miner.ts
+++ b/packages/client/lib/miner/miner.ts
@@ -211,7 +211,7 @@ export class Miner {
 
     // Set the state root to ensure the resulting state
     // is based on the parent block's state
-    await vmCopy.eei.state.setStateRoot(parentBlock.header.stateRoot)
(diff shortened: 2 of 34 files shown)


acolytec3 (approved):
This looks good to me.  Is the basic thing that we're just making EEI extend VmState?  I didn't really see anything else that jumped out at me.

holgerd77 (approved):
Great work, thanks. 👍 🙂 I have to say I am a bit nervous on this last minute decision to merge EEI and the state VM implementation, but, ok, let's do it.

Let v7 correct everything where we might have taken a bit of a wrong side path or confirm where things stood the test of time. 🤓

Will merge.

holgerd77:
(have temporarily deactivated the client tests for this)

g11tech (commented):
LGTM!

jochem-brouwer:
@acolytec3 This cleanups the EEI interface. If I would not merge the VmState + EEI then the EEI itself would be very shallow, so to me it made sense to merge both of them.

@holgerd77 You are absolutely right, I think the V7 interface would make more sense. (The current state is a huge improvement but it is not yet at a phase where we would be satisfied - there are some weird leftovers which do not make much sense especially upon reviewing these things)

holgerd77:
@jochem-brouwer yes, maybe we also want to communicate this towards the community this way ("huge improvement but still inconsistencies and only half-way there") instead of just "not mention" or sell as "now everything is great and modular".

PR #2004, opened 2022-06-30
Post-Beta 1 Releases Clean-Up

holgerd77:
Some clean-up and README additions after the Beta 1 releases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,7 @@
 [![Code Coverage][coverage-badge]][coverage-link]
 [![Discord][discord-badge]][discord-link]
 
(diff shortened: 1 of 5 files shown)


g11tech (approved):

holgerd77:
Ah, something strage happens here in the UI, the "Squash and merge" button persisted for me to be pressable, and I interpreted this as "Merging has not yet happened", so I pressed twice. Button is now still in this state, so apparently there is something broken with GitHub UI. We now have two commits with these changes in `master`.

After giving this a short thought I think it is not that grave in this case respetively I wouldn't want to do a "Remove branch protection" -> "Delete commit" operation with all the risks involved for that.

So I will close here (changes applied) and let this as the "PR applied and everything ok" state.

PR #2005, opened 2022-06-30
common: Update sepolia and ropsten configs for merge

g11tech:
Add `td` values for `ropsten` and `sepolia`  as well as `mergeForkIdTransition` entries which will be provided later which is when the `forkHash` will change.


holgerd77 (approved):
LGTM

holgerd77:
Rebased this via UI.

PR #2006, opened 2022-07-01
client: Fix rpc import broken after tuning off esModuleInterop

g11tech:
Fix the cors lib rpc import broken after turning off esModuleInterop

image


holgerd77 (approved):
LGTM

PR #2007, opened 2022-07-02
Export all exports (VM)

ScottyPoi:
In reference to #1946 
```
 Export all exports from top level. Micah suggested this is better for browser support since browser deep imports still has its challenges. We already do this for many libs like Common, Tx, Block, etc., but it could be done for VM (and blockchain's new genesisStates).
```


holgerd77:
Hi Scotty,
this is getting too broad to re-export everything from `runBlock()`, `buildBlock()`, whatever, this issue of VM exports has really been somewhat solved by #1991. Let's please keep it this way now so close for the final releases unless you stumble upon something very very targeted, where only a single additional item needs to be exported or similar. In this case with this one-place client usage of a VM dist folder reference only by us internally this is not so dramatic and we don't need to change/update our import structure on this for now.

Will close here.

PR #2010, opened 2022-07-04
evm: Add check for zero nonce in create

### LINKED ISSUE #2009: Error in runCall due to negative nonce
I'm using the beta release. I may be using this wrong.

Here's a silly example that I think should work but it doesn't:

```javascript
const vm = await VM.create();
vm.evm.runCall({ to: undefined });
```

It raises the error:

```
Error: Cannot convert string to buffer. toBuffer only supports 0x-prefixed hex strings and this string was given: 0x-1
    at toBuffer (/node_modules/@ethereumjs/util/dist/bytes.js:147:19)
    at bigIntToBuffer (/node_modules/@ethereumjs/util/dist/bytes.js:197:33)
    at EVM._generateAddress (/node_modules/@ethereumjs/evm/dist/evm.js:634:94)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async EVM.runCall (/node_modules/@ethereumjs/evm/dist/evm.js:503:40)
```

It's happening here in `evm.ts`:

```
const acc = await this.eei.getAccount(message.caller)
const newNonce = acc.nonce - BigInt(1)
addr = generateAddress(message.caller.buf, bigIntToBuffer(newNonce))
```

`acc.nonce` is `0n` which seems correct given that the caller (the zero address) is an unused account. I'm not sure why it tries to do `acc.nonce - 1`, but in this case it results in a negative integer that can't be converted to a buffer.

Given that this works through `runTx`, I'm guessing that the nonce is meant to be incremented before execution. Is this the issue?

acolytec3:
Fixes #2009 

In `generateAddress`, we are not currently checking if the `FROM` account's nonce is 0 (i.e. an unused account) which leads to the error documented in #2009.  In practice, this should only ever occur if someone is making a direct call to `evm.runCall` independent of a running instance of the `VM` since the `FROM` account's nonce is incremented in `runTx` and you can't initiate a contract deployment from a non-existent account in an actual blockchain context.

### DIFF
--- a/packages/evm/src/evm.ts
+++ b/packages/evm/src/evm.ts
@@ -691,7 +691,15 @@ export class EVM extends AsyncEventEmitter<EVMEvents> implements EVMInterface {
         origin: opts.origin ?? opts.caller ?? Address.zero(),
       }
 
-      const caller = opts.caller ?? Address.zero()
+      let caller = opts.caller
+      if (!caller) {
+        // Set sensible default with correct nonce if no caller provided
+        caller = Address.zero()
+        const callerAccount = await this.eei.getAccount(caller)
+        callerAccount.nonce += BigInt(1)
+        await this.eei.putAccount(caller, callerAccount)
+      }
+
       const value = opts.value ?? BigInt(0)
       if (opts.skipBalance === true) {
         // if skipBalance, add `value` to caller balance to ensure sufficient funds
--- a/packages/evm/tests/runCall.spec.ts
+++ b/packages/evm/tests/runCall.spec.ts
@@ -13,6 +13,19 @@ function create2address(sourceAddress: Address, codeHash: Buffer, salt: Buffer):
   return new Address(Buffer.from(keccak256(hashBuffer)).slice(12))
 }
 
+tape('Create where FROM account nonce is 0', async (t) => {
+  const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Constantinople })
+  const eei = await getEEI()
+  const evm = await EVM.create({ common, eei })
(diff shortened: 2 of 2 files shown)


holgerd77:
Have rebased this via UI.

holgerd77:
Ah sorry, hope I didn't mess up your update with the rebase, thought this was final since there already was a review requested.

acolytec3:
> Ah sorry, hope I didn't mess up your update with the rebase, thought this was final since there already was a review requested.

All good.  I noticed I had left a `tape.only` in my new test that I needed to remove so did that and then put the comments in the right place since I accidentally moved them :sweat_smile:

holgerd77:
So, to read the state of this PR from the latest commit message @jochem-brouwer has sent, I would assume that this is still be worked on, right? So I would put this in a "WIP" state again.

Would be great if you can very explicitly state here once you are ready.

For now I would like to skip digging into this whole discussion, since I would rather like to concentrate on other things like the README updates and the like. Let me know if this is not sufficient and you need an additional opinion here though.

jochem-brouwer:
This is indeed WIP, consensus tests are failing.

holgerd77:
Rebased this via UI.

Will merge this going along with the decision outcome of our team call yesterday (so, we agreed to merge but let this - to some extend - an open design question with the possibility to later adopt (and e.g. add a flag) if we are totally not happy with it.

acolytec3 (commented):
LGTM!  Took 6.5s on the modern Intel I7 in my LG laptop so not bad.  I turned the one test off after a couple minutes of running it on `master` so it's definitely improving the performance.

PR #2012, opened 2022-07-05
Util: Add recovery to ECDSA interface

ScottyPoi:
Re: #1961

Adds a recovery value to the ECDSA interface.
Updates EC functions to use new interface.
Updates tests to use updated EC functions and interface.


packages/tx/src/legacyTransaction.ts
```diff
@@ -367,6 +361,15 @@ export default class Transaction extends BaseTransaction<Transaction> {
...
       }
+      if (
+        common?.hardfork() === 'chainstart' ||
+        common?.hardfork() === 'homestead' ||
+        common?.hardfork() === 'tangerineWhistle'
+      ) {
+        if (v > 28) {
+          throw new Error('Legacy txs need either v = 27/28 or v >= 37 (EIP-155 replay protection)')
+        }
+      }
```
holgerd77, line 372: This is both not correct (legacy txs can still be used on other hardforks) and not reachable, since all v cases caught here are already included in the conditional clause from above.

So this can be completely removed. 🙂

holgerd77 (changes requested):
Hi there,
thanks for reworking, already looks a lot cleaner respectively pretty clean and cleaned-up in general 👍. Some further things to address.

packages/tx/src/legacyTransaction.ts
ScottyPoi: Without this, 90 TX tests fail like this for Homestead, Frontier, and EIP-150. 
```
not ok 2937 Transaction should be invalid on Homestead
  ---
    operator: ok
    expected: true
    actual:   false
```
So either the tests themselves are wrong, or something is not being caught by the rest of the conditional for another reason

ScottyPoi: The way that the TX test work, using the `transactionRunner`, I'm finding it very hard to track down exactly which test is which

packages/util/src/signature.ts
```diff
@@ -4,10 +4,23 @@ import { toBuffer, setLengthLeft, bufferToHex, bufferToInt, bufferToBigInt } fro
 import { SECP256K1_ORDER, SECP256K1_ORDER_DIV_2 } from './constants'
 import { assertIsBuffer } from './helpers'
 
+/**
+ * Interface for ECDSA signature
+ *
+ * @interface ECDSASignature
+ *
```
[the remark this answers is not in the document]
holgerd77: Ah, I actually only meant the `@interface ECDSASignature` part, where TypeScript already knows a) that it's an interface and b) the name, so didn't mean the text comment itself.

Rather a bit of a nit here, since the additional information in the text is also not that big, so guess we can keep it this way.

packages/util/src/signature.ts
```diff
@@ -30,26 +30,32 @@ export interface ECDSASignature {
...
-  if (v === BigInt(0) || v === BigInt(1)) return v
-
-  if (chainId === undefined) {
+export function calculateSigRecovery(v: bigint): bigint {
+  if (v > BigInt(28) && v < BigInt(35)) {
+    return v
+  }
+  if (v < BigInt(27) && v > BigInt(1)) {
+    return v
+  }
```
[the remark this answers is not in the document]
holgerd77: Can you give a reasoning why you took the path of not throwing and instead return an (arbitray) invalid signature value?

I do not find this clean and would strongly prefer the other solution if there is no really strong argument why we should do this here.

holgerd77:
Hmm, just tested.

Strangely enough this branch works for me locally, so running the tx tests with `npm run test`, I also made sure to rebuild both Common and Util before. 🤔

Hmm. Needs some investigation I fear.

holgerd77:
It would be great if this can be finished this week. 🙂

Issue #2017, opened 2022-07-05
`vm.copy()` doesn't preserve all options, like `allowUnlimitedContractSize`.


davidmurdoch:
I expected initial options to be preserved when calling `vm.copy()`, but this does not seem to be the case:

```javascript
const assert = require("assert");
const VM = require("@ethereumjs/vm").default;

const vm = new VM({
  allowUnlimitedContractSize: true
});
const vmCopy = vm.copy();
assert.strictEqual(vm._allowUnlimitedContractSize, true); // passes
assert.strictEqual(vmCopy._allowUnlimitedContractSize, true); // fails

```

holgerd77:
@davidmurdoch Hi there, yes that's a bug, thanks for reporting!

I guess this is also a candidate which we want to do towards `v5-maintenance` and `master` (currently containing the v5 breaking release beta code), with the `master` changes likely affecting both the `VM` and the `EVM` package.

Here are some references:

- `v5-maintenance` [VM.copy()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/v5-maintenance/packages/vm/src/index.ts#L487) function
- `master` [VM.copy()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/vm.ts#L217) function
- `master` [EVM.copy()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/evm/src/evm.ts#L910) (there is some opts stuff in there, can't judge on first look, this would at least need a confirmation if things are passed. Current assumption: nevertheless rather not)

I will assign @ScottyPoi here since I think this might fit, Scotty, feel free to unassign if you have other plans or think this is not so fitting.

If you do it would be great if you can add a simple test case (in the form of the code David posted, or at least similar) for every option which needs to be passed.

With these activational options we need to be careful, so `activatePrecompiles` and `activateGenesisState`, so that there is no re-creation/overwriting of state in an already initialized state manager. As an update and when thinking about it: these should not be passed over along with copy().

I guess the very most (all?) of the other options as well, but would be good to give this some additional own thinking. I would think at least the "simple" flag and flag-like options like `hardforkByBlockNumber`, `hardforkByTD`, `allowUnlimitedContractSize`, ... should be safe to pass, also things like the custom opcodes and precompiles (this is actually instead really necessary and would otherwise be a bug as well).

---

[...]

PR #2022, opened 2022-07-07
Vm-fix-copy-bug

ScottyPoi:
Fixes for #2017 on master

User found a bug in vm.copy().  Options such as `allowsUnlimitedContractSize` were not being passed on to the copy.

Fixed by editing the `VM.create({...})  ` call in `VM.copy()` to specifically include these options.

### DIFF
--- a/packages/vm/src/vm.ts
+++ b/packages/vm/src/vm.ts
@@ -223,14 +223,18 @@ export class VM extends AsyncEventEmitter<VMEvents> {
     // rather than deep copying the original ones since the copy of the `StateManager`
     // inside the EVM and EEI will be different than the `VM` level copy otherwise
     const eeiCopy = new EEI(stateCopy, commonCopy, blockchainCopy)
-    const evmCopy = new EVM({ eei: eeiCopy, common: commonCopy })
+    const evmCopy = new EVM({
+      eei: eeiCopy,
+      common: commonCopy,
+      allowUnlimitedContractSize: (this.evm as any)._allowUnlimitedContractSize,
+    })
     return VM.create({
       stateManager: stateCopy,
       blockchain: blockchainCopy,
       common: commonCopy,
       evm: evmCopy,
       eei: eeiCopy,
-      hardforkByBlockNumber: this._hardforkByBlockNumber ? this._hardforkByBlockNumber : undefined,
+      hardforkByBlockNumber: this._hardforkByBlockNumber ? true : undefined,
       hardforkByTD: this._hardforkByTD ? this._hardforkByTD : undefined,
     })
   }
--- a/packages/vm/src/vm.ts
+++ b/packages/vm/src/vm.ts
@@ -230,6 +230,8 @@ export class VM extends AsyncEventEmitter<VMEvents> {
       common: commonCopy,
       evm: evmCopy,
       eei: eeiCopy,
+      hardforkByBlockNumber: this._hardforkByBlockNumber ? this._hardforkByBlockNumber : undefined,
+      hardforkByTD: this._hardforkByTD ? this._hardforkByTD : undefined,
     })
   }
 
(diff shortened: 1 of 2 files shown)


packages/vm/src/vm.ts
```diff
@@ -223,13 +223,19 @@ export class VM extends AsyncEventEmitter<VMEvents> {
     // rather than deep copying the original ones since the copy of the `StateManager`
     // inside the EVM and EEI will be different than the `VM` level copy otherwise
     const eeiCopy = new EEI(stateCopy, commonCopy, blockchainCopy)
-    const evmCopy = new EVM({ eei: eeiCopy, common: commonCopy })
+    const evmCopy = new EVM({
+      eei: eeiCopy,
+      common: commonCopy,
+      allowUnlimitedContractSize: (this.evm as any)._allowUnlimitedContractSize,
```
acolytec3, line 229: We should also add `customPrecompiles` and `customOpcodes` from the `EVMOpts` property list here.

ScottyPoi: okay

packages/vm/src/vm.ts
```diff
@@ -223,13 +223,21 @@ export class VM extends AsyncEventEmitter<VMEvents> {
...
+      customOpcodes: (this.evm as any)._customOpcodes,
+      customPrecompiles: (this.evm as any)._customPrecompiles,
+    })
```
holgerd77, line 232: Hmm, this whole structure here is unlucky and not clean (it was before, nothing you introduced).

So one basic thing which also goes in here to note before: this issue was also about adopting the EVM `copy()` function, not only the VM one, so e.g. for these `customOpcodes` and `customPrecompiles` functions.

So: what is unlucky here that this is actually *not* using the (an) `EVM.copy()` function but doing a `new EVM(...)` instantiation.

The same goes for the `EEI`. The `EEI` has a copy function as well in the interface (`EEIInterface` in `types.ts` in the EVM package).

This is actually somewhat tricky due to what the comment above hints to (we want to have the same StateManager and Blockchain in the copied VM/EVM), but I think I got it now. 😋

So here we want to end up with the following:

1. Only do the EVM copy (so `EVM.copy()`), no other copies whatsoever (not state, blockchain, common, EEI).
2. For `VM.create()` use state, blockchain, common, EEI directly from the EVM with similar `any` casts as above

Can you please resubmit this PR and do things in the following order:

1. Adopt this `VM.copy()` function here as described (so using `EVM.copy()` + other adoptions), see if things are still working respectively tests are passing
2. Add the 3 missing parameters to `EVM.copy()` (allowUnlimitedContractSize, customOpcodes, customPrecompiles) (there should nothing of these needed to be made public, would rather be good if _allowUnlimitedContractSize can be made `protected` again)
3. Add new tests for EVM, see if things are working
4. Add the new options to `VM.copy()`, again, tests, see if things are working

Thanks! 🙂 

*first* adopt the `EVM.copy()` function with the

packages/vm/tests/api/index.spec.ts
```diff
@@ -196,6 +203,121 @@ tape('VM -> hardforkByBlockNumber, hardforkByTD, state (deprecated), blockchain'
...
+      evm: evm,
+    })
+    vmCopy = await vm.copy()
+    st.deepEqual(
+      (vm.evm as any)._customPrecompiles,
+      (vmCopy.evm as any)._customPrecompiles,
+      'vm.copy() successfully copied customPrecompiles option'
+    )
+    console.log((vm.evm as any)._customPrecompiles)
+  })
```
holgerd77, line 320: These tests are misplaced here and should be added to the EVM tests.

holgerd77 (changes requested):
@ScottyPoi thanks a lot for tackling, some problems in this whole design space which are a bit more fundamental than expected 😋, so this would need a resubmission to deal with these on a more fundamental level.

packages/vm/tests/api/index.spec.ts
holgerd77: Ah no, just the ones with the EVM options, sorry.

PR #2023, opened 2022-07-07
Update @ethereumjs/vm USAGE example

reginaldlong:
There is no longer an `evm` field on the vm object, and `runCode` + the event emitter are defined directly on the `vm` object.


holgerd77:
Hi there,
thanks for the PR, that's nice, I have a huge PR open though rewriting the whole EVM/VM README files here #2016. I would close here since things would conflict.

I will admin-merge the other PR though (was waiting a bit for a review), to prevent that more people are getting confused by the state of the README files.

PR #2027, opened 2022-07-07
master-vm-copy-bug

ScottyPoi:
Edits VM.copy() to use EVM.copy() instead of running copy() on each parameter.  

Addings missing options to vm.copy().  So now options `allowUnlimitedContractSize`, `hardforkByBlockNumber`, `hardforkByTD`, `customOpcodes`, and `customPrecompiles` should successfully copy 

Adds tests to vm and evm to test copied options.


holgerd77 (approved):
Thanks for the resubmission, looks great! 😄 👍 

Making `customOpcodes` and `customPrecompiles` public should actually not be necessary, I removed this commit, also force-pushed a rebased version.

Will merge if CI passes.

PR #2029, opened 2022-07-08
Rename evm loggers

### LINKED ISSUE #2021: EVM: adopt Logger Names
The logger names in the EVM are still named after the vm, so e.g. `vm:eei:gas`, `vm:opts:SSTORE` (dynamic) and so on. This should be adopted and instead `evm:` prefixed logger names used, this would also need an update in the README docs.

A bit unsure about the `vm:evm` prefixed loggers (2 I guess), if this should become `evm:evm` or rather named somewhat differently (`evm:exec` or so? just a very quick shot though).

acolytec3:
Fixes #2021

### DIFF
--- a/packages/evm/README.md
+++ b/packages/evm/README.md
@@ -204,11 +204,11 @@ The following loggers are currently available:
 
 | Logger                            | Description                                         |
 | --------------------------------- | --------------------------------------------------- |
-| `vm:evm`                          |  EVM control flow, CALL or CREATE message execution |
-| `vm:evm:gas`                      |  EVM gas logger                                     |
-| `vm:eei:gas`                      |  EEI gas logger                                     |
-| `vm:ops`                          |  Opcode traces                                      |
-| `vm:ops:[Lower-case opcode name]` | Traces on a specific opcode                         |
+| `evm`                             |  EVM control flow, CALL or CREATE message execution |
+| `evm:gas`                         |  EVM gas logger                                     |
+| `evm:eei:gas`                     |  EEI gas logger                                     |
+| `evm:ops`                         |  Opcode traces                                      |
+| `evm:ops:[Lower-case opcode name]`| Traces on a specific opcode                         |
 
 Here are some examples for useful logger combinations.
 
--- a/packages/evm/src/evm.ts
+++ b/packages/evm/src/evm.ts
@@ -35,8 +35,8 @@ import {
 import { EEIInterface } from './types'
 import { TransientStorage } from './transientStorage'
 
-const debug = createDebugLogger('vm:evm')
-const debugGas = createDebugLogger('vm:evm:gas')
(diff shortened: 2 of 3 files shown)


acolytec3:
Should we also go ahead and update references in code comments from `VM` to `EVM` where it makes sense within the EVM?  There aren't too many but wasn't sure what y'all think.  It's small things but maybe useful since we now have to distinct but closely related packages.

holgerd77:
> Should we also go ahead and update references in code comments from `VM` to `EVM` where it makes sense within the EVM? There aren't too many but wasn't sure what y'all think. It's small things but maybe useful since we now have to distinct but closely related packages.

Yes, I would think so. 👍

acolytec3:
Cool, have finished cleaning up the docs and code comments.  I missed a few instances in the docs previously but this should be ready for review now.

packages/evm/README.md
```diff
@@ -138,10 +138,10 @@ It is possible to individually activate EIP support in the EVM by instantiate th
 
 ```typescript
 import { Chain, Common } from '@ethereumjs/common'
-import { VM } from '@ethereumjs/vm'
+import { EVM } from '@ethereumjs/evm'
 
 const common = new Common({ chain: Chain.Mainnet, eips: [2537] })
-const vm = new VM({ common })
+const vm = new EVM({ common })
```
holgerd77, line 144: Ah, no, this is not correct, read the initial sentence from this section. We cannot instantiate a standalone EVM right now (and this code just would not work), therefore the example with the VM.

packages/evm/README.md
```diff
@@ -216,38 +216,38 @@ The following loggers are currently available:
...
-DEBUG=vm:*,vm:*:*,-vm:ops ts-node test.ts
+DEBUG=evm:*,evm:*:*,-evm:ops ts-node test.ts
 ```
 
-Run some specific loggers including a logger specifically logging the `SSTORE` executions from the VM (this is from the screenshot above):
+Run some specific loggers including a logger specifically logging the `SSTORE` executions from the EVM (this is from the screenshot above):
 
 ```shell
-DEBUG=vm:evm,vm:ops:sstore,vm:*:gas ts-node test.ts
+DEBUG=evm,evm:ops:sstore,vem:*:gas ts-node test.ts
```
holgerd77, line 250: `v` sneaked in. 😋

holgerd77 (changes requested):
Nice, two small things to address.

packages/evm/README.md
```diff
@@ -141,7 +141,7 @@ import { Chain, Common } from '@ethereumjs/common'
 import { EVM } from '@ethereumjs/evm'
```
holgerd77, line 141: Import needs an update as well. 😜

holgerd77 (approved):
LGTM

Issue #1935, opened 2022-06-05
Additional lint rules to consider


ryanio:
After working in the monorepo extensively, I compiled some ESLint rules that should help improve dev ex, code quality and style.

These are nothing urgent or breaking and can be picked up over time. (These rules should be added to the top-level `config/eslint.js` to automatically be included in all packages.)

* [x] [strict-boolean-expressions](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md) https://github.com/ethereumjs/ethereumjs-monorepo/pull/2030
* [ ] [require-await](https://eslint.org/docs/rules/require-await) Extracted to https://github.com/ethereumjs/ethereumjs-monorepo/issues/2156 due to its size
* [x] [import-sort-order](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md) https://github.com/ethereumjs/ethereumjs-monorepo/pull/2086
* [x] [object-shorthand](https://eslint.org/docs/rules/object-shorthand) https://github.com/ethereumjs/ethereumjs-monorepo/pull/2131
* [x] [array-foreach](https://github.com/github/eslint-plugin-github/blob/main/docs/rules/array-foreach.md) https://github.com/ethereumjs/ethereumjs-monorepo/pull/2130
* [x] [no-sparse-arrays](https://eslint.org/docs/rules/no-sparse-arrays) included via `eslint:recommended`

ryanio:
nothing breaking, also nothing super urgent, just some suggestions to improve code quality/style, can be done over time 👍

gabrocheleau:
+1 [strict-boolean-expressions](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md), would have caught an issue that I just spent quite a bit of time troubleshooting :)

ryanio:
Here's some of my preferred config, hardhat's [eslintrc rules](https://github.com/NomicFoundation/hardhat/blob/master/config/eslint/eslintrc.js) are a great resource.

```js
    "@typescript-eslint/consistent-type-imports": "error",
    "import/order": [
      "error",
      {
        alphabetize: {
          order: "asc",
        },
        groups: [
          "object",
          ["builtin", "external"],
          "parent",
          "sibling",
          "index",
          "type",
        ],
        "newlines-between": "always",
      },
    ],
[...]

PR #2030, opened 2022-07-08
Monorepo: eslint strict boolean expressions

gabrocheleau:
This PR adds the [strict-boolean-expressions](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md) eslint rule and applies it throughout the monorepo. I also added `isTruthy` and `isFalsy` type guard utils for convenience. Eventually, I think we can refactor most of those  `isTruthy` and `isFalsy` by providing more accurate types upstream and explicitly checking for truthy/falsy cases instead of relying on JS truthiness.  

 Related to #1935

### DIFF
--- a/packages/util/src/types.ts
+++ b/packages/util/src/types.ts
@@ -118,3 +118,35 @@ export function toType<T extends TypeOutput>(
       throw new Error('unknown outputType')
   }
 }
+
+type Falsy = false | '' | 0 | null | undefined | 0n
+
+/**
+ * Returns true if a value is falsy
+ *
+ * @param value - Value to check for falseness
+ *
+ * @deprecated This helper function should only be used temporarily until the monorepo types are explicit enough
+ */
+export function isFalsy(value: unknown): value is Falsy {
+  return !!(
+    value === false ||
+    value === '' ||
+    value === 0 ||
+    Number.isNaN(value) ||
+    value === null ||
+    typeof value === 'undefined' ||
+    value === BigInt(0)
+  )
+}
+
+/**
+ * Returns true if a value is truthy
+ *
+ * @param value - Value to check for truthiness
+ *
+ * @deprecated This helper function should only be used temporarily until the monorepo types are explicit enough
+ */
+export function isTruthy<T>(value: T | Falsy): value is T {
+  return !isFalsy(value)
+}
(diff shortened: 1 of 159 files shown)


holgerd77:
Ugh. What a PR. 😜

Great though! 🙏 ❤️

packages/util/src/types.ts
```diff
@@ -118,3 +118,20 @@ export function toType<T extends TypeOutput>(
       throw new Error('unknown outputType')
   }
 }
+
+type Falsy = false | '' | 0 | null | undefined
+
+export function isFalsy(value: unknown): value is Falsy {
```
holgerd77, line 124: Can you add a very very explicit note here (and for `isTruthy` as well) that this is a function which should only temporarily be used and otherwise be replaced with the direct checks (as you mentioned as well), together with some overall context why this function exists?

I otherwise have the strong fear that this looks like as something "how we do things" and other developers rather align with this.

packages/util/src/types.ts
```diff
@@ -118,3 +118,20 @@ export function toType<T extends TypeOutput>(
...
+    value === 0 ||
+    value === null ||
+    typeof value === 'undefined' ||
+    value === BigInt(0)
+  )
+}
+
+export function isTruthy<T>(value: T | Falsy): value is T {
+  return !isFalsy(value)
+}
```
holgerd77, line 137: Also, on another note: are you confident that this code is correct, since we would be using this all over the place now in our libraries (for now)? So does this match 1:1 the previous behavior?

I guess we very much would want to have some tests (or rather: somewhat complete, type-wise) on this, to be really sure this behaves correctly in all cases.

gabrocheleau: Yeah I am confident, but can definitely add tests since as you say this is used throughout

holgerd77: Thanks, yeah, guess this should be some really compact LoCs to cover the various cases.

gabrocheleau: Added some tests

packages/util/src/types.ts
gabrocheleau: Good point. I added a note and marked them as deprecated. The goal should definitely to get rid of those eventually, and it should not be used in new code (I think the `@deprecated` natspec will help achieve that)

holgerd77: Thanks, that's a good solution! 👍 😄

holgerd77:
Ah, interestingly enough it is not working here (GitPOAP), but activated here #2035.

Two theories:

1. The other PR got submitted later and the Bot also takes the submission point into account or something like that.
2. PRs are only counted when not submitted from the original repo but from a fork.

We'll see (or someone has already the answer). 🙂

PR #2034, opened 2022-07-12
Common, Client: Sepolia DNS config and activation

holgerd77:
This PR adds Sepolia DNS discovery config to Common and DNS discovery activation to the client which solves the Sepolia initial sync connection problem.

For context see the [discv4-dns-lists](https://github.com/ethereum/discv4-dns-lists) repo and the [bootnode config](https://github.com/ethereum/go-ethereum/blob/master/params/bootnodes.go#L108) from Go Ethereum.


g11tech (approved):
awesome!

PR #2043, opened 2022-07-15
Trie: allow customizing hash algorithm

libotony:
Keccak256 is the default hash algorithm for Ethereum, I use blake2b256 in my case. It's would be better that the underlying package supports customizing the hash algorithm rather than forking it.So I added the package level configurable hash function, by default it's keccak256.

### DIFF
--- a/packages/statemanager/src/stateManager.ts
+++ b/packages/statemanager/src/stateManager.ts
@@ -320,14 +320,14 @@ export class DefaultStateManager extends BaseStateManager implements StateManage
    */
   async getProof(address: Address, storageSlots: Buffer[] = []): Promise<Proof> {
     const account = await this.getAccount(address)
-    const accountProof: PrefixedHexString[] = (await Trie.createProof(this._trie, address.buf)).map(
-      (p) => bufferToHex(p)
+    const accountProof: PrefixedHexString[] = (await this._trie.createProof(address.buf)).map((p) =>
+      bufferToHex(p)
     )
     const storageProof: StorageProof[] = []
     const storageTrie = await this._getStorageTrie(address)
 
     for (const storageKey of storageSlots) {
-      const proof = (await Trie.createProof(storageTrie, storageKey)).map((p) => bufferToHex(p))
+      const proof = (await storageTrie.createProof(storageKey)).map((p) => bufferToHex(p))
       let value = bufferToHex(await this.getContractStorage(address, storageKey))
       if (value === '0x') {
         value = '0x0'
@@ -365,7 +365,7 @@ export class DefaultStateManager extends BaseStateManager implements StateManage
 
     // This returns the account if the proof is valid.
     // Verify that it matches the reported account.
-    const value = await Trie.verifyProof(rootHash, key, accountProof)
+    const value = await new Trie().verifyProof(rootHash, key, accountProof)
 
     if (value === null) {
       // Verify that the account is empty in the proof.
@@ -411,7 +411,7 @@ export class DefaultStateManager extends BaseStateManager implements StateManage
       const storageProof = stProof.proof.map((value: PrefixedHexString) => toBuffer(value))
       const storageValue = setLengthLeft(toBuffer(stProof.value), 32)
       const storageKey = toBuffer(stProof.key)
-      const proofValue = await Trie.verifyProof(storageRoot, storageKey, storageProof)
+      const proofValue = await new Trie().verifyProof(storageRoot, storageKey, storageProof)
       const reportedValue = setLengthLeft(
         Buffer.from(RLP.decode(Uint8Array.from((proofValue as Buffer) ?? [])) as Uint8Array),
         32
--- a/packages/trie/README.md
+++ b/packages/trie/README.md
@@ -73,8 +73,8 @@ const trie = new Trie()
 
 async function test() {
   await trie.put(Buffer.from('test'), Buffer.from('one'))
-  const proof = await Trie.createProof(trie, Buffer.from('test'))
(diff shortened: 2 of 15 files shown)


holgerd77:
Ah, that's so insane, we actually talked *yesterday* on a call that we would like to have *exactly* this functionality!!! 🙂 🙃 That's actually hyper-strange, never had this auto-fullfilling before, thanks a lot for submitting! ❤️

One thing: our idea was to pass in the hash function via the constructor in the options. I would still have some tendency to prefer this, or do you think that your approach has substantial advantages I might not see atm?

Again: really great, nice!

Short note: we are currently in a stage where we have our upcoming beta releases on `master`, so this would be released in an upcoming - likely - Beta 3 release for Trie with final releases still a couple of weeks away, just to let you know.

libotony:
Thank you for your quick response, I am very excited to see you have the same thoughts 😄. I started with the same idea of adding hash to constructor options, it was fine till I get to the static functions in SecureTrie `verifyProof` and `verifyRangeProof`. Especially there is new Trie called inside `verifyRangeProof`, the hash function might need to pe passed in these functions, I was afraid that this changes too much in the API level, and might be rejected. So I turned to another approach, which to be honest is not the most elegant one. I would like to hear your thoughts regarding this.

holgerd77:
> `verifyRangeProof`

Would it have stupid side effects if we just *leave* the default hash function to be used there for these cases? There is not much hashing needed for this, and I would otherwise think - in a very first round - that this might be the easiest and most pragmatic solution? 🤔

WDYT?

holgerd77:
Ah, sorry, but you would also want to use just *another* hash function, right? We are coming more from the corner "using a more performant implementation".

Hmm, let's rethink this again.

libotony:
> Ah, sorry, but you would also want to use just _another_ hash function, right? We are coming more from the corner "using a more performant implementation".
> 
> Hmm, let's rethink this again.

You are right.

holgerd77:
Update: hmm, I would cautiously say it would be no drama if we make all these methods non-static (then it would fit very well with these breaking releases we are doing right now). It's a *little* bit more inconvenient but I guess doesn't have any substantial side effect (?).

But would like to have an additional confirmation on that.

libotony:
> Update: hmm, I would cautiously say it would be no drama if we make all these methods non-static (then it would fit very well with these breaking releases we are doing right now). It's a _little_ bit more inconvenient but I guess doesn't have any substantial side effect (?).
> 
> But would like to have an additional confirmation on that.

That would be great, please let me know. I would like to contribute once we got confirmation.

holgerd77:
> That would be great, please let me know. I would like to contribute once we got confirmation.

Ok, cool!

Just for some context: I'll be on holidays in 3 days and I am *right now* in the process of preparing Beta 2 releases here #2045 - so also for the Trie library - to be released today or tomorrow. If you want to have this released early it might be a good idea to rework this in this direction directly without waiting, then I will relatively likely still be able to take this in. And I guess chances that there are substantial reservations on this is relatively low.

acolytec3:
> > That would be great, please let me know. I would like to contribute once we got confirmation.
> 
> Ok, cool!
> 
> Just for some context: I'll be on holidays in 3 days and I am _right now_ in the process of preparing Beta 2 releases here #2045 - so also for the Trie library - to be released today or tomorrow. If you want to have this released early it might be a good idea to rework this in this direction directly without waiting, then I will relatively likely still be able to take this in. And I guess chances that there are substantial reservations on this is relatively low.

Agree on this. Given we're doing another beta in a few weeks, adding it now will give us the chance to experiment and see if there are any drawbacks and we could always revert or revise further.

holgerd77:
Also, some implementation note: in case you do the static to member-based method updates, you would need to do these updates across the whole monorepo (which in this case likely sounds worse than it is, the proof functionality is really only used a handful of times or so).

So do a search `Trie.verifyProof` (e.g.) throughout the monorepo libraries, and these occurrences would then be needed to be replaced with `new Trie().verifyProof` I guess.

Also: please then directly make *all* proof functions non-static (so I guess: prove, createProof, verifyProof, verifyRangeProof) and not only the hash-function-using ones.

libotony:
> Also, some implementation note: in case you do the static to member-based method updates, you would need to do these updates across the whole monorepo (which in this case likely sounds worse than it is, the proof functionality is really only used a handful of times or so).
> 
> So do a search `Trie.verifyProof` (e.g.) throughout the monorepo libraries, and these occurrences would then be needed to be replaced with `new Trie().verifyProof` I guess.
> 
> Also: please then directly make _all_ proof functions non-static (so I guess: prove, createProof, verifyProof, verifyRangeProof) and not only the hash-function-using ones.

Thanks for the kind note. I'll start right away.

holgerd77:
👍

Cool. 🙂 We use linting on the source code so you will likely need to run "npm run lint:fix" (or without :fix to just check) on your code, otherwise CI will break in many cases.

Also, please add minimally one test case - e.g. using a dummy hash function (e.g. just a custom function adding a prefix "hash_" to keys should already workI guess?).

Ok, happy coding! 😃

libotony:
Ready for review.

holgerd77 (approved):
Thanks a lot, that's impressively clean and generalized work! 🤗🎉

Will directly merge, love it!

holgerd77:
@libotony released (as Beta 2, see [@ethereumjs/trie v5.0.0-beta.2](https://github.com/ethereumjs/ethereumjs-monorepo/releases/tag/%40ethereumjs%2Ftrie%405.0.0-beta.2)), thanks a lot for the contribution! 😀 ❤️

PR #2044, opened 2022-07-15
client: Add instructions for sepolia merge

g11tech:
Modify the testnet instructions for sepolia and cleanup kiln as it will be deprecated soon.


holgerd77:
Rebased this via UI.

packages/client/merge/README.md
```diff
@@ -0,0 +1,93 @@
...
+docker build . --tag ethereumjs:latest
+```
+You may now do appropriate directory/file mounts for `data` dir and `jwtsecret` file and provide their path appropriately in the `client` run command.
+
+#### Run client
+
+This the the bare minimum command required for interop with the CL `lodestar`. You may add further args to expose functionality for your use case.
+
+```shell
+npm run client:start -- --network sepolia --datadir /path/to/sepolia/data/dir --rpcEngine --jwt-secret /path/to/jwtsecret/file
```
holgerd77, line 67: I will just ask some questions here, otherwise merge.

Depending on the answers I would then directly integrate/update along the release PR.

So here: do we maybe want to skip the `--datadir` option now? Same for Lodestar? This is now totally optional and if people have any datadir change needs then they are just in the mode of "unrelatedly using the datadir option".

holgerd77: Second question on this: is the JWT secret generated on first usage when used like this?

If not: then I would prefer to describe a slightly alternative way to use (so: EthereumJS without the option -> Lodestar taking over the path). Guess otherwise people might be a bit lost here.

If yes: then I would like to add a sentence on this.

packages/client/merge/README.md
```diff
@@ -0,0 +1,93 @@
...
+This will start `ethereumjs` client on `sepolia` network and expose `engine` endpoints at `8551`(default port, can be modified see cli help via `--help`).
+
+In case you want to host `engine_*` without auth (not recommended, only for debugging purposes), pass `--rpcEngineAuth false` as extra argument in the above run command (and you may skip `--jwt-secret`).
+
+
+### Consensus - Lodestar Setup
+
+#### Beacon
+
+1. Use lodestar branch `unstable` and run `yarn && yarn build`
```
holgerd77, line 79: Is it still necessary respectively recommended to use `unstable`? I just used `master` and that worked fine as well? Might unstable have dedicated disadvantages (regarding - tada - stability 😋)?

packages/client/merge/README.md
```diff
@@ -0,0 +1,93 @@
...
+2. Run cmd: `./lodestar beacon --network sepolia --rootDir /path/to/sepolia/data/dir --jwt-secret /path/to/jwtsecret/file`
+
+This will by default try reaching out `ethereumjs` over the endpoint `8551`. (You may customize this in conjuction with `ethereumjs`, see lodestar cli help via `--help`).
+
+You may provide `--weakSubjectivityServerUrl` (with a synced `sepolia` beacon node endpoint as arg value) and  `--weakSubjectivitySyncLatest` to start directly off the head on the `sepolia` beacon chain, possibly triggering (backfill) beacon sync on ethereumjs.
```
holgerd77, line 84: Just to double-check here: are these options still working/up-to-date?

packages/client/merge/README.md
g11tech: >So here: do we maybe want to skip the --datadir option now?
yes we may skip :+1:

g11tech: >Second question on this: is the JWT secret generated on first usage when used like this? 
I added a jwt-secret step (in above section) although we generate one randomly and use it, which can then be provided to lodestar, but from what i have seen, many people get confused here and its better to just have an explicit instruction.

packages/client/merge/README.md
g11tech: yes, these options work! we need a working sepolia synced beacon node exposing debug endpoints to download state, but i don't have one handy hosted by lodestar for sepolia

PR #2048, opened 2022-07-15
Disable hardhat tests CI job

acolytec3:
Disables the hardhat tests for now since they are autofailing and won't work for the forseeable future (until Hardhat upgrades to work with our new breaking releases.


holgerd77 (approved):
LGTM

PR #2049, opened 2022-07-15
Monorepo: only lint changed files

acolytec3:
Updates our lint script to only lint changed files (i.e. the diff between your local branch and HEAD of the remote tracking branch you're working with)

The upside of this is it will allow us to do `git push` and it will quickly run the linter without linting every single package top to bottom.


acolytec3:
Okay, this is ready for review.  This makes one major change to to our lint rules:
1) When you run `git push`, the `lint:diff` script is now run instead of `lint` and it checks to see if there is a remote tracking branch and if so, only lints the files that have changed since the last push to the remote branch.  If no remote branch is detected, it runs the full linter.  
2) The CI continues to run the `lint` script on all jobs (as currently).

The case for this is that it dramatically speeds up the linter when you're push commits and we no longer have the tendency (or at least I no longer will) to run `git push --no-verify`. :grimacing:

ScottyPoi (commented):
I made a test branch off of this and pushed simple changes to it.  The pre-push linting was indeed much quicker, and only actually proceeded on the package with changes.

However, when I made a change to a function in the `util` package that broke code in another package, this was not caught in the pre-push linting.  Since it only looked at files that changed, it missed something it would have otherwise caught.

acolytec3:
> I made a test branch off of this and pushed simple changes to it. The pre-push linting was indeed much quicker, and only actually proceeded on the package with changes.
> 
> However, when I made a change to a function in the `util` package that broke code in another package, this was not caught in the pre-push linting. Since it only looked at files that changed, it missed something it would have otherwise caught.

What sort of a change did you make?  The linter only addresses code style that can detected via static analysis tools so if your code change was valid but "breaking" for another package (e.g. making a sync function async), those changes wouldn't be caught by the linter.  That would only get picked up by the tests.

ScottyPoi:
> What sort of a change did you make? The linter only addresses code style that can detected via static analysis tools so if your code change was valid but "breaking" for another package (e.g. making a sync function async), those changes wouldn't be caught by the linter. That would only get picked up by the tests.

Just altered the spelling of a function to see what would happen.  
I thought the linter would pick up that a call was made to a function that doesn't exist, but maybe it wouldn't anyway

ScottyPoi (approved):
Gotcha.  Ok, then LGTM

acolytec3:
Anybody else want to give this a test run or weigh in on the concept before I merge?

g11tech (approved):
awesome!

acolytec3:
Better DX for the win!

holgerd77:
Cool stuff. 😎

PR #2051, opened 2022-07-16
Add "recovery" to ECDSA Signature Interface

ScottyPoi:
Fixes Issue #1961

Adds `recovery: bigint` as a value in `ECDSA Signature` interface.

`recovery` is calculated from `v` and `chainId`, or just `v` if no chainId is provided, and `EIP-155 Replay Protection` is enabled.

Updates `signature.spec.ts` test in `util`
and `eip-3074-authcall.spec.ts` in `vm` to use new interface/API

Resubmit of PR #1995 and PR #2012

### DIFF
--- a/packages/util/src/signature.ts
+++ b/packages/util/src/signature.ts
@@ -52,33 +52,19 @@ export function ecsign(msgHash: Buffer, privateKey: Buffer, chainId?: bigint): E
  */
 
 export function calculateSigRecovery(v: bigint): bigint {
-    if (v === BigInt(0) || v === BigInt(1)) {
-      return v
-    } else if (v === BigInt(27) || v === BigInt(28)) {
-      return v - BigInt(27)
-    } else {
-      return v - (chainId * BigInt(2) + BigInt(35))
-    }
-  } else {
-    if (v === BigInt(0) || v === BigInt(1)) {
-      return v
-    } else if (v < BigInt(27)) {
-      throw new Error('invlaid v value < 27')
-    } else if (v > BigInt(28) && v < BigInt(35)) {
-      throw new Error('invlaid v value >28 & <35 ')
-    } else if (v > BigInt(36)) {
-      throw new Error('invlaid v value > 36')
-    }
-
-    if (v === BigInt(27) || v === BigInt(28)) {
-      return v - BigInt(27)
-    } else if ((v - BigInt(35)) % BigInt(2) === BigInt(0)) {
-      return BigInt(1)
-    } else if ((v - BigInt(35)) % BigInt(2) === BigInt(1)) {
+  if (v === BigInt(0) || v === BigInt(1)) {
+    return v
+  } else if (v === BigInt(27) || v === BigInt(28)) {
+    return v - BigInt(27)
+  } 
+  if (v >= BigInt(35)) {
+    if ((v - BigInt(35)) % BigInt(2) === BigInt(0)) {
       return BigInt(0)
     } else {
-      throw new Error(`Invalid v value`)
-    }
+      return BigInt(1)
+  }
+  } else {
+    throw new Error(`Invalid V value ${v}`)
   }
 }
 
--- a/packages/util/src/signature.ts
+++ b/packages/util/src/signature.ts
@@ -52,36 +52,25 @@ export function ecsign(msgHash: Buffer, privateKey: Buffer, chainId?: bigint): E
  */
 
 export function calculateSigRecovery(v: bigint, chainId?: bigint): bigint {
-  if (chainId !== undefined) {
     if (v === BigInt(0) || v === BigInt(1)) {
       return v
-    } else if (v === BigInt(27) || v === BigInt(28)) {
+  }
+  if (v === BigInt(27) || v === BigInt(28)) {
       return v - BigInt(27)
-    } else {
+  }
+  if (chainId !== undefined) {
       return v - (chainId * BigInt(2) + BigInt(35))
-    }
-  } else if (chainId === undefined) {
-    if (v === BigInt(0) || v === BigInt(1)) {
-      return v
-    } else if (v < BigInt(27)) {
+  } else {
+    if (v < BigInt(27)) {
       // Returns an invalid signature value instead of throwing error
-      throw new Error('invlaid v value < 27')
+      throw new Error('invlaid v value')
     } else if (v > BigInt(28) && v < BigInt(35)) {
       // Returns an invalid signature value instead of throwing error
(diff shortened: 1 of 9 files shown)


ScottyPoi:
@holgerd77 

Finally got everything to pass, without touching anything outside of `util`

Could use a sanity check but I think this is done.

holgerd77:
Rebased this via UI.

packages/util/src/signature.ts
```diff
@@ -18,30 +31,56 @@ export interface ECDSASignature {
...
+    }
+  } else {
+    if (v === BigInt(0) || v === BigInt(1)) {
+      return v
+    } else if (v < BigInt(27)) {
+      throw new Error('invlaid v value < 27')
+    } else if (v > BigInt(28) && v < BigInt(35)) {
+      throw new Error('invlaid v value >28 & <35 ')
+    } else if (v > BigInt(36)) {
+      throw new Error('invlaid v value > 36')
```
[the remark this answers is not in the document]
holgerd77: Also can you put minimally the error cases for the *generally*  invalid v values out of this conditional here and put at the beginning of the function code, since these are generic and apply for all cases. So this would be:

- `v > BigInt(1) && v < BigInt(27)` and
- `v > BigInt(28) && v < BigInt(35)`

(correct me if I am not correct here or missing something, also writing this down somewhat quickly)

I would also be a fan if you compatify the check to one conditional for all values and then just give one generic error message just passing in the actual v value, so `Invalid signature v value ${v}`.

packages/util/src/signature.ts
```diff
@@ -72,11 +107,11 @@ export const ecrecover = function (
  * @returns Signature
  */
 export const toRpcSig = function (v: bigint, r: Buffer, s: Buffer, chainId?: bigint): string {
-  const recovery = calculateSigRecovery(v, chainId)
-  if (!isValidSigRecovery(recovery)) {
+  try {
+    calculateSigRecovery(v, chainId)
+  } catch {
     throw new Error('Invalid signature v value')
```
holgerd77, line 113: Maybe add the actual v value here as well as suggested above.

packages/util/src/signature.ts
```diff
@@ -86,12 +121,7 @@ export const toRpcSig = function (v: bigint, r: Buffer, s: Buffer, chainId?: big
  * NOTE: Accepts `v == 0 | v == 1` for EIP1559 transactions
  * @returns Signature
  */
-export const toCompactSig = function (v: bigint, r: Buffer, s: Buffer, chainId?: bigint): string {
-  const recovery = calculateSigRecovery(v, chainId)
-  if (!isValidSigRecovery(recovery)) {
-    throw new Error('Invalid signature v value')
-  }
```
holgerd77, line 93: This removes the recovery check completely if I read the code correctly which shouldn't be the case I guess.

packages/util/src/signature.ts
```diff
@@ -103,41 +133,47 @@ export const toCompactSig = function (v: bigint, r: Buffer, s: Buffer, chainId?:
...
 /**
  * Convert signature format of the `eth_sign` RPC method to signature parameters
- *
  * NOTE: For an extracted `v` value < 27 (see Geth bug https://github.com/ethereum/go-ethereum/issues/2053)
  * `v + 27` is returned for the `v` value
  * NOTE: After EIP1559, `v` could be `0` or `1` but this function assumes
  * it's a signed message (EIP-191 or EIP-712) adding `27` at the end. Remove if needed.
  */
-export const fromRpcSig = function (sig: string): ECDSASignature {
+export const fromRpcSig = function (sig: string, chainId?: bigint): ECDSASignature {
```
holgerd77, line 141: Why do we need to add the `chainId` here? I think `calculateSigRecovery()` should also work without, so it shouldn't be necessary to change the function signature?

packages/util/src/signature.ts
```diff
@@ -103,41 +133,47 @@ export const toCompactSig = function (v: bigint, r: Buffer, s: Buffer, chainId?:
...
+    recovery = calculateSigRecovery(v, chainId)
     s[0] &= 0x7f
   } else {
     throw new Error('Invalid signature length')
   }
 
   // support both versions of `eth_sign` responses
-  if (v < 27) {
-    v = v + BigInt(27)
-  }
```
holgerd77, line 135: It's not necessary to double this code above since `recovery` isn't used up till here at all, so recovery calculation could be just added here, or am I missing something?

holgerd77 (changes requested):
Sorry, it's a lot again, hope you don't get frustrated (no need to). 🙂 

Just work down the things in some light-weight manner and do not try to "be perfect" or something, there will likely be 1-3 iterations more anyhow. We just slowly need to get this to a point where the code is as simple, robust and transparent as possible.

ScottyPoi:
@holgerd77 

I did another round on this and have everything passing.  

* I needed to tweak something in `LegacyTransaction` to allow the `tx` tests to pass.
* Changing this in `tx` caused some `client` tests to fail
* To fix these, I made changes to `miner.ts` and `pendingBlock.ts` in the client.
  * Both of these changes had to do with `setHardforkByBlockNumber`

ScottyPoi:
To achieve the same result without changing any defaults, I created a config option called `disableMinerHardforkByBlockNumber`, and applied it to the failing tests in `miner.spec.ts` and `pendingBlock.spec.ts`

The way the tests were set up, with custom chains and custom hardforks, the miner would fail to pass the custom hardfork information down to the miner, so the miner would change it back to `chainstart` every time.

holgerd77:
Thanks for the efforts, really appreciate the contiuous work! 🙏 Nevertheless I am still not getting why this needs to affect the upper level libraries in any way. We are *adding* just an additional field to Util - `recovery` - with `v` just staying exactly the same as it was before (that's at least how it should be). Why (the hell) does this lead to changes in something like `LegacyTransaction` and even a new client option??? 🤔

This should just be some additional convenience for the Util method users who can then choose to either use the `v` value or `recovery` depending on their needs.

I would put this on hold for now, I do not have the energy/ressources to properly review this in time before our final releases so this will likely not make it in, so be it. Eventually it is possible to later do in a non-breaking way.

ScottyPoi:
> I would put this on hold for now, I do not have the energy/ressources to properly review this in time before our final releases so this will likely not make it in, so be it. Eventually it is possible to later do in a non-breaking way.

Ok.  I got in over my head with this one, to be honest.  I will move on.

holgerd77:
Thanks, no worries, this happens, we just have to be honest about it things are ready or not and not wave things through which just need more thought it polishing. 🙂

PR #2054, opened 2022-07-18
Ensure EVM runs when nonce is zero

jochem-brouwer:
This PR is based on an old version of master. In here I address the nonce bug in #2009. 

I think that pre-EVM calls the state should be exactly as it is pre-EVM. Since we have introduced this "nonce - 1" logic at some point this should be removed (and I think this might be easy). This first version does the depth check and removes the 'nonce - 1' if the depth is zero.

I tried removing it by also moving the "increment nonce" logic in `interpreter.ts` the same way how I moved it now in `runTx`, but this has a failing chainstart test `RecursiveCreateContractsCreate4Contracts_d0g0v0_Frontier` which I should look into.


jochem-brouwer:
Ah right, we cannot update `interpreter.ts` to update the nonce *before* `runCall`. The reason is: if the current account A which executes a CREATE then creates an account B which recursively calls back into account A, which then does another CREATE, will then not have the "new nonce" and thus will try to create a new account at B, while it should create an account at another address C.

If you recursively call back into the CREATE-ing account, then the nonce never updates, so it always tries to CREATE at the same address.

acolytec3:
@jochem-brouwer I reverted the changes I made around setting the default `caller` and fixed the 2 broken `evm` tests based on it producing the correct address/code so should be good to go unless you have more changes to make.

acolytec3 (approved):
Not sure if there is more work to be done but this looks good to me!

jochem-brouwer:
Thanks!

PR #2055, opened 2022-07-19
evm: Remove dubious booleans in evm

acolytec3:
Following on #2030, I noticed the linter complaining about some implicit boolean truthiness being checked in `evm.ts` so trying to come up with better typing to satisfy the new explicit boolean linter rule and also remove the `isTruthy`/`isFalsy` methods @gabrocheleau introduced for us to use as a fallback.  I'm on the fence about whether I like the "improvements."  Since alot of our booleans were previously using the existence of an object in a check as below:
```js
if (someObject.Property) {
// do something
}
```
or 
```js
if (!someObjectOrVariable) {
/// do something
}
```
we basically have to check if something is undefined or not, and according to multiple stackOverflow questions, the only correct way to do this is to make our code look like this:
```js
if (typeof someObject.Property !== 'undefined') {
// do something
}
// or
if (typeof someOjbectOrVariable === 'undefined`) {
// do something
}
```

What do y'all think?

### DIFF
--- a/packages/evm/src/evm.ts
+++ b/packages/evm/src/evm.ts
@@ -548,7 +548,7 @@ export class EVM extends AsyncEventEmitter<EVMEvents> implements EVMInterface {
           if (
             EOF.validOpcodes(
               result.returnValue.slice(codeStart, codeStart + eof1CodeAnalysisResults.code)
-            ) !== false
+            ) === false
           ) {
             result = {
               ...result,
--- a/packages/evm/src/evm.ts
+++ b/packages/evm/src/evm.ts
@@ -548,9 +548,9 @@ export class EVM extends AsyncEventEmitter<EVMEvents> implements EVMInterface {
           // index 7 (if no data section is present) or index 10 (if a data section is present)
           // in the bytecode of the contract
           if (
-            !EOF.validOpcodes(
+            EOF.validOpcodes(
               result.returnValue.slice(codeStart, codeStart + eof1CodeAnalysisResults.code)
-            )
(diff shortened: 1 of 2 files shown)


jochem-brouwer:
We should not check if certain properties are the string `"undefined"`, right, we should check if they are `undefined`. I see you used `typeof` in the commits though (so in that case we check undefined)? 

Can you link this stack overflow question?

acolytec3:
> We should not check if certain properties are the string `"undefined"`, right, we should check if they are `undefined`. I see you used `typeof` in the commits though (so in that case we check undefined)?
> 
> Can you link this stack overflow question?

There are numerous but [this one](https://stackoverflow.com/questions/27509/detecting-an-undefined-object-property) makes the most sense to me.   The argument for using `typeof someVar === 'undefined' that we're explicitly testing the type of `someVar` (though I think that's arguably the same as `if (someVar === undefined)`.  This line `Note: this last method is the only way to refer to an undeclared identifier without an early error` from the article is what motivated me to use the `typeof someVar...` pattern since it guards against trying to access an undefined and undeclared identifier that would result in an error if that identifier is referenced directly.

holgerd77:
I guess using some code like `typeof someObject.Property !== 'undefined'` as a standard in our code base is a total no go, let's keep this a bit in discussion, eventually we need to adopt the linting rules again respectively evolve.

acolytec3:
Okay, I'm gonna close for now so we don't clutter up our pull request log but we can revisit when needed.

Issue #2032, opened 2022-07-11
Releases tracking issue


holgerd77:
This issue tracks problems of consumers of our beta-releases libraries which we should fix.

**General**
- [ ] Some error messages check `expected === actual` but do not report what the `expected` and `actual` values are (handy for debugging)

**Block**
- [x] Should add a `BlockOption` to allow consensus format validation to be skipped as there may be a case where you want to construct a block that doesn't conform to currently defined consensus rules defined in `common` - done #2139

**Common**
- ~~[ ] It should be possible to query what ChainIDs we supporrt: there should be a map `ChainId: number -> CustomChain` so that one can query whatever custom chain is meant by what chainId and then pass this enum in the `Common.custom` method.~~  Established to not be an issue since the existing `Chain` enum can be indexed on a `chainId` to validate if a chain is supported or not.
- [x] Merge total terminal difficulty is `td` in the lib, should be `ttd` #2075

**EVM**

- [x] Verify that examples in README are correct
- [x] `@types/async-eventemitter` is listed as `devDependency`: this should be a `dependency`, otherwise consumers cannot access the `on` property of EVM. #2077 
- [x] Async eventemitter type when using an `async` function to listen to events is not right (second argument could be undefined, and the function needs an argument when called).

**VM**

- [x] To create EEI, one does `import { EEI } from "@ethereumjs/vm"; const eei = new EEI.EEI(...);`, this is ugly and it should be `new EEI`, probably caused by `export * as EEI from './eei/eei'` should probably be `export * from './eei/eei'` -- in process with #2078
- [x] `@types/async-eventemitter` is listed as `devDependency`: this should be a `dependency`, otherwise consumers cannot access the `on` property of VM. #2077 
- [x] When creating the VM, the `evm` property is cast as `EVMInterface`. If VM is created without injecting the EVM, it thus creates `EVM`, which should also be cast as such. It would also be nice that if the EVM is injected, it automatically infers the returned type (so if you create our EVM then it should also infer it as EVM)
- [x] It seems that `BlockchainInterface` and `Blockchain` type are mixed internally (it is hard to import a custom `BlockchainInterface` -> `_init()` for instance does not exist -- in process with #2069 
[...]

PR #2057, opened 2022-07-19
blockchain: add documentation for custom consensus

acolytec3:
Adds some new verbiage and links around custom consensus.  Fixes one item from #2032


holgerd77 (approved):
LGTM, will admin-merge (only docs).

PR #2059, opened 2022-07-20
Add `EthersStateManager`

### LINKED ISSUE #2019: VM: Add EthersStateManager
It really get's on my nerve that it is so so hard to run the VM in a "real world" state context, so e.g. I just wanted to add a new more fitting VM (so: not EVM) example to the README code along the #2016 work - by using `VM.runBlock()` or `VM.runTransaction()` - and it is just so hard to do due to all the (pre-)state context missing. One is also re-realizing this by looking at the respective test setups (where there is just no "simple" runBlock() or runTx() example).

Maybe we should really rather sooner than later add a new `EthersStateManager`, which just takes in an Ethers instance using one of the providers (Infura, Alchemy, whatever) and which then allows to just read state from an external source an allows for such use cases mentioned above.

This should really be not such a hard task now having our StateManager fully extracted and also with a nicely extendable BaseStateManager one can build upon.

I would even be a fan of directly integrating into the `@ethereumjs/statemanger` package. This would be such a super-useful addition where people will have myriads of use cases (and which would additionally also be a good selling point for the breaking releases 😋).

Ethers is such a tiny package (and we have Richard on the team) and I would say that it wouldn't hurt here to add this as an additional dependency. We can even go one step further and think if we rather make this a "standard" package which is "allowed" to be used in various packages (in a synced version number) if useful, similarly to the `ethereum-cryptography` library e.g.). Then we will likely be better able to integrate in general - e.g. also on use cases like loading an Ethers tx in our tx library - see #1099 - which would improve interoperability between these ecosystems, guess users will likely benefit since there are lots of use cases around that (especially in this field: reading RPC data with Ethers, continue to use/process with EthereumJS libs).

So to come back: maybe this is really some nice holiday task for someone. Again, nothing pressing, but likely also not too difficult to realize.

acolytec3:
Fixes #2019

This is preliminary but takes the first step towards an EthersStateManager. 

- [x] Use `@ethersproject/provider` to source data for `getAccount`/`getContractCode`/`getContractStorage`
- [x] Add support for setting the block number to reference when retrieving state from the provider
- [x] Leverage `getProof` (where available from the provider) to validate account existence
- [x] Add support for getStateRoot methods
- [x] Handle `getContractStorage` correctly
- [x] Add more integration tests with `runBlock`/`runTx` to verify that data sourced from the Ethers provider is being handled correctly
- [x] Add some tests that recreate a transaction and block from Mainnet using data sourced from Ethers provider
- [x] Add better unit tests with truly fake provider rather than relying on cloudflare
- [x] Fix browser tests (I fixed them by skipping all the API tests :grimacing: )
- [x] Validate that `putContractStorage` works right


acolytec3:
I'm sure there's more work to be done but this should be a good first step.  Will work on adding some more tests to try and recreate real transactions and blocks using data sourced from mainnet.

acolytec3:
I'm not sure why the karma tests are failing but will look at that once I get some comprehensive tests.

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,103 @@
...
+export class EthersStateManager extends BaseStateManager implements StateManager {
+  private provider: JsonRpcProvider
+  private contractCache: Map<string, Buffer>
+  private storageCache: Map<string, string>
+
+  constructor(opts: EthersStateManagerOpts) {
+    super({ common: opts.common })
+    this.provider = opts.provider
+    this.contractCache = new Map()
+    this.storageCache = new Map()
```
g11tech, line 24: should we do this for basestatemanager downstream?

then the task will only come down to just cache load/flush.

acolytec3: I haven't thought that deeply on it yet.  I need to look more into the underlying functionality of the `baseStateManager` class versus our existing `DefaultStateManager` and see how it works.  The reason I'm storing them in a cache in this instance is because this state manager is "readonly" and we don't persist results that occur through VM execution beyond the immediate context of the code this is running in and I didn't want any contracts created/storage slots changed to get lost in the immediate context since the `DefaultStateManager` is storing all the updates directly to the various tries.

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+    return true
+  }
+
+  async getAccount(address: Address): Promise<Account> {
+    const account = await this._cache.getOrLoad(address)
+    return account
+  }
+  async getAccountFromProvider(address: Address): Promise<Account> {
+    const balance = await this.provider.getBalance(address.toString())
```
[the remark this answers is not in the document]
acolytec3: Correct, there is no API call for that in the JsonRPC so therefore `ethers` doesn't have one either.

acolytec3:
I'm going to pause work on this.  We now have a reasonably complete proof of concept where a user can leverage any `ethers.jsonRpcProvider` to source data and run at least simple tests with the vm without data preconditioning.  My initial tests are just a simple `runTx` with a transfer of some ETH to Vitalik and then running a block that has no transactions in it from the ancient chain history, but I think this demonstrates the concept is sound.

Next steps on this are to decide if we want to integrate this as an alternative state manager implementation exposed by our `statemanager` class or just split it out as a standalone representation of what's possible.

If we want to integrate it in `statemanager`, I will make the tests truly unit tests with mocked responses from the provider and add some better integration tests using the cloudflareprovider (which is a subclass of `ethers.jsonrpcprovider`) so we can get a sense for how it works.

@holgerd77 and @ethereumjs/committers, over to you for opinions at this point.

packages/statemanager/tests/ethersStateManager.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+import { CloudflareProvider, JsonRpcProvider } from '@ethersproject/providers'
+import * as tape from 'tape'
+import { blockFromRpc } from '@ethereumjs/block/dist/from-rpc'
+import { VM } from '@ethereumjs/vm'
+import { Chain, Common, Hardfork } from '@ethereumjs/common'
+import { FeeMarketEIP1559Transaction } from '@ethereumjs/tx'
+
+import { EthersStateManager } from '../src/ethersStateManager'
+
+const provider = new CloudflareProvider()
```
holgerd77, line 11: Hmm, this is doing real requests on the Ethers standard cloudflare account on our CI, is this correct? Not sure if this is a good idea, Richard is always concerned about over-use.

acolytec3: Oh really? I thought this was the public cloudflare gateway? These tests are only proof on concept to see that it works anyway. I would switch to mocked responses in a final merged version of we go with this approach.

holgerd77:
Great stuff, cool! ❤️

Just a general note since I stumbled upon this now a couple of times already: we might want to split our StateManager interfaces in a slightly different/evolved way and do a StatelessStateManager interface and then a full (Statefull)StateManager interface, with the first one taking out the stateful operations.

This needs some expansion respectively some deeper thinking I guess.

holgerd77:
(updated the last post on site)

packages/statemanager/tests/ethersStateManager.spec.ts
holgerd77: Or to phrase it differently: we generally should make sure that we do not leech on public Ethers API endpoints.

acolytec3: [We're good] (https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/cloudflare-provider.ts#L23). I looked at the code behind the cloudflare provider and it's calling the main public gateway and not associated with an ethers api key.

holgerd77: 👍

packages/statemanager/tests/ethersStateManager.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+  })
+  const vm = await VM.create({ common, stateManager: state })
+  const blockData = await provider.send('eth_getBlockByNumber', [bigIntToHex(blockTag), true])
+  const block = blockFromRpc(blockData, undefined, { common, hardforkByBlockNumber: true})
+  const res = await vm.runBlock({
+    block,
+    skipHeaderValidation: true,
+    skipBlockValidation: true,
+    skipBalance: true,
+    skipNonce: true,
```
holgerd77, line 109: Just for my understanding, `skipBalance` and `skipNonce` is still needed in this get-the-state-from-a-central-source scenario?

acolytec3: I removed those from the test and you actually don't need them in this scenario.  I think you would need those if you were constructing a custom transaction and didn't want to deal with balance and nonce checks on the caller account but they pass just fine if you set up the preconditions correctly.  In the case of just trying to run a block from mainnet, the preconditions you need are to set up the state manager to make it's calls against the state at the [block number - 1](https://github.com/ethereumjs/ethereumjs-monorepo/blob/0b7f966ff570160b40518a2b9db0a86297c45a17/packages/statemanager/tests/ethersStateManager.spec.ts#L98) and then retrieve the block and try and run it.  if the state manager is pulling valid state from the provider, everything should pass and the block state root you get from the VM should match what's in the block pulled from the provider.  If you point the state manager to the wrong block height, you may get nonce validation failures (for example) if you run a block with a transaction from a calling account in it and the state manager is looking pulling state from the chain at some other height.
In this precise case, you still need `skipHeaderValidation` because the parent block isn't in the vm's blockchain so it returns header not found.  We could make a separate blockchain implementation if we wanted to pull the headers from the provider as well.

There's still some nuance to conditioning the tests, but really just setting up which validation rules you want to use.

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,162 @@
...
+    log(`getting account data for ${address.toString()}`)
+    let accountData
+    try {
+      accountData = await this.provider.send('eth_getProof', [
+        address.toString(),
+        [],
+        this.blockTag,
+      ])
+    } catch (e) {
+      accountData.balance = await this.provider.getBalance(address.toString(), this.blockTag)
```
jochem-brouwer, line 113: I think we should remove this catch block and throw entirely instead, since we have no way to get `storageRoot` for an account over RPC.

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,163 @@
...
+      const dataBuffer = Buffer.from(proofItem.slice(2), 'hex')
+      const hash = keccak256(dataBuffer)
+      await (this.trie as any)._db.put(arrToBufArr(hash), dataBuffer)
+    }
+
+    const account = Account.fromAccountData({
+      balance: BigInt(accountData.balance),
+      nonce: BigInt(accountData.nonce),
+      codeHash: toBuffer(accountData.codeHash),
+      storageRoot: await this.getStateRoot(),
```
jochem-brouwer, line 131: This is not right, the stateRoot of the account should be taken from the accountData. The state root of the account is the root of the underlying trie DB which are the key/value pairs of the storage.

jochem-brouwer, line 131:
```suggestion
      storageRoot: accountData.storageHash
```

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,253 @@
...
+    const accountData = await this.provider.send('eth_getProof', [
+      address.toString(),
+      [],
+      this.blockTag,
+    ])
+    const rawData = accountData.accountProof
+
+    await this.trie.fromProof(rawData.map((e: string) => hexToBytes(e)))
+
+    const account = Account.fromAccountData({
```
jochem-brouwer, line 142: Does this work if we have 2 txs from the same account in the block? Does it save the account in cache, or does it get this from the external provider again and thus overrides the earlier operation?

acolytec3: It should never be called twice for the same account.  `getAccountFromProvider` is called within `getAccount` which checks the cache first and only calls out to the provider if the account isn't found.

acolytec3:
Alright all, I think this is ready for a semi-final round of review.  I'd like for folks to build and run it locally and see if you're satisfied with what it offers or let me know if there are things we should change.  At the moment, it supports retrieving state from the provider and running "most" blocks, unless you happen to be one of those edge cases where a storage slot is zeroed or an account self-destruct occurs and that results in a branch node having only a single leaf node that gets deleted (which also means the branch node gets deleted). In this edge case, we don't have a way via the web3Provider interface to retrieve the surrounding inner nodes we need to update the state trie so we wouldn't be able to compute the proper state root for those blocks.  Beyond that edge case, it seems to work pretty reliably as far as running blocks, running transactions, etc.

One thing I haven't done yet is implemente a "reset trie" method which would let you completely reset the trie, start the provider on a different block, and then start running things again, but that should be trivial.  

Look at the tests I've written and see if that covers all the ground you'd like to cover in web3provider driven State Manager and if it looks good, I'll rework the tests with mocked provider responses so we have proper unit tests and then this might be about merge ready.

packages/block/src/from-rpc.ts
```diff
@@ -28,8 +30,7 @@ function normalizeTxParams(_txParams: any) {
   // v as raw signature value {0,1}
   // v is the recovery bit and can be either {0,1} or {27,28}.
   // https://ethereum.stackexchange.com/questions/40679/why-the-value-of-v-is-always-either-27-11011-or-28-11100
-  const v: number = txParams.v
-  txParams.v = v < 27 ? v + 27 : v
+  txParams.v = toType(txParams.v, TypeOutput.BigInt)!
```
gabrocheleau, line 33: Does this actually work? Why can we now remove the `v < 27 ? v + 27 : v` logic?

packages/block/src/from-rpc.ts
```diff
@@ -49,11 +50,13 @@ export function blockFromRpc(
...
 
   const transactions: TypedTransaction[] = []
-  const opts = { common: header._common }
-  for (const _txParams of blockParams.transactions) {
-    const txParams = normalizeTxParams(_txParams)
-    const tx = TransactionFactory.fromTxData(txParams as TxData, opts)
-    transactions.push(tx)
+  if (blockParams.transactions !== undefined) {
+    const opts = { common: header._common }
+    for (const _txParams of blockParams.transactions) {
```
gabrocheleau, line 55:
```suggestion
    const opts = { common: header._common }
    for (const _txParams of blockParams.transactions ?? []) {
```
not a big deal, but one less line of code

packages/block/src/from-rpc.ts
gabrocheleau: I see you've updated the test data so that `v`s are now >=27. Does that fit with what we'd expect in a "real" context?

packages/block/test/testdata/testdata-from-rpc.json
```diff
@@ -1,5 +1,4 @@
 {
-  "author": "0x4bb96091ee9d802ed039c4d1a5f6216f90f81b01",
```
gabrocheleau, line 2: We also have those non-standard fields (e.g. author, creates, etc.) in other `from-rpc` testdata JSON files. I have no insight into why these are there, but from what I've read they are part of the Ethereum JSON-RPC standard (see https://ethereum.github.io/execution-apis/api-documentation/) 

image

Could we completely remove them? If so, we could as remove the `as unknown as JsonRpcBlock` typecastings from the `from-rpc` test file.

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,400 @@
...
+  blockTag?: bigint | string
+}
+
+export class EthersStateManager extends BaseStateManager implements StateManager {
+  private provider: JsonRpcProvider
+  private contractCache: Map<string, Buffer>
+  private storageTries: { [key: string]: Trie }
+  private externallyRetrievedStorageKeys: Map<string, Map<string, boolean>>
+  private blockTag: string
+  private root: Buffer | undefined
```
gabrocheleau, line 39: `root` is unusued

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,400 @@
...
+import type { StorageDump } from './interface'
+import type { StorageProof } from './stateManager'
+import type { Common } from '@ethereumjs/common'
+import type { Address, PrefixedHexString } from '@ethereumjs/util'
+import type { JsonRpcProvider } from '@ethersproject/providers'
+
+const log = debug('statemanager')
+export interface EthersStateManagerOpts {
+  provider: JsonRpcProvider
+  blockTag?: bigint | string
```
gabrocheleau, line 30: Should we change `string` to the literal `latest`| `earliest` | `pending` (or whichever ones we support) and then validate against those in the constructor (perhaps with an `unimplemented` error message for valid parameters that we don't yet support like we do for some RPC calls in the client)

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,400 @@
...
+   */
+  async putContractCode(address: Address, value: Buffer): Promise<void> {
+    // Store contract code in the cache
+    this.contractCache.set(address.toString(), value)
+  }
+
+  /**
+   * Gets the storage value associated with the provided `address` and `key`. This method returns
+   * the shortest representation of the stored value.
+   * @param address -  Address of the account to get the storage for
```
gabrocheleau, line 98:
```suggestion
   * @param address - Address of the account to get the storage for
```

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,400 @@
...
+   * @returns {Buffer} - The storage value for the account
+   * corresponding to the provided address at the provided key.
+   * If this does not exist an empty `Buffer` is returned.
+   */
+  async getContractStorage(address: Address, key: Buffer): Promise<Buffer> {
+    // Retrieve storage slot from provider if not found in cache
+    await this.getContractStorageFromProvider(address, key)
+
+    const storageTrie = await this._getStorageTrie(address)
+    const foundValue = await storageTrie.get(key) //@ts-ignore
```
gabrocheleau, line 109:
```suggestion
    const foundValue = await storageTrie.get(key)
```
Doesn't seem to error locally if I remove this

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,400 @@
...
+    const storageTrie = await this._getStorageTrie(address)
+
+    await storageTrie.fromProof(storageData.proof.map((e: string) => hexToBytes(e)))
+
+    let map = this.externallyRetrievedStorageKeys.get(address.toString())
+    if (!map) {
+      this.externallyRetrievedStorageKeys.set(address.toString(), new Map())
+    }
+    map = this.externallyRetrievedStorageKeys.get(address.toString())!
+    map.set(key.toString('hex'), true)
```
gabrocheleau, line 147:
```suggestion
    let map = this.externallyRetrievedStorageKeys.get(address.toString())
    if (!map) {
      map = this.externallyRetrievedStorageKeys
        .set(address.toString(), new Map())
        .get(address.toString())!
    }
    map.set(key.toString('hex'), true)
```
I think this works and avoids `get`ting twice when map is retrieved from the first `get`

gabrocheleau: Actually this is even simpler
```suggestion
     const map =
      this.externallyRetrievedStorageKeys.get(address.toString()) ??
      this.externallyRetrievedStorageKeys
        .set(address.toString(), new Map())
        .get(address.toString())!
    map.set(key.toString('hex'), true)
```

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,400 @@
...
+  }
+
+  /**
+   * Adds value to the state trie for the `account`
+   * corresponding to `address` at the provided `key`.
+   * @param address -  Address to set a storage value for
```
gabrocheleau, line 153:
```suggestion
   * @param address - Address to set a storage value for
```
can't believe my eyes are spotting these

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,400 @@
...
+   * @param key - Key to set the value at. Must be 32 bytes long.
+   * @param value - Value to set at `key` for account corresponding to `address`. Cannot be more than 32 bytes. Leading zeros are stripped. If it is a empty or filled with zeros, deletes the value.
```
gabrocheleau, line 155:
```suggestion
   * @param value - Value to set at `key` for account corresponding to `address`. 
   Cannot be more than 32 bytes. Leading zeros are stripped. 
   If it is empty or filled with zeros, deletes the value.
```

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,400 @@
...
+        })
+        .catch((e) => {
+          reject(e)
+        })
+    })
+  }
+
+  /**
+   * Checks if the `account` corresponding to `address`
+   * exists
```
gabrocheleau, line 213:
```suggestion
   * Checks if an `account` exists at `address`
   ```

File: packages/statemanager/src/ethersStateManager.ts [hunk 19]
```diff
@@ -0,0 +1,400 @@
...
+      // if not verified (i.e. verifyProof returns null), account does not exist
+      if (verified === null) return false
+    }
+    return true
+  }
+
+  /**
+   * Gets the code corresponding to the provided `address`.
+   * @param address - Address to get the `code` for
+   * @returns {Promise<Buffer>} -  Resolves with the code corresponding to the provided address.
```
gabrocheleau, line 236:
```suggestion
   * @returns {Promise<Buffer>} - Resolves with the code corresponding to the provided address.
```

gabrocheleau (commented):
Reviewed most of the logic in ethersStateManager, have to run off, will resume the review early next week.

packages/block/src/from-rpc.ts
acolytec3: I can look at the responses again but Infura/Alchemy appeared to be returning 27 or 28 for the old transaction type and it was breaking at this point.

packages/statemanager/src/ethersStateManager.ts
acolytec3: We certainly can and there shouldn't be any thing "unimplemented" here since we should expect the provider to support the RPC spec, right?

packages/statemanager/src/ethersStateManager.ts
acolytec3: Hmm, at the cost of one additional boolean, I'd prefer just moving the second get inside the `!map` check.  I know we can chain all that stuff together but it just makes it less legible to me and I don't think any material difference in code executed, right?

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,400 @@
...
+    log(`Verify if ${address.toString()} exists`)
+    const account = this._cache.get(address)
+    if (account.isEmpty()) {
+      // Get latest block (or block specified in `this.blockTag`)
+      const block = await this.provider.send('eth_getBlockByNumber', [this.blockTag, false])
+      // Get merkle proof for `address` from provider
+      const proof = await this.provider.send('eth_getProof', [address.toString(), [], block.number])
+      const proofBuf = proof.accountProof.map((proofNode: string) => toBuffer(proofNode))
+      const trie = new Trie({ useKeyHashing: true })
+      const verified = await trie.verifyProof(toBuffer(block.stateRoot), address.buf, proofBuf)
```
acolytec3, line 226: @jochem-brouwer @g11tech  Do we need the block's `stateRoot` here or can we if we have already requested the proof can we just hash the first node and use that as the state root in the `verifyProof` call?  Seems like the latter should be correct, right?

packages/statemanager/src/ethersStateManager.ts
gabrocheleau: Yeah you're right, think I wrote this before understanding the full picture of the ethersStateManager

packages/statemanager/src/ethersStateManager.ts
gabrocheleau: Yes, sure! No difference indeed, let's go with what makes it most legible

packages/block/test/testdata/testdata-from-rpc.json
acolytec3: I think we could.  Only the ones you have expanded there are in the standard responses I see when interacting with providers.  Should I go ahead and do it in this PR?

packages/statemanager/src/ethersStateManager.ts
g11tech: yes, that should give you the stateroot if you don't want to verify/double check

acolytec3:
Alright team, this one is now ready for a first/final round of review.  It does the following:
1) Fixes the `fromRpc` methods and test data in `block` to work with our current code base and the data provided by `ethers` providers
2) Adds a new `EthersStateManager` class that implements the `StateManager` interface.  
- This state manager can execute transactions and blocks in the `vm` package using data sourced from any `ethers` `JsonRpcProvider` - works best with archive nodes so ideally alchemy or infura (though will work with CloudFlare as long as you aren't looking at ancient chain history).
- Adds a new `mockProvider` class for use in unit testing the new state manager

cc: @ricmoo would be happy for any additional input you have on this.

Thanks to @jochem-brouwer for many hours of help troubleshooting the work to allow us to run blocks (or at least most blocks) with data only sourced from a provider (and not devp2p)

packages/statemanager/src/ethersStateManager.ts
g11tech: i think two txs refering the same account should work, because first time it fetches and writes to cache, and if there are modifications, that are written to the cache, so for the next tx, it will correctly get the modified account

packages/statemanager/tests/testdata/providerData/mockProvider.ts
```diff
@@ -0,0 +1,36 @@
...
+      case 'eth_getCode':
+        return 0
+      default:
+        throw new Error(`method ${method} not implemented`)
+    }
+  }
+
+  private getProofValues = async (params: [address: string, _: [], blockTag: bigint | string]) => {
+    const [address, _, blockTag] = params
+    const account = await import(`./accounts/${address}.json`)
```
g11tech, line 21: ohh this is where you needed to bundle all the jsons in the dist :thinking:

g11tech (commented):
looks really good :heart: :fire: :rocket: , will try most probably by tomorrow and update

packages/block/src/from-rpc.ts
gabrocheleau: Taking a random transaction (`0x4e9c81762459fd1ecec67d632dbe3fdb0accb33608e300246547692db4087484`) from a recent block on an alchemy node, I'm getting:

``` 
{
	"jsonrpc": "2.0",
	"id": 67,
	"result": {
		"blockHash": "0xa8a73a4d76924e8e6bb32ea6f7416c92abab6fbca701ffeb5e0ccf93915be258",
		"blockNumber": "0xec6807",
		"from": "0x1bc851490add02632f2b41c87a0b8dc3cba4cf0f",
		"gas": "0x4daef",
		"gasPrice": "0x2fa3be745",
		"maxFeePerGas": "0x50f53742f",
		"maxPriorityFeePerGas": "0x77359400",
		"hash": "0x4e9c81762459fd1ecec67d632dbe3fdb0accb33608e300246547692db4087484",
		"input": "0x5f5755290000000000000000000000000000000000000000000000000000000000000080000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb480000000000000000000000000000000000000000000000000000000011ad50f400000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000000000136f6e65496e6368563446656544796e616d69630000000000000000000000000000000000000000000000000000000000000000000000000000000000000007c0000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb4800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000011ad50f400000000000000000000000000000000000000000000000002788ce887bbefe200000000000000000000000000000000000000000000000000000000000001200000000000000000000000000000000000000000000000000005a3dbe52e10d5000000000000000000000000f326e4de8f66a0bdc0970b79e0924e33c79f1915000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000006887c025200000000000000000000000000f2f400c138f9fb900576263af0bc7fcde2b1b8a800000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000180000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48000000000000000000000000eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee000000000000000000000000f2f400c138f9fb900576263af0bc7fcde2b1b8a800000000000000000000000074de5d4fcbf63e00296fd95d33236b97940166310000000000000000000000000000000000000000000000000000000011ad50f4000000000000000000000000000000000000000000000000027e22542580950700000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004e0000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000030000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000016000000000000000000000000000000000000000000000000000000000000002808000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000064eb5625d9000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48000000000000000000000000e68c1d72340aeefe5be76eda63ae2f4bc75141100000000000000000000000000000000000000000000000000000000011ad50f400000000000000000000000000000000000000000000000000000000800000000000000000000000e68c1d72340aeefe5be76eda63ae2f4bc7514110000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000084fe029156000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb4800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000011ad50f4000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000018414284aab00000000000000000000000000000000000000000000000000000000000000808000000000000000000000000000000000000000000000000000000000000044000000000000000000000000eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee00000000000000000000000000000001000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000064d1660f99000000000000000000000000eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee0000000000000000000000001111111254fb6c44bac0bed2854e76f90643097d00000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000ab4991fe000000000000000000000000000000000000000000000000cc",
		"nonce": "0xae",
		"to": "0x881d40237659c251811cec9c364ef91dc08d300c",
		"transactionIndex": "0xb",
		"value": "0x0",
		"type": "0x2",
		"accessList": [],
		"chainId": "0x1",
		"v": "0x0",
		"r": "0x585a5019401700c284a1e73cb617a1a7b3c7c0f9f30bcf7a3888b49ac8cda983",
		"s": "0x72b03f87e10a1ee05ec90bbcafb35fab253bc918384482b0864e0df53956b4de"
	}
}
```

acolytec3: Hmm, maybe I was imagining things then. Let me dig into it again. Now I'm wondering if the 27/28 was only for legacy transactions.

holgerd77: Yes, v 27/28 is only legacy.

g11tech:
Wow! really amazing PR! :heart: :fire: :rocket:  

successfully able to hookup the jsonrpc to my local mainnet node and able to run this block successfully: https://etherscan.io/block/15496421

image
Feels really close

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,420 @@
...
+  private provider: JsonRpcProvider
+  private contractCache: Map<string, Buffer>
+  private storageTries: { [key: string]: Trie }
+  private externallyRetrievedStorageKeys: Map<string, Map<string, boolean>>
+  private blockTag: string
+  private trie: Trie
+
+  constructor(opts: EthersStateManagerOpts) {
+    super({})
+    this.trie = new Trie({ useKeyHashing: true })
```
g11tech, line 43:
```suggestion
    // useKeyHashing = true since the web3 api provides proof nodes which are hashed
    // If there were direct api access to devp2p stack, a normal Trie could have been constructed
    this.trie = new Trie({ useKeyHashing: true })
```

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,420 @@
...
+export interface EthersStateManagerOpts {
+  provider: string | JsonRpcProvider
+  blockTag?: bigint | 'latest' | 'earliest'
+}
+
+export class EthersStateManager extends BaseStateManager implements StateManager {
+  private provider: JsonRpcProvider
+  private contractCache: Map<string, Buffer>
+  private storageTries: { [key: string]: Trie }
+  private externallyRetrievedStorageKeys: Map<string, Map<string, boolean>>
```
g11tech, line 37: can we add a nice comment detailing the "structure" of this map (what it refers to), its not really obvious

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,420 @@
...
+import type { Proof, StateManager } from '.'
+import type { StorageDump } from './interface'
+import type { StorageProof } from './stateManager'
+import type { Common } from '@ethereumjs/common'
+import type { Address, PrefixedHexString } from '@ethereumjs/util'
+
+const log = debug('statemanager')
+export interface EthersStateManagerOpts {
+  provider: string | JsonRpcProvider
+  blockTag?: bigint | 'latest' | 'earliest'
```
g11tech, line 30: rather than having block number/tag as part of constructor opts, can we have a `setBlock` fn that does the job, in this way it can be used recurringly

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,420 @@
...
+      throw new Error(`valid JsonRpcProvider or url required; got ${opts.provider}`)
+    }
+
+    if (typeof opts.blockTag === 'bigint') {
+      this.blockTag = bigIntToHex(opts.blockTag)
+    } else {
+      this.blockTag = opts.blockTag ?? 'latest'
+    }
+
+    this.contractCache = new Map()
```
g11tech, line 59: if we have a separate `init` fn, this allocation could become a part of it, and we can call it here directly

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,420 @@
...
+   * @private
+   */
+  async getAccountFromProvider(address: Address): Promise<Account> {
+    const accountData = await this.provider.send('eth_getProof', [
+      address.toString(),
+      [],
+      this.blockTag,
+    ])
+
+    const rawData = accountData.accountProof
```
g11tech, line 267: should the account proof be verified here?

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,420 @@
...
+    return this.trie.root()
+  }
+
+  /**
+   * Sets the state of the instance to that represented
+   * by the provided `stateRoot`.
+   * @param stateRoot - The state-root to reset the instance to
+   */
+  async setStateRoot(stateRoot: Buffer): Promise<void> {
+    this._cache.flush
```
g11tech, line 305:
```suggestion
    this._cache.flush()
```

packages/statemanager/src/ethersStateManager.ts
acolytec3: I think I'll keep it as optional so people can set it once and forget it if they are only running one call and then add a second initalization method to allow you to set the blockTag (which in turn will reset the local state/storage cache.

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,420 @@
...
+  }
+
+  /**
+   * Dumps the RLP-encoded storage values for an `account` specified by `address`.
+   * @param address - The address of the `account` to return storage for
+   * @returns {Promise<StorageDump>} - The state of the account as an `Object` map.
+   * Keys are are the storage keys, values are the storage values as strings.
+   * Both are represented as `0x` prefixed hex strings.
+   */
+  dumpStorage(address: Address): Promise<StorageDump> {
```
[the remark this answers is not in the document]
acolytec3: I think I'll leave in for now as discussed on discord.  It's part of the interface so I don't see any reason not to support it.

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,420 @@
...
+    } else {
+      throw new Error(
+        `expected blockTag to be block hash, bigint, or hex prefixed string; got ${blockTag}`
+      )
+    }
+
+    const uncleHeaders = []
+    if (blockData.uncles.length > 0) {
+      for (let x = 0; x < blockData.uncles.length; x++) {
+        const headerData = await this.provider.send('eth_getUncleByBlockHashAndIndex', [
```
g11tech, line 411: i think this is not really required to be fetched for the purpose of the statemanager exec of the block.

packages/statemanager/src/ethersStateManager.ts
acolytec3: I don't think so.  If the account doesn't exist in the state trie, the state manager should create a virtual account locally and the response from the provider is just an empty account (0 balance, 0 nonce, empty code/storage hash) with a proof of non existence.  We don't care if it doesn't already exist.  It would still be nice to have those proof nodes in the local state trie.

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,420 @@
...
+    })
+  }
+
+  /**
+   * Checks if an `account` exists at `address`
+   * @param address - Address of the `account` to check
+   */
+  async accountExists(address: Address): Promise<boolean> {
+    log(`Verify if ${address.toString()} exists`)
+    const account = this._cache.get(address)
```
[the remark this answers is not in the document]
acolytec3: I don't think so.  As above, `getAccount` is fulfilling a different purpose.  

@jochem-brouwer any other thoughts here?

packages/statemanager/src/ethersStateManager.ts
acolytec3: You have to compute the correct uncle hash as part of the process of computing the final block hash.  I was getting an incorrect block hash without this piece.

packages/statemanager/src/ethersStateManager.ts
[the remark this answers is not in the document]
g11tech: you you seem to be doing even much more than getaccount here (and below) whats the difference?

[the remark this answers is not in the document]
acolytec3: `accountExists` and `getAccount` are fulfilling totally different use cases as far as I understand.

`accountExists` is validating if an account currently exists in the state trie.  `getAcccount` returns an account's current state.  If that account doesn't have an entry in the state trie, a new `Account` object with zero nonce/balance is added to the cache and then returned by the method.

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,451 @@
...
+
+    const rawData = accountData.accountProof
+
+    await this.trie.fromProof(rawData.map((e: string) => hexToBytes(e)))
+
+    const account = Account.fromAccountData({
+      balance: BigInt(accountData.balance),
+      nonce: BigInt(accountData.nonce),
+      codeHash: toBuffer(accountData.codeHash),
+      storageRoot: accountData.storageHash,
```
gabrocheleau, line 307: Shouldn't we `toBuffer` this too? 
```suggestion
      storageRoot: toBuffer(accountData.storageHash),
```

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,451 @@
...
+    return returnValue
+  }
+
+  /**
+   * Helper method to retrieve a block from the provider to use in the VM
+   * @param blockTag block hash or block number to be run
+   * @param common Common instance used in VM
+   * @returns the block specified by `blockTag`
+   */
+  getBlockFromProvider = async (blockTag: string | bigint, common: Common) => {
```
gabrocheleau, line 394: Could we also add the `latest`, `earliest`, `pending`, cases here?

gabrocheleau (commented):
Looks good, reviewed everything except the tests (will try to do those later) and left a couple comments

packages/statemanager/tests/util.ts
```diff
@@ -1,3 +1,4 @@
+//@ts-nocheck
```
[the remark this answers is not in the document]
acolytec3: Nope.  Just a miss.  Will remove on next commit

packages/statemanager/src/ethersStateManager.ts
acolytec3: I believe you're right.  Will add an additional check for those cases.

packages/statemanager/src/ethersStateManager.ts
acolytec3: Yes we should.

packages/statemanager/src/ethersStateManager.ts
g11tech: what will ethers api return if address doesn't exist in the state? and do you mean to say the rawData proof that you got will already be validated in `await this.trie.fromProof(rawData.map((e: string) => hexToBytes(e)))`?

acolytec3: Infura returns the same proof object with the account balance/nonce being 0 and the proof being one of non existence

gabrocheleau:
General comment that the new deps increase the npm audit issues, 
before:
Screenshot from 2022-09-09 09-13-43

after:
Screenshot from 2022-09-09 09-15-05

the additional high severity issue is used in production:

Screenshot from 2022-09-09 09-30-18

I checked some of these but am not clear on if they are problematic for us or not.

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,451 @@
...
+import {
+  Account,
+  arrToBufArr,
+  bigIntToHex,
+  bufferToHex,
+  intToHex,
+  isHexPrefixed,
+  toBuffer,
+} from '@ethereumjs/util'
+import { JsonRpcProvider } from '@ethersproject/providers'
```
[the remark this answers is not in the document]
acolytec3: I agree with you in the case where we're instantiating the provider for the user (i.e. when they just pass in an RPC endpoint url).  That said, I want the user to be able to provide any JsonRpc compliant endpoint if they pass in their own provider and the other subclasses (e.g. `CloudflareProvider`, etc) don't inherit from the `StaticJsonRpcProvider` so I'm going to set the type as one of the two in the opts.

packages/block/src/from-rpc.ts
```diff
@@ -28,8 +30,7 @@ function normalizeTxParams(_txParams: any) {
   // v as raw signature value {0,1}
   // v is the recovery bit and can be either {0,1} or {27,28}.
   // https://ethereum.stackexchange.com/questions/40679/why-the-value-of-v-is-always-either-27-11011-or-28-11100
```
gabrocheleau, line 32: We could update this comment in light of: https://ethereum.stackexchange.com/questions/113499/why-is-value-of-ecdsa-v-often-neither-27-nor-28

packages/block/test/testdata/testdata-from-rpc.json
gabrocheleau: Extracted some of your work + additional cases in the PR here: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2281

gabrocheleau (approved):
Left a few minor comments, rest looks good to me!

packages/block/src/from-rpc.ts
acolytec3: So I'm still stumped about this. I've been looking at `tx` trying to understand what `v` should be when it's passed in as part of the `fromTxData` constructor we need to use when pulling transactions from the provider.  If I understand things, v should always be { 0, 1, 27, 28 }.  For legacy transactions, we have this `meets155` helper that checks the to make sure it matches the legacy expectations and for Type 1 and 2 transactions, it should always be 0 or 1. So, to my way of thinking, our "from-rpc" helper that formats the data should just take the value of v that's passed in and let the `tx` constructors error out if something's wrong, right?  I shouldn't need to do any special massaging of data passed in from the RPC endpoint since those should just be formatting helpers to make sure that values that come in as hex strings are converted to bigints or what have you, right?

holgerd77: So there are 3 Ethereum "periods" for the v values:

1. Before [EIP-155](https://eips.ethereum.org/EIPS/eip-155): `v` was `27` or `28` back then (up to `SpuriousDragon` HF, block `2,675,000`).
2. After `EIP-155`: v has now the chain ID baked in according to the `EIP-155` formula `{0,1} + CHAIN_ID * 2 + 35`. So the minimal `v` is `0 + 1*2 +35` which is `37`
3. Along with typed txs introduced with [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) (first concrete one: `EIP-2930` with these access lists introduced in the `Berlin` HF (block `12,244,000`) `v` went back to simply `0` and `1` (so what the signature ECDSA function expects as an input) and is should technically also not called `v` any more but something like `yParity` or something (but still is for RPC for backwards-compatibility reasons)

So: if we keep this code with `v + 27` we are killing off the typed tx v values. If I am not totally overlooking something we should update here, so basically - as Andrew suggested - just letting the `v` values "through".

I think this `v + 27` code was for some very old Geth bug behavior where it returned (I think in the very old per-SpuriousDragon days) `0` or `1` instead of `27` or `28` in RPC (so - attention! - big source of confusion if one stumbles upon such a transaction: values which are *now*  correct but which weren't back in the days).

Since this bug has been fixed a long time ago I think we can safely upgrade here (again: just passing the `v` values through).

gabrocheleau: Great! Thanks for the detailed explanation holger!

packages/statemanager/README.md
```diff
@@ -42,6 +42,28 @@ await stateManager.commit()
...
+
+Refer to [this test script](./tests/ethersStateManager.spec.ts) for complete examples of running transactions and blocks in the `vm` with data sourced from a provider.
+
+**WARNING** When using the `EthersStateManager`, any time an account self destruct occurs or a storage slot is deleted while running transactions and blocks, do not be surprised to see an error due to a fundamental limitation of the `web3provider` interface whereby we are not able to retrieve all of the necessary trie nodes needed to update the parent node of a deleted leaf resulting from the storage slot deletion/account self destruct. Many blocks should execute successfully and report the correct updated state root but it is not possible to predict a head of time if any particular transaction will result in this error since all keys for values in the underlying trie are hashed.
```
gabrocheleau, line 65:
```suggestion
**WARNING** When using the `EthersStateManager`, any time an account self destruct occurs or a storage slot is deleted while running transactions and blocks, do not be surprised to see an error. This is due to a fundamental limitation of the `web3provider` interface whereby we are not able to retrieve all of the necessary trie nodes needed to update the parent node of a deleted leaf resulting from the storage slot deletion/account self destruct. Many blocks should execute successfully and report the correct updated state root. However, it is not possible to predict ahead of time if any particular transaction will result in this error since all keys for values in the underlying trie are hashed.
```
slightly more readable by breaking up the sentence imo 

Concerning this part:
```
However, it is not possible to predict ahead of time if any particular transaction will result in this error since all keys for values in the underlying trie are hashed.
```
Is it actually not possible? My understanding of this is not 100% clear, but doesn't the issue only arise when the first branch-node parent of the deleted node has 2 branches (and therefore has to be converted to a leaf node/extension node when it goes from 2 branches to 1 branch)?

acolytec3:
I think this is ready for final review and/or merging at this point.  I made some last edits to the readme to note the gotchas related to slot deletions/account self-destruct.  It doesn't really get into the nuts & bolts of how it works but hopefully it at least alerts folks to the potential outcomes.

packages/statemanager/tests/ethersStateManager.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+    const tx = FeeMarketEIP1559Transaction.fromTxData(normedTx, { common })
+    const res = await vm.runTx({ tx })
+    t.equal(res.totalGasSpent, 21000n, 'calculated correct total gas spent for simple transfer')
+    t.end()
+  }
+})
+
+/** To run the block test with an actual provider, you will need a provider URL (like alchemy or infura) that provides access to an archive node.
+ *  Pass it in the PROVIDER=[provider url] npm run tape -- 'tests/ethersStateManager.spec.ts'
+ *  Note: Cloudflare only provides access to the last 128 blocks on these tests so will fail on these tests.
```
gabrocheleau, line 214:
```suggestion
 *  Note: Cloudflare only provides access to the last 128 blocks so it will fail on these tests.
```

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,519 @@
...
+    }
+
+    const returnValue: Proof = {
+      address: address.toString(),
+      balance: bigIntToHex(account.balance),
+      codeHash: bufferToHex(account.codeHash),
+      nonce: bigIntToHex(account.nonce),
+      storageHash: bufferToHex(account.storageRoot),
+      accountProof,
+      storageProof: [],
```
gabrocheleau, line 412:
```suggestion
      storageProof,
```
is it intentional that we're returning an empty array?

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,519 @@
...
+
+import type { Proof, StateManager } from '.'
+import type { StorageDump } from './interface'
+import type { StorageProof } from './stateManager'
+import type { Common } from '@ethereumjs/common'
+import type { Address, PrefixedHexString } from '@ethereumjs/util'
+
+const log = debug('statemanager')
+
+type blockTagType = bigint | 'latest' | 'earliest'
```
gabrocheleau, line 28:
```suggestion
type BlockTagType = bigint | 'latest' | 'earliest' | 'pending'
```

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,519 @@
...
+export interface EthersStateManagerOpts {
+  provider: string | StaticJsonRpcProvider | JsonRpcProvider
+  blockTag?: blockTagType
```
gabrocheleau, line 31:
```suggestion
  blockTag?: BlockTagType
```

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,519 @@
...
+      this.provider = opts.provider
+    } else {
+      throw new Error(`valid JsonRpcProvider or url required; got ${opts.provider}`)
+    }
+
+    if (typeof opts.blockTag === 'bigint') {
+      this.blockTag = bigIntToHex(opts.blockTag)
+    } else {
+      this.blockTag = opts.blockTag ?? 'latest'
+    }
```
gabrocheleau, line 63: Could we get rid of that duplicate logic and just do this:
```suggestion
this.setBlockTag(opts.blockTag ?? 'latest')
```
might complain about not assigning this.blockTag in the constructor though

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,519 @@
...
+    ;(newState as any).storageTries = { ...this.storageTries }
+    return newState
+  }
+
+  /**
+   * Sets the new block tag used when querying the provider and clears the
+   * internal cache.
+   * @param blockTag - the new block tag to use when querying the provider
+   */
+  setBlockTag(blockTag: blockTagType): void {
```
gabrocheleau, line 83:
```suggestion
  setBlockTag(blockTag: BlockTagType): void {
```

gabrocheleau (commented):
Looks great. A few minor comments.

packages/statemanager/README.md
acolytec3: That's precisely the scenario but there's no way to know ahead of time until you actually run the transaction and get the proof nodes into the trie that you're going to hit this scenario and even then, there's no way to recover from it since we don't have the sibling nodes to the parent node and no way to get they keys associated with those nodes.

packages/block/src/from-rpc.ts
g11tech: thanks @gabrocheleau that link was informational! since this would depend on chainId too as per my understanding, we should have a helper fn that can handle the range

packages/statemanager/src/ethersStateManager.ts
g11tech: or use tertiary op: `this.blockTag = (typeof opts.blockTag === 'biging')?bigIntToHex(opts.blockTag): opts.blockTag ?? 'latest'`

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,519 @@
...
+      await storageTrie.put(key, encodedValue)
+    } else {
+      try {
+        await storageTrie.del(key)
+      } catch (err: any) {
+        if (
+          err.message === 'Missing node in DB' &&
+          (err.stack as string).includes('async Trie.del')
+        ) {
+          throw new Error(
```
g11tech, line 219: can we also attach the original stack of the `err` here as it would be informational

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,519 @@
...
+      for (let x = 0; x < blockData.uncles.length; x++) {
+        const headerData = await this.provider.send('eth_getUncleByBlockHashAndIndex', [
+          blockData.hash,
+          intToHex(x),
+        ])
+        uncleHeaders.push(headerData)
+      }
+    }
+
+    return blockFromRpc(blockData, uncleHeaders, { common, hardforkByBlockNumber: true })
```
g11tech, line 453: this would fail for post merge blocks

packages/statemanager/README.md
```diff
@@ -42,6 +42,28 @@ await stateManager.commit()
...
+
+Refer to [this test script](./tests/ethersStateManager.spec.ts) for complete examples of running transactions and blocks in the `vm` with data sourced from a provider.
+
+**WARNING** When using the `EthersStateManager`, any time an account self destruct occurs or a storage slot is deleted while running transactions and blocks, do not be surprised to see an error. This is due to a fundamental limitation of the `web3provider` interface whereby we are not able to retrieve all of the necessary trie nodes needed to update the parent node of a deleted leaf resulting from the storage slot deletion/account self destruct. Many blocks should execute successfully and report the correct updated state root. However, it is not possible to predict ahead of time if any particular transaction will result in this error since all keys for values in the underlying trie are hashed.
```
g11tech, line 65:
```suggestion
**WARNING** When using the `EthersStateManager`, any time an account self destruct occurs or a storage slot is deleted while running transactions and blocks, do not be surprised to see an error. This is due to a fundamental limitation of the `web3provider` interface whereby we are not able to retrieve all of the necessary trie nodes needed to update the parent node of a deleted leaf resulting from the storage slot deletion/account self destruct as the keys in the underlying trie *backing* this statemanager are hashed.
However, many blocks should execute successfully and report the correct updated state root.
```

packages/statemanager/src/ethersStateManager.ts
acolytec3: Nope, just missed this :facepalm:

holgerd77:
Superseeded by #2315, will close.

Issue #1938, opened 2022-06-06
Update CI to support `workflow_dispatch` option


acolytec3:
Our CI currently only runs on PRs and pushes to master.  We should add [`workflow_dispatch`](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) to allow us to manually trigger the CI as needed.

acolytec3:
Have a look [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows) and [here](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) for an idea. We don't need any special inputs, just to be able to trigger them from the UI

PR #2060, opened 2022-07-20
ci: add `workflow_dispatch` event trigger

faustbrian:
> Resolves https://github.com/ethereumjs/ethereumjs-monorepo/issues/1938

Not sure if you had any inputs in mind that should be possible to be entered.

CleanShot 2022-07-20 at 16 00 26

### DIFF
--- a/.github/workflows/block-build.yml
+++ b/.github/workflows/block-build.yml
@@ -5,6 +5,7 @@ on:
     tags: ['*']
   pull_request:
     types: [opened, reopened, synchronize]
+  workflow_dispatch:
 
 env:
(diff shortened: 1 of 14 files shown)


acolytec3 (approved):
Thanks andLGTM.  I don't think we need any inputs at this point but will read further on what's available to think through if there might be any use cases where we'd want them.  CI just runs on push to master and PRs right now so there's no additional intelligence needed.

holgerd77:
Please always label, either submitter or latest reviewer, thanks! 🙏

holgerd77:
(Otherwise great addition, thanks a lot! 😄)

PR #2063, opened 2022-07-21
evm: add new exports requested by hardhat

### LINKED ISSUE #2062: Add new exports needed for Hardhat
Hardhat requested we export the following items from the `evm` as part of the final breaking release
- `ERROR` enum
- `ExecResult`
- `InterpreterStep`
- `Message`

acolytec3:
Fixes #2062

### DIFF
--- a/packages/evm/src/index.ts
+++ b/packages/evm/src/index.ts
@@ -1,15 +1,20 @@
 import { getActivePrecompiles } from './precompiles'
 import { EEIInterface, EVMInterface, EVMStateAccess, Log } from './types'
-import { EVM, EVMResult } from './evm'
-import { EvmError } from './exceptions'
-
+import { EVM, EVMResult, ExecResult } from './evm'
+import { EvmError, ERROR as EvmErrorMessage } from './exceptions'
+import { InterpreterStep } from './interpreter'
+import { Message } from './message'
 export {
   getActivePrecompiles,
   EEIInterface,
   EVM,
   EVMInterface,
   EvmError,
+  EvmErrorMessage,
   EVMStateAccess,
+  ExecResult,
   Log,
   EVMResult,
+  InterpreterStep,
+  Message,
 }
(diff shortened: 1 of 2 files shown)


paulmillr:
Can't they just import the files themselves instead of index.ts?

acolytec3:
> Can't they just import the files themselves instead of index.ts?

@alcuadrado Can you or someone from your team comment here?  I don't see an issue with us adding these as toplevel exports since I believe you were having to import them from `@etheremjs/evm/dist/...` which was kind of hideous.

@paulmillr Is there a reason for keeping these out of the top level exports?

paulmillr:
@acolytec3 reason is simple: tree shaking. For example, [this](https://github.com/paulmillr/noble-hashes/tree/main/src) has 10 files, if a user wants one, they import one; other files won't be included into the build; which will make the result file much smaller.

acolytec3:
> Can't they just import the files themselves instead of index.ts?

Got it.  I'm not decrying the value of tree-shaking but that would be a complete departure from how we do exports right now and I wouldn't want to start piecemeal changing our approach. 

Beyond that, if we're going to go the route you suggest, would it make more sense to use an `exports` directive in `package.json` and define all the export paths that way?

paulmillr:
Yeah, better keep the approach as-is before the release.

> and define all the export paths that way

I think you will need to do this in any case in order to support common.js + esm at the same time.

acolytec3:
@alcuadrado @fvictorio  So any opinion whether to add these imports or do want to leave them where they were and we can maybe go the route of adding import directives in the package.json?

holgerd77:
Please label

holgerd77:
I think if we have users which need theses imports it makes a lot of sense to add respectively we should do.

g11tech (approved):
LGTM!

fvictorio:
> I think can do the deep imports, and @fvictorio was mostly worried about importing from '/dist/'. Can you confirm, Franco?

Sorry, I just saw this, but yes, I confirm.

acolytec3:
> > I think can do the deep imports, and @fvictorio was mostly worried about importing from '/dist/'. Can you confirm, Franco?
> 
> Sorry, I just saw this, but yes, I confirm.

Okay, for now, they've been added to the top level import.  Obviously, this could be reworked if needed along the way.

PR #2064, opened 2022-07-21
Merge tests

### LINKED ISSUE #2038: VM: Update ethereum/tests to "Tests update 11: Merge PoS"
A new `ethereum/tests` version has just been [released](https://github.com/ethereum/tests/releases/tag/v11) focusing on the Merge.

We should update here during the next 1-2 weeks or so.

There are actually three types of tests included, being 1. the pure State/blockchain tests for Merge HF rules (VM), 2. transition tests for ArrowGlacier -> Merge (VM as well) and 3. ArrowGlacier difficulty tests (Block).

First part 1. is likely a classical test runner update on the VM, for 2. I am not sure if this is already fully supported or if this need #1901 needs to be merged for this (maybe @jochem-brouwer can clarify), for 3. the import of the GrayGlacier testdata in `difficulty.spec.ts` in Block can now be updated to directly take in this data and the temporary local test files can be deleted again.

jochem-brouwer:
Closes #2038 

This PR focusses on Merge tests, and especially on the ArrowGlacier -> Merge transition (this would in production most likely be GrayGlacier -> Merge, of course). 

The bugs were in the test runner, @g11tech and I had a conversation about what should be done externally and what should be done internally. This PR had focussed on fixing bugs internally in the packages but in the end we figured out we could fix these bugs in the test runner (which we did). However this might get some follow-ups to also harden these transitions inside our libraries.


jochem-brouwer:
Ok this is a though one, will stop for today, have tried fixing the merge transition tests but this keeps finding bugs (both in VM but also in the test runner), which mainly have to do with correctly setting Common. Since we copy common all over the places this common should be set to the right HF before the common is copied, the question is: where should this happen? In the VM, blockchain, or should it happen in the test runner...?

packages/blockchain/src/blockchain.ts
```diff
@@ -1172,8 +1176,13 @@ export class Blockchain implements BlockchainInterface {
...
   }
 
-  protected checkAndTransitionHardForkByNumber(number: bigint, td?: BigIntLike): void {
+  protected checkAndTransitionHardForkByNumber(
+    number: bigint,
+    blockHeader: BlockHeader,
+    td?: BigIntLike
+  ): void {
     this._common.setHardforkByBlockNumber(number, td)
+    blockHeader._common.setHardforkByBlockNumber(number, td)
```
jochem-brouwer, line 1185: Because `common` is copied, now we have to also explicitly set hardfork in the blockHeader. There is a high chance we have more bugs which are the same (common hardfork changes in one package, where it should also change in another package, but because it is copied this is not done and we should thus explicitly change the "chain" of all common hardforks of relevant package objects).

g11tech (changes requested):
blocking the PR till can figure out what problems its trying to solve, and if this is the correct way to solve it

g11tech (dismissed):
LGTM! lets go :rocket:

g11tech (approved):

packages/vm/tests/tester/runners/GeneralStateTestsRunner.ts
```diff
@@ -116,6 +122,7 @@ async function runTestCase(options: any, testData: any, t: tape.Test) {
         await vm.runTx({ tx, block })
         execInfo = 'successful tx run'
       } catch (e: any) {
+        console.log(e)
```
acolytec3, line 125:
```suggestion
```

Can we remove this console log?

acolytec3 (changes requested):
I this all looks good, provided CI passes.  There is one presumably extraneous console log we should remove but looks good otherwise.

g11tech (approved):

g11tech:
rebased via UI

acolytec3 (approved):

packages/vm/tests/tester/runners/GeneralStateTestsRunner.ts
jochem-brouwer: We should ook in our eslint rules, you get a warning if it happens in /src but not if it is in /tests. (Which is annoying)

acolytec3: I think Ryan turned it off for tests but that's an easy fix if so.

acolytec3: > We should ook in our eslint rules, you get a warning if it happens in /src but not if it is in /tests. (Which is annoying)

Just to round this out.  The `no-console` lint rule is only off in the VM (and only in the test runner) since we have a lot of custom logging output there.  We could switch it all to use tape comments or something if it's important enough to turn the warning back on for this section of the code.

image

PR #2068, opened 2022-07-22
fix(trie): pass down `hash` function for hashing on trie copies

faustbrian:
> Resolves https://github.com/ethereumjs/ethereumjs-monorepo/pull/2067#discussion_r927240858 via https://github.com/ethereumjs/ethereumjs-monorepo/pull/2067#discussion_r927255515
> Needed for https://github.com/ethereumjs/ethereumjs-monorepo/pull/2067 to write proper tests

Something I noticed here https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/trie/src/trie/secure.ts#L109 is that `hash: this.hash` seems to be missing. The `Trie` passes it down when it calls `copy` but `CheckpointTrie` and `SecureTrie` don't. Is this intentional or an oversight?

```ts
  copy(): Trie {
    return new Trie({
      db: this.db.copy(),
      root: this.root,
      deleteFromDB: this._deleteFromDB,
      hash: this._hash,
    })
  }
```

I assume this could result in corrupted data if you copy a `CheckpointTrie` or `SecureTrie` that for example uses `sha256` but then it doesn't pass down the hash function and the copy works with `keccak256` hashes because `const _hash = opts?.hash ?? keccak256` will fall back to the `keccak256` function?

```ts
  it.test('created copy uses the correct hash function', async function (t) {
    const trie = new SecureTrie({
      db: new LevelDB(),
      hash: (value) => createHash('sha256').update(value).digest(),
    })

    await trie.put(Buffer.from('key1'), Buffer.from('value1'))
    trie.checkpoint()
    await trie.put(Buffer.from('key2'), Buffer.from('value2'))
    const trieCopy = trie.copy()
    const value = await trieCopy.get(Buffer.from('key1'))
    t.equal(value!.toString(), 'value1')
    t.end()
  })
```

```sh
# created copy uses the correct hash function

not ok 183 TypeError: Cannot read properties of null (reading 'toString')
  ---
    operator: error
    stack: |-
      TypeError: Cannot read properties of null (reading 'toString')
          at Test.<anonymous> (/Users/brianfaust/Work/bearmint/ethereumjs-monorepo/packages/trie/test/trie/secure.spec.ts:167:20)
```

I think the above reproduces the issue. If I then change to the following `copy` function for `SecureTrie` it'll pass.

```ts
  copy(includeCheckpoints = true): SecureTrie {
    const secureTrie = new SecureTrie({
      db: this.dbStorage.copy(),
      root: this.root,
      deleteFromDB: (this as any)._deleteFromDB,
      hash: this._hash,
    })
    if (includeCheckpoints && this.isCheckpoint) {
      secureTrie.db.checkpoints = [...this.db.checkpoints]
    }
    return secureTrie
  }
```

The above test will fail because it will store `key1` hashed via `sha256` and `key2` hashed via `keccak256` but then the **copy** of the trie will use `keccak256` again because `this.hash` wasn't passed down which results in it looking up `key1` hashed via `keccak256` instead of the `sha256` it was stored as.


acolytec3 (approved):
LGTM

PR #2069, opened 2022-07-22
vm/blockchain: change and use BlockchainInterface

jochem-brouwer:
This PR updates the BlockchainInterface. It also guards any methods which are not exposed by the interface.

Could someone give me a go/no-go? I have to fix a lot of tests due to this type change and I don't want to waste any time if someone raises objections here.


acolytec3:
Given that we've committed in writing multiple times to supporting a `BlockchainInterface`, I say move forward with it.

holgerd77:
Ah, that's nice, yes, we should indeed use the interface, for e.g. Hardhat is doing their own blockchain implementation (not sure if I am having the whole context on this though, just wanted to express my first-sight assessment).

acolytec3:
I've added the remaining missing methods (and adjusted the js docs) to the interface and all the tests are passing now.  @alcuadrado @fvictorio does this cover what you need for the blockchain interface?  I'm assuming there may have to be some dummy functions (like that `_init`) in your implementation to make this work?

jochem-brouwer:
Great, thanks @acolytec3! :)

packages/vm/tests/api/runBlock.spec.ts
```diff
@@ -229,6 +229,22 @@ tape('runBlock() -> API parameter usage/data errors', async (t) => {
       })
   })
 
+  t.test('should fail when no `validateHeader` method exists on blockchain class', async (t) => {
+    const vm = await VM.create({ common })
+    const blockRlp = toBuffer(testData.blocks[0].rlp)
+    const block = Object.create(Block.fromRLPSerializedBlock(blockRlp))
+    ;(vm.blockchain as any).validateHeader = undefined
```
jochem-brouwer, line 236: I have first tried to keep the blockchain interface at minimal since most methods are used somewhat internally and they are technically not necessary for a blockchain to support (e.g. if you dont have `genesisState` then only a small subset of features do not work). However, since you've gone forward and added these methods to the interface, then we should also remove those errors right, since now the interface specifies that these methods are necessary. (An implementer can add the method but always throw a `NotImplementedException`)

acolytec3: Ah, I see.  Ugh, sorry, that makes a lot of sense.  Were you able to get the VM tests passing locally if we remove those other methods?

jochem-brouwer: I did not check if tests were passing since the tests did not compile at that point. But they should pass, I think we use our own blockchain object internally and not mock blockchains.

acolytec3: At a minimum, we would need to add a check for that `validateHeader` like you did for `genesisState` if we want to remove that from the interface.  Is there anything else we could leave out of the interface without breaking the VM?

jochem-brouwer: I actually think we should keep `validateHeader` mandatory. The other methods, I'm not sure if we want to make those mandatory. Maybe let's not remove those from the interface but instead make them optional, i.e. add a `?` to them. I think all methods below `copy` in the interface except `validateHeader` should be made optional (and then we would indeed need those checks if we want to call that method!). This will also most likely break a lot of tests.

packages/client/lib/execution/vmexecution.ts
```diff
@@ -296,6 +321,9 @@ export class VMExecution extends Execution {
         )
       }
       startHeadBlock = endHeadBlock
+      if (typeof this.vm.blockchain.getCanonicalHeadBlock !== 'function') {
+        throw new Error('cannot get iterator head: blockchain has no getTotalDifficulty function')
```
jochem-brouwer, line 325: getTotalDifficulty -> getCanonicalHeadBlock

fvictorio:
Thank you folks, this seems to be a good improvement. I will give it a try and let you know how it goes.

> I'm assuming there may have to be some dummy functions (like that _init) in your implementation to make this work?

From what I see in the code, you are checking first if `_init` exists before calling it, so I guess we don't need to add the dummy method anymore?

acolytec3:
> Thank you folks, this seems to be a good improvement. I will give it a try and let you know how it goes.
> 
> > I'm assuming there may have to be some dummy functions (like that _init) in your implementation to make this work?
> 
> From what I see in the code, you are checking first if `_init` exists before calling it, so I guess we don't need to add the dummy method anymore?

That's the thought.  The new methods we added to the interface are the minimal set needed for the `vm` to work with the interface.

acolytec3:
@holgerd77 @jochem-brouwer @gabrocheleau given the positive feedback from Hardhat on this approach, I think this is ready for review.  I've rebased on current master so just waiting for tests to pass but should be good at this point.  I won't approve/merge since I've made a number of commits to it so don't feel right approving my own work.

packages/blockchain/src/types.ts
```diff
@@ -34,9 +35,45 @@ export interface BlockchainInterface {
...
+  /**
+   * Returns a copy of the blockchain
+   */
+  copy(): BlockchainInterface
+
+  /**
+   * Returns the specified iterator head.
+   *
+   * @param name - Optional name of the iterator head (default: 'vm')
+   */
```
holgerd77, line 52: This is not aligned with the associated method.

holgerd77 (approved):
LGTM, one comment, but - relatively - tiny, will add to tiny-issues list.

packages/blockchain/src/types.ts
holgerd77: Have added to #2105

packages/blockchain/src/types.ts
```diff
@@ -34,9 +35,45 @@ export interface BlockchainInterface {
    *
    * @param name - Name of the state root head
    * @param onBlock - Function called on each block with params (block: Block,
```
jochem-brouwer, line 37: Not part of this PR but this comment seems incomplete?

packages/blockchain/src/types.ts
```diff
@@ -34,9 +35,45 @@ export interface BlockchainInterface {
...
+   * @param name - Optional name of the iterator head (default: 'vm')
+   */
+
+  /**
+   * Validates a block header, throwing if invalid. It is being validated against the reported `parentHash`.
+   * @param header - header to be validated
+   * @param height - If this is an uncle header, this is the height of the block that is including it
+   */
+  validateHeader(header: BlockHeader, height?: bigint): Promise<void>
+  getIteratorHead?(name?: string): Promise<Block>
```
jochem-brouwer, line 60: Missing docs

holgerd77: Docs are a bit more above, have added to the tiny-fixes issue (that was the trigger 😋)

PR #1897, opened 2022-05-18
client: Add client protocol and synchronizer for snap sync

g11tech:
This PR will enable fetching snap data for accounts,storage,codeby using the devp2p interface exposed in the PR: https://github.com/ethereumjs/ethereumjs-monorepo/pull/1883.  

Currently this PR is pointing to #1883, but when that gets merged, this PR will be updated to pointing master. This is so we can leverage the devp2p work being done in 1883 while independently developing some client capabilities with regard to consuming and fetching snap data (and get some early feedback)

### Curent status
`snap` protocol availability in the peer:
```typescript
[05-18|16:28:57] DEBUG Peer connected: id=e9631196 address=10.244.0.112:30303 transport=rlpx protocols=eth,snap inbound=false 
[05-18|16:28:57] DEBUG Found full peer: id=e9631196 address=10.244.0.112:30303 transport=rlpx protocols=eth,snap inbound=false 
[05-18|16:28:57] DEBUG Peer added: id=e9631196 address=10.244.0.112:30303 transport=rlpx protocols=eth,snap inbound=false 
```
Next Steps (all with a static peer):
- [x] Call getAccountRange on snap capable peer
- [x] Add remaining methods
 
Tests and coverage will follow via 
- https://github.com/ethereumjs/ethereumjs-monorepo/pull/1933

PR #2070, opened 2022-07-23
client: snap protocol fetchers

scorbajio:
Continues the snap protocol implementation work. Work left to do:
- [ ] Build fetchers for accounts, storage slots, and bytecode, and just dump it into a mock MPT
- [ ] Build MPT and successfully run move the chain forward on a normal eth sync
- [ ] Chart out behavior of non static peer and map next steps


holgerd77:
You are somewhat off with where you branched off here (lot's of old commits in), can you resubmit in a clean state?

scorbajio:
@holgerd77 Should I be continuously syncing with master on this branch?

holgerd77:
> @holgerd77 Should I be continuously syncing with master on this branch?

I am not sure if I understand your question, so if you want to continue to work here, you have to bring this to the latest base branch state at some point (of course).

Can you maybe write a bit what's the state of this PR, what is left to do, when/if you would plan to continue.

scorbajio:
> > @holgerd77 Should I be continuously syncing with master on this branch?
> 
> I am not sure if I understand your question, so if you want to continue to work here, you have to bring this to the latest base branch state at some point (of course).
> 
> Can you maybe write a bit what's the state of this PR, what is left to do, when/if you would plan to continue.
Thanks for clarifying, some of the changes from #1897 aren't in this PR (like the `forceSnapSync` option) but do show up in master. Not sure if I missed something while I was branching. Might need to just pull from `master` to include those changes.

Work that needs to be done:
- [ ] First, we need to capture data for `AccountRange`, `StorageRanges`, `ByteCodes`, and `TrieNodes` to get practical visibility into what the data for those calls include and how to use them
- [ ] Create account fetcher
- [ ] Create storage fetcher
- [ ] Create byte code fetcher
- [ ] Create trie node fetcher

My understanding is that there needs to be one fetcher per each `get` call defined in the snap specs. Open to suggestions on this design plan.

scorbajio:
I see so `snap-client` has been merged and deleted and I need to reraise a new PR to merge with `master`.

scorbajio:
Work is being continued at #2107 .

Issue #2061, opened 2022-07-20
Trie values returning null after reboot?


acolytec3:
Following on the below discussion, we should add a new option to the base `trie` constructor to store the root hash under a `root` key in the db so it can be retrieved if a persistent DB is provided in the constructor.

### Discussed in https://github.com/ethereumjs/ethereumjs-monorepo/discussions/2056

<div type='discussions-op-text'>

<sup>Originally posted by **faustbrian** July 19, 2022</sup>
Indeed me being dense the root has to be manually restored. Maybe this could be improved by introducing a decorator like `new PersistedTrie(new SecureTrie())` which then takes care of writing the root to and reading the root from disk? Might be a nicer DX than everyone coming up with their own way of persisting the root and manually restoring it. Is there any mention in docs that it has to be restored manually?

---

So I'm attempting to use the trie package to store some persistent data that I need to access as soon as my application launches but something seems to be amiss here, probably me being dense. I'm creating a `ClassicLevel` instance that I pass to `SecureTrie`, then log that the value is `null`, then store `key` with a value of `one` and finally log that the value is `one`. That's all fine but then when I run the exact same script again `key` is again `null` instead of `one` as I would expect because in the first run we finished the script with `key` being `one` and we are using a LevelDB implementation with persistence.

```js
const { SecureTrie, LevelDB } = require('@ethereumjs/trie')
const { ClassicLevel } = require('classic-level')

const level = new ClassicLevel('./db')
const root = new SecureTrie({ db: new LevelDB(level) })

async function test() {
  // the first time we run this script this should be 'null'
  // the second time we run this script it should be 'one' (because of the 'put' call) but it isn't
  console.log(await root.get(Buffer.from('key'))) // null

  await root.put(Buffer.from('key'), Buffer.from('one'))

  console.log((await root.get(Buffer.from('key'))).toString()) // one
}

test()
```

The value can be accessed directly via Level so it is in the database. I assume it's because the root is not set after relaunching so it falls back to the `KECCAK256_RLP` value but how would the root be initialised so that all data from previous runs is accessible? Should this happen automatically or do I need to manually restore the root?

```js
try {
    console.log(await level.get(Buffer.from('401fb497d3e347e3d480e3137b1372491314d54a46f764e0bf00336649feb6c6', 'hex')));
} catch(e) {
    console.log(e)
}
```

[...]

PR #2071, opened 2022-07-23
feat(trie): implement root hash persistence

faustbrian:
Resolves https://github.com/ethereumjs/ethereumjs-monorepo/issues/2061

### DIFF
--- a/packages/trie/src/trie/checkpoint.ts
+++ b/packages/trie/src/trie/checkpoint.ts
@@ -71,7 +71,7 @@ export class CheckpointTrie extends Trie {
       db: this.dbStorage.copy(),
       root: this.root,
       deleteFromDB: (this as any)._deleteFromDB,
-      persistRoot: (this as any)._persistRoot,
+      persistRoot: this._persistRoot,
       hash: (this as any)._hash,
     })
     if (includeCheckpoints && this.isCheckpoint) {


acolytec3:
> Should I throw an exception if put methods of any trie are called with the ROOT_DB_KEY key, meaning a user manually wanted to modify that value?

My vote is yes (provided that the `trie.persistRoot` option is set.  @jochem-brouwer @gabrocheleau What do you think about this?  If we're persisting a root, it doesn't make sense to allow the user to modify that root, right?

>I used __root__ instead of __persistedRootHash__ since the __ prefix and suffix should already distinguish it enough from user space wording.

This is fine.

acolytec3:
> @acolytec3 one concern I have with defaulting the behaviour to `true` is that it means that everyone that has an existing database and upgrades to 5.0 needs to explicitly set it to `false` so this should probably be made very clear in the changelogs or upgrade guide because someone might assume that nothing but their own keys will enter the database by default. For example if you expect the root to be XYZ with your keys but then after upgrading to 5.0 and it writing the root to the database the first time you perform a write operation you would get a different result than you expect.

Good point. @jochem-brouwer @gabrocheleau Any thoughts on the defaults here? I still think default to persisting the root makes sense and I'm happy to write the updates to the readme if y'all are for with going this route

holgerd77:
I have some mild tendency to keep this on `false` but can see that there are somewhat good arguments for both ways.

acolytec3 (changes requested):
Looks great, especially the new tests.  Left some minor feedback but excited to have this added in.  I'm going to open an issue to remind us to update the usage guidelines once this is merged.

acolytec3 (commented):
Looks like my comments may have gotten lost somehow.  Let me re do them.

acolytec3 (changes requested):
Not sure where they went but just left the substance of the feedback in the two comments I added.  Let me know if any of this doesn't make sense.

packages/trie/src/trie/trie.ts
```diff
@@ -46,11 +57,29 @@ export class Trie {
...
+    } else {
+      this._persistRoot = false
+    }
+
     if (opts?.root) {
       this.root = opts.root
     }
   }
 
+  static async make(opts?: TrieOpts) {
```
acolytec3, line 71: Our pattern for static constructors in the monorepo is to call them create (see `blockchain.create` and `vm.create`). Can you rename this and all the other references to follow that pattern?

packages/trie/src/trie/trie.ts
```diff
@@ -116,6 +145,10 @@ export class Trie {
    * @returns A Promise that resolves once value is stored.
    */
   async put(key: Buffer, value: Buffer): Promise<void> {
+    if (this._persistRoot) {
+      ok(!key.equals(ROOT_DB_KEY), `Attempted to set '__root__' key but it is not allowed.`)
```
acolytec3, line 149: Let's directly `throw new Error` here instead of using assert utilities. We recently removed all references to `assert` across the monorepo and don't want to reintroduce them.

ScottyPoi:
LGTM.  Especially the tests.

One note is that we have tests for SecureTrie.copy(), but not of Trie.copy().  Things can get weird when making copies.

packages/trie/src/trie/trie.ts
```diff
@@ -46,11 +56,29 @@ export class Trie {
     this._root = this.EMPTY_TRIE_ROOT
     this._deleteFromDB = opts?.deleteFromDB ?? false
 
+    if (opts?.db) {
+      this._persistRoot = opts.persistRoot ?? true
```
jochem-brouwer, line 60: Why must one pass a `db` if we want to persist root?

packages/trie/src/trie/trie.ts
```diff
@@ -46,11 +56,29 @@ export class Trie {
     this._root = this.EMPTY_TRIE_ROOT
     this._deleteFromDB = opts?.deleteFromDB ?? false
 
+    if (opts?.db) {
```
jochem-brouwer, line 59:
```suggestion
    if (opts?.db !== undefined) {
```

packages/trie/test/trie/trie.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+    'creates an instance via the static constructor `make` function and defaults to `false` without a database',
+    async function (st) {
+      st.false(((await Trie.create()) as any)._persistRoot)
+    }
+  )
+
+  t.test('persist the root if the `persistRoot` option is not `false`', async function (st) {
+    const trie = await Trie.create({ db: new LevelDB(new MemoryLevel()) })
+
+    st.equal(await trie.db.get(ROOT_DB_KEY), null)
```
jochem-brouwer, line 35: Should this not return the roothash of the empty trie instead of null?

ScottyPoi: Elsewhere you have this:
```
st.notEqual(await copy.db.get(ROOT_DB_KEY), null)
```
Did you mean to use `notEqual` here?

packages/trie/src/trie/trie.ts
jochem-brouwer: (Shouldn't this also trigger the linter..?)

packages/trie/src/trie/trie.ts
ScottyPoi: I think this means that if you pass a db, it will persist unless `persistRoot` is set to false?

jochem-brouwer: Well if a custom DB is not passed in it will default `persistRoot` to `false` :thinking:

acolytec3: > Why must one pass a `db` if we want to persist root?

Persisting the root just means we right a specific key in the DB with the root value from the `trie` so it can be retrieved if we reinstantiate the trie.  So, if you have an application where you want to keep a trie around across runs of the script, the only way to find the actual root of the `trie` is to set the root hash in the DB with the `__root__` key (or whatever key you wanted to use, but that's what's used here).

acolytec3: So, if you don't pass a db in the constructor, there's no underlying data persisted from that trie so no reason to "persist" the trie since you would just be writing it to an in-memory database that gets destroyed when the script ends.

packages/trie/test/trie/trie.spec.ts
acolytec3: > Should this not return the roothash of the empty trie instead of null?

This is just retrieving the value stored under the special `__root__` key in the DB.  The roothash of the empty trie is returned if you call `trie.root` (which is different).  I guess we could programmatically set the `__root__` key to the hash of the empty trie if a DB is provided but not sure what value that adds.

acolytec3:
> @acolytec3 are the client tests flaky? Sometimes they pass instantly and sometimes they need to be rerun because they run for ages.

The goerli full and light client ones are quite flaky since they start a node and try to start syncing the goerli testnet which is hit or miss. Is that what you're talking about? Those two aren't required to pass on order for us to merge anyway.

packages/trie/test/trie/trie.spec.ts
```diff
@@ -0,0 +1,89 @@
+import { Level } from 'level'
+import { MemoryLevel } from 'memory-level'
+import { tmpdir } from 'os'
+import * as tape from 'tape'
+import { LevelDB, ROOT_DB_KEY, Trie } from '../../src'
+
+tape('Trie', function (t) {
+  t.test(
+    'creates an instance via the static constructor `make` function and defaults to `true` with a database',
```
acolytec3, line 9: Can you update all these test comments to reference `create` instead of `make`? After that, I'll approve and we can get this merged. Great work!

acolytec3 (changes requested):

acolytec3 (dismissed):
LGTM! @jochem-brouwer any last questions before we merge?

packages/trie/test/trie/trie.spec.ts
acolytec3: > @acolytec3 @jochem-brouwer I guess we could call `persistRoot()` in `create` if no root has been stored yet? Don't really see the benefit of that since the value will always be the same for an empty trie but could be an option.

If there is an `opts.root`, I think it makes sense. So I would just add a call to persist root in the constructor where you set root of that opt is provided.

packages/trie/src/trie/trie.ts
```diff
@@ -46,11 +56,29 @@ export class Trie {
...
     this._deleteFromDB = opts?.deleteFromDB ?? false
 
+    if (opts?.db !== undefined) {
+      this._persistRoot = opts.persistRoot ?? true
+    } else {
+      this._persistRoot = false
+    }
+
     if (opts?.root) {
       this.root = opts.root
```
acolytec3, line 66:
```suggestion
      this.root = opts.root
      this.persistRoot()
```

I mean this.  That way, you don't do any every logic and just persist it the same way you do elsewhere in the class.

packages/trie/test/trie/trie.spec.ts
acolytec3: Sorry, was hard to do on my phone.  LIke [this](https://github.com/ethereumjs/ethereumjs-monorepo/pull/2071/files#r931122810)

acolytec3: Given my other suggestion won't work (can't do async ops inside sync constructor), take the above route except write to the DB in the else branch instead of doing that boolean check.  The DB will be still keep the key when you instantiate the trie so no need to have a separate boolean to check what we already know to be true (i.e. `shouldPersistRoot`).

packages/trie/src/trie/trie.ts
jochem-brouwer: Ok I follow your reasoning, since on `copy` we also copy the root, so if we use memory LevelDB then it does not matter if we persist or not

packages/trie/src/trie/secure.ts
```diff
@@ -29,6 +29,10 @@ export class SecureTrie extends CheckpointTrie {
    * @param value
    */
   async put(key: Buffer, val: Buffer): Promise<void> {
+    if (this._persistRoot && key.equals(ROOT_DB_KEY)) {
+      throw new Error(`Attempted to set '__root__' key but it is not allowed.`)
```
jochem-brouwer, line 33: This is technically not necessary, since in `SecureTrie` we hash the `key` before putting it in the DB. So the key can never be `__root__`

jochem-brouwer: Super nit, but in the error messages probably `ROOT_DB_KEY.toString()` should be used?

packages/trie/src/trie/secure.ts
```diff
@@ -107,11 +111,21 @@ export class SecureTrie extends CheckpointTrie {
       db: this.dbStorage.copy(),
       root: this.root,
       deleteFromDB: (this as any)._deleteFromDB,
+      persistRoot: (this as any)._persistRoot,
```
jochem-brouwer, line 114: `_persistRoot` is `protected` so why do we cast this as `any`?

jochem-brouwer (commented):
Some small questions and nits

packages/trie/test/trie/trie.spec.ts
acolytec3: LGTM

packages/trie/src/trie/checkpoint.ts
```diff
@@ -69,6 +71,7 @@ export class CheckpointTrie extends Trie {
       db: this.dbStorage.copy(),
       root: this.root,
       deleteFromDB: (this as any)._deleteFromDB,
+      persistRoot: (this as any)._persistRoot,
```
jochem-brouwer, line 74:
```suggestion
      persistRoot: this._persistRoot,
```

jochem-brouwer:
When CI passes I will approve, thanks a lot @faustbrian :tada: :+1:

acolytec3 (approved):
LGTM!  Thanks for seeing this one through.

acolytec3:
CI all passed so approving and merging.  Thanks @faustbrian!

holgerd77:
That was really great work here, thanks again a lot! 😄

So, just to recap, if I understand the code correctly we now finally settled to persisting by default (so: set to `true`). This is correct, right?

holgerd77:
It would be good if we can make this a bit more explicit in the code docs (maybe you can even side-add along your current work on #2092 or so).

holgerd77:
I am just writing the Beta 3 release notes for this feature and TBH I find the default options here so confusing that I can barely describe them.

So it would roughly be:

- Trie + default DB: persistRoot === false
- Trie + permanent DB passed: persistRoot === true
- Trie + memory DB passed: persistRoot === true

Would this be correct???

Can't we just skip the DB check at all and then have just the plain same-default-boolean-setting-for-all behavior?

acolytec3:
> Yeah they are quite confusing, that's why I had proposed to always default to `false` and make it opt-in sweat_smile In general not a fan of side-effects like this where behaviours change implicitly based on unrelated input.

I guess this doesn't feel complicated to me (but maybe that's because I pushed for it).  The basic logic I had was, if I've provided a DB (i.e. I'm writing the trie nodes to disk), by default I want the root node persisted so that if I can pick it up where I left off when I next start up the script.  If you're using an memory DB (whether passed in or not), there's no point in persisting the root node since it will disappear when the script ends.  All the checks in the constructor do is see whether you passed in a db or not.  If you did, we assume it's a disk db and not in memory.  We aren't hurting anything by "persisting" the root on an in memory DB that someone passes in.

holgerd77:
> > Yeah they are quite confusing, that's why I had proposed to always default to `false` and make it opt-in sweat_smile In general not a fan of side-effects like this where behaviours change implicitly based on unrelated input.
> 
> I guess this doesn't feel complicated to me (but maybe that's because I pushed for it). The basic logic I had was, if I've provided a DB (i.e. I'm writing the trie nodes to disk), by default I want the root node persisted so that if I can pick it up where I left off when I next start up the script. If you're using an memory DB (whether passed in or not), there's no point in persisting the root node since it will disappear when the script ends. All the checks in the constructor do is see whether you passed in a db or not. If you did, we assume it's a disk db and not in memory. We aren't hurting anything by "persisting" the root on an in memory DB that someone passes in.

Yes, I get why this might be preferred in the one direction in the one scenario and in the other direction in the other scenario, the confusing part is still to have an option which changes the default based on some other input.

Just read the incoming comment from @faustbrian as well while writing, yes, I would also strongly pledge for just doing one thing for all, being it `true` by default or `false`.

holgerd77:
(I would rather also favor `false` here)

acolytec3:
That's fine. Not something I have super strong opinions on.  Was just going with what made sense to me at the time (not always very reliable :grimacing: )

PR #2075, opened 2022-07-25
common: Change td(total difficulty) config to be of bigint

### LINKED ISSUE #2066: Merge tracking issue
Tracking issue for things which need to change which I noticed in #2064 

- [ ] Rename `td` to `ttd` where appropriate. TD: total difficulty. TTD: total terminal difficulty. In `merge` hardfork this should be `ttd`, it is currently `td` (which is confusing). Inputs to most (if not all) methods in common should remain `td` #2075 
- [ ] `td` is currently of type `number` in Common, I think we should change to either string or bigint #2075 

Mainnet TD is currently larger than max safe number, so we should definitely change the type of `number` of `ttd` to either `string` or `bigint` (probably string for JSON)

g11tech:
Pursuant to https://github.com/ethereumjs/ethereumjs-monorepo/issues/2066 this PR changes the type of `td` in `HardforkConfig` to `bigint`

However, decided against changing `td` to `ttd` in merge because as of now it seems that other hardforks could be based on `td` to bring about consensus changes, but `ttd` = total terminal difficulty has associations with only `merge` of Execution + Consensus chains. 

So decided against the name change as `td` = `total difficulty` at which `merge` hardfork is triggered, sort of makes sense. let me know @jochem-brouwer , @holgerd77 how you feel about this?

Closes #2066


acolytec3:
>However, decided against changing td to ttd in merge because as of now it seems that other hardforks could be based on td to bring about consensus changes, but ttd = total terminal difficulty has associations with only merge of Execution + Consensus chains.

I'm not sure this makes sense to me.  After the merge, total difficulty becomes constant since the difficulty on all PoS blocks is 0 so unless someone is running a chain with custom hardforks and extending our code, there will never be another hardfork based on terminal total difficulty.  As such, I'd still vote to change the name to `ttd`

jochem-brouwer:
+1, `TTD` is used as trigger of the merge, when the total difficulty >= total terminal difficulty then at that point the merge is triggered and subsequent blocks will have difficulty zero.

g11tech:
ok, makes sense that there could be only one such hardfork since difficult 0 post it renders td/ttd not usable again. will update

acolytec3 (approved):
LGTM

packages/common/src/types.ts
```diff
@@ -47,7 +47,7 @@ export interface GenesisBlockConfig {
 export interface HardforkConfig {
   name: Hardfork | string
   block: number | null
-  td?: number
+  ttd?: bigint | string
```
holgerd77, line 50: Do we really need a bigint type here? If I see correctly `HardforkConfig` is solely(as the other *Config types) for data coming from JSON and JSON doesn't support bigint? This is then not consistent with the other number related types, e.g. in `GenesisBlockConfig` we do `number` for numbers and `string` for values which we later use as BigInts.

PR #2076, opened 2022-07-26
Point hardhat tests to the upgrade branch

acolytec3:
WIP


holgerd77:
Ah, interesting. Cool! 🙂

acolytec3:
It's only cool if I get it to work :grimacing: I'm stumped at the moment though I suspect it's currently related to our beta releases somehow messing up the versioning in verdaccio.

holgerd77:
Is this more about "helping HH to update" (and so even eventually leading to PRs on the HH side)? Or rather for finding inconsistencies in our own implementation?

(also asking because first possibility is likely also the job for the HH people to some greater extend?)

acolytec3:
The main goals were to see if the job even worked, which it wasn't and then to give me a tracking sense of where their work was in terms of integrating (and maybe reveal any bugs in our code if it surfaces)

holgerd77:
Outdated, will close.

PR #2077, opened 2022-07-26
Move @types/async-eventemitter from devDependencis to dependencies

ScottyPoi:
VM / EVM

From tracking issue #2032

evm:
```
@types/async-eventemitter is listed as devDependency: this should be a dependency, otherwise consumers cannot access the on property of EVM.
```
vm:
```
 @types/async-eventemitter is listed as devDependency: this should be a dependency, otherwise consumers cannot access the on property of VM.
```


jochem-brouwer (approved):
Thanks!

PR #2078, opened 2022-07-26
EVM/VM fixes

acolytec3:
Fixes below two items from #2032 
- `To create EEI, one does import { EEI } from "@ethereumjs/vm"; const eei = new EEI.EEI(...);, this is ugly and it should be new EEI, probably caused by export * as EEI from './eei/eei' should probably be export * from './eei/eei'`

- `When creating a VM and passing an @ethereumjs/evm instance as evm option, one should also ensure that the same EEI in EVM is reused when creating the VM. Otherwise, a new EEI will be created and the warm accounts and storage slots will thus never be cleared (there will most likely also problems with touched accounts, etc.). Probably, EEI should be a public property of the EVM interface, and if the EVM interface is passed as option, then take the EEI from there (and ensure no custom EEI is set, if so, then throw)`


packages/vm/src/vm.ts
```diff
@@ -104,9 +104,16 @@ export class VM extends AsyncEventEmitter<VMEvents> {
 
     // TODO tests
     if (opts.eei) {
+      if (opts.evm) {
+        throw new Error('cannot specify EEI if EVM opt provided')
```
jochem-brouwer, line 108: Could you add a test for this case? And a test to show that VMs EEI is indeed the same EEI as the one used in EVM if one is passed on.

acolytec3: Done!

jochem-brouwer (approved):
LGTM, thanks! :smile:

g11tech (approved):

PR #2081, opened 2022-07-29
block: Set hardforkbyblocknumber true on rlp block constructor

acolytec3:
We encountered a scenario where instantiating a block header using `Blockheader.fromRLPSerializedHeader` was producing incorrect block hashes for block RLPs pulled from mainnet prior to London if the `blockOpt.hardforkByBlockNumber` is not set to true because the default hardfork is now London and if no baseFee is provided in the header data and london is active, the blockheader constructor adds a default basefee field which causes the block hash to be calculated with the additional field included in the hash.  Defaulting `hardforkByBlockNumber` to true in cases where it is not provided and we are using RLP serialized header data seems like a reasonable default setting to avoid strange scenarios where a user expects the constructor to produce the correct block but it has a basefee field attached that wasn't in the RLP data.


g11tech (dismissed):
LGTM

g11tech (approved):
LGTM

packages/block/src/header.ts
```diff
@@ -99,6 +99,15 @@ export class BlockHeader {
       throw new Error('Invalid serialized header input. Must be array')
     }
 
+    // If an RLP serialized block header is provided and no `hardforkByBlockNumber` opt is provided, default true to
+    // avoid scenarios where no `opts.common` or `opts.hardforkByBlockNumber` is provided and a serialized blockheader
+    // is provided that predates London result in a default base fee being added to the block
+    // (resulting in an erroneous block hash since the default `common` hardfork is London and the blockheader constructor
+    // adds a default basefee if EIP-1559 is active and no basefee is provided in the `headerData`)
+    if (opts.hardforkByTTD === undefined) {
```
holgerd77, line 107: `hardforkByTTD` doesn't exist, this should be `hardforkByTD`.

acolytec3: :facepalm: Ack.  Will add it to the fixes list.

holgerd77: Wait, I have another comment... 😋

holgerd77: I will soon write a more general comment on this solution itself, but *if* we would take the solution here, wouldn't we need to also check that `Common` is *not* provided? otherwise we'll overwrite in many cases (which is likely not intended by the user)? 🤔

holgerd77:
I am generally not so sure if I like this solution so much, since this will often lead to false results in unknown chain scenarios where no common is provided (this leads on the side line also to the `Common` question I just posted), since this will always falsely assume mainnet block numbers, also for blocks from custom chains or unknown sources.

Would the following mechanims an alternative?

```
if (noCommonProvided && values.length === 15) { // old block format without base fee
  Initialize with a Common with a pre-London HF (likely: Berlin?)
}
```

On a first round I would think this has substantially less side effects.

acolytec3:
Hmm, that makes sense and I'd be open to that route.  One other edge case to consider (my original issue) is that if we have an RLP serialized block in the pre-london format and someone passes in common with london or later hardfork set, should we throw or just let the user proceed?  In this instance, the user is giving a completely packaged up RLP encoded block (which we assume they retrieved from some RPC endpoint or something similar rather than encoding by hand) so if there is incomplete data in the block (i.e. no basefee) and they've set the hardfork as london, should we throw or let the constructor provide the default base fee?  I'm mainly thinking about the exact edge case I encountered in Ultralight where I wasn't thinking about setting the hardfork and assumed the `fromRLPSerializedHeader` constructor would handle all of that for me since I'm retrieving serialized blocks from the network and just trying to deserialize them and verify the hashes match for purposes of putting into the header accumulator.

holgerd77:
Ok, nice! 🙂

> One other edge case to consider (my original issue) is that if we have an RLP serialized block in the pre-london format and someone passes in common with london or later hardfork set

I had a first thinking of generalize here and doing these kind of things in the main header constructor, but then draw back for various reasons (our current base fee default behavior e.g. which is also convenient I guess and also makes sense in various use case scenarios).

So, yes, I guess it makes sense to directly throw in these cases in the RLP static constructor, so both for:

1. Pre-London Common, base fee provided
2. Post-London Common, no base fee provided

Does this make sense and is an accurate task description?

acolytec3:
Yep!

PR #2082, opened 2022-07-30
Optimize the github workflow checkouts

g11tech:
Optimize the github workflow checkouts


acolytec3 (approved):
Looks great. I checked the run times and the jobs that don't need the submodules run something like 1/3 faster than before.  DX improvements for the win!

holgerd77:
Cool work, thanks a lot! 🙏

PR #2084, opened 2022-07-31
client: allow eth_call without to for testing contract creation

### LINKED ISSUE #1987: Let validator allow eth_call to test creation code
The validator check field by 

https://github.com/ethereumjs/ethereumjs-monorepo/blob/74d5937b115d419a7d3fd697bc9c321355204a33/packages/client/lib/rpc/validation.ts#L199

and eth_call require `to` field

https://github.com/ethereumjs/ethereumjs-monorepo/blob/74d5937b115d419a7d3fd697bc9c321355204a33/packages/client/lib/rpc/modules/eth.ts#L318

The only way to let `to` be undefined(not a address) is

https://github.com/ethereumjs/ethereumjs-monorepo/blob/74d5937b115d419a7d3fd697bc9c321355204a33/packages/client/lib/rpc/modules/eth.ts#L452

The only way to test _executeCreate is let `!!to` be false.

https://github.com/ethereumjs/ethereumjs-monorepo/blob/74d5937b115d419a7d3fd697bc9c321355204a33/packages/evm/src/evm/evm.ts#L815

That is to say only `!!to` is allowed by validator, but it should allow one value that `!to` to test creation code.

g11tech:
client: allow eth_call without to for testing contract creation 

Allows contract creation via eth_call without providing `to`.

Closes #1987

### DIFF
--- a/packages/client/lib/rpc/modules/eth.ts
+++ b/packages/client/lib/rpc/modules/eth.ts
@@ -313,7 +313,7 @@ export class Eth {
     this.blockNumber = middleware(this.blockNumber.bind(this), 0)
 
     this.call = middleware(this.call.bind(this), 2, [
-      [validators.transaction(['to'])],
+      [validators.transaction()],
       [validators.blockOption],
     ])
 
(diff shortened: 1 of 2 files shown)


acolytec3 (commented):
LGTM

acolytec3 (approved):

PR #2086, opened 2022-08-01
chore: set up `eslint-plugin-simple-import-sort` and `eslint-plugin-import`

faustbrian:
- Sets up the first batch of rules from https://github.com/ethereumjs/ethereumjs-monorepo/discussions/2073#discussioncomment-3230792 via https://github.com/lydell/eslint-plugin-simple-import-sort and https://github.com/import-js/eslint-plugin-import/
- The rules for sorting various types of data structures like objects will be submitted after this
- There will be more PRs to set the rules that are set to `warn` to `error` but they require a lot more work because there are hundreds of circular dependencies


acolytec3:
I haven't looked into either this one or #2085 yet but I'm inferring there's a fair amount of overlap since @gabrocheleau referenced your list from #2073. Assuming these two PRs cover the same territory, can y'all consolidate?

gabrocheleau:
> I haven't looked into either this one or #2085 yet but I'm inferring there's a fair amount of overlap since @gabrocheleau referenced your list from #2073. Assuming these two PRs cover the same territory, can y'all consolidate?

Indeed, they both overlap significantly but this one covers some additional rules, so I suggest we keep this one. I'll review.

config/eslint.js
```diff
@@ -46,7 +57,32 @@ module.exports = {
...
+    'import/no-extraneous-dependencies': 'error',
+    'import/no-mutable-exports': 'error',
+    'import/no-namespace': 'off',
+    'import/no-restricted-paths': 'error',
+    'import/no-self-import': 'error',
+    'import/no-unresolved': 'off',
+    'import/no-unused-modules': 'error',
+    'import/no-useless-path-segments': 'error',
+    'import/no-webpack-loader-syntax': 'error',
+    'import/order': 'error',
```
gabrocheleau, line 85: Do we actually need all those additional rules? My intuition would be to only include rules that currently result in an update when applied for performance and general simplicity. I haven't looked up what each of these rules does in detail but here's a couple examples of rules that I suspect don't result in changes:
```
    'import/no-absolute-path': 'error',
    'import/no-anonymous-default-export': 'error',
    'import/no-restricted-paths': 'error',
    'import/no-self-import': 'error',
    
    Curious to hear your thoughts on that, and more generally, I personally feel like it'd be great if the additional rule was justified, so that we avoid bloating up our linting configuration.

File: packages/block/test/block.spec.ts [hunk 2]
```diff
@@ -1,18 +1,19 @@
...
+import { Chain, Common, Hardfork } from '@ethereumjs/common'
 import { NestedUint8Array, toBuffer, zeros } from '@ethereumjs/util'
 import { RLP } from 'rlp'
-import { Chain, Common, Hardfork } from '@ethereumjs/common'
+import * as tape from 'tape'
+// eslint-disable-next-line import/order, @typescript-eslint/no-unused-vars
+import util from 'util'
+
+// explicitly import util, needed for karma-typescript bundling
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
gabrocheleau, line 9:
```suggestion
// explicitly import util, needed for karma-typescript bundling
// eslint-disable-next-line import/order, @typescript-eslint/no-unused-vars
import util from 'util'
```
Also, why do we want to disable import/order for this one? Does having that import on top of the file break something?

packages/client/lib/net/peer/rlpxpeer.ts
```diff
@@ -1,16 +1,18 @@
-import { randomBytes } from 'crypto'
 import {
   Capabilities as Devp2pCapabilities,
   ETH as Devp2pETH,
   LES as Devp2pLES,
   Peer as Devp2pRlpxPeer,
   RLPx as Devp2pRLPx,
 } from '@ethereumjs/devp2p'
+import { isTruthy } from '@ethereumjs/util'
+// eslint-disable-next-line import/order
```
gabrocheleau, line 9: I've noticed a couple of instances where the import/order rule was disabled with some imports. Why is that necessary?

packages/devp2p/src/dpt/dpt.ts
```diff
@@ -1,15 +1,15 @@
-import ms = require('ms')
+import { randomBytes } from 'crypto'
 import { EventEmitter } from 'events'
+import ms = require('ms')
+import { Debugger } from 'debug'
 import { getPublicKey } from 'ethereum-cryptography/secp256k1'
-import { randomBytes } from 'crypto'
+
+import { DNS } from '../dns'
 // import { debug as createDebugLogger } from 'debug'
```
gabrocheleau, line 8: Not related to your work but this comment can be removed.
```suggestion
```

packages/client/lib/net/peer/rlpxpeer.ts
gabrocheleau: Cool thanks for the note!

packages/evm/src/precompiles/index.ts
```diff
@@ -1,6 +1,12 @@
...
 import { Common, Hardfork } from '@ethereumjs/common'
-import { PrecompileInput, PrecompileFunc } from './types'
+import { Address, isTruthy } from '@ethereumjs/util'
+
+import { precompile0a } from './0a-bls12-g1add'
+import { precompile0b } from './0b-bls12-g1mul'
+import { precompile0c } from './0c-bls12-g1multiexp'
+import { precompile0d } from './0d-bls12-g2add'
+import { precompile0e } from './0e-bls12-g2mul'
+import { precompile0f } from './0f-bls12-g2multiexp'
```
gabrocheleau, line 9: In this particular case I think those 6 should still go after the { precompile09 } line

packages/statemanager/tests/stateManager.spec.ts
```diff
@@ -1,22 +1,22 @@
...
 } from '@ethereumjs/util'
-import { Chain, Common, Hardfork } from '@ethereumjs/common'
-import { DefaultStateManager } from '../src'
-import { createAccount } from './util'
-
+import { keccak256 } from 'ethereum-cryptography/keccak'
+import { bytesToHex } from 'ethereum-cryptography/utils'
+import * as tape from 'tape'
 // explicitly import `inherits` to fix karma-typescript issue
 import { inherits } from 'util' // eslint-disable-line
```
gabrocheleau, line 15: Not part of your work either, but could we figure out which eslint rule needs to be disabled specifically on this line, instead of disabling all linting?

packages/trie/test/index.spec.ts
```diff
@@ -1,13 +1,13 @@
-import * as tape from 'tape'
+import { Buffer } from 'buffer'
 import { bufArrToArr, KECCAK256_NULL, KECCAK256_RLP_S } from '@ethereumjs/util'
-import { RLP } from 'rlp'
-import { CheckpointTrie, HashFunc, LeafNode, LevelDB, Trie } from '../src'
-
 // explicitly import buffer,
 // needed for karma-typescript bundling
```
gabrocheleau, line 4: These comments should be moved up along with import { Buffer }

packages/vm/tests/api/index.spec.ts
```diff
@@ -1,20 +1,19 @@
-import * as tape from 'tape'
-import { Account, Address, KECCAK256_RLP } from '@ethereumjs/util'
+import { Buffer } from 'buffer' // eslint-disable-line @typescript-eslint/no-unused-vars
+import * as util from 'util' // eslint-disable-line @typescript-eslint/no-unused-vars
 import { Chain, Common, Hardfork } from '@ethereumjs/common'
 import { DefaultStateManager } from '@ethereumjs/statemanager'
+import { Account, Address, KECCAK256_RLP } from '@ethereumjs/util'
+import * as tape from 'tape'
+// explicitly import util and buffer,
+// needed for karma-typescript bundling
```
gabrocheleau, line 8: move comments up above import { Buffer }

gabrocheleau (commented):
Great work overall, really satisfying to see the monorepo get tidied up :) Left a couple comments

packages/block/test/block.spec.ts
```diff
@@ -1,18 +1,20 @@
...
+import util from 'util'
+
+import { Chain, Common, Hardfork } from '@ethereumjs/common'
 import { NestedUint8Array, toBuffer, zeros } from '@ethereumjs/util'
 import { RLP } from 'rlp'
-import { Chain, Common, Hardfork } from '@ethereumjs/common'
+import * as tape from 'tape'
+
+// explicitly import util, needed for karma-typescript bundling
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
gabrocheleau, line 10: Move this up

config/eslint.js
gabrocheleau: Yeah I can totally get behind the preventative approach, I guess my main concern is I'd like to make sure that the added rules are relevant, mostly because I don't currently have a clear view on why some of these rules were added to start with.

I think it's worth thinking about what our criteria for adding new rules should be. Some examples that come to mind:

-  we are adding `'import/default'` (which ensures that default imports are coupled with exports) but not `'import/named'` (which ensures that named imports are coupled with exports). 
- `group-exports` and `exports-last`, is this something we actually want? I tend to lean in favor, but it requires a non-trivial amount of work, and I wouldn't consider it a no-brainer.
- `no-restricted-paths`, from what I understand this is only useful in the context where "restricted zones" are defined, but I don't think we have those

I've ran into situations, on other repos, where an excessive amount of resource-intensive linting rules led to performance issues for some contributors, hence my preoccupation with ensuring that additional rules serve a useful purpose.

Tagging the rest of the team in case they'd like to voice an opinion on that. @holgerd77 @acolytec3 @jochem-brouwer @ScottyPoi @g11tech

gabrocheleau:
> @gabrocheleau think I addressed all of them.

Looks good, thanks for the quick fixes! I'll wait for additional input from some other team members on the discussion above, and give it another review later this week :)

config/eslint.js
acolytec3: I don't have past experience with complex linting setups beyond the monorepo so can't comment on performance, but I don't notice any differences between running the linter on master vs this PR on my VSCode setup.

As far as individual rules in this set go, the only one we'll have to revisit is the `import/extensions` if/when we add ESM support.  Otherwise, this list looks good to me and I like having all the imports organized.

gabrocheleau: Thanks guys for the additional comments! Looks great @faustbrian, I'm on board with the remaining eslint rules! Good to know that there are solutions if we ever run into performance issues.

acolytec3:
@holgerd77 Any thoughts on this before we move forward?

config/eslint.js
```diff
@@ -46,7 +57,32 @@ module.exports = {
     '@typescript-eslint/no-redeclare': ['error'],
     '@typescript-eslint/restrict-plus-operands': 'off',
     'import/no-default-export' : ['error'],
-    '@typescript-eslint/strict-boolean-expressions': ['error']
+    '@typescript-eslint/strict-boolean-expressions': ['error'],
+    // 'simple-import-sort/imports': 'error',
```
gabrocheleau, line 61: Instead of commenting out, I'd suggest we just remove it and add it to our linting issue/discussion.

packages/vm/tests/api/index.spec.ts
```diff
@@ -1,20 +1,19 @@
-import * as tape from 'tape'
-import { Account, Address, KECCAK256_RLP } from '@ethereumjs/util'
+// explicitly import util and buffer,
+// needed for karma-typescript bundling
+import { Buffer } from 'buffer' // eslint-disable-line @typescript-eslint/no-unused-vars
```
gabrocheleau, line 3: I think the eslint-disable might not be necessary for Buffer in this instance, as Buffer is used in the file.

gabrocheleau (approved):
Approving, looks good to me. Left two minor comments.

Let's still wait for @holgerd77 's input before merging however.

gabrocheleau (approved):

holgerd77 (approved):
Ok. 🙂

PR #2087, opened 2022-08-01
common: set the default hard fork to merge

### LINKED ISSUE #2015: Monorepo: Make Merge Default Hardfork
Since we are now getting so close to the Merge HF date with our breaking releases (or might be even do *after* the Merge, we'll see) I would be a big fan of directly making the Merge HF the default HF on the libraries.

There would be a very very tiny window of where we are "ahead" (and even this wouldn't be bad), and otherwise we would be aligned with the current HF during very much most of the time of the release lifecycle. 🙂 

Inspiration and guidance for update locations can be taken from the London HF work #1749 done by @emersonmacro. I would be very much expecting though that this upgrade work is not as extensive any more (but still some work to do), since Emmett hardened e.g. various code parts like test initializations along the way (by explicitly providing a certain HF e.g.).

This doesn't need to be finished before Beta 2 but can e.g. be done if someone gets bored while I am and other are on holidays e.g.. 😅

g11tech:
set the default hard fork to merge

Closes #2015

### DIFF
--- a/packages/block/src/types.ts
+++ b/packages/block/src/types.ts
@@ -18,7 +18,7 @@ export interface BlockOptions {
    * Default: {@link Common} object set to `mainnet` and the HF currently defined as the default
    * hardfork in the {@link Common} class.
    *
-   * Current default hardfork: `london`
+   * Current default hardfork: `merge`
    */
   common?: Common
   /**
--- a/packages/client/test/integration/merge.spec.ts
+++ b/packages/client/test/integration/merge.spec.ts
@@ -37,7 +37,7 @@ tape('[Integration:Merge]', async (t) => {
         },
       ],
     },
-    { baseChain: ChainCommon.Goerli }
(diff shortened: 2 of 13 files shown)


gabrocheleau:
Looks good! However some merged-related tests are now failing in `merge.spec.ts`:
- 'should mine and stop at the merge (PoA)'
- 'should mine and stop at the merge (PoW)'

Didn't check in detail but perhaps all that's required is to manually provide a pre-merge HF for those test cases

g11tech:
thanks @acolytec3 for fixes, got distracted after pushing initial PR :wink:

gabrocheleau (dismissed):
LGTM! 👍🏼

gabrocheleau (approved):

holgerd77:
Nice, cool, didn't assume this to be so narrowly scoped respectively going through with so few changes. 😀 🤩 Seems Emmett already did a lot of base work here with adding the explicit HF (Common) instantiations in the round before upgrading to London.

Great!

PR #2091, opened 2022-08-02
Make the client browser implementation work again...

acolytec3:
...since it doesn't right now :-(


acolytec3:
Philosophical question on this, I've got the client to where it will build again and start in the browser. If you start it in light sync mode, it can connect to a peer but I'm not able to retrieve block headers at the moment. I'm not really sure why it's not syncing at the moment and I'm wondering how much more effort I should put towards this. Any thoughts?

holgerd77:
> Philosophical question on this, I've got the client to where it will build again and start in the browser. If you start it in light sync mode, it can connect to a peer but I'm not able to retrieve block headers at the moment. I'm not really sure why it's not syncing at the moment and I'm wondering how much more effort I should put towards this. Any thoughts?

I'd say if we have one sync mode - somewhat - working that would be enough for now.

acolytec3:
> > Philosophical question on this, I've got the client to where it will build again and start in the browser. If you start it in light sync mode, it can connect to a peer but I'm not able to retrieve block headers at the moment. I'm not really sure why it's not syncing at the moment and I'm wondering how much more effort I should put towards this. Any thoughts?
> 
> I'd say if we have one sync mode - somewhat - working that would be enough for now.

Okay, cool.  I wouldn't call it "working" yet since it's not actually fetching headers but I think we're not too far off.

g11tech:
success!!!! :tada: 

I think the PR is good to go!
image

acolytec3:
Can anyone else confirm they're able to get the browser client to sync headers?  I'm not able to make it work and still get some consensus validation errors but if everyone else can make it work, I say we go ahead and merge.  Full sync still doesn't work due to some unexplained polyfill issue I haven't been able to identify but it in theory at least sorta works again.

g11tech (approved):
Approving as my runs were with Sepolia (and look into consensus issue as a separate fix)

holgerd77:
Have admin-merged since codecov was not reporting here (not sure if this wil happen again or was a one-time thing, we'll see).

PR #2093, opened 2022-08-03
chore: sort `package.json` files

faustbrian:
Uses https://github.com/keithamus/sort-package-json to sort every `package.json` file the same way to make it easier to dig through them. Can be triggered by running `npm run sort-package-json` in the root directory.


holgerd77 (approved):
LGTM!

One small note: please always be a bit careful with dev dependencies and e.g. do not add dependencies with a too low level of stars/engagement/activity. I am always a bit afraid of some repo getting abandoned and then taken over by someone with malicious intentions.

In this case it should still be ok I guess, but from my feeling some dependency with stars in the hundreds (rather than the thousands) is already the lowest limit which should/can be accepted.

holgerd77:
Oh, how annoying, you are right. Not sure how this happened, normally the "Squash and Merge" option is show directly upfront? 🤔

Anyhow, I won't do anything about it. These were (luckily) only three commits, and too dangerous to mess up with `master` for this.

PR #2094, opened 2022-08-03
test(trie): run persistence suite for all types of tries

faustbrian:


acolytec3 (changes requested):
See my comment on line33.  If @holgerd77 agrees, we should apply this logic everywhere in this file `tape("some test", (t) => { (st) => { st.end() } t.end() }`

acolytec3:
These reworked tests look great though, apart from my stylistic nit.

packages/trie/test/trie/trie.spec.ts
```diff
@@ -1,104 +1,135 @@
...
+]) {
+  let dbTmpDir: string
+  try {
+    dbTmpDir = mkdtempSync(join(tmpdir(), title))
+  } catch (error) {
+    dbTmpDir = tmpdir()
+  }
+
+  tape(`${title} (Persistence)`, function ({ test }) {
+    test('creates an instance via the static constructor `create` function and defaults to `true` with a database', async function (assert) {
```
acolytec3: Sorry, this is what I was talking about.  For `tape` tests, our naming pattern across the monorepo is `tape( "some test", (t) => { (st) => { st.end() } t.end() }`.  I don't like calling subtests `assert` since that's too easily confused with the nodejs builtin.  Our usual pattern would be like below.  Can you apply this pattern (or something similar) across this test suite?
```suggestion
    t('creates an instance via the static constructor `create` function and defaults to `true` with a database', async (st) => {
```

acolytec3: Great question.  If there is, it's woefully out of date as I certainly wasn't aware of anything when I started with the team.  I'll add a new issue for us to draft/update our contributor documentation as it's not something we've maintained (mostly since we've not had many outside contributors of late so yours are much much appreciated!).

acolytec3 (dismissed):
LGTM

acolytec3:
Looks great, thanks for the updates. Will approve

acolytec3 (approved):
Looks great as always!

PR #2096, opened 2022-08-03
client: Run execution in batches when not near head

g11tech:
Run execution in batches when not near head

### DIFF
--- a/packages/client/lib/sync/fetcher/blockfetcher.ts
+++ b/packages/client/lib/sync/fetcher/blockfetcher.ts
@@ -48,7 +48,7 @@ export class BlockFetcher extends BlockFetcherBase<Block[], Block> {
     }
     const headers = headersResult[1]
     const bodiesResult = await peer!.eth!.getBlockBodies({ hashes: headers.map((h) => h.hash()) })
-    if (isFalsy(bodiesResult) || bodiesResult[1].length === 0) {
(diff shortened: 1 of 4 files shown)


acolytec3 (commented):
LGTM!

acolytec3 (approved):

acolytec3 (approved):
still lgtm

packages/client/lib/sync/fetcher/blockfetcher.ts
```diff
@@ -48,7 +48,7 @@ export class BlockFetcher extends BlockFetcherBase<Block[], Block> {
     }
     const headers = headersResult[1]
     const bodiesResult = await peer!.eth!.getBlockBodies({ hashes: headers.map((h) => h.hash()) })
-    if (isFalsy(bodiesResult) || bodiesResult[1].length === 0) {
+    if (isFalsy(bodiesResult) || isFalsy(bodiesResult[1]) || bodiesResult[1].length === 0) {
```
holgerd77, line 51: Can we try not to use respectively even introduce these `isFalsy` (and friends 🙂) methods but directly go to the explicit version?

This was really intended to just be a placeholder until we do better.

holgerd77: (or is the whole concept not working/practicable and we would need to revise the linting rule update at this point?)

PR #2097, opened 2022-08-03
chore: add scripts for dependency updates

faustbrian:
Adds 3 scripts that use https://github.com/raineorshine/npm-check-updates to update dependencies to to various semantic version points. I'll use those in some upcoming PRs and maybe we could even set up a weekly workflow that runs the `minor` script?

### DIFF
--- a/packages/client/lib/execution/vmexecution.ts
+++ b/packages/client/lib/execution/vmexecution.ts
@@ -145,7 +145,7 @@ export class VMExecution extends Execution {
    * @param loop Whether to continue iterating until vm head equals chain head (default: true)
    * @returns number of blocks executed
    */
-  async run(loop = true): Promise<number> {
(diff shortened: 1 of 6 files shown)


paulmillr:
What is the point in having dependency versions? You can just put `*`. 

I think dependencies should be upgraded rarely, like, once per quarter. This is what i'm doing in personal repos. I'm locking down **all** dependencies including `devDeps`. The more you mess with it, the more the chance of breakage. Better leave it alone.

Even more than that: any package upgrade could bring some malware in. Because you don't review the deps code manually.

holgerd77:
I have also a bit the fear that this is getting too broad and will eventually be misused by entice people to do updates all over the board without checking enough.

Can we settle for now that you (or whoever needs) use these scripts locally and we do not merge in?

PR #2103, opened 2022-08-04
refactor(trie): remove unnecessary temp variables and buffer creation

faustbrian:
Removes unnecessary calls to `Buffer.from` and also some temporary variables that seem unnecessary

PR #2104, opened 2022-08-05
ci: retry `test-client` and `test-client-cli` jobs up to 3 times

faustbrian:
As mentioned in https://github.com/ethereumjs/ethereumjs-monorepo/pull/2103 I'm adding a retry of up to 3 times for the `test-client-cli` jobs that randomly fail


acolytec3:
Can you also add the same logic to the other two client builds for goerli light and full sync?

acolytec3 (approved):
Perfect, looks great, thanks!

holgerd77:
Thanks a lot, this is one of these small but super-super valuable and everyday-life-easing improvements! 🙏

(never thought about doing something about it thougn, lol)

PR #2109, opened 2022-08-07
chore(trie): add database example scripts

faustbrian:
Adds example scripts for the default `level` implementation and also an implementation with https://github.com/DoctorEvidence/lmdb-js to illustrate a different implementation


holgerd77 (approved):
Nice, LGTM! 🙂 👍

PR #2114, opened 2022-08-07
chore: update `husky` to `8.0.0`

faustbrian:
Updates `husky` from the v4 major release to the v8 major release. The biggest difference is that tasks are now stored in bash files rather than the `package.json` file


holgerd77:
Hi @faustbrian short nit, unrelated to this specific PR: can you please always label with one or more `package:*` labels? Rule is somewhat that every library which has been touched in a non-trivial way should be labeled. Generic all-library touching PRs like this one (mostly CI PRs) should get a `package: monorepo` label.

I need these labels for doing the release rounds and otherwise I will forget things to include into the release notes or even forget complete releases, so I am otherwise always doing a relatively laborious pre-releases labelling round for all missing PRs. 😋

Thanks a lot!
Holger

holgerd77:
(so "trivial" would mean e.g.: some typos corrected, smaller code clean-ups, things like that)

PR #2116, opened 2022-08-07
VM, EVM, Common: Implement EIP-2330 Extsload Opcode

zemse:
This PR implements [EIP-2330](https://eips.ethereum.org/EIPS/eip-2330) Extsload at 0x5c opcode.

If the opcode is supported as an experimental opcode, it would help with easier experimentation and research on the EIP.

Currently, there is a clash at the same opcode 0x5c `BEGINSUB` from [EIP-2315](https://eips.ethereum.org/EIPS/eip-2315), hence currently either EIP could be activated at a time. I'll discuss with EIP-2330 authors if there can be another candidate for the opcode number.

Please let me know of any changes that are needed.

### DIFF
--- a/packages/evm/src/opcodes/functions.ts
+++ b/packages/evm/src/opcodes/functions.ts
@@ -766,11 +766,24 @@ export const handlers: Map<number, OpHandler> = new Map([
   ],
   // 0x5b: JUMPDEST
   [0x5b, function () {}],
-  // 0x5c: BEGINSUB
+  // 0x5c: BEGINSUB or EXTSLOAD
   [
     0x5c,
-    function (runState) {
-      trap(ERROR.INVALID_BEGINSUB + ' at ' + describeLocation(runState))
+    function (runState, common) {
+      if (common.isActivatedEIP(2315)) {
+        // BEGINSUB is sync
+        trap(ERROR.INVALID_BEGINSUB + ' at ' + describeLocation(runState))
+      } else if (common.isActivatedEIP(2330)) {
+        // EXTSLOAD is async
+        const addressBigInt = runState.stack.pop()
+        const address = new Address(addressToBuffer(addressBigInt))
+        const keyBigInt = runState.stack.pop()
+        const key = setLengthLeft(bigIntToBuffer(keyBigInt), 32)
+        return runState.interpreter.externalStorageLoad(address, key).then((value) => {
+          const valueBigInt = value.length ? bufferToBigInt(value) : BigInt(0)
+          runState.stack.push(valueBigInt)
+        })
+      }
     },
   ],
   // 0x5d: RETURNSUB
--- a/packages/evm/src/opcodes/gas.ts
+++ b/packages/evm/src/opcodes/gas.ts
@@ -199,7 +199,13 @@ export const dynamicGasHandlers: Map<number, AsyncDynamicGasHandler | SyncDynami
         const keyBuf = setLengthLeft(bigIntToBuffer(key), 32)
 
         if (common.isActivatedEIP(2929) === true) {
-          gas += accessStorageEIP2929(runState, keyBuf, false, common)
+          gas += accessStorageEIP2929(
+            runState,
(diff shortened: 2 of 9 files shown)


holgerd77:
Hi Soham,
that's pretty impressive work here, cool! 👍 

I guess we need some discussion if we want to integrate and if the EIP is relevant and up-to-date enough to integrate (or bear with me if we had this discussion already? I sometimes forget these kind of things.)

Thing is that we can't integrate all sorts of EIPs into the main code base, otherwise our code base would get too fragmented.

jochem-brouwer:
Very cool zemse! I have checked the EIP and I think the EIP is rather outdated. For instance, it does not mention gas costs like how much gas it would cost if one tries to extsload a cold account / cold slot (See EIP 2930 for warm/cold definitions). But on the other hand it now matches the spec, so I guess that is OK.

packages/evm/src/opcodes/functions.ts
```diff
@@ -766,11 +766,24 @@ export const handlers: Map<number, OpHandler> = new Map([
...
+  // 0x5c: BEGINSUB or EXTSLOAD
   [
     0x5c,
-    function (runState) {
-      trap(ERROR.INVALID_BEGINSUB + ' at ' + describeLocation(runState))
+    function (runState, common) {
+      if (common.isActivatedEIP(2315)) {
+        // BEGINSUB is sync
+        trap(ERROR.INVALID_BEGINSUB + ' at ' + describeLocation(runState))
+      } else if (common.isActivatedEIP(2330)) {
```
jochem-brouwer, line 776: Hmm I'm not sure maybe we can do this in another way (no if/else branch), like inject them in `codes.ts`, the same way as how one would inject custom opcodes..?

packages/evm/src/opcodes/gas.ts
```diff
@@ -251,7 +257,27 @@ export const dynamicGasHandlers: Map<number, AsyncDynamicGasHandler | SyncDynami
...
+    ],
+    [
+      /* EXTSLOAD */
+      0x5c,
+      async function (runState, gas, common): Promise<bigint> {
+        if (common.isActivatedEIP(2330) && common.isActivatedEIP(2929)) {
+          const [addressBigInt, keyBigInt] = runState.stack.peek(2)
+          const address = new Address(addressToBuffer(addressBigInt))
+          const key = setLengthLeft(bigIntToBuffer(keyBigInt), 32)
+          gas += accessAddressEIP2929(runState, address, common)
```
jochem-brouwer, line 279: I'm not sure here, this gas schedule does not seem to match the EIP?

packages/evm/src/opcodes/functions.ts
zemse: I agree the if-else doesn't look good here. I will explore moving logic in `codes.ts`. If you have an example can you please share?

zemse:
Just wanted to add to the conversation here that this EIP was recently updated. Please view the latest version here: https://eips.ethereum.org/EIPS/eip-2330. Also, some community members see the value from EIP (e.g. [discord link](https://discord.com/channels/595666850260713488/706868829900505180/921827809763155998)).

packages/evm/src/opcodes/functions.ts
jochem-brouwer: I think we might want to leverage this logic of "clashing opcodes" into `codes.ts` (we might also have this at some point with precompiles). The API is not (directly) ready for this, but https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/evm/tests/customOpcodes.spec.ts these tests might point a bit in the right direction.

packages/evm/src/opcodes/gas.ts
jochem-brouwer: Oh that's really great! :+1:

holgerd77:
Since it is pretty unlikely that we will take this in (EIP fully stagnant and not discussed/considered), I will close here!

Thanks @zemse again for the interesting experiment!

PR #2119, opened 2022-08-09
client: snap protocol tests

scorbajio:
This PR adds tests for `snapprotocol` and `snapsync`.

### DIFF
--- a/packages/client/test/net/protocol/snapprotocol.spec.ts
+++ b/packages/client/test/net/protocol/snapprotocol.spec.ts
@@ -1,0 +1,175 @@
+import * as tape from 'tape'
+
+import { Account, bigIntToBuffer } from '@ethereumjs/util'
+import { Chain } from '../../../lib/blockchain'
+import { Config } from '../../../lib/config'
+import { SnapProtocol } from '../../../lib/net/protocol'
+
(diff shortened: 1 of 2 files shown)


g11tech (changes requested):
@scorbajio i would love to see some logs of the other type of requests, and then encoding/decoding compared against some actual sample.

scorbajio:
> @scorbajio i would love to see some logs of the other requests, and then encoding/decoding compared against some actual sample. Right now using our conver encoder => decoder sort of is self reinforcing, so would be awesome to test them against some actual sample data received from a geth/nethermind or other snap capable peer.

I'm not sure I understand what sample data you're referring to for testing encoding/decoding. Could you explain or give an example?

g11tech:
> > @scorbajio i would love to see some logs of the other requests, and then encoding/decoding compared against some actual sample. Right now using our conver encoder => decoder sort of is self reinforcing, so would be awesome to test them against some actual sample data received from a geth/nethermind or other snap capable peer.
> 
> I'm not sure I understand what sample data you're referring to for testing encoding/decoding. Could you explain or give an example?

ok, i will add the same on top of these tests, for now we can merge these ones

g11tech (approved):
LGTM! will add more tests on top regarding encoding decoding with real data

PR #2107, opened 2022-08-06
client: create snap sync fetcher to sync accounts

scorbajio:
AccountFetcher
Capture data now done/part of  
- https://github.com/ethereumjs/ethereumjs-monorepo/pull/2157  
    -  `getAccountRange`
    - `getStorageRanges`
    - `getByteCodes`
    - `getTrieNodes`

My understanding is that there needs to be one fetcher per each `get` call defined in the snap specs. Open to suggestions on this design plan.

PR #2120, opened 2022-08-09
client: Make fetcher typesafe in syncronizers

g11tech:
With the new fetchers for AccountRange etc in the snapsync PR https://github.com/ethereumjs/ethereumjs-monorepo/pull/2107, we need fetcher access to be typesafe in their respective synchronizers.

This PR enables the same.

### DIFF
--- a/packages/client/lib/sync/beaconsync.ts
+++ b/packages/client/lib/sync/beaconsync.ts
@@ -44,7 +44,7 @@ export class BeaconSynchronizer extends Synchronizer {
     return 'beacon'
   }
 
-  get fetcher():ReverseBlockFetcher|null{
+  get fetcher(): ReverseBlockFetcher | null {
     if(this._fetcher!==null && !(this._fetcher instanceof ReverseBlockFetcher)){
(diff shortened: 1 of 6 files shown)


packages/client/lib/sync/beaconsync.ts
```diff
@@ -44,6 +44,17 @@ export class BeaconSynchronizer extends Synchronizer {
     return 'beacon'
   }
 
+  get fetcher():ReverseBlockFetcher|null{
```
acolytec3, line 47: What's going on with the spacing in here?  Is `prettier` misconfigured or something?

acolytec3 (dismissed):
This generally looks good so just need to clean up the spacing before merging.

packages/client/lib/sync/beaconsync.ts
g11tech: hmmm, linter isn't catching/fixing it, added manually

holgerd77 (approved):
LGTM, linting seems to pass now, will merge!

Issue #2074, opened 2022-07-25
Update `trie` usage/readme


acolytec3:
With the breaking changes, we're making some fairly substantial updates to the `trie` package so we should go over the docs/readme in detail and make sure everything is up to date once #2071 is merged.

helderjnpinto:
Yes, i needed to check tests to get the things working, it would be great to have a readme with some examples fully working.

PR #2121, opened 2022-08-10
docs: update README formatting

faustbrian:
Reorganises a few bits, adds information about the root hash persistence and updates all information to be compatible with the 5.0.0 release. Should resolve https://github.com/ethereumjs/ethereumjs-monorepo/issues/2074


packages/trie/README.md
```diff
@@ -6,23 +6,27 @@
...
 
-> The modified Merkle Patricia tree (trie) provides a persistent data structure to map between arbitrary-length binary data (byte arrays). It is defined in terms of a mutable data structure to map between 256-bit binary fragments and arbitrary-length binary data. The core of the trie, and its sole requirement in terms of the protocol specification is to provide a single 32-byte value that identifies a given set of key-value pairs.
+> The modified Merkle Patricia tree (trie) provides a persistent data structure to map between arbitrary-length binary data (byte arrays). It is defined in terms of a mutable data structure to map between 256-bit binary fragments and arbitrary-length binary data. The core of the trie, and its sole requirement in terms of the protocol specification, is to provide a single 32-byte value that identifies a given set of key-value pairs.
 
-The only backing store supported is LevelDB through the `levelup` module.
+## Installation
 
-# INSTALL
```
holgerd77, line 15: This is the header naming scheme in (more or less) all our other packages, and I wouldn't want to change this for one package here along the sidelines.

holgerd77: So this would be for all before-level-1-after-PR-level2 headlines I guess. Also this level changing get's really rather even more inconsistent if we only change on one package.

So I am open to whatever formatting you come up with and things like capitalization or not or whatever, but we should do across the reops.

holgerd77: (if you want to do I would be also fine to do this within this PR)

packages/trie/README.md
```diff
@@ -56,17 +60,39 @@ import { Level } from 'level'
...
+  persistRoot: false,
+})
+```
 
 ## Proofs
 
 ### Merkle Proofs
 
-The `createProof` and `verifyProof` functions allow you to verify that a certain value does or does not exist within a Merkle-Patricia trie with a given root.
+The `createProof` and `verifyProof` functions allow you to verify that a certain value does or does not exist within a Merkle Patricia trie with a given root.
```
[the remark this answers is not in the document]
holgerd77: Yes, again - as you already suggested not trie in this context but `Merkle Patricia Tree`, the other version sneaked in.

packages/trie/README.md
```diff
@@ -181,76 +208,89 @@ async function test() {
...
 - Videos
   - [Trie and Patricia Trie Overview](https://www.youtube.com/watch?v=jXAHLqQthKw&t=26s)
 
-# EthereumJS
-
-See our organizational [documentation](https://ethereumjs.readthedocs.io) for an introduction to `EthereumJS` as well as information on current standards and best practices.
+## EthereumJS
 
-If you want to join for work or do improvements on the libraries have a look at our [contribution guidelines](https://ethereumjs.readthedocs.io/en/latest/contributing.html).
+See our organizational [documentation](https://ethereumjs.readthedocs.io) for an introduction to `EthereumJS` as well as information on current standards and best practices. If you want to join for work or carry out improvements on the libraries, please review our [contribution guidelines](https://ethereumjs.readthedocs.io/en/latest/contributing.html) first.
```
holgerd77, line 289: This is also a generic section below all READMEs for all packages, if you do changes here you should "broadcast" to the other packages as well.

holgerd77 (changes requested):
Nice, some comments to address and/or discuss.

packages/block/README.md
```diff
@@ -169,34 +173,31 @@ const block = Block.fromBlockData(
...
-If you want to join for work or do improvements on the libraries have a look at our [contribution guidelines](https://ethereumjs.readthedocs.io/en/latest/contributing.html).
-
-# LICENSE
+## License
 
 [MPL-2.0](<https://tldrlegal.com/license/mozilla-public-license-2.0-(mpl-2)>)
-
 [discord-badge]: https://img.shields.io/static/v1?logo=discord&label=discord&message=Join&color=blue
 [discord-link]: https://discord.gg/TNwARpR
 [block-npm-badge]: https://img.shields.io/npm/v/@ethereumjs/block.svg
```
holgerd77, line 203: Something is broken in this file and the badge links don't render correctly.

holgerd77: Update: seems this is the case in other files as well, e.g. Blockchain (I would recommend a markdown review plugin for Visual Studio Code (if you use that one)).

holgerd77 (changes requested):
Markdown is not rendering correctly, will leave it here till fixed.

holgerd77 (approved):
Nice, thanks for the quick update - yes - this looks a lot better now! 🙏

Will merge.

holgerd77:
Ok, I'll be hitting "Squash and merge" here...

holgerd77:
Seems to have worked: https://github.com/ethereumjs/ethereumjs-monorepo/commits/master 🙂

(the title should always be taken from the PR title, if that was the case? 🤔 So: in the case of this PR, it is)

PR #2123, opened 2022-08-10
refactor(trie): default to `false` for `persistRoot` option

faustbrian:
See discussion in https://github.com/ethereumjs/ethereumjs-monorepo/pull/2071

PR #2124, opened 2022-08-10
Revert "refactor(trie): default to `false` for `persistRoot` option"

gabrocheleau:
Reverts the package-lock.json changes from ethereumjs/ethereumjs-monorepo#2123. Let's keep such changes separately scoped (i.e. in their own PR) so that they can be properly reviewed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,10 +56,9 @@
       }
     },
     "node_modules/@babel/compat-data": {
-      "version": "7.18.8",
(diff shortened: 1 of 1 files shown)


holgerd77:
We now had several PRs with a `package-lock.json` included and if we now revert this last one this would feel a bit arbitrary to me. But from the cause you are bringing up the table you definitely have a point, we have no clear and practical procedure yet how best to deal with `package-lock.json` updates.

PR #2127, opened 2022-08-12
fix(trie): explicitly check for `persistRoot` option being `true` in `SecureTrie`

faustbrian:
Small issue I noticed when upgrading to `beta.3` and a test with key count failing on my end. The `Trie` class already had the correct check https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/trie/src/trie/trie.ts#L752

### DIFF
--- a/packages/trie/src/trie/secure.ts
+++ b/packages/trie/src/trie/secure.ts
@@ -119,7 +119,7 @@ export class SecureTrie extends CheckpointTrie {
    * Persists the root hash in the underlying database
    */
   async persistRoot() {
-    if (this._persistRoot !== undefined) {
+    if (this._persistRoot === true) {
       await this.db.put(this.hash(ROOT_DB_KEY), this.root)
     }
   }


packages/trie/src/trie/secure.ts
```diff
@@ -119,7 +119,7 @@ export class SecureTrie extends CheckpointTrie {
    * Persists the root hash in the underlying database
    */
   async persistRoot() {
-    if (this._persistRoot !== undefined) {
+    if (this._persistRoot === true) {
```
holgerd77, line 122: Sorry, side topic: with the linting I would actually personally prefer if we would allow `if (this._persistRoot)` or similar for all boolean values (since there it is just a binary choice anyhow and the short version is just more elegant, and only enforce an explicit comparison 2nd value for all non-boolean values.

Would this be possible? 🤔 //cc @gabrocheleau

holgerd77 (approved):
LGTM, thanks! 🙂

packages/trie/src/trie/secure.ts
holgerd77: Ah ok, that's true. Meanwhile often forget about JS completely. 😬 (which is also not that good, we should really - also - keep to have the JS users in mind, so in the sense of JS Developers, and at the end every *end* user is a JS user anyhow).

holgerd77:
> Looks like this PR was again merged without using the PR title. Will keep that in mind in the future and squash everything to the PR title 😅

Ah, that's also not the solution, since for more complex PRs we need the separated commits to ease the reviews.

I have still not clue why this is happening though, I am 100% sure that I didn't merge in any other way here than on any other PR.

PR #2129, opened 2022-08-12
Upgrade jayson to v4.0

acolytec3:
Upgrades the `jayson` client to the recently released version 4.0 which cuts that dep's size in half and removes some transitive dependencies in the process

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -78,7 +78,7 @@
     "fs-extra": "^10.0.0",
     "it-pipe": "^1.1.0",
     "it-pushable": "^1.4.2",
-    "jayson": "^3.4.4",
+    "jayson": "^4.0.0",
     "jwt-simple": "^0.5.6",
     "level": "^8.0.0",
     "libp2p": "^0.30.7",


g11tech (approved):
LGTM!

holgerd77:
Oh, that's a great step ("cuts that dep's size in half")! 🙂 ❤️ I was already wondering if we would need to remove (replace) the logger completely (as it was already brought on the table by Paul Rumkin 1-2 years ago where he already mentiond the large number of dependencies).

acolytec3:
> Oh, that's a great step ("cuts that dep's size in half")! 🙂 ❤️ I was already wondering if we would need to remove (replace) the logger completely (as it was already brought on the table by Paul Rumkin 1-2 years ago where he already mentiond the large number of dependencies).

I'm not sure about the logger but it may be feasible to just use the Winston one. Will need to research further

holgerd77:
Ah, sorry, I was confused, yes, Jayson is the RPC client, right? Having this confused with the logger, yes, was referencing to Winston with the post above.

PR #2128, opened 2022-08-12
refactor(trie): use `micro-bmark` for benchmarks

faustbrian:
For a smaller dependency footprint you could use https://github.com/paulmillr/micro-bmark and https://github.com/bestiejs/benchmark.js is also unmaintained for good so might be time to drop it anyway. Mostly a suggestion so can close it right away if you don't want it 😅

```
❯ npm run benchmarks

> @ethereumjs/trie@5.0.0-beta.3 benchmarks
> node -r ts-node/register --max-old-space-size=8024 benchmarks

-------
Benchmarking
1k-3-32-ran x 4 ops/sec @ 232ms/op
1k-5-32-ran x 4 ops/sec @ 201ms/op
1k-9-32-ran x 5 ops/sec @ 197ms/op
1k-1k-32-ran x 4 ops/sec @ 203ms/op
1k-1k-32-mir x 4 ops/sec @ 218ms/op
Checkpointing: 100 iterations x 6,378 ops/sec @ 156μs/op ± 11.16% (min: 79μs, max: 702μs)
Checkpointing: 500 iterations x 5,583 ops/sec @ 179μs/op ± 5.18% (min: 117μs, max: 1ms)
Checkpointing: 1000 iterations x 5,398 ops/sec @ 185μs/op ± 2.73% (min: 122μs, max: 928μs)
Checkpointing: 5000 iterations x 5,119 ops/sec @ 195μs/op
RAM: rss=355.3mb heap=209.7mb used=185.4mb ext=42.6mb arr=41.4mb
```

PR #2132, opened 2022-08-15
chore: enable `consistent-type-imports` rule for `eslint`

faustbrian:
Enables the `@typescript-eslint/consistent-type-imports` rule mentioned in https://github.com/ethereumjs/ethereumjs-monorepo/issues/1935


holgerd77:
Yeah, that's a nice one too. I would first merge #2128, then update here and merge.

So the PR is ready, right?

holgerd77 (approved):
LGTM

PR #2133, opened 2022-08-15
chore: enable `strict-boolean-expressions` rule for `eslint`

faustbrian:
Enable the `@typescript-eslint/strict-boolean-expressions` rule mentioned in https://github.com/ethereumjs/ethereumjs-monorepo/issues/1935. As mentioned in https://github.com/ethereumjs/ethereumjs-monorepo/issues/1935#issuecomment-1214628861 enabling this feels quite sketchy because of various factors. The main one is the mixed use of `null` and `undefined` which means you can never be sure what you're handling and need those hacky `isTruthy` and `isFalsy` functions to have any kind of insurance that everything is handled.

Opened this mainly to show the amount of changes this rule causes. It might be best to introduce this rule package-by-package rather than repository-wide and then when every package is done enable it repository-wide.


holgerd77:
Oh, yes, this is a hard one. Likely worth to work through this step-by-step though.

PR #2138, opened 2022-08-15
Adjust `skipBalance` logic

acolytec3:
Updates changes in #1849 with some recommendations from hardhat.  Primarily, instead of just adding the tx cost or message value to the `caller` balance when `skipBalance` is set, we now only touch the balance if it's less than the transaction's cost (i.e. gasLimit *gasPrice + tx.value or maxFeePerGas * gasLimit + value if an EIP 1559 tx).

### DIFF
--- a/packages/evm/src/evm.ts
+++ b/packages/evm/src/evm.ts
@@ -699,10 +699,12 @@ export class EVM extends AsyncEventEmitter<EVMEvents> implements EVMInterface {
 
       const value = opts.value ?? BigInt(0)
       if (opts.skipBalance === true) {
-        // if skipBalance, add `value` to caller balance to ensure sufficient funds
         const callerAccount = await this.eei.getAccount(caller)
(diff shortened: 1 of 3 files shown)


jochem-brouwer:
2 notes:

(1) this PR should be rebased/`merge` master (there is a conflict)

(2) I think we should add documentation about the behavior of `skipBalance`. The current documentation does not really explain what is going on;


https://github.com/ethereumjs/ethereumjs-monorepo/blob/c4b7c91c758ee13e06d7558b8da5d2b94dd31875/packages/vm/src/types.ts#L234-L237

https://github.com/ethereumjs/ethereumjs-monorepo/blob/c4b7c91c758ee13e06d7558b8da5d2b94dd31875/packages/evm/src/types.ts#L128-L131

acolytec3:
Closing and replacing with #2172

PR #2139, opened 2022-08-15
block: add new consensusFormatValidation flag

### LINKED ISSUE #2135: Block: Add flag for consensus-specific format validation to Block Header
As pointed out by @fvictorio from Hardhat it would be good if there would be an option to enable/disable the [consensusFormatValidation()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/header.ts#L305) check in the BlockHeader constructor.

By default this option should be set to `true` (so keep the behavior as we have now), I would be inclined to name as the method, so `consensusFormatValidation`.

acolytec3:
Fixes #2135


packages/block/src/header.ts
```diff
@@ -26,6 +26,7 @@ import { CLIQUE_EXTRA_SEAL, CLIQUE_EXTRA_VANITY } from './clique'
 import type { BlockHeaderBuffer, BlockOptions, HeaderData, JsonHeader } from './types'
 import type { CliqueConfig } from '@ethereumjs/common'
 
+
```
gabrocheleau, line 29: Extra space

packages/block/src/types.ts
```diff
@@ -1,3 +1,4 @@
+
```
gabrocheleau, line 1: Extra space

packages/block/src/types.ts
```diff
@@ -69,6 +69,10 @@ export interface BlockOptions {
    * Will throw if provided on a non-PoA chain.
    */
   cliqueSigner?: Buffer
+  /**
+   * Perform consensus validation checks on header if set.  Defaults to true.
```
gabrocheleau, line 73:
```suggestion
   * Perform consensus validation checks on header if set. Defaults to true.
```

packages/block/test/header.spec.ts
```diff
@@ -290,6 +290,66 @@ tape('[Block]: Header functions', function (t) {
...
+      () => BlockHeader.fromHeaderData({ stateRoot: badHash }),
+      (err: any) => err.message.includes('stateRoot must be 32 bytes'),
+      'throws on invalid state root hash length'
+    )
+    st.throws(
+      () => BlockHeader.fromHeaderData({ transactionsTrie: badHash }),
+      (err: any) => err.message.includes('transactionsTrie must be 32 bytes'),
+      'throws on invalid transactionsTrie root hash length'
+    )
+    let nonce = Buffer.alloc(5)
```
gabrocheleau, line 328: Not a big deal, but I find the nonce reassignment a bit weird. Could we just either scope it differently or directly assign `fromHeaderData({nonce: Buffer.alloc(5)})` ?

gabrocheleau (approved):

gabrocheleau (approved):

holgerd77:
Just a short additional LGTM from my side as well, @acolytec3 thanks for addressing! 🙂

Issue #2137, opened 2022-08-15
Enable `@typescript-eslint/strict-boolean-expressions` rule


faustbrian:
Opening a separate issue from https://github.com/ethereumjs/ethereumjs-monorepo/issues/1935 for enabling the rule mentioned in the title. Keeping a checklist here to track what was done:

- [x] block
- [x] blockchain
- [x] client
- [x] common
- [x] devp2p
- [x] ethash
- [x] evm
- [x] rlp
- [x] statemanager
- [x] trie
- [x] tx
- [x] util
- [x] vm

PR #2144, opened 2022-08-16
chore(blockchain): enable `strict-boolean-expressions` rule for `eslint`

faustbrian:
Part of https://github.com/ethereumjs/ethereumjs-monorepo/issues/2137 and depends on https://github.com/ethereumjs/ethereumjs-monorepo/pull/2143 being merged

### DIFF
--- a/packages/blockchain/src/consensus/clique.ts
+++ b/packages/blockchain/src/consensus/clique.ts
@@ -221,7 +221,7 @@ export class CliqueConsensus implements Consensus {
     const limit = this.CLIQUE_SIGNER_HISTORY_BLOCK_LIMIT
     const blockSigners = this._cliqueLatestBlockSigners
     const lastBlockNumber = blockSigners[blockSigners.length - 1]?.[0]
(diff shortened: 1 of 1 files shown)


acolytec3 (approved):
LGTM.  General question, are you going to fold these PRs into the following ones since all of the previous package changes are included in each?

holgerd77:
> @acolytec3 yeah I want them merged in a specific order as I created them on top of each other branches. That's why all but 1 are marked as draft.

Where is the non-DRAFT one? 🙂

holgerd77:
> Used to be #2143 but since you want to tag the stable releases soon I would rather wait with those and move them to develop once that branch is updated. There just won't be enough time to test these PRs properly.

Ah, no, we won't use `develop` for quite some time any more. Our normal procedure is to just develop towards `master`, which always represents our latest state of the work. The `develop` branch was just an exception to prepare for the breaking releases (we should have named this `develop-v6` or something) and won't be used any time soon (we'll forget about *any* breaking changes in the next 4+ months 😋).

Issue #2100, opened 2022-08-04
Master tracker for snap sync


g11tech:
This is a master tracker for snap-sync PR/tasks

- [x] https://github.com/ethereumjs/ethereumjs-monorepo/pull/1883
  - [x] https://github.com/ethereumjs/ethereumjs-monorepo/pull/1893
- [x] https://github.com/ethereumjs/ethereumjs-monorepo/pull/1731
- [x] https://github.com/ethereumjs/ethereumjs-monorepo/pull/1897
  - [x] https://github.com/ethereumjs/ethereumjs-monorepo/pull/1933
  - [x] https://github.com/ethereumjs/ethereumjs-monorepo/pull/2119
- [x] Test call and verify other methods
  - [x] Test Call GetAccountRange: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2157
  - [x] Test Call GetByteCodes: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2157
  - [x] Test Call GetTrieNodes: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2282
- [ ] Build fetchers and just dump it into a mock MPT 
  - [x] GetAccountRange
      - [x] https://github.com/ethereumjs/ethereumjs-monorepo/pull/2070
      - [x] https://github.com/ethereumjs/ethereumjs-monorepo/pull/2107
  - [x] GetStorageRange: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2345
  - [x] GetByteCodes: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2602
  - [ ] GetTrieNodes: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2623
- [ ] Build MPT and successfully run move the chain forward on a normal eth sync
- [ ] Chart out behavior of non static peer and map next steps

cc @scorbajio @jochem-brouwer @holgerd77 @acolytec3

scorbajio:
#1897 is being continued at #2107.

scorbajio:
Closing this and continuing on #2874.

scorbajio:
Closing this and continuing on https://github.com/ethereumjs/ethereumjs-monorepo/issues/2874.

PR #2157, opened 2022-08-16
client: Enhance snap protocol spec tests

g11tech:
Add tests for responses, and validating with real sample data. 
Master tracker: https://github.com/ethereumjs/ethereumjs-monorepo/issues/2100
- [x] GetAccountRange/AccountRange
- [x] GetStorageRanges/StorageRanges
- [x] GetByteCodes/ByteCodes

Add Encoder/decoder tests for 
- [x] AccountRange
- [x] StorageRanges
- [x] ByteCodes

Will address  GetTrieNodes/TrieNodes  in a separate PR as it directly deal with MPT nodes, and needs a bit more digging. However all for a static MPT the methods verified by this PR should be sufficient.


packages/client/test/net/protocol/snapprotocol.spec.ts
```diff
@@ -82,7 +85,56 @@ tape('[SnapProtocol]', (t) => {
...
+    t.end()
+
+  })
+
+  t.test('GetAccountRange/AccountRange should verify real sample',async(t)=>{
+    const config = new Config({ transports: [] })
+    const chain = new Chain({ config })
+    const p = new SnapProtocol({ config, chain })
+    /* eslint-disable @typescript-eslint/no-use-before-define */
+    const reqData= RLP.decode(Buffer.from(getAccountRangeRLP,'hex'));
```
holgerd77, line 110: @faustbrian why are these semicolon additions not caught by linting? Did this get lost somewhere along the way? Any insight?

g11tech: @faustbrian yea i push with `--no-verify` but after running lint:fix, which is sort of my primary workflow to push. And the `lint:fix` doesn't seem to be prettifying/formating these things. would really appreciate if we can fix  the "lint:fix"

packages/client/test/net/protocol/snapprotocol.spec.ts
```diff
@@ -82,7 +85,88 @@ tape('[SnapProtocol]', (t) => {
...
+    // accountRangeRLP is the corresponding response to getAccountRangeRLP
+    const resData = RLP.decode(Buffer.from(accountRangeRLP, 'hex')) as unknown
+    const { accounts, proof } = p.decode(
+      p.messages.filter((message) => message.name === 'AccountRange')[0],
+      resData
+    )
+
+    const trie = new CheckpointTrie({ db: new LevelDB() })
+    try {
+      const accountRLP = await trie.verifyProof(
```
jochem-brouwer, line 151: If possible could this be changed to the `verifyRangeProof`? This should contain the logic which we should have to verify the entire proof. (Note: there could be bugs here, since I got those locally)

g11tech: we can replace with verifyRangeProof once the bugs are resolved and merged (change can be part of bug resolving PR itself)

packages/client/test/net/protocol/snapprotocol.spec.ts
```diff
@@ -149,7 +233,84 @@ tape('[SnapProtocol]', (t) => {
...
+      p.messages.filter((message) => message.name === 'StorageRanges')[0],
+      data
+    )
+    // storageRangesRLP response is to the lastAccount's slots so slots[0] are the slots of
+    // lastAccount
+    const lastAccountSlots = slots[0]
+    const lastAccountStorageRoot = (lastAccount.body as any)[2]
+    const trie = new CheckpointTrie({ db: new LevelDB() })
+    try {
+      const slotRLP = await trie.verifyProof(
```
jochem-brouwer, line 295: Same here for the `verifyRangeProof`

jochem-brouwer (commented):
These tests are very clean, only got 2 questions

scorbajio (dismissed):
utACK

packages/client/test/net/protocol/snapprotocol.spec.ts
g11tech: replaced with verifyRangeProof :+1:

packages/client/test/net/protocol/snapprotocol.spec.ts
g11tech: replaced by verifyRangeProof :+1:

jochem-brouwer (approved):
LGTM!

PR #2162, opened 2022-08-17
chore: enable `prettier/prettier` rule for `eslint`

faustbrian:
`utils` was using `6.8.0` while all others used `8.0.0` and that is what caused `utils` to blow up with `TypeError: Error while loading rule 'prettier/prettier': context.getPhysicalFilename is not a function` because the `prettier` plugin assumes `eslint@8.0.0` or higher. eslint now lives at the root package.json so this is no longer an issue.


gabrocheleau:
Shouldn't we move that dep to the root package.json like you've done in #2163 ?

gabrocheleau (approved):

acolytec3 (approved):
Making linting great again, LGTM!

holgerd77:
It's really hard to follow up with what's going on in this repository. 😜

Phew. But great stuff everywhere. Nice. 🔥

PR #2163, opened 2022-08-17
chore: move commonly shared `devDependencies` to root `package.json`

faustbrian:
> This is part of a series of PRs that aim to pin dependencies and update them where appropriate.

Moves dependencies that are used across almost every single package into the root to make it easier to keep versions consistent. Having shared development dependencies in the root makes things a lot simpler for that as I already multiple times ran into issues with upgrading things and then having things blow up because one package uses 2.8.4 and another 5.9.8 of the same dependency. Using another package manager than npm would also solve this.


packages/vm/package.json
```diff
@@ -80,25 +80,13 @@
...
     "memory-level": "^1.0.0",
     "minimist": "^1.2.5",
     "node-dir": "^0.1.17",
-    "nyc": "^15.1.0",
     "solc": "^0.8.1",
-    "standard": "^10.0.0",
-    "tape": "^5.3.1",
-    "ts-node": "^10.2.1",
-    "typescript": "^4.4.2"
+    "standard": "^10.0.0"
```
gabrocheleau, line 90:
```suggestion
```

packages/block/package.json
```diff
@@ -46,18 +46,6 @@
     "ethereum-cryptography": "^1.1.2"
   },
   "devDependencies": {
-    "@types/lru-cache": "^5.1.0",
```
gabrocheleau, line 49: This one has not been added to the root package.json. Is that intended?

gabrocheleau (dismissed):
One small comment, rest looks good!

packages/vm/package.json
```diff
@@ -80,25 +80,12 @@
...
-    "karma-chrome-launcher": "^3.1.0",
-    "karma-firefox-launcher": "^2.1.0",
-    "karma-tap": "^4.2.0",
-    "karma-typescript": "^5.5.3",
     "level": "^8.0.0",
     "memory-level": "^1.0.0",
     "minimist": "^1.2.5",
     "node-dir": "^0.1.17",
-    "nyc": "^15.1.0",
     "solc": "^0.8.1",
```
gabrocheleau, line 89:
```suggestion
    "solc": "^0.8.1"
```
extra comma

gabrocheleau (approved):

config/eslint.js
```diff
@@ -108,4 +108,13 @@ module.exports = {
     sourceType: 'module',
     project: './tsconfig.json',
   },
+  overrides: [
+    {
+      files: ['test/**/*.ts', 'tests/**/*.ts'],
+      rules: {
+        'implicit-dependencies/no-implicit': 'off',
```
acolytec3, line 115: Are there any risks you are aware of turning these off in tests?  I'm assuming this is here since the linter throws errors about tape not being present (since it's moving to the root)?

acolytec3 (commented):
I'm good with the changes though it does run the risk of us not getting proper linting errors in our tests around implicit dependencies.  Would it be better to leave the linting rules in place and just disable them for each instance of tape?  I know it's a little bit cumbersome so not sure what people's thoughts are here.

PR #2164, opened 2022-08-17
chore: add `engines` property to all `package.json` files

faustbrian:
Adds the `engines` property to each `package.json` file to make sure the minimum version is met when installing.


packages/trie/package.json
```diff
@@ -55,11 +55,11 @@
     "semaphore-async-await": "^1.5.1"
   },
   "devDependencies": {
+    "0x": "^4.9.1",
     "@types/benchmark": "^1.0.33",
     "@types/node": "^16.11.7",
     "@types/readable-stream": "^2.3.13",
     "@types/tape": "^4.13.2",
-    "0x": "^4.9.1",
```
gabrocheleau, line 62: Curious why this dep was removed? Was it not used?

gabrocheleau: Ah! wow totally missed that it had only been moved

gabrocheleau:
Looks good!
Would be best however is someone else could confirm that we're fine with only supporting Node v16 and greater, ast there might be retro-compatibility considerations that I'm unaware of. @holgerd77  @acolytec3  @jochem-brouwer

gabrocheleau (approved):

acolytec3:
We should support v14 and beyond for now.  We try to stick with supporting all NodeJS LTS versions not currently at EOL (so currently 14 and above) https://nodejs.org/en/about/releases/

acolytec3 (approved):
Lgtm

holgerd77:
Yes, think 14 is a good choice, we don't want to be too aggressive here.

PR #2166, opened 2022-08-17
chore: move commonly shared `devDependencies` to root `package.json`

faustbrian:
> This is part of a series of PRs that aim to pin dependencies and update them where appropriate.

- Moves dependencies that are used across almost every single package into the root to make it easier to keep versions consistent. Having shared development dependencies in the root makes things a lot simpler for that as I already multiple times ran into issues with upgrading things and then having things blow up because one package uses 2.8.4 and another 5.9.8 of the same dependency. Using another package manager than npm would also solve this.
- https://github.com/istanbuljs/nyc has been superseded by https://github.com/bcoe/c8 so I replaced `nyc` with `c8`


gabrocheleau (approved):
Nice, that seems to fix the code coverage CI issue :)

acolytec3 (approved):
LGTM and yay for fixing codecov in the process!

holgerd77:
Great PR, thanks! 😃👍

PR #2167, opened 2022-08-17
refactor(trie): drop `level` dependency completely

faustbrian:
After reviewing the trie package a bit more I realised that with the way Level 8.0.0 works and the abstractions introduced for the database that we are able to drop the default `LevelDB` implementation completely. This removes more dependencies and we can use a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) with hex keys to get the same in-memory behaviour for storage.

The `client` package relied on `LevelDB` for persistence so a modified implementation of the `LevelDB` class has been moved to it. After this PR I'll make some further changes to reduce the dependency footprint of the trie package.


holgerd77:
I am a bit undecided here, is it worth to give up the convenience of having Level more tightly integrated? This *does* put some burden to users who *do* want to use level DB? 🤔 I thought "browser" e.g. would not be such an issue any more with the latest changes?

On the other hand I see the beauty of this to have this completely freed up.

Any more voices here?

acolytec3:
> I am a bit undecided here, is it worth to give up the convenience of having Level more tightly integrated? This _does_ put some burden to users who _do_ want to use level DB? thinking I thought "browser" e.g. would not be such an issue any more with the latest changes?
> 
> On the other hand I see the beauty of this to have this completely freed up.
> 
> Any more voices here?

I'd vote to remove entirely.  It will shrink the size of that package somewhat.

Could we provide an example of what the levelDB adapter would look like for `trie` based on what @faustbrian put in the client package and leave it at that?

holgerd77:
> > I am a bit undecided here, is it worth to give up the convenience of having Level more tightly integrated? This _does_ put some burden to users who _do_ want to use level DB? thinking I thought "browser" e.g. would not be such an issue any more with the latest changes?
> > On the other hand I see the beauty of this to have this completely freed up.
> > Any more voices here?
> 
> I'd vote to remove entirely. It will shrink the size of that package somewhat.
> 
> Could we provide an example of what the levelDB adapter would look like for `trie` based on what @faustbrian put in the client package and leave it at that?

Ok, then let's do it. 🙂 

I think Brian has integrated various examples in the PR (or are you thinking about yet another one or in another context or whatever?).

acolytec3:
One question that just occurred to.  @faustbrian are you able to provide any performance difference between `level-memory` and a per javascript map?  I'm assuming the `map` would be faster but this would be one more thing to check.

acolytec3:
That answers my question.  Looks great!

acolytec3 (dismissed):

holgerd77 (changes requested):
Let's please put this on hold until the questions from the Ganache side from @davidmurdoch from https://github.com/ethereumjs/ethereumjs-monorepo/pull/2159#issuecomment-1218843400 are settled.

davidmurdoch:
I misunderstood this change previously; I thought this change was removing the support for level from the Trie implementation entirely. I see now that it is not; I don't think this change will be problematic for us (at least not any more problematic than the Trie changes have already been :-) ).

Since Ganache will likely need to still use `LevelDB` (or we will have to implement the exact same class ourselves) and we don't otherwise have a need for `@ethereumjs/client`, moving the `LevelDB` implementation there will be mildly annoying for us.

holgerd77:
Ok, after the clarifications I guess we can merge this now once this is ready.

holgerd77 (approved):
I really love this when looking over this a second and third time 💯 🙂. It will likely be *some* burden from the one or the other user. But I would otherwise also very much say that it's worth the change, seeing the current state of the "dependencies" section is really eye opening.

PR #2170, opened 2022-08-17
Remove state tests from coverage run

acolytec3:
Code Coverage shouldn't be dependent on the `ethereum/tests` state tests so removing it from the CI since it takes a long time to run for no gain.


holgerd77:
Hmm, the reasoning for this at least *was* at some point that large parts of the VM *are* actually tested by the state tests (there was a time where we had no API tests *at all* and this was our sole test suite), so this would lead to a substantial coverage drop.

One could argue that the API tests are the main tests which *should* be considered for coverage - since the other test suite is rather some kind of a black box where we do not really have the coverage branches under control.

I am not sure yet though if this would be preferrable since I guess a lot of special cases are likely only covered by the state tests.

Update while (after) writing this: hmm, seems this PR is *not* decreasing coverage, just discovered. Why is this? 🤔

acolytec3:
> Hmm, the reasoning for this at least _was_ at some point that large parts of the VM _are_ actually tested by the state tests (there was a time where we had no API tests _at all_ and this was our sole test suite), so this would lead to a substantial coverage drop.
> 

Ah, that makes sense from looking at the history of those lines.

> One could argue that the API tests are the main tests which _should_ be considered for coverage - since the other test suite is rather some kind of a black box where we do not really have the coverage branches under control.
> 
> I am not sure yet though if this would be preferrable since I guess a lot of special cases are likely only covered by the state tests.
> 
> Update while (after) writing this: hmm, seems this PR is _not_ decreasing coverage, just discovered. Why is this? thinking

This change wouldn't stop the state tests from running.  They're already running so we're effectively running them twice in the CI at present.  This is an experiment to see if we lose any test coverage in the VM by not running the state tests.  Initial results seem positive on that front and if we can not run the state tests twice, it'll definitely improve our CI timeline.  I either want to remove it from the coverage line or the separate task we have for running the state tests.

holgerd77:
Ah, just caught me: the reason why this is not changing anything *here* is that the state tests are solely targeted towards the EVM (so that's why coverage there is so low). So for coverage reasons we can actually drop here.

I wonder if we *can* actually just move this over to the EVM coverage format and - from there - first changing the directly and actually *use* this for EVM coverage? Would this work? 🤔

acolytec3:
> Ah, just caught me: the reason why this is not changing anything _here_ is that the state tests are solely targeted towards the EVM (so that's why coverage there is so low). So for coverage reasons we can actually drop here.
> 
> I wonder if we _can_ actually just move this over to the EVM coverage format and - from there - first changing the directly and actually _use_ this for EVM coverage? Would this work? thinking

Interesting, I didn't realize that but possibly?  The issue there is that the tester isn't currently designed to run on the EVM standalone so will take some additional work to make that possible.  Is that worth doing you think?  I'm happy to take a whack at it but it still doesn't seem likely that will have anything to do with coverage.  For now, could we at least remove it here since it isn't impacting test coverage in the VM at all (as far as I can tell)?

holgerd77:
> > Ah, just caught me: the reason why this is not changing anything _here_ is that the state tests are solely targeted towards the EVM (so that's why coverage there is so low). So for coverage reasons we can actually drop here.
> > I wonder if we _can_ actually just move this over to the EVM coverage format and - from there - first changing the directly and actually _use_ this for EVM coverage? Would this work? thinking
> 
> Interesting, I didn't realize that but possibly? The issue there is that the tester isn't currently designed to run on the EVM standalone so will take some additional work to make that possible. Is that worth doing you think? I'm happy to take a whack at it but it still doesn't seem likely that will have anything to do with coverage. For now, could we at least remove it here since it isn't impacting test coverage in the VM at all (as far as I can tell)?

Yes, we can remove here for sure.

Not sure if it is *necessary* to redesign to run EVM standalone, does it hurt if this is run via the VM? Would be enough if the EVM code touches are measured (or is this then not the case? 🤔). The VM is just some boilerplate entrypoint in this context I would say.

acolytec3:
> > > Ah, just caught me: the reason why this is not changing anything _here_ is that the state tests are solely targeted towards the EVM (so that's why coverage there is so low). So for coverage reasons we can actually drop here.
> > > I wonder if we _can_ actually just move this over to the EVM coverage format and - from there - first changing the directly and actually _use_ this for EVM coverage? Would this work? thinking
> > 
> > 
> > Interesting, I didn't realize that but possibly? The issue there is that the tester isn't currently designed to run on the EVM standalone so will take some additional work to make that possible. Is that worth doing you think? I'm happy to take a whack at it but it still doesn't seem likely that will have anything to do with coverage. For now, could we at least remove it here since it isn't impacting test coverage in the VM at all (as far as I can tell)?
> 
> Yes, we can remove here for sure.
> 
> Not sure if it is _necessary_ to redesign to run EVM standalone, does it hurt if this is run via the VM? Would be enough if the EVM code touches are measured (or is this then not the case? thinking). The VM is just some boilerplate entrypoint in this context I would say.

It may not.  I need to look at how the state test runner works today.  If we can just execute `evm.runCall` as a dropin replacement for `vm.runTx` it shouldn't be a big deal but will see.

acolytec3:
I've added a new issue to address the EVM/ state tests question.  If we're good with removing the duplicate state test runs from the current VM CI, this should be ready for final review.

acolytec3:
WDYGT?  I'm not sure whether to merge or not, given all the conversation on discord.  If we're not going to remove the state tests from coverage, we should definitely remove the other run so we don't duplicate our run time on the CI.

holgerd77:
Updated the branch via UI.

holgerd77 (approved):
LGTM

jochem-brouwer (commented):
LGTM

jochem-brouwer (approved):

jochem-brouwer:
Unrelated, but not sure why GH shows 9 commits while there is only one change..?

acolytec3:
It's just all the merge commits 🙃

PR #2172, opened 2022-08-17
adjust skipbalance logic

acolytec3:
Opening as a replacement for #2138 that had its git history corrupted for some reason.


jochem-brouwer (approved):
LGTM!

PR #2174, opened 2022-08-18
ci: use `npm ci` instead of `npm i` to install dependencies

faustbrian:
> https://docs.npmjs.com/cli/v8/commands/npm-ci

This command is similar to [npm install](https://docs.npmjs.com/cli/v8/commands/npm-install), except it's meant to be used in automated environments such as test platforms, continuous integration, and deployment -- or any situation where you want to make sure you're doing a clean install of your dependencies.

### DIFF
--- a/.github/workflows/block-build.yml
+++ b/.github/workflows/block-build.yml
@@ -35,7 +35,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: 'npm'
 
-      - run: npm i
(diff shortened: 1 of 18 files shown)


holgerd77 (approved):
Personally not familiar with the command but trust your assessment here and very much seems to make sense since the command seems to be very much targeted for the situation we use it here.

Will approve and merge.

PR #2175, opened 2022-08-18
ci: run `docs:build` weekly and create a pull request

faustbrian:
See https://github.com/ethereumjs/ethereumjs-monorepo/pull/2175/commits/cb283a8ed7f02e3e7621c11575cfb80a9ff20f4c for committed changes. This ensures that documentation is **_always_** in sync with committed code.


holgerd77:
Hi there, for me this is three steps or so too far ahead. 😋 😬

Can we please - minimally, might be other steps necessary in between - do a manual round of updating the docs, with a dedicated look at each repo doc creation, eventually even one-by-one. It's absolutely unclear in what condition doc creation is at the moment, some packages have their particularities and we might have doc creation problems like in the trie library here and there. Also we had these structural package changes.

So my priority before the final releases would be to have a set of up-to-date docs *at all* for all libraries 😜, this will already be demanding enough given the timeframe we have. Then we can see as a second or third step if we can and want to structuraly improve here.

Will mark this as "blocked" for now. 🙂

holgerd77:
Could you expand a bit here how this actually works? When and how often are docs built and how are they added to the PR? Could this interfere with other PRs and lead to various additional merge conflicts if every PR is building their own docs?

Some of my current reservations. 🙂

holgerd77:
Ah ok, thanks a lot! 🙂 Will give this some thought, still have some reservations regarding interference with everyday work.

PR #2180, opened 2022-08-18
Bump EVM coverage

jochem-brouwer:
Let's see if this works.

### DIFF
--- a/.github/workflows/evm-build.yml
+++ b/.github/workflows/evm-build.yml
@@ -26,6 +26,8 @@ jobs:
         node-version: [16]
     steps:
       - uses: actions/checkout@v3
+        with:
+          submodules: recursive
 
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v2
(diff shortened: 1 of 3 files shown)


jochem-brouwer:
OK, no idea why this does not increase coverage.

holgerd77:
Maybe we need to reference the EVM on the local file level in the VM test runner so that it is not referenced/seen as a package? 🤔

Can you try creating an explicit EVM in `GeneralStateTestRunner.ts` and pass this to the VM and import via the local path (Interpreter as well), so `import { EVM, Interpreter } from '../../evm/src/'` (something like this, not sure if path is correct)?

jochem-brouwer:
Thanks a lot @holgerd77, this change made a lot of sense! EVM coverage now up +16.10%. I'd say this is ready for review.

holgerd77:
Rebased this via UI.

holgerd77 (approved):
Ok, cool, great that this worked! 👍🙂

Then let's merge.

PR #2182, opened 2022-08-18
VM update type casting

jochem-brouwer:
This PR attempts to solve the problem that VMs EVM property is of type `EVMInterface`. It is now possible to add types to the `create` method of VM such that the EVM type is correctly inferred.

However it would be much nicer if this type can be taken from the option type which is passed in in the constructor (not sure if this is possible at TS).

By doing the casting we can access the `.on` method for instance;

image

If no type is given at the constructor, it defaults to `EVMInterface`.

The same happens with EEI;

image

### DIFF
--- a/packages/vm/src/vm.ts
+++ b/packages/vm/src/vm.ts
@@ -128,7 +128,7 @@ export class VM<
       this.eei = opts.eei
     } else {
       if (opts.evm) {
-        this.eei = <any>opts.evm.eei
+        this.eei = <EEIType>opts.evm.eei
       } else {
         this.eei = <any>new EEI(this.stateManager, this._common, this.blockchain)
       }
--- a/packages/vm/src/vm.ts
+++ b/packages/vm/src/vm.ts
@@ -113,9 +113,7 @@ export class VM<
     if (opts.stateManager) {
       this.stateManager = opts.stateManager
     } else {
-      this.stateManager = new DefaultStateManager({
-        common: this._common,
-      })
+      this.stateManager = new DefaultStateManager({})
     }
 
     this.blockchain = opts.blockchain ?? new (Blockchain as any)({ common: this._common })
--- a/packages/vm/src/vm.ts
+++ b/packages/vm/src/vm.ts
@@ -31,7 +31,7 @@ import type { StateManager } from '@ethereumjs/statemanager'
  *
  * This class is an AsyncEventEmitter, please consult the README to learn how to use it.
  */
-export class VM<EVMType extends EVMInterface> extends AsyncEventEmitter<VMEvents> {
+export class VM<EVMType extends EVMInterface = EVMInterface> extends AsyncEventEmitter<VMEvents> {
   /**
    * The StateManager used by the VM
    */
@@ -253,7 +253,7 @@ export class VM<EVMType extends EVMInterface> extends AsyncEventEmitter<VMEvents
       stateManager: (eeiCopy as any)._stateManager,
       blockchain: (eeiCopy as any)._blockchain,
       common: (eeiCopy as any)._common,
-      evm: <any>evmCopy,
+      evm: <EVMType>evmCopy,
       hardforkByBlockNumber: this._hardforkByBlockNumber ? true : undefined,
       hardforkByTTD: isTruthy(this._hardforkByTTD) ? this._hardforkByTTD : undefined,
     })
--- a/packages/vm/src/vm.ts
+++ b/packages/vm/src/vm.ts
@@ -31,7 +31,10 @@ import type { StateManager } from '@ethereumjs/statemanager'
  *
  * This class is an AsyncEventEmitter, please consult the README to learn how to use it.
  */
-export class VM<EVMType extends EVMInterface = EVMInterface> extends AsyncEventEmitter<VMEvents> {
+export class VM<
+  EVMType extends EVMInterface = EVMInterface,
+  EEIType extends EEIInterface = EEIInterface
+> extends AsyncEventEmitter<VMEvents> {
   /**
    * The StateManager used by the VM
    */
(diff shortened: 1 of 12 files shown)


acolytec3:
@jochem-brouwer I think I solved the errors where declaring a parameter to be type VM wants you to always do the `VM<EVMType>` or whatever.  You can set a default subtype in the class definition so added it there and set to `EVMInterface` and the API tests are all passing now.

jochem-brouwer:
@davidmurdoch Would this work for you and fix these type errors/problems you had with evm/eei? (See first post)

packages/vm/src/vm.ts
```diff
@@ -122,17 +128,17 @@ export class VM extends AsyncEventEmitter<VMEvents> {
       this.eei = opts.eei
     } else {
       if (opts.evm) {
-        this.eei = opts.evm.eei
+        this.eei = <any>opts.evm.eei
```
jochem-brouwer, line 131: It would be great if we could remove this `any`

packages/vm/src/vm.ts
```diff
@@ -122,17 +128,17 @@ export class VM extends AsyncEventEmitter<VMEvents> {
...
       } else {
-        this.eei = new EEI(this.stateManager, this._common, this.blockchain)
+        this.eei = <any>new EEI(this.stateManager, this._common, this.blockchain)
```
jochem-brouwer, line 133: It would be great if we could remove this `any`

packages/vm/src/vm.ts
```diff
@@ -122,17 +128,17 @@ export class VM extends AsyncEventEmitter<VMEvents> {
...
       }
     }
 
     // TODO tests
     if (opts.evm) {
       this.evm = opts.evm
     } else {
-      this.evm = new EVM({
+      this.evm = <any>new EVM({
```
jochem-brouwer, line 141: It would be great if we could remove this `any`

jochem-brouwer:
Ready for review, but do not merge until we get a :+1: from @davidmurdoch

davidmurdoch:
I haven't pulled this locally to try it, but this looks like it defaults the types to `EVMInterface` and `EEIInterface`, but the _actual_ defaults are `EVM` and `EEI`. The same can be said of the `stateManager` field, whose type is`StateManager`, but the default is actually `DefaultStateManager`.

```
const vm = VM.create(); // no options
vm.eei // should be type EEI, not EEIInterface
```

If the types are cascading through the options correctly this should work:

```
class MyEVM extends EVM {
   public myCustomField: number;
}

const vm = VM.create({ evm: new MyEVM() });
vm.evm; // should be type MyEVM
vm.evm.myCustomField // should be type number
```

jochem-brouwer:
I agree that the defaults are indeed EVM and EEI, but if someone imports a new type of EVM or EEI and forget to cast the interface, then it will default to EVM/EEI and they are just on the wrong type?

Ideally we would indeed want that these types cascade through the options, but I am not sure how to do this. This is not yet a feature.

holgerd77:
> I agree that the defaults are indeed EVM and EEI, but if someone imports a new type of EVM or EEI and forget to cast the interface, then it will default to EVM/EEI and they are just on the wrong type?
> 
> Ideally we would indeed want that these types cascade through the options, but I am not sure how to do this. This is not yet a feature.

Another option would/might be to shield the events by (extract to) an object property of EVM and EEI instead of making these event emitters directly, so that we would have something like:

```typescript
evm.events.on('step', () = { // do something })
```

Not sure if I have thought this through but this might encapsulate the whole event mechanism better (a bit more like the event bus in the client) and then `evm` (and EEI) are freed from being event emitters themselves, eventually `events` can also made optional (maybe simply with the behavior, no `events` property -> no events emitted. 🙂 ).

davidmurdoch:
@jochem-brouwer See https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgKIDUCyBJc15LIDeAUMucnAApQD2ADtGAJ4BcyIArgLYBG0AbhIBfEiQQAbOAGdpaLMmDd6EiNwjg5GHHliIUpCsnqdeE4Ako0GTNhx78oyALzIADEKMmzF5BABu3AAqABagAObsvLS0qnAgLshgUJwQQqLiUrLIAMKc0mC03Np+AB6QIAAmWgqGFN7mlgj5hcFhIJHIBVARyAA+9hISiVxD6WIsjMhYAPL0YMC0INIAPNpBzFMQ5RrV8jqQekgAfIl15AHcAPzs65sQIiQwnCAIC0vIITIAYsAQEpUVkFjgAKODseLMACUEMUciCxDIFCgEDAnCgCTgyAAhK5RhJxpIZHIsGssBstjsqjUDvh9IltKdzsZTI1kCi4JUlhJmH5ArdyfchEjyAglt1OG9aFAQQx3st2LN5otlmTMBSIMcocyjMAYMhZcrxQA6S5QxFGS0UMBhaSmwKJOUqu2XTxW5DCPwSaQGEXu8g24Auh14iAAd32IPNMmQLwA1iBaGHMTSNW6raIjJn6j1-HBIJRpMxXooQMAwFGLe6APTV5DGht+7PkAr53wyYtNDmQNUasoVPbaXCHAgoVyMw3y6SKzBzKe9+6nVxEYTav1GMXLMDIEMccPTTCT51Q9OWuBhuDlnfcY2gctR09GFFojHX0+iDJgosl56veVR5lNwKZBKggeBOAkMB0G4RJz0vbcsGNBBuwgB8-VA8DIOg+1uFPWt-QAPSuP0-SA7dmgKIoSlcOCr0Q5CIHzVCiEudgQH3PJKOKLAo1XU8KNabQcLwut3SIkQoQfIA for how this can be done

packages/vm/src/vm.ts
```diff
@@ -31,7 +31,10 @@ import type { StateManager } from '@ethereumjs/statemanager'
  *
  * This class is an AsyncEventEmitter, please consult the README to learn how to use it.
  */
-export class VM extends AsyncEventEmitter<VMEvents> {
+export class VM<
+  EVMType extends EVMInterface = EVMInterface,
+  EEIType extends EEIInterface = EEIInterface
```
davidmurdoch, line 36: I don't think this is correct; I think the default types are `EVM` and `EEI`, right?

holgerd77: We are (or: @jochem-brouwer is) testing out a new approach here at #2208 which might be able to leverage the resulting type problems by preventing this type split in the first place. We'll see. 🙂

jochem-brouwer:
This one is merged into #2208, see there for updates

PR #2183, opened 2022-08-18
vm: make `touchAccount` of state manager non-protected

jochem-brouwer:
`touchAccount` is `protected` which it should not be.


acolytec3 (approved):
LGTM

PR #2184, opened 2022-08-18
evm/vm: update event signatures

jochem-brouwer:
If you listen to an event in `evm` or `vm` asynchronously, then you need to call a continuation function once you are done;

```typescript
vm.on('afterTx',  async (e, cont) => {
... do async stuff
cont()
})
```

However, our type signatures stated that this `cont` function should always be called with an argument, which is not the case.


acolytec3 (approved):
LGTM

PR #2186, opened 2022-08-19
refactor(trie): move `Trie#_findValueNodes` to `TrieReadStream`

faustbrian:
Moves the `Trie#_findValueNodes` function to `TrieReadStream` because it is only called there.


holgerd77 (approved):
LGTM

PR #2187, opened 2022-08-19
refactor(trie): replace `semaphore-async-await` with simpler implementation

faustbrian:
Drops the `semaphore-async-await` dependency and replaces it with only the bits that we're actually using (which is very little). I left comments with links to the original implementations in the code.

**The trie package is now almost free of third-party dependencies, only the `readable-stream` remains for an unused class. Assuming https://github.com/ethereumjs/ethereumjs-monorepo/pull/2167 gets merged.**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21484,8 +21484,7 @@
         "ethereum-cryptography": "^1.1.2",
         "level": "^8.0.0",
         "memory-level": "^1.0.0",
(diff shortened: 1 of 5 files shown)


holgerd77:
Rebased this via UI.

holgerd77 (approved):
Pretty cool, LGTM, just checked, dependency is MIT licensed, so this should be ok. 👍 🙂

Issue #1114, opened 2021-02-16
Client JSON-RPC Support


ryanio:
This is a tracking issue for [JSON-RPC](https://eth.wiki/json-rpc/API) support in the client.

| Endpoint     | Status       |
| ----------- | ----------- |
| web3_clientVersion | :white_check_mark: |
| web3_sha3 | :white_check_mark: |
| net_version | :white_check_mark: |
| net_peerCount | :white_check_mark: |
| net_listening | :white_check_mark: |
| eth_protocolVersion | :white_check_mark: |
| eth_syncing | |
| eth_coinbase | | 
| eth_mining | |
| eth_hashrate | |
| eth_gasPrice | |
| eth_accounts | |
| eth_blockNumber | :white_check_mark: |
| eth_getBalance | |
| eth_getStorageAt | |
| eth_getTransactionCount | |
| eth_getBlockTransactionCountByHash | :white_check_mark: |
| eth_getBlockTransactionCountByNumber | |
| eth_getUncleCountByBlockHash | |
| eth_getUncleCountByBlockNumber | |
| eth_getCode | |
| eth_sign | |
| eth_signTransaction | |
| eth_sendTransaction | |
| eth_sendRawTransaction | |
| eth_call | |
| eth_estimateGas | |
| eth_getBlockByHash | |
| eth_getBlockByNumber | :white_check_mark: |
| eth_getTransactionByHash | :white_check_mark: |
| eth_getTransactionByBlockHashAndIndex  | |
| eth_getTransactionByBlockNumberAndIndex | |
| eth_getTransactionReceipt | |
| eth_getUncleByBlockHashAndIndex | |
| eth_getUncleByBlockNumberAndIndex | |
| eth_getCompilers | |
| eth_compileLLL | |
| eth_compileSolidity | |
| eth_newFilter | |
| eth_newBlockFilter | |
| eth_newPendingTransactionFilter | |
| eth_uninstallFilter | |
| eth_getFilterChanges | |
| eth_getFilterLogs | |
| eth_getLogs | |
| eth_getWork | |
| eth_submitWork | |
| eth_submitHashrate | |
| db_putString | |
| db_getString | | 
| db_putHex | |
| db_getHex | |

#### Namespace: Real-time Events ([Publish-Subscribe](https://geth.ethereum.org/docs/rpc/pubsub))

| Endpoint     | Status       |
| ----------- | ----------- |
| eth_subscribe | | 
[...]

PR #2188, opened 2022-08-19
client: RPC get block transaction count by number

scorbajio:
Implements `eth_getBlockTransactionCountByNumber` rpc endpoint for #1114.


jochem-brouwer:
Could you add a test for this one? There's also a console.log left :)

scorbajio:
> 

Yes, I'll take care of that. Picking this PR back up.

acolytec3:
So I didn't realize we had left this one open but this was completed in #2379.  Sorry we didn't see and merge this earlier!  I'm going to close for now as it accomplishes the same thing.

PR #2192, opened 2022-08-19
chore: replace `nyc` configurations with `c8` configurations

faustbrian:
`c8` supports reading the old `nyc` configurations but it makes more sense to rename them to the names that `c8` itself uses.


holgerd77 (approved):
LGTM, small nit: might have been nice to use `git mv` for renaming to preserve the file history, but I guess in the cases of these super small configuration files this is not absolutely necessary.

Issue #2193, opened 2022-08-19
Remove unused dependencies


paulmillr:
Based on @faustbrian analysis from #2190

- `buffer-xor` - (1 sub-dep) can be easily replaced with 5-line function
- `core-js-pure` - used for shimming `Set`, our targets should support `Set` as-is
- `fs-extra` - used for `ensureDirSync` in `client`, also `readFileSync` but that's available in `fs`. I would argue to replace it, since it has 3 deps, graceful-fs etc.
- `inherits`, `it-pipe`, `it-pushable`  - used as devDependency, while it's specified as prod dependency
- `miller-rabin` - (3 sub-deps) not used anywhere

paulmillr:
unused devDeps:

- `@babel/plugin-transform-spread`, `node-fetch`, `pino-pretty`, `pull-pair`
- `node-dir` in `evm`, `tx`
- `web3` in `trie` is only used in `examples`, but not in tests or src. Can we create examples package.json so that it's not fetched by default when tests are executed?

Not sure if used, from webpack.config.js:

```
constants: require.resolve('constants-browserify'),
crypto: require.resolve('crypto-browserify'), // used by: rlpxpeer, bin/cli.ts
os: require.resolve('os-browserify/browser'), // used by: bin/cli.ts, web3_clientVersion rpc
stream: require.resolve('stream-browserify'), // used by: fetcher
```

paulmillr:
- `master` npm install: `added 1925 packages, and audited 1939 packages in 40s`
- After removing all those packages: `removed 217 packages, and audited 1722 packages`
- After replacing `fs-extra` (a bit more complex): `removed 4 packages, and audited 1718 packages`

holgerd77:
Totally love it that this gets some serious attention! ❤️

PR #2194, opened 2022-08-19
Remove unused deps

paulmillr:
#2193

Before: `audited 1939 packages`, after: `1585 packages`

1. Why is `lib` used instead of `src` in `client`? I have ignore rules for `lib` dirs, because they contain compiled js code all of the time.
2. `verdaccio@4` is the only package, it seems, that requires downloading node.js binaries/headers from nodejs.org site. Does it make sense to extract it to a separate directory? Or, would you rather keep it top-level, to be installed in all cases? UPD: v5 seems to be ok. also v4 has security warnings


holgerd77:
On 1. Historical inconsistency
Can be changed, but please in a separate PR

holgerd77:
Side note: it's very much ok if this task will take 6 weeks or so, you do not have to finish this all in one go. 😆

Maybe we can also merge something in in-between and split this up into some rounds.

acolytec3:
> 2\. `verdaccio@4` is the only package, it seems, that requires downloading node.js binaries/headers from nodejs.org site. Does it make sense to extract it to a separate directory? Or, would you rather keep it top-level, to be installed in all cases? UPD: v5 seems to be ok. also v4 has security warnings


As far as I know, verdaccio is only used in one CI script (specifically for the hardhat e2e tests that we were running as a sanity check when pushing changes to verify if they were breaking or not) so it ought to be removable from the top level package.json entirely.  Let me try that in my branch that updates the hardhat tests and see if it still works.

>@acolytec3 do you know what specific issue required adding karma-typescript-es6-transform in https://github.com/ethereumjs/ethereumjs-monorepo/pull/2091?

Unless something has changed, Karma doesn't support ES6 so uses Babel to transpile to ES5 compatible code - https://github.com/monounity/karma-typescript/tree/master/packages/karma-typescript-es6-transform

Looks like removing that transform is what is causing those client tests to fail.  I had to add that in to get karma to work for client tests since it depends on an ESM module in the imports.

paulmillr:
I've ensured all devDependencies in root package have specific versions.

1. A broken typescript, prettier, or anything else — could bring malware to the compiled code, before packages are published.
2. I have went through package-lock.json to ensure there are no duplicates and that specified versions are OK
3. This change should have no effect on client code, since it's only devDependencies. We will think if it would make sense to fixate prod deps in the next major release, and i'm not sure if it would make sense at all. Keeping devDeps the new way is a good thing and security improvement in any case.

acolytec3:
Also, depending on how this [CI run](https://github.com/ethereumjs/ethereumjs-monorepo/runs/7925155491?check_suite_focus=true) goes, we may be able to remove verdaccio from the deps as well.

acolytec3:
@paulmillr looks like Verdaccio still works if we just install it directly in that CI script.  You should be able to remove it from the dependencies here since we're not currently running the hardhat tests at all (since they're broken).

holgerd77:
> @paulmillr looks like Verdaccio still works if we just install it directly in that CI script. You should be able to remove it from the dependencies here since we're not currently running the hardhat tests at all (since they're broken).

Oh, that's great, actually the most significant share of the dependency alert noise is coming from Verdaccio! 🙏 🎉

paulmillr:
The PR is ready now.

paulmillr:
@faustbrian i've tried this, see earlier commits, there is additional more complicated logic in fs-extra, so maybe in the next pr

acolytec3 (approved):
Yay for less deps.

holgerd77:
@paulmillr congrats to the substantial achievements here! 💯

Great to see this merged! 🙂

PR #2195, opened 2022-08-20
fix(trie): add static `create` constructor to `CheckpointTrie` and `SecureTrie`

faustbrian:
Same as `Trie.create` but for the subclasses so that they can support root restoration. Without this the underlying `Trie` was constructed instead of the subclass.


holgerd77:
> ~@holgerd77 will there be any more beta releases before the stable tag? Would be good to test #2167 and this PR in a release before tagging a stable, just to make sure there's no oddities left.~
> 
> Just tried the mentioned PRs via `"@ethereumjs/trie": "link:../../../ethereumjs-monorepo/packages/trie"` locally in my project and works as expected and all type errors are gone.

Cool 👍, but also to answer on the above: we now had so many unexpected additional changes (and change requests) that I am very inclined to drop my initial we-should-directly-go-to-final-releases opinion and do another round of Beta releases instead before, I guess that ver much makes sense.

holgerd77:
Oh. Tried to rebase via UI but this failed. 😬 @faustbrian could you please manually update here? Thanks!

holgerd77 (approved):
LGTM, nice fix, great that you caught that!

PR #2196, opened 2022-08-20
docs(trie): add upgrade guide for v5

faustbrian:


gabrocheleau (approved):
Looks good to me, I think this should bring much more clarity to some of the issues discussed in #2159

holgerd77:
@faustbrian thanks, this is a great guide! 🎉

Can we have some link towards this from the main README? I am a bit afraid that just no one (or, at least: not many people) will find it like this and this and that would be a bit of a pity.

I think it would also make sense to add an additional note to the v5 Beta 1 CHANGELOG.

PR #2197, opened 2022-08-20
Remove common from state manager

acolytec3:
The instance of `Common` in the `BaseStateManager` class is unused except for being copied around.  This PR is an experiment to see if it can be safely removed entirely.


g11tech (approved):
Makes sense, we should remove `common` from other such places as well where its not required.

packages/statemanager/tests/stateManager.spec.ts
```diff
@@ -24,7 +23,6 @@ tape('StateManager', (t) => {
     const stateManager = new DefaultStateManager()
 
     st.deepEqual(stateManager._trie.root, KECCAK256_RLP, 'it has default root')
-    st.equal(stateManager._common.hardfork(), 'petersburg', 'it has default hardfork')
```
jochem-brouwer, line 27: How come this defaulted to `petersburg` and this test passed!?

jochem-brouwer (approved):
LGTM, not sure why the petersburg test passed before though!?

packages/statemanager/tests/stateManager.spec.ts
acolytec3: :thinking: No idea but it's gone now :-)

holgerd77:
Oh nice, didn't realize that this was now freed from any HF-specific logic after the refactoring! 💯

@g11tech I would very much assume that this is a rare exception and there will close-to-for-sure no other places where Common could be removed. This was just because of the latest "slimming" along the StateManager refactoring. 😋

PR #2200, opened 2022-08-20
trie: reorganize examples and update mpt tutorial

gabrocheleau:
This PR:

- Reorganizes the examples folder so that the readme/examples from the MPT are grouped together
- Updates the broken MPT examples 
- Removes the reliance on web3, examples can now all be run without installing an additional package
- Adds an general examples README with a table of contents of the different examples


acolytec3 (approved):
LGTM

PR #2201, opened 2022-08-21
ci: run `lintlock` if `package-lock.json` has been modified

faustbrian:


gabrocheleau (approved):
Awesome!

holgerd77:
Just to make sure we are on the same page: I would assume that this `lockfile-lint` too will bring the `package-lock.json` to a unified format to make PR diffs on `package-lock.json` smaller which allows to easier review differences.

a) is this correct and b) is this the sole purpose or did I forget something?

holgerd77:
And also maybe for completeness: where is the first run where this has been applied and lead to an updated `package-lock.json` file?

holgerd77:
(respectively: the link from above (so: the CI run) was triggered by this PR. I am wondering why this didn't lead to a changed `package-lock.json` file? It is highly unlikely that such a lint run changes plainly *nothing* on first run?)

holgerd77:
> As far as I'm aware it only makes sure that your lockfile [adheres to certain rules](https://github.com/lirantal/lockfile-lint#usage) and errors if it doesn't. I don't think it performs any writes to the lockfile. The linting name might be misleading.

Ah, ok, interesting 👍, this indeed totally mislead me.

PR #2202, opened 2022-08-21
vm updates types improvements

gabrocheleau:
This brings potential improvements to the types of #2182.

The general reasoning is the following:
EEI can actually come from 3 places:
1. From opts.eei
2. From opts.evm.eei
3. From the EVM constructor if none of the above is provided

So based on the existing logic, the goal of the updates `create` types is the following:

- if a custom "evm" is provided, then it also needs to come with `evm.eei` (and EEIType should then come from that).
- if no custom "evm" is provided, then a custom eei can be provided through `opts.eei`
- EVMType and EEIType default to EVM/EEI respectively (matches the logic of the constructor)

An additional type improvement was also made in the `copy` method, where the `eei` is always taken from the `evm` (whether or not this is the behavior that we want is another question, discussion ongoing in discord).

### DIFF
--- a/packages/evm/src/types.ts
+++ b/packages/evm/src/types.ts
@@ -8,12 +8,12 @@ import type { Account, Address, PrefixedHexString } from '@ethereumjs/util'
 /**
  * API of the EVM
  */
-export interface EVMInterface<EEIType extends EEIInterface = EEIInterface> {
+export interface EVMInterface {
   runCall(opts: EVMRunCallOpts): Promise<EVMResult>
   runCode?(opts: EVMRunCodeOpts): Promise<ExecResult>
   precompiles: Map<string, any> // Note: the `any` type is used because EVM only needs to have the addresses of the precompiles (not their functions)
   copy(): EVMInterface
-  eei: EEIType
+  eei: EEIInterface
 }
 
 /**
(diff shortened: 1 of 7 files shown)


holgerd77:
@gabrocheleau can something like this please *always* be reasoned? What are you doing here, in what form did you think this improves the status quo and how/where was this motivated (at least some link reference)?

This might be obvious to you but in many cases it is not for a reviewer or others who want to follow along.

Thanks! 🙂

Will give this a blocked label for now to avoid side effects by accidental merging and changes not yet sufficiently discussed/understood.

holgerd77:
(and if there was some discussion on Discord or so on this, it is also a good way to just copy over the Discord reasoning to make this transparent)

gabrocheleau:
> @gabrocheleau can something like this please _always_ be reasoned? What are you doing here, in what form did you think this improves the status quo and how/where was this motivated (at least some link reference)?
> 
> This might be obvious to you but in many cases it is not for a reviewer or others who want to follow along.
> 
> Thanks! slightly_smiling_face
> 
> Will give this a blocked label for now to avoid side effects by accidental merging and changes not yet sufficiently discussed/understood.

In case that was missed, this PR points to #2182, just didn't want to add a commit to @jochem-brouwer's PR directly. Indeed the discussion for that happened in Discord, but I've added the reasoning to the PR description.

jochem-brouwer:
I made a bit of a mess, but I have merged this into #2229 

Thanks a lot @gabrocheleau!! :smile:

PR #2203, opened 2022-08-21
trie: implement trie pruning

### LINKED ISSUE #2102: Add trie support for "pruned tries"
There are many (?) cases where one would just want to get the root of the trie and not care about older versions of the trie (i.e. `setRoot` to rollback to an older version of the root). However, it seems that currently the trie always keeps track of older versions ("archive" mode), which is not desired in some situations. There should be a flag which allows pruning older versions of the trie - if one writes to the same `key` then the old value of the key should be destroyed and the new value should be added.

```typescript
const { Trie, LevelDB } = require('@ethereumjs/trie')
const { Level } = require('level')

const trie = new Trie({ db: new LevelDB(new Level('MY_TRIE_DB_LOCATION')) })

async function test() {
  await trie.put(Buffer.from('test'), Buffer.from('1'))
  await trie.put(Buffer.from('test'), Buffer.from('2'))
  await trie.put(Buffer.from('test'), Buffer.from('3'))
  await trie.put(Buffer.from('test'), Buffer.from('4'))
  await trie.put(Buffer.from('test'), Buffer.from('5'))
  await trie.put(Buffer.from('test'), Buffer.from('6'))

  // this logs 7 keys but I only want 1 key to be stored with the value 6
  console.log(await trie.db._leveldb.keys().all())
}

test()
```

Thanks @faustbrian for this test case.

jochem-brouwer:
Closes #2102 

TODOs

- [x] Document code better
- [x] Write more tests


packages/trie/test/trie/prune.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    for (let i = 0; i < keys.length; i++) {
+      await trie.put(Buffer.from(keys[i], 'hex'), Buffer.from(values[i], 'hex'))
+    }
+
+    for (let i = 0; i < keys.length; i++) {
+      //console.log(await trie.get(Buffer.from(keys[i], 'hex')))
+    }
+
+    // TODO
+    st.end()
```
jochem-brouwer, line 57: In the end we should add a test on a random trie with 100 keys and 1000 random values which are all set to random keys (so some keys will be updated) and then finally also dropping 20 keys (deleting them). What we do here is: check if all the DB keys in the DB are reached (or the key is the root). If it cannot be reached (using `walkTrie` search using DFS/BFS) then the test has failed as we now have unreachable keys in the DB (which should be pruned).

packages/trie/src/trie/trie.ts
```diff
@@ -176,9 +194,22 @@ export class Trie {
   async del(key: Buffer): Promise<void> {
     await this.lock.wait()
     const { node, stack } = await this.findPath(key)
+    let ops: BatchDBOp[]
+    if (this._pruneTrie) {
+      const deleteHashes = stack.map((e) => this.hash(e.serialize()))
```
jochem-brouwer, line 199: Ah I just realized an edge case; do NOT delete keys if the value is the current value at `key`

packages/trie/test/trie/prune.spec.ts
ScottyPoi: Wrote a test here that I think does what you suggest.

I definitely see that pruning has an effect when keys are updated.  If I run the test with `pruneTee: true`, the size of the DB doesn't grow when keys are updated, but it does if set to `pruneTree: false`.

Also when deleting.  with `pruneTree: true`,  the DB size shrinks when keys are deleted.  With `pruneTree: false`, it does not.

From your comment, though, I expected `walkTrie` to fail if `pruneTree: false` was set, but `walkTrie` runs either way.

holgerd77:
I guess the ultimate test here would be if we run this PR with the client through some couple of thousand tx-including historical blocks (so to not have to deal with reorgs or the like)? 🤔

packages/trie/test/trie/prune.spec.ts
```diff
@@ -52,8 +52,48 @@ tape('Pruned trie tests', function (tester) {
...
+        val = Math.floor(Math.random() * 16384)
+      }
+      values.push(val.toString(16))
+    }
+    for (let i = 0; i < keys.length; i++) {
+      const idx = i < 100 ? i : Math.floor(Math.random() * 100)
+      const key = keys[idx]
+      await trie.put(Buffer.from(key), Buffer.from(values[i]))
+    }
+    st.equals((<any>trie.db)._database.size, 20, 'DB size correct')
```
jochem-brouwer, line 79: How come we know beforehand how many entries are in the DB if the test is random..?

packages/trie/src/trie/checkpoint.ts
```diff
@@ -13,6 +13,9 @@ export class CheckpointTrie extends Trie {
 
   constructor(opts?: TrieOpts) {
     super(opts)
+    if (opts?.pruneTrie === true) {
+      throw new Error('It is not possible to set `pruneTrie` to `true` on a CheckpointTrie')
```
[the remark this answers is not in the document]
jochem-brouwer: You are right, I should have checked if checkpoints wrote to disk or not, but yes they are indeed cached in memory so this error can be removed.

jochem-brouwer:
@holgerd77 You are absolutely right, that would be a great test! I will test locally once this PR is in a ready state.

packages/trie/test/trie/prune.spec.ts
jochem-brouwer: This makes absolutely no sense to me, if I run the test 1000 times it always passes!?

jochem-brouwer:
OK, sadly the tests are failing, so it is not as "trivial" as I thought this would be. There's probably a gotcha somewhere which I'm missing (my initial guess is that some trie nodes are merged, but right now this would also delete these trie nodes)

packages/trie/test/trie/prune.spec.ts
jochem-brouwer: Yes, the `walkTrie` does not fail in this case, because it starts at the root (if we wold not be able to get the root from the db then the trie is definitely corrupted since we then cannot get any key/value pair). It also only checks the root (walkTrie needs you to reschedule tasks, see the `verifyPrunedTrie` test in newer commits)

packages/trie/src/trie/trie.ts
```diff
@@ -59,8 +60,12 @@ export class Trie {
     this.EMPTY_TRIE_ROOT = this.hash(RLP_EMPTY_STRING)
     this._hashLen = this.EMPTY_TRIE_ROOT.length
     this._root = this.EMPTY_TRIE_ROOT
+    if (opts?.persistRoot === true) {
+      this._deleteFromDB = true
```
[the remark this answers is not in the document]
jochem-brouwer: Yes, in the past if you deleted keys (so `Trie.del`) this would also delete the keys from the DB. However, this means you cannot rollback into past tries, since you now have deleted tries. This lead to us not being able to handle reorgs in the client / tests. This is essentially the non-active version of trie pruning, since in active trie pruning you actively delete all nodes which cannot be reachable, but in the `deleteFromDb` it only deletes the node.

jochem-brouwer:
At this point, the trie succesfully prunes on all `put` operations, but not yet on `del` operations.

packages/trie/src/trie/trie.ts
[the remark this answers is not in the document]
jochem-brouwer: I have no super strong opinions here, someone else has to chime in here. The `deleteFromDB` is essentially the pruned-"lite" version, so I think it might make some sense to remove/deprecate it in favor of this new pruning?

jochem-brouwer:
After doing some research it seems that this turns down to whenever we delete a node, and this results in the remaining BranchNode only having one Branch. So we then have to collapse this BranchNode. However this branch node is currently not deleted from the DB. I have to wrap my head around all the cases there are in whatever the BranchNode can collapse into, this is not trivial and it is not yet clear to me when I should delete and when I should not do that.

jochem-brouwer:
I have a super dirty commit here which does a deterministic test and this crashes

https://github.com/ethereumjs/ethereumjs-monorepo/commit/02268520ccae826296bb3205cebcb7128b21472c

Run the `spec.bak.ts` test

packages/trie/src/trie/trie.ts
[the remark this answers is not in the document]
jochem-brouwer: @faustbrian what name would you suggest for this option? (We can tackle this in another PR)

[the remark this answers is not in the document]
gabrocheleau: Also wondering if we actually need to keep deleteFromDB.

packages/trie/src/trie/trie.ts
```diff
@@ -176,9 +206,26 @@ export class Trie {
   async del(key: Buffer): Promise<void> {
     await this.lock.wait()
     const { node, stack } = await this.findPath(key)
+    let ops: BatchDBOp[] = []
+    // Only delete if the `key` currently has any value
+    if (this._useNodePruning && node !== null) {
+      const deleteHashes = stack.map((e) => this.hash(e.serialize()))
+      // Just as with `put`, the stack items all will have their keyhashes updated
+      // So after deleting the node, one can safely delete these from the DB
+      ops = <BatchDBOp[]>deleteHashes.map((e) => {
```
gabrocheleau, line 215:
```suggestion
      ops = deleteHashes.map((e) => {
```

packages/trie/src/trie/trie.ts
```diff
@@ -160,8 +165,33 @@ export class Trie {
...
+        const val = await this.get(key)
+        // Only delete keys if it either does not exist, or if it gets updated
+        // (The update will update the hash of the node, thus we can delete the original leaf node)
+        if (val === null || !val.equals(value)) {
+          // All items of the stack are going to change.
+          // (This is the path from the root node to wherever it needs to insert nodes)
+          // The items change, because the leaf value is updated, thus all keyhashes in the
+          // stack should be updated as well, so that it points to the right key/value pairs of the path
+          const deleteHashes = stack.map((e) => this.hash(e.serialize()))
+          ops = <BatchDBOp[]>deleteHashes.map((e) => {
```
gabrocheleau, line 179:
```suggestion
          ops = deleteHashes.map((e) => {
```

packages/trie/src/trie/trie.ts
```diff
@@ -505,6 +552,19 @@ export class Trie {
       const branchNode = branchNodes[0][1]
       const branchNodeKey = branchNodes[0][0]
 
+      // Special case where one needs to delete an extra node:
```
gabrocheleau, line 555: Does the `_formatNode` method also need to be updated? Not clear on what it does but it makes use of `_deleteFromDB` so perhaps it should also consider `useNodePruning`?

packages/trie/src/types.ts
```diff
@@ -32,6 +32,8 @@ export interface TrieOpts {
   /**
    * Delete nodes from DB on delete operations (disallows switching to an older state root)
    * Default: `false`
+   * Note: this only deletes final leaf nodes from the DB. For more active pruning
+   * (deleting all unreachable key hashes from the DB), use `useNodePruning`.
```
gabrocheleau, line 36: In a couple of instances it is written, in comments:
`(delete operations are only executed on DB with `deleteFromDB` set to `true`)`
I don't think that's true anymore with `useNodePruning`

gabrocheleau (commented):
Review most of the core logic, left some comments. Will do a 2nd round of review tomorrow

packages/trie/src/trie/trie.ts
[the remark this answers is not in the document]
jochem-brouwer: I will remove it, it indeed does not make much sense to have two modes.

packages/trie/src/trie/trie.ts
jochem-brouwer: Hmm this is weird, I have no idea why tests passed, I intended to set `_deleteFromDB` to `true` if `useNodePruning` is true, but I actually messed it up and only did so when `persistRoot` was true. 

However I think that if `formatNode` is called with `remove = true`, then earlier on we already have put that node on the delete stack, since it is part of the stack if you retrieve it by `findPath`

jochem-brouwer:
Addressed the comments, thanks @gabrocheleau @faustbrian :smile: :+1:. Now I have to merge master which is probably going to be a hard one :sweat_smile:

jochem-brouwer:
It took me two rounds to merge master, first I tried in the new VSCode merge editor (this resulted in a lot of conflicts and a lot of failing tests). Then I tried it in the old editor, which resulted in just 4 failing tests (made a single mistake at the merge). The tests should now pass. To ensure I would not have to a lot of conflicts for each commit, I have squashed all original commits prior to merging.

I will now spin up the client with trie pruning enabled to see if this also works.

jochem-brouwer:
Hmm this is not good

```
[08-25|14:24:12] INFO Executed blocks count=50 first=49900 hash=0x0d06…b6c3 td=48261962686294634 hardfork=chainstart last=49950 hash=0xbc7a…939d txs=20 
[08-25|14:24:13] INFO New sync target height=15409086 hash=0xb351…b727 
[08-25|14:24:13] INFO Imported blocks count=50 first=49951 last=50000 hash=0x8680…42fe hardfork=chainstart peers=10 
/home/jochem/Documents/ejs/ethereumjs-monorepo/packages/trie/src/trie/trie.ts:368
      throw new Error('Stack underflow')
            ^
Error: Stack underflow
    at Trie._updateNode (/home/jochem/Documents/ejs/ethereumjs-monorepo/packages/trie/src/trie/trie.ts:368:13)
    at Trie.put (/home/jochem/Documents/ejs/ethereumjs-monorepo/packages/trie/src/trie/trie.ts:193:18)
    at async /home/jochem/Documents/ejs/ethereumjs-monorepo/packages/statemanager/src/stateManager.ts:273:9
```

jochem-brouwer:
`Error: State trie does not contain state root` :cry:

jochem-brouwer:
Ahhhh right we can prune the account trie, but we should not prune storage tries.

jochem-brouwer:
I will sync to block 1M and if there are no errors then I think we can safely assume that it is correct!

jochem-brouwer:
Okay I left the client running for over a hour, and it did not report any errors. I am up to block 173991, there are plenty of txs in there, and there are also contracts being created/written to.

I think we can safely say that this is OK.

jochem-brouwer:
Could you dump your code somewhere? This should definitely not be happening. I tested on client, so accounts there get updated a lot (thus key exists), and no error happens.

jochem-brouwer:
BTW please note, also in my random test: this will use the same key (with a very high probability) at least multiple times.

Also check the first test, this dumps the value 1-6 in the same key and then verifies that DB only has 1 key item, this test also passes..?

jochem-brouwer:
Sounds great @faustbrian, there's probably a gotcha we are missing here :smile: :+1: I do also assume that it is in this "double checkpointed" logic though.

jochem-brouwer:
OK, can unblock this one. @faustbrian and me found a bug in copying the checkpoints, this was a shallow copy and not a deep copy, which resulted in checkpoint-copied Tries actually editing the checkpoints of the original trie as well. Very exciting, this was super interesting to research and I think this is a great feature.

packages/trie/src/types.ts
```diff
@@ -60,13 +54,20 @@ export interface TrieOpts {
    * Store the root inside the database after every `write` operation
    */
   useRootPersistence?: boolean
+
+  /**
+   * Flag to prune the trie. When this is set to `true`,
+   * each time a value is overridden, then the (intermediate) nodes which are now not
+   * reachable anymore will be pruned (deleted) from the trie
```
gabrocheleau, line 61:
```suggestion
   * Flag to prune the trie. When set to `true`, each time a value is overridden,
   * unreachable nodes will be pruned (deleted) from the trie
```
A bit more concise imo

packages/trie/test/index.spec.ts
```diff
@@ -170,11 +170,7 @@ tape('testing deletion cases', function (tester) {
     trie: new Trie(),
     msg: 'without DB delete',
   }
-  const trieSetupWithDBDelete = {
-    trie: new Trie({ deleteFromDB: true }),
-    msg: 'with DB delete',
-  }
-  const trieSetups = [trieSetupWithoutDBDelete, trieSetupWithDBDelete]
+  const trieSetups = [trieSetupWithoutDBDelete]
```
gabrocheleau, line 173: Since we only have one setup, I think we could now do without that array and the `for (const tripeSetup of trieSetups)` below

packages/trie/test/trie/checkpoint.spec.ts
```diff
@@ -197,4 +198,84 @@ tape('testing checkpoints', function (tester) {
     t.deepEqual(await trie.get(k1), v1, 'after revert (first CP): v1')
     t.end()
   })
+
+  /*
+    In this educational example, it is shown how operations on a clone of a trie
+    can be copied into the original trie. This also includes pruning.
+    A practical use-case of this could be to take two clones of a trie
+    and apply different operations on it. Based on the outcome, pick one.
+    If this "checkpoint copy trick" (see test for this "trick") is not used,
```
gabrocheleau, line 207: Perhaps we could link to the "trick" and to the PRs mentioned below?

gabrocheleau (approved):
Looks good to me. Left a few minor comments

holgerd77:
> release

Undecided.

Not totally against inclusion, but I think the somewhat more natural way for such a big and relatively intrusive feature addition would indeed be a minor release 2-3 weeks after, likely with other clean-up and fixing changes from after release publication.

But we can also weight in how annoying this is for you. 🙂

holgerd77:
> @holgerd77 no rush for me. I've already fully integrated it so I can just merge things in my project whenever this gets released and run with the repository until then. Rather have this properly tested instead of creating corrupted tries.

Yeah, thanks, then let's wait with merging.

packages/trie/test/trie/checkpoint.spec.ts
jochem-brouwer: I'm not sure what you mean here, the test itself is the trick?

jochem-brouwer: Or do you mean adding URLs?

gabrocheleau: Yeah I meant adding URLs sorry!

jochem-brouwer:
Great to hear @faustbrian :smile:

jochem-brouwer:
You might have to also explicitly use the `verifyPrunedTrie` method from `prune.spec.ts` so that you can verify that all nodes are reachable in trie :)

jochem-brouwer:
Let's say we are convinced that the pruning works, then what would be the purpose of the method?

jochem-brouwer:
OK, I agree, will add.

jochem-brouwer:
I've changed the name and added is as member of Trie, plus added tests. Like this @faustbrian ?

packages/trie/src/trie/trie.ts
```diff
@@ -730,6 +785,58 @@ export class Trie {
     )
   }
 
+  // This method verifies if all keys in the trie (except the root) are reachable
+  // If one of the key is not reachable, then that key could be deleted from the DB
+  // (i.e. the Trie is not correctly pruned)
+  // If this method returns `true`, the Trie is correctly pruned and all keys are reachable
+  async verifyIsPruned(): Promise<boolean> {
+    const trie = this
```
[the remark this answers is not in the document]
jochem-brouwer: No idea why I added this :thinking:

packages/trie/src/trie/trie.ts
```diff
@@ -730,6 +785,58 @@ export class Trie {
...
+          if (node instanceof ExtensionNode) {
+            // If the value of the ExtensionNode points to the dbkey, then it is found
+            if (node.value().toString('hex') === dbkey) {
+              found = true
+              return
+            }
+            controller.allChildren(node, key)
+          }
+        })
+      } catch (e: any) {
```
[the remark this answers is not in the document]
jochem-brouwer: Interesting, I did not know this, thanks! :smile:

jochem-brouwer:
Addresses your review @faustbrian, could you check and if OK, re-approve?

jochem-brouwer:
@holgerd77 are you OK merging this (removing the blocked label?)?

packages/trie/src/trie/trie.ts
```diff
@@ -740,6 +795,57 @@ export class Trie {
...
+  // This method verifies if all keys in the trie (except the root) are reachable
+  // If one of the key is not reachable, then that key could be deleted from the DB
+  // (i.e. the Trie is not correctly pruned)
+  // If this method returns `true`, the Trie is correctly pruned and all keys are reachable
+  async verifyPrunedIntegrity(): Promise<boolean> {
+    const root = this.root().toString('hex')
+    for (const dbkey of (<any>this)._db.db._database.keys()) {
+      if (dbkey === root) {
+        // The root key can never be found from the trie, otherwise this would
+        // convert the tree from a directed acyclic graph to a directed cycling graph
```
gabrocheleau, line 807: nice, didn't know about those terms

packages/trie/src/trie/trie.ts
```diff
@@ -180,7 +180,31 @@ export class Trie {
       // First try to find the given key or its nearest node
       const { remaining, stack } = await this.findPath(appliedKey)
       // then update
```
gabrocheleau, line 182:
```suggestion
```
Duplicate comment (was moved to 202)

gabrocheleau (dismissed):
LGTM! Left one super minor comment

gabrocheleau (approved):

jochem-brouwer:
Great, thanks for this merge, this was a super cool PR to create :smile:

holgerd77:
Thanks for merging! 🙂

PR #2205, opened 2022-08-22
Block: do not auto-activate hardforkByBlockNumber in static BlockHeader constructor

holgerd77:
This PR reverts on #2081 after a discussion with @acolytec3 on Discord, some extracts:

Example posted by @acolytec3, which was not working any more after the changes:

grafik

Reasoning to revert by @holgerd77:

---

I am coming more and more to the conclusion that this in `BlockHeader.fromRLPSerializedHeader()` was the wrong choice:

```typescript
// If an RLP serialized block header is provided and no `hardforkByBlockNumber` opt is provided, default true to
    // avoid scenarios where no `opts.common` or `opts.hardforkByBlockNumber` is provided and a serialized blockheader
    // is provided that predates London result in a default base fee being added to the block
    // (resulting in an erroneous block hash since the default `common` hardfork is London and the blockheader constructor
    // adds a default basefee if EIP-1559 is active and no basefee is provided in the `headerData`)
    if (opts.hardforkByTTD === undefined) {
      opts.hardforkByBlockNumber = opts.hardforkByBlockNumber ?? true
    }
```

This is again one of these implicit-options behaviors which is not clear from the outside and this also breaks the above example. If someone wants a `hardforkByBlockNumber` behavior, it is very much possible to set the respective option (that's where the option is for). If someone wants a "HF has precedense" behavior, he/she sets the explicit HF (here: London). There one would except that everything stays in the London realm (this expectation is broken here).

On top I am already pretty sure that there *are* these use cases out there where people (mainly these dev tool guys like HH/Ganache) have some mainnet RLP serialized block with some artificial block number which *should* behave in a London (e.g.) context. We had such cases various times. This forcing into the block number logic will likely lead to problems for them.

I would somewhat be inclined to revert (unless I might have forgotten the strong reasoning why we introduced).

---

### DIFF
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -95,15 +95,6 @@ export class BlockHeader {
       throw new Error('Invalid serialized header input. Must be array')
     }
 
-    // If an RLP serialized block header is provided and no `hardforkByBlockNumber` opt is provided, default true to
-    // avoid scenarios where no `opts.common` or `opts.hardforkByBlockNumber` is provided and a serialized blockheader
-    // is provided that predates London result in a default base fee being added to the block
-    // (resulting in an erroneous block hash since the default `common` hardfork is London and the blockheader constructor
-    // adds a default basefee if EIP-1559 is active and no basefee is provided in the `headerData`)
-    if (opts.hardforkByTTD === undefined) {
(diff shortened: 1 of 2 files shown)


acolytec3 (commented):
LGTM

acolytec3 (approved):

PR #2206, opened 2022-08-22
Common, Block, Blockchain, Client: removed Kovan chain support (EOL)

holgerd77:
This PR removes support for Kovan which is EOL for quite some time and support for the only supporting client (Open Ethereum) has also ended.

I would still be inclined to keep Ropsten and Rinkeby since - while both networks are now [deprecated](https://blog.ethereum.org/2022/06/21/testnet-deprecation/) - deprecation announcement is still pretty recent and both networks are still up and running and will likely still be used for testing purposes.

This is WIP, first push is for seeing if Kovan removal works for all packages on CI.

Will add some Common default HF update to the PR.


holgerd77:
Haha, `mainnet` wasn't actually defaulting yet to `Merge`. 😜 Have fixed this in the last commits, lot's of tests to fix, at least one client test is still failing, ready to be picked up if someone wants to.

holgerd77:
Side note: I also thought about removing the `defaultHardfork` entries completely in Common JSON files where the file default HF is the same as the class default HF (so: `merge`) (so the basic mechanism is that it is possible to override the class default HF with one from the chain file for cases where one would want to deviate from the "normal" default HF, e.g. for Rinkeby which didn't (and won't) transition to the Merge).

I then decided to keep the `defaultHardfork` setting also in the files where this is set to `merge` since e.g. the `mainnet` file is often taken as a basis for own HF files and then people can see that this setting exists and they can change.

holgerd77:
@gabrocheleau finally, the real "success" is entering the repository, @ryanio will be reliefed. 😂

Nice, I guess there should be no reason to not merge this here (actually I will do)?

holgerd77 (commented):
LGTM.

Will admin-merge since technically reviewed by @gabrocheleau by continuing on the PR (and I in turn now reviewed the additional changes by Gabriel).

PR #2208, opened 2022-08-22
Update VM/EVM types; remove AsyncEventEmitter extension

jochem-brouwer:
This PR:

- [x] Removes the AsyncEventEmitter extension from VM/EVM and instead puts it into a property `events`
- [x] Update the EVM interface such that it now has an optional `events` property
- [ ] Update VM/EVM such that it is now more typesafe (see #2182)

WIP

### DIFF
--- a/packages/vm/src/vm.ts
+++ b/packages/vm/src/vm.ts
@@ -154,7 +154,9 @@ export class VM {
 
     // We cache this promisified function as it's called from the main execution loop, and
     // promisifying each time has a huge performance impact.
-    this._emit = <(topic: string, data: any) => Promise<void>>promisify(this.events.emit.bind(this))
+    this._emit = <(topic: string, data: any) => Promise<void>>(
+      promisify(this.events.emit.bind(this.events))
+    )
   }
 
   async init(): Promise<void> {
--- a/packages/vm/src/vm.ts
+++ b/packages/vm/src/vm.ts
@@ -31,7 +31,10 @@ import type { StateManager } from '@ethereumjs/statemanager'
  *
  * This class is an AsyncEventEmitter, please consult the README to learn how to use it.
  */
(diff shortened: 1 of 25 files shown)


packages/vm/src/vm.ts
```diff
@@ -31,7 +31,7 @@ import type { StateManager } from '@ethereumjs/statemanager'
  *
  * This class is an AsyncEventEmitter, please consult the README to learn how to use it.
  */
-export class VM extends AsyncEventEmitter<VMEvents> {
+export class VM<EVMType extends EVMInterface = EVM, EEIType extends EEIInterface = EEI> {
```
jochem-brouwer, line 34: So I have now changed `extends EVMInterface = EVMInterface, EEIType extends EEIInterface = EEIInterface` to now use `EVM` and `EEI`. However now I get errors down below on types (see comment) and also tests now throw:

```
tests/api/buildBlock.spec.ts:307:22 - error TS2345: Argument of type 'VM<EVMInterface, EEIInterface>' is not assignable to parameter of type 'VM<EVM, EEI>'.

307     await setBalance(vm, address)
```

I have no idea what causes this..?

packages/vm/src/vm.ts
```diff
@@ -236,20 +243,20 @@ export class VM extends AsyncEventEmitter<VMEvents> {
    * - {@link BlockBuilder.revert}
    */
   async buildBlock(opts: BuildBlockOpts): Promise<BlockBuilder> {
-    return buildBlock.bind(this)(opts)
+    return buildBlock.bind(<any>this)(opts)
```
jochem-brouwer, line 246: Have to cast this to `any` for some reason :thinking:

jochem-brouwer:
I made a git mess, updates go into #2229

PR #2213, opened 2022-08-23
Fix VM Nightly

acolytec3:
The VM Nightly run is intended to use a fresh `package-lock.json` and so must use `npm i` to install deps.  This was accidentally changed in a recent update to the CI.  This PR:
- Reverts the changes to `npm i`
- Adds a workflow_dispatch trigger so we can manually trigger this action when testing
- Fixes a validation error in `common.setHardforkByNumber` that checks for `null` but not `undefined` that was causing some edge case tests to fail in the blockchain test runner.

### DIFF
--- a/.github/workflows/vm-nightly-test.yml
+++ b/.github/workflows/vm-nightly-test.yml
@@ -3,8 +3,6 @@ on:
   schedule:
     - cron: 0 0 * * *
   workflow_dispatch:
-  push:
-    branches: ['fix-vm-nightly-again']
 
 env:
   cwd: ${{github.workspace}}/packages/vm
--- a/.github/workflows/vm-nightly-test.yml
+++ b/.github/workflows/vm-nightly-test.yml
@@ -26,7 +26,7 @@ jobs:
       - run: rm package-lock.json
         working-directory: ${{github.workspace}}
 
-      - run: npm ci
+      - run: npm i
         working-directory: ${{github.workspace}}
 
       - run: npm run test:API
--- a/.github/workflows/vm-nightly-test.yml
+++ b/.github/workflows/vm-nightly-test.yml
@@ -3,6 +3,8 @@ on:
   schedule:
     - cron: 0 0 * * *
   workflow_dispatch:
+  push:
+    branches: ['fix-vm-nightly-again']
 
 env:
   cwd: ${{github.workspace}}/packages/vm
--- a/.github/workflows/vm-nightly-test.yml
+++ b/.github/workflows/vm-nightly-test.yml
@@ -2,6 +2,7 @@ name: VM Nightly
 on:
   schedule:
     - cron: 0 0 * * *
+  workflow_dispatch:
 
 env:
   cwd: ${{github.workspace}}/packages/vm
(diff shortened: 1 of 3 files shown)


acolytec3:
[Proof](https://github.com/ethereumjs/ethereumjs-monorepo/actions/runs/2910328847) that the Nightlies are fixed.  Once all the runs complete, I will remove the PR trigger and push a final commit.

holgerd77:
> [Proof](https://github.com/ethereumjs/ethereumjs-monorepo/actions/runs/2910328847) that the Nightlies are fixed. Once all the runs complete, I will remove the PR trigger and push a final commit.

Note that this needs a branch update now at some point as well.

acolytec3:
> > [Proof](https://github.com/ethereumjs/ethereumjs-monorepo/actions/runs/2910328847) that the Nightlies are fixed. Once all the runs complete, I will remove the PR trigger and push a final commit.
> 
> Note that this needs a branch update now at some point as well.

Yup, once the nightly finishes, I'll rebase and remove the unnecessary triggers.

acolytec3:
I've fixed the original issue but we have some additional issues that have surfaced.  

Running `npx ts-node ./tests/tester --stack-size=1500 "--blockchain" "--fork=HomesteadToEIP150At5" "--verify-test-amount-alltests"` is failing in the long VM blockchain tests.  The issue is that the `hardforkByBlockNumber` is getting a hardfork `chainstart` with an undefined block number where the method is looking for `hf.block === null` instead of `undefined`.

acolytec3:
@holgerd77 can we open a new issue for the `common.hardforkByBlockNumber` issue or do you want me to fix it here?  I'm assuming this got broken by #2206 but since it's a test that isn't run in the CI runs for PRs, it wasn't caught.

acolytec3:
This is ready for final review.  [This run](https://github.com/ethereumjs/ethereumjs-monorepo/runs/7971777746?check_suite_focus=true) proves the VM Nightly is working again.

gabrocheleau (approved):
LGTM. 
Nice catch with the hf.block

holgerd77:
@acolytec3 thanks for the fixes here! 😄

This Common fix still makes me a bit nervous TBH since I am not understanding yet what was the problem and what has then been fixed. Could you describe this a bit more to the point?

So what function call to Common was previously producing a wrong result? Was it `common.getHardforkByBlockNumber(0)` not returning the Chainstart HF? 🤔 Anything else?

Can we eventually add one or two test cases to Common for these cases?

acolytec3:
The common hardfork array had `{ name: chainstart: block: undefined}` for chainstart in the failing tests I noted.

I'm not 100% sure where that undefined block number came from but we should check either way to my way of thinking.

holgerd77:
> The common hardfork array had `{ name: chainstart: block: undefined}` for chainstart in the failing tests I noted.
> 
> I'm not 100% sure where that undefined block number came from but we should check either way to my way of thinking.

Can we please nail this down to a specific test case in Common (which should fail before this PR and pass after)?

Otherwise this still feels dangerous too me.

PR #2215, opened 2022-08-23
refactor(trie): integrate `CheckpointTrie` behaviour into main `Trie` class

faustbrian:
After this I'll have some cleanup pull requests for the trie package in general.


packages/trie/src/trie/trie.ts
```diff
@@ -39,16 +38,18 @@ interface Path {
 export class Trie {
   /** The root for an empty trie */
   EMPTY_TRIE_ROOT: Buffer
-  protected lock: Semaphore
+  private lock: Semaphore
```
holgerd77, line 41: It's mostly better to keep things `protected` rather then `private`, then people can still use this if they want to extend and customize the base class in some way.

holgerd77:
But to add on this a bit: really be prepared that this might not make it any more into the final releases. 🙂 This is so so close before the final finish line and this has some more-than-trivial complexity. So we'll see.

packages/trie/src/types.ts
```diff
@@ -24,11 +25,18 @@ export interface TrieOpts {
   /**
    * A database instance.
    */
-  db?: DB
+  db?: DB | CheckpointDB
+
+  /**
+   * A database instance.
+   */
+  dbStorage?: DB
```
holgerd77, line 33: Ugh.

This makes things highly complicated and I do not think that people "get" why we (they) need two database options here and where their trie is ultimately safed (to be honest: I am not getting both questions either right now, might "get it" when further looking at the code.

So - again writing this without having the internal picture: what I would want here is that there is just *one* database to be passed in and everything else is handled internally and shielded from the user - if e.g. a temporary DB for state changes or whatever needs to be created.

Some side context info: I found this already not so lucky in the original CheckpointTrie code base, so this separate naming for the storage DB with `dbStorage` - I can remember that I stumbled upon this along original PR review - but I could live with that as long as it stayed internal.

holgerd77: Ah.

I had to look at this a couple of times but I think I am slowly getting confidence. 🙂 

Yes, that seems to be at least going in a very good direction - first couldn't believe it - but maybe that's already there to a greater extend.

I would be inclined to also not allow passing in a `CheckpointDB` directly for `db` in the constructor (at least for now, we could add this later if we want). This would make it slicker and more clear and we do not have to explain to people how to create their own CheckpointDBs. (or is there a concrete use case (e.g. also by you) for this *right now*?).

For now always using a CheckpointDB: so is it double/triple-checked that a CP DB (so in particular the `db/checkpoint.ts` class) behaves exactly as a non-CP-DB in an non-checkpointed context? 🤔 Should very muhc be since you could use this without CPing before, but better triple-ask here to get really really sure.

It is just super-essential that people do not get their DBs corrupted.

packages/trie/src/trie/trie.ts
```diff
@@ -72,7 +75,23 @@ export class Trie {
...
+
+    if (opts?.db !== undefined && opts?.persistRoot === true) {
+      if (opts?.root === undefined) {
+        opts.root = (await opts?.db.get(key)) ?? undefined
+      } else {
+        await opts?.db.put(key, opts.root)
+      }
+    }
+
+    return new Trie(opts)
```
holgerd77, line 93: Ah, directly re-internalized 🙂, nice.

packages/trie/test/trie/trie.spec.ts
```diff
@@ -1,30 +1,47 @@
...
+  } else {
+    ROOT_DB_KEY = BASE_DB_KEY
+  }
+
+  let EXPECTED_ROOTS: string
+  if (IS_SECURE_TRIE) {
+    EXPECTED_ROOTS = '8204723ce0fb452b130a282ecc727e07295c18cbd2c2eef33ba9eb9c7a9ded9b'
+  } else {
+    EXPECTED_ROOTS = '99650c730bbb99f6f58ce8b09bca2a8d90b36ac662e71bf81ec401ed23d199fb'
+  }
```
holgerd77, line 38: Ah, thanks for reintroducing these tests and giving this a bit more thought than I had. 👍

holgerd77 (approved):
Thanks a lot, I think I can give this an approval right now! 💯

This is really a super great refactoring, I do realize that I am super glad that we/you now brought this to completion with this. This will just ease a lot of things to not have this three-class-separation any more.

I think people from the community wo are doing deeper work with the Trie library will love this too! 🎉 🙂 ❤️

holgerd77:
> @holgerd77 let me know when this is merged so I can update my other PRs since there will be conflicts

This is merged. 🙂

PR #2216, opened 2022-08-23
refactor(trie): always use `CheckpointDB` as internal database

faustbrian:
The `CheckpointDB` class contains internal checks to either checkpoint data in-memory or call the `DB` implementation so we can always rely on it as the internal database of the trie which will enable some further refactors.


holgerd77:
So, could this be used for not needing the CP option any more? (not have given up on this yet 😜)

PR #2207, opened 2022-08-22
vm: improve tester types and remove isTruthy/isFalsy checks

gabrocheleau:
This PR:

- Improves the types from the vm/tester folder and improve the comments so that the types of the cli args are more obvious. This brings more clarity in usage and also helps document the CLI options a bit better
- Removes all isTruthy/isFalsy from the tester folder

PR #2217, opened 2022-08-23
Verify that hardforks without block number or ttd are not applied in common

acolytec3:
#2213 slightly changed the way that hardforks are checked in `common` when calling `getHardforkBlockNumber` or `setHardforkByBlockNumber` so that a hardfork can be passed in the `ChainConfig` object with either `null` or `undefined` set for the `block` property and these hardforks will be treated as being "not applied" in that `common` instance.  This addresses a change introduced in #2207 that broke certain edge cases in the `vm` test runner where some hardforks were getting set to `undefined` in the transition network tests due to missed type checking.

This PR does two things:
- Updates the test runner to explicitly set a hardfork block to 0 when transitionTests are being run and the hardfork is earlier than the last applied hardfork in the `transitionNetworks` set (which was being done implicitly before #2207).
- Adds new tests in `common` to validate that hardforks passed in the `ChainConfig` object with no defined or null `block` property are tried as "not applied".

### DIFF
--- a/packages/vm/tests/tester/config.ts
+++ b/packages/vm/tests/tester/config.ts
@@ -375,14 +375,12 @@ export function getCommon(network: string): Common {
           | null
         testHardforks.push({
           name: hf.name,
-          // forkHash: hf.forkHash,
-          block: forkBlockNumber,
+          block: forkBlockNumber ?? 0, // If hardfork isn't in transitionForks, activate at 0
         })
       } else {
         // disable the hardfork
         testHardforks.push({
           name: hf.name,
-          // forkHash: hf.forkHash,
           block: null,
         })
       }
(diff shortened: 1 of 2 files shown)


packages/vm/tests/tester/config.ts
```diff
@@ -375,14 +375,12 @@ export function getCommon(network: string): Common {
           | null
```
gabrocheleau, line 375: Let's add `undefined` here also?

gabrocheleau (approved):
Looks good, left one small comment

packages/vm/tests/tester/config.ts
acolytec3: It's already covered by the `forkBlockNumber ?? 0` option below methinks.  We can make it explicit though for future reference.

gabrocheleau: Yeah, I think making it explicit is nice, avoids someone else running into a similar issue eventually

gabrocheleau (approved):

holgerd77:
@acolytec3 thanks for going on with this, finally have some descent relief here. 👍 😋 (sorry, but these kind of changes can just go *so* wrong if we do not pay enough attention. Some code running on the wrong HF in certain cases (which sometimes might trigger something and sometimes not) is just such a nightmare. 😬

PR #2218, opened 2022-08-24
 refactor(trie): change `isCheckpoint` getter to `hasCheckpoints` checker

faustbrian:

`get/set` functions are generally considered black magic in terms of DX because it gives the illusion that you're working with a property and developers might try to set the property too even though only a getter is available. Now you could argue that an IDE can catch this but you shouldn't build your projects around that the assumption that everyone runs an IDE with all kinds of features and plugins. All that can be prevented by simply using functions that make it clear that you're not working with a property.

Besides that the name wasn't very descriptive because the function body reads as `database has more than 0 checkpoints`.


holgerd77 (commented):
This looks good, thanks! 🙂 

This PR still has 34 commits, this is always a situation which always makes me nervous (might be unjustified, anyhow).

Also, you have now opened *a lot* of PRs, lol. 😋 While totally agree that - theoretically - it is really nice and the way "it should be" to have feature changes atomic in different PRs through our situation with the monorepo this unfortunately means a large waiting-cascade to go through one PR after the other, always with this 20-30 min CI wait time in between.

Can we therefore maybe settle to open up *one* new PR and I go through the PRs now one by one and review and then you cherry-pick the respective relevant 1-3 commits from the single PRs and push towards this one new PR? 🤔

That would help a lot to speed up the process, if we at the end only have to wait one time for CI. I would also say this is very much justifiable regarding the type of the changes.

holgerd77:
(also regarding our tight timing regarding the releases)

So, if you are ok with this, you could directly start with this PR and already pack #2226 on top (or other way around).

holgerd77 (approved):
LGTM

PR #2219, opened 2022-08-24
refactor(trie): combine `get/set` for `root` into function

faustbrian:

Same as https://github.com/ethereumjs/ethereumjs-monorepo/pull/2218 to avoid black magic and improve DX.


holgerd77 (commented):
Would be ok with this as well (no approval yet).

Another thing: please please do not open *any* more breaking refactoring PRs beyond the ones already open. This is all really really great and I am confident that we get this managed, but otherwise this will overload us and the whole release process. 😋

Also others (@jochem-brouwer e.g.) also need to do some separate work towards the releases on Friday, and they need to have some stable code basis for this and do not be distracted/overloaded with merge conflicts or rebases too much.

PR #2220, opened 2022-08-24
refactor(trie): replace `get/set` for `key/value` for nodes with function

faustbrian:

Same as https://github.com/ethereumjs/ethereumjs-monorepo/pull/2218 to avoid black magic and improve DX. This together with https://github.com/ethereumjs/ethereumjs-monorepo/pull/2219 and https://github.com/ethereumjs/ethereumjs-monorepo/pull/2218 would be my last pet peeve about the trie package 😅


holgerd77 (commented):
Would be ok with this as well (no approval yet).

PR #2221, opened 2022-08-24
refactor(trie): mark `db` as `protected` to avoid leaky properties

faustbrian:

The `db` property shouldn't be accessible outside of the trie instance because we wrap it so `opts.db` and `this.db` aren't equal and in general everything that can be non-public should be non-public by default. Always start with private and only work up if necessary.

### DIFF
--- a/packages/blockchain/src/genesisStates/index.ts
+++ b/packages/blockchain/src/genesisStates/index.ts
@@ -21,7 +21,7 @@ export interface GenesisState {
  * Derives the stateRoot of the genesis block based on genesis allocations
  */
 export async function genesisStateRoot(genesisState: GenesisState) {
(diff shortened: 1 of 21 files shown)


holgerd77 (commented):
This looks good too, won't give an approval yet.

holgerd77:
@faustbrian ah, can you rather just cherry-pick the essential commits? These are only 7 or so, and then you could just solve some eventual conflicts one-by-one? With this muddling all together I am somewhat afraid that this will really head to some "big messy bunch". 😬

holgerd77:
(so on top of a fresh `master`?)

holgerd77:
Update: ah ok, maybe this works as well, we'll see. 😬

PR #2222, opened 2022-08-24
refactor: replace `semaphore-async-await` dependency with smaller implementation

faustbrian:


holgerd77:
Ah, now you want to go all in, right? 😆

holgerd77 (commented):
Ok, MIT licensed, this should be ok, we can take this in I guess.

PR #2223, opened 2022-08-24
refactor(trie): rename `persistRoot` to `useRootPersistence`

faustbrian:

Bringing this property in line with the naming of other functional modifiers. Also see https://github.com/ethereumjs/ethereumjs-monorepo/pull/2226.

### DIFF
--- a/packages/blockchain/src/genesisStates/index.ts
+++ b/packages/blockchain/src/genesisStates/index.ts
@@ -21,7 +21,7 @@ export interface GenesisState {
  * Derives the stateRoot of the genesis block based on genesis allocations
  */
 export async function genesisStateRoot(genesisState: GenesisState) {
-  const trie = new Trie({ useHashedKeys: true })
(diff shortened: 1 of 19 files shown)


holgerd77 (commented):
This looks good as well (no approval yet).

holgerd77:
Hmm, everything is blocked here. Maybe give this combined-PR-suggestion another thought, I just simply otherwise can't guarantee that we still include these changes in the release since I might not have the time any more to approve one-by-one with an hour or so in-between-wait-time. This is just our CI very very much at its limits, since a single merge is triggering re-runs all over the place, not necessarily the right ones queued first.

I am not seeing so much the revert difficulty problem. All things remain in single commits and if one would run at least the Trie tests after each cherry-pick this should remain under control I guess.

So maybe give this another thought, also important to close the taken-over PRs.

holgerd77:
(and things have mostly already proven to be working by CI having passed on close to all (all?) separate PRs)

PR #2224, opened 2022-08-24
refactor(trie): store `opts` in private property with defaults

faustbrian:

Instead of having separate properties for everything we can use a default object, merge properties and then access everything through that object.

### DIFF
--- a/packages/trie/src/trie/trie.ts
+++ b/packages/trie/src/trie/trie.ts
@@ -37,19 +37,19 @@ interface Path {
 export class Trie {
   private readonly _opts: TrieOptsWithDefaults = {
     deleteFromDB: false,
-    persistRoot: false,
-    useHashedKeys: false,
-    useHashedKeysFunction: keccak256,
+    useKeyHashing: false,
+    useKeyHashingFunction: keccak256,
+    useRootPersistence: false,
   }
 
   /** The root for an empty trie */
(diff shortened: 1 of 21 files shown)


packages/trie/src/trie/trie.ts
```diff
@@ -35,38 +35,39 @@ interface Path {
  * The API for the base and the secure interface are about the same.
  */
 export class Trie {
+  #opts: TrieOptsWithDefaults = {
```
holgerd77, line 38: What is this `#` standing for respectively doing? 🤔

holgerd77 (commented):
Oh *this* is a beauty, wonder if we want to last-minute do for the other libraries too? 🤔 (some substantial amount of work though, so at least 1 person, 1 day, but might be worth it, eventually we can also still be doing between RC and final but before would likely be better).

Anyhow, so would give an approval here as well (not formally yet).

packages/trie/src/trie/trie.ts
holgerd77: Ah, just googled a bit, this is an ES2022 property which is beyond our build target, let's please not introduce any new language constructs here last minute.

This property is super useful, but not anything any more for this round of breaking releases. I can very well imagine a broad set of refactoring around this along the next breaking release round. 🙂

holgerd77:
So I could take this in for inclusion as well.

PR #2226, opened 2022-08-24
refactor(trie): rename `useHashedKeys` to `useKeyHashing`

faustbrian:
The naming convention for `use` properties should be `[use, property, what]` to keep it consistent. This will result in the following properties:

- `useKeyHashing`
- `useRootPersistence`
- `useNodePruning`

This is not only more descriptive but also reads better than:

- `useHashedKeys`
- `usePersistedRoot`
- `prune`

Because you can now read the keys and first see `which` property or component is affected and `what` will happen to or with it.

### DIFF
--- a/packages/trie/src/db/checkpoint.ts
+++ b/packages/trie/src/db/checkpoint.ts
@@ -20,7 +20,7 @@ export class CheckpointDB implements DB {
   /**
    * Is the DB during a checkpoint phase?
    */
-  get isCheckpoint() {
+  hasCheckpoints() {
     return this.checkpoints.length > 0
   }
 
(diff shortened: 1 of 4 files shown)


holgerd77 (changes requested):
Argh, was waiting for this to finalize on CI but I now realized that this accidentally re-introduces the `trie/checkpoint.ts` and `trie/util.ts` files.

This needs to be fixed. 🙂

holgerd77:
(otherwise I totally love the renaming together with the clear reasoning on it - yes, can very much confirm from my own reading experience what you wrote in the initial comment)

holgerd77:
> (otherwise I totally love the renaming together with the clear reasoning on it - yes, can very much confirm from my own reading experience what you wrote in the initial comment)

Lol, I accidentally "thumbed" myself on this (instead of the comment below), but it seems I can't "unthumb" myself. Apparently a GitHub bug. Love that one is actually also repeatedly running into some things like this also on "the big ones" (GitHub) once you leave the trotted path and go into some edge cases. 🙂

holgerd77:
Ok, then let's do 1-by-1 for now. Have updated this branch and would merge this one next.

holgerd77:
> @holgerd77 no conflicts with the checkpoint PR so could admin merge

Ah, I won't admin merge feature PRs with code changes.

PR #2227, opened 2022-08-24
Trie: Property Renaming & Visibility / Getter-Setter & Dependency Removals

faustbrian:
Don't have a title since this is just a big ball of mud 😅


holgerd77:
> Don't have a title since this is just a big ball of mud 😅

I've found one. 😂

holgerd77:
> @holgerd77 maybe we could merge this normal instead of squashing? That way the commits of the squashed PRs will be preserved and still have the ability to revert them one by one.

Ah, yes, for sure, that makes sense!

holgerd77 (approved):
LGTM. 🙂 

And now please write to a preferred chalkboard near you 100x:

"I WILL NOT SUBMIT ANY MORE BREAKING CHANGE PRS TO THE TRIE LIBRARY WITHIN THE NEXT 2 DAYS."

😂 😂 😂

PR #2228, opened 2022-08-24
Common: Add explicit HF block number null checks back, throw on undefined

acolytec3:
Reverts previous changes made in #2217 that allowed a `hardfork.block` value of `undefined` to indicate a not applied hardfork.

Also explicitly checks all the hardforks in the `common.custom` constructor to verify that there are no `undefined` block numbers and throws if so.

### DIFF
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -46,7 +46,7 @@ export interface GenesisBlockConfig {
 
 export interface HardforkConfig {
   name: Hardfork | string
-  block?: number | null
+  block: number | null // null is used for hardforks that should not be applied -- since `undefined` isn't a valid value in JSON
   ttd?: bigint | string
   forkHash?: string | null
 }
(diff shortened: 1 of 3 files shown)


packages/common/src/types.ts
```diff
@@ -46,7 +46,7 @@ export interface GenesisBlockConfig {
 
 export interface HardforkConfig {
   name: Hardfork | string
-  block?: number | null
+  block: number | null // null is used for hardforks that should not be applied -- since `undefined` isn't a valid value in JSON
```
jochem-brouwer, line 49: If I don't want to apply the hf, then I can also just not create the object itself, right?

acolytec3: Yes.

holgerd77: Yes, we mainly have this option in to allow for defining a future HF like `shanghai` and already be able to switch to this HF in a static context and so activate the HF rules without knowing the block number (yet).

holgerd77:
Rebased this via UI.

holgerd77 (approved):
LGTM

PR #2229, opened 2022-08-24
Update VM/EVM types; remove AsyncEventEmitter extension (old)

jochem-brouwer:
This PR:

- [x] Removes the AsyncEventEmitter extension from VM/EVM and instead puts it into a property `events`
- [x] Update the EVM interface such that it now has an optional `events` property
- [x] Update VM/EVM such that it is now more typesafe (see #2182)

WIP


packages/vm/tests/api/buildBlock.spec.ts
```diff
@@ -9,12 +9,15 @@ import { VM } from '../../src/vm'
...
+import type { EEI } from '../../src'
+import type { EVM } from '@ethereumjs/evm'
+
 tape('BlockBuilder', async (t) => {
   t.test('should build a valid block', async (st) => {
     const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Istanbul })
     const genesisBlock = Block.fromBlockData({ header: { gasLimit: 50000 } }, { common })
     const blockchain = await Blockchain.create({ genesisBlock, common, validateConsensus: false })
-    const vm = await VM.create({ common, blockchain })
+    const vm = await VM.create<EVM, EEI>({ common, blockchain })
```
jochem-brouwer, line 20: Why is this necessary?

acolytec3 (dismissed):
Aside from that one annoying `<any>new EVM` typecast, this looks great!

packages/vm/tests/api/buildBlock.spec.ts
acolytec3: It's not :facepalm: .  I misread VSCode's red squigglies.  Will revert.

packages/client/lib/rpc/modules/engine.ts
```diff
@@ -628,11 +629,15 @@ export class Engine {
     if (payloadAttributes) {
       const { timestamp, prevRandao, suggestedFeeRecipient } = payloadAttributes
       const parentBlock = this.chain.blocks.latest!
-      const payloadId = await this.pendingBlock.start(await this.vm.copy(), parentBlock, {
-        timestamp,
-        mixHash: prevRandao,
-        coinbase: suggestedFeeRecipient,
-      })
+      const payloadId = await this.pendingBlock.start(
+        (await this.vm.copy()) as VM<EVM, EEI>,
```
jochem-brouwer, line 633: Does this need to be cast?

jochem-brouwer (commented):
One question, great work @gabrocheleau @acolytec3 :smile:

jochem-brouwer:
I'll have to pull this in to VSCode later today to verify that the type casting as end-user also looks good :+1:

packages/client/lib/rpc/modules/engine.ts
gabrocheleau: Since the `start` method only works with `EVM` and `EEI` (and not the generic types) we have to either typecast it here, or add generics to the start method, which is a massive rabbit hole as it basically requires adding generics in dozens of instances of the client package. That seemed overkill so I went for the typecast .

holgerd77:
Superseeded by #2235, will close.

PR #2230, opened 2022-08-24
client: Fix hardfork issues in beacon sync

g11tech:
After `goerli` merge when the beacon sync was tried on it, multiple hardfork related issues appeared on beacon sync
- Engine's assemble block was  not using correct fork
- Skeleton's call to getBlockHeaders wasn't resulting in correct block and hence the block bodies were also not correct because of which skeleton backfill was breaking
- Reverse block fetcher sync was picking up a peer which seemed to be providing `london` blocks instead of merge causing the fetcher to break again and again instead of just banning that peer.
- Skeleton's getBlock was not using the correct hardfork

This PR fixes the same.

Also plan to refactor the usage of hardforkBy... options in a separate PR

Verifying beacon sync
- [x] Sepolia -- beacon sync init, full skeleton backfilling, skeleton filling canonical chain
- [x] Goerli - beacon sync init, skeleton backfill started (chain too long to verify rest of sequence of events)

### DIFF
--- a/packages/block/src/block.ts
+++ b/packages/block/src/block.ts
@@ -50,14 +50,23 @@ export class Block {
     // parse uncle headers
     const uncleHeaders = []
     const uncleOpts: BlockOptions = {
-      hardforkByBlockNumber: true,
-      ...opts, // This potentially overwrites hardforkByBlocknumber
+      ...opts,
       // Use header common in case of hardforkByBlockNumber being activated
       common: header._common,
       // Disable this option here (all other options carried over), since this overwrites the provided Difficulty to an incorrect value
       calcDifficultyFromHeader: undefined,
-      // Uncles are obsolete post-merge (no use for hardforkByTTD)
+      // This potentially overwrites hardforkBy options but we will set them cleanly just below
+      hardforkByBlockNumber: undefined,
       hardforkByTTD: undefined,
+      hardforkByChainTTD: undefined,
+    }
+    // Uncles are obsolete post-merge, any hardfork by option implies hardforkByBlockNumber
+    if (
+      opts?.hardforkByBlockNumber !== undefined ||
+      opts?.hardforkByTTD !== undefined ||
+      opts?.hardforkByChainTTD !== undefined
+    ) {
+      uncleOpts.hardforkByBlockNumber = true
     }
     for (const uhData of uhsData ?? []) {
       const uh = BlockHeader.fromHeaderData(uhData, uncleOpts)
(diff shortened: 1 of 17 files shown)


holgerd77:
There is only of the two things possible:

1. The PR is merge ready or
2. You are actively working on the PR

What is the case in this case here? 😂

packages/block/src/block.ts
```diff
@@ -114,14 +119,18 @@ export class Block {
...
       calcDifficultyFromHeader: undefined,
+      // This potentially overwrites hardforkBy options but we will set them cleanly just below
+      hardforkByTTD: undefined,
+      hardforkByChainTTD: undefined,
     }
-    if (isTruthy(uncleOpts.hardforkByTTD)) {
-      delete uncleOpts.hardforkByBlockNumber
+    // Uncles are obsolete post-merge, any hardfork by option implies hardforkByBlockNumber
+    if (opts?.hardforkByTTD !== undefined || opts?.hardforkByChainTTD !== undefined) {
+      uncleOpts.hardforkByBlockNumber = true
```
holgerd77, line 133: Ah, that seems like a lot better solution!

g11tech:
> There is only of the two things possible:
> 
>     1. The PR is merge ready or
> 
>     2. You are actively working on the PR
> 
> 
> What is the case in this case here? joy

merge ready! just trying to enhance the coverage :smile:

packages/block/src/types.ts
```diff
@@ -42,6 +42,13 @@ export interface BlockOptions {
    * pointing to a Shanghai block: this will lead to set the HF as Shanghai and not the Merge).
    */
   hardforkByTTD?: BigIntLike
+  /**
+   * Infer the total difficulty by inspecting the difficulty field and determine the HF.
+   *
+   * This option is used specially when we cannnot calculate the total difficulty
+   * for e.g. in the blocks constructed in skeleton/beacon sync
+   */
+  hardforkByChainTTD?: BigIntLike
```
holgerd77, line 51: I would really not want to side introduce yet another `hardforkBy*` option here. This is such a special case and we already have two options here.

Remember, we are *not* developing these libraries for ourselves respectively - at least regarding the priority - these are *not* libraries for the client or something but primarily stand-alone libraries which are used by community developers for various different purposes (so there also shouldn't any client construct/internals mentions in the code docs e.g., like "block in skeleton/beacon sync").

Can you please go with another solution here, e.g. just create a respective Common to be passed in manually in the applying client code? Guess that should be possible.

holgerd77 (changes requested):
One bigger change request on first round.

packages/block/src/types.ts
g11tech: ummmm i actually want to "remove" the hardforkBy options in a followup PR where properly set common needs to be passed from "outside" itself as the only purpose of these options is to set the hardfork. just didn't want to intermix those changes here.

holgerd77: You can also give this a bit of extra time, things are also getting dense on my side regarding the releases since so many things side stepped in today that I think I'll rather not do tomorrow any more but postpone to Monday morning. So you would have tomorrow as well for an update, then we can hopefully still take this in (and otherwise it would also be possible to do the client release 2-3 days later since this technically not necessarily needs to be published in sync with the other releases).

holgerd77: Ah, sorry, changes like these are not possible any more anyhow. Are you aware that we are planning breaking release release candidates tommorow (or now eventually on monday)? These kind of changes would be breaking and after tomorrow there is a development stop on breaking changes.

Generally: can you *please* discuss these kind of ideas on issues before you just go for it? It would really be good if you give people a chance to give some opinion and open some room for discussion.

holgerd77 (approved):
Will only roughly review, since there are time constraints and it seems you have tested this practically thoroughly.

LGTM 🙂

PR #2231, opened 2022-08-25
devp2p: rename tests to spec.ts suffix

gabrocheleau:
Some tests were suffix by .ts instead of .spec.ts


gabrocheleau:
> Why not just remove the two `index.ts` files entirely and just update the `test` script in `package.json` to `npm run tape -- './test/**/*.spec.ts'` like we do in other packages? One less thing to worry about maintaining if we decide to add more devp2p tests.

Done! Can be reviewed again

acolytec3 (approved):
LGTM

PR #2235, opened 2022-08-25
Remove AsyncEventEmitter extension from EVM and VM

jochem-brouwer:
Counter PR against https://github.com/ethereumjs/ethereumjs-monorepo/pull/2229

I'm not sure anymore why we needed the generic types, especially now we can just access these fields from the interface directly (yes, the events are optional, but you can just `vm.events!.on(`)

### DIFF
--- a/packages/vm/tests/api/EIPs/eip-2315.spec.ts
+++ b/packages/vm/tests/api/EIPs/eip-2315.spec.ts
@@ -12,7 +12,7 @@ tape('Berlin: EIP 2315 tests', (t) => {
     let i = 0
     const vm = await VM.create({ common })
 
-    ;(<EVM>vm.evm).on('step', function (step: any) {
+    ;(<EVM>vm.evm).events.on('step', function (step: any) {
       if (test.steps.length > 0) {
         st.equal(step.pc, test.steps[i].expectedPC)
         st.equal(step.opcode.name, test.steps[i].expectedOpcode)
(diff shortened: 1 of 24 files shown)


jochem-brouwer:
Note: the original problem we tried to solve in #2229, #2182 was that before this PR, each time if you wanted to use EVM events, you had to

`(vm.evm as unknown as EVM).on("step", () => { ... });`

This is not what we want. Instead, if we add the events optionally to the interface, one can now;

`vm.evm.events!.on(`

This does therefore not require the generic types as was introduced in the other PRs.

packages/vm/tests/api/EIPs/eip-2315.spec.ts
```diff
@@ -3,16 +3,13 @@ import * as tape from 'tape'
...
-
 tape('Berlin: EIP 2315 tests', (t) => {
   const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Berlin, eips: [2315] })
 
   const runTest = async function (test: any, st: tape.Test) {
     let i = 0
     const vm = await VM.create({ common })
-
-    ;(<EVM>vm.evm).on('step', function (step: any) {
+    vm.evm.events!.on('step', function (step: any) {
```
holgerd77, line 12: Have updated all these unnecessary castings in the tests, should be good to go now. 🙂

jochem-brouwer: Yep this is much cleaner, thanks :smile: :+1:

holgerd77 (approved):
LGTM

PR #2239, opened 2022-08-26
evm: add error which triggers if code size deposit exceeds the maximum size

### LINKED ISSUE #2065: Review EVM error handling messages
In EVM, sometimes the reported error is Out-Of-Gas (OOG), but that is not the case. For instance, https://github.com/ethereumjs/ethereumjs-monorepo/blob/d276fcc103533fdf7eec12b55f202efc74ceeea4/packages/evm/src/evm.ts#L582 violates the code length limit, but it is reported as OOG error.

To do: review all OOG errors and if it is clearly not OOG, change it to something else which is more descriptive.

jochem-brouwer:
Closes #2065

Adds a new error type to EVM, which allows EVM consumers to figure out the difference between (Homestead or later) contract creation goes OOG, or if the deposited code is too large (exceeds maximum code size).

Checked other errors too, but did not find anything "weird" there.

CC @alcuadrado

### DIFF
--- a/packages/evm/src/evm.ts
+++ b/packages/evm/src/evm.ts
@@ -577,7 +577,7 @@ export class EVM implements EVMInterface {
         if (this.DEBUG) {
           debug(`Not enough gas or code size not allowed (>= Homestead)`)
         }
-        result = { ...result, ...OOGResult(message.gasLimit) }
+        result = { ...result, ...CodesizeExceedsMaximumError(message.gasLimit) }
       } else {
         // we are in Frontier
         if (this.DEBUG) {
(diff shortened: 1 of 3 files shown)


holgerd77:
Rebased this via UI.

holgerd77 (approved):
Yeah, I also stumbled upon this strange combination at some point, nice improvement (also somewhat breaking actually, but so be it, small and local enough to still let in I guess, even from RC to final).

alcuadrado:
Late to the party, but thanks for this @jochem-brouwer!

PR #2240, opened 2022-08-27
feat(trie): support manual setting of checkpoints

faustbrian:
## Checkpoints

This is based on https://github.com/ethereumjs/ethereumjs-monorepo/pull/2203/commits/c5fab21431fded59fe8fc050692ce263296901c6 from @jochem-brouwer but packages it with a nicer API for some upcoming additions I would need for more complex uses of the checkpointing mechanism. This also fixes the shallow copy issue mentioned in https://github.com/ethereumjs/ethereumjs-monorepo/pull/2203.

## Database

More complex use-cases of https://github.com/ethereumjs/ethereumjs-monorepo/pull/2203 like mine with a trie and multiple copies require access to the `CheckpointDB` to be able to access pending checkpoints and grab them. The new `database` function introduced in this PR serves as a `getter` and `setter` like the `root` function. This also has the added benefit of enabling the ability to change the database at run-time if someone has that requirement for some reason.

Combined with `setCheckpoints` this will allow you to do things like `trie.database().setCheckpoints(copy.database().checkpoints)` instead of having to hack around on protected and private APIs _(which they should be, don't consider changing that!)_ and it doesn't require adding more proxy functions to the `Trie` class since people with the need for advanced use-cases can access the `CheckpointDB` directly now.

### DIFF
--- a/packages/trie/src/db/checkpoint.ts
+++ b/packages/trie/src/db/checkpoint.ts
@@ -18,6 +18,23 @@ export class CheckpointDB implements DB {
   }
 
   /**
+   * Flush the checkpoints and set the requested checkpoints.
+   *
+   * @param {Checkpoint[]} checkpoints
+   * @memberof CheckpointDB
+   */
+  setCheckpoints(checkpoints: Checkpoint[]) {
+    this.checkpoints = []
+
+    for (let i = 0; i < checkpoints.length; i++) {
+      this.checkpoints.push({
+        root: checkpoints[i].root,
+        keyValueMap: new Map(checkpoints[i].keyValueMap),
+      })
+    }
+  }
+
+  /**
    * Is the DB during a checkpoint phase?
    */
   hasCheckpoints() {
--- a/packages/trie/src/db/checkpoint.ts
+++ b/packages/trie/src/db/checkpoint.ts
@@ -18,10 +18,8 @@ export class CheckpointDB implements DB {
   }
 
   /**
-   * Flush the checkpoints and set the requested checkpoints.
-   *
+   * Flush the checkpoints and use the given checkpoints instead.
    * @param {Checkpoint[]} checkpoints
-   * @memberof CheckpointDB
    */
   setCheckpoints(checkpoints: Checkpoint[]) {
     this.checkpoints = []
--- a/packages/trie/src/trie/trie.ts
+++ b/packages/trie/src/trie/trie.ts
@@ -92,7 +92,7 @@ export class Trie {
     return new Trie(opts)
   }
 
-  public database(db?: DB) {
+  database(db?: DB) {
     if (db !== undefined) {
       if (db instanceof CheckpointDB) {
         throw new Error('Cannot pass in an instance of CheckpointDB')
--- a/packages/trie/src/trie/trie.ts
+++ b/packages/trie/src/trie/trie.ts
@@ -759,7 +759,7 @@ export class Trie {
       root: this.root(),
     })
     if (includeCheckpoints && this.hasCheckpoints()) {
-      trie._db.checkpoints = [...this._db.checkpoints]
+      trie._db.setCheckpoints(this._db.checkpoints)
     }
     return trie
   }
--- a/packages/trie/src/trie/trie.ts
+++ b/packages/trie/src/trie/trie.ts
@@ -13,6 +13,7 @@ import { BranchNode, ExtensionNode, LeafNode, decodeNode, decodeRawNode, isRawNo
 
 import type {
   BatchDBOp,
+  DB,
   EmbeddedNode,
   FoundNodeFunction,
   Nibbles,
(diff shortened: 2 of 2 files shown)


packages/trie/src/db/checkpoint.ts
```diff
@@ -17,6 +17,21 @@ export class CheckpointDB implements DB {
     this.checkpoints = []
   }
 
+  /**
+   * Flush the checkpoints and use the given checkpoints instead.
+   * @param {Checkpoint[]} checkpoints
+   */
+  setCheckpoints(checkpoints: Checkpoint[]) {
+    this.checkpoints = []
```
jochem-brouwer, line 25: I find this rather dangerous, should we not check if we currently are uncheckpointed (i.e. `this.checkpoints.length === 0`)?

packages/trie/src/trie/trie.ts
```diff
@@ -94,6 +92,18 @@ export class Trie {
     return new Trie(opts)
   }
 
+  database(db?: DB) {
```
jochem-brouwer, line 95: I'm not sure do we want the feature that one can set the DB after creating a Trie (so put some values, then change DB, and put more values..?). This seems rather error prone since the new DB most likely does not have the right key value pairs?

packages/trie/src/trie/trie.ts
```diff
@@ -94,6 +92,18 @@ export class Trie {
...
+    if (db !== undefined) {
+      if (db instanceof CheckpointDB) {
+        throw new Error('Cannot pass in an instance of CheckpointDB')
```
jochem-brouwer, line 98: I have asked this before but I forgot, why can we not use a CheckpointDB?

packages/trie/src/trie/trie.ts
```diff
@@ -749,7 +759,7 @@ export class Trie {
       root: this.root(),
     })
     if (includeCheckpoints && this.hasCheckpoints()) {
-      trie._db.checkpoints = [...this._db.checkpoints]
+      trie._db.setCheckpoints(this._db.checkpoints)
```
jochem-brouwer, line 762: Ah right. Yes this is nice. I think you should probably target the branch of #2203 here?

jochem-brouwer (commented):
Some Qs :)

packages/trie/src/db/checkpoint.ts
```diff
@@ -17,6 +17,21 @@ export class CheckpointDB implements DB {
...
+
+    for (let i = 0; i < checkpoints.length; i++) {
+      this.checkpoints.push({
+        root: checkpoints[i].root,
+        keyValueMap: new Map(checkpoints[i].keyValueMap),
```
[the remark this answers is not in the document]
jochem-brouwer: I am not yet super sure here, what if the Map is kind of large, then if the caller does not want to deep-copy it, the method does it anyways which will be expensive :thinking:

packages/trie/src/db/checkpoint.ts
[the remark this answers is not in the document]
jochem-brouwer: Yes, you are right, it is also the responsibility of the caller to check this :)

packages/trie/src/trie/trie.ts
```diff
@@ -46,7 +47,7 @@ export class Trie {
   EMPTY_TRIE_ROOT: Buffer
 
   /** The backend DB */
-  protected _db: CheckpointDB
+  protected _db!: CheckpointDB
```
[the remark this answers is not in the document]
jochem-brouwer: Yup, I realized after I saw that you introduced this new method :)

packages/trie/src/trie/trie.ts
jochem-brouwer: Right, yes, I agree.

packages/trie/src/trie/trie.ts
jochem-brouwer: Yes, it does not really matter anyways.

packages/trie/src/trie/trie.ts
jochem-brouwer: Ah right, I think my confusion comes from the fact that you pass in a DB, and this actually gets converted into a CheckpointDB later on. (I think this is somewhat confusing in our codebase)

packages/trie/src/db/checkpoint.ts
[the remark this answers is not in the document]
jochem-brouwer: Upon giving this some more thought, if someone really cares about this then they should probably just hack it in.

jochem-brouwer (approved):
All Qs are resolved, thanks @faustbrian. I will approve, but let someone else review this as well :) This looks really good.

holgerd77:
Side question (without having a closer look myself): is this breaking in any way? 🤔

Or could we also bundle with the pruning-PR minor release?

jochem-brouwer:
@holgerd77 This is a new feature, and I would say it also fixes a bug, since trie "copies" shallow copied the checkpoints, which mean if you commit on a "copied" trie, it also gets committed on the original trie.

holgerd77:
Rebased this via UI.

holgerd77:
> @holgerd77 This is a new feature, and I would say it also fixes a bug, since trie "copies" shallow copied the checkpoints, which mean if you commit on a "copied" trie, it also gets committed on the original trie.

Ok, then let's take this in.

holgerd77 (approved):
LGTM

Issue #1054, opened 2019-02-04
Trie: Batching multiple leaf proofs


s1na:
When proof for multiple leaves of the trie are required simultaneously, batching the proofs together as proposed [here](https://ethresear.ch/t/optimizing-merkle-tree-multi-queries/4912) can reduce the total overhead. A new method can be added, e.g. `proveBatch` or `proveMultiple` that accepts an array of keys and returns the batched proof. This can potentially be useful in stateless and light clients.

jochem-brouwer:
So this essentially means that we can now create the proof to feed in `verifyRangeProof`, right? Very exciting.

holgerd77:
> So this essentially means that we can now create the proof to feed in `verifyRangeProof`, right? Very exciting.

The above comment is hard to read, this here is an issue, what do you mean with "this" from "so this essentially..."? 🤔

jochem-brouwer:
Yeah I actually meant the issue itself :)

holgerd77:
I will actually close here before it gets forgotten again, this should - latest - be implemented with the completion of #2821 I guess (let me know if I am wrong here).

PR #2243, opened 2022-08-27
Trie: implement createRangeProof

jochem-brouwer:
Related to #1054 

`createRangeProof` allows users to feed a start key and an end key (unhashed), where the returned proof can be fed into `verifyRangeProof` in order to prove the range. (This can also be used in snap sync)

Note: this is not optimal, because the optimized version would require a flat key/value database. This implementation naively walks the trie, and is thus I/O heavy.

Super WIP.

### DIFF
--- a/packages/trie/src/trie/trie.ts
+++ b/packages/trie/src/trie/trie.ts
@@ -800,6 +800,47 @@ export class Trie {
       }
     })
 
+    if (keyValueItems.length === 0) {
+      await this.walkTrie(this.root(), async (_, node, keyProgress, walkController) => {
+        if (keyValueItems.length > 0) {
+          return
+        }
+        const keyCopy = [...keyProgress]
+        // TODO edge case: there are no more keys, how to prove this?
+        // Check if there are nodes right of limitHash, if not, then return proof that there are no keys right of key X
+        // X is the final value
+        if (node instanceof BranchNode) {
+          const children = node.getChildren()
+          for (let i = 0; i < children.length; i++) {
+            const cpy = [...keyCopy]
+            cpy.push(children[i][0])
+            if (nibblesCompare(highKeyNibbles, cpy) < 0)
+              // This node matches the range, so go deeper
+              walkController.onlyBranchIndex(node, keyProgress, children[i][0])
+          }
+        } else if (node instanceof ExtensionNode) {
+          const extKey = [...keyProgress, ...node._nibbles]
+          if (nibblesCompare(highKeyNibbles, extKey) < 0) {
+            walkController.allChildren(node, keyProgress)
+          }
+        }
+        if (node !== null && node.value() !== null && nibblesCompare(highKeyNibbles, keyProgress)) {
+          let targetKey: Nibbles
+          if (node instanceof BranchNode) {
+            targetKey = keyProgress
+          } else if (node instanceof ExtensionNode) {
+            targetKey = [...keyProgress, ...node._nibbles]
+          } else if (node instanceof LeafNode) {
+            targetKey = [...keyProgress, ...node._nibbles]
+          }
+          keyValueItems.push({
+            key: nibblesToBuffer(targetKey!),
+            value: node.value()!,
+          })
+        }
+      })
+    }
+
     let maxValueIndex = -1
     let maxKeyBuffer = Buffer.from('')
     for (let i = 0; i < keyValueItems.length; i++) {
--- a/packages/trie/src/trie/trie.ts
+++ b/packages/trie/src/trie/trie.ts
@@ -824,8 +824,33 @@ export class Trie {
     for (let i = 0; i < keyValueItems.length; i++) {
       // TODO keys should be sorted (not sure how the promise event scheduler schedules all events)
       // Could be rather flakey, so need to sort
-      keys.push(keyValueItems[i].key)
-      values.push(keyValueItems[i].value)
+
+      // For some reason keys get added multiple times
+      // (How is this possible? Need to research, if this also happens in findPath then we need to optimize it!!)
+      let found = false
+      for (let j = 0; j < keys.length; j++) {
+        if (keys[j].equals(keyValueItems[i].key)) {
+          found = true
+          break
+        }
+      }
+      if (!found) {
+        keys.push(keyValueItems[i].key)
+      }
+    }
+
+    keys.sort((a, b) => {
+      return Buffer.compare(a, b)
+    })
+
+    for (let i = 0; i < keys.length; i++) {
+      const key = keys[i]
+      for (let j = 0; j < keyValueItems.length; j++) {
+        if (keyValueItems[j].key.equals(key)) {
+          values.push(keyValueItems[j].value)
+          break
+        }
+      }
     }
 
     return {
--- a/packages/trie/src/trie/trie.ts
+++ b/packages/trie/src/trie/trie.ts
@@ -764,45 +764,11 @@ export class Trie {
       return false
     }
 
-    await this.walkTrie(this.root(), async (_, node, keyProgress, walkController) => {
-      const keyCopy = [...keyProgress]
-      // TODO edge case: there are no more keys, how to prove this?
-      // Check if there are nodes right of limitHash, if not, then return proof that there are no keys right of key X
-      // X is the final value
-      if (node instanceof BranchNode) {
-        const children = node.getChildren()
-        for (let i = 0; i < children.length; i++) {
-          const cpy = [...keyCopy]
-          cpy.push(children[i][0])
-          if (keyChk(cpy))
-            // This node matches the range, so go deeper
-            walkController.onlyBranchIndex(node, keyProgress, children[i][0])
-        }
-      } else if (node instanceof ExtensionNode) {
-        const extKey = [...keyProgress, ...node._nibbles]
-        if (keyChk(extKey)) {
-          walkController.allChildren(node, keyProgress)
-        }
-      }
-      if (node !== null && node.value() !== null && keyChk(keyProgress)) {
-        let targetKey: Nibbles
-        if (node instanceof BranchNode) {
-          targetKey = keyProgress
-        } else if (node instanceof ExtensionNode) {
-          targetKey = [...keyProgress, ...node._nibbles]
-        } else if (node instanceof LeafNode) {
-          targetKey = [...keyProgress, ...node._nibbles]
-        }
-        keyValueItems.push({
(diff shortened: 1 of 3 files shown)


jochem-brouwer:
```
not ok 1 Error: invalid two edge elements proof: the length of firstKey should be equal to the length of lastKey
  ---
    operator: error
    stack: |-
      Error: invalid two edge elements proof: the length of firstKey should be equal to the length of lastKey
          at verifyRangeProof (/home/jochem/Documents/ejs/ethereumjs-monorepo/packages/trie/src/proof/range.ts:494:11)
          at Trie.verifyRangeProof (/home/jochem/Documents/ejs/ethereumjs-monorepo/packages/trie/src/trie/trie.ts:729:28)
          at Test.<anonymous> (/home/jochem/Documents/ejs/ethereumjs-monorepo/packages/trie/test/proof.spec.ts:181:16)
          at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

Right we do not support this. (Geth also does not support this at this point)

jochem-brouwer:
Hmm OK now I get 

```
not ok 1 Error: invalid keys order
  ---
    operator: error
    stack: |-
      Error: invalid keys order
          at verifyRangeProof (/home/jochem/Documents/ejs/ethereumjs-monorepo/packages/trie/src/proof/range.ts:427:13)
          at Trie.verifyRangeProof (/home/jochem/Documents/ejs/ethereumjs-monorepo/packages/trie/src/trie/trie.ts:729:28)
          at Test.<anonymous> (/home/jochem/Documents/ejs/ethereumjs-monorepo/packages/trie/test/proof.spec.ts:184:16)
          at processTicksAndRejections (node:internal/process/task_queues:96:5)
  ...
```

The keys seem to be in right order though, so thats weird.

jochem-brouwer:
OK had to take some time to find some gotcha's here, but this is ready for a first round of reviews. I'm not entirely sure about the implemented API and the code is also a bit dirty, but the tests at least show that the 4 cases (have to check if these are all cases).

@g11tech @scorbajio Do you guys want to give this a first look? I need to add tests same as in #2203 so we can randomly start to test big random tries.

packages/trie/src/trie/trie.ts
```diff
@@ -670,8 +677,8 @@ export class Trie {
    * Creates a proof from a trie and key that can be verified using {@link Trie.verifyProof}.
    * @param key
    */
-  async createProof(key: Buffer): Promise<Proof> {
-    const { stack } = await this.findPath(this.appliedKey(key))
+  async createProof(key: Buffer, ignoreAppliedKey = false): Promise<Proof> {
```
jochem-brouwer, line 680: I'm debating to remove this property and to just add API for:

`createProof(key: Buffer)`

And

`private _createProof(key: Buffer, ignoreAppliedKey = false)`

jochem-brouwer: Note that without this API we get proof for the wrong keys on hashed tries, if we try to proof key X then it is actually going to prove `hash(key)` which has value `null`. (We also do not have the preimages)

packages/trie/src/trie/trie.ts
```diff
@@ -730,6 +737,175 @@ export class Trie {
...
 
+  /**
+   * Return a range proof of trie items between startingHash and limitHash
+   * This proof can be verified using `verifyRangeProof`.
+   * Note, that the proof will always include a proof for the startingHash (even if it does not exist)
+   * The limitHash could not be proved,
+   * @param startingHash This is the starting (hash) key item (note on non-hashed tries this is just a key)
+   * @param limitHash This is the limit (hash) key item (note on non-hashed tries this is just a key)
+   */
+  async createRangeProof(startingHash: Buffer, limitHash: Buffer) {
```
jochem-brouwer, line 748: TODO probably add a limit to how much we want to return also (devp2p/snap compat)

jochem-brouwer:
OK, the random range proof is failing, so there is something wrong here.

g11tech:
> OK had to take some time to find some gotcha's here, but this is ready for a first round of reviews. I'm not entirely sure about the implemented API and the code is also a bit dirty, but the tests at least show that the 4 cases (have to check if these are all cases).
> 
> @g11tech @scorbajio Do you guys want to give this a first look? I need to add tests same as in #2203 so we can randomly start to test big random tries.

thanks @jochem-brouwer for this awesome PR! will have a look and try it out and hopefully we can push this to the goal line this week!

packages/trie/src/trie/trie.ts
holgerd77: Can we give an example here, what a "range" actually *is* in the context of a range proof here? So maybe referencing a super simple trie and give an example start and end and say what keys would be "in range"?

Also not 100% clear to me here just from "reading" the API and comments: is `limitHash` included or not? Why not just call this `startX` and `endX`?

Another naming suggestion: since this *can* be non hashed keys (a naming case we also have in other parts of the libray) why not rather call this (I'll take my naming suggestions from above, but feel free to decide separately) `startKey` and `endKey`? This would be the more generic case covering both base and secure trie cases and the comment additions could also go away?

Generally, thanks a lot here, seems that you guys are really starting to understand the dephts of the trie library, makes me really happy! 🙂

holgerd77: (so, I was e.g. first thinking: so range likely would mean if we have a trie with keys 1,3,5 and 7 a "range" with start 1 and 5 would likely be 1,3,5, but then I thought: no, this very likely isn't right and to simple and "range" must be defined somewhat differently being based on the structure of the trie)

jochem-brouwer: `limitHash` is included, but your name suggestions are great. Your example is right though, if you have keys 1,3,5,7 and you request range request 1-5, you get 1,3,5. (These are all ascending keys and all are included)

g11tech: +1 for startKey and endKey, its clear that way, and I think range directly refers to the actual key structure of the trie, and yes limit (endKey) should be including if there exits an endKey, else the key  before that.
similarly for startKey as per my understanding

g11tech:
@jochem-brouwer i have digged into the PR, looks incredible :fire: , will try next running it to get a feel of the `walkTrie` helper :+1:

scorbajio:
> invalid keys order

I'm not getting this when I run the tests. Maybe one of the randomized trie tests failed.

scorbajio:
> Hmm OK now I get
> 
> ```
> not ok 1 Error: invalid keys order
>   ---
>     operator: error
>     stack: |-
>       Error: invalid keys order
>           at verifyRangeProof (/home/jochem/Documents/ejs/ethereumjs-monorepo/packages/trie/src/proof/range.ts:427:13)
>           at Trie.verifyRangeProof (/home/jochem/Documents/ejs/ethereumjs-monorepo/packages/trie/src/trie/trie.ts:729:28)
>           at Test.<anonymous> (/home/jochem/Documents/ejs/ethereumjs-monorepo/packages/trie/test/proof.spec.ts:184:16)
>           at processTicksAndRejections (node:internal/process/task_queues:96:5)
>   ...
> ```
> 
> The keys seem to be in right order though, so thats weird.

I'm not seeing this error. It might be from the randomized trie tests, though.

packages/trie/test/proof.spec.ts
```diff
@@ -154,3 +155,254 @@ tape('simple merkle proofs generation and verification', function (tester) {
...
+    await trie.put(Buffer.from('3000', 'hex'), Buffer.from('c'))
+    await trie.put(Buffer.from('3300', 'hex'), Buffer.from('c'))
+    await trie.put(Buffer.from('3330', 'hex'), Buffer.from('c'))
+
+    const lKey = Buffer.from('00'.repeat(32), 'hex')
+    const rKey = Buffer.from('9999' + '00'.repeat(30), 'hex')
+
+    const proof = await trie.createRangeProof(lKey, rKey)
+
+    // This should report a part of the trie, somewhere between 1 and 9 elements
```
scorbajio, line 247: Is the return length deterministic? Is it unknown how many key-values are going to be returned or is this just an estimation?

packages/trie/test/proof.spec.ts
```diff
@@ -154,3 +155,254 @@ tape('simple merkle proofs generation and verification', function (tester) {
...
+
+    const proof = await trie.createRangeProof(lKey, rKey)
+
+    // We do NOT need the multiproof here
+    // However, if we want to check this, we have to:
+    // (1) check if there is no item left of lKey
+    // (2) check if there is no item right of rKey
+    // This is trivial in a flat DB, but this is not yet supported
+    // Therefore, probably do not implement this.
+    t.ok(proof.proof.length === 0)
```
scorbajio, line 335: I get an error here:
```
# creates all elements proof
not ok 23 should be truthy
  ---
    operator: ok
    expected: true
    actual:   false
    at: Test.<anonymous> (/ethjs/ethereumjs-monorepo/packages/trie/test/proof.spec.ts:335:7)
    stack: |-
      Error: should be truthy
          at Test.assert [as _assert] (/ethjs/ethereumjs-monorepo/node_modules/tape/lib/test.js:314:54)
          at Test.bound [as _assert] (/ethjs/ethereumjs-monorepo/node_modules/tape/lib/test.js:99:32)
          at Test.assert (/ethjs/ethereumjs-monorepo/node_modules/tape/lib/test.js:433:10)
          at Test.bound [as ok] (/ethjs/ethereumjs-monorepo/node_modules/tape/lib/test.js:99:32)
          at Test.<anonymous> (/ethjs/ethereumjs-monorepo/packages/trie/test/proof.spec.ts:335:7)
          at processTicksAndRejections (node:internal/process/task_queues:95:5)
          at runNextTicks (node:internal/process/task_queues:64:3)
          at processImmediate (node:internal/timers:442:9)
  ...
  ```
  Is there a reason empty proof is expected here?

scorbajio (commented):
Looking great!

packages/trie/test/proof.spec.ts
jochem-brouwer: It is indeed not known beforehand how much items it will return, I should update this test to better reflect was is going on.

packages/trie/test/proof.spec.ts
jochem-brouwer: We do not need these proof items, because /all/ key/values are reported in the key/values :) (So to verify it is trivial, just dump in all key/values and check the root)

jochem-brouwer:
Follow-up PR #2672

PR #2248, opened 2022-08-30
vm: remove isTruthy and isFalsy

gabrocheleau:
Extracts the vm package changes from #2233

### DIFF
--- a/packages/vm/src/runBlock.ts
+++ b/packages/vm/src/runBlock.ts
@@ -2,15 +2,7 @@ import { Block } from '@ethereumjs/block'
 import { ConsensusType, Hardfork } from '@ethereumjs/common'
 import { RLP } from '@ethereumjs/rlp'
 import { Trie } from '@ethereumjs/trie'
-import {
-  Account,
-  Address,
-  bigIntToBuffer,
-  bufArrToArr,
-  intToBuffer,
-  isTruthy,
-  short,
-} from '@ethereumjs/util'
+import { Account, Address, bigIntToBuffer, bufArrToArr, intToBuffer, short } from '@ethereumjs/util'
 import { debug as createDebugLogger } from 'debug'
 
 import { Bloom } from './bloom'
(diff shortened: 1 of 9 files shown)


packages/vm/src/runBlock.ts
```diff
@@ -53,8 +45,8 @@ export async function runBlock(this: VM, opts: RunBlockOpts): Promise<RunBlockRe
 
   if (
     this._hardforkByBlockNumber ||
-    isTruthy(this._hardforkByTTD) ||
-    isTruthy(opts.hardforkByTTD)
+    this._hardforkByTTD !== undefined ||
```
jochem-brouwer, line 48: In `vm.ts` this field is marked as optional, but it looks to me that it is always being set? I think we can improve here by making that field non-optional?

jochem-brouwer (commented):
One possible minor improvement

acolytec3 (approved):
LGTM

packages/vm/src/runBlock.ts
gabrocheleau: I tried making this.hardforkByTTD non-optional (which ends up removing that if altogether) but ended up with a couple failing API tests. I've created an issue for it in case we want to investigate that further (#2260)

holgerd77: No no, just to add here: there are a lot of development use cases where this is not wanted. So these hardforkBy* options should always be present and available to set.

jochem-brouwer: The type of `hardforkByTTD` is `undefined | bigint`, and with `optional` I meant that it should be of type `bigint` (so it should never be `undefined`), but this ruins the logic of setting the hardforks (I realized this when checking #2260). I'm not sure what you mean "should always be available to set", if you want to set the `hardforkByBlockNumber` from `false` to `true` or vice-versa you have to either hack it in, or you have to create a new VM since it is `readonly`.

holgerd77: Plainly meant this should remain an option (in the case this would be in question). 🙂

PR #2250, opened 2022-08-30
tx: remove isTruthy and isFalsy

gabrocheleau:
Extracts the tx package from #2233

### DIFF
--- a/packages/client/lib/rpc/modules/eth.ts
+++ b/packages/client/lib/rpc/modules/eth.ts
@@ -26,7 +26,7 @@ import type { RpcTx } from '../types'
 import type { Block } from '@ethereumjs/block'
 import type { Log } from '@ethereumjs/evm'
 import type { Proof } from '@ethereumjs/statemanager'
-import type { FeeMarketEIP1559Transaction, JsonTx, TypedTransaction } from '@ethereumjs/tx'
+import type { FeeMarketEIP1559Transaction, JsonRpcTx, TypedTransaction } from '@ethereumjs/tx'
 import type { Account } from '@ethereumjs/util'
 import type { PostByzantiumTxReceipt, PreByzantiumTxReceipt, TxReceipt, VM } from '@ethereumjs/vm'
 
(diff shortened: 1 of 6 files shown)


packages/client/lib/rpc/modules/eth.ts
```diff
@@ -26,7 +26,7 @@ import type { RpcTx } from '../types'
 import type { Block } from '@ethereumjs/block'
 import type { Log } from '@ethereumjs/evm'
 import type { Proof } from '@ethereumjs/statemanager'
-import type { FeeMarketEIP1559Transaction, JsonTx, TypedTransaction } from '@ethereumjs/tx'
+import type { FeeMarketEIP1559Transaction, JsonRpcTx, TypedTransaction } from '@ethereumjs/tx'
```
jochem-brouwer, line 29: This is a file in client, I think this is the wrong PR?

jochem-brouwer: Ah OK I see that the interface is moved into tx.

jochem-brouwer (approved):
LGTM!

packages/tx/test/index.ts
```diff
@@ -1,20 +1,19 @@
-import { isTruthy } from '@ethereumjs/util'
 import * as minimist from 'minimist'
 
 const argv = minimist(process.argv.slice(2))
 
-if (isTruthy(argv.b)) {
+if (argv.b === true) {
```
acolytec3, line 5: I'm good with this but just want to confirm, we're not looking to enforce a hard rule on how we write boolean checks where the value we're checking is indeed a boolean like here, right?  So stylistically for us, it's fine to write either:
`if (argv.b)` or `if (argv.b === true)` right?  Or does this eslint rule disallow the first?

gabrocheleau: You are correct! In this case, since `argv.b`, is not typed, I made it explicit, but in normal cases where we're dealing with a `boolean`, there's no need to be explicit. Explicitness is only needed when truthiness/falsiness is involved (and boolean is true/false, not "truthy/falsy").

holgerd77: Ah, that's good to have this confirmed! 🙂

PR #2252, opened 2022-08-30
evm: remove isTruthy and isFalsy

gabrocheleau:
Extracts evm from #2233

### DIFF
--- a/packages/evm/src/evm.ts
+++ b/packages/evm/src/evm.ts
@@ -575,7 +575,7 @@ export class EVM implements EVMInterface {
         if (this.DEBUG) {
           debug(`Not enough gas or code size not allowed (>= Homestead)`)
         }
-        result = { ...result, ...OOGResult(message.gasLimit) }
+        result = { ...result, ...CodesizeExceedsMaximumError(message.gasLimit) }
       } else {
         // we are in Frontier
         if (this.DEBUG) {
@@ -1018,6 +1018,14 @@ export function INVALID_EOF_RESULT(gasLimit: bigint): ExecResult {
     returnValue: Buffer.alloc(0),
     executionGasUsed: gasLimit,
     exceptionError: new EvmError(ERROR.INVALID_EOF_FORMAT),
+  }
+}
+
+export function CodesizeExceedsMaximumError(gasUsed: bigint): ExecResult {
+  return {
+    returnValue: Buffer.alloc(0),
+    executionGasUsed: gasUsed,
+    exceptionError: new EvmError(ERROR.CODESIZE_EXCEEDS_MAXIMUM),
   }
 }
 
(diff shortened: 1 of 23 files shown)


packages/evm/src/evm.ts
```diff
@@ -577,7 +575,7 @@ export class EVM implements EVMInterface {
         if (this.DEBUG) {
           debug(`Not enough gas or code size not allowed (>= Homestead)`)
         }
-        result = { ...result, ...CodesizeExceedsMaximumError(message.gasLimit) }
+        result = { ...result, ...OOGResult(message.gasLimit) }
```
jochem-brouwer, line 578: Why is this removed? The error message is handy, because then you know the reason why the contract could not be created.

jochem-brouwer (commented):
One Q

packages/evm/src/evm.ts
gabrocheleau: Hmm strange, I don't remember changing that. Will take a deeper look

jochem-brouwer: How come this test is not failing? https://github.com/ethereumjs/ethereumjs-monorepo/blob/f1cf6b31e6f1be3c6f71a1b8fe0696fb23cedf13/packages/evm/tests/runCall.spec.ts#L574

gabrocheleau: Ah now I see what happened, I extracted these commits with my branch not updated with that commit: https://github.com/ethereumjs/ethereumjs-monorepo/commit/de21d490e60999f667733625a49cc72dc23f21b8 so these changes were missed

gabrocheleau: Hmm but yeah, not sure why the test wasn't failing?

jochem-brouwer: I checked locally, it failed for me, and the CI should also pick it up (`npm run coverage`)

acolytec3 (approved):
LGTM

packages/evm/src/precompiles/0e-bls12-g2mul.ts
```diff
@@ -13,8 +11,6 @@ const {
 } = require('./util/bls12_381')
 
 export async function precompile0e(opts: PrecompileInput): Promise<ExecResult> {
-  if (isFalsy(opts.data)) throw new Error('opts.data missing but required')
```
holgerd77, line 16: Hmm, I already overlooked a similar case on some other PR, but since it is reoccuring here: I vaguely remember that we introduced this for the JS side of things, since there error messages are unreliable if some parameter is not passed in correctly and things are not guarded by type checking.

Not sure, eventually worth to re-add.

On the other hand, to admit, we do not really have a consistent JS guarding strategy throughout the monorepo anyhow.

Hmm.

Anhow: just wanted to at least mention.

PR #2256, opened 2022-08-30
block: remove isTruthy and isFalsy

gabrocheleau:
Extracts block from #2233

### DIFF
--- a/packages/block/src/header-from-rpc.ts
+++ b/packages/block/src/header-from-rpc.ts
@@ -1,9 +1,7 @@
-import { isTruthy } from '@ethereumjs/util'
-
 import { BlockHeader } from './header'
 import { numberToHex } from './helpers'
 
-import type { BlockOptions } from './types'
+import type { BlockOptions, JsonRpcBlock } from './types'
 
 /**
  * Creates a new block header object from Ethereum JSON RPC.
@@ -11,14 +9,17 @@ import type { BlockOptions } from './types'
  * @param blockParams - Ethereum JSON RPC of block (eth_getBlockByNumber)
  * @param options - An object describing the blockchain
  */
-export function blockHeaderFromRpc(blockParams: any, options?: BlockOptions) {
+export function blockHeaderFromRpc(
+  blockParams: Omit<JsonRpcBlock, 'receiptsRoot'> & { receiptRoot?: string; receiptsRoot?: string },
+  options?: BlockOptions
+) {
   const {
     parentHash,
     sha3Uncles,
     miner,
     stateRoot,
     transactionsRoot,
-    receiptRoot,
+    receiptRoot, // TODO: Investigate dual receiptRoot/receiptsRoot usage.
     receiptsRoot,
     logsBloom,
     difficulty,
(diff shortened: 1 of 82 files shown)


packages/block/src/header-from-rpc.ts
```diff
@@ -1,24 +1,25 @@
...
+  options?: BlockOptions
+) {
   const {
     parentHash,
     sha3Uncles,
     miner,
     stateRoot,
     transactionsRoot,
-    receiptRoot,
+    receiptRoot, // TODO: Investigate dual receiptRoot/receiptsRoot usage.
```
jochem-brouwer, line 22: This TODO can be removed, right?

packages/block/test/block.spec.ts
```diff
@@ -176,7 +176,7 @@ tape('[Block]: block functions', function (t) {
     const common = new Common({ chain: Chain.Goerli, hardfork: Hardfork.Chainstart })
 
     try {
-      blockFromRpc(testDataFromRpcGoerli, [], { common })
+      blockFromRpc(testDataFromRpcGoerli as any, [], { common })
```
jochem-brouwer, line 179: Can cast this as `JsonRPCBlock`?

packages/block/test/from-rpc.spec.ts
```diff
@@ -83,9 +83,13 @@ tape('[fromRPC]:', function (t) {
     'should create a block given json data that includes a difficulty parameter of type integer string',
     function (st) {
       const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.London })
-      const blockDifficultyAsInteger = blockFromRpc(blockDataDifficultyAsInteger, undefined, {
-        common,
-      })
+      const blockDifficultyAsInteger = blockFromRpc(
+        blockDataDifficultyAsInteger as any,
```
jochem-brouwer, line 87: Same in this file, guess those `any`s can be removed/changed?

jochem-brouwer (commented):
Two points about types.

packages/block/test/from-rpc.spec.ts
gabrocheleau: Indeed! done

packages/block/test/block.spec.ts
gabrocheleau: yes! done

packages/block/src/header-from-rpc.ts
gabrocheleau: Yep, done by merging with master

jochem-brouwer (approved):
If CI passes then OK! Thanks!

packages/block/src/header-from-rpc.ts
```diff
@@ -1,15 +1,18 @@
...
 
 /**
  * Creates a new block header object from Ethereum JSON RPC.
  *
  * @param blockParams - Ethereum JSON RPC of block (eth_getBlockByNumber)
  * @param options - An object describing the blockchain
  */
-export function blockHeaderFromRpc(blockParams: any, options?: BlockOptions) {
+export function blockHeaderFromRpc(
+  blockParams: Omit<JsonRpcBlock, 'receiptsRoot'> & { receiptRoot?: string; receiptsRoot?: string },
```
acolytec3, line 13: Do we still need this `receiptRoot` in here?  I thought we decided to do away with it in #2256?  I looked at the Ethereum JSON RPC specs and it's definitely `receiptsRoot` in the [JSON-RPC definition](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getblockbyhash)

gabrocheleau: oh no we don't, messed that up during the "extraction" of the large PR into package-specific PRs, fixed

Issue #2262, opened 2022-08-31
Implement eth/67


jochem-brouwer:
This just removes two methods from `devp2p` so that should not be a super big problem.

rashidmya:
I can give it a go. which two methods to remove?

jochem-brouwer:
Hi @rashidmya, see https://github.com/ethereum/devp2p/blob/master/caps/eth.md#eth67-eip-4938-march-2022 :)

ernestosperanza:
Is this issue open to work?

jochem-brouwer:
Hi @ernestosperanza, we have an open PR from @rashidmya by this, so I would say this is not open at this point.

PR #2263, opened 2022-08-31
removed GetNodeData & NodeData

rashidmya:
Implemented eth/67

Removed:
- GetNodeData 
- NodeData

Issue: #2262

### DIFF
--- a/packages/devp2p/src/protocol/eth.ts
+++ b/packages/devp2p/src/protocol/eth.ts
@@ -84,8 +84,6 @@ export class ETH extends Protocol {
         if (this._version >= ETH.eth62.version) break
         return
 
-      case ETH.MESSAGE_CODES.GET_NODE_DATA:
-      case ETH.MESSAGE_CODES.NODE_DATA:
       case ETH.MESSAGE_CODES.GET_RECEIPTS:
       case ETH.MESSAGE_CODES.RECEIPTS:
         if (this._version >= ETH.eth63.version) break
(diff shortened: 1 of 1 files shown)


packages/devp2p/src/protocol/eth.ts
```diff
@@ -299,8 +297,6 @@ export class ETH extends Protocol {
         if (this._version >= ETH.eth62.version) break
         throw new Error(`Code ${code} not allowed with version ${this._version}`)
 
-      case ETH.MESSAGE_CODES.GET_NODE_DATA:
-      case ETH.MESSAGE_CODES.NODE_DATA:
```
jochem-brouwer, line 303: Hiya, we will need access to these message codes if we want to run anything `< eth/67`

rashidmya:
is this what you mean? @jochem-brouwer

jochem-brouwer:
Yup, if you also could include a test which shows these methods are not reachable if one is on `eth/67` then we can start to review it :)

holgerd77:
Please let's not merge this in any more before the final major version monorepo releases early next week.

Will give this a "Blocked" label for now, can be removed once final releases have been published.

packages/devp2p/src/protocol/eth.ts
rashidmya: shouldn't this also be if `>= eth/63`?

jochem-brouwer: Yes, so any anything between 63 and 66 should be able to access it.

rashidmya: fixed it & added a test to show methods not reachable on `eth/67`

jochem-brouwer:
I will go ahead and rebase this, @holgerd77 I think we can remove the blocked label here? I would like to go ahead and implement eth/68 as well. (Need this for hive tests)

acolytec3:
Will approve once CI passes

PR #2247, opened 2022-08-29
Invalid README examples

rodrigoherrerai:
1. Blockchain cannot be used to initiate an EVM, it is not "EVMOpts". 
2. "gasUsed" is not of type "ExecResult"

PR #2266, opened 2022-08-31
evm: Update examples

acolytec3:
Following @holgerd77's well made point from #2247, this addresses the out of date EVM examples (also one of the last open items from #2032).  

- Updates example in README to reflect current usage
- Updates `npm run examples` script in package.json to run EVM examples and not VM examples
- Adds complete example of `runCode` to example scripts
~~- [ ] Add functioning browser example~~


acolytec3:
@holgerd77 @jochem-brouwer @g11tech how important is it to have a specific functioning browser example right now?  I've spent a couple of hours experimenting with `browserify` and `rollup` trying to get a corrected version of `run-code-browser.js` to work with no luck.  Browserify struggles with some oddity in a node_modules dependency and rollup just doesn't seem able to correctly bundle everything together.

I'm sure webpack would work (since we use it for the client today and I was able to get that to build) but that seems like overkill for this since it would require a whole setup with a webpack config and various additional details.  Does it make sense to just remove the browser specific example at this point and put a `todo` to add one back in the future or should I get a setup together with webpack that will work since we do want the EVM to run in the browser (and know it does based on the browser client example working)

jochem-brouwer:
OK I've also tested this for a few minutes but I'm no browser expert so I will not continue here. However did we remove most `tsconfig.browser.json` files from the packages? I thought that most of them got browser builds + tests as well?

I think this is semi-OK if we do not update it (but then lets remove the file entirely) and immediately create an issue, and point to the deleted file: restore browser examples. (This should most likely be done for all packages). On the other hand we are not getting much opened issues where people are having problems with browser builds.

acolytec3:
> OK I've also tested this for a few minutes but I'm no browser expert so I will not continue here. However did we remove most `tsconfig.browser.json` files from the packages? I thought that most of them got browser builds + tests as well?

We had separate `browser` configs mainly for various specific things we had to tell Typescript to do since we were on ES5.  Now that we're targeting ES2020, those are no longer an issue so had no need for browser specific configuration (except for the client but that's a special case).

I know it "works" in the browser because all of our karma tests are passing and those run in the browser.  It's more about how do I get all the dependencies to bundle correctly in a single file without setting up a whole project.  We used to be able to just say "run `browserify some-example.js -o index.js` and then serve that via a script tag in the browser, but I can't convince browserify to bundle everything correctly when we include VM, EVM, Blockchain, and Common because of some obscure ESM thing in one of the dependencies in `ethash` of all places (since I guess `vm` or `blockchain` depends on Ethash somewhere along the way).

> I think this is semi-OK if we do not update it (but then lets remove the file entirely) and immediately create an issue, and point to the deleted file: restore browser examples. (This should most likely be done for all packages). On the other hand we are not getting much opened issues where people are having problems with browser builds.

This is what I was thinking I would do, more or less.  We definitely want to figure out a clean way to provide an example script of the EVM in the browser (I know it works because we're using the in Ultralight now via `vm.runTx`).

acolytec3:
> OK I agree here, the main issue is that we cannot bundle "a single file" without setting up the whole project. Could you open an issue for this where you describe the problem in detail? This looks handy for browser develops so we should take a look ASAP.

Sure though I don't know if it's really a problem per se.  In my experience, "most" web3 front-end/browser oriented developers are already using a framework like React or Svelte or whatever and by necessity they won't be using single file bundles the way this example is structured.  As such, they'll be using webpack or parcel or some other bundler and then it's just a matter of letting your bundler handle the bits needed to package up your deps (including `evm`) into a single package.  I think the bigger issue for us is that `evm` requires `statemanager`, `blockchain`, and `vm` to work since we need an `EEI` in order to stand up the `evm`.  Getting that dummy EEI implemented would be a big win.

jochem-brouwer:
Oh yes right, the dummy EEI, I forgot about it, will pick it up tomorrow!

jochem-brouwer:
I am OK with this but someone else has to weigh in on this browser thing.

packages/evm/package.json
```diff
@@ -33,7 +33,7 @@
     "coverage": "c8 --all --reporter=lcov --reporter=text --reporter=text-lcov npm run coverage:test",
     "coverage:test": "npm run test; cd ../vm; npm run tester -- --state",
     "docs:build": "typedoc --options typedoc.js",
-    "examples": "ts-node ../../scripts/examples-runner.ts -- vm",
+    "examples": "ts-node ../../scripts/examples-runner.ts -- evm",
```
holgerd77, line 36: Ah, nice. 🙂

holgerd77 (approved):
Cool, this looks good, thanks a lot! 👍

PR #2268, opened 2022-09-01
trie: adds profiling script back

acolytec3:
Adds back a script used for profiling the `trie` package that seems to have been removed at some point during the latest round of breaking release work.

Now does some benchmarking with `keccak256` from `noble/hashes` vs `cryptocoin`

Run `npx ts-node benchmarks/random.ts` to see the output.

### DIFF
--- a/packages/trie/benchmarks/random.ts
+++ b/packages/trie/benchmarks/random.ts
@@ -2,6 +2,14 @@
 import { keccak256 } from 'ethereum-cryptography/keccak'
 import { Trie } from '../dist'
 
+const createKeccak = require('keccak')
+const bench = require('micro-bmark')
+const { run, mark } = bench
+
+const oldHash = (val: Buffer) => {
+  const hash = createKeccak('keccak256').update(val).digest()
+  return hash
+}
 // References:
 // https://eth.wiki/en/fundamentals/benchmarks#the-trie
 // https://gist.github.com/heikoheiko/0fa2b322560ba7794f22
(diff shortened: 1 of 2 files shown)


acolytec3:
Ah, you're right and I missed that when I was adding this.  This was triggered by a conversation with Hardhat about performance and I'm happy to close once we wrap up our discussion.  My script as written will show the performance differences between a pure JS hashing function versus one that uses the NodeJS builtins. and that's the piece we need to iron out (i.e. whether to do something like you did for trie but expand it to all the libraries that do hashing (`tx`, `block`, `evm`).

acolytec3:
> @acolytec3 when you say `NodeJS builtins`, do you mean https://github.com/cryptocoinjs/keccak? I don't see any access to the native `crypto` module.

Yeah, you're right. I didn't read closely enough. But even if not nodejs built ins, it's using CPP code for the hashing and the performance  benefits are there in your numbers.

paulmillr:
@faustbrian bcrypto fetches node.js headers from nodejs.org, compiles them, a huge mess, really. Adding a native dependency to the `npm i` is meh. Also it can't be used in browsers. All for - what? 20% perf boost?

macOS 12.5, node.js 18.8, Apple M2

```
noble-unrolled x 15 ops/sec @ 64ms/op ± 5.06% (min: 60ms, max: 89ms)
cryptocoin x 12 ops/sec @ 80ms/op ± 3.09% (min: 78ms, max: 100ms)
bcrypto x 18 ops/sec @ 52ms/op ± 6.85% (min: 48ms, max: 84ms)
```

acolytec3:
I don't have strong opinions at this point as this is in line with my past experience going from native deps to noble.  But it does prove the point that the current noble implementation does put a performance drag on our libraries when hashing intensive code is run. That's not a criticism but just the reality of using pure JavaScript instead of CPP, as @paulmillr has noted elsewhere. I think the near term outcome for us should be to add an optional hashing parameter to each of our libraries that do hashing or require secp256k1 signing/verifying so that users can opt for more performant hashing implementations if desired (similarly to what is currently done for `trie`. I'll open a new issue for it accordingly and we can discuss further on the particulars.

acolytec3:
Brian, care to add `https://www.npmjs.com/package/unrolled-nbl-hashes-sha3` to your comparison and run the performance checks one more time?

acolytec3:
Closing for now as this code is already covered elsewhere.

holgerd77:
Cool, thanks for this extensive investigation! 👍 I talked with Cayman from Lodestar yesterday and they are taking (have taken) a similar for the crypto libraries, so provide a base library setup/config but then allow to override this (also other crypto primitives beside hashing) in the constructor of a library.

PR #2269, opened 2022-09-05
client: Fix snap protocol encoding decoding as per spec

g11tech:
Match snap protocol's expected encoding/decoding as per the spec.

### DIFF
--- a/packages/client/lib/net/protocol/snapprotocol.ts
+++ b/packages/client/lib/net/protocol/snapprotocol.ts
@@ -1,19 +1,30 @@
-import { bigIntToBuffer, bufferToBigInt } from '@ethereumjs/util'
+import {
+  accountBodyFromSlim,
+  accountBodyToSlim,
+  bigIntToBuffer,
+  bufferToBigInt,
+  setLengthLeft,
+} from '@ethereumjs/util'
 
 import { Protocol } from './protocol'
 
 import type { Chain } from '../../blockchain'
 import type { Message, ProtocolOptions } from './protocol'
-import type { AccountData as AccountDataBody } from '@ethereumjs/util'
+import type { AccountBodyBuffer } from '@ethereumjs/util'
 
 interface SnapProtocolOptions extends ProtocolOptions {
   /* Blockchain */
   chain: Chain
+  /**
+   * If to convert slim body received of an account to normal while decoding.
+   * Encoding is always converted to slim
+   */
+  convertSlimBody?: boolean
 }
 
 type AccountData = {
   hash: Buffer
-  body: AccountDataBody
+  body: AccountBodyBuffer
 }
 
 type GetAccountRangeOpts = {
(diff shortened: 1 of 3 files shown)


packages/client/lib/net/protocol/snapprotocol.ts
```diff
@@ -1,19 +1,30 @@
...
+import type { AccountBodyBuffer } from '@ethereumjs/util'
 
 interface SnapProtocolOptions extends ProtocolOptions {
   /* Blockchain */
   chain: Chain
+  /**
+   * If to convert slim body received of an account to normal while decoding.
+   * Encoding is always converted to slim
+   */
+  convertSlimBody?: boolean
```
jochem-brouwer, line 22: Why would we want to keep the slim body?

packages/client/lib/net/protocol/snapprotocol.ts
```diff
@@ -124,11 +154,12 @@ export class SnapProtocol extends Protocol {
       name: 'GetStorageRanges',
       code: 0x02,
       response: 0x03,
+      // [reqID: P, rootHash: B_32, accountHashes: [B_32], startingHash: B, limitHash: B, responseBytes: P]
       encode: ({ reqId, root, accounts, origin, limit, bytes }: GetStorageRangesOpts) => {
         return [
           bigIntToBuffer(reqId ?? ++this.nextReqId),
-          root,
-          accounts,
+          setLengthLeft(root, 32),
```
jochem-brouwer, line 161: I am honestly a bit surprised that the spec marks this as a `B_32`. We could save some bytes if the hash has leading zeros, why not shave those off?

packages/client/lib/net/protocol/snapprotocol.ts
```diff
@@ -239,6 +286,7 @@ export class SnapProtocol extends Protocol {
     super(options)
 
     this.chain = options.chain
+    this.convertSlimBody = options.convertSlimBody
```
jochem-brouwer, line 289: This defaults to `undefined` / `false`?

jochem-brouwer (commented):
Some questions regarding the convert slim body option.

packages/client/lib/net/protocol/snapprotocol.ts
g11tech: yes, i wasn't sure what we want default to be, so chose that this be passed as explicit flag to make any conversions

packages/client/lib/net/protocol/snapprotocol.ts
g11tech: is slim body not useful at all? then we can always convert and remove this flag totally

jochem-brouwer (approved):
I will approve here, I am not entirely sure if we want to have the slim account feature support and if it defaults to `false` (when decoding), but I think we will quickly learn in the case that we do/do not want it to remove/keep it. This is one of these cases which we will "learn by practice" :)

Thanks @g11tech :smile:

holgerd77:
Just to note, if you guys want fee free to merge in again.

PR #2271, opened 2022-09-07
feat(util): implement `assertIsUint8Array`

faustbrian:


holgerd77:
I don't think this is a safe thing to do since one can't anticipate all the use cases here. If this is used for specifically checking for a Buffer on a value and after that Buffer-specific functionality is called things would break.

So my current tendency is to not merge this in.

We can work our way through the libraries and see where we can take in `Uint8Array` instead of `Buffer` in a first round for feature functionality methods though without side effects, at least that would be my first-thought suggestion where it would be possible to start on this a bit.

holgerd77:
Can we then instead just add another method `assertIsUint8Array()` and switch to that where it is possible? I guess I just don't like the fact that the function after this change just doesn't do any more what it claims to do "by name".

Just searched, this method is only used within Util it seems anyhow (doesn't mean we can delete though).

packages/util/src/account.ts
```diff
@@ -194,9 +194,9 @@ export const isValidChecksumAddress = function (
  * @param from The address which is creating this new address
  * @param nonce The nonce of the from account
  */
-export const generateAddress = function (from: Buffer, nonce: Buffer): Buffer {
-  assertIsBuffer(from)
-  assertIsBuffer(nonce)
+export const generateAddress = function (from: Buffer, nonce: Uint8Array): Buffer {
```
holgerd77, line 197: `from` forgotten

packages/util/src/address.ts
```diff
@@ -77,7 +77,7 @@ export class Address {
    * @param salt A salt
    * @param initCode The init code of the contract being created
    */
-  static generate2(from: Address, salt: Buffer, initCode: Buffer): Address {
+  static generate2(from: Address, salt: Buffer, initCode: Uint8Array): Address {
```
holgerd77, line 80: `salt`

packages/util/src/bytes.ts
```diff
@@ -107,8 +106,8 @@ const stripZeros = function (a: any): Buffer | number[] | string {
  * @param a (Buffer)
  * @return (Buffer)
  */
-export const unpadBuffer = function (a: Buffer): Buffer {
-  assertIsBuffer(a)
+export const unpadBuffer = function (a: Uint8Array): Buffer {
```
holgerd77, line 109: I have again this naming problem here. Additionally I don't think it's a good thing for conversion functions to change the type in between, so here, to then have `Uint8Array` as input and return a `Buffer`. This is also hard to deprecate at some point.

I wouldn't want to do too quick decisions on these kind of things, since implications on how we can deal with this later on can be significant.

So I think if we do we should likely go the a bit harder way and also add here parallel `unpadUint8Array()` and the like functionality and then switch over one-by-one at some point deprecate (in the sense of: write a deprecation note, so not directly remove) the Buffer functions (if we decide that we want that. We can also have both in parallel).

holgerd77:
Generally, do not give too much expectations on having this merged in soon. We have spent months and months on these Buffer vs Uint8Array discussions.

If we want to pick this up again I want to make sure that we have a gentle, consistent and non-breaking strategy (optimally written down as an issue or something) throughout the whole monorepo before we act upon this, especially so close after the breaking releases.

paulmillr:
Thoughts on renaming to assertIsUi8a?

jochem-brouwer (commented):
General Q just to confirm: all `Buffer`s are `Uint8Array`s but not all `Uint8Array`s are `Buffer`s?

paulmillr:
Note: Buffer.slice is not compatible with Uint8array.slice — first returns original memory, second copies it.

holgerd77:
Find it a bit of a pity that this has been closed. Maybe we can revive parts of it at some point.

PR #2272, opened 2022-09-07
chore: enable `@typescript-eslint/return-await` rule for `eslint`

faustbrian:
Enables https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/return-await.md which ensures that `await` is used appropriately, like for example `return await` inside a `try/catch` to make sure the catch can handle any errors.


gabrocheleau (approved):
lgtm!

PR #2274, opened 2022-09-07
Update upgrade guide to match recipe

### LINKED ISSUE #2273: Upgrade guide recommends errant import
Hi! I'm working on upgrading Ganache to the latest release. The `@ethereumjs/trie` package's [first line of code in the upgrade guide](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/trie/UPGRADING.md#implementation) to implement a `LevelDB` class has the following:
```TypeScript
import { isTruthy } from '@ethereumjs/util'
```

This line doesn't work because `@ethereumjs/util` doesn't have an `isTruthy` export. In fact, a search of the codebase indicates that an example is the only reference to `isTruthy`. Looking at previous versions, I'm not finding any definition there either.

acolytec3:
Fixes #2273 

I tested this revised example locally with the `index.spec.ts` test to verify that importing a DB that implements this example still passes our tests.

### DIFF
--- a/packages/trie/UPGRADING.md
+++ b/packages/trie/UPGRADING.md
@@ -123,7 +123,6 @@ Note that the `--save-exact` flag will pin these dependencies to exact versions 
 Fortunately the implementation does not require any input from you other than copying and pasting the below code into a file of your choosing in any given location. You will then import this to any area in which you need to instantiate a trie.
 
 ```ts
-import { isTruthy } from '@ethereumjs/util'
 import { MemoryLevel } from 'memory-level'
 
 import type { BatchDBOp, DB } from '@ethereumjs/trie'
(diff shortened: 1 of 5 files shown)


gabrocheleau:
This could also be updated in `packages/trie/examples/level.js`
Also noticed that we have the isTruthy/isFalsy methods still referenced in `packages/util/docs/README.md`. Perhaps that could be removed while we're at it.

gabrocheleau (changes requested):
See comment

acolytec3:
> This could also be updated in `packages/trie/examples/level.js` Also noticed that we have the isTruthy/isFalsy methods still referenced in `packages/util/docs/README.md`. Perhaps that could be removed while we're at it.

Good catches.  I updated the example and re-ran the docs builder to update.  Should be clear now.

gabrocheleau (approved):

PR #2276, opened 2022-09-09
Remove extra computation

rodrigoherrerai:
We extend the Memory inside of the write function, so it is unnecessary to do it here.


gabrocheleau (approved):

gabrocheleau:
Looks good to me!

Issue #2105, opened 2022-08-05
Tiny Fixes Meta Issue


holgerd77:
Will reopen this issue in favor of #905 to give this more visibility again since this has proven to be very handy.

To recap: this issue is for collecting small tasks which are too tiny to address on its own. Instead they can be done along other PRs, also possible to go into this list and work down tasks in batch from time to time.

So feel free to directly edit this initial post and add, check or remove. 🙂 

## StateManager

- ~See if `eslint-disable-next-line implicit-dependencies/no-implicit` for `js-sdsl` depdendency import from [PR #2285](https://github.com/ethereumjs/ethereumjs-monorepo/pull/2285/files#diff-27995baca2e3cab4168cca2adf0592d1ef14e2feb853cd540972551098c8f612) can be removed (should be possible)~

## EVM

- [ ] Investigate if EVM-step-event stop issue from #2290 could need additional code docs along the `step` event docs

## Common

- [ ] Mark `shanghai` HF as experimental/not-finished as mentioned in https://github.com/ethereumjs/ethereumjs-monorepo/pull/2459/files#r1059136654

## Util

- [ ] Document new `intToBuffer()` method in `bytes` module, see https://github.com/ethereumjs/ethereumjs-monorepo/pull/2409/files#diff-82756c063b1cbe4840166da5b6381648c96064b1867c3ce972196581fd0b4b66

gabrocheleau:
I guess this can be closed?

holgerd77:
> I guess this can be closed?

No, this is a permanent issue to track tiny issues which are not worth to be dealt with on its own. Just add to the initial comment if you have something.

PR #2281, opened 2022-09-13
block: clean up block RPC test data

### LINKED ISSUE #2280: Block: Clean up block RPC test data
Some test data from the block package does not seem to conform with the standard RPC interface, for e.g. `"sealFields"`. which from a quick research seems to be a now unused parity-specific field ([src](https://pkg.go.dev/github.com/INFURA/go-ethlibs/eth)).

Since we don't make use of them, we could easily clean those up and remove the need for typecasting in some of the tests:
e.g. `blockDataDifficultyAsInteger as unknown as JsonRpcBlock`

gabrocheleau:
Resolves #2280


holgerd77:
Rebased this via UI.

acolytec3 (approved):
LGTM!

packages/block/test/testdata/testdata-from-rpc-difficulty-as-integer.json
```diff
@@ -25,77 +20,73 @@
...
       "input": "0x",
       "nonce": "0xfc02d",
-      "publicKey": "0x5d8af02c52b1fb40e45d6e04ae2351d443d1f7327e9b6dbb5ae1d918414899c625690c4c236889e23d9705f9db88876542b959d947a3497a30d0d922ee30015f",
       "r": "0x059934eeace580cc2bdc292415976692c751f0bcb025930bd40fcc31e91208f3",
       "raw": "0xf870830fc02d8504a817c80083015f9094c4f381af25c41786110242623373cc9c7647f3f1880e139507cd50c0188026a0059934eeace580cc2bdc292415976692c751f0bcb025930bd40fcc31e91208f3a077ff34a10a3de0d906a0363b4bdbc0e9a06cb4378476d96dfd446225d8d9949c",
       "s": "0x77ff34a10a3de0d906a0363b4bdbc0e9a06cb4378476d96dfd446225d8d9949c",
       "to": "0xc4f381af25c41786110242623373cc9c7647f3f1",
       "transactionIndex": "0x3",
-      "v": 1,
+      "type": "0x0",
```
holgerd77, line 88: Really cool clean-up PR, thanks @gabrocheleau! ❤️ 🙂

Short half-nit: so these are legacy txs if I get this right, correct? Now with the `v` values somewhat corrected to 27/28. I wonder about this `type` field: is this an allowed combination, is this confirmed? Respectively do some (all) clients do this (now?), to send back a type 0 for legacy txs? 🤔

I am just wondering because I also wouldn't want to introduce some data combinations in test data where "future generations" [TM] (lol) will assume this to be correct and wonder about the combination. 😋

In case this can (or even: should) be removed, you can just add the task TODO to the tiny-fixes-issue #2105.

acolytec3: I can't speak for all clients but at least Infura returns in this format for legacy transactions.  I just pulled one from the dao fork era and it came back with `type: 0x0, v: 0x1b`

holgerd77: Ah, interesting.

PR #2284, opened 2022-09-14
TX: fix TX decoding when some values are actually arrays

jochem-brouwer:
This PR fixes:

-> Decoding of any Tx where any of a value is an array instead of a value
-> Decoding EIP2930/EIP1559 in the case that `v` or `chainId` is an array

In all cases, these TXs are wrongly formatted and they should throw when trying to decode.


packages/tx/src/eip2930Transaction.ts
```diff
@@ -158,6 +160,11 @@ export class AccessListEIP2930Transaction extends BaseTransaction<AccessListEIP2
       gasPrice: this.gasPrice,
     })
 
+    const notArray = { ...txData }
```
jochem-brouwer, line 163: TODO we copy this, how to optimize?

jochem-brouwer: @acolytec3 we can now just dump in txData into the static func right?

packages/tx/test/inputValue.spec.ts
```diff
@@ -136,3 +146,30 @@ tape('[Transaction Input Values]', function (t) {
     st.end()
   })
 })
+
+tape('[Invalid Array Input values]', (t) => {
+  const txTypes = [0x0, 0x1, 0x2]
+  for (const txType of txTypes) {
+    const tx = TransactionFactory.fromTxData({ type: txType })
```
jochem-brouwer, line 153: OK this is a great test. Shall we generalize this and do a check for all keys which we do not want array?

Bonus: we should also check for the access lists, that in the access list the account is not array. Also for storage array: the elements should be keys and not arrays. (I think we do not check this ATM)

packages/tx/src/eip2930Transaction.ts
acolytec3: Should be able to. That's what the tests are doing and why since were failing since we were just blindly checking everything that was passed in

packages/tx/test/inputValue.spec.ts
acolytec3: I thought it already did, unless I'm missing something. It creates a transaction with default values and then overrides one by one with an array. Unless I'm imagining things, that should hit all of the `txData` interface properties which shouldn't be arrays.

jochem-brouwer:
Ready for review.

packages/tx/src/eip1559Transaction.ts
```diff
@@ -174,6 +175,11 @@ export class FeeMarketEIP1559Transaction extends BaseTransaction<FeeMarketEIP155
       maxPriorityFeePerGas: this.maxPriorityFeePerGas,
     })
 
+    const notArray = { ...txData }
+    delete notArray.accessList
+
+    BaseTransaction._validateNotArray(notArray)
```
g11tech, line 181: you could just do `BaseTransaction._validateNotArray(txData)` , was typescript giving issues for doing this?

packages/tx/src/eip2930Transaction.ts
g11tech: hmmm seems like you are talking the same thing here, what issues are you facing if you just use txData in the validation call?

Issue #2283, opened 2022-09-14
Migrate functional-red-black-tree to js-sdsl


ZLY201:
Hey! I'm the developer of Js-sdsl. Official website: https://js-sdsl.github.io/

Now, we published the version 4.1.4.

I see you are using functional-red-black-tree.

In [benchmark](https://js-sdsl.github.io/#/test/benchmark-result?id=rbtree), we have confirmed that Js-sdsl is several times faster than functional-red-black-tree.

I've submitted a PR to eslint and has been accepted.

See https://github.com/eslint/eslint/issues/16255 and https://github.com/eslint/eslint/pull/16267.

I am willing to submit a pull request for this change.

Looking forward to your reply! :D

acolytec3:
Thanks for the note. We're definitely open to a PR though will want to see the benchmarks using our own code before merging. Note, this dependency is also used in our ` statemanager` package.

ZLY201:
OK, I'll show you the benchmark result later.

ZLY201:
Benchmarks here:

- perf/migrate-rbtree

https://github.com/ZLY201/ethereumjs-monorepo/actions/runs/3054050866/jobs/4925454999

```bash
> @ethereumjs/vm@6.0.0 benchmarks
> node --max-old-space-size=4096 ./benchmarks/run.js benchmarks mainnetBlocks:10

Benchmarking started...
Running 'mainnetBlocks':
Total number of blocks in data set: 15
Number of blocks to sample: 10
Block 9422905 x 1[8](https://github.com/ZLY201/ethereumjs-monorepo/actions/runs/3054050866/jobs/4925454999#step:6:9),375 ops/sec ±2.02% (82 runs sampled)
Block [9](https://github.com/ZLY201/ethereumjs-monorepo/actions/runs/3054050866/jobs/4925454999#step:6:10)422906 x 17,719 ops/sec ±4.70% (82 runs sampled)
Block 9422907 x 17,477 ops/sec ±4.75% (85 runs sampled)
Block 9422908 x 16,721 ops/sec ±9.67% (85 runs sampled)
Block 94229[10](https://github.com/ZLY201/ethereumjs-monorepo/actions/runs/3054050866/jobs/4925454999#step:6:11) x [17](https://github.com/ZLY201/ethereumjs-monorepo/actions/runs/3054050866/jobs/4925454999#step:6:18),523 ops/sec ±1.95% (80 runs sampled)
Benchmark run finished.
```

[...]

PR #2285, opened 2022-09-14
perf: migrate rbtree to js-sdsl

ZLY201:
Linked issue: https://github.com/ethereumjs/ethereumjs-monorepo/issues/2283.

### DIFF
--- a/packages/statemanager/src/cache.ts
+++ b/packages/statemanager/src/cache.ts
@@ -1,8 +1,10 @@
 import { Account } from '@ethereumjs/util'
 
-import type { Address } from '@ethereumjs/util'
+// eslint-disable-next-line implicit-dependencies/no-implicit
+import { OrderedMap } from 'js-sdsl'
 
-const Tree = require('functional-red-black-tree')
+import type { Address } from '@ethereumjs/util'
+import type { OrderedMapIterator } from 'js-sdsl'
 
 export type getCb = (address: Address) => Promise<Account | undefined>
 export type putCb = (keyBuf: Buffer, accountRlp: Buffer) => Promise<void>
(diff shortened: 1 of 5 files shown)


packages/statemanager/src/cache.ts
```diff
@@ -1,8 +1,10 @@
 import { Account } from '@ethereumjs/util'
+// eslint-disable-next-line implicit-dependencies/no-implicit
+import { OrderedMap } from 'js-sdsl'
 
 import type { Address } from '@ethereumjs/util'
-
-const Tree = require('functional-red-black-tree')
+// eslint-disable-next-line implicit-dependencies/no-implicit
```
holgerd77, line 6: Hmm, these shouldn't be necessary I guess, maybe you've forgotten to rebuild or to restart your editor or something? 🤔

Anyhow: won't trigger a new CI run just for that, have added to our #2105 issue for small fixes.

packages/vm/package.json
```diff
@@ -69,7 +69,6 @@
     "async-eventemitter": "^0.2.4",
     "debug": "^4.3.3",
     "ethereum-cryptography": "^1.1.2",
-    "functional-red-black-tree": "^1.0.1",
```
holgerd77, line 72: Oh. Seems this was simply forgotten after the state manager extraction work. 🙂 

Good catch! 👍

holgerd77 (approved):
Hmm, have looked at the local code changes, all looks reasonable.

Also the broader VM test suite is passing with this, so this is a strong indicator that things are working as expected since StateManager is really used heavily in there.

So: think we can merge. 😄 

Again, thanks a lot, great PR! 🙏

PR #2289, opened 2022-09-15
statemanager: remove eslint-disable from imports

ScottyPoi:
Fixes tiny issue from #2105

* See if eslint-disable-next-line implicit-dependencies/no-implicit for js-sdsl depdendency import from PR #2285 can be removed (should be possible)

Removing the eslint-disable comments had no effect as expected.


holgerd77 (approved):
LGTM, thanks @ScottyPoi! 👍 🙂

PR #2293, opened 2022-09-18
client: Abstract out cl manager from the engine methods impl

g11tech:
Instead of mixing the connection manager status updates with the implementation of the engine methods (and was not being consistently called before all returns in the engine methods), 
this PR "middleware" the connection method updates keeping the implementation cleaner and consistent


packages/client/lib/rpc/modules/engine.ts
```diff
@@ -484,7 +494,7 @@ export class Engine {
    */
   async forkchoiceUpdatedV1(
     params: [forkchoiceState: ForkchoiceStateV1, payloadAttributes: PayloadAttributesV1 | undefined]
-  ): Promise<ForkchoiceResponseV1> {
+  ): Promise<ForkchoiceResponseV1 & { headBlock?: Block }> {
```
acolytec3, line 497: The return type definition should be updated to include the optional headblock

acolytec3 (changes requested):
Looks good except for one place where the method documentation should be updated.

acolytec3 (approved):
LGTM

holgerd77:
Ah, that's a nice clean up here. 🙂

PR #2296, opened 2022-09-20
fix(trie): handle root key and value during pruned integrity verification

faustbrian:
Resolves an issue where the function would fail when only the root key is present as a key but not the root value.

### DIFF
--- a/packages/trie/src/trie/trie.ts
+++ b/packages/trie/src/trie/trie.ts
@@ -799,7 +799,7 @@ export class Trie {
   // (i.e. the Trie is not correctly pruned)
   // If this method returns `true`, the Trie is correctly pruned and all keys are reachable
   async verifyPrunedIntegrity(): Promise<boolean> {
-    const roots = [this.root().toString('hex'), this.hash(ROOT_DB_KEY).toString('hex')]
+    const roots = [this.root().toString('hex'), this.appliedKey(ROOT_DB_KEY).toString('hex')]
     for (const dbkey of (<any>this)._db.db._database.keys()) {
       if (roots.includes(dbkey)) {
         // The root key can never be found from the trie, otherwise this would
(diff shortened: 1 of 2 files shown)


packages/trie/test/trie/prune.spec.ts
```diff
@@ -176,4 +176,65 @@ tape('Pruned trie tests', function (tester) {
...
+      }
+
+      st.ok(await trie.verifyPrunedIntegrity(), 'trie is correctly pruned')
+      st.ok(trie.root().equals(KECCAK256_RLP), 'trie is empty')
+
+      let dbKeys = 0
+      for (const _dbkey of (<any>trie)._db.db._database.keys()) {
+        dbKeys++
+      }
+      st.ok(dbKeys === 0, 'db is empty')
```
jochem-brouwer, line 237:
```suggestion
      st.ok(dbKeys === 1, 'db is empty')
```

This fails since root is saved, key count should be 1.

packages/trie/src/trie/trie.ts
```diff
@@ -799,9 +799,9 @@ export class Trie {
   // (i.e. the Trie is not correctly pruned)
   // If this method returns `true`, the Trie is correctly pruned and all keys are reachable
   async verifyPrunedIntegrity(): Promise<boolean> {
-    const root = this.root().toString('hex')
+    const roots = [this.root().toString('hex'), this.hash(ROOT_DB_KEY).toString('hex')]
```
jochem-brouwer, line 802:
```suggestion
    const roots = [this.root().toString('hex'), this.appliedKey(ROOT_DB_KEY).toString('hex')]
```

Should this not be the applied key? If trie is not hashed then this returns the wrong `ROOT_DB_KEY`.

jochem-brouwer (approved):
LGTM

holgerd77:
Just for my understanding: what are the practical implications of this issue?

Issue #2171, opened 2022-08-17
Add `state` tests to EVM package


acolytec3:
We just explore moving the `State Tests` from `ethereum/tests` from the `vm` to the `evm` package since notionally we don't need the `vm` package to run these tests.  This would involve cloning the `stateTestRunner` from `vm` and then making whatever modifications are needed to allow it to run the state tests with `evm.runCall` instead of `vm.runTx`

jochem-brouwer:
This is a super great idea which I also somewhat had when working on implementing retesteth. I think this is what we would want in the end. Especially the state tests are indeed made for this. This would also greatly improve EVM coverage :+1:

holgerd77:
> This is a super great idea which I also somewhat had when working on implementing retesteth. I think this is what we would want in the end. Especially the state tests are indeed made for this. This would also greatly improve EVM coverage 👍

Ah, this gives me some confirmation, thanks! 🙂 

@acolytec3 also, yes, makes a lot of sense to just adopt the test runner to use EVM directly and not do this circumvents I suggested with going into the VM folder or the like.

PR #2299, opened 2022-09-21
EVM-State-Tests

ScottyPoi:
Referencing issue #2171

Run `State Tests` from `ethereum-tests` using just the EVM.

This PR adapts the `test runner` from VM/tests for this use.
  * tester/config.ts
  * tester/testLoader.ts
  * tester/index.ts
  * tester/util.ts
  * tester/scripts/state-test-run-test.sh

* `tester/runners/GeneralStateRunner.ts`
  * Use of `VM` was removed
  * `VM.runTX()` was replaced by `EVM.runCall()*`
    * `*EVM.runCall()` does not, on it's own, change the state.   
    * `runAltTx.ts`  is adapted from `VM.runTx()` to simulate all the state change that happens in the transaction.

### DIFF
--- a/packages/evm/package.json
+++ b/packages/evm/package.json
@@ -43,6 +43,11 @@
     "tape": "tape -r ts-node/register --stack-size=1500",
     "test": "npm run tape -- './tests/**/*.spec.ts'",
     "test:browser": "karma start karma.conf.js",
+    "tester": "ts-node ./tests/tester --stack-size=1500",
+    "test:state": "npm run tester -- --state",
+    "test:state:allForks": "echo 'Chainstart Homestead dao TangerineWhistle SpuriousDragon Byzantium Constantinople Petersburg Istanbul MuirGlacier Berlin London ByzantiumToConstantinopleFixAt5 EIP158ToByzantiumAt5 FrontierToHomesteadAt5 HomesteadToDaoAt5 HomesteadToEIP150At5 BerlinToLondonAt5' | xargs -n1 | xargs -I v1 npm run test:state -- --fork=v1 --verify-test-amount-alltests",
+    "test:state:selectedForks": "echo 'Homestead TangerineWhistle SpuriousDragon Petersburg Berlin London' | xargs -n1 | xargs -I v1 npm run test:state -- --fork=v1 --verify-test-amount-alltests",
+    "test:state:slow": "npm run test:state -- --runSkipped=slow",
     "tsc": "../../config/cli/ts-compile.sh"
   },
   "dependencies": {
--- a/packages/evm/tests/tester/config.ts
+++ b/packages/evm/tests/tester/config.ts
@@ -1,0 +1,504 @@
+import { Chain, Common, Hardfork } from '@ethereumjs/common'
+import * as path from 'path'
+
+/**
+ * Default tests path (git submodule: ethereum-tests)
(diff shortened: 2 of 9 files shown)


ScottyPoi:
The CI won't automatically run this test yet.
Can be tested manually with (from packages/evm): 
`npm run test:state`
`npm run test:state: allForks`
`npm run test:state:SelectedForks`
`npm run test:state:slow`

holgerd77:
Hi Scotty,
thanks for starting with this, great! 😄

The greatest problem I have for now is that we replicate far (far) too much code with this for now. So > 1.000 LOC is by far too much and couldn't justify such a refactor.

So let's take a step back here again: I guess the main thing which we want here is to have the state tests attribute to the test coverage for the EVM. Is this correct (also question to everyone else)? 🤔

Somewhat of a related side question: is it really settled that the state tests should rather "work" for the EVM coverage and the blockchain tests for the VM coverage (also rather a question for everyone)? An eventual alternative would be to have *both* test types rather work "for" the EVM and have the remaining VM code solely covered by API tests. I am a bit indifferent, I would think the more important step is to have at least one of these tests work for EVM in the first place.

So my current judgement is that these test runners (state/blockchain) are too tighly coupled and we shouldn't tear this apart, seeing here to how much code replication this leads (even taking into account that e.g. the state runner file can likely be deleted later on the VM side).

Since it seems even necessary - as seen in the PR - to replicate close to all of the `VM.runTx()` logic this really makes no sense. So I would rather pledge that we (as we do now, lol) take the `VM.runTx()`  logic for the test execution and keep the whole test runner "suite" together and the sole thing we do is to take the state test execution commands `test:state:*` out of the VM `package.json` and move it over to the EVM and eventually adopt a bit if this is not directly working. Then we have "the best" [TM] of both worlds, getting the state coverage for the EVM, the blockchain coverage for the VM, and have/keep the test runner integrated without the need of code replication.

The one thing we can think about is to extract the whole test runner (everything in `tester`, not sure @jochem-brouwer, `retesteth` as well?) into a separate folder `evm-vm-test-runner` or the like on the `packages` file level.

We can also think about a generally somewhat adopted file structure regarding (VM/EVM) test related logic integrating the `ethereum/tests` submodule as well like (again on `packages` level):

- `evm-vm-tests`
  - `test-runner`
  - `retesteth`
  - `ethereum-tests`

What do you guys think? 🤔

ScottyPoi:
> The greatest problem I have for now is that we replicate far (far) too much code with this for now. So > 1.000 LOC is by far too much and couldn't justify such a refactor.

I agree, this seems impractical / unhelpful

ScottyPoi:
> The one thing we can think about is to extract the whole test runner (everything in tester, not sure @jochem-brouwer, retesteth as well?) into a separate folder evm-vm-test-runner or the like on the packages file level.

I though about this, and it does solve the issue of replicating the test runner code.

To run the State tests with just EVM without reproducing VM.runTx()... I don't know :thinking:

holgerd77:
> To run the State tests with just EVM without reproducing VM.runTx()... I don't know 🤔

No no, we *run* with `VM.runTx()` (or at least that would be the suggestion), this would then also execute the EVM internally which then counts to code coverage when triggered from the EVM `package.json` which I assume should then be fine.

PR #2300, opened 2022-09-21
common: Relocate geth genesis, state parsers to common, blockchain respectively

### LINKED ISSUE #1708: Common: add support for genesis to add code/storage/nonces to genesis accounts

g11tech:
Relocate geth genesis, state parsers to common, blockchain respectively

Partially addresses  #1708 as `parseGethGenesisState` can set these objects
- [x] code
- [x] storage

Nonces still need to be checked

Note (from @holgerd77): this also fixes a severe bug in the `Blockchain.create()` method.


packages/common/README.md
```diff
@@ -195,6 +195,24 @@ const common1 = new Common({
...
+```typescript
+import { Common, parseGethGenesis } from '@ethereumjs/common'
+const chainParams = await parseGethGenesis(chainData, 'customChain')
+const common = new Common({
+  chain: 'customChain',
+  customChains: [chainParams],
+})
+// You either have genesis block hash or calculate it from using the geth state via `blockchain` module
+// const genesisBlockHash = ...
+common.setForkHashes(genesisBlockHash)
```
holgerd77, line 213: Hi Gajinder,
this looks already pretty great all over the place! 😄 

Just for context/understanding: are fork hashes generally not included in `genesis.json` files? 🤔 Or is this just an "emergency mode" addition in case they are not included but normally they are?

g11tech: hey @holgerd77 , thanks :heart:  the geth genesis doesn't has any forkHashes, so yes this would need to be set explicitly for custom chains based out of geth configs, and also can work as the `fallback` as you mentioned :+1:

holgerd77:
Can you add at least one test in `Common` directly using the new functionality?

Also just to mention: when I suggested this I thought about adding a new static constructor like `Common.fromGethConfig()` or the like. If this has no logic hickups I would still find this a bit more elegant, but I can also life with the current solution. Also to mention that I guess we can still always additionally add such a constructor afterwards in addition to the existing helper function (that might generally be good to have both ways/options for flexibility).

Just some thoughts, if you also think this is a good idea to streamline things feel free to additionally add such a constructor if this doesn't collide with option types or the like. If you don't then just leave for now.

g11tech:
> Can you add at least one test in `Common` directly using the new functionality?
> 
> Also just to mention: when I suggested this I thought about adding a new static constructor like `Common.fromGethConfig()` or the like. If this has no logic hickups I would still find this a bit more elegant, but I can also life with the current solution. Also to mention that I guess we can still always additionally add such a constructor afterwards in addition to the existing helper function (that might generally be good to have both ways/options for flexibility).
> 
> Just some thoughts, if you also think this is a good idea to streamline things feel free to additionally add such a constructor if this doesn't collide with option types or the like. If you don't then just leave for now.

sounds good @holgerd77 :slightly_smiling_face:  will update :+1: and add some tests as well

g11tech:
added a test for kiln genesis's block root match (sourced from an old geth build), seems to be not matching :thinking: , currently debugging for fix

packages/blockchain/src/genesisStates/index.ts
```diff
@@ -36,7 +36,7 @@ export async function genesisStateRoot(genesisState: GenesisState) {
         account.codeHash = Buffer.from(keccak256(toBuffer(code)))
       }
       if (storage !== undefined) {
-        const storageTrie = new Trie()
+        const storageTrie = new Trie({ useKeyHashing: true })
```
holgerd77, line 39: Oh. The v6 final release clean-up bugs keep trickling in (still on a very modest level though).

Great find! 🙂

g11tech: yea it took quite a digging to find this one :sweat_smile:

g11tech:
@holgerd77 PR should be good to merge :pray:

packages/blockchain/src/genesisStates/index.ts
holgerd77: Just following up the call-path here,so this is actually not such a "mild" bug being triggered directly in `Blockchain.create()` in certain circumstances, right? 🤔 "Only" in genesis storage scenarios, but anyhow. Guess some bugfix releases are really getting closer to be realized/done at some point.

g11tech: yes in genesis storage scenarios only! thats why i digged around and used kiln genesis as well as kiln genesis block (finally had to find an old geth build to get it :sweat_smile: )

packages/common/tests/utils.spec.ts
```diff
@@ -0,0 +1,89 @@
+import * as tape from 'tape'
+
+import { Common } from '../src/common'
+import { parseGethGenesis } from '../src/utils'
+
+tape('[Utils/Parse]', (t) => {
+  t.test('should parse geth params file', async (t) => {
+    const json = require(`../../client/test/testdata/geth-genesis/testnet.json`)
```
holgerd77, line 8: Hmm. Won't make it a blocker here but this is a bit dangerous.

So if someone reorganizes test data in client, this will (at least locally) remain unnoticed when someone is working "client only" and running the client tests and the like. Will break on CI then so somewhat ok, but nevertheless we should generally avoid and in doubt rather take over.

At some point we might also want to consolidate testdata a bit more (or at least discuss), and do a dedicated directory `testdata` on `packages` folder level or the like and in turn keep the test data a bit more curated and high quality level. Just some thoughts for now though.

holgerd77 (approved):
Nice, looks really really great, we have been waiting for such a functionality/constructor for quite some time, @ryanio would/will love this as well 🙂, quite some pre-work being done to enable this and this will really ease (expand on) use cases for the other libraries (easily instantiate a `Block` from a custom chain where you only have a genesis.json file and the block RPC e.g., these kind of things).

So, great! Will merge.

PR #2302, opened 2022-09-22
block: Fix from rpc methods

acolytec3:
This addresses several typing issues in the `blockFromRpc` method and its helpers so that block responses from RPC endpoints are now correctly handled. 

Changes pulled from #2059


g11tech (dismissed):
LGTM!

packages/block/src/from-rpc.ts
```diff
@@ -1,35 +1,30 @@
...
     txParams.to !== null && txParams.to !== undefined
       ? setLengthLeft(toBuffer(txParams.to), 20)
       : null
 
-  // v as raw signature value {0,1}
-  // v is the recovery bit and can be either {0,1} or {27,28}.
-  // https://ethereum.stackexchange.com/questions/40679/why-the-value-of-v-is-always-either-27-11011-or-28-11100
-  const v: number = txParams.v
-  txParams.v = v < 27 ? v + 27 : v
+  txParams.v = toType(txParams.v, TypeOutput.BigInt)!
```
gabrocheleau, line 27:
```suggestion
  txParams.v = toType(txParams.v, TypeOutput.BigInt)
```
I don't think the ! is necessary

gabrocheleau (dismissed):

g11tech (approved):

packages/block/src/from-rpc.ts
```diff
@@ -1,35 +1,30 @@
...
 import { blockHeaderFromRpc } from './header-from-rpc'
-import { numberToHex } from './helpers'
 
 import { Block } from './index'
 
 import type { BlockOptions, JsonRpcBlock } from './index'
 import type { TxData, TypedTransaction } from '@ethereumjs/tx'
 
-function normalizeTxParams(_txParams: any) {
+export function normalizeTxParams(_txParams: any) {
```
holgerd77, line 11: Is it really necessary to export this function (always problematic from an API point of view, since the "we cannot change surface" will grow)? I am not seeing any external usage.

packages/block/src/from-rpc.ts
```diff
@@ -1,35 +1,30 @@
...
     txParams.to !== null && txParams.to !== undefined
       ? setLengthLeft(toBuffer(txParams.to), 20)
       : null
 
-  // v as raw signature value {0,1}
-  // v is the recovery bit and can be either {0,1} or {27,28}.
-  // https://ethereum.stackexchange.com/questions/40679/why-the-value-of-v-is-always-either-27-11011-or-28-11100
-  const v: number = txParams.v
-  txParams.v = v < 27 ? v + 27 : v
+  txParams.v = toType(txParams.v, TypeOutput.BigInt)
```
holgerd77, line 27: Hmm, after your comment somewhere (I guess in #2059) which I do not find any more unfortunately I wonder if we *do* still need this `v + 27` part - since it seems that there is this use case of legacy txs coming in with 0/1 - but limiting this to the legacy case? 🤔

So something like:

```typescript
if (!(txParams.type !== undefined && BigInt(txParams.type) > 0n) && v < 27) {
  txParams.v = v + 27
}
```

Something like that?

packages/block/src/from-rpc.ts
acolytec3: I used it in the Ethers State Manager and I think it just got carried over when I copied this code from that PR.  I think we could remove the export for now and add it back later if/when we proceed with an Ethers State Manager PR that would get merged.

packages/block/src/from-rpc.ts
acolytec3: I guess I'm still not sure.  I haven't seen the v show up as a 27/28 in any of the transaction responses I've seen from Infura/Alchemy (who are presumably running geth nodes).  Let me run geth and nethermind locally and look at some old blocks with transactions and see what v looks like in their rpc responses.

acolytec3: So I spoke wrong here.  Looks like Infura/Cloudflare/Geth are all returning v either 27/28 for legacy transactions (at least in very old blocks, e.g. 2000004). In a more modern block (e.g. 14000000), in the same block I see two different legacy transactions, one with a v = 0 and one with it 38.  If I run them through the current version of `blockFromRpc` it creates the block just fine.  So, I think our current code in the legacy transaction constructor is handling all of the various edge cases of V (including 0/1, 27/28, and 35+ for EIP-155 txns with a Chain ID included).

acolytec3: All of that to say, I don't think we need to do any adjusting v since our Transaction constructor already checks all the edge cases and throws if it's invalid.

holgerd77: Ok, great. If this is for some reason some issue for some special cases (I would rather assume: it is more likely than not that it is not), I guess someone will just stumble upon it and eventually open an issue if there is a need.

So I guess we are fine here to just keep as we have and move on.

packages/block/src/from-rpc.ts
holgerd77: Hmm, I guess from an API point-of-view this is not an optimal method to export regarding method location (block library), purpose, typing, docs....

So I would really like to remove again, in doubt just copy over the code to `EthersStateManager`. Or alternatively, if you have a solid and generic way to implement and document you can also move over to `Transaction` (I think this would make more sense in case we do?), but then we also need some tests and the whole way down. 🙂

packages/block/src/from-rpc.ts
acolytec3: I would assume so. I don't know all the edge cases that could exist but our existing transaction library seems to account for them, at least what we find when consuming transactions from major rpc providers and Geth directly. For other providers, I would first wonder if they aren't following the spec before I assume we've done something wrong.

PR #2303, opened 2022-09-22
evm: update imports/exports

### LINKED ISSUE #2295: Build errors with newest ethereumjs/evm
Hey @jochem-brouwer @holgerd77,

I encountered 2 build errors using the newest package of `@ethereumjs/evm` (1.0.0) that prevent me from deploying my `evm.codes` app to Vercel. While I am able to easily solve these errors locally (by changing the sources of the `evm` package, this is obviously not ideal and I would like to know what is the correct way to handle these errors so that I can deploy to Vercel using the original package.

The first error:

```
./node_modules/@ethereumjs/evm/src/evm.ts:12:1
Type error: Import assignment cannot be used when targeting ECMAScript modules. Consider using 'import * as ns from "mod"', 'import {a} from "mod"', 'import d from "mod"', or another module format instead.

  10 |   zeros,
  11 | } from '@ethereumjs/util'
> 12 | import AsyncEventEmitter = require('async-eventemitter')
     | ^
  13 | import { debug as createDebugLogger } from 'debug'
  14 | import { promisify } from 'util'
  15 | 
error Command failed with exit code 1.
```
I am able to solve this easily by changing the erroneous line to `import AsyncEventEmitter from 'async-eventemitter'`

The second error:

```
./node_modules/@ethereumjs/evm/src/precompiles/index.ts:204:3
Type error: Re-exporting a type when the '--isolatedModules' flag is provided requires using 'export type'.

  202 | 
  203 | export {
> 204 |   AddPrecompile,
      |   ^
  205 |   CustomPrecompile,
  206 |   DeletePrecompile,
  207 |   getActivePrecompiles,
error Command failed with exit code 1.
```

I am able to solve this one by changing the erroneous block of code to:

```typescript
export {
  getActivePrecompiles, precompiles,
  ripemdPrecompileAddress
}
export type {
  AddPrecompile,
  CustomPrecompile,
  DeletePrecompile, PrecompileFunc,
  PrecompileInput
}
```

Also, I cannot disable `--isolatedModules` since I am using SWC, which requires it to be enabled, in my project.

Below you can see my `tsconfig`:

```json
{
  "compilerOptions": {
    "baseUrl": ".",
    "target": "ESNext",
    "lib": [
      "dom",
      "dom.iterable",
      "ESNext"
    ],
    "allowJs": true,
    "skipLibCheck": true,
    "strict": true,
    "forceConsistentCasingInFileNames": true,
    "noEmit": true,
    "esModuleInterop": true,
    "module": "ESNext",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "isolatedModules": true,
    "jsx": "preserve",
    "incremental": true
  },
  "include": [
    "next-env.d.ts",
    "**/*.ts",
    "**/*.tsx"
  ],
  "exclude": [
    "node_modules"
  ]
}
```

Any thoughts on these errors? would appreciate your help! :)

acolytec3:
Fixes #2295


holgerd77:
Just read up on the associated issue, what is the status of this here? I would assume from your latest comment ("will pick up on Monday" or similar) not yet ready for review?

acolytec3:
Correct. I need to figure out how to cleanly update the types for `asyncEventEmitter`. The hacks described in this thread all work in some sense but I haven't been able to make any of them without either casting to `any` or else having other weird side affects on the `EVMInterface` typing so far where it also comes for evm.codes

acolytec3:
I ran the VM benchmarks with the new `eventemitter2` dependency and below are the results vs current master (basically identical).
```
eventemitter2
Number of blocks to sample: 10
Block 9422905 x 27,769 ops/sec ±1.84% (85 runs sampled)
Block 9422906 x 26,526 ops/sec ±3.12% (87 runs sampled)
Block 9422907 x 26,901 ops/sec ±0.80% (91 runs sampled)
Block 9422908 x 25,902 ops/sec ±0.80% (88 runs sampled)
Block 9422910 x 24,201 ops/sec ±4.45% (86 runs sampled)

async-eventemitter
Number of blocks to sample: 10
Block 9422905 x 28,308 ops/sec ±1.21% (88 runs sampled)
Block 9422906 x 26,785 ops/sec ±2.05% (89 runs sampled)
Block 9422907 x 26,336 ops/sec ±1.01% (92 runs sampled)
Block 9422908 x 24,039 ops/sec ±3.81% (83 runs sampled)
Block 9422910 x 24,225 ops/sec ±0.82% (86 runs sampled)
```

So, on performance at least, looks good.  Will rebase the PR on current master and we can verify everything is working ok.

holgerd77 (approved):
Actually I just can't do much more on this review than sitting here before the monitor for 5 minutes and reflecting on the changes and impacts. 😀

I do think that changing on the event emitter class "on the go" has some risks and might trigger some not-thought-of consequenses for some use cases and we somewhat have to count in that some people might complain (or at least run into some issues).

On the other hand I think this is so much the right respectively an overdue and at the end good move to switch from this super outdated package to a modern, used and maintained library, that the benefits - at least mid-term - strongly outweight some short-term hickups.

So, long story short: will approve, also directly integrate into releases (will prepare right after merging in here).

packages/evm/src/evm.ts
```diff
@@ -150,7 +148,7 @@ export class EVM implements EVMInterface {
 
   public readonly _transientStorage: TransientStorage
 
-  public readonly events: AsyncEventEmitter<EVMEvents>
+  public readonly events: AsyncEventEmitter
```
holgerd77, line 151: Ugh.

@jochem-brouwer Seems we are loosing this typing here. 😬

I've tried to re-add, but seems evenemitter2 is not generic so this is not an option.

Hmm, sorry for that, I know this was some great improvement. Since we were so stuck on this import problem finally solved here and having this super-old dependency removed and updated would generally be really good I would have some stronger tendency for now to nevertheless go for it. Are you ok with this? 😬

I am currently not seeing an alternative.

Or are there eventual solutions here to get the typing back? Can't think of any right now?

holgerd77: Seems this is being worked on (respectively: already done) here? 🤩

https://github.com/EventEmitter2/EventEmitter2/pull/294

So maybe we can just bring this back in a follow-up bugfix release once this is ready to be published?

davidmurdoch:
This is a breaking change, if EthereumJS follows semver. Ganache uses the `next` method of EtheruemJS vm's usage of `AsyncEventEmitter`'s `emitter.on` method. Example:

```javascript
vm.evm.events.on("step", async (event, next) => {
  // ... some async things ...
  next();
});
```

It doesn't appear that the new library you use has this option. Can you revert this change in the v6 release? Then maybe re-release with this PR as a v7?

holgerd77:
@davidmurdoch we'll discuss internally on how to proceed here and reach back here once we have settled.

PR #2308, opened 2022-09-24
blockchain: Fix and enhance the blockchain iterator

g11tech:
Blockchain's reorg handling was incorrect as well as client execution's usage of iterator. Fixed those.

Also enhanced the iterator with the option to run callback outside the `runWithlock` as executing the block doesn't require one to lock the blockchain, which may very well be used to get/make blockchain updates while the block is executing. This may make concurrency more efficient.


holgerd77:
Rebased this via UI.

packages/blockchain/test/iterator.spec.ts
```diff
@@ -0,0 +1,211 @@
+import * as tape from 'tape'
+
+import { Blockchain } from '../src'
+
+import { createTestDB, generateBlockchain, generateConsecutiveBlock } from './util'
+
+import type { Block } from '@ethereumjs/block'
+
+tape('blockchain test', (t) => {
+  t.test('should iterate through 24 blocks without reorg', async (st) => {
```
holgerd77, line 10: Ah, great to have these tests extracted! 👍

holgerd77 (approved):
Nice, this LGTM!

Feel free to merge.

packages/blockchain/test/iterator.spec.ts
g11tech: :slightly_smiling_face:  also enhanced the tests with reorg creation and checking

holgerd77: Ah sorry, didn't have such a close look and wasn't totally obvious through this extraction and the diff therefore not displaying, but, yes: that's great too for sure! 🙂

Issue #1882, opened 2022-05-13
DevP2P debug strings optimization


jochem-brouwer:
In devp2p at least in eth and les we always generate a debug string even if we are not in debug mode, e.g. here: https://github.com/ethereumjs/ethereumjs-monorepo/blob/8e1c3cf9c47bfc037182c275f46eef405ef70dca/packages/devp2p/src/protocol/eth.ts#L43

Guards should be added so we don't unnecessarily generate debug strings in case we are not in debug mode.

holgerd77:
See e.g. VM [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/runTx.ts#L63) as reference for such a guard implementation/addition.

ndrozd:
My first attempt, would be happy to hear what i did wrong :)

PR #2311, opened 2022-09-25
Added env check for debug mode + optimised the variable usage #1882

ndrozd:
Solves DevP2P debug strings optimization #1882


packages/devp2p/src/protocol/eth.ts
```diff
@@ -47,29 +53,33 @@ export class ETH extends Protocol {
   static eth65 = { name: 'eth', version: 65, length: 17, constructor: ETH }
   static eth66 = { name: 'eth', version: 66, length: 17, constructor: ETH }
 
+  
+
   _handleMessage(code: ETH.MESSAGE_CODES, data: any) {
     const payload = arrToBufArr(RLP.decode(bufArrToArr(data)))
     const messageName = this.getMsgPrefix(code)
-    const debugMsg = `Received ${messageName} message from ${this._peer._socket.remoteAddress}:${this._peer._socket.remotePort}`
+    const debugMsg = this.DEBUG ?  `Received ${messageName} message from ${this._peer._socket.remoteAddress}:${this._peer._socket.remotePort}` : undefined;
```
acolytec3, line 61: Let's move this inside the `if (code... && this.DEBUG)` code section and then you can remove the `this.DEBUG ? ... : undefined` and only create this debug message when the DEBUG flag is iset.

packages/devp2p/src/protocol/eth.ts
```diff
@@ -47,29 +53,33 @@ export class ETH extends Protocol {
...
-          'Uncontrolled status message',
-          this.debug.bind(this),
-          'STATUS'
-        )
-        this._peerStatus = payload as ETH.StatusMsg
-        const peerStatusMsg = `${
-          this._peerStatus !== undefined ? this._getStatusString(this._peerStatus) : ''
-        }`
-        this.debug(messageName, `${debugMsg}: ${peerStatusMsg}`)
+        if (this.DEBUG) {
```
acolytec3, line 69: This should move below line 77 below as we need the `this._peerStatus...` stuff regardless of whether we're doing debug logging or not.

acolytec3 (changes requested):
Thanks for this PR!  I've commented on a couple of fixes in line.  There are also still an instance of `this.debug` that isn't set behind the `if (this.DEBUG)` guard in `_validateForkId` (around line 122) so it'd be nice to add that too for consistency.

packages/devp2p/src/protocol/eth.ts
ndrozd: That was what I did initially, but then project won't build then because debugMsg is not defined for all the other conditions, but maybe I'm just no doing this right. 
Logically it's fine since all the conditions will only run when var is defined, but I guess linter is just going crazy about that.

acolytec3: Ah, didn't catch that. In that case, this approach is good.

packages/devp2p/src/protocol/eth.ts
```diff
@@ -127,7 +137,9 @@ export class ETH extends Protocol {
     const peerFork: any = c.hardforkForForkHash(peerForkHash)
     if (peerFork === null) {
       const msg = 'Unknown fork hash'
-      this.debug('STATUS', msg)
+      if (this.DEBUG) {
```
acolytec3, line 140: Can you use this same check above at line 132?

ndrozd:
Thank you, I'll fix that ASAP :)

acolytec3:
Just checking back in.  Do you have time to finish this up?  Otherwise, I can pick it up and finish things off so we can get this merged.  Thanks!

ndrozd:
Should get this done by Monday

ndrozd:
Done @acolytec3

acolytec3 (approved):
LGTM!  Thanks for the PR.

acolytec3:
@ndrozd I fixed a couple of our devp2p examples and committed to this PR since it touches the same codebase.  I've since run those example scripts and validated that the `devp2p:eth` messages are only appearing when the DEBUG flag is set to true.  Thanks for the PR!

PR #2313, opened 2022-09-26
common: Refactor and update common.getHardforkByBlockNumber to address post merge hfs

g11tech:
The sepolia network even post this 
```typescript
{
      "//_comment": "Post merge forks should always have ttd set with block to distinguish that they come post merge hf (block = null)",
      "name": "mergeForkIdTransition",
      "block": 1735371,
      "forkHash": "0xb96cbd13"
    },
``` 
was only giving `merge` hardfork on `common.getHardforkByBlockNumber(1735371,"17000000000000000")` 
and would give `mergeForkIdTransition` on `common.getHardforkByBlockNumber(1735371)`

This PR refactors the `getHardforkByBlockNumber` as the case handling was not clear and upfront

TODO: 
 - [x] Add test cases covering the post merge hardforks as well as validations


packages/common/src/common.ts
```diff
@@ -285,57 +285,89 @@ export class Common extends EventEmitter {
...
       }
-      if (td && (typeof hf.ttd === 'string' || typeof hf.ttd === 'bigint')) {
-        if (td >= BigInt(hf.ttd)) {
-          minTdHF = hf.name
-        } else {
-          maxTdHF = previousHF
+
+      // If merge hardfork has block null set, we will make determination if this is merge hardfork or preMergeHF
+      //  - if no future hardforks qualify by block number
+      // So we assume it to quality for now to move to check future post merge hardfork qualitfication
```
acolytec3, line 321:
```suggestion
      // So we assume it to qualify for now to move to check future post merge hardfork qualitfication
```

packages/common/src/common.ts
```diff
@@ -285,57 +285,89 @@ export class Common extends EventEmitter {
...
-        }
+      // Check if hf can be applied, if not then break and the last hardfork set is the response
+      if (hf.block === null && (hf.ttd === undefined || hf.ttd === null)) {
+        // Not a hardfork which has been scheduled yet so just skip
         continue
       }
-      if (blockNumber >= BigInt(hf.block)) {
-        hardfork = hf.name as Hardfork
+
+      if (hf.ttd !== undefined && hf.ttd !== null && preMergeHF === undefined) {
```
acolytec3, line 314: Shouldn't we throw here if `preMergeHF !== undefined`?

acolytec3 (changes requested):
Will do more review later but here are some initial questions/feedback.

packages/common/src/common.ts
g11tech: ummm...  can chain start from pre merge hardfork specification (i.e. genesi post merge), followed by hard forks still activated at genesis (`0`), just that the user jumbled them up, In long term there can be a dag dependency that we can validate

acolytec3: If you find a preMergeHf at this point, doesn't that mean you found a second HF with a ttd specified (which is invalid)?  And if the user jumbled up the hardfork sequence, that's still invalid.  We shouldn't allow hardforks to be out of sequence.

g11tech: no I need to set `preMergeHf` here if this `hf` is the merge one so that if post merge hardforks don't quality based on block numbers, the choice has to be made from `preMergeHf` and the merge hf (which will be set in `hardfork` if the next hardfork don't quality and we break out of the loop)

the check of double ttd hardfork is made  down (towards loop end) as `if (passedMergeHF !== undefined && hf.ttd !== undefined && hf.ttd !== null) {`

packages/common/tests/hardforks.spec.ts
```diff
@@ -63,6 +63,82 @@ tape('[Common]: Hardfork logic', function (t: tape.Test) {
...
+      )
+      st.fail('should have thrown as specified td < merge ttd for a post merge hardfork')
+    } catch (error) {
+      st.pass('throws error as specified td < merge ttd for a post merge hardfork')
+    }
+
+    st.end()
+  })
+
+  t.test('should reject two hardforks  with ttd specified', function (st: tape.Test) {
```
g11tech, line 117: @acolytec3  see this

packages/common/src/common.ts
g11tech: > We shouldn't allow hardforks to be out of sequence.
Ideally we should have a dag, which should impose partial order on the hardfork specification, but for now let user handle it. The order isn't really fixed for e.g. `mergeForkIdTransition` could come before (like it came in `kiln` network) or could come after `merge` (like sepolia and then other networks will followup)

so its better to apply an open approcah here, and have some sort of dag validation when we are setting the `hardforks` in the common while instantiation.

acolytec3: I guess maybe I'm not tracking something then.  This is the first validation step and the first thing you do is check to see if the hardfork being checked as a TTD.  If it does, you check to see if you've set a "preMergeHf" and if not, you set it.  If the preMergeHf already exists, then you have another hf with TTD, which we agree is an invalid case.  My point is that I think this validation step takes care of the "no two hardforks with TTD allowed" and you don't need a separate check.

g11tech: ohh thanks for explaing , yes it should, will update and see if test file runs

g11tech: yes it works! updated

packages/common/src/common.ts
```diff
@@ -285,57 +285,89 @@ export class Common extends EventEmitter {
...
+    /**
+     * Merge hardfork config if we have passed the merge hardfork while finding hardfork
+     * because next hardfork's block <= blockNumber
+     */
+    let passedMergeHF: HardforkConfig | undefined = undefined
+    /** Candiate hardfork config, updated as we iterate amongst the hardforks */
+    let hardfork: HardforkConfig | undefined = undefined
+
+    // Hardforks specified are strickly in the order of application, so we can make some intelligent decisions
+    // based on that especially regarding the merge hardfork where block might not be specified
```
acolytec3, line 306:
```suggestion
    // Hardforks specified are strictly in the order of application, so we can make some intelligent decisions
    // based on that, specifically regarding the merge hardfork where only TTD and not block might be specified
```

packages/common/src/common.ts
```diff
@@ -285,57 +285,89 @@ export class Common extends EventEmitter {
...
-        hardfork = hf.name as Hardfork
-      }
-      if (td && (typeof hf.ttd === 'string' || typeof hf.ttd === 'bigint')) {
-        if (td >= BigInt(hf.ttd)) {
-          minTdHF = hf.name
-        } else {
-          maxTdHF = previousHF
+
+      if (hf.ttd !== undefined && hf.ttd !== null) {
+        // Ok, this is the hardfork with ttd
```
acolytec3, line 315:
```suggestion
        // This is the merge hardfork since only merge can have TTD
```

packages/common/src/common.ts
```diff
@@ -285,57 +285,89 @@ export class Common extends EventEmitter {
...
-        if (!this.hardforkGteHardfork(hardfork, minTdHF)) {
-          const msg = 'HF determined by block number is lower than the minimum total difficulty HF'
-          msgAdd += `total difficulty: ${td} (-> ${minTdHF})`
-          throw new Error(`${msg}: ${msgAdd}`)
+
+      // If merge hardfork has block null set, we will make determination if this is merge hardfork or preMergeHF
+      //  - if no future hardforks qualify by block number
+      // So we assume it to qualify for now to move to check future post merge hardfork qualitfication
+      //
+      // For the hardforks with block specified we check for qualification
```
acolytec3, line 329:
```suggestion
      // If merge fork has fork block set, we've found the correct hardfork
      // for `blockNumber` so break here.  Otherwise, check any additional forks.
```

packages/common/src/common.ts
```diff
@@ -285,57 +285,89 @@ export class Common extends EventEmitter {
...
       }
-      if (maxTdHF !== undefined) {
-        if (!this.hardforkGteHardfork(maxTdHF, hardfork)) {
-          const msg = 'Maximum HF determined by total difficulty is lower than the block number HF'
-          msgAdd += `total difficulty: ${td} (-> ${maxTdHF})`
-          throw new Error(`${msg}: ${msgAdd}`)
+
+      // hf qualifies else we would have broken out of the loop till now
+      if (hardfork?.ttd !== undefined && hardfork?.ttd !== null) {
+        passedMergeHF = hardfork!
```
acolytec3, line 338: Why do we need this in addition to `preMergeHf`?  I think I must not be tracking all of the logic yet because there can be only one hardfork with TTD set.  So, if we have that, can't we just use that hardfork (set above as `preMergeHf`).  Also, it seems really odd to call the merge hardfork `preMergeHf`.  The "pre-merge hardfork" is London so maybe we can find a better way to name this?

packages/common/tests/hardforks.spec.ts
```diff
@@ -63,6 +63,86 @@ tape('[Common]: Hardfork logic', function (t: tape.Test) {
...
   })
 
+  t.test('test post merge hardforks using Sepolia', function (st: tape.Test) {
+    const c = new Common({ chain: Chain.Sepolia })
+    let msg = 'should get HF correctly'
+
+    st.equal(c.getHardforkByBlockNumber(0), Hardfork.London, msg)
+    //using Hardfork.London even though Merge is expected after terminal block 1450408
+    st.equal(c.getHardforkByBlockNumber(1450409), Hardfork.London, msg)
+    st.equal(c.getHardforkByBlockNumber(1450409, BigInt('17000000000000000')), Hardfork.Merge, msg)
```
acolytec3, line 73: I think this result is counterintuitive.  The function is called `getHardforkByBlockNumber` and yet when we set TTD and a block number prior to the actual merge block, it's telling me the hardfork is the Merge hardfork.  If we're checking the hardfork and specifying a block number prior to the actual block where the merge happens, we shouldn't report a hardfork that is happening later.  I know there's screwiness around Merge and MergeForkTransitionId but this doesn't make a ton of sense to me.

acolytec3 (changes requested):

packages/common/src/common.ts
g11tech: yes, thats what `preMergeHf` should represent, its `london`, and `passedMergedHF` is assigned `merge` hardfork if its passed, i guess i can change it to boolean and store the merge hardfork as `mergeHF` as this later needs to be accessed.

holgerd77:
TBH: I have no idea what we are actually solving here.

> was only giving merge hardfork on common.getHardforkByBlockNumber(1735371,"17000000000000000")
> and would give mergeForkIdTransition on common.getHardforkByBlockNumber(1735371)

This seems like the correct behavior to me. When a TTD is given the TTD takes precedence (and therefore returns the Merge HF with the TTD set, otherwise respectively in the only-block-number case the mergeForkIdTransition HF is returned.

What is the problematic behavior respectively what case can not be modelled with the current method behavior? 🤔

g11tech:
> This seems like the correct behavior to me. When a TTD is given the TTD takes precedence (and therefore returns the Merge HF with the TTD set, otherwise respectively in the only-block-number case the mergeForkIdTransition HF is returned.
> 
> What is the problematic behavior respectively what case can not be modelled with the current method behavior? thinking

It  (`common.getHardforkByBlockNumber(1735371,"17000000000000000")`) should give `mergeForkIdTransition` on/after `1735371` as `mergeForkIdTransition` is scheduled post merge. i.e. if the `merge` block is not specified in hardfork config, it will always return `merge` even for hardforks scheduled post merge.

The current change is to get the correct `mergeForkIdTransition` even if one has not specified the `block` of the merge hardfork.

holgerd77:
> It (common.getHardforkByBlockNumber(1735371,"17000000000000000")) should give mergeForkIdTransition on/after 1735371 as mergeForkIdTransition is scheduled post merge. i.e. if the merge block is not specified in hardfork config, it will always return merge even for hardforks scheduled post merge.
> The current change is to get the correct mergeForkIdTransition even if one has not specified the block of the merge hardfork.

Yes, you should only call the `getHardforkByBlockNumber` method with TTD if you want to get the Merge HF itself, in all other cases, you should call it without TTD (also for `mergeForkIdTransition`). That's how I intended it when writing this method. Is this behavior problematic? 🤔

You should just be able to call with TTD on all code parts where you want the Merge HF and otherwise do without.

g11tech:
OR you mean to say, we selectively call with/without `td` based on if we want to access some merge related functionality?

acolytec3:
> Yes, you should only call the `getHardforkByBlockNumber` method with TTD if you want to get the Merge HF itself, in all other cases, you should call it without TTD (also for `mergeForkIdTransition`). That's how I intended it when writing this method. Is this behavior problematic? thinking

If we pass with TTD only, it just becomes a boolean of `provided ttd >= mergefork.ttd`, right?  That doesn't seem like it fits the intended use case for this method and we should create a special purpose method for such.  

To my way of thinking, if a user wants to know if a given block is passed the merge hardfork using this method, we should require them to pass both block number and ttd, run through the list of hardforks and find the one that we're on by block number, then check ttd only if the hf by blocknumber says we're on london.  If we find a later hardfork with a block number (e.g. in some future world where Shanghai has passed already), then we shouldn't need to look at ttd at all.

Does that logic make sense at all?

acolytec3:
Could we replace the whole thing with something like this?
```js
 // Filter out hardforks with no block number and no ttd (i.e. unapplied hardforks)
  const hfs = common
    .hardforks()
    .filter((hf) => hf.block !== null || (hf.ttd !== null && hf.ttd !== undefined))
  // The index of the hardfork in the hf array that matches our hardfork based on block number
  const hfIndex = hfs.findIndex((entry) => entry.block !== null && entry.block > blockNumber) - 1
  if (hfs[hfIndex].block === null) {
    // We're on the merge hardfork.  Let's check the TTD
    if (BigInt(hfs[hfIndex].ttd!) > td) {
      // Merge ttd greater than current td so we're on hardfork before merge
      hf = hfs[hfIndex - 1]
    } else {
      // Merge ttd equal or less than current td so we're on merge hardfork
      hf = hfs[hfIndex]
    }
  }
  return hf
```

packages/common/tests/hardforks.spec.ts
acolytec3: I still think this result is counterintuitive.  If you provide a block number and the merge hasn't happened yet, regardless of what you pass in for TD, this method should return the previous hardfork.  Block number should trump hardfork in terms of which fork is returned.

acolytec3 (changes requested):

packages/common/tests/hardforks.spec.ts
g11tech: merge HAS happened if we provide td > ttd (L73), and hasn't happened if we don't provide td L72

packages/common/tests/mergePOS.spec.ts
```diff
@@ -187,4 +189,122 @@ tape('[Common]: Merge/POS specific logic', function (t: tape.Test) {
...
+    const c = new Common({ chain: Chain.Sepolia })
+    let msg = 'should get HF correctly'
+
+    st.equal(c.getHardforkByBlockNumber(0), Hardfork.London, msg)
+    // Make it null manually as config could be updated later
+    const mergeHf = c.hardforks().filter((hf) => hf.ttd !== undefined && hf.ttd !== null)[0]
+    const prevMergeBlockVal = mergeHf.block
+    mergeHf.block = null
+
+    // should get Hardfork.London even though happened with 1450408 as terminal as config doesn't have that info
```
g11tech, line 211: @acolytec3  see if these comments help making sense of the situation

g11tech: yes but for this test case i made block null, see L209, i have made this test to be independent of what we specify config in json

acolytec3 (changes requested):

acolytec3 (approved):
LGTM!

holgerd77:
Ok, cool, thanks for finalizing!

PR #2315, opened 2022-09-26
Add Ethers State manager v2

acolytec3:
This started with #2059 and has replaced the local `trie` with a `storageCache` which is just a simple map of maps to handle storage slots and then uses the existing `cache` from the `BaseStateManager` for account storage.  I've probably got some rough edges yet to handle but this should be in a good starting place for review.


acolytec3:
This should be ready for review now.

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,382 @@
...
+    }
+    const deleteCb = async (_keyBuf: Buffer) => {
+      return Promise.resolve()
+    }
+    this._cache = new Cache({ getCb, putCb, deleteCb })
+  }
+
+  copy(): EthersStateManager {
+    const newState = new EthersStateManager({ provider: this.provider })
+    ;(newState as any).contractCache = new Map(this.contractCache)
```
holgerd77, line 79: Just mentioning some stuff I notice along comparison to old EthersStateManager, not necessarily something to act upon (judge yourself): is the state itself (before: trie) actually copied here?

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,382 @@
...
+  }
+
+  /**
+   * Clears the internal cache so all accounts, contract code, and storage slots will
+   * initially be retrieved from the provider
+   */
+  clearCache(): void {
+    this.contractCache.clear()
+    this.storageCache.clear()
+    this._cache.clear()
```
holgerd77, line 105: Difference here: `this._cache.clear()` hasn't been in the "old" code.

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,382 @@
...
+   * @param address - Address to get the `code` for
+   * @returns {Promise<Buffer>} - Resolves with the code corresponding to the provided address.
+   * Returns an empty `Buffer` if the account has no associated code.
+   */
+  async getContractCode(address: Address): Promise<Buffer> {
+    let codeBuffer = this.contractCache.get(address.toString())
+    if (codeBuffer !== undefined) return codeBuffer
+    const code = await this.provider.getCode(address.toString(), this.blockTag)
+    codeBuffer = toBuffer(code)
+    if (codeBuffer === undefined || codeBuffer.length === 0) return Buffer.from([])
```
holgerd77, line 119: This line seems to be new.

holgerd77 (commented):
Ah sorry, wanted to finish another post, so need to do an early stop: just a few first mentions of some stuff.

packages/statemanager/src/ethersStateManager.ts
acolytec3: Yep, and now that I look at it again, we don't actually need it.

acolytec3: Since `getCode` will always return a hex string (just `0x` if there's no code stored there) and `toBuffer` converts 0x to the empty buffer already.

packages/statemanager/src/ethersStateManager.ts
acolytec3: Yep, we don't have a `cache` in the trie-backed version of the PR so I added this hear to ensure the cache was cleared

acolytec3:
Barring any issues in the tests.  Are we ready to move forward with final review on this?  Would be cool to include in the next bug fix release if so.

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -251,23 +205,15 @@ export class EthersStateManager extends BaseStateManager implements StateManager
...
-        })
-    })
+    const addressStorage = this.storageCache.get(address.toString())
+    if (addressStorage === undefined) {
+      return Promise.resolve({} as StorageDump)
+    }
+    const dump: StorageDump = {}
+    for (const slot of addressStorage) {
+      dump[slot[0]] = bufferToHex(slot[1])
+    }
```
gabrocheleau, line 215:
```suggestion
    const dump: StorageDump = {}
    if (addressStorage !== undefined) {
      for (const slot of addressStorage) {
        dump[slot[0]] = bufferToHex(slot[1])
      }
    }
```
slightly simpler imo

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -452,67 +339,53 @@ export class EthersStateManager extends BaseStateManager implements StateManager
...
-  /**
-   * Dummy method needed by base state manager interface
-   */
   async flush(): Promise<void> {
-    return Promise.resolve()
+    await this._cache.flush()
+  }
+
+  getStateRoot = async () => {
+    return this.stateRoot !== undefined ? this.stateRoot : setLengthLeft(Buffer.from([]), 32)
```
gabrocheleau, line 380:
```suggestion
    return this.stateRoot ?? setLengthLeft(Buffer.from([]), 32)
```

gabrocheleau (approved):
Looks good! A few nits, but makes sense to me to merge the PR in its current form

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,391 @@
...
+   * @param address - Address to get the `code` for
+   * @returns {Promise<Buffer>} - Resolves with the code corresponding to the provided address.
+   * Returns an empty `Buffer` if the account has no associated code.
+   */
+  async getContractCode(address: Address): Promise<Buffer> {
+    let codeBuffer = this.contractCache.get(address.toString())
+    if (codeBuffer !== undefined) return codeBuffer
+    const code = await this.provider.getCode(address.toString(), this.blockTag)
+    codeBuffer = toBuffer(code)
+    this.contractCache.set(address.toString(), codeBuffer)
```
gabrocheleau, line 125:
```suggestion
    this.putContractCode(address, codeBuffer)
```
nbd, but this aligns better with what we're doing in getContractStorage

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,391 @@
...
+    const accountStorage: Map<string, Buffer> | undefined = this.storageCache.get(
+      address.toString()
+    )
+    let storage: Buffer | string | undefined
+    if (accountStorage !== undefined) {
+      storage = accountStorage.get(key.toString('hex'))
+      if (storage !== undefined) {
+        return storage
+      }
+    }
```
gabrocheleau, line 160: I think we'd encounter issues in some scenarios, for example the following (maybe that's an intentional limitation of the current ethersStateManager):

- `this.blockTag` = `"latest"`
- we call `getContractStorage()` with an arbitrary address/key pair (it's therefore retrieved at the state of the latest block)
- some time passes, and that address/key pair storage is updated 
- we call `getContractStorage()` again with the same address/key pair, the (outdated) cache is retrieved and returned 

What if we bypassed the cache is one of the "dynamic" values (pending, earliest, latest)?

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -60,15 +54,36 @@ export class EthersStateManager extends BaseStateManager implements StateManager
...
     this.contractCache = new Map()
-    this.externallyRetrievedStorageKeys = new Map<string, Map<string, boolean>>()
+    this.storageCache = new Map()
+
+    /*
+     * For a custom StateManager implementation adopt these
+     * callbacks passed to the `Cache` instantiated to perform
+     * the `get`, `put` and `delete` operations with the
+     * desired backend.
+     */
```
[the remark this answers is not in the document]
acolytec3: Accidentally copied over from the base state manager class.  I'm just going to remove entirely.

packages/statemanager/src/ethersStateManager.ts
acolytec3: I've updated the readme section related to this state manager to hopefully clarify these points.

packages/statemanager/src/ethersStateManager.ts
g11tech: it seems like changes in cache below will actually persist between copes, as well as for storage cache's elements

packages/statemanager/src/ethersStateManager.ts
g11tech: I think the point is valid, between consecutive calls of fetches in race conditions for the same run, dynamic tag will bring inconsistent data, the easy way to handle is when some one `inits`/`setTag` the "run" starts and the etherstatemanager should get the block corresponding to that and run it (I am not sure how pending can be supported unless the node is not syncing more blocks, and also not sure about getProofs for pending)

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,384 @@
...
+  async flush(): Promise<void> {
+    await this._cache.flush()
+  }
+
+  getStateRoot = async () => {
+    return this.stateRoot ?? setLengthLeft(Buffer.from([]), 32)
+  }
+
+  setStateRoot = async (root: Buffer) => {
+    if (root.length !== 32) throw new Error('Root must contain 32 bytes')
```
g11tech, line 377: there should ideally be the check here that this matches with the blockTag since this seems to be on the public api, or could just throw that the stateroot can't be set outside `init/setTag` function

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,384 @@
...
+    let storage: Buffer | string | undefined
+    if (accountStorage !== undefined) {
+      storage = accountStorage.get(key.toString('hex'))
+      if (storage !== undefined) {
+        return storage
+      }
+    }
+
+    // Retrieve storage slot from provider if not found in cache
+    storage = await this.provider.getStorageAt(
```
g11tech, line 157: ideally we should also use getProof with storage slots in consistency with how to get account, and verify the proof

acolytec3: The only reason I'm using `getProof` in `getAccountFromProvider` is to provide the correct data for the `cache` which needs balance/nonce/codeHash/storageRoot.  I'm not sure I see the value of doing getProof for storage slots.  Just feels like a lot of extra work unless we don't trust the provider and that seems beyond the scope of this code.

packages/statemanager/src/ethersStateManager.ts
acolytec3: Resolved by removing the option to pass in dynamic block tags.

g11tech:
ran a block off mainnet: the total gas matches perfectly! so i say good to go! :rocket: 

https://etherscan.io/block/15669066
image
image

g11tech (approved):
lgtm!

packages/statemanager/src/ethersStateManager.ts
scorbajio: Proof validation can happen at another layer and not inside the state-manager itself. Could also mention that the state-manager expects trusted data.

packages/evm/package.json
```diff
@@ -48,6 +48,7 @@
   "dependencies": {
     "@ethereumjs/common": "^3.0.0",
     "@ethereumjs/util": "^8.0.0",
+    "@ethersproject/providers": "^5.7.1",
```
holgerd77, line 51: Is this correct to have this dependency now in the EVM main dependencies? 🤔

packages/statemanager/package.json
```diff
@@ -41,12 +41,32 @@
     "tsc": "../../config/cli/ts-compile.sh"
   },
   "dependencies": {
+    "@ethereumjs/block": "^4.0.0",
```
holgerd77, line 44: I am not seeing on first sight (and would generally want to avoid it) why the Block library is added here as a main dependency?

holgerd77 (changes requested):
Two last dependency questions, otherwise looks really great and would say as well ready for merge. 🙂

packages/statemanager/package.json
acolytec3: I had it here because I have the `getBlockFromRpc` method that takes care of returning pulling a block from the provider, getting the uncle headers (if needed), and then putting it in our `block` format so it's ready to go in a VM run.  We could certainly remove the dependency on `block` and just have it return an RPC format block that then put the onus on the user to call `Block.blockFromRpc` if they want to use `vm.runBlock`.  Is that your preference here?

packages/evm/package.json
acolytec3: That's the package within `ethers` that contains the provider interface used by the new State Manager to retrieve data.  So, we either bundle all of `ethers` or that specific package or I suppose I could duplicate the interface as an internal type though that might lead to typing issues for user.  Not sure unless I try it.

packages/statemanager/package.json
holgerd77: I think we should move this whole method over to the `Block` library, this is the part which finally overloads the `StateManager` and is misplaced, also this `Block` dependency is not fitting (a state manager should not depend on a block?).

Since we wanted to improve on Ethers compatibility anyhow: I would suggest that we directly integrate this as a new static constructor `Block.fromEthersRPCProvider(provider, blockTag, opts?: BlockOptions)` (or similar). This is likely generally useful for people. I would have a tendency to do for Tx as well the same thing.

Optimally this should be done in a separate PR and then this PR rebased on top. 😋

And regarding a monorepo-wide Ethers strategy:

I would have a tendency to take `Ethers` in into the monorepo as a relatively "standard" dependency like we have with `ethereum-cryptography`, so also aligning on the version numbers.

This would lead to the following two points:

1. I think we should directly **take "whole" Ethers in** an not only the provider. Then we are more future proof on future use cases and can e.g. easily add a static constructor `Block.fromEthers()` (so one for just the block object, also re-occuring use case). And Ethers is just not such a heavy burden as a dependency, I guess this should be ok.
2. I would have a strong tendency to **directly go for Ethers v6** to be future proof. Final release should be very very imminent.

WDYGT?

packages/evm/package.json
holgerd77: See other comment on Ethers

holgerd77:
Will move here back to "WIP".

packages/statemanager/src/ethersStateManager.ts
jochem-brouwer: `new Map(x)` where `x` is of type `Map` actually copies the entire map so these are copied. But the `_cache` itself is not copied, so this looks a bit inconsistent?

packages/statemanager/src/ethersStateManager.ts
jochem-brouwer: I don't think people should be allowed to use `latest` or `pending` since those are dynamic; if inbetween calls to ethers the block changes then you get incorrect results.

packages/statemanager/package.json
acolytec3: Have made all these moves in block and tx.  Now I just need to figure out what's going on when running some blocks.  I found that 2000005 is returning an incorrect gas calculation (likely related somehow to deleting trie nodes or something but not sure).

packages/block/src/block.ts
```diff
@@ -137,6 +147,65 @@ export class Block {
...
   }
 
+  /**
+   * Creates a new block object from Ethereum JSON RPC.
+   *
+   * @param blockParams - Ethereum JSON RPC of block (eth_getBlockByNumber)
+   * @param uncles - Optional list of Ethereum JSON RPC of uncles (eth_getUncleByBlockHashAndIndex)
+   * @param options - An object describing the blockchain
+   */
+  public static fromRPC(blockData: any, uncles?: any[], opts?: BlockOptions) {
```
holgerd77, line 157: blockParams or blockData? And why is this `any` typed?

packages/block/src/block.ts
```diff
@@ -137,6 +147,65 @@ export class Block {
...
+   *  method to retrieve a block from the provider to use in the VM
+   * @param provider an Ethers JsonRPCProvider
+   * @param blockTag block hash or block number to be run
+   * @param common Common instance used in VM
+   * @returns the block specified by `blockTag`
+   */
+  public static fromEthersProvider = async (
+    provider: ethers.providers.JsonRpcProvider | string,
+    blockTag: string | bigint,
+    common: Common
```
holgerd77, line 171: Can this also get the full options?

packages/block/src/block.ts
```diff
@@ -137,6 +147,65 @@ export class Block {
...
+   * @param blockParams - Ethereum JSON RPC of block (eth_getBlockByNumber)
+   * @param uncles - Optional list of Ethereum JSON RPC of uncles (eth_getUncleByBlockHashAndIndex)
+   * @param options - An object describing the blockchain
+   */
+  public static fromRPC(blockData: any, uncles?: any[], opts?: BlockOptions) {
+    return blockFromRpc(blockData, uncles, opts)
+  }
+
+  /**
+   *  method to retrieve a block from the provider to use in the VM
```
holgerd77, line 162: method -> Method

Description shouldn't be targeted towards VM usage, this is a standalone library.

One additional newline below

holgerd77 (commented):
Some questions and remarks.

packages/statemanager/README.md
```diff
@@ -63,11 +63,14 @@ The `EthersStateManager` can be be used with an `ethers` `JsonRpcProvider` or on
...
-- If you update the block tag (block number or `latest`/`earliest`/`pending`) using `stateManager.setBlockTag()`, the cached accounts/storage/contract code will be cleared and retrieved from the provider at the specified block tag.
-- If you are using the `latest`/`pending` block tag in your use case, ensure you clear the state manager cache between execution runs using `stateManager.clearCache()` to ensure the state manager retrieves correct values for accounts/contract/storage.
 - The Ethers State Manager cannot compute valid state roots when running blocks as it does not have access to the entire Ethereum state trie so can not compute correct state roots, either for the account trie or for storage tries.
-  Refer to [this test script](./tests/ethersStateManager.spec.ts) for complete examples of running transactions and blocks in the `vm` with data sourced from a provider.
+- You cannot use `latest`/`earliest`/`pending` as a block tag and must specify an exact block number for the provider.
```
jochem-brouwer, line 72: Why can you not use `earliest`?

packages/statemanager/README.md
```diff
@@ -63,11 +63,14 @@ The `EthersStateManager` can be be used with an `ethers` `JsonRpcProvider` or on
...
+- If you are replaying mainnet transactions and an account is touched by multiple transactions in a block, you must replay those transactions in order (with regard to their position in that block) or calculated gas will likely be different than actual gas consumed.
```
jochem-brouwer, line 73: Same with storage hits

packages/block/src/block.ts
acolytec3: Good catch.  I meant to update this using `JsonRpcBlock`.  Now updated

packages/block/src/block.ts
acolytec3: Yes.  Updated

packages/block/src/block.ts
acolytec3: Yep, forgot to update when I copied it over

packages/statemanager/README.md
acolytec3: You can.  Added it back

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -0,0 +1,335 @@
...
+    super({})
+    if (typeof opts.provider === 'string') {
+      this.provider = new ethers.providers.StaticJsonRpcProvider(opts.provider)
+    } else if (opts.provider instanceof ethers.providers.JsonRpcProvider) {
+      this.provider = opts.provider
+    } else {
+      throw new Error(`valid JsonRpcProvider or url required; got ${opts.provider}`)
+    }
+
+    this.blockTag = opts.blockTag === 'earliest' ? opts.blockTag : bigIntToHex(opts.blockTag)
```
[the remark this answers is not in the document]
acolytec3: I reworked the usage notes in the Readme somewhat to address this.

jochem-brouwer (approved):
LGTM

holgerd77 (approved):
LGTM

PR #2317, opened 2022-09-27
evm-vm-tests

ScottyPoi:
Continuing work on issue: #2171 
New approach to work done in PR #2299 

Moves (via `git mv`) `vm/tests/tester` into new `evm-vm-tests` package.


holgerd77:
Ah, files would need to be moved with `git mv`, otherwise we'll loose git file history here for the test runner and other files.

ScottyPoi:
> Ah, files would need to be moved with `git mv`, otherwise we'll loose git file history here for the test runner and other files.

Did I otherwise set up the package correctly in the monorepo?

holgerd77:
> > Ah, files would need to be moved with `git mv`, otherwise we'll loose git file history here for the test runner and other files.
> 
> Did I otherwise set up the package correctly in the monorepo?

Do not have the immediate oversight over this TBH. You can analyze the results in the respective Codecov runs (do you have got access?) like https://app.codecov.io/gh/ethereumjs/ethereumjs-monorepo/pull/2317.

This is kind of the baseline why we are doing this (so improve EVM - this -16% on Codecov already looks suspicious? Anyhow: needs some deeper look I guess.).

holgerd77:
Just to note: some of the renamings are not yet working here.

ScottyPoi:
Re: Codecov

I'm trying to figure out how to get codecov to recognize that a piece of one package is tested in another.  In this case, EVM components being tested by `evm-vm-tests`.    Before they were tested from within `vm` tests, so it doesn't make sense to me that the codecov for EVM should go down the way it has.

ScottyPoi:
Oh, actually -- I think I got it. :smile:

ScottyPoi:
I'm stumped on how to rebase this after https://github.com/ethereumjs/ethereumjs-monorepo/pull/2339 was merged.

This PR involves `git mv` for files in the VM/tests directory.  With the name change -- I'm not sure how to resolve this.  If I try to rebase, git cannot apply the `git mv` commit onto master, because the file names have changed.

ScottyPoi:
Closing to be replaced by #2341

PR #2319, opened 2022-09-28
common: use a default chain name  in fromGethGenesis if no chain name specified

g11tech:
seems like files might not have name property in them and hence would cause common constructor to error when undefined chain name gets passed.

To prevent this scenario `chain` representing custom chain name is now mandatory field

### DIFF
--- a/packages/common/src/common.ts
+++ b/packages/common/src/common.ts
@@ -166,7 +166,7 @@ export class Common extends EventEmitter {
   ): Common {
     const genesisParams = parseGethGenesis(genesisJson, chain)
     const common = new Common({
-      chain: genesisParams.name,
+      chain: genesisParams.name ?? 'custom',
       customChains: [genesisParams],
       hardfork,
     })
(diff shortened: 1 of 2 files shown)


acolytec3:
Should we make the chain `name` mandatory or just pass in a default name in the common constructor?  You could just do `const common = new Common ({ chain: genesisParams.name ?? 'testnet'...` or something like that.  I doubt most people are going to care too much about the chain name in these custom parameters situations.

g11tech:
> 

hmm, can pass `custom` less people have to type better it is

packages/common/src/common.ts
```diff
@@ -157,7 +157,7 @@ export class Common extends EventEmitter {
   /**
    * Static method to load and set common from a geth genesis json
    * @param genesisJson json of geth configuration
-   * @param { chain, genesisHash, hardfork } to futher configure the common instance
+   * @param { chain, genesisHash, hardfork } to futher configure the common instance, mandatory option: `chain` representing custom chain name`
```
acolytec3, line 160: This comment is no longer relevant.

acolytec3 (approved):
LGTM

holgerd77 (approved):
LGTM

PR #2320, opened 2022-09-28
Fix SHA3 opcode with KECCAK256

ethosdev:
Reference: https://github.com/ethereum/yellowpaper/pull/825

### DIFF
--- a/packages/evm/src/opcodes/codes.ts
+++ b/packages/evm/src/opcodes/codes.ts
@@ -79,7 +79,7 @@ const opcodes: OpcodeEntry = {
   0x1a: { name: 'BYTE', isAsync: false, dynamicGas: false },
 
   // 0x20 range - crypto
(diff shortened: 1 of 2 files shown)


holgerd77:
Same here, @jochem-brouwer.

I would assume we can just do? 🤔 Or has this some (too) breaking notion (is anyone anywhere relying on these names / otherwise framed: can we ignore the eventual very few (attention: unproven thesis) use cases?)?

jochem-brouwer:
We can merge this without problems, but if there are tracing tools which rely on it being called `SHA3`, this will break.

However, this PR should change all `SHA3` occurences to `KECCAK256`.

holgerd77:
Have internalized this within #2706, hope that's ok, didn't want to take away your commits but this was just a lot easier from a procedural perspective to take these small changes directly in.

This is for our breaking releases (to be released in ~ July, will be a pre-release before).

holgerd77:
Will close here.

Issue #2306, opened 2022-09-23
Request for EthereumJS libraries to stop using ENV variables internally


davidmurdoch:
If a user sets the `DEBUG` environment variable when using CLI tools like `ganache` and `truffle` (even if they do `DEBUG=false` or `DEBUG=0`), EthereumJS libraries take that as a signal to do their own internal debug-related things. Our CLI users most certainly don't want that. And while this can generally be worked around by CLI users, it's inconvenient for users to do so, and it leaks internal abstractions that our users don't care about.

I know Ganache could set `ejsThing.DEBUG = false;` after we instantiate, but this isn't a foolproof solution, as we would be relying on the internals of EJS package to _not_ use the flag set by `process.env.DEBUG` in its own constructor/`create`/`init`-like methods (like `await VM.create();`).

Previously we worked around this by setting `process.env.DEBUG` to `false` (to the boolean value, not as a string) when we webpacked the library; this only worked because the EJS libraries checked for truthiness (`if (this.process.env.DEBUG) {`), but now that the libraries check that the _type_ of the value of `process.env.DEBUG` is not equal to `"undefined"` our little hack doesn't work.

Anyway, we at truffle feel strongly that libraries shouldn't alter behavior based on environment variables and it'd help tremendously if EthereumJS libraries stopped listening to `process.env.DEBUG`.

holgerd77:
Yes, that totally makes sense, I guess we haven't thought about these side effects sufficiently.

Should have no dramatic side effects if we rename to `DEBUG_EM` (for: "EthereumJS Monorepo", alternative naming suggestions welcome from everyone, should just be both expressive and short to type, not so easy 😋).

This should have no side effects on our side I would say on first guess, since we only use (and recommend this to use) for local development purposes.

I will likely plan some bug fix releases for the monorepo for next week or so with some other fixes where this could get in.

Update: just thinking, we just need to make sure that this properly works with the `debug` package in some way, will need some investigation I guess.

ScottyPoi:
Hi @davidmurdoch I'm looking into it.
Could you provide an example for me to replicate this?

holgerd77:
Just giving this some additional thoughts.

[...]

PR #2322, opened 2022-09-29
Debug - Stop relying on ENV variables

ScottyPoi:
Responding to user issue #2306 `Request for EthereumJS libraries to stop using ENV variables internally`

> If a user sets the DEBUG environment variable when using CLI tools like ganache and truffle (even if they do DEBUG=false or DEBUG=0), EthereumJS libraries take that as a signal to do their own internal debug-related things.

This happens due to a flaw in a performance feature designed to reduce the overhead of debug calls if no debugging is desired.  We skip the debug calls only if `process.env.DEBUG === undefined`.  When our libraries are used within other CLI tools, setting the DEBUG environment variable for anything will override this.

The fix is to detach `debug` enable / disable from the environmental variable DEBUG, and instead add `debug?: boolean` to the constructor options, and only initializing the `debug` modules if the option is set to true, and otherwise skipping those calls.

### DIFF
--- a/packages/blockchain/src/consensus/clique.ts
+++ b/packages/blockchain/src/consensus/clique.ts
@@ -7,8 +7,7 @@ import type { Blockchain } from '..'
 import type { Consensus, ConsensusOptions } from './interface'
 import type { Block, BlockHeader } from '@ethereumjs/block'
 import type { CliqueConfig } from '@ethereumjs/common'
-
-const debug = createDebugLogger('blockchain:clique')
+import type { Debugger } from 'debug'
 
 // Magic nonce number to vote on adding a new signer
 export const CLIQUE_NONCE_AUTH = Buffer.from('ffffffffffffffff', 'hex')
@@ -98,8 +97,23 @@ export class CliqueConsensus implements Consensus {
    */
   public _cliqueLatestBlockSigners: CliqueLatestBlockSigners = []
 
-  constructor() {
+  private debug?: Debugger
+
+  constructor(debug: boolean = false) {
     this.algorithm = ConsensusAlgorithm.Clique
+    if (debug === true) {
+      this.debug = createDebugLogger('blockchain:clique')
+      createDebugLogger.enable('blockchain:clique')
+    } else if (process.env.DEBUG !== undefined) {
+      if (
+        process.env.DEBUG.includes('blockchain:*') ||
+        process.env.DEBUG.includes('blockchain*') ||
+        process.env.DEBUG.includes('blockchain:clique')
+      ) {
+        this.debug = createDebugLogger('blockchain:clique')
+        createDebugLogger.enable('blockchain:clique')
+      }
+    }
   }
 
   /**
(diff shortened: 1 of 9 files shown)


ScottyPoi:
One thought is that if we DO want to retain the ability to activate specific debuggers from the command line even with the debug option off, we could still check the process.env.DEBUG string as well.

ScottyPoi:
That way instructions like this can still be valid: https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/vm#understanding-the-vm

packages/blockchain/src/consensus/clique.ts
```diff
@@ -98,8 +97,23 @@ export class CliqueConsensus implements Consensus {
    */
   public _cliqueLatestBlockSigners: CliqueLatestBlockSigners = []
 
-  constructor() {
+  private debug?: Debugger
+
+  constructor(debug: boolean = false) {
```
holgerd77, line 102: Ah, that's a bit of a special case, generally good catch that we are not guarding at all the debug calls here for the consensus implementations.

So: we can't add single constructor options here, since all consensus implementations should adhere to the same interface.

So the `debug` option should be added to `ConsensusOptions` in `interface.ts` in this case, and then a local debug guard should be initialized in the `setup()` method (I guess clique-only), since this method is taking in the `ConsensusOptions` as we have implemented it.

packages/vm/src/types.ts
```diff
@@ -139,6 +139,11 @@ export interface VMOpts {
    * Use a custom EVM to run Messages on. If this is not present, use the default EVM.
    */
   evm?: EVMInterface
+
+  /**
+   * Activate VM debug logging
+   */
```
holgerd77, line 145: The code docs on this should be adjusted to better describe what the option is doing.

When giving the whole option structure another thought: I think it is technically not true that we have this optional since we *do* have this actually activated by default (printing out CL debug statements). So we should rather model this with non-optional but defaulting to `true`.

I guess we also want to rename the option a bit to make it already clear by the option name what is being done here. Maybe we really go with something very explicit like `printCLDebugOutput`.

holgerd77: (if someone has something shorter but similarly expressive let me know)

packages/vm/src/runTx.ts
```diff
@@ -56,7 +54,7 @@ export async function runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
   }
 
   await state.checkpoint()
-  if (this.DEBUG) {
+  if (debug) {
```
holgerd77, line 57: These should all stay on `this.DEBUG`.

holgerd77 (changes requested):
Ok, some additional comments.

Generally this issue implementation should only change some little code in the constructors of the affected libraries and (mostly) not much else and there should be this new (somewhat renamed, see comment) option.

Every library should get 1-2 tests for the option (very simple, instantiating with the option to `false` and `true` and manually set `process.env.DEBUG` respectively and see if `this.DEBUG` is set correctly).

(one side note: in my code example I used lower capital letters for `this.DEBUG` in one conditional, this was just a typo, not to confuse you)

PR #2324, opened 2022-09-29
common: Add merge block numbers for merged networks and fix forkhash calc

g11tech:
Add block numbers where merge hf activates in merge testnet for ease of use (without `td`) requirement.

Also fix forkhash calculation to ignore the merge hardfork even if it might have block number assigned

Tests already cover the forkhash calc change: 
- https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/tests/hardforks.spec.ts#L270 
as the forkhash calc tests was failing for these networks upon addition of block number

### DIFF
--- a/packages/common/src/chains/mainnet.json
+++ b/packages/common/src/chains/mainnet.json
@@ -91,7 +91,7 @@
       "//_comment": "The forkHash will remain same as mergeForkIdTransition is post merge, terminal block: https://etherscan.io/block/15537393",
       "name": "merge",
       "ttd": "58750000000000000000000",
-      "block": null,
+      "block": 15537394,
       "forkHash": "0xf0afd0e3"
     },
     {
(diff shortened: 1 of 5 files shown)


acolytec3 (approved):
LGTM - but the forkhash not changing is so weird!

packages/common/src/chains/mainnet.json
```diff
@@ -91,7 +91,7 @@
       "//_comment": "The forkHash will remain same as mergeForkIdTransition is post merge, terminal block: https://etherscan.io/block/15537393",
       "name": "merge",
       "ttd": "58750000000000000000000",
-      "block": null,
+      "block": 15537394,
```
holgerd77, line 94: https://etherscan.io/block/15537394 🎉

So cool to see this getting in and have a block number and everything. 🥳🥳🥳

💯🙏

So nice to spot the various Etherscan changes between 15537393 and 15537394. 🙂

holgerd77 (commented):
Nice, LGTM (post-PR-merge review).

PR #2325, opened 2022-09-29
Debug - Remove ENV check

ScottyPoi:
### Request for EthereumJS libraries to stop using ENV variables internally

> If a user sets the DEBUG environment variable when using CLI tools like ganache and truffle (even if they do DEBUG=false or DEBUG=0), EthereumJS libraries take that as a signal to do their own internal debug-related things.

This happens due to a *performance feature* designed to reduce the overhead of debug calls if `process` does not exist.  

Rather than execute lines of code - 
#####  `this.debugger("some debug output")`  
when there was no such debug log enabled...

- `this.DEBUG` would default to `false`, and
- all `this.debugger()` calls would be wrapped inside a `if (this.DEBUG === true)...` conditional

#### Issue

Users may also want to utilize the performance feature while operating in an environment where `process` does exist, and `DEBUG` variables are being set for other reasons.  

Currently, the "skip debug" mode only checks if process is `undefined`

```
    // Safeguard if "process" is not available (browser)
    if (typeof process?.env.DEBUG !== 'undefined') {
      this.DEBUG = true
    }
```

This PR introduces a constructor option `disableCLDebugOutput` to `VM`, `EVM`, and `VmState`.

```

  /**
   * Option to disable command line debugging
   * Will take precedence over process.env.DEBUG variable usage
   * Default: `undefined`
   * Set to `true` to disable debugging.
   */
  disableCLDebug?: boolean
```

Setting this option to `true` will allow users to set `env.DEBUG` variables while still enabling this performance feature.
```
    // Optimize performance by skipping debug() calls when debugging has been deactivated
    if (typeof process?.env.DEBUG !== 'undefined' && opts.disableCLDebug !== true) {
      this.DEBUG = true
    }
```

### DIFF
--- a/packages/evm/src/evm.ts
+++ b/packages/evm/src/evm.ts
@@ -298,16 +298,16 @@ export class EVM implements EVMInterface {
       }
     }
 
+    // Optimize performance by skipping debug() calls when debugging has been deactivated
+    if (typeof process?.env.DEBUG !== 'undefined' && opts.disableCLDebug !== true) {
+      this.DEBUG = true
+    }
+
     // We cache this promisified function as it's called from the main execution loop, and
     // promisifying each time has a huge performance impact.
     this._emit = <(topic: string, data: any) => Promise<void>>(
       promisify(this.events.emit.bind(this.events))
     )
-
-    // Safeguard if "process" is not available (browser)
-    if (typeof process?.env.DEBUG !== 'undefined') {
-      this.DEBUG = true
-    }
   }
 
   protected async init(): Promise<void> {
--- a/packages/evm/src/evm.ts
+++ b/packages/evm/src/evm.ts
@@ -129,6 +129,14 @@ export interface EVMOpts {
    * The External Interface Factory, used to build an External Interface when this is necessary
    */
   eei: EEIInterface
+
+  /**
+   * Option to disable command line debugging
+   * Will take precedence over process.env.DEBUG variable usage
+   * Default: `false`
+   * Set to `true` to disable debugging.
+   */
+  disableCLDebug?: boolean
 }
 
 /**
--- a/packages/evm/src/evm.ts
+++ b/packages/evm/src/evm.ts
@@ -206,8 +206,7 @@ export class EVM implements EVMInterface {
    * EVM is run in DEBUG mode (default: false)
    * Taken from DEBUG environment variable
    *
-   * Safeguards on debug() calls are added for
-   * performance reasons to avoid string literal evaluation
+   * debug() calls can be deactivated with deactivateCLDebug
    * @hidden
    */
   readonly DEBUG: boolean = false
--- a/packages/evm/test/debugOutput.spec.ts
+++ b/packages/evm/test/debugOutput.spec.ts
@@ -1,0 +1,26 @@
+import { Chain, Common, Hardfork } from '@ethereumjs/common'
+import debug from 'debug'
+import * as tape from 'tape'
+
+import { EVM } from '../src'
+
+import { getEEI } from './utils'
+
+tape('Tests EVM constructor option to disable command line debugging output', (t) => {
+  t.test('Should default to DEBUG: true', async (t) => {
+    debug.enable('evm')
+    const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Constantinople })
+    const eei = await getEEI()
+    const evm = await EVM.create({ eei, common })
+    t.notok(evm.DEBUG, 'DEBUG defaults to false')
+    t.end()
+  })
+  t.test('Should set DEBUG: false from `deactivateCLDebug: true` constructor option', async (t) => {
+    const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Constantinople })
+    const eei = await getEEI()
+    const evm = await EVM.create({ eei, common, disableCLDebug: true })
+    t.notok(evm.DEBUG, 'DEBUG set to false')
+    t.end()
+  })
+  t.end()
+})
--- a/packages/vm/src/types.ts
+++ b/packages/vm/src/types.ts
@@ -139,6 +139,14 @@ export interface VMOpts {
    * Use a custom EVM to run Messages on. If this is not present, use the default EVM.
    */
   evm?: EVMInterface
+
+  /**
+   * Disable command line debugging.
+   * This setting will override DEBUG environment variables
+   *
+   * Default: 'undefined'
+   */
+  disableCLDebug?: boolean
 }
 
 /**
--- a/packages/vm/src/vm.ts
+++ b/packages/vm/src/vm.ts
@@ -145,16 +145,15 @@ export class VM {
     this._hardforkByBlockNumber = opts.hardforkByBlockNumber ?? false
     this._hardforkByTTD = toType(opts.hardforkByTTD, TypeOutput.BigInt)
 
+    if (typeof process.env.DEBUG !== 'undefined' && opts.disableCLDebug !== true) {
+      this.DEBUG = true
+    }
+
     // We cache this promisified function as it's called from the main execution loop, and
     // promisifying each time has a huge performance impact.
     this._emit = <(topic: string, data: any) => Promise<void>>(
       promisify(this.events.emit.bind(this.events))
     )
-
-    // Safeguard if "process" is not available (browser)
-    if (process !== undefined && typeof process.env.DEBUG !== 'undefined') {
-      this.DEBUG = true
-    }
   }
 
   async init(): Promise<void> {
--- a/packages/vm/src/vm.ts
+++ b/packages/vm/src/vm.ts
@@ -67,8 +67,9 @@ export class VM {
    * VM is run in DEBUG mode (default: false)
    * Taken from DEBUG environment variable
    *
-   * Safeguards on debug() calls are added for
-   * performance reasons to avoid string literal evaluation
+   * debug() calls can be skipped when to optimize performance
+   * By setting VMOpts.disableCLDebug: true
+   * This setting will override the environment variable
    * @hidden
    */
   readonly DEBUG: boolean = false
--- a/packages/vm/src/eei/vmState.ts
+++ b/packages/vm/src/eei/vmState.ts
@@ -55,8 +55,8 @@ export class VmState implements EVMStateAccess {
     this._accessedStorage = [new Map()]
     this._accessedStorageReverted = [new Map()]
 
-    // Safeguard if "process" is not available (browser)
-    if (process !== undefined && typeof process.env.DEBUG !== 'undefined') {
+    // Skip debug() calls if
+    if (!disableCLDebug && typeof process.env.DEBUG !== 'undefined') {
       this.DEBUG = true
     }
     this._debug = createDebugLogger('vm:state')
--- a/packages/vm/src/eei/vmState.ts
+++ b/packages/vm/src/eei/vmState.ts
@@ -37,7 +37,15 @@ export class VmState implements EVMStateAccess {
 
   protected readonly DEBUG: boolean = false
 
-  constructor({ common, stateManager }: { common?: Common; stateManager: StateManager }) {
+  constructor({
+    common,
+    stateManager,
+    disableCLDebug = false,
+  }: {
+    common?: Common
+    stateManager: StateManager
+    disableCLDebug?: boolean
+  }) {
     this._checkpointCount = 0
     this._stateManager = stateManager
     this._common = common ?? new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Petersburg })
--- a/packages/vm/test/api/index.spec.ts
+++ b/packages/vm/test/api/index.spec.ts
@@ -45,6 +45,12 @@ tape('VM -> basic instantiation / boolean switches', (t) => {
     st.end()
   })
 
+  t.test('deactivateCLDebug option should set vm.DEBUG: false', async (t) => {
+    const vm = await VM.create({ disableCLDebug: true })
+    t.notok(vm.DEBUG, 'DEBUG should be disabled')
+    t.end()
+  })
+
   t.test('should be able to activate precompiles', async (st) => {
     const vm = await VM.create({ activatePrecompiles: true })
     st.notDeepEqual(
--- a/packages/vm/test/api/vmState.spec.ts
+++ b/packages/vm/test/api/vmState.spec.ts
@@ -34,6 +34,21 @@ tape('vmState', (t) => {
     }
   )*/
 
+  t.test('Should default to DEBUG: true', async (st) => {
+    const common = new Common({ chain: Chain.Mainnet })
+    const stateManager = new StateManager({})
+    const vmState = new VmState({ common, stateManager })
+    st.notok((vmState as any).DEBUG, 'DEBUG set to true')
+    st.end()
+  })
+  t.test('DEBUG should be disabled by constructor option', (st) => {
+    const common = new Common({ chain: Chain.Mainnet })
+    const stateManager = new StateManager({})
+    const vmState = new VmState({ common, stateManager, disableCLDebug: true })
+    st.notok((vmState as any).DEBUG, 'DEBUG set to false')
+    st.end()
+  })
+
   t.test('should generate the genesis state root correctly for mainnet from common', async (st) => {
     if (isRunningInKarma()) {
       st.skip('skip slow test when running in karma')


packages/evm/src/evm.ts
```diff
@@ -296,8 +304,8 @@ export class EVM implements EVMInterface {
     }
 
     // Safeguard if "process" is not available (browser)
-    if (typeof process?.env.DEBUG !== 'undefined') {
-      this.DEBUG = true
+    if (process === undefined || opts.printCLDebugOutput === false) {
```
holgerd77, line 307: I think we do not want this to "activate" DEBUG by this `printCLDebugOutput`, since this has no value on it's own and is only useful if `DEBUG` is used on the command line. So this leads to cases where debugging "activates" if the option is set to true and `process` is available, this is not adding any value but rather in doubt consumes performance.

The `printCLDebugOutput` option should really just be a - completely separated - guard towards the `DEBUG` usage on the command line.

Please switch the conditional logic back to the "old"/proposed way here, so leaving `  readonly DEBUG: boolean = false` and then set `true` as the default value for `printCLDebugOutput`.

Thanks! 🙂 🙏

packages/evm/src/evm.ts
```diff
@@ -129,6 +129,14 @@ export interface EVMOpts {
    * The External Interface Factory, used to build an External Interface when this is necessary
    */
   eei: EEIInterface
+
+  /**
+   * Option to disable Debug logging
```
holgerd77, line 134: Update: sorry for the twists, I have yet another naming suggestion here with `deactivateCLDebug`, then set to `false` (so adopt the other comment suggestion accordingly), since this is really about deactivation of the whole debugging and not just printing.

Does that make sense?

The doc description should be a bit more descriptive, adding that the option takes precedence over the `DEBUG` CL variable usage.

holgerd77 (changes requested):
Hi there,
thanks for reworking! 🙏

Some comments, can you please integrate and then expand to the other libraries?

Thanks!

holgerd77:
Ah, and not to forget a test case for the option (as described in some older comment).

ScottyPoi:
Something I realized along the way is that the original "safeguard" may not actually be necessary.
```
    // Safeguard if "process" is not available (browser)
    if (typeof process?.env.DEBUG !== 'undefined') {
      this.DEBUG = true
    }
```
`debug` module ***does***  work in the browser.  So the *option*  to deactivate the debugging for *performance*  reasons seems totally valuable, but I don't think we need the "safeguard" for the browser context.

packages/evm/src/evm.ts
```diff
@@ -129,6 +129,15 @@ export interface EVMOpts {
...
    */
   eei: EEIInterface
+
+  /**
+   * Option to disable command line debugging
+   * Will take precedence over process.env.DEBUG variable usage
+   * Default: `false`
+   * Set to `true` to disable debugging.
+   */
+
```
gabrocheleau, line 139:
```suggestion
```

packages/evm/src/evm.ts
```diff
@@ -202,14 +211,13 @@ export class EVM implements EVMInterface {
   public readonly _mcl: any //
 
   /**
-   * EVM is run in DEBUG mode (default: false)
+   * EVM is run in DEBUG mode (default: true)
    * Taken from DEBUG environment variable
    *
-   * Safeguards on debug() calls are added for
-   * performance reasons to avoid string literal evaluation
+   * debug() calls can be deactivated with disableCLDebug
```
gabrocheleau, line 217:
```suggestion
   * debug() calls can be deactivated with deactivateCLDebug
```

packages/evm/test/debugOutput.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+    t.end()
+  })
+  it('Should set DEBUG: false from `deactivateCLDebug: true` constructor option', async (t) => {
+    const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Constantinople })
+    const eei = await getEEI()
+    const evm = await EVM.create({ eei, common, deactivateCLDebug: true })
+    t.notok(evm.DEBUG, 'DEBUG set to false')
+    t.end()
+  })
+  test.end()
```
gabrocheleau, line 26:
```suggestion
tape('Tests EVM constructor option to disable command line debugging output', (t) => {
  t.test('Should default to DEBUG: true', async (st) => {
    debug.enable('evm')
    const common = new Common({ chain: Chain.Mainnet })
    const eei = await getEEI()
    const evm = await EVM.create({ eei, common })
    st.ok(evm.DEBUG, 'DEBUG defaults to true')
    st.end()
  })
  it('Should set DEBUG: false from `deactivateCLDebug: true` constructor option', async (st) => {
    const common = new Common({ chain: Chain.Mainnet })
    const eei = await getEEI()
    const evm = await EVM.create({ eei, common, deactivateCLDebug: true })
    st.notok(evm.DEBUG, 'DEBUG set to false')
    st.end()
  })
  t.end()
```
Current version is fine, just updated for style consistency with the rest of our tests. Also removing the hardfork common option since it's not necessary.

packages/evm/src/evm.ts
```diff
@@ -295,9 +302,10 @@ export class EVM implements EVMInterface {
       }
     }
 
-    // Safeguard if "process" is not available (browser)
-    if (typeof process?.env.DEBUG !== 'undefined') {
-      this.DEBUG = true
```
holgerd77, line 300: The env.DEBUG check shouldn't be removed here, instead both should now count in:

```typescript
if (typeof process?.env.DEBUG !== 'undefined' && !opts.deactivateCLDebug) {
  this.DEBUG = true
}
```

packages/evm/src/evm.ts
```diff
@@ -202,14 +210,13 @@ export class EVM implements EVMInterface {
...
+   * EVM is run in DEBUG mode (default: true)
    * Taken from DEBUG environment variable
    *
-   * Safeguards on debug() calls are added for
-   * performance reasons to avoid string literal evaluation
+   * debug() calls can be deactivated with deactivateCLDebug
    * @hidden
    */
-  readonly DEBUG: boolean = false
+  readonly DEBUG: boolean = true
```
holgerd77, line 219: This should remain `false` by default.

holgerd77: (all other occurrences as well)

holgerd77 (changes requested):
Some change requests.

packages/evm/test/debugOutput.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+tape('Tests EVM constructor option to disable command line debugging output', (t) => {
+  t.test('Should default to DEBUG: true', async (t) => {
+    debug.enable('evm')
+    const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Constantinople })
+    const eei = await getEEI()
+    const evm = await EVM.create({ eei, common })
+    t.notok(evm.DEBUG, 'DEBUG defaults to false')
+    t.end()
+  })
+  t.test('Should set DEBUG: false from `deactivateCLDebug: true` constructor option', async (t) => {
```
davidmurdoch, line 18:
```suggestion
  t.test('Should set DEBUG: false from `disableCLDebug: true` constructor option', async (t) => {
```

packages/vm/src/types.ts
```diff
@@ -139,6 +139,14 @@ export interface VMOpts {
    * Use a custom EVM to run Messages on. If this is not present, use the default EVM.
    */
   evm?: EVMInterface
+
+  /**
+   * Disable command line debugging.
+   * This setting will override DEBUG environment variables
+   *
+   * Default: 'undefined'
```
davidmurdoch, line 147: Is the default supposed to be `false`?

packages/vm/src/vm.ts
```diff
@@ -67,8 +67,9 @@ export class VM {
    * VM is run in DEBUG mode (default: false)
    * Taken from DEBUG environment variable
    *
-   * Safeguards on debug() calls are added for
-   * performance reasons to avoid string literal evaluation
+   * debug() calls can be skipped when to optimize performance
```
davidmurdoch, line 70:
```suggestion
   * debug() calls can be skipped to optimize performance
```

davidmurdoch:
If a `VM` is instantiated with the option `disableCLDebug: true` I think that flag should be passed to all other EJS modules that then get instantiated by the VM by default (like `EVM`, and maybe `VmState`, `Common`, and whatever else uses the flag, etc). Is there are reason not to do this?

Also, I imaging that in most cases the users of these libraries won't want the library's debug statements to be enabled when they set `DEBUG`; so setting the to enabled by default is a bit unexpected. Is this just to avoid a (breaking) change in behavior?

MicaiahReid:
Hi @ScottyPoi is there any update on this?

holgerd77:
I have to say that I am moving more and more away from the solution we tested out here, this is getting ugly and causes too many unnecessary complexities with these new constructor options which *then* even have to be passed around from library to library. Sorry that it took me so long to get to this conclusion. 😬

I would alternatively suggest we limit this early on, e.g. by adding a generic (not too long -> otherwise inconvenient!) EthereumJS prefix like `ethjs:*` (`ethereumjs:` too long, `ejs` to generic) which we then relatively dumb-string-check in our little activation code `if 'ethjs:' in 'DEBUG'` before we activate our own debugger.

This should have very (very) little (respectively: likely rather no) external side effect. At the end this whole debugging thing on our side is for the very big part really just for us, eventually also for some very selected very direct users of the libraries. So on *our* side we can really easily and without broader community effects do this kind of changes here.

What do you think? This just remains so much more local with this and would already do the job I guess.

Sorry again, took ages, I know. But I always felt uneasy when coming back here, but was not able to give this an eventual constructive direction until this renewed reflection round.

davidmurdoch:
**A somewhat unrelated opinion on your comment about options:**

`VM` v6's options passing isn't as full-featured as previous versions. It now necessitates that the we read the source code of all constructors downstream from `VM` because the abstractions are a bit more leaky now.

I think downstream options, like for `EVM` and `EEI`, _should_ be present in `VM`'s options, and then passed downstream when it instantiates the defaults. Because we now have to do our own `EVM`/`EEI` instantiation and pass things into the `VM`, instead of using defaults, we now also have to re-read the constructors at each release -- just in case the `VM` changed the default instantiation of these downstream classes. This is because we _do_ want the default `VM` configuration, mostly, but also with some small tweaks; but we can't do that because the VM doesn't pass the relevant options downstream.

Anyway, _it's not a big deal for us_, really, just a minor inconvenience. I only wanted to mention it here because you said passing options through from one library to another makes the code ugly and complex... but the opposite is actually true in the grand scheme of things. It would make the code of the users of the library cleaner and less complex. There is only one implementation of the VM itself that would have "ugly" or "complex" code, but _thousands_ of uses of the VM (in user code, like Ganache) that would not.

---

**Back to things more directly on topic:**

Ganache ships as a bundle (meaning we webpack nearly everything) to improve security, determinism, and installation time. Previously, during bundling, Ganache would remove EthereumJS's debugging blocks by setting `process.env.DEBUG` to `false` (a boolean, which is actually not a possible value during runtime) so that when the AST is evaluated by webpack it would treat `if (process.env.DEBUG) {...}` as dead code and remove it all together.

It sounds like a hack, but this exact use-case is [documented by webpack](https://webpack.js.org/plugins/environment-plugin/#usage-with-default-values), and successfully prevented EthereumJS's `DEBUG` statements from running for users who use `DEBUG=*` in the environment.

In the latest version we can now set `process.env.DEBUG` to `undefined`, which has a similar effect, but it only removes the `this.DEBUG = true;` block, which isn't quite as good as before because it doesn't remove the actual `debug` blocks (`if (this.DEBUG) { debug(...); }`), even though they are now unused, though technically still reachable if Ganache itself did something like `vmInst.DEBUG = true;` internally.

**All that to say:** we are hoping for a solution that makes it possible for Ganache to prevent EthereumJS's debug statements from ever running while also allowing us to use a minifier to remove the debug blocks altogether. If instead of using `if (this.DEBUG) { debug(...) }` you used something like `if (utils.debugEnabled()) { debug(...) }` we could make use of dead code removal AND users who want to enable DEBUG could toggle `DEBUG` on and off at any time, rather than only before instantiation.

ScottyPoi:
> I would alternatively suggest we limit this early on, e.g. by adding a generic (not too long -> otherwise inconvenient!) EthereumJS prefix like ethjs:* (ethereumjs: too long, ejs to generic) which we then relatively dumb-string-check in our little activation code if 'ethjs:' in 'DEBUG' before we activate our own debugger.

@holgerd77 -- While checking for the prefix would work, we would have to add the 'ethjs:' prefix to all of the debug loggers, which also feels a little ugly, and wasteful of terminal space.  

Also inconvenient for package testing.
This example from `vm/README.md` :
```
DEBUG=vm:tx,vm:evm,vm:ops:sstore,vm:*:gas ts-node test.ts
```
would turn into: 
```
DEBUG=ethjs:vm:tx,ethjs:vm:evm,ethjs:vm:ops:sstore,ethjs:vm:*:gas ts-node test.ts
```

If we're attached to not changing any constructor options, we could just require an extra DEBUG variable called `ethjs` to activate debugging.  Something like:
```
DEBUG=ethjs,vm:tx,vm:evm,vm:ops:sstore,vm:*:gas ts-node test.ts
```
Then in the code we just look for that master 'ethjs' variable to activate the debug logging:
```
    if ( typeof process?.env.DEBUG !== 'undefined' &&
      process.env.DEBUG.includes('ethjs')
    ) {
      this.DEBUG = true
    }
```

holgerd77:
> DEBUG=ethjs,vm:tx,vm:evm,vm:ops:sstore,vm:*:gas ts-node test.ts

@ScottyPoi Yes, that's a lot better, let's do that. 🙂 

@davidmurdoch I am not sure if I follow with all you write since my AST/bundling experience is limited: so do I understand correctly that if we would wrap our `if (this.DEBUG) { // print debug statement }` calls in the VM (and other libs) with an additional method `isDebugActiavated()` - which just returns the state of the `this.DEBUG` variable - the AST removal work would work better? 🤔 Would a `getter` for `this.DEBUG` also do it?

Anyhow, I guess this switch to the `ethjs` solution is unrelated to this respectively the above is a follow-up/additional problem (feature) for you guys. So I guess we can independently do - and eventually release already - the `ethjs` switch.

---

@davidmurdoch regarding the VM -> EVM (EEI) options chaining: can you point me to the code part in Ganache where this takes place to get a better view on how this looks now?

davidmurdoch:
@holgerd77 Webpack treats `process.env` as static (which isn't technically correct, but for all intents and purposes it is fine), and this allows it to replace any reference to things like `process.env.DEBUG` with values we provide at transpilation time, like `false` or `undefined`. Using environment variables to remove debug code, or change behavior, is a common pattern in compiled languages and JavaScript bundlers.

Webpack can't replace `vmInstance.DEBUG` in the same way presumably because `DEBUG` can be changed from outside of the class, and since webpack and its minifier (Terser) aren't actually _optimizing_ compilers, despite the ability to do some dead code removal, it can't determine that other code will never change the value of `DEBUG` to `true`.

So, if a new `isDebugActivated` function was:

```
function isDebugActivated() {
  return typeof process.env.DEBUG !== "undefined";
}
```

and Ganache set `process.env.DEBUG` to `undefined` when transpiling, webpack will turn the above function into:

```
function isDebugActivated() {
  return false;
}
```

and the minifier, Terser, will inline the function's result, `false`, into the call site, and then perform a dead code removal pass:

```
// this line
if (isDebugActivated()) {
  debug(...)
}

// gets inlined into:
if (false) {
  debug(...)
}

// which turns into:
// nothing :-)
```

Unfortunately: webpack and terser do NOT do inline replacement across file boundaries, so Ganache would need that helper function to be redefined in each file its used in (unless Ganache switched to another bundler+minifier that allows for this sort of cross-file optimization). That'd be an annoying thing for EthereumJS to do, so I 1000% understand if you wouldn't want to do this :-)

So, I'll backtrack a bit and suggest that a check like `if (typeof process.env.DEBUG !== undefined){ debug(...); }`, similar to what you used to do, works best for Ganache, as we can then perform debug/dead code removal.

__Just to reiterate: this isn't a huge issue for us, as we've found ways to work around it. We all seriously appreciate the time you all have put it to helping us, discussing these things with us, and trying to make things work.__

---

I don't want to hijack yet another of the EthJS issues with unrelated conversation, but since you asked for details i'll oblige 😛:

Here is our VM instantiation now: https://github.com/trufflesuite/ganache/blob/8016f7c076b13e1ca15c961569fa4edb8a0a2c47/src/chains/ethereum/ethereum/src/blockchain.ts#L699-L707

It could have been something like the following, if the `allowUnlimitedContractSize` option was passed through like it was before (IIRC):

```typescript
const vm = await VM.create({
      activatePrecompiles: false,
      allowUnlimitedContractSize: true,
      common,
      blockchain,
      stateManager
    });
```

ScottyPoi:
Resubmitted as PR #2433

PR #2328, opened 2022-10-04
StateManager: stateless verkle statemanager

gabrocheleau:
This PR builds on #1721 to resume work on the stateless verkle state manager.


Pemburu88 (dismissed):

paulmillr:
I know the current version is basically prototyping, but at some point the wasm binary will need to be replaced with JS, because wasm is not supported in all environments, due to CSP, or other reasons.

One example that cannot use wasm is Metamask.

holgerd77:
I know this will be some substantial amount of work (would count in 2 days or so) but it would also be good for this to make the switch to `develop-v7` relatively soon (rather weeks than months, so: before the releases). This would generally be necessary at some point and if we do early this could help a lot to see if the structures we built up there (or: changed) are suitable to allow for an easy integration of other state managers.

holgerd77:
> > I know this will be some substantial amount of work (would count in 2 days or so) but it would also be good for this to make the switch to `develop-v7` relatively soon (rather weeks than months, so: before the releases). This would generally be necessary at some point and if we do early this could help a lot to see if the structures we built up there (or: changed) are suitable to allow for an easy integration of other state managers.
> 
> Absolutely! I was thinking about that while contributing to the buffers -> uint8array transition. Definitely makes sense to do this sooner rather than latter. I am planning on tackling this over the weekend.

Pretty cool! 🙂 👍

gabrocheleau:
Superseded by #2643

PR #2333, opened 2022-10-06
client: Provide txPool.txsByPriceAndNonce with correct vm for fetching txs to build block

g11tech:
Currently the txpool.txsByPriceAndNonce filter transaction based on txpool's vm's eei, which could be different from the eei/statemanager which is building the block (and set to different root). 

This lead to exclusion of subsequent transactions build on top of previous one in the `shandong` testnet experiments.

This PR passes the correct vm to apply to filter the transactions from the pool.

Test case for this will be automatically included in the `shandong` sim test which builds consecutive transactions on same address.


jochem-brouwer (approved):
LGTM!

PR #2341, opened 2022-10-08
Create evm-vm-tests package

ScottyPoi:
Replacing PR #2317 
Working on issue #2171 

Moves `state` and `blockchain` tests from `VM/test` into new package `evm-vm-tests`.

Edits github workflows to run tests from evm-vm-tests

Edits EVM test script to run state test from evm-vm-tests


ScottyPoi:
@holgerd77 

I'm not sure why there are still 10 "Expected Checks" that do not run.

Other than that, everything else passes, and the codecov report does not change.

I know that the ultimate goal is better coverage for EVM by altering the state tests.  Should I continue that work on this PR, or open a new one to work on improving the coverage?

holgerd77:
This should not be a dedicated package on its own (or has there been any discussion on this that I missed? Otherwise I am not seeing a reason for this?).

So `evm-vm-tests` would just be a folder with the files moved. Test triggering would just work "as normal" from the respective `evm` and `vm` packages, just with the file paths slightly adopted.

(again: or do I miss any additional advantages of this particular solution? I feel this rather as substantial additional overhead with not too much pracitcal benefit.)

ScottyPoi:
> This should not be a dedicated package on its own (or has there been any discussion on this that I missed? Otherwise I am not seeing a reason for this?).

@holgerd77  I based this on your comment in https://github.com/ethereumjs/ethereumjs-monorepo/pull/2299


> The one thing we can think about is to extract the whole test runner (everything in tester...) into a separate folder evm-vm-test runner or the like on the packages file level.

Happy to redo differently if you want.

jochem-brouwer:
Note: in the near future I will get #2000 ready, and this will also allow us to run the tests directly in the `EVM` package. What we will most likely do in the mid-term future is that we run retesteth against it, and thus use that as a "test package" and delete our own test package.

holgerd77:
No follow-up and not considered practical, will close

Issue #2343, opened 2022-10-10
Support using hashed account and storage tries


scorbajio:
Currently, in order to put an account into the state manager, it has to be done so using the `putAccount` function, which requires an account address. During snap sync, account addresses are not known, only their hash. In order to store state from snap sync in state manager, it is necessary to add this functionality.

Also see #2345.

jochem-brouwer:
The same would also apply for storages. Can we not directly dump these into an unhashed trie into snap sync, or do we really need to use StateManager?

scorbajio:
> The same would also apply for storages. Can we not directly dump these into an unhashed trie into snap sync, or do we really need to use StateManager?

@jochem-brouwer it is possible to put the data directly into tries, but it seems like we would be recreating a significant amount of StateManager functionality. It may be advantageous for code reuse to use a modified StateManager.

holgerd77:
@jochem-brouwer @scorbajio is it eventually possible to combine both and initialize a StateManager "as is" with an unhashed trie? 🤔 Just to throw another option in the ring.

jochem-brouwer:
Hmm that would most likely fix the problem in an elegant way. Thoughts @scorbajio ?

scorbajio:
@holgerd77 What do you mean by "combine both"? Do you mean support both hashed and unhashed trie puts for StateManager?

scorbajio:
> @jochem-brouwer @scorbajio is it eventually possible to combine both and initialize a StateManager "as is" with an unhashed trie? thinking Just to throw another option in the ring.

@holgerd77 What do you mean by "combine both"? Do you mean support both hashed and unhashed trie puts for StateManager?

holgerd77:
I meant: combine both ideas 🙂, so using the Trie library in an unhashed fashion (idea 1) and then put this in the StateManager (idea 2).

That was not a super-thought-through works-for-sure suggestion though, so I am neither sure if this has caveats or works at all. I just threw this in the ring as an additional eventual option, feel free to discard immediately if not fitting. 😋

scorbajio:
@holgerd77 If I understand correctly, you are saying initialize the `StateManager` with `useKeyHashing` set to `false`, which means that if an address is passed in, it will have to be hashed before putting it, and if an already hashed key is passed in, it can just be put immediately, since it's not necessary to rehash it. Please correct me if I have misunderstood you. 

[...]

PR #2345, opened 2022-10-10
client: Add support for storage and bytecode fetching

scorbajio:
This PR continues where #2107 left off. The goals for this PR include the following:
* Create storage fetcher
* Create bytecode fetcher
* Create a queuing layer where tasks can be efficiently batched and queued depending on accounts fetched so far for fetching storage and bytecodes - A simpler place to start is queuing contract accounts as soon as they are identified in the `process()` function of account fetcher for fetching

It is possible to store data directly into tries, but we may be able to use `StateManager` if support is added for putting accounts by hash. Also see #2343 and #2389.


holgerd77:
Also here, what's the state, how does this relate to the other PRs, a short update would be nice!

scorbajio:
> Also here, what's the state, how does this relate to the other PRs, a short update would be nice
Currently, I am investigating the results of the account fetcher. I have not been able to confirm that the account fetcher can fetch a known account by address. I have upgraded my dev setup to include a hackable boot node, so I should be able to get more clarity into what's going on on the geth side. After resolving the account fetching issue and being able to fetch a contract account and seeing a non-null `storageRoot`, I should be able to continue where I left off with the storage fetcher.

scorbajio:
> Also here, what's the state, how does this relate to the other PRs, a short update would be nice!



> Also here, what's the state, how does this relate to the other PRs, a short update would be nice!

Also here, what's the state, how does this relate to the other PRs, a short update would be nice
Currently, I am investigating the results of the account fetcher. I have not been able to confirm that the account fetcher can fetch a known account by address. I have upgraded my dev setup to include a hackable boot node, so I should be able to get more clarity into what's going on on the geth side. After resolving the account fetching issue and being able to fetch a contract account and seeing a non-null storageRoot, I should be able to continue where I left off with the storage fetcher.

packages/client/lib/config.ts
```diff
@@ -266,6 +266,8 @@ export interface ConfigOptions {
    * The time after which synced state is downgraded to unsynced
    */
   syncedStateRemovalPeriod?: number
+
+  maxStorageRange?: bigint
```
g11tech, line 270: can we move it below maxAccountRange for some context clustering

scorbajio: done

packages/client/lib/sync/fetcher/fetcher.ts
```diff
@@ -312,7 +312,7 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable
     this.nextTasks()
     const job = this.in.peek()
     if (!job) {
-      if (this.finished !== this.total) {
+      if (this.finished >= this.total) {
```
g11tech, line 315: when can finished be > total? if this is happening indicates a bug somewhere

scorbajio: Yeah, I had a feeling that it would be more of a bug than a feature. It happens with the storage fetcher. I'll take another look at this.

g11tech: ok i have found the source of bug :sweat_smile: , will fix and update PR

g11tech (approved):
lgtm! further improvements and enhancements to be done in the followup PRs

g11tech (changes requested):
self todo: blocking to furher increase coverage for storagefetcher

g11tech (approved):
approving further coverage to be enhanced in followup PRs

holgerd77:
Great stuff, congrats on the merge here! 🙏 🎉 🤩 ❤️

Thanks Amir for all the patience when working through the intricacies here and Gajinder for the help bringing this over the finish line. 🙂

PR #2351, opened 2022-10-12
common: Update EIP-3675 to final

acolytec3:
Update status of EIP-3675 to Final since the merge is now no longer soon :tm: but here.

### DIFF
--- a/packages/common/src/eips/3675.json
+++ b/packages/common/src/eips/3675.json
@@ -3,7 +3,7 @@
   "number": 3675,
   "comment": "Upgrade consensus to Proof-of-Stake",
   "url": "https://eips.ethereum.org/EIPS/eip-3675",
-  "status": "Review",
+  "status": "Final",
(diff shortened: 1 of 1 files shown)


jochem-brouwer (approved):

gabrocheleau (approved):

holgerd77:
Just some additional comment: the `status` field was meant to - somewhat - reflect the status from the EIP, so that should be the main trigger for `status` changes rather then some loose "this has already happened" argumentation, since it happens that this is "out of sync" (HF has happened, but no one has updated the EIP 😋).

So in these cases we would update our status before the EIP status, which also feels a bit weird to me.

Just checked though, [EIP-3675](https://eips.ethereum.org/EIPS/eip-3675) also moved to `Final` some time ago.

That said: this `status` field on our side is also not so bloody important. 😆 (just some rough reference for orientation for us as well as Common users).

PR #2352, opened 2022-10-12
Fix `tests` directory references 

acolytec3:
Fixes various references to `tests` in packages where the test directory was renamed from `tests` to `test`.  The most impactful one is the update to the `retesteth` configuration which I have tested and verified that it works again.


jochem-brouwer (approved):
:+1:

PR #2353, opened 2022-10-12
Implement EIP4895: Beacon Chain withdrawals

jochem-brouwer:
Implements [EIP 4895](https://eips.ethereum.org/EIPS/eip-4895)

TODOs:

- [ ] Add JSON support to block/header

### DIFF
--- a/packages/block/src/block.ts
+++ b/packages/block/src/block.ts
@@ -3,12 +3,10 @@ import { RLP } from '@ethereumjs/rlp'
 import { Trie } from '@ethereumjs/trie'
 import { Capability, TransactionFactory } from '@ethereumjs/tx'
 import {
-  Address,
   KECCAK256_RLP,
   arrToBufArr,
   bigIntToHex,
   bufArrToArr,
-  bufferToBigInt,
   bufferToHex,
   intToHex,
   isHexPrefixed,
@@ -20,7 +18,14 @@ import { ethers } from 'ethers'
 import { blockFromRpc } from './from-rpc'
 import { BlockHeader } from './header'
 
-import type { BlockBuffer, BlockData, BlockOptions, JsonBlock, JsonRpcBlock, Withdrawal } from './types'
+import type {
+  BlockBuffer,
+  BlockData,
+  BlockOptions,
+  JsonBlock,
+  JsonRpcBlock,
+  Withdrawal,
+} from './types'
 import type { Common } from '@ethereumjs/common'
 import type {
   FeeMarketEIP1559Transaction,
@@ -28,6 +33,7 @@ import type {
   TxOptions,
   TypedTransaction,
 } from '@ethereumjs/tx'
+import type { Address } from '@ethereumjs/util'
 
 /**
  * An object that represents the block.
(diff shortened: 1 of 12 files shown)


jochem-brouwer:
This would be ready for a first round of reviews (but still WIP).

g11tech:
picking this up

g11tech (approved):
@jochem-brouwer we can merge this, will start a separate PR to add the engine api methods and the respective handling

holgerd77:
That was a pretty quick merge for such a big and extensive PR here. 🤔 Let's please keep this a bit "under review" and under observation for some time post-merge, eventually we can also add some additional tests.

This touches a lot of code parts of production libraries like the Block library.

packages/block/src/block.ts
```diff
@@ -33,6 +42,7 @@ export class Block {
   public readonly header: BlockHeader
   public readonly transactions: TypedTransaction[] = []
   public readonly uncleHeaders: BlockHeader[] = []
+  public readonly withdrawals?: Withdrawal[]
```
holgerd77, line 45: Thinking about this for some time now: `withdrawals` have this type `Withdrawal` here with all these non-deterministic types like `BigIntLike`:

```typescript
export type Withdrawal = {
  index: BigIntLike
  validatorIndex: BigIntLike
  address: AddressLike
  amount: BigIntLike
}
```

While this always comes in handy for input data I wonder if this is an optimal design choice for the internal representation of this data. In the case of withdrawals, this `Withdrawal` type atm actually serves for *both* puroposes.

With this kind of design we just can't be sure what type these internal withdrawal parameters like `index` or `address` really is. My fear is that this will lead to a myriad of follow-up problems. For test cases, client usages,... one can never really count on what the type really is here and either cast here, make code comments, add complex procedural choice conditional switches for values on processing, or minimally: one always need to convert with `withdrawalToBufferArray()`.

I would have a relatively strong tendency that we should switch here to deterministic values, I guess that will safe us from some substantial amount of follow-up trouble.

So I would suggest to actually have *two* types here, the one from above renamed to `WithdrawalData` or similar and used for withdrawal inputs.

And another one `Withdrawal` with deterministic types. Following our tendency to represent internal values not always with Buffer any more but with more "matching" types this would be something like the following I guess:


```typescript
export type Withdrawal = {
  index: BigInt
  validatorIndex: BigInt
  address: Address
  amount: BigInt
}
```

I know that this is some extra effort to rewrite now with various code already in place but I am relatively sure that this more than pay off in the future.

holgerd77: Update: Haha https://github.com/ethereumjs/ethereumjs-monorepo/pull/2401/files#diff-6ef8827503f0272ab6d926170261d659f841b39a00e354abcd69ea5635c8eae1

Just discovered on the follow up PR from @g11tech.

At least we think in impressively similar lines. 🤩 😂

packages/vm/test/api/EIPs/eip-4895-BeaconChainWithdrawals.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+  chain: Chain.Mainnet,
+  hardfork: Hardfork.Merge,
+})
+
+const pkey = Buffer.from('20'.repeat(32), 'hex')
+
+tape('EIP4895 tests', (t) => {
+  t.test('EIP4895: withdrawals execute as expected', async (st) => {
+    const vm = await VM.create({ common })
+    vm._common.setEIPs([4895])
```
holgerd77, line 21: This should be set using the`eips` constructor option along `Common` initialization and not by injecting this afterwards to the semi-private `_common` object.

tuanggo (commented):
open

PR #2358, opened 2022-10-14
client: fix validation logic for optional params in rpc

acolytec3:
Fixes a bug discovered in `eth_estimateGas` where the parameter logic was being applied to an optional parameter when the optional parameter didn't exist.


g11tech (approved):
lets go!

PR #2361, opened 2022-10-17
New Releases

holgerd77:
First round of new releases - mainly bugfix releases - following the final breaking release round from September 6, 2022. Some awaited bug fixes in EVM and Trie package e.g., the EVM switches the event emitter package along. Another noteworthy changes and features, new experimental EthersStateManager and Geth genesis.json import functionality for the Common library.

Additional changes:

- EVM: added getActiveOpcodes() in ac16d71783dcc0a582245384c05433a40c3f654c, see #2363 

Open for review, thanks everyone for the continued work on this. 🙂

This also includes a small Client release.

Note: optimally please push fixes directly on the branch and then re-approve, thanks! 🙏

PR #2365, opened 2022-10-18
evm: expose getActiveOpcodes

### LINKED ISSUE #2363: Expose getActiveOpcodes to the VM object
Before v6, which separated the `vm` and `evm` packages, one could instantiate a `vm` object and use: `vm.getActiveOpcodes` ot get the list of available opcodes.
Now that these 2 packages had been separated, the `getActiveOpcodes` function is no longer available in the context of a `vm` object, but only if you instantiate a separate `evm` object and access it via that object.

This creates an awkward situation for us, where we would instantiate both `vm` and `evm` objects, but the `evm` object will only be used for accessing the `getActiveOpcodes`.

See [this](https://github.com/comitylabs/evm.codes/pull/180#discussion_r996768877) comment for reference.

jochem-brouwer:
This now allows to access `getActiveOpcodes` if `EVMInterface` is of type `EVM`. For instance, in VM:

`const opcodes = vm.evm.getActiveOpcodes!()`

Closes #2363

### DIFF
--- a/packages/evm/src/types.ts
+++ b/packages/evm/src/types.ts
@@ -1,7 +1,7 @@
 import type { EVM, EVMResult, ExecResult } from './evm'
 import type { InterpreterStep } from './interpreter'
 import type { Message } from './message'
-import type { OpHandler } from './opcodes'
+import type { OpHandler, OpcodeList } from './opcodes'
 import type { AsyncDynamicGasHandler, SyncDynamicGasHandler } from './opcodes/gas'
 import type { Account, Address, PrefixedHexString } from '@ethereumjs/util'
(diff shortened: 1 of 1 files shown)


holgerd77:
Oh, sorry, have done exactly the same here https://github.com/ethereumjs/ethereumjs-monorepo/pull/2361/commits/ac16d71783dcc0a582245384c05433a40c3f654c 2 minutes ago. 😆 Is it ok if we close here (otherway around is also fine to me)?

jochem-brouwer:
Yeah I just noticed haha. I will close :)

holgerd77:
Lol, yeah, funny sometimes! 🙂 Thanks for the flexibility!

PR #2368, opened 2022-10-18
EIP3540 run EOF code correctly

jochem-brouwer:
After the new updates in https://github.com/ethereum/tests/pull/847, we have many failing tests, and one of the causes is that we do not run the actual EOF code, but we run the entire container (so EOF code immediately runs into 0xEF, which is the invalid opcode). 

Also thanks to @pk910 who discovered this on Shangdong testnet :)

To run:

```
cd ./packages/ethereum-tests
gh pr checkout 847
cd ../vm
```

Run tests;

Current:

```
npm run test:state -- --fork=London+3540+3670
# tests 293
# pass  171
# fail  122
```

```
npm run test:blockchain -- --fork=London+3540+3670
# tests 2426
# pass  1588
# fail  838
```


packages/evm/src/evm.ts
```diff
@@ -858,6 +858,9 @@ export class EVM implements EVMInterface {
       } else {
         message.code = await this.eei.getContractCode(message.codeAddress)
         message.isCompiled = false
+        if (this._common.isActivatedEIP(3540)) {
+          message.code = getEOFCode(message.code)
```
acolytec3, line 862: Do we really need this?  `runInterpreter` checks the code passed in as part of the message for the EOF header and sets the PC to the correct position based on the sections header.

acolytec3: I see that more tests fail when I comment it out but from how we handle code in the interpreter, it seems like they shouldn't be failing.  I looked back at the original PR and the idea was that we handled all the bits about defining the code section inside `interpreter.run` so this theoretically should be duplicative.

jochem-brouwer:
Now have 4 tests failing. This is due to CODECOPY and likely also CODESIZE. If you run these if you `CALL` a contract, then `interpreter.code` is not the entire container, it is the runtime code. At this point I do not know how to cleanly fix this though. It seems that we are inconsistent in situations, because if you try to create a contract (so, the `to` field is `undefined`) then the `.code` is actually set to the entire container...

acolytec3:
> Now have 4 tests failing. This is due to CODECOPY and likely also CODESIZE. If you run these if you `CALL` a contract, then `interpreter.code` is not the entire container, it is the runtime code. At this point I do not know how to cleanly fix this though. It seems that we are inconsistent in situations, because if you try to create a contract (so, the `to` field is `undefined`) then the `.code` is actually set to the entire container...

See my latest commit.  I think this relatively cleanly solves it by adding a new `codeContainer` field that's only populated when you call `_loadCode`.  The `codeContainer` only gets referenced by `CODESIZE` and `CODECOPY`.  `EXTCODESIZE` and `EXTCODECOPY` both reference the entire code pulled from the state manager so should work as currently coded with no changes.

holgerd77:
Cool!

Why the hell didn't we pack this in the hotfix releases?  😋

acolytec3:
Never too late 🙃

acolytec3:
I added a test for the new `getEOFCode` function so hopefully that will help with the code coverage.  Are we ready to merge this?  It'd be awesome to restart the Shandong network early next week with EIP-3540 working correctly.

jochem-brouwer:
I think in general we also want another client (i.e. Geth) to run alongside our Shangdong testnet (we should run this ourselves) so we can quickly figure out if there is a consensus bug.

@acolytec3 can you confirm that the current EIP3540 tests pass?

acolytec3:
I need to run the state tests but all of the Blockchain tests pass except for that over where I raised the question about the container length which appears to be invalid. I'll check the state tests today and see if there are any remaining failures.

acolytec3:
> I think in general we also want another client (i.e. Geth) to run alongside our Shangdong testnet (we should run this ourselves) so we can quickly figure out if there is a consensus bug.

I believe Nethermind is planning to join and they have implemented EIP3540/3670 as well. I'm not sure if Geth has it integrated in their master branch yet

holgerd77:
> I need to run the state tests but all of the Blockchain tests pass except for that over where I raised the question about the container length which appears to be invalid.

@acolytec3 has this been confirmed (e.g. by the test authors on ethereum/tests or similar) that this test is invalid? Can you provide a reference here?

holgerd77:
I wonder if we should also prepare a quick EVM-only bugfix release on this? Could do this tomorrow morning.

Issue #2370, opened 2022-10-19
Trie integrity verification is leaking database implementation details and forces sync


faustbrian:
https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/trie/src/trie/trie.ts#L803 is leaking implementation details by assuming that the underlying database of every `DB` implementation is exposed as a public variable `_database` and that the `keys` function is synchronous.

The fix is fairly straightforward. Do `for (const dbkey of await this._db.db.keys())` instead of accessing the `keys` function on the underlying storage _(which is not guaranteed to even have such a method)_ and assuming that it us synchronous. Then add the `keys` function to the `DB` interface and have each database implement it. _This can cause applications to crash if they use a custom database implementation that doesn't leak its internals like the default implementations do._

**I'll submit a PR for this tomorrow morning or on Friday.**

PR #2372, opened 2022-10-19
refactor(trie): introduce `DB#keys` function to avoid database implementation leakage

faustbrian:
Resolves https://github.com/ethereumjs/ethereumjs-monorepo/issues/2370


jochem-brouwer (commented):
This looks good, but I do think it is better performance-wise if you create an iterator method. Now, these `keys()` return *all* keys, while in some cases you would only need the first key, or you have a for loop which breaks if a certain condition is met?

holgerd77:
Can this branch eventually get an update according to the latest comment?

PR #2376, opened 2022-10-20
Switch eventemitter2 to local asynceventemitter

### LINKED ISSUE #2366: EventEmitter2 changes how async events work
In old `async-eventemitter`:

`.on('step', (e) => {})` -> immediately resolves
`.on('step', (e, fn) => {setInterval(fn,1000)}` -> hangs interpreter for 1 s
`.on('step', async (e, fn) => {setInterval(fn,1000)}` -> hangs interpreter for 1 s
`.on('step', (e, fn) => {}` -> quits js (?) (I guess this is because node detects that there are only hanging promises left, and nothing is ran, so it deadlocks and therefore just quits)

In `EventEmitter2`:

```
.on('step') (e: InterpreterStep) => {
return new Promise((resolve) => {setTimeout(resolve, 1000)})
}
```

Also, the `EventEmitter2` events are now not typed, i.e. `.on('step', (e) => {})`, now `e` is cast as `any`

This is breaking and should be resolved.

TODO

- [ ] Ensure async events work as before
- [ ] Ensure events are typed
- [ ] Add a test which ensures async events work as in the past

acolytec3:
Fixes #2366 

This borrows code from the old `async-eventemitter` module, migrates it to typescript, and replaces `eventemitter2` with our original API usage (hopefully).

### DIFF
--- a/packages/evm/test/asyncEvents.spec.ts
+++ b/packages/evm/test/asyncEvents.spec.ts
@@ -16,7 +16,7 @@ tape.only('async events', async (t) => {
     const startTime = Date.now()
     setTimeout(() => {
       t.ok(Date.now() > startTime, 'evm paused on step function for one second')
-      next()
+      next?.()
     }, 1000)
   })
   const runCallArgs = {
--- a/packages/evm/test/asyncEvents.spec.ts
+++ b/packages/evm/test/asyncEvents.spec.ts
@@ -6,7 +6,7 @@ import { EVM } from '../src'
 
 import { getEEI } from './utils'
 
-tape.only('async events', async (t) => {
+tape('async events', async (t) => {
   t.plan(2)
   const caller = new Address(Buffer.from('00000000000000000000000000000000000000ee', 'hex'))
   const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Constantinople })
--- a/packages/evm/test/asyncEvents.spec.ts
+++ b/packages/evm/test/asyncEvents.spec.ts
@@ -1,0 +1,28 @@
+import { Chain, Common, Hardfork } from '@ethereumjs/common'
+import { Address } from '@ethereumjs/util'
+import * as tape from 'tape'
+
+import { EVM } from '../src'
+
+import { getEEI } from './utils'
+
+tape.only('async events', async (t) => {
+  t.plan(2)
+  const caller = new Address(Buffer.from('00000000000000000000000000000000000000ee', 'hex'))
+  const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Constantinople })
+  const eei = await getEEI()
+  const evm = await EVM.create({ common, eei })
+  evm.events.on('step', async (event, next) => {
+    const startTime = Date.now()
+    setTimeout(() => {
+      t.ok(Date.now() > startTime, 'evm paused on step function for one second')
+      next()
+    }, 1000)
+  })
+  const runCallArgs = {
+    caller, // call address
+    gasLimit: BigInt(0xffffffffff),
+    data: Buffer.from('600000', 'hex'),
+  }
+  await evm.runCall(runCallArgs)
+})
--- a/packages/evm/test/asyncEvents.spec.ts
+++ b/packages/evm/test/asyncEvents.spec.ts
@@ -15,7 +15,7 @@ tape.only('async events', async (t) => {
   evm.events.on('step', async (event, next) => {
     const startTime = Date.now()
     setTimeout(() => {
-      t.ok(Date.now() > startTime, 'evm paused on step function for one second')
+      t.ok(Date.now() > (startTime + 999), 'evm paused on step function for one second')
       next?.()
     }, 1000)
   })
--- a/packages/evm/test/asyncEvents.spec.ts
+++ b/packages/evm/test/asyncEvents.spec.ts
@@ -15,7 +15,7 @@ tape.only('async events', async (t) => {
   evm.events.on('step', async (event, next) => {
     const startTime = Date.now()
     setTimeout(() => {
-      t.ok(Date.now() > (startTime + 999), 'evm paused on step function for one second')
+      t.ok(Date.now() > startTime + 999, 'evm paused on step function for one second')
       next?.()
     }, 1000)
   })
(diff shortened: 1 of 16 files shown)


acolytec3:
@thevaizman, I re-implemented `async-eventemitter` in Typescript within our repo and slimmed it down to just overwrite the `emit` property (which is the only bit of the old `async-eventemitter` module we used internally).  When you have a chance, could you try pointing your local evm.codes build to this branch and see if everything works correctly (including your async event logic)?

thevaizman:
@acolytec3 hey, unfortunately, I am not able to test this.
there is no easy way to simply point all packages to the git branch since this repo is a monorepo. this means that i have to build and pack using `yarn pack` each package and point to the local `.tgz` file on my machine from the `package.json`.
while this works fine, the inner dependencies are then fetched remote and i am not able to make `yarn` resolve those locally (since i can't use `yarn link` as it is incompatible with SWC 😞 )

i think i will have to wait until this change is "released" to test it...

acolytec3:
Interesting. What were you doing before that allowed you to test against local builds of our code?

acolytec3:
@holgerd77 provided others on the team validate this, could we do a beta release with this PR merged to verify that everything works correctly with evm.codes (and hopefully ganache too)?

thevaizman:
> Interesting. What were you doing before that allowed you to test against local builds of our code?

changes before didn't include any inner dependencies changes (i.e `@ethereumjs/util`) afaik. therefore, i could simply use `yarn pkg` and install the local packages without problems.
however, when trying to do the same thing now, i get the following error:
```typescript
./node_modules/@ethereumjs/evm/src/evm.ts:4:3
Type error: Module '"@ethereumjs/util"' has no exported member 'AsyncEventEmitter'.

  2 | import {
  3 |   Address,
> 4 |   AsyncEventEmitter,
    |   ^
  5 |   KECCAK256_NULL,
  6 |   MAX_INTEGER,
  7 |   bigIntToBuffer,
error Command failed with exit code 1.
```

which as far as i understand stems from the inner dependency `@ethereumjs/util` being resolved remotely as part of the `evm` package. obviously, the remote `util` package is different than the local one, as it doesn't have the `AsyncEventEmitter` class, which causes this error.

packages/util/src/asyncEventEmitter.ts
```diff
@@ -0,0 +1,194 @@
+import { eachSeries } from 'async'
```
holgerd77, line 1: Just so that we do not forget: please add license and a source reference here, both to the async-eventemitter package as well as the corresponding definitely-typed TS files.

gabrocheleau (approved):

gabrocheleau (dismissed):

packages/evm/test/asyncEvents.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+tape.only('async events', async (t) => {
+  t.plan(2)
+  const caller = new Address(Buffer.from('00000000000000000000000000000000000000ee', 'hex'))
+  const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Constantinople })
+  const eei = await getEEI()
+  const evm = await EVM.create({ common, eei })
+  evm.events.on('step', async (event, next) => {
+    const startTime = Date.now()
+    setTimeout(() => {
+      t.ok(Date.now() > startTime, 'evm paused on step function for one second')
```
jochem-brouwer, line 18:
```suggestion
      t.ok(Date.now() > (startTime + 999), 'evm paused on step function for one second')
```

More explicit check

jochem-brouwer (commented):
This looks great, but this is an error we did not have in the past: 

image

gabrocheleau:
> This looks great, but this is an error we did not have in the past:
> 
> image

I thought I had addressed this in an earlier commit. Which file is that?

packages/evm/test/asyncEvents.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { Chain, Common, Hardfork } from '@ethereumjs/common'
+import { Address } from '@ethereumjs/util'
+import * as tape from 'tape'
+
+import { EVM } from '../src'
+
+import { getEEI } from './utils'
+
+tape.only('async events', async (t) => {
```
gabrocheleau, line 9:
```suggestion
tape('async events', async (t) => {
```

jochem-brouwer:
@gabrocheleau `asyncEvents.spec.ts`

gabrocheleau:
> @gabrocheleau `asyncEvents.spec.ts`

I don't hav ethat locally. Did you pull latest?

jochem-brouwer:
@gabrocheleau I have and you should too; https://github.com/ethereumjs/ethereumjs-monorepo/blob/localized-asynceventemitter/packages/evm/test/asyncEvents.spec.ts

gabrocheleau:
> @gabrocheleau I have and you should too; https://github.com/ethereumjs/ethereumjs-monorepo/blob/localized-asynceventemitter/packages/evm/test/asyncEvents.spec.ts

I see `next?.()`. Did you mean that optional chaining wasn't necessary in the past?

jochem-brouwer:
@gabrocheleau I thought it was, but I just rechecked (`git checkout 6eb5a85a48c3b6027ccf1d59a177c4103eaca0b1` - this was the commit right before we switched to eventemitter2). It indeed raises the same error, so dismiss my comment.

jochem-brouwer (approved):
I have checked the behavior and the type casting, this seems to match how all worked before we went to eventemitter2, so I will approve here. Thanks @acolytec3 :+1: :smile:

gabrocheleau (approved):

acolytec3:
With great fear and trembling, I merge this. 😬

@holgerd77 would it be possible to do a beta release of this so we can see if it works for evm.codes? I could try the same with ganache if I can figure out how to build their monorepo from source.

holgerd77:
> With great fear and trembling, I merge this. 😬
> 
> @holgerd77 would it be possible to do a beta release of this so we can see if it works for evm.codes? I could try the same with ganache if I can figure out how to build their monorepo from source.

Great, Andrew (and everyone else), thanks for working through this! 

I will directly do a "normal" release. Ganache is already in a broken state if I am correct, so it can only get better from here upon. If this still do not work, we'll just do a follow-up release.

(weekend is also a reason, since I am not sure if I'll be able to work during the next 2 days)

PR #2371, opened 2022-10-19
Release Fixes (new Util release v8.0.1, new client release v0.6.5 (v0.6.4 broken))

holgerd77:
This PR fixes releases from #2361 by submitting a forgotten Util release v8.0.1 which caused releases from Blockchain upwards to be broken as well as doing a new Client release v0.6.5 since v0.6.4 has a hardcoded Util v8.0.0 dependency which wouldn't been updated (so v0.6.4 is also broken).

Open for review. 🙂

PR #2377, opened 2022-10-21
New Releases (EVM/VM Event Emitter Fix)

holgerd77:
These releases are fixes for the previous EVM v1.1.0 and VM v6.1.0 releases from #2361 and #2371 where the async event emitter package had been updated which turned out to be breaking in some parts of the functionality, see comments made here https://github.com/ethereumjs/ethereumjs-monorepo/pull/2303#issuecomment-1284377395.

The new releases switch back to a modernized version of the initially used `async-eventemitter` package, see #2376.

Open for review. 🙂

//cc @davidmurdoch @thevaizman


packages/evm/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ethereumjs/evm",
-  "version": "1.1.0",
+  "version": "1.2.0",
```
acolytec3, line 3: I guess I just noticed this but any reason we're using minor version updates for `evm` and point releases for the other packages?

acolytec3 (approved):
One question about version numbers but LGTM!

packages/evm/package.json
holgerd77: Yes. I already used a minor for the first release due to the significance of the event emitter update, for the VM as well, so to indicate that this would be a release with substantial changes (which was obviously still not enough 😋).

acolytec3: Fair.  Lets merge it!

davidmurdoch: Thanks for fixing this so quickly!

One note: for npm packages, package version numbers don't communicate significance (except version `1.0.0`) or provide any indication of how substantial the changes are. They indicate if the release contains fixes/refactors (0.0.x), new features (0.x.0), or breaking changes (x.0.0). Of course, if ethereumjs isn't adhering to [Semver](https://semver.org/) this isn't true, but we've been operating under the impression that it does.

holgerd77:
Just published the following releases on npm:

- `@ethereumjs/util` [v8.0.2](https://github.com/ethereumjs/ethereumjs-monorepo/releases/tag/%40ethereumjs%2Futil%408.0.2)
- `@ethereumjs/evm` [v1.2.0](https://github.com/ethereumjs/ethereumjs-monorepo/releases/tag/%40ethereumjs%2Fevm%401.2.0)
- `@ethereumjs/vm` [v6.2.0](https://github.com/ethereumjs/ethereumjs-monorepo/releases/tag/%40ethereumjs%2Fvm%406.2.0)

packages/evm/package.json
holgerd77: Hi there, yeah, Andrew did a really great job here! 🙂 👍 

I've just published all the versions, hope everything works now, otherwise let us know (respectively: some general feedback would likely be good in this case).

Yes, I know that (semver). We DO adhere to Semver. For the bugfix vs minor versions I perceive these lines as a bit blurry and not as important as for minor vs breaking. Feature additions are often so insignifcant for the user - if a new function is mainly for internal usage e.g., or if their is a slight API parameter change or so. If one would always call this out a minor version, one would very quickly land at v6.45. 😋

So I switched a bit (more) to a distinction to "what feels right to me" in these cases and I reserve minor versions more for the a bit more substantal feature additions.

Richard (from Ethers) is doing the opposite and is putting everything in bugfix releases. 😆

thevaizman:
@acolytec3  @holgerd77 - thanks a lot!
I can confirm that everything seems to function well with our app too (evm.codes).

Only annoying bit is that I still need to add the following line to my `package.json` to prevent the `blockchain` package from throwing errors due to it picking up the `8.0.0` by default (even though it has `^8.0.0` as an inner dependency):
```json
"resolutions": {
    "@ethereumjs/util": "8.0.2"
  }
```

But that is a minor hiccup, I really appreciate all the work on the event emitter! 🙌

holgerd77:
> @acolytec3 @holgerd77 - thanks a lot! I can confirm that everything seems to function well with our app too (evm.codes).
> 
> Only annoying bit is that I still need to add the following line to my `package.json` to prevent the `blockchain` package from throwing errors due to it picking up the `8.0.0` by default (even though it has `^8.0.0` as an inner dependency):
> 
> ```json
> "resolutions": {
>     "@ethereumjs/util": "8.0.2"
>   }
> ```
> 
> But that is a minor hiccup, I really appreciate all the work on the event emitter! 🙌

Yeah, sorry, this *will* get better on the next round of full-scale releases, I was honestly too lazy to do the full round of 13 releases and chose to limit to the core 3 libraries.

Super great to hear that things are working on the event emitter side!!! 🙏

We didn't have such a side-kicking bug/malfunction for quite some time, super-glad that this seemed to have worked out as expected/hoped, thanks again @acolytec3 for working through this.

PR #2382, opened 2022-10-26
Integrate txpool fixes found in Shandong

acolytec3:
- Fixes the `txPool.validate` method so it checks transactions against the correct stateroot when validating accounts
- Adds back missing `eth` protocol handlers for transaction gossip

### DIFF
--- a/packages/client/lib/net/protocol/ethprotocol.ts
+++ b/packages/client/lib/net/protocol/ethprotocol.ts
@@ -95,11 +95,7 @@ export class EthProtocol extends Protocol {
       encode: (txs: TypedTransaction[]) => {
         const serializedTxs = []
         for (const tx of txs) {
-          if (tx.type === 0) {
-            serializedTxs.push(tx.raw())
-          } else {
-            serializedTxs.push(tx.serialize())
-          }
+          serializedTxs.push(tx.serialize())
         }
         return serializedTxs
       },
@@ -197,10 +193,6 @@ export class EthProtocol extends Protocol {
     {
       name: 'NewPooledTransactionHashes',
       code: 0x08,
-      encode: ({ hashes }: { hashes: Buffer[] }) => [hashes],
-      decode: ([hashes]: [Buffer[]]) => {
-        hashes
-      },
     },
     {
       name: 'GetPooledTransactions',
--- a/packages/client/lib/net/protocol/ethprotocol.ts
+++ b/packages/client/lib/net/protocol/ethprotocol.ts
@@ -103,11 +103,13 @@ export class EthProtocol extends Protocol {
         }
         return serializedTxs
       },
-      decode: ([txs]: [Buffer[]]) => {
-        // TODO: add proper Common instance (problem: service not accessible)
-        //const common = this.config.chainCommon.copy()
-        //common.setHardforkByBlockNumber(this.config.syncTargetHeight, this.chain.headers.td)
-        return txs.map((txData) => TransactionFactory.fromBlockBodyData(txData))
+      decode: (txs: Buffer[]) => {
+        const common = this.config.chainCommon.copy()
+        common.setHardforkByBlockNumber(
+          this.config.syncTargetHeight ?? BigInt(0),
+          this.chain.headers.td
+        )
+        return txs.map((txData) => TransactionFactory.fromSerializedData(txData, { common }))
       },
     },
     {
@@ -192,6 +194,10 @@ export class EthProtocol extends Protocol {
     {
       name: 'NewPooledTransactionHashes',
       code: 0x08,
+      encode: ({ hashes }: { hashes: Buffer[] }) => [hashes],
+      decode: ([hashes]: [Buffer[]]) => {
+        hashes
+      },
     },
     {
       name: 'GetPooledTransactions',
--- a/packages/client/lib/net/protocol/ethprotocol.ts
+++ b/packages/client/lib/net/protocol/ethprotocol.ts
@@ -104,10 +104,13 @@ export class EthProtocol extends Protocol {
         return serializedTxs
       },
       decode: (txs: Buffer[]) => {
+        if (!this.config.synchronized) return
         const common = this.config.chainCommon.copy()
         common.setHardforkByBlockNumber(
-          this.config.syncTargetHeight ?? BigInt(0),
-          this.chain.headers.td
+          this.chain.headers.latest?.number ?? // Use latest header number if available OR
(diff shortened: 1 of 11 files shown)


packages/client/lib/net/protocol/ethprotocol.ts
```diff
@@ -103,11 +103,13 @@ export class EthProtocol extends Protocol {
...
         return serializedTxs
       },
-      decode: ([txs]: [Buffer[]]) => {
-        // TODO: add proper Common instance (problem: service not accessible)
-        //const common = this.config.chainCommon.copy()
-        //common.setHardforkByBlockNumber(this.config.syncTargetHeight, this.chain.headers.td)
-        return txs.map((txData) => TransactionFactory.fromBlockBodyData(txData))
+      decode: (txs: Buffer[]) => {
+        const common = this.config.chainCommon.copy()
+        common.setHardforkByBlockNumber(
```
g11tech, line 108: i don't think we need to set hardfork by number  here, chainCommon should always have updated hf as its updated whenever putblock is done by client

acolytec3: Don't we want to ensure we don't try to instantiate a transaction with a default hardfork that could be invalid for that transaction type (e.g. we don't instantiate a 1559 tx if the current hardfork is < london)?  In theory this should never happen but it seems like it's worth validating that when instantiating the transaction, right?

jochem-brouwer: We should definitely add in the hardfork by number. The reason is this: if we have a HF at height `A`, and the chain is at height `A-1`, then the txs should be validated against `A` (which is thus the HF). This could invalidate some txs, or actually validate other txs! (For instance if the HF creates a new Tx type!)

packages/client/lib/service/txpool.ts
```diff
@@ -282,6 +282,9 @@ export class TxPool {
         `Tx gaslimit of ${tx.gasLimit} exceeds block gas limit of ${block.gasLimit} (exceeds last block gas limit)`
       )
     }
+
+    // Set state root to latest block so that account balance is correct when doing balance check
+    await this.vm.stateManager.setStateRoot(block.stateRoot)
```
jochem-brouwer, line 287: Doesn't this add a race condition if we are running blocks at the same time? (Or is this a different VM than we use for running the blocks?)

jochem-brouwer (commented):
Two comments.

How did this turn op on Shandong?

acolytec3:
The one regarding the validate function I found when I was trying to add transactions to the txpool and they were being dropped. It turned out the account balance was getting compared to Genesis instead of the current stateroot so it was failing validation checks

packages/client/lib/service/txpool.ts
acolytec3: Looks like it's the same VM from looking at the txpool constructor.  I can copy it here and pull state that way since the underlying db should be the same.

packages/client/lib/service/txpool.ts
```diff
@@ -282,6 +282,10 @@ export class TxPool {
         `Tx gaslimit of ${tx.gasLimit} exceeds block gas limit of ${block.gasLimit} (exceeds last block gas limit)`
       )
     }
+
+    const vmCopy = await this.vm.copy()
```
g11tech, line 286: i don't think we need copy here, txpool's vm is already a copy and its only being used in validate where the stateroot is already being set

acolytec3: It's just using the execution module's VM as [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/0b569e67daf9275d5e34a6674ab7a3afb4117296/packages/client/lib/service/txpool.ts#L153) so I believe it's all the same instance across the whole client.

packages/client/lib/net/protocol/ethprotocol.ts
acolytec3: Hmm, I'm not really sure what to do here then.  We don't store the current height of the chain anywhere else other than this `syncTargetHeight` which isn't guaranteed to be defined and I can't call any async methods to retrieve the latest header here since these encode/decode helpers are synchronous.  I would think we would want to accept transactions as long as the txPool is running and just take our best guess at a block height based on the syncTarget (or just 0 since we don't know otherwise).  I could also just take whatever the current hardfork is set in common and use that if we don't know the block number maybe?

g11tech: this.headers.latest.number should be fine, but `chainCommon` should be set at the correct hf corresponding to that, thats my point, no need to set it further since copy will copy its set hardfork

packages/client/lib/service/txpool.ts
g11tech: ok, will review but feels like vmExecution is running on its own copy, but for the sake of the PR, we can copy vm in the constructor itself

packages/client/lib/net/protocol/ethprotocol.ts
holgerd77: Yeah, I would also think that a "best guess" should be appropriate here, since - technically - these txs do not have *any HF at all* yet since they are not bound to a block yet. So there is just no correct td (e.g.) at this point to assign.

So this should be ok (?) as long as it is assured that the common HF is eventually re-assigned/re-set when a tx then eventually make it in a block through the block builder?

packages/client/lib/service/txpool.ts
g11tech: ok, will review but feels like vmExecution is running on its own copy, but for the sake of the PR, we can copy vm in the constructor itself

packages/client/lib/net/protocol/ethprotocol.ts
g11tech: this could be ok if headers.td is paired with headers.latest.number else we might get td invalidation errors, which as per my understanding is the common already set in chainCommon.

packages/client/lib/service/txpool.ts
acolytec3: Wouldn't you end up operating on out of sync state manager if you do that since the txpool's state manager wouldn't get updated when the VM runs blocks?

g11tech: that won't be the issue as setStateRoot will set the correct one, however i do see that the VMs are shared between txpool and execution, but i am wondering why would this vm's state not being updated, but i am ok with the copy here :+1:

g11tech (approved):
lgtm!

PR #2383, opened 2022-10-26
EVM: non-empty "empty" account bugfixes

jochem-brouwer:
Thanks to @kchojn who found this bug. It was introduced by https://github.com/ethereumjs/ethereumjs-monorepo/pull/2054

The PR fixes 4 bugs in our lib.

This is the situation:

EOA Account `A` has balance `B`, nonce 0, and (of course) no code and no storage.
Account `A` now sends a Tx to contract `C` and it forwards ALL balance upfront to pay for gas. (So, at this point, account `A` has 0 balance as it starts running the contract).

Now JS has 3 bugs in these situations:
Calling `EXTCODEHASH` on account A will put 0 on the stack, not the empty code hash, because we report that account is empty.
`CALL` will charge `callNewAccount` gas when we forward nonzero amount from the contract back into the account
`SELFDESTRUCT` will do the same if we selfdestruct nonzero value to account A.

### DIFF
--- a/packages/evm/src/opcodes/functions.ts
+++ b/packages/evm/src/opcodes/functions.ts
@@ -513,15 +513,15 @@ export const handlers: Map<number, OpHandler> = new Map([
     async function (runState) {
       const addressBigInt = runState.stack.pop()
       const address = new Address(addressToBuffer(addressBigInt))
-      const empty = (await runState.eei.getAccount(address)).isEmpty()
-      if (empty) {
+      const account = await runState.eei.getAccount(address)
+      const empty = account.isEmpty()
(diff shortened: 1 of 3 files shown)


kchojn:
Awesome, I am glad that you helped to solve the problem so quickly! it looks like there are no more problems with it, gj

acolytec3 (approved):
LGTM

qbzzt:
Did you run https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stExtCodeHash/extcodehashEmptyFiller.yml? It should have detected this bug because after EOA A sends its balance it should have a non-zero nonce.

Issue #2391, opened 2022-10-27
Cleanup `handleEth`


acolytec3:
The [`handleEth`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/1e903db3f29f7b4f93db3b2c61248a3c1babad27/packages/client/lib/service/fullethereumservice.ts#L228) function is a mess of `if...else if...` that should be converted to use a switch statement based on the `message.name` to improve the clarity of this code.

PR #2392, opened 2022-10-28
Clean up `handleEth`

ScottyPoi:
Fix for issue #2391 
#### Clean up `handleEth`

> The [handleEth](https://github.com/ethereumjs/ethereumjs-monorepo/blob/1e903db3f29f7b4f93db3b2c61248a3c1babad27/packages/client/lib/service/fullethereumservice.ts#L228) function is a mess of if...else if... that should be converted to use a switch statement based on the message.name to improve the clarity of this code.

### DIFF
--- a/packages/client/lib/service/fullethereumservice.ts
+++ b/packages/client/lib/service/fullethereumservice.ts
@@ -313,8 +313,6 @@ export class FullEthereumService extends EthereumService {
         peer.eth?.send('Receipts', { reqId, receipts })
         break
       }
-      default:
-        return
     }
   }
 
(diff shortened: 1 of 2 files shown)


ScottyPoi:
Unclear on why this broke the test double tests
```
not ok 12 Error: Unsatisfied verification on test double. Wanted: - called with `({}, undefined)`. But there were no invocations of the test double.
  ---
    operator: error
    stack: |-
      Error: Unsatisfied verification on test double.
      
        Wanted:
          - called with `({}, undefined)`.
      
        But there were no invocations of the test double.
```

acolytec3 (commented):
This has gone from hideous to beautiful.  Thanks for taking this on!  I've added a couple of tests for some of the message handlers to bump coverage.  I'm surprised we hadn't added these bfore so good that we're doing it here.  Will approve if the coverage check passes (or will add some more).

PR #2394, opened 2022-10-29
client: Multifetcher synchronizer support

### LINKED ISSUE #2389: Modify Synchronizer class to support multi-fetcher syncs
Currently, in order to run a fetcher, the `fetcher` field of a `Synchronizer` object has to be set to a `Fetcher` object that has been initialized. This creates a one-to-one relationship between the `Synchronizer` and `Fetcher` classes. For snap sync, it is necessary to create and run many fetchers - specifically, four (storage, account, bytecode, and trie node). Modify the `Synchronizer` class to accept an array of fetchers. It can call the `fetch` method of each, and resolve only when all fetchers are done fetching.

Also see #2345

scorbajio:
Fixes #2389.

### DIFF
--- a/packages/client/lib/sync/beaconsync.ts
+++ b/packages/client/lib/sync/beaconsync.ts
@@ -46,14 +46,27 @@ export class BeaconSynchronizer extends Synchronizer {
   }
 
   get fetcher(): ReverseBlockFetcher | null {
-    if (this._fetcher !== null && !(this._fetcher instanceof ReverseBlockFetcher)) {
+    if (
+      this._fetchers !== null &&
+      this._fetchers.length > 0 &&
+      !(this._fetchers[0] instanceof ReverseBlockFetcher)
+    ) {
       throw Error(`Invalid Fetcher, expected ReverseBlockFetcher`)
     }
-    return this._fetcher
+    if (this._fetchers !== null) {
+      return this._fetchers![0] as ReverseBlockFetcher
+    }
+    return null
   }
 
   set fetcher(fetcher: ReverseBlockFetcher | null) {
-    this._fetcher = fetcher
+    if (fetcher !== null) {
+      if (this._fetchers != null) {
+        this._fetchers.push(fetcher)
+      } else {
+        this._fetchers = [fetcher]
+      }
+    }
   }
 
   /**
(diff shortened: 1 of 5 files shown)


g11tech (changes requested):
I would not extend the fetcher to be fetches, but the better solution will be to have snap's own fetcher that might span their own fetchers, and idealy that should be handled in job creation and not multiple fechers.

scorbajio:
> I would not extend the fetcher to be fetches, but the better solution will be to have snap's own fetcher that might span their own fetchers, and idealy that should be handled in job creation and not multiple fechers.

I don't understand the part where you say "that should be handled in job creation and not multiple fetchers." Can you give a full explanation of the alternative solution that you are proposing and explain what makes it better?

holgerd77:
> I would not extend the fetcher to be fetches, but the better solution will be to have snap's own fetcher that might span their own fetchers, and idealy that should be handled in job creation and not multiple fechers.

@g11tech Yeah, I would agree with @scorbajio here, could you please give a reasoning for the alternative design suggestion? What would be the shortcomings respectively what are the strucural advantages of the proposed alternative? 🙂

g11tech:
> > I would not extend the fetcher to be fetches, but the better solution will be to have snap's own fetcher that might span their own fetchers, and idealy that should be handled in job creation and not multiple fechers.
> 
> @g11tech Yeah, I would agree with @scorbajio here, could you please give a reasoning for the alternative design suggestion? What would be the shortcomings respectively what are the strucural advantages of the proposed alternative? slightly_smiling_face

There should be an "umberalla" snapfetcher that houses all the fetchers related to snap, current PR really introduces a  possible dangerous mix of fetchers from across synchronizers and might really introduce hard to find bugs

Also the contribution of this PR is to add fetchers array without adding how they will be used in snap. I am not conviced of this design but @scorbajio can certainly use it to prototype. This is however not merge-able/worthy  in the current state (in my opinion).

When a client actually runs, it becomes very hard to reason why something is happening as sync is primarily very hard in my opinion. So unless there is a very convincing reason this is how it should be done, it would be better to not have array over here (and also not an unbounded array inside the snapfetcher)

holgerd77:
What is the state of this?

g11tech:
closing this in favour of the nested fetcher designed as merged in https://github.com/ethereumjs/ethereumjs-monorepo/pull/2345

Issue #2359, opened 2022-10-14
Implement `eth_feeHistory` RPC method


acolytec3:
We should implement the `eth_feeHistory` RPC method so that our client will be fully compatible with the testnet faucets which use `feeHistory` to construct their transactions.  The [Alchemy documentation is up to date](https://docs.alchemy.com/reference/eth-feehistory) so don't rely on the Ethereum.org RPC spec as it doesn't include the `GasUsedRatio` property in the response.

A stub implementation was included in #2316 to allow the Shandong testnet to proceed.

scorbajio:
Does `eth_feeHistory` still need to be implemented? This ticket hasn't been updated in a while, is there any new context around this issue?

acolytec3:
I'd say it's still open. We don't have an inventory need for it but always good to fill out missing rpc endpoints.

May 30, 2023 7:48:15 PM Scorbajio ***@***.***>:

holgerd77:
Closed by #3295

PR #2396, opened 2022-11-01
Implementation of 'eth_gasPrice'

rodrigoherrerai:
Hi! 

This implements 'eth_gasPrice'. 

As discussed in issue #1114 , the implementation consists of the following: 

If the chain is not 1559 compatible, it iterates over the last 20 blocks with a limit of 500 transactions. It then gets the average price of those transactions. 

If the chain is 1559 compatible, it iterates over the last block to get the average priority fee and adds it to the next base fee.


packages/client/lib/rpc/modules/eth.ts
```diff
@@ -986,4 +988,56 @@ export class Eth {
...
+          const maxPriorityFeePerGas = (tx as FeeMarketEIP1559Transaction).maxPriorityFeePerGas
+          priorityFee += maxPriorityFeePerGas
+        }
+      }
+      priorityFee =
+        priorityFee !== BigInt(0) ? priorityFee / BigInt(blockTransactionsLength) : BigInt(1)
+      gasPrice = baseFee + priorityFee
+    } else {
+      // For chains that don't support EIP-1559 we iterate over the last 20
+      // blocks to get an average gas price. We cap the tx lookup to 500.
```
acolytec3, line 1017:
```suggestion
      // blocks to get an average gas price. We cap the tx lookup per block to 500.
```
Not sure if you're actually intending to cap the transaction look up in the block or total transactions examined.  If you're intending to do it as looking at a total number of transactions, you'll need to move the `break` in line 1030 after 1031.  Otherwise, it will just break from the inner loop and not the outer one.

acolytec3 (changes requested):
This looks great so far (left two specific comments on things I think need to be revised).

Also, can you add a test that runs 20 blocks and computes the gasPrice?  Maybe run 21 blocks and validate the computation produces an average of only the last 20 (maybe put some astronomical gas fee in the first block that moves the average up and then make the gas fee just like 20 or something in the following 20) and then validate that the price returned by the RPC call is 20 and not 20 + some large number divided/# txns?

rodrigoherrerai:
> 

Added the tests for 21 blocks !

packages/client/lib/rpc/modules/eth.ts
jochem-brouwer: @acolytec3 capping it per block makes no sense (unless I'm in the wrong here), we can get the entire block body from the LevelDB immediately so just getting the avg gas price should be super fast once we got the block from the DB. (This would ofc. be different if we would have to get tx-by-hash each time)

packages/client/lib/rpc/modules/eth.ts
```diff
@@ -986,4 +988,55 @@ export class Eth {
...
+        const block = await this._chain.getBlock(latest.number - BigInt(i))
+        if (block.transactions.length === 0) {
+          continue
+        }
+
+        for (const tx of block.transactions) {
+          const txGasPrice = (tx as Transaction).gasPrice
+          gasPrice += txGasPrice
+          txCount++
+          if (txCount >= 500) break
```
jochem-brouwer, line 1028: I really do not think we should do a per-block check, and I also do not think we need a tx cap limit. I might be wrong here, but the heaviest operation here is pulling the block body from the DB. Iterating over the txs should be much faster than that.

jochem-brouwer (commented):
I got some comments, but I am really happy to see that this one is being picked up on, thanks a lot! :smile:

packages/client/lib/rpc/modules/eth.ts
rodrigoherrerai: Yes this makes sense. I removed the tx cap.

packages/client/lib/rpc/modules/eth.ts
```diff
@@ -986,4 +988,53 @@ export class Eth {
...
+
+  /**
+   * Gas price oracle.
+   *
+   * Returns a suggested gas price.
+   * @returns a hex code of an integer representing the suggested gas price in wei.
+   */
+  async gasPrice() {
+    let gasPrice = BigInt(0)
+    const latest = await this._chain.getCanonicalHeadHeader()
```
holgerd77, line 1000: I generally wonder if we want to make this dependent from the `synchronized` status of the client `this.client.config.synchronized`? 🤔 it likely doesn't make sense to return anything (respectively: is not useful) if the client is still syncing and we are not calculating upon the latest blocks?

Not sure, maybe to test this in Geth what is responded in such a case?

Also, then: any reason to not also take the chain common object here for the 1559 check (below)?

holgerd77 (commented):
Short comment

packages/client/lib/rpc/modules/eth.ts
rodrigoherrerai: 1. Also, then: any reason to not also take the chain common object here for the 1559 check (below)?

->Made a change where the min gas price from the chain common object is now checked on both (1559 and non 1559) chains.

acolytec3: >Not sure, maybe to test this in Geth what is responded in such a case?

I checked this in geth and it does return a gas price even when syncing.  Not sure where that gas price comes from but it's producing something.  The spec doesn't give any specifics about how to calculate current gas price so I guess I would just we go with compute best gas price available based on the current view of the chain that we have (i.e. what is implemented here) and leave it to the consumer to validate the sync status of the client via the `eth_syncing` when evaluating gas price.

packages/client/lib/rpc/modules/eth.ts
```diff
@@ -1005,4 +1007,54 @@ export class Eth {
...
+
+        for (const tx of block.transactions) {
+          const txGasPrice = (tx as Transaction).gasPrice
+          gasPrice += txGasPrice
+          txCount++
+        }
+      }
+
+      if (txCount > 0) {
+        gasPrice = gasPrice / txCount > minGasPrice ? gasPrice / txCount : minGasPrice
```
acolytec3, line 1052:
```suggestion
        const avgGasPrice = gasPrice / txCount
        gasPrice = avgGasPrice > minGasPrice ? avgGasPrice : minGasPrice
```

holgerd77 (approved):
Ok, this looks good now! 🙂 Thanks a lot @rodrigoherrerai for the additional iterations and the contribution in general! 💯 🙏

Will merge.

holgerd77:
Rebased this via UI.

rodrigoherrerai:
> Ok, this looks good now! 🙂 Thanks a lot @rodrigoherrerai for the additional iterations and the contribution in general! 💯 🙏
> 
> Will merge.

Great ! Is there any other are that you guys need help ? Would love to contribute more if possible :)

acolytec3:
> > Ok, this looks good now! slightly_smiling_face Thanks a lot @rodrigoherrerai for the additional iterations and the contribution in general! 100 pray
> > Will merge.
> 
> Great ! Is there any other are that you guys need help ? Would love to contribute more if possible :)

Definitely!  If you're looking for another RPC endpoint to implement, the `eth_feeHistory` could be an interesting one.  I opened an issue with initial comments/a proof of concept implementation (very incomplete) in #2359.  If you're looking for something else, feel free to join our discord and we can discuss further on what makes the most sense!

rodrigoherrerai:
> > > Ok, this looks good now! slightly_smiling_face Thanks a lot @rodrigoherrerai for the additional iterations and the contribution in general! 100 pray
> > > Will merge.
> > 
> > 
> > Great ! Is there any other are that you guys need help ? Would love to contribute more if possible :)
> 
> Definitely! If you're looking for another RPC endpoint to implement, the `eth_feeHistory` could be an interesting one. I opened an issue with initial comments/a proof of concept implementation (very incomplete) in #2359. If you're looking for something else, feel free to join our discord and we can discuss further on what makes the most sense!

Perf! Can you provide the link to the discord ?

acolytec3:
https://discord.gg/aT78qk3u

PR #2404, opened 2022-11-04
Fix nonce problem

### LINKED ISSUE #2399: VM updates nonce after running a transaction, but should do right before executing it
There are three phases of a transaction:

1. The initialization phase. In this phase, the tx is unpacked and it is being read, and the execution environment is being set up. (I.e. we create the right parameters to pass on to `EVM` when running a transaction in `runTx`)
2. The execution phase. In this phase, the transaction is executed: if a contract is called, we run the code.
3. The final phase. In this phase, coinbase is credited and any gas which was not used is refunded to the caller.

Right between phase 1 and 2, the nonce of the account should be bumped. If a EOA account which does not have transactions (nonce is 0) tries to create a contract, then the parameters passed on to generate the contract address will use nonce 0 (not 1). However, when we now enter the execution frame, the nonce should be 1, not zero.

Right now this is a problem: we update the nonce after the account. This led to a consensus bug where, if the nonce is zero and the account sends all funds to a contract, the EVM now thinks that the account is empty (it has no code, no storage, no balance, and nonce is zero). This leads to problems in (1) `EXTCODEHASH` thinks the account is empty and thus pushes 0 on the stack, not the hash of the empty string, (2) `CALL`ing the account will lead to `callNewAccount` gas being charged if nonzero value is sent and (3) the same for `SELFDESTRUCT` which charges `callNewAccount` when the contract has nonzero balance.

These bugs are fixed in https://github.com/ethereumjs/ethereumjs-monorepo/pull/2397, but the method to fix those bugs could be said to be a bit on the hacky-ish side.

We have had a lengthy discussion about this nonce problem in the past when we separated VM and EVM. It is thus not trivial to fix. But, we should ensure that the nonce is updated right before we enter the EVM execution: otherwise, this will haunt us in the future.

jochem-brouwer:
Closes #2399

Do not merge, as I just realized when creating this PR why we had this nonce stuff change in the first place and why we had lengthy discussions. Discussion goes into #2399

Note: this rollbacks the changes made to EVM/VM source of #2383 (I consider this rather hacky), but it keeps the tests of that PR.


jochem-brouwer:
OK, for some reason the CI goes out-of-memory on ShanghaiLove test, but only on Homestead fork. It is not clear to me why any changes to this nonce logic should influence this, but OFC this has to be solved.

acolytec3 (approved):
This all looks good as far as it goes but is it going to have unintended knock-on effects when we start really using the EVM as a standalone execution environment?

PR #2405, opened 2022-11-04
EVM optimizations

jochem-brouwer:
This PR is aimed towards optimizing EVM. In particular, see: https://github.com/NomicFoundation/hardhat/issues/3316#issuecomment-1304064298

When updating the EVM using chunk-sized extensions (so do not extend it with 32 bytes each time, but instead of containers of 8192 / 8 kb), MSTORE gets about 8 times faster (based upon above problem ^). It also almost halves the runtime of aboves' problem.

Still WIP because I might want to introduce more optimizations (plus this might cause a bunch of test errors).


packages/evm/src/memory.ts
```diff
@@ -30,7 +32,10 @@ export class Memory {
     const newSize = ceil(offset + size, 32)
     const sizeDiff = newSize - this._store.length
     if (sizeDiff > 0) {
-      this._store = Buffer.concat([this._store, Buffer.alloc(sizeDiff)])
+      this._store = Buffer.concat([
+        this._store,
+        Buffer.alloc(Math.ceil(sizeDiff / CONTAINER_SIZE) * CONTAINER_SIZE),
+      ])
```
holgerd77, line 38: Ok, just did a quick search `._store.length` for the EVM to see if there are any external accesses to memory length where we would have side effects, but this seems not to be the case (apart from some memory-internal tests which can eventually be adopted if needed). 🙂

So this seems like a really clever solution to reduce the buffer allocations drastically along memory extensions! 👍

(still am a bit cautious and wonder though if this might have some side effect(s) *somewhere*? 🤔

jochem-brouwer: I agree that we should very carefully examine this. (I also did the same search as you did since I was afraid that there indeed might be external reads to it)

axic:
@jochem-brouwer for speed benchmarks you could try to plug in https://github.com/ipsilon/evm-benchmarks (it uses the general state tests format). If you do please let us know the results (or just tag me in the issue/PR).

acolytec3 (approved):
LGTM!

PR #2316, opened 2022-09-27
pre-shanghai: Eof testnet setup

g11tech:
### Eof sim setup

- https://github.com/ethereumjs/ethereumjs-monorepo/issues/2298

This sim test is to setup a single ethereumjs<>lodestar instance to allow executing testvectors for the EIPs targetting Shanghai hardfork.

### EIP(s) testing

Target EIPs are: `3540,3651,3670,3855,3860`

### Prerequisite(s)

- Docker since lodestar docker image is used to run CL to drive the post-merge sim run
- Docker should run without `sudo`, else prefix `sudo` in from of docker commands in `test/sim/single-run.sh`
- Mac users should skip `docker` and install directly install and use lodestar binary as instructed below

### How to run

1. `npm install` the `ethereumjs-monorepo`
2. `cd packages/client`
3. `docker pull chainsafe/lodestar:latest`, for MAC users: `npm install -g @chainsafe/lodestar:latest`
4. Install linux package `jq` is not installed
5. Create a data directory `data/eof` (or any other place convinient for you).

There are three ways to run tests with varying degree of automation:

#### Fully automated sim run

Just run:

```
DATADIR=/data/eof npm run tape -- test/sim/eof.spec.ts
```

for MAC add extra env variable `LODE_BINARY="lodestar"` (or `LODE_BINARY="npx lodestar"` if you didn't install lodestar globally) , i.e.

```
DATADIR=/data/eof LODE_BINARY="lodestar" npm run tape -- test/sim/eof.spec.ts
```

It will auto run the script to run custom network, run tests and in end tear it down. You can modify `test/sim/eof.spec.ts` transaction scenarios to test your custom functionality.

#### Semi-Auto run using script to start custom network

Currently you can just start the local instance by
`DATADIR=data/eof test/sim/./single-run.sh`

MAC users:
`DATADIR=data/eof LODE_BINARY="lodestar" test/sim/./single-run.sh`

This command run should start both ethereumjs and lodestar in terminal. Soon you should see lodestar driving ethereumjs in PoS configuration.

Once the network looks synced to you, you can run tests:

```
EXTERNAL_RUN=true npm run tape -- test/sim/eof.spec.ts
```

#### Manual run to setup network

1. Clean DATADIR (`rm -rf data/eof/*`) in start of each run and do `mkdir data/eof/ethereumjs && mkdir data/eof/lodestar`
[...]

PR #2407, opened 2022-11-05
client: Fix enode to ip4 and write the same to disk

g11tech:
Extracted from shandong

One of the issues while peering ethereumjs to iself was, enodes we created were ipv6 but we only listen on ipv4 interface in devp2p hence peers were connecting.

This PR fix that, also adds validation for ip6 user ip input, plus write the address (rlxp + libp2p) onto disk for easy fetch and use


packages/client/lib/util/parse.ts
```diff
@@ -80,3 +88,8 @@ export function parseKey(input: string | Buffer) {
   }
   return Buffer.from(input, 'hex')
 }
+
+// From: https://community.helpsystems.com/forums/intermapper/miscellaneous-topics/5acc4fcf-fa83-e511-80cf-0050568460e4
```
acolytec3, line 92: That is some intense regex.  I'm glad I didn't have to write it.

packages/client/lib/client.ts
```diff
@@ -110,6 +112,14 @@ export class EthereumClient {
       this.config.logger.warn(`Server error: ${error.name} - ${error.message}`)
     })
     this.config.events.on(Event.SERVER_LISTENING, (details) => {
+      const networkDir = this.config.getNetworkDirectory()
+      // Write the transport into a file
+      try {
+        writeFileSync(`${networkDir}/${details.transport}`, details.url)
```
acolytec3, line 118: Do we have an actual use case for this?  Feels like we're just arbitrarily writing something to disk and not using it anywhere.

acolytec3 (commented):
Added one test for coverage and left one question we should think through before we merge but looks good otherwise!

packages/client/lib/client.ts
g11tech: yes, i am using it to do a multi peer setup in shandong with script reading from file, as well as picking it up  to populate enodes

packages/client/lib/util/parse.ts
g11tech: ha ha, thats what i felt while i copied and tested it

holgerd77: Short nit: this URL is forwarding to https://community.fortra.com/forums/intermapper/miscellaneous-topics/5acc4fcf-fa83-e511-80cf-0050568460e4 now (maybe they rebranded in the last few days?).

Maybe let's directly update the URL in case the old domain name goes over to some other (malicious) usage at some point (likely not, but who knows? 🙂).

holgerd77: Anyhow, really a super-nitty nit. 😎

holgerd77:
Rebased this via UI.

PR #2408, opened 2022-11-05
client: Fix skeleton reset scenario when head announced before subchain 0 tail

g11tech:
Extracted from shandong:

In the scenario for e.g. when the CL might start from genesis it will signal head which might be before subchain 0's tail. This fix appropriately rests the chain to new head declaration


holgerd77 (approved):
I tried to understand what happens here and I think this LGTM! 👍

Issue #2402, opened 2022-11-04
Eth/66 request ID `0` RLP encoded improperly


shemnon:
For an Eth/66 response where the sender sends a request ID of `0` the response Id is being encoded in an invalid scalar.

```
0xf901fe00f901faf901f7a021d614c165442d4bd5097f825ac857f31f2df54a8cc1e0b3d8d4d609f0cc3ae8a01dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d4934794f97e180c050e5ab072211ad2c213eb5aee4df134a095f5929c80fd2a875796d2a10af2a6f4ba9913a791efa98d113bcf7121f3ff30a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421b90100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008082d21a83400000808463645fa080a0a7144184e31164c7f806a1eec4160c54ca4d541955dbf4657ae01a955cef279788000000000000000007
```

Is an in-the-wild response.  The fourth byte is `00` which per RLP is an invalid leading character for a scalar value, which the request ID of Eth/66 is defined to be.  The [correct encoding of an RLP zero](https://github.com/ethereum/tests/blob/develop/RLPTests/rlptest.json#L34) is the empty byte string: `80`.

The above should read:

```
[...]

PR #2409, opened 2022-11-05
client: Use unpadded int/bigint to buffer in net protocols

g11tech:
Extracted and a bit enhanced from shandong 

In the devp2p protocol, integer 0 is to be coded as `0x80` and not as buffered `0x00`

Closes: #2402

### DIFF
--- a/packages/client/lib/net/protocol/snapprotocol.ts
+++ b/packages/client/lib/net/protocol/snapprotocol.ts
@@ -1,7 +1,7 @@
 import {
   accountBodyFromSlim,
   accountBodyToSlim,
-  devP2PBigIntToBuffer as bigIntToBuffer,
+  bigIntToUnpaddedBuffer,
   bufferToBigInt,
   setLengthLeft,
 } from '@ethereumjs/util'
@@ -97,11 +97,11 @@ export class SnapProtocol extends Protocol {
       // [reqID: P, rootHash: B_32, startingHash: B_32, limitHash: B_32, responseBytes: P]
       encode: ({ reqId, root, origin, limit, bytes }: GetAccountRangeOpts) => {
         return [
-          bigIntToBuffer(reqId ?? ++this.nextReqId),
+          bigIntToUnpaddedBuffer(reqId ?? ++this.nextReqId),
           setLengthLeft(root, 32),
           setLengthLeft(origin, 32),
           setLengthLeft(limit, 32),
-          bigIntToBuffer(bytes),
+          bigIntToUnpaddedBuffer(bytes),
         ]
       },
       decode: ([reqId, root, origin, limit, bytes]: any) => {
@@ -128,7 +128,7 @@ export class SnapProtocol extends Protocol {
         proof: Buffer[]
       }) => {
         return [
-          bigIntToBuffer(reqId ?? ++this.nextReqId),
+          bigIntToUnpaddedBuffer(reqId ?? ++this.nextReqId),
           accounts.map((account) => [
             setLengthLeft(account.hash, 32),
             accountBodyToSlim(account.body),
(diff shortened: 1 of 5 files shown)


packages/util/src/bytes.ts
```diff
@@ -225,6 +226,19 @@ export function bigIntToBuffer(num: bigint) {
...
 
+/**
+ * Converts a {@link bigint} to a {@link Buffer} for devp2p rlp encodings
+ * excluding `0` as rlp has its own special encoding for the same
+ * see: https://github.com/ethereumjs/ethereumjs-monorepo/issues/2402
+ */
+export function devP2PBigIntToBuffer(input: bigint | number): BufferLike {
+  const inputNum = BigInt(input)
+  if (inputNum === 0n) {
+    return 0
```
jochem-brouwer, line 237:
```suggestion
    return Buffer.alloc(0)
```

packages/util/src/bytes.ts
```diff
@@ -225,6 +226,19 @@ export function bigIntToBuffer(num: bigint) {
   return toBuffer('0x' + num.toString(16))
 }
 
+/**
+ * Converts a {@link bigint} to a {@link Buffer} for devp2p rlp encodings
+ * excluding `0` as rlp has its own special encoding for the same
+ * see: https://github.com/ethereumjs/ethereumjs-monorepo/issues/2402
+ */
+export function devP2PBigIntToBuffer(input: bigint | number): BufferLike {
```
jochem-brouwer, line 234:
```suggestion
export function devP2PBigIntToBuffer(input: bigint | number): Buffer {
```

packages/util/src/bytes.ts
```diff
@@ -225,6 +226,19 @@ export function bigIntToBuffer(num: bigint) {
...
+  const inputNum = BigInt(input)
+  if (inputNum === 0n) {
+    return 0
+  } else {
+    return bigIntToBuffer(inputNum)
```
jochem-brouwer, line 239: I find it weird that `bigIntToBuffer` of `BigInt(0)` returns `<Buffer 00>`, it should IMO return the empty Buffer? Why did we choose to return this zero byte buffer?

packages/util/src/bytes.ts
jochem-brouwer: Now it always returns a Buffer type.

Note: `RLP.encode(Buffer.alloc(0))` = `<Buffer 80>` (as we expect)

jochem-brouwer (changes requested):

packages/util/src/bytes.ts
g11tech: i could still find normal bigIntToBuffer handy in snap sync's jobs where origin we map from bigint ranges to buffers but yea usage is very limited.

packages/util/src/bytes.ts
```diff
@@ -2,6 +2,7 @@ import { assertIsArray, assertIsBuffer, assertIsHexString } from './helpers'
 import { isHexPrefixed, isHexString, padToEven, stripHexPrefix } from './internal'
 
 import type {
+  BufferLike,
```
g11tech, line 5:
```suggestion
```

packages/client/lib/net/protocol/snapprotocol.ts
```diff
@@ -1,7 +1,7 @@
 import {
   accountBodyFromSlim,
   accountBodyToSlim,
-  bigIntToBuffer,
+  devP2PBigIntToBuffer as bigIntToBuffer,
```
holgerd77, line 4: See my other comment, we can use the existing `bigIntToUnpaddedBuffer` function for these cases.

I would also suggest to *not* rename but to use plainly, then it gets more explicit what conversion takes place here, renaming could easily mislead to think that plain `bigIntToBuffer` is used, which is also an existing function in `Util`.

holgerd77 (changes requested):
See other comments

packages/util/src/bytes.ts
```diff
@@ -225,6 +225,19 @@ export function bigIntToBuffer(num: bigint) {
...
 }
 
+/**
+ * Converts a {@link bigint} to a {@link Buffer} for devp2p rlp encodings
+ * excluding `0` as rlp has its own special encoding for the same
+ * see: https://github.com/ethereumjs/ethereumjs-monorepo/issues/2402
+ */
+export function devP2PBigIntToBuffer(input: bigint | number): Buffer {
+  const inputNum = BigInt(input)
+  if (inputNum === 0n) {
```
holgerd77, line 235: Just a reminder here (not directly relevant any more): we have decided on a policy to never use the *n notation for now for BigInt but always use `BigInt(*)` for compatibility reasons.

packages/util/src/bytes.ts
g11tech: awesome will swap out the fn

packages/util/src/bytes.ts
```diff
@@ -386,3 +386,7 @@ export const bigIntToHex = (num: bigint) => {
 export function bigIntToUnpaddedBuffer(value: bigint): Buffer {
   return unpadBuffer(bigIntToBuffer(value))
 }
+
```
holgerd77, line 389: Won't make this a blocker but generally please always document methods which become part of the public API.

Have added to the tiny-fixes-issue #2105 so that we won't forget.

holgerd77 (approved):
LGTM, thanks a lot! 👍 🙂

PR #2410, opened 2022-11-05
Add txpool_content rpc method for pending txs fetch

g11tech:
Extracted from shandong

developed by @acolytec3 

This allows blockscout to inspect txpool and fetch the pending transactions to list in explorer

TODO:
- [x] add the spec test


packages/client/lib/rpc/modules/txpool.ts
```diff
@@ -0,0 +1,43 @@
...
+  /**
+   * Returns the contents of the transaction pool
+   * @param params An empty array
+   */
+  content(_params = []) {
+    const pending = new Map()
+    for (const pool of this._txpool.pool) {
+      const pendingForAcct = new Map<bigint, any>()
+      for (const tx of pool[1]) {
+        pendingForAcct.set(tx.tx.nonce, tx.tx.toJSON())
```
jochem-brouwer, line 35: Can't we just `pool[1].map((e) => e.tx.toJSON()`?

acolytec3:
I've fixed a missing export so the new namespace is now exposed and have added a test.  Should be pretty close to ready at this point.

packages/client/lib/rpc/modules/txpool.ts
acolytec3: Yes, yes we can.  :-)  Good catch.

acolytec3:
Looks like it breaks something in the client tests though.  I can take a further look tomorrow as needed.

acolytec3:
The fix in almost all cases on the failing client tests is adding an `includeVM:true` to the client constructor in the rpc tests.  I can take a stab at this tomorrow.

acolytec3:
@holgerd77 Any opinions on this commit?  602f8b6903ae0c70b722304ca8cbc69a642fff1d  I don't know the history here as to why we don't always instantiate the VM when we're doing this client setup for the rpc tests.  The idea of this commit is to include the VM by default since otherwise I have to make tons of references to `createClient({ includeVM: true})`.  We need this because the `txpool_content` RPC class requires a VM in the constructor.  We can probably take different approaches if if desired bso don't want to merge this without further discussion.

holgerd77:
> @holgerd77 Any opinions on this commit? [602f8b6](https://github.com/ethereumjs/ethereumjs-monorepo/commit/602f8b6903ae0c70b722304ca8cbc69a642fff1d) I don't know the history here as to why we don't always instantiate the VM when we're doing this client setup for the rpc tests. The idea of this commit is to include the VM by default since otherwise I have to make tons of references to `createClient({ includeVM: true})`. We need this because the `txpool_content` RPC class requires a VM in the constructor. We can probably take different approaches if if desired bso don't want to merge this without further discussion.

Hmm, I just looked a bit in the RPC test `helper.ts` file where this gets applied: I have to say I also can't recall the complete history and reasoning for this. I would say since this is such an undocumented switch we can't easily find out anymore anyhow. So we should just do the switch (as you suggested) and see what happens. Might have some side effects on other tests (maybe we wanted to test in some other parts without VM?), but I have absolutely no idea or certainty here.

So, try out, rather make sure to *then* run the tests early on completely. Then we'll see.

acolytec3:
> So, try out, rather make sure to _then_ run the tests early on completely. Then we'll see.

Done.  And, all the tests pass.  This feels merge ready unless @g11tech or @holgerd77 or someone else has another opinion.

> Hmm, I just looked a bit in the RPC test helper.ts file where this gets applied: I have to say I also can't recall the complete history and reasoning for this. I would say since this is such an undocumented switch we can't easily find out anymore anyhow. So we should just do the switch (as you suggested) and see what happens. Might have some side effects on other tests (maybe we wanted to test in some other parts without VM?), but I have absolutely no idea or certainty here.

I may follow this up tomorrow or the next day with a PR to clarify all of those `createClient` opts as an options dictionary so we know what we're looking at.

acolytec3:
I tested this locally against blockscout and discovered we'd missed a couple of things from the Shandong branch so pending txns weren't showing up.  I've brought those over from Shandong and tested with blockscout and things are now working correctly.  This should be ready for final review.

g11tech (commented):
looks pretty neat! @acolytec3 :heart:  you will have to approve/merge!

acolytec3 (approved):
Approving my own work but yay for this!

PR #2412, opened 2022-11-05
client: Fix sendTransactions peer loop and enchance txpool logs 

g11tech:
- Fix sendTransactions peer loop
- Enhance the txpool stat logs to give an indication of the activity happening on txpool with success/failure activity of add, broadcast and block selection logs

### DIFF
--- a/packages/client/lib/service/txpool.ts
+++ b/packages/client/lib/service/txpool.ts
@@ -34,16 +34,19 @@ type TxPoolObject = {
   tx: TypedTransaction
   hash: UnprefixedHash
   added: number
+  error?: Error
 }
 
 type HandledObject = {
   address: UnprefixedAddress
   added: number
+  error?: Error
 }
 
 type SentObject = {
   hash: UnprefixedHash
   added: number
+  error?: Error
 }
 
 type UnprefixedAddress = string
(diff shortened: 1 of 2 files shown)


packages/client/lib/service/txpool.ts
```diff
@@ -718,8 +751,41 @@ export class TxPool {
   }
 
   _logPoolStats() {
+    let broadcasts = 0
+    let broadcasterrors = 0
+    let knownpeers = 0
+    for (const sendobjects of this.knownByPeer.values()) {
```
acolytec3, line 757: Should we contepmplate putting this whole method behind a debug flag or something?  I'm just wondering since this is a sort of a lot of iterating over arrays just report broadcast stats.  Maybe at a minimum, we could add a check [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/6c429278713ab6447b98ac236e02854de32f131e/packages/client/lib/service/txpool.ts#L190) to only turn this interval on if the loglevel is set to `info` or higher?  These logs kinda feel superfluous and I wouldn't want to run them by default.  So, we could maybe even turn the log level for these messages to debug?  I dunno, just a thought.

packages/client/lib/service/txpool.ts
```diff
@@ -438,8 +450,23 @@ export class TxPool {
...
+        const newHashesHex = newHashes.map((txHash) => txHash.toString('hex'))
+        const nexTxs = txs.filter((tx) => newHashesHex.includes(tx.hash().toString('hex')))
+        peer.eth?.request('Transactions', nexTxs).catch((e) => {
+          this.markFailedSends(peer, newHashes, e as Error)
+        })
+      }
+    }
+  }
+
+  private markFailedSends(peer: Peer, failedHashes: Buffer[], e: Error): void {
```
acolytec3, line 463: Is it hard to add a test for this function?  I can dig into it maybe on Monday but feels like we should add coverage for this and the error conditions that lead to it.

acolytec3 (changes requested):
We should bump the coverage here on the error conditions for failed tx broadcasts and I can try and help with that.  I think we should also place a guard on the tx pool logging to only run it if our loglevel is set to the txpool stats logging level (currently `info`) or higher (and preferably bump the loglevel to `debug` for tx pool stats)

g11tech:
> We should bump the coverage here on the error conditions for failed tx broadcasts and I can try and help with that. I think we should also place a guard on the tx pool logging to only run it if our loglevel is set to the txpool stats logging level (currently `info`) or higher (and preferably bump the loglevel to `debug` for tx pool stats)

hmmm we can try improving coverage, but not comfortable bumping them onto debug for reasons:
that they don't flood the screen
and provide very useful info for normal course of operation which helped me for e.g. looking at shandong nodes.

packages/client/lib/service/txpool.ts
g11tech: hmmm i can understand the reasoning, i would be ok to put them onto a debug level functioanlity where we can switch on/off module debug levels with the key press without requiring restarts, 
would it be possible to enhance `debug` with this?

packages/client/lib/service/txpool.ts
g11tech: i can test it when i take a break from withdrawals work

holgerd77:
What is the status of this PR?

acolytec3:
> What is the status of this PR?

Should be ready to go.  I'm ready to approve.  Lemme rebase and see where the test coverage is low.

g11tech:
> > What is the status of this PR?
> 
> Should be ready to go. I'm ready to approve. Lemme rebase and see where the test coverage is low.

trying to add few more testcases to improve coverage

acolytec3 (approved):
LGTM.  Great test updates!

PR #2413, opened 2022-11-05
client: add feeHistory method

g11tech:
Extracted from shandong 
-
Add the missing methods for faucet/wallets to price and submit transactions

### DIFF
--- a/packages/client/lib/rpc/modules/eth.ts
+++ b/packages/client/lib/rpc/modules/eth.ts
@@ -360,6 +360,9 @@ export class Eth {
       1,
       [[validators.blockOption]]
     )
+
+    this.feeHistory = middleware(this.feeHistory.bind(this), 0, [])
+    this.gasPrice = middleware(this.gasPrice.bind(this), 0, [])
   }
 
   /**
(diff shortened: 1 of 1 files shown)


holgerd77:
As mentioned by @acolytec3 in the chat we should not merge `feeHistory` as is since implementation was only rudimentary, also `gasPrice` is being worked on right now at #2396, will mark this as "blocked" for now to avoid accidental merge.

packages/client/lib/rpc/modules/eth.ts
```diff
@@ -986,4 +989,37 @@ export class Eth {
...
+  async gasPrice() {
+    let gasPrice: bigint = BigInt(0)
+    const latest = await this._chain.getCanonicalHeadHeader()
+    if (this._vm !== undefined && this._vm._common.isActivatedEIP(1559)) {
+      gasPrice = latest.calcNextBaseFee()
+    } else {
+      if (latest.gasUsed > BigInt(0)) gasPrice = latest.gasUsed
+      else {
+        while (gasPrice === BigInt(0)) {
+          const nextBlock = await this._chain.getBlock(latest.number - BigInt(1))
```
monstermax, line 1018: are you looping always on the same block ?

acolytec3: This is out of date. We merged a separate PR recently for `eth_gasPrice`.

holgerd77: So to add: see comment here https://github.com/ethereumjs/ethereumjs-monorepo/pull/2413#issuecomment-1305497686

acolytec3:
Closing as `gasPrice` has been implemented separately and the implementation of `feeHistory` here is just a stub and should be properly reworked from first principles in a separate PR.

PR #2419, opened 2022-11-11
stateManager: add EthersStateManager to exports

### LINKED ISSUE #2416: `EthersStateManager` isn't exported by package
The `EthersStateManager` isn't currently exported from `index.ts` so requires users to deep import it:
`import { EthersStateManager } from '@ethereumjs/statemanager/dist/ethersStateManager`

Let's fix this.

acolytec3:
Closes #2416 

Adds the `EthersStateManager` to the `statemanager` exports so people don't have to deep import it.

### DIFF
--- a/packages/statemanager/src/index.ts
+++ b/packages/statemanager/src/index.ts
@@ -1,3 +1,4 @@
 export { BaseStateManager } from './baseStateManager'
+export { EthersStateManager, EthersStateManagerOpts } from './ethersStateManager'
 export { AccountFields, StateAccess, StateManager } from './interface'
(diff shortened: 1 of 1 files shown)


acolytec3:
@holgerd77 I think this was an oversight in my `EthersStateManager` PR but it looks like we forgot to add it to the package exports in `index.ts`.  Once this is merged, we'll need to do another point release to update so people can import it without having to do `import { EthersStateManager } from '@ethereumjs/statemanager/dist/ethersStateManager'`

holgerd77 (approved):
LGTM

PR #2433, opened 2022-11-26
EthJS DEBUG variable

ScottyPoi:
Followup to PR #2325 
Addressing issue #2306 

The EVM, VM, StateManager, and devp2p packages all contain a performance feature that will skip calls to debug logging if the Debug output is not desired.

This was implemented like:
```
this.DEBUG = false
if (process.env.DEBUG !== undefined) { 
   this.DEBUG === true 
}
```
with a wrapper around calls to debug()
```
if (this.DEBUG) { 
    debug('message') 
}
```
This proved problematic for users who wish to enable DEBUG variables for other reasons without incidentally enabling EthJS debuggers.

----
This PR changes the way this.DEBUG is set, by checking instead for a master variable `ethjs`, and enabling this.DEBUG only if that variable is present.
```
    this.DEBUG = process?.env?.DEBUG?.includes('ethjs') ?? false
```

The example scripts in the various `README.md` files were updated to include the `ethjs` variable.

`DEBUG=devp2p:* npx ts-node ./examples/script.ts` 
becomes
`DEBUG=ethjs,devp2p:* npx ts-node ./examples/script.ts`

### DIFF
--- a/packages/client/README.md
+++ b/packages/client/README.md
@@ -309,13 +309,13 @@ ethereumjs --loglevel=debug
 For more in-depth debugging on networking the underlying [devp2p](../devp2p) library integrates with the [debug](https://github.com/visionmedia/debug) package and can also be used from within a client execution context:
 
 ```shell
-DEBUG=*,-babel [CLIENT_START_COMMAND]
+DEBUG=ethjs,*,-babel [CLIENT_START_COMMAND]
 ```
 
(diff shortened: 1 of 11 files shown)


holgerd77 (approved):
Ok, cool 🙂, have locally tested various on/off possibilities, seems to work fine, will merge! 👍

PR #2434, opened 2022-11-28
In readme, remove references to LevelDB abstraction

TimDaub:
`LevelDB` abstraction was removed and can now not be imported through @ethereumjs/trie >= 5.0.0 anymore. Hence, for most examples in the readme.md file, we're removing references to a top-level exported `LevelDB` abstraction, and we're replacing it with the generic and in-memory `MapDB` implementation.


holgerd77 (approved):
Thanks for submitting, LGTM! 👍 

Will admin-merge without running the tests since no production code touched.

PR #2437, opened 2022-11-30
common: Implement hardfork by time

g11tech:
The new post merge hardforks which need to be coordinated with CL are based on timestamp.

this PR attempts the same.

WIP


packages/common/src/common.ts
```diff
@@ -505,8 +534,13 @@ export class Common extends EventEmitter {
    * @param td Total difficulty
    *    * @returns The value requested or `BigInt(0)` if not found
    */
-  paramByBlock(topic: string, name: string, blockNumber: BigIntLike, td?: BigIntLike): bigint {
-    const hardfork = this.getHardforkByBlockNumber(blockNumber, td)
+  paramByBlock(
+    topic: string,
+    name: string,
+    blockNumber: BigIntLike,
+    { td, timestamp }: { td?: BigIntLike; timestamp?: BigIntLike }
```
acolytec3, line 541: I didn't know you could declare parameters like this. Will have to experiment with it and see if it makes sense.

packages/common/src/common.ts
```diff
@@ -375,8 +401,11 @@ export class Common extends EventEmitter {
    * @param td
    * @returns The name of the HF set
    */
-  setHardforkByBlockNumber(blockNumber: BigIntLike, td?: BigIntLike): string {
-    const hardfork = this.getHardforkByBlockNumber(blockNumber, td)
+  setHardforkByBlockNumber(
+    blockNumber: BigIntLike,
+    { td, timestamp }: { td?: BigIntLike; timestamp?: BigIntLike } = {}
```
acolytec3, line 406: Doesn't this change of `td` to an object make this a breaking change? I'm assuming it would be preferable to just add an additional optional parameter got timestamp.

acolytec3 (commented):
I'll have experiment more deeply with it but it looks okay so far. Let a couple of questions.

packages/common/src/common.ts
holgerd77: Yes, changing to an object is a breaking change (also for the `getHardforkByBlockNumber` method), so we would need to switch to an additional parameter.

g11tech:
the withdrawals devnet based on shanghai time successfully synced!
image
 (some cleanup still wip)

packages/common/src/common.ts
```diff
@@ -325,30 +331,41 @@ export class Common extends EventEmitter {
 
     // Find the first hardfork that has a block number greater than `blockNumber` (skips the merge hardfork since
```
acolytec3, line 332: We should update this to note `blockNumber or timestamp`

packages/common/src/common.ts
```diff
@@ -636,67 +702,45 @@ export class Common extends EventEmitter {
     return BigInt(ttd)
   }
```
acolytec3, line 704: I don't think we can remove this function as this is part of the public API without this being a breaking change.  What's driving removal of this?

packages/common/src/common.ts
```diff
@@ -636,67 +702,45 @@ export class Common extends EventEmitter {
...
-      const mergeIndex = hfs.findIndex((hf) => hf.ttd !== null && hf.ttd !== undefined)
-      if (mergeIndex < 0) {
-        throw Error(`Merge hardfork should have been found`)
-      }
-      hfBlock = this.hardforkBlock(hfs[mergeIndex - 1].name)
-    }
-    if (hfBlock === null) {
+    const hfs = this.hardforks()
+    let hfIndex = hfs.findIndex((hf) => hf.name === hardfork)
+    // If the current hardfork is merge go, one behind as merge hf is not part of these
```
acolytec3, line 714: The meaning of this comment isn't clear to me.  Can you clarify?

packages/common/src/common.ts
```diff
@@ -636,67 +702,45 @@ export class Common extends EventEmitter {
...
+    // calcs even if the merge hf block is set
+    if (hardfork === Hardfork.Merge) {
```
acolytec3, line 716: Can we move this above current line 712 since this check is independent of looking up the hfIndex.

packages/common/src/common.ts
```diff
@@ -636,67 +702,45 @@ export class Common extends EventEmitter {
...
-    hardfork = hardfork ?? this._hardfork
-    const block = this.hardforkBlock(hardfork)
-    return typeof block === 'bigint' && block !== BigInt(0) ? block === blockNumber : false
-  }
-
   /**
    * Returns the change block for the next hardfork after the hardfork provided or set
    * @param hardfork Hardfork name, optional if HF set
-   * @returns Block number or null if not available
+   * @returns Block timestamp, number or null if not available
```
acolytec3, line 708: I'm a little anxious about the naming of this method now.  I know we don't really want a breaking release for `common` but its a complete misnomer for us to call `nextHardforkBlock` if we start returning timestamps as one possibility.  We should think very carefully about this and consider maybe adding a new function nameed something like `nextHardforkBlockOrTimestamp` (and then deprecate `nextHardForkBlock` in the same release with an intent to remove it in the next breaking release).  My biggest concern is that there's no way for the library consumer to know programmatically whether or not we're returning a block number or a timestamp.  I realize there are other factors in play whereby library consumers can know this (check the consensus type and see if PoS or other) but that wouldn't be reliable since it could vary across with network you're on.

packages/common/src/types.ts
```diff
@@ -48,6 +48,7 @@ export interface HardforkConfig {
   name: Hardfork | string
   block: number | null // null is used for hardforks that should not be applied -- since `undefined` isn't a valid value in JSON
   ttd?: bigint | string
+  timestamp?: number | string
```
acolytec3, line 51: Is it possible for this to be `null` if a timestamp hardfork is read in from a geth genesis config file?

packages/common/src/common.ts
```diff
@@ -375,8 +422,12 @@ export class Common extends EventEmitter {
    * @param td
    * @returns The name of the HF set
```
acolytec3, line 423: Need to update the documentation to add the optional `timestamp` parameter

acolytec3 (changes requested):
Looking good overall.  I think we need to carefully consider some of these changes though and make sure we're handling edge cases.  I know I haven't thought of all of them but will try and spin up the withdrawals devnet and validate directly to look for more edge cases.

acolytec3 (changes requested):
Looking good overall.  I think we need to carefully consider some of these changes though and make sure we're handling edge cases.  I know I haven't thought of all of them but will try and spin up the withdrawals devnet and validate directly to look for more edge cases.

packages/common/src/common.ts
```diff
@@ -307,13 +307,19 @@ export class Common extends EventEmitter {
...
     blockNumber = toType(blockNumber, TypeOutput.BigInt)
     td = toType(td, TypeOutput.BigInt)
+    timestamp = toType(timestamp, TypeOutput.Number)
 
-    // Filter out hardforks with no block number and no ttd (i.e. unapplied hardforks)
+    // Filter out hardforks with no block number, no ttd or no timestamp (i.e. unapplied hardforks)
     const hfs = this.hardforks().filter(
-      (hf) => hf.block !== null || (hf.ttd !== null && hf.ttd !== undefined)
+      (hf) =>
+        hf.block !== null || (hf.ttd !== null && hf.ttd !== undefined) || hf.timestamp !== undefined
```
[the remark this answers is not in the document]
g11tech: unless someone specifies them as null and tries to use them outside typescript since typescript would complain while trying to match with `HardforkConfig` type.
But open to having/allowing them as null and introduce appropriate checks, but may be as a separate PR

packages/common/src/common.ts
g11tech: we don't use it at all, let me know if this needs restoration

packages/common/src/common.ts
g11tech: the comment means is to ignore merge hardfork to determine what is current fork and what is the next, its a rehash of what was previously being done as `hfBlock = this.hardforkBlock(hfs[mergeIndex - 1].name`

packages/common/src/types.ts
g11tech: haven't seen null specified in geth genesis, but i can make sure we remove those hardforks from consideration where their block/time is specified as null

packages/common/src/common.ts
acolytec3: It's not whether we use it or not. It's part of the public API so any downstream consumer of our libraries could use it and this would break their usage. We can't remove functions or change their parameters without it being a breaking change. I'm not opposed it it in principle but we have to be really careful about breaking releases.

packages/common/src/common.ts
acolytec3: Let's move this check above line 730.  It's a small thing but it will give us one slight bit of efficiency provided we're on the merge fork.

acolytec3 (changes requested):
Just a couple of small nits at this point.  Looks great!

packages/common/src/common.ts
g11tech: but this updates the hfIndex to one hardfork back, sorry can't understand how to move it to L730

acolytec3: 🤦 Disregard. I just can't read.

packages/common/src/enums.ts
```diff
@@ -25,6 +25,7 @@ export enum Hardfork {
   Merge = 'merge',
   Eof = 'eof',
   Shanghai = 'shanghai',
+  ShardingFork = 'shardingFork',
```
holgerd77, line 28: I think I am still not happy that we are mixing the super-official mainnet forks like `Shanghai` with our own inventions so transparently here and I wouldn't want to release it this way.

Can we rename this to something which clearly separates, e.g. `Eof_INTERNAL`, `ShardingFork_INTERNAL`, so that this is not sneaking in into peoples production code somehow and will unnecessarily break their code if we change here?

holgerd77 (commented):
Yeah, first: thanks a lot Gajinder, great work!!! 🎉 Super-important building block for future Ethereum updates.

I nevertheless already pray for the day when we can clean this code up again. 😋 This yet-another-mechanism addition gets really ugly, regarding the existing naming, the overall bloat of the code, chances that we have some edge cases in there are relatively high.

I would nevertheless have a tendency to for now take this in as we have, I guess that's very much a candidate for a first-round refactor on the next breaking release round, e.g. in a first round changing the names of the `Common` functions and take in a dictionary as Gajinder already did at some point.

Ok, so I would have one specific comment to address on our new internal forks, otherwise I am fine with merging! 🙂

packages/common/src/enums.ts
acolytec3: Done.  I also removed the shardingFork reference from the EVM list of supported hardforks for now.  It will stay in my 4844 branch for now.

acolytec3 (approved):
LGTM - will merge once tests pass given @holgerd77 's written approval of the changes as well.

holgerd77:
@acolytec3 Thanks a lot for updating, not sure - might not be pressing - just to make sure: weren't there more references of these hardforks - e.g. `Eof` - in the testnet scripts from @g11tech e.g.? 🤔

holgerd77:
What is this sharding HF for, is this for your 4844 work? Do we need this (in master) at all?

acolytec3:
>https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/test/sim/eof.spec.ts

I don't see anything in here that references the `hardfork` enum so not sure what you're pointing to.

> What is this sharding HF for, is this for your 4844 work? Do we need this (in master) at all?

Yeah, that can be removed.  I wasn't thinking too closely about it and I think we've been so focused on the 4844 work it just got conflated with the dummy EOF hardfork so left in.

holgerd77:
> > https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/test/sim/eof.spec.ts
> 
> I don't see anything in here that references the `hardfork` enum so not sure what you're pointing to.
> 
> > What is this sharding HF for, is this for your 4844 work? Do we need this (in master) at all?
> 
> Yeah, that can be removed. I wasn't thinking too closely about it and I think we've been so focused on the 4844 work it just got conflated with the dummy EOF hardfork so left in.

Ah, you are right, this is using the string directly, didn't look to closely.

Ok, I might delete the HF along release if you don't contradict.

Also: still planning to do the releases, maybe end of the week or beginning of next week.

acolytec3:
> Ok, I might delete the HF along release if you don't contradict.

Sounds good to me.  Definitely support keeping master as clean as possible.

g11tech:
yes eof.spec.ts doesn't directly refer the hardfork, but its refered in the geth genesis used for it in the config, `eofBlock`, since you updated the map for parsing geth genesis to the new Hardfork.eof_INTERNAL, so that should be correctly used now

PR #2441, opened 2022-12-02
Fix logic bug in txPool.validate

acolytec3:
Fixes a logic bug in the `txPool.validate` function where we are retrieving an account balance from the wrong vm and stateManager instance.  This caused an issue where `txPool.validate` would fail when submitting a transaction via the RPC `eth_sendRawTransaction` endpoint after restarting the client on an existing stateDB where no transactions/blocks were being run.  `setStateRoot` was being called on the `vmCopy` but the account balance was being retrieved from the `vm.stateManager` which had an unknown stateroot and the account might not have the correct balance.


gabrocheleau (dismissed):
Looks good to me.

acolytec3:
Fixed the other broken client test here.  Should be ready for final review.

holgerd77 (approved):
Cool, nice find 🙂, LGTM! 👍

Issue #2438, opened 2022-11-30
Implement an ethers provider that wraps `ethersStateManager` 


acolytec3:
Implementing a new provider that wraps the `ethersStateManager` would allow for simplified testing of continuous simulated transactions against any block height where the state changes made by any given transaction can be maintained.

ScottyPoi:
Took a stab at this following the model of the Ultralight provider.

Biggest question has to do with `getAccount` - and what the equivalent `ethers` call is.

acolytec3:
> Took a stab at this following the model of the Ultralight provider.
> 
> Biggest question has to do with `getAccount` - and what the equivalent `ethers` call is.

There isn't a 1-1 equivalent there. You would use it inside of `ethers.provider.getBalance` if the main one I can think of.

acolytec3:
I just realized my original issue description wasn't exactly correct be there is no `getAccount` in the rpc. It's really just `getBalance`. You might also have to implement a wrapper around `getTransactionCount` and then use `stateManager.getAccount` to retrieve the nonce and return that.

ScottyPoi:
> There isn't a 1-1 equivalent there. You would use it inside of `ethers.provider.getBalance` if the main one I can think of.

So i should include a wrapper for getBalance and others that look up accounts?

acolytec3:
> 
> > There isn't a 1-1 equivalent there. You would use it inside of `ethers.provider.getBalance` if the main one I can think of.
> 
> So i should include a wrapper for getBalance and others that look up accounts?

I think `getBalance` and `getTransactionCount` are the two. Would need to look through the rest of the rpc endpoint list to see if there are any others.

ScottyPoi:
@acolytec3 can you take a peek at #2442 and tell me if I'm doing what you had in mind?

In working on it I found a couple inconsistencies in `EthersStateManager`.   Between what the comments said should happen, and what was happening in the methods.

acolytec3:
Will do today

PR #2442, opened 2022-12-03
Ethers forked state provider

ScottyPoi:
Implementing idea from #2438


packages/statemanager/src/ethersStateManager.ts
```diff
@@ -137,6 +137,8 @@ export class EthersStateManager extends BaseStateManager implements StateManager
       storage = accountStorage.get(key.toString('hex'))
       if (storage !== undefined) {
         return storage
+      } else {
+        return Buffer.from([])
```
acolytec3, line 141: Same here, add a test to the state manager test?

packages/statemanager/src/ethersForkedStateProvider.ts
```diff
@@ -0,0 +1,81 @@
...
+
+import type { Account } from '@ethereumjs/util'
+
+export class EthersForkedStateProvider extends ethers.providers.JsonRpcProvider {
+  private fallbackProvider: ethers.providers.JsonRpcProvider
+  private ethersStateManager: EthersStateManager
+  constructor(provider: string | ethers.providers.JsonRpcProvider) {
+    super(typeof provider === 'string' ? provider : provider.connection)
+    this.fallbackProvider =
+      typeof provider === 'string' ? new ethers.providers.JsonRpcProvider(provider) : provider
```
acolytec3, line 14: As above, let's just remove this entirely.

acolytec3 (changes requested):
Great start!  TLDR on my feedback is we shouldn't need the fallback provider to be instantiated in this class as it's already handled by the state manager.

packages/statemanager/src/ethersStateManager.ts
ScottyPoi: Strange behavior here.

Without the change -- it seems clear that `undefined` is a potential return value.
But if you add `undefined` to the return types -- `async getContractStorage(address: Address, key: Buffer): Promise<Buffer | undefined> {...`,
typescript complains that it now incorrectly implements the BaseStateManager method.

Is there some under-the-hood typescript magic happening that causes this to return a buffer no matter what??

ScottyPoi:
Great thanks for the feedback.
Got rid of the `fallback provider` stuff, and the unnecessary `getAccount`.

In trying to write tests for the `EthersStateManager` fixes, i encountered some weirdness that was (i think) giving us some false positives in the test.

```
 async putContractStorage(address: Address, key: Buffer, value: Buffer): Promise<void> {
    let accountStorage = this.storageCache.get(address.toString())
    if (accountStorage === undefined) {
      this.storageCache.set(address.toString(), new Map<string, Buffer>())
      accountStorage = this.storageCache.get(address.toString())
    }
    accountStorage?.set(key.toString('hex'), value)
  }
```

For unknown reasons, the conditional as written did not actually kick in if the first `this.storageCache.get()` returned undefined.  So the empty Map() would not be set, and `accountStorage` would remain `undefined`.  The `?` at the end allowed this to happen without error.  This method was also missing the function of `deleting` a key from the Map when  a value is deleted.

Then because of an unknown fallback mechanism happening in `EthersStateManager.getContractStorage()`, we would end up with the correct return value (an empty buffer) anyway, further masking the issue.  The way it was written (below), it should return nothing (`undefined`) unless both `accountStorage` and `storage` are found.  But it would not return `undefined`... it would return an empty buffer.  Which is what it *should* return -- but where is that empty buffer coming from?   

```
  async getContractStorage(address: Address, key: Buffer): Promise<Buffer> {
    // Check storage slot in cache
    const accountStorage: Map<string, Buffer> | undefined = this.storageCache.get(
      address.toString()
    )
    let storage: Buffer | string | undefined
    if (accountStorage !== undefined) {
      storage = accountStorage.get(key.toString('hex'))
      if (storage !== undefined) {
        return storage
      }
    }
```

By rewriting the first method like this:
```
  async putContractStorage(address: Address, key: Buffer, value: Buffer): Promise<void> {
    const accountStorage = this.storageCache.get(address.toString()) ?? new Map<string, Buffer>()

    if (value.length === 0 || value.equals(Buffer.alloc(value.length).fill(0))) {
      accountStorage.delete(key.toString('hex'))
    } else {
      accountStorage.set(key.toString('hex'), value)
    }
    this.storageCache.set(address.toString(), accountStorage)
  }
```

We add the function of actually deleting the key, while also removing the potential for "nothing at all" to happen.
Making this change caused this test to fail:
```
    // Verify that provider is not called
    ;(state as any).getAccountFromProvider = function () {
      throw new Error('should not have called this!')
    }
```
Which was then fixed by adding the missing return value (`...} else {return Buffer.from([])`)  to `getContractStorage`:
```
  async getContractStorage(address: Address, key: Buffer): Promise<Buffer> {
    // Check storage slot in cache
    const accountStorage: Map<string, Buffer> | undefined = this.storageCache.get(
      address.toString()
    )
    let storage: Buffer | string | undefined
    if (accountStorage !== undefined) {
      storage = accountStorage.get(key.toString('hex'))
      if (storage !== undefined) {
        return storage
      } else {
        return Buffer.from([])
      }
    }
```

packages/statemanager/src/ethersStateManager.ts
acolytec3: You can't add `undefined` because that would break the interface that we're implementing.  The interface [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/25223887ed2f7ec8bfd62473f165d332223b687d/packages/statemanager/src/interface.ts#L22) requires that this function returns a `buffer`. That is interesting though that Typescript didn't complain about not covering that edge case.  I think your conditional works fine though.

acolytec3:
So I think I misread your original comment and I think the short answer to your question is that the reason the correct empty buffer is being returned is that if a call to `getContractStorage` doesn't find the storage value in the local cache, it always calls out to the provider and retrieves a value from there.  If the provider returns undefined, we create an empty buffer at that point and return it.  So, I think we need to revert your change in `getContractStorage`.

ScottyPoi:
> So I think I misread your original comment and I think the short answer to your question is that the reason the correct empty buffer is being returned is that if a call to `getContractStorage` doesn't find the storage value in the local cache, it always calls out to the provider and retrieves a value from there. If the provider returns undefined, we create an empty buffer at that point and return it. So, I think we need to revert your change in `getContractStorage`.

Where does that happen?

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -167,7 +169,11 @@ export class EthersStateManager extends BaseStateManager implements StateManager
       this.storageCache.set(address.toString(), new Map<string, Buffer>())
       accountStorage = this.storageCache.get(address.toString())
     }
-    accountStorage?.set(key.toString('hex'), value)
+    if (value.length === 0 || value.equals(Buffer.alloc(value.length).fill(0))) {
```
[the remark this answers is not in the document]
ScottyPoi: More dyslexia -- "delete value" does not mean delete key, just to replace value with empty buffer.
This still needed a tweak, because it did not account for buffers full of zeros triggering a deletion.

I believe the reason there is to prevent users from stockpiling storage space with buffers full of nothing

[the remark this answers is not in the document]
acolytec3: > because it did not account for buffers full of zeros triggering a deletion.
>I believe the reason there is to prevent users from stockpiling storage space with buffers full of nothing

Good catch.  I think you're right.  This is where gas tokens come from most likely.

ScottyPoi:
The tests I've written do not test the "blockTag" parameter for `getBalance`, `getStorageAt`, and `getTransactionCount`, and I'm having wondering what you think the best approach to that might be?

acolytec3:
> The tests I've written do not test the "blockTag" parameter for `getBalance`, `getStorageAt`, and `getTransactionCount`, and I'm having wondering what you think the best approach to that might be?

I wouldn't lose any sleep over that.  I covered all of that in the original `ethersStateManager` tests and I don't see anything new you're doing with that parameter that should require us to test it again.

packages/statemanager/test/ethersForkedStateProvider.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+})
+
+tape('getCode / getStorageAt', async (t) => {
+  if (isBrowser() === true) {
+    // The `MockProvider` is not able to load JSON files dynamically in browser so skipped in browser tests
+    t.end()
+  } else {
+    const provider =
+      process.env.PROVIDER !== undefined
+        ? new StaticJsonRpcProvider(process.env.PROVIDER, 1)
```
acolytec3, line 102: Is this what the Mock Provider returns for the uni contract code?  :grimace:

acolytec3:
You know what just occurred to me that may be missing from this whole implementation is that we don't have a way to to actually simulate the transactions without the dapp developer actually having to implement and run our VM.  The whole idea here is to be able to simulate and run transactions just from this provider and I totally missed that when I was sketching out the initial solution.  As such, we're also going to have to override `provider.sendTransaction`.

In thinking about it, the implementation should be something like this:
1.  We need to add a private `vm` property to the provider and instantiate it using the `ethersStateManager` property.
2. Override `sendTransaction` as follows:
- Deserialize the signed transaction parameter
- Call this.vm.runTx(decodedTransaction)
- Return the result in the format provided 

If I'm thinking about things correctly, since the private `vm` will run the transactions using the `ethersStateManager`, the state updated applied by running the transaction should magically be updated so if you call the `provider.getBalance` method after running a transaction, you should see the expected balance net of transaction fees.  There's probably some edge case I'm not contemplating but that seems like it should work as roughly described above. 

A fruitful set of tests would be to run a couple of transactions and then verify that the post account balances match up.  I can probably help devise a contract with some storage updates as well if needed to verify that everything really works properly.

ScottyPoi:
> A fruitful set of tests would be to run a couple of transactions and then verify that the post account balances match up. I can probably help devise a contract with some storage updates as well if needed to verify that everything really works properly.

Thanks.  I'll get it going and let you know where I need help.

ScottyPoi:
> We need to add a private vm property to the provider and instantiate it using the ethersStateManager property.

Is there a good way to instantiate a VM in a constructor? since VM.create( )  client/vmexecution does this:
```ts
    this.vm = new (VM as any)({
      stateManager: this.ethersStateManager,
    }) as VM
```

ScottyPoi:
> If I'm thinking about things correctly, since the private vm will run the transactions using the ethersStateManager, the state updated applied by running the transaction should magically be updated so if you call the provider.getBalance method after running a transaction, you should see the expected balance net of transaction fees

Yes, this is working as expected.  
I have to figure out how to turn our `TxResult` from vm.runTx() into the `TransactionResponse` interface that the ethers provider expects to return

roninjin10:
This is super useful! I need this as well gonna implement this for my viem state manager

holgerd77:
Oh, this got a bit forgotten. 😳

I have to admit that I only now (better late than never) getting the whole idea of this. Yes, this is really great!

So how do we want to proceed?

My biggest fear here is the "yet another package to maintain" thing. Would it be a way that we integrate this into `StateManager` instead and then live with an optional Ethers dependency, as we do for EthersStateManager?

Then we can do everything one level smaller, do not need to setup an own README, tests, CI, and all this stuff.

WDYGT?

PR #2443, opened 2022-12-04
Add eth_getTransactionByBlockHashAndIndex RPC Endpoint

JowieXiang:
Adding `eth_getTransactionByBlockHashAndIndex` RPC endpoint.


acolytec3:
This looks good so far. Will review again once the tests are added.

acolytec3 (approved):
LGTM!  Will merge once the tests pass.

JowieXiang:
Seems one existing test is failing 
```
# should sync to next subchain head or chain height
ok 677 should sync block 5 and 4
ok 678 should target syncing all the way to chain
not ok 679 TypeError: Cannot read properties of null (reading 'first')
  ---
    operator: error
    stack: |-
      TypeError: Cannot read properties of null (reading 'first')
          at Test.<anonymous> (/home/runner/work/ethereumjs-monorepo/ethereumjs-monorepo/packages/client/test/sync/beaconsync.spec.ts:146:27)
  ...
```

acolytec3:
There's a race condition somewhere in one of the client sync tests where this occasionally fails and we just haven't chased it down yet.  I'm rerunning it now and will merge once it passes.

JowieXiang:
> There's a race condition somewhere in one of the client sync tests where this occasionally fails and we just haven't chased it down yet. I'm rerunning it now and will merge once it passes.

I can look into that if it does not require too much insider knowledge and is not very pressing. Or I can proceed with the other JSON-RPC API. Thx :)

acolytec3:
> > There's a race condition somewhere in one of the client sync tests where this occasionally fails and we just haven't chased it down yet. I'm rerunning it now and will merge once it passes.
> 
> I can look into that if it does not require too much insider knowledge and is not very pressing. Or I can proceed with the other JSON-RPC API. Thx :)

I honestly don't even know how to reproduce it so probably not a simple thing to chase down unfortunately.  Maybe the other RPC endpoint would be a good next step.

JowieXiang:
> > > There's a race condition somewhere in one of the client sync tests where this occasionally fails and we just haven't chased it down yet. I'm rerunning it now and will merge once it passes.
> > 
> > 
> > I can look into that if it does not require too much insider knowledge and is not very pressing. Or I can proceed with the other JSON-RPC API. Thx :)
> 
> I honestly don't even know how to reproduce it so probably not a simple thing to chase down unfortunately. Maybe the other RPC endpoint would be a good next step.

Understood. Thx~

PR #2447, opened 2022-12-12
Exclude sim from client unit tests

acolytec3:
Excludes `sim` directory from `client` tests since this isn't really a unit test but an integrated end-to-end test for validating experimental hardforks.


g11tech (approved):
lgtm

PR #2451, opened 2022-12-15
Fix EVM test scripts

acolytec3:
The `evm` tests have been failing since #2405 but the `npm run coverage` script wasn't picking this up due to the way it was constructed and missing the error code returned by the initial call to run the tests.

TODO
- [ ] - Fix broken tests in `evm/test/memory.spec.ts`

### DIFF
--- a/packages/evm/package.json
+++ b/packages/evm/package.json
@@ -31,7 +31,7 @@
     "build": "../../config/cli/ts-build.sh",
     "clean": "../../config/cli/clean-package.sh",
     "coverage": "c8 --all --reporter=lcov --reporter=text --reporter=text-lcov npm run coverage:test",
-    "coverage:test": "npm run test; cd ../vm; npm run tester -- --state",
+    "coverage:test": "npm run test && cd ../vm && npm run tester -- --state",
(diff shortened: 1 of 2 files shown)


acolytec3:
@jochem-brouwer In looking at the changes from #2405 again, why do we even bother creating a memory store of anything other than 8192 bytes?  I know this is ancient history but it seems like it might make it might be more efficient to just have the private `store` property be a `buffer` of 8192 bytes and then have a second `length` property that keeps track of the length of the "memory" for a given contract?  That way, we're never having to create new buffers at all.

jochem-brouwer:
We do, because every call frame needs a new buffer (so up to 1024 buffers). And we need to clear out the created buffers. If we keep "caching" this memory it means that we keep the memory allocated for the tx which used the most memory per frame (so we do not "garbage collect"). So I am not sure if this is the way the go?

acolytec3:
As best I can tell, the code you introduced in #2404 just automatically expands the memory buffer to 8192 bytes every you read from it, if it's not already 8192 bytes. So maybe I'm not sure the difference?

jochem-brouwer (approved):
LGTM!

PR #2457, opened 2022-12-17
client: Add blockValue to the getPayload response

g11tech:
on/post shanghai i.e. getPayloadV2 onwards, the getPayload should also return blockValue which natively is to be implemented as fees accrued  to the miner

See 
- https://github.com/ethereum/execution-apis/pull/314/files

This is going to be part of next withdrawal testnet amongst other things


acolytec3 (approved):
Looks good to me.

acolytec3:
Looks good and will merge here.

PR #2458, opened 2022-12-19
common: Update forkhash calculation for timebased hardforks

g11tech:
Update the forkhash calculation for time based hardforks as specified here: 
- https://github.com/ethereum/EIPs/pull/6122


holgerd77 (approved):
LGTM

PR #2459, opened 2022-12-21
common: Add eips 3651,3855,3860 to shanghai hf for Zhejiang shanghai testnet

g11tech:
Adding the eips for shanghai hf which are now part of the Zhejiang testnet

see discussion here: https://discord.com/channels/595666850260713488/892088344438255616/1055081293450317894

### DIFF
--- a/packages/common/src/hardforks/shanghai.json
+++ b/packages/common/src/hardforks/shanghai.json
@@ -1,7 +1,7 @@
 {
   "name": "shanghai",
-  "comment": "Next feature hardfork after the merge hardfork",
+  "comment": "Next feature hardfork after the merge hardfork having withdrawals, warm coinbase, push0, limit/meter initcode",
   "url": "https://github.com/ethereum/pm/issues/356",
   "status": "Pre-Draft",
-  "eips": [4895]
+  "eips": [4895, 3651, 3855, 3860]
 }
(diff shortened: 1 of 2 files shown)


holgerd77 (approved):
Looks good, feel free to merge, just a note: not following so closely but can we at some point make sure that we are not re-using the eof.spec.ts file and other eof-named stuff for other testnet but rename properly?

g11tech:
> Looks good, feel free to merge, just a note: not following so closely but can we at some point make sure that we are not re-using the eof.spec.ts file and other eof-named stuff for other testnet but rename properly?

yes, as of now we are not reusing the eof spec, may be we can rename it back to shangdong.spec.ts without introducing the shandong hf and programatically applying the eips on shanghai hf

packages/common/src/hardforks/shanghai.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "shanghai",
-  "comment": "Next feature hardfork after the merge hardfork",
+  "comment": "Next feature hardfork after the merge hardfork having withdrawals, warm coinbase, push0, limit/meter initcode",
```
jochem-brouwer, line 3: Could we add that this fork is still experimental?

holgerd77: Yes, that would be important so people can clearly distinguish, we should definitely do before next releases.

Have added to tiny-fixes issue: https://github.com/ethereumjs/ethereumjs-monorepo/issues/2105

PR #2461, opened 2022-12-23
common: Update setForkHashes to update timebased hardfork forkhashes

g11tech:
While trying to sync the eip4844 devnet 3, ethereumjs was disconnecting from peer as it wasn't able to identiy the peer's forkhash which was on sharding hf based on time. 
On debugging realized that the we were not setting forkhashes for timestamp based hardfork.
This PR fixes the same, and as a result was able to peer successfully with the geth which was on sharding hf


acolytec3 (approved):
Good catch! Lgtm

PR #2462, opened 2022-12-23
client: Handle withdrawal bodies in the blockfetcher and skeleton sync fixes

g11tech:
Handle withdrawal bodies in the blockfetcher and skeleton sync fixes for beacon syncing from a post sharding eip4844 devnet 3.

### DIFF
--- a/packages/client/lib/sync/fetcher/blockfetcher.ts
+++ b/packages/client/lib/sync/fetcher/blockfetcher.ts
@@ -66,17 +66,24 @@ export class BlockFetcher extends BlockFetcherBase<Block[], Block> {
       `Requested blocks=${blocksRange} from ${peerInfo} (received: ${headers.length} headers / ${bodies.length} bodies)`
     )
     const blocks: Block[] = []
-    for (const [i, [txsData, unclesData]] of bodies.entries()) {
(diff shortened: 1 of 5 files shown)


acolytec3 (approved):
Looks good to me!

packages/client/lib/sync/fetcher/blockfetcher.ts
```diff
@@ -66,17 +66,24 @@ export class BlockFetcher extends BlockFetcherBase<Block[], Block> {
       `Requested blocks=${blocksRange} from ${peerInfo} (received: ${headers.length} headers / ${bodies.length} bodies)`
     )
     const blocks: Block[] = []
-    for (const [i, [txsData, unclesData]] of bodies.entries()) {
+    for (const [i, [txsData, unclesData, withdrawalsData]] of bodies.entries()) {
```
acolytec3, line 69: I'm assuming referencing `withdrawalsData` on a pre-Shanghai block won't draw an error referencing an undefined object, right?  All the tests are passing so this appears to be the case and you explicitly check for `withdrawalsData` existing below.

g11tech: yes, it will come undefined

Issue #2398, opened 2022-11-03
Update EIP 3540 code


acolytec3:
Our current implementation of EIP-3540 EOF code validation in `evm/src/eof.ts` uses an older version of the reference implementation that is no longer present in the EIP.  The code still works with all current EIP3540 tests passing but it would be fruitful to update our implementation to match the current spec.  There are several items to be addressed
- [ ] `eofCodeAnalysis` should be updated to match the [current reference implementation](https://eips.ethereum.org/EIPS/eip-3540#reference-implementation) for code validation
- [ ] The current checks for EOF code check for `FORMAT` (0xEF) and `MAGIC` (0x00) separately and this has been merged in the EIP into [the `MAGIC`](https://eips.ethereum.org/EIPS/eip-3540#the-magic) (0xEF00).  These checks should be updated in `evm.ts` and `interpreter.ts` and the `FORMAT` and `MAGIC` constants exported by `eof.ts` should be consolidated to just `MAGIC`
- [ ] `loadEOFCode` in `evm/src/eof.ts` should be revised to just pull the code section from the EOF container directly by reading the `EOF` header to get the code section and do `subArray` (or `slice`) to get the code section rather than using the `codeAnalysis` method to figure out where the code section is since this duplicates the code analysis done at deploy time and is a small, but important performance drag since the whole point of the EIP is to not have to revalidate the code container every time we run the code.

acolytec3:
Superseded by #2454

PR #2453, opened 2022-12-16
Implement Big EOF EIPs

jochem-brouwer:
Implements the EIPs according to https://github.com/ethereum/EIPs/tree/4e5cd9fe6dd168165a7a5cfef68ead439ef27f17

The EIPs in question are:

- 3540 (EOF header format)
- 4750 (EOF functions)
- 5450 (EOF Stack validation)
- 3670 (EOF code validation)
- 4200 (EOF static relative jumps)

Note that 3860 (limit and meter initcode) is also part of this one, but it does not seem that we need changes to that one here.

This PR aims to create code which passes the consensus tests (not yet available). The code is not optimized. Optimizing the code will make it less readable and this is not part of this PR.

Some notes regarding optimization:

- `checkOpcodes` (essentially 3670 + 5450 checks) passes the code twice
- When creating an `EOFContainer` the container is always verified. This is only necessary when checking `initcode` (e.g. create transaction, `CREATE` opcode), or when deploying a new EOF contract.

TODOS

- [ ] Write tests for 5450
- [ ] Write tests for 4750
- [ ] Rewrite tests for 4200 (or at helper methods to convert old tests)
- [ ] Ensure that EOF1 contracts cannot deploy legacy contracts
- [ ] Ensure that EOF1 checks on initcode/code on both txs and CREATE/CREATE2 consume the right amount of gas when the validation fails (see EIP 4200, 3540)
- [ ] Stack delta stuff should be part of `Opcode` class (`codes.ts`)
- [ ] Change stack delta stuff to only update inputs/outputs. For swap/dup, (e.g. swap16) mark swap16 as 16 inputs and 16 outputs (delta = 0, but min stack height of 16 is now implicitly verified)
- [ ] Ensure the newly introduced opcodes (RJUMP, RJUMPV, RJUMPI, CALLF, RETF) are INVALID in legacy contracts
- [ ] Await until `ethereum/tests` gets a final spec. Note: we should wait until the tests are also updated with this (currently unmerged)  PR: https://github.com/ethereum/EIPs/pull/6249. Currently we fail the EIP 3860 tests, and it makes no sense currently to update those, as we are not know if the tests are incorrect, or our implementation is incorrect. (We need 3860 to test EOF).

**NOTE: since Big EOF is highly evolving, always check above ^ URL to see to which version of the EIPs it points**

Version updates:

[...]

PR #2465, opened 2022-12-24
Update EOF to current spec

ScottyPoi:
Addressing issue #2398 

> Our current implementation of EIP-3540 EOF code validation in `evm/src/eof.ts` uses an older version of the reference implementation that is no longer present in the EIP.... 

1. `eofCodeAnalysis`  updated to match the [current reference implementation](https://eips.ethereum.org/EIPS/eip-3540#reference-implementation) for code validation

1. EOF checks updated in `evm.ts` and `interpreter.ts` and the `FORMAT` and `MAGIC` constants exported by `eof.ts` consolidated to just `MAGIC`

1. `loadEOFCode` in `evm/src/eof.ts`  revised to just pull the code section from the EOF container directly by reading the EOF header to get the code section and do subArray (or slice) to get the code section rather than using the `codeAnalysis`.

----

#### `/evm/eofContainer.ts`
- Class objects for EOF components
  - SectionHeader
    - sectionKind
    - sectionSize(s)
  - EOFHeader
    - typeSize
    - numCodeSections
    - codeSize
    - dataSize
  - TypeSection
    - inputs
    - outputs
    - maxStackHeight
  - EOFBody
    - typeSections
    - codeSections
    - dataSection
  - EOFContainer
    - eofHeader
    - eofBody
- validate()
- toBuffer()
- fromBuffer()
- *helper methods*
  - `header.sectionHeaders()`
  - `header.sectionSizes()`
  - `body.size()`
  - `body.sections()`
  - `body.sectionSizes()`


jochem-brouwer:
Thanks @ScottyPoi, but I think we should do a complete overhaul now that the EIPs have been updated again... (I plan to do this in https://github.com/ethereumjs/ethereumjs-monorepo/pull/2453)

ScottyPoi:
Cool -- I have gone a bit farther. I'll compare with what you've done.   Even if I didn't do anything new, it was useful to me 😋

jochem-brouwer:
Could you push your changes? :smile:

jochem-brouwer:
I have somewhat started doing this overhaul but a lot, if not all, needs to be rewritten I think (because the spec has changed almost everything)

jochem-brouwer:
See this commit, this should bring eof validation up-to-date with latest spec https://github.com/ethereumjs/ethereumjs-monorepo/pull/2453/commits/8163a81d0f8aabdc7f633581902d20869e2e9bdc

(Note: this does not build, because I have un-exported some methods)

ScottyPoi:
@jochem-brouwer - 

Let me know how you feel about the approach I took in `eofTypes.ts`.

jochem-brouwer:
@ScottyPoi this is very well structured and looks very good. I have some remarks though:

The `buffer()` method seems to convert the contents of an EOF header (or section) back into a Buffer. This is nice, but does it have practical usage..? This would actually be the task of the compiler (e.g. solidity) to create these buffers. It is our task to only verify it (although this would be a very nice helper function I'd say to write tests).

I think your approach is the way to go, I now have a very ugly verifier: see https://github.com/ethereumjs/ethereumjs-monorepo/pull/2453/files and then `eof.ts`

However it seems like we are now somewhat doing the same work at almost the same time, could you ping me if you work on EOF on discord? :smile: And should we start targeting PR #2453 to move all big EOF stuff in there?

I will continue with my rather ugly verifier for now but we should update it to a more structured way (like how you wrote the eof container!)

jochem-brouwer:
@ScottyPoi also see https://github.com/shemnon/besu/blob/b39ada871bbc95638b451a2f50e437a6028e5883/evm/src/main/java/org/hyperledger/besu/evm/code/EOFLayout.java

jochem-brouwer:
I am going to merge this format into #2465 if that is ok :)

jochem-brouwer:
I will close here, so we can continue on #2453?

PR #2467, opened 2023-01-01
block: Handle hardfork defaults consistently

g11tech:
Noticed in eip4844 and elsewhere that its better to set defaults properly for the required fields which makes the UX of initializing a default object easy with just `Block.fromBlockData/Block.fromHeaderData` with correct hardfork defaults.

This fixes the  addTransaction flow in the `vm/src/buildBlock.ts` which has this code flow: 
```
    const blockData = { header, transactions: this.transactions }
    const block = Block.fromBlockData(blockData, this.blockOpts)
```
where withdrawal defaults are not being explicity added in and hence resulting into the block building failure in shanghai with txs.

### DIFF
--- a/packages/block/src/block.ts
+++ b/packages/block/src/block.ts
@@ -260,10 +260,11 @@ export class Block {
     withdrawals?: Withdrawal[]
   ) {
     this.header = header ?? BlockHeader.fromHeaderData({}, opts)
-    this.transactions = transactions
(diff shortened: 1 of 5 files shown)


acolytec3 (approved):
LGTM!  Glad we're providing a default for withdrawals now.

holgerd77:
Yes, would agree, this is better UX-wise, cool!

PR #2469, opened 2023-01-03
common: add eip option to GethConfigOpts

gabrocheleau:
This adds the EIP option to the GethConfigOpts. This change makes the workflow a bit easier for verkle trees work (can set directly the EIPs with the `Common.fromGethConfig` helper instead of using `.setEIPs` in a second step) so I'm extracting this in a separate PR. I think this could also be useful for further testnet work.


ScottyPoi (approved):
LGTM

Issue #2414, opened 2022-11-08
`fromGethGenesis` sets `DEFAULT_HARDFORK` to `merge` when no `hardfork` is provided.


tenthirtyone:
There may be a small bug in the way `fromGethGenesis` sets the default hardfork when creating a new `Common`. 

The `parseGethGenesis` function it relies on to convert from `genesis.json` into `CommonOpts` relies on a [forkmap](https://github.com/ethereumjs/ethereumjs-monorepo/blob/5738e2762b260df279a26d924d12348ce8a9d3ce/packages/common/src/utils.ts#L83) to determine which hardforks are available to the chain.

Providing `fromGethGenesis` with a hardfork that is not defined in `genesis.json` will, correctly, throw an error. E.g. specifying `london` with a matching `londonBlock` in `genesis.json` throws. However, if a `hardfork` is not provided Common's [constructor](https://github.com/ethereumjs/ethereumjs-monorepo/blob/9d470d336fa2cd606dbac31c4c6f8a6a1981f878/packages/common/src/common.ts#L220) sets `DEFAULT_HARDFORK` to `merge`, regardless of whether a corresponding `mergeForkBlock` was parsed from `genesis.json`

I believe the default behavior to choose the most recent fork makes sense. Since the available hardforks are defined by `genesis.json`  when using `fromGethGenesis` I think it might be better for the default hardfork value to be set to the most recent hardfork defined in `genesis.json` by block height.  

To reproduce add these tests to https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/test/utils.spec.ts 

``` javascript
  t.test('Throws', (t) => {
    const json = require(`../../client/test/testdata/geth-genesis/testnet.json`)

    // This is allowed - but no mergeForkBlock exists in testnet.json
    const common1 = Common.fromGethGenesis(json, {
      chain: 'customChain1',
    })

    //common1.hardfork() === 'merge'

    // This will throw - Hardfork with name merge not supported
    const common2 = Common.fromGethGenesis(json, {
      chain: 'customChain2',
      hardfork: 'merge',
    })


  })

  t.test('Not equal', (t) => {
    const json = require(`../../client/test/testdata/geth-genesis/testnet.json`)

    const common1 = Common.fromGethGenesis(json, {
      chain: 'customChain1',
    })

    const common2 = Common.fromGethGenesis(json, {
      chain: 'customChain2',
      hardfork: 'london',
    })

[...]

PR #2470, opened 2023-01-04
common: infer latest hardfork from geth genesis file

gabrocheleau:
This PR solves #2414 by attempting to get the latest hardfork from the geth genesis file as a fallback when the `fromGethGenesis` constructor isn't provided with an explicit hardfork. It also adds some type improvements to the geth genesis helpers.

### DIFF
--- a/packages/common/src/common.ts
+++ b/packages/common/src/common.ts
@@ -183,7 +183,7 @@ export class Common extends EventEmitter {
       chain: genesisParams.name ?? 'custom',
       customChains: [genesisParams],
       eips,
-      hardfork,
+      hardfork: hardfork ?? genesisParams.hardfork,
     })
     if (genesisHash !== undefined) {
       common.setForkHashes(genesisHash)
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -19,9 +19,9 @@ export interface ChainConfig {
   name: string
   chainId: number | bigint
   networkId: number | bigint
-  defaultHardfork: string
-  comment: string
-  url: string
+  defaultHardfork?: string
+  comment?: string
+  url?: string
   genesis: GenesisBlockConfig
   hardforks: HardforkConfig[]
   bootstrapNodes: BootstrapNodeConfig[]
--- a/packages/common/src/utils.ts
+++ b/packages/common/src/utils.ts
@@ -156,10 +156,6 @@ function parseGethParams(json: any, mergeForkIdPostMerge: boolean = true) {
     return (a.timestamp ?? genesisTimestamp) - (b.timestamp ?? genesisTimestamp)
   })
 
-  const latestHardfork = params.hardforks.length > 0 ? params.hardforks.slice(-1)[0] : undefined
-  params.hardfork = latestHardfork?.name
-  params.hardforks.unshift({ name: Hardfork.Chainstart, block: 0 })
-
   if (config.terminalTotalDifficulty !== undefined) {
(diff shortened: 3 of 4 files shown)


holgerd77 (commented):
Hi there, cool, thanks for submission! 😄 

Haven't completely followed the instantiation and (default) HF setting path, so just to make sure this is on the right track regarding HF setting semantics (also see this comment https://github.com/ethereumjs/ethereumjs-monorepo/issues/2414#issuecomment-1306856429), rules should be:

If a Merge HF (respectively: the default HF from Common, so this should be implemented in a generic way and not break once we change the default HF to something else) is present, HF should be set to the Merge HF. This is to align the behavior of this constructor with the other constructors we have. If a Merge HF (so, counts for the above as well: in the Geth config file) is *not* present, the last HF from the Geth config file should be taken.

First review comment:

Can you add some test cases here by adding the respective Geth config files, use the static constructor and test on the HF set?

Minimally the following cases should be covered I guess:

1. "Normal" config file with subsequent block numbers + Merge HF included
2. Config file with subsequent block numbers + Merge HF not included
3. All block numbers 0 + Merge HF included
4. All block numbers 0 + Merge HF not included

Might make sense to add 1-2 additional mixed cases as well (only Merge or post-Merge HF changing on the block number e.g.). Post-Merge HF case should generally be tested as well (so: a Geth config file with a HF after the Merge).

I guess with these test cases we can also live with this fork map reliance regarding the 0 block number HFs? 🤔

g11tech:
> Hi there, cool, thanks for submission! smile
> 
> Haven't completely followed the instantiation and (default) HF setting path, so just to make sure this is on the right track regarding HF setting semantics (also see this comment [#2414 (comment)](https://github.com/ethereumjs/ethereumjs-monorepo/issues/2414#issuecomment-1306856429)), rules should be:
> 
> If a Merge HF (respectively: the default HF from Common, so this should be implemented in a generic way and not break once we change the default HF to something else) is present, HF should be set to the Merge HF. This is to align the behavior of this constructor with the other constructors we have. If a Merge HF (so, counts for the above as well: in the Geth config file) is _not_ present, the last HF from the Geth config file should be taken.
> 
> First review comment:
> 
> Can you add some test cases here by adding the respective Geth config files, use the static constructor and test on the HF set?
> 
> Minimally the following cases should be covered I guess:
> 
>     1. "Normal" config file with subsequent block numbers + Merge HF included
> 
>     2. Config file with subsequent block numbers + Merge HF not included
> 
>     3. All block numbers 0 + Merge HF included
> 
>     4. All block numbers 0 + Merge HF not included
> 
> 
> Might make sense to add 1-2 additional mixed cases as well (only Merge or post-Merge HF changing on the block number e.g.). Post-Merge HF case should generally be tested as well (so: a Geth config file with a HF after the Merge).
> 
> I guess with these test cases we can also live with this fork map reliance regarding the 0 block number HFs? thinking

if merge hardfork is not present, then should be set to last hardfork or the one before merge?

g11tech:
also assuming we generalize this to whatever the repo's current default hardfork is (which is merge for now)
and hence testCases for that

holgerd77:
> if merge hardfork is not present, then should be set to last hardfork or the one before merge?

I would say last hardfork makes the most sense? 🤔 Since this is likely "what people want" [TM] (respectively: expect) in that case?

g11tech (approved):
I see you went with the last scheduled hardfork, i guess its a valid resolution. lgtm

PR #2474, opened 2023-01-07
client: reuse jwt-token from default path

### LINKED ISSUE #2473: Client: Reuse JWT secret if present in the default location
The current JWT creation behavior if the `--jwt-secret` flag is *not* provided is that there is always a new secret generated (or at least the log says so with `Wrote a hex encoded random jwt secret to path=../../../ethereumjs-datadir/jwtsecret` even if there is a secret present in the default location.

I think we should switch here to only generate if *not* present and otherwise reuse the existing secret, since this is what one would expect and this is also more user friendly, there should still be a log output though (`Using secret from default location [LOCATION]` or something similar). If people want to regenerate (which is rather the exception case) they can still manually delete the file.

gabrocheleau:
Resolves #2473

### DIFF
--- a/packages/client/bin/startRpc.ts
+++ b/packages/client/bin/startRpc.ts
@@ -1,4 +1,4 @@
-import { existsSync, readFileSync, writeFileSync } from 'fs'
+import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'fs'
 
 import { RPCManager, saveReceiptsMethods } from '../lib/rpc'
 import * as modules from '../lib/rpc/modules'
@@ -58,8 +58,13 @@ function parseJwtSecret(config: Config, jwtFilePath?: string): Buffer {
     )
     jwtSecret = Buffer.from(jwtSecretHex, 'hex')
   } else {
+    const folderExists = existsSync(config.datadir)
+    if (!folderExists) {
+      mkdirSync(config.datadir, { recursive: true })
+    }
+
     jwtSecret = Buffer.from(Array.from({ length: 32 }, () => Math.round(Math.random() * 255)))
-    writeFileSync(defaultJwtPath, jwtSecret.toString('hex'))
+    writeFileSync(defaultJwtPath, jwtSecret.toString('hex'), {})
     config.logger.info(`Wrote a hex encoded random jwt secret to path=${defaultJwtPath}`)
   }
   return jwtSecret
--- a/packages/client/bin/startRpc.ts
+++ b/packages/client/bin/startRpc.ts
@@ -1,4 +1,4 @@
-import { readFileSync, writeFileSync } from 'fs'
+import { existsSync, readFileSync, writeFileSync } from 'fs'
 
 import { RPCManager, saveReceiptsMethods } from '../lib/rpc'
 import * as modules from '../lib/rpc/modules'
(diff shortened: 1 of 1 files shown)


packages/client/bin/startRpc.ts
```diff
@@ -33,24 +33,28 @@ type RPCArgs = {
...
+ * Returns a jwt secret from a provided file path, otherwise saves a randomly generated one to datadir if none already exists
  */
 function parseJwtSecret(config: Config, jwtFilePath?: string): Buffer {
-  let jwtSecret
-  if (jwtFilePath !== undefined) {
-    const jwtSecretContents = readFileSync(jwtFilePath, 'utf-8').trim()
+  let jwtSecret: Buffer
+  const defaultJwtPath = `${config.datadir}/jwtsecret`
+  if (jwtFilePath !== undefined || existsSync(defaultJwtPath)) {
+    const jwtSecretContents = readFileSync(jwtFilePath ?? defaultJwtPath, 'utf-8').trim()
```
jochem-brouwer, line 42: Check if `jwtFilePath` exists? I don't know what it will throw otherwise, but probably a less-descriptive error (user might not know that this jwt does not exist)

packages/client/bin/startRpc.ts
```diff
@@ -33,24 +33,28 @@ type RPCArgs = {
...
+        jwtFilePath !== undefined ? `path=${jwtFilePath}` : `default path=${defaultJwtPath}`
+      }`
+    )
     jwtSecret = Buffer.from(jwtSecretHex, 'hex')
   } else {
-    jwtFilePath = `${config.datadir}/jwtsecret`
     jwtSecret = Buffer.from(Array.from({ length: 32 }, () => Math.round(Math.random() * 255)))
-    writeFileSync(jwtFilePath, jwtSecret.toString('hex'))
-    config.logger.info(`Wrote a hex encoded random jwt secret to path=${jwtFilePath}`)
+    writeFileSync(defaultJwtPath, jwtSecret.toString('hex'))
```
jochem-brouwer, line 56: Does this also write if the directory/directories do not exist?

jochem-brouwer (commented):
Two comments.

I also think this needs a test..?

packages/client/bin/startRpc.ts
gabrocheleau: Good point, I think that if jwtFilePath is provided but does not exist, we should throw (and not default to the other one, as that would likely confuse users).

gabrocheleau: Resolved!

jochem-brouwer: Just checked, if file does not exist, you get an error: 

```
fs.readFileSync('abacdsadf.ts')
Uncaught Error: ENOENT: no such file or directory, open 'abacdsadf.ts'
    at Object.openSync (node:fs:585:3)
    at Object.readFileSync (node:fs:453:35) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: 'abacdsadf.ts'
```

(But for the user, it might not be directly clear that this is their `jwt` stuff, so it would be more descriptive to just throw `jwt: file does not exists`, or something like that)

packages/client/bin/startRpc.ts
jochem-brouwer: It does not:
```
fs.writeFileSync('abc/def/efg/out.ts', 'abc')
Uncaught Error: ENOENT: no such file or directory, open 'abc/def/efg/out.ts'
    at Object.openSync (node:fs:585:3)
    at Object.writeFileSync (node:fs:2170:35) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: 'abc/def/efg/out.ts'
```

packages/client/bin/startRpc.ts
```diff
@@ -38,7 +38,13 @@ type RPCArgs = {
...
   let jwtSecret: Buffer
   const defaultJwtPath = `${config.datadir}/jwtsecret`
-  if (jwtFilePath !== undefined || existsSync(defaultJwtPath)) {
+
+  // If jwtFilePath is provided, it should exist
+  if (typeof jwtFilePath === 'string' && !existsSync(jwtFilePath)) {
+    throw new Error(`No file exists at provided jwt secret path=${jwtFilePath}`)
+  }
+
+  if (typeof jwtFilePath === 'string' || existsSync(defaultJwtPath)) {
```
jochem-brouwer, line 47: Can it `existsSync` in case that type of `jwtFilePath` is not a string?

gabrocheleau: I might be misunderstanding your comment, but the `existsSync` condition refers to the `defaultJwtPath`, while the first just checks if the user has provided a path. So this basically checks "has the user provided a path? if not, is there a file at the defaultJwtPath?"

packages/client/bin/startRpc.ts
gabrocheleau: Oh nice catch, this means that this logic actually throws in the master branch. Will address.

gabrocheleau: Added logic to ensure that the dir exists before writing

packages/client/bin/startRpc.ts
gabrocheleau: This should not happen with the latest update though right? Since we're only reading if either the first exists, or the second exists.

gabrocheleau:
Thanks for reviewing @jochem-brouwer. I think I've addressed everything, should be ready for re-review

holgerd77:
Rebased this via UI.

packages/client/bin/startRpc.ts
holgerd77: Given that we here had yet another case where we were close to accidentally introduce something structurally client breaking (with the last time having a bug preventing mainnet from sync which actually made it into the release), we should really still look at increase client integration testing. This is otherwise too fragile if one needs to manually do `npm run client:start` before release for a pre-release test and this is breaking (from my historical experiences) in 15-25% of the trials (and not even catching everything like this one here).

packages/client/bin/startRpc.ts
```diff
@@ -33,24 +33,39 @@ type RPCArgs = {
...
     const jwtSecretHex = hexPattern.exec(jwtSecretContents)?.groups?.jwtSecret
     if (jwtSecretHex === undefined || jwtSecretHex.length !== 64) {
       throw Error('Need a valid 256 bit hex encoded secret')
     }
-    config.logger.debug(`Read a hex encoded jwt secret from path=${jwtFilePath}`)
+    config.logger.debug(
+      `Read a hex encoded jwt secret from ${
+        typeof jwtFilePath === 'string' ? `path=${jwtFilePath}` : `default path=${defaultJwtPath}`
+      }`
+    )
```
holgerd77, line 58: Ah, that's a nice and welcome enhancement on the logging here! 🙂

holgerd77 (approved):
Ok, nice, cool, LGTM! 👍 

Will merge.

PR #2475, opened 2023-01-07
client: typed client opts

### LINKED ISSUE #2415: `createClient` opts are not typed
The [`clientOpts`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/9d470d336fa2cd606dbac31c4c6f8a6a1981f878/packages/client/test/rpc/helpers.ts#L59) for `createClient` in the rpc test helper are not typed and kind of a black box.  We should define an interface/type for these opts so we can understand what each test is doing when using this helper.

A good solution will be a new `clientOpts` interface with all optional properties where every option is defined that's referenced in `createClient`.

gabrocheleau:
Resolves #2415

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -21,6 +21,8 @@ import { helprpc, startRPCServers } from './startRpc'
 
 import type { Logger } from '../lib/logging'
 import type { FullEthereumService } from '../lib/service'
+import type { ClientOpts } from '../lib/types'
+import type { RPCArgs } from './startRpc'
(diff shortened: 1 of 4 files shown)


acolytec3:
Self-serving but this looks great to me.  @gabrocheleau, care to look at the changes I added and thumbs up/thumbs down them so we can approve and merge?

acolytec3:
As a note, I ran the client on mainnet and using various combinations of the `dev` and `mine` commands to validate that stuff was working correctly.

holgerd77:
Rebased this via UI.

holgerd77 (approved):
LGTM

PR #2476, opened 2023-01-09
rlp: avoid copy in encode

wemeetagain:
Avoid an intermediate copy when `encode` is called with an Array.


acolytec3 (approved):
LGTM, thanks for the catch!

wemeetagain:
Curious when this would land in a release?

acolytec3:
> Curious when this would land in a release?

Unfortunately, hard to say exactly. Might be able to get a point release out in a couple of weeks but it depends on what other PRs we've merged that might need to get included and what all of our downstream packages also have to be released. Is this urgent for lodestar?

wemeetagain:
not urgent at all

PR #2477, opened 2023-01-10
client: Clean stop sync and execution to allow client shutdown

g11tech:
If the client is midway syncing, sending shutdown signal to the client via Ctrl + C for e.g. wouldn't clean stop the sync or execution leading to client hanging mid way (or sometimes just keep processing the jobs it has in queue)

This PR fixes the same by clean stopping the syncronizer and execution and resolves the problem with proper stopping on client

### DIFF
--- a/packages/client/lib/execution/vmexecution.ts
+++ b/packages/client/lib/execution/vmexecution.ts
@@ -92,6 +92,10 @@ export class VMExecution extends Execution {
    */
   async open(): Promise<void> {
     return this.runWithLock<void>(async () => {
+      // id already opened or stopping midway
+      if (this.started || this.vmPromise !== undefined) {
+        return
+      }
       await this.vm.init()
       if (typeof this.vm.blockchain.getIteratorHead !== 'function') {
         throw new Error('cannot get iterator head: blockchain has no getIteratorHead function')
@@ -111,6 +115,7 @@ export class VMExecution extends Execution {
         }
         await this.vm.eei.generateCanonicalGenesis(this.vm.blockchain.genesisState())
       }
+      await super.open()
       // TODO: Should a run be started to execute any left over blocks?
       // void this.run()
     })
@@ -185,7 +190,7 @@ export class VMExecution extends Execution {
    * @returns number of blocks executed
    */
   async run(loop = true, runOnlybatched = false): Promise<number> {
-    if (this.running) return 0
+    if (this.running || !this.started) return 0
     this.running = true
     let numExecuted: number | undefined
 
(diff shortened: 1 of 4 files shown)


acolytec3 (commented):
General question on this: what is the difference between `vmexecution.running` and `vmexecution.started`?  I've been trying to decipher from the code but it's just not apparent to me.

g11tech:
> General question on this: what is the difference between `vmexecution.running` and `vmexecution.started`? I've been trying to decipher from the code but it's just not apparent to me.

running is when the execution is actually running in the `run()` function, but i have added started to see if its the execution has been stopped or not so as to exit out of the `run` which could be looping over all the pending blocks for execution

g11tech:
rebased via UI

packages/client/lib/execution/vmexecution.ts
```diff
@@ -92,6 +92,10 @@ export class VMExecution extends Execution {
    */
   async open(): Promise<void> {
     return this.runWithLock<void>(async () => {
+      // id already opened or stopping midway
```
g11tech, line 95:
```suggestion
      // if already opened or stopping midway
```

holgerd77:
It seems there is now a "double stopping" - at least as seen from the logs - tested on the (I guess, also concluding from the log state) not-in-syncing-progress state.

So here is the log with the code from this PR:

grafik

(see the doubled "Stopped execution/synchronization" output)

Here in comparison from before (took client from latest release):

grafik

Can you have a look? These kind of things often points to something structurally not clean which then can cause other problems as well.

holgerd77:
(some comment updates in the browser, please read on site)

g11tech:
> grafik
> 
> Can you have a look? These kind of things often points to something structurally not clean which then can cause other problems as well.

I think in the first image you did two ctrl + c since it has two lines of "Caught interrupt signal. Shutting down..." 
so the stop was processed for each signal.

Here is a screen shot of execution stopping mid way with one ctrl + c
image

holgerd77:
Have updated this via UI.

@g11tech just retested, no, I actually only pressed once, double tested this again. Not sure, if you have no time to further investigate we can also merge this in in a first round, double-closing seems better than no closing at all. 😜 At least we should have this in for the interop.

g11tech:
> Have updated this via UI.
> 
> @g11tech just retested, no, I actually only pressed once, double tested this again. Not sure, if you have no time to further investigate we can also merge this in in a first round, double-closing seems better than no closing at all. stuck_out_tongue_winking_eye At least we should have this in for the interop.

weird, for me it just closes with one log. Are your runs logging :
`"Caught interrupt signal. Shutting down..."`
once or twice? if twice it seems two signals are being send to the client.

holgerd77 (approved):
Ah, I'll merge this in for now (yeah, I am having two "Caught interrupt signal."s).

If annoying or something we can side-investigate further. 😋

PR #2479, opened 2023-01-10
Fix: typos

omahs:
Fix: typos


acolytec3 (approved):
Thanks for the PR!  I put one of your corrections in the original jsdoc instead of the interfaces file since that's a generated file.  Will merge once the CI runs.

PR #2481, opened 2023-01-11
Fix beaconsync test race condition

### LINKED ISSUE #2480: Beaconsync tests often fail on CI
[This beacon sync test](https://github.com/ethereumjs/ethereumjs-monorepo/blob/70dbdd132f284fa25068c2766e8c761e50f837bf/packages/client/test/sync/beaconsync.spec.ts#L148) intermittently fails on CI because it's a time dependent test that inspects the contents of the fetcher while it's in the process of running (I think).  We should ideally rewrite these tests to not use time dependent validation so we don't have random failures in our CI.

acolytec3:
Fixes #2480

### DIFF
--- a/packages/client/test/sync/beaconsync.spec.ts
+++ b/packages/client/test/sync/beaconsync.spec.ts
@@ -110,7 +110,8 @@ tape('[BeaconSynchronizer]', async (t) => {
     t.end()
   })
 
-  t.test('should sync to next subchain head or chain height', async (t) => {
+  t.test('should sync to next subchain head or chain height', async (st) => {
+    st.plan(3)
     const config = new Config({
       transports: [],
       safeReorgDistance: 0,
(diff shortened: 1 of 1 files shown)


g11tech (approved):
lgtm!

packages/client/test/sync/beaconsync.spec.ts
```diff
@@ -137,26 +137,31 @@ tape('[BeaconSynchronizer]', async (t) => {
...
-    void sync.sync()
-    await wait(50)
-    t.equal(sync.fetcher!.first, BigInt(5), 'should sync block 5 and 4')
-    t.equal(sync.fetcher!.count, BigInt(5), 'should target syncing all the way to chain')
-    await wait(51)
+    sync.config.logger.addListener('data', (data: any) => {
+      if ((data.message as string).includes('first=5 count=5'))
+        st.pass('should sync block 5 and target chain start')
+    })
+    await sync.sync()
```
holgerd77, line 144: Ah, these kind of robustness fixes are always super nice!! 🤗 👍

PR #1842, opened 2022-04-07
Utils: Amount formatter (new)

holgerd77:
Cherry-picked re-submission of #1794 after `develop` rebase, see #1835.

PR #2483, opened 2023-01-13
util: Change withdrawal amount representation for Wei to Gwei

g11tech:
util: Change withdrawal amount representation for Wei to Gwei

ref:
- https://github.com/ethereum/EIPs/pull/6325/files


holgerd77 (changes requested):
Hmm, given that we had these endless discussions and drafts on unit conversion, latest with #1842, I would feel uneasy if we now side introduce such a constant `GWEI_TO_WEI`, and then even in the `withdrawals` Util module.

On the other hand I agree that this is super-helful, even in this raw "just the factor" constant. 😋

Can we at least move this to a dedicated new module `units`, where we can eventually expand from, maybe add 1-2 other small helper methods e.g. (and level down a bit from our previous ambitions)?

That would feel better to me, having this in `withdrawals` would cause some level of technical debt - since people might directly import - if we finally want to expand a bit on the topic and include in a release.

g11tech:
> Hmm, given that we had these endless discussions and drafts on unit conversion, latest with #1842, I would feel uneasy if we now side introduce such a constant `GWEI_TO_WEI`, and then even in the `withdrawals` Util module.
> 
> On the other hand I agree that this is super-helful, even in this raw "just the factor" constant. yum
> 
> Can we at least move this to a dedicated new module `units`, where we can eventually expand from, maybe add 1-2 other small helper methods e.g. (and level down a bit from our previous ambitions)?
> 
> That would feel better to me, having this in `withdrawals` would cause some level of technical debt - since people might directly import - if we finally want to expand a bit on the topic and include in a release.

sounds good!

g11tech:
@holgerd77 i have created a minimal module `@ethereumjs/units` and move the `GWEI_TO_WEI` constant there, for now i guess this should be sufficient and can extend withdrawal along the way if needed

g11tech:
rebased via ui

holgerd77:
Rebased this via UI

holgerd77 (approved):
Looks good now! 👍

PR #2486, opened 2023-01-15
client: Apply correct hf to peer fetched txes as well as filter and remove mismatching hf txs while building blocks

g11tech:
In the devnet-2 ethereumjs produced blocks which were rejected by the peers, on debugging by @jochem-brouwer it was deduced that the hf of the added txs in the block didn't seem correct.

On debugging, figured out there was a way txes can be fetched from txhash annoucement using eth getPooledTransactions, in which hardfork was incorrectly being set as proper chain opts were not being used.

This PR fixes the same as well as also adds filteration and removal of txs which do not match the  block hf being build.

TODO: test cases for coverage


holgerd77:
Thanks a lot for taking on this! 🙏

packages/vm/src/runTx.ts
```diff
@@ -33,6 +33,21 @@ export async function runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
...
+      // update the tx as per new hardfork
+      const msg = _errorMsg('tx has a different hardfork than the vm', this, opts.block, opts.tx)
+      throw new Error(msg)
+    }
+    if (opts.block._common.hardfork() !== this._common.hardfork()) {
+      // Block and VM's hardfork should match as well
+      const msg = _errorMsg('block has a different hardfork than the vm', this, opts.block, opts.tx)
+      throw new Error(msg)
+    }
+  }
```
holgerd77, line 49: I was first irritated by this new `skipHardForkValidation` option but now that I am getting is and see how this is applied I find this a super-great solution to guard the current behavior under this flag and at the same time to allow to go narrower and deactivate, eventually one by one respectively over time! 🤩

Really nice, since we have talked about this a lot already but didn't dare yet to just plainly introduce these checks, this is a super nice and gentle way to finally do! 🙏 👍

packages/client/lib/net/protocol/ethprotocol.ts
```diff
@@ -225,13 +225,21 @@ export class EthProtocol extends Protocol {
...
+      decode: ([reqId, txs]: [Buffer, any[]]) => {
+        const common = this.config.chainCommon.copy()
+        common.setHardforkByBlockNumber(
+          this.chain.headers.latest?.number ?? // Use latest header number if available OR
+            this.config.syncTargetHeight ?? // Use sync target height if available OR
+            common.hardforkBlock(common.hardfork()) ?? // Use current hardfork block number OR
+            BigInt(0), // Use chainstart,
+          undefined,
+          this.chain.headers.latest?.timestamp ?? Math.floor(Date.now() / 1000)
+        )
```
holgerd77, line 237: 👍 

Ok, this will definitely be an improvement.

packages/client/lib/miner/pendingBlock.ts
```diff
@@ -148,14 +170,28 @@ export class PendingBlock {
...
+          // but for simplicity just remove the tx as the sender can always retransmit
+          // the tx
+          this.txPool.removeByHash(txs[index].hash().toString('hex'))
+          this.config.logger.error(
+            `Pending: Removed from txPool tx 0x${txs[index]
+              .hash()
+              .toString('hex')} having different hf=${txs[
+              index
+            ].common.hardfork()} than block vm hf=${vm._common.hardfork()}`
+          )
```
holgerd77, line 194: 👍

holgerd77 (approved):
Really cool, I wasn't aware that this will get such a profoundly-improving-PR 🙏, so nice, this adds so much additional control around this HF-matching problem, also love the way this new `skipHardforkValidation` option is introduced and used to slowly tighten on this (as mentioned)! 🤩 ❤️

Will merge.

packages/vm/src/runTx.ts
g11tech: yea! i was also apprehensive to add this but then realized it was a more progressive way to go for our tests! glad you feel the same :slightly_smiling_face:

PR #2487, opened 2023-01-17
Update Dockerfiles to use node 18

g11tech:
Using node18 locally and having no issues, it makes sense to update the dockerfiles to use node18


holgerd77 (approved):
LGTM

PR #2488, opened 2023-01-18
util: Add ssz roots capability for withdrawals hash tree root

g11tech:
In preparation for moving the ssz root gen capability for withdrawals, this PR adds ssz types and root gen utility method which can be later easily deployed

will just need to switch all Block.genWithdrawlsTrieRoot to Block.genWithdrawalsSSZRoot method when the spec moves to ssz root

### DIFF
--- a/packages/block/src/block.ts
+++ b/packages/block/src/block.ts
@@ -54,7 +54,7 @@ export class Block {
   }
 
   /**
-   * Returns the withdrawals ssz root for array of Withdrawal.
+   * Returns the ssz root for array of withdrawal transactions.
    * @param wts array of Withdrawal to compute the root of
    */
   public static async genWithdrawalsSSZRoot(wts: Withdrawal[]) {
--- a/packages/block/src/block.ts
+++ b/packages/block/src/block.ts
@@ -11,6 +11,7 @@ import {
   bufferToHex,
   intToHex,
   isHexPrefixed,
+  ssz,
 } from '@ethereumjs/util'
 import { keccak256 } from 'ethereum-cryptography/keccak'
 import { ethers } from 'ethers'
@@ -50,6 +51,14 @@ export class Block {
       await trie.put(Buffer.from(RLP.encode(i)), arrToBufArr(RLP.encode(wt.raw())))
     }
     return trie.root()
+  }
+
+  /**
+   * Returns the withdrawals ssz root for array of Withdrawal.
+   * @param wts array of Withdrawal to compute the root of
+   */
+  public static async genWithdrawalsSSZRoot(wts: Withdrawal[]) {
+    ssz.Withdrawals.hashTreeRoot(wts.map((wt) => wt.toValue()))
   }
 
   /**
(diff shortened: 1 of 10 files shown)


packages/util/src/withdrawal.ts
```diff
@@ -103,6 +103,15 @@ export class Withdrawal {
     return Withdrawal.toBufferArray(this)
   }
 
+  toValue() {
```
acolytec3, line 106: This is nice!  I wonder if we should add something like this for blob txns?

packages/util/src/ssz.ts
```diff
@@ -0,0 +1,24 @@
...
+  ContainerType,
+  ListCompositeType,
+  UintBigintType,
+  UintNumberType,
+} from '@chainsafe/ssz'
+
+import { MAX_WITHDRAWALS_PER_PAYLOAD } from './constants'
+
+export const UintNum64 = new UintNumberType(8)
+export const UintBn64 = new UintBigintType(8)
```
acolytec3, line 12: Can we changes this to `UintBigInt64`?  I don't like `BN` references.

packages/block/src/block.ts
```diff
@@ -52,6 +53,14 @@ export class Block {
     return trie.root()
   }
 
+  /**
+   * Returns the withdrawals ssz root for array of Withdrawal.
```
acolytec3, line 57:
```suggestion
   * Returns the ssz root for array of withdrawal transactions.
```

packages/block/src/block.ts
```diff
@@ -52,6 +53,14 @@ export class Block {
...
+   * @param wts array of Withdrawal to compute the root of
+   */
+  public static async genWithdrawalsSSZRoot(wts: Withdrawal[]) {
```
acolytec3, line 60:
```suggestion
  public static async generateWithdrawalsSSZRoot(withdrawals: Withdrawal[]) {
```

The naming convention for the short version isn't 100% clear to me.  Prefer this.

acolytec3 (changes requested):
Basically nits but would like some of the naming conventions a little bit clearer.  Code looks great though!

packages/util/src/withdrawal.ts
g11tech: yes its good :+1: picked idea from ssz code structure

packages/util/src/ssz.ts
g11tech: ok sounds good :+1:

acolytec3 (approved):
Looks great!

packages/util/test/withdrawal.spec.ts
```diff
@@ -83,12 +83,18 @@ tape('Withdrawal', (t) => {
...
     )
     const withdrawalsJson = withdrawals.map((wt) => wt.toJSON())
     st.deepEqual(withdrawalsGethVector, withdrawalsJson, 'Withdrawals json should match')
+
+    const withdrawalsValue = withdrawals.map((wt) => wt.toValue())
+    st.deepEqual(
+      withdrawalsValue.map((wt) => `0x${wt.address.toString('hex')}`),
+      withdrawalsJson.map((wt) => wt.address)
+    )
     st.end()
```
HBIT-2A2C-JRAX-HAWJ-9DRCK, line 98: 0x990De5Cc9Cc644F0703f6A09e084470a949eC370

PR #2489, opened 2023-01-18
Fix bug in how precompiles activated by EIP are identified

acolytec3:
Fixes a bug in the EVM where precompiles that are activated by EIP might not be activated if their trigger EIP is activated on a hardfork rather than by passing an EIP into the `common` constructor.  Was discovered while working on 4844.


jochem-brouwer:
Ah right looks good to me. Maybe add a test?

holgerd77 (dismissed):
I'll give this a go here also without an additional test case, I guess at the moment we should prioritize getting our code base ready and updated.

Test case would be nice, but this is also very much implicitly tested by precompile calling, and when the PR then additionally fixes stuff in the 4844 branch it is basically "proven" [TM] that the status quo has been improved by this.

acolytec3:
Could I get another approval on this one? The actual change in the EVM was also included in the 4844 PR but we should still move these 2537 tests over to the `evm` folder so would like to go ahead and merge this.

g11tech (approved):
lgtm

PR #2490, opened 2023-01-19
client: validate CLI args

gabrocheleau:
This PR makes the CLI args parsing strict, meaning that unknown args will result in yargs throwing instead of them being ignored.

`npm run client:start -- --invalidOption` now results in:

image

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -284,7 +284,9 @@ const args: ClientOpts = yargs(hideBin(process.argv))
     describe:
       'Block number to start syncing from. Must be lower than the local chain tip. Note: this is destructive and removes blocks from the blockchain, please back up your datadir before using.',
     number: true,
-  }).argv
+  })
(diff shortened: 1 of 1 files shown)


gabrocheleau:
@holgerd77, do you think it's worth writing a test for this? I'm tempted to leave testing out, as we'd basically just be testing `yargs` behavior.

packages/client/bin/cli.ts
```diff
@@ -284,7 +284,9 @@ const args: ClientOpts = yargs(hideBin(process.argv))
     describe:
       'Block number to start syncing from. Must be lower than the local chain tip. Note: this is destructive and removes blocks from the blockchain, please back up your datadir before using.',
     number: true,
-  }).argv
+  })
+  // strict() ensure that yargs throws when an invalid arg is provided
```
gabrocheleau, line 288:
```suggestion
  // strict() ensures that yargs throws when an invalid arg is provided
```

PR #2493, opened 2023-01-20
Interop 4844 Docs Preparation & Alignment

holgerd77:
Currently generalizing the 4844 devnet setup and docs a bit by introducing a dedicated folder and document structure which is a bit generalized to easier "grow" with additional client setup instructions, also to align things with the Shapella interop notes.

Generally and for the interop I would really like to get client runs out of this "script magic" (actually, it's not so much magic, but out of this automation level), this is to get more flexible on the runs itself (best to run with the client start commands directly to easier e.g. add a `debug` flag or something), also to allow to directly set up - optimally - on the orignal set of testnet configuration files, which will likely change and evolve during interop.

For the EthereumJS client that's easy for 4844, will just remove the script since it is not doing a lot.

For prysm not 100% sure yet, but from a first look maybe just keep the replacement part in the script - since this might be useful - and move the start commands over to documentation.


holgerd77:
This will align with the following setup instructions: #2494

g11tech (dismissed):
looks good!

holgerd77:
Rebased this via UI. Will merge this in via admin-merge after test runs finished for procedural reasons, no need to keep this separate and instructions and everything will just keep to evolve.

(or: short-term approval also welcomed!)

PR #2499, opened 2023-01-24
tx: Handle json fields in fromTxData for eip4844 tx

g11tech:
Handle json fields in fromTxData for eip4844 tx

A tx provided to test was not getting being loaded and giving errors on serialization as maxDataFee was still a string. This PR fixes it. 
It also adds a test for hash, which currently doesn't seem to be correct, has been commented out to be fixed in separate issue


acolytec3 (approved):
LGTM!  Will need a follow-up PR to do proper handling of the kzgProof in `txData` since we're just disregarding that present.

PR #2500, opened 2023-01-24
Client: ensure that block builder / miner correctly switches to Shanghai after timestamp has been reached

jochem-brouwer:


packages/block/src/header.ts
```diff
@@ -178,8 +178,16 @@ export class BlockHeader {
...
-      this._common.setHardforkByBlockNumber(number, options.hardforkByTTD, timestamp)
+    if (
+      hardforkByBlockNumber ||
+      options.hardforkByTTD !== undefined ||
+      options.hardforkByTimestamp !== undefined
+    ) {
+      this._common.setHardforkByBlockNumber(
+        number,
+        options.hardforkByTTD,
+        options.hardforkByTimestamp ?? timestamp
```
acolytec3, line 189: I'm not sure this makes sense to me.  We should just pass in `options.hardforkByTimestamp` since `timestamp` will never be undefined and this will cause the first half of [this line](https://github.com/ethereumjs/ethereumjs-monorepo/blob/109d1c9b65da5e8339136295c85aa71fedbea12d/packages/common/src/common.ts#L341) in `getHardforkByBlockNumber` to always be true, right?

PR #2501, opened 2023-01-24
tx: make kzgProof required field for network wrapper

acolytec3:
Removes the internal computation of the `aggregateKzgProof` for the network wrapper version of blob transactions and requires that field to be present when instantiating a blob transaction in the network wrapper format.


holgerd77 (approved):
LGTM

PR #2502, opened 2023-01-24
client/rpc/engine: fix forkchoiceUpdateV2 shanghai

jochem-brouwer:
This PR fixes `forkchoiceUpdateV2`. See also https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#engine_forkchoiceupdatedv2

If a payload with withdrawals is thus passed to fcuV2 then timestamp should be >= Shanghai. 

This PR also fixes the hive test:

`go run ./hive.go --client ethereumjs --sim ethereum/engine --sim.limit "engine-withdrawals/Withdrawals Fork on Block 2" `

### DIFF
--- a/packages/client/lib/rpc/modules/engine.ts
+++ b/packages/client/lib/rpc/modules/engine.ts
@@ -547,6 +547,17 @@ export class Engine {
   }
 
   async newPayloadV2(params: [ExecutionPayloadV2]): Promise<PayloadStatusV1> {
+    const shanghaiTimestamp = this.chain.config.chainCommon
+      .hardforks()
+      .find((hf) => hf.name === 'shanghai')?.timestamp!
+
+    if (parseInt(params[0].timestamp) < shanghaiTimestamp) {
+      throw {
+        code: INVALID_PARAMS,
+        message: 'Cannot invoke newPayloadV2 before Shanghai is activated',
+      }
+    }
+
     return this.newPayload(params)
   }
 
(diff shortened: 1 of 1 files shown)


packages/client/lib/rpc/modules/engine.ts
```diff
@@ -748,8 +759,42 @@ export class Engine {
   }
 
   private async forkchoiceUpdatedV2(
-    params: [forkchoiceState: ForkchoiceStateV1, payloadAttributes: PayloadAttributesV2 | undefined]
+    params: [
+      forkchoiceState: ForkchoiceStateV1,
+      payloadAttributes: PayloadAttributesV1 | PayloadAttributesV2 | undefined
+    ]
   ): Promise<ForkchoiceResponseV1 & { headBlock?: Block }> {
+    console.log(params)
```
g11tech, line 767: console.log

jochem-brouwer: Ah right the linter warns instead of throws

packages/client/lib/rpc/modules/engine.ts
```diff
@@ -547,6 +547,17 @@ export class Engine {
   }
 
   async newPayloadV2(params: [ExecutionPayloadV2]): Promise<PayloadStatusV1> {
+    const shanghaiTimestamp = this.chain.config.chainCommon
+      .hardforks()
+      .find((hf) => hf.name === 'shanghai')?.timestamp!
```
holgerd77, line 552: I guess at some point we should replace these kind of retrievals with a new method `hardforkTimestamp()` analog to `hardforkBlock()` in Common (which throws if requested for a block HF).

But I guess for now and to getting these kind of fixes in quickly it is ok.

holgerd77 (approved):
Looks good, I am very much assuming that devp2p failure test is a one time timeout issue (also looking at the CI run), will admin-merge.

packages/client/lib/rpc/modules/engine.ts
ScottyPoi: @g11tech added a method with that name last month in `common.ts`

https://github.com/ethereumjs/ethereumjs-monorepo/blob/7d7b0ad0915b560093a7e1ac5802432e40f436f5/packages/common/src/common.ts#L669-L676

PR #2503, opened 2023-01-24
Add checks for replacement data gas too low for blob txs

acolytec3:
Adds one more check missed from the original 4844 implementation to check that data gas is increased by 1.1x on replacement transactions submitted via `sendRawTransaction`.


holgerd77 (approved):
Ok, cool and solid test setup, thanks a lot!

PR #2504, opened 2023-01-25
engine-api-validators

ScottyPoi:
Adding to work done in #2502

Adds more specific type labels and validators for engine api based on type descriptions in https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#engine_forkchoiceupdatedv2

hex string parameters are differentiated into bytevector length (or bitlength) giving us:
- Bytes8
- Bytes20
- Bytes32
- Root
  - *Root is a Bytes32 that represents the merkle root of a nested ssz container*
- VariableBytes32
  - 0 to 32 bytes
- Bytes48
- Bytes256
- Uint64
- Uint256
- Blob
  - because `Bytes131072` is silly

Typenames are chosen to mimic those used in SSZ typing.  They are all just `type === 'string'`.  The actual bytelength check happens in the validators

A new `validator` has been created for each type, checking:
1. type === string
2. prefix === 0x
3. bytelength === expected

#### newPayloadV2 and newPayloadV3

Both of these methods were updated according to:
https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md
and
https://github.com/ethereum/execution-apis/blob/main/src/engine/experimental/blob-extension.md

`newPayloadV2` accepts either `executionPayloadV1` or `executionPayloadV2`.
`newPayloadV3` accepts either `executionPayloadV1` or `executionPayloadV2` or `executionPayloadV3`.

The methods have been updated to handle each type, and `validators.either()` used to enable validation for multiple type possibilities

### DIFF
--- a/packages/client/lib/rpc/modules/engine.ts
+++ b/packages/client/lib/rpc/modules/engine.ts
@@ -367,7 +367,13 @@ export class Engine {
 
     this.newPayloadV3 = cmMiddleware(
       middleware(this.newPayloadV3.bind(this), 1, [
-        [validators.object(executionPayloadV3FieldValidators)],
+        [
+          validators.either(
+            validators.object(executionPayloadV1FieldValidators),
+            validators.object(executionPayloadV2FieldValidators),
+            validators.object(executionPayloadV3FieldValidators)
+          ),
+        ],
       ]),
       ([payload], response) => this.connectionManager.lastNewPayload({ payload, response })
     )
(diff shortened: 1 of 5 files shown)


packages/client/lib/rpc/modules/engine.ts
```diff
@@ -546,23 +568,65 @@ export class Engine {
...
-    const shanghaiTimestamp = this.chain.config.chainCommon
-      .hardforks()
-      .find((hf) => hf.name === 'shanghai')?.timestamp!
-
-    if (parseInt(params[0].timestamp) < shanghaiTimestamp) {
-      throw {
-        code: INVALID_PARAMS,
-        message: 'Cannot invoke newPayloadV2 before Shanghai is activated',
+  async newPayloadV2(params: [ExecutionPayloadV2 | ExecutionPayloadV1]): Promise<PayloadStatusV1> {
+    const shanghaiTimestamp = this.chain.config.chainCommon.hardforkTimestamp('shanghai')
```
jochem-brouwer, line 572: Ah I did not realize we already had this method in Common

jochem-brouwer (approved):
LGTM great tests!

holgerd77:
Thanks Scotty, so awesome overnight work, great!!! 🎉

PR #2505, opened 2023-01-25
vm: Add and use execHardfork while running a tx

g11tech:
since we don't have access to td in all places while tx is submitted, we can safely fallback to the pre merge hf for tx matching in runtx as merge doesn't introduce any changes in the tx behavior and vm run capabilities.

This PR adds the concept of "execHardfork" which ignores the mergeHf for matching in the vm. this will resolves error like these:
```typescript
[01-24|22:56:31] DEBUG [ CL ] eth_getTransactionReceipt called with params: [ '0xd5455662e76b193a84ce57d4c0a3b6fd609fdfca21cc93b93408d
e62be3c5708']
[01-24|22:56:31] DEBUG [ CL ] eth_getTransactionReceipt responded with: error: tx has a different hardfork than the vm (vm hf=merge ->
 block number=102 hash=0xeb7f045c845d787080fb73a8e64c940af443585237912b848d81d92792ce3dc3 hf=shardingFork baseFeePerGas=1219 txs=1 uncles=0 -> tx type=5 hash=0xd54556
62e76b193a84ce57d4c0a3b6fd609fdfca21cc93b93408de62be3c5708 nonce=0 value=12345678 signed=true hf=shardingFork maxFeePerGas=5000000000 maxPriorityFeePerGas=5000000000)
```

### DIFF
--- a/packages/block/src/block.ts
+++ b/packages/block/src/block.ts
@@ -569,10 +569,6 @@ export class Block {
     let hf = ''
     try {
       hf = this._common.hardfork()
-      if (hf === Hardfork.Merge) {
-        // Add info of execHardFork if relevant
-        hf = `${hf}(${this._common.execHardfork()})`
-      }
     } catch (e: any) {
       hf = 'error'
     }
--- a/packages/block/src/block.ts
+++ b/packages/block/src/block.ts
@@ -1,4 +1,4 @@
-import { ConsensusType, Hardfork } from '@ethereumjs/common'
+import { ConsensusType } from '@ethereumjs/common'
 import { RLP } from '@ethereumjs/rlp'
 import { Trie } from '@ethereumjs/trie'
 import { BlobEIP4844Transaction, Capability, TransactionFactory } from '@ethereumjs/tx'
--- a/packages/block/src/block.ts
+++ b/packages/block/src/block.ts
@@ -1,4 +1,4 @@
-import { ConsensusType } from '@ethereumjs/common'
+import { ConsensusType, Hardfork } from '@ethereumjs/common'
 import { RLP } from '@ethereumjs/rlp'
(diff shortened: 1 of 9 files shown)


packages/common/src/common.ts
```diff
@@ -953,6 +961,15 @@ export class Common extends EventEmitter {
     return this._hardfork
   }
 
+  /**
+   * Returns the hardfork excluding the merge hf which has
+   * no effect on the execution capabilities
+   * @returns Hardfork name
```
holgerd77, line 967: Could you give this a more extensive description? I am starting to "get it" 😋 but I am pretty sure "normal people" won't be able to grasp the concept yet what this method does and in what contexts it can and can't be used.

packages/block/src/block.ts
```diff
@@ -569,6 +569,10 @@ export class Block {
     let hf = ''
     try {
       hf = this._common.hardfork()
+      if (hf === Hardfork.Merge) {
+        // Add info of execHardFork if relevant
+        hf = `${hf}(${this._common.execHardfork()})`
```
acolytec3, line 574: This feels dangerous to me since the merge hardfork has plenty of functional differences for block headers and I can imagine lots of places where we want to report that the block header error stems from the hardfork being `Merge` and not `GrayGlacier` or whatever previous fork.

g11tech: yes it still retains the merge hf now it comes as: `Merge (GrayGlacier)` so the user has more context

acolytec3:
Pushed an empty commit and removed the test-all-hardforks since the shangahaiLove test that's failing has nothing to do with this PR to retrigger the CI.

acolytec3 (approved):

PR #2508, opened 2023-01-25
4844 Engine API Fixes

holgerd77:
Some fixes on the extended HF agnostic Engine API semantics.


acolytec3 (approved):
This resolves the early set of issues identified today and I've verified that we're able to sync devnet 4 up to the point where the chain stalls (at block 1109).

PR #2510, opened 2023-01-26
More Interop Log Improvements

holgerd77:
Some more log improvements related to understand/monitor post-Merge HF switches and get a better feeling for what happens during sync along the new payload calls by adding some comprehensive payload to payload aggregated stats (numbers of blocks processed/assembled, min/max block numbers involved, tx counts per tx types).

Open for review and merge I guess (might add more though, nevertheless: open for merge, then I'll submit on another PR).

### DIFF
--- a/packages/client/lib/rpc/modules/engine.ts
+++ b/packages/client/lib/rpc/modules/engine.ts
@@ -329,6 +329,9 @@ export class Engine {
   private remoteBlocks: Map<String, Block>
   private connectionManager: CLConnectionManager
 
+  private lastNewPayloadHF: string = ''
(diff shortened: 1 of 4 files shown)


jochem-brouwer (approved):
LGTM

acolytec3:
I've pushed a commit to this branch that fixes a logic bug in how we were calculating the data gas fee.  Will merge once CI passes

acolytec3 (approved):
Looks good over all though we have some `NaN` in the logging I noticed.  Will fix in a future PR since this also includes a fix for a 4844 logic bug that prevents us from syncing devnet-4.

PR #2513, opened 2023-01-27
tx: reduce number of ethers dependencies

mpetrunic:


acolytec3 (approved):
Thanks for the PR.  LGTM!

holgerd77:
I am a bit sceptical on this PR respectively I would have a tendency to roll this back, following reasoning:

We have now taken Ethers in as a "standard" dependency within the monorepo (at the moment being present in block, statemanager and tx), similarly to `ethereum-cryptography` or - our own - `RLP`. For these kind of dependencies we somewhat have the best practice to have the same dependency reference and version number in `package.json` files where the dependencies are used, e.g.:

```javascript
"ethereum-cryptography": "^1.1.2",
"ethers": "^5.7.1"
```

On updating version numbers this is (by me for sure, in general very likely) done by a monorepo wide full-text search, e.g. by searching for `"ethers": "^5.7.1"` and then replace the version number with VS Code replace or something. If we now deviate from this structure on single libraries, e.g. by referencing submodules/packages or the like, it is extremely likely that we will forget to update these single deviating entities and loose the advantages of deduplication along having the same library version and the like. In fact this has already happened with this PR.

From my weighting the argument from above outweights the slight advantage of having a somewhat leaner dependency for one of the various using libraries.

So - again - I would have a somewhat strong tendency to roll back here. Open for additional perspectives though.

mpetrunic:
> I am a bit sceptical on this PR respectively I would have a tendency to roll this back, following reasoning:
> 
> We have now taken Ethers in as a "standard" dependency within the monorepo (at the moment being present in block, statemanager and tx), similarly to `ethereum-cryptography` or - our own - `RLP`. For these kind of dependencies we somewhat have the best practice to have the same dependency reference and version number in `package.json` files where the dependencies are used, e.g.:
> 
> ```js
> "ethereum-cryptography": "^1.1.2",
> "ethers": "^5.7.1"
> ```
> 
> On updating version numbers this is (by me for sure, in general very likely) done by a monorepo wide full-text search, e.g. by searching for `"ethers": "^5.7.1"` and then replace the version number with VS Code replace or something. If we now deviate from this structure on single libraries, e.g. by referencing submodules/packages or the like, it is extremely likely that we will forget to update these single deviating entities and loose the advantages of deduplication along having the same library version and the like. In fact this has already happened with this PR.
> 
> From my weighting the argument from above outweights the slight advantage of having a somewhat leaner dependency for one of the various using libraries.
> 
> So - again - I would have a somewhat strong tendency to roll back here. Open for additional perspectives though.

I see your point. For us on web3.js side, we end up with entire ethers in bundle. Maybe it could have been reduced by just importing "ethers/providers".

Ideally, this method could be outside of TransactionFactory so it can be tree-shaked^^

holgerd77:
@mpetrunic short note: we have now merged the PR removing the Ethers dependency from Block/Tx (@acolytec3 found a good solution here) in #2633 and will do a full release round (generally planned) sometime during the next week.

holgerd77:
(again, as mentioned in the other thread: it is not as likely that we'll remove SSZ as well. Let us know how much of a problem this is for you guys)

PR #2515, opened 2023-01-30
New client ci run

acolytec3:
Replaces the old `test-client-ci` run with a single ethjs <> lodestar pair and does some real simple sanity check transactions to verify we can produce blocks and mine transactions


g11tech (approved):

holgerd77:
Nice, great improvement, thanks for working this through!! 🙏 🤩

Issue #2511, opened 2023-01-26
`getBlock` throws when block not found by hash


acolytec3:
`getBlock` is currently typed to return a [`block` or `null`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/7fa64079578b24e791c62be3fe3d6e634bb10f13/packages/blockchain/src/types.ts#L29) according to the `blockchain` interface.  In our implementation, if you look up a block by its hash and the corresponding DB key isn't in the database, the underlying `level` database throws a `NotFound` error and the dbManager doesn't handle this per the type so we get an exception instead.  

We should discuss whether to fix our implementation to match the interface or else change the interface since our general goal in recent history has to move away from returning `null`  but rather throwing an error or else returning undefined in these cases.  At a minimum, we should catch the error and then throw something more descriptive since `NotFound` doesn't tell the library consumer what is going on and is against the type suggestions provided.

jochem-brouwer:
To do so, we should review all cases where we `getBlock` and see if any behavior diverges if we return `null` instead of throw. I honestly think that indeed returning `null` makes more sense than throwing.

ScottyPoi:
Started working on this.

ScottyPoi:
> To do so, we should review all cases where we `getBlock` and see if any behavior diverges if we return `null` instead of throw. I honestly think that indeed returning `null` makes more sense than throwing.

I think I addressed all of the calls to `getBlock`, which required a bunch of `if (block === null) throw new Error("Block Not Found")`, or ` if (block === null) return undefined`

I don't know whether it feels like an improvement or not.

gabrocheleau:
Linking to to #2516

gabrocheleau:
Marking this as closed as this was reported as fixed in #2516

PR #2516, opened 2023-02-01
blockchain.getBlock - return Null instead of throwing

ScottyPoi:
Fixes bug mentioned in #2511
> `getBlock` throws when block not found by hash


acolytec3 (commented):
Looks pretty good to me though still need to fix something in the client tests.  I like that we're making the code match the interface, even if it is a tiny bit clunkier.  

Where is it breaking?  Is it the `getIteratorHead` return type change?

acolytec3:
Looks like at least some of the failing client tests are just not accounting for the new `null` return value and the test was written expecting it to throw.  So...fixing the tests will likely mean some slight modifications to how those RPC calls are handled (i.e. checking for null like you do elsewhere) and then throw the appropriate error.

ScottyPoi:
> 

I thought i covered everything, but maybe `coverage` runs some test that the other test commands don't?
Anyway, should be an easy fix like the others if I can find them.

ScottyPoi:
@acolytec3 thanks, I didn't see that you had started working on it.  

running `coverage` one more time locally, but I think i have it all

ScottyPoi:
The method `blockchain.getHead()` is marked as deprecated, replaced by `blockchain.getIteratorHead()`.
It is not called anywhere in the monorepo.

Updating `getBlock` required adding some lines of code to `getHead()` to handle the null return.  Since we already removed any tests that use `getHead()`, adding lines of code to it brought the `codecov` score further down.  

This question might be for @holgerd77 
When is it safe to delete  **deprecated** methods?

packages/blockchain/src/blockchain.ts
```diff
@@ -309,11 +309,14 @@ export class Blockchain implements BlockchainInterface {
...
    * @param name - Optional name of the iterator head (default: 'vm')
    */
-  async getIteratorHead(name = 'vm'): Promise<Block> {
-    return this.runWithLock<Block>(async () => {
+  async getIteratorHead(name = 'vm'): Promise<Block | Buffer> {
+    return this.runWithLock<Block | Buffer>(async () => {
       // if the head is not found return the genesis hash
       const hash = this._heads[name] ?? this.genesisBlock.hash()
       const block = await this.getBlock(hash)
+      if (block === null) {
```
acolytec3, line 317: Is it possible to reach this code block?  I traced down through the subsequent calls in `getBlock` and it looks like we don't have any places where the genesis block isn't instantiated as an actual `block`.  I tried contriving a test to call `getIteratorHead` without calling `blockchain._init` (where the genesis block gets created) and the `genesisBlock` is always in the db so the initial call to `getBlock` in line 316 will always return an actual `block`.
Here's my contrived test.  If I comment out the call to `_init`, `getIteratorHead` throws with an error complaining I haven't run `_init`.

```
tape('check blockchain constructor', async (t) => {
  const blockchain = new (Blockchain as any)()
  await blockchain._init()
  console.log(await blockchain.getIteratorHead())
  t.end()
})
```

All that to say, I'm wondering if we really need to have the return type of `getIteratorHead` be `<Block | Buffer>` since I don't think there's a place where this call to `getBlock` would actually return `null`.

ScottyPoi: I wonder the same thing.

Where does the "return the genesis hash if the head not found" rule come from?

acolytec3: Not sure, but I think we can safely remove that code block and revert the interface change there. Are there any other breaking changes?

acolytec3: Not sure, but I think we can safely remove that code block and revert the interface change there. Are there any other breaking - changes?

ScottyPoi: I don't think so.

I'm about to push a commit that rolls back that change and the downstream effects.
Seems like everything is still fine

holgerd77:
Rebased this and made a few clean-ups.

holgerd77 (approved):
@ScottyPoi @acolytec3 thanks a lot for working this through (definitely right decision to revert this `getIteratorHead()` type change) and generally for the clean implementation on this.

This whole change had me scratching my head a bit since I think we *will* eventually/likely get into some trouble changing/adopting the method signature typing for our `getBlock()` implementation. But I honestly also have to say that I do not see an alternative. At the end this is some kind of bug fixing, so to align this with the corresponding interface.

And the thing is (this fact I only got in a second round of reviewing/looking at this here): in the VM e.g. we do *have* implemented our stuff (and that's what others eventually have done to) our blockchain interaction behavior according to the assumed behavior of the interface, I e.g. stumbled upon this code part in `VM.runTx()`:

```typescript
const parentBlock = await this.blockchain.getBlock(opts.block?.header.parentHash)
if (parentBlock === null) {
  const msg = _errorMsg(
    `Parent block not found in blockchain so cannot compute data gas price`,
    this,
    block,
    tx
  )
  throw new Error(msg)
}
```

I was puzzled because I first thought "ah, Scotty has added this `if (parentBlock === null)` part, but then I didn't find it in the change summary here and realized that we just implemented this stuff *already*  after the interface logic and - if we would keep the old deviating behavior - the above code part would just not work since we didn't implement this null-returning logic.

So I guess this is a really important inconsistency/bug fix also for ourself, since these cases where we otherwise just "overlook" these (in doubt pretty common) block-is-not-there-for-whatever-reasons cases this can lead to serious indefined/wrong behavior.

So I will merge. Already seeing the complaints though. 🙂 

Will try to mitigate a bit by releasing at least under a minor version bump. Maybe this helps a bit to draw some attention to the fact that an update might need some attention.

Have also done some final clean-ups since some casts where unnecessary any more after going back to the old type signature for `getIteratorHead()`.

PR #2519, opened 2023-02-07
client: Fix client's sync state on startup or while mining or single node (test) runs

### LINKED ISSUE #2517: merge at genesis oddities
When we start the client with genesis parameters that have the merge at the genesis block, I have observed a couple of oddities with how the client responds to RPC calls that should be addressed.
- `eth_syncing` - When `eth_syncing` is called, if merge is at genesis and running as a solo EL/CL pair, the client throws [this error](https://github.com/ethereumjs/ethereumjs-monorepo/blob/52dbfc2026ae42ba6e607d287d616a23aa21bbe5/packages/client/lib/rpc/modules/eth.ts#L996) since there will never be another EL layer peer to sync with.  We should either add a new option to the CLI for running solo client pairs (i.e. local devnet) or else find another way to address this one (maybe add a check to see if we're using `BeaconSync` vs one of the other synchronizers?
- `eth_sendRawTransaction` - [Here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/52dbfc2026ae42ba6e607d287d616a23aa21bbe5/packages/client/lib/rpc/modules/eth.ts#L874) - when a transaction is submitted, the client throws an error about not knowing the current chain height.  If merge is at genesis, it seems like we should just assume height = 0 until a paired CL client provides a FcU and accept the transaction.

g11tech:
Fix client's sync state on startup or while mining or single node (test) runs

Closes #2517

### DIFF
--- a/packages/client/lib/config.ts
+++ b/packages/client/lib/config.ts
@@ -463,6 +463,14 @@ export class Config {
         }
       }
     }
+
+    this.logger.debug(
+      `Client synchronized=${this.synchronized}${
+        latest ? ' height= ' + latest.number : ''
+      } syncTargetHeight=${this.syncTargetHeight} lastSyncDate=${
+        (Date.now() - this.lastSyncDate) / 1000
+      } secs ago`
+    )
   }
 
   /**
--- a/packages/client/lib/config.ts
+++ b/packages/client/lib/config.ts
@@ -4,11 +4,12 @@ import { Level } from 'level'
 
 import { getLogger } from './logging'
 import { Libp2pServer, RlpxServer } from './net/server'
-import { EventBus } from './types'
-import { parseTransports } from './util'
+import { Event, EventBus } from './types'
+import { parseTransports, short } from './util'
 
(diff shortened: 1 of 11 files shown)


packages/client/lib/config.ts
```diff
@@ -275,6 +288,7 @@ export class Config {
   public static readonly SAFE_REORG_DISTANCE = 100
   public static readonly SKELETON_FILL_CANONICAL_BACKSTEP = 100
   public static readonly SKELETON_SUBCHAIN_MERGE_MINIMUM = 1000
+  public static readonly SYNCED_STATE_REMOVAL_PERIOD = 60000
```
holgerd77, line 291: That's really a nice and more explicit configuration parameter on this! 👍

holgerd77 (approved):
This is a really welcomed refactor of the synced state detection, really somewhat central to have this clean under a single source of truth, already realized this when I muddled a bit along these code parts at interop, great! 🤩

Will merge.

PR #2523, opened 2023-02-09
Revert PR #2516 Blockchain.getBlock() return Null instead of throwing

holgerd77:
This reverts commit c5bb62a5ac38ee052a0c4be54e77f2b37e3a6619.

After an additional analysis with @g11tech on the way we have implemented we came to the conclusion that this is too breaking in its current form, since we would have had already various necessary code changes following this in our internal code base.

We will then instead go the path and change the interface instead (removing the `null` return option), which should have substantially less side effects.

@ScottyPoi can you please go your PR #2516 and extract (and adopt a bit) valuable unrelated updates like the additional tests you wrote? Thanks! 🙏

So, some context for review: this is undoing the changes from the PR with the command:

```shell
git revert c5bb62a5ac38ee052a0c4be54e77f2b37e3a6619
```


g11tech (approved):
lgtm

PR #2524, opened 2023-02-09
Fix blockchain interface getBlock() signature (remove null return option)

holgerd77:
Replacement candidate for reverted #2516 after some additional discussion on the breaking implications of this.

This PR now instead adjusts the Blockchain getBlock() interface signature, adds some tests and removes some non-reachable null asserting conditional clauses.

PR also improves on the error message of `getBlock()` as suggested in #2511.

Open for review.

### DIFF
--- a/packages/blockchain/src/blockchain.ts
+++ b/packages/blockchain/src/blockchain.ts
@@ -708,7 +708,18 @@ export class Blockchain implements BlockchainInterface {
     // in the `VM` if we encounter a `BLOCKHASH` opcode: then a bigint is used we
     // need to then read the block from the canonical chain Q: is this safe? We
     // know it is OK if we call it from the iterator... (runBlock)
-    return this.dbManager.getBlock(blockId)
+    try {
+      return await this.dbManager.getBlock(blockId)
+    } catch (error: any) {
+      if (error.code === 'LEVEL_NOT_FOUND') {
+        if (typeof blockId === 'object') {
+          error.message = `Block with hash ${blockId.toString('hex')} not found in DB (NotFound)`
+        } else {
+          error.message = `Block number ${blockId} not found in DB (NotFound)`
+        }
+      }
+      throw error
+    }
   }
 
   /**
--- a/packages/blockchain/src/blockchain.ts
+++ b/packages/blockchain/src/blockchain.ts
@@ -653,9 +653,6 @@ export class Blockchain implements BlockchainInterface {
     let parentHash = block.header.parentHash
     for (let i = 0; i < getBlocks; i++) {
       const parentBlock = await this.getBlock(parentHash)
-      if (parentBlock === undefined) {
-        throw new Error(`could not find parent block ${block.errorStr()}`)
-      }
       canonicalBlockMap.push(parentBlock)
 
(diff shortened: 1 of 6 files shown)


packages/blockchain/src/blockchain.ts
```diff
@@ -708,7 +705,18 @@ export class Blockchain implements BlockchainInterface {
     // in the `VM` if we encounter a `BLOCKHASH` opcode: then a bigint is used we
     // need to then read the block from the canonical chain Q: is this safe? We
     // know it is OK if we call it from the iterator... (runBlock)
-    return this.dbManager.getBlock(blockId)
+    try {
+      return await this.dbManager.getBlock(blockId)
+    } catch (error: any) {
+      if (error.code === 'LEVEL_NOT_FOUND') {
+        if (typeof blockId === 'object') {
+          error.message = `Block with hash ${blockId.toString('hex')} not found in DB (NotFound)`
```
holgerd77, line 713: I've kept the `NotFound` here to increase backwards compatibility of the change, in case someone compares the string of this error message (respectively: checks for inclusion).

packages/vm/src/eei/eei.ts
```diff
@@ -12,7 +12,7 @@ type Block = {
 }
 
 type Blockchain = {
-  getBlock(blockId: number): Promise<Block | null>
+  getBlock(blockId: number): Promise<Block>
```
holgerd77, line 15: I just changed this to align, not 100% sure if this might have some effects somewhere.

Generally seems to me the right thing to do.

packages/vm/src/runTx.ts
```diff
@@ -326,15 +326,6 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
...
     const parentBlock = await this.blockchain.getBlock(opts.block?.header.parentHash)
-    if (parentBlock === null) {
-      const msg = _errorMsg(
-        `Parent block not found in blockchain so cannot compute data gas price`,
-        this,
-        block,
-        tx
-      )
-      throw new Error(msg)
-    }
```
holgerd77, line 337: I was a bit lazy here (since code could not be reached before anyhow), we could also expand here though and wrap in a try/catch and recreate the error message if wished.

g11tech (approved):
looks good :+1:

packages/vm/src/eei/eei.ts
acolytec3: This is what I originally wanted to do so glad we're able to incorporate this.  I'm assuming this does indeed make this breaking change?

holgerd77: I actually don't think that this is so breaking, having dealt with similar code parts. If someone has prepared for dealing with Block or `null` in his/her code base, all well, now it's just limited to only Blocks. Especially regarding the fact that `null` de facto never was a valid option respectively never occurred.

So I think we can label this as a bugfix and pack this in at most a minor release version.

ScottyPoi (commented):
This is already merged, but still wanted to give my :+1: 

Feels like there are some minor optimizations that could be done here, but nothing urgent

PR #2522, opened 2023-02-08
Remove trusted setups from build

acolytec3:
Moves the `trusted_setup` out of `src` so it's no longer included in the `tx` bundle by default.  Also updates the readme to provide guidance on how to set up the dev environment to work with blob transactions.

PR #2526, opened 2023-02-10
Client trusted setup path fixes

holgerd77:
This fixes two bugs related to reading the trusted setup file updated in #2522 and breaking the client build, discovered this while trying to locally build the docker image for playing around with Hive.

1. The trusted setup file had not been included in the build, have added a script addition to the Client `npm run build`  command
2. the relative path references were not working properly especially for the trusted setup file reference in `bin/cli.ts` where path hierarchy would change when copied over to `dist/bin/cli.ts` (one level deeper). This is now mitigated by using client package internal references which - in the dist folder case - leads to taking the - now copied over, see 1. - devnet4.txt file - from `dist/trustedSetups/devnet4.txt`

Open for review. 🙂

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -579,7 +579,7 @@ async function run() {
 
   // TODO sharding: Just initialize kzg library now, in future it can be optimized to be
   // loaded and initialized on the sharding hardfork activation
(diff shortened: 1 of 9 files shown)


holgerd77:
This is open for review, have updated the initial description with context on what is fixed here.

g11tech (approved):
lgtm

PR #2507, opened 2023-01-25
Fix engine-withdrawals hive tests 

jochem-brouwer:

PR #2529, opened 2023-02-10
allforks: Hive withdrawal fixes

### LINKED ISSUE #2498: Hive tests failing
Currently, we have failing hive tests (on hive `master` branch). 

In the [hive](https://github.com/ethereum/hive) repository, in order to run local code, change `clients/ethereumjs/Dockerfile` to:

```
FROM node:16-alpine as build
RUN apk update && apk add --no-cache bash git jq curl python3 gcc make g++ && rm -rf /var/cache/apk/*
ADD ethereumjs-monorepo /ethereumjs-monorepo
RUN cd ethereumjs-monorepo && npm i

RUN cd ethereumjs-monorepo/packages/client

ADD ethereumjs.sh /ethereumjs.sh
ADD mapper.jq /mapper.jq
ADD jwtsecret /jwtsecret
RUN chmod +x /ethereumjs.sh

# Inject the enode id retriever script
RUN mkdir /hive-bin
ADD enode.sh /hive-bin/enode.sh
RUN chmod +x /hive-bin/enode.sh

ADD genesis.json /genesis.json

# Export the usual networking ports to allow outside access to the node
EXPOSE 8545 8546 8551 8547 30303 30303/udp

# NodeJS applications have a default memory limit of 2.5GB.
# This limit is bit tight, it is recommended to raise the limit
# since memory may spike during certain network conditions.
ENV NODE_OPTIONS=--max_old_space_size=6144

ENTRYPOINT ["/ethereumjs.sh"]
```

You will need to copy/clone `ethereumjs-monorepo` into the `clients/ethereumjs` directory of hive. Note: tests are still somewhat slow since each time hive is started it will copy the entire directory into the docker container.

g11tech:
Fixes regarding failing hive testvectors

Closes #2498

 - [x] tx hardfork mismatch fix where the starting client hardfork was not being properly set as the timestamp was not being used, fixed
 - [x] check for withdrawals/excessdatagas missing on newPayload
 - [x] Skip altering the canonical chain number to hash mapping on runWithoutSetHead / newPayload
 - [x] Fix setting the hardfork on vm in getTransactionReceipt
 - [x] Fix setting evm hardfork on copy
 - [x] fix loading block/header which might not be in canonical chain

### DIFF
--- a/packages/vm/src/eei/eei.ts
+++ b/packages/vm/src/eei/eei.ts
@@ -94,6 +94,8 @@ export class EEI extends VmState implements EEIInterface {
   }
 
   public copy() {
-    return new EEI(this._stateManager.copy(), this._common.copy(), this._blockchain.copy())
+    const common = this._common.copy()
+    common.setHardfork(this._common.hardfork())
+    return new EEI(this._stateManager.copy(), common, this._blockchain.copy())
   }
 }
--- a/packages/blockchain/src/blockchain.ts
+++ b/packages/blockchain/src/blockchain.ts
@@ -1256,7 +1256,7 @@ export class Blockchain implements BlockchainInterface {
    */
   createGenesisBlock(stateRoot: Buffer): Block {
     const common = this._common.copy()
-    common.setHardforkByBlockNumber(0)
+    common.setHardforkByBlockNumber(0, undefined, common.genesis().timestamp)
 
     const header: BlockData['header'] = {
       ...common.genesis(),
--- a/packages/blockchain/src/db/helpers.ts
+++ b/packages/blockchain/src/db/helpers.ts
@@ -69,9 +69,11 @@ function DBSetHashToNumber(blockHash: Buffer, blockNumber: bigint): DBOp {
   })
 }
 
-function DBSaveLookups(blockHash: Buffer, blockNumber: bigint): DBOp[] {
+function DBSaveLookups(blockHash: Buffer, blockNumber: bigint, skipNumIndex?: boolean): DBOp[] {
   const ops = []
-  ops.push(DBOp.set(DBTarget.NumberToHash, blockHash, { blockNumber }))
+  if (skipNumIndex !== true) {
+    ops.push(DBOp.set(DBTarget.NumberToHash, blockHash, { blockNumber }))
+  }
 
   const blockNumber8Bytes = bufBE8(blockNumber)
   ops.push(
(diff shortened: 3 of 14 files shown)


jochem-brouwer:
Some (more) fixes might be in this PR: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2507

(I honestly thought this was merged already)

acolytec3 (approved):
LGTM

g11tech:
> Some (more) fixes might be in this PR: #2507
> 
> (I honestly thought this was merged already)

thanks @jochem-brouwer , yea we had incorporated fixes from there into this :+1:

packages/vm/src/eei/eei.ts
```diff
@@ -94,6 +94,8 @@ export class EEI extends VmState implements EEIInterface {
   }
 
   public copy() {
-    return new EEI(this._stateManager.copy(), this._common.copy(), this._blockchain.copy())
+    const common = this._common.copy()
+    common.setHardfork(this._common.hardfork())
```
holgerd77, line 98: Valid solution here, but I guess this hardfork setting should be functionality directly integrated into the `Common` `copy()` function? 🤔

holgerd77:
Had another look here, thanks🙏, this all looks great!

packages/vm/src/eei/eei.ts
acolytec3: I wasn't sure on this one.  It might warrant further research on whether or not there is other "state" inside of `common` that needs to be copied along with the current `hardfork`.  That is the most impactful one (as identified by our failing tests) but I'm not sure if there's more to be thought through here.

holgerd77: This got me a bit curious and a had another look, since on second thought it seemed extremely unlikely to me that our libraries would work at all if in the former setup `Common.copy()` wouldn't preserve the hardfork state. Tested this out and it turns out that it *does*, since it does copy property values as well in `Common.copy()`.

You can test out the following code:

```typescript
import { Common, Chain, Hardfork } from '@ethereumjs/common'
const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Shanghai, eips: [4844] })
const common2 = common.copy()
common2.hardfork() // 'shanghai'
```

So seems we can delete these additional lines here again, the fixes seem to come from some other code changes in this PR. 🙂

EIPs are actually preserved as well, `common2.eips() // [ 4844 ]`.

acolytec3: Ah, okay. That makes sense. I wasn't sure if the EEI was impacted or not. We certainly had a bug where copying VM resulted in separate `common` instances being created for the internal evm and EEI but maybe that was the limit of the issue.

PR #2530, opened 2023-02-14
allforks: Allow genesis to be post merge

g11tech:
This PR builds on the top of other withdrawal fixes  for #2498
- https://github.com/ethereumjs/ethereumjs-monorepo/pull/2529

since this needs its own separate consideration, however this branch can be used to continue fixing further withdrawal hive fixes

We have been having various oddies regarding merge at genesis, where we didn't allow genesi block to attain merge, and forcefully putting merge hf just after genesis hardforks. So even with shanghai genesis we would get merge hf at block 1,  and other issues which would not be as clear to the developer.

Post discussions with @acolytec3 while hive debugging, we concluded that its ok or rather desirable to allow genesis to attain merge.

This PR allows merge at genesis as well as skips some validations for PoS genesis which hive post merge shanghai genesis were failing.


acolytec3 (approved):
LGTM!

PR #2532, opened 2023-02-15
Various Hive related fixes

acolytec3:
New fixes related to Hive tests

- Adds a new `--loadBlocksFromRlp` CLI parameter that allows for loading RLP encoded blocks on chain start that are fed via the `./chain.rlp` path in various Hive tests.  [Needed by the Hive tests](https://github.com/ethereum/hive/blob/master/docs/clients.md#files)
- Add better handling of reorgs that go pre-merge -- fixes hive tests run by this command - `go run ./hive.go --client ethereumjs --sim ethereum/engine --sim.limit "engine-transition/Single Block PoW Re-org to Higher-Total-Difficulty Chain, Equal Height"`
- Skip extradata check on PoW genesis blocks (since some Hive tests with PoW genesis parameters have `extradata` that does not conform to PoW extradata requirements

### DIFF
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -354,8 +354,10 @@ export class BlockHeader {
     if (this._common.consensusAlgorithm() === ConsensusAlgorithm.Ethash) {
       // PoW/Ethash
       if (
+        number > BigInt(0) &&
         this.extraData.length > this._common.paramByHardfork('vm', 'maxExtraDataSize', hardfork)
       ) {
+        // Check length of data on all post-genesis blocks
         const msg = this._errorMsg('invalid amount of extra data')
         throw new Error(msg)
       }
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -421,7 +421,7 @@ async function startClient(config: Config, customGenesisState?: GenesisState) {
     await startBlock(client)
   }
 
-  if (typeof args.loadBlocksFromRlp === 'string') await client.open()
+  await client.open()
   // update client's sync status and start txpool if synchronized
   client.config.updateSynchronizedState(client.chain.headers.latest)
   if (client.config.synchronized) {
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -1,9 +1,11 @@
 #!/usr/bin/env node
 
+import { Block } from '@ethereumjs/block'
 import { Blockchain, parseGethGenesisState } from '@ethereumjs/blockchain'
 import { Chain, Common, ConsensusAlgorithm, Hardfork } from '@ethereumjs/common'
+import { RLP } from '@ethereumjs/rlp'
 import { initKZG } from '@ethereumjs/tx'
-import { Address, toBuffer } from '@ethereumjs/util'
+import { Address, arrToBufArr, toBuffer } from '@ethereumjs/util'
 import * as kzg from 'c-kzg'
 import { randomBytes } from 'crypto'
 import { existsSync, writeFileSync } from 'fs'
@@ -25,6 +27,7 @@ import type { Logger } from '../lib/logging'
 import type { FullEthereumService } from '../lib/service'
 import type { ClientOpts } from '../lib/types'
 import type { RPCArgs } from './startRpc'
+import type { BlockBuffer } from '@ethereumjs/block'
 import type { GenesisState } from '@ethereumjs/blockchain/dist/genesisStates'
 import type { AbstractLevel } from 'abstract-level'
 
@@ -292,6 +295,10 @@ const args: ClientOpts = yargs(hideBin(process.argv))
       'To run client in single node configuration without need to discover the sync height from peer. Particularly useful in test configurations. This flag is automically activated in the "dev" mode',
     boolean: true,
   })
+  .option('loadBlocksFromRlp', {
+    describe: 'path to a file of RLP encoded blocks',
+    string: true,
+  })
   // strict() ensures that yargs throws when an invalid arg is provided
   .strict().argv
 
@@ -414,7 +421,7 @@ async function startClient(config: Config, customGenesisState?: GenesisState) {
     await startBlock(client)
   }
 
-  await client.open()
+  if (typeof args.loadBlocksFromRlp === 'string') await client.open()
   // update client's sync status and start txpool if synchronized
   client.config.updateSynchronizedState(client.chain.headers.latest)
   if (client.config.synchronized) {
@@ -422,6 +429,17 @@ async function startClient(config: Config, customGenesisState?: GenesisState) {
     // The service might not be FullEthereumService even if we cast it as one,
     // so txPool might not exist on it
     ;(fullService as FullEthereumService).txPool?.checkRunState()
+  }
+
+  if (args.loadBlocksFromRlp !== undefined) {
+    // Specifically for Hive simulator, preload a first block provided in RLP format
+    const blockRlp = readFileSync(args.loadBlocksFromRlp)
+    const decodedBlock = RLP.decode(blockRlp)
+    const block = Block.fromValuesArray(arrToBufArr(decodedBlock) as BlockBuffer, {
+      common: config.chainCommon,
+      skipConsensusFormatValidation: true,
+    })
+    await client.chain.putBlocks([block])
   }
 
   if (args.executeBlocks !== undefined) {
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -431,6 +431,14 @@ async function startClient(config: Config, customGenesisState?: GenesisState) {
     ;(fullService as FullEthereumService).txPool?.checkRunState()
   }
 
+  if (args.executeBlocks !== undefined) {
+    // Special block execution debug mode (does not change any state)
+    await executeBlocks(client)
+  } else {
+    // Regular client start
+    await client.start()
+  }
+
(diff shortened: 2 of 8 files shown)


acolytec3:
Additional things needed to properly run this code with hive:
1) Replace the dockerfile in `hive/clients/ethereumjs/Dockerfile` with below which adds the new reading of `./chain.rlp`
```
FROM node:18-alpine as build
RUN apk update && apk add --no-cache bash git jq curl python3 gcc make g++ rsync && rm -rf /var/cache/apk/*

RUN mkdir /ethereumjs-monorepo
WORKDIR /ethereumjs-monorepo

COPY ethereumjs-monorepo/package.json .
COPY ethereumjs-monorepo/package-lock.json .
RUN npm i --ignore-scripts

COPY ethereumjs-monorepo-master/ ./
RUN npm i

COPY ethereumjs-monorepo /ejsrepo
RUN rsync -a /ejsrepo/* .
RUN git status
RUN npm i

WORKDIR /

ADD ethereumjs.sh /ethereumjs.sh
ADD mapper.jq /mapper.jq
ADD jwtsecret /jwtsecret
RUN chmod +x /ethereumjs.sh


# Inject the enode id retriever script
RUN mkdir /hive-bin
ADD enode.sh /hive-bin/enode.sh
RUN chmod +x /hive-bin/enode.sh

ADD genesis.json /genesis.json

# Export the usual networking ports to allow outside access to the node
EXPOSE 8545 8546 8551 8547 30303 30303/udp

# NodeJS applications have a default memory limit of 2.5GB.
# This limit is bit tight, it is recommended to raise the limit
# since memory may spike during certain network conditions.
ENV NODE_OPTIONS=--max_old_space_size=6144

ENTRYPOINT ["/ethereumjs.sh"]
```

2.  Replace `hive/clients/ethereumjs/ethereumjs.sh` with this:
```
#!/bin/bash

# Startup script to initialize and boot a ethereum-js instance.
#
# This script assumes the following files:
#  - `genesis.json` file is located in the filesystem root (mandatory)
#  - `chain.rlp` file is located in the filesystem root (optional)
#  - `blocks` folder is located in the filesystem root (optional)
#  - `keys` folder is located in the filesystem root (optional)
#
# This script assumes the following environment variables:
#
#  - HIVE_BOOTNODE                enode URL of the remote bootstrap node
#  - HIVE_NETWORK_ID              network ID number to use for the eth protocol
#  - HIVE_TESTNET                 whether testnet nonces (2^20) are needed
#  - HIVE_NODETYPE                sync and pruning selector (archive, full, light)
#
# Forks:
#
#  - HIVE_FORK_HOMESTEAD          block number of the homestead hard-fork transition
#  - HIVE_FORK_DAO_BLOCK          block number of the DAO hard-fork transition
#  - HIVE_FORK_DAO_VOTE           whether the node support (or opposes) the DAO fork
#  - HIVE_FORK_TANGERINE          block number of Tangerine Whistle transition
#  - HIVE_FORK_SPURIOUS           block number of Spurious Dragon transition
#  - HIVE_FORK_BYZANTIUM          block number for Byzantium transition
#  - HIVE_FORK_CONSTANTINOPLE     block number for Constantinople transition
#  - HIVE_FORK_PETERSBURG         block number for ConstantinopleFix/PetersBurg transition
#  - HIVE_FORK_ISTANBUL           block number for Istanbul transition
#  - HIVE_FORK_MUIRGLACIER        block number for Muir Glacier transition
#  - HIVE_FORK_BERLIN             block number for Berlin transition
#  - HIVE_FORK_LONDON             block number for London
#
# Clique PoA:
#
#  - HIVE_CLIQUE_PERIOD           enables clique support. value is block time in seconds.
#  - HIVE_CLIQUE_PRIVATEKEY       private key for clique mining
#
# Other:
#
#  - HIVE_MINER                   enable mining. value is coinbase address.
#  - HIVE_MINER_EXTRA             extra-data field to set for newly minted blocks
#  - HIVE_SKIP_POW                if set, skip PoW verification during block import
#  - HIVE_LOGLEVEL                client loglevel (0-5)
#  - HIVE_GRAPHQL_ENABLED         enables graphql on port 8545
#  - HIVE_LES_SERVER              set to '1' to enable LES server


# Immediately abort the script on any error encountered
set -e

ethereumjs="node /ethereumjs-monorepo/packages/client/dist/bin/cli.js"
FLAGS="--gethGenesis ./genesis.json --rpc --rpcEngine --saveReceipts --rpcEnginePort 8551 --ws --logLevel debug --rpcDebug --transports rlpx --isSingleNode"


# Configure the chain.
mv /genesis.json /genesis-input.json
jq -f /mapper.jq /genesis-input.json > /genesis.json

# Dump genesis
echo "Supplied genesis state:"
cat /genesis.json

# Import clique signing key.
if [ "$HIVE_CLIQUE_PRIVATEKEY" != "" ]; then
    # Create password file.
    echo "Importing clique key..."
    echo "$HIVE_CLIQUE_PRIVATEKEY" > ./private_key.txt
    # Ensure password file is used when running ethereumjs in mining mode.
    if [ "$HIVE_MINER" != "" ]; then
        FLAGS="$FLAGS --mine --unlock ./private_key.txt --minerCoinbase $HIVE_MINER"
    fi
fi

if [ "$HIVE_TERMINAL_TOTAL_DIFFICULTY" != "" ]; then
    FLAGS="$FLAGS --jwt-secret ./jwtsecret"
fi


# Load the test chain if present
echo "Loading initial blockchain..."
if [ -f /chain.rlp ]; then
  FLAGS="$FLAGS --loadBlocksFromRlp=/chain.rlp"
  else
  echo "Warning: chain.rlp not found."
fi

FLAGS="$FLAGS --bootnodes=$HIVE_BOOTNODE"
echo "Running ethereumjs with flags $FLAGS"
$ethereumjs $FLAGS
```

acolytec3:
Use this command -- `go run ./hive.go --client ethereumjs --sim ethereum/engine --sim.limit "engine-transition/Single Block PoW Re-org to Higher-Total-Difficulty Chain, Equal Height" --docker.output` to run one of the `engine-transition` tests that loads block data on client start.  It's currently failing because we don't handle the reorg case correctly when going back from PoS to a different PoW terminal block.

packages/block/src/header.ts
```diff
@@ -354,8 +354,10 @@ export class BlockHeader {
     if (this._common.consensusAlgorithm() === ConsensusAlgorithm.Ethash) {
       // PoW/Ethash
       if (
+        number > BigInt(0) &&
```
jochem-brouwer, line 357: Ok, this disables genesis extradata check by default, I am not sure if we want this?

packages/client/bin/cli.ts
```diff
@@ -292,6 +295,10 @@ const args: ClientOpts = yargs(hideBin(process.argv))
       'To run client in single node configuration without need to discover the sync height from peer. Particularly useful in test configurations. This flag is automically activated in the "dev" mode',
     boolean: true,
   })
+  .option('loadBlocksFromRlp', {
+    describe: 'path to a file of RLP encoded blocks',
```
jochem-brouwer, line 299: We should take a look how other clients import these blocks. I would be more in favor to only allow this if some kind of "test network" flag is provided.

packages/client/bin/cli.ts
```diff
@@ -431,6 +438,46 @@ async function startClient(config: Config, customGenesisState?: GenesisState) {
...
+        })
+        blocks.push(block)
+        buf = RLP.decode(buf.remainder, true)
+        config.logger.info(
+          `Preloading block 0x${short(block.header.hash().toString('hex'))} - ${
+            block.header.number
+          }`
+        )
+      } catch {
+        config.logger.info('Encountered invalid block while preloading chain')
```
jochem-brouwer, line 462: If this is the case, should we not immediately quit the client, instead of proceeding?

packages/client/bin/cli.ts
```diff
@@ -431,6 +438,46 @@ async function startClient(config: Config, customGenesisState?: GenesisState) {
...
+        done = true
```
jochem-brouwer, line 463: `break`?

jochem-brouwer (commented):
Some remarks :)

packages/client/bin/cli.ts
acolytec3: https://github.com/ethereum/hive/blob/master/docs/clients.md No, we're supposed to import to to the last valid block, per the [hive docs](https://github.com/ethereum/hive/blob/master/docs/clients.md)

packages/block/src/header.ts
acolytec3: This could be tricky since this check is in common and doesn't have access to the clients config flags that would indicate a test environment. We could have the hive simulator remove the extradata field if we're on PoW and that should circumvent the issue

packages/client/bin/cli.ts
acolytec3: It's pretty similar. They look for a specific flag and load them if so. Why would this be an issue for us as currently designed? This flag never gets set for a production run.

packages/client/bin/cli.ts
acolytec3: Ahh, yes, that should have been obvious to me. Will update.

holgerd77:
Ok, I would regard this as too much work-in-progress as to consider for tomorrows releases.

Please do not merge before releases are published.

g11tech (approved):
lgtm!

holgerd77 (approved):
Looks good

holgerd77 (approved):
LGTM

PR #2533, opened 2023-02-16
withdrawals and transaction inclusion logic updates

g11tech:
- Changes `engine_forkchoiceUpdatedV2` withdrawals parameter to `optional` to ensure we return the correct error message if a preShanghai payload is sent
- Adds some caching logic for pending payloads so as to prune pending block payloads once one is built and provided to the CL
- Modifies the logic for including transactions in a pendingBlock by using the pending block's hardfork (instead of the transaction hardfork since transactions are generally hardfork agnostic) when running the transaction in the VM to ensure that it can be executed and included in the block.

Targets Hive tests that can be run using this command:
`go run ./hive.go --client ethereumjs --sim ethereum/engine --sim.limit "engine-withdrawals/Withdrawals Fork on Block 2"`

### DIFF
--- a/packages/client/lib/miner/pendingBlock.ts
+++ b/packages/client/lib/miner/pendingBlock.ts
@@ -1,6 +1,13 @@
 import { BlockHeader } from '@ethereumjs/block'
 import { BlobEIP4844Transaction } from '@ethereumjs/tx'
-import { TypeOutput, bigIntToUnpaddedBuffer, bufferToHex, toBuffer, toType } from '@ethereumjs/util'
+import {
+  TypeOutput,
+  bigIntToUnpaddedBuffer,
+  bufferToHex,
+  toBuffer,
+  toType,
+  zeros,
+} from '@ethereumjs/util'
 import { BuildStatus } from '@ethereumjs/vm/dist/buildBlock'
 import { keccak256 } from 'ethereum-cryptography/keccak'
 
@@ -86,9 +93,9 @@ export class PendingBlock {
     const { gasLimit } = parentBlock.header
 
     // payload is uniquely defined by timestamp, gasLimit and the header
-    const timestampBuf = bigIntToUnpaddedBuffer(toType(timestamp, TypeOutput.BigInt))
+    const timestampBuf = bigIntToUnpaddedBuffer(toType(timestamp ?? 0, TypeOutput.BigInt))
     const gasLimitBuf = bigIntToUnpaddedBuffer(gasLimit)
-    const mixHashBuf = toType(mixHash!, TypeOutput.Buffer)
+    const mixHashBuf = toType(mixHash!, TypeOutput.Buffer) ?? zeros(32)
     const payloadIdBuffer = toBuffer(
       keccak256(Buffer.concat([parentBlock.hash(), mixHashBuf, timestampBuf, gasLimitBuf])).slice(
         0,
--- a/packages/client/lib/miner/pendingBlock.ts
+++ b/packages/client/lib/miner/pendingBlock.ts
@@ -172,18 +172,6 @@ export class PendingBlock {
               `Pending: Assembled block full (gasLeft: ${gasLimit - builder.gasUsed})`
             )
           }
-        } else if ((error as Error).message.includes('tx has a different hardfork than the vm')) {
-          // We can here decide to keep a tx in pool if it belongs to future hf
-          // but for simplicity just remove the tx as the sender can always retransmit
-          // the tx
-          this.txPool.removeByHash(txs[index].hash().toString('hex'))
-          this.config.logger.error(
-            `Pending: Removed from txPool tx 0x${txs[index]
-              .hash()
-              .toString('hex')} having different hf=${txs[
-              index
-            ].common.hardfork()} than block vm hf=${vm._common.hardfork()}`
-          )
         } else {
           // If there is an error adding a tx, it will be skipped
           this.config.logger.debug(
--- a/packages/client/lib/miner/pendingBlock.ts
+++ b/packages/client/lib/miner/pendingBlock.ts
@@ -304,8 +304,6 @@ export class PendingBlock {
       this.constructBlobsBundle(payloadId, blobTxs, block.header.hash())
     }
 
-    // Remove from pendingPayloads
-    // this.pendingPayloads = this.pendingPayloads.filter((p) => !p[0].equals(payloadId))
 
     return [block, builder.transactionReceipts, builder.minerValue]
   }
--- a/packages/client/lib/miner/pendingBlock.ts
+++ b/packages/client/lib/miner/pendingBlock.ts
@@ -1,6 +1,7 @@
 import { BlockHeader } from '@ethereumjs/block'
 import { BlobEIP4844Transaction } from '@ethereumjs/tx'
-import { randomBytes } from 'crypto'
+import { TypeOutput, bigIntToUnpaddedBuffer, bufferToHex, toBuffer, toType } from '@ethereumjs/util'
+import { keccak256 } from 'ethereum-cryptography/keccak'
 
 import type { Config } from '../config'
 import type { TxPool } from '../service/txpool'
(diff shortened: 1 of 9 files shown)


holgerd77:
Could you guys mention the names of the Hive test cases being fixed in the PR and optimally also link to where these tests are run online (if a server test run exists for the specific tests)?

That would make things more transparent and also builds up on excitement when looking for the server test runs the next time! 🙂 🙏 😋

packages/vm/src/buildBlock.ts
```diff
@@ -216,10 +229,9 @@ export class BlockBuilder {
    * Reverts the checkpoint on the StateManager to reset the state from any transactions that have been run.
    */
   async revert() {
-    this.checkStatus()
     if (this.checkpointed) {
       await this.vm.stateManager.revert()
-      this.reverted = true
+      this.blockStatus = { status: BuildStatus.Reverted }
```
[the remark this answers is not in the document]
g11tech: well we can always `revert` even if its build (my thinking behind not throwing)

packages/client/lib/miner/pendingBlock.ts
```diff
@@ -33,20 +42,42 @@ interface BlobBundle {
...
-    this.blobBundles = new Map()
     this.skipHardForkValidation = opts.skipHardForkValidation
   }
 
+  pruneSetToMax(maxItems: number): number {
+    let itemsToDelete = this.pendingPayloads.size - maxItems
+    const deletedItems = Math.max(0, itemsToDelete)
+
+    if (itemsToDelete > 0) {
+      // keys are in fifo order
```
jochem-brouwer, line 69: OK, had to verify this, but indeed map keys are fifo order

packages/client/lib/miner/pendingBlock.ts
```diff
@@ -58,9 +89,29 @@ export class PendingBlock {
...
   ) {
     const number = parentBlock.header.number + BigInt(1)
-    const { timestamp } = headerData
+    const { timestamp, mixHash } = headerData
     const { gasLimit } = parentBlock.header
 
+    // payload is uniquely defined by timestamp, gasLimit and the header
+    const timestampBuf = bigIntToUnpaddedBuffer(toType(timestamp ?? 0, TypeOutput.BigInt))
+    const gasLimitBuf = bigIntToUnpaddedBuffer(gasLimit)
+    const mixHashBuf = toType(mixHash!, TypeOutput.Buffer) ?? zeros(32)
```
jochem-brouwer, line 98: Doesn't this crash if `mixHash` is `undefined`?

packages/client/lib/miner/pendingBlock.ts
```diff
@@ -260,7 +305,7 @@ export class PendingBlock {
     }
 
     // Remove from pendingPayloads
-    this.pendingPayloads = this.pendingPayloads.filter((p) => !p[0].equals(payloadId))
+    // this.pendingPayloads = this.pendingPayloads.filter((p) => !p[0].equals(payloadId))
```
jochem-brouwer, line 308: This is commented out?

packages/client/lib/miner/pendingBlock.ts
```diff
@@ -272,13 +317,13 @@ export class PendingBlock {
    * @param blockHash the blockhash of the pending block (computed from the header data provided)
    */
   private constructBlobsBundle = (
-    payloadId: Buffer,
+    payloadId: string,
```
jochem-brouwer, line 320: These payloadIds are 8-byte strings, right? Should we change the string type to a `buffer8string` or something similar for clarity?

packages/client/lib/miner/pendingBlock.ts
```diff
@@ -163,33 +201,40 @@ export class PendingBlock {
...
    * Stops a pending payload
    */
-  stop(payloadId: Buffer) {
-    const payload = this.pendingPayloads.find((p) => p[0].equals(payloadId))
-    if (!payload) return
+  stop(payloadIdBuffer: Buffer | string) {
+    const payloadId =
+      typeof payloadIdBuffer !== 'string' ? bufferToHex(payloadIdBuffer) : payloadIdBuffer
+    const builder = this.pendingPayloads.get(payloadId)
+    if (!builder) return
```
jochem-brouwer, line 214: In other code (which is not edited here but is in the same files) we explicitly check `builder !== undefined`, should we change? (This is also part of an eslint discussion we had a while back to make these checks more strict)

packages/vm/src/buildBlock.ts
```diff
@@ -216,11 +229,7 @@ export class BlockBuilder {
    * Reverts the checkpoint on the StateManager to reset the state from any transactions that have been run.
    */
   async revert() {
-    this.checkStatus()
-    if (this.checkpointed) {
-      await this.vm.stateManager.revert()
-      this.reverted = true
-    }
+    this.blockStatus = { status: BuildStatus.Reverted }
```
jochem-brouwer, line 232: This removes a checkpoint mechanism check, what are the implications of this?

jochem-brouwer (commented):
Some remarks and questions :)

packages/vm/src/buildBlock.ts
acolytec3: This particular one shouldn't matter because once reverted, the builder can never be used again and will get deleted on the call to ForkChoiceUpdated. Since the underlying checkpoints are in a copy of state manager attached to the block builder, it gets thrown away so extra work to revert the checkpoints

acolytec3: Does that matter? Once a block builder has reverted, should anyone care about the state manager attached to it? You can't do anything more with the builder at that point.

packages/client/lib/miner/pendingBlock.ts
g11tech: mixhash (prevRandao) is a mandatory field in fcU

packages/client/lib/miner/pendingBlock.ts
g11tech:
```suggestion
```

packages/client/lib/miner/pendingBlock.ts
```diff
@@ -260,7 +305,7 @@ export class PendingBlock {
     }
 
     // Remove from pendingPayloads
```
g11tech, line 307:
```suggestion
```

packages/client/lib/miner/pendingBlock.ts
g11tech: so just have a `type buffer8string= string`?

packages/vm/src/buildBlock.ts
holgerd77: I am also sceptical on this one or at least might not have the complete picture: if someone uses the same state for (other) VM tasks and block building this non-reverting of the StateManager state would definitely change the behavior - as Jochem pointed out - right? 🤔

What is/was the reason to remove this StateManager revert in the first place? Was this necessary respectively does this fix anything?

holgerd77:
Would also have a tendency here, to not try to squeeze this into tomorrow's releases any more, @g11tech let me know if you feel otherwise and you see an urgent need to get in.

g11tech:
> Would also have a tendency here to not try to squeeze this into tomorrow's releases any more, @g11tech let me know if you feel otherwise and you see an urgent need to get in.

sounds good :+1:

packages/vm/src/buildBlock.ts
holgerd77: Update: so I do not find this VM/StateManager copy in the VM itself (unless I am overlooking something), so from the perspective of the VM this is a breaking API (respectively: behavioral change) we should revert. Even in the case copy happens in the VM there still might be some use cases of the block builder (Hardhat using this for their tests or something and relying on this assumption of the state being reset) where this breaks current usages.

Let me know if I am talking non-sense here, but as I read between the lines in the chat this also hasn't been such a high profile change, rather some optimization? So in doubt we should be able to easily revert? 🤔

g11tech: the copy happens in client in `miner/miner.ts` for pre merge and for postmerge `miner/pendingBlock.ts`'s `start` recieves a `vm` copy,
but anyway added the revert logic back as suuggested :+1:

holgerd77 (approved):
LGTM

packages/vm/src/buildBlock.ts
holgerd77: Yes, but relevant for deciding what is breaking in the VM scope is the code inside of the VM package since people are using this stand-alone.

Issue #2528, opened 2023-02-10
Zhejiang empty withdrawals EthereumJS client encoding issue


holgerd77:
On the [Zhejiang](https://zhejiang.ethpandaops.io/) testnet seems to be malformed/wrongly encoded blocks sent by the EthereumJS client in the case that there are no withdrawals associated with a block, see [this](https://discord.com/channels/595666850260713488/1073208138309840926) Eth R&D Discord thread and an associated issue https://github.com/ethereum/go-ethereum/issues/26647 opened on the go-ethereum side.

If I replicate the core of the issue correctly, I think we create blocks at some point where the withdrawals root in the block is correctly being set to the empty trie root but the withdrawals field is decoded (by the other-side-client) as `undefined` or empty instead of an empty list `[]`, which would be the correct structure to use.

Therefore such a block is reported as invalid by e.g. Geth while originally this was a valid block.

I followed some encoding paths already but couldn't imminently find some weak spots, following places to look at:

- Block encoding preparation with `raw()` and then `serialize()` in the Block package
- Eventually some initialization path in the Block library along the static constructors (unlikely though I would say)
- Maybe it's a thing in combination with a wrong Common, unlikely as well if the withdrawals root is present I would say?
- Also a place to look: the block builder in the VM or the outer block building code in the Client package
- And finally: devp2p encoding of block structures in the Client

Happy debugging, I will also continue to try things out! 🧑🏽‍💻

g11tech:
t shouldn't be the issue: since the code flow is: 

```typescript
try {
      body = await this.getBody(hash, number)
    } catch (error: any) {
      if (error.code !== 'LEVEL_NOT_FOUND') {
        throw error
      }
      // If this block had empty withdrawals push an empty array in body
      if (header.withdrawalsRoot !== undefined && header.withdrawalsRoot.equals(KECCAK256_RLP)) {
        body.push([])
      }
    }
```
i.e. if no bodies has been pushed (hats when it will error and come to this catch), then it could be the scenario of empty transactions and empty withdrawals which is why we push []

if there was something in the body, then body would have just been assigned from getBody

holgerd77:
> t shouldn't be the issue: since the code flow is:
> 
> ```ts
> try {
>       body = await this.getBody(hash, number)
>     } catch (error: any) {
>       if (error.code !== 'LEVEL_NOT_FOUND') {
>         throw error
>       }
>       // If this block had empty withdrawals push an empty array in body
[...]

PR #2534, opened 2023-02-17
blockchain: Add extra validations for assuming nil bodies in getBlock

g11tech:
Add extra validations for assuming nil bodies in getBlock

While debugging for 
- https://github.com/ethereumjs/ethereumjs-monorepo/issues/2528

No discrepancy was found: 

synced two ethereumjs peers 
- peer1 from geth  to have full local chain 
  (our EF node has the bug of not serving canonical chain in beacon sync like how we discovered in shandong but should have been fixed with latest hive fixes regarding DB ops)
- peer2 from local ethereumjs and added extra validation for the downloaded block

but haven't found the nil withdrawal body issue so far.

However discovered that we need extravalidations 

i) while importing block from peer
ii) Also while assuming nil bodies when bodies not found in the DB (could be some DB bodies write bug)

Hence  PR for the same

Closes #[2528](https://github.com/ethereumjs/ethereumjs-monorepo/issues/2528)

### DIFF
--- a/packages/blockchain/test/index.spec.ts
+++ b/packages/blockchain/test/index.spec.ts
@@ -475,6 +475,32 @@ tape('blockchain test', (t) => {
     st.end()
   })
 
+  t.test('should test nil bodies / throw', async (st) => {
+    const blocks = generateBlocks(3)
+    const blockchain = await Blockchain.create({
+      validateBlocks: false,
+      validateConsensus: false,
+      genesisBlock: blocks[0],
+    })
+    await blockchain.putHeader(blocks[1].header)
+    // Should be able to get the block
+    await blockchain.getBlock(BigInt(1))
+
+    const block2HeaderValuesArray = blocks[2].header.raw()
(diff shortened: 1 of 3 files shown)


packages/blockchain/test/index.spec.ts
```diff
@@ -475,6 +475,32 @@ tape('blockchain test', (t) => {
...
+      common: blocks[2]._common,
+    })
+    await blockchain.putHeader(block2Header)
+    try {
+      await blockchain.getBlock(BigInt(2))
+      st.fail('block should not be constucted')
+    } catch (e) {
+      st.pass('block not constructed from empty bodies')
+    }
+    st.end()
```
holgerd77, line 501: Nice test setup! 👍

holgerd77 (approved):
Reviewed relatively carefully (since `Blockchain.getBlock()` is relatively central and broadly used functionality, looks good! 🙂

PR #2535, opened 2023-02-17
devp2p: Client Peer Connection SUBPROTOCOL Error Analysis

holgerd77:
This analysis has been triggered by trying to connect with the client to `Sepolia` with `npm run client:start -- --network=sepolia` and being dissatisfied by the level of connectivity getting on startup.

When doing some client runs with the devp2p logger being activated I (re-)discovered that we often have the situation with peer connections "looking good" but then nevertheless terminating with a `SUBPROTOCOL_ERROR`.

This is a problem we have for ages, and having these peer connections go through would be especially valuable, since there *are* only a few with some option for success - many many others refuse with `TOO_MANY_PEERS` - and if these few are terminating unexpectedly to some greater extend that is a substantial loss for client connectivity.

I tried to isolate one of these peer runs to a peer with IP `172.105.167.71` and the standard port `30303`.

On the devp2p side I wrote a small script which we might want to generalize/improve later on and which allows to run one peer connection in isolation.

This script can be run with:

```shell
DEBUG=ethjs,devp2p:* ts-node scripts/tester.ts
```

And gives the following output (the last DISCONNECT is expected):

grafik

The one thing which I realized at some point. On the DPT side after our `ping` answered by a `pong` from the node, there is another `ping` send from the node which we then answer back with a `pong` (so lines 5 and 6). This is not happening *at all* within a client run context (neither for this nor for other peers) (or minimally it is not logged (unlikely)).

I then went ahead and isolated the peer connection in the client context as well by manipulating the Common bootstrap nodes and then run the client with:

```shell
cd ../common && npm run build && cd ../devp2p && npm run build && cd ../client && DEBUG=ethjs,devp2p:*,devp2p:*:* npm run client:start -- --network=sepolia --discV4=false --discDns=false
```

This gives the following output:

grafik

As one can see, the second ping/pong is just not there, instead connection repeatedly halts with a received `DISCONNECT` from the remote node with no additional communication (compared to the devp2p example) in between.

So there is some larger likelihood that the missing ping/pong communication is the root cause of this lost connection.

This makes me somewhat excited since - if we got this tracked - this might substantially/sustainably improve client connectivity (might be something else as well though 😋).

If someone wants to continue with an analysis that is therefore highly welcomed, I will stop for now but leave these notes and the code as being written here.

### DIFF
--- a/packages/devp2p/src/dpt/server.ts
+++ b/packages/devp2p/src/dpt/server.ts
@@ -4,7 +4,7 @@ import { EventEmitter } from 'events'
 import LRUCache = require('lru-cache')
 import ms = require('ms')
 
-import { createDeferred, devp2pDebug, formatLogId, keccak256, pk2id } from '../util'
+import { createDeferred, devp2pDebug, formatLogId, pk2id } from '../util'
 
 import { decode, encode } from './message'
 
@@ -46,7 +46,6 @@ export class Server extends EventEmitter {
   _timeout: number
   _endpoint: PeerInfo
   _requests: Map<string, any>
-  _parityRequestMap: Map<string, string>
   _requestsCache: LRUCache<string, Promise<any>>
   _socket: DgramSocket | null
   _debug: Debugger
@@ -60,7 +59,6 @@ export class Server extends EventEmitter {
     this._timeout = options.timeout ?? ms('10s')
     this._endpoint = options.endpoint ?? { address: '0.0.0.0', udpPort: null, tcpPort: null }
     this._requests = new Map()
-    this._parityRequestMap = new Map()
     this._requestsCache = new LRUCache({ max: 1000, maxAge: ms('1s'), stale: false })
 
     const createSocket = options.createSocket ?? dgram.createSocket.bind(null, { type: 'udp4' })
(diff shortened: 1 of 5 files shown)


packages/devp2p/src/dpt/server.ts
```diff
@@ -149,20 +147,7 @@ export class Server extends EventEmitter {
...
-    // 2018-02-28
-    if (typename === 'ping') {
-      const rkeyParity = keccak256(msg.slice(98)).toString('hex')
-      this._parityRequestMap.set(rkeyParity, msg.slice(0, 32).toString('hex'))
-      setTimeout(() => {
-        if (this._parityRequestMap.get(rkeyParity) !== undefined) {
-          this._parityRequestMap.delete(rkeyParity)
-        }
-      }, this._timeout)
-    }
```
holgerd77, line 165: This is unrelated to the original problem, I just removed "on occasion", but not to confuse anyone.

g11tech (approved):
looks good to me! 

yea with generally peer debugging most of the times issue comesout to be fork-id/genesis mismatch if there is one and very rarely any issue with devp2p (except one where our eth message codes were not correct but rectified log ago by @jochem-brouwer )

else i have seen similar issues of just pure disconnects or too many peers disconnect.

holgerd77:
This was *not* intended to be merged, please read the PR descriptions more carefully, this also has a `do-not-merge` label attached. This is just a test setup replacing/injecting a somewhat random node `172.105.167.71` into the Sepolia chain confiuration where I could replicate the problem with.

This was just very much a hack and intended to be used for further analysis, so we need to revert this PR.

Will directly submit a revert-PR.

PR #2536, opened 2023-02-20
Patch-1

zgayjjf:
340282366920938463463374607431768211455 equals to BigInt(2) ** BigInt(128) - BigInt(1).

If we use formular instead of the direct result, in some case, boundler like webpack will turn `A ** B` into `Math.pow(A, B)`, which doesn't support BigInt type thus causing error.

Changing the `browserlist` field in `package.json` as below can be a workaround. But it's better not to ask the developer to change their building config.

https://github.com/paulmillr/noble-ed25519/issues/23#issuecomment-1109248320

This is a known issue as https://github.com/facebook/create-react-app/issues/10785


holgerd77 (dismissed):
Hi there,
makes sense to me, thanks for the PR! 🙂 

Will approve and merge once tests pass, last chance for others to have a look. 😋

holgerd77 (approved):
Have added a lint fix, LGTM!

holgerd77:
Rebased this via UI.

PR #2539, opened 2023-02-21
Fix broken URL to `packages/statemanager`

ernestognw:
## Description

Just found a broken link while reading :)


acolytec3 (approved):
Thanks for the catch!  Will merge this once our next round of releases rolls out.

PR #2540, opened 2023-02-21
Client: Move to async Blockchain constructor / fix --startBlock option bug

holgerd77:
This moves the client to a safe Blockchain initialization by switching to async static create() constructors for the `EthereumClient` and `Chain` classes. This should be generally beneficial to client stability cause with this safe and finalized Blockchain initialization can be guaranteed.

In this case this fixes a problem with the `--startBlock` option which does an early blockchain access and was before failing.

@g11tech Ready for review, feel free to approve and merge or alternatively push own work on top, whatever is better/more convenient.

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -398,6 +398,7 @@ async function startClient(config: Config, customGenesisState?: GenesisState) {
   const dbs = initDBs(config)
 
   let blockchain
+  console.log(customGenesisState)
   if (customGenesisState !== undefined) {
     const validateConsensus = config.chainCommon.consensusAlgorithm() === ConsensusAlgorithm.Clique
     blockchain = await Blockchain.create({
@@ -411,7 +412,7 @@ async function startClient(config: Config, customGenesisState?: GenesisState) {
     config.chainCommon.setForkHashes(blockchain.genesisBlock.hash())
   }
 
-  const client = new EthereumClient({
+  const client = await EthereumClient.create({
     config,
     blockchain,
     ...dbs,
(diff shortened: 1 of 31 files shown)


packages/client/bin/cli.ts
```diff
@@ -391,6 +391,7 @@ async function startClient(config: Config, customGenesisState?: GenesisState) {
   const dbs = initDBs(config)
 
   let blockchain
+  console.log(customGenesisState)
```
holgerd77, line 394: Ah, shit. 😋

Will remove.

packages/client/bin/cli.ts
```diff
@@ -404,20 +404,20 @@ async function startClient(config: Config, customGenesisState?: GenesisState) {
     config.chainCommon.setForkHashes(blockchain.genesisBlock.hash())
   }
 
-  const client = new EthereumClient({
+  const client = await EthereumClient.create({
     config,
     blockchain,
     ...dbs,
   })
+  await client.open()
```
holgerd77, line 412: This moving over was also necessary and I already thought "Oh would this have been the one-line fix already?" 🤪 but then I did the counter-test by using `master` and just moving the line and it actually didn't work (thanks god 😂) but broke with a `LEVEL_NOT_FOUND` error, actually another indicator that this new from-the-top-async-flow is more healthy to have things in place.

g11tech (approved):
lgtm

g11tech (changes requested):
blocking for further fixes (reuse of startblock just after first use breaks if chain doesn't sync)

g11tech (approved):
lgtm

g11tech (approved):

PR #2541, opened 2023-02-21
Fix the fetcher errored property colliding with Readable class

g11tech:
As encountered by @holgerd77 , fetcher (and client) would break on error handling because seemingly `errored` property of fetcher was conflicting with the `Readable` class from which fetcher extends causing such errors:
https://nodejs.org/api/stream.html#readableerrored

image

this fix renames the property to a non conflicting one which doesn't exit the client and keeps it going (forced error throw here to test)

image


acolytec3 (commented):
LGTM!

holgerd77 (approved):
LGTM

Issue #2390, opened 2022-10-27
Support receipts when mining/sealing blocks


acolytec3:
The client currently does not honor the `saveReceipts` CLI parameter when mining/sealing blocks with either PoW or PoA consensus.  Ideally, the receipts should be caught [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/1e903db3f29f7b4f93db3b2c61248a3c1babad27/packages/client/lib/miner/miner.ts#L286) and stored around [this point](https://github.com/ethereumjs/ethereumjs-monorepo/blob/1e903db3f29f7b4f93db3b2c61248a3c1babad27/packages/client/lib/miner/miner.ts#L308) when the miner is done building the block and has put it in the chain.

gabrocheleau:
Closing this since this has been marked as fixed by #2544

PR #2544, opened 2023-02-22
Miner-support-for-saveReceipts

ScottyPoi:
Fixes issue #2390
`Support receipts when mining/sealing blocks`

### DIFF
--- a/packages/client/lib/miner/miner.ts
+++ b/packages/client/lib/miner/miner.ts
@@ -293,7 +293,9 @@ export class Miner {
         const txResult = await blockBuilder.addTransaction(txs[index], {
           skipHardForkValidation: this.skipHardForkValidation,
         })
-        this.config.saveReceipts && receipts.push(txResult.receipt)
+        if (this.config.saveReceipts) {
+          receipts.push(txResult.receipt)
+        }
       } catch (error) {
         if (
           (error as Error).message ===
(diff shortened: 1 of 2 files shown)


packages/client/lib/miner/miner.ts
```diff
@@ -286,11 +287,13 @@ export class Miner {
...
     let index = 0
     let blockFull = false
+    const receipts: TxReceipt[] = []
     while (index < txs.length && !blockFull && !interrupt) {
       try {
-        await blockBuilder.addTransaction(txs[index], {
+        const txResult = await blockBuilder.addTransaction(txs[index], {
           skipHardForkValidation: this.skipHardForkValidation,
         })
+        this.config.saveReceipts && receipts.push(txResult.receipt)
```
holgerd77, line 296: Short nit: won't make it a blocker but I think it is better to write these kind of things "in the old fashioned way" with `if (...) { // push the receipts }`, this is otherwise for a lot of developers unnecessarily hard to read (on a very very rough estimate I would assume that 50% of people know this and 50% do not or are at least not so sure or getting wonky on these kind of code parts).

holgerd77 (approved):
Looks good, cool, I guess we can still take this in (the releases, if we will ever do 😅), very local code changes and good test coverage! 🤩

packages/client/lib/miner/miner.ts
holgerd77: Update: I needed to do a small update to fix client test runs anyhow, so I have updated to easier to read conditionals along.

ScottyPoi: I didn't realize that the `&&`  was such a hip new thing! :stuck_out_tongue: 

I can avoid this in the future

ScottyPoi:
@holgerd77 does the one test I added feel sufficient?

packages/client/lib/miner/miner.ts
holgerd77: The hip new programmers for sure know this, but I e.g. already needed two additional brain cycles to get this deciphered. 😂

holgerd77 (approved):
Thanks for updating, looks good!

PR #2546, opened 2023-02-23
devp2p: Fix DNS Discovery ENR record decoding

holgerd77:
This fixes errors like `devp2p:dns:dns Errored searching DNS tree at subdomain YNSF7U3LFSTT4YF7Z37KP5XRZU: Error: Invalid padding: string should have whole number of bytes +1ms` for client DNS discovery.

Turned out that our decoding of ENR text entries what slightly wrong assume decoding would be plain base64 (had to google myself) while this was some modification. Figured this out in exchange with the Go Ethereum team (thanks for that! 🙏), especially Martin got the crucial hint with providing the following information:

> RawURLEncoding, which is the unpadded alternate base64 encoding defined in RFC 4648
> Records need to prepared like the following: replace - wth +, replace _ with / and add padding

which I directly took over into the code docs.

While also not being a connection "game changer", this should substantially (and substainably) improve client connections on primarly the testnets (which are most dependent on good DNDS discovery) since before 30-50% of the DNS peers were unnecessarily lost due to this bug.

Ah, associated EIP is actually [EIP-1459](https://eips.ethereum.org/EIPS/eip-1459), which is an interesting read, easier to grasp then I first assumed, gave this a first somewhat deeper read yesterday.


g11tech (approved):
awesome!

PR #2548, opened 2023-02-23
blockchain: Revert to previous sane heads if block or header put fails

g11tech:
Earlier when the putblockheader errored on hf bug, the blockchain would be corrupted because headblockhash/headheaderhash have dirty updates and were not being reverted

This PR reverts any head changes in case there is any putblock error

### DIFF
--- a/packages/blockchain/src/blockchain.ts
+++ b/packages/blockchain/src/blockchain.ts
@@ -451,100 +451,113 @@ export class Blockchain implements BlockchainInterface {
    */
   private async _putBlockOrHeader(item: Block | BlockHeader) {
     await this.runWithLock<void>(async () => {
-      const block =
-        item instanceof BlockHeader
-          ? new Block(item, undefined, undefined, {
-              common: item._common,
-            })
-          : item
-      const isGenesis = block.isGenesis()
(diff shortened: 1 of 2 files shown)


packages/blockchain/src/blockchain.ts
```diff
@@ -451,100 +451,113 @@ export class Blockchain implements BlockchainInterface {
...
+          // save hash to number lookup info even if rebuild not needed
+          dbOps.push(DBSetHashToNumber(blockHash, blockNumber))
+        }
+
+        const ops = dbOps.concat(this._saveHeadOps())
+        await this.dbManager.batch(ops)
+
+        await this.consensus.newBlock(block, commonAncestor, ancestorHeaders)
+      } catch (e) {
+        // restore head to the previouly sane state
```
g11tech, line 555: main diff is this setting old heads and assigning oldheads back, rest is because of moving inside try catch here

holgerd77 (approved):
Looks good, thanks for digging so deep here, these are generally valuable structural changes, makes so much sense to have this wrapped catching and eventually go back to the old head on errors. 🙏

Will merge and then update the release PRs.

PR #2550, opened 2023-02-23
Fix invalid head block reset bug

acolytec3:
Fixes a logic bug related to reorg in skeleton sync where during a reorg that occurs during `fillCanonicalHead` where a skeleton subchain extends backwards passed the current canonical head block (due to a reorg scenario).
Currently, we reset the canonical head block on the chain to match the new skeleton block that we're putting in the chain when a reorg occurs via skeleton sync (i.e. the CL client reports our current head block in an FcU, then reports a different head block in the next FcU and then our client finds via the `ReverseFetcher` that the new head block announced in the FcU involves a reorg away from our current head block).  If that first skeleton block on the reorg chain is invalid for whatever reason, the chain's canonical head block is left as the original head's parent block (which is incorrect since the whole reorg was invalid).  This PR puts the original head block back as the canonical head.

### DIFF
--- a/packages/client/lib/sync/skeleton.ts
+++ b/packages/client/lib/sync/skeleton.ts
@@ -657,6 +657,7 @@ export class Skeleton extends MetaDBManager {
       } catch (e) {
         this.config.logger.error(`fillCanonicalChain putBlock error=${(e as Error).message}`)
         if (resetHead === true && oldHead !== null) {
+          // Put original canonical head block back if reorg fails
           await this.chain.putBlocks([oldHead], true)
         }
       }
--- a/packages/client/lib/sync/skeleton.ts
+++ b/packages/client/lib/sync/skeleton.ts
@@ -608,7 +608,6 @@ export class Skeleton extends MetaDBManager {
 
     let canonicalHead = this.chain.blocks.height
     let oldHead = null
-    let resetHead = false
     const subchain = this.status.progress.subchains[0]!
     if (this.status.canonicalHeadReset) {
       oldHead = this.chain.blocks.latest // Grab previous head block in case of resettng canonical head
@@ -627,7 +626,6 @@ export class Skeleton extends MetaDBManager {
       canonicalHead = newHead
       await this.chain.resetCanonicalHead(canonicalHead)
       this.status.canonicalHeadReset = false
-      resetHead = true
     }
 
     const start = canonicalHead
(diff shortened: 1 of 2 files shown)


packages/client/lib/sync/skeleton.ts
```diff
@@ -652,6 +656,10 @@ export class Skeleton extends MetaDBManager {
         numBlocksInserted = await this.chain.putBlocks([block], true)
       } catch (e) {
         this.config.logger.error(`fillCanonicalChain putBlock error=${(e as Error).message}`)
+        if (resetHead === true && oldHead !== null) {
+          // Put original canonical head block back if reorg fails
+          await this.chain.putBlocks([oldHead], true)
```
g11tech, line 661: this should again rebuild canonical references till this oldhead right?

packages/client/lib/sync/skeleton.ts
```diff
@@ -652,6 +656,10 @@ export class Skeleton extends MetaDBManager {
         numBlocksInserted = await this.chain.putBlocks([block], true)
       } catch (e) {
         this.config.logger.error(`fillCanonicalChain putBlock error=${(e as Error).message}`)
+        if (resetHead === true && oldHead !== null) {
```
g11tech, line 659:
```suggestion
        if (oldHead !== null) {
```
i think this should be sufficient since old head is only assigned in reset scenarios

packages/client/lib/sync/skeleton.ts
acolytec3: I'm not 100% sure since the reorg scenario we're talking about here is just a 1 block reorg where `fillCanonical` resets the head by one block.  It resolves this hive test - `engine-transition/Transition to a Chain with Invalid Terminal Block, Higher Configured Total Difficulty` which only tests where there is a one block reorg.  That said, it seems like it should work since the reorg only resets the canonical head by 1 block I think.

packages/client/lib/sync/skeleton.ts
```diff
@@ -652,6 +656,10 @@ export class Skeleton extends MetaDBManager {
         numBlocksInserted = await this.chain.putBlocks([block], true)
       } catch (e) {
         this.config.logger.error(`fillCanonicalChain putBlock error=${(e as Error).message}`)
+        if (oldHead !== null) {
```
g11tech, line 659:
```suggestion
        if (oldHead !== null && oldHead >= block.header.number) {
```
should this also be done?

packages/client/lib/sync/skeleton.ts
g11tech: may be we extend the resetHead test i put in blockchain to restore the head and see if we can do block by number in the mid section (of reset head and the old head)

acolytec3: Yep, lemme take a look and work on it some.  Your second check fixed the current broken test.

g11tech (approved):

PR #2552, opened 2023-02-24
Fix kzg imports in `tx`

acolytec3:
The latest release of `tx` breaks installs where `c-kzg` isn't also included because the `blobHelpers` file points to `c-kzg` instead of our own internal `kzg` interface.  This resolves the import breakage.


g11tech (approved):
lgtm

g11tech (approved):

PR #2557, opened 2023-02-28
Turn off libp2p transport by default

acolytec3:
Makes the default transport in the client rlpx only for now since libp2p transport is only semi-functional and can lead to hard to debug port conflicts for code that is not currently being used

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -38,7 +38,7 @@
     "clean": "../../config/cli/clean-package.sh",
     "client:start": "ts-node bin/cli.ts",
     "client:start:dev1": "npm run client:start -- --discDns=false --discV4=false --bootnodes",
-    "client:start:dev2": "npm run client:start -- --discDns=false --discV4=false --transports=rlpx --port=30304 --dataDir=datadir-dev2",
+    "client:start:dev2": "npm run client:start -- --discDns=false --discV4=false --port=30304 --dataDir=datadir-dev2",
     "coverage": "c8 --all --reporter=lcov --reporter=text npm run test:unit",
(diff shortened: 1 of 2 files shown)


packages/tx/tx
```diff
@@ -0,0 +1 @@
+./tx
```
holgerd77, line 1: You've got some strange thing in here which (likely) leads the (tx) tests to fail.

packages/client/package.json
```diff
@@ -38,7 +38,7 @@
     "clean": "../../config/cli/clean-package.sh",
     "client:start": "ts-node bin/cli.ts",
     "client:start:dev1": "npm run client:start -- --discDns=false --discV4=false --bootnodes",
-    "client:start:dev2": "npm run client:start -- --discDns=false --discV4=false --transports=rlpx --port=30304 --dataDir=datadir-dev2",
+    "client:start:dev2": "npm run client:start -- --discDns=false --discV4=false --port=30304 --dataDir=datadir-dev2",
```
holgerd77, line 41: Oh, THAT'S an impressive catch. 🙂 🙃

packages/tx/tx
acolytec3: Interesting. I must have accidentally committed something else and didn't see it when I was writing the commit message

acolytec3: OKay, removed the errant commit.  Should be good now.

packages/client/package.json
acolytec3: Only because I'm pretty sure I originally added that.  :wink:

g11tech (approved):
:+1:

holgerd77 (approved):
LGTM

PR #2560, opened 2023-03-02
monorepo: upgrade to ethers v6

### LINKED ISSUE #2559: Switch `ethers` to peer dependency in `block`/`tx` and upgrade to v6
We currently import `ethers` in `block`, `tx`, and `statemanager` and have it pinned to version 5.7.  We should do the following:
- Move ethers to an optional peer dependency in `block` and `tx`
- Update the `fromEthersProvider` static constructors in `block` and `tx` to check if `ethers` is instantiated and throw a helpful error message if not (i.e. 'please install Ethers to use this function')
- Update Ethers to v6 (and make any necessary changes based on the updated API - probably most substantial in `statemanager`)

gabrocheleau:
Fixes #2559 

### TODO

- [ ] Update the fromEthersProvider static constructors in block and tx to check if ethers is instantiated and throw a helpful error message if not (i.e. 'please install Ethers to use this function')
- [ ] Fix build error TS1452 (see discord discussion)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,6 +45,11 @@
         "npm": ">=7"
       }
     },
+    "node_modules/@adraffy/ens-normalize": {
+      "version": "1.8.9",
+      "resolved": "https://registry.npmjs.org/@adraffy/ens-normalize/-/ens-normalize-1.8.9.tgz",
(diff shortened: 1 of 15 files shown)


holgerd77:
Thanks for start experimenting on this. Generally this is a breaking change though, and we need to wait at least until breaking change work has started and then submit towards the new branch. Also read up on your discussions around problems with Ethers ESM build, so might generally be best to start/continue with this a bit later in the pipeline.

holgerd77:
Superseeded by #2633, will close.

PR #2565, opened 2023-03-02
Client: LMDB Database Test

holgerd77:
This is a test to replace the LevelDB state database with an [LMDB](https://github.com/kriszyp/lmdb-js) to get an impression about the effect on the performance of VM execution.

WIP

Client is building, client run is hanging after the following steps:

grafik

Seems this is not processing further at the moment in the first state DB accessing method (`VmState.generateCanonicalGenesis()`).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2743,15 +2743,75 @@
       }
     },
     "node_modules/@lmdb/lmdb-darwin-arm64": {
-      "version": "2.6.8",
+      "version": "2.7.9",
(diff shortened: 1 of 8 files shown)


acolytec3:
Did you try just running the trie tests with the lmdb database passed in? I'm curious if it works at all

holgerd77:
> Did you try just running the trie tests with the lmdb database passed in? I'm curious if it works at all

Yeah, that's what I would try next. Going directly to client might be some steps too much. I would now go from bottom to top next round. So start with Trie, then Blockchain, then VM, then Client. 🙂

PR #2567, opened 2023-03-04
tx: Update eip4844 txs to decoupled blobs spec

g11tech:
Blobs and now decoupled and don't come in a sidecar, which has implications for eip4844 txs replacing kzg_aggregate_proof with proofs (confirmed on the c-kzg tele channel)

Still waiting on the EIP-4844 update for the same and hence will remain WIP till eip updated
Ref:
 - https://github.com/ethereum/EIPs/pull/6610


acolytec3:
This looks good to me as far as it goes.

acolytec3 (dismissed):
LGTM!

acolytec3:
Unless I broke something else, this should be good to go now.  I fixed the karma test issues we were having in `block` and `tx`.

acolytec3 (approved):
Approved (with concurring opinion from @g11tech on my edits)

PR #2568, opened 2023-03-06
VM: bnadd/bnmul Precompile Performance Optimization

holgerd77:
Replaces #1570

(second commit optimizations from the above PR already done along some subsequent jumpdest analysis/optimization work)


acolytec3 (approved):

acolytec3:
Looks great!  I tested this on my Intel Silicon based laptop and it definitely improves performance on these precompiles.

PR #2570, opened 2023-03-06
EVM: Memory Fix & Other Optimizations

holgerd77:
This cuts performance test 1 and 3 in roughly half by avoiding an additional buffer copy on memory read (so 0f0e2828c3eebe0c0f43378bb54784cb9b8df72c is the essential commit here). Additionally there are some more performance optimizations and debug logger enhancements included.

### DIFF
--- a/packages/evm/src/memory.ts
+++ b/packages/evm/src/memory.ts
@@ -70,7 +70,7 @@ export class Memory {
     const returnBuffer = Buffer.allocUnsafe(size)
     // Copy the stored "buffer" from memory into the return Buffer
 
-    const loaded = Buffer.from(this._store.slice(offset, offset + size))
+    const loaded = this._store.slice(offset, offset + size)
     returnBuffer.fill(loaded, 0, loaded.length)
 
     if (loaded.length < size) {
--- a/packages/evm/src/evm.ts
+++ b/packages/evm/src/evm.ts
@@ -789,25 +789,15 @@ export class EVM implements EVMInterface {
       result.execResult.selfdestruct = {}
       result.execResult.gasRefund = BigInt(0)
     }
-    if (err) {
-      if (
-        this._common.gteHardfork(Hardfork.Homestead) ||
-        err.error !== ERROR.CODESTORE_OUT_OF_GAS
-      ) {
-        result.execResult.logs = []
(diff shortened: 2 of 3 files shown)


packages/evm/src/memory.ts
```diff
@@ -70,7 +70,7 @@ export class Memory {
     const returnBuffer = Buffer.allocUnsafe(size)
     // Copy the stored "buffer" from memory into the return Buffer
 
-    const loaded = Buffer.from(this._store.slice(offset, offset + size))
+    const loaded = this._store.slice(offset, offset + size)
```
jochem-brouwer, line 73: Had to explicitly check that this could not lead to unwanted writes:

- `slice` is a view point, so if you edit this buffer, you edit the "original" buffer too
- `loaded` is not written to in this context
- The `returnBuffer` is `fill`ed, and this actually copies the value (so if you edit it, it does not change the original)

So this is great :)

jochem-brouwer (approved):
LGTM

packages/evm/src/memory.ts
holgerd77: This is half-great. Martin (Holst Swende) explicitly said, that one of the core things we are doing not optimal in our VM is that we copy things around, while we should use the original pointer to memory (the test this half-fixes was actually that there was huge call data created and we copy this over and over on each repeated call).

So we should rather make sure that we also remove this `loaded` copy over time and double check that there *are* just no writes happening. This will eventually/likely give another huge/significant speed bump if we do this right.

Independent from the approval on this PR, since this only eliminates *one* unnecessary copy. So safe anyhow. 🙂 

Will merge in.

holgerd77: (updated the comment above for better readability)

jochem-brouwer: I think this is fine, `loaded` is not a copy, it is a view, so this should be a super fast operation.

holgerd77: Update: just did the test, these are the numbers of the performance tests when we remove this `returnBuffer` copy (this is not out of the box working yet though and needs subsequent fixes, tested on some mainnet blocks).

So these are the new values:

grafik

And these are the old ones:

grafik

PR #2577, opened 2023-03-09
Client: ensure safe/finalized blocks are part of the canonical chain on forkchoiceUpdated

jochem-brouwer:
This fixes these hive tests:

- `engine-api/Inconsistent Head in ForkchoiceState`
- `engine-api/Inconsistent Safe in ForkchoiceState`
- `engine-api/Inconsistent Finalized in ForkchoiceState`

To run: `go run ./hive.go --client ethereumjs --sim ethereum/engine --sim.limit "engine-api/Inconsistent" --docker.output`

(This runs above three tests)

### DIFF
--- a/packages/client/test/rpc/engine/forkchoiceUpdatedV1.spec.ts
+++ b/packages/client/test/rpc/engine/forkchoiceUpdatedV1.spec.ts
@@ -271,7 +271,7 @@ tape(`${method}: latest block after reorg`, async (t) => {
   ])
 
   expectRes = (res: any) => {
-    t.equal(res.body.result.payloadStatus.latestValidHash, blocks[1].blockHash)
+    t.equal(res.body.error.code, -32602)
   }
   await baseRequest(t, server, req, 200, expectRes)
 })
--- a/packages/client/test/rpc/engine/forkchoiceUpdatedV1.spec.ts
+++ b/packages/client/test/rpc/engine/forkchoiceUpdatedV1.spec.ts
@@ -1,15 +1,19 @@
 import { Block, BlockHeader } from '@ethereumjs/block'
+import { Chain, Common, Hardfork } from '@ethereumjs/common'
 import { bufferToHex, zeros } from '@ethereumjs/util'
 import * as tape from 'tape'
 import * as td from 'testdouble'
 
 import { INVALID_PARAMS } from '../../../lib/rpc/error-code'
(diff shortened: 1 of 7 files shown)


g11tech (approved):
lgtm

jochem-brouwer:
Client test fails :thinking:

packages/client/test/rpc/engine/forkchoiceUpdatedV1.spec.ts
```diff
@@ -271,7 +271,7 @@ tape(`${method}: latest block after reorg`, async (t) => {
   ])
 
   expectRes = (res: any) => {
-    t.equal(res.body.result.payloadStatus.latestValidHash, blocks[1].blockHash)
+    t.equal(res.body.error.code, -32602)
```
jochem-brouwer, line 274: Note; this FCU payload is indeed invalid, because `headBlock` has block number 1, while  `safeBlock` has block number 2 (and `finalBlock` also)

g11tech:
@jochem-brouwer can we push the coverage a bit up here, i think we just need test for the failure cases of being non canonical i think

packages/client/test/rpc/engine/forkchoiceUpdatedV1.spec.ts
```diff
@@ -271,7 +295,93 @@ tape(`${method}: latest block after reorg`, async (t) => {
...
+
+  const expectRes = (res: any) => {
+    t.equal(res.body.error.code, -32602)
+    t.ok(res.body.error.message.includes('Safe'))
+    t.ok(res.body.error.message.includes('canonical'))
+  }
+  await baseRequest(t, server, req, 200, expectRes)
+})
+
+tape(`${method}: validate finalizedBlockHash is part of canonical chain`, async (t) => {
```
jochem-brouwer, line 346: Note: had to copy the previous test, because if `baseRequest` encounters an error, it ends the test suite thus you cannot test anything else afterwards.

jochem-brouwer:
Added tests (which was quite a bit of work), which essentially recreate the hive tests. This should also bump the coverage.

Side note: need to somewhat cleanup these tests at some point:

- Ensure `baseRequest` has an option to not close the server in case of an error (otherwise in a single test it is only possible to check for a single error, and this can only be the final request)
- Ensure imports are moved to an util, if methods are imported from test files, this will run all these tests again.

g11tech (approved):
superb! :+1:

PR #2578, opened 2023-03-10
Client: Shanghai dDoS Resilience -> New Trie CP Mechanism, VM State CP Refactorings, Diff-based StateManager Cache

holgerd77:
WIP

Issue #1536, opened 2021-10-20
Client/VM/MPT: Shanghai dDoS Attacks Resilience


holgerd77:
Our client is currently unsyncable along the Shanghai dDoS attack blocks with not having passed this ~ 500,000 blocks range even after a couple of days syncing.

The Shanghai dDoS attacks happened at the time of Devcon 2 during September/October 2016 affecting blocks 2,286,910 to 2,717,576, see e.g. [Defeating the Ethereum DDos Attacks](https://tjayrush.medium.com/defeating-the-ethereum-ddos-attacks-d3d773a9a063), Feb 12 2018, Thomas Jay Rush.

The Shanghai dDoS attacks are not a single attack but a series of several subsequent attacks all with different or at least adopted and modified attack vectors.

While there were some things structurally fixed by the following `TangerineWhistle` and `SpuriousDragon` HFs there have been several client (geth) bug fix releases with non-consensus-touching fixes which are just removing implementation suboptimalities in data structures and memory management and the like.

To make our client more robust and not be vulnerable to similar attacks in the future (and practically hopefully be able again to sync the affected block ranges) we should at least have a short look at each of the attacks + fixes, see how complex an implementation would be and if we might have already implemented a similar solution or not.

The different attack vectors can be followed upon in Martin Holst Swende's talk [here](https://www.youtube.com/watch?v=xl06E2kwi-U) from EDCON in March 2017. [This presentation](https://archives.sector.ca/presentations18/SarahAndMaurelianEthereum.pdf) is also a valuable source for some insight.

References:

- https://www.bokconsulting.com.au/blog/ethereum-network-attackers-ip-address-is-traceable/

## From Shanghai With Love

State caching mechanism attack with recursive `DELEGATECALL`.

Date: September 18 2016
Block(s): [2,283,416](https://etherscan.io/block/2283416)
Talk Martin: [1:20](https://youtu.be/xl06E2kwi-U?t=80)
Bugfix Release: Geth [v1.4.12](https://github.com/ethereum/go-ethereum/releases/tag/v1.4.12) "From Shanghai With Love"
Fixing PR: https://github.com/ethereum/go-ethereum/pull/3006

### Vulnerability Analysis ("Are we vulnerable?")

[...]

Issue #2406, opened 2022-11-04
Homestead: ShanghaiLove_Homestead failing


jochem-brouwer:
When running tests with `test-all-hardforks` label, CI now always fails since `ShanghaiLove_Homestead` makes node run out of memory. 

When I go back into time, for instance: c70a1585bb0ce3c8ad912189955c468677875d2d, which is #2244 (this did test all hardforks, including `ShanghaiLove_Homestead` which passed), it now locally also fails.

Starting to think this is not our problem, but the latest node versions might have a problem?

Note, to test;

```
cd packages/vm
npm run test:blockchain -- --fork=Homestead --test=ShanghaiLove_Homestead
```

The problem in this test is that in ShanghaiLove, the contract keeps `DELEGATECALL`ing itself, this copies a ton of code. In TangerineWhistle this is solved because now the cost of `DELEGATECALL` goes from 40 to 700, so you cannot reach a super deep stack.

Node sample err;

```
<--- Last few GCs --->

[50229:0x592db50]    37417 ms: Scavenge 2026.8 (2079.0) -> 2020.8 (2080.0) MB, 5.1 / 0.0 ms  (average mu = 0.668, current mu = 0.382) allocation failure; 
[50229:0x592db50]    37433 ms: Scavenge 2027.6 (2080.0) -> 2021.6 (2080.7) MB, 4.5 / 0.0 ms  (average mu = 0.668, current mu = 0.382) allocation failure; 
[50229:0x592db50]    37449 ms: Scavenge 2028.5 (2080.7) -> 2022.4 (2096.7) MB, 6.3 / 0.0 ms  (average mu = 0.668, current mu = 0.382) allocation failure; 


<--- JS stacktrace --->

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
 1: 0xb6dd00 node::Abort() [node]
 2: 0xa7db1c  [node]
 3: 0xd474e0 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
 4: 0xd47887 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [node]
 5: 0xf24c45  [node]
 6: 0xf3712d v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [node]
[...]

PR #2581, opened 2023-03-13
VM: Diff-based Touched Accounts Checkpointing

holgerd77:
Part of #1536

Extracted from #2578

PR switches to diff-based touched accounts checkpointing in the VM. This helps to solve the first of Shanghai dDoS attacks going into a recursive `DELEGATECALL` starting at block `2283416` (DAO HF) and repeated in subsequent block txs (with "normally" running blocks in between) and leading our client to crash with heap out of memory ("Javascript Ineffective mark-compacts near heap limit Allocation failed") (a bit in contrast to the attack in #1536 described by me as well as just stating the block execution getting very slow interestingly, not sure if we might have added additional (memory) weight here in the meantime. So this is on an 8GB Apple MacBook Air M1 machine) (I do know about this option to just increase the Node.js memory limit but I explicity *did not* want to just do *that*, seems not a "solution" to me).

Here is a screenshot (don't get confused by block-number-misleading "Latest local block" message):

grafik

When looking at the debug logs of the block/tx run by running:

```shell
DEBUG=ethjs,evm:*,evm:*:*,vm:*,vm:*:*,statemanager:* npm run client:start -- --discDns=false --discV4=false --maxPeers=0 --executeBlocks=2283416
```

one can see that hundreds of `DELEGATECALL`s are executed subsequently, each one happening from within the previous delegate call and each creating a new checkpoint without committing for 100s of delegate calls. With the previous checkpointing implementation for touched accounts this lead to the whole account structure being copied over and over again, even when there are no changes, leading to a lot of memory bloat.

Situation will get better at the [Tangerine Whistle](https://eips.ethereum.org/EIPS/eip-608) HF at block `2463000` increasing the gas costs for call operations and other I/O heavy opcodes (in case of `DELEGATECALL` from `40` to `700`, so pretty substantially), so that such an attack is not that effective any more.

This is nevertheless a lasting and structural change improving the overall VM efficiency. The functionality (change) should be sufficiently covered by existing tests.

Note that this change alone is *not* solving the out-of-memory crash though yet, I will submit a subsequent PR with a separate cache improvement which will completely solve the issue. PR is nevertheless open for review and can be merged. 🙂

We have a structurally similar problem with the way we handle accessed storage checkpointing in the VM, which we should also tackle (also open for grasps, a bit more complex but also not too much, solution should be structurally very similar). For now I "solved" this for the current context I am working on by guarding the cache to just get used for Berlin HF and upwards, which generally makes sense since before it is just an unnecessarily build-up structure.

Small third change in this PR is to use the direct memory access (see #2573 for more context) for the `MLOAD` opcode, which should be very much safe since this will be converted to `BigInt` directly afterwards anyhow. So this just saves an unnecessary in-between copy.

### DIFF
--- a/packages/vm/src/eei/journaling.ts
+++ b/packages/vm/src/eei/journaling.ts
@@ -1,0 +1,82 @@
+export class Journaling<T> {
+  public journal: Set<T>
+  protected journalStack: { [key: number]: Set<T> }
+  protected journalHeight: Map<T, number>
+  protected height: number
+
+  constructor() {
+    this.journal = new Set()
+    this.journalStack = {}
+    this.journalHeight = new Map()
+    this.height = 0
+  }
+
+  clear() {
+    this.journal.clear()
+    this.journalStack = {}
+    this.journalHeight = new Map()
+    this.height = 0
+  }
+
+  checkpoint() {
+    this.height++
+  }
+
+  revert(ignoreItem?: T) {
+    const height = this.height
+    if (height in this.journalStack) {
+      for (const key of this.journalStack[height]) {
+        // Exceptional case due to consensus issue in Geth and Parity.
+        // See [EIP issue #716](https://github.com/ethereum/EIPs/issues/716) for context.
+        // The RIPEMD precompile has to remain *touched* even when the call reverts,
+        // and be considered for deletion.
+        if (key === ignoreItem) {
+          continue
+        }
+
+        if (this.journal.has(key) && this.journalHeight.get(key)! >= height) {
+          this.journal.delete(key)
+          this.journalHeight.delete(key)
+        }
+      }
+      delete this.journalStack[height]
+    }
+    this.height--
+  }
+
+  commit() {
+    const height = this.height
+    if (height in this.journalStack) {
+      // Copy the items-to-delete in case of a revert into one level higher
+      if (height !== 1) {
+        if (this.journalStack[height - 1] === undefined) {
+          this.journalStack[height - 1] = new Set()
+        }
+        for (const address of this.journalStack[height]) {
+          this.journalStack[height - 1].add(address)
+          if (this.journalHeight.get(address) === height) {
+            this.journalHeight.set(address, height - 1)
+          }
+        }
+      } else {
+        this.journal = new Set()
+        this.journalHeight = new Map()
+      }
+      delete this.journalStack[height]
+    }
+    this.height--
+  }
+
+  addJournalItem(input: T) {
+    const height = this.height
+    if (!(height in this.journalStack)) {
+      this.journalStack[height] = new Set()
+    }
+    this.journalStack[height].add(input)
+
+    this.journal.add(input)
+    if (this.journalHeight.get(input) === undefined) {
+      this.journalHeight.set(input, height)
+    }
+  }
+}
--- a/packages/vm/src/eei/journaling.ts
+++ b/packages/vm/src/eei/journaling.ts
@@ -48,7 +48,7 @@ export class Journaling<T> {
     const height = this.height
     if (height in this.journalStack) {
       // Copy the items-to-delete in case of a revert into one level higher
-      if (height !== 1) {
+      if (height > 0) {
         if (this.journalStack[height - 1] === undefined) {
           this.journalStack[height - 1] = new Set()
         }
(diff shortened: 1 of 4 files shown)


holgerd77:
Ah, interestingly some selected tests are failing here. Would be glad on some support on this (during the week would be enough), if someone has some ressources to spare (@jochem-brouwer or @ScottyPoi e.g.?).

jochem-brouwer:
Will take a look this afternoon! :smile:

packages/vm/src/eei/vmState.ts
```diff
@@ -286,6 +325,7 @@ export class VmState implements EVMStateAccess {
       }
     }
     this._touched.clear()
+    this._touchedStack = {}
```
jochem-brouwer, line 328: The `touchedStack` is not empty for some reason here, which makes me not super confident that this is all correct. After running `npm run test:blockchain -- --fork=Istanbul --test=RevertPrecompiledTouch_d0g0v0_Istanbul` there are still fields at the `0` and `2` keys.

jochem-brouwer:
I think that the current approach is not correct @holgerd77. Here is a situation:

Tx calls a contract. This contract touches the address `A`. Now, the contract calls itself again (`checkpoint`), it now touches `A` again, and now this call frame reverts. This **deletes** `A` from the touched accounts. However, it is clear that `A` should be touched. After finishing the Tx, `A` should be checked and if it is empty, delete the account from the state.

jochem-brouwer:
I think in order to fix:

- Remember at which height the account was touched first
- If a frame `revert`s, only delete accounts from `_touched` if the current height matches the height when it was first touched.

jochem-brouwer:
I tried integrating it yesterday, but this still fails tests (introduced some new ones and fixed some old ones). I will definitely have time to work on this this week!

jochem-brouwer:
Ok, this logic is starting to get very confusing and very complex. Let's see if CI passes now though.

In this state, I do not want to merge this (in case this passes) because the logic is all over the place. Maybe we should create a more general journal class which does this diff-based journaling, with the commit/revert/checkpoint logic?

jochem-brouwer:
Ok, Istanbul state tests passed locally so I expect CI to pass.

acolytec3 (approved):
This looks great.  Have worked through the code and I think this approach makes sense!

jochem-brouwer:
Ok great! Let's discuss in the call this afternoon if we want to merge as-is, because I think this current approach (1) works but is (2) rather hackish :)

jochem-brouwer:
Added general journaling. I have not yet looked if we can also add this to the access lists, so the journaling might change in another PR. Ready for review.

holgerd77:
> BTW, side note, since this targets the Shanghai DDoS attacks - we still have this problem: #2406
> 
> (For this specific attack we will run out-of-memory on mainnet also)

Ah, didn't know that this had an issue already. Yeah, that's what I also mentioned in the call and what is solved by the StateManager cache rewrite now along #2590.

packages/vm/src/eei/journaling.ts
```diff
@@ -0,0 +1,82 @@
...
+      this.journalStack[height] = new Set()
+    }
+    this.journalStack[height].add(input)
+
+    this.journal.add(input)
+    if (this.journalHeight.get(input) === undefined) {
+      this.journalHeight.set(input, height)
+    }
+  }
+}
```
holgerd77, line 82: Oh, this looks super-clean. Great stuff! 🙂 👍

holgerd77:
Ah, client tests are failing here as well (as in my totally different-set PR #2586, so we have something which got broken along the road.

holgerd77:
Rebased this via UI.

acolytec3:
I've started looking at the logs.  Looks like the jorunaling isn't properly catching accounts that need to be cleaned up on revert (at least in some instances).  I've compared the VM output between master and this branch and see where a touched account is being cleaned up in master and not in this branch so will need to research further.

jochem-brouwer:
Hmm, weird, the tests failing do not seem super related to the touched accounts. I also ran the Istanbul blockchain test locally, and it passed, so I assumed all was OK (but I rechecked, also fails here now, so not sure what went on there)

acolytec3 (approved):
This looks great.  Pending passing CI with the fix on the journal height, lets merge this.

jochem-brouwer (approved):

PR #2585, opened 2023-03-13
client: Add safe and finalized blockoptions to the chain

g11tech:
Add safe and finalized blockoptions to the chain

fix: `engine-api/Safe Re-Org to Side Chain`

Test is passing: 
image

### DIFF
--- a/packages/client/lib/blockchain/chain.ts
+++ b/packages/client/lib/blockchain/chain.ts
@@ -267,7 +267,8 @@ export class Chain {
     }
 
     headers.latest = await this.getCanonicalHeadHeader()
-    // finalized and safe have to block only since execution has to be valid
+    // finalized and safe are always blocks since they have to have valid execution
+    // before they can be saved in chain
     headers.finalized = (await this.getCanonicalFinalizedBlock()).header
     headers.safe = (await this.getCanonicalSafeBlock()).header
 
--- a/packages/client/lib/blockchain/chain.ts
+++ b/packages/client/lib/blockchain/chain.ts
@@ -37,6 +37,18 @@ export interface ChainBlocks {
   latest: Block | null
 
   /**
+   * The block as signalled `finalized` in the fcU
+   * This corresponds to the last finalized beacon block
+   */
+  finalized: Block | null
+
+  /**
+   * The block as signalled `safe` in the fcU
+   * This corresponds to the last justified beacon block
+   */
+  safe: Block | null
+
+  /**
    * The total difficulty of the blockchain
    */
   td: bigint
@@ -55,6 +67,18 @@ export interface ChainHeaders {
    * The latest header in the chain
    */
   latest: BlockHeader | null
+
+  /**
+   * The block as signalled `finalized` in the fcU
+   * This corresponds to the last finalized beacon block
+   */
+  finalized: BlockHeader | null
+
+  /**
+   * The block as signalled `safe` in the fcU
+   * This corresponds to the last justified beacon block
+   */
+  safe: BlockHeader | null
 
   /**
    * The total difficulty of the headerchain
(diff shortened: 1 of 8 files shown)


packages/client/lib/rpc/modules/engine.ts
```diff
@@ -825,45 +884,6 @@ export class Engine {
...
-     */
-    const zeroBlockHash = zeros(32)
-    const safe = toBuffer(safeBlockHash)
-    if (!safe.equals(headBlock.hash()) && !safe.equals(zeroBlockHash)) {
-      const msg = 'Safe block not in canonical chain'
-      try {
-        const safeBlock = await this.chain.getBlock(safe)
-        const canonical = await this.chain.getBlock(safeBlock.header.number)
-        if (!canonical.hash().equals(safe)) {
-          throw new Error(msg)
```
jochem-brouwer, line 840: This logic is now gone?

g11tech: no its moved to setHead

packages/client/lib/execution/vmexecution.ts
```diff
@@ -161,27 +161,72 @@ export class VMExecution extends Execution {
...
         }
       }
+
+      // check if the head, safe and finalized are now canonical
+      for (const [blockName, block] of chainPointers) {
+        if (block === null) {
+          continue
+        }
+        const blockByNumber = await this.chain.getBlock(block.header.number)
+        if (!blockByNumber.hash().equals(block.hash())) {
```
g11tech, line 218: canonicalness of head, safe and finalized is not checked here

packages/client/lib/blockchain/chain.ts
```diff
@@ -221,17 +253,27 @@ export class Chain {
...
     const blocks: ChainBlocks = {
       latest: null,
+      finalized: null,
+      safe: null,
       td: BigInt(0),
       height: BigInt(0),
     }
 
     headers.latest = await this.getCanonicalHeadHeader()
+    // finalized and safe have to block only since execution has to be valid
```
acolytec3, line 270: This comment is a little bit unclear.  Should it say something like `finalized and safe blocks are equivalent to canonical header since block has been executed`?

packages/client/lib/blockchain/chain.ts
```diff
@@ -414,6 +471,22 @@ export class Chain {
     return this.blockchain.getCanonicalHeadBlock()
   }
 
+  /**
+   * Gets the latest block in the canonical chain
+   */
+  async getCanonicalSafeBlock(): Promise<Block> {
```
acolytec3, line 477: Why do we have separate functions if they do the same thing?  Are we going to get to a place in our code where we might have different safe or finalized blocks?

packages/client/lib/rpc/modules/engine.ts
```diff
@@ -826,7 +878,15 @@ export class Engine {
       }
 
       const blocks = [...parentBlocks, headBlock]
-      await this.execution.setHead(blocks)
+      try {
+        await this.execution.setHead(blocks, { safeBlock, finalizedBlock })
+      } catch (error) {
+        console.log(error)
```
acolytec3, line 884: Let's remove this

packages/client/lib/blockchain/chain.ts
g11tech: safe and finalized are not same... safe can become finalized (or in rare re-org scenarios not)

acolytec3 (changes requested):
Looks fine though left a couple of questions to resolve before we merge.

packages/client/lib/blockchain/chain.ts
acolytec3: Ah, sorry, misread the code.

acolytec3 (approved):

PR #2590, opened 2023-03-15
Statemanager: Cache Rewrite with Access and Diff Cache Separation

holgerd77:
WIP

PR #2592, opened 2023-03-17
Statemanager: Cache Rewrite (Account/Storage) & Refactoring (Interface/Tests) (old)

holgerd77:
This PR replaces the previous two cache rewrite PRs #2528 - allowing for cache deactivation - and #2590 - introducing a broad-scale refactor with a separation of access and diff (account) cache and preparing for a subsequent deeper cache integration and persistance.

PR is now independent from #2581 and targeting `master`, so that these two improvements can be handled and reviewed independently. Changes from here still fix the recursive DELEGATECALL out-of-memory issue from block 2283516.

Currently there is still a bug in the code since some blocks are quitting with a receipt root error. I have added a more extensive dedicated test suite for checkpointing tests to StateManager, this helped but something is still going wrong. Eventually the CI run here will point to something, since interdependencies with the other PR are now removed.

Additionally PR contains two small fixes:

- Bug in `StateManager.copy()` not taking in the options
- ~Bug in `Util` `Account.isEmtpy()` ignoring empty storage root~


packages/util/src/account.ts
```diff
@@ -122,7 +122,12 @@ export class Account {
    * "An account is considered empty when it has no code and zero nonce and zero balance."
    */
   isEmpty(): boolean {
-    return this.balance === _0n && this.nonce === _0n && this.codeHash.equals(KECCAK256_NULL)
+    return (
+      this.balance === _0n &&
+      this.nonce === _0n &&
+      this.codeHash.equals(KECCAK256_NULL) &&
+      this.storageRoot.equals(KECCAK256_RLP)
```
jochem-brouwer, line 129: I remember changing this in the past as well too. This breaks tests. I think it is related to gas costs, if all fields are zero except `storageRoot` then the account, for EVM purposes, is still empty. 

If you remove this line, `npm run test:state -- --fork=London` changes from `fail  1778` to `fail  186`

packages/statemanager/src/cache.ts
```diff
@@ -14,74 +15,176 @@ export interface CacheOpts {
...
-    const it = this._cache.find(keyStr)
-    if (!it.equals(this._cacheEnd)) {
-      return it.pointer[1].deleted
+  keyIsDeleted(address: Address): boolean {
+    const account = this.get(address)
+
+    if (account.isEmpty()) {
+      return false
+    } else {
+      return true
```
jochem-brouwer, line 186: I am confused here, if the account is not empty, the key is deleted?

packages/vm/src/eei/vmState.ts
```diff
@@ -266,6 +266,7 @@ export class VmState implements EVMStateAccess {
       }
     }
     await this._stateManager.flush()
+    this.touchedJournal.clear()
```
jochem-brouwer, line 269: I think we need to investigate this a bit further. The problem with the client test failing (`getTransactionReceipt.spec.ts`) was that:

- The blockchain is initialized with a Geth genesis state
- This creates accounts for a big precompile range (0x00 - 0x00..00ff)`
- However, when creating genesis (in `vmexecution.open()`), these accounts are all created and dumped into `touchedJournal`. This is /not/ cleared
- However, on current master, the touched accounts are empty

This sounds a bit error prone to me and I want to investigate why this change has happened.

holgerd77:
Rebased this and removed the `mainnetOptimizations` option (didn't work on follow-up blocks due to SELFDESTRUCT being executed repeatedly with a beneficiary in the targeted zero-based address range).

packages/util/src/account.ts
holgerd77: Have reverted.

packages/client/lib/execution/vmexecution.ts
```diff
@@ -48,6 +54,10 @@ export class VMExecution extends Execution {
 
       const stateManager = new DefaultStateManager({
         trie,
+        cacheOptions: {
+          deactivate: false,
+          size: 2500000,
```
jochem-brouwer, line 59: This immediately allocates these items, do we want this?

holgerd77: Hmm, yeah, maybe this can get optimized, I would also say that it doesn't hurt too much right now (and I am also not so motivated to improve here atm).

Are you generally sure about this behavior? Where did you "see" that?

jochem-brouwer: image

It's in the docs :) https://www.npmjs.com/package/lru-cache

jochem-brouwer: Quick note: if you set the cache size to `1` then the tests all fail (this implies that somewhere we use the cache wrong). If you use the current setting, the tests will go out-of-memory.

jochem-brouwer: (at least on the version I tested yesterday)

holgerd77: Ah, that's really interesting, even 10, 100, and 1000 fail for `npm run tape -- test/execution/vmexecution.spec.ts`, 10000 is the first setting where this goes through.

Will have a look.

holgerd77: Ah, one thing I am side-realizing while debugging: when we are running the client tests on mainnet, there is always (on every test run) the need to re-fill the state with all these hundres of  genesis/presale accounts, this will very likely slow down test runs significantly. So we might want to switch over to sepolia or so where ever we can.

holgerd77: Update: ah, so these numbers from above, succeed or not, is likely the number of presale wallets.

holgerd77: Ok, yeah, it is that the current design is just not working properly if the cache size is lower than number of accounts modified during a block execution.

This is since these modifications are only put into the trie on `flush()` execution, and if there are more puts than the cache can hold in between the old (not yet saved) values are getting out of the cache again before this is persisted.

This would need a substantial cache re-write (again) if this should be tackled, I wouldn't want to do that. But we should definitely point this out in the docs.

holgerd77: [49df9605133642d40b456505188ea2edad7002ab](https://github.com/ethereumjs/ethereumjs-monorepo/pull/2592/commits/b6eafac9cdd62637a3f7274c924d63d4c75d11d6)

holgerd77:
Ready for review (together with #2618). 🙂 

Do not merge, PRs will be rebased on top of `develop-v7` after review.

acolytec3 (commented):
Partial review done, will continue on the state manager itself next.

packages/statemanager/src/stateManager.ts
```diff
@@ -76,8 +121,10 @@ export interface DefaultStateManagerOpts {
 export class DefaultStateManager extends BaseStateManager implements StateManager {
   _trie: Trie
   _storageTries: { [key: string]: Trie }
+  _codeCache: { [key: string]: Buffer }
```
acolytec3, line 124: This is remarkable to me that we didn't have a codecache before.  Great add!

packages/statemanager/src/stateManager.ts
```diff
@@ -87,39 +134,102 @@ export class DefaultStateManager extends BaseStateManager implements StateManage
...
+          account.balance
+        } contract=${account.isContract() ? 'yes' : 'no'} empty=${account.isEmpty() ? 'yes' : 'no'}`
+      )
     }
-    const deleteCb = async (keyBuf: Buffer) => {
+    if (this._cacheSettings.deactivate) {
       const trie = this._trie
-      await trie.del(keyBuf)
+      // This is fixing a bug in the VM GeneralStateTestsRunner passing undefined here for selected accounts
+      // and which breaks when account cache is being deactivated
```
acolytec3, line 210:
```suggestion
      // and which breaks when account cache is deactivated
```

packages/statemanager/src/stateManager.ts
```diff
@@ -139,9 +249,15 @@ export class DefaultStateManager extends BaseStateManager implements StateManage
     // @ts-expect-error
     await this._trie._db.put(key, value)
 
+    const keyHex = key.toString('hex')
+    this._codeCache[keyHex] = value
+
     if (this.DEBUG) {
       this._debug(`Update codeHash (-> ${short(codeHash)}) for account ${address}`)
     }
+    if (!(await this.getAccount(address))) {
```
acolytec3, line 258:
```suggestion
    if ((await this.getAccount(address) === undefined)) {
```

I'm surprised the linter isn't complaining here about an explicit boolean check.

packages/statemanager/src/stateManager.ts
```diff
@@ -153,15 +269,25 @@ export class DefaultStateManager extends BaseStateManager implements StateManage
...
-    // @ts-expect-error
-    const code = await this._trie._db.get(key)
-    return code ?? Buffer.alloc(0)
+
+    const keyHex = key.toString('hex')
+    if (keyHex in this._codeCache) {
+      return this._codeCache[keyHex]
+    } else {
+      // @ts-expect-error
+      const code = (await this._trie._db.get(key)) ?? Buffer.alloc(0)
```
acolytec3, line 287: Just a question about best practice on Typescript errors like this, should we use the above `@ts-expect-error` which makes it not obvious to me what the actual error is (since VS Code respects this and doesn't give you the red squiggles), or should we do `(await (this._trie as any)._db...`?  I'm fine with this approach but maybe you could add a comment explaining what the expected error is (i.e. `db` is a protected member of the `trie` class)

packages/statemanager/src/stateManager.ts
```diff
@@ -260,6 +392,9 @@ export class DefaultStateManager extends BaseStateManager implements StateManage
     }
 
     value = unpadBuffer(value)
+    if (!(await this.getAccount(address))) {
```
acolytec3, line 395:
```suggestion
    if ((await this.getAccount(address)) === undefined) {
```

packages/statemanager/src/stateManager.ts
```diff
@@ -329,6 +465,9 @@ export class DefaultStateManager extends BaseStateManager implements StateManage
    */
   async getProof(address: Address, storageSlots: Buffer[] = []): Promise<Proof> {
     const account = await this.getAccount(address)
+    if (!account) {
```
acolytec3, line 468: Here again, very surprised the linter isn't complaining.  I thought we had a rule for explicit boolean checks where a variable contains either a value or `undefined`?
```suggestion
    if (account === undefined) {
```

packages/statemanager/src/stateManager.ts
```diff
@@ -465,8 +608,11 @@ export class DefaultStateManager extends BaseStateManager implements StateManage
     }
 
     this._trie.root(stateRoot)
-    this._cache.clear()
+    if (this._cache && clearCache) {
+      this._cache.clear()
```
acolytec3, line 612:
```suggestion
    if (clearCache) {
      this._cache?.clear()
```

packages/statemanager/src/stateManager.ts
```diff
@@ -509,17 +655,29 @@ export class DefaultStateManager extends BaseStateManager implements StateManage
...
-    if (
-      account &&
-      ((account as any).virtual === undefined || (account as any).virtual === false) &&
-      !this._cache.keyIsDeleted(address)
-    ) {
-      return true
-    }
-    if (await this._trie.get(address.buf)) {
-      return true
+    if (this._cache) {
```
acolytec3, line 658:
```suggestion
    if (this._cache !== undefined) {
```

packages/statemanager/src/cache.ts
```diff
@@ -1,192 +1,366 @@
...
+        } for account ${addressHex}`
+      )
+      this._diffCache[this._checkpoints].setElement(addressHex, oldElem)
+    }
   }
 
   /**
    * Puts account to cache under its address.
-   * @param key - Address of account
+   * @param key - Address of account or undefined if account doesn't exist in the trie
```
acolytec3, line 118: Can we rewrite the docs to match the existing function spec?  You now have `address` and `account` where `account` can be either an account or `undefined` so not sure if the above docs still match.

packages/statemanager/src/cache.ts
```diff
@@ -1,192 +1,366 @@
...
+    } else {
+      this._orderedMapCache!.setElement(addressHex, elem)
+    }
+    this._stats.cache.writes += 1
   }
 
   /**
-   * Returns the queried account or an empty account.
+   * Returns the queried account or undefined if account doesn't exist
    * @param key - Address of account
```
acolytec3, line 142:
```suggestion
   * @param address - Address of account
```

packages/statemanager/src/cache.ts
```diff
@@ -1,192 +1,366 @@
...
+      this._stats.cache.hits += 1
+    }
+    return elem
   }
 
   /**
-   * Returns the queried account or undefined.
-   * @param key - Address of account
+   * Marks address as deleted in cache.
+   * @param key - Address
```
acolytec3, line 163:
```suggestion
   * @param address - Address
```

packages/statemanager/src/cache.ts
```diff
@@ -1,192 +1,366 @@
...
+        accountRLP: undefined,
+      })
     }
+
+    this._stats.cache.dels += 1
   }
 
   /**
    * Returns true if the key was deleted and thus existed in the cache earlier
    * @param key - trie key to lookup
```
acolytec3, line 184:
```suggestion
   * @param address - trie key to lookup
```

packages/statemanager/src/cache.ts
```diff
@@ -1,192 +1,366 @@
...
+    } else {
+      return false
     }
-    return false
   }
 
   /**
    * Looks up address in cache, if not found, looks it up
    * in the underlying trie.
    * @param key - Address of account
```
acolytec3, line 199:
```suggestion
   * @param address - Address of account
```

acolytec3 (commented):
Finished an initial review.  Need to go back over the cache again and really make sure i understand it.

packages/client/lib/rpc/modules/eth.ts
```diff
@@ -514,6 +513,9 @@ export class Eth {
     const vm = await this._vm.copy()
     await vm.stateManager.setStateRoot(block.header.stateRoot)
     const account = await vm.stateManager.getAccount(address)
+    if (account === undefined) {
+      throw new Error(`could not read account`)
```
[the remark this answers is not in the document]
holgerd77: TBH, I have no idea why I added this here, that was very much a mistake, yes, will update.

packages/client/test/sync/beaconsync.spec.ts
```diff
@@ -195,7 +197,7 @@ tape('[BeaconSynchronizer]', async (t) => {
   })
 
   t.test('should extend and set with a valid head', async (t) => {
-    const config = new Config({ transports: [] })
+    const config = new Config({ transports: [], cacheSize: 10000 })
```
holgerd77, line 200: Not sure, an eventual more elegant way to solve this need of always passing in the - lowered - cache sizes in the tests might be to default the client to an ORDERED_MAP cache (and so get rid of this size problem) and only from the `bin/cli.ts` instantiation instantiate with LRU for cache_type.

Somewhat of a downside might be that tests are bit less "realistic" when we take another cache type for the tests.

No strong opinion here though, let me know if you have any thoughts or tendencies.

packages/evm/src/interpreter.ts
```diff
@@ -918,6 +928,9 @@ export class Interpreter {
       Object.assign(this._result.selfdestruct, selfdestruct)
       // update stateRoot on current contract
       const account = await this._eei.getAccount(this._env.address)
+      if (!account) {
+        throw new Error('could not read contract account')
```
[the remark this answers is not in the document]
holgerd77: Generally - yes - we can use to detect systemic issues early and not have this disguided by account creation (which then goes on with a wrong nonce or so e.g.). That said, I removed most of these patterns (throw) again in the storage cache PR, since these places tend to need some dedicated attention and the initial test failures were too overwhelming.

That often doesn't mean that it would be plain wrong to change, in fact at the end of my storage cache work I re-introduced some of these (also for finding my own bugs in the code).

packages/statemanager/test/cache.spec.ts
```diff
@@ -2,7 +2,7 @@ import { Trie } from '@ethereumjs/trie'
 import { Account, Address } from '@ethereumjs/util'
 import * as tape from 'tape'
 
-import { Cache } from '../src/cache'
+import { Cache, CacheType } from '../src/cache'
```
acolytec3, line 5: General question on this, should we be testing with both cache types here (so just run these tests in a loop over eac cache type)?

packages/client/test/sync/beaconsync.spec.ts
acolytec3: Not sure if I have an opinion either way.  Doesn't the LRU cache type give significant performance improvements we would when running the client?

holgerd77:
Also some procedural note: will keep this "un-rebased" during the review period and then squash the commits from here as Andrew suggested and do a first `master` rebase (primarily to get the snap sync stuff integrated, @scorbajio might need some help there) and then do the `develop-v7` rebase in the next step.

holgerd77:
(also eventually possible: some mixed procedure, so might already start with the rebases during review period and then do the review fixes on top of the rebased and squashed PR. This unsquashed PR should nevertheless remain the main review location)

packages/client/lib/rpc/modules/eth.ts
```diff
@@ -684,7 +686,10 @@ export class Eth {
     await vm.stateManager.setStateRoot(block.header.stateRoot)
 
     const address = Address.fromString(addressHex)
-    const account: Account = await vm.stateManager.getAccount(address)
+    const account = await vm.stateManager.getAccount(address)
+    if (account === undefined) {
+      throw new Error(`could not read account`)
```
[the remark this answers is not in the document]
holgerd77: Same, have updated.

packages/evm/src/evm.ts
```diff
@@ -615,8 +630,9 @@ export class EVM implements EVMInterface {
         // It is thus an unnecessary default item, which we have to save to dik
         // It does change the state root, but it only wastes storage.
         //await this._state.putContractCode(message.to, result.returnValue)
+        //const account = await this.eei.getAccount(message.to)
```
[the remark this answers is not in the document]
holgerd77: I can do that but for the other commented out line with `putContractCode()` I am not sure why this is there and if this is maybe a reminder that something is still off here or if this is safe to remove? 🤔

Maybe @jochem-brouwer can shed some light here? Or do you have any context as well?

packages/statemanager/src/stateManager.ts
holgerd77: Wrong parantheses with await, online code suggestions are always a bit dangerous 😋.

But will generally update, thanks, still not sure about this linter thing.

packages/statemanager/src/stateManager.ts
holgerd77: Just seeing, thanks for updating. 

Update: ah, and just also seeing that we have a public access method for the database in the Trie library (`database()`).

Will just update with this, so can prevent the whole thing.

packages/statemanager/src/cache.ts
holgerd77: Generally did some additional round through the code docs.

packages/client/test/sync/beaconsync.spec.ts
holgerd77: Client should always be LRU - yes - this is just a question if the test setup would be a bit easier (and also somewhat less error prone, if one e.g. forgets to add the lower cache limits for config) if realized in the above-described way.

Anyhow, will leave for now, but still somewhat inclined to do, no super strong opinion though.

packages/statemanager/test/cache.spec.ts
holgerd77: Yes, makes sense, have updated.

PR #2594, opened 2023-03-20
EVM+Tx: Add `allowUnlimitedInitcodeSize` option to partially disable EIP-3860

### LINKED ISSUE #2588: Add `allowUnlimitedContractSize` option
I think it's plausible that users that are already making use of `allowUnlimitedContractSize` to deploy "initcode" (tx.data) larger than this new `MAX_INITCODE_SIZE` limitation, so they'd be stuck without a way of also bailing out of this init code size check (Ganache's own tests actually fall victim to this limitation). I like the idea of using `allowUnlimitedInitCodeSize` as the option name and would update Ganache options to match.

I started on the work to do this myself, but I ran into a complication with the transactions, as it looks like the options for either `Common` or the transaction classes will need to be extended. This sort of interface change seems to better suited to a maintainer of the repo, so here I am, creating an issue instead.

We can work around this in Ganache by disabling EIP-3860 in our `Common` if the user sets our `allowUnlimitedInitCodeSize` flag, so it's not super urgent, though passing in an option to the evm, like we do for `allowUnlimitedContractSize` would be ideal.

related: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2575

jochem-brouwer:
Closes #2588 

This PR adds the `allowUnlimitedInitcodeSize` option. If EIP 3860 is disabled, this does nothing. If it is active, then this only disables the initcode length check, but it will still charge the initcode data gas cost.

This PR also somewhat refactors tx: in all of the tx types the same initcode size check is added, so it makes sense to move this into `baseTransaction`. (However, due to legacy transaction we do not have access to a common there, so either take this from the options or create the default one)

### DIFF
--- a/packages/evm/src/evm.ts
+++ b/packages/evm/src/evm.ts
@@ -99,6 +99,12 @@ export interface EVMOpts {
   allowUnlimitedContractSize?: boolean
 
   /**
+   * Allows unlimited contract code-size init while debugging. This (partially) disables EIP-3860.
+   * Gas cost for initcode size analysis will still be charged. Use with caution.
+   */
+  allowUnlimitedInitcodeSize?: boolean
+
+  /**
    * Override or add custom opcodes to the EVM instruction set
    * These custom opcodes are EIP-agnostic and are always statically added
    * To delete an opcode, add an entry of format `{opcode: number}`. This will delete that opcode from the EVM.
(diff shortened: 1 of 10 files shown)


holgerd77 (approved):
LGTM

PR #2598, opened 2023-03-22
evm: clamp step event memory to actual size

davidmurdoch:
This PR fixes the `memory` field in `step` events: it previously reported the internal memory state (which is expanded by 8kb each time it gets into a new 8kb section), but now it reports the actual memory which EVM sees.


acolytec3 (approved):
LGTM!  Good test too!

holgerd77:
Thanks 🙏, yeah, great fix!

PR #2600, opened 2023-03-23
Fix Block Hash Calculation When Creating a New Block Object From JSON RPC (Shanghai)

solimander:
This PR fixes an issue in which the withdrawals root would not be passed to block header creation when creating a block from JSON RPC, which caused the block hash to be computed incorrectly.

The withdrawals are also passed to the `Block` class via `fromBlockData` when creating a block from RPC, which enables `validateWithdrawalsTrie` to function as expected.


jochem-brouwer:
Great catch! Could you also add a test? :smile:

solimander:
> Great catch! Could you also add a test? 😄

Done 🙂

PR #2601, opened 2023-03-24
Update ethereum/tests to v12

jochem-brouwer:
This PR:

- Updates the test suite to v12 (this includes Shanghai tests)
- Fixes these tests
- Also add support for Merge on the CI in case we test all forks
- Temporarily disables ShanghaiLove test on Homestead, such that now we can run the `test all hardforks` label on Homestead as well (without crashing node). See #2406


jochem-brouwer:
Retesteth:

```
--------
*** TOTAL ERRORS DETECTED: 11 errors during all test execution!
--------
info: (GeneralStateTests/stQuadraticComplexityTest/Create1000, fork: Shanghai, TrInfo: d: 0, g: 1, v: 0, TrData: ` 0x..`)
info: (GeneralStateTests/stTransactionTest/HighGasPrice, fork: EIP150, TrInfo: d: 0, g: 0, v: 0, TrData: ` 0x..`)
info: (GeneralStateTests/stTransactionTest/ValueOverflow, fork: EIP150, TrInfo: d: 0, g: 0, v: 0, TrData: ` 0x..`)
info: (GeneralStateTests/stExample/accessListExample, fork: EIP150, TrInfo: d: 0, g: 0, v: 0, TrData: `:label declaredKeyWrite 0x00..`)
info: (GeneralStateTests/stExample/basefeeExample, fork: EIP150, TrInfo: d: 0, g: 0, v: 0, TrData: `:label declaredKeyWrite 0x00..`)
info: (Shanghai/stEIP3860-limitmeterinitcode/create2InitCodeSizeLimit, fork: Shanghai, TrInfo: d: 1, g: 0, v: 0, TrData: `:label invalid 0x000000..`)
info: (Shanghai/stEIP3860-limitmeterinitcode/createInitCodeSizeLimit, fork: Shanghai, TrInfo: d: 1, g: 0, v: 0, TrData: `:label invalid 0x000000..`)
info: (Shanghai/stEIP3860-limitmeterinitcode/creationTxInitCodeSizeLimit, fork: Shanghai, TrInfo: d: 1, g: 0, v: 0, TrData: `:label invalid 0x601080..`)
info: (GeneralStateTests/stExample/eip1559, fork: EIP150, TrInfo: d: 0, g: 0, v: 0, TrData: ` 0x00..`)
info: (GeneralStateTests/stRandom/randomStatetest307, fork: Shanghai, TrInfo: d: 0, g: 0, v: 0, TrData: ` 0x7f0000..`)
info: (GeneralStateTests/stRandom2/randomStatetest646, fork: Shanghai, TrInfo: d: 0, g: 0, v: 0, TrData: ` 0x785196..`)
Terminated
```

jochem-brouwer:
Shanghai blockchain tests failing:

```
Errors thrown in file: GeneralStateTests/Shanghai/stEIP3860-limitmeterinitcode/creationTxInitCodeSizeLimit.json test: creationTxInitCodeSizeLimit_d1g0v0_Shanghai:
	tx should fail, but did not fail
Errors thrown in file: InvalidBlocks/bc4895-withdrawals/amountIs0TouchAccount.json test: amountIs0TouchAccount_Shanghai:
	Error: invalid block stateRoot (vm hf=shanghai -> block number=1 hash=0x1d03b6adf02f3c78dcd1f8a164eab82d39af920791bb8d23d367a7f5a2a2d1c8 hf=shanghai baseFeePerGas=9 txs=0 uncles=0)
	correct last header block
Errors thrown in file: InvalidBlocks/bc4895-withdrawals/shanghaiWithoutWithdrawalsRLP.json test: shanghaiWithoutWithdrawalsRLP_Shanghai:
	expected exception but test did not throw an exception: RLP_ExpectedAsList
```

There are also EVM tests failing regarding EIP3860. The tests which are failing on Chainstart/Homestead/TangerineWhistle are most likely wrong tests.

jochem-brouwer:
Shanghai tests should be fixed now, awaiting confirm that the old tests (chainstart/homestead/tangerine whistle) are wrong.

packages/block/src/block.ts
```diff
@@ -140,6 +140,16 @@ export class Block {
       throw new Error('Invalid serialized block input. Must be array')
     }
 
+    if (
+      opts?.common !== undefined &&
+      opts?.common?.isActivatedEIP(4895) &&
+      (values[3] === undefined || !Array.isArray(values[3]))
+    ) {
+      throw new Error(
```
jochem-brouwer, line 148: TODO also check if we have this problem in header (and maybe also for `excessDataGas` in header..?)

jochem-brouwer: Also check if we should move this into `fromValuesArray` (in header this is the case..?)

jochem-brouwer: I do not feel super good that `fromHeaderData` / `fromBlockData` is more "permissive" than the other methods. For instance, if you `fromBlockData` with 4895 (withdrawals) active, then it will not complain and just instantiate the block with an empty withdrawals list. However, currently `fromRLPSerializedBlock` will throw (and probably `fromValuesArray` should also throw). Not super sure what I think about these "two levels" of "security" of these methods.

holgerd77: I know what you mean but the thing is that we actually use this *a lot* - in tx there is a similar pattern and we use this for all sorts of convenience instantiations spread over hundreds of places - things like:

```typescript
const tx = FeeMarketEIP1559Transaction.fromTxData(
  { to: B.address, maxFeePerGas: 6 },
  { common }
)
```

This differentiation also *has* some logic, since the other two constructors (RLP, values array) *are* just used for the very most part in contexts (client, devp2p, other) where a "full" tx/block is expected, while - I would assume - the `from*Data` constructor prevails for testing or non-complete setups.

I think we do at least want to keep this more permissive behavior, at least as one option. One option to improve here might be an additional flag, just for this one constructor (or for all constructors but with a different default setting?) where one can set the checks to be more strict or something.

jochem-brouwer: I agree, but I think we should indeed add this strict check flag on the `from*Data`, such that users /must/ provide all fields.

jochem-brouwer:
Ok this will fix almost all, except the <= TangerineWhistle test suite, since it appears there are two situations:

- tx is invalid, but create coinbase account anyways, e.g. `npm run test:state -- --fork=TangerineWhistle --test=HighGasPrice`
- tx is invalid, but do not create coinbase (???) `npm run test:state -- --fork=TangerineWhistle --test=transactionCollidingWithNonEmptyAccount_calls`

jochem-brouwer:
I think CI will pass, (except homestead, but we know this), ready for review.

jochem-brouwer:
Ok, also temporarily removed the ShanghaiLove test from Homestead such that we can run the blockchain suite on Homestead as well. See #2406, we should address.

CI should pass now. Ready for review.

packages/block/src/block.ts
g11tech: shouldn't you just call `validateData` here? no reason to duplicate checks

acolytec3 (approved):
LGTM

PR #2604, opened 2023-03-27
Move EEI into EVM

jochem-brouwer:
The purpose of this PR is to move the EEI into EVM and perform cleanups.

Note: this targets the `statemanager-cache-rewrite-new`, which currently points to `master`, but this PR is super breaking and it should eventually land in `dev-v7`.


packages/common/src/types.ts
```diff
@@ -120,3 +121,109 @@ export interface GethConfigOpts extends BaseOpts {
...
+  verifyProof?(proof: Proof): Promise<boolean>
+  hasStateRoot(root: Buffer): Promise<boolean>
+}
+
+export interface StateManagerInterface extends StateAccess {
+  cache?: CacheInterface
+  copy(): StateManagerInterface
+  flush(): Promise<void>
+  dumpStorage(address: Address): Promise<StorageDump>
+}
```
holgerd77, line 229: I think we should be really sparse with taking over interfaces to Common and not "throwing things all over here".

So these should really only be the main interfaces, like `StateManagerInterface`, eventually `BlockchainInterface`, and not much else.

So these should be very clear, very well-managed interfaces, where there is a lot of attention onto. Otherwise I fear that we will likely overload and produce too much complexity which won't make things easier to oversee respectively more stable.

I would particularly suggest that we do not expose the cache so prominently, I guess I added this (right?), but anyhow, it would likely be better if we just hack this a bit on the very few direct cache accesses we have internally than build up such heavy structures like this.

I would suggest the following:

- Remove cache from SM Interface, delete all related interfaces
- (CacheClearingOpts is outdated anyhow)
- Remove the `dumpStorage` and `StorageDump` from the interface in general, this is only used in one very obscure client location, we can hack this in there
- Throw the `StateAccess` and `StateManagerInterface` interfaces together (never gotten a big fan of this)
- Definitely remove this `AccountFields` partial
- Eventually keep the Proof types, but really only if stable enough

Two other things:

1. This could also very well be an own PR, totally independent from other stuff
2. I think we should rather do a dedicated file `interfaces.ts` for this

packages/evm/src/evm.ts
```diff
@@ -135,9 +138,21 @@ export interface EVMOpts {
...
+   *
+   */
+  blockchain?: Blockchain
+
+  /**
+   * This optional flag should be set to `true` if no blockchain is provided
+   * This is used the warn users that if they do not provide a blockchain,
+   * the default blockchain will be used, which always returns the 0 hash if a block is
+   */
+  enableDefaultBlockchain?: boolean
```
holgerd77, line 155: I don't think we need this option.

I somewhat understand the intention to "warn" users. I would argue that this is somewhat intuitive though (if an EVM has no connection to whatever blockchain, no information from a/the blockchain can be read). So I think if we have this prominently in the docs that really should be enough.

At the same time the option causes a lot of bloat - like in the tests.

holgerd77:
Hi Jochem,
could you please reason a bit what you want to achieve with the PR respectively what the design goals are? We actually did a lot of work to *extract* the EEI from the EVM and move this over to VM, now we are moving back here, at least to some extend? This might be justified if we have a clearer view here now and this will come with an adopted structure, I would like to have some more clear reasoning though so that we can get a bit more assured that we are on a good and sustainable path with this.

I also have the impression that there are several subtasks involved here which might be able to be separated, like:

1. Extracting the interfaces (and discuss them)
2. Go directly "through" to StateManager and not use the EEI
3. Not sure if there are others, e.g. related to this "Fake/Mock/Default Blockchain" idea

Not sure if it might be worth to think a bit again about the order of tasks and eventually submit separately? Whatever is best and easier to realize. Just feels a bit several things are mixed here together.

holgerd77:
Maybe we should just do some call on this and discuss this directly 😋, maybe on Thursday morning?

PR #2607, opened 2023-03-29
Buffer to Uint8Array cleanup

acolytec3:
- Cleans up various `buffer` instantiations that were missed in previous `Uint8Array` pr.
- Fixes `devp2p` logging bug
- Addresses multiple places where `Uint8Array` values that are `put` in the database are transformed into JSON objects consisting of key/value pairs where the key is the array index and the value is that uint8 from the original `uint8Array`.  This occurs because we have several places where the `valueEncoding` of our data being put in the DB is `json` which always maps to a JSON object (or nestred object) of utf-8 elements.
- Adds a new PoW miner integration test to the `client` test suite


packages/client/test/miner/ethash.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+      }
+    })
+  })
+}
+
+const restartClient = async (client: EthereumClient, t: tape.Test) => {
+  await client.start()
+  await new Promise((resolve) => {
+    client.config.logger.on('data', (data) => {
+      if (data.message.includes('Miner: Found PoW solution') === true && client.started) {
```
jochem-brouwer: This test looks good, but the only problem I have here is that if this test fails (we broke the miner) this test will run endlessly and thus our CI will not finish. Maybe add a time limit of a minute or so, which rejects the promise if no solution is found?

packages/client/test/miner/ethash.spec.ts
```diff
@@ -110,7 +110,10 @@ tape('start client', async (t) => {
   const client = await setupPowDevnet(minerAddress, true)
   t.ok(client.started, 'client started successfully')
   await stopClient(client, t)
-  const restartedClient = await setupPowDevnet(minerAddress, false)
-  await restartClient(restartedClient, t)
+  await client.open()
+  await ((client.service('eth') as FullEthereumService).execution as any).stateDB!.open()
```
jochem-brouwer: In general this is a great test too, so if we stop the client some internal variables are not cleaned up (so we cannot restart correctly). But this is a different issue, should be addressed though.

packages/client/test/integration/pow.spec.ts
jochem-brouwer, line 1: Shouldn't the test be:

- Create inline client
- Mine a PoW block
- Stop the client
- Restart the client
- Mine a PoW block?

Right now it starts the client, does not mine, then stops and starts again and then mine a PoW block (this would reflect what we tested using the cli)

acolytec3: I think there must be some unclearness in how the function is designed. The `setupPowDevnet` function instantiates and starts the client.  I've renamed the `stopClient` function to `mineBlockAndStopClient` to more clearly identify what happens in that helper.  I set it up as an async function to make it easier to work with the event listener (which uses a callback rather than a promise interface and isn't naturally async).  What actually happens is:
1) `setupPowDevnet` creates and starts the client
2) Test to verify the client is started
3) `mineBlockAndStopClient` sets a listener that waits for the first block to be mined and then passes its test and stops the client once it is mined.

If we try to stop and then restart the client.  A second pow solution is never found so a next block is never mined.

packages/client/test/miner/ethash.spec.ts
acolytec3: Added a 60 second timeout to the test so the whole suite fails if it doesn't complete in 60 seconds (which I think should be sufficient).

packages/client/test/integration/pow.spec.ts
jochem-brouwer: Ok right this ethash problem is beyond the scope of this pr right?

packages/ethash/test/miner.spec.ts
```diff
@@ -6,11 +6,11 @@ import * as tape from 'tape'
 import { Ethash } from '../src'
 
 import type { BlockHeader } from '@ethereumjs/block'
-
+const cacheDb = new MemoryLevel()
 const common = new Common({ chain: Chain.Ropsten, hardfork: Hardfork.Petersburg })
 
 tape('Check if miner works as expected', async function (t) {
-  const e = new Ethash(new MemoryLevel())
+  const e = new Ethash(cacheDb as any)
```
g11tech, line 13: wouldn't it be good to alloc new db each fime? may be in beforeeach?

g11tech (dismissed):
looks good, overall minor queries

packages/ethash/test/miner.spec.ts
acolytec3: That's a matter of opinion I guess.  The above change just has this code match the original test structure (which reused the same DB each time).

packages/evm/test/eips/eip-3860.spec.ts
```diff
@@ -135,7 +135,7 @@ tape('EIP 3860 tests', (t) => {
     const eei = await getEEI()
     const evm = await EVM.create({ common, eei, allowUnlimitedInitCodeSize: true })
 
-    const buffer = Buffer.allocUnsafe(1000000).fill(0x60)
+    const buffer = new Uint8Array(1000000).fill(0x60)
```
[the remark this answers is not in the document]
acolytec3: Aren't we filling with `0x60` or is that equivalent to zero here?

packages/devp2p/src/rlpx/peer.ts
```diff
@@ -602,6 +602,7 @@ export class Peer extends EventEmitter {
       }
       protocolObj.protocol._handleMessage(msgCode, payload)
     } catch (err: any) {
+      console.log(err)
```
g11tech, line 605: console.log?

g11tech (approved):
:+1:

jochem-brouwer (approved):
LGTM

PR #2610, opened 2023-03-30
Clique: get signers by block num

jochem-brouwer:
I added a test, which currently fails, it messes up the `_cliqueLatestSignerStates` somehow (after adding block 2, it adds a new signer to block 2, but also edits the signers of block 0?)

Note: we most likely have this problem with the "latest block signers", and the votes as well.

This PR currently allows to sync Goerli > 5476.

There is another problem with the current implementation: in case of a reorg, the active signers are all still remembered. However, if due to this reorg the signers at block X (where block X is now remembered of the previous change) change, this messes up the signer state internally. It should on reorg actually delete these signers above the block (need to investigate this claim though).

Note to self: local branch `clique-by-blocknum-new`

### DIFF
--- a/packages/blockchain/src/consensus/clique.ts
+++ b/packages/blockchain/src/consensus/clique.ts
@@ -47,6 +47,10 @@ type CliqueLatestBlockSigners = CliqueBlockSigner[]
 
 /**
  * This class encapsulates Clique-related consensus functionality when used with the Blockchain class.
+ * Note: reorgs which happen between epoch transitions, which change the internal voting state over the reorg
+ * will result in failure and is currently not supported.
+ * The hotfix for this could be: re-load the latest epoch block (this has the clique state in the extraData of the header)
+ * Now replay all blocks on top of it. This should validate the chain up to the new/reorged tip which previously threw.
  */
 export class CliqueConsensus implements Consensus {
   blockchain: Blockchain | undefined
@@ -243,8 +247,6 @@ export class CliqueConsensus implements Consensus {
         this._cliqueLatestSignerStates.push(lastItem)
       }
     }
-
-    this._cliqueLatestSignerStates.sort((a, b) => (a[0] > b[0] ? 1 : -1))
 
     // save to db
     const formatted = this._cliqueLatestSignerStates.map((state) => [
@@ -478,7 +480,6 @@ export class CliqueConsensus implements Consensus {
     this._cliqueLatestSignerStates = this._cliqueLatestSignerStates.filter(
       (s) => s[0] <= blockNumber
     )
-    this._cliqueLatestSignerStates.sort((a, b) => (a[0] > b[0] ? 1 : -1))
     await this.cliqueUpdateSignerStates()
 
     this._cliqueLatestVotes = this._cliqueLatestVotes.filter((v) => v[0] <= blockNumber)
--- a/packages/blockchain/src/consensus/clique.ts
+++ b/packages/blockchain/src/consensus/clique.ts
@@ -287,7 +287,7 @@ export class CliqueConsensus implements Consensus {
           (header.number %
             BigInt((this.blockchain!._common.consensusConfig() as CliqueConfig).epoch))
         const limit = this.cliqueSignerLimit(header.number)
-        let activeSigners = this.cliqueActiveSigners(header.number)
+        let activeSigners = [...this.cliqueActiveSigners(header.number)]
         let consensus = false
 
         // AUTH vote analysis
@@ -424,7 +424,7 @@ export class CliqueConsensus implements Consensus {
       return []
     }
     for (let i = signers.length - 1; i >= 0; i--) {
-      if (signers[i][0] <= blockNum) {
+      if (signers[i][0] < blockNum) {
         return signers[i][1]
       }
     }
--- a/packages/blockchain/src/consensus/clique.ts
+++ b/packages/blockchain/src/consensus/clique.ts
@@ -126,7 +126,7 @@ export class CliqueConsensus implements Consensus {
     }
 
     const { header } = block
-    const valid = header.cliqueVerifySignature(this.cliqueActiveSigners())
+    const valid = header.cliqueVerifySignature(this.cliqueActiveSigners(header.number))
     if (!valid) {
       throw new Error('invalid PoA block signature (clique)')
     }
@@ -140,7 +140,7 @@ export class CliqueConsensus implements Consensus {
       // only active (non-stale) votes will counted (if vote.blockNumber >= lastEpochBlockNumber
 
       const checkpointSigners = header.cliqueEpochTransitionSigners()
-      const activeSigners = this.cliqueActiveSigners()
+      const activeSigners = this.cliqueActiveSigners(header.number)
       for (const [i, cSigner] of checkpointSigners.entries()) {
         if (activeSigners[i]?.equals(cSigner) !== true) {
           throw new Error(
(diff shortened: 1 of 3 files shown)


jochem-brouwer:
Ok, I think I fixed the CI for blockchain.

The usage of `cliqueActiveSigners(blockNum: bigint)` are: if you input `blockNum` this will return whatever signers are active on that `blockNum`. In case a new signer gets activated/deactivated, this is **referenced in the next block**. This change of behavior is not a problem if there are no reorgs. However, if there are reorgs, this change should be taken into account.

packages/blockchain/src/consensus/clique.ts
```diff
@@ -233,17 +244,24 @@ export class CliqueConsensus implements Consensus {
       }
     }
 
+    this._cliqueLatestSignerStates.sort((a, b) => (a[0] > b[0] ? 1 : -1))
```
acolytec3, line 247: Is it likely that the clique signer states will get out of order and require us to potentially sort them twice in a single call to this update method?

packages/blockchain/src/consensus/clique.ts
```diff
@@ -233,17 +244,24 @@ export class CliqueConsensus implements Consensus {
...
-      i++
+    if (signerState !== undefined) {
+      let i = 0
+      try {
+        for (const signer of this.cliqueActiveSigners(signerState[0])) {
+          debug(`Clique signer [${i}]: ${signer} (block: ${signerState[0]})`)
+          i++
+        }
+        // eslint-disable-next-line no-empty
+      } catch (e) {}
```
acolytec3, line 264: What's the error that could occur here?

packages/blockchain/src/consensus/clique.ts
```diff
@@ -400,12 +418,17 @@ export class CliqueConsensus implements Consensus {
...
    */
-  cliqueActiveSigners(): Address[] {
+  cliqueActiveSigners(blockNum: bigint): Address[] {
     const signers = this._cliqueLatestSignerStates
     if (signers.length === 0) {
       return []
     }
-    return [...signers[signers.length - 1][1]]
+    for (let i = signers.length - 1; i >= 0; i--) {
+      if (signers[i][0] < blockNum) {
```
acolytec3, line 427: Just clarify my understanding here, we're just returning the most recent signer state, (which might be from a block older than the current chain head if no signer state change occurs on the current block, right?)

acolytec3 (approved):
This all looks great.  Have a few clarifying questions but would love to merge this once final.

packages/blockchain/src/consensus/clique.ts
jochem-brouwer: I added this just to be sure, but looking at the code, it trims the `cliqueLatestSignerStates` after the first sort (so they are already ordered by block number) and thus removes latest signer states if they are below a certain threshold number (so the order should still be fine). I think we can remove.

packages/blockchain/src/consensus/clique.ts
jochem-brouwer: I should have commented because now I don't remember. I think it was something when I tried to sync Goerli, and then for some reason `signerState[0]` was undefined, which throws the method. (If you want I can re-investigate)

packages/blockchain/src/consensus/clique.ts
jochem-brouwer: I think you got the idea here, but just for clarity:

- This solves the reorg problem, because before it always returned whatever active signers there were at current chain head. (On client this could result in execution vs block import being out of sync which thus require different active signers if you verify them)
- Note that `_cliqueLatestSignerStates` is sorted
- We return the first candidate which has active signers < requested block num (so the state we want)
- Note that signers always has 1 member

packages/blockchain/src/consensus/clique.ts
```diff
@@ -455,6 +478,7 @@ export class CliqueConsensus implements Consensus {
     this._cliqueLatestSignerStates = this._cliqueLatestSignerStates.filter(
       (s) => s[0] <= blockNumber
     )
+    this._cliqueLatestSignerStates.sort((a, b) => (a[0] > b[0] ? 1 : -1))
```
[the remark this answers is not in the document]
jochem-brouwer: On each reorg they will be out-of-order and will mess up the latest states (this fixes reorgs in case voting changes). However since this is in the delete method and the assumption is that the signer state is always sorted, this should only pop items from the end of the array, so might indeed remove.

packages/blockchain/test/clique.spec.ts
```diff
@@ -310,19 +348,40 @@ tape('Clique: Initialization', (t) => {
...
+    await addNextBlock(blockchain, blocks, B, [D, true])
+    await addNextBlock(blockchain, blocks, C)
+    await addNextBlock(blockchain, blocks, A, [E, true])
+    await addNextBlock(blockchain, blocks, B, [E, true])
+    await addNextBlock(blockchain, blocks, C)
+
+    for (let i = 1; i < blocks.length; i++) {
+      await blockchain.putBlock(blocks[i])
+    }
+    st.end()
```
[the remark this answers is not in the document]
jochem-brouwer: Oh wow, the diff got completely messed up. Now it looks like I wrote a gigantic amount of new tests which is not the case. I must have accidentally moved some stuff around. I only added reorg tests and this specific test does not reorg. Note that if you `putBlock` and the state is wrong (either on reorg or not), then `putBlock` will throw. I will try to fix the diff.

packages/blockchain/test/clique.spec.ts
```diff
@@ -698,24 +778,188 @@ tape('Clique: Initialization', (t) => {
...
+      )
+      st.deepEqual(
+        (blockchain.consensus as CliqueConsensus).cliqueActiveSigners(
+          blocks[blocks.length - 1].header.number + BigInt(1)
+        ),
+        [A.address, B.address, C.address],
+        'address C added to signers'
+      )
+      st.ok(
+        equalsBytes((await blockchain.getCanonicalHeadBlock()).hash(), headBlockUnforked.hash())
```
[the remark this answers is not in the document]
jochem-brouwer: Oh really? I did not know that :)

jochem-brouwer:
I will pick up these points and cleanup. I will also not fix this epoch reorg thing, but I will comment out the test and leave it there for educational purposes, and explicitly comment that we do not support certain reorg situations, like epoch reorg. (And possibly others but I will not research).

In case you linearly import a PoA chain now, this should be fine. However, at tip of chain if there are reorgs, failure might occur. (Note that this is still somewhat fixable: you can rewind the chain back to latest epoch change block, read clique state from there, and then just replay the chain, which should now fix the issue on the reorg). So it might be unstable at tip of PoA chains.

packages/blockchain/test/clique.spec.ts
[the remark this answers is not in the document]
jochem-brouwer: Nvm, I had to move some logic into methods to create the `addNextBlockReorg` function. This specific test, I think I added this to test things internally. I think for completeness, lets keep it, but I think if we run this on the old version (without reorg logic) this test still passes anyways.

acolytec3:
Great, the couple of edits look good.  Are we ready to merge?

jochem-brouwer:
Yup.

jochem-brouwer:
(sorry did not update the PR status, just did :smile: )

acolytec3 (approved):
Lgtm

PR #2612, opened 2023-03-31
Implement EIP6780

jochem-brouwer:
Implements [EIP6780](https://eips.ethereum.org/EIPS/eip-6780). This is considered for inclusion by ACD.

Test currently fails, and I have no idea why, the storage root of the contract is the empty storage root, but somehow the contract still returns that the slot is non-empty.

### DIFF
--- a/packages/statemanager/src/stateManager.ts
+++ b/packages/statemanager/src/stateManager.ts
@@ -122,6 +122,11 @@ export class DefaultStateManager extends BaseStateManager implements StateManage
     })
   }
 
+  async deleteAccount(address: Address) {
+    delete this._storageTries[address.buf.toString('hex')]
+    await super.deleteAccount(address)
+  }
+
   /**
    * Adds `value` to the state trie as code, and sets `codeHash` on the account
    * corresponding to `address` to reference this.
--- a/packages/common/src/eips/6780.json
+++ b/packages/common/src/eips/6780.json
@@ -1,0 +1,12 @@
+{
+  "name": "EIP-6780",
+  "number": 6780,
+  "comment": "SELFDESTRUCT only in same transaction",
+  "url": "https://eips.ethereum.org/EIPS/eip-6780",
+  "status": "Draft",
+  "minimumHardfork": "london",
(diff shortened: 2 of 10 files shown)


packages/statemanager/src/stateManager.ts
```diff
@@ -122,6 +122,11 @@ export class DefaultStateManager extends BaseStateManager implements StateManage
     })
   }
 
+  async deleteAccount(address: Address) {
+    delete this._storageTries[address.buf.toString('hex')]
```
jochem-brouwer, line 126: In the case that a contract gets selfdestructed, this method gets called. However, before this change, it would not clear the storage trie cache. This means that while the contract should have empty storage, StateManager thinks the storage has not been wiped.

Does this lead to a consensus bug if you selfdestruct a CREATE2 contract and then CREATE2 the contract again? If EVM reads storage, does this contract then have storage or is it empty? It actually is, if we create a contract we ensure we wipe the storage first.

jochem-brouwer: The problem was this:

- Create a contract, which sets storage slot 1 to value 1. Selfdestruct this contract
- If you `getAccount` on the contract, it correctly returns an account which has nonce 0, balance 0, codeHash empty, storage empty
- However, if you `getContractStorage` on that address, and that key, it will return `Buffer < 01 >`. This is because the storage trie cache was not deleted, so storage trie still points to the contract storage state right before the selfdestruct.

This thus means that in some very exotic situations, if test runners (ganache, hardhat) try to test certain expected values, StateManager thus returns the wrongly cached values (so if you inspect StateManager it reports the wrong values). You could thus write a test which will not match the expected values due to not deleting the cached trie.

holgerd77:
I guess it likely makes sense to put this on hold a bit and proceed once we have a version of #2592 squashed and rebased towards `develop-v7` (so: somewhat of the final version) and then re-target this PR towards that work. Then we are safe that this remains stable.

jochem-brouwer:
Followed-up by #2733

PR #2613, opened 2023-03-31
Tx: correctly decode rpc txs

### LINKED ISSUE #2605: Tx: different behavior when decoding serialized txs vs decoding JSON txs
If you decode this tx: 

https://optimistic.etherscan.io/tx/0xf4304cb09b3f58a8e5d20fec5f393c96ccffe0269aaf632cb2be7a8a0f0c91cc

(serialized: https://optimistic.etherscan.io/getRawTx?tx=0xf4304cb09b3f58a8e5d20fec5f393c96ccffe0269aaf632cb2be7a8a0f0c91cc)

Then if you decode it using `TransactionFactory`, it will decode fine. However, if you decode the JSON, then it will feed '0x' to the TransactionFactory (instead of empty buffers). In `baseTransaction` we then decode this as `v=0`, while if you would feed the serialized tx it would set `v=undefined`.

jochem-brouwer:
Closes #2605 

On the Optimism chain there are system txs like this one: https://optimistic.etherscan.io/tx/0xf4304cb09b3f58a8e5d20fec5f393c96ccffe0269aaf632cb2be7a8a0f0c91cc. If you get the serialized tx and decode it, you will notice that v/r/s values (signature values) are empty Buffers. However, if you get the tx over RPC, then some/all RPCs will report v/r/s values as "0x0". This should (in this case?) thus be interpreted as empty/undefined values, and not be fed into `toBuffer`, because then this will decode as `Buffer < 00 >`.

Side note: to run these txs into VM, override `getSenderAddress` to return the zero address, and ensure that `isSigned` returns `true`.

Note: should investigate side effects, so lets not merge this super quickly.


jochem-brouwer:
This will need some more discussion, because due to the failing VM test it uncovered a problem:

If we create a FeeMarket tx, and then set `v = 0`, then if we convert it to JSON it will report back `v = '0x0'`. If we handle this the same as the `r` and `s` values, this would then result in the empty Buffer, which will set `v` to undefined. However, previously, if you would feed `v = '0x0'` to the tx constructor, it would set `v = Buffer < 00 >`. The difference between the two is that the first one is unsigned, while the latter one is.

So maybe we should not do this change after all, this gets confusing? Especially with different behavior with `r` and `s`.

acolytec3:
> This will need some more discussion, because due to the failing VM test it uncovered a problem:
> 
> If we create a FeeMarket tx, and then set `v = 0`, then if we convert it to JSON it will report back `v = '0x0'`. If we handle this the same as the `r` and `s` values, this would then result in the empty Buffer, which will set `v` to undefined. However, previously, if you would feed `v = '0x0'` to the tx constructor, it would set `v = Buffer < 00 >`. The difference between the two is that the first one is unsigned, while the latter one is.
> 
> So maybe we should not do this change after all, this gets confusing? Especially with different behavior with `r` and `s`.

So if a transaction received over RPC has `0x0` for r, s, and v, is that indicative of it being signed?  I'm trying to wrap my mind around what the actual question is.

packages/tx/src/transactionFactory.ts
```diff
@@ -110,4 +110,18 @@ export class TransactionFactory {
...
   }
+
+  /**
+   * Method to decode data retrieved from RPC, such as `eth_getTransactionByHash`
+   * Note that this normalizes some of the parameters
+   * @param txData The RPC-encoded data
+   * @param txOptions The transaction options
+   * @returns
+   */
+  public static async fromRPCTx(
```
acolytec3, line 121: Should we consider reusing this logic inside `fromEthersProvider`?  They're doing basically the same thing except `fromEthersProvider` does an additional fetch call to actually retrieve the data.

jochem-brouwer: Yup, great idea

acolytec3 (approved):
LGTM!

PR #2620, opened 2023-04-05
Add dataGasUsed to txReceipt and result

acolytec3:
This adds the `dataGasUsed` field to the `TxResult` and also adds a new receipt type for blob transactions which includes the DataGasUsed  and `DataGasPrice` values in the receipt.


packages/vm/src/types.ts
```diff
@@ -48,6 +48,13 @@ export interface PostByzantiumTxReceipt extends BaseTxReceipt {
   status: 0 | 1
 }
 
+export interface EIP4844BlobTxReceipt extends PostByzantiumTxReceipt {
+  /**
+   * Data gas consumed by a transaction
+   */
+  dataGasUsed: bigint
```
jochem-brouwer, line 55: I do not see this in the EIP, unless I am missing something. In `runBlock` we have `encodeReceipt` and in there we currently do not encode the `dataGasUsed`. I find it a bit confusing if for `PreByzantiumTxReceipt` and `PostByzantiumTxReceipt` the encoded values are RLPd (in a certain order) and put in the receipt trie (this is part of block header and must be validated), while for `EIP4844BlobTxReceipt` we add this extra field, but this is not encoded in the receipt.

Is there a pending PR for 4844 to add this?

g11tech: I think this is the PR @acolytec3 raised but yes makes sense to also include it in the encoded values

acolytec3: > I do not see this in the EIP, unless I am missing something. In `runBlock` we have `encodeReceipt` and in there we currently do not encode the `dataGasUsed`. I find it a bit confusing if for `PreByzantiumTxReceipt` and `PostByzantiumTxReceipt` the encoded values are RLPd (in a certain order) and put in the receipt trie (this is part of block header and must be validated), while for `EIP4844BlobTxReceipt` we add this extra field, but this is not encoded in the receipt.
> 
> Is there a pending PR for 4844 to add this?

Correct, the change is on the execution APIs rather than the EIP.  The discussion from the implementers call was that since it's receipt metadata that doesn't change the receipts root on the block, it shouldn't go in the EIP.

jochem-brouwer: Ah right now I see the PR. Is this automatically added in client `eth_getTransactionReceipt`?

g11tech: can we add here just a comment that this is not to be a part of the rlp encoding just as a reference for future

g11tech (approved):
lgtm

acolytec3:
I vote we don't merge this for now.  It shouldn't be too much work to rebase it if'/when the corresponding spec update happens so no point in adding it now when there's no real world use for it.

g11tech (approved):
lgtm!

PR #2622, opened 2023-04-06
Bump `@chainsafe/ssz` to `0.11.0`

FrederikBolding:
Bumps `@chainsafe/ssz` to `0.11.0`. 

The current version of `@chainsafe/ssz` uses an AssemblyScript based SHA-256 implementation that causes issues on some platforms since it requires `WebAssembly` for execution (e.g. in sites or extensions with strict CSPs). 

This new version uses a pure JS implementation (`@noble/hashes`) as the default hasher for SHA-256 and should fix some of these aforementioned incompatibility issues.

For more context see: https://github.com/ChainSafe/ssz/issues/313

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1695,9 +1695,9 @@
       "license": "MIT"
     },
     "node_modules/@chainsafe/as-sha256": {
-      "version": "0.3.1",
-      "resolved": "https://registry.npmjs.org/@chainsafe/as-sha256/-/as-sha256-0.3.1.tgz",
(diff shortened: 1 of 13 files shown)


FrederikBolding:
Hmm, looks like the `exports` field in package.json is not supported by `karma` 🤔

acolytec3:
I think I've resolved the karma import issue.  Will approve if CI passes

FrederikBolding:
There could be an argument that this should be solved in the SSZ package itself. Since other tools may struggle with the exports configuration too.

holgerd77 (approved):
Thanks a lot, cool 🙂, will approve and merge!

holgerd77:
> There could be an argument that this should be solved in the SSZ package itself. Since other tools may struggle with the exports configuration too.

If you solve this in the SSZ package let us know, then we likely also want to do another update on this. We have releases planned for roughly in 10 days or so, if this will happen it would be great if we could already take the updated version in.

PR #2621, opened 2023-04-05
Fix `pyspec` hive tests

jochem-brouwer:
This PR fixes the `pyspec` hive tests. These tests are tests from the `execution-spec-tests` [repo](https://github.com/ethereum/execution-spec-tests) and re-execute these tests, but now using the engine API of the client.

No consensus bugs were found. However, some missing features and some annoying stuff has been fixed in this PR: (most of those are edge cases, except genesis files where nonce is nonzero for some accounts)

- When importing genesis state (for instance, from a Geth genesis), also read the nonce (if available)
- - This caused some pyspec tests not to run, because the provided state after putting this genesis has a different state root than what the genesis block actually points to (which results in client reporting: parent block not executed yet)
- Ensures that if genesis blocks have very large (larger than `Number.MAX_SAFE_INTEGER`) `gasLimit` or `difficulty`, this is parsed right by using `BigInt`
- Ensures in `Trie` that if we request if the empty state root (so 0 accounts, not even empty ones) is available, it returns `true` (this value is not in the DB). Fixes a test which starts with empty genesis allocation.
- Fixes setting up the genesis trie db in VM where, if you put empty accounts, this account is actually in the touched list and thus deleted on the first tx run (only exception of course if this tx creates a non-empty account)
- Updates engine-only RPC to only require the strict methods which should be available by current spec
- - Found "bugs" on hive, where some methods which are not required were called on the engine-only port: https://github.com/ethereum/hive/pull/752
- Fixes a nasty instance where you create a private key file which ends with a newline (private key import then thinks private key is 33 bytes instead of expected 32 bytes). Now client removes all newlines from the private key file.

Before this PR: fail 93/94. Current: fail 0/94

To run:

- Ensure you have updated hive to the latest version
- From hive: `go run hive.go --docker.output --client ethereumjs --sim ethereum/pyspec`

PR #2630, opened 2023-04-11
Statemanager: Cache Rewrite (Account/Storage) & Refactoring (Interface/Tests) (develop-v7 Rebase)

holgerd77:
This PR is a rebased version of #2592 towards `develop-v7` and replaces the old PR.

Rebase is done, some tests are not passing yet.

Continued reviews can be done here or on the old PR (if easier e.g.), will do the review updates here.


packages/statemanager/src/stateManager.ts
```diff
@@ -157,43 +318,41 @@ export class DefaultStateManager extends BaseStateManager implements StateManage
...
-    storageTrie.root(account.storageRoot)
-    storageTrie.flushCheckpoints()
-    return storageTrie
+    const keyHex = bytesToHex(key)
+    if (keyHex in this._codeCache) {
+      return this._codeCache[keyHex]
+    } else {
+      // @ts-expect-error
+      const code = (await this._trie._db.get(key)) ?? new Uint8Array(0)
+      this._codeCache[keyHex] = code
```
jochem-brouwer, line 337: Had to take a look, but it seems like we do not persist code cache / storage trie cache if we set state root. So this code/storage tries cache only works within a block. Beyond the scope of this PR but we could also (possibly re-use the account cache class?) to do the same for codeCache/storage trie cache?

packages/statemanager/src/stateManager.ts
```diff
@@ -157,43 +318,41 @@ export class DefaultStateManager extends BaseStateManager implements StateManage
...
     const addressHex = bytesToHex(address.bytes)
-    let storageTrie = this._storageTries[addressHex]
-    if (storageTrie === undefined || storageTrie === null) {
-      // lookup from state
-      storageTrie = await this._lookupStorageTrie(address)
+    const storageTrie = this._storageTries[addressHex]
+    if (storageTrie === undefined) {
+      const storageTrie = this._trie.copy(false)
+      storageTrie.root(account.storageRoot)
+      storageTrie.flushCheckpoints()
```
jochem-brouwer, line 354: Hmm should it not push this `storageTrie` in `_storageTries` cache here?

packages/statemanager/src/stateManager.ts
```diff
@@ -284,12 +444,49 @@ export class DefaultStateManager extends BaseStateManager implements StateManage
...
+
   /**
    * Clears all storage entries for the account corresponding to `address`.
    * @param address -  Address to clear the storage of
    */
   async clearContractStorage(address: Address): Promise<void> {
-    await this._modifyContractStorage(address, (storageTrie, done) => {
+    const account = await this.getAccount(address)
+    if (!account) {
+      throw new Error(`clearContractStorage() called on non-existing account (${address})`)
```
jochem-brouwer, line 486: If you selfdestruct when you create a contract, would this not throw?

packages/statemanager/src/stateManager.ts
```diff
@@ -333,20 +570,20 @@ export class DefaultStateManager extends BaseStateManager implements StateManage
    */
   async getProof(address: Address, storageSlots: Uint8Array[] = []): Promise<Proof> {
     const account = await this.getAccount(address)
+    if (!account) {
+      throw new Error(`getProof() can only be called for an existing account`)
```
jochem-brouwer, line 574: I disagree, you can create a proof that an account does not exist in the trie. (non-existance proof)

packages/statemanager/test/cache/account.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    cache.revert()
+
+    elem = cache.get(addr)
+    st.ok(elem !== undefined && elem.accountRLP && equalsBytes(elem.accountRLP, accRLP))
+
+    st.end()
+  })
+
+  t.test('cache clearing', async (st) => {
```
jochem-brouwer, line 82: There is no test which checks the `commit`? (Although this is covered by various other tests under-the-hood of course)

jochem-brouwer: Nvm, this is in the checkpointing tests file.

packages/statemanager/test/proofStateManager.spec.ts
```diff
@@ -17,16 +17,17 @@ tape('ProofStateManager', (t) => {
     const code = hexStringToBytes('6000')
     const stateManager = new DefaultStateManager()
     await stateManager.checkpoint()
+    await stateManager.putAccount(address, new Account())
     await stateManager.putContractStorage(address, key, value)
     await stateManager.putContractCode(address, code)
     const account = await stateManager.getAccount(address)
```
jochem-brouwer, line 23: I think you can do this:
```suggestion
    const account = (await stateManager.getAccount(address))!
```

This lets typescript know that it will not return `undefined`.

packages/statemanager/test/stateManager.account.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+        const res1 = await stateManager.getAccount(address)
+
+        st.equal(res1!.balance, BigInt(0xfff384))
+
+        await stateManager.flush()
+        stateManager._accountCache?.clear()
+
+        const res2 = await stateManager.getAccount(address)
+
+        st.ok(equalsBytes(res1!.serialize(), res2!.serialize()))
```
jochem-brouwer, line 105: This test does not test if the account is added to the cache as far as I can see.

packages/statemanager/test/stateManager.account.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+      await stateManager.putAccount(address, account)
+
+      const account0 = await stateManager.getAccount(address)
+      st.equal(account0!.balance, account.balance, 'account value is set in the cache')
+
+      await stateManager.commit()
+      const account1 = await stateManager.getAccount(address)
+      st.equal(account1!.balance, account.balance, 'account value is set in the state trie')
+
+      await stateManager.setStateRoot(initialStateRoot)
```
jochem-brouwer, line 60: I think before this, add a test if the cache is non-empty, and after setting state root then test if the cache is empty.

packages/statemanager/test/stateManager.code.spec.ts
jochem-brouwer, line 1: It looks like this file (and possibly other test files) are moved from somewhere (or is this splitting `stateManager.spec.ts` into multiple files?). If this is moved from somewhere it would be nice if this would be moved with `git mv` to preserve history.

packages/vm/src/eei/eei.ts
```diff
@@ -39,7 +39,10 @@ export class EEI extends VmState implements EEIInterface {
    * @param address - Address of account
    */
   async getExternalBalance(address: Address): Promise<bigint> {
-    const account = await this.getAccount(address)
+    let account = await this.getAccount(address)
+    if (!account) {
+      account = new Account()
```
jochem-brouwer, line 44:
```suggestion
      return BigInt(0)
```

packages/vm/src/eei/vmState.ts
```diff
@@ -267,6 +267,7 @@ export class VmState implements EVMStateAccess {
     }
     await this._stateManager.flush()
     // If any empty accounts are put, these should not be marked as touched (when first tx is ran, this account is deleted when it cleans up the accounts)
+    // TODO: not 100% sure if to keep this line or not along cache refactor rebase, 2023-04-11
```
jochem-brouwer, line 270: I added this in #2621 :) 

We definitely need this. It is an edge case when dumping genesis:

1. Create a genesis with allocation. In this, there is at least one account which is empty. 
2. Now run a block with txs. These txs do not touch the empty account.
3. After the first tx, the account which is empty is cleaned up.

Note that the empty state trie vs. a state trie with 1 empty account has a different state root.

(Removing this line will break `pyspec` hive tests)

packages/vm/src/runBlock.ts
```diff
@@ -42,6 +42,7 @@ const DAORefundContract = DAOConfig.DAORefundContract
 export async function runBlock(this: VM, opts: RunBlockOpts): Promise<RunBlockResult> {
   const state = this.eei
   const { root } = opts
+  const clearCache = opts.clearCache ?? true
```
jochem-brouwer, line 45: I think what we should do here is:

1. Check current state root
2. Check parent header state root
3. If this is equal, then do not clear cache
4. If it is non-equal then clear cache
5. Of course if `opts.clearCache` is defined this overrides above logic.

packages/vm/src/vm.ts
```diff
@@ -179,10 +179,11 @@ export class VM {
       // put 1 wei in each of the precompiles in order to make the accounts non-empty and thus not have them deduct `callNewAccount` gas.
       for (const [addressStr] of getActivePrecompiles(this._common)) {
         const address = new Address(hexToBytes(addressStr))
-        const account = await this.eei.getAccount(address)
+        let account = await this.eei.getAccount(address)
         // Only do this if it is not overridden in genesis
         // Note: in the case that custom genesis has storage fields, this is preserved
-        if (account.isEmpty() === true) {
+        if (account === undefined) {
```
jochem-brouwer, line 185: I think this should be
```suggestion
        if (account === undefined || account.isEmpty()) {
```

Such that we do not change behavior.

jochem-brouwer (commented):
Some questions and comments but this looks great overall! :smile:

packages/statemanager/src/cache/storage.ts
```diff
@@ -0,0 +1,374 @@
...
+/**
+ * key -> storage mapping
+ *
+ * undefined: storage value is known not to exist in the cache
+ */
+type DiffStorageCacheMap = OrderedMap<string, Uint8Array | undefined>
+type StorageCacheMap = OrderedMap<string, Uint8Array>
+
+export class StorageCache extends Cache {
+  _lruCache: LRUCache<string, StorageCacheMap> | undefined
```
acolytec3, line 23: As a general question, is there any advantage to using the `LRUCache` for the top level cache if we are using `OrderedMaps` for each individual contract storage cache?  Would it be possible to make the `StorageCacheMap` an LRUCache as well?

acolytec3 (changes requested):
Would like to clarify the usage of a few things before we merge.

1) Can we use LRUCaches for the `StorageCacheMap` so that cache type is truly an `LRUCache`  or do we need the `OrderedMaps`?

2) For the `EthersStateManager`, I think it's dangerous that we're not actually copying the cache object but just having the copy use the same cache as the original.  We should deep copy this if possible or else just not copy at all (if we're going this route)

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -61,7 +48,7 @@ export class EthersStateManager extends BaseStateManager implements StateManager
     })
     ;(newState as any).contractCache = new Map(this.contractCache)
     ;(newState as any).storageCache = new Map(this.storageCache)
-    ;(newState as any)._cache = this._cache
+    ;(newState as any)._accountCache = this._accountCache
```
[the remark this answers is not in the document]
acolytec3: Will open a separate issue to remind me to revisit the `EthersStateManager`.  Let's not worry about it here.

holgerd77:
Will admin merge now here after exchange on chat, so that this is not blocking other work.

I'll address the review comments in a follow-up PR.

packages/client/lib/rpc/modules/eth.ts
```diff
@@ -517,6 +516,9 @@ export class Eth {
     const vm = await this._vm.copy()
     await vm.stateManager.setStateRoot(block.header.stateRoot)
     const account = await vm.stateManager.getAccount(address)
+    if (account === undefined) {
+      throw new Error(`could not read account`)
```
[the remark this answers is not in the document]
holgerd77: Tried to confirm "by spec" but helpless 😬, think we are just not that far with the RPC specification, the one on https://github.com/ethereum/execution-apis/ is not doing the job yet (so not sure where you have your secret information from, haha).

Anyhow, tested with infura on a non-existing account with `{"jsonrpc":"2.0","method":"eth_getBalance","params":["0x1122334455667788991011121314151617181920", "0x100"],"id":1}` and this indeed returns `{"jsonrpc":"2.0","id":1,"result":"0x0"}`.

Will update. 🙂 

(side note: along implementation I didn't think about this at all what should be the general behavior here, so thanks for the mention respectively for catching this)

packages/client/lib/rpc/modules/eth.ts
```diff
@@ -687,7 +689,10 @@ export class Eth {
     await vm.stateManager.setStateRoot(block.header.stateRoot)
 
     const address = Address.fromString(addressHex)
-    const account: Account = await vm.stateManager.getAccount(address)
+    const account = await vm.stateManager.getAccount(address)
+    if (account === undefined) {
+      throw new Error(`could not read account`)
```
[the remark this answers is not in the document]
holgerd77: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2634/commits/28b698faeb7ad9cdd09c419698023446d71cb32b

packages/client/test/rpc/eth/getStorageAt.spec.ts
```diff
@@ -109,8 +113,11 @@ tape(`${method}: call with valid arguments`, async (t) => {
   await vm.runBlock({ block: block2, generate: true, skipBlockValidation: true })
   await vm.blockchain.putBlock(ranBlock2!)
 
+  // TODO: fix tests
```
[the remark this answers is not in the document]
holgerd77: Yes, I am not able to do this, maybe someone can help, I think @scorbajio might have set up these tests?

packages/evm/src/evm.ts
```diff
@@ -617,8 +633,9 @@ export class EVM implements EVMInterface {
         // It is thus an unnecessary default item, which we have to save to dik
         // It does change the state root, but it only wastes storage.
         //await this._state.putContractCode(message.to, result.returnValue)
+        //const account = await this.eei.getAccount(message.to)
         const account = await this.eei.getAccount(message.to)
-        await this.eei.putAccount(message.to, account)
+        await this.eei.putAccount(message.to, account!)
```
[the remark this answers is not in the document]
holgerd77: Ah ok, just re-read, this is exactly what is written in the comment (guess it's from you, right?), so that pre-Homestead there is explicitly an empty account created in this situation, have updated https://github.com/ethereumjs/ethereumjs-monorepo/pull/2634/commits/87e4541bb21cd5116fde0efe55b1b4b73d14965a.

Have directly pushed, interested in what this is doing with the tests.

packages/statemanager/test/stateManager.code.spec.ts
holgerd77: Ah no, not possible. These tests are extracted from the old `stateManager.spec.ts` as you mentioned as a possibility. So I could only preserve history for one of the files.

packages/evm/src/evm.ts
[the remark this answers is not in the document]
holgerd77: (unfortunately I didn't have the all-HFs label activated on that push. Now added so we'll see on push after that. Might eventually fix the remaining 6 blockchain tests? 🤔)

packages/evm/src/interpreter.ts
```diff
@@ -474,6 +485,9 @@ export class Interpreter {
   async storageStore(key: Uint8Array, value: Uint8Array): Promise<void> {
     await this._eei.storageStore(this._env.address, key, value)
     const account = await this._eei.getAccount(this._env.address)
+    if (!account) {
+      throw new Error('could not read account along persisting memory')
```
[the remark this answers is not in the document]
holgerd77: Yes, so it is more of an error to be meant to discover eventual implementation flaws early on.

packages/vm/src/eei/eei.ts
holgerd77: Have updated, thanks.

packages/vm/src/eei/vmState.ts
holgerd77: Ok, have removed my comment.

packages/vm/src/runBlock.ts
holgerd77: Hmm, I am not seeing how we can do this without going out of the scope of this method. I think storing the block from the previous `runBlock()` execution and then comparing to that is definitely not the right approach respectively happening "at the wrong place", this would rather be something for the client (if at all).

For VM one can have all sorts of setups (execute `runBlock()`, manually set to an own state root, run another block, whatever). So this narrows things down to much.

Let me know if I understand something wrong about your suggestion though.

packages/vm/src/vm.ts
holgerd77: Hmm, before we had the behavior that `getAccount()` *was* just always returning an empty account also for a non-existing one, so we had no choice.

I would argue that coming from the situation here it would just make no sense that an empty account is already present for a precompile, so removing this (previously dirty-included) `account.isEmpty()` seems rather like a clean-up respectively bug fix to me, pretty much safe to be done.

Let me know if I am missing some scenarios.

packages/statemanager/src/cache/storage.ts
```diff
@@ -0,0 +1,374 @@
...
+      if (storageMap !== undefined) {
+        const itDiffStorage = diffStorageMap.begin()
+        while (!itDiffStorage.equals(diffStorageMap.end())) {
+          const keyHex = itDiffStorage.pointer[0]
+          const value = storageMap.getElementByKey(keyHex)
+          items.push([addressHex, keyHex, value])
+          itDiffStorage.next()
+        }
+      } else {
+        throw new Error('Inconsistent cache state')
```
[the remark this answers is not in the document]
holgerd77: This is just an internal error similar to the one Jochem mentioned above signaling an implementation error if thrown since normally `storageMap` should be defined at that point.

packages/statemanager/src/cache/storage.ts
holgerd77: Yes, this is an intentional decision.

So currently/now the storage cache is implemented in a two way structure. There is the outer account structure, this is the main cache structure which can be switched from LRU to ORDERED_MAP. Every account entry then holds this `StorageCacheMap` where storage slots requested are cached/stored.

This leads to the overall functionality of the cache that - for LRU - cache items are thrown away if the account - as a whole, so any storage slot - hasn't been touched for a certain while. One can argue about that, if this is the most optimal or not, I would say this is at least "optimal enough" and *might* even be the superior solution (with the argument that - if some storage slots from an account are touched *at all*, it is likely that some other slots are also requested).

But these optimal-or-not questions are not the main reason I implemented it this way. I actually had the full other version going at some point in time, this was doing a map (or LRU) from `address_slotKey`-> `slotValue`.

This was then very much throwing (for LRU) just the oldest not-used specific slots away, also structure was somewhat simpler. The unsolvable hurdle I came across along implementation was that we have functionality like `clearTouchedAccounts()`, where it is necessary to re-retrieve all cached slots for one specific account: for this one-layer structure this is just not possible any more (without building up complex side structures).

So far on the context. 😋

For this new structure we only store a couple of storage slots for each account in this additional structure. If we would use an LRU cache for these objects as well that would create a huge overhead, especially on the memory side, if we instantiate these fixed size objects for each and every account, especially regarding that a reasonable cache size is roughly 200.000 or so (that's the client default), so we would create 200.000 LRU objects with this (and we already had memory problems with 1, lol 😋).

Yeah, so. That's the long version of the reasoning togehter with some context.

holgerd77:
@acolytec3 I guess 2. (EthersStateManager copy()) is a bit extracted and we will solve separately, right?

packages/statemanager/src/cache/account.ts
```diff
@@ -0,0 +1,276 @@
...
+    if (this.DEBUG) {
+      this._debug(`Flushing cache on checkpoint ${this._checkpoints}`)
+    }
+
+    const diffMap = this._diffCache[this._checkpoints]!
+    const it = diffMap.begin()
+
+    const items: [string, AccountCacheElement][] = []
+
+    while (!it.equals(diffMap.end())) {
```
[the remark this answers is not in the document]
holgerd77: Ah, likely yes. At the moment I wouldn't want to do these kind of changes, too much stuff not completely stable and I wouldn't want to break anything additionally. So might do later on as an optimization (likely for various occasions in the code).

packages/statemanager/src/cache/account.ts
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    this._stats.dels += 1
+  }
+
+  /**
+   * Flushes cache by returning accounts that have been modified
+   * or deleted and resetting the diff cache (at checkpoint height).
+   */
+  async flush(): Promise<[string, AccountCacheElement][]> {
```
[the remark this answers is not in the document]
holgerd77: 👍

packages/statemanager/src/stateManager.ts
holgerd77: Yes, definitely. 🙂 Already thought about doing this but refrained since things are complex enough. But we can and likely want to add this third cache piece as well I guess.

packages/statemanager/src/stateManager.ts
holgerd77: Oh, super good point, definitely, have added. 👍 

(will push combined later on)

packages/evm/src/opcodes/gas.ts
```diff
@@ -521,7 +524,7 @@ export const dynamicGasHandlers: Map<number, AsyncDynamicGasHandler | SyncDynami
         if (value > BigInt(0)) {
           gas += common.param('gasPrices', 'authcallValueTransfer')
           const account = await runState.eei.getAccount(toAddress)
-          if (account.isEmpty()) {
+          if (!account) {
```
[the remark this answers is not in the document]
holgerd77: Can't 100% answer this in the context to be honest. Again, regarding that `getAccount()` is now returning `undefined` and not a combined empty account for (undefined || empty) I would assume for this situation that `undefined` is meant since the gas price param is also called `callNewAccount`.

Does this make sense?

packages/evm/src/interpreter.ts
```diff
@@ -850,6 +864,9 @@ export class Interpreter {
       Object.assign(this._result.selfdestruct, selfdestruct)
       // update stateRoot on current contract
       const account = await this._eei.getAccount(this._env.address)
+      if (!account) {
+        throw new Error('could not read contract account')
```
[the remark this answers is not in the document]
holgerd77: Yeah, similar like the other case, then this is just a guard towards an implementation failure (or wrong test setup hack or something).

packages/statemanager/src/stateManager.ts
holgerd77: Feeling a bit shaky here, also can vaguely remember that I battled a bit with this addition here (can't fully recall the context though).

I followed the `SELFDESTRUCT` call chain again, seems reasonable to me that a SELFDESTRUCT can also be executed on a non-existing account and at the end we get into this situation, so I updated here: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2634/commits/b13f85b04a625298301236caaac9f2581fb43eb6

I also ran EVM and VM (API) tests which passed.

packages/statemanager/src/stateManager.ts
holgerd77: Ah, this non-existance thing 😋, thanks again @acolytec3 for updating! 🔥 🙏

(I am generally still no proof expert, so the proof parts are the weakest parts of this PR)

packages/statemanager/test/proofStateManager.spec.ts
holgerd77: Not sure about the context here, is this solved, respectively what was the problem to be solved here?

packages/statemanager/src/cache/storage.ts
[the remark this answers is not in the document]
acolytec3: It is, though maybe we could make the message a little more descriptive like `storageMap should be defined at this point` or something so I or other team members who don't yet grasp all of the cache inner workings? :grimacing:

packages/statemanager/test/stateManager.account.spec.ts
holgerd77: No, but this is also not a cache test, so goal here was to test StateManager functionality.

packages/statemanager/test/stateManager.account.spec.ts
holgerd77: Again, didn't want to mix StateManager API and Cache tests too much.

packages/statemanager/src/cache/storage.ts
holgerd77: 👍 

I guess I chose `OrderedMap` since I assumed it to be faster (at least I think that was the reason, I CAN remember that I HAD a reason and thought about that, lol). We can at some point do experiments, switch. Not totally easy since these iterations would need to be completely rewritten.

Side note: I haven't forgotten all your other review comments in the old PR, just accidentally started by addressing the ones in this newer PR! 🙂

packages/statemanager/src/cache/storage.ts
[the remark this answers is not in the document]
holgerd77: Yes, makes sense, have updated.

packages/evm/src/opcodes/gas.ts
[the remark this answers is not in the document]
jochem-brouwer: Oh right good point. I think the new version is actually better than the old version :thinking:

packages/statemanager/test/proofStateManager.spec.ts
jochem-brouwer: It's not a problem at all, it is more a bit of a code helper / stylish thing. There are a lot of `account!.<field-to-index>` below, but if you let TS know that the account exists (so is not undefined) then all those `account!` lines below can just be renamed to `account`.

packages/vm/src/runBlock.ts
jochem-brouwer: I have no strong opinion either, current implementation is fine!

packages/evm/src/opcodes/gas.ts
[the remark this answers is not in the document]
holgerd77: That was the intent at least. 🙂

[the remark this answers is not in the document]
holgerd77: Guess we will stumble upon several occasions where we have to (and are now able to) re-answer this same or somewhat adopted question: what was actually meant? Undefined or empty? (which should be a good thing)

packages/client/test/rpc/eth/getStorageAt.spec.ts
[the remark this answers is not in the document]
scorbajio: I didn't set them up. Git blame shows them being 2 years old.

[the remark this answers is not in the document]
scorbajio: Not sure, they don't look familiar. These tests were committed 2 years ago according to git blame.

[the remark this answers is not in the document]
holgerd77: Ah, thanks for the context. We might just want to replace with a simpler setup, seems the easiest way to do, so something like just directly set some storage with the statemanager from `vmexecution` after client initialization and see if this storage can get accessed by `getStorageAt` and leave all this outer setup. Seems to be a lot more future-proof regarding maintenance of these tests.

PR #2633, opened 2023-04-12
Kick `ethers` off the `block`

acolytec3:
Removes `ethers` from `block` and `tx` libraries as a dependency and replace `fromEthersProvider` internals with `fetch` calls to retrieve data from provider
- Creates new `fetchFromProvider` utility in `util` for retrieving data from a generic JSON RPC provider
- Removes dependency on `ethers` from both `tx` and `block`
- Fixes various tests
- Updates testdouble to latest version and fixes `td.replace` typing in `client`/`devp2p`


g11tech (commented):
looks good, assuming you will substitute micro-ftch

holgerd77 (commented):
One additional review question

packages/block/test/from-rpc.spec.ts
```diff
@@ -185,13 +186,29 @@ tape('[fromRPC] - Alchemy/Infura API block responses', (t) => {
...
+    if (
+      req.method === 'eth_getBlockByHash' &&
+      req.params[0] === '0x1850b014065b23d804ecf71a8a4691d076ca87c2e6fb8fe81ee20a4d8e884c24'
+    ) {
+      const block = await import(`./testdata/infura15571241wtxns.json`)
+      return block
+    }
+  }
+
+  const providerUtils = require('@ethereumjs/util/dist/provider')
```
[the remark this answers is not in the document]
acolytec3: I have to get a concrete reference to the function I want to stub using `testdouble` and this is the only way I could come up with to do it.  I spent several hours trying to stub the `fetch` calls directly without success when I was using `cross-fetch` as the backend.  I might be able to do more direct stubbing now that we're using `micro-ftch` so will see if I can make it work correctly.

packages/block/src/block.ts
```diff
@@ -230,37 +231,50 @@ export class Block {
...
   public static fromEthersProvider = async (
-    provider: ethers.providers.JsonRpcProvider | string,
+    provider: any,
     blockTag: string | bigint,
     opts: BlockOptions
   ) => {
     let blockData
-    const prov =
-      typeof provider === 'string' ? new ethers.providers.JsonRpcProvider(provider) : provider
+    const providerUrl = getProvider(provider)
```
g11tech, line 239: so the provider is essentially being stripped out. not sure if this is the right approach as the provider might be providing some wrapper functionality (like proof seeking and verification)

May be we can later revisit this on how to best interface a provider

packages/client/test/service/lightethereumservice.spec.ts
```diff
@@ -12,12 +12,12 @@ tape('[LightEthereumService]', async (t) => {
...
   PeerPool.prototype.close = td.func<any>()
-  td.replace('../../lib/net/peerpool', { PeerPool })
+  td.replace<any>('../../lib/net/peerpool', { PeerPool })
   const MockChain = td.constructor([] as any)
   MockChain.prototype.open = td.func()
-  td.replace('../../lib/blockchain', { MockChain })
+  td.replace<any>('../../lib/blockchain', { MockChain })
   const LesProtocol = td.constructor([] as any)
-  td.replace('../../lib/net/protocol/lesprotocol', { LesProtocol })
+  td.replace<any>('../../lib/net/protocol/lesprotocol', { LesProtocol })
```
g11tech, line 20: just wondering why do we have to now explicitly type it to any?

g11tech (approved):
two comments/observations, but looks good to me

packages/client/test/service/lightethereumservice.spec.ts
acolytec3: Testdouble did a minor version upgrade that got picked up when I added it to `tx` and `block` as a devDependency and their `td.replace` function now explicitly returns an object of type `unknown` instead `any` if you don't specify the type to return.  I can back this version bump out and pin to the previous version if people are concerned about it.

packages/block/src/block.ts
acolytec3: Where would a provider do proof verification?  We could put cleanly define the provider interface (at least for the minimal functions we'd need to call) I supposed but that feels like more work than it's worth and very brittle if `ethers` or `web3js` (if we support that provider type too in the future) ever change their API.  At the end of the day, we're just making an HTTP call to some remote RPC provider to get the json blob back for the blob or the tx.  But, that said, happy to discuss further.  We need to add a todo on the breaking release to rename the `fromEthersProvider` functions anyway.

holgerd77: Can we eventually test for the Ethers import to work (in a try/catch block) and only do this stripping out as a fallback solution? Then we might somewhat have best of both worlds.

acolytec3: I guess we could but all we were doing before was calling the `provider.send` method which is just abstracting a raw JSON RPC call.  Here again, we weren't using any provider-specific functionality, just a simple wrapper around HTTP requests.  As such, I'm not sure what we gain since we'd have to blindly call the provider and hope nothing in the API has changed.  I don't think what I've done in this PR is any worse than that.

holgerd77: Ok, good reasoning, then let's merge here.

I would now plan a round of "normal" releases for sometime mid next week, will prepare release notes next Monday or so.

holgerd77 (approved):
LGTM

PR #2635, opened 2023-04-13
client: Fix breaking tests because of scheduled mainnet shanghai hf

g11tech:
Fix breaking tests because of scheduled mainnet shanghai hf


holgerd77 (approved):
LGTM

PR #2640, opened 2023-04-14
tx: Fix the deserialization of blob tx and add no empty blobs validation

g11tech:
In the Uint8Array conversion a bug might have gotten  introduced for deserialization. This PR fixes that as well as ban zero blob tx

ref: 
- https://github.com/ethereum/EIPs/pull/6863

### DIFF
--- a/packages/tx/src/eip4844Transaction.ts
+++ b/packages/tx/src/eip4844Transaction.ts
@@ -50,6 +50,10 @@ const validateBlobTransactionNetworkWrapper = (
   if (!(versionedHashes.length === blobs.length && blobs.length === commitments.length)) {
     throw new Error('Number of versionedHashes, blobs, and commitments not all equal')
   }
+  if (versionedHashes.length === 0) {
+    throw new Error('Invalid transaction with empty blobs')
+  }
+
   try {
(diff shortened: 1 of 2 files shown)


packages/tx/src/eip4844Transaction.ts
```diff
@@ -265,7 +269,8 @@ export class BlobEIP4844Transaction extends BaseTransaction<BlobEIP4844Transacti
       const accessListItem: AccessListBytesItem = [listItem.address, listItem.storageKeys]
       accessList.push(accessListItem)
     }
-    const to = tx.to.value === null ? undefined : Address.fromString(bytesToHex(tx.to.value))
+    const to =
+      tx.to.value === null ? undefined : Address.fromString(bytesToPrefixedHexString(tx.to.value))
```
acolytec3, line 273: Great catch!

packages/tx/test/eip4844.spec.ts
```diff
@@ -108,6 +108,19 @@ tape('fromTxData using from a json', (t) => {
   }
 })
 
+tape('fromSerializedTx - from bytes', (t) => {
```
acolytec3, line 111: Thanks for adding this test!

acolytec3 (approved):
LGTM!

PR #2642, opened 2023-04-14
client: add proofs to blobsbundle

g11tech:
Add proofs to blobsbundle and consequently to getblobsbundle engine api call so that CL doesn't need to compute proof which ideally would be coming packaged in the network wrapper serialized tx

ref:
- https://github.com/ethereum/execution-apis/pull/392

### DIFF
--- a/packages/client/test/miner/pendingBlock.spec.ts
+++ b/packages/client/test/miner/pendingBlock.spec.ts
@@ -5,6 +5,7 @@ import {
   Account,
   Address,
   blobsToCommitments,
+  blobsToProofs,
   bytesToHex,
   bytesToPrefixedHexString,
   commitmentsToVersionedHashes,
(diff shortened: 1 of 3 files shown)


holgerd77:
Rebased this via UI

holgerd77 (changes requested):
It would be good to have at least 1-2 small test cases for this so to minimally make sure that data is passed around and encoded/decoded in the correct format.

packages/client/test/miner/pendingBlock.spec.ts
```diff
@@ -298,8 +301,13 @@ tape('[PendingBlock]', async (t) => {
...
     const payloadId = await pendingBlock.start(vm, parentBlock)
     await pendingBlock.build(payloadId)
-    const pendingBlob = pendingBlock.blobBundles.get(bytesToPrefixedHexString(payloadId))?.blobs[0]
+
+    const blobBundle = pendingBlock.blobBundles.get(bytesToPrefixedHexString(payloadId))!
+    st.ok(blobBundle !== undefined)
+    const pendingBlob = blobBundle.blobs[0]
     st.ok(pendingBlob !== undefined && equalsBytes(pendingBlob, blobs[0]))
+    const blobProof = blobBundle.proofs[0]
+    st.ok(blobProof !== undefined && equalsBytes(blobProof, proofs[0]))
```
g11tech, line 310: @holgerd77 this checks if the bundle has the proof and valdiates it as well

g11tech:
rebased via UI again

holgerd77 (approved):
LGTM

PR #2646, opened 2023-04-17
Client: Fix eth_getStorage RPC method

holgerd77:
Just looked a bit into the eth_getStorage RPC method, this is respectively was generally broken respectively over-promising already before, since it worked on the simple storageTrie cache datastructure and therefore couldn't deliver anything beyond `latest`.

For now I would also suggest that we fix along that line, since anything else will be getting complicated regarding the interplay with the cache (no new thing to fix) and either hitting hard on performance or introducing needs to flush the cache for just one getStorage call. All relatively complex questions to answer.

There were other bugs like return on no-account or the need to always pad the returned value, at least that's what I got when testing with Geth, e.g.:

```shell
{"jsonrpc":"2.0","id":1,"result":"0x0000000000000000000000000000000000000000000000000000000000000000"}
```

Still WIP.

Don't want to spend too much time on this right now but there is still a real/saved storage slot not returned, might be something not working with SM copy or something.

I've now done an easier test setup which can be easier maintained and expanded with new test cases, still a lot of stuff untested here.


packages/client/lib/rpc/modules/eth.ts
```diff
@@ -613,25 +613,30 @@ export class Eth {
    *   3. integer block number, or the string "latest", "earliest" or "pending"
    */
   async getStorageAt(params: [string, string, string]) {
-    const [addressHex, positionHex, blockOpt] = params
-    const block = await getBlockByOption(blockOpt, this._chain)
+    const [addressHex, keyHex] = params
+    const emptySlotStr = `0x${'00'.repeat(32)}`
```
jochem-brouwer, line 617: This should be cached instead of being set on each call

packages/client/lib/rpc/modules/eth.ts
```diff
@@ -613,25 +613,30 @@ export class Eth {
...
 
     if (this._vm === undefined) {
       throw new Error('missing vm')
     }
 
     const vm = await this._vm.copy()
-    await vm.stateManager.setStateRoot(block.header.stateRoot)
+    // TODO: this needs more thought, keep on latest for now
+    // const block = await getBlockByOption(blockOpt, this._chain)
```
jochem-brouwer, line 625: This does not work?

jochem-brouwer (commented):
2 comments. If I understood it correctly this only support latest block, right? (Also as per the PR message on top?)

packages/client/test/rpc/eth/getStorageAt.spec.ts
```diff
@@ -1,154 +1,53 @@
...
-    from: address.toString(),
-    data: `0x${funcHash}`,
-    gasLimit: bigIntToHex(BigInt(530000)),
-    nonce: 1,
-  }
-  const storeTx = Transaction.fromTxData(storeTxData, { common, freeze: false })
-  storeTx.getSenderAddress = () => {
-    return address
+  req = params(method, [address.toString(), '0x0', 'latest'])
+  expectRes = (res: any) => {
```
ScottyPoi, line 28: currently this is returning empty slot value because of 'nonexistent account' as well.

ScottyPoi: I think "putAccount" assumes that it will be not be a contract.  

replace: `await execution.vm.stateManager.putAccount(...)` 
with:  `await execution.vm.stateManager.putContractCode(address, Uint8Array.from([]))
`

will fix this test and the next.
but should we change the way `putAccount` works instead?

ScottyPoi: hmm, i thought it was an issue with the test, but i cant seem to fix it

ScottyPoi:
The tests are passing now, after editing `eth_getStorageAt`

The method began by making a copy of `this._vm`

  Does anybody know why? 

It works if we just use `this._vm`, but not with the copy:

```
await this._vm.stateManager.accountExists(address)
// true
```

```
await vm.stateManager.accountExists(address)
// false
```

ScottyPoi:
i don't know why `miner.spec.ts` is failing...
it passes locally on my machine

acolytec3:
> The `getStorageAt.spec.ts` tests are passing now, after editing `eth_getStorageAt`
> 
> The method began by making a copy of `this._vm`
> 
>   Does anybody know why? 
> 
> It works if we just use `this._vm`, but not with the copy:
> 
> ```
> await this._vm.stateManager.accountExists(address)
> // true
> ```
> 
> ```
> const vm = await this._vm.copy()
> await vm.stateManager.accountExists(address)
> // false
> ```
> 
> 

I suspect because the copy didn't preserve the state of the cache. Under the older code, we didn't have a storage cache so it would still have all the same data.

ScottyPoi:
Notes from team call 4/19/23 -

* Need to use `vm.copy` to avoid doing anything that updates the actual VM
* `vm.copy` probably has a bug related to `statemanager.copy()`
* test coverage is insufficient for copy methods

TODO:
* revert back to using `this._vm.copy()` in `eth_getStorageAt`
* Debug `vm.copy()` and/or `statemanager.copy()`
* improve test coverage for both copy methods
* implement getStorageAt for `blockOpt`

ScottyPoi:
Introduced a simple, failing test to VM that exposes the vm.copy() bug.  This is a pretty glaring gap in the test coverage for vm, and I expect that i'll find the same gap in the state manager tests.  

so, beyond this PR, a full review of test coverage for these packages should be a goal

holgerd77:
Updated this via UI.

holgerd77:
> Decide if we want to leave it this way, or refactor the copy methods to include uncommitted changes

No, the copy() method should not include uncommitted changes, that would be overblown.

For the client context this is also enough, the eth_getStorage method shouldn't (can't) obviously be called during execution of a block, so state is guaranteed to be committed and flushed. So that's a good fix here, to just add this to the tests.

packages/vm/test/api/copy.spec.ts
```diff
@@ -0,0 +1,36 @@
+import { Account, Address } from '@ethereumjs/util'
```
acolytec3, line 1: Do we need this test script?  Don't we already have tests for `vm.copy`?

acolytec3 (commented):
LGTM except not clear on if we need the new `vm.copy` tests are not.  Feels duplicative.  But feel free to merge anyway.

acolytec3 (approved):
LGTM except not clear on if we need the new `vm.copy` tests are not.  Feels duplicative.  But feel free to merge anyway.

PR #2647, opened 2023-04-18
Statemanager: Storage Cache Unknown Slot Fix

holgerd77:
This fixes a bug where the state manager storage cache was wrongly marking a storage slot as known-to-be-empty when in fact it was just not known by the cache before at all.


holgerd77:
Can this be merged? This is definitely a bug, regardless of additional receiptTrie errors.

acolytec3 (approved):
I can confirm this fixed the identified bug. Great catch!

PR #2649, opened 2023-04-18
Remove EEI from EVM v2

jochem-brouwer:
Follow-up of #2604

The goal of this PR is to remove the EEI. All the logic (except blockchain logic) can be moved into StateManager.

This PR:

- Removes EEI
- Creates an interface in Common
- Setups a mock blockchain in EVM by default (blockchain in EVM only used for BLOCKHASH opcode: mockchain always returns `zeros(32)`)

Follow-up PRs;

1. Move all the touched/warm account journaling and friends into a journaler class
2. All touched/warm accounts logic is now removed (no separate journals). These are instead tracked from the account and storage cache

Another PR which will build on top of this but can be done at any place before above PRs;

- Ensure EVM runs `ethereum/tests` state tests (so remove `@ethereumjs/vm` dev dependency)

TODO in this PR:

- [x] Remove enableDefaultBlockchain logic
- [ ] Move EVM logic from DefaultStateManager into EVM
- [ ] Update StateManagerInterface such that it only exposes the "minimal" interface


packages/evm/src/evm.ts
```diff
@@ -673,7 +699,13 @@ export class EVM implements EVMInterface {
       versionedHashes: message.versionedHashes ?? [],
     }
 
-    const interpreter = new Interpreter(this, this.eei, env, message.gasLimit)
+    const interpreter = new Interpreter(
+      this,
+      this.stateManager,
```
acolytec3, line 704: Is it really necessary to pass `this.stateManager` and `this.blockchain` in this constructor since they are both properties of `this` (i.e. the EVM object)?

packages/statemanager/src/stateManager.ts
```diff
@@ -137,10 +152,36 @@ export class DefaultStateManager implements StateManager {
...
+  // It is possible that the storage slots set is empty. This means that the address is warm.
+  // It is not possible to have an accessed storage slot on a cold address (which is why this structure works)
+  // Each call level tracks their access themselves.
+  // In case of a commit, copy everything if the value does not exist, to the level above
+  // In case of a revert, discard any warm slots.
+  //
+  // TODO: Switch to diff based version similar to _touchedStack
+  // (_accessStorage representing the actual state, separate _accessedStorageStack dictionary
+  // tracking the access diffs per commit)
+  protected _accessedStorage: Map<string, Set<string>>[]
```
acolytec3, line 167: How much of (if any) this caching/journaling stuff is duplicating the various bits of caching that has built out in the recent caching work merged into v7?  I only ask because it feels like we're duplicating a lot of work here (though I do recognize that the `stateManager` is doing different work with the caching it does).  What I'm wondering if is we need a generalized caching mechanism in the `stateManager` when we are tracking the state changes in each call frame as a checkpoint through the journaling here so seems like there's an opportunity to streamline caching.

acolytec3 (commented):
Have just done an initial look at things based on the current state of the PR.  Looks generally pretty good!

packages/evm/src/evm.ts
jochem-brouwer: Interpreter does currently not have access to EVM currently IIRC (am on mobile)

packages/statemanager/src/stateManager.ts
jochem-brouwer: Yes I agree we should check if we can generalize it. Note that the PR is still super WIP. I am now at the phase where I can really start refactoring out things from the old statemanager.

acolytec3: Sounds good! Was just giving this an early look

jochem-brouwer: Yup, great and thanks :smile: :+1:

jochem-brouwer:
This is ready for a first round of reviews. Note that `enableDefaultBlockchain` flag will be removed (will do so ASAP).

jochem-brouwer:
Note: state manager tests fail, because EthersStateManager does not have the now newly required methods which are necessary in the state manager if you want to run it in EVM.

Shall we temporarily disable those tests..?

packages/evm/test/eips/eip-3860.spec.ts
```diff
@@ -132,8 +144,12 @@ tape('EIP 3860 tests', (t) => {
       hardfork: Hardfork.London,
       eips: [3860],
     })
-    const eei = await getEEI()
-    const evm = await EVM.create({ common, eei, allowUnlimitedInitCodeSize: true })
+    const evm = await EVM.create({
+      common,
+      stateManager: new DefaultStateManager(),
+
```
g11tech, line 150: extra new line?

packages/evm/src/evm.ts
g11tech: it seems to have as far the the Interpreter constructor goes, but the question is will there be a usecase where to pass a different statemanager than the evm's?

g11tech:
> Note: state manager tests fail, because EthersStateManager does not have the now newly required methods which are necessary in the state manager if you want to run it in EVM.
> 
> Shall we temporarily disable those tests..?

ha ha this was the exact question that was coming in my mind: what now happens to/implications are for `EthersStateManager`
since the EVM's intermittent state (previous vmState with checkpoints, warmed addresses etc) is now part of the state manager (which makes sense, but should the underlying `committed` data access layer be also carved out into a separate class, which can be plugged by `EthersStateManager`)

If we have an answer to how to best handle this, we can disable `EthersStateManager` to be refactored and re-activated in a followup PR.

jochem-brouwer:
Yup, so we can basically go two ways, 

1. Create a new BaseStateManager which handles this so we can just `super.method()` on both state managers (Holger does not like this since we just removed it)
2. Create a new class which handles this. The methods should call into this class. We can create some subclass like `StateManager.EVMJournal` / `EthersStateManager.EVMJournal` and call directly into that from EVM.

packages/evm/src/evm.ts
jochem-brouwer: Well, using EthersStateManager for instance :)

jochem-brouwer:
Wait, there is also a third option, which is moving the journal logic inside EVM. :thinking: Which upon thought might be the cleanest..? Thoughts?

g11tech:
> Wait, there is also a third option, which is moving the journal logic inside EVM. thinking Which upon thought might be the cleanest..? Thoughts?

I like this the best and to restrict state manager to dealing with `committed` data to provide us nice clean abstractions to do things like `EthersStateManager` (or `PortalNetworkStateManager`) etc

jochem-brouwer:
I think I agree, I don't know why I had not thought about this earlier. It makes more sense to move this EVM-related journal logic inside EVM and not outsource it to another package. When I find time I will move it into EVM 😄

packages/evm/src/evm.ts
holgerd77: I think you are missing the point here, the question was not about using another StateManager but a different one for EVM and StateManager, so if we need to pass both „this“ and „this.StateManager“, where I would also think the answer is clearly: not.

(„This“ is the EVM, just not to overlook)

I think for this.blockchain question goes as well.

packages/vm/src/runTx.ts
```diff
@@ -356,7 +346,7 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
     if (opts.skipBalance === true && fromAccount.balance < maxCost) {
       // if skipBalance, ensure caller balance is enough to run transaction
       fromAccount.balance = maxCost
-      await this.stateManager.putAccount(caller, fromAccount)
+      await this.stateManager.putAccount(caller, fromAccount, true)
```
holgerd77, line 349: Is there any false case here?

(so is the parameter necessary?)

holgerd77:
> I think I agree, I don't know why I had not thought about this earlier. It makes more sense to move this EVM-related journal logic inside EVM and not outsource it to another package. When I find time I will move it into EVM 😄

Can we take on this part a bit slow, from a performance POV it is relatively crucial that we get this right and I‘m still not 100% convinced that this could not just „fall out from the cache“, especially things like getOriginalContractStorage().

when I did some measurements these things were insanely slow and if we keep unnecessary double structures here we will never able to fully optimize this eventually.

Maybe rather something for next week, but eventually let’s do another call here.

packages/vm/src/runTx.ts
jochem-brouwer: `putAccount` defaults to `touched = false`. This fixes edge cases like: generating canonical accounts where empty accounts are put in trie (note that if we have empty account at address A, where this account exists in trie B and not in trie C, we get a different state root). In previous versions where `putAccount` on EEI was called, this would touch the account (and `generateCanonicalGenesis` called this). This would mark these empty accounts as touched. If you now execute a tx and you do not touch these accounts, they get wiped in the end anyways.

packages/evm/src/evm.ts
jochem-brouwer: I don't think I understand the question. The passed StateManager is part of EVM constructor, and Interpreter also needs access to it?

holgerd77: Yes, but we do not need to pass this extra on since we already pass `this` (the EVM) so we can do `evm.stateManager` or `evm.blockchain` inside `Interpreter` and so do not need three Interpreter constructor paramters for this, that's is the whole point here. 🙂

packages/vm/src/runTx.ts
holgerd77: Hmm, from an API standpoint this is not so beautiful to have this extra `touched` option which puts the burden to think about what to put there also for more generic StateManager users, or keep too different method signatures, which is also not so nice.

I would think that it's worth to think about if we can just track these (very very few I guess) exceptions EVM internally in an inside data-structure `_nonTouched` or so and then also handle internally? 🤔

holgerd77:
@jochem-brouwer Update: I will just write down my thoughts and design suggestions in a first round to discuss, this is just too much stuff and touching too many thing to "just do". 🙂 

Then we can discuss and eventually work on it together during the next week!

holgerd77:
Another update: ok, I am trippleing down step by step, this overloads me right now to be plainly honest. I can now also definitely say that moving to EVM is definitely the better option than the current status quo with this too overloaded StateManager interface. On the other hand I am seeing so many things which are just quite as the same as what I have done in the caches. Hmm.

I guess we are just coming from these very different angles right now, I am from my caching implementation experiences and additionally all these performance experiments, you being the expert on all these touched/warmed/EVM stuff.

I think it's likely most productive if we just do a longer call on this early next week (directly Monday or Tuesday?, both ok for me) than we can exchange the different perspectives and brainstorm a bit on this together.

Will lay this aside for now.

packages/vm/src/runTx.ts
jochem-brouwer: Yep, this is the general idea also thus promoting the idea of moving this EVM logic  which now resides in the StateManager (before this PR, it was inside EEI) back into EVM to track it there and not let StateManager track this EVM logic.

jochem-brouwer:
I agree that we come from different angles, but don't get me wrong, I care a lot about performance as well. I think the fact that we are both working on these things (so: me extracting EEI, and you working on the performance/caches) is a bit doing work which touches the same stuff and therefore we might both get a bit confused on what to do next.

Shall we do a call Monday afternoon? @holgerd77

holgerd77:
Hmm, since this will likely be a substantial round of additional work, I wonder: should we just merge this in "as is", eventually comment out 1-2 tests or so or fix some last stuff? 🤔

Then we can finally do this `develop-v7` merge and bring this a bit forward, otherwise I fear that we'll always loose this point again.

jochem-brouwer:
We can do this, I will move the cleanup work into another PR.

acolytec3 (approved):
LGTM

PR #2651, opened 2023-04-19
Statemanager: Expand Checkpointing Tests

holgerd77:
This PR expands the StateManager checkpointing test suites, both for account and for storage checkpointing, with some stronger emphasis on the storage tests.

Two test cases failing at the moment (one is a double failure due to 2 consecutive runs), one failure with directly reading a storage slot saved to cache, one other involving an empty/non-empty storage slot write.

Still needs investigation and fixes.


jochem-brouwer (approved):
LGTM

PR #2653, opened 2023-04-19
Statemanager: Fix Cache Commit Bug

holgerd77:
This fixes a bug in both caches (account and storage) where the diff cache was wrongly updated if the changed account/storage key was not present in the lower CP diff cache.


jochem-brouwer (approved):
Before this PR:

`ts-node cli.ts --maxPeers 0 --executeBlocks 721873-721874`

Fail: invalid receipt trie.

After this PR: passes.

Also passes on `721830-721874`

holgerd77:
Ok, in between update before I'll do a merge update here: all relevant tests passing 🎉, only 3 client tests failing (these base fee ones), seems something old, right?

acolytec3 (approved):
I'ma go ahead and very optimistically approve.  Hopefully we've licked it at this point!

packages/statemanager/test/ethersStateManager.spec.ts
```diff
@@ -182,7 +182,7 @@ tape('runTx custom transaction test', async (t) => {
     // The `MockProvider` is not able to load JSON files dynamically in browser so skipped in browser tests
     t.end()
   } else {
-    const common = new Common({ chain: Chain.Mainnet })
+    const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.London })
```
jochem-brouwer, line 185: Why do we set this to London?

holgerd77: Test was failing with the latest HF updates to Shanghai default, so I set the to something fixed (if I recall correctly it was something test setup related).

PR #2655, opened 2023-04-20
common: Make Shanghai the default hardfork

g11tech:
Shift the default hf to Shanghai 

(build on top of #2652 ), would be rebased once 2652 merges

PR #2657, opened 2023-04-21
common: Fix the block's fromValuesArray common refernces

g11tech:
While running devnet5 ethereumjs started crashing on restarts with the error on genesis init

image

because genesis was not shanghai, but being past shanghai time it, the blockchain's/dbmanager's common was shanghai which was being passed in opts, but along with hardforkByBlock opts which was correctly setting hardfork for the header, but failing for block hf validations as it was not using the new updated and properly set header's hardfork

This PR fixes the bug

pre-devent5 genesis
 - https://github.com/ethpandaops/4844-testnet/blob/skylenet/prep-devnet-5/network-configs/devnet-5/genesis.json#L17

This PR is based off on

- https://github.com/ethereumjs/ethereumjs-monorepo/pull/2655

and hence should be merged after the above merges

### DIFF
--- a/packages/block/src/block.ts
+++ b/packages/block/src/block.ts
@@ -152,19 +152,20 @@ export class Block {
     if (values.length > 4) {
       throw new Error('invalid block. More values than expected were received')
     }
+
+    // First try to load header so that we can use its common (in case of hardforkByBlockNumber being activated)
+    // to correctly make checks on the hardforks
+    const [headerData, txsData, uhsData, withdrawalBytes] = values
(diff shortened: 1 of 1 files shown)


g11tech:
> Hmm, not sure what happened, base changed (since I merged the other PR), but still all the Paris -> Shanghai changes displaying. thinking
> 
> Also: is this still in draft, conflicts a bit with ready-for-review?

yea, it was build on top of the shanghai default one, rebased it should be good now! also updated the label

jochem-brouwer (approved):
LGTM!

PR #2659, opened 2023-04-24
common: Rename shardingForkDev hf to Cancun

g11tech:
With Cancun hf shaping up nicely, this PR updates the references of the hf to `Cancun` (corresponding CL hf `Deneb`)

### DIFF
--- a/packages/common/src/chains/mainnet.json
+++ b/packages/common/src/chains/mainnet.json
@@ -104,6 +104,11 @@
       "block": null,
       "timestamp": "1681338455",
       "forkHash": "0xdce96c2d"
+    },
+    {
+      "name": "cancun",
+      "block": null,
+      "forkHash": null
     }
   ],
   "bootstrapNodes": [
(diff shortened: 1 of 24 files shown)


packages/common/src/chains/mainnet.json
```diff
@@ -104,6 +104,11 @@
       "block": null,
       "timestamp": "1681338455",
       "forkHash": "0xdce96c2d"
+    },
+    {
+      "name": "cancun",
+      "block": null,
+      "forkHash": null
```
holgerd77, line 111: I wonder if me might always want to take the sematics here:

If there is an entry on something with `null`, there is (will be) some value, we just don't know (yet) what it is, and completely leaving when there is just not a value.

This would be an argument to also add `"timestamp": null` for these HF entries (while having (obviously) left the `timestamp` entry for a HF like London).

I have absolutely no idea/overview, if this has practical implications or not, but it might help that we can do some things slightly cleaner already or in the future.

Totally no strong opinion here though, might help a bit, might be a total "non-thing", feel free to add or leave (if you e.g. do not want to re-run CI).

holgerd77 (approved):
I can already approve here independently from the one thing I mentioned.

If you want to still add I can directly re-approve, otherwise just merge in.

PR #2662, opened 2023-04-24
Trie: Refactor findPath() Core Retrieval Logic

holgerd77:
WIP

### DIFF
--- a/packages/trie/src/trie/trie.ts
+++ b/packages/trie/src/trie/trie.ts
@@ -144,10 +144,11 @@ export class Trie {
    * @returns A Promise that resolves to `Uint8Array` if a value was found or `null` if no value was found.
    */
   async get(key: Uint8Array, throwIfMissing = false): Promise<Uint8Array | null> {
-    const { node, remaining } = await this.findPath(this.appliedKey(key), throwIfMissing)
+    const { remaining, stack } = await this.findPath2(this.appliedKey(key), throwIfMissing)
     let value: Uint8Array | null = null
-    if (node && remaining.length === 0) {
-      value = node.value()
+    if (remaining.length === 0) {
+      const last = stack.pop()!
+      value = last.value()
     }
     return value
   }
@@ -311,6 +312,76 @@ export class Trie {
       // Resolve if walkTrie finishes without finding any nodes
       resolve({ node: null, remaining: [], stack })
     })
+  }
+
+  async processNode(nodeKey: Uint8Array, stack: TrieNode[], progress: Nibbles, remaining: Nibbles) {
+    const node = await this.lookupNode(nodeKey)
+
+    if (node === null) {
+      throw new Error('Path not found')
+    }
+    stack.push(node)
+    const current = remaining[0]!
+    progress.push(current)
+    remaining = remaining.slice(1)
+
+    if (node instanceof BranchNode) {
+      if (remaining.length === 0) {
+        // we exhausted the key without finding a node
+        return
+      } else {
+        const branchNode = node.getBranch(current)
+        if (!branchNode) {
+          // there are no more nodes to find and we didn't find the key
+          return
+        } else {
+          // node found, continuing search
+          // this can be optimized as this calls getBranch again.
+          //walkController.onlyBranchIndex(node, progress, branchIndex)
+          await this.processNode(branchNode as Uint8Array, stack, progress, remaining)
+        }
+      }
+    } else if (node instanceof LeafNode) {
+      if (doKeysMatch(remaining, node.key())) {
+        return
+      }
+    } else if (node instanceof ExtensionNode) {
+      const matchingLen = matchingNibbleLength(remaining, node.key())
+      if (matchingLen !== node.key().length) {
+        return
+      } else {
+        // keys match, continue search
+        const children = [[node.key(), node.value()]]
+        for (const child of children) {
+          const childRef = child[1] as Uint8Array
+          //this.pushNodeToQueue(childRef, childKey, priority)
+          await this.processNode(childRef, stack, progress, remaining)
+        }
+      }
+    }
+  }
+
+  /**
+   * Tries to find a path to the node for the given key.
+   * It returns a `stack` of nodes to the closest node.
+   * @param key - the search key
+   * @param throwIfMissing - if true, throws if any nodes are missing. Used for verifying proofs. (default: false)
+   */
+  async findPath2(key: Uint8Array, throwIfMissing = false): Promise<Path> {
+    const stack: TrieNode[] = []
+    const nodeKey = this.root()
+    const progress: Nibbles = []
+    const remaining = bytesToNibbles(key)
+    try {
+      await this.processNode(nodeKey, stack, progress, remaining)
+    } catch (error: any) {
+      if (error.message === 'Missing node in DB' && !throwIfMissing) {
+        // pass
+      } else {
+        throw error
+      }
+    }
+    return { node: null, stack, remaining }
   }
 
   /**
--- a/packages/trie/src/util/walkController.ts
+++ b/packages/trie/src/util/walkController.ts
@@ -66,7 +66,7 @@ export class WalkController {
    * @param node - Node to get all children of and call onNode on.
    * @param key - The current `key` which would yield the `node` when trying to get this node with a `get` operation.
    */
-  allChildren(node: TrieNode, key: Nibbles = []) {
+  allChildren(node: TrieNode, progress: Nibbles = []) {
     if (node instanceof LeafNode) {
       return
     }
@@ -82,7 +82,7 @@ export class WalkController {
     for (const child of children) {
       const keyExtension = child[0] as Nibbles
       const childRef = child[1] as Uint8Array
-      const childKey = key.concat(keyExtension)
+      const childKey = progress.concat(keyExtension)
       const priority = childKey.length
       this.pushNodeToQueue(childRef, childKey, priority)
     }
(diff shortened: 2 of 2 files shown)


ScottyPoi:
Got all but one of the Trie tests passing with the new findPath().

the failure is weird.  it doesn't do anything too different that the other tests around it, so i can't quite tell why it's failing

packages/trie/src/trie/trie.ts
```diff
@@ -376,16 +376,17 @@ export class Trie {
...
+      node = await this.processNode(nodeKey, stack, progress, remaining)
     } catch (error: any) {
       if (error.message === 'Missing node in DB' && !throwIfMissing) {
         // pass
       } else {
         throw error
       }
     }
-    return { node: null, stack, remaining }
+    return { node, stack, remaining }
```
holgerd77, line 386: At the end I thought that we might be able to skip this `node` return value since the node returned is just the last item on the stack (if the stack is complete), so this information is just redundant. I generally had the impression that this could be significantly simplified (so: the return format), not sure e.g., is `remaining` necessary or used? Might be though.

packages/trie/src/trie/trie.ts
```diff
@@ -935,8 +932,8 @@ export class Trie {
    * Returns a copy of the underlying trie.
    * @param includeCheckpoints - If true and during a checkpoint, the copy will contain the checkpointing metadata and will use the same scratch as underlying db.
    */
-  copy(includeCheckpoints = true): Trie {
-    const trie = new Trie({
+  async copy(includeCheckpoints = true): Promise<Trie> {
```
holgerd77, line 935: Note: if we keep this we need to mention in the docs (breaking).

Good cath though, we might want to re-visit the other `copy()` methods as well.

holgerd77:
> Got all but one of the Trie tests passing with the new findPath().
> 
> the failure is weird. it doesn't do anything too different that the other tests around it, so i can't quite tell why it's failing

🎉

Something is off with CI though, would be great if you can fix as some first next step, so that one gets an overview of the tests still failing.

packages/trie/src/trie/trie.ts
ScottyPoi: i'll dig into it a little bit more, to see if it's actually necessary for it to be async.  

we encourage users to use the `await Trie.create()` method, but many of our tests are written with `new Trie()`.  So yes, either a bunch of rewrites to make it al async, or maybe i will find that the create method doesn't actually need to be async.

packages/trie/src/trie/trie.ts
ScottyPoi: you're probably right.

ScottyPoi:
> Note: if we keep this we need to mention in the docs (breaking).

The issues here run deep, and while I can do my best to limit this PR to non-breaking changes, we may find it impossible to fix one issue without also fixing a few others, which makes breaking changes all the more likely.  

one helpful non-breaking change i'd love to make right away would be to give the Trie class a `debugger`.  does it not have one be design?  like as a performance choice or something?

packages/trie/src/trie/trie.ts
holgerd77: No, the `create()` method needs to be async, that's the whole point why we introduced this method, to allow for async trie instantiation when using e.g. LevelDB as a database which has async reads and writes.

So it is rather our laziness that we still use (our own discouraged) way of using `new Trie()` directly and we should optimally rewrite the tests (goes for some other libraries as well, I guess at least VM and Blockchain? 🤔)

packages/trie/src/trie/trie.ts
```diff
@@ -942,8 +942,8 @@ export class Trie {
    * Returns a copy of the underlying trie.
    * @param includeCheckpoints - If true and during a checkpoint, the copy will contain the checkpointing metadata and will use the same scratch as underlying db.
    */
-  async copy(includeCheckpoints = true): Promise<Trie> {
-    const trie = await Trie.create({
+  copy(includeCheckpoints = true): Trie {
+    const trie = new Trie({
```
holgerd77: Ah, please, as mentioned, do not revert this. This was a great change! 🤩

ScottyPoi: oh -- sure.  even though it makes this PR breaking?

ScottyPoi: ah, just saw your comment above

holgerd77:
Just a general note: when I was starting with this, I was not yet 100% sure if "this is the way to go", this was just started as some (promising, attention: personal judgement 😋) experiment.

If we do it this way, this would throw away a lot of logic, in particular this job queuing thing with this PrioritizedTaskExecutor - or however this is named.

I am currently not seeing why such a task executor is needed, from my understanding Trie node retrieval seems to be a totally deterministic process with no surprises (this `findPath` name (and other semantics from wording as well) is sometimes putting one on the path thinking that there is some "oh let's look here, no nothing, let's take the other route, ah, nothing as well" process involved. 😜 That's completely not true unless I am missing something.

So anyhow: to confirm that we are not the right track here it would be good if you do substantial mainnet sync trials with the PR on the sideline, just let this running a bit, see how it goes, if the client goes out of memory at some point e.g. with these changes (there is some mention of OOM prevention in "the old code", I am personally not seeing yet where this should happen with max 64 (?) subsequent DB reads).

For the performance perspective it would be good if you take some couple of 100/1000 tx loaded blocks and take this version and then the old one and see how sync times compare (optimally post here).

packages/trie/src/trie/trie.ts
holgerd77: Yes, now it's the time to make these kind of breaking changes if they are useful. 🙂

jochem-brouwer:
I do remember this prioritized task executor and I tried to remove it at some point (or at least how I remember is by  changing it to an optimized version with binary sort). However this had a side effect why it is still this task executor instead of the binary search one. We should maybe find out what the original motivation for this task executor.

holgerd77:
> I do remember this prioritized task executor and I tried to remove it at some point (or at least how I remember is by changing it to an optimized version with binary sort). However this had a side effect why it is still this task executor instead of the binary search one. We should maybe find out what the original motivation for this task executor.

Hmm, yeah, this thing is really "interesting". 🤔 🙂

I injected a console log for `findPath` start and then for `priority` in `pushNodeToQueue(nodeRef: Uint8Array, key: Nibbles = [], priority?: number)` along block execution in a client run and this looks like this:

```shell
findPath
1
2
3
4
5
6
findPath
1
2
3
4
findPath
1
2
3
4
5
findPath
1
findPath
1
2
3
4
5
findPath
1
2
3
4
findPath
1
2
3
4
5
findPath
```

So totally linear and simple (and low number, will go a bit higher when in state has gronw, this is in block ~500.000 or so, anyhow).

So this would keep me on the track that this is not doing so much useful.

One theory would be that this was mitigating some flaws of the complicated Promise structrue before (things couldn't execute or something?). But just a theory, does not necessarily need to be true.

I guess we just need to observe this a bit, client sync is likely a good testbed here, solid amount of state + real world scenario. Generally of course also: the higher (block number), the better (on the other hand: testbed before was likely not so ambitious? 🤔)

jochem-brouwer:
I agree on the above ^ comment @holgerd77. My gut also says that this prioritized queue thing is a leftover of a very old implementation (it would be somewhat likely this indeed has to do with the old callback structure) and that we can safely remove it, but for completeness just wanted to mention my old experiences. I think if we just let the client run for a while and it does not hiccup we are safe.

ScottyPoi:
I am so glad to hear all of this!  @jochem-brouwer and @holgerd77 

The git history on the Trie library is a little vague on exactly who wrote what parts of it, and I don't want to offend any current team members by suggesting big changes.  

I totally agree that much of the weird complexity to this library can be stripped away, it just requires some finesse, and some basic fundamental changes.  

I'm feeling empowered to tear this library apart, and remake it as a more user friendly and "general purpose" set of tools.  Should I treat this PR as a vehicle for all of that? 

 Or should I try to make granular incremental changes in separate PR's?

jochem-brouwer:
I think I wrote/rewrote most of the trie library, go ahead refactoring it :) 

If you edit a single method, I think it is great if this is a single PR. However if the refactor is "all over the place" this should be self contained in a single PR also, IMO.

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -189,9 +189,9 @@ export class EthersStateManager implements StateManager {
     const proof = await this.provider.send('eth_getProof', [address.toString(), [], this.blockTag])
 
     const proofBuf = proof.accountProof.map((proofNode: string) => toBytes(proofNode))
-
-    const trie = new Trie({ useKeyHashing: true })
-    const verified = await trie.verifyProof(keccak256(proofBuf[0]), address.bytes, proofBuf)
+    const veritrie = await Trie.fromProof(keccak256(proofBuf[0]), proofBuf)
```
holgerd77, line 192: Ah, that's a nice new API! 🙂 👍 

(not sure if final but at least going into a good direction, always looked wrong/out-of-place to me with the explicit Trie instantiation)

packages/statemanager/test/stateManager.code.spec.ts
```diff
@@ -140,18 +139,18 @@ tape('StateManager -> Code', (t) => {
...
+    //   })
+    //   const address = new Address(hexStringToBytes('a94f5374fce5edbc8e2a8697c15331677e6ebf0b'))
+    //   const code = hexStringToBytes('80')
+    //   try {
+    //     await stateManager.putContractCode(address, code)
+    //     st.fail('should throw')
+    //   } catch (e) {
+    //     st.pass('successfully threw')
+    //   }
+    // })
```
holgerd77, line 154: Same here, please use `/* */` commenting out.

holgerd77 (commented):
Hi Scotty,
had a first look on this, some first-round comments! 🙂 

This is an impressive amount of new code! 😋

Can you please prioritize branch updates on this, also try to get as much tests as possible (also and in particular non-Trie related) working and continue to do so, so that it gets more transparent where the PR is atm?

Please let us know if you need any help on branch updates or the like, often people responsible for the respective PR can assist a bit.

Ugh. Really excited to see what will come out of this!

What a change set! 😜 😆

PR #2663, opened 2023-04-24
Migrate `fromEthersProvider` to `fromJsonRpcProvider` constructors

acolytec3:
This PR, when complete, will replace the `fromEthersProvider` static constructors in `block` and `tx` with a more generalized `fromJsonRpcProvider` constructor that will take the URL string for a JSON-RPC provider and then fetch the corresponding block or tx from the provider using native `fetch` or `http` request and will also remove the` micro-ftch` dependency which requires several polyfills in a browser/webpack context (i.e. `zlib-browserify`,`https-browserify`, and `stream-http`).

- [x] Rewrite `fetchFromProvider` to use native `fetch` or `http` calls instead of `micro-ftch`
- [x] Replace `fromEthersProvider` with `fromJsonRpcProvider` in `block` and `tx`
- [x] Clean up various `testdouble` typing items across tests

### DIFF
--- a/packages/block/README.md
+++ b/packages/block/README.md
@@ -31,7 +31,7 @@ There are five static factories to instantiate a `Block`:
 - `Block.fromRLPSerializedBlock(serialized: Buffer, opts?: BlockOptions)`
 - `Block.fromValuesArray(values: BlockBuffer, opts?: BlockOptions)`
 - `Block.fromRPC(blockData: JsonRpcBlock, uncles?: any[], opts?: BlockOptions)`
-- `Block.fromEthersProvider(provider: ethers.providers.JsonRpcProvider | string, blockTag: string | bigint, opts: BlockOptions)`
+- `Block.fromJsonRpcProvider(provider: ethers.providers.JsonRpcProvider | string, blockTag: string | bigint, opts: BlockOptions)`
 
 For `BlockHeader` instantiation analog factory methods exists, see API docs linked below.
 
--- a/packages/block/README.md
+++ b/packages/block/README.md
@@ -31,7 +31,7 @@ There are five static factories to instantiate a `Block`:
 - `Block.fromRLPSerializedBlock(serialized: Buffer, opts?: BlockOptions)`
 - `Block.fromValuesArray(values: BlockBuffer, opts?: BlockOptions)`
 - `Block.fromRPC(blockData: JsonRpcBlock, uncles?: any[], opts?: BlockOptions)`
-- `Block.fromJsonRpcProvider(provider: ethers.providers.JsonRpcProvider | string, blockTag: string | bigint, opts: BlockOptions)`
+- `Block.fromJsonRpcProvider(provider: string | EthersProvider, blockTag: string | bigint, opts: BlockOptions)`
 
 For `BlockHeader` instantiation analog factory methods exists, see API docs linked below.
 
--- a/packages/block/src/block.ts
+++ b/packages/block/src/block.ts
@@ -8,12 +8,13 @@ import {
   bigIntToHex,
   bytesToHex,
   equalsBytes,
+  fetchFromProvider,
+  getProvider,
   intToHex,
   isHexPrefixed,
   ssz,
 } from '@ethereumjs/util'
 import { keccak256 } from 'ethereum-cryptography/keccak'
-import { ethers } from 'ethers'
 
 import { blockFromRpc } from './from-rpc'
 import { BlockHeader } from './header'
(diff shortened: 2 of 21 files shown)


holgerd77:
Great, there is also still an Ethers dependency in VM/EVM, maybe you can remove along.

paulmillr:
micro-ftch is indeed experimental and its usage is not encouraged.

However, `https` module import feels like bundlers will be shimming the module with polyfills in browsers?

For a context:

node.js v16 is getting ditched in september 2023 (https://endoflife.date/nodejs). Probably around the time of monorepo v7 release. node.js v18 and v20 both have native `fetch`, just like the one in browsers, as a global variable. Maybe it's worth targetting node.js v18 for fetch and other great features?

acolytec3:
> micro-ftch is indeed experimental and its usage is not encouraged.
> 
> However, `https` module import feels like bundlers will be shimming the module with polyfills in browsers?
> 
> For a context:
> 
> node.js v16 is getting ditched in september 2023 (https://endoflife.date/nodejs). Probably around the time of monorepo v7 release. node.js v18 and v20 both have native `fetch`, just like the one in browsers, as a global variable. Maybe it's worth targetting node.js v18 for fetch and other great features?

I'm happy to go either way on the `https` import (especially given the above about Node 16 being close to EOL, though ideally bundlers would just alias it to false since we already check for global fetch first and only fall back to to the native https call if it it's not present.  Either case is better than the current variant with the several polyfills it requires in browser for `micro-ftch` (which itself is still probably better than the whole ethers dependency tree we were importring before as far as our bundle size goes).

paulmillr:
> ideally bundlers would just alias it to false

ideally yes, in practice users would run into all kinds of issues. For example, noble libs have been using `2n ** 3n` exponentiation operator, and many users don't bother setting proper browser target, which leads to their bundlers transpiling it to `Math.pow(2n, 3n)` - which is an error. I'm just guessing something similar could happen here.

Moving swiftly with regards to node.js version upgrade will allow monorepo to lay out a great strategy for new features. Some time in the future, when v18 is dropped, v20+ will allow using pure-esm modules without bundlers - which is huge. Imagine writing `import {strip0x} from "https://ethereum.org/js/util.js"` (or from a local alternative) - not possible right now.

acolytec3:
> > ideally bundlers would just alias it to false
> 
> ideally yes, in practice users would run into all kinds of issues. For example, noble libs have been using `2n ** 3n` exponentiation operator, and many users don't bother setting proper browser target, which leads to their bundlers transpiling it to `Math.pow(2n, 3n)` - which is an error. I'm just guessing something similar could happen here.
> 
> Moving swiftly with regards to node.js version upgrade will allow monorepo to lay out a great strategy for new features. Some time in the future, when v18 is dropped, v20+ will allow using pure-esm modules without bundlers - which is huge. Imagine writing `import {strip0x} from "https://ethereum.org/js/util.js"` (or from a local alternative) - not possible right now.

Yup, definitely things for consideration down the line.  Probably the best we can do for now is to just try and provide good docs on how to use our stuff in browser.  Our goal with the next set of breaking releases is to minimize the amount of polyfill/aliasing required to use our libraries in browser.  Unfortunately, they're still pretty clunky at the moment, though getting rid of buffers will definitely help.  My goal is that maybe by 2024, someone can do `create-react-app` and drop our code directly into it.

packages/block/README.md
```diff
@@ -31,7 +31,7 @@ There are five static factories to instantiate a `Block`:
 - `Block.fromRLPSerializedBlock(serialized: Buffer, opts?: BlockOptions)`
 - `Block.fromValuesArray(values: BlockBuffer, opts?: BlockOptions)`
 - `Block.fromRPC(blockData: JsonRpcBlock, uncles?: any[], opts?: BlockOptions)`
-- `Block.fromEthersProvider(provider: ethers.providers.JsonRpcProvider | string, blockTag: string | bigint, opts: BlockOptions)`
+- `Block.fromJsonRpcProvider(provider: ethers.providers.JsonRpcProvider | string, blockTag: string | bigint, opts: BlockOptions)`
```
jochem-brouwer, line 34: By the current method signature of `Block`, this is an `any` and could thus be something else than `ethers.providers.JsonRpcProvider`

packages/block/src/block.ts
```diff
@@ -223,44 +224,57 @@ export class Block {
...
-      blockData = await prov.send('eth_getBlockByNumber', [bigIntToHex(blockTag), true])
+      blockData = await fetchFromProvider(providerUrl, {
+        method: 'eth_getBlockByNumber',
+        params: [bigIntToHex(blockTag), true],
+      })
     } else if (
       isHexPrefixed(blockTag) ||
       blockTag === 'latest' ||
       blockTag === 'earliest' ||
       blockTag === 'pending'
```
jochem-brouwer, line 255: Beyond the scope of this PR but maybe we should check if `blockTag` is of any kind of whatever valid string values there are? I think `syncing` is a new param you can use as blockTag as of, I think, the merge?

packages/block/test/mockProvider.ts
jochem-brouwer, line 1: Does this file actually do anything?

packages/tx/README.md
```diff
@@ -282,7 +282,7 @@ The correct tx type class for instantiation will then be chosen on runtime based
 - `public static fromTxData(txData: TxData | AccessListEIP2930TxData, txOptions: TxOptions = {}): TypedTransaction`
 - `public static fromSerializedData(data: Buffer, txOptions: TxOptions = {}): TypedTransaction`
 - `public static fromBlockBodyData(data: Buffer | Buffer[], txOptions: TxOptions = {})`
-- `public static async fromEthersProvider(provider: string | ethers.providers.JsonRpcProvider, txHash: string, txOptions?: TxOptions)`
+- `public static async fromJsonRpcProvider(provider: string | ethers.providers.JsonRpcProvider, txHash: string, txOptions?: TxOptions)`
```
jochem-brouwer, line 285: Same here with type; `provider` type is currently `any`

packages/tx/test/mockProvider.ts
jochem-brouwer, line 1: Same here does this still do anything?

jochem-brouwer (commented):
Some notes and qs

packages/block/src/block.ts
```diff
@@ -223,44 +224,57 @@ export class Block {
...
+   *  Method to retrieve a block from a JSON-RPC provider and format as a {@link Block}
+   * @param provider either a url for a remote provider or an Ethers JsonRpcProvider object
    * @param blockTag block hash or block number to be run
    * @param opts {@link BlockOptions}
    * @returns the block specified by `blockTag`
    */
-  public static fromEthersProvider = async (
-    provider: ethers.providers.JsonRpcProvider | string,
+  public static fromJsonRpcProvider = async (
+    provider: any,
```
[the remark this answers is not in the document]
acolytec3: Will do.  That's a good call.

packages/block/README.md
acolytec3: Good catch.  I've added an `EthersProvider` interface in `util` that addresses this.

packages/statemanager/test/ethersStateManager.spec.ts
```diff
@@ -224,8 +224,8 @@ tape('runTx test: replay mainnet transactions', async (t) => {
 
     const blockTag = 15496077n
     common.setHardforkByBlockNumber(blockTag)
-    const txHash = '0xed1960aa7d0d7b567c946d94331dddb37a1c67f51f30bf51f256ea40db88cfb0'
-    const tx = await TransactionFactory.fromEthersProvider(provider, txHash, { common })
+    const txData = require('./testdata/providerData/transactions/0xed1960aa7d0d7b567c946d94331dddb37a1c67f51f30bf51f256ea40db88cfb0.json')
```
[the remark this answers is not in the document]
acolytec3: These tests aren't focused on testing the `fetchFromProvider` calls in `util` and it was proving tricky to mock just the readonly `connection.url` property of the `ethers.JsonRpcProvider` interface.  We don't gain anything by testing the block/tx static constructors here as that coverage is elsewhere.  If you feel strongly, I can bring it back in but it will take some further messing about to get the `mockprovider` to work correctly on these tests if I do.

acolytec3 (commented):
Thanks for the feedback, @jochem-brouwer.  I think I've completed the updates so ready for another round of review.

packages/block/src/block.ts
g11tech: ummm what would `syncing` point to? since merge we can now use `finalized` and `safe`, we can add them
```suggestion
      blockTag === 'pending' ||
      blockTag === 'finalized' ||
      blockTag === 'safe'
```

g11tech (commented):
looks good added  a suggestion

packages/block/src/block.ts
jochem-brouwer: Sorry, I meant `safe`.

jochem-brouwer (commented):
If the CI is fixed and @g11tech 's suggestion is included then I can approve :smile:

jochem-brouwer (approved):
Pushed fix-CI commit, will pre-approve :smile:

PR #2665, opened 2023-04-25
Use JS maps in cache

acolytec3:
Replace `OrderedMaps` in account and storage `diffCache` with pure JS maps.  Appears to have a small performance gain (as well as readability improvements).

Can use a script like this to validate performance:
```
import { Address } from '@ethereumjs/util'
import { randomBytes } from 'crypto'

import { createAccount } from '../../test/util'

import { AccountCache } from './account'
import { CacheType } from './types'

const samples = 10000
const accountCache = () => {
  const cache = new AccountCache({ type: CacheType.LRU, size: 10000 })
  for (let x = 0; x < samples; x++) {
    const address = Address.fromPrivateKey(randomBytes(32))
    const account = createAccount(undefined, 1n)
    cache.get(address)
    cache.put(address, account)
    if ((x ^ 10) === 0) {
      cache.checkpoint()
    }
    if ((x ^ 20) === 0) {
      cache.commit()
    }
  }
}

for (let x = 0; x < 10; x++) {
  console.time('accountcache')
  accountCache()
  console.timeEnd('accountcache')
}
```

### DIFF
--- a/packages/statemanager/src/cache/account.ts
+++ b/packages/statemanager/src/cache/account.ts
@@ -33,8 +33,7 @@ export class AccountCache extends Cache {
    * to the account), the element didn't exist in the cache
    * before.
    */
-  _diffCache: OrderedMap<string, AccountCacheElement | undefined>[] = []
-
+  _diffCache: Map<string, AccountCacheElement | undefined>[] = []
   constructor(opts: CacheOpts) {
     super()
     if (opts.type === CacheType.LRU) {
@@ -46,20 +45,20 @@ export class AccountCache extends Cache {
       this._orderedMapCache = new OrderedMap()
     }
 
-    this._diffCache.push(new OrderedMap())
+    this._diffCache.push(new Map<string, AccountCacheElement | undefined>())
     this._debug = createDebugLogger('statemanager:cache:account')
   }
 
   _saveCachePreState(cacheKeyHex: string) {
-    const it = this._diffCache[this._checkpoints].find(cacheKeyHex)
-    if (it.equals(this._diffCache[this._checkpoints].end())) {
+    const it = this._diffCache[this._checkpoints].get(cacheKeyHex)
+    if (it === undefined) {
       let oldElem: AccountCacheElement | undefined
       if (this._lruCache) {
         oldElem = this._lruCache!.get(cacheKeyHex)
       } else {
         oldElem = this._orderedMapCache!.getElementByKey(cacheKeyHex)
       }
-      this._diffCache[this._checkpoints].setElement(cacheKeyHex, oldElem)
+      this._diffCache[this._checkpoints].set(cacheKeyHex, oldElem)
     }
   }
 
@@ -142,12 +141,11 @@ export class AccountCache extends Cache {
     }
 
     const diffMap = this._diffCache[this._checkpoints]!
-    const it = diffMap.begin()
 
     const items: [string, AccountCacheElement][] = []
 
-    while (!it.equals(diffMap.end())) {
-      const cacheKeyHex = it.pointer[0]
+    for (const entry of diffMap.entries()) {
+      const cacheKeyHex = entry[0]
       let elem: AccountCacheElement | undefined
       if (this._lruCache) {
         elem = this._lruCache!.get(cacheKeyHex)
@@ -158,9 +156,8 @@ export class AccountCache extends Cache {
       if (elem !== undefined) {
         items.push([cacheKeyHex, elem])
       }
-      it.next()
     }
-    this._diffCache[this._checkpoints] = new OrderedMap()
+    this._diffCache[this._checkpoints] = new Map<string, AccountCacheElement | undefined>()
     return items
   }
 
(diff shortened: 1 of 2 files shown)


acolytec3:
Performance tests using the above script:
```
<----JS Maps for diffCache
accountcache: 3.343s
accountcache: 3.283s
accountcache: 3.255s
accountcache: 3.248s
accountcache: 3.287s
accountcache: 3.218s
accountcache: 3.232s
accountcache: 3.246s
accountcache: 3.389s
accountcache: 3.255s

<----- OrderedMap for diffCache
accountcache: 3.440s
accountcache: 3.354s
accountcache: 3.371s
accountcache: 3.347s
accountcache: 3.417s
accountcache: 3.259s
accountcache: 3.332s
accountcache: 3.327s
accountcache: 3.382s
accountcache: 3.323s
```

holgerd77:
Updated this branch.

Great, looks generally good! 👍 🙂 

For these kind of cache changes it is always good to run the first ~1 Mio mainnet blocks (things starting to get interesting at around block 700.000 or so, so first 200.000 or so just won't do it), to see both:

1. If there are block spanning consensus errors due to permanent cache and
2. if client goes OOM

I had a lot of these cases when doing these kind of changes.

Can you please trigger such a run on the sideline and run for 2 hours or so and give some short feedback? Then I am very happy to merge here!

(side note: not sure maybe it was just me who didn't know but I found out during this week that one can very well continue to work with the code, switching branches,... when doing such runs since Node is one-time-loading all code parts into memory at start which helped me a lot in these kind of situations. 😋 (one can of course not (so easily) do parallel client runs during that time though).

holgerd77:
Not sure, didn't have a closer look: can we remove the `js-sdsl` package from the dependencies after this refactor?

jochem-brouwer:
LGTM, have triggered a CI run with all hardforks though to see if we are OK (most likely yes)

holgerd77:
> LGTM, have triggered a CI run with all hardforks though to see if we are OK (most likely yes)

Great idea, note that this cannot replace the long-run-sync test though.

acolytec3:
> Not sure, didn't have a closer look: can we remove the `js-sdsl` package from the dependencies after this refactor? 

Unless we remove the `OrderedMap` cache type entirely, that package needs to remain.  My code here was purely to remove the use of `OrderedMaps` in the "diffCache" (aka the checkpoint journal).  I'm not sure if there's value in maintaining multiple cache type options or not over the longer term since the LRU cache seems be more performant but I'm not sure I'm a good judge of that at present.

holgerd77:
> > Not sure, didn't have a closer look: can we remove the `js-sdsl` package from the dependencies after this refactor?
> 
> Unless we remove the `OrderedMap` cache type entirely, that package needs to remain. My code here was purely to remove the use of `OrderedMaps` in the "diffCache" (aka the checkpoint journal). I'm not sure if there's value in maintaining multiple cache type options or not over the longer term since the LRU cache seems be more performant but I'm not sure I'm a good judge of that at present.

Ah ok, thanks.

Ah, no, ORDERED_MAP (or at least: non-LRU) cache is needed, for LRU memory is pre-allocated and so not suitable for short-lived scenarios. E.g. the VM blockchain tests go out of memory if LRU cache is used.

Also there is this thing that - if the LRU cache is set too small - and usage is bigger than size within the scope before things are flushed, LRU cache is not working. We had this case (also discovered in VM tests) where LRU cache was set to something under 10.000 and things were breaking already in genesis block, since 10.000+ accounts from genesis were added to the cache with some already falling out again. With ORDERED_MAP these things cannot happen.

(so LRU is good for long-lived and large-cache-size scenarios)

jochem-brouwer:
image

We can use this method to test this?

holgerd77:
> Regarding VM breaking if the LRU cache is too small:
> 
> What was the reason again why it breaks? I have somewhat a hard time to imagine why it would break. If an item falls out of the cache, we SHOULD dump that value into the non-cache (so either disk or level-mem whatever is used for DB). Then upon lookup, what happens?
> 
>     * Is it in cache? No
> 
>     * Read it from DB (should have the cached value)
> 
>     * Done
> 
> 
> This should work, right? I think if people start using LRUCache without knowing the side effects people might get really confused why suddenly results start to diverge..? 🤔

This has structural problems on various ends, I did think this through to some level but came to the conclusion that it's not worth it.

The first case is this - extreme - version that the cache is too small for items handled during one before-flush-period (a transaction). At the moment the cache does not know about the database any more. This was a design goal, before we had this strange callback-passing in structure (which I also introduced), which made things too coupled and the code hard to read. If we now try to solve this (very rare, only if one specific cache is too small) case "what happens if the cache is running out already in this period" problem, we would basically need to structurally re-introduce this two-way-binding respectively do large restructures again, from my PoV not for the better. The other alternative would be to introduce some additional backup structures within the cache itself (keep this thrown out stuff for this before-flush() period. I thought about that, but this is more complex (due to checkpointing) than one would think in first round and would introduce a lot of extra complexity.

Second, a bit more general thing: if we use this `dispose()` method, this would very likey put a large performance burden on the cache, which would be unfortunate. Furthermore: the cache will run full "by-design", since this is an LRU cache. So, when the cache is full, we would then (very very unnecessarily) re-write long commited state again to Trie, totally unnecessarily and again introducing a huge performance overhead. Or there will be again the need for - likely ugly - additional tracking structures.

So, I would very likely think: we should keep it this way. 🙂 

In the case people are considering switching the cache they will have likely given this some explicit thought and have their reasoning for it, the main/only good reason here is actually to use a larger and more permanent cache (where the problem doesn't exist). There is also a very good chance that people will *then* also have read to code docs on the LRU cache (to read up on the benefits), where this theoretical problem is clearly described.

So also from a practical PoV I am in good spirits that this will be not that much of a problem. 😀

holgerd77:
Have done various edits to the post above, please read on site. 😄

PR #2666, opened 2023-04-25
Fix handling of nested uint8Arrays in JSON in DB

acolytec3:
Switches the handling of uint8Arrays nested inside JSON objects put in the DB to store as hex strings since LevelDB incorrectly converts these uint8Arrays to nested JSON objects and reassembling the Uint8Array is slower than converting to and from hex strings.

### DIFF
--- a/packages/blockchain/src/blockchain.ts
+++ b/packages/blockchain/src/blockchain.ts
@@ -1211,8 +1211,14 @@ export class Blockchain implements BlockchainInterface {
    * @hidden
    */
   private _saveHeadOps(): DBOp[] {
+    // Convert DB heads to hex strings for efficient storage in DB
+    // LevelDB doesn't handle Uint8Arrays properly when they are part
+    // of a JSON object being stored as a value in the DB
+    const hexHeads = Object.fromEntries(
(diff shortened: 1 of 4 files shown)


acolytec3:
```
import { bytesToHex } from '@ethereumjs/util'
import { Level } from 'level'

const main = async () => {ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss
    const db = new Level<string | Uint8Array, any>(process.argv[3])
    await db.open()
    const heads = JSON.parse(await db.get('heads'))
    console.log(heads)
    for (const key of Object.keys(heads)) {
        heads[key] = Uint8Array.from(Object.values(heads[key])) as any
    }
    const newHeads = Object.fromEntries(Object.entries(heads).map((el) => [el[0], bytesToHex(el[1] as Uint8Array)]))
    console.log(newHeads)
    try {
        await db.put('heads', JSON.stringify(newHeads))
    } catch (err) {
        console.log(err)
    }

}
void main()
```

If you run this script and pass in a chain datadir, it will convert a `develop-v7` branch based chain DB `heads` value to the new hex string based approach.

g11tech (approved):
lgtm!

PR #2667, opened 2023-04-26
Trie/Client: Add permanent Trie Node Cache + Client Integration

holgerd77:
This PR adds a permanent Node cache to the Trie checkpoint DB class which increases client block execution in particular and repetitive trie reads in general.

Cache design is extremely simple since trie node keys are unique, so there is no need for this back-and-forth related to checkpointing, since there is no risk to overwrite and old entry and therefore accidentally introduce a wrong key. I tested this with longer client sync ranges and there were no problems.

It is also "by design" that reverted node entries are kept in the cache, since this might help to mitigate repeated revert-attacks with the same attack setup, always re-reading the same accounts/slots e.g. and then revert.

I in-between planned to activate this LRU cache by default, since in most cases there should be no downsides. I then discovered several ad-hoc Trie instantiations in our own code base (mostly related to proofs), so I restrained since there might be side effects if regarding too frequent instantiations.

(Side question: do these Proof methods need any Trie object state or might it generally be an option (respectively occasion now during breaking period) to make these methods static?)

Another note: I first added an active cache to the Trie checkpointing tests and these were actually failing. I then had a closer look and realized that these were just creating "un-realistic" trie scenarios - by directly operating on the db and then adding a key and updating the same key with a different value. This scenario can - by Merkle Patricia Tree design - just not happen, since keys are always unique (in relation to the content/subtree they refer to). I already thought if these tests should generally be deleted, but refrained since they do test the general DB functionality nevertheless. All other tests (which I ran in between with the Trie-cache-active-by-default work state) are passing with the cache though.

Open for review and merge.

### DIFF
--- a/packages/trie/src/db/checkpoint.ts
+++ b/packages/trie/src/db/checkpoint.ts
@@ -100,7 +100,7 @@ export class CheckpointDB implements DB {
       }
       await this.batch(batchOp)
     } else {
-      // dump everything into the current (higher level) cache
+      // dump everything into the current (higher level) diff cache
       const currentKeyValueMap = this.checkpoints[this.checkpoints.length - 1].keyValueMap
       for (const [key, value] of keyValueMap.entries()) {
         currentKeyValueMap.set(key, value)
@@ -130,15 +130,14 @@ export class CheckpointDB implements DB {
       }
     }
 
-    // Lookup the value in our cache. We return the latest checkpointed value (which should be the value on disk)
+    // Lookup the value in our diff cache. We return the latest checkpointed value (which should be the value on disk)
     for (let index = this.checkpoints.length - 1; index >= 0; index--) {
       const value = this.checkpoints[index].keyValueMap.get(bytesToHex(key))
       if (value !== undefined) {
         return value
       }
     }
-    // Nothing has been found in cache, look up from disk
-
+    // Nothing has been found in diff cache, look up from disk
     const value = await this.db.get(key)
     this._stats.db.reads += 1
     if (value !== null) {
@@ -146,7 +145,8 @@ export class CheckpointDB implements DB {
     }
     this._cache?.set(keyHex, value)
     if (this.hasCheckpoints()) {
-      // Since we are a checkpoint, put this value in cache, so future `get` calls will not look the key up again from disk.
+      // Since we are a checkpoint, put this value in diff cache,
+      // so future `get` calls will not look the key up again from disk.
       this.checkpoints[this.checkpoints.length - 1].keyValueMap.set(bytesToHex(key), value)
     }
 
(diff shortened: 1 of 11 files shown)


packages/trie/src/db/checkpoint.ts
```diff
@@ -114,19 +156,27 @@ export class CheckpointDB implements DB {
...
+      this._stats.db.writes += 1
     }
   }
 
   /**
    * @inheritDoc
    */
   async del(key: Uint8Array): Promise<void> {
+    const keyHex = bytesToHex(key)
+    this._cache?.set(keyHex, null)
```
acolytec3, line 179: Should we replace this with 
```js
      this._cache.set(keyHex, null)
      this._stats.cache.writes += 1
    }
```

to ensure the cache stats are updated?

acolytec3 (changes requested):
Looks good over all.  Left a couple of comments we should consider addressing.

packages/trie/src/db/checkpoint.ts
```diff
@@ -114,19 +156,27 @@ export class CheckpointDB implements DB {
...
    */
-  async put(key: Uint8Array, val: Uint8Array): Promise<void> {
+  async put(key: Uint8Array, value: Uint8Array): Promise<void> {
+    const keyHex = bytesToHex(key)
+    if (this._cache !== undefined) {
+      this._cache.set(keyHex, value)
+      this._stats.cache.writes += 1
+    }
     if (this.hasCheckpoints()) {
       // put value in cache
```
acolytec3, line 166: What is the difference between this cache and the one above?  I guess maybe we should refer to this as the "diffCache" or something like that to match what we do in the `statemanager` cache comments.

packages/trie/src/db/checkpoint.ts
```diff
@@ -93,6 +120,16 @@ export class CheckpointDB implements DB {
...
+    if (this._cache !== undefined) {
+      const value = this._cache.get(keyHex)
+      this._stats.cache.reads += 1
+      if (value !== undefined) {
+        this._stats.cache.hits += 1
+        return value
+      }
+    }
+
     // Lookup the value in our cache. We return the latest checkpointed value (which should be the value on disk)
```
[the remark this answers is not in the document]
holgerd77: Yes, see my initial comment, that all doesn't matter since keys are unique, so it will always get the right value.

packages/trie/src/db/checkpoint.ts
holgerd77: Good point, have updated mainly (only) in comments, since we generally had no code cache references in this file before.

packages/trie/src/db/checkpoint.ts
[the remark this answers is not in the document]
acolytec3: Ah, yes, that makes sense.  Thanks!

acolytec3 (approved):
Looks great pending CI passing.

jochem-brouwer (commented):
LGTM

PR #2668, opened 2023-04-27
Client short-term "Execution Mode" Mitigations

holgerd77:
Ok, this PR now starts with some low-level things to ease respectively improve (aka: speed up) "execution mode", so to catch up with execution if things are behind.

It does so by:

- Allowing to disable transports (actually by updating the Yargs version, `--transports` didn't work as expected before)
- Allowing to completely disable sync (`syncMode` option also renamed to `sync` since easier to write, new option `none`, so now: `--sync=full` (normally not used), `sync=none`)

So this allow to run the client with:

```shell
npm run client:start -- --transports --sync=none
```

Haven't explicitly tested execution speed-up, since at the end it doesn't really matter for this PR, think it's generally good to have at least these possibilities from above, independtly from the purpose. Should help for execution too though.

PR also cautiously sets up the block per request and execution iteration from 50 to 100 following Jochems suggestion respectively mentioning. We can see how this goes, generally one can already see that this better reflects execution/download speed now.

Open for review and merge.

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -58,7 +58,7 @@ const args: ClientOpts = yargs(hideBin(process.argv))
     choices: networks.map((n) => parseInt(n[0])),
     default: undefined,
   })
-  .option('syncMode', {
+  .option('sync', {
     describe: 'Blockchain sync mode (light sync experimental)',
     choices: Object.values(SyncMode),
     default: Config.SYNCMODE_DEFAULT,
@@ -754,7 +754,7 @@ async function run() {
     multiaddrs,
     port: args.port,
     saveReceipts: args.saveReceipts,
-    syncmode: args.syncMode,
+    syncmode: args.sync,
     disableBeaconSync: args.disableBeaconSync,
     forceSnapSync: args.forceSnapSync,
     transports: args.transports,
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -21,6 +21,8 @@ import { Level } from 'level'
 import { homedir } from 'os'
 import * as path from 'path'
 import * as readline from 'readline'
+import yargs from 'yargs'
+import { hideBin } from 'yargs/helpers'
 
 import { EthereumClient } from '../lib/client'
 import { Config, DataDirectory, SyncMode } from '../lib/config'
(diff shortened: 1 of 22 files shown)


packages/client/bin/cli.ts
```diff
@@ -38,15 +40,13 @@ import type { BlockBytes } from '@ethereumjs/block'
...
-const { hideBin } = require('yargs/helpers')
-const yargs = require('yargs/yargs')
-
 type Account = [address: Address, privateKey: Uint8Array]
 
 const networks = Object.entries(Common._getInitializedChains().names)
 
 let logger: Logger
 
+// @ts-ignore
```
holgerd77, line 49: There is some type problem (which was just hidden before) occurring now that yargs has types: the `--dev` option is not so much liked by Yargs allowing for boolean and string values at the same time.

I nevertheless didn't dare to change anything here since I wasn't sure about the implications. Maybe @g11tech can chime in here, eventually also in a follow-up PR, since he has introduced most of these `--dev` switches for devnets I guess.

packages/client/bin/cli.ts
```diff
@@ -58,7 +58,7 @@ const args: ClientOpts = yargs(hideBin(process.argv))
     choices: networks.map((n) => parseInt(n[0])),
     default: undefined,
   })
-  .option('syncMode', {
+  .option('sync', {
```
acolytec3, line 61: I'm not really sure why this but with this change the `--dataDir` parameter no longer accepts relative paths (e.g. `--dataDir=../../datadir`) but only absolute ones.  I don't think that's necessarily a bad thing but we should probably note in the parameter help about "only accepts absolute paths" or something to that effect since `yargs` gives a weird error about "unknown parameter --dataDir, --datadir, --data-dir`

holgerd77:
@acolytec3 have force-pushed and removed the "opened removal" commit, sorry, only for practical reasons I had some local changes and didn't want to go through a stash or whatever process.

acolytec3:
That's fine. I just happened to notice it in there and I'm not sure what it's for. I don't see that particular opt referenced anywhere in the codebase unless I'm missing something.

packages/client/bin/cli.ts
g11tech: will review although we have not been using `--dev` for a long time now

jochem-brouwer:
bb50f723237a176c9d0bdbc2ed3c4d79b3d40faf fixes the failing test.

Note that with this commit we now actually support client tests which have more than two `baseRequest`s per test. Before, the first `baseRequest` always shut downs the client. Note that in https://github.com/ethereumjs/ethereumjs-monorepo/blob/bb50f723237a176c9d0bdbc2ed3c4d79b3d40faf/packages/client/lib/execution/vmexecution.ts#L260 , by this PR, we do NOT run any blocks if client is in shutdown state. Therefore, the block with empty txs is not being ran and therefore also the receipts/txs of block 1 have been pruned and we thus return the tx.

g11tech (approved):
lgtm!

PR #2669, opened 2023-04-27
blockchain: remove level dependency from blockchain

gabrocheleau:


holgerd77:
Cool 🤩, had some first look over the changes, I would generally say this already very much goes into a good direction! 👍

packages/blockchain/src/consensus/clique.ts
```diff
@@ -253,7 +248,7 @@ export class CliqueConsensus implements Consensus {
       bigIntToBytes(state[0]),
       state[1].map((a) => a.toBytes()),
     ])
-    await this.blockchain!.db.put(CLIQUE_SIGNERS_KEY, RLP.encode(formatted), DB_OPTS)
+    await this.blockchain!.db.put(CLIQUE_SIGNERS_KEY, RLP.encode(formatted) /* , DB_OPTS */)
```
holgerd77, line 251: Clean-up.

packages/blockchain/src/consensus/clique.ts
```diff
@@ -414,7 +409,7 @@ export class CliqueConsensus implements Consensus {
       bigIntToBytes(v[0]),
       [v[1][0].toBytes(), v[1][1].toBytes(), v[1][2]],
     ])
-    await this.blockchain!.db.put(CLIQUE_VOTES_KEY, RLP.encode(formatted), DB_OPTS)
+    await this.blockchain!.db.put(CLIQUE_VOTES_KEY, RLP.encode(formatted) /* , DB_OPTS */)
```
holgerd77, line 412: Here as well.

packages/blockchain/src/consensus/ethash.ts
```diff
@@ -45,7 +44,7 @@ export class EthashConsensus implements Consensus {
   public async genesisInit(): Promise<void> {}
   public async setup({ blockchain }: ConsensusOptions): Promise<void> {
     this.blockchain = blockchain
-    this._ethash = new Ethash(this.blockchain.db as unknown as EthashCacheDB)
+    this._ethash = new Ethash(this.blockchain.db as any)
```
holgerd77, line 47: Not for this PR, but should we really take this occasion and loosen the binding between blockchain and the Ethash class/package? So eventually make this an optional dependency? 🤔

I think this is not an easy side-shot though but needs some additional design thoughts, regarding consensus instantiation (eventually) on mainnet, thinks like that.

But at the end should not be that complicated/invasive and I guess it's maybe/likely the time for that.

packages/blockchain/src/db/manager.ts
```diff
@@ -109,30 +102,27 @@ export class DBManager {
...
-        if (!equalsBytes(header.withdrawalsRoot, KECCAK256_RLP)) {
-          throw error
+        if (
+          !equalsBytes(header.withdrawalsRoot, KECCAK256_RLP) &&
+          (body.length !== 3 || body[2]?.length === 0)
+        ) {
+          throw new Error('withdrawals root shoot be equal to hash of null when no withdrawals')
         }
-        body.push([])
+        if (body.length !== 3) body.push([])
```
holgerd77, line 125: Not have the complete picture here, is the case at this point that body length might equal 3 sufficiently covered or alternatively not possible to reach due to previous code?

packages/blockchain/src/db/map.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  copy(): DB<TKey, TValue> {
+    return new MapDB<TKey, TValue>(this._database)
+  }
+
+  open() {
+    return Promise.resolve()
+  }
+}
```
holgerd77, line 46: Is this 1:1 the same as the Trie version?

Wonder if we want to move this to Util as well? (Code) bloat seems limited to me, not such an extensive implementation.

Would this make sense?

packages/client/bin/cli.ts
```diff
@@ -432,7 +433,7 @@ async function startClient(config: Config, customGenesisState?: GenesisState) {
   if (customGenesisState !== undefined) {
     const validateConsensus = config.chainCommon.consensusAlgorithm() === ConsensusAlgorithm.Clique
     blockchain = await Blockchain.create({
-      db: dbs.chainDB,
+      db: new LevelDB(dbs.chainDB) as any,
```
holgerd77, line 436: `as any` necessary here?

packages/client/lib/execution/level.ts
```diff
@@ -20,16 +28,32 @@ export class LevelDB implements DB {
...
+      encoding = 'view'
+    if (key === 'heads' || typeof key === 'number') {
+      encoding = 'json'
+    }
     try {
-      value = await this._leveldb.get(key, ENCODING_OPTS)
+      value = await this._leveldb.get(key, {
+        valueEncoding: encoding,
+      })
+      if (value === null) return undefined
```
holgerd77, line 56: Ah, this is too hacky here, we need to find another solution for that. 🙂

(so I am getting this right that this would also be the "default" implementation for other Blockchain LevelDB users, right?)

My tendency would be that we make encoding a mandatory method parameter in the `DB` interface and then just always pass this explicitly in our implementations and one can then decide in the wrapper implementation s (like `LevelDB`, so this one here) to use this information or not.

Does this make sense?

packages/client/lib/sync/fetcher/accountfetcher.ts
```diff
@@ -202,7 +203,7 @@ export class AccountFetcher extends Fetcher<JobTask, AccountData[], AccountData>
       }
     }
 
-    const trie = new Trie({ db: new LevelDB() })
+    const trie = new Trie({ db: new LevelDB() as DB })
```
holgerd77, line 206: Same. Necessary?

holgerd77 (changes requested):
Great refactor, totally excited about this! 🎉

Some comments, biggest thing likely this encoding thing (can also continue to discuss in the chat).

packages/client/lib/execution/level.ts
acolytec3: We can though I don't know if we should or not.  The nicest sort ofd "default" way would be that the DB just takes what we pass in without doing anything to it and returns that same value when we pass in the same key (more or less how our `MapDB` does).  The reason I have to do all this special handling is that `level` assumes that all data should be a certain datatype and format when you dump it into the DB.  I'm not an expert on key/value DBs but if other key/value DBs require encoding details and if the names that specify those encoding types is identical, I think it's fine.  If not, I'd prefer to err on the side of not overspecifying what should be in the wrapper.

packages/ethash/src/level.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  copy(): DB {
+    return new LevelDB(this._leveldb)
+  }
+
+  open() {
+    return this._leveldb.open()
+  }
+}
```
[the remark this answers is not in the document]
acolytec3: I'm not actually sure.  Let me go back and remove some of them and see how it goes.

packages/client/bin/cli.ts
acolytec3: It isn't here.  It is in certain other places since we re-use the `db` from the client across lower level modules and they have their own particular key/value typing that is a subset (or different in the `Ethash` case) than what the `client` uses, though the underlying database can support either.

packages/client/lib/execution/vmexecution.ts
```diff
@@ -50,7 +51,7 @@ export class VMExecution extends Execution {
 
     if (this.config.vm === undefined) {
       const trie = new Trie({
-        db: new LevelDB(this.stateDB),
+        db: new LevelDB(this.stateDB) as DB,
```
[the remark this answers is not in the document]
acolytec3: It is in this instance since the `client` DB allows more key and value types than does the `trie`.

packages/client/lib/execution/level.ts
holgerd77: The wrapper is a code part which is not in our control. People will just copy over this code part if they want to use Blockchain with LevelDB. So this should have a very stable API and be robust and generic in the inside. We do not want to tell people "oh, and please add `if (key === NewKeyWeAddedToBlockchain)` in line 34 of your wrapper" when we publish an update of our Blockchain library where we add some new functionality. 🙂

packages/client/lib/execution/vmexecution.ts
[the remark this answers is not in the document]
holgerd77: Can we optimally align this so that we have one single interface for Trie, Blockchain and eventual other usages? 

(this goes a bit together with the special-handling topic for the encoding values)

holgerd77 (approved):
Ok, looks great, I think I will merge this in here now.

Feel free to do continued refinements though.

PR #2671, opened 2023-04-29
tx: Fix EIP-155 transaction encoding on chain ID 0.

MicahZoltu:
When preparing Ethereum data for RLP encoding, you must convert all values to either byte arrays or lists.  Scalar values (positive integers) are converted to byte arrays by using the fewest bytes possible to represent the value in big endian format.  The canonical form of any scalar value prepped for RLP encoding is that any leading `0` bytes stripped.  In the case of `0`, this means that the only byte (`0x00`) is stripped and you are left with an empty array.

Most of the values in this code are properly 0-stripped using `bigIntToUnpaddedBuffer`.  Unfortunately, it appears that the `chainId` was added without stripping leading zeros.  In most cases this doesn't matter, but if someone is running a chainId 0 blockchain and they want to EIP-155 encode their transaction, the previous code would result in an invalid transaction because only the canonical form is allowed.

This change fixes this bug and correctly prepares the chainId in EIP-155 transactions for RLP encoding.

### DIFF
--- a/packages/tx/src/legacyTransaction.ts
+++ b/packages/tx/src/legacyTransaction.ts
@@ -190,7 +190,7 @@ export class Transaction extends BaseTransaction<Transaction> {
     ]
 
     if (this.supports(Capability.EIP155ReplayProtection)) {
-      values.push(toBuffer(this.common.chainId()))
+      values.push(bigIntToUnpaddedBuffer(this.common.chainId()))
       values.push(unpadBuffer(toBuffer(0)))
       values.push(unpadBuffer(toBuffer(0)))
     }


g11tech (approved):
:+1:

holgerd77:
Just to note: we have somewhat of a `master` branch merge stop since we did our last planned round of releases for this branch last week.

So we might take this change over to our `develop-v7` branch where we are planning the next round of breaking releases.

(@MicahZoltu or is this an urgent bug fix for you?)

MicahZoltu:
It is not an urgent bugfix.

The ethers test suite used this library to generate its test cases which resulted in a large number of test cases that are invalid/incorrect, and I ran across it when testing my library against that suite.  Getting this fixed would let ethers fix its test suite which would let me test my library more completely.  😄

01SMD (commented):
Chain id(1)

PR #2672, opened 2023-04-30
Trie rangeproof v2

jochem-brouwer:
Related to https://github.com/ethereumjs/ethereumjs-monorepo/issues/1054

createRangeProof allows users to feed a start key and an end key (unhashed), where the returned proof can be fed into verifyRangeProof in order to prove the range. (This can also be used in snap sync)

Note: this is not optimal, because the optimized version would require a flat key/value database. This implementation naively walks the trie, and is thus I/O heavy.

Super WIP.

Supersedes #2243

(OOPS, I messed up! Was: 28e2e556d)


jochem-brouwer:
See #2821

PR #2673, opened 2023-05-01
WIP removal of level DB stuff

acolytec3:


packages/trie/src/db/checkpoint.ts
```diff
@@ -177,12 +178,12 @@ export class CheckpointDB implements DB {
   async del(key: Uint8Array): Promise<void> {
     const keyHex = bytesToHex(key)
     if (this._cache !== undefined) {
-      this._cache.set(keyHex, null)
```
holgerd77, line 180: This will likely not fully work. There are different semantics in the cache for „known to not exist“ (as meant here) which was null before and „not in the cache“ which was already undefined before, I battled with this a long time.

This comes into play when the cache get method is called, because this gives undefined when the key is not in the cache, now it also gives undefined for the above case (so there is this together-throwing).

This is why I added a somewhat more complex dict structure in the account cache for the account value or undefined, so there is an „outer“ and an „inner“ undefined, which makes up for the differentiation there.

Maybe it is easier to keep null here, I was actually never that happy with the account cache structure regarding this, since it creates some overhead.

acolytec3: Understood, I haven't really dug into the `trie` piece of it yet.  I've been focused on trying to get `blockchain` passing all the tests while removing as much of the dependence on `try/catch` logic as possible.

packages/blockchain/src/db/map.ts
```diff
@@ -39,4 +39,8 @@ export class MapDB<TKey extends Uint8Array | string, TValue extends Uint8Array |
   copy(): DB<TKey, TValue> {
     return new MapDB<TKey, TValue>(this._database)
   }
+
+  open() {
+    return Promise.resolve()
+  }
```
gabrocheleau, line 45: Does this actually do anything? Or is this added in to make the interface consistent with the expected types?

packages/client/lib/execution/vmexecution.ts
```diff
@@ -49,7 +50,7 @@ export class VMExecution extends Execution {
 
     if (this.config.vm === undefined) {
       const trie = new Trie({
-        db: new LevelDB(this.stateDB),
+        db: new LevelDB(this.stateDB) as DB<Uint8Array, Uint8Array>,
```
gabrocheleau, line 53:
```suggestion
        db: new LevelDB(this.stateDB) as DB,
```
I don't think the types are necessary since this defaults to Uint8Array.

packages/client/lib/sync/fetcher/accountfetcher.ts
```diff
@@ -202,7 +203,7 @@ export class AccountFetcher extends Fetcher<JobTask, AccountData[], AccountData>
       }
     }
 
-    const trie = new Trie({ db: new LevelDB() })
+    const trie = new Trie({ db: new LevelDB() as DB<Uint8Array, Uint8Array> })
```
gabrocheleau, line 206: same

gabrocheleau (approved):

PR #2675, opened 2023-05-03
Client Memory Optimizations

holgerd77:
Memory optimizations for the client and in general, can either be directly merged or used as a base for additional pushes, don't really care, what fits better.

### DIFF
--- a/packages/client/lib/execution/execution.ts
+++ b/packages/client/lib/execution/execution.ts
@@ -56,11 +56,7 @@ export abstract class Execution {
    */
   async open(): Promise<void> {
     this.started = true
-    if (this.config.execution) {
-      this.config.logger.info('Setup EVM execution.')
-    } else {
-      this.config.logger.info('EVM execution skipped.')
-    }
+    this.config.logger.info('Setup EVM execution.')
   }
 
   /**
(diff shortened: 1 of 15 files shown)


packages/client/lib/execution/execution.ts
```diff
@@ -56,15 +56,21 @@ export abstract class Execution {
    */
   async open(): Promise<void> {
     this.started = true
-    this.config.logger.info('Setup EVM execution.')
+    if (this.config.execution) {
```
g11tech, line 59: i think we should not change the log UX here as the execution could still be triggered by beacon sync for e.g. and then stop would not log the stop message.

instead we should just add a log when we trigger execution run in full service start

packages/client/lib/service/fullethereumservice.ts
```diff
@@ -162,7 +162,7 @@ export class FullEthereumService extends EthereumService {
     }
     await super.start()
     this.miner?.start()
-    if (!this.config.execCommon.gteHardfork(Hardfork.Paris)) {
+    if (!this.config.execCommon.gteHardfork(Hardfork.Paris) && this.config.execution) {
       void this.execution.run(true, true)
```
g11tech, line 166:
```suggestion
     this.logger.info("Triggering execution of already downloaded blocks if any")
      void this.execution.run(true, true)
```
something like this...

packages/client/lib/execution/execution.ts
holgerd77: 👍

Could you eventually directly push ah update?

holgerd77:
Will admin merge here for procedural reasons, we can see if codecov not reporting was eventually a one time thing on the next PR.

PR #2677, opened 2023-05-04
client: Fix unclean shutdown scenario where SIGINT may come before client fully started

g11tech:
It may take client a little while to fully start and bootstrap its services (especially if there are pending blocks to be executed). If SIGINT/CTRL + C comes before that, client may end up in an unclean shutdown

image

This PR hooks up shutdown handler as soon as client is started, and also makes sure client is properly started before initiating a clean shutdown:

image

PS: Build on top of client memory optimizations PR and is pointing to that branch
 - https://github.com/ethereumjs/ethereumjs-monorepo/pull/2675

So pls rebase and merge afterwards 2675 is merged and with the base branch of this PR updated to `develop-v7`

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -785,22 +785,42 @@ async function run() {
     config.logger.info(`Reading custom genesis state accounts=${numAccounts}`)
   }
 
-  const client = await startClient(config, customGenesisState)
-  const servers =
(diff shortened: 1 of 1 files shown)


holgerd77:
Have admin merged 2675, thought this would ease the process, PR here is still off with commits though, might need an update.

g11tech:
> Have admin merged 2675, thought this would ease the process, PR here is still off with commits though, might need an update.

yea noticed :slightly_smiling_face: , updated

PR #2680, opened 2023-05-05
blockchain: Save iterator head to last successfuly executed even on errors

g11tech:
On further debugging and testing another case of `vmHead` corruption was observed:

1. client was executing 749796-749896 range but got an interrup signal and 749890 failed

2. next time client started vmHead was already updated to 749896 and execution started from block 749897 which obviously would fail
image

The synopsis of the issue:

In vmExecution  iterator, block error was being ignored leading to incorrect vmHead forward movement. So the right thing to do was to throw error. 
However in blockchain,  iterator head was not being properly saved to last successfully executed if `onBlock` callbacks errored. 

This PR addressed both the issues and also moves out iterator error handling out side of the vm iterator callback for further simplicity.

### DIFF
--- a/packages/blockchain/src/blockchain.ts
+++ b/packages/blockchain/src/blockchain.ts
@@ -1016,7 +1016,7 @@ export class Blockchain implements BlockchainInterface {
               }
             }
 
-            // if there was no reorg, updated head
+            // if there was no reorg, update head
             if (!reorgWhileOnBlock) {
               this._heads[name] = nextBlock.hash()
               lastBlock = nextBlock
--- a/packages/blockchain/src/blockchain.ts
+++ b/packages/blockchain/src/blockchain.ts
@@ -995,6 +995,9 @@ export class Blockchain implements BlockchainInterface {
               nextBlockNumber = headBlockNumber + BigInt(1)
               nextBlock = await this.getBlock(nextBlockNumber)
             }
+
+            // While running onBlock with released lock, reorgs can happen via putBlocks
+            let reorgWhileOnBlock = false
             if (releaseLockOnCallback === true) {
               this._lock.release()
             }
@@ -1003,13 +1006,23 @@ export class Blockchain implements BlockchainInterface {
             } finally {
               if (releaseLockOnCallback === true) {
                 await this._lock.acquire()
+                // If lock was released check if reorg occured
+                const nextBlockMayBeReorged = await this.getBlock(nextBlockNumber).catch(
+                  (_e) => null
+                )
+                reorgWhileOnBlock = nextBlockMayBeReorged
+                  ? !equalsBytes(nextBlockMayBeReorged.hash(), nextBlock.hash())
+                  : true
               }
             }
 
+            // if there was no reorg, updated head
+            if (!reorgWhileOnBlock) {
+              this._heads[name] = nextBlock.hash()
+              lastBlock = nextBlock
+              nextBlockNumber++
+            }
             // Successful execution of onBlock, move the head pointer
-            this._heads[name] = nextBlock.hash()
-            lastBlock = nextBlock
-            nextBlockNumber++
             blocksRanCounter++
           } catch (error: any) {
             if (error.code === 'LEVEL_NOT_FOUND') {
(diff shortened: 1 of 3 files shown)


packages/blockchain/src/blockchain.ts
```diff
@@ -981,44 +981,48 @@ export class Blockchain implements BlockchainInterface {
...
-            headBlockNumber = await this.dbManager.hashToNumber(headHash)
-            nextBlockNumber = headBlockNumber + BigInt(1)
-            nextBlock = await this.getBlock(nextBlockNumber)
-          }
-          this._heads[name] = nextBlock.hash()
-          lastBlock = nextBlock
-          if (releaseLockOnCallback === true) {
-            this._lock.release()
-          }
+      try {
```
g11tech, line 984: lots of diff here is because of formatting change introduced because of this `try`

packages/client/lib/execution/vmexecution.ts
```diff
@@ -302,82 +302,105 @@ export class VMExecution extends Execution {
...
-            }
-          }
-          // run block, update head if valid
-          try {
-            const { number, timestamp } = block.header
-            if (typeof blockchain.getTotalDifficulty !== 'function') {
-              throw new Error(
-                'cannot get iterator head: blockchain has no getTotalDifficulty function'
-              )
+      this.vmPromise = blockchain
```
g11tech, line 305: again most of diff here is because of formatting changed because of introduction to the `catch` to the iterator promise below

packages/blockchain/src/blockchain.ts
```diff
@@ -981,44 +981,48 @@ export class Blockchain implements BlockchainInterface {
...
-          blocksRanCounter++
-        } catch (error: any) {
-          if (error.code === 'LEVEL_NOT_FOUND') {
-            break
-          } else {
-            throw error
           }
         }
+        return blocksRanCounter
+      } finally {
```
acolytec3, line 1023: To make sure I understand what's going on here, this outer `try...finally` block just makes sure save the current head and then throws the triggering error out to the client so it can exit gracefully?

packages/client/lib/execution/vmexecution.ts
```diff
@@ -302,82 +302,105 @@ export class VMExecution extends Execution {
...
+              errorBlock = block
+              throw error
+            }
+          },
+          this.config.numBlocksPerIteration,
+          // release lock on this callback so other blockchain ops can happen while this block is being executed
+          true
+        )
+        // Ensure to catch and not throw as this would lead to unCaughtException with process exit
+        .catch(async (error) => {
```
acolytec3, line 402: If we catch the error here, what error is thrown that tells the client to exit?

acolytec3 (commented):
Left a couple of questions to clarify my understanding of what the new catch blocks do but looks fine so far as it goes.

Is there any way to easily test that this works?

g11tech:
> Left a couple of questions to clarify my understanding of what the new catch blocks do but looks fine so far as it goes.
> 
> Is there any way to easily test that this works?

yes : first run client with just sync: `./node_modules/.bin/ethereumjs --network mainnet --execution false`

after it has synced couple thousand blocks, just not test with execution and keep doing Ctrl C and restart to 

`./node_modules/.bin/ethereumjs --network mainnet --sync none`

packages/blockchain/src/blockchain.ts
g11tech: yes correct! this.heads['vm'] always have the last successfuly executed block

packages/client/lib/execution/vmexecution.ts
acolytec3: Ah, right, this is just handling an error that occurs when the client is already in the process of exiting.  Makes sense.

acolytec3:
There are a couple of reorg related `blockchain` tests failing so I wonder if there's some minor parentheses or bracket that got moved around the reorg logic.

g11tech:
> There are a couple of reorg related `blockchain` tests failing so I wonder if there's some minor parentheses or bracket that got moved around the reorg logic

fixed an edge case :+1:

acolytec3 (approved):
Lgtm!

PR #2681, opened 2023-05-08
Trie: More Conservative Trie Cache Behavior

holgerd77:
It turned out that the current trie cache substantially slows down block execution in certain client block ranges respectively call/attack situations, likely in combination with late reverts since the trie cache just lazily stores all nodes in the hope that these will be re-usable eventually also in case of a revert.

Block range where slowing has been tested was the following, with and without a `--trieCache=0` addition:

```shell
node dist/bin/cli.js --dataDir=/Volumes/Eth-SSD/sync/ --executeBlocks=2433000-2433121
```

Trie cache enabled version was substantially slower here.

This PR somewhat mitigates this by switching to a more conservative cache behavior by only storing:

a) values which at least once made it to the trie (in `get`)
b) values which are in the process of being written to the trie after last CP has been comitted

A re-run has shown that this significantly improves on the above and very likely other block ranges.

Cache behavior should be continued to be monitored, likely additional tweaks necessary.

### DIFF
--- a/packages/trie/src/db/checkpoint.ts
+++ b/packages/trie/src/db/checkpoint.ts
@@ -158,16 +158,17 @@ export class CheckpointDB implements DB {
    */
   async put(key: Uint8Array, value: Uint8Array): Promise<void> {
     const keyHex = bytesToHex(key)
-    if (this._cache !== undefined) {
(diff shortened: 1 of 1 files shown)


g11tech (approved):
lgtm!

PR #2684, opened 2023-05-10
block: Add helper to construct block from beacon payload

g11tech:
Add a handy static method to construct block from beacon payload which is more readily available via beacon apis: https://ethereum.github.io/beacon-APIs/#/Beacon/getBlockV2


holgerd77:
Just labelled this on occasion, hope that's ok.

Interesting PR! 🙂

packages/block/src/from-beacon-payload.ts
```diff
@@ -0,0 +1,65 @@
...
+  timestamp: string
+  extra_data: string
+  base_fee_per_gas: string
+  block_hash: string
+  transactions: string[]
+  withdrawals?: BeaconWithdrawal[]
+  excess_data_gas?: string
+}
+
+export function beaconToExecutionPayload(payload: BeaconPayload): ExecutionPayload {
```
acolytec3, line 33: Can we rename this `beaconBlockToExecutionPayload` and add some docs to explain what it does and how to use it?

acolytec3 (changes requested):
Code looks great.  Just need a little bit of additional docs in the new helper method and this should be ready to merge!

packages/block/src/from-beacon-payload.ts
g11tech: its not a beacon block, its execution payload in beacon block ( see the jq path here: `curl localhost:5052/eth/v2/beacon/blocks/56610 | jq .data.message.body.execution_payload`), let me know what you finally want to keep

acolytec3: So maybe name it something like `executionPayloadFromBeaconBlock` and then add TypeDocs specifying what the input/outputs are.  The doc could be something like:
```
/**
 * Converts a beacon block execution payload JSON object to the {@link ExecutionPayload} data needed to construct a {@link Block}.
 * The JSON data can be retrieved from a consensus layer (CL) client on this Beacon API `/eth/v2/beacon/blocks/[block number]`
 */
 ```

g11tech: applied changes

holgerd77:
Oh. Something very fishy going on with `test-client`. 🙂

acolytec3 (approved):
Looks great, especially with the new docs!  Will merge once the CI passes.

holgerd77:
Really nice PR! 🙂👍

PR #2686, opened 2023-05-10
evm: Fix the gasCost logs in op code trace

g11tech:
While matching op code trace of ethereumjs with the geth it was observed that ethereumjs's gasCost logs were incorrect w.r.t the actual gas used by the opcode.

This was happening because the dynamicFee should be the one to be logged which is the actual gas consumed by the opcode 

Post this the gasCost mismatch diff with geth greatly reduced and the `0` gasCost entries disappeared from the trace


holgerd77 (approved):
LGTM

PR #2696, opened 2023-05-10
Abstract "keys" method

TimDaub:
- the ethereumjs trie now uses these db adapters
- e.g. I'm using the same lmdb adapter that is used here: https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/trie/recipes
- and e.g. lmdb doesn't implement a "keys" function: https://github.com/kriszyp/lmdbx-js#dbgetkeysoptions-rangeoptions-iterableany it's called "getKeys" and it's also a promise
- However, before merging this PR, I think we should source the respective "keys" methods for all exemplified db adapters


holgerd77:
No follow-up, will close

PR #2697, opened 2023-05-11
Use `storageCache` in `ethersStateManager`

acolytec3:
Updates the `ethersStateManager` to use the newly implemented `storageCache` instead of the custom JS Map we were using before.

### DIFF
--- a/packages/statemanager/src/ethersStateManager.ts
+++ b/packages/statemanager/src/ethersStateManager.ts
@@ -46,7 +46,7 @@ export class EthersStateManager implements StateManagerInterface {
       blockTag: BigInt(this.blockTag),
     })
     ;(newState as any).contractCache = new Map(this.contractCache)
-    ;(newState as any).storageCache = this.storageCache
+    ;(newState as any).storageCache = new StorageCache({ size: 10000, type: CacheType.LRU })
     ;(newState as any)._accountCache = this._accountCache
     return newState
   }
--- a/packages/statemanager/src/ethersStateManager.ts
+++ b/packages/statemanager/src/ethersStateManager.ts
@@ -1,10 +1,11 @@
+import { RLP } from '@ethereumjs/rlp'
 import { Trie } from '@ethereumjs/trie'
 import { Account, bigIntToHex, bytesToBigInt, bytesToHex, toBytes } from '@ethereumjs/util'
 import { debug } from 'debug'
 import { keccak256 } from 'ethereum-cryptography/keccak'
 import { ethers } from 'ethers'
 
-import { AccountCache, CacheType } from './cache'
+import { AccountCache, CacheType, StorageCache } from './cache'
 
 import type { Proof } from '.'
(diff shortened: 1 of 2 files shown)


packages/statemanager/src/ethersStateManager.ts
```diff
@@ -47,7 +47,7 @@ export class EthersStateManager implements StateManagerInterface {
       blockTag: BigInt(this.blockTag),
     })
     ;(newState as any).contractCache = new Map(this.contractCache)
-    ;(newState as any).storageCache = new Map(this.storageCache)
+    ;(newState as any).storageCache = this.storageCache
```
jochem-brouwer, line 50: This doesn't copy the storage cache while it was copied before, is this intended?

packages/statemanager/src/ethersStateManager.ts
```diff
@@ -109,24 +109,22 @@ export class EthersStateManager implements StateManagerInterface {
...
 
     // Retrieve storage slot from provider if not found in cache
-    storage = await this.provider.getStorageAt(
+    const storage = await this.provider.getStorageAt(
       address.toString(),
       bytesToBigInt(key),
       this.blockTag
     )
-    const value = toBytes(storage)
+    value = toBytes(storage)
```
jochem-brouwer, line 127: Should it also put the value in the storage here?

jochem-brouwer (commented):
Two questions

g11tech (dismissed):
looks neat :+1:

packages/statemanager/src/ethersStateManager.ts
acolytec3: It does two lines below

packages/statemanager/src/ethersStateManager.ts
acolytec3: Hmm, interesting point.  What should we do here?  The `DefaultStateManager` just creates a brand new cache using the same settings as the original (but not actually copying the cached values).  This is a bit of a different beast as the cache is the "actual" changed values.

g11tech: I think current copy semantics is to copy original cache, (we use vmcopy for e.g. to get a new copy but not with any uncommitted (to disk) modifications)

acolytec3: Hmm, ok, I'll have to think how to do this.  Not sure exactly the best way to handle this.

packages/statemanager/src/ethersStateManager.ts
jochem-brouwer: Ah ok great :+1:

packages/statemanager/src/ethersStateManager.ts
jochem-brouwer:
```suggestion
    ;(newState as any).storageCache = new StorageCache({ size: 10000, type: CacheType.LRU })
```

I would suggest to do this.

jochem-brouwer (approved):
LGTM

PR #2700, opened 2023-05-15
client: Enable default debug logging to file

g11tech:
Most if not all clients have this as default policy:
 - info logs on stdout
 - debug logs on file with filerotate

The default enabling of debug logs help in retrospective debugging even if the user forgets to enable logging. This PR enables the same and fixes a previous issue of `--logFile true` not working which according to help should have logged to `ethereumjs.log`

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -734,7 +734,7 @@ async function run() {
   ensureDirSync(configDirectory)
   const key = await Config.getClientKey(datadir, common)
 
-  // logFile is either filename or string "true"
+  // logFile is either filename or boolean true or false to enable (with default) or disable
   if (typeof args.logFile === 'boolean') {
     args.logFile = args.logFile ? `${networkDir}/ethereumjs.log` : undefined
   }
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -182,12 +182,14 @@ const args: ClientOpts = yargs(hideBin(process.argv))
     default: 'info',
   })
   .option('logFile', {
-    describe: 'File to save log file (pass true for `ethereumjs.log`)',
+    describe:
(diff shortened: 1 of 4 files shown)


holgerd77 (dismissed):
Yes, makes sense, LGTM!

packages/client/bin/cli.ts
```diff
@@ -727,9 +729,16 @@ async function run() {
...
 
   const datadir = args.dataDir ?? Config.DATADIR_DEFAULT
-  const configDirectory = `${datadir}/${common.chainName()}/config`
+  const networkDir = `${datadir}/${common.chainName()}`
+  const configDirectory = `${networkDir}/config`
   ensureDirSync(configDirectory)
   const key = await Config.getClientKey(datadir, common)
+
+  // logFile is either filename or string "true"
+  if (typeof args.logFile === 'boolean') {
```
acolytec3, line 738: How does this `typeof` call know it's a `boolean` as opposed to a string?

g11tech: the comment is misleading because of some refactoring i did:

`// logFile is either filename or boolean true/false`

packages/client/bin/cli.ts
```diff
@@ -727,9 +729,16 @@ async function run() {
   }
 
   const datadir = args.dataDir ?? Config.DATADIR_DEFAULT
-  const configDirectory = `${datadir}/${common.chainName()}/config`
+  const networkDir = `${datadir}/${common.chainName()}`
+  const configDirectory = `${networkDir}/config`
   ensureDirSync(configDirectory)
   const key = await Config.getClientKey(datadir, common)
+
+  // logFile is either filename or string "true"
```
g11tech, line 737:
```suggestion
  // logFile is either filename or boolean true or false to enable (with default) or disable
```

g11tech:
rebased via UI

holgerd77 (approved):
LGTM, thanks for the additional reviews.

PR #2702, opened 2023-05-15
EVM, remove EEI: cleanup statemanager

jochem-brouwer:
Follow-up of #2649

This PR intends to remove all EVM-related logic from state manager, in particular:

- [x] Touched accounts
- [x] Warm addresses
- [x] Warm slots

This PR will also:

- [x] Optimize `getOriginalContractStorage`
- [x] Cleanup the EVMStateManagerInterface in `common`
- [ ] Ensure `EthersStateManager` is compatible with `EVMStateManagerInterface` and is thus able to also run VM (@holgerd77 could you check this?)
- - [x] (fixes the TODO tests in the ethers state manager tests)

Super WIP


jochem-brouwer:
Will use this comment as a scratchpad otherwise I will lose my line of thinking, so ignore this comment.

For `_accessedStorageReverted` we only use this in generate access list, this should not be a Map, it should just be a Set and each time a frame is reverted, put all items on the `_accessedStorageReverted`. (Also most likely add a flag to disable this behavior since we only need it to generate the access list).

Maybe for the access list just change the behavior: keep track of a set, each time a new slot is touched (whether it is reverted or not) just put it on the access list, this simplifies it a lot.

jochem-brouwer:
Few gotcha's:

**New journal**

The new `evmJournal` has multiple journals which it keeps track of. Each time a new call frame is entered (this can be done by either creating a new call frame OR reverting/finishing a call frame) a new item is pushed on the `journalDiff` stack. This includes the current `journalHeight` (current call depth, thus +1 when a new call frame is entered and -1 if it is committed/revert). This **tracks the items which were added to the `journal` of the `evmJournal`.

Note that `journal` is a single Map. Only items are added to the `journalDiff` stack/array if the item added did not exist yet in the journal. The items are: set of warm addresses, map of warm address => warm storage, set of touched accounts. I tested and it is not possible to assume that warm addresses / touched addresses are equal (`EXTCODE*` warms the address but does not touch it!). 

The journal diff is only used on a revert. If this happens, start at the end of the `journalDiff` array and move back, until the reported journal height is lower than the current journal height. For each of these items (except if the reported journal height is lower than journal height then break the loop), loop over the diff items and remove these items from warm/touched. This ensures that the journal is always the current state of the items in there.

This works a bit the same as the StateManager storage cache (with the `diffCache`, except that now the diff cache items are **not** merged, because this is unnecessary). Each item in the `journalDiff` will only be there once.

**Original storage**

The gotcha here is that a new class is introduced (rather simple), the `OriginalStorageCache`. You need to initialize this with a method to actually get the current contract storage. This is ONLY invoked if the original storage cache is unknown (so get contract storage and use that as original storage). This class is necessary because both EthersStateManager and DefaultStateManager need it for EVM execution. Note that in their constructors, the respective `getContractStorage` is passed as constructor argument to the new class and is bound to that respective state manager.

**Generate Access List**

We previously had a lot of logic regarding accessed storage, accessed warm accounts, with a whole stack mechanism which then needed to merge together at the end. This is unnecessary. We can simply, in the new `evmJournal`, track which items are added to warm storage or warm addresses. If it reverts, we don't care, since it was warmed previously anyways and we thus want to keep this on the reported access list. This logic is now fully moved into EVM.

packages/vm/test/retesteth/transition-child.ts
```diff
@@ -125,7 +125,7 @@ async function runTransition(argsIn: any) {
   const logsBloom = builder.logsBloom()
   const logsHash = keccak256(logsBloom)
 
-  await vm.stateManager.cleanupTouchedAccounts()
+  await vm.evm.evmJournal.cleanupTouchedAccounts()
```
acolytec3, line 128: Shouldn't this just be `evmJournal.cleanup()`?

packages/vm/test/api/EIPs/eip-3529.spec.ts
```diff
@@ -139,7 +139,7 @@ tape('EIP-3529 tests', (t) => {
       )
 
       await vm.stateManager.getContractStorage(address, key)
-      vm.stateManager.addWarmedStorage(address.bytes, key)
+      vm.evm.evmJournal.addPreWarmedSlot(bytesToHex(address.bytes), bytesToHex(key))
```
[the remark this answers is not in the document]
jochem-brouwer: Yes, it warms the slot. The reason why this is a new method, is that these slots are always warm in the current context, as opposed to slots which get warm during the transaction (if a call frame reverts, the newly added warm slots get deleted from the warm slots list, while the pre-warmed slots can never be deleted and should thus be tracked separately).

packages/vm/test/retesteth/transition-child.ts
jochem-brouwer: Yup, `cleanupTouchedAccounts` is not part of the interface anymore. Good catch.

acolytec3: Thank you VS Code.

packages/evm/src/evmJournal.ts
```diff
@@ -0,0 +1,283 @@
...
+
+/**
+ * Journal Diff Item:
+ * Index 0: remove warm address
+ * Index 1: remove warm slots for this warm address
+ * Index 2: remove touched
+ */
+
+type JournalDiffItem = [Set<AddressString>, Map<AddressString, Set<SlotString>>, Set<AddressString>]
+//AddressString | Map<AddressString, SlotString>
```
gabrocheleau, line 29: Should this be removed?

packages/statemanager/test/vmState.spec.ts
```diff
@@ -147,7 +148,10 @@ tape('Original storage cache', async (t) => {
     const res = await stateManager.getContractStorage(address, key)
     st.deepEqual(res, newValue)
 
-    const origRes = await (<any>stateManager).getOriginalContractStorage(address, key)
+    const origRes = await (<any>stateManager).originalStorageCache.getOriginalContractStorage(
```
gabrocheleau, line 151: Is the `<any>` still necessary?

packages/statemanager/test/vmState.spec.ts
```diff
@@ -160,28 +164,40 @@ tape('Original storage cache', async (t) => {
 
     let res = await stateManager.getContractStorage(address, key2)
     st.deepEqual(res, value2)
-    let origRes = await (<any>stateManager).getOriginalContractStorage(address, key2)
+    let origRes = await (<any>stateManager).originalStorageCache.getOriginalContractStorage(
```
gabrocheleau, line 167: Is the `<any>` still necessary?

packages/statemanager/test/vmState.spec.ts
```diff
@@ -160,28 +164,40 @@ tape('Original storage cache', async (t) => {
...
+      key2
+    )
     st.deepEqual(origRes, value2)
 
     await stateManager.putContractStorage(address, key2, value3)
 
     res = await stateManager.getContractStorage(address, key2)
     st.deepEqual(res, value3)
-    origRes = await (<any>stateManager).getOriginalContractStorage(address, key2)
+    origRes = await (<any>stateManager).originalStorageCache.getOriginalContractStorage(
```
gabrocheleau, line 177: Same

packages/statemanager/test/vmState.spec.ts
```diff
@@ -160,28 +164,40 @@ tape('Original storage cache', async (t) => {
...
+      address,
+      key2
+    )
     st.deepEqual(origRes, value2)
 
     // Check previous key
     res = await stateManager.getContractStorage(address, key)
     st.deepEqual(res, hexToBytes('1235'))
-    origRes = await (<any>stateManager).getOriginalContractStorage(address, key)
+    origRes = await (<any>stateManager).originalStorageCache.getOriginalContractStorage(
```
gabrocheleau, line 186: Same

packages/vm/src/runTx.ts
```diff
@@ -144,45 +128,31 @@ export async function runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
     const castedTx = <AccessListEIP2930Transaction>opts.tx
 
     for (const accessListItem of castedTx.AccessListJSON) {
-      const address = toBytes(accessListItem.address)
-      state.addWarmedAddress(address)
+      this.evm.evmJournal.addPreWarmedAddress(accessListItem.address, true)
       for (const storageKey of accessListItem.storageKeys) {
-        state.addWarmedStorage(address, toBytes(storageKey))
+        this.evm.evmJournal.addPreWarmedSlot(accessListItem.address, storageKey, true)
```
gabrocheleau, line 133: Is it intentional that we're refactoring `addWarmedStorage` to `addPreWarmedSlot` (since we have a `addWarmedStorage` method on the evmJournal)?

packages/vm/test/api/runTx.spec.ts
```diff
@@ -474,7 +474,7 @@ tape('runTx() -> runtime behavior', async (t) => {
       await vm.runTx({ tx }) // this tx will fail, but we have to ensure that the cache is cleared
 
       t.equal(
-        (<any>vm.stateManager)._originalStorageCache.size,
+        (<any>vm.stateManager).originalStorageCache.map.size,
```
gabrocheleau, line 477: Is the `<any> necessary?

packages/vm/test/util.ts
```diff
@@ -366,7 +366,7 @@ export async function setupPreConditions(state: EVMStateManagerInterface, testDa
   await state.commit()
   // Clear the touched stack, otherwise untouched accounts in the block which are empty (>= SpuriousDragon)
   // will get deleted from the state, resulting in state trie errors
-  ;(<any>state).touchedJournal.clear()
+  //;(<any>state).touchedJournal.clear()
```
gabrocheleau, line 369: Intentional?

jochem-brouwer: Ah yes, in this method the state is setup via StateManager, not EVM (so there are no touched accounts). Therefore this cleanup action is not necessary. Should remove anyways.

packages/common/src/interfaces.ts
```diff
@@ -40,8 +40,9 @@ export type AccessList = AccessListItem[]
 
 export interface StateManagerInterface {
   accountExists(address: Address): Promise<boolean>
+  accountIsEmptyOrNonExistent(address: Address): Promise<boolean>
```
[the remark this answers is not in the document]
holgerd77: Short context addition: I added (+ named) this method in a first round, @jochem-brouwer feel very free to rename or remove.

So this is the "account being dead" definition from [EIP-161](https://eips.ethereum.org/EIPS/eip-161), that's the state trie clearing HF included in Spurious Dragon. One alternative was also to name as `accountisDead()` or take the EIP in, `accountIsDeadEIP161()` or so, didn't like both versions so I chose this more declarative one.

That said: I would also be very much ok if we totally remove and sleeken the interface here and do the explicit check Gabriel suggested, would have a slight preference for this but totally no strong opinion.

So very much up to you! 🙂

packages/statemanager/src/cache/originalStorageCache.ts
```diff
@@ -0,0 +1,46 @@
...
+    const keyHex = bytesToHex(key)
+    if (map!.has(keyHex) === false) {
+      map!.set(keyHex, value)
+    }
+  }
+
+  clearOriginalStorageCache(): void {
+    this.map = new Map()
+  }
+}
```
holgerd77, line 46: I would have a tendency to drastically shorten the names here, so just use `get`, `put` and `clear`. 😋

Think that makes the API more elegant and readable and what this does can be taken very much from the context, so e.g. `this._stateManager.originalStorageCache.get(...)` (still long enough 🙂).

packages/common/src/interfaces.ts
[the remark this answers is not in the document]
jochem-brouwer: I will check in what contexts this is used (and I doubt there are a lot). We might just remove this and then use it, at the context, by either checking `getAccount(addr) === undefined || getAccount(addr).IsEmpty()` (of course written down a bit more stylish). I think the only context relevant is the state clearing, and I do not think that there is an external use case. There would be if you would check if the account exists, but since all dead accounts are cleared (no dead accounts exists on mainnet) this also doesn't seem to have any external use case.

packages/evm/src/evmJournal.ts
```diff
@@ -0,0 +1,283 @@
...
+type JournalHeight = number
+
+export class EvmJournal {
```
holgerd77, line 32: Here as well, I would suggest to just name `Journal` and also simplify the file name `evmJournal.ts` -> `journal.ts`.

jochem-brouwer: Should I rename `evm.evmJournal` to `evm.journal` as well?

holgerd77: I would vote yes! 🙂

jochem-brouwer:
Ok, I should have resolved most review items except if I should rename `evm.evmJournal` to `evm.journal` (I think I have the tendency to do so).

Furthermore, I have removed `accountExists` and `accountIsEmptyOrNonExistent` from the interface. However, `accountExists` is still in EthersStateManager, since this is rather educational: it shows how to know, from `eth_getProof` if an account exists in the trie or not. You **cannot** use `eth_getAccount` for this, since it will always return an empty account if the account does not exist in the trie, or it exists in the trie but is empty.

I have also cleaned up the method names of `originalStorageCache`.

holgerd77:
> And possibly `preWarmed` -> `alwaysWarm`?

Was very much a quick-shot 😋, I think I still find it substantially easier to grasp though, also on second thought (preWarmed can lead thoughts somewhat in a wrong direction).

jochem-brouwer:
Yup, have addressed!

holgerd77:
Short in-between note: I now synced with this branch > block 1Mio with the client, no anomalies to detect. 🙂

packages/evm/src/interpreter.ts
```diff
@@ -10,6 +10,7 @@ import { trap } from './opcodes'
 import { Stack } from './stack'
 
 import type { EVM, EVMResult } from './evm'
+import type { Journal } from './journal'
 import type { AsyncOpHandler, OpHandler, Opcode } from './opcodes'
 import type { Block, Blockchain, Log } from './types'
 import type { Common, EVMStateManagerInterface } from '@ethereumjs/common'
```
holgerd77, line 16: Debug logger from below, currently `evm:eei:gas` (so annying this untouched-lines-comment-deactivation), should be renamed.

Simply `evm:gas`?

jochem-brouwer:
> Short in-between note: I now synced with this branch > block 1Mio with the client, no anomalies to detect. 🙂 

Cool! I also wonder slightly about performance, did you notice anything?

We could also do a quick test with disk reads before/after this PR. It should decrease (and hopefully a lot! :) )

packages/evm/src/interpreter.ts
jochem-brouwer: Yes, `evm:gas` is perfect. I will do a quick check to see if there are any other eei leftovers.

holgerd77:
> > Short in-between note: I now synced with this branch > block 1Mio with the client, no anomalies to detect. 🙂
> 
> Cool! I also wonder slightly about performance, did you notice anything?
> 
> We could also do a quick test with disk reads before/after this PR. It should decrease (and hopefully a lot! :) )

Performance is always hard to grasp, so I am getting more and more cautious to do any early statements here.

Doing catch-up-execution right now (by deactivating block sync), at ~1.2 Mio (Homestead) atm, everything still running smoothly atm, though block execution has gotten a bit slower, but that's also always dependent of the respective usage of the respective block parts.

I wanted to sync into the "problematic ranges" again, that start in the ~ 1.6 Mio or so block numbers. These are always easier to compare, since base execution numbers for a single block are already pretty high.

Also not totally meaningful then (since single blocks/txs are often targeting very specfic EVM functionality). So at the end best is always to do 2-4 different ranges.

gabrocheleau:
> Ok, I should have resolved most review items except if I should rename `evm.evmJournal` to `evm.journal` (I think I have the tendency to do so).
> 
> Furthermore, I have removed `accountExists` and `accountIsEmptyOrNonExistent` from the interface. However, `accountExists` is still in EthersStateManager, since this is rather educational: it shows how to know, from `eth_getProof` if an account exists in the trie or not. You **cannot** use `eth_getAccount` for this, since it will always return an empty account if the account does not exist in the trie, or it exists in the trie but is empty.
> 
> I have also cleaned up the method names of `originalStorageCache`.

Nice, I really like the API simplifications!

acolytec3 (approved):
Looks great with the additional edits!

PR #2705, opened 2023-05-16
tx: ensure rpc txs are correctly decoded for v=0

jochem-brouwer:
In https://github.com/ethereumjs/ethereumjs-monorepo/pull/2613 I "solved" decoding the optimism system txs wich v=0, but forgot that EIP2930 or EIP1559 have 50% chance to have v=0 as well. If we then set v=undefined, then these txs are suddenly unsigned while they clearly are. Test included.

Also renames `fromRPCTx` -> `fromRPC` in line with the Block lib.

### DIFF
--- a/packages/tx/src/fromRpc.ts
+++ b/packages/tx/src/fromRpc.ts
@@ -28,7 +28,7 @@ export const normalizeTxParams = (_txParams: any): TxData => {
   txParams.r = txParams.r === '0x0' ? '0x' : txParams.r
   txParams.s = txParams.s === '0x0' ? '0x' : txParams.s
 
-  if (txParams.v !== '0x') {
+  if (txParams.v !== '0x' || txParams.r !== '0x' || txParams.s !== '0x') {
     txParams.v = toType(txParams.v, TypeOutput.BigInt)
   }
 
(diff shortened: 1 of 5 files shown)


holgerd77 (approved):
Nice simplified version of this, LGTM!

jochem-brouwer:
That was quick :tada: :smile:

holgerd77:
Rebased this via UI

jochem-brouwer:
This one is ready for merge, so if someone could (re-)review? 😄

g11tech (approved):
lgtm!

packages/tx/src/fromRpc.ts
```diff
@@ -28,7 +28,7 @@ export const normalizeTxParams = (_txParams: any): TxData => {
   txParams.r = txParams.r === '0x0' ? '0x' : txParams.r
   txParams.s = txParams.s === '0x0' ? '0x' : txParams.s
 
-  if (txParams.v !== '0x') {
+  if (txParams.v !== '0x' || txParams.r !== '0x' || txParams.s !== '0x') {
```
holgerd77, line 31: Since I am now an Optimism expert (lol): Optimism system transactions (aka "deposited transactions", I would very much say these are synonyms, might need some confirmation though) have their own tx type (`0x7E` respectively `126`), so might be the cleaner option here to do this check based on a tx type check.

holgerd77:
A bit unfortunate but I will revert this "fix" along #2713, since this turned out to be too much of a hack. Yes, this works on the surface. Underneath this randomly instantiates a 1559 tx, having totally different properties than an Optimism [system tx](https://github.com/ethereum-optimism/optimism/blob/develop/specs/deposits.md#the-deposited-transaction-type).

While this might work for selected use cases, this will give totally unpredictable (and wrong) results on others.

Furthermore in the more recent blocks the tx type is added to Optimism system tx, and this leads our tx library to throw ("tx type not supported") anyhow. So the solution here is also not future proof at all.

PR #2707, opened 2023-05-17
tx: Normalize toJson for different tx types

g11tech:
While sharing a block via json with @acolytec3 it was noticed that jsons didnt have type info in many instances, althrough one could interpret the json type via other fields, but a more verbose `type` can certainly help in easy identification. 

This PR aims to add normalized base transaction json fields in every toJson as well as removes some serialization redundancy of these common fields

Also fixes `versionedHashes` serialization which didn't have prefixed 0x


jochem-brouwer (approved):
LGTM! Also interesting that we never reported `type` on the JSON :thinking: Great addition.

PR #2708, opened 2023-05-17
tx: De-sszify 4844 blob transaction

g11tech:
Implements

- https://github.com/ethereum/EIPs/pull/6985


g11tech:
> I'm on it. slightly_smiling_face
> 
> Do we still have SSZ usage somewhere or could/should we also remove the Util dependency for now? grimacing
> 
> (since we have the code we can easily re-add later once this is introduced)

sounds good, as per discord discussion will do a followup PR removing ssz as for now we don't need it anywhere and can re-add whenever required :slightly_smiling_face:

holgerd77 (approved):
LGTM

PR #2709, opened 2023-05-17
WIP - Rework CI for build jobs

acolytec3:
This PR attempts to rework our entire CI workflow so that all of the jobs that build the monorepo are consolidated into a single workflow where step 1 is a `build` that installs all dependencies, builds all of the packages and then caches them.  Then, each of the individual package workflows just `restore` that build image and run that package's tests and related workflows on it.
I also added one global `lint` job to run all of the lint jobs at once in its own step.


holgerd77:
Just revisting this here, cannot completely remember, what was the status of this?

Did I kill this off preliminary by being too sceptical about it? 😬 Or have have there been some technical problems.

Now when revisiting I find this actually pretty solid and not really sure anymore where my biggest doubts were. 😬 😬 😬 
(sorry for that eventually)

acolytec3:
> Just revisting this here, cannot completely remember, what was the status of this?
> 
> Did I kill this off preliminary by being too sceptical about it? 😬 Or have there been some technical problems?
> 
> Now when revisiting I find this actually pretty solid and not really sure anymore where my biggest doubts were. 😬 😬 😬 
> (sorry for that eventually)

I can certainly update it and see if it still works. I don't remember your specific concerns but I think it just got forgotten about.

holgerd77:
> I can certainly update it and see if it still works. I don't remember your specific concerns but I think it just got forgotten about.

👍🙂🙏

holgerd77:
Super-seeded by #3592

PR #2710, opened 2023-05-17
Fix 4844 test set

jochem-brouwer:
This PR fixes the 4844 test set and also will introduce some features which consumers will want at some point anyways (and we likely need them for building these blocks as well).

To test:

- Get the 4844 tests from here: https://github.com/ethereum/execution-spec-tests/releases/tag/v0.2.5
- Put the tests inside `./packages/ethereum-tests/BlockchainTests`
- Run `npm run test:blockchain -- --fork=Cancun`

Currently down from 73 to ~~46~~ ~~26~~ 0 failing tests.

TODO:

- [x] The `excessDataGas` field of the block header is a field which is calculated (just like the `baseFee`). Just like `baseFee`, we want an option to calculate this field if a parent header is given (or some method to do so) -> if we want this, in another PR
- [x] ~~Put `dataGasPerBlob * tx.numBlobs()` gas into upfront costs~~


packages/block/src/block.ts
```diff
@@ -564,13 +564,25 @@ export class Block {
     }
   }
 
+  private calcExcessDataGas(parentHeader: BlockHeader, newBlobs: number) {
```
acolytec3, line 567: I think this duplicates [this](https://github.com/ethereumjs/ethereumjs-monorepo/blob/e4f5a42cd403d624029b0d8cac58a832a3219c05/packages/block/src/helpers.ts#L91) and I think this makes more sense as a standalone function than as part of the block class since there's nothing intrinsic to the block class required for this function to work.

packages/evm/src/precompiles/14-kzg-point-evaluation.ts
```diff
@@ -37,14 +49,6 @@ export async function precompile14(opts: PrecompileInput): Promise<ExecResult> {
   const commitment = opts.data.subarray(96, 144)
   const kzgProof = opts.data.subarray(144, 192)
```
acolytec3, line 51: Why did you remove this check?  This is part of the EIP.

Also, if we do agree to remove this, we should also remove the associated EVM error for inputs greater than the moduls.

acolytec3 (changes requested):
Left a couple of questions that I think need addressing.

packages/evm/src/precompiles/14-kzg-point-evaluation.ts
jochem-brouwer: It is part of the EIP, but this check is actually done inside the C-KZG. If I keep it, then we have 26 failing tests. If I remove, it passes. (So our pre-input check which we have is wrong, it likely takes the wrong inputs to check against). I will remove the error message, indeed.

jochem-brouwer: image

I asked this sometime ago on R&D

packages/block/src/block.ts
jochem-brouwer: You are right, but that method is not called anywhere from inside the block package (which caused failing tests). It does not handle the edge case that parent block does not have 4844 activated and the block itself is the first 4844 block, will fix.

jochem-brouwer:
I addressed the comments and updated, ready for re-review @acolytec3

packages/evm/src/precompiles/14-kzg-point-evaluation.ts
g11tech: ahh yes, this was anyway wrong sice y and z were coming in little endian and here our bytesToBigInt is big endian, and verify from kzg has these checks as mentioned by Jochem so we can remove them and let underlying libs handle the correctness

packages/evm/src/precompiles/14-kzg-point-evaluation.ts
```diff
@@ -79,7 +83,6 @@ export async function precompile14(opts: PrecompileInput): Promise<ExecResult> {
 
   // Return value - FIELD_ELEMENTS_PER_BLOB and BLS_MODULUS as padded 32 byte big endian values
   const fieldElementsBuffer = setLengthLeft(bigIntToBytes(fieldElementsPerBlob), 32)
-  const modulusBuffer = setLengthLeft(bigIntToBytes(BLS_MODULUS), 32)
```
g11tech, line 82: nice to move this out to global :+1:

g11tech (approved):
looks good to me :+1:

packages/block/src/block.ts
jochem-brouwer: Have removed this method and instead used the helper one

jochem-brouwer:
Thanks @g11tech :smile: 

@acolytec3 Could you re-approve?

packages/evm/src/precompiles/14-kzg-point-evaluation.ts
```diff
@@ -29,6 +30,17 @@ export async function precompile14(opts: PrecompileInput): Promise<ExecResult> {
     )
   }
 
+  if (opts.gasLimit < gasUsed) {
+    if (opts._debug !== undefined) {
+      opts._debug(`KZG_POINT_EVALUATION (0x14) failed: OOG`)
+    }
+    return OOGResult(opts.gasLimit)
```
acolytec3, line 37: Can you add a quick test for this? Should help with coverage and should be easy enough to add? Otherwise, I can do it later today

packages/evm/src/precompiles/14-kzg-point-evaluation.ts
```diff
@@ -29,6 +30,17 @@ export async function precompile14(opts: PrecompileInput): Promise<ExecResult> {
...
+  }
+
+  if (opts.data.length !== 192) {
```
acolytec3, line 40: Same here. Can we add tests for this error condition?

acolytec3 (changes requested):
Looks good. Can we add test coverage on the precompile error conditions?

packages/evm/src/precompiles/14-kzg-point-evaluation.ts
jochem-brouwer: I didn't want to add those tests, since these are in the Cancun fixtures.

jochem-brouwer: But if you want we can add of course.

acolytec3: No, that's fine. Will approve

acolytec3 (approved):

packages/evm/src/precompiles/14-kzg-point-evaluation.ts
jochem-brouwer: Thanks!

PR #2712, opened 2023-05-19
4844 transition tests

jochem-brouwer:
This test introduces the "ShanghaiToCancunAtTime15k" fork network.

To run the specific 4844 tests, download https://github.com/ethereum/execution-spec-tests/releases/tag/v0.2.5 `fixtures-cancun.tar.gz` and put `eips/eip4844` folder inside `ethereum-tests`.

Then run using `npm run test:blockchain -- --fork=ShanghaiToCancunAtTime15k`

This PR also fixes the failing tests. Note: in order to run the tests, disable `        const msg = this._errorMsg('versioned hash does not start with KZG commitment version')
` inside the 4844 blob transaction file in the tx package. The tests were filled without the requirement of this versioned hash.

WIP


jochem-brouwer:
Ok this passes the tests now, but this PR is far from clean. The fixes are now rather dirty and I will further use this PR to cleanup the implementation. I also want to investigate how common switches to Cancun, then a Shanghai block is imported (common points to Cancun still) and then Shanghai block is used again. So somewhere we are not setting the common correctly to the right HF.

packages/block/src/header.ts
```diff
@@ -117,6 +117,20 @@ export class BlockHeader {
...
       }
     }
+    if (opts.common && number !== undefined && timestamp !== undefined) {
+      if (opts.common.isActivatedEIP(4844) && excessDataGas === undefined) {
+        // There are situations where the common points to the wrong HF. Clone the common and
+        // set the HF by timestamp to ensure we are at the right fork.
+        // Note, for instance in blockchain/db/manager in `getHeader` we use the common of the manager
+        // This could thus point to the wrong HF (for instance, at the Shanghai -> Cancun fork block)
+        // The HF will be set to Cancun, but the parent block should point to Shanghai, but instead points to Cancun.
+        const clone = opts.common?.copy()
```
jochem-brouwer, line 127: I am not sure if I am doing this right here. The common could also be instantiated correctly in `blockchain/db/manager.ts`, but this adds boilerplate code..? Not sure how to proceed here?

holgerd77: Ah, yes, this feels too hacky. We should very much try to avoid to add additional Common/HF hacks into the code.

Will give this some closer look.

holgerd77: Update: yes, we should definitely look what is happening here and why the HF is not set correctly.

So the blockchain DB manager [getHeader()](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/blockchain/src/db/manager.ts#L153) method *does* various preparations to set the HF correctly and uses the `opts.hardforkByTTD` option and everything.

And in Block header constructor `timestamp` e.g. [is used](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/header.ts#L186) for HF setting.

Again: so something is off.

We should remove the code fix below and analyze this along the failing scenario which uncovered here, happy to help!

holgerd77 (changes requested):
See other comment.

packages/block/src/header.ts
g11tech: yes from my understanding of the currrent code working, db/manager.ts sets the correct fork.
if opts have any hardforkby options set, the resulting header's common will have hf correctly set which can be used to validate against he value's expected length.

packages/block/src/helpers.ts
```diff
@@ -158,7 +164,7 @@ export const getDataGasPrice = (header: BlockHeader) => {
  * @returns the total data gas fee for a transaction assuming it contains `numBlobs`
  */
 export const calcDataFee = (numBlobs: number, parent: BlockHeader) => {
```
g11tech, line 166: this fn has been refactored in https://github.com/ethereumjs/ethereumjs-monorepo/pull/2750 as per latest specs and will need to be rebased

packages/block/src/helpers.ts
```diff
@@ -158,7 +164,7 @@ export const getDataGasPrice = (header: BlockHeader) => {
...
-  if (parent.excessDataGas === undefined) {
+  if (parent.excessDataGas === undefined && parent._common.isActivatedEIP(4844)) {
```
g11tech, line 167: i think this is already validated in constructor of parent so not sure why we need extra validation here

packages/blockchain/src/blockchain.ts
```diff
@@ -603,7 +603,9 @@ export class Blockchain implements BlockchainInterface {
     // TODO: Rethink how validateHeader vs validateBlobTransactions works since the parentHeader is retrieved multiple times
     // (one for each uncle header and then for validateBlobTxs).
     const parentBlock = await this.getBlock(block.header.parentHash)
-    block.validateBlobTransactions(parentBlock.header)
+    if (parentBlock._common.isActivatedEIP(4844)) {
```
g11tech, line 606:
```suggestion
    if (block._common.isActivatedEIP(4844)) {
```
parent might not be 4844

packages/block/src/header.ts
jochem-brouwer: `db/manager.ts` takes the common from `blockchain`, and the problem here is that if you put the Cancun fork block then the common is thus at Cancun, but if you retrieve the parent header, it should be Shanghai but it is Cancun (which then throws if you try to deserialize the RLP). 

Note that the Common / fork is not saved into the DB, only the RLP. (So upon typing this it would very make much sense to setup the right common in `db/manager.ts` and not here)

jochem-brouwer:
Follow-up by #2770

PR #2716, opened 2023-05-21
client: Extend newPayloadV3 for blob versioned hashes checks

g11tech:
Post blob tx conversion to RLP, CL's commitments check via versioned hashes is now passed over to EL with new payloads having the computed CL versioned hashes.

Implements:
- https://github.com/ethereum/execution-apis/pull/407

This PR implements the same

Depends on 
 - https://github.com/ethereumjs/ethereumjs-monorepo/pull/2708

and should be merged (post rebase) after #2708 is merged

### DIFF
--- a/packages/client/lib/rpc/modules/engine.ts
+++ b/packages/client/lib/rpc/modules/engine.ts
@@ -451,7 +451,9 @@ export class Engine {
    *      valid block in the branch defined by payload and its ancestors
    *   3. validationError: String|null - validation error message
    */
-  private async newPayload(params: [ExecutionPayload, Bytes32[]?]): Promise<PayloadStatusV1> {
+  private async newPayload(
+    params: [ExecutionPayload, (Bytes32[] | null)?]
+  ): Promise<PayloadStatusV1> {
     const [payload, versionedHashes] = params
     if (this.config.synchronized) {
       this.connectionManager.newPayloadLog()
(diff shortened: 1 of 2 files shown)


packages/client/lib/rpc/modules/engine.ts
```diff
@@ -467,6 +469,48 @@ export class Engine {
...
+    if (block._common.isActivatedEIP(4844)) {
+      let validationError: string | null = null
+      if (versionedHashes === undefined) {
+        validationError = `Error verifying versionedHashes: received none`
+      } else {
+        // Collect versioned hashes to match
+        const txVersionedHashes = []
+        for (const tx of block.transactions) {
+          if (tx instanceof BlobEIP4844Transaction) {
+            for (const vHash of tx.versionedHashes) {
```
holgerd77, line 481: maybe rename to `vHashes`?

holgerd77 (commented):
Some small comments on naming and doc strings

packages/client/lib/rpc/modules/engine.ts
g11tech: but this is individual hash being pushed to create a flat txVersionedHashes

holgerd77: Ah sorry, misread the code

holgerd77 (approved):
LGTM

PR #2722, opened 2023-05-23
Client: Rename lib to src Folder

### LINKED ISSUE #2199: Rename lib directory to src in client
This is a legacy inconsistency issue, we've mentioned it should be renamed in #2193

holgerd77:
Closes #2199


acolytec3 (approved):
Looks great!  I fixed the one linter bug in `client` so we should merge once CI passes.

PR #2724, opened 2023-05-23
Update `rustbn.js` to latest

acolytec3:
Updates `rustbn.js` to v0.3.0 and removes a hack in `node_modules` required to pass in hex strings.


holgerd77 (approved):
Ah, nice that this works, LGTM

PR #2728, opened 2023-05-25
Add fully resolved paths with file extensions for all internal imports

acolytec3:
Part 2 of our ESM support process adding fully resolved paths with file extensions to all internal imports

- Currently tests are all broken because `tape` can't find stuff


acolytec3:
The more I dig into this, the less confident I am we'll be able to do this without fully switching our default code environment to ESM.  The `tape` module is unable to resolve fully specified paths with file extensions for any of our code and I suspect it has something to do with the way ts-node is compiling stuff but I'm not 100% sure.  Oddly enough, the browser tests pass just fine.  I've opened an issue on the `tape` repo to look for some clarifications.

acolytec3:
Superseded by #2764

PR #2730, opened 2023-05-26
client: Fix double runs of the block execution

g11tech:
On the 4844 devnet5 a weird error was noticed:

1. on new payload the client runs `vmExecution.runWithoutSetBlock` first and 
  image
2. then on fcU once the block is part of canonical chain it runs `vmExecution.run` which runs block in an iterator loop
  image

Run on 1. gives `VALID` execution while 2nd fails with `invalid state root`

This PR aims to solve it:
Synopsis of the issue: the vmExecution's `run` and `runWithoutSetHead` can end up running the same block one after another leading to state manager's incorrect caches

This PR updates `clearCache` flag in runBlock if such a scenario comes up during client execution.

- [x] Add a testcase to repro
- [x] Fix the issue
- [x] Testing
  - [x] sync mainnet
  - [x] sync local post merge devnet
  - [x] beacon sync and stay sync in a devnet 

PS:
the issue is resolved with this diff which might give an idea about where the problem could be i.e. by running the block again in the `vm` copy of vmexecution

```diff
--- a/packages/client/src/execution/vmexecution.ts
+++ b/packages/client/src/execution/vmexecution.ts
@@ -353,7 +353,8 @@ export class VMExecution extends Execution {
                 }
                 const beforeTS = Date.now()
                 this.stats(this.vm)
-                const result = await this.vm.runBlock({
+                const ivm = await this.vm.copy()
+                const result = await ivm.runBlock({
                   block,
                   root: parentState,
                   clearCache: reorg ? true : false,
```

### DIFF
--- a/packages/blockchain/src/types.ts
+++ b/packages/blockchain/src/types.ts
@@ -63,6 +63,14 @@ export interface BlockchainInterface {
   getIteratorHead?(name?: string): Promise<Block>
 
   /**
+   * Set header hash of a certain `tag`.
+   * When calling the iterator, the iterator will start running the first child block after the header hash currently stored.
+   * @param tag - The tag to save the headHash to
+   * @param headHash - The head hash to save
+   */
+  setIteratorHead(tag: string, headHash: Uint8Array): Promise<void>
+
+  /**
    * Gets total difficulty for a block specified by hash and number
    */
   getTotalDifficulty?(hash: Uint8Array, number?: bigint): Promise<bigint>
--- a/packages/client/test/execution/vmexecution.spec.ts
+++ b/packages/client/test/execution/vmexecution.spec.ts
@@ -1,14 +1,18 @@
+import { Block } from '@ethereumjs/block'
 import { Blockchain } from '@ethereumjs/blockchain'
 import { Chain as ChainEnum, Common, Hardfork } from '@ethereumjs/common'
+import { bytesToHex } from '@ethereumjs/util'
 import { VM } from '@ethereumjs/vm'
 import * as tape from 'tape'
 
 import { Chain } from '../../src/blockchain'
 import { Config } from '../../src/config'
 import { VMExecution } from '../../src/execution'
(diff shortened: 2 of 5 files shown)


g11tech:
a lot of diff is because of the formatting changes by moving runWithLock a bit out, but the real PR work is till this commit: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2730/commits/810aa54bd49f0ef46c98f3dc76e595ba90d4655e

packages/blockchain/src/types.ts
```diff
@@ -62,6 +62,14 @@ export interface BlockchainInterface {
    */
   getIteratorHead?(name?: string): Promise<Block>
 
+  /**
+   * Set header hash of a certain `tag`.
+   * When calling the iterator, the iterator will start running the first child block after the header hash currently stored.
+   * @param tag - The tag to save the headHash to
+   * @param headHash - The head hash to save
+   */
+  setIteratorHead(tag: string, headHash: Uint8Array): Promise<void>
```
holgerd77, line 71: 👍 🙏

packages/client/test/execution/vmexecution.spec.ts
```diff
@@ -104,4 +108,94 @@ tape('[VMExecution]', async (t) => {
...
+
+    await blockchain.putBlock(block)
+    await execution.run()
+
+    const newHead = await blockchain.getIteratorHead()
+    t.equal(bytesToHex(block.hash()), bytesToHex(newHead.hash()), 'vm execution head should be on')
+
+    closeRPC(server)
+    t.end()
+  })
```
holgerd77, line 136: Nice new test case, VM set iterator head execution tests very much needed, great! 👍

holgerd77 (approved):
Lot of great smallish additions/clean-ups and nice partial VM execution refactoring, this is still very much needed to have improvements there!

Code looks good, and I guess execution-flow-code can also best be tested by the test cases we have + continued client runs, where I am at right now anyhow.

So will approve and merge. 🙂

Issue #2731, opened 2023-05-26
`@ethereumjs/wallet` integration TODO list


acolytec3:
- [x] initial integration of `wallet` into monorepo - #2729 
- [x] Replace all `buffer` references with `Uint8Array` #2739 
- [x] Integrate tests into CI #2738 
- [x] Update `ethereum-cryptography` dep to v2.0.0 #2737 
- [x] Clean up docs - #2732 
- [x] Update `ethers` devDependency to v6 - #2747
- [x] Ensure linter/prettier works correctly with `wallet`- partly finished by #2739 
  -Review the `eslint` rules I disabled in #2739 to see if they can be switched back on
- [x] Switch from `mocha` to `tape` in tests #2742 
- [ ] Remove conflicting types/node devdependency

Coollaitar:
Updated dependencies of ethereum-cryptography to v2.0.0 #2734 

- [x] Update ethereum-cryptography dep to v2.0.0

Coollaitar:
Can you please elaborate in which files `buffer` should be replaced with `Uint8Array`?

acolytec3:
> Can you please elaborate in which files `buffer` should be replaced with `Uint8Array`?

Anywhere `buffers ` are used in this new package, they need to be placed with `Uint8Arrays`.

Coollaitar:
> > Can you please elaborate in which files `buffer` should be replaced with `Uint8Array`?
> 
> Anywhere `buffers ` are used in this new package, they need to be placed with `Uint8Arrays`.

Okay, I will make a PR for that.

Coollaitar:
Updated dependencies of `ethereum-cryptography` to v2.0.0 https://github.com/ethereumjs/ethereumjs-monorepo/pull/2737

- [x] Update ethereum-cryptography dep to v2.0.0

gabrocheleau:
Can this be closed @acolytec3 ? Seems like all todos have been completed.

acolytec3:
We should also clean up the conflicting `types/node` devDependency. I don't think there's any actual work other than removing the ones from `wallet` but we should do this so we don't have duplicate conflicting dependencies in our repo.

gabrocheleau:
> We should also clean up the conflicting `types/node` devDependency. I don't think there's any actual work other than removing the ones from `wallet` but we should do this so we don't have duplicate conflicting dependencies in our repo.

[...]

PR #2734, opened 2023-05-28
Updated `ethereum-cryptography`dep to v2.0.0

Coollaitar:
Solves #2731 Updated dependencies of  `ethereum-cryptography`  to v2.0.0
- [x] Update `ethereum-cryptography` dep to v2.0.0


Coollaitar:
> Thanks for the PR. Can you update this in `package.json` and not `package-lock.json`? Then, ensure that the `[monorepo-root]packages/wallet/node_modules` does not include anything other than `ethers` and related types?

I will make an another PR

PR #2740, opened 2023-05-31
client: add another vm execution re-run test scenario

g11tech:
adds another test case as followup to
- https://github.com/ethereumjs/ethereumjs-monorepo/pull/2730


holgerd77:
Rebased this via UI

PR #2744, opened 2023-06-01
tx: Add serialized blob tx spec test

g11tech:
Adds a hardcoded serialized blob tx testcase for test and reference with other EL clients


holgerd77:
Rebased this via UI

holgerd77 (approved):
LGTM

PR #2746, opened 2023-06-01
tx: Update c-kzg to big endian implementation

g11tech:
c-kzg npm (and other kzg impls/bindings) have been updated to `big endian`, this PR aims to update the same

Ref
 - https://github.com/ethereum/consensus-specs/pull/3354
 

TODO:

~~- [ ] Update the setup file~~ confirmed setup file remain sames
- [x] Update and fix tests


holgerd77:
Have updated this via UI

packages/tx/test/eip4844.spec.ts
```diff
@@ -441,4 +441,4 @@ tape('Network wrapper deserialization test', async (t) => {
...
...000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```
holgerd77, line 444: Maybe this file gets more readable and we will save some disk space if we do something like `'00'.repeat(1000000)` or so. 😂

Won't make it a blocker though, but might really be worth to update on the sideline on one of the next updates.

holgerd77: (or is this then hindering others from other clients to take the data over? 🤔)

holgerd77 (approved):
LGTM

packages/tx/test/eip4844.spec.ts
g11tech: yea we can do that, but for now the idea was to make sure consensus with other clients
we can def cleanup now

packages/tx/test/eip4844.spec.ts
```diff
@@ -387,16 +387,16 @@ tape('Network wrapper deserialization test', async (t) => {
       chainId: '0x1',
       accessList: [],
       maxFeePerDataGas: '0x5f5e100',
-      versionedHashes: ['0x01317228841f747eac2b4987a0225753a4f81688b31b21192ad2d2a3f5d252c5'],
+      versionedHashes: ['0x01a638b488178191c51eb548b89ce6e144f897b250b8cba834a7e4accd1cf297'],
```
holgerd77, line 390: Hi @g11tech, just having another look at this PR for release notes. Out of interest: why is the changeset for this PR and this tx tests actually so small? 🤔

Wouldn't e.g. the following from above (just totally abitrary example) also change and therefore a lot of hash and the like comparisons?

```typescript
t.equal(
  bytesToHex(unsignedTx.getMessageToSign(true)),
  '8ce8c3544ca173c0e8dd0e86319d4ebfe649e15a730137a6659ba3a721a9ff8b',
  'produced the correct transaction hash'
)
```

Just wondering what "I am not getting" here. 🙂

g11tech: @holgerd77  the txMeta (see at L392) changes exactly what you been mentioning

holgerd77: 👍

PR #2749, opened 2023-06-02
tx: Add nethermind network blob tx in 4844 spec test for validation

g11tech:
Adds a blob tx generated from `nethermind` client for validation


holgerd77 (approved):
LGTM

PR #2750, opened 2023-06-02
block: Add data gas used and refactor cal excess data gas cal

g11tech:
Implements
 - https://github.com/ethereum/EIPs/pull/7062
 
as part of devnet 6 target spec


packages/block/src/block.ts
```diff
@@ -515,6 +515,12 @@ export class Block {
       }
     }
 
+    if (this._common.isActivatedEIP(4844) === true) {
+      if (dataGasUsed !== (this.header.dataGasUsed ?? BigInt(0))) {
```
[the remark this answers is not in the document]
g11tech: true, just to keep linter happy  can `!` here

packages/block/src/helpers.ts
```diff
@@ -152,16 +146,15 @@ export const getDataGasPrice = (header: BlockHeader) => {
 }
```
holgerd77, line 146: I am still a bit shaky here on all the changes from this whole price calculation, but if I grasp things correctly I think code docs and error msgs from the `getDataGasPrice()` helper method above with all this parent references might not be correct.

g11tech: ahh yes, will update, as the dataprice is no more based on parent, but the current header

packages/block/src/block.ts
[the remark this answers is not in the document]
g11tech: updated to direct comparision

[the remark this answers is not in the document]
holgerd77: Yeah, that's better, thanks

packages/block/test/eip4844block.spec.ts
```diff
@@ -49,6 +56,26 @@ tape('EIP4844 header tests', function (t) {
...
+            common: earlyCommon,
+          }
+        )
+      },
+      (err: any) => {
+        return (
+          err.message.toString() === 'data gas used can only be provided with EIP4844 activated'
+        )
+      },
+      'should throw when setting excessDataGas with EIP4844 not being activated'
```
acolytec3, line 76:
```suggestion
      'should throw when setting dataGasUsed with EIP4844 not being activated'
```

packages/vm/test/api/EIPs/eip-4844-blobs.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+      headerData: {
+        gasLimit: 0xffffn,
+      },
+    })
+
+    // Set up tx
+    const blobs = getBlobs('hello world')
+    const commitments = blobsToCommitments(blobs)
+    const versionedHashes = commitmentsToVersionedHashes(commitments)
+    const proofs = blobsToProofs(blobs, commitments)
```
acolytec3, line 69: Do you think it would be worthwhile creating a generic test utility method for creating blob transactions (like [here]https://github.com/ethereumjs/ethereumjs-monorepo/blob/6d28106ce098e0f445e33dad88f49b208e27d24a/packages/client/test/sim/simutils.ts#L362) but more generic)?  Or maybe we update the static constructor to let you pass in the data for one or more blobs and it generates all of it?  It's just a lot of boilerplate to create these blob transactions (the more I think about it).

acolytec3 (changes requested):

packages/block/src/header.ts
```diff
@@ -523,6 +534,24 @@ export class BlockHeader {
     return nextBaseFee
   }
 
+  /**
+   * Calculates the excess data gas for next (hopefully) post EIP 4844 block.
+   *
+   * Note: This function expects that it is only being called on a valid 4844 parent as it does not have
+   * access to the "current" block's common instance to verify if 4844 is active or not.
```
acolytec3, line 541:
```suggestion
```

This comment can be deleted since it doesn't apply to the refactored method.

acolytec3 (commented):
Left one additional comment (just cleaning up n/a docs) but this looks good!

packages/block/src/helpers.ts
```diff
@@ -152,16 +126,15 @@ export const getDataGasPrice = (header: BlockHeader) => {
...
 
 /**
- * Returns the total fee for data gas spent on `numBlobs` in the current/pending block
- * @param numBlobs
- * @param parent parent header of the current/pending block
- * @returns the total data gas fee for a transaction assuming it contains `numBlobs`
+ * Returns the total fee for data gas spent in the current/pending block
+ * @param header of the current/pending block
+ * @param numBlobs number of blobs in the transaction
```
acolytec3, line 131: Isn't this the `numBlobs` in the "block" and not the tx?

packages/block/src/helpers.ts
```diff
@@ -152,16 +126,15 @@ export const getDataGasPrice = (header: BlockHeader) => {
...
+ * @returns the total data gas fee for all
```
acolytec3, line 132:
```suggestion
 * @returns the total data gas fee for block
```

packages/block/src/helpers.ts
g11tech: it can be used for any (so will update), but in the EIP second arg is the `tx`

acolytec3: Ah, that makes sense, sorry.  I was thinking this helper had been added to `blockheader`.

packages/vm/test/api/EIPs/eip-4844-blobs.spec.ts
g11tech: yes thats a very good idea to have a static method in the tx

acolytec3: > yes thats a very good idea to have a static method in the tx OR we can modify fromTxData to just accept blobs string[] and just generate the rest?

Yup, that's more or less what I was thinking.  Let's tackle as a separate PR though.  Not crucial for this one.

acolytec3 (approved):
LGTM.  Lets merge pending CI passing.

holgerd77:
Great PR overall! 😃🎉❤️

PR #2752, opened 2023-06-05
Client: Fix PeerPool Memory Leak

holgerd77:
This fixes a memory leak in the client being triggered when a peer is banned who is then sending continued messages to our client. These messages are the not answered but added to the `messageQueue` in `BoundProtocol`. If the other client is staying connected and continues to send messages (which happens in practice) `messageQueue` is growing indefinitely.

Bildschirmfoto 2023-06-05 um 14 14 53

I have tested that the functionality both triggers and is not breaking anything by adding `console.log()` output to both the `server.ban()` method in the client and the then executed `rlpx.disconnect()` method in devp2p, see screenshot above, and then waiting for a respective event to trigger during client run.

I have also added a general reasonable guard for `messageQueue` to generally avoid this list to grow without bounds.

Ready for review. 🙂

### DIFF
--- a/packages/client/src/net/peerpool.ts
+++ b/packages/client/src/net/peerpool.ts
@@ -191,7 +191,7 @@ export class PeerPool {
     // Reconnect to peer after ban period if pool is empty
     this._reconnectTimeout = setTimeout(async () => {
       if (this.running && this.size === 0) {
-        await peer.server?.connect(peer.id)
+        await peer.connect()
         this.connected(peer)
       }
     }, maxAge + 1000)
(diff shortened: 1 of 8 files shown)


packages/client/src/net/server/rlpxserver.ts
```diff
@@ -188,6 +188,7 @@ export class RlpxServer extends Server {
       return false
     }
     this.dpt!.banPeer(peerId, maxAge)
+    this.rlpx!.disconnect(hexToBytes(peerId))
```
g11tech, line 191: :+1:

g11tech (approved):
lgtm :+1:

packages/client/src/net/peerpool.ts
```diff
@@ -191,7 +191,7 @@ export class PeerPool {
     // Reconnect to peer after ban period if pool is empty
     this._reconnectTimeout = setTimeout(async () => {
       if (this.running && this.size === 0) {
-        await peer.server?.connect(peer.id)
+        await peer.connect()
```
acolytec3, line 194: What does this do?  Looks like this is a noop added in `peer.ts`?

acolytec3 (commented):
Fix looks fine but need clarity on the noop for `peer.connect`.

packages/client/src/net/peerpool.ts
holgerd77: It was a noop before (see connect() method in server, now it's calling into the existing connect() from RlpxPeer.

g11tech:
rebase via UI

PR #2755, opened 2023-06-06
Add low overhead blob tx constructor option

acolytec3:
Adds a new optional `blobsData` property to the `fromTxData` constructor for blob transactions to simplify tx construction

### DIFF
--- a/packages/tx/src/eip4844Transaction.ts
+++ b/packages/tx/src/eip4844Transaction.ts
@@ -184,7 +184,10 @@ export class BlobEIP4844Transaction extends BaseTransaction<BlobEIP4844Transacti
         throw new Error('cannot have both raw blobs data and KZG commitments in constructor')
       }
       if (txData.versionedHashes !== undefined) {
-        throw new Error('cannot have both raw blobs data and encoded blobs in constructor')
+        throw new Error('cannot have both raw blobs data and versioned hashes in constructor')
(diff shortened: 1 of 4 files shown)


packages/tx/src/eip4844Transaction.ts
```diff
@@ -172,6 +176,28 @@ export class BlobEIP4844Transaction extends BaseTransaction<BlobEIP4844Transacti
...
+      if (txData.kzgCommitments !== undefined) {
+        throw new Error('cannot have both raw blobs data and KZG commitments in constructor')
+      }
+      if (txData.versionedHashes !== undefined) {
+        throw new Error('cannot have both raw blobs data and versioned hashes in constructor')
+      }
+      if (txData.kzgProofs !== undefined) {
+        throw new Error('cannot have both raw blobs data and KZG proofs in constructor')
+      }
+      txData.blobs = getBlobs(txData.blobsData.reduce((acc, cur) => acc + cur))
```
g11tech, line 192: nice, so blobsData can be any length array  that needs to be chunked into blobs!

g11tech (approved):
lgtm

holgerd77:
The good thing about writing release notes is that you finally *really* need to grasp what is going on! 😋

This is so cool here 🎉, never realized this to full extend before!

PR #2756, opened 2023-06-06
tx: Update the kzg valiation and replace trusted setup with latest

g11tech:
- Updated the kzg validation
- updating  https://github.com/ethereum/c-kzg-4844/blob/main/src/trusted_setup.txt as the devnet6 setup as c-kzg team indicated that the trusted setup encoding has been updated (and new trusted setup is also different from the devnet4)
- Also adds a cli option to specify trusted setup for client run  (option has been requested by EF devops for testing ease)


acolytec3 (approved):
Lgtm

holgerd77:
Oh, nice, only now learning about this new `--trustedSetup` option, cool! 😎 😝

PR #2758, opened 2023-06-06
Make `libp2p` optional

acolytec3:
Moves `libp2p` dependencies in `client` to optional peer  dependencies and makes necessary tweaks to ensure Typescript doesn't complain about them missing.


packages/client/libp2pBrowserBuild/util/index.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * @module util
+ */
+import { bytesToPrefixedHexString } from '@ethereumjs/util'
```
g11tech, line 4: we can use client's util, it has browser handling, no reason to duplicate i think

packages/client/src/net/protocol/index.ts
```diff
@@ -6,7 +6,7 @@ export * from './boundprotocol'
 export * from './ethprotocol'
 export * from './flowcontrol'
 export * from './lesprotocol'
-export * from './libp2psender'
+// export * from '../../../browser/net/libp2psender'
```
g11tech, line 9: should cleanup these comments?

acolytec3: Done

acolytec3:
To test this and verify the browser client works, do:
1. `npm run build:browser` from client package
2. `echo '<script src="/dist/bundle.js"></script>' > index.html`
3. `npx http-server`
4. Open a browser window and point to 127.0.0.1
5. Open the developer console and run 
```js
ethereumjs.run({
  network: 'goerli',
  syncmode: 'none',
})
```

You should see the client start up with some logs that look like this:
```
Full sync mode
bundle.js:2 Initializing account cache size=400000
bundle.js:2 Initializing storage cache size=200000
bundle.js:2 Initializing trie cache size=200000
bundle.js:2 Initializing Ethereumjs client version=v0.7.1 network=goerli chainId=5
bundle.js:2 Starting FullEthereumService with no syncing.
bundle.js:2 Initializing VM execution hardfork=petersburg
bundle.js:2 Setup EVM execution.
bundle.js:2 Setup networking and services.
bundle.js:2 Started eth service.
bundle.js:2 Skipped execution run vmHead=0 canonicalHead=0 hardfork=petersburg execution=true
```

g11tech (approved):
lgtm

PR #2762, opened 2023-06-07
client: fix the newPayloadV3 validations

g11tech:
fix newPayloadV3 validation errors as uncovered in devnet6

### DIFF
--- a/packages/client/src/rpc/modules/engine.ts
+++ b/packages/client/src/rpc/modules/engine.ts
@@ -693,23 +693,33 @@ export class Engine {
         }
       }
     } else if (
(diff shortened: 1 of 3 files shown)


holgerd77 (approved):
LGTM, wonder if this way we are going the validations here is so future proof, but anyhow: good for now.

Also we should ramp up on a good HF comparison method my timestamp (that's the future [tm]).

Will merge.

g11tech:
I have been asking to make these methods 1:1 (i.e. v3 for cancun, v2 for shanghai, v1 for paris only) i.e. no backward forward compatibility, and then we only need to check for Hardfork, rest fields are being validated by our `validators` module.

However this backward forrward compatibility forces us with this kind of conditions.

PR #2763, opened 2023-06-08
client: Fixes for block and blob building uncovered in devnet6

g11tech:
Couple of issues discovered for block and blob building in devnet6:

- excessDataGas was always being set to 0 
- blobs were not included for a blob tx causing versionedhash , to blobs mismatch

This PR fixes the same


holgerd77 (approved):
LGTM

PR #2765, opened 2023-06-08
vm: Discard blob txs with missing blobs for block building

g11tech:
makes sure that a block is not build with missing blobs (i.e. from non network wrapper tx)

TODO:
- [x] reject tx with missing blob in build block
~~- [ ] ensure same in gossip~~ TO be dealt in a different PR after discussion
- [x] recheck in rpc submission
- [x] enhance coverage

### DIFF
--- a/packages/client/src/miner/miner.ts
+++ b/packages/client/src/miner/miner.ts
@@ -2,7 +2,7 @@ import { BlockHeader } from '@ethereumjs/block'
 import { ConsensusType, Hardfork } from '@ethereumjs/common'
 import { Ethash } from '@ethereumjs/ethash'
 import { bytesToPrefixedHexString } from '@ethereumjs/util'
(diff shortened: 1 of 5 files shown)


packages/client/src/miner/miner.ts
```diff
@@ -325,18 +325,6 @@ export class Miner {
               `Miner: Assembled block full (gasLeft: ${gasLimit - blockBuilder.gasUsed})`
             )
           }
-        } else if ((error as Error).message.includes('tx has a different hardfork than the vm')) {
```
g11tech, line 328: this is a case no longer required as we stopped throwing on tx have different hardfork than vm, but instead set it to vm's if vm's and block's hf match

gabrocheleau (approved):

gabrocheleau:
Looks good to me, can be merged once tests pass.

Issue #2726, opened 2023-05-25
Tx: Replace TypedTransaction type with generic Interface


holgerd77:
When working on improving Optimism Bedrock compatibiliy in #2713 I realized that we still have got this unholy [TypedTransaction](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/types.ts#L100) type being used as a generic basis for tx input in the upper libraries, e.g. [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/block.ts#L371) as constructor option typing in the `Block` library.

This is how this type looks like, so it is just a junction of all tx types we have got implemented so far.

```typescript
export type TypedTransaction =
  | Transaction
  | AccessListEIP2930Transaction
  | FeeMarketEIP1559Transaction
  | BlobEIP4844Transaction
```

This is very much not sustainable and severly limits the possibility to use custom tx types in our other libraries, particularly the `Block`, `Blockchain` and `VM` libraries, `Client` is a bit more secondary but should be updated nevertheless.

This type should be completely removed and updated with a proper interface as we have for the StateManager (recently moved and now in the `@ethereumjs/common` package [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/interfaces.ts#L41)).

This interface should expose a common set of shared methods, randomly citing `raw()`, `isSigned()`, `hash()` as safe candidates - optimally the (shared) static constructors like `fromValuesArray()` as well (not sure, is this possible?) - but safe to say various others as well.

Larger part of typing in the mentioned upstream libraries should then be updated to use this new interface, basically all use cases where things should generically work for the various tx types.

Here are the most prominent update cases I can think of right now:

- As mentioned, constructor options (e.g. [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/block.ts#L371)) for the `Block` library
- VM [runTx options](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/types.ts#L317), various others
- VM [block builder](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/buildBlock.ts)
- Client: everything in [tx pool](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/src/service/txpool.ts), various others

[...]

PR #2767, opened 2023-06-09
tx: Generic Transaction Interface

gabrocheleau:
Resolves issue #2726 by implementing a generic Transaction interface and additional types/interfaces for related ones (e.g. `TxData`, `TxValuesArray`)

### Notable changes

- Renamed `Transaction` to `LegacyTransaction`. There were some comments in the repo that the `Transaction` naming was kept for retro-compatibility, so not sure if we can allow ourselves to rename it for this breaking release.
- Added a `TransactionType` enum for the 4 transaction types
- Added a `TransactionInterface` (similar to the state manager interface) that specifies the methods that should be implement by all transaction classes & uses generics to significantly improve type inference. 
- The `BaseTransaction` abstract class now implements `TransactionInterface` 
- Added a `Transaction[TransactionType], TxData[Transaction[TransactionType], TxValuesArray[TransactionType]` interfaces that can be used to map a transaction type to its corresponding data types.
- Added TypeGuard methods to verify if Tx/TxData belongs to a certain TxType (e.g. `isLegacyTx(txData: TypedTransaction): tx is LegacyTransaction`)

### DIFF
--- a/packages/client/src/net/protocol/ethprotocol.ts
+++ b/packages/client/src/net/protocol/ethprotocol.ts
@@ -79,9 +79,7 @@ export interface EthProtocolMethods {
 }
 
 function exhaustiveTypeGuard(_value: never): never {
-  throw new Error(
-    `ERROR! Reached forbidden guard function with unexpected value: ${JSON.stringify(_value)}`
-  )
+  throw new Error(`ERROR! Reached forbidden guard function with unexpected value`)
 }
 
 /**
--- a/packages/client/src/net/protocol/ethprotocol.ts
+++ b/packages/client/src/net/protocol/ethprotocol.ts
@@ -4,7 +4,9 @@ import { RLP } from '@ethereumjs/rlp'
 import {
   BlobEIP4844Transaction,
   TransactionFactory,
+  isAccessListEIP2930Tx,
   isBlobEIP4844Tx,
+  isFeeMarketEIP1559Tx,
   isLegacyTx,
 } from '@ethereumjs/tx'
 import {
@@ -74,6 +76,12 @@ export interface EthProtocolMethods {
   getBlockBodies: (opts: GetBlockBodiesOpts) => Promise<[bigint, BlockBodyBytes[]]>
   getPooledTransactions: (opts: GetPooledTransactionsOpts) => Promise<[bigint, TypedTransaction[]]>
   getReceipts: (opts: GetReceiptsOpts) => Promise<[bigint, TxReceipt[]]>
+}
+
+function exhaustiveTypeGuard(_value: never): never {
+  throw new Error(
+    `ERROR! Reached forbidden guard function with unexpected value: ${JSON.stringify(_value)}`
+  )
 }
 
 /**
@@ -224,18 +232,17 @@ export class EthProtocol extends Protocol {
       encode: ({ reqId, txs }: { reqId: bigint; txs: TypedTransaction[] }) => {
         const serializedTxs = []
         for (const tx of txs) {
-          if (isLegacyTx(tx)) {
-            serializedTxs.push(tx.raw())
-            break
-          }
-
+          // serialize txs as per type
           if (isBlobEIP4844Tx(tx)) {
             serializedTxs.push(tx.serializeNetworkWrapper())
-            break
+          } else if (isFeeMarketEIP1559Tx(tx) || isAccessListEIP2930Tx(tx)) {
+            serializedTxs.push(tx.serialize())
+          } else if (isLegacyTx(tx)) {
+            serializedTxs.push(tx.raw())
+          } else {
+            // type gurard to force typescript to throw build errors if any type is missed above
+            exhaustiveTypeGuard(tx)
           }
-
-          serializedTxs.push(tx.serialize())
-          break
         }
 
         return [bigIntToUnpaddedBytes(reqId), serializedTxs]
--- a/packages/client/src/net/protocol/ethprotocol.ts
+++ b/packages/client/src/net/protocol/ethprotocol.ts
@@ -1,7 +1,12 @@
 import { Block, BlockHeader, getDifficulty, valuesArrayToHeaderData } from '@ethereumjs/block'
 import { Hardfork } from '@ethereumjs/common'
 import { RLP } from '@ethereumjs/rlp'
-import { BlobEIP4844Transaction, TransactionFactory } from '@ethereumjs/tx'
+import {
+  BlobEIP4844Transaction,
+  TransactionFactory,
+  isBlobEIP4844Tx,
+  isLegacyTx,
+} from '@ethereumjs/tx'
 import {
   bigIntToUnpaddedBytes,
   bytesToBigInt,
@@ -17,7 +22,7 @@ import type { TxReceiptWithType } from '../../execution/receipt'
 import type { Message, ProtocolOptions } from './protocol'
 import type { BlockBodyBytes, BlockBytes, BlockHeaderBytes } from '@ethereumjs/block'
 import type { Log } from '@ethereumjs/evm'
-import type { TransactionsArray } from '@ethereumjs/tx'
+import type { TypedTransaction } from '@ethereumjs/tx'
 import type { BigIntLike, NestedUint8Array } from '@ethereumjs/util'
 import type { PostByzantiumTxReceipt, PreByzantiumTxReceipt, TxReceipt } from '@ethereumjs/vm'
 
@@ -67,7 +72,7 @@ type GetReceiptsOpts = {
 export interface EthProtocolMethods {
   getBlockHeaders: (opts: GetBlockHeadersOpts) => Promise<[bigint, BlockHeader[]]>
   getBlockBodies: (opts: GetBlockBodiesOpts) => Promise<[bigint, BlockBodyBytes[]]>
-  getPooledTransactions: (opts: GetPooledTransactionsOpts) => Promise<[bigint, TransactionsArray]>
+  getPooledTransactions: (opts: GetPooledTransactionsOpts) => Promise<[bigint, TypedTransaction[]]>
   getReceipts: (opts: GetReceiptsOpts) => Promise<[bigint, TxReceipt[]]>
 }
 
@@ -91,7 +96,7 @@ export class EthProtocol extends Protocol {
     {
       name: 'Transactions',
       code: 0x02,
-      encode: (txs: TransactionsArray) => {
+      encode: (txs: TypedTransaction[]) => {
         const serializedTxs = []
         for (const tx of txs) {
           // Don't automatically broadcast blob transactions - they should only be announced using NewPooledTransactionHashes
@@ -216,21 +221,23 @@ export class EthProtocol extends Protocol {
     {
       name: 'PooledTransactions',
       code: 0x0a,
-      encode: ({ reqId, txs }: { reqId: bigint; txs: TransactionsArray }) => {
+      encode: ({ reqId, txs }: { reqId: bigint; txs: TypedTransaction[] }) => {
         const serializedTxs = []
         for (const tx of txs) {
-          switch (tx.type) {
-            case 0:
-              serializedTxs.push(tx.raw())
-              break
-            case 3:
-              serializedTxs.push((tx as BlobEIP4844Transaction).serializeNetworkWrapper())
-              break
-            default:
-              serializedTxs.push(tx.serialize())
-              break
+          if (isLegacyTx(tx)) {
+            serializedTxs.push(tx.raw())
+            break
           }
+
+          if (isBlobEIP4844Tx(tx)) {
+            serializedTxs.push(tx.serializeNetworkWrapper())
+            break
+          }
+
+          serializedTxs.push(tx.serialize())
+          break
         }
+
         return [bigIntToUnpaddedBytes(reqId), serializedTxs]
       },
       decode: ([reqId, txs]: [Uint8Array, any[]]) => {
--- a/packages/client/src/net/protocol/ethprotocol.ts
+++ b/packages/client/src/net/protocol/ethprotocol.ts
@@ -78,8 +78,8 @@ export interface EthProtocolMethods {
   getReceipts: (opts: GetReceiptsOpts) => Promise<[bigint, TxReceipt[]]>
 }
 
-function exhaustiveTypeGuard(_value: never): never {
-  throw new Error(`ERROR! Reached forbidden guard function with unexpected value`)
+function exhaustiveTypeGuard(_value: never, errorMsg: string): never {
+  throw new Error(errorMsg)
 }
 
 /**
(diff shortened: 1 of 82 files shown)


packages/tx/src/baseTransaction.ts
```diff
@@ -46,7 +43,9 @@ interface TransactionCache {
  *
  * It is therefore not recommended to use directly.
  */
-export abstract class BaseTransaction<TransactionObject> {
+export abstract class BaseTransaction<TTransactionType extends TransactionType>
```
g11tech, line 46:
```suggestion
export abstract class BaseTransaction<T extends TransactionType>
```
just using `T` should be sufficient and understandable for generics while keeping it short

packages/tx/src/eip2930Transaction.ts
```diff
@@ -138,7 +145,7 @@ export class AccessListEIP2930Transaction extends BaseTransaction<AccessListEIP2
    * the static factory methods to assist in creating a Transaction object from
    * varying data types.
    */
-  public constructor(txData: AccessListEIP2930TxData, opts: TxOptions = {}) {
+  public constructor(txData: TxData[TransactionType.AccessListEIP2930], opts: TxOptions = {}) {
```
g11tech, line 148: I would like to define all ` ...[TransactionType.AccessListEIP2930]` references on top of each transaction class file as:
```
type TxData = AllTypesTxData[TransactionType.AccessListEIP2930]
```
and then use just `TxData`, `TxValuesArray` etc etc in the class def for brevity

packages/tx/src/types.ts
```diff
@@ -93,20 +95,91 @@ export function isAccessList(input: AccessListBytes | AccessList): input is Acce
...
+}
+
+export interface TxData {
+  [TransactionType.Legacy]: LegacyTxData
+  [TransactionType.AccessListEIP2930]: AccessListEIP2930TxData
+  [TransactionType.FeeMarketEIP1559]: FeeMarketEIP1559TxData
+  [TransactionType.BlobEIP4844]: BlobEIP4844TxData
+}
+
+export type UnknownTxData =
```
g11tech, line 154: may be you can do this: `TxData[TransactionType]` and should give the same meaning

also would not like to call this as `UnknownTxData` but `TypedTransaction` was a better name imo

packages/tx/src/types.ts
```diff
@@ -93,20 +95,91 @@ export function isAccessList(input: AccessListBytes | AccessList): input is Acce
...
+  isSigned(): boolean
+  verifySignature(): boolean
+  getSenderAddress(): Address
+  getSenderPublicKey(): Uint8Array
+  sign(privateKey: Uint8Array): Transaction[TTransactionType]
+  toJSON(): JsonTx
+  errorStr(): string
+}
+
+export interface TxData {
```
g11tech, line 147: like this idea very much :heart:

packages/tx/src/types.ts
```diff
@@ -93,20 +95,91 @@ export function isAccessList(input: AccessListBytes | AccessList): input is Acce
...
+  const txType = Number(bytesToBigInt(toBytes(txData.type)))
+  return txType === 0
+}
+export function isAccessListEIP2930TxData(
+  txData: UnknownTxData
+): txData is AccessListEIP2930TxData {
+  const txType = Number(bytesToBigInt(toBytes(txData.type)))
+  return txType === 1
+}
+export function isFeeMarketEIP1559TxData(txData: UnknownTxData): txData is FeeMarketEIP1559TxData {
```
g11tech, line 170: very nice typeguards :clap:

packages/tx/src/types.ts
```diff
@@ -93,20 +95,91 @@ export function isAccessList(input: AccessListBytes | AccessList): input is Acce
...
+
+export function isLegacyTxData(txData: UnknownTxData): txData is LegacyTxData {
+  const txType = Number(bytesToBigInt(toBytes(txData.type)))
+  return txType === 0
+}
+export function isAccessListEIP2930TxData(
+  txData: UnknownTxData
+): txData is AccessListEIP2930TxData {
+  const txType = Number(bytesToBigInt(toBytes(txData.type)))
+  return txType === 1
```
g11tech, line 168: Instead of 1,2,3 num comparisions, can do we enum `TransactionType` in a proper way assinging the type numbers to it explicitly and then use the enum for comparision.

packages/tx/src/types.ts
```diff
@@ -93,20 +95,91 @@ export function isAccessList(input: AccessListBytes | AccessList): input is Acce
...
+  [TransactionType.BlobEIP4844]: BlobEIP4844Transaction
+}
+
+export type UnknownTransaction =
+  | Transaction[TransactionType.Legacy]
+  | Transaction[TransactionType.AccessListEIP2930]
+  | Transaction[TransactionType.FeeMarketEIP1559]
+  | Transaction[TransactionType.BlobEIP4844]
+
+export type TransactionsArray = Array<UnknownTransaction>
```
g11tech, line 119: i think the `TypedTransaction[]` is more meaningful and code intuitive than `TransactionsArray` and would like to retain its usage even in the new refactored form

packages/tx/src/types.ts
gabrocheleau: Good point, will do!

gabrocheleau: Done! Also added additional typeguards and replaced a switch statement with that.

packages/tx/src/baseTransaction.ts
gabrocheleau: Good point, done.

packages/tx/src/types.ts
holgerd77: Yes, would agree here.

gabrocheleau: Renamed to TypedTransaction & TypedTransaction[]

packages/tx/src/types.ts
gabrocheleau: Oh indeed, nice. Implemented and updated the naming to TypedTxData

packages/tx/src/eip2930Transaction.ts
gabrocheleau: Good idea, done!

packages/client/src/net/protocol/ethprotocol.ts
```diff
@@ -219,18 +224,20 @@ export class EthProtocol extends Protocol {
...
-              serializedTxs.push(tx.serialize())
-              break
+          if (isLegacyTx(tx)) {
+            serializedTxs.push(tx.raw())
+            break
+          }
+
+          if (isBlobEIP4844Tx(tx)) {
+            serializedTxs.push(tx.serializeNetworkWrapper())
+            break
```
g11tech, line 234: break here will break the for loop, continue is better , but overall it should be chained in if else if...

g11tech: updated

g11tech (approved):

packages/client/src/net/protocol/ethprotocol.ts
gabrocheleau: Thanks!

g11tech (approved):

gabrocheleau:
Thanks for approving @g11tech! I'll wait for @holgerd77 's final approval to merge to make sure this PR properly addresses the issue he had created. (#2726)

packages/tx/src/eip1559Transaction.ts
```diff
@@ -51,7 +57,7 @@ export class FeeMarketEIP1559Transaction extends BaseTransaction<FeeMarketEIP155
    *
    * @hidden
    */
-  protected DEFAULT_HARDFORK = 'london'
+  protected DEFAULT_HARDFORK = Hardfork.London
```
holgerd77, line 60: Uhg.

We still have got "London" in here? 🤔

So this is definitely a left over. Just had a look to (re-)get the whole picture, some context: we introduced this "sub"-default HF for tx types which are not yet activated on the current broader default HF (so historically these were actually not "back" but "forward" 😋).

So for 1559 (London) and 2930 (Berlin) we should definitely remove completely.

We can eventually think if we rather want to add for 4844 and set to "Cancun" (so *that* was actually the original intended use case 😋).

On thinking about I would rather think we do not want to do though, seems to have a tendency to get forgotten and think we should be fine with the concrete ´this.common.isActivatedEIP(4844)` capability/EIP checks in the constructor (which is more tx-type-adequate anyhow).

Nevertheless, so: we should remove these two sub-default-HFs before breaking releases!

packages/tx/src/types.ts
```diff
@@ -93,20 +95,98 @@ export function isAccessList(input: AccessListBytes | AccessList): input is Acce
...
+  return tx.type === TransactionType.BlobEIP4844
+}
+
+export interface TransactionInterface<T extends TransactionType> {
+  supports(capability: Capability): boolean
+  type: number
+  validate(): boolean
+  validate(stringError: false): boolean
+  validate(stringError: true): string[]
+  validate(stringError: boolean): boolean | string[]
```
holgerd77, line 137: Ah, we should urgently have a look there and see that we align this typing to a single version, pretty sure (not 100% though, so this would need some confirmation) these overload-typings have been introduced for some backwards-compatibility reasons.

packages/tx/src/types.ts
```diff
@@ -93,20 +95,98 @@ export function isAccessList(input: AccessListBytes | AccessList): input is Acce
...
+  getBaseFee(): bigint
+  getDataFee(): bigint
+  getUpfrontCost(): bigint
+  toCreationAddress(): boolean
+  raw(): TxValuesArray[T]
+  serialize(): Uint8Array
+  getMessageToSign(hashMessage: false): Uint8Array | Uint8Array[]
+  getMessageToSign(hashMessage?: true): Uint8Array
```
holgerd77, line 145: Same here.

If I remember correctly there should be some valuable and extensive additional context in the comments for these methods.

holgerd77 (approved):
🤯

Oh oh oh.

I honestly didn't expect something so wonderful come out of this task in such an ad hoc manner!!! 🎉

Thanks so so much, all these simplifications are a delight to look at, throughout the whole PR. 🙏

Two remarks, but rather very much unrelated and for a follow-up PR, either for you or someone else.

Also, just to confirm respectively to keep some confirmative discussion going a bit: so one of the larger goals for this task was that it I guess both gets easier a) for people in the tx library itself to implement their own tx type and as well b) *use* this tx type in our other libraries.

For a) the direct question: how much did you have this in mind and would you think the structure is better prepared for this now? In particular, is this tx type enum eventually somewhere in the way to do so, or is this well integrated/expandable?

For b) I guess we need to have some differentiated look anyhow: so for VM this is (at least for now and for the short-term future) very much "bound" regarding compatibility to the existing tx types (generally interesting discussion though if this *could* theoretically be generalized, will leave this here though, going too far).

So I guess we can start the compatibility question on Block - eventually Blockchain (but rather Block for now) - question. Is Block now (better) prepared to eventually take a custom tx type? 🤔

Still thinking a bit loud myself: we likely also *do not want* to have Block accept all sorts of types e.g. for a predefined chain (so e.g. for a mainnet Cancun chain accept Optimism system txs).

I wonder if would want to just explicitly add the accepted tx types to chain files in Common, so that this gets:

```json
{
  "name": "london",
  "block": 12965000,
  "txTypes": [0, 1, 2],
  "forkHash": "0xb715077d"
}
```

This would make things more consistent. Wouldn't be much to add (likely also more clean for our own checks) and then people could define their own chain files with additional type numbers, implement their tx type and block and blockchain validations can then likely (?) be written to check for this respective types in Common.

Anyhow: this is going a lot further and up for a separate discussion! 😀

Again, great work!!!

Will merge.

packages/tx/src/eip1559Transaction.ts
gabrocheleau: Sure, will push a PR for that!

gabrocheleau:
> exploding_head
> 
> Oh oh oh.
> 
> I honestly didn't expect something so wonderful come out of this task in such an ad hoc manner!!! tada
> 
> Thanks so so much, all these simplifications are a delight to look at, throughout the whole PR. pray
> 
> Two remarks, but rather very much unrelated and for a follow-up PR, either for you or someone else.
> 
> Also, just to confirm respectively to keep some confirmative discussion going a bit: so one of the larger goals for this task was that it I guess both gets easier a) for people in the tx library itself to implement their own tx type and as well b) _use_ this tx type in our other libraries.
> 
> For a) the direct question: how much did you have this in mind and would you think the structure is better prepared for this now? In particular, is this tx type enum eventually somewhere in the way to do so, or is this well integrated/expandable?
> 
> For b) I guess we need to have some differentiated look anyhow: so for VM this is (at least for now and for the short-term future) very much "bound" regarding compatibility to the existing tx types (generally interesting discussion though if this _could_ theoretically be generalized, will leave this here though, going too far).
> 
> So I guess we can start the compatibility question on Block - eventually Blockchain (but rather Block for now) - question. Is Block now (better) prepared to eventually take a custom tx type? thinking
> 
> Still thinking a bit loud myself: we likely also _do not want_ to have Block accept all sorts of types e.g. for a predefined chain (so e.g. for a mainnet Cancun chain accept Optimism system txs).
> 
> I wonder if we would want to just explicitly add the accepted tx types to chain files in Common, so that this gets:
> 
> ```json
> {
>   "name": "london",
>   "block": 12965000,
>   "txTypes": [0, 1, 2],
>   "forkHash": "0xb715077d"
> }
> ```
> 
> This would make things more consistent. Wouldn't be much to add (likely also more clean for our own checks) and then people could define their own chain files with additional type numbers, implement their tx type and block and blockchain validations can then likely (?) be written to check for this respective types in Common.
> 
> Anyhow: this is going a lot further and up for a separate discussion! grinning
> 
> Again, great work!!!
> 
> Will merge.

Hey @holgerd77!

Thanks for the detailed comments and reflections!

Regarding a) and b) 

I did not directly focus on expandability with this PR, but rather more on refactoring and improving types. However, I think this PR still brings us closer to enabling custom a) (custom tx types). 

I'm not actually sure about the enum itself, makes me wonder if we could also add a TransactionType.Custom (?), although I'm not sure how helpful that would be. One nice thing about this updated structure is that a user that would add a 4th transaction type to the enum (e.g. `TransactionType.Optimism`) would then get a cascade of type issues throughout the repo allowing them to easily spot instances where some implementation is missing (e.g. switch statements that are missing the case for that 4th txtype). They could also easily create a new Transaction class that extends BaseTransaction, and therefore comes with the constraints of the TransactionInterface (just like when implementing a new StateManager). 

I'm not sure how much ease of extensibility we can realistically aim for with other packages though. As you imply in your comments regarding b), txs are quite a fundamental building block, and new tx types would I assume come with all sorts of idiosyncratic behavior that affect blocks, vm, etc. I feel like there is a limit to how much generalization we can do. As a thought experiment, pre-4844, how could we have generalized Txs in a way that allowed for 4844 transactions? Some of these are bound to be so "outside the box" that I can hardly see how we could generalize transactions in a way that's robust and future-proof.

Really interesting thing to reflect on though! We could eventually add an example of implementing a custom tx type, just like we do for some of the other packages.

PR #2769, opened 2023-06-10
block: add missing tx type

gabrocheleau:
This adds a missing 4844 txData type to the block `transactions` type

### DIFF
--- a/packages/block/src/types.ts
+++ b/packages/block/src/types.ts
@@ -2,6 +2,7 @@ import type { BlockHeader } from './header'
 import type { Common } from '@ethereumjs/common'
 import type {
   AccessListEIP2930TxData,
+  BlobEIP4844TxData,
   FeeMarketEIP1559TxData,
   JsonRpcTx,
   JsonTx,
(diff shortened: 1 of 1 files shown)


packages/block/src/types.ts
```diff
@@ -116,7 +117,9 @@ export interface BlockData {
    * Header data for the block
    */
   header?: HeaderData
-  transactions?: Array<TxData | AccessListEIP2930TxData | FeeMarketEIP1559TxData>
+  transactions?: Array<
+    TxData | AccessListEIP2930TxData | FeeMarketEIP1559TxData | BlobEIP4844TxData
```
jochem-brouwer, line 121: What we could also do here is, export this type (so all types combined) from our `tx` library, and then import that type here..? (As a slight refactor)

gabrocheleau: Agreed but I'm actually doing this WIP PR as part of a larger tx interfaces refactoring: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2767

jochem-brouwer: Allright :+1:

jochem-brouwer (approved):

PR #2733, opened 2023-05-27
Implement EIP6780: SELFDESTRUCT only in same transaction

jochem-brouwer:
Follow-up of #2612

Implements https://eips.ethereum.org/EIPS/eip-6780

This EIP is currently included in Cancun per https://github.com/ethereum/execution-specs/pull/763

Note: target branch is `cleanup-statemanager-v2` (PR #2702), do not merge yet. (But ready for review)

PR #2771, opened 2023-06-10
Implement EIP6780: SELFDESTRUCT only in same transaction (v2)

jochem-brouwer:
Follow-up of #2733 

Implements https://eips.ethereum.org/EIPS/eip-6780

This EIP is currently included in Cancun per https://github.com/ethereum/execution-specs/pull/763


packages/vm/test/api/EIPs/eip-6780-selfdestruct-same-tx.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+const GWEI = BigInt(1000000000)
+const sender = new Address(privateToAddress(pkey))
+
+const common = new Common({
+  chain: Chain.Mainnet,
+  hardfork: Hardfork.London,
+  eips: [6780],
+})
+
+const payload = hexToBytes('60016001556001FF')
```
scorbajio, line 19: A comment like "Payload includes `SELFDESTRUCT` `FF` opcode" above this line would make it more clear that the instruction is being executed as part of the transaction.

scorbajio (commented):
Overall, looks great!

scorbajio:
Also, not sure why the linter is failing the vm tests: It passes locally. Maybe a rerun will do the trick.

packages/vm/src/runTx.ts
```diff
@@ -506,6 +506,11 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
     const keys = Object.keys(results.execResult.selfdestruct)
     for (const k of keys) {
       const address = new Address(hexToBytes(k))
+      if (this._common.isActivatedEIP(6780)) {
+        if (!results.execResult.createdAddresses![address.toString()]) {
+          continue
+        }
+      }
```
holgerd77, line 513: Oh, this is interesting, that this is the "only" change on the procedural code (apart from the address tracking) to get this working/implemented. 🤓 🙂

packages/vm/test/api/EIPs/eip-6780-selfdestruct-same-tx.spec.ts
jochem-brouwer: Have addressed!

acolytec3 (commented):
This looks great as far as it goes.  Do we need to consider anythiung else with regard to the call frame?  Just to ensure I'm thinking this through correctly.  Addresses hit by self-destruct in a call frame higher than "0" (i.e. the first call in the call stack), get pushed into the list of `createdAddresses` in the callResult for the lower level calls, right?  Just want to make sure I'm thinking this through correctly.

packages/vm/src/runTx.ts
```diff
@@ -506,6 +506,11 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
     const keys = Object.keys(results.execResult.selfdestruct)
     for (const k of keys) {
       const address = new Address(hexToBytes(k))
+      if (this._common.isActivatedEIP(6780)) {
+        if (!results.execResult.createdAddresses![address.toString()]) {
```
g11tech, line 510:
```suggestion
        if (results.execResult.createdAddresses![address.toString()] !== true) {
```
would this be better?

jochem-brouwer: Hmm - actually, if the address is not in the object, it would return false. (If the address is not created in this tx, and we want to selfdestruct it, this does not happen). So the check should be `=== undefined`. (Why does the linter not catch this? It should throw on non-boolean expressions?)

packages/evm/src/types.ts
```diff
@@ -95,6 +95,10 @@ export interface EVMRunCallOpts {
    * Addresses to selfdestruct. Defaults to none.
    */
   selfdestruct?: { [k: string]: boolean }
+  /**
+   * Created addresses in current context. Used in EIP 6780
+   */
+  createdAddresses?: { [k: string]: boolean }
```
g11tech, line 101: map/set would perform better than object

g11tech (approved):
lgtm

PR #2773, opened 2023-06-11
RPC method eth_feeHistory implementation

GitMark0:
Implements: `eth_feeHistory` method for tracking historical gas information

Method is used as described in: https://docs.infura.io/networks/ethereum/json-rpc-methods/eth_feehistory

### DIFF
--- a/packages/client/src/rpc/modules/eth.ts
+++ b/packages/client/src/rpc/modules/eth.ts
@@ -3,6 +3,8 @@ import { BlobEIP4844Transaction, Capability, TransactionFactory } from '@ethereu
 import {
   Address,
   TypeOutput,
+  // eslint-disable-next-line import/named
+  bigIntMax,
   bigIntToHex,
   bytesToPrefixedHexString,
   hexStringToBytes,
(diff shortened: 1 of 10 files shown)


packages/client/src/rpc/modules/eth.ts
```diff
@@ -234,6 +237,82 @@ const getBlockByOption = async (blockOpt: string, chain: Chain) => {
...
+    ) {
+      blockRewards.push(tx.effectivePriorityFee)
+      rewardPercentileIndex++
+    }
+  }
+
+  return blockRewards
+}
+
+const calculateNextBaseFee = (parentBlock: Block, londonHardforkNumber: bigint | null) => {
```
jochem-brouwer, line 287: You can use `calcNextBaseFee` of `parentBlock.header` here

packages/client/src/rpc/modules/eth.ts
```diff
@@ -1119,4 +1204,75 @@ export class Eth {
 
     return bigIntToHex(gasPrice)
   }
+
+  async feeHistory(params: [string | number | bigint, string, Array<number>?]) {
```
jochem-brouwer, line 1208: I don't think the `params` type is correct here. It should be `[bigint, string, [bigint]?]` per the spec

packages/client/test/rpc/eth/getFeeHistory.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+    t.equal(increase, -0.125)
+  }
+
+  await baseRequest(t, server, req, 200, expectRes)
+})
+
+tape(
+  `${method}: should return initial base fee if the block number is london hard fork`,
+  async (t) => {
+    const initialBaseFee = 1000000000n
```
jochem-brouwer, line 92: This const should probably be retrieved from common

packages/client/test/rpc/eth/getFeeHistory.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+      const [baseFee] = res.body.result.baseFeePerGas as [string]
+
+      t.equal(bytesToBigInt(hexToBytes(baseFee)), initialBaseFee)
+    }
+
+    await baseRequest(t, server, req, 200, expectRes)
+  }
+)
+
+tape(`${method}: should return 0x0 for base fees requested before eip-1559`, async (t) => {
```
jochem-brouwer, line 107: Not super sure about the `eth_feeHistory` spec, does this only work for 1559 chains, or also for non-1559 chains..?

packages/wallet/src/util/bytes.ts
```diff
@@ -379,6 +379,18 @@ export const bigIntToHex = (num: bigint) => {
   return '0x' + num.toString(16)
 }
 
+/**
+ * Calculates max bigint from an array of bigints
+ * @param args array of bigints
+ */
+export const bigIntMax = (...args: bigint[]) => args.reduce((m, e) => (e > m ? e : m))
```
jochem-brouwer, line 386: This should be required from `@ethereumjs/util`

jochem-brouwer (commented):
Great PR, some first comments.

holgerd77:
Hi @GitMark0, hope we didn't push you away here, but remarks like the one from @jochem-brouwer are simply part of our normal review process we also apply internally? 😬 🙂

Would it be possible to get this over the finish line? This would be somewhat of a pity otherwise, there is already so much great and extensive work in the PR here!

GitMark0:
> Hi @GitMark0, hope we didn't push you away here, but remarks like the one from @jochem-brouwer are simply part of our normal review process we also apply internally? 😬 🙂
> 
> 
> 
> Would it be possible to get this over the finish line? This would be somewhat of a pity otherwise, there is already so much great and extensive work in the PR here!

Hey Holger, thanks for reaching out! Unfortunately I just didn't find time to finish the PR. I got really busy with a new (unplanned) project a day or two after I opened it... I'd love to finish the PR and continue contributing to the EthereumJS community and I hope that I'll have time to do it by the end of the month. Cheers!

holgerd77:
Cool, no problem and no rush! 🙂👍🙏

packages/client/test/rpc/eth/getFeeHistory.spec.ts
GitMark0: If by non-1559 chains you mean backward compatibility, then yes. The `eth_feeHistory` is fully backward compatible as per spec. 

`Legacy Ethereum transactions will still work and be included in blocks, but they will not benefit directly from the new pricing system. This is due to the fact that upgrading from legacy transactions to new transactions results in the legacy transaction’s gas_price entirely being consumed either by the base_fee_per_gas and the priority_fee_per_gas.`

Source: https://eips.ethereum.org/EIPS/eip-1559

packages/client/test/rpc/eth/getFeeHistory.spec.ts
holgerd77: This can be retrieved with `common.param('gasConfig', 'initialBaseFee')` (so it's there, guess you have overlooked).

jochem-brouwer:
Follow-up: #3295

holgerd77:
Super-seeded by #3295, will close

PR #2775, opened 2023-06-12
common: update the blob target and limit to 3/6 respectively

g11tech:
ref to updated devnet 6 specs:
 - https://github.com/ethereum/EIPs/pull/7154

TODO:
 - [x] fix tests

### DIFF
--- a/packages/block/test/eip4844block.spec.ts
+++ b/packages/block/test/eip4844block.spec.ts
@@ -133,16 +133,17 @@ tape('data gas tests', async (t) => {
     t.equal(excessDataGas, 0n, 'excess data gas should be 0 for small parent header data gas')
     t.equal(dataGasPrice, 1n, 'data gas price should be 1n when low or no excess data gas')
     const highGasHeader = BlockHeader.fromHeaderData(
-      { number: 1, excessDataGas: 4194304, dataGasUsed: BigInt(4) * dataGasPerBlob },
+      { number: 1, excessDataGas: 6291456, dataGasUsed: BigInt(6) * dataGasPerBlob },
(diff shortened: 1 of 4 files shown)


packages/block/test/eip4844block.spec.ts
```diff
@@ -133,16 +133,17 @@ tape('data gas tests', async (t) => {
     t.equal(excessDataGas, 0n, 'excess data gas should be 0 for small parent header data gas')
     t.equal(dataGasPrice, 1n, 'data gas price should be 1n when low or no excess data gas')
     const highGasHeader = BlockHeader.fromHeaderData(
-      { number: 1, excessDataGas: 4194304, dataGasUsed: BigInt(4) * dataGasPerBlob },
+      { number: 1, excessDataGas: 6291456, dataGasUsed: BigInt(6) * dataGasPerBlob },
```
gabrocheleau, line 136: Could be worth declaring constants BLOB_PER_BLOCK_TARGET and BLOB_PER_BLOCK_MAX ? Imo it would make it clearer to follow what's going on + easier to play with parameters & update when necessary.

g11tech: BLOB_PER_BLOCK_TARGET and BLOB_PER_BLOCK_MAX are already calculated from the hardfork params (which is same)

tbh i don't see the need unless you feel strongly

gabrocheleau: Ah fair enough then, I agree, no need to add that.

acolytec3 (approved):
LGTM!

PR #2776, opened 2023-06-12
tx: remove default hardfork from tx classes

gabrocheleau:
Addresses the issue that was brought up in this comment: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2767#discussion_r1226359471

### DIFF
--- a/packages/tx/src/baseTransaction.ts
+++ b/packages/tx/src/baseTransaction.ts
@@ -1,4 +1,4 @@
-import { Chain, Common, Hardfork } from '@ethereumjs/common'
+import { Chain, Common } from '@ethereumjs/common'
 import {
   Address,
   MAX_INTEGER,
@@ -25,6 +25,7 @@ import type {
   TxOptions,
   TxValuesArray,
 } from './types'
+import type { Hardfork } from '@ethereumjs/common'
 import type { BigIntLike } from '@ethereumjs/util'
 
 interface TransactionCache {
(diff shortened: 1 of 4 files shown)


acolytec3:
Should we also remove the entire [`DEFAULT_HARDFORK`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/0a6bf5e70cf0b7ee68bb67be76a7d528df6d1af7/packages/tx/src/baseTransaction.ts#L92) from the `BaseTransaction` class?  Shouldn't we just go straight to the default hardfork specified by `common` rather than continue to retain this default hardfork in `BaseTransaction`?

scorbajio (dismissed):
ACK

gabrocheleau:
> Should we also remove the entire [`DEFAULT_HARDFORK`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/0a6bf5e70cf0b7ee68bb67be76a7d528df6d1af7/packages/tx/src/baseTransaction.ts#L92) from the `BaseTransaction` class? Shouldn't we just go straight to the default hardfork specified by `common` rather than continue to retain this default hardfork in `BaseTransaction`?

Oh, definitely, I just missed that BaseTransaction has a Default hardfork too

acolytec3 (approved):
LGTM

holgerd77:
Nice, yes, this whole `DEFAULT_HARDFORK` thing was some overthinking when new tx types came in and we were still on an older HF and wanted people to just instantiate a new tx type "out of the box".

So basically the same situation as we have with 4844 blob tx right now, but I now feel it's very much ok and even more transparent and clear that in these cases there is the need to instantiate a Common.

PR #2778, opened 2023-06-13
fixed requests.msgTypes in devp2p/examples/peer-communication.ts

simone1999:
the intended behaviour from the requests.msgTypes dict of counters seems to be different from what was implemented. With this fix the message counter should probably count the amount of received messages per peer and message type. This counter later is used to disconect from peers that did not broadcast any blocks while submitting 8 other requests with the same message type to this node.


holgerd77:
Hi there, thanks for submitting and sorry for the age-long non-review, I guess no one has touched these example code parts for quite some time.

I now had a closer look, I think your fix is correct, the current version makes not so much sense.

Would you be able to re-submit this on top of current master, either as a new PR or by force-push here?

That would be great than we can merge this! 🙏 🙂

holgerd77:
Fixed a merge conflict and updated this via UI

(I have no idea why I requested a PR update before (see above), since I now could just simply do this myself via UI??)

Anyhow. 🙂

PR #2779, opened 2023-06-13
vm: fix the dataGasPrice calculation in running the tx

g11tech:
fix the bug in vm runTx where the dataGasPrice calc was still being based on the block's parent from ssz times.

this was caught in devnet-6 runs where for the first time blocks with consecutive blobs were produced


acolytec3 (approved):
LGTM!  Yay for testnets catching bugs.

PR #2781, opened 2023-06-13
Utils tests

scorbajio:


scorbajio:
`MapDB`'s copy function does not create a new `Map`, but rather passes the current one by reference, so the newly copied `MapDB` still uses the old one's backing `Map` and so any changes to it will reflect for both the newly copied and old `MapDB` objects. It would be less confusing to rename the function to reflect this behavior or update/add documentation to mention this so that developers are not confused.

packages/util/test/mapDB.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    })
+  })
+
+  t.test('open', (t) => {
+    t.test('should return a resolved promise', async (t) => {
+      const mapDB = new MapDB()
+
+      const promise = await mapDB.open()
+
+      t.equal(promise, undefined, 'should return a resolved promise')
```
acolytec3, line 76:
```suggestion
      t.equal(promise, undefined, 'should return undefined since this is a stub function')
```

acolytec3 (approved):

PR #2782, opened 2023-06-13
common: default common chain to Mainnet

gabrocheleau:


holgerd77:
I think I have a tendency to not favor this change, with some similar feeling like Jochem expressed in the call. A bit other context, but when someone wants to instantiate a tx e.g. using a specific HF for whatever reason, chances are then (too) high that the chain gets forgotten, if this tx e.g. is intended to be send/used on Goerli. And then the signature gets wrong. Same for running a block within the VM, to take one other example, there are more though.

So my tendency moves stronger towards the do-not-change side of things.

holgerd77:
Open to take in other voices though, still not 100% settled.

holgerd77:
We decided to not adopt due to keep chain setting more explicit, will close.

PR #2784, opened 2023-06-14
client: fix new payload 4844 validations

g11tech:
Another discovery from devnet-6 test runs:

The verifying of excessDataGas/4844 validations were not fully happening in new payload because of requirement of parent, which in our normal flow would happen once we do a putBlock into chain (which happens in fcU update)

This PR performs those validations


holgerd77 (approved):
LGTM

PR #2789, opened 2023-06-16
Devp2p file extensions

acolytec3:
The fully defined file extensions got left off for `devp2p` when we merged #2786.  This PR adds them in for the `src` folder, updates the `examples` scripts to run using the latest changes, updates the readme references, and also cleans up a couple of dependencies (remove one `types` and move another to `devDependencies`).


packages/devp2p/examples/peer-communication.cts
```diff
@@ -7,10 +7,10 @@ import chalk from 'chalk'
 import type LRUCache from 'lru-cache'
 
 const LRU = require('lru-cache')
```
gabrocheleau, line 9: Can't LRU be imported with ESM syntax?

gabrocheleau: Their docs seem to indicate that the import syntax should work: http://isaacs.github.io/node-lru-cache/

packages/devp2p/src/rlpx/index.ts
```diff
@@ -1,6 +1,6 @@
 'use strict'
```
gabrocheleau, line 1: Is use strict still necessary?

gabrocheleau (approved):
Approved but left a few minor comments

packages/devp2p/examples/peer-communication.cts
acolytec3: Doesn't seem to work in this instance.  I think since I've changed these scripts to be `cts` (ie explicitly CommonJS), we have to continue to require it.  I'm sure there's a way to do it with import but the exports in the version of LRUCache we're using here don't play nicely with Typescript's `import ...` resolution.

packages/devp2p/src/rlpx/index.ts
acolytec3: Indeed not.  Thanks!

gabrocheleau (approved):

PR #2792, opened 2023-06-17
tx: simplify validate methods

### LINKED ISSUE #2726: Tx: Replace TypedTransaction type with generic Interface

gabrocheleau:
resolves #2726 

This PR mainly removes the Transaction.validate overloaded methods, and replaces them with these two:
- an `isValid()`  method that returns a boolean
- a `getValidationErrors` method that returns the validation error strings

It also simplifies the corresponding `Block.validateTransactions` methods in the same way, and unifies the naming of a few other block validation methods to conform to the following semantics:

- isValid() returns true if tx is valid, false if not
- validate() throws if invalid
- getValidationErrors() returns an array of error strings

 I also suspect those updates will bring performance improvements, although I'm not sure how significant. Currently we were going through all validation checks, populating an errors: string[], and only then returning a boolean if the errorStr: boolean was undefined/false. This is quite wasteful as when only a boolean is return, we can return false as soon as a validation error is detected.


gabrocheleau:
There was also mention in the #2767 PR comments that we should also take a look at the `tx.getMessageToSign` method.
```
  getMessageToSign(hashMessage: false): Uint8Array | Uint8Array[]
  getMessageToSign(hashMessage?: true): Uint8Array
  ```

I am less convinced that this simplification would be meaningful. If we strictly want to avoid overloading, we could split it in two methods, namely `getMessageToSign` and `getHashedMessageToSign`.

**Comment by holgerd77** [comment]
I'm a bit undecided on this PR, will give this another round of thought.

**Review by holgerd77 — CHANGES_REQUESTED** [review]
Hey there, had a first looks, two comments.

Also, for `getMessageToSign()`:

Yes, can we please do this, to split this up into two methods `getMessageToSign()` and `getHashedMessageToSign()`?

I can remember various confusing discussions (so: library users where confused 😋) around this method, and this would mitigate this, this would be also relevant since this is a very central/sensitive method people use for ledger signing and the like, it's good to reduce confusion here.

And we then also remove this overloading, which is good.

Can you also integrate in this PR? Thanks! 🙂

File: packages/block/src/block.ts [hunk 1]
```diff
@@ -603,17 +644,19 @@ export class Block {
...
   }
 
   /**
    * Validates the withdrawal root
+   * @returns true if the withdrawals trie root is valid, false otherwise
    */
-  async validateWithdrawalsTrie(): Promise<boolean> {
+  async withdrawalsTrieIsValid(): Promise<boolean> {
     if (!this._common.isActivatedEIP(4895)) {
       throw new Error('EIP 4895 is not activated')
```
jochem-brouwer, line 661: This throws? Shouldn't this `return false`?

packages/tx/src/baseTransaction.ts
```diff
@@ -151,26 +151,58 @@ export abstract class BaseTransaction<T extends TransactionType>
...
-      errors.push(`gasLimit is too low. given ${this.gasLimit}, need at least ${this.getBaseFee()}`)
+      return false
     }
 
+    // Invalid signature
+    if (this.isSigned() && !this.verifySignature()) {
+      return false
+    }
+    return true
+  }
```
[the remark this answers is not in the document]
jochem-brouwer: Yes, and I think the same holds for block.

jochem-brouwer (commented):
Some comments.

In general: shouldn't we generalize this interface? We now have 

`isValid` + `getValidationErrors` on tx,

`isValid` + `validate` + `getValidationErrors` on block

We could thus also generalize those by writing an internal validation method which validates all rules.`isValid` loops over these rules via a `try catch`, if one of the rules throws return false else true. `validate` just loops over the rules, if one of the methods throws `validate` then also throws. `getValidationErrors` loops over the rules via `try catch` and pushes errors onto an array and finally returns those (or empty array if no errors). 

(Alternatively we could also create two arrays of the validation methods, where the validation methods return either true or false and we lookup an error message from the array if we want the error message - this would be slightly more performant since we don't have to create the strings if we throw)

WDYGT?

packages/block/src/block.ts
jochem-brouwer:
```typescript
  /**
   * Validates the withdrawal root
   * @returns true if the withdrawals trie root is valid, false otherwise
   */
  async withdrawalsTrieIsValid(): Promise<boolean> {
    if (!this._common.isActivatedEIP(4895)) {
      throw new Error('EIP 4895 is not activated')
```
     
The (new) docs say otherwise :thinking:

jochem-brouwer: Hmm, wait, it makes no sense if you call the method if the EIP is activated so I guess it is correct

packages/tx/src/baseTransaction.ts
```diff
@@ -151,26 +151,58 @@ export abstract class BaseTransaction<T extends TransactionType>
...
+  //     errors.push('Invalid Signature')
+  //   }
+
+  //   if (this.getBaseFee() > this.gasLimit) {
+  //     errors.push(`gasLimit is too low. given ${this.gasLimit}, need at least ${this.getBaseFee()}`)
+  //   }
+
+  //   return errors
+  // }
+
```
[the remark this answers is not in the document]
gabrocheleau: Removed

packages/block/src/block.ts
gabrocheleau: Yeah exactly, makes sense that it throws if 4895 isn't activated as this basically is a misuse of the method.

gabrocheleau:
> Some comments.
> 
> In general: shouldn't we generalize this interface? We now have
> 
> `isValid` + `getValidationErrors` on tx,
> 
> `isValid` + `validate` + `getValidationErrors` on block
> 
> We could thus also generalize those by writing an internal validation method which validates all rules.`isValid` loops over these rules via a `try catch`, if one of the rules throws return false else true. `validate` just loops over the rules, if one of the methods throws `validate` then also throws. `getValidationErrors` loops over the rules via `try catch` and pushes errors onto an array and finally returns those (or empty array if no errors).
> 
> (Alternatively we could also create two arrays of the validation methods, where the validation methods return either true or false and we lookup an error message from the array if we want the error message - this would be slightly more performant since we don't have to create the strings if we throw)
> 
> WDYGT?

I think the approach you suggest makes sense, but I doubt it is necessary at this point, as we don't have to my knowledge cases where we have all 3 methods, only two instances where we have "isValid" and "getValidationErrors". I feel it's sufficient to just have the isValid method call `this.getValidationErrors` and just return `errors.length === 0`. I've done the same for `getTransactionsValidationErrors` and `transactionAreValid` from block. If others feel like your approach would be more useful though I can implement it, just seems a bit overkill at this stage in my view.

gabrocheleau:
I refactored `getMessageToSign(true)` and `getMessageToSign(false)` into two separate methods (`getMessageToSign` and `getHashedMessageToSign` respectively) and simplified the code of some validate methods to make the code more DRY.

packages/block/src/block.ts
jochem-brouwer: Definitely, will close this thread

holgerd77:
Have given this a branch update, will have another look at this PR now.

holgerd77 (approved):
I have to say I am now totally in love with this PR and all the changes it contain after having a third look with some distance, guess it was also just the amount of things changed here which needed to sink in a bit.

But these new structured naming semantics make so much sense, also (to also tap a bit on my shoulder 😅) I really think that this second overlading removal is such a big win, `getHashedMessageToSign()` is just so much more explicit and readable than `tx.getMessageToSign(true)` (where you can literally "read" nothing from about what this call actually does).

So, great great, will approve and merge, thanks for tackling! 🙂 👍

jochem-brouwer (approved):
I agree here, my proposal would make everything complicated which is not necessary. Great work :tada:

jochem-brouwer:
Forgot to note: these method names changes are breaking ( @holgerd77 ). But we can (?) release those?

holgerd77:
Failing tests are unrelated, will merge here.

There seems to be some timeout now in some blockchain tests, @acolytec3 did you actually introduce some additional caching now? 🤔 Might be related (building cache takes to long or something).

holgerd77:
> Forgot to note: these method names changes are breaking ( @holgerd77 ). But we can (?) release those?

Yes, super-breaking for sure, that's why I needed three rounds to let this sink in if it's worth it. But - yes - we do breaking releases anyhow, it's a bit tedious for people but I would now say that this clear semantics (everything `*IsValid` gives a boolean) is worth it, since this helps a lot on easy-code understanding.

acolytec3:
> Failing tests are unrelated, will merge here.
> 
> There seems to be some timeout now in some blockchain tests, @acolytec3 did you actually introduce some additional caching now? 🤔 Might be related (building cache takes to long or something).

Not that I'm aware of. I have some experimental PRs with build caches but nothing that got merged into master that I'm aware of

jochem-brouwer:
I re-ran the test, it was likely just the runner which was being super slow :)

PR #2795, opened 2023-06-18
util: Rename mapdb copy function

scorbajio:
The current implementation of the `copy` function of the `DB` interface does not create a new underlying db but rather reuses the same db of the original object being copied. This PR renames the `copy` function to `shallowCopy` in order to reflect this behavior and make it's name more intuitive and less confusing.


acolytec3 (approved):
LGTM

PR #2796, opened 2023-06-18
util: Add tests for blob helpers

scorbajio:


packages/util/test/blobHelpers.spec.ts
```diff
@@ -0,0 +1,42 @@
+import { assert, describe, it } from 'vitest'
+
+import { commitmentsToVersionedHashes, computeVersionedHash, getBlobs } from '../src/index.js'
+
+describe('YourClass', () => {
```
acolytec3, line 5: Total nit but can we rename this to something that reflects the test case, just `blob helper tests` at a minimum?

acolytec3 (changes requested):
Looks great, with one edit on the test case name

packages/util/test/blobHelpers.spec.ts
scorbajio: Fixed.

gabrocheleau (approved):

acolytec3 (approved):
Lgtm

Issue #2788, opened 2023-06-16
`vitest` and ESM todos


acolytec3:
After merging the big PR with vitest and ESM path extensions, we have a few todos left

- [ ] Rewrite the Ethereum -Tests runner to use ESM and run in vitest
- [x] Replace tape with vitest in client
- [ ] Get browser tests working for all packages

holgerd77:
Very small issue and we will never re-find this anyhow and some stuff addressed.

Will close.

PR #2797, opened 2023-06-18
client: migrate tests to vite

gabrocheleau:
Addresses part of #2788


ScottyPoi:
Client-CLI was failing due to an `exclude` setting in `vitest.config.json`

Deleting that line fixed the test run.  Was it set that way for a reason?

packages/client/vitest.config.js
```diff
@@ -0,0 +1,9 @@
+import { defineConfig } from 'vitest/config'
+
+export default defineConfig({
+  test: {
+    exclude: ['test/(integration|cli|sim)/**/*.spec.ts'],
```
ScottyPoi: Why this line?  it caused the Client-CLI test to crash

acolytec3: Sorry, I should have come back to this and hasn't gotten to it. This config sets up the unit tests correctly. Otherwise, vitest will just run everything that looks a test since it doesn't respect glob patterns from the CLI. We can update the integration and sim test npm scripts separately so they only target the specific folders. Will do that this morning.

acolytec3:
A comments on the things I changed when migrating things to vitest that I think are important:
In general, I started the process of removing `testdouble` and replacing with `vitest`'s inbuilt mocking functionality.  It's not completely done by a long shot but that can be taken up in further PRs.
- I completely removed the `test/integration/lightethereumservice.spec.ts' since there were no actual tests in that file
- I tweaked [this test](https://github.com/ethereumjs/ethereumjs-monorepo/blob/9ba9d2dfbb5a52e373f6e4829b34363773486042/packages/client/test/net/protocol/flowcontrol.spec.ts#L20) to make it work by removing an element from the array.  I honestly don't know if this materially impacts the test output but I didn't want to spend hours trying to solve this when LES/light sync is a little used protocol and I'm reasonably confident our implementation works.  There was some weirdness with the fake timers used that I couldn't sort out with an hour's research.
- There are a couple of places (that I can't remember precisely) where switching from testdouble to vitest mocks isn't an exact like for like replacement (basically there were places where I wasn't sure how to exactly match the `td.verify(something was called with some set of params)` using similar `vitest` functionality so I opted for more general `vitest` "this was called" type functionality.
- I modified how the `superagent` setup helpers create the servers for the RPC tests to randomize the port used (basically by just not specifying a port) so the tests won't conflict when multiple RPC tests are run in parallel.

holgerd77 (approved):
LAGTM (Looks absurdly great to me)

Will merge.

PR #2799, opened 2023-06-19
Typescript KZG proof of concept usage with ESM

acolytec3:
Supersedes #2725.  I won't do further testing on this until we confirm that the typescript implementation has been updated to the latest KZG spec but this at least answers the question of whether we can import this `@crate-crypto/crypto-4844` module directly since its ESM only.

- [x] Validate that import of ESM only module works now - [d96562f](https://github.com/ethereumjs/ethereumjs-monorepo/pull/2799/commits/d96562f90b9f16807feb259ffb04270bdaac1cd1)
- [ ] Determine if the CJS output of our code will allow this direct import to still work or if it breaks
- [ ] See how this works in browser
- [ ] If we verify it works everywhere, replace `c-kzg` usage across monorepo

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1736,6 +1736,41 @@
         "node": ">=0.1.90"
       }
     },
+    "node_modules/@crate-crypto/crypto-4844": {
+      "version": "0.2.0",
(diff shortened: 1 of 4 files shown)


acolytec3:
Going to close this for now.  Will revisit once I figure out how to compile `c-kzg` to WASM.

PR #2798, opened 2023-06-19
Common, Block, VM: Unify and Refactor getHardforkBy Options

holgerd77:
Partly solves #2774

This PR does the following:

- `Common.getHardforkByBlockNumber()` -> `Common.getHardforkBy()` ~(+ strict bigint types)~
- `Common.setHardforkByBlockNumber()` -> `Common.setHardforkBy()` ~(+ strict bigint types)~ 
- ~`Blockchain.checkAndTransitionHardForkByNumber()` BigIntLike -> strict bigint types~

Issue #2774, opened 2023-06-12
Monorepo: Unify/Refactor getHardforkByTTD,... Options and Common Methods


holgerd77:
Ok, so this is a somewhat small task we really should do for breaking releases as well, since things are so chaotic right now.

I think I've got a proposal here which is both manageable and un-wiring.

To recap, we have:
-`Common` method `setHardforkByBlockNumber(blockNumber: BigIntLike, td?: BigIntLike, timestamp?: BigIntLike)`
- `Common` method `getHardforkByBlockNumber(blockNumber: BigIntLike, td?: BigIntLike, timestamp?: BigIntLike)`
- `Block` option `hardforkByBlockNumber` (boolean)
- `Block` option `hardforkByTTD` taking precedence over the former if used
- `Common` `hardforkIsActiveOnBlock() / activeOnBlock()` taking in only block number and having just one (DAO) usage (so: not even a "use case")
- `hardforkByTTD?: BigIntLike` option for VM
- `hardforkByTTD?: bigint` option for VM.runBlock() (haha, it's getting ridiculous 😂)
- `_hardforkByBlockNumber: boolean` and `_hardforkByTTD?: bigint` internal VM properties. lOL. Double ROFl.

Ugh.

Ok, all pretty chaotic, but I think fixing this is not as rough as one would think.

So what we can do here is:

1. Add a new dict interface `hardforkByOpts` or similar to Common like:

```typescript
interface hardforkByOpts = {
  blockNumber?: bigint,
  timestamp?: bigint,
  td?: bigint
}
```

(I would be a fan here (and everywhere else) to limit to BigInt and remove the `BigIntLike` thing)

Then we can simplify and better name the Common methods to:

- `Common.setHardforkBy(byOpts: HardforkByOpts)`
- `Common.getHardforkBy(by: HardforkByOpts)` // by or byOpts (aligned of course), both works I guess

(open for better/alternative namings but I liked this after some thought and look)

2. `Block` options should be unified to - name suggestion! - `setHardfork: boolean | bigint` with the same semantics as the previous two options

3. Same alignment for `VM` and `VM.runBlock()` as well as internal properties:

- `setHardfork: boolean | bigint` as VM option
- `setHardfork: boolean | bigint` as VM.runBlock() option
- `_setHardfork: boolean | bigint` as internal property

[...]

PR #2800, opened 2023-06-19
Block, VM: Unify hardforkBy Options

holgerd77:
Second part of #2774 following #2798

This PR does the following:

- Block: unify `hardforkByBlockNumber`, `hardforkByTTD` options -> `setHardfork`
- VM (constructor/runBlock()): unify `hardforkByBlockNumber`, `hardforkByTTD` options -> `setHardfork`


jochem-brouwer (dismissed):
LGTM

packages/block/src/header.ts
```diff
@@ -182,11 +177,17 @@ export class BlockHeader {
...
-    const hardforkByBlockNumber = options.hardforkByBlockNumber ?? false
-    if (hardforkByBlockNumber || options.hardforkByTTD !== undefined) {
+    const setHardfork = options.setHardfork ?? false
+    if (setHardfork === true) {
+      this._common.setHardforkBy({
+        blockNumber: number,
+        timestamp,
+      })
+    }
+    if (typeof setHardfork !== 'boolean') {
```
acolytec3, line 187: Could this be simplified to just the below?

```suggestion
    else {
```

acolytec3, line 187:
```suggestion
    else if (typeof setHardfork !== 'boolean') {
```

packages/vm/src/runBlock.ts
```diff
@@ -56,16 +57,21 @@ export async function runBlock(this: VM, opts: RunBlockOpts): Promise<RunBlockRe
...
-    })
+  if (setHardfork !== false || this._setHardfork !== false) {
+    const setHardforkUsed = setHardfork ?? this._setHardfork
+    if (setHardforkUsed === true) {
+      this._common.setHardforkBy({
+        blockNumber: block.header.number,
+        timestamp: block.header.timestamp,
+      })
+    }
+    if (typeof setHardforkUsed !== 'boolean') {
```
acolytec3, line 68:
```suggestion
    else if (typeof setHardforkUsed !== 'boolean') {
```

acolytec3 (changes requested):
I could be missing the significance but requested one change in how we evaluate the `setHardfork` opt so we don't check it twice (if it's indeed `true`).  Small efficiency gain if I'm not missing something.  Otherwise, looks great!

holgerd77:
@jochem-brouwer @acolytec3 thanks for reviewing 🙂, have applied the suggested changes, needs re-approval now! 🙂

jochem-brouwer (dismissed):
Ah right, good catch @acolytec3. LGTM still! :smile: (assuming CI passes)

jochem-brouwer:
Nvm, CI fails. I should not approve so fast :sweat_smile: (how do I dismiss my review?)

jochem-brouwer:
Will re-approve if CI passes

g11tech (approved):
looks amazing! :rocket: :heart:

PR #2801, opened 2023-06-20
VM:  test:API skipping tests

ScottyPoi:
The NPM script `npm run test:API` is meant to run all tests in the `vm/test/api/` dir.
 
as written, it looks to match: `test/api/**/*.spec.ts`

This matches tests in `test/api/EIPs`, `test/api/Istanbul`, etc. but does not hit the `spec.ts` files that were directly in the `test/api/` directory, and not in a subdirectory.

~~Solved by moving those skipped tests via `git mv`  into a new subdirectory `test/api/API/`~~

Updated test script to  ` "test:API": "npx vitest run ./test/api/*.ts ./test/api/**/*.spec.ts" `

Now the `npm run test:API` runs all expected tests (which passed :relieved: )

---

~~Also added a dependency `"@esbuild/linux-x64"` which has been necessary to run vitest on my machine.  Installing locally did not work.~~


holgerd77 (changes requested):
Hi there, great find! 🤩

I wouldn't want to have the whole test files moved for this. I find this overblown and the `api/API` test folder doesn't look so nice.

Can we please rather solve this by just adopting the test script command?

holgerd77:
This @esbuild/linux thing is then in turn killing other platforms (mine e.g. with `Unsupported platform for @esbuild/linux-x64@0.18.4: wanted {"os":"linux","arch":"x64"} (current: {"os":"darwin","arch":"arm64"})` 😜), so we need to find another solution for this and this needs to be removed from the PR (Trie as well, at least before finishing up).

jochem-brouwer:
I had this esbuild error yesterday, and `npm run clean && npm i` in monorepo root dir fixed it for me :+1: :smile:

holgerd77:
> I had this esbuild error yesterday, and `npm run clean && npm i` in monorepo root dir fixed it for me 👍 😄

Not sure if this is a suggestion for me or for Scotty, but if it's for me: I just tested and still get the same error. 🙂

ScottyPoi:
Fixed by #2802

PR #2803, opened 2023-06-20
Add extensions for new `util` file

acolytec3:
Adds missing file extensions on internal imports in `genesis.ts`


acolytec3:
I found these missing through trying to integrate our breaking releases into Ultralight.  Doubtless will find more along the way.

holgerd77 (approved):
LGTM

PR #2804, opened 2023-06-21
Monorepo: Update lru-cache dependencies (ESM) / Switch Browser Test Provider

holgerd77:
This PR updates different require-using dependencies, in particular the lru-cache dependency.

I've now "solved" the undefined/null-now-forbidden question for lru-cache by just using `@ts-ignore`, see also comment on trie library for some reasing, also opened this issue https://github.com/isaacs/node-lru-cache/issues/308 on the `lru-cache` library, since - while also having some intuition why restricting on undefined and null makes sense - I also feel that e.g. for our use case this is still the cleanest way to do things.

Update: `@ts-ignore` doesn't work since it is stripped in the declaration files in the `dist` folder. I am now using `any` for this. Still feel this is the better solution than using an `"UNDEFINED"` string or whatever one could come up with.


packages/vm/package.json
```diff
@@ -19,6 +19,7 @@
   "contributors": [
     "Alex Beregszaszi <alex@rtfs.hu>"
   ],
+  "type": "commonjs",
```
holgerd77, line 22: I would want to keep this for now. It is generally not yet 100% clear what setting is appropriate here and on the VM browser tests work better when setting here to `"commonjs"`(lol).

packages/trie/src/db/checkpoint.ts
```diff
@@ -16,7 +14,17 @@ export class CheckpointDB implements DB {
...
+  // Starting with lru-cache v8 undefined and null are not allowed any more
+  // as cache values. At the same time our design works well, since undefined
+  // indicates for us that we know that the value is not present in the
+  // underlying trie database as well (so it carries real value).
+  //
+  // Solution here seems therefore adequate, other solutions would rather
+  // be some not so clean workaround.
+  //
+  // (note that @ts-ignore doesn't work since stripped on declaration (.d.ts) files)
+  protected _cache?: LRUCache<string, any>
```
Thenkei, line 26:
```suggestion
  protected _cache?: LRUCache<string, Uint8Array>
```

Is this enough in your case?

holgerd77: No, we need undefined as a potential value within the current code setup, this is not allowed anymore by the lru-cache library, therefore the any typing as some kind of hack here.

acolytec3 (dismissed):
LGTM, as a starting point for getting browser tests running again.

acolytec3 (approved):
LGTM

PR #2806, opened 2023-06-21
VM/EVM: move selfdestruct type to set

jochem-brouwer:
As noticed in #2771, @g11tech and me noted that the EIP6780 created addresses object could be transformed in a set, but the same is the case for the selfdestruct addresses.

This PR updates the type. It also updates the documentation: the previous type was an object where the keys were the address-to-selfdestruct and the value either a boolean (marks address to selfdestruct) or an Uint8Array (the address to selfdestruct to). This is wrong: when operating SELFDESTRUCT, funds are immediately sent to the recipient. After finishing the transaction, *then* the account is actually deleted. (This thus also means that if funds are somehow (e.g. via a CALL to that address via another call frame of another contract) in the contract, these are burned - this was the only way to burn ETH pre-1559). Most likely this tape came from the assumption that at the end of the tx, the funds were still sent to the recipient (which is not the case).


g11tech (approved):
looks amazing!

g11tech (dismissed):

g11tech (approved):
lgtm

PR #2807, opened 2023-06-21
Updates the crc32 dependency in `common` 

acolytec3:
Replaces the existing crc32 dependency in `common` with one that exports an ESM build


acolytec3:
Have cherry-picked the important commits into #2809

PR #2810, opened 2023-06-22
Implement EIP4788: Beacon block root in EVM

jochem-brouwer:
Implements https://eips.ethereum.org/EIPS/eip-4788

WIP

Note: keep eyes on https://github.com/ethereum/EIPs/pull/7178

From the EIP (specifically https://github.com/ethereum/EIPs/pull/7178:)

> When verifying a block, execution clients MUST ensure the root value in the block header matches the one provided by the consensus client.

This will need a PR on the CL side for the execution API, so will not implement here


jochem-brouwer:
Referenced PR is merged. This is ready for review.

jochem-brouwer:
Am on mobile, cannot see if this can be merged without conflicts. (how do I see this on mobile?)

holgerd77:
Yes, no, so this needs a branch update.

holgerd77:
Have also dropped a short question on the readiness-state of the EIP on https://discord.com/channels/595666850260713488/856943705209569291/1127869874354524240.

packages/block/src/header.ts
```diff
@@ -56,6 +56,7 @@ export class BlockHeader {
   public readonly withdrawalsRoot?: Uint8Array
   public readonly dataGasUsed?: bigint
   public readonly excessDataGas?: bigint
+  public readonly beaconRoot?: Uint8Array
```
g11tech, line 59:
```suggestion
  public readonly parentBeaconBlockRoot?: Uint8Array
```
to be one to one with eip and specs

packages/evm/src/precompiles/0b-beaconroot.ts
```diff
@@ -0,0 +1,76 @@
...
+  )
+
+  if (bytesToBigInt(recordedTimestamp) !== timestampInput) {
+    return {
+      executionGasUsed: gasUsed,
+      returnValue: zeros(32),
+    }
+  }
+  const timestampExtended = timestampIndex + historicalRootsLength
+  const returnData = setLengthLeft(
```
g11tech, line 60: does the output of `getContractStorage` needs `setLengthLeft`?

g11tech (changes requested):
filed name change is requested to be consistent with eip and specs

packages/evm/src/precompiles/0b-beaconroot.ts
jochem-brouwer: Note that this writes to CALLs `out` memory, so we need to write 32 bytes. This is not explicit in `getContractStorage`;

```typescript
    const value = await trie.get(key)
    if (!this._storageCacheSettings.deactivate) {
      this._storageCache?.put(address, key, value ?? hexToBytes('0x80'))
    }
    const decoded = RLP.decode(value ?? new Uint8Array(0)) as Uint8Array
    return decoded
    ```

File: packages/block/src/header.ts (hunk 1 above)
Comment (jochem-brouwer) [reply packages/block/src/header.ts:59]: Hmm weird how I did not pick up this change to the EIP, thanks :smile:

File: packages/evm/src/precompiles/0b-beaconroot.ts (hunk 2 above)
Comment (g11tech) [reply packages/evm/src/precompiles/0b-beaconroot.ts:60]: ahh makes sense, thank you!

**Review by g11tech — APPROVED** [review]
lgtm :rocket:

File: packages/common/src/eips/4788.json [hunk 3]
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "EIP-4788",
+  "number": 4788,
+  "comment": "Beacon block root in the EVM",
+  "url": "https://eips.ethereum.org/EIPS/eip-4788",
+  "status": "Draft",
+  "minimumHardfork": "cancun",
```
holgerd77, line 7: I know it's somewhat counter intuitive and I wonder if we ever "get this in" reliably 😋, but the minimum HF here is not the inclusion HF but St least one lower (do here Shanghai would be a good choice), otherwise this can't be used in a Shanghai+EIP way.

So this is to read: the state of the network where all preconditions are fulfilled so that this EIP can be activated.

packages/evm/src/evm.ts
```diff
@@ -935,6 +936,7 @@ export class EVM {
       common: this.common,
       _EVM: this,
       _debug: this.DEBUG ? debugPrecompiles : undefined,
+      stateManager: this.stateManager,
```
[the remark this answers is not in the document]
holgerd77: Out of interest because I'm totally not getting the whole picture: what's the "super" part here? 😋 What does this do/change?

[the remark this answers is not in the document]
holgerd77: Update: just only getting this now, is this whole thing a precompile and not an opcode??

packages/common/src/eips/4788.json
holgerd77: So just to re-iterate on this and make this very concrete: this is the situation which breaks otherwise (and currently does) if the minimum HF is set to Cancun here:

```shell
➜  vm git:(master) ts-node
> import { Chain, Common, Hardfork } from '@ethereumjs/common'
undefined
> let c = new Common({ chain: Chain.Mainnet, eips: [ 4788 ] })
/ethereumjs-monorepo/packages/common/dist/cjs/common.js:361
                throw new Error(`${eip} cannot be activated on hardfork ${this.hardfork()}, minimumHardfork: ${minHF}`);
                ^

Uncaught Error: 4788 cannot be activated on hardfork shanghai, minimumHardfork: false
    at Common.setEIPs (/ethereumjs-monorepo/packages/common/src/common.ts:455:15)
    at new Common (/ethereumjs-monorepo/packages/common/src/common.ts:241:12)
    at /ethereumjs-monorepo/packages/vm/<repl>.ts:2:9
    at Script.runInThisContext (node:vm:129:12)
    at runInContext (/node/v18.14.2/lib/node_modules/ts-node/src/repl.ts:673:19)
    at Object.execCommand (/node/v18.14.2/lib/node_modules/ts-node/src/repl.ts:639:28)
    at /node/v18.14.2/lib/node_modules/ts-node/src/repl.ts:661:47
    at Array.reduce (<anonymous>)
    at appendCompileAndEvalInput (/node/v18.14.2/lib/node_modules/ts-node/src/repl.ts:661:23)
    at evalCodeInternal (/node/v18.14.2/lib/node_modules/ts-node/src/repl.ts:222:12)
>
```

holgerd77: Not sure, since this is such a re-occuring place for stumbling over it might be really worth to rename to something like the following? 🤔

`minimumHardforkForIsolatedEIP`

Even if the name is a bit bulky.

PR #2811, opened 2023-06-22
EVM: move KZG precompile

jochem-brouwer:
Implements 

https://github.com/ethereum/EIPs/pull/7172 (move KZG)

https://github.com/ethereum/EIPs/pull/7214 (move BLS)

This PR also makes it possible to have multiple precompiles at the same address (of course, only one is active at the same time). 

It also exports this precompileAvailability.

Note: we could consider removing the `precompiles` as well (we export it but I don't think it's used anywhere)


jochem-brouwer:
Ok, note: https://github.com/ethereum/EIPs/pull/7214

We can also remove this multiple-precompiles-at-same-address logic :sweat_smile: (if that PR gets merged)

g11tech (approved):
lgtm!

holgerd77:
Oh, yeah, that's a really nice PR! 👍

PR #2812, opened 2023-06-22
Export precompileAvailability from evm/precompiles/index.ts

CedarMist:
This is a trivial fix to make the `precompileAvailability` list available, given than `precompiles` is also exported.

### Why?

Unfortunately there are situations where it's necessary to monkey-patch custom precompiles into an application which uses ethereumjs and no other method is easily available.

While `precompiles` is exported the `getPrecompile` function will not let you use anything added to it as it does not pass the availability check. See: https://github.com/ethereumjs/ethereumjs-monorepo/blob/b2df1473b8e3fbf9315236d1099bde516a3d002b/packages/evm/src/precompiles/index.ts#L160

An alternative method is to override the `getPrecompile` function, with a fallback to the original `getPrecompile` method.

### DIFF
--- a/packages/evm/src/precompiles/index.ts
+++ b/packages/evm/src/precompiles/index.ts
@@ -208,6 +208,6 @@ function getActivePrecompiles(
   return precompileMap
 }
 
-export { getActivePrecompiles, precompiles, ripemdPrecompileAddress }
+export { getActivePrecompiles, precompileAvailability, precompiles, ripemdPrecompileAddress }
 
 export type { AddPrecompile, CustomPrecompile, DeletePrecompile, PrecompileFunc, PrecompileInput }


jochem-brouwer:
That's true, but `precompiles` themselves are exported and you can add them as custom precompiles in the EVM.

But, nevertheless, it makes sense to export it.

CedarMist:
> and you can add them as custom precompiles in the EVM.

This is in the context where you don't have control over the EVM constructor without very intrusive patches to several third-party apps, or maintaining a fork of ethereumjs and patching the other apps to use it.

jochem-brouwer:
Yes, I agree. Since we are (likely) to change the name of this object (including the type), I have cherry-picked your commit in https://github.com/ethereumjs/ethereumjs-monorepo/pull/2811

jochem-brouwer:
Merged the PR :)

CedarMist:
Awesome thanks!

CedarMist:
@jochem-brouwer  Just to follow up, here is an example use case where custom precompiles are added without patching a third-party product: https://github.com/xmzheng/hardhat-precompiles-plugin/commit/c7d39ad681ff1f3845339f574bab9deac55af8cc

PR #2813, opened 2023-06-22
Remove deprecated networks

scorbajio:
This PR removes the deprecated ropsten and rinkeby networks.


holgerd77:
Oh, this is actually more than I thought (I take this back with the 1 hour work estimate, lol). 😋

Attention, most of the tests removed are actually *not* about Ropsten but about some specific functionality of a library (sometimes the test comments are a bit misleading) and should optimally not be just deleted but replaced by some alternative test scenario from another network.

Also, as a side note: Goerli is likely the next deprecation candidate with the replacing Holesky network already getting into the ring, so might not be the most future-proof choice.

jochem-brouwer:
I would argue to use Sepolia for non-mainnet tests regarding the Goerli end-of-life coming soon.

scorbajio:
> Oh, this is actually more than I thought (I take this back with the 1 hour work estimate, lol). yum
> 
> Attention, most of the tests removed are actually _not_ about Ropsten but about some specific functionality of a library (sometimes the test comments are a bit misleading) and should optimally not be just deleted but replaced by some alternative test scenario from another network.
> 
> Also, as a side note: Goerli is likely the next deprecation candidate with the replacing Holesky network already getting into the ring, so might not be the most future-proof choice.

Yea I was hoping for a quick search-and-replace but some of the tests are specific to ropsten or rinkeby. My plan is to remove those tests and get all of the tests passing, and then come back and rewrite them as a part of the same PR.

scorbajio:
> I would argue to use Sepolia for non-mainnet tests regarding the Goerli end-of-life coming soon.

Noted, I'll opt for Sepolia over Goerli.

holgerd77:
Ah yes, the tests are really its very own task, didn't take that into account.

One thing I just thought of: it might also be an option for the tests to just then instantiate the Common as a custom chain with then the chain ID from Ropsten or Rinkeby (guess we have more Ropsten tests?) and setting the HF explicitly to the HF which would apply (so e.g. if there is a block with a certain block number used as test data one can see which HF this would be).

This might be the substantially easier way to go - e.g. if there is specific test data with explicit fixtures or hashes or whatever. And from my perspective this is a feasible solution, in most (all?) cases this should preserve the goal of the original test.

scorbajio:
> Nice, this already looks pretty good! :-)

Almost there, I am going to take one more stab at the failing client tests (the wasm miner errors are hard to debug because they don't provide helpful messages) and then it should be ready for review.

scorbajio:
There is a single client test still failing - it's the "assembleBlocks() -> with a hardfork mismatching tx" one. The error is "Error: Maximum HF determined by block number/ttd is lower than timestamp HF". It can't be resolved (to my knowledge) by using a custom chain since the test is testing if a transaction with a mismatching hardfork will be properly ignored, so if you use a custom chain for creating the transaction, it won't have any hardforks to be set to for the test.

g11tech (approved):
lgtm

PR #2818, opened 2023-06-23
Fix retesteth runner

jochem-brouwer:
WIP. And this is likely temporary since if we switch tests to `import x from foo.js` then it will likely start to fail too.

### DIFF
--- a/packages/vm/test/retesteth/transition-child.cts
+++ b/packages/vm/test/retesteth/transition-child.cts
@@ -8,7 +8,8 @@ import { hexToBytes } from 'ethereum-cryptography/utils'
 import { readFileSync, writeFileSync } from 'fs'
 import { join } from 'path'
 
-import { BlockBuilder, VM } from '../../dist/cjs'
+import { VM } from '../../dist/cjs'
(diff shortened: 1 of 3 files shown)


packages/vm/test/retesteth/transition-child.cts
```diff
@@ -8,12 +8,12 @@ import { hexToBytes } from 'ethereum-cryptography/utils'
 import { readFileSync, writeFileSync } from 'fs'
 import { join } from 'path'
 
-import { VM } from '../../src'
-import { BlockBuilder } from '../../src/buildBlock'
+import { VM } from '../../dist/cjs'
```
jochem-brouwer, line 11: Should we import from src? (as we did before?)

jochem-brouwer:
I think this is ready for review.

packages/vm/test/retesteth/transition-child.cts
acolytec3: You know or may not matter since we are using `cts` to force it to be CommonJS. But let's be consistent one way or the other

g11tech (approved):
lgtm

PR #2819, opened 2023-06-23
util: Async event emitter tests

scorbajio:


holgerd77:
Updated this via UI

holgerd77 (approved):
Nice, definitely useful to have this code port tested a bit, to prevent unrealized breakage along maintenance updates e.g.!

Will merge.

PR #2821, opened 2023-06-25
Trie: implement createRangeProof v4

jochem-brouwer:
See: #2672 (this PR is messed up), and commit refs (this PR has them all squashed) at #2820

----

This PR implements `Trie` class method:  `createRangeProof`.

***createRangeProof***

- params:
  - startingKey: Uint8Array
  - limitKey: Uint8Array

- returns
  - keys: Uint8Array[ ]
  - values: Uint8Array[ ]
  - proof: Uint8Array[ ] | null

*details*

1. Lookup path to `startingKey`
  - If no node exists for `startingKey`, 
    - If starting key === '0x000...'
      - return `AllElementsRangeProof`
    - else 
      - Walk back up through path to find the first value node with a key higher than the `startingKey`
      - Treat this node as the new starting point

2. Lookup path to `limitKey`
  - If no node exists for `limitKey`
    - Walk back up through path to find the first value node with a key higher than the `limitKey`
    - Treat this node as the new range limit
    - If no node is found, the range effectively becomes all nodes to the right of `startingKey`
     
 3. Walk the section of the Trie between `startingNode` and `limitNode`, collecting the `keys` and `values`, and the total set of nodes included in the proofs for each of them.

4. Return `RangeProof` object:
```
  keys: Uint8Array[]
  values: Uint8Array[]
  proof: Uint8Array[]
```

**If 0 value nodes exist between `startingKey` and `limitKey`, return *zeroElementsProof***
**`{keys: null, values: null, proof: null}`**

**If 1 value node exists between `startingKey` and `limitKey`, return *oneElementProof***
**`{keys: [ node.key ], values: [ node.value ] , proof: trie.createProof(node)  }`**

**If `startingKey` = `0x000...000` and `limitKey` = `0xFFF...FFF`, return *allElementsProof***
**`{keys: [ *all keys* ], values: [ *all values* ], proof: null}`**


jochem-brouwer:
Before I continue locally, 2ce929be24b56e64422d6f43962b679005e61270 is a commit which passes the (old) tests.

Not included: empty proofs where trie still has a key with value right of the limitHash (this is a prerequiste for snap sync: always report at least one key-value, or none for a non-existence proof of items right of limitHash)

ScottyPoi:
@jochem-brouwer 

i'm a little confused by the comments, and labeling here (v4...?)
happy to continue work on it, i just want to understand the status so i work on the right thing :smirk:

jochem-brouwer:
This v4 is the fourth time this PR has been cherry-picked over others.

The comment about "logic of finding right key" is such that, if one requests a range of nodes, and no nodes are found, then in the snap protocol the key right of the "limitHash" key should be returned. Thus we need logic such that given key X, one returns the key (plus value) of the key right of X (so with key higher than X).

The other logic is about using `nibblesCompare` or `compareBytes` (these methods are slightly different) to make the logic correct.

ScottyPoi:
> Thus we need logic such that given key X, one returns the key (plus value) of the key right of X (so with key higher than X).

:+1:   i can tinker around with a couple approaches to that, and see what works well.

a couple questions
1. are we trying to implement this *generally* for MPT with keys of any size?   or can we assume these will be hashed 32 bytes keys?
2. do we just need to return the key/value? or will we ultimately want the path/proof as well?

packages/trie/test/proof.spec.ts
```diff
@@ -143,3 +154,283 @@ describe('simple merkle proofs generation and verification', () => {
     assert.equal(bytesToUtf8(val!), 'c')
   })
 })
+
+describe('createRangeProof()', function () {
```
ScottyPoi, line 158: Any particular reason to add this test here and not in `test/proof/range.spec.ts`?

jochem-brouwer: No reason. This PR is rather dirty. It should be cleaned up :)

jochem-brouwer:
> > Thus we need logic such that given key X, one returns the key (plus value) of the key right of X (so with key higher than X).
> 
> 👍 i can tinker around with a couple approaches to that, and see what works well.
> 
> a couple questions
> 
> 1. are we trying to implement this _generally_ for MPT with keys of any size?   or can we assume these will be hashed 32 bytes keys?
> 2. do we just need to return the key/value? or will we ultimately want the path/proof as well?

1) Regarding this: this is non-trivial. If we have keys of equal size (e.g. 32 bytes), then it is well defined if keys are higher/lower than other keys. For tries with dynamically sized keys: is `0x01` higher than `0x0100`? Or Is it higher than `0x0001`? This is not trivial and is currently not implemented (note: Geth also does not implement it)

2) This is part of the get range proof logic: if there are no key/values found in the range, then return the rightmost key/value. So this is part of the protocol and not necessarily part of the method.

ScottyPoi:
@jochem-brouwer 

Take a look at this when you get the chance.  
I'm doing a bit more cleanup on the PR, but `createRangeProof` is passing the tests that were already written when I picked this up.  

There are some obvious in-efficiencies to the current solution that I could continue to work on, but if you agree that the method works as is, we could address that stuff in a future PR

jochem-brouwer:
@ScottyPoi I will review today. I just started on it, but I noticed there are not many comments around, which makes it rather hard to review, since some lines are rather complex and I have a hard time figuring out what they do :thinking: Could you add some comments + documentation of the added methods?

packages/trie/src/proof/range.ts
```diff
@@ -382,6 +384,180 @@ async function hasRightElement(trie: Trie, key: Nibbles): Promise<boolean> {
...
+    try {
+      const b = node.getBranch(i)
+      if (b !== null) {
+        return [i, b]
+      }
+    } catch (err) {
+      //
+    }
+  }
+  return undefined
```
jochem-brouwer, line 408: There is syntax somewhere `hasRightBranch() ?? null`, so maybe this should be changed to `return null`.

packages/trie/src/proof/range.ts
```diff
@@ -382,6 +384,180 @@ async function hasRightElement(trie: Trie, key: Nibbles): Promise<boolean> {
...
+      if (isValueNode(rightNode!)) {
+        return {
+          node: rightNode!,
+          currentKey: [...current, rightBranch[0]],
+        }
+      }
+    }
+    let i = childIdx + 1
+    while ((rightBranch === null || rightBranch.length <= 0) && i < 16) {
+      rightBranch = (par as BranchNode).getBranch(i) ? [i, (par as BranchNode).getBranch(i)!] : null
```
jochem-brouwer, line 514: Is this the same as `hasRightBranch` here?

packages/trie/src/proof/range.ts
```diff
@@ -520,3 +701,121 @@ export async function verifyRangeProof(
...
+  if (startingNode !== null && limitNodeRight === null) {
+    const walk = this.walkTrieIterable(this.root())
+    for await (const { node, currentKey } of walk) {
+      if (isValueNode(node)) {
+        const nodeKey = [...currentKey, ...(<any>node)._nibbles]
+        if (nibblesCompare(nodeKey, bytesToNibbles(startingKey)) >= 0) {
+          keyvals.set(bytesToHex(nibblestoBytes(nodeKey)), node._value!)
+          const path = await this.findPath(nibblestoBytes(nodeKey))
+          for (const p of path.stack) {
+            proofSet.add(bytesToHex(p.serialize()))
```
jochem-brouwer, line 771: I'm not sure, what does this do? Prove the start key? I think we only need the key of the rightmost key (so in this cast, left of `limitKey`)

packages/trie/src/proof/range.ts
```diff
@@ -520,3 +701,121 @@ export async function verifyRangeProof(
...
+    // If the starting node does not exist, find the next value node to the right
+    const rightNode = await returnRightNode.bind(this)(bytesToNibbles(startingKey))
+    if (rightNode !== null) {
+      proofSet.add(bytesToHex(rightNode.node.serialize()))
+      startingKey = nibblestoBytes([...rightNode.currentKey, ...(rightNode.node as any)._nibbles])
+
+      keyvals.set(bytesToHex(startingKey), rightNode.node.value()!)
+      startingNode = rightNode.node
+    } else {
+      //
```
jochem-brouwer, line 757: Should something happen here?

packages/trie/src/proof/range.ts
```diff
@@ -520,3 +701,121 @@ export async function verifyRangeProof(
...
+      ...(limitNodeRight.node as any)._nibbles,
+    ])
+    limitNodeRight.node._value &&
+      keyvals.set(
+        bytesToHex(
+          nibblestoBytes([...limitNodeRight.currentKey, ...(limitNodeRight.node as any)._nibbles])
+        ),
+        limitNodeRight.node.value()!
+      )
+    proofSet.add(bytesToHex(limitNodeRight.node.serialize()))
```
jochem-brouwer, line 804: Nothing is being done later on with this proof set?

packages/trie/src/proof/range.ts
```diff
@@ -520,3 +701,121 @@ export async function verifyRangeProof(
...
+        values: [startingNode!._value as Uint8Array],
+        keys: [startingKey],
+        proof: [...path.stack.map((p) => p.serialize()), startingNode.serialize()],
+      }
+    }
+  }
+  return {
+    values: [],
+    keys: [],
+    proof: await this.createProof(startingKey),
```
jochem-brouwer, line 819: What is the case where this return statement gets reached?

packages/trie/src/trie.ts
```diff
@@ -97,13 +97,13 @@ export class Trie {
 
     if (opts?.db !== undefined && opts?.useRootPersistence === true) {
       if (opts?.root === undefined) {
-        const rootHex = await opts?.db.get(bytesToUnprefixedHex(key), {
+        const rootHex = await opts?.db.get(bytesToHex(key), {
```
jochem-brouwer, line 100: Why is this changed to `bytesToHex`?

packages/trie/test/proof/rightNode.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  })
+  for (const { key, value } of entries) {
+    await trie.put(key, value)
+  }
+
+  const sortedByKey = entries.sort((a, b) =>
+    nibblesCompare(bytesToNibbles(a.key), bytesToNibbles(b.key))
+  )
+  it('passes', () => {
+    expect(true).toBe(true)
```
jochem-brouwer, line 23: What is this test?

packages/trie/test/proof/rightNode.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  })
+  for (const { key, value } of sortedByKey) {
+    it(`returns RightNode for key ${key} (${value})`, async () => {
+      const val = await trie.get(key)
+      const node = (await trie.findPath(key)).node
+      expect(val).toEqual(value)
+      expect(isValueNode(node!)).toBe(true)
```
jochem-brouwer, line 30: Ok, but the final key should not return any node, right?

jochem-brouwer: Oh wait, this tests `isValueNode`. Test name should be updated.

packages/trie/test/proof/rightNode.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+})
+
+describe('hasRightBranch', async () => {
+  const branchNode = new BranchNode()
+  for (const i of Array.from({ length: 7 }, (_, i) => i)) {
+    branchNode.setBranch(i, Uint8Array.from([i]))
+  }
+  branchNode.setBranch(8, Uint8Array.from([8]))
+  for (const i of Array.from({ length: 16 }, (_, i) => i)) {
```
jochem-brouwer, line 41: This is very fancy syntax for `for (let i = 0; i < 16; i++)`

packages/trie/test/proof/rightNode.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+      })
+    } else {
+      const rightNode = await returnRightNode.bind(trie)(
+        bytesToNibbles(nibblestoBytes([0, i, i, i]))
+      )
+      it(`returnRightNode for ${i} = ${i === 5 ? i + 2 : i + 1}`, async () => {
+        expect(rightNode, ``).not.toBeNull()
+        i === 5
+          ? expect(rightNode?.node.value()).toEqual(Uint8Array.from([i + 2]))
+          : expect(rightNode?.node.value()).toEqual(Uint8Array.from([i + 1]))
```
jochem-brouwer, line 79: I have no idea what's going on here. How does the trie look? Why are there conditionals here for these values? Why is index 5 special? I don't see this in the trie setup.

packages/trie/test/proof/range.spec.ts
```diff
@@ -481,4 +838,64 @@ describe('simple merkle range proofs generation and verification', () => {
...
+                assert.ok('succesfully verified')
+              } catch (e: any) {
+                assert.fail('could not verify')
+              }
+            }
+          }
+        }
+      }
+    },
+    { retry: 3 }
```
jochem-brouwer, line 899: We should definitely not retry this test and fail immediately if a random trie + proofs fail.

jochem-brouwer (commented):
Okay first round, mainly a lot of (initial) questions. I think first this PR needs to be cleaned up + documented before I can do a more complete review. Also, the changes `bytesToHex` instead of to unprefixed should be rolled back (this is an internal optimization and is intended to use the deprecated to unprefixed hex).

packages/trie/src/proof/range.ts
```diff
@@ -382,6 +384,180 @@ async function hasRightElement(trie: Trie, key: Nibbles): Promise<boolean> {
...
+    }
+  }
+  return undefined
+}
+
+export async function walkRight(this: Trie, key: Nibbles, hash: Uint8Array, childIdx: number) {
+  const rightWalk = this.walkTrieIterable(hash, [childIdx])
+  for await (const { node, currentKey } of rightWalk) {
+    if (nibblesCompare(key, [...currentKey, ...('_nibbles' in node ? node._nibbles : [])]) < 0) {
+      if (node instanceof LeafNode && node.value()!.length < 0) {
```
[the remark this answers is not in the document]
ScottyPoi: ah, no -- leftover hack from an `indexOf` that no longer exists

packages/trie/test/proof/rightNode.spec.ts
acolytec3: > true -- i was under the impression that `for (let i...)` had gone out of fashion.  i am ambivalent

I just have missed the memo. I still do that all the time.

holgerd77: Yes, please change back. I couldn't even read this new thing. 😂

packages/trie/test/proof/rightNode.spec.ts
ScottyPoi: sometimes i just need an easy win

packages/trie/test/proof/rightNode.spec.ts
ScottyPoi: yeah -- i have no idea... :facepalm:

packages/trie/test/proof/rightNode.spec.ts
ScottyPoi: *
                                          BRANCH_NODE
    0     1     2     3     4     5     6     7    8    9    10    11   12   13   14   15
    *     *     *      *     *     *     _     *     _    _      _      _      _      _     _      _
    ^     ^     ^      ^     ^     ^            ^
                       LEAF_NODES
    
    
    We have a branch node, with leaves on branches 0, 1, 2, 3, 4, 5, and 7.  This tests the basic ability of `returnRightNode` to work from different starting conditions:
    
  -  Keys 0 - 4 :   There is a node in the Trie at that address, and the next node to the right is one index higher.
  -  Key 5:  There is a node in the Trie at that address, and a node to its right, but with an empty branch between them.  This tests that `rightNode` isn't simply looking 1 index higher, but will continue searching to find a value node.
  - Key 6:  There is NOT a node in the Trie at that address, but there is a node to its right.  This tests that `rightNode` can start its search correctly with or without finding a node for that key.
  - Key 7, there is a node at the address, but 0 nodes to the right.
  - Key 8 - 15, there is no node at the address, and 0 nodes to the right.

packages/trie/src/proof/range.ts
```diff
@@ -382,6 +384,180 @@ async function hasRightElement(trie: Trie, key: Nibbles): Promise<boolean> {
...
+        ) {
+          return { node, currentKey: [] }
+        }
+      }
+    }
+  }
+  return null
+}
+
+export async function rightNodeFromNull(this: Trie, key: Nibbles, path: Path) {
```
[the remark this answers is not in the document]
ScottyPoi: Nothing really.  I think we'll always be looking for 'rightNodeFromNull` in the `rangeProof` methods anyway.

packages/trie/src/proof/range.ts
```diff
@@ -382,6 +384,180 @@ async function hasRightElement(trie: Trie, key: Nibbles): Promise<boolean> {
...
+    }
+  }
+  return null
+}
+
+export async function rightNode(
+  this: Trie,
+  key: Nibbles,
+  path: Path
+): Promise<{ node: TrieNode; currentKey: Nibbles } | null> {
```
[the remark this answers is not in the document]
ScottyPoi: If the keys are all equal size, then yes, because all value nodes will terminate.

If keys were not the same size, then we would end up finding value nodes that did not terminate, and end up with leftover nibbles.

packages/trie/test/proof/range.spec.ts
ScottyPoi: Yes, I thought you wrote it...

packages/trie/src/proof/range.ts
```diff
@@ -520,3 +701,121 @@ export async function verifyRangeProof(
...
+}> {
+  if (nibblesCompare(bytesToNibbles(startingKey), bytesToNibbles(limitKey)) > 0) {
+    throw new Error('invalid keys order')
+  }
+  let startingNode = (await this.findPath(startingKey)).node
+  let limitNode = (await this.findPath(limitKey)).node
+
+  if (
+    startingNode === null &&
+    limitNode === null &&
```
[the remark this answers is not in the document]
ScottyPoi: Yes, i should think so.

packages/trie/src/proof/range.ts
ScottyPoi: Do you mean that all we actually need to return is a proof for the limit-node?
Yes that would be much smaller....and make a lot more sense

jochem-brouwer: Yes, only the proof of the limitHash should be included in the range proof

packages/trie/test/proof/rightNode.spec.ts
jochem-brouwer: I think this looks somewhat similar to how you use for loops in python (i.e. `range()`) however this is not necessary in JS :)

packages/trie/test/proof/range.spec.ts
```diff
@@ -481,4 +838,60 @@ describe('simple merkle range proofs generation and verification', () => {
 
     assert.equal(await verify(trie, entries, 0, entries.length - 1), false)
   })
+
+  it('passes randomly created tries with randomly selected ranges', async () => {
```
jochem-brouwer, line 842: I just resolved the other convo because I thought the copied test was solved, but it seems not. This test seems to appear twice? (and yes, I wrote it :smile: )

packages/trie/test/proof/rightNode.spec.ts
scorbajio: points for creativity

ScottyPoi: > points for creativity

like the points you get for hitting a golf ball into a tree...

packages/trie/src/proof/range.ts
```diff
@@ -382,6 +384,180 @@ async function hasRightElement(trie: Trie, key: Nibbles): Promise<boolean> {
...
+      //
+    }
+  }
+  return undefined
+}
+
+export async function walkRight(this: Trie, key: Nibbles, hash: Uint8Array, childIdx: number) {
+  const rightWalk = this.walkTrieIterable(hash, [childIdx])
+  for await (const { node, currentKey } of rightWalk) {
+    if (nibblesCompare(key, [...currentKey, ...('_nibbles' in node ? node._nibbles : [])]) < 0) {
```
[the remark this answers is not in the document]
ScottyPoi: walk iterator returns a stream of:
`{ node, currentKey }`

`currentKey` = nibbles collected on path to `node`.  Essentially the "full key" address of the parent of `node` ( + `branch_index` if parent is `branch` node)

`...('_nibbles' in node ? node._nibbles : [])]` 
is just a compact way of saying "if `node` is a leaf node, add the remaining nibbles, but if it's a branch node, don't even look for nibbles, just pretend it's an empty array".

In practice, though, since we're using hashed keys of equal length, this conditional should not be necessary, because all value nodes should be Leaf nodes.  

using nibbles compare here works because we're comparing to the actual key of the starting node that we found, which should always be the same length as the full key of the node.

holgerd77:
What is the state of this PR??

jochem-brouwer:
I think we should drop this PR. The idea of all this createRangeProof was to create an intermediate `createRangeProof` before we had a flat database. Since this is getting there soon (TM), the flat DB implementation is more trivial than using the current tree-based library. (I initially thought with a tree database, it would not be trivial, but also not super complex, but this turned out to be rather complex especially with these edge cases)

jochem-brouwer:
We can address this with snap DB.

PR #2825, opened 2023-06-26
Hive: fix `pyspec` tests

jochem-brouwer:
We currently fail 20/98 pyspec tests.

Fixes:

- `eth_getStorageAt` assumes that the values returned from `StateManager.getStorageAt` are actually RLP-encoded. This is not the case. (Why were tests passing before? Well, the encoded value was 0x64, which is RLP-encoded and RLP-decoded the same, so for that case it works)

To run in hive:

`./hive --client ethereumjs --sim ethereum/pyspec`


acolytec3 (approved):
LGTM!  I ran the pyspec tests and confirm that only 2 fail.

PR #2826, opened 2023-06-26
VM/SM/Blockchain/Trie/EVM: Copy() -> shallowCopy() function renaming

scorbajio:
TODO:
- [x] VM (see #2833)
- [x] StateManager
- [x] Blockchain
- [x] Trie
- [x] Evm
- [ ] Ethash


holgerd77:
Quick note: after an exchange with @jochem-brouwer we realized that the EVM copy() method also has state and therefore should be renamed as well.

packages/client/lib/sync/fetcher/accountfetcher.ts
```diff
@@ -0,0 +1,98 @@
+import { Fetcher, FetcherOptions } from './fetcher'
+import { Account } from '@ethereumjs/util'
+// import { Chain } from '../../blockchain'
```
gabrocheleau, line 3: can be removed?

scorbajio: yea I don't know how this crept it's way in. will remove.

holgerd77:
Did a branch update here via UI

holgerd77:
Another branch update via UI

holgerd77 (approved):
This looks pretty good, thanks a lot! 🤩

One thing: it would be good if we give this a mention in the direct code docs above the - now - `shallowCopy()` methods - what the "shallow" part of the copy actually is, so that this gets a bit of a description, something like the following for EVM:

```typescript
/**
* This method copies the EVM, current HF and EIP settings
* and returns a new EVM instance.
* 
* Note: this is only a shallow copy and both EVM instances
* will point to the same underlying state DB.
* 
* @returns EVMInterface
*/
```

Will merge though, this can be done in a small follow-up PR (also ok: alongside another PR).

scorbajio:
Working on documentation update in #2855.

PR #2830, opened 2023-06-27
Util, Monorepo: hexStringToBytes -> prefixedHexStringToBytes

### LINKED ISSUE #2827: Util: rename/adjust hexStringToBytes to prefixedHexStringToBytes
We have made bad experiences with being loose on hex string prefixes - also on the input side - and allow for lazy input and addition or removal, getting more strict here was therefore a substantial part of one of former breaking release rounds.

The current `hexStringToBytes()` method *does* a lazy `if (hex.startsWith('0x')) { hex = hex.slice(2) }` conversion here.

This leads to the problematic situation that it is just not deterministically decidable if a `0x` string is either a prefix or part of the string itself, in case both is allowed.

So let's say one has the - let's assume - prefixed hex string:

```typescript
0x0x1234
```

The following would then be the - **unprefixed** hex string:

```typescript
0x1234
```

So it is just not decidable if `0x` is part of the string or it is a prefix and it is highly recommendable to request from users to decide on the format before-hand. The prefixed version is preferred here since once can always throw if a non-prefixed version is provided, and so do an always-working format check.

This is not only a theoretical example. People *can* put arbitrary byte data on some fields (e.g. `extraData`) or might even do this on purpose 😋 to annoy developers and cause things to break.

So, long story short: we should definitely update here before the breaking releases and rename to `prefixedHexStringToBytes()`, remove the `0x` stripping and throw for all non-0x-strings passed.

I will already add/reference the updated name to the release notes.

jochem-brouwer:
Closes #2827 

Note: if we have methods which return hex strings, the return type should be either `PrefixedHexString` or `NotPrefixedHexString`. Rewriting this in all libraries was fine, except for client, since I am not sure if the return values of the methods are prefixed or not, which is very error-prone.

TODO:

- [x] Ensure client runs (try to sync mainnet)
- [x] Fix tests (I ran them locally and thought they passed :cry: )


packages/trie/src/db/checkpoint.ts
```diff
@@ -1,10 +1,30 @@
...
 import type { Checkpoint, CheckpointDBOpts } from '../types.js'
 import type { BatchDBOp, DB, DelBatch, PutBatch } from '@ethereumjs/util'
 
+/**
+ * In order to not-prefix the key/values in the DB each time (this copies the key/values each time: expensive)
+ * Use this "private" method (see `prefixedHexStringToBytes` from @ethereumjs/util for reference)
+ * @param hex
+ * @returns
+ */
+function _nonPrefixedHexStringToBytes(hex: string) {
```
jochem-brouwer, line 14: Alternative: move this to util but shield it so no one uses it

jochem-brouwer:
Ok, have renamed all, tests should be fixed, however, I am totally not sure regarding client, since we have either hex-prefixed strings or non-hex-prefixed strings there, so we should take a thorough look there (or: alternatively, temporarily use an internal `hexStringToBytes` method which reflects the old version of util: check if hex prefixed then strip the 0x, otherwise immediately convert).

packages/util/src/bytes.ts
```diff
@@ -19,15 +19,18 @@ export const bytesToPrefixedHexString = (bytes: Uint8Array): string => {
...
-export const hexStringToBytes = (hex: string): Uint8Array => {
+export const prefixedHexStringToBytes = (hex: string): Uint8Array => {
   if (typeof hex !== 'string') {
     throw new Error(`hex argument type ${typeof hex} must be of type string`)
   }
 
-  if (hex.startsWith('0x')) {
-    hex = hex.slice(2)
+  if (!hex.startsWith('0x')) {
+    console.trace()
```
jochem-brouwer, line 28: Oops, debug artifact :sweat:

gabrocheleau:
Minor question, but since we're making that change, would it be better to name this:
`unprefixedHexString` or 
`notPrefixedHexString`?
The first (`unprefixed`) sounds more fluent to me. And googling unprefixed seems to indicate that it's more commonly used in a coding context.

gabrocheleau:
> > we're making that change, would it be better to name this:
> > `unprefixedHexString` or
> 
> I'm not sure what you are referring to here?
> 
> The PR is open for review now, just tested locally that client works (you might have to use a custom datadir or wipe your datadir)

Oh sorry, I thought you had implemented explicit naming for prefixed and unprefixed hex strings, had only looked at the description and not the code. Will review later!

packages/vm/test/retesteth/README.md
```diff
@@ -10,7 +10,7 @@ Note: All paths are relative paths to the `VM` package root.
 
 1. If you change the port number in `transition-cluster.ts` to anything other than 3000 (or run `transition-cluster` on a separate machine or different IP address from retesteth), update `test/vm/retesteth/clients/ethereumjs/start.sh` to reflect the right IP and port.
 
-2. From VM package root directory, run `npx ts-node test/retesteth/transition-cluster.ts`
+2. From VM package root directory, run `npx ts-node test/retesteth/transition-cluster.cts`
```
jochem-brouwer, line 13: This is not relevant for this PR, but added it here, since it felt silly to open another PR for this.

holgerd77 (changes requested):
I would really prefer if we do *not* introduce an extra type `UnprefixedHexString` and so to not encourage/diffuse people to go into a wild mix again.

If we have - for whatever reason - this 1-2 occasions where it is necessary that we use ourselves we should rather keep this local.

holgerd77:
Otherwise this looks really great! 🤩 Wasn't aware that this has gotten so extensive!

jochem-brouwer:
Alright, I will remove this `UnprefixedHexString` (it has no use in VSCode since it still shows up as `string` (*sigh*)). I kept it mainly as tracker/reminder for myself but I will remove them.

And regarding the amount of changes: yes this was much more work than I anticipated :sweat_smile: I thought I could just "find-and-replace" but that was not the case :joy:

jochem-brouwer:
Ok, removed (I thought I had put it in much more places :joy: )

holgerd77 (dismissed):
LGTM

packages/block/test/eip4895block.spec.ts
```diff
@@ -50,7 +52,7 @@ describe('EIP4895 tests', () => {
       () => {
         BlockHeader.fromHeaderData(
           {
-            withdrawalsRoot: hexStringToBytes('00'.repeat(32)),
+            withdrawalsRoot: prefixedHexStringToBytes('0x' + '00'.repeat(32)),
```
gabrocheleau, line 55: Could use the `zeros` util

gabrocheleau: (For this and other ones in the file)

packages/trie/test/proof/range.spec.ts
```diff
@@ -141,8 +141,8 @@ describe('simple merkle range proofs generation and verification', () => {
     }
 
     // Special case, two edge proofs for two edge key.
-    const startKey = hexStringToBytes('00'.repeat(32))
-    const endKey = hexStringToBytes('ff'.repeat(32))
+    const startKey = prefixedHexStringToBytes('0x' + '00'.repeat(32))
```
gabrocheleau, line 144: zeros util

gabrocheleau (approved):

gabrocheleau (approved):

gabrocheleau (approved):

gabrocheleau:
One test failing, looking into it.

gabrocheleau:
> One test failing, looking into it.

Ah nvm, just noticed it is also failing on master.

packages/trie/test/proof/range.spec.ts
jochem-brouwer: I think in this case it is cleaner to keep it like this, since `endKey` uses `prefixedHexStringToBytes` and `startKey` would now use `zeros` - makes it slightly harder to read (I do agree with block tests though these should+will be changed in 25f29dc9fbbbdf3d75b080bb9803e8a6eb97f7ad)

holgerd77 (approved):
LGTM

PR #2833, opened 2023-06-27
VM: research copy behavior

jochem-brouwer:
VM has weird copy behavior, it does not use `evm.copy()` but instead uses re-uses some of the copy logic and then performs the copy itself. This PR researches the behavior of `VM.copy()`

Note: the EVM copy behavior is from this PR: #2529, most likely that on a fork switch the EVM pointed to the wrong HF than VM, and thus needs to point to the same common.


scorbajio:
Can this PR be closed?

PR #2834, opened 2023-06-27
Replace `rustbn.js` with wasm-compiled `rustbn` module

acolytec3:
Replaces the `rustbn.js` dependency with a newer version compiled from WASM that works in browser and in Nodejs

- Replaces `rustbn.js` with [`rustbn.wasm`](https://github.com/acolytec3/rustbn.wasm)
- Updates global `module` setting to `Node16` - needed for the WASM instantiation bit in `rustbn.wasm` (though we might be able to revert if the team doesn't want to make this change now; this also has Typescript use Node's native ESM loader which feels like the right way forward for us
- Turns `esModuleInterop` back on for `client` so since `Node16` disallows the `import something = require('someModule')` Typescript syntax.
  - Client imports have been updated to use the altered syntax `import someData from 'some.json'` or `import someDefault from 'someModule'` and removes namespace imports where needed (mainly `import * as tape from 'tape'`)

PR #2842, opened 2023-06-29
Revert default to namespace imports

acolytec3:
Not important changes: 
Revert `esModuleInterop` flag in `client's` tsconfig and then fix **many** regressions in #2834 where namespace imports were converted to default imports which breaks when running in native nodejs.

Important changes:
1. Adjust the import syntax for `debug` to 
```js
const { debug: createDebugLogger } = debugDefault
```
to make the node ESM loader happy.

2. Adjust import syntax for `qheap` to:
change `import qheap = require('qheap')` (which is an odd Typescript only syntax that breaks when you switch tsconfig's module resolution to `Node16` to
```js
import type QHeap from 'qheap'
const Heap = require('qheap')
```
and then adjust `Heap` constructor calls to something like:
```js
    const byPrice = new Heap({
      comparBefore: (a: TypedTransaction, b: TypedTransaction) =>
        this.normalizedGasPrice(b, baseFee) - this.normalizedGasPrice(a, baseFee) < BigInt(0),
    }) as QHeap<TypedTransaction>
```

It's slightly more verbose but makes the Nodejs ESM loader happy.

3.  Adds a new `cli.spec.js` under the client `test/integration` folder that is invoked in CI as the `test-client-cli` job (replacing the `test/sim/mainnet.spec.ts` job.  The logic here is that the new `cli.spec.js` simply runs the client in native JS rather than via ts-node and is a canary test to verify that changes we've made don't break running the client directly in NodeJS without the crutches that ts-node sometimes provides (which have masked various import errors in the past)


holgerd77:
Client tests are failing here, so I assume for now this is not ready yet.

acolytec3:
> Client tests are failing here, so I assume for now this is not ready yet.

Right.  I'm doing some more experimentation locally to figure the best way to resolve.  I need to revert a few things I changed in #2834

jochem-brouwer:
If possible and not super-time-consuming could we also add a CI test for this here?

(I assume this is to check if our client runs in JS mode ..?)

holgerd77 (approved):
LGTM

PR #2845, opened 2023-06-30
Util: bytes cleanup + do not import from ethereum-cryptography/utils

### LINKED ISSUE #2843: Util/Monorepo: Clean-Up bytes module Usage / Remove ethereum-cryptography helper usage
With the new releases the `@ethereumjs/util` `bytes` module helper methods like `prefixedHexStringToBytes()` will get an extremely prominent role when using our library.

We should therefore make sure, that things are really consistent here and this provides a consistent basis for conversions with a stable API to use.

Not a super big job, we are mostly there and Andrew laid out great foundations here.

Nevertheless some left over TODOs I realized when writing release notes and bringing this together in #2832:

- [x] (1) Maybe the biggest one: we use and also re-export in Util some helper methods from `ethereum-cryptography`, namely `bytesToHex`, `bytesToUtf8`, `concatBytes`, `equalsBytes`, `utf8ToByte`. Throughout our libraries we nevertheless have a wild mix of using these methods directly by importing from `ethereum-cryptography`, so e.g. by having `import { bytesToHex, equalsBytes, hexToBytes } from 'ethereum-cryptography/utils.js'` in `Blockchain`, just to give a random example. This should be unified and cleaned up to only use the imports from `Util`, so that this doesn't establish as a common practice to do this wild mix here
- [x] (2) The `bytesToHex` re-exported from `ethereum-cryptography` is (somewhat) redundant respectively inconsistent with our `0x` prefix handing and current naming scheme. The re-export should be removed in favor of a `prefixedHexStringToBytes` usage (I have looked through the occurrences, a lot of places a "just" console.log() or debug outputs and can be easily converted, some usages need some relatively close look though to not miss something (should nevertheless be done), e.g. if hex strings are used as keys for a cache (needs some close look on both all writes and accesses so to make sure that cache keys are accidentally never found again by always looking for the unprefixed version at some place))
- [x] (3) The methods `bytesToBigInt`, `bigIntToBytes`, `bytesToInt` are misplaced (in the `bytes` module) and should be moved up and re-ordered properly
- [x] (4) Function declaration format is a wild mix of the short and `function` containing format, this should be unified to the `() => {}` format (so e.g. `export const bigIntToBytes = (num: bigint): Uint8Array => {...}`
- [x] (5) Discussion: is the `concatBytesNoTypeCheck()` method really worth to keep? 🤔 From the code docs: "Can give small performance gains on large arrays". From the formulation this sounds like an eventual not so worth trade off, can't judge on this though, might be also just a formulation thing and gains are more significant in reality. Likely @acolytec3 can shed a light here -> Update: decision to simplify (only use the concatBytesNoCheck method, rename this to concatBytes and skip the re-export from ethereum-cryptography), see thread below
- [x] (6) The `toUtf8()` method doesn't fit into the naming scheme and is not used anywhere in the code base. I would pledge to have this removed
- [x] (7) `bigIntToHex()` name is inconsistent and might need to be renamed to `bigIntToPrefixedHexString()` (pending the below discussion)
- [x] (8) The `compareBytes()` method seems to be not used within the whole code base. Is this necessary?

That should be it. Lol. 😋

And *then* - since we now have a unifed way of working with 0x-prefixed hex strings within the whole code base we can finally discuss (came to this in a second round after writing this together)`:

Do we *now* actually want to simplify these `bytesToPrefixedHexString()` names again and go with the shorter version? 🤔

So this would be:

- `bytesToPrefixedHexString()` -> `bytesToHex()`
- `prefixedHexStringToBytes()` -> `hexToBytes()`
- `intToPrefixedHexString()` -> `intToHex()`
- `unpadHexString()` -> `unpadHex()` (a bit optional maybe)
- `bigIntToHex()` -> can then stay this way! 🙂

I think this would *then* be ok if we have this consistent in the code base and these other names really are "monsters" to deal with. 😆 (So I would have a relatively strong tendency to do)

Side note: I would nevertheless highly recommend to sequentially go through the TODO steps before. This is not so much more extra work and otherwise there is a high danger to not do this right (e.g. by forgetting to go from non-prefixed to prefixed for `bytesToHex` in the first round.

jochem-brouwer:
Closes #2843 

This PR:

- Renames `prefixedHexStringToBytes` to `hexStringToBytes`
- Renames `bytesToPrefixedHexString` to `bytesToHex`
- Ensures that there are no imports from `ethereum-crypto/utils` (re-exported from util's `bytes`)
- All remaining task from the #2843 issue 

TODOs

- [x] All tasks in the linked issue
- [x] Add lint rule preventing imports from `ethereum-cryptography/utils`
- [x] Verify in entire repo there are no more occurences of `prefixedHexStringToBytes` and `bytesToPrefixedHexString`
- [x] ~~Upgrade performance of hexToBytes (do not slice + pad, this copies the string)~~ - will go in a different PR, this commit b079375af115c3b673b7064bf3d26e4267a04bb0 breaks tests. See #2848


jochem-brouwer:
Hmm great, merge conflicts.

Anyways, will stop for today. There are 2 test suits failing: wallet (seems to hang?) and client.

To debug keep this in mind:

The new bytesToHex is what was (and is re-exported) `bytesToUnprefixedHex`. 

The new hexToBytes is what was (and is re-exported) `unprefixedHexToBytes`

jochem-brouwer:
Remove utf8 in 21d26fe21bfd2a2afdad4bbfc1a7bc450ffd6c80

jochem-brouwer:
Performance will go in a different PR because my changes in b079375af115c3b673b7064bf3d26e4267a04bb0 broke tests

packages/client/genesis.json
```diff
@@ -0,0 +1,16 @@
+{
```
scorbajio, line 1: Did you mean to include this file?

packages/client/jwtsecret
```diff
@@ -0,0 +1 @@
+aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
scorbajio, line 1: Did you mean to include this file?

packages/util/src/bytes.ts
```diff
@@ -1,16 +1,35 @@
...
+// eslint-disable-next-line no-restricted-imports
+import {
+  bytesToHex as _bytesToUnprefixedHex,
+  hexToBytes as _unprefixedHexToBytes,
+} from 'ethereum-cryptography/utils.js'
+
+// Note: ethereum-cryptography/hexToBytes is imported as "unprefixedHexToBytes"
+// This name is not correct, since it supports prefixed strings. However, internally here we
+// explicitly use it when the strings are known to be unprefixed (for readability and avoid
+// confuction why we would not use our own `hexToBytes`)
```
scorbajio, line 11: The first word here should be "confusion"?

packages/client/jwtsecret
jochem-brouwer: Ugh this (plus genesis) is from running docker in a `ADD` mode instead of `COPY`. Will remove

packages/util/src/bytes.ts
jochem-brouwer: Yes, fixed

holgerd77 (approved):
OMG is this an insane PR.

Have nevertheless forced me to go at least once through the whole thing.

Beautiful.

🤩

LGTM.

PR #2846, opened 2023-07-01
Fix Genesis parsing for 4844

jochem-brouwer:
@spencer-tb found this bug when running 4844 pyspec tests on hive. We do not parse the geth genesis correctly.

### DIFF
--- a/packages/blockchain/src/blockchain.ts
+++ b/packages/blockchain/src/blockchain.ts
@@ -1322,7 +1322,7 @@ export class Blockchain implements BlockchainInterface {
       number: 0,
       stateRoot,
       withdrawalsRoot: common.isActivatedEIP(4895) ? KECCAK256_RLP : undefined,
-      excessDataGas: common.isActivatedEIP(4844) ? BigInt(0) : undefined,
+      excessDataGas: common.isActivatedEIP(4844) ? common.genesis().excessDataGas : undefined,
     }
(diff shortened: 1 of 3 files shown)


packages/blockchain/src/blockchain.ts
```diff
@@ -1322,7 +1322,7 @@ export class Blockchain implements BlockchainInterface {
       number: 0,
       stateRoot,
       withdrawalsRoot: common.isActivatedEIP(4895) ? KECCAK256_RLP : undefined,
-      excessDataGas: common.isActivatedEIP(4844) ? BigInt(0) : undefined,
+      excessDataGas: common.isActivatedEIP(4844) ? common.genesis().excessDataGas : undefined,
```
jochem-brouwer, line 1325: TODO (check): I think this line is unnecessary, it takes params from `common.genesis()` first and copies those?

acolytec3:
Not sure if there's a specific hive test we should be passing but I was able to run Cancun tests from Spencer's PR 765 that has all the 4844/cancun tests and they pass so seems like this resolves the issue.  Will approve and let @jochem-brouwer merge if I"m correct.

acolytec3 (approved):
LGTM

jochem-brouwer:
Don't have write rights on the external repo, closing in favour of #2847

PR #2854, opened 2023-07-03
util/bytes: make unprefixedHexToBytes specific

jochem-brouwer:
This PR adds an extra guard to the proxy `hexToBytes` from `ethereum-cryptography`. That method supports both prefixed and unprefixed hex strings. The new (deprecated) `unprefixedHexToBytes` now explicitly only accepts unprefixed hex strings.

(Tests seem to pass locally so ready for review)

Follow up of #2845

### DIFF
--- a/packages/evm/src/precompiles/06-ecadd.ts
+++ b/packages/evm/src/precompiles/06-ecadd.ts
@@ -1,4 +1,10 @@
-import { bytesToHex, bytesToUnprefixedHex, short, unprefixedHexToBytes } from '@ethereumjs/util'
+import {
+  bytesToHex,
+  bytesToUnprefixedHex,
+  hexToBytes,
+  short,
+  unprefixedHexToBytes,
+} from '@ethereumjs/util'
 import { ec_add } from 'rustbn-wasm'
 
 import { OOGResult } from '../evm.js'
@@ -24,7 +30,8 @@ export function precompile06(opts: PrecompileInput): ExecResult {
     return OOGResult(opts.gasLimit)
   }
 
-  const returnData = unprefixedHexToBytes(ec_add(inputData))
+  const ret = ec_add(inputData)
+  const returnData = ret.slice(0, 2) === '0x' ? hexToBytes(ret) : unprefixedHexToBytes(ret)
 
   // check ecadd success or failure by comparing the output length
   if (returnData.length !== 64) {
--- a/packages/evm/src/precompiles/06-ecadd.ts
+++ b/packages/evm/src/precompiles/06-ecadd.ts
@@ -1,4 +1,4 @@
-import { bytesToHex, bytesToUnprefixedHex, short, unprefixedHexToBytes } from '@ethereumjs/util'
+import { bytesToHex, bytesToUnprefixedHex, hexToBytes, short } from '@ethereumjs/util'
 import { ec_add } from 'rustbn-wasm'
 
 import { OOGResult } from '../evm.js'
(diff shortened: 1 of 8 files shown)


packages/evm/src/precompiles/06-ecadd.ts
```diff
@@ -24,7 +30,8 @@ export function precompile06(opts: PrecompileInput): ExecResult {
     return OOGResult(opts.gasLimit)
   }
 
-  const returnData = unprefixedHexToBytes(ec_add(inputData))
+  const ret = ec_add(inputData)
+  const returnData = ret.slice(0, 2) === '0x' ? hexToBytes(ret) : unprefixedHexToBytes(ret)
```
jochem-brouwer, line 34: Note: (for all `ec*` files), it seems that if there is a return value, the string is not prefixed, but if there is no return value (error?) then 0x is returned. (So we need to do this check)

jochem-brouwer:
Just verified (see prev. commit): all tests complete on if we merge https://github.com/ethereumjs/rustbn-wasm/pull/5. This PR is on hold until it is merged and released.

jochem-brouwer:
(this PR will need an update to that new pkg version in order to run CI)

packages/util/src/bytes.ts
```diff
@@ -23,7 +18,13 @@ export const bytesToUnprefixedHex = _bytesToUnprefixedHex
 /**
  * @deprecated
  */
-export const unprefixedHexToBytes = _unprefixedHexToBytes
+export const unprefixedHexToBytes = (inp: string) => {
+  if (inp.slice(0, 2) === '0x') {
+    throw new Error('hex string is prefixed with 0x, should be unprefixed')
+  } else {
```
holgerd77, line 24: We can‘t do this check here since an unprefixed hex string can very well start with 0x. 😅 🤯

That‘s one of the reasons why we take prefixed as preferred: one can have a positive check that the format is correct.

holgerd77 (changes requested):

packages/util/src/bytes.ts
holgerd77: Actually that also means that you cannot use the method from ethereum-cryptography, now that I am thinking about it, right? So what you can do is always to add 0x and then use our own prefixed method I would think.

jochem-brouwer: Sorry, I thought we decided we wanted to explicitly check here if it was unprefixed? But, I can change, and then call into our own method (so we do not import `hexToBytes` from `ethereum-cryptography` anymore)

holgerd77: Ah wait, maybe let's give this another round. Or would the Ox-in-string case only count for the utf8ToBytes version? 🤔

acolytec3: >We can‘t do this check here since an unprefixed hex string can very well start with 0x

How is this possible, or have I misunderstood something along the way?  Hex doesn't include `x`, right, so unprefixed would explicitly mean no `0x`, right?

holgerd77: Sorry, I think I am actually off here all the way down.

Have to some extend mixed this up with arbitrary (UTF-8) strings (fee very free to challenge me more on these kind of things, I stated this explicitly in the whole reasoning for the prefixed-hex work).

Anyhow: I think the outcome is (gladly enough, puh) still desired to have this unified since this reduces the risk of errrors when working in an inconsistent way with prefixed and unprefixed hex strings.

Ok, will update the branch here and merge.

holgerd77:
Rebased this via UI

holgerd77 (approved):
LGTM

Issue #2852, opened 2023-07-03
Monorepo: Mark all _* named methods as protected (some less-strict flavor of private)


holgerd77:
Sorry everyone that I am even coming up with *this* one 😂, but the UX/DX benefits are just too dramatic to *not* do it in this round.

Current code completion for - basically - our whole library items/classes look like the following, here for the EVM:

![grafik](https://github.com/ethereumjs/ethereumjs-monorepo/assets/931137/cc52689a-7e96-405f-8ffd-f222bf27060b)

So bascially *EVERTHING* is displayed but *not* the official API. 😆

We should urgently update here, this would so drastically improve the whole experience and make API usage so much better.

Some pre-comments here:

1. We have in the very most cases `_*` names for methods or properties which should not be accessed. Quickly thought if we would want to switch to "normal" naming here, but then I remembered that this distinction is actually good for the JavaScript users where these keywords like `protected` doesn't exist. So we should keep this as the standard.

2. We should switch to `protected` (for the most part) and not to `private`. This also "does the job" and this helps if people want to extend a class or something, so it is just a bit more on the safe side of things.

3. We should nevertheless take this task a bit loose. This will be a lot to do, if we get from 30% to 90% that would be already a huge achievement.

So what we should do is:

- [ ] Go through the libraries and mark all `_*` methods as `protected`, sometimes `private` might also be a good choice. This should be accompanied by a lot of test runs, since one *can* actually break several test setups with this.
- [ ] If there are only 1-2, maybe 3 accesses of some side variable in the tests, we should rather go with  `any` casting or something (or is there a better modifier with some "access nevertheless" for such a case which preserves the typing? 🤔)
- [ ] For more frequently accessed methods/properties (I guess this counts more for the properties part) one can also ask the other way around: does it make sense to expose this within the public API? (this can also be a frequent discussion point during the next days, if someone takes on the task)
- [ ] I guess a special case is `_common` access from other libraries, which is a very common pattern within the monorepo and we might want to discuss separately. Do we want the Common objects to be generally exposed within the API?
- [ ] Also, careful other way around as well: are there non `_*` methods/properties which are in fact rather protected/private? These should then *get* an underscore! I think `DEBUG` is a prominent example.
- [ ] Special (easier) case here: the above and already marked as `protected`. So this should be checked by a package or monorepo search: are all methods/properties marked as `protected` written with `_`?

[...]

PR #2857, opened 2023-07-04
Monorepo: make _* methods protected / _common -> common

holgerd77:
Addresses #2852 

~This is a first test on the above issue to get a feeling on the extent of the task, will start with the `Common` library.~ 

This also refactors the `EventEmitter` structure from `Common` aligning this with EVM and client by separating the `EventEmitter` from the main class and make this an own property to allow for a cleaner API (code completion).

Update: this is now taking on all the libraries. This is on purpose a "rough" update, primarily targeting the main classes where the API is most visible. Everything else would just take too much time and effort at this point.


holgerd77:
Rebased this via UI

packages/vm/test/api/istanbul/eip-152.deactivated.ts
```diff
@@ -97,7 +97,7 @@ describe('Istanbul: EIP-152', () => {
       const res = precompile09({
         data: hexToBytes('0x' + testCase.input),
         gasLimit: BigInt(20),
-        _common: common,
+        common: common,
```
jochem-brouwer, line 100:
```suggestion
        common,
```

Super nit but we can now do this

packages/vm/test/api/istanbul/eip-152.deactivated.ts
```diff
@@ -108,7 +108,7 @@ describe('Istanbul: EIP-152', () => {
       const res = precompile09({
         data: hexToBytes('0x' + testCase.input),
         gasLimit: BigInt(10000000),
-        _common: common,
+        common: common,
```
jochem-brouwer, line 111:
```suggestion
        common,
```

jochem-brouwer (approved):
Two super-nit comments, but will approve anyways since it is just a style-ish comment. LGTM!

packages/vm/test/api/istanbul/eip-152.deactivated.ts
holgerd77: Will ignore this for now, we should really activate linter for test files though, this would then be caught and fixed.

PR #2859, opened 2023-07-05
monorepo/Util: deprecate node 16 / remove Node.js fetch primitive

### LINKED ISSUE #2850: Util/Monorepo: Deprecate Node 16 / Switch to native fetch() in Util
This is a follow-up more specific issue to #2837 , will close the old one.

We have got *one* single Node.js primitive usage left we want to get rid off, which is the usage of the `https` package for the `fetch` implementation in the Util [provider.ts](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/src/provider.ts) package.

For this we decided the following in one of the last calls: Node.js 18 introduces [native fetch](https://blog.logrocket.com/fetch-api-node-js/). We will therefore deprecate (in the sense of: remove) support for Node 16 on this occasion, which only has extended support for a very few months left anyhow, which will the free us to use native fetch in the Util provider module to replace the Node native stuff.

This should be done. 🙂 

The PR should also make sure that CI/Actions is not run on Node 16 any more (then go to Node 18 I would say?), particularly in the Node versions tests Node 16 should be removed.

Update: just did a simple "16" search in the `/.github` folder, actually there are *a lot* of references to Node 16 which can either get updated or on several occasions I think would also just lead to a clean-up/deletion of code parts (if there are conditionals around Node 16 e.g.).

gabrocheleau:
Resolves #2850


gabrocheleau:
To get the CI to pass, someone with the relevant repository permissions would need to update the "required workflows" so that they look for the node18 tests rather than the node16 ones.

acolytec3:
Can you also update the test in `provider.spec.ts` to remove the check for global fetch?

gabrocheleau:
> Can you also update the test in `provider.spec.ts` to remove the check for global fetch?

ah yes good point, done!

holgerd77 (approved):
Thanks for tackling, also for doing this so thoroughly with all the additional clean-ups and findings (resetting the `engines` property in `package.json` is a great and important one e.g.)!

LGTM, have adjusted CI and will merge.

PR #2861, opened 2023-07-05
EVM runCode/runCall type cleanup

### LINKED ISSUE #2860: EVM: Consolidate EVMRunCallOpts/EVMRunCodeOpts types
Small clean-up task: above option types are highly redundand and should be unified by inheritance or something.

(do not have the 100%-full picture: if it is necessary to create one base option type for this this artificial type should *not* be exported, not sure though, only necessary if one-way inheritance doesn't work because there might be distinct properties on each type)

jochem-brouwer:
Closes #2860 

This PR unifies the types for the `EVMRunCodeOpts` and `EVMRunCallOpts`.

Discussion item: in order to not create a base class I had to unify the `pc` option from `runCodeOpts`. I cannot think of a practical reason why someone would want to runCode (which they supply) but they do not want to start at the starting point of the code. I suggest we remove it.

Also side-note: this whole runCode and runCall should at some point be merged. We can refactor EVM in such way that we allow the external method call to `runCall` (or maybe we should rename this, `runEVM` or something) to provide the complete system state. In these, almost all options are optional, but they provide access to the internal run state. This would extend to providing things like current stack, the current pc, etc. (This is an idea I had a long time ago, and seeing the current state of EVM it is not clean and we should clean it up, it is kind of messy)


gabrocheleau (dismissed):
lgtm

holgerd77:
For the `pc` option I think there was a use case when this was introduced (can vaguely remember that someone explicitly requested).

So for these kind of options we should really be careful and/or really dig this out again. Normally no one is just introducing options "for fun" and had some argument why to do. 😜

holgerd77:
On this EVM refactor thing: maybe you want to do another quick round of reflection and estimate here. Is this at the end a rather somewhat small thing to do, in terms of the amount of work? Or is this more extensive? I personally completely can't judge.

So if this is rather small, we might even want to consider to just take this in now (but absolutely not the outcome I am pushing here for, totally in the open, just wanted to at least mention the possibility).

jochem-brouwer:
> For the `pc` option I think there was a use case when this was introduced (can vaguely remember that someone explicitly requested).
> 
> So for these kind of options we should really be careful and/or really dig this out again. Normally no one is just introducing options "for fun" and had some argument why to do. stuck_out_tongue_winking_eye

Alright, I will block this PR and re-introduce the `pc` option.

Regarding the EVM refactor: this will not be a small task. Nevertheless, I will prioritize this.

holgerd77:
> Regarding the EVM refactor: this will not be a small task. Nevertheless, I will prioritize this.

Ah, no, please no big refactorings any more.

jochem-brouwer:
@holgerd77 It's an internal refactor, see discord :)

jochem-brouwer:
(and the side-note of this PR :) )

holgerd77 (approved):
LGTM

PR #2863, opened 2023-07-05
devp2p: type improvements & cleanup

### LINKED ISSUE #2849: Devp2p: Consolidate Types/Interfaces in types.ts
In the `devp2p` package the various exported types and interfaces are still cluttered throughout the library, e.g. `DPTOptions` directly in the `dpt.ts` module.

These should be consolidated and put into a unifed `src/types.ts` file as we have in the other packages.

So this would go for 1) exported types/interaces which are 2) somewhat "important"[tm] (to be judged by implementer) for the user, so this goes for all types/interfaces I guess which are part of the public API (e.g. like `DPTOptions`) and so used as either input or output value for a publicly callable method or constructor or similar.

gabrocheleau:
Resolves #2849 and overall tightens a lot of the typings from devp2p. 

### Todo

- [x] Refactor exported and used types/interfaces/enums to `devp2p/src/types.ts`
- [x] PeerOptions interface
- [x] KBucket types and KBucketOptions interface
- [ ] Improve types for events (for another PR) 
- [x] Go through `any` instances and replace when possible
- [x] Cleanup insteances where debug messages & data were wastefully declared


packages/devp2p/src/ext/kbucket.ts
```diff
@@ -35,12 +35,14 @@ OTHER DEALINGS IN THE SOFTWARE.
...
 
+import type { KBucketOptions } from '../types'
+
 /**
  * @param  {Uint8Array} array1
  * @param  {Uint8Array} array2
  * @return {Boolean}
  */
-function arrayEquals(array1: any, array2: any) {
+function arrayEquals(array1: Uint8Array, array2: Uint8Array): boolean {
```
acolytec3, line 45: Maybe this is a dumb question but how is this used? Is this used when we're trying to verify if two arrays point to the same underlying memory?

packages/devp2p/src/types.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  /**
+   * Function for socket creation
+   *
+   * Default: dgram-created socket
+   */
+  createSocket?: Function
+}
+
+export enum EthProtocol { // What does this represent?
```
acolytec3, line 129: Agree with this comment slightly. The `enum` name is weird given one of the options is `eth`. Would it be bad to rename it to just `Protocol` like we do in `client`?

acolytec3 (commented):
This is looking great so far! Left a couple of questions but great improvements overall.

packages/devp2p/src/types.ts
gabrocheleau: Renaming to protocol would cause conflicts as we already have a Protocol class, but I've renamed it to ProtocolLabel, which I feel is clearer.

packages/devp2p/src/ext/kbucket.ts
gabrocheleau: This just seems like a re-implementation of equalsBytes. Refactored it!

acolytec3: Okay, the code that was there certainly didn't implement true value level equality for arrays, and I'm on mobile so wasn't able to search and see how it's used.

packages/devp2p/src/types.ts
gabrocheleau: We actually also have another interface called Capabilities!
```
export interface Capabilities {
  name: string
  version: number
  length: number
  constructor: ProtocolConstructor
}

``

gabrocheleau:
All right I think this is pretty much done.

I would like to leave the events typings for another PR as this PR is already quite large and I'd prefer not to bloat it further.

holgerd77 (approved):
Thanks a lot, this looks great and moves this library substantially closer to our other libraries, "quality wise" 😋. So that's really nice, typing for devp2p in particular is really important since devp2p is so very unforgiving!

One small naming nit: I perceive `ProtocolLabel` as targeting a somewhat a bit (too) superficial level, I would find `ProtocolName` or `ProtocolType` a bit more fitting, indicating more that "this is the protocol itself".

I've now directly pushed this, taken `ProtocolType` since we have this `*Type` naming already on various occasions in the monorepo, so this is also somewhat fitting from this side.

Hope you are ok with this change, admittedly this stays somewhat subjective!

Will approve and merge once CI passes again.

PR #2864, opened 2023-07-05
monorepo: fix tsc errors

gabrocheleau:
This PR fixes several errors that were showing up when running `npm run tsc --workspaces`. Likely a side-effect of some of the recent changes.


acolytec3 (approved):
Solid fixes. I was wondering about some of these.

PR #2866, opened 2023-07-06
evm: rename EVMOpts -> EVMCreateOpts

jochem-brouwer:
Note: static `create` will be removed if we remove mcl-wasm

### DIFF
--- a/packages/evm/src/evm.ts
+++ b/packages/evm/src/evm.ts
@@ -59,7 +59,7 @@ if (isBrowser() === false) {
 /**
  * Options for instantiating a {@link EVM}.
  */
-export interface EVMOpts {
+export interface EVMCreateOpts {
   /**
    * Use a {@link Common} instance for EVM instantiation.
    *
@@ -203,7 +203,7 @@ export class EVM {
 
   protected _precompiles!: Map<string, PrecompileFunc>
 
-  protected readonly _optsCached: EVMOpts
+  protected readonly _optsCached: EVMCreateOpts
 
   public get precompiles() {
     return this._precompiles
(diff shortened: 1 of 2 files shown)


gabrocheleau:
A few client tests are failing, not sure why the renaming would change anything, re-running to ensure it's not a false-failure...

gabrocheleau (approved):

gabrocheleau (approved):

gabrocheleau:
> A few client tests are failing, not sure why the renaming would change anything, re-running to ensure it's not a false-failure...

That fixed it, lgtm

packages/evm/src/evm.ts
```diff
@@ -239,13 +239,13 @@ export class EVM {
...
    */
-  static async create(opts: EVMOpts = {}): Promise<EVM> {
+  static async create(opts: EVMCreateOpts = {}): Promise<EVM> {
     const evm = new this(opts)
     await evm.init()
     return evm
   }
 
-  constructor(opts: EVMOpts) {
+  constructor(opts: EVMCreateOpts) {
```
holgerd77, line 248: This option should for *not* be renamed. 😜

Needs to be (half-)reverted.

jochem-brouwer: I don't think I follow, why should this not be renamed? Should I introduce `EVMOpts = EVMCreateOpts`? :thinking:

holgerd77: Because this is the other data structure for which we started this renaming in the first place???

So to recap: you named the thing in types.ts EVMOpts, which was the same name as the general options which we have here for the constructor. For that reason the name in types.ts should be named to something else. Am I missing something?

jochem-brouwer: Oh right, I completely missed this point. Note that the data structure itself has not changed (we do not import `EVMCreateOpts` from `types.ts`). I have addressed this in #2866

holgerd77: Thanks, I was getting a bit desparte. 😂

PR #2868, opened 2023-07-08
evm/types: rename EVMCreateOpts -> EVMRunOpts

jochem-brouwer:
The original PR #2861 changed `EVMOpts` -> `EVMCreateOpts` (in two places, for both the constructor EVM as the argument type for `runCall` and `runCode`)

Now, the constructor argument for EVM is `EVMCreateOpts`. The runtime arguments are now `EVMRunCallOpts` and `EVMRunCreateOpts`, which have a base type which it extends: `EVMRunOpts`.

@holgerd77 If wanted I can also revert the `EVMCreateOpts` (so EVM constructor argument type) back to `EVMOpts`. (But, I think this type name is slightly more clear)

### DIFF
--- a/packages/evm/src/evm.ts
+++ b/packages/evm/src/evm.ts
@@ -59,7 +59,7 @@ if (isBrowser() === false) {
 /**
  * Options for instantiating a {@link EVM}.
  */
-export interface EVMCreateOpts {
+export interface EVMOpts {
   /**
    * Use a {@link Common} instance for EVM instantiation.
    *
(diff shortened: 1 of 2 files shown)


ScottyPoi (dismissed):
Nice change.  :+1:

holgerd77 (approved):
Ah, sorry, yes, I changed `EVMCreateOpts` back to `EVMOpts`, I have a relatively strong opinion here. We use this naming scheme `[LIBRARY_NAME]Opts` for the constructor options for all other libraries in the monorepo and for my taste to change here would need a stronger justification than something being slightly cleaner or the like. 🙂 

Ok, will approve and merge otherwise.

jochem-brouwer:
Ok, thanks for the quick convert, had no strong opinion here so the revert is fine :smile: Thanks for the merge :pray:

Issue #2867, opened 2023-07-07
Reintroduce EVMInterface


holgerd77:
@jochem-brouwer So we decided to re-introduce, right?

jochem-brouwer:
Not sure here, I don't know what would be the use case of it, I have not yet seen anyone who wishes to use "their own" EVM instead of ours? :thinking:

holgerd77:
That's not so much for people having their completely own EVM implementation but rather for users who maintain a fork of our library and pass this to the `VM` with the `evm` option. The moment they add a custom property for this or change some (non-called but publicly exposed) method name, things will break for them.

This can be avoided by keeping the interface (which would be an easy win).

Actually this is more or less the same argumentation we have for Blockchain or StateManager, where people (Hardhat e.g.) have their own modfied forks.

Side note: have you seen the comment here https://github.com/ethereumjs/ethereumjs-monorepo/pull/2866#discussion_r1254872876 ?

This would need a fix! 😆

jochem-brouwer:
Ok, will re-introduce the interface :+1: :smile:

jochem-brouwer:
I will create a PR tomorrow or Monday for this interface. However, the `EVMJournal` is now deeply integrated (`evm/journal.ts`) in EVM + VM. So this would add extra items to the EVM interface (it is not as shallow as it was in the past with only a few methods). Note: the original journal calls didn't go to EVM, they went to StateManager/EEI directly, but since we removed EEI now these methods have moved into EVM.

jochem-brouwer:
Closed by #2869

PR #2869, opened 2023-07-09
evm/vm: reintroduce evm interface

jochem-brouwer:
Addresses #2867 

Will await an approval that this is OK before continuing.

Should we also introduce to use this interface in client?


holgerd77:
So, just to give some feedback here: for me this looks pretty healthy, generally make things a lot more transparent to see at one place what is actually exposed by the EVM (and used). And e.g. the journal then *is* used anyhow - being it implicit or named/stated in the interface - that's just how things are implemented right now and we just also do not want to change any more at this point (right?).

So, tldr;: I would be a fan of adding this here.

jochem-brouwer:
> So, just to give some feedback here: for me this looks pretty healthy, generally make things a lot more transparent to see at one unifed place what is actually exposed by the EVM (and used). And e.g. the journal then _is_ used anyhow - being it implicit or named/stated in the interface - that's just how things are implemented right now and we just also do not want to change any more at this point (right?).
> 
> So, tldr;: I would be a fan of adding this here.

Ok, I wanted to just get this approval in order to continue, thanks :smile: 

Will finish today!

holgerd77:
Updated this via UI

holgerd77:
Client tests failing here.

packages/vm/src/vm.ts
```diff
@@ -20,6 +20,7 @@ import type {
 } from './types.js'
 import type { BlockchainInterface } from '@ethereumjs/blockchain'
 import type { EVMStateManagerInterface } from '@ethereumjs/common'
+import type { EVMInterface } from '@ethereumjs/evm/dist/cjs/types.js'
```
holgerd77, line 23: These imports are still a total no-go for merging, I guess that's obvious?

jochem-brouwer: Huh, how did this get introduced? :thinking: I think VSCode auto-imported this

jochem-brouwer:
When I got the approval I saw this comment on this PR (see first message): 

> Will await an approval that this is OK before continuing.

I was a bit confused why I had this comment, but now if I review my own PR I see many TODOs. This is also why this PR looks dirty (CC @holgerd77 )

packages/vm/src/vm.ts
jochem-brouwer: Nvm see my new comment on the PR, PR was not yet ready to be reviewed (sorry for marking it as ready-to-review)

jochem-brouwer:
Have changed a method name (`reportAccessList` -> `startReportingAccessList`) in EVM journal interface since the name was misleading (not super related to this PR but this is an improvement IMO)

packages/vm/src/vm.ts
```diff
@@ -231,7 +232,7 @@ export class VM {
       blockchain,
       stateManager,
     }
-    const evmCopy = new EVM(evmOpts)
+    const evmCopy = new EVM(evmOpts) // TODO fixme (should copy the EVMInterface, not default EVM)
```
gabrocheleau, line 235: Imo we could do this in this PR by just adding a `copy` method to the EVMInterface, and implemting it in EVM in the way you've done here. Then replace this line with `evmCopy = evm.copy()`. What do you think?

packages/evm/src/types.ts
```diff
@@ -134,6 +136,26 @@ export type EVMEvents = {
...
+    deleteAccount(address: Address): Promise<void>
+    accessList?: Map<string, Set<string>>
+    addAlwaysWarmAddress(address: string, addToAccessList?: boolean): void
+    addAlwaysWarmSlot(address: string, slot: string, addToAccessList?: boolean): void
+    startReportingAccessList(): void
+  }
+  stateManager: EVMStateManagerInterface
+  precompiles: Map<string, PrecompileFunc>
+  runCall(opts: EVMRunCallOpts): Promise<EVMResult>
+  events?: AsyncEventEmitter<EVMEvents>
```
gabrocheleau, line 156: Why are we making events optional? Could there be a valid EVM implementation that does not emit events?

gabrocheleau (commented):
Looks good, left 2 comments.

packages/evm/src/types.ts
jochem-brouwer: It is not super-necessary to make EVM run. If we make this non-optional we need to have "external" EVMs import our AsyncEventEmitter and implement the events it defines

packages/vm/src/vm.ts
jochem-brouwer: I think we should keep this as a TODO for now (?)

gabrocheleau: Ah sure yes, I thought this was a simple implementation but since it's more complicated than that let's leave for another PR :)

packages/vm/test/tester/runners/GeneralStateTestsRunner.ts
```diff
@@ -129,7 +129,7 @@ async function runTestCase(options: any, testData: any, t: tape.Test) {
       const block = makeBlockFromEnv(testData.env, { common })
 
       if (options.jsontrace === true) {
-        vm.evm.events.on('step', stepHandler)
+        vm.evm.events!.on('step', stepHandler)
```
gabrocheleau, line 132: Would it make sense in these two instances to check if `vm.evm.events` is defined, and to throw if not?

jochem-brouwer: No, in tests we are sure we use default EVM, so we explicitly are sure that we use `EVM` and thus the events are defined :smile:

gabrocheleau: Ok sure!

gabrocheleau (approved):

holgerd77:
Have updated this via UI

holgerd77:
Maybe just to throw in the ring: do we want to keep this interface on EVM or also move to Common?

Side note: in case we move to Common I would be a fan of a slightly different structure there and do folder "interfaces/" and then have a dedicated named file for every major interface. This just scales better and is more transparent and very well worth to have an own directory for this, these interfaces now play a very central role in our monoprepo and can very well get a visible place.

jochem-brouwer:
I indeed thought of moving this to Common and I think regarding cleani-ness this is great :smile:

holgerd77:
Ok, can you maybe tackle directly today or tomorrow? Would you also be ok with this slightly adopted folder structure (and could you integrate as well)? So I would then name the files after the packages for simplicity: so e.g. `interfaces/evm.ts`.

PR #2870, opened 2023-07-10
Remove BLS EIP 2537 (remove mcl-wasm package)

### LINKED ISSUE #2858: Remove mcl-wasm
Also remove EVM.create()

jochem-brouwer:
Closes #2858


packages/evm/src/evm.ts
```diff
@@ -234,18 +215,7 @@ export class EVM {
...
-   * @param opts EVM engine constructor options
-   */
-  static async create(opts: EVMOpts = {}): Promise<EVM> {
-    const evm = new this(opts)
-    await evm.init()
-    return evm
-  }
-
-  constructor(opts: EVMOpts) {
+  constructor(opts: EVMOpts = {}) {
```
holgerd77, line 218: This is a bit of a side discussion, but since I am seeing this change/taking-over here:

Is there any difference in semantics between doing `opts: EVMOpts = {}` and `opts?: EVMOpts`?

I can vaguely remember a discussion around this, but absolutely not the outcome any more.

Atm we use this a bit mixed - especially for the `opts` part. Would be interesting to recall and see what benefits/drawbacks/differences are (if there are any).

holgerd77 (approved):
Wild mixture of sadness and hapiness to see all this code removed! 😅 🙂 😜

Thanks for adressing so quickly. Will merge.

packages/evm/src/evm.ts
jochem-brouwer: I think the difference is that if you use `opts?` it is undefined by default, while now it is `{}` by default. (I had to add this because some tests did not pass a param, and therefore if you read `opts.common` (opts is thus `undefined` if you do not default it to `{}`) it throws.

jochem-brouwer:
I have pushed to `monorepo-with-bls` the commit before this PR which thus has monorepo state with BLS integrated in it.

packages/evm/src/evm.ts
holgerd77: Hmm, yeah, makes sense. Maybe we want to unify this before the releases (to default to `{}` for `opts` also for other libraries).

PR #2875, opened 2023-07-11
Ethereum cryptography version fix

puneettiwariofficial:
removed carrot symbol, as new version not compatible.

Screenshot 2023-07-11 at 3 38 58 PM


holgerd77:
Hi there,
I am pretty sure there will be hotfix releases soon (in the next 1-2 days), we would then directly jump onto the new releases. Will close here.

PR #2879, opened 2023-07-11
Use version 2.1.1 for ethereum-cryptography

scorbajio:


holgerd77:
Hi there, not sure if you misread, but 2.1 is actually the broken version. We need to wait for a hot fix release here.

scorbajio:
> 

I indeed misread.

holgerd77:
2.1.1 is released, this can (and should, since we need for releases) now be updated quickly.

jochem-brouwer (approved):
Lgtm!

holgerd77:
Please let's not start pinning dependencies now. This just does happen occasionally. But actually it is not happening that often and on the other hand it can happen on any of the other 50 or so dependencies we have and generally we have good reasons not to pin.

E.g. for this library there is already a 2.1.2 release out (we should update 😂)

PR #2880, opened 2023-07-12
client: improve new payload and fcu block executions

g11tech:
fix and optimize the block execution for engine's new payload and fcUs, supersedes https://github.com/ethereumjs/ethereumjs-monorepo/pull/2787

- chain head was being used for getting `vmHead` instead of the actual `vm` head
- optimize and correct checks if the parent(s) have been executed (track and use `executedBlocks` ) 
- recursively find parents would only find just 1 parent, `vmHead` hash should have been used as its used to find parents till `vmHead`
- add `executedBlocks` map in engine for easy figuring out what has been executed (rather than checking `stateRoot` in state manager because a bad block can be formed with previously executed `stateRoot` which would of course be declared valid) ,
- and better pruning of `executedBlocks` and `remoteBlocks` based on `finalized` and `vm` head
- skip block executions if too many blocks (including parents) are pending to be executed for new payload or if the `VmExecution` s already busy - and return optimistic `SYNCING` response which would be resolved later to valids or invalids
- make sure that `running` flag is switched to false even if `run(..)` of `VmExecution` fails

A branch of this (devnet 7 compatible) seems to be working fine on devnet-7

### DIFF
--- a/packages/client/src/execution/vmexecution.ts
+++ b/packages/client/src/execution/vmexecution.ts
@@ -153,13 +153,22 @@ export class VMExecution extends Execution {
    * It relies on the additional {@link VMExecution.setHead} call to finalize
    * the entire procedure.
    * @param receipts If we built this block, pass the receipts to not need to run the block again
+   * @param optional param if runWithoutSetHead should block for execution
+   * @returns if the block was executed or not, throws on block execution failure
    */
-  async runWithoutSetHead(opts: RunBlockOpts, receipts?: TxReceipt[]): Promise<boolean> {
-    if (this.running || !this.started || this.config.shutdown) return false
-    this.running = true
+  async runWithoutSetHead(
+    opts: RunBlockOpts,
+    receipts?: TxReceipt[],
+    blocking: boolean = false
+  ): Promise<boolean> {
+    // if its not blocking request then return early if its already running else wait to grab the lock
+    if ((!blocking && this.running) || !this.started || this.config.shutdown) return false
 
-    try {
-      await this.runWithLock<void>(async () => {
+    await this.runWithLock<void>(async () => {
+      try {
+        // running should be false here because running is always changed inside the lock and switched
+        // to false before the lock is released
+        this.running = true
         const { block, root } = opts
 
         if (receipts === undefined) {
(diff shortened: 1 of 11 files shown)


packages/client/src/execution/vmexecution.ts
```diff
@@ -154,38 +154,46 @@ export class VMExecution extends Execution {
...
-        // Check if we need to pass flag to clear statemanager cache or not
-        const prevVMStateRoot = await this.vm.stateManager.getStateRoot()
-        // If root is not provided its mean to be run on the same set state
-        const parentState = root ?? prevVMStateRoot
-        const clearCache = !equalsBytes(prevVMStateRoot, parentState)
+  async runWithoutSetHead(opts: RunBlockOpts, receipts?: TxReceipt[]): Promise<boolean> {
+    if (this.running || !this.started || this.config.shutdown) return false
+    this.running = true
 
-        const result = await this.vm.runBlock({ clearCache, ...opts })
```
g11tech, line 168: the diff here and below is because of the putting the code inside try catch

packages/client/src/execution/vmexecution.ts
```diff
@@ -271,168 +279,170 @@ export class VMExecution extends Execution {
     if (this.running || !this.started || this.config.shutdown) return 0
     this.running = true
 
-    // run inside a lock so as to not be entangle with runWithoutSetHead or setHead
-    return this.runWithLock<number>(async () => {
```
g11tech, line 275: the diff here is because of the try catch added so that `this.running` can be turned to `false` in the `finally` section

holgerd77:
Wanted to review, is this ready? 🙂

holgerd77:
(and could we merge on positive review?)

packages/client/src/execution/vmexecution.ts
holgerd77: 👍

holgerd77 (approved):
Did a somewhat rough review commit-by-commit, I think this is very much the type of work though which strongly „proves itself“ just by using it in practical scenarios without too much side effects.

Will approve and merge.

PR #2885, opened 2023-07-13
Move EVMOpts to types in EVM

### LINKED ISSUE #2883: EVM: Move EVMOpts from evm.ts -> types.ts
I think we should really still do this short-term, this is such a stupid runaway from how this is placed in all other libraries.

Stumble upon it each time I open the file.

jochem-brouwer:
Closes #2883

Also fixes docs building in EVM (tests now comply with TS rules)


holgerd77 (approved):
Nice, thanks! 🙂

PR #2886, opened 2023-07-13
VM/genesis: Fix small genesis API inconsistency

holgerd77:
This PR fixes a small inconsistency with the genesis package API I discovered while writing release notes. So the example I added as a test now would break before, because the `getGenesis()` API would give either `GenesisState` or `null` while the other `genesisState` input APIs in VM and Blockchain would take `GenesisState` or `undefined`.

This didn't play well together on the type level and so the example would break.

PR fixes this.


acolytec3 (approved):
Clean. Lgtm

g11tech (approved):
lgtm

PR #2887, opened 2023-07-13
Fix the typedoc scripts across the monorepo

acolytec3:
Fixes the typedoc config across the monorepo


holgerd77:
Nice, thanks! 🙂

Issue #2872, opened 2023-07-10
Update `devp2p` API methods


acolytec3:
We should upgrade `devp2p`s' API to match changes done in the other packages

- [ ] Change `_common` to `common` on any/all classes exposed by Devp2p (notably `RLPx`)
- [ ] Change any classes that extend the `EventEmitter` class to have an internal `events` property instead so the main class API isn't cluttered up with `EventEmitter` properties

holgerd77:
One note: this is very well breaking though, so this should be done relatively imminently after RC1 so that we can get it into the final releases.

holgerd77:
Hi Amir,
so this is the respective issue. 🙂 

So to expand a bit on the initial description: the `_common` to `common` change (rename) follows the example from #2857 and should be pretty straight forward I guess.

Update: I am now looking a bit into this. For `devp2p` we have actually *no* external accesses of `_common` at all. So I would think it makes sense to switch here and keep this as `_common` for now, no necessity to artificially expose when not needed.

I would in turn add the following to the issue: in `devp2p` we have all these `_` properties, like e.g. `_privateKey: Uint8Array`. Also following #2857 these should be made `protected` (so somewhat the same as `private`, but `protected` gives more flexibility on subclassing and the like).

This can very likely not be done in just one go but will likely need some one-by-one approach (or 2-4 at once or so) and then re-running tests (or better: CI with client) re-occuringly, otherwise one runs the risk to "guard" too many properties at once and have things breaking on various fronts since properties are accessed anyhow.

If properties are *only* accessed by tests please do *not* make public but rather access by `any`-cast or something (there was this alternative approach doing e.g. `rlpx['_privateKey']` which circumvents the `protected` keyword and keeps the typing, so maybe even better).

For the EventEmitter thing (might make sense to do in a separte PR now with the above task expanded):

So there is also an example in #2857 as well on this.

For `devp2p` there are seven classes which at the moment inherit from `EventEmitter` directly, so e.g. `export class RLPx extends EventEmitter`.

These classes should instead get an additional (public) property `events` which instantiates an `EventEmitter` and the events should be listened to from this property instead of the class object itself.

So `rlpx.on('event', ...)` changes to `rlpx.events.on('event', ...)`.

Likely also good to do a one-by-one approach here, there is a lot of event-listening in the client and there things might break otherwise. Or minimally client tests should be run every now and then I guess.

PR #2889, opened 2023-07-13
Update devp2p API names and access specifiers

scorbajio:
This PR addresses part of #2872 by renaming and rethinking access specifiers for properties of devp2p package classes.


holgerd77 (commented):
Super cool, going in a good direction 👍 and this will so dramatically improve the API (e.g. for code completion) of this package 🙏, phew, wasn’t aware that there are SO many properties floating around. 😅

packages/devp2p/src/dns/dns.ts
```diff
@@ -21,8 +21,8 @@ type SearchContext = {
 }
 
 export class DNS {
-  private _DNSTreeCache: { [key: string]: string }
-  private readonly _errorTolerance: number = 10
+  protected _DNSTreeCache: { [key: string]: string }
+  protected readonly _errorTolerance: number = 10
```
holgerd77, line 25: 👍

packages/devp2p/src/dpt/ban-list.ts
```diff
@@ -14,19 +14,19 @@ const debug = createDebugLogger('devp2p:dpt:ban-list')
 const verbose = createDebugLogger('verbose').enabled
 
 export class BanList {
-  private lru: LRUCache<string, boolean>
+  private _lru: LRUCache<string, boolean>
```
holgerd77, line 17: 👍

packages/devp2p/src/dpt/server.ts
```diff
@@ -22,14 +22,14 @@ const verbose = createDebugLogger('verbose').enabled
...
-  _socket: DgramSocket | null
-  _debug: Debugger
+  protected _dpt: DPT
+  protected _privateKey: Uint8Array
+  protected _timeout: number
+  protected _endpoint: PeerInfo
+  protected _requests: Map<string, any>
+  protected _requestsCache: LRUCache<string, Promise<any>>
+  protected _socket: DgramSocket | null
+  private _debug: Debugger
```
holgerd77, line 32: Likely not practically relevant and only a small nit, but I would think that it would make sense to just not distinguish between "even more internal" or so variables and just also make `_debug` protected, maybe someone wants to customize the devp2p package and add own debug messages, whatever.

holgerd77 (approved):
🙏 🙏 🙏

Absolutely love this! 😍

Super clean and consistent, thanks a lot for this straight forward tackling!

Will merge.

packages/devp2p/src/dpt/server.ts
scorbajio: I was on the fence about this one, but since debuggers are usually initialized specific to a class, I opted for private.

holgerd77: The `protected` version is also remaining in the class scope, so this is just a bit broader so that people can e.g. subclass the `Server` class, e.g. for a custom binance devp2p implementation (super random example) or whatever. And then on the subclass one might want to adjust some properties and it is just a bit unlucky if these are then not accessible. So at some point in time [tm] we generally decided to favor `protected` over `private`, since it (more or less) achieves the same thing and has fewer side effects.

PR #2893, opened 2023-07-16
Devp2p EventEmitter refactor

scorbajio:
This PR addresses part of https://github.com/ethereumjs/ethereumjs-monorepo/issues/2872 by changing the classes in the devp2p package that extend the `EventEmitter` class to rather keep the emitter in a class field. This PR builds off of #2889 and should be merged after #2889 has been merged.


holgerd77:
Ok, thanks a lot for the other PR 👍, RC releases tomorrow (Tuesday), I might or might not be able to still take this one in as well. Definitely no stress though, as you make it. Otherwise we'll just take this into the final breaking releases.

packages/client/test/net/peer/rlpxpeer.spec.ts
```diff
@@ -89,9 +89,9 @@ tape('[RlpxPeer]', async (t) => {
     peer.config.events.on(Event.PEER_DISCONNECTED, (rlpxPeer) =>
       t.equals(rlpxPeer.pooled, false, 'got disconnected')
     )
-    peer.rlpx!.emit('peer:error', rlpxPeer, new Error('err0'))
-    peer.rlpx!.emit('peer:added', rlpxPeer)
-    peer.rlpx!.emit('peer:removed', rlpxPeer, 'reason')
+    peer.rlpx!.events.emit('peer:error', rlpxPeer, new Error('err0'))
```
acolytec3, line 92: I think the tests in this file are failing because the mock `devp2p` peer the test is connecting to doesn't have the `events` property on it.  So, you'd need to adjust the properties of the `RLPx` class that starts on line 11 to resolve I think.  That said, I say we correct this once we merge the `vitest` PR so we don't have to spend lots of time rewriting these tests later.

acolytec3 (approved):
LGTM!

holgerd77:
To-tal-ly love this!!! 🙂👍

Thanks a lot!

Issue #2865, opened 2023-07-06
Node 18/20 Nightly CI Tests failing


holgerd77:
Currently the Node 20 tests from the respective [node-versions.yml](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/.github/workflows/node-versions.yml) workflow running nightly are [failing](https://github.com/ethereumjs/ethereumjs-monorepo/actions/runs/5470665981/jobs/9961007079) with:

![grafik](https://github.com/ethereumjs/ethereumjs-monorepo/assets/931137/2265c68d-8f9b-4f8f-88e6-e5484f53b683)

This should be examined and fixed. 🙂 

Procedural note: it works best to temporarily add the workflow to the PR run, otherwise it is really hard to decide (or rather: not really possible) if things are really working (or if someone has an alternative (better) approach, please drop a note below).

jochem-brouwer:
(should we add a prio urgent here?)

holgerd77:
@jochem-brouwer seems we *need* to add a prio urgent here, otherwise noone is picking this up, yes. 😂

And - yes - this should really be adressed before we do the releases. Making sure that our libraries run with the current latest Node versions is something we should at least substantially *try* to achieve. Sometimes there are blockers but this might in fact be a small thing.

Have no idea though. Needs some investigation.

holgerd77:
So, and - yes - CI still failing, [here](https://github.com/ethereumjs/ethereumjs-monorepo/actions/runs/5503134068) is the latest run.

jochem-brouwer:
This is weird, I thought I addressed this, and did a search:

https://github.com/ethereumjs/ethereumjs-monorepo/issues/2829

https://github.com/ethereumjs/ethereumjs-monorepo/pull/2839

So something re-broke?

holgerd77:
This issue is from 4 days ago. 😋

holgerd77:
(so, yes, something likely re-broke, but which is really not unlikely, we just did so many changes in the last 2 weeks)

scorbajio:
Should keep an eye out on tests to see if relaxing timing constraints results in less time outs on test runs.

holgerd77:
Closed by #2928

PR #2895, opened 2023-07-18
Troubleshoot failing node versions tests: Adjust timeouts in devp2p tests to avoid race conditions

scorbajio:
While investigating #2865, after making a busy wait and timeout longer, the devp2p tests would pass more consistently than before when running them locally. This will give more consistent results for test runs.


holgerd77:
Rebased this via UI

gabrocheleau:
Any insight as to why tests would run slower on Node20 vs Node18?

acolytec3:
Can you rename the PR to represent what it's actually doing?  Something like "Adjust timeouts in devp2p tests to avoid race conditions"

acolytec3 (approved):
LGTM

scorbajio:
> Any insight as to why tests would run slower on Node20 vs Node18? Seems surprising.

It seems to be unrelated to versions, but rather related to the vitest refactors. Tests in devp2p and client packages have shown to time out intermittently. Relaxing timing constraints keeps tests from timing out, for now.

PR #2898, opened 2023-07-18
Resolve access errors from doc build process

scorbajio:


packages/devp2p/test/integration/rlpx-simulator.spec.ts
```diff
@@ -34,10 +34,13 @@ describe('RLPx simulator tests', () => {
         tcpPort: null,
       }
       assert.notOk(
+        // @ts-ignore
         rlpxs[0]._dpt!.banlist.has(peer),
```
acolytec3, line 38:
```suggestion
    rlpxs[0]['_dpt']!['_banlist'].has(peer),
```

Instead of putting `ts-ignore` in there, you can reference these protected members using bracket notation like above.

PR #2899, opened 2023-07-18
docker: catch signals like SIGTERM by running node via dumb-init

skylenet:
### Problem:

When a container is stopped, the node process isn't terminated like it should. The reason for that is that signals aren't handled well by node. 

## Solution:

[dumb-init](https://github.com/Yelp/dumb-init) enables you to simply prefix your command with dumb-init. It acts as PID 1 and immediately spawns your command as a child process, taking care to properly handle and forward signals as they are received.

### Reproducing: 

Without the fix, run `docker run --rm -it ethereumjs`. You'll see ethereumjs starting. Now press CTRL+C to send a SIGTERM. You'll notice that it won't work. 

Build the image now with the fix and retry. Pressing CTRL+C will send the SIGTERM which will be propagated correctly now via dumb-init and the ethereumjs process will terminate gracefully.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,7 @@ RUN npm install @ethereumjs/client@$VERSION
 
 FROM node:18-alpine
 WORKDIR /usr/app
+RUN apk --no-cache add dumb-init
 COPY --from=build /usr/app .
 
 # Sanity check
@@ -18,4 +19,4 @@ RUN node /usr/app/node_modules/.bin/ethereumjs --help
 # since memory may spike during certain network conditions.
 ENV NODE_OPTIONS=--max_old_space_size=6144
 
-ENTRYPOINT ["node", "/usr/app/node_modules/.bin/ethereumjs"]
+ENTRYPOINT ["dumb-init", "node", "/usr/app/node_modules/.bin/ethereumjs"]
--- a/Dockerfile.fromSource
+++ b/Dockerfile.fromSource
@@ -11,6 +11,7 @@ RUN npm i
 
 FROM node:18-alpine
 WORKDIR /usr/app
+RUN apk --no-cache add dumb-init
 COPY --from=build /usr/app .
 
 # Sanity check
@@ -23,4 +24,4 @@ RUN node /usr/app/node_modules/.bin/ethereumjs --help
 # since memory may spike during certain network conditions.
 ENV NODE_OPTIONS=--max_old_space_size=6144
 
-ENTRYPOINT ["node", "/usr/app/node_modules/.bin/ethereumjs"]
+ENTRYPOINT ["dumb-init", "node", "/usr/app/node_modules/.bin/ethereumjs"]


acolytec3:
Can you advise how you are running this image to recreate this behavior?  I built our Docker image and then ran it and did not observe `ctrl+c` being disregarded by the container.  It appeared to shutdown as expected.

skylenet:
Right. Maybe the CTRL+C example isn't working as expected but try the following:

Building the image with the fix:

```sh
➜  ethereumjs-monorepo git:(skylenet/fix-signals-docker) $ docker build -t ethereumjs:fix-signals .
```

Building the image without the fix (based on master)

```sh
➜  ethereumjs-monorepo git:(skylenet/fix-signals-docker) $ git checkout master
➜  ethereumjs-monorepo git:(master) $ docker build -t ethereumjs:master .
```


Have 2 tabs open, one where we run the container and another one where we'll trigger some signals:

On tab 1:

```sh
docker run --rm --name ethereumjs -it ethereumjs:fix-signals
```

On tab 2:

```sh
docker kill --signal=SIGTERM ethereumjs
# Alternatively you can also just run the kill command on PID 1 from within the container (Which by default sends the SIGTERM signal):
docker exec -it ethereumjs kill 1
```

You'll notice that the process will terminate. 

Now repeat the same steps, but use the `ethereumjs:master` image instead. You should notice that nothing happens. 

---

Also here some recommendations from the official best practices on [how to deal with signals on node](https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#handling-kernel-signals): 
 > Node.js was not designed to run as PID 1 which leads to unexpected behaviour when running inside of Docker.

skylenet:
Another problem: While going through this exercise I've noticed that it doesn't terminate gracefully when sending a SIGTERM, which should be fixed on a separate PR. It does handle well the SIGINT case. 

```sh
docker run --rm --name ethereumjs -it ethereumjs:fix-signals
```

Send a SIGINT

```sh
## Sending a SIGINT (What CTRL+C does): 
docker kill --signal=SIGINT ethereumjs
 
## Logs:
[07-21|06:44:08] INFO Caught interrupt signal. Obtaining client handle for clean shutdown...
[07-21|06:44:08] INFO (This might take a little longer if client not yet fully started)
[07-21|06:44:08] INFO Shutting down the client and the servers...
[07-21|06:44:16] INFO Stopped synchronization.
[07-21|06:44:16] INFO Stopped execution.
[07-21|06:44:16] INFO Stopped eth service.
[07-21|06:44:16] INFO Stopped rlpx server.
[07-21|06:44:16] INFO Exiting.
```

If you try to do the same using **SIGTERM**, the process will just terminate without logging those lines.

```sh
## Sending a SIGINT (What CTRL+C does): 
docker kill --signal=SIGTERM ethereumjs
 
## Logs:
(Nothing, container just exists)
```

acolytec3:
Thanks for the feedback.  I have been now able to recreate the issue you described.  I can certainly add handling for the SIGTERM in our client and already have it coded up in another PR I'm working on.  That said, from reading the docs you linked, I don't think we need this `dumb-init` dependency and can just use the `--init` parameter when running the docker image and it will accomplish the same thing with out adding in new dependencies.  I'll push the other PR shortly and link here so you can review and see if you concur or not.

acolytec3:
@skylenet See #2921 and let me know if this addresses your concerns.

holgerd77:
Will cautiously close here, @skylenet a final (?) feedback note would still be appreciated! 🙂

skylenet:
Sorry for the delay on the response @holgerd77 @acolytec3 , I was offline for some time. 

The fix in #2921 is great and definitely needed 🎉  
 
Regarding the solution of delegating the process wrapping to the `--init`flag, I'm not sure.... 
I do understand that you don't want to add one additional dependency and prefer to delegate the responsibility of that to the container runtime. But from my side, I now have to additionally check how kubernetes (our main container orchestration platform) is dealing with this. There seems to be a big conversation ongoing in https://github.com/kubernetes/kubernetes/issues/84210. I'll be testing a bit around on kubernetes and see how it behaves. There seems to be a "workaround" to use ` shareProcessNamespace: true` for the kubernetes POD. But that's doing way more things than what dumb-init was doing. 

IMO the application should be responsible for dealing with the termination signals and not using some additional flag/feature that "might" be built into some container runtimes. This is also something that the [12 Factor App Manifesto] regarding [Disposability](https://12factor.net/disposability) says:

> Processes shut down gracefully when they receive a [SIGTERM](http://en.wikipedia.org/wiki/SIGTERM) signal from the process manager.

acolytec3:
Makes sense I think.  WIll talk a little further with @g11tech about the best approach and will see if either `tini` or `dumb-init` will work.  Do you have a preference?  Some googling seems to indicate that both are generally suitable though I'd be inclined to use `tini` just because it's already packaged with the docker runtime so theoretically blessed by Docker :tm:

skylenet:
No preference from me. For this use case any of them work.

jochem-brouwer:
@acolytec3 haven't we fixed this currently? Since we now also dont have hive timeouts? (Maybe we should update our local docker package though, to the hive one?)

acolytec3:
I'm not sure if Hive is indicative or not. @skylenet If you build an image from the latest master branch of our repo, do you still see it hang or not shutdown correctly when you send SIGTERM in your environment?

PR #2901, opened 2023-07-19
Adjust client unit test timeouts

### LINKED ISSUE #2902: RPC won't start in client on current master
If you start the client using `npm run client:start -- --rpc` on the current master branch, you will see this error in the logs.
```sh
[07-19|22:01:55] ERROR Error starting client (0 , connect_1.default) is not a function
```

[This line](https://github.com/ethereumjs/ethereumjs-monorepo/blob/c1950acac1518ff15dc128d58253df4de81760b8/packages/client/src/util/rpc.ts#L167) and [this line](https://github.com/ethereumjs/ethereumjs-monorepo/blob/c1950acac1518ff15dc128d58253df4de81760b8/packages/client/src/util/rpc.ts#L168) are now being compiled to an undefined reference in the client so the RPC server doesn't start at all.
 
 The final javascript looks like this:
 ```js
     const app = (0, connect_1.default)();
    if (typeof rpcCors === 'string')
        app.use((0, cors_1.default)({ origin: rpcCors }));
 ```

 If you remove the `.default` in the first and third lines, everything works.  So, we need to rework the default imports for the `connect` and `cors` modules that are imported at the top of this file to be compiled correctly.

acolytec3:
This PR does several things:
- Removes unit test timeouts and replaces with global timeout in `vitest.config.unit.ts`
- Ups the timeouts on a number of integration tests to give more time to pass
- Reverts some namespace imports in `src/util/rpc.ts` to `require` statements that Typescript is compiling in strange ways and causing the RPC to break in the compiled JS code - this will resolve #2902 
- Steramlines the `cli` tests down to one test that runs independently in the client CI and starts the client on a local devnet and verifies that the RPC starts up.  It also implicitly tests PoA block building.


acolytec3:
So vitest doesn't like namespace imports it seems.  For the moment, I'm reverting those two imports to `require` since the `client` module is currently a CJS build only.  If we switch to ESM, I guess we can tackle this then.

gabrocheleau (approved):

holgerd77:
So but to confirm: our client release is currently broken until we release this here, right?

acolytec3:
The RPC doesn't work. We can deal with the `requires` I added if/when we convert the client to ESM.

Issue #2856, opened 2023-07-03
Feature request for WalkController: extracting nodes by level


TimDaub:
To use the ethereumjs/trie library for set reconciliation, we go breadth-first, level by level, take those nodes as an array and then send them to remote node for comparing their hashes; we implement a function descend as follows

https://github.com/attestate/kiwistand/blob/e2c8562fc202a7b5450a1d28e9774b52ca512409/src/store.mjs#L162-L220

For that, we keep track of a decrementing value we call level until it reaches zero, in a naive attempt to figure out how far in the trie we can traverse into. Note that the key length of a node itself cannot be used as a substitute for this, as a PMT in Ethereum features several optimizations e.g. ExtensionNodes, which I don't know how to e.g. handle in terms of their key length.

While recently coming across this while investigating a synchronization bug, I noticed that the level variable won't decrement often enough if the trie has a big branching factor, as this unit test shows.

```js
test("store.descend function should correctly handle high branching factor", async (t) => {
  env.DATA_DIR = "dbtestA";
  const trieA = await store.create();

  // Manually create a trie with a high branching factor
  await trieA.put(Buffer.from("0100", "hex"), Buffer.from("A", "utf8"));
  await trieA.put(Buffer.from("0101", "hex"), Buffer.from("B", "utf8"));
  await trieA.put(Buffer.from("0102", "hex"), Buffer.from("C", "utf8"));
  await trieA.put(Buffer.from("0103", "hex"), Buffer.from("D", "utf8"));
  await trieA.put(Buffer.from("0104", "hex"), Buffer.from("E", "utf8"));
  await trieA.put(Buffer.from("0200", "hex"), Buffer.from("F", "utf8"));

  const root = trieA.root();
  const originalLevel = 2;
  const currentLevel = 2;
  const nodes = await store.descend(trieA, root, originalLevel, currentLevel);

  // Check that all nodes are included in the result
  t.is(nodes.length, 6); // it fails and nodes.length is 2

  await rm("dbtestA", { recursive: true });
});
```

[...]

PR #2904, opened 2023-07-20
Trie - async walk generator

ScottyPoi:
## Trie -- `walkTrieIterable`

 **QuickStart:** 
> from:  `packages/trie`
> run:    `npx vitest test/util/asyncWalk.spec.ts`

observe  console logs to see details of walk.
test includes walk of "Sparse" Trie and comparison to `walk controller` API.

### Intention:
* To establish more user friendly API for walking a Trie
  * And do so without altering the `walk controller` 
* To enable users to walk a Sparse* Trie
  * *Sparse meaning a valid merkle trie built from a proof
  * "Missing" nodes in a sparse trie cause the `walk controller` to throw an Error, and end the walk.
  * So users cannot use the `walk controller` to navigate sparse Tries.

### Feature

This PR introduces a new approach to Trie walking, via `AsyncIterable`

This is a **new** method, independent from the existing `WalkController` class.  

**It does not presume to replace the `WalkController` in any active context at this time.  **
*details on why in comments*

**Main differences from `WalkController`:**

1. Iterator API is more straightforward and intuitive.  User create the async iterator object, and then proceed via `for...await`
1. **Sparse Trie** walking 
1. "CurrentKey" is yielded by the iterator along with each node, instead of just `_nibbles`
1. Inner and Outer `OnFound` functions, combined with a `NodeFilter` parameter enable very customizable walks.

### Async Generator / Async Iterable / Async Iterator
Uses the iteration protocols described in the MDN docs here: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols

### Function

the `_walkTrie` function is exported form `trie/src/util/asyncWalk.ts`, but also internalized as a bound method for the Trie class: `trie.walkTrieIterablle`

Do not fear the **`bind`**

This is not bound to an interface, just to an exported function.  This is so that the new code will not add length to the already very long `trie.ts` file.  Hyperlinks is VSCode, etc. will link to the function implementation iteself with one jump.  You will not have to search for it.  The class method should inherit the typedoc comments as well.

binding the function context to `this`  allows the external function to access private class methods.
For example:
* `trie.hash()` is a private method, that can only be used inside the class itself.
* by binding the `_walkTrie` function to a Trie instance, `_walkTrie` can access `this.hash()`
* This avoids patterns like `(trie as any).hash()` and other hacks

### _walkTrie

`_walkTrie`  generates the `Async Iterable`:

```
export async function* _walkTrie(
  this: Trie,
  nodeHash: Uint8Array = this.root(),
  currentKey: number[] = [],
  onFound: OnFound = async (_trieNode: TrieNode, _key: number[]) => {},
  filter: NodeFilter = async (_trieNode: TrieNode, _key: number[]) => true,
  visited: Set<string> = new Set<string>()
): AsyncIterable<{ node: TrieNode; currentKey: number[] }>
```

### OnFound

This is a callback that will be called on every node encountered on the walk ("pre-filter").
Useful when `NodeFilter` is defined, as it will be called on all nodes, not only the filtered nodes that are yielded.

### NodeFilter

Pass a filter function that will return true for desired nodes.  The walk will actually `yield` only nodes that pass the filter.

### NodeHash / CurrentKey / Visited

With these params left `undefined`, the walk will begin at the root, and later use these parameters to call itself recursively during the walk.  If defined from the start, they can be used to begin the walk from another place.

### Use

`trie: Trie`

```
const walk = trie.walkTrieIterable(trie.root())

for await ( const { node, currentKey } of walk )  {
  // ... do something i.e. console.log( { node, currentKey } )
}
```

### Helpers
Included are 2 examples of helper functions that setup and execute the walk iterator for specific common uses.  These are simply two of many possible wrapper functions that could be designed for common uses.  I have listed a few more ideas below.

**walkAllNodes**
*Walk the Trie from the root, and execute `onFound` on every node in the Trie.*
This will run through the whole walk with default parameters and no filter.

`trie.walkAllNodes(onFound: OnFound) => Promise<void>`

**walkAllValueNodes**
*Walk the Trie from the root, and execute `onFound` on all Trie nodes that store a VALUE*
This will run through the whole walk from the root with a pre-set filter for VALUE_NODES ( LeafNodes and BranchNodes with non-null values )

`trie.walkAllValueNodes(onFound: OnFound) => Promise<void>`

additional filter-based helpers could include:

* allLeafNodes
* allBranchNodes
* allExtensionNodes
* getSubTree

possible `OnFound`-based helpers could include:

* logNodes
* copyNodes
* sortNodes
* garbageCollect
* getAllPaths
* getAllKeys


acolytec3:
>Do not fear the bind 

It's not fear but great dislike. If this behaves as described, I fear no longer.

acolytec3:
Just to make sure I understand how this works, this isn't changing the walk strategy, which is depth first, left to right, correct?  It's more or less the same as the current walk controller, except that it's an async generator that allows the library consumer to trigger the next step, right?

acolytec3 (changes requested):
Looks good so far.  One error message I think needs fixing (see comment)

ScottyPoi:
> Just to make sure I understand how this works, this isn't changing the walk strategy, which is depth first, left to right, correct? It's more or less the same as the current walk controller, except that it's an async generator that allows the library consumer to trigger the next step, right?

yes exactly.  same walk pattern, different API.  
the `walk controller` API can be hard to grasp, and this should be slightly more intuitive to use.

The `walk controller` method is designed to throw an Error, and certain parts of the Trie and other libraries that use the Trie rely on those Errors.   Those expectations run deep, so the walk controller needs to stay as it is.

This way, we can build up a more user friendly Trie walking API without touching the `walk controller`

ScottyPoi:
> Looks good so far. One error message I think needs fixing (see comment)

where is comment?  it's not showing up for me

packages/trie/src/util/asyncWalk.ts
```diff
@@ -0,0 +1,63 @@
...
+      return
+    }
+    visited.add(toHex(this.hash(node!.serialize())))
+    await onFound(node!, currentKey)
+    if (await filter(node!, currentKey)) {
+      yield { node: node!, currentKey }
+    }
+    if (node instanceof BranchNode) {
+      for (const [nibble, childNode] of node._branches.entries()) {
+        if (childNode === null) {
```
ScottyPoi, line 47: - reported as untouched by coverage
TODO: - why does union type for node._branches include null?
      - empty branches are returned as empty Uint8Array, not null.

acolytec3:
> > Looks good so far. One error message I think needs fixing (see comment)
> 
> where is comment? it's not showing up for me

Sorry, was commenting on the wrong PR :facepalm:

acolytec3 (changes requested):
This looks fantastic as an addition.  The one thing I think would be helpful would be to add some of the commentary from the PR to the docs for `trie` so people are aware of this option.  I know we have library consumers who use `trie` as is and this will definitely be helpful for them and we need to surface it so they know it exists.

holgerd77:
Yeah, cool, also a big fan of such API additions to the Trie library! 🤩 Would be great if the originator of this feature request could also have a look before we merge, so this would be @TimDaub and this #2856 issue, right?

holgerd77:
Updated this via UI

ScottyPoi:
> Yeah, cool, also a big fan of such API additions to the Trie library! star_struck Would be great if the originator of this feature request could also have a look before we merge, so this would be @TimDaub and this #2856 issue, right?

Yep, @TimDaub -- This does not explicitly offer the feature that you suggested, but may offer a simpler approach to your use case from #2856

TimDaub:
I‘ll try to take a look in an hour or so

packages/trie/src/util/asyncWalk.ts
```diff
@@ -0,0 +1,60 @@
...
+        const _childNode: Uint8Array =
+          childNode instanceof Uint8Array ? childNode : this.hash(RLP.encode(childNode))
+        yield* _walkTrie.bind(this)(_childNode, nextKey, onFound, filter, visited)
+      }
+    } else if (node instanceof ExtensionNode) {
+      const childNode = node.value()
+      const nextKey = [...currentKey, ...node._nibbles]
+      yield* _walkTrie.bind(this)(childNode, nextKey, onFound, filter, visited)
+    }
+  } catch (e) {
```
TimDaub, line 57: Why apply this try catch on the entire function and not log the error. Don't want to evolve myself into coding standards of ethereumjs too much but IMO a good practice is to do 1 statement per try catch and then match the statement's respective errors or allow errors to bubble up further.

packages/trie/src/trie.ts
```diff
@@ -350,6 +352,37 @@ export class Trie {
...
+  }
+
+  /**
+   * Executes a callback for each value node in the trie.
+   * @param onFound - callback to call when a node is found.
+   * @returns Resolves when finished walking trie.
+   */
+  async walkAllValueNodes(onFound: OnFound): Promise<void> {
+    for await (const { node, currentKey } of this.walkTrieIterable(
+      this.root(),
```
TimDaub, line 375: Is it possible to make this user-definable so that e.g. we could start pass in a branch root on a lower level to skip a bunch of upper-level walking?

packages/trie/src/util/asyncWalk.ts
```diff
@@ -0,0 +1,60 @@
...
+): AsyncIterable<{ node: TrieNode; currentKey: number[] }> {
+  if (equalsBytes(nodeHash, this.EMPTY_TRIE_ROOT)) {
+    return
+  }
+  try {
+    const node = await this.lookupNode(nodeHash)
+    if (node === undefined || visited.has(toHex(this.hash(node!.serialize())))) {
+      return
+    }
+    visited.add(toHex(this.hash(node!.serialize())))
```
TimDaub, line 40: From working with trie.ts, I think it'd be awesome if there were a canonical representation of the trie's and node's hash values and if trie.ts would provide this serialization of values directly instead of having "users" implement their own thing.

packages/trie/src/trie.ts
ScottyPoi: Yes -- this is one example of a method that can be built using `walkTrieIterable`
       
The walk iterator can be set up to start from any node.  However note that the `currentKey` value returned for each node will be rooted from the starting node.  so to string together a node's "full key", you would still need the nibbles from the path to the starting node somehow.

packages/trie/src/util/asyncWalk.ts
ScottyPoi: Really the only potential error here is something like 
"DB: Node Missing".  We want the walk to ignore these errors and keep going, and I would prefer it happened silently.

I don't want the error to throw, because that stops the walk from happening.  I want it to simply continue on.

The reason here is that we want to walk through a sparse trie -- a Trie that has been built from a proof.  
A walk through a trie like this will find a hash on the branch of a proof node for a node that is not in the database, 

said node was not part of the proof, but it's hash was part of the RLP encoding of one of the proof nodes.  So the walk will look for that hash in the DB, and the DB will throw an error.  Which in turn would stop the walk prematurely, even though it is a valid Trie with nodes remaining.

ScottyPoi:
@TimDaub 

Thanks for the feedback.  I responded to a couple comments, but would love to continue the conversation.

Re:
> From working with trie.ts, I think it'd be awesome if there were a canonical representation of the trie's and node's hash values and if trie.ts would provide this serialization of values directly instead of having "users" implement their own thing. (not necessarily actionable in this PR specifically)

Can you be more specific?  This is certainly for a different PR, but if you can help me understand what you mean, I can look into it.

acolytec3 (approved):
Lgtm!

PR #2907, opened 2023-07-20
Fix and streamline client CLI tests

acolytec3:
This fixes and simplifies the current client CLI tests and also updates the CI workflow to run the CLI test only in the `client-test-cli` job and run the unit and integration tests separately.


acolytec3:
A couple of comments here:
1) I removed the `test/integration/cli.spec.ts` tests since this does essentially the same thing as `test/cli/cli-rpc.spec.ts`
2) I removed `cli-sync.spec.ts` as this pings mainnet geth bootndes in an attempt to sync. We agreed previously that this wasn't being a good citizen of the ecosystem by consuming mainnet bootnode bandwidth for our constantly running CI jobs

This does not fix #2902 as vitest seems to be able to resolve the `connect` dependency as is for some reason.

Also, I think it would be good to expand these tests to include starting up 2 clients locally, have one as leader and the other as follower and then have them sync blocks.  That would mimic what the former `cli.sync` test did without needing to go out to the external network.

acolytec3:
Have merged this into #2901

PR #2908, opened 2023-07-20
Monorepo: Post-Build JSON Type Assertion Addition

### LINKED ISSUE #2900: Dealing with ESM requirement for import assertions  JSON imports
On our new:tm: ESM builds, we proudly proclaim we are "fully browser compatible", which may well be true, but I suspect it will break as soon as you try to load one of our modules in browser that also imports JSON..  The issue is that if you try to import the ESM build (at least in NodeJS) without a bundler step that either intentionally or magically handles the import assertion requirement, you get unfriendly errors like this:
```sh
TypeError [ERR_IMPORT_ASSERTION_TYPE_MISSING]: Module "file:///home/jim/development/ultralight/node_modules/@ethereumjs/common/dist/esm/chains/goerli.json" needs an import assertion of type "json"
    at new NodeError (node:internal/errors:387:5)
    at validateAssertions (node:internal/modules/esm/assert:82:15)
    at defaultLoad (node:internal/modules/esm/load:84:3)
    at nextLoad (node:internal/modules/esm/loader:173:28)
    at /home/jim/development/ultralight/node_modules/ts-node/src/esm.ts:255:45
    at async addShortCircuitFlag (/home/jim/development/ultralight/node_modules/ts-node/src/esm.ts:409:15)
    at async nextLoad (node:internal/modules/esm/loader:173:22)
    at async ESMLoader.load (node:internal/modules/esm/loader:616:20)
    at async ESMLoader.moduleProvider (node:internal/modules/esm/loader:472:11) {
  code: 'ERR_IMPORT_ASSERTION_TYPE_MISSING'
```

So, we need to make sure our JSON imports look something like this in the ESM output of our code:
```js
import * as goerli from './chains/goerli.json' assert {type: "json"};
import * as mainnet from './chains/mainnet.json' assert {type: "json"};
import * as sepolia from './chains/sepolia.json' assert {type: "json"};
```

Vite seems to do this magically for our our browser examples but ideally we add some sort of code transform that would be applied by a bundler (either rollup or ts-patch or webpack or whatever).  

One additional note:  we can't just add these into our sources or Typescript complains about them not being allowed unless we change the `module` setting in our `tsconfig` from `Node16` to `NodeNext` (which in turn will open a bunch of other challenges I haven't fully explored.

One simple hack that works for the Ultralight build process is to use this `sed` command from whatever project directory root where you have imported the `@ethereumjs` modules in your `node_modules` directory.

`sed -i  '/from \S\+\.json'\''/ s/;//;/from \S\+\.json/ s/.*/& assert {type: \"json\"};/' node_modules/@ethereumjs/**/dist/esm/**/*.js `

Unfortunately, this also is brittle and probably not an ideal solution for our build process (proved problematic when run repeatedly because my regex would not recognize that the type assertion had already been added on previous builds).

holgerd77:
Closes #2900 

This PR adds a post-build step to add JSON type assertion additions based on a modified version of the `sed` regex command from @acolytec3.

Phew, this was tricky. Together with another commit I will push after PR submission I am now confident that things work, I tested both on Ubuntu (CLI) and Mac OS (CLI + Script).

Also checked for the `.js-e` files being gone and the changes being applied on the JSON imports.

Open for review.


holgerd77:
Superseded by #2911 (alternative solution to remove JSON imports all together), will close.

Issue #2903, opened 2023-07-20
Fix Security Vulnerabilities


holgerd77:
Currently when installing npm packages with `npm i` from monorepo root we have 12 high severe security vulnerabilities reported:

<img width="476" alt="grafik" src="https://github.com/ethereumjs/ethereumjs-monorepo/assets/931137/391ba908-f186-422c-9df1-daec6d2fc0bf">

Even while most of these will likely "only" occur in dev dependencies, we should nevertheless try to fix most - optimally all - of them, also for the reason to get a better picture and actually not *loose* any of the production dependency vulnerabilities.

If someone tackles, bonus task *would* actually be to find out if there is a good/easy way (especially regarding our more complex monorepo setup), to separate the reporting in a way that we at least have the possibility to just get a report on the production dependencies, so that we can run *that* more frequently and imminently fix if there is some vulnerability showing up.

ScottyPoi:
On this :muscle:

holgerd77:
I would call this sufficiently addressed and there is this one more specific issue Scotty opened on a remaining dependency in client.

Will close here.

PR #2912, opened 2023-07-21
Address security vulnerabilities

ScottyPoi:
Addressing **12** vulnerabilities re issue: #2903 

1. Devp2p
* Removed dependency: @types/chalk@2.2.0: 
* Update import of multiaddr/convert to remove `Require`

2. Client: 
* Update `peer-id` dependency
* ~~Update `multiaddr` dependency to `@multiformats/multiaddr`~~

3. Trie
 * Remove `0x` dependency

**ALL REMAINING VULNERABILITIES**
*caused by:*

`multiaddr` v10.0.01 is deprecated
to be replaced by `@multiformats/multiaddr`

> currently unable to update due to ESM conflicts.


ScottyPoi:
Halfway there!

Screenshot from 2023-07-21 06-05-43

ScottyPoi:
Screenshot from 2023-07-21 07-18-00

packages/client/package.json
```diff
@@ -83,8 +83,9 @@
     "jwt-simple": "^0.5.6",
     "level": "^8.0.0",
     "memory-level": "^1.0.0",
-    "multiaddr": "^10.0.1",
-    "peer-id": "^0.14.3",
+    "@multiformats/multiaddr": "^12.1.3",
+    "@libp2p/interface-peer-id": "2.0.2",
+    "@libp2p/peer-id-factory": "2.0.4",
```
holgerd77, line 88: Maybe @acolytec3 can confirm, but we removed libp2p completely from the core build, so I would think these two dependencies shouldn't show up here at all? 🤔

packages/devp2p/src/dns/enr.ts
```diff
@@ -1,16 +1,15 @@
...
 import { base32, base64url } from '@scure/base'
 import { ecdsaVerify } from 'ethereum-cryptography/secp256k1-compat.js'
-import { Multiaddr } from 'multiaddr'
 import { sscanf } from 'scanf'
 
 import { keccak256, toNewUint8Array } from '../util.js'
 
 import type { PeerInfo } from '../types.js'
 
-const Convert = require('multiaddr/src/convert')
```
holgerd77, line 12: Ah, another require gone, great! 😃

packages/trie/package.json
```diff
@@ -61,7 +61,6 @@
   },
   "devDependencies": {
     "@ethereumjs/genesis": "0.1.0-rc.1",
-    "0x": "^4.9.1",
```
holgerd77, line 64: I am personally ok with dropping the `0x` dependency, this was added for convenience I guess when we were very much into "flame-graphing" 😂, but I guess it's totally ok respectively likely better if such a tools are just installed in a global scope by everyone who uses them.

holgerd77 (commented):
Looks already pretty good, thanks for tackling! 🤩

(1-2 CI runs failing, just to mention)

packages/client/package.json
acolytec3: Depends on how we're using `peer-id`. That library got radically changed a while back so you might need that factory for even our limited usage.

packages/trie/package.json
ScottyPoi: it wasn't in use anywhere that i could find :shrug: 

and it was responsible for like 9 vulnerabilities :laughing:

packages/client/package.json
ScottyPoi: Mostly found in `client/libp2pBrowserBuild`

acolytec3:
So some potential bad news on the front of trying to upgrade the `multiaddr` dep.  That module is now ESM only, so the reason we're seeing this `Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in...` error in the devp2p examples scripts and in the client is that those being run as CommonJS modules and like so many other challenges with ESM, you can't statically import ESM modules in CJS code.  So...since the IPFS/libp2p Javascript world has elected to go ESM only (and not provide CJS builds), we either need to internalize all the multiaddr usage in our code or just pin this old dependency.  Multiaddrs are futureproof so in theory our usage should never change.

acolytec3:
So just to follow up, I'd be a fan of backing out the `multiaddr` and merging this with just the other stuff removed for now and we tackle multiaddrs in a future update.

ScottyPoi:
> So just to follow up, I'd be a fan of backing out the `multiaddr` and merging this with just the other stuff removed for now and we tackle multiaddrs in a future update.

Ok, I'll rewind that change and make an issue about it for future us.

ScottyPoi:
strangely -- reverting back to the deprecated `mutliaddr v10.0.1` did not bring back the vulnerability warning :thinking:

acolytec3 (approved):
LGTM! I think we might be able to get rid of the peerId stuff entirely but that can be a further PR as needed.

Issue #2753, opened 2023-06-05
implement debug_traceCall


drortirosh:
allow the use of debug_traceCall()
This is required for ERC4337 implementation

holgerd77:
I have not enough context on this issue, maybe @acolytec3 can shed a light here? Didn't you have implemented some trace RPC call lately? 🤔 Is this the same/related/whatever?

acolytec3:
Are you possibly referring to [`debug_traceCall`](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debug_traceCall)? I don't see anything regarding debug tracing in the `eth_call` docs.

holgerd77:
Yes.

How close is this to the requested feature? Is this just a small integration?

(again, completely can't judge)

acolytec3:
Ah, sorry, shouldn't try to respond to questions before coffee...my instinct is this should be pretty simple to build out.  I can put it on the todo list.

acolytec3:
@drortirosh 6 weeks late but I've started on this.  Hopefully won't be too much of a lift to get tests implemented and then get this integrated.,

PR #2913, opened 2023-07-21
Add `debug_traceCall` RPC method

acolytec3:
Adds the [`debug_traceCall`](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debugtracecall) method from the geth `debug` namespace which mirrors `eth_call` except runs in a specified block execution context and then returns traces matching `debug_traceTransaction`

Checklist:
- [x] - Implement rpc method
- [x] - Add tests
- [x] - Update docs with jsdoc links


acolytec3:
Once finished, this will address #2753

acolytec3:
Thanks to tests magically appearing courtesy of @ScottyPoi, this ought to be ready for review.

packages/client/test/rpc/debug/traceCall.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+} from '../helpers'
+
+const method = 'debug_traceCall'
+
+describe(method, () => {
+  const manager = createManager(createClient({ opened: true }))
+  const methods = manager.getMethods()
+  const server = startRPC(methods)
+
+  it(' debug_traceCall method exists', async () => {
```
scorbajio, line 25: Seems like there is an unnecessary leading space, here:
```suggestion
  it('debug_traceCall method exists', async () => {
```

packages/client/test/rpc/debug/traceCall.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+  })
+  it(`expects param[0] to be type "object"`, async () => {
+    const req = params(method, ['', ''])
+    const expectRes = (res: any) => {
+      expect(res.body.error.message).toBe('invalid argument 0: argument must be an object')
+      return res
+    }
+    await baseRequest(server, req, 200, expectRes)
+  })
+  it(`expects param[1] to be type string`, async () => {
```
scorbajio, line 36: For consistency:
```suggestion
  it(`expects param[1] to be type "string"`, async () => {
```

packages/client/test/rpc/debug/traceCall.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+      rpcTx = toRpcTx(t)
+    }
+    await baseRequest(server, rpcTxReq, 200, expectResTx, true)
+    const req2 = params('debug_traceCall', [rpcTx, '0x1', {}])
+    const expectRes2 = (res2: any) => {
+      expectTypeOf(res2.body.result)
+        .toHaveProperty('gas')
+        .toHaveProperty('returnValue')
+        .toHaveProperty('failed')
+        .toHaveProperty('structLogs')
```
holgerd77, line 94: Ugh. This reads really nicely. 👍🙂

packages/client/src/rpc/modules/debug.ts
```diff
@@ -170,4 +184,98 @@ export class Debug {
...
+    if (this.vm === undefined) {
+      throw new Error('missing vm')
+    }
+
+    const opts = validateTracerConfig(tracerOpts)
+
+    const block = await getBlockByOption(blockOpt, this.chain)
+    const parentBlock = await this.service.chain.getBlock(block.header.parentHash)
+
+    const vm = await this.vm.shallowCopy()
```
jochem-brouwer, line 216: This has an underlying Trie which writes to disk - right? I don't think we want that, we want that the Trie is in memory for these calls, otherwise we clog the disk with calls which are possibly never made in the chain. (Also a DoS vector since you can spam the RPC with random txs and thus hog disk space)

jochem-brouwer (commented):
One question

packages/client/src/rpc/modules/debug.ts
holgerd77: We need to operate on the same datastructure (and actually: state) and it is not possible/practical to copy over the state to memory on these kind of instantiations.

So I would say on first round memory read is not possible. This *does* use the statemanager + all initialized caches though, so depending on the request this might not need direct trie/state reads.

That said: If I am not mistaken my understanding of RPC (as it is designed for Ethereum) is that it is *always* a potential DoS surface you cannot hide from, and therefore it is just not recommended to completely expose an RPC interface in non-testing/devnet situations.

Calling RPC methods just *does* use additional ressources, and calling them often uses even more. 😋

There are likely more general solutions and mitigations - e.g. block certain request IPs after x calls or after ressouce-intensive calls. Not sure if this is so useful for us atm with no real production users.

acolytec3: This is why we need to implement dynamic snapshots! 😀

Issue #2906, opened 2023-07-20
Client Options CLI Test Suite


holgerd77:
For the client we are still in the situation that too often things at the very usage level (CLI) break unnoticed, especially starting the client in different configurations and with different options do not always trigger one of the unit tests, and if so, this is mostly somewhat accidentally (which is better than nothing).

We noticed this last round with the `--rpc` flag - see #2902. While for this specific issue it turned out that our client CLI tests were just not running *at all* (haha, sigh) atm this kind of breakage could literally happen with all other not tested option (combinations).

We should therefore slowly (or quickly 😋) build up a somewhat complete CLI options test suite.

We *do* have working CLI tests for the client in `test/cli/`, e.g. [test/cli/cli-rpc.spec.ts](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/client/test/cli/cli-rpc.spec.ts).

The RPC test setup already provides some working boilerplate on how to test a CLI option.

We should expand this to have at least a basic on/off test for basically all options we have, to *minimally* ensure that the client is not completely breaking if an option is used. Better if at least some part of the functionality is also confirmed of course though if easily possible.

I think it would likely make sense to group these tests not in a single file but just somewhat by loose topics, somewhat like "networking", "rpc", "engine", "logs" - eventually 1-3 more - and so then `networking.spec.ts` (as a side task we should really remove the redundand `cli-*` name part for these test files respectively rename with git).

(another side-side task: it would be *generally* good if our options would get some grouping. Not sure if this can be done with the current CLI parser library. If it can it might make sense to directly integrate into this work here and *then* align the grouping from the help command `npm run client:start -- --help` (so this would be ideal if options are grouped there) with the groupings from the tests)

Another thing to consider: it is essential to make sure that these CLI runs do not trigger actual sync processes (with the eventual exception of 1-2 tests for the live sync iteself).

Currently I think (might need confirmation) this is achieved by using the `--dev` flag - see `cli-rpc.spec.ts`. It might be more robust and is likely more transparent on what the effect is to use `'--sync=none'` (sometimes maybe also `'--execution=false'`), now that we have these option selections.

acolytec3:
One comment here, using `--dev` by itself just has the client start in the default mainnet setting (which isn't ideal).  You have to pass it a parameter like `--dev=poa` for it to actually start the devnet.  I think another improvement might be to have the client throw an error if `--dev` is passed without an option.

PR #2916, opened 2023-07-25
Add tests for client cli options

scorbajio:
See #2906.


packages/client/test/cli/cli.spec.ts
```diff
@@ -3,19 +3,46 @@ import { Client } from 'jayson/promise'
...
+  it('should successfully start client with a custom network and network id', async () => {
     const file = require.resolve('../../dist/bin/cli.js')
-    const cliArgs = ['--rpc', '--dev=poa']
+    const cliArgs = ['--network=sepolia', '--networkId=11155111', '--dev=poa']
+    const child = spawn(process.execPath, [file, ...cliArgs])
+    return new Promise((resolve) => {
+      child.stdout.on('data', async (data) => {
+        const message: string = data.toString()
+        if (message.includes('Client started successfully')) {
+          assert.ok('Client started successfully', 'read from HTTP RPC')
```
acolytec3, line 15: These test success messages seem a bit imprecise.  Can you adjust the second message so it reflects the what the test is actually doing?  As an example, in this one, the test isn't that that the RPC started but that the client started successfully and that it's using a custom network setup.

acolytec3: Also, this test isn't actually working correctly.  If you combine the network parameters with `--dev=poa`, the devnet parameters take precedent.  You can log the output of the messages to the console by adding something like
```js
      child.stdout.on('data', async (data) => {
        console.log(data.toString())
```

Then, you can adjust the tests accordingly to check for valid log messages that correspond to the test parameters you're working with.

In this test, we should be validating that the below log message has:
`[07-27|06:13:01] INFO Initializing Ethereumjs client version=v0.8.0 network=sepolia chainId=11155111`

but right now it is showing:
`[07-27|06:13:01] INFO Initializing Ethereumjs client version=v0.8.0 network=devnet chainId=123456`

packages/client/test/cli/cli.spec.ts
```diff
@@ -24,25 +51,145 @@ describe('[CLI]', () => {
...
+    const file = require.resolve('../../dist/bin/cli.js')
+    const cliArgs = ['--rpcEngine', '--rpcEngineAuth=false', '--dev=poa']
+    const child = spawn(process.execPath, [file, ...cliArgs])
+    return new Promise((resolve) => {
+      child.stdout.on('data', async (data) => {
+        const message: string = data.toString()
+        if (message.includes('http://')) {
+          // if http endpoint startup message detected, call http endpoint with RPC method
+          assert.ok(message.includes('engine'), 'read from HTTP RPC')
+          assert.ok(message.includes('rpcEngineAuth=false'), 'read from HTTP RPC')
```
acolytec3, line 75: Could we adjust this test to actually call the engine endpoint and confirm you can read from it rather than just reading the log messages?  Here again, the test success message should be a bit more precise, indicating that the engine RPC endpoints were started.

packages/client/test/cli/cli.spec.ts
```diff
@@ -24,25 +51,145 @@ describe('[CLI]', () => {
...
+  it('should start engine rpc on custom port', async () => {
+    const file = require.resolve('../../dist/bin/cli.js')
+    const cliArgs = ['--rpcEngine', '--rpcEnginePort=8552', '--dev=poa']
+    const child = spawn(process.execPath, [file, ...cliArgs])
+    return new Promise((resolve) => {
+      child.stdout.on('data', async (data) => {
+        const message: string = data.toString()
+        if (message.includes('http://')) {
+          // if http endpoint startup message detected, call http endpoint with RPC method
+          assert.ok(message.includes('8552'), 'read from HTTP RPC')
```
acolytec3, line 104: Same as above, can you try calling the engine endpoint on the custom port and verify it's available there?

acolytec3 (changes requested):
This is a good starting point.  

I haven't gone in great detail through every single test (though I'm happy to if helpful) but a lot of these tests all have the same test success message, `read from HTTP RPC`, which doesn't reflect what the test is actually doing.  Will try and add some more specific comments later this morning.

packages/client/test/cli/cli.spec.ts
```diff
@@ -3,19 +3,46 @@ import { Client } from 'jayson/promise'
...
-
     return new Promise((resolve) => {
       child.stdout.on('data', async (data) => {
         const message: string = data.toString()
         if (message.includes('http://')) {
           // if http endpoint startup message detected, call http endpoint with RPC method
-          const client = Client.http({ port: 8545 })
-          const res = await client.request('web3_clientVersion', [], 2.0)
-          assert.ok(res.result.includes('EthereumJS'), 'read from HTTP RPC')
+          assert.ok(message.includes('engine'), 'read from HTTP RPC')
```
acolytec3, line 45: Same here, let's add an RPC call to an engine endpoint and get a valid response

packages/client/test/cli/cli.spec.ts
```diff
@@ -24,25 +51,145 @@ describe('[CLI]', () => {
...
+  it('should start engine rpc on custom address', async () => {
+    const file = require.resolve('../../dist/bin/cli.js')
+    const cliArgs = ['--rpcEngine', '--rpcEngineAddr="0.0.0.0"', '--dev=poa']
+    const child = spawn(process.execPath, [file, ...cliArgs])
+    return new Promise((resolve) => {
+      child.stdout.on('data', async (data) => {
+        const message: string = data.toString()
+        if (message.includes('http://')) {
+          // if http endpoint startup message detected, call http endpoint with RPC method
+          assert.ok(message.includes('0.0.0.0'), 'read from HTTP RPC')
```
acolytec3, line 133: As above, add an RPC call that verifies you get the engine authentication error message when calling an engine endpoint on the custom address and then clarify the test message

packages/client/test/cli/cli.spec.ts
```diff
@@ -24,25 +51,145 @@ describe('[CLI]', () => {
...
+      '--wsEngineAddr="0.0.0.0"',
+      '--dev=poa',
+    ]
+    const child = spawn(process.execPath, [file, ...cliArgs])
+    return new Promise((resolve) => {
+      child.stdout.on('data', async (data) => {
+        const message: string = data.toString()
+        if (message.includes('ws://') && message.includes('engine')) {
+          // if http endpoint startup message detected, call http endpoint with RPC method
+          assert.ok(message.includes('ws://0.0.0.0:8552'), 'read from HTTP RPC')
```
acolytec3, line 168: Change the success message to `read from WS RPC on custom address` and add an actual call to the endpoint (and verify you get the invalid credentials error)

packages/client/test/cli/cli.spec.ts
```diff
@@ -61,20 +208,177 @@ describe('[CLI]', () => {
...
+    const child = spawn(process.execPath, [file, ...cliArgs])
+    return new Promise((resolve) => {
+      child.stdout.on('data', async (data) => {
+        const message: string = data.toString()
+        if (message.includes('ws://')) {
+          // if ws endpoint startup message detected, call ws endpoint with RPC method
+          const client = Client.websocket({ url: 'ws://localhost:8546' })
+          ;(client as any).ws.on('open', async function () {
+            const res = await client.request('web3_clientVersion', [], 2.0)
+            assert.ok(res.result.includes('EthereumJS'), 'read from WS RPC')
```
acolytec3, line 234:
```suggestion
            assert.ok(res.result.includes('EthereumJS'), 'read from WS RPC on custom port')
```

packages/client/test/cli/cli.spec.ts
```diff
@@ -61,20 +208,177 @@ describe('[CLI]', () => {
...
+    const child = spawn(process.execPath, [file, ...cliArgs])
+    return new Promise((resolve) => {
+      child.stdout.on('data', async (data) => {
+        const message: string = data.toString()
+        if (message.includes('ws://')) {
+          // if ws endpoint startup message detected, call ws endpoint with RPC method
+          const client = Client.websocket({ url: 'ws://0.0.0.0:8546' })
+          ;(client as any).ws.on('open', async function () {
+            const res = await client.request('web3_clientVersion', [], 2.0)
+            assert.ok(res.result.includes('EthereumJS'), 'read from WS RPC')
```
acolytec3, line 267:
```suggestion
            assert.ok(res.result.includes('EthereumJS'), 'read from WS RPC on custom address')
```

packages/client/test/cli/cli.spec.ts
```diff
@@ -97,17 +401,281 @@ describe('[CLI]', () => {
...
+  // logging and documentation tests
+  it('should start HTTP RPC and return valid responses', async () => {
+    const file = require.resolve('../../dist/bin/cli.js')
+    const cliArgs = ['--rpc', '--helpRpc=true', '--dev=poa']
+    const child = spawn(process.execPath, [file, ...cliArgs])
+    return new Promise((resolve) => {
+      child.stdout.on('data', async (data) => {
+        const message: string = data.toString()
+        if (message.includes('JSON-RPC: Supported Methods')) {
+          assert.ok(message, 'read from HTTP RPC')
```
acolytec3, line 424:
```suggestion
          assert.ok(message, 'logged out supported RPC methods')
```

packages/client/test/cli/cli.spec.ts
```diff
@@ -97,17 +401,281 @@ describe('[CLI]', () => {
...
-
       child.on('close', (code) => {
         if (typeof code === 'number' && code > 0) {
           assert.fail(`child process exited with code ${code}`)
         }
       })
     })
   }, 20000)
+  // logging and documentation tests
+  it('should start HTTP RPC and return valid responses', async () => {
```
acolytec3, line 416:
```suggestion
  it('should log out available RPC methods', async () => {
```

packages/client/test/cli/cli.spec.ts
```diff
@@ -97,17 +401,281 @@ describe('[CLI]', () => {
...
+    const file = require.resolve('../../dist/bin/cli.js')
+    const cliArgs = ['--mine=true', '--forceSnapSync=true', '--dev=poa', '--port=30304']
+    const child = spawn(process.execPath, [file, ...cliArgs])
+    return new Promise((resolve) => {
+      child.stdout.on('data', async (data) => {
+        const message: string = data.toString()
+        if (message.includes('Client started successfully')) {
+          assert.ok(
+            message.includes('Client started successfully'),
+            'Clients started with experimental feature options'
```
acolytec3, line 579: If there's a way to infer that snap sync did start (from the logs), it would be nice to capture that here.

packages/client/test/cli/cli.spec.ts
```diff
@@ -97,17 +401,281 @@ describe('[CLI]', () => {
...
+      '--dev=poa',
+    ]
+    const child = spawn(process.execPath, [file, ...cliArgs])
+    return new Promise((resolve) => {
+      child.stdout.on('data', async (data) => {
+        const message: string = data.toString()
+        if (message.includes('Client started successfully')) {
+          assert.ok(
+            message.includes('Client started successfully'),
+            'Clients starts with client execution limits options'
```
acolytec3, line 660:
```suggestion
            'Clients starts with custom network parameters'
```

packages/client/test/cli/cli.spec.ts
```diff
@@ -97,17 +401,281 @@ describe('[CLI]', () => {
...
+      '--extIP=0.0.0.0',
+      '--rpcCors=https://foo.example',
+      '--dnsAddr=8.8.8.8',
+      '--dev=poa',
+    ]
+    const child = spawn(process.execPath, [file, ...cliArgs])
+    return new Promise((resolve) => {
+      child.stdout.on('data', async (data) => {
+        const message: string = data.toString()
+        if (message.includes('Client started successfully')) {
```
acolytec3, line 657: Can you validate the `Server listener up` instead and verify that we're getting the correct `extIP` and `port`?
The log message will look like:
`[07-27|10:02:22] INFO Server listener up transport=rlpx url=enode://7a03e93feceb3f9905f82c2fe498eb4d13c48ad62e3880b79a7cc8c0021e32487b11e5318044e9a5617d9e977dba9633bcf63bce60dbf5aff0b4caee4078a798@0.0.0.0:2100`

YOu can just split the string on the `@` and parse the IP and port address to verify.

acolytec3 (commented):
Added more complete review comments.

packages/client/test/cli/cli.spec.ts
scorbajio: fixed

packages/client/test/cli/cli.spec.ts
scorbajio: fixed

packages/client/test/cli/cli.spec.ts
scorbajio: fixed

packages/client/test/cli/cli.spec.ts
scorbajio: fixed

packages/client/test/cli/cli.spec.ts
scorbajio: fixed

packages/client/test/cli/cli.spec.ts
scorbajio: fixed

packages/client/test/cli/cli.spec.ts
scorbajio: fixed

packages/client/test/cli/cli.spec.ts
scorbajio: fixed

packages/client/test/cli/cli.spec.ts
scorbajio: fixed

packages/client/test/cli/cli.spec.ts
scorbajio: fixed

packages/client/test/cli/cli.spec.ts
scorbajio: Currently, the messages still aren't updated to log that snap sync is happening. Will change that and update the tests in a different PR.

packages/client/test/cli/cli.spec.ts
scorbajio: fixed

holgerd77:
Hi @scorbajio, thanks a lot, great new tests! 🤩

The only basic thing that bothers me around this PR is that it is adding so much new code. Can we do one additional round and try to minimize?

I think one generic helper method where the `child` objects can be passed and which adds the two generic listeners for `.stderr.on('data',...)` and `child.on('close', (code),...)` should definitely work?

And maybe parts for the `child.stdout.on('data',...)` listener can be generalized as well, would be nice, haven't looked so closely though yet. Maybe by passing in a callback for the custom part to a helper function, or by adding 2-3 other helper functions for reoccuring stuff, not sure, again, didn't have such a close look here.

Additionally a small helper for populating the `cliArgs` array might also make sense.

Alternatively @acolytec3 might also be able to directly do if he is working on some stuff here, just to mention the possibility.

acolytec3 (approved):
This looks great!  Thanks for sticking it through.

holgerd77:
This is beyond what I expected we would start with, thanks for this fantastic test suite! 🤩🙏

PR #2921, opened 2023-07-28
Handle SIGTERM in client

acolytec3:
This PR supersedes #2899 (by documenting the Docker approved way to start node in Docker containers) and also adds a handler for the `SIGTERM` kernel signal that is often used to shut down Docker containers.


g11tech (approved):
aha so SIGTERM was not being handled, lgtm

PR #2922, opened 2023-07-28
Client/StateManager: storageRangeAt() RPC call / EVMStateManager Interface Extension

KoningR:
This PR implements `debug_storageRangeAt()` as discussed in ethereum/go-ethereum#14350. An implementation is only provided for `DefaultStateManager`, not for `EthersStateManager`. Additionally, hashed key preimages are always `null` because there does not seem to be a good way to retrieve them currently.


packages/client/src/rpc/modules/debug.ts
```diff
@@ -92,6 +101,7 @@ export class Debug {
       [validators.transaction()],
       [validators.blockOption],
     ])
+    this.storageRangeAt = middleware(this.storageRangeAt.bind(this), 5, [[validators.hex]])
```
g11tech, line 104: the validators here will only validate the first arg, reading from comments need 5 validators

packages/client/src/rpc/modules/debug.ts
```diff
@@ -278,4 +288,80 @@ export class Debug {
       message: err.message.toString(),
     }
   }
+
+  /**
+   * Returns a limited set of storage keys belonging to an account.
+   * @param params An array of 5 parameters:
+   *    1. The hash of the block that contains the requested account storage.
```
g11tech, line 295:
```suggestion
   *    1. The hash of the block at which to get storage from the state.
```
just to make it less ambiguous

g11tech, line 295:
```suggestion
   *    1. The hash of the block at which to get storage from the state.
```
just making it less ambiguous

packages/client/src/rpc/modules/debug.ts
```diff
@@ -278,4 +288,80 @@ export class Debug {
...
+   *    2. The number index of the requested transaction within the block.
```
g11tech, line 296:
```suggestion
   *    2. The transaction index of the requested block post which to get the storage.
```

packages/client/src/rpc/modules/debug.ts
```diff
@@ -278,4 +288,80 @@ export class Debug {
...
+   *    3. The address of the account.
+   *    4. The smallest (hashed) storage key that will be returned. To include the entire range, pass '0x00'.
```
g11tech, line 298:
```suggestion
   *    4. The starting (hashed) key from which storage will be returned. To include the entire range, pass '0x00'.
```

packages/client/src/rpc/modules/debug.ts
```diff
@@ -278,4 +288,80 @@ export class Debug {
...
+   *    5. The maximum number of storage values that will be returned.
```
g11tech, line 299:
```suggestion
   *    5. The maximum number of storage values that could be returned.
```

packages/client/src/rpc/modules/debug.ts
```diff
@@ -278,4 +288,80 @@ export class Debug {
...
+    } catch (err: any) {
+      throw {
+        code: INVALID_PARAMS,
+        message: err.message.toString(),
+      }
+    }
+
+    let block: Block
+    try {
+      block = await this.service.chain.getBlock(blockHashBytes)
```
g11tech, line 329:
```suggestion
      block = await this.chain.getBlock(blockHashBytes)
```
this.chain is available

packages/client/src/rpc/modules/debug.ts
```diff
@@ -278,4 +288,80 @@ export class Debug {
...
+    } catch (err: any) {
+      throw {
+        code: INTERNAL_ERROR,
+        message: 'Could not get requested block hash.',
+      }
+    }
+
+    if (txIndex + 1 > block.transactions.length) {
```
g11tech, line 337:
```suggestion
    if (txIndex  >= block.transactions.length) {
```
more comprehensible about the check

packages/client/src/rpc/modules/debug.ts
```diff
@@ -278,4 +288,80 @@ export class Debug {
...
+      throw {
+        code: INTERNAL_ERROR,
+        message: 'txIndex cannot be larger than the number of transactions in the block.',
+      }
+    }
+
+    try {
+      const block = await this.service.chain.getBlock(hexToBytes(blockHash))
```
g11tech, line 345: isn't the block already read above (in params check) also `this.chain` can be directly refereed

packages/client/src/rpc/modules/debug.ts
```diff
@@ -278,4 +288,80 @@ export class Debug {
...
+      }
+    }
+
+    let address: Address
+    let blockHashBytes: Uint8Array
+    let startKeyBytes: Uint8Array
+    try {
+      address = Address.fromString(account)
+      blockHashBytes = hexToBytes(blockHash)
+      startKeyBytes = hexToBytes(startKey)
```
g11tech, line 319: `startKeyBytes` could be directly converted to bigint using a hexToBig... helper

packages/client/src/rpc/modules/debug.ts
```diff
@@ -278,4 +288,80 @@ export class Debug {
...
+   * @param params An array of 5 parameters:
+   *    1. The hash of the block that contains the requested account storage.
+   *    2. The number index of the requested transaction within the block.
+   *    3. The address of the account.
+   *    4. The smallest (hashed) storage key that will be returned. To include the entire range, pass '0x00'.
+   *    5. The maximum number of storage values that will be returned.
+   * @returns A {@link StorageRange} object that will contain at most `limit` entries in its `storage` field.
+   * The object will also contain `nextKey`, the next (hashed) storage key after the range included in `storage`.
+   */
+  async storageRangeAt(params: [string, number, string, string, number]) {
```
g11tech, line 303:
```suggestion
  async storageRangeAt(params: [string, string, string, string, string]) {
```
most like it will come as a hex encoded number ... not sure about that but do check and update

holgerd77:
Thanks a lot for this PR! 🙏🙂❤️

KoningR:
It appears that parameters `txIndex` and `limit` are indeed passed as `number` instead of as hex strings. Currently, I have omitted validation for these parameters because I could not find a suitable validator in `validation.ts`. Is it suggested that I parse the input myself or is there a proper validator for the `number` type?

packages/client/src/rpc/modules/debug.ts
KoningR: There does not appear to exist a `hexToBigInt()` helper. I have however moved the function calls to the same line.

packages/client/src/rpc/modules/debug.ts
```diff
@@ -92,6 +101,14 @@ export class Debug {
       [validators.transaction()],
       [validators.blockOption],
     ])
+    // TODO: txIndex and limit are currently not input validated.
```
ScottyPoi, line 104: do we need new validator types for these?

packages/client/test/rpc/debug/storageRangeAt.spec.ts
```diff
@@ -0,0 +1,480 @@
...
+    const block = await blockBuilder.build()
+    await chain.putBlocks([block], true)
+
+    context.server = server
+    context.block = await chain.getCanonicalHeadBlock()
+    context.createdAddress = result.createdAddress!!
+    context.createdAddressNoStorage = thirdResult.createdAddress!!
+  })
+
+  it<TestSetup>('Should return the correct (number of) key value pairs.', async ({
```
ScottyPoi, line 159: What does the `<TestSetup>` type parameter do?  is it just a label, or does it change the `it` function?

packages/client/src/rpc/modules/debug.ts
```diff
@@ -278,4 +295,70 @@ export class Debug {
...
+   * The object will also contain `nextKey`, the next (hashed) storage key after the range included in `storage`.
+   */
+  async storageRangeAt(params: [string, number, string, string, number]) {
+    const [blockHash, txIndex, account, startKey, limit] = params
+
+    if (this.vm === undefined) {
+      throw new Error('Missing VM.')
+    }
+
+    if (txIndex < 0) {
```
ScottyPoi, line 317: if we don't already have a validator for `index` parameters, it could be worth adding one. especially if there are other methods are doing this themselves

packages/client/src/rpc/modules/debug.ts
```diff
@@ -278,4 +295,70 @@ export class Debug {
...
+      const vmCopy = await this.vm.shallowCopy()
+      await vmCopy.stateManager.setStateRoot(parentBlock.header.stateRoot)
+      for (let i = 0; i <= txIndex; i++) {
+        await vmCopy.runTx({ tx: block.transactions[i], block })
+      }
+
+      return await vmCopy.stateManager.dumpStorageRange(
+        // Validator already verified that `account` and `startKey` are properly formatted.
+        Address.fromString(account),
+        bytesToBigInt(hexToBytes(startKey)),
```
ScottyPoi, line 354:
```suggestion
        BigInt(startKey),
```

packages/client/src/rpc/modules/debug.ts
ScottyPoi: if the hex is '0x' prefixed, you can just use the BigInt constructor like:  `BigInt('0xabcde')

packages/client/src/rpc/modules/debug.ts
ScottyPoi: The reference implementation does use `int` for both of these.

https://github.com/ethereum/go-ethereum/blob/4e9775668efc329cd4d12644f5310c1b5ac2ab41/eth/api_debug.go#L212

ScottyPoi:
> > It appears that parameters `txIndex` and `limit` are indeed passed as `number` instead of as hex strings. Currently, I have omitted validation for these parameters because I could not find a suitable validator in `validation.ts`. Is it suggested that I parse the input myself or is there a proper validator for the `number` type?
> 
> If you are open to do maybe add a very simple `get int()` validator to `validation.ts`, as simple as the `bool` one:
> 
> ```ts
> /**
>    * bool validator to check if type is boolean
>    * @param params parameters of method
>    * @param index index of parameter
>    */
>   get bool() {
>     return (params: any[], index: number) => {
>       if (typeof params[index] !== 'boolean') {
>         return {
>           code: INVALID_PARAMS,
>           message: `invalid argument ${index}: argument is not boolean`,
>         }
>       }
>     }
>   }
> ```

Are there methods that use signed integers (negative numbers)?  "index" will always be unsigned, so it might be better to add `getUint`, which throws if the input is < 0

packages/client/test/rpc/debug/storageRangeAt.spec.ts
KoningR: It is a typesafe way of passing variables from `beforeEach()` to tests (see [here](https://vitest.dev/guide/test-context.html#typescript-1)). For legibility, I had already annotated the `TestSetup` interface.

ScottyPoi (approved):
Looks great!  Thanks for including all the tests, and for adding the validator!  super helpful :+1:

jochem-brouwer (approved):
LGTM

g11tech (approved):
lgtm

holgerd77:
@KoningR Hi there, thanks again for this great work! 🤩

We'll directly that this into the upcoming releases (so: today)!

PR #2923, opened 2023-07-29
verkle: verkle package

gabrocheleau:
This PR adds a new package: @ethrereumjs/verkle. Similarly to the trie package, this package will implement verkle tries as specified https://notes.ethereum.org/@vbuterin/verkle_tree_eip, enabling us to create verkle tries, do operations on them, and use them to verify and generate proofs. 

### Notes

I've identified cryptographic primitives that we are missing for a complete implementation (based on Go's implementation here: https://github.com/gballet/go-ethereum/blob/kaustinen-with-shapella/trie/trie.go). 

- `Point` class. Comes from `banderwagon.Element`. Quite fundamental as it represents a polynominal commitment, with a number of methods used repeatedly in the implementation.
- `toFrMultiple`. Implementation comes from Banderwagon's `MultiMapToScalarField` method. 
- `Fr` class. Field representation class. Comes from Banderwagon's `fr.Element`. 
- CommitToPoly. Comes from `ipa.IPAConfig` in `go-ipa/ipa`. ([link](https://github.com/gballet/go-verkle/blob/master/config_ipa.go#L91)). Used for example when creating new leaf nodes.


packages/client/CHANGELOG.md
```diff
@@ -562,5 +562,3 @@ showing the relations between the main components as well as the initialization
 ## [0.0.1] - 2018-10-26
 
 - Initial development release
-
-[0.0.1]: https://github.com/ethereumjs/ethereumjs-client/tree/v0.0.1
```
jochem-brouwer, line 566: This URL is removed? (but URL still works?)

packages/verkle/src/db/checkpoint.ts
```diff
@@ -0,0 +1,275 @@
...
+    // Lookup the value in our diff cache. We return the latest checkpointed value (which should be the value on disk)
+    for (let index = this.checkpoints.length - 1; index >= 0; index--) {
+      if (this.checkpoints[index].keyValueMap.has(keyHex)) {
+        return this.checkpoints[index].keyValueMap.get(keyHex)
+      }
+    }
+    // Nothing has been found in diff cache, look up from disk
+    const valueHex = await this.db.get(keyHex, {
+      keyEncoding: KeyEncoding.String,
+      valueEncoding: ValueEncoding.String,
```
jochem-brouwer, line 157: I think with Verkle we use a completely new database, right? (So not related to old MPT DB). If that's the case, can we immediately use `valueEncoding` set to `ValueEncoding.Bytes`? Then we don't have to convert `bytesToUnprefixedHex` early on and we don't have to deal with a breaking release for the underlying database later on.

jochem-brouwer: (same maybe also for keys directly)

packages/verkle/src/util/lock.ts
```diff
@@ -0,0 +1,42 @@
+// Based on https://github.com/jsoendermann/semaphore-async-await/blob/master/src/Semaphore.ts
+export class Lock {
```
jochem-brouwer, line 2: This file also exists in `packages/util/src/lock.ts`?

jochem-brouwer (commented):
Ok, I have given this a look (but not in-depth), some minor comments and questions.

In general regarding the structure: my first thought would be that this is great, it is essentially MPT but now with Verkle support (so I assume it would be rather "easy" (?) to swap MPT for Verkle - of course only at the start of a client not having to do transition MPT -> Verkle).

I will need to study Verkle a bit to also review the internals, since I cannot comment on that right now.

holgerd77:
Very very very cool!!! ❤️ 🤩 This looks like an extremely solid fundament already where we can further build upon! Would vote on taking this in pretty soon (this or next week)!

packages/util/src/bytes.ts
```diff
@@ -444,5 +444,66 @@ export const concatBytes = (...arrays: Uint8Array[]): Uint8Array => {
...
+  bytes[2] = Number((lo >> BigInt(16)) & BigInt(0xff))
+  bytes[3] = Number((lo >> BigInt(24)) & BigInt(0xff))
+
+  bytes[4] = Number(hi & BigInt(0xff))
+  bytes[5] = Number((hi >> BigInt(8)) & BigInt(0xff))
+  bytes[6] = Number((hi >> BigInt(16)) & BigInt(0xff))
+  bytes[7] = Number((hi >> BigInt(24)) & BigInt(0xff))
+
+  return bytes
+}
```
holgerd77, line 506: These new helper methods look more "advanced" (lol) than our classic conversion helpers for BE! 🙂 🤔 Also love that they explicitly target already 32/64 bit e.g..

Can you say something how they compare and relate to our existing methods, e.g. performance wise? This explicit bit-length also has the nice (and eventually performance-increasing) side effect that it omits the 0-padding if I see correctly?

If these methods are better than our old ones: might it be a mid-term option to also adopt this for our BE methods?

(would like to at least high-level settle on these kind of questions before we take new methods with new naming schemes and everything into Util)

packages/verkle/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Initialization and Basic Usage
+
+```typescript
+import { VerkleTrie } from '@ethereumjs/verkle'
+
+const trie = new VerkleTrie()
+
+async function test() {
+  await trie.put(utf8ToBytes('test'), utf8ToBytes('one'))
```
scorbajio, line 40: If `utf8ToBytes` and `bytesToUtf8` are used in the example it would be more clear to include them in the imports as well:
```
import { bytesToUtf8, utf8ToBytes } from 'ethereum-cryptography/utils.js'
```

packages/verkle/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+The EthereumJS Verkle package is still in its infancy, and as such, it does not currently support Verkle proof creation and verification. Support for Verkle proofs will be added eventually.
+
+## Examples
+
+You can find additional examples complete with detailed explanations [here](./examples/README.md).
+
+## Browser
+
+With the breaking release round in Summer 2023 we have added hybrid ESM/CJS builds for all our libraries (see section below) and have eliminated many of the caveats which had previously prevented a frictionless browser usage.
```
scorbajio, line 60:
```suggestion
With the breaking release round in Summer 2023 we have added hybrid ESM/CJS builds for all our libraries (see section below) and have eliminated many of the caveats which had previously prevented frictionless browser usage.
```

packages/verkle/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Docs
+
+Generated TypeDoc API [Documentation](./docs/README.md)
+
+### Hybrid CJS/ESM Builds
+
+With the breaking releases from Summer 2023 we have started to ship our libraries with both CommonJS (`cjs` folder) and ESM builds (`esm` folder), see `package.json` for the detailed setup.
+
+If you use an ES6-style `import` in your code files from the ESM build will be used:
```
scorbajio, line 74:
```suggestion
If you use an ES6-style `import` in your code, files from the ESM build will be used:
```

packages/verkle/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+```typescript
+import { EthereumJSClass } from '@ethereumjs/[PACKAGE_NAME]'
+```
+
+If you use Node.js specific `require`, the CJS build will be used:
```
scorbajio, line 80:
```suggestion
If you use Node.js-specific `require`, the CJS build will be used:
```

packages/verkle/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+```typescript
+const { EthereumJSClass } = require('@ethereumjs/[PACKAGE_NAME]')
+```
+
+Using ESM will give you additional advantages over CJS beyond browser usage like static code analysis / Tree Shaking which CJS can not provide.
```
scorbajio, line 86:
```suggestion
Using ESM will give you additional advantages over CJS beyond browser usage like static code analysis / Tree Shaking, which CJS cannot provide.
```

scorbajio (commented):
This is looking great! I'm not sure if something like this already exists for this package, but since it is in development, something worth considering that might attract help/contributors is if we could keep an updated list of issues with an overall master tracker on work that still needs to get done.

packages/verkle/src/util/lock.ts
gabrocheleau: Didn't realized that (had copied from trie), will refactor.

packages/util/src/bytes.ts
gabrocheleau: I've researched this a bit, and I think we could strike a balance between high performance and readability (the current implementation is quite opaque) by using DataView to go from Uint8Arrays to ints/bigints. I'm reading that DataViews are very performant & optimized, and the methods have built-in support for BE/LE. See this reference for more info: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/getBigInt64

gabrocheleau: When you mention "adopting this for our BE methods", what methods are you referring to? `bytesToBigInt` and `bytesToInt`? I can implement alternative versions using DataViews and compare performance.

packages/client/CHANGELOG.md
gabrocheleau: Removed by mistake, have reverted.

packages/verkle/src/util/lock.ts
gabrocheleau: Done

packages/verkle/src/db/checkpoint.ts
gabrocheleau: Good point, will go with bytes.

packages/verkle/README.md
gabrocheleau: Agreed

gabrocheleau: Done

packages/verkle/src/db/checkpoint.ts
gabrocheleau: Updated

gabrocheleau: I've updated all key/value types to Uint8Arrays (db, cache), but kept the Checkpoint `keyValueMap` as `Map<string, Uint8Array | undefined>` since Uint8Arrays can't directly be used as Map keys (since two uint8arrays are not equal even if they contain the same data)

gabrocheleau: I'm now reading that using Uint8Array as key for the cache (the cache uses the LRUCache package) might cause issues. I will do some testing but I think I'll revert that part, so we'd end up with:

- A DB that uses Uint8Arrays for keys and values
- A cache (LRU-cache) that uses `string` for keys and Uint8Array as values
- A keyValueMap (`Checkpoint` type) that uses `string` as keys and Uint8Array as values.

gabrocheleau: Indeed, this fails:

image

So I'll switch the cache to use hexstrings for keys.

holgerd77 (approved):
LGTM

Let's take this in and continue to refine and evolve. 🙂

jochem-brouwer:
`Test-verkle` and `test-examples` runs failed on this CI and they now also fail on master + PRs (should we make these mandatory so we cannot accidentally merge if those fail? Since it is rather error-prone if two are failing, this is in some cases codecov messing up :sweat_smile: )

https://github.com/ethereumjs/ethereumjs-monorepo/actions/runs/6645612344/job/18057315347
https://github.com/ethereumjs/ethereumjs-monorepo/actions/runs/6645612366/job/18057315572

PR #2928, opened 2023-07-31
Extend fullsync test timeout

acolytec3:
The `fullsync` integration test routinely timeouts so experiments with extending that test timeline to see if this resolves the issue.


holgerd77 (approved):
Tests are passing, will give this a tentative approval, feel free to merge if everything is solid from your side.

PR #2930, opened 2023-08-01
Fix rpc custom address bug

### LINKED ISSUE #2924: Properly apply `rpcAddress` parameter in `client` CLI
In the `client`, we list `rpcAddress`, `rpcEngineAddr`, `wsAddr`, and `wsEngineAddr` as parameters that allow the user to specify a custom IP address to open the JSON-RPC server on.  While looking at #2916, I discovered that parameter inputs for these 4 parameters are ignored when setting up the servers.  We should be passing `rpcAddr` as the second paramter [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/654c67aa411d41b7a7a31bb5b9131f59e7da857f/packages/client/bin/startRpc.ts#L132) when we start the server (and similarly for the engine and websocket RPC servers).  I've validated that simply updating this code to say `rpcHttpServer.listen(rpcPort, rpcAddr)` makes the basic HTTP JSON-RPC work but we need to go through all the branches of the logic for starting the RPC servers and verify that the custom address parameter for each variant is honored when starting the servers and then write appropriate tests.

acolytec3:
Fixes #2924


scorbajio (approved):
ACK

holgerd77:
Cool PR 😎, impressive that this slipped through so long.

Issue #2927, opened 2023-07-31
StateManager: Re-use proof Trie / static Proof Functionality


holgerd77:
In the DefaultStateManager `verifyProof()` implementation a Trie object is newly created on two occasions each time the method is called.

This can and should be optimized I guess by one-time-instantiate a `_proofTrie` as a protected member of DefaultStateManager and re-use this trie object for the calls.

Another thought line I had was if it would rather make sense to make both Trie and SM `verifyProof()` methods static, didn't completely followed the thought but that might be an alternative option.

jochem-brouwer:
By #2932 it seems like this should be re-opened? Will do.

holgerd77:
Issue description is relatively short and generic, and proof functionality has been reworked on some broad basis in mentioned PRs, will close.

If there are follow-up needs please open a more specific issue.

PR #2932, opened 2023-08-01
Reuse proof trie

scorbajio:


acolytec3 (approved):
LGTM, though I think we should really go with making `verifyProof` a static method on the `trie` class and not have this random `trie` object hanging around the state manager.

holgerd77:
Partly addresses #2927 

@acolytec3 ok, will keep the issue open and change the title.

PR #2933, opened 2023-08-01
Use version 12.3 of ethereum/test

scorbajio:


jochem-brouwer:
Can we also locally run the EIP tests to check if we pass those?

holgerd77:
> Can we also locally run the EIP tests to check if we pass those?

Can you add some instructions on how to do?

scorbajio:
The blockchain and state tests all pass from the EIP test suit when run locally.

jochem-brouwer:
Ah no, sorry, the tests in `./packages/ethereum-tests/EIPTests`. We do **not** cover those in the CI (and we should not do either since these EIPs get updated a lot, and therefore the tests should also change). It would however be interesting to see if we diverge somewhere, and we should figure out if the tests are wrong (by the spec of the EIP) or we are wrong.

(This can be done beyond the scope of this PR, but we should do at some point to prepare for devnet-8)

holgerd77 (approved):
LGTM

PR #2931, opened 2023-08-01
Cancun tests for devnet-8

jochem-brouwer:
This PR is used to track work on the Cancun tests for devnet-8.

To run;

Git clone https://github.com/marioevz/hive/tree/cancun-dn8-4844-4788

Merge this PR: https://github.com/ethereum/hive/pull/809

Change `hive/clients/ethereumjs/ethereumjs.sh`: `FLAGS="--gethGenesis ./genesis.json --rpc --rpcEngine --saveReceipts --rpcAddr 0.0.0.0 --rpcEngineAddr 0.0.0.0 --rpcEnginePort 8551 --ws false --logLevel debug --rpcDebug --transports rlpx --isSingleNode"`

`./hive --sim ethereum/engine --client ethereumjs --sim.limit "engine-blobs/"`

More output;

`./hive --sim ethereum/engine --client ethereumjs --docker.output --loglevel 5`

Do not merge since this likely finds bugs in hive.
TODOS

- [ ] Get #2935 ready
- [x] Fix above tests (likely devp2p related regarding `eth/68`)
- [ ] Open a PR for hive to patch `ethereumjs.sh`. NOTE: this should allow the websockets to be open. We currently disable the websockets in order to run these tests. (i.e. `-- ws false` should be replaced by the default `--ws`)

PR #2935, opened 2023-08-03
Fix engine-blobs hive tests

acolytec3:
Addresses a failing test noted in #2931 where we should be broadcasting the contents of our transaction pool to newly connected peers.  Still WIP

This PR is a continuation of #2931, to test hive and other notes: instructions are there.

This also fixes a performance issue in the `sendNewTransactionHashes`function in `txPool` where it was `awaiting` a devp2p message that never gets a response and was causing some devp2p message handlers to time out silently.


holgerd77:
Actually that’s not only “a failing test” but real misbehavior in our client. 😋

jochem-brouwer:
For docs: this is relevant and super nice to read https://github.com/ethereum/devp2p/blob/master/caps/eth.md#transaction-exchange

packages/client/src/service/fullethereumservice.ts
```diff
@@ -143,6 +145,24 @@ export class FullEthereumService extends EthereumService {
...
+      // TODO: Should we do this if the txPool isn't started?
+      const txs: [number[], number[], Uint8Array[]] = [[], [], []]
+      for (const addr of this.txPool.pool) {
+        for (const tx of addr[1]) {
+          const rawTx = tx.tx
+          txs[0].push(rawTx.type)
+          if (rawTx.type !== 3) {
+            txs[1].push(rawTx.serialize().byteLength)
+          } else {
+            txs[1].push((rawTx as BlobEIP4844Transaction).serializeNetworkWrapper().byteLength)
```
g11tech, line 159: shouldn't we annouce tx hashes over there?

packages/client/src/service/service.ts
```diff
@@ -77,9 +77,9 @@ export class Service {
     this.config.events.on(Event.POOL_PEER_BANNED, (peer) =>
       this.config.logger.debug(`Peer banned: ${peer}`)
     )
-    this.config.events.on(Event.POOL_PEER_ADDED, (peer) =>
+    this.config.events.on(Event.POOL_PEER_ADDED, (peer) => {
```
g11tech, line 80: i think the previous version was also fine

g11tech (approved):
lgtm :partying_face:

PR #2941, opened 2023-08-06
Snap Sync Fetchers: Highest-Known-Hash Optimization

scorbajio:
Currently, the account and storage fetchers statically partition large hash ranges of accounts and storage slots to fetch each range sequentially. During fetching, we do not know if any or what hashes exist in a given range, so there is the potential that during a range request, a hash is returned that is higher than multiple range partitions that are queued for requesting. If this happens, the fetchers currently will still request ranges that are known to be empty after having received a hash value that is higher than the limits of queued ranges. This change will track the highest known key hash for the affected fetchers and skip requesting any queued ranges that have limit lower than the known highest hash.


scorbajio:
Optimizing the storage fetcher is more complicated since the response for `getStorageRanges` is a double array. Will handle that in a future PR.

g11tech (changes requested):
blocking till 
- https://github.com/ethereumjs/ethereumjs-monorepo/pull/2950 

is merged so that the PR can be merged with keyHashing enabled

g11tech:
will test this out today @scorbajio :+1:

packages/client/src/sync/fetcher/accountfetcher.ts
```diff
@@ -287,6 +290,11 @@ export class AccountFetcher extends Fetcher<JobTask, AccountData[], AccountData>
     const origin = this.getOrigin(job)
     const limit = this.getLimit(job)
 
+    if (this.highestKnownHash && compareBytes(limit, this.highestKnownHash) < 0) {
```
g11tech, line 293: ahh so this solves the problem that if some range is empty it won't be requested :+1:

g11tech:
from the code review it seems that we would need similar optimization in storage fetcher?

scorbajio:
> from the code review it seems that we would need similar optimization in storage fetcher?

Yes, but since the optimization for the storage fetcher is more complicated, I'm planning on doing it in a followup PR.

g11tech (approved):

PR #2946, opened 2023-08-08
Fix block validation order for EIP4844

jochem-brouwer:
The problem this fixes is that we first validate the blob transactions (this calls into fake_exponential) and then validate the excess blob gas. This is problematic, because if the excess blob gas check is incorrect it calls with very large numbers into fake_exponential, essentially an infinite loop.

This also optimizes the blob gas price calculation: this is constant, but we calculated it for each blob tx before.


acolytec3 (dismissed):
Nice fixes (especially the `blobGasPrice` one).

PR #2948, opened 2023-08-08
Client: If "dev" option is used, it should have a value and not be defaulted to pow

scorbajio:


holgerd77 (approved):
Guess we can still take this in, thanks, nice and clean addition (didn't know about this `check` possibility)! 👍

PR #2949, opened 2023-08-09
Trie proof API refactor

ScottyPoi:
This PR improves upon the proof methods in the `Trie` library.

The Trie methods  `fromProof` and `verifyProof` require a new Trie object to be instantiated by the user simply for the purpose of calling the methods.  This can be confusing, and seems unnecessarily indirect.

With static versions of the same methods, users can now accomplish these tasks in one step.

1. `Trie.fromProof`
- Works just like `trie.fromProof`, with an additional `opts` parameter to pass in keyHashing settings.

Now instead of:
```
const trie = new Trie(opts)
await trie.fromProof(proof)
```
We can just do:
```
const trie = await Trie.fromProof(proof, opts)
```

2. `Trie.verifyProof`
- Works just like `trie.verifyProof`, with an additional `opts` parameter to pass in keyHashing settings.
- Does not require the `rootHash` parameter as expected by class method.

Now instead of:
```
const trie = new Trie(opts)
const val = trie.verifyProof(trie.root(), key, proof, opts)
```
We can just do:
```
const val = Trie.verifyProof(key, proof, opts)
```


scorbajio:
Linking #2927

packages/trie/src/trie.ts
```diff
@@ -113,6 +113,50 @@ export class Trie {
     return new Trie(opts)
   }
 
+  static async fromProof(proof: Proof, opts?: TrieOpts): Promise<Trie> {
```
jochem-brouwer, line 116: What if I want to import two proofs? Or, I want to import a new proof in my trie later? This is now not possible anymore, right?

jochem-brouwer (commented):
I'm not sure if we want to make `fromProof` static

packages/trie/src/trie.ts
holgerd77: Ok, let's bring this back to the drawing board and first decide if this is a change we want respectively which makes sense and is thought through to the end.

holgerd77: Will give this a "blocked" label for now.

jochem-brouwer: (this change (i.e. make `fromProof` static) does not seem to be referred in the linked issue)

scorbajio: > What if I want to import two proofs? Or, I want to import a new proof in my trie later? This is now not possible anymore, right?

It seems like these tasks are made possible with the new `updateFromProof` function.

jochem-brouwer: Ah yes you are right, I somehow missed that.

holgerd77: Feel free to "unblock" this again or course of you guys think you've gotten answers for most of the questions here and things can be introduced in a non breaking way, you guys have a lot better overview than me here what's useful and makes sense regarding all this proof stuff!

packages/trie/src/trie.ts
```diff
@@ -770,22 +814,19 @@ export class Trie {
    * Saves the nodes from a proof into the trie.
    * @param proof
    */
-  async fromProof(proof: Proof): Promise<void> {
+
+  async updateFromProof(proof: Proof): Promise<void> {
+    if (!equalsBytes(this.root(), this.hash(proof.shift()!))) {
```
jochem-brouwer, line 819: I am not sure if I like this check, have to think about the practical downsides of this :thinking:

scorbajio (changes requested):

packages/trie/src/trie.ts
ScottyPoi: yes, `updateFromProof` is just the old `fromProof` with the word `update` tagged on to distinguish from the other.  To avoid actual breaking changes, we could instead use `createFromProof` for the static method, and leave `fromProof` as it was.

packages/trie/src/trie.ts
ScottyPoi: If you try to update a sparse trie with an invalid proof, this will throw.
shift() because the root node is already in the trie, so we'll never need to add it.

packages/trie/src/trie.ts
```diff
@@ -113,6 +113,50 @@ export class Trie {
...
+   * @param proof
+   * @param opts Trie options
+   * @throws If proof is found to be invalid.
+   * @returns The value from the key, or null if valid proof of non-existence.
+   */
+  static async verifyProof(
+    rootHash: Uint8Array,
+    key: Uint8Array,
+    proof: Proof,
+    opts?: TrieOpts
```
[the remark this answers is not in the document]
ScottyPoi: Three, actually, because the first node in the proof is also the root node.  I didn't write the method, I just made it static, and left it mostly the way it was.  The top `rootHash` parameter is definitely not needed, and the "opts" here really just needs to set the `keyHashing` behavior.

packages/statemanager/src/stateManager.ts
```diff
@@ -717,7 +719,10 @@ export class DefaultStateManager implements EVMStateManagerInterface {
       const storageProof = stProof.proof.map((value: PrefixedHexString) => hexToBytes(value))
       const storageValue = setLengthLeft(hexToBytes(stProof.value), 32)
       const storageKey = hexToBytes(stProof.key)
-      const proofValue = await this._proofTrie.verifyProof(storageRoot, storageKey, storageProof)
+      const proofValue = await Trie.verifyProof(storageRoot, storageKey, storageProof, {
+        useKeyHashing: true,
+        root: this._proofTrie.root(),
```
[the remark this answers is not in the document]
ScottyPoi: agreed

ScottyPoi:
Should be non-breaking now :egg:

packages/trie/src/trie.ts
ScottyPoi: :thinking:  I'd have to hear the upside to allowing a trie to update itself from a proof to a different trie!

packages/client/test/sim/beaconsync.md
jochem-brouwer, line 1: Random changes in this file?

packages/client/test/testdata/blocks/kaustinen2.json
jochem-brouwer, line 1: Also random changes here?

packages/statemanager/src/rpcStateManager.ts
```diff
@@ -9,7 +9,8 @@ import {
   toBytes,
 } from '@ethereumjs/util'
 import debugDefault from 'debug'
-import { keccak256 } from 'ethereum-cryptography/keccak.js'
+// import { ethers } from 'ethers'
+// import { keccak256 } from 'ethereum-cryptography/keccak.js'
```
jochem-brouwer, line 13: These comments should be deleted

packages/trie/src/trie.ts
jochem-brouwer: We now have https://github.com/ethereumjs/ethereumjs-monorepo/pull/3186. I think we want to make the methods static in some cases? (But this is breaking)

jochem-brouwer (commented):
I am not entirely sure about the state of this PR. We now have https://github.com/ethereumjs/ethereumjs-monorepo/pull/3186, it seems that if we want this in this seems to be merely marking some methods as `static`? (Note: we should make this non-breaking, so it should both be static and non-static). I will remove the `blocked` label until we resolve what we want with this PR. (I think the gist is to make the methods static which should have been static in the first place? :thinking: )

holgerd77:
> I will remove the blocked label until we resolve what we want with this PR

@jochem-brouwer I assume that you intended to write "I will not remove..." if not otherwise stated since that would make more sense 🙂 

@scorbajio don't know, are you getting along with the things we get from #3186 ? If so, @jochem-brouwer: would you then recommend to close this PR and rather not merge?

jochem-brouwer:
Yes indeed I did intend to write: I will not remove :)

I think that this PR can be updated with the changes of #3186, but I think we should update this PR to also include small changes such as also making `verifyProof` static (besides keeping it non-static, because otherwise it is breaking). So that would change this PR to a few lines of changes.

( @holgerd77 I think you intended to ping @ScottyPoi in your last post)

holgerd77:
@jochem-brouwer 👍

Ah, no, I indeed intended to mention @scorbajio since he requested this work to be updated and reviewed for some other follow up work.

packages/client/test/sim/beaconsync.md
scorbajio: Must have been from one of the merges. I fixed it so it's just the single extra space removed and any other unnecessary changes excluded.

packages/client/test/testdata/blocks/kaustinen2.json
scorbajio: Fixed.

packages/trie/src/trie.ts
scorbajio: Yes, the goal is to have a static version of `verifyProof` so that a new trie instantiation isn't needed for each proof verification, or even more awkward is the proofTrie pattern where we keep a trie around in both the statemanager and snap sync fetchers to perform proof verification.

The changes fro #3186 should be included, so we should have a static and non-static version of `verifyProof` and `fromProof`. This shouldn't be breaking, right? Since users can just use the previously available non-static versions if they chose to.

scorbajio:
@jochem-brouwer There is currently both a static and non-static version of both `verifyProof` and `fromProof` available, but it might not show up since it's been rebased on `master`, so it should no longer be breaking, right? I've updated the function docstrings to make this more visible.

holgerd77:
I've completely lost track here who wants what.  😋 

It would nevertheless be good if we bring this conversation and PR to a close.

Some things to keep in mind: we should both...

a) work towards a clean API and
b) remain backwards compatible and
c) see how we can fullfil the needs from Amir in some way (either he needs some changes or his code can be adjusted)

packages/trie/src/trie.ts
```diff
@@ -172,6 +172,54 @@ export class Trie {
...
+      return {
+        type: 'put',
+        key: Uint8Array.from(trie.hash(nodeValue)),
+        value: nodeValue,
+      } as PutBatch
+    })
+    await trie._db.batch(opStack)
+    trie.root(trie.hash(proof[0]))
+    await trie.persistRoot()
+    return trie
```
jochem-brouwer, line 195: This code is now redundant, can this call into `trie.fromProof`? The `if (opts?.root && !equalsBytes(opts.root, trie.hash(proof[0]))) {` check should stay though.

However, another question then here is, we have `createTrieFromProof` which almost does exactly the same, except that this `fromProof` has the extra check that the `trie.hash(proof[0])` matches the `opts.root` :thinking: (if it is provided :thinking: )

scorbajio (commented):
LGTM!

packages/trie/src/trie.ts
jochem-brouwer: I think this one needs to be checked @scorbajio

packages/trie/src/trie.ts
```diff
@@ -147,6 +147,40 @@ export class Trie {
     || ----------------`)
   }
 
+  static async createFromProof(proof: Proof, trieOpts?: TrieOpts) {
```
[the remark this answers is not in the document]
scorbajio: Added.

packages/client/src/sync/fetcher/accountfetcher.ts
```diff
@@ -318,11 +319,21 @@ export class AccountFetcher extends Fetcher<JobTask, AccountData[], AccountData>
...
     const values = accounts.map((acc: any) => accountBodyToRLP(acc.body))
     // convert the request to the right values
-    return trie.verifyRangeProof(stateRoot, origin, keys[keys.length - 1], keys, values, <any>proof)
+    return Trie.verifyRangeProof(
+      stateRoot,
+      origin,
+      keys[keys.length - 1],
+      keys,
+      values,
+      <any>proof,
```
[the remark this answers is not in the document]
scorbajio: Good point.

packages/trie/src/trie.ts
scorbajio: I've switched it over to `updateFromProof` to remove redundancies.

holgerd77:
Updated this via UI

I think we can take this in after having done a rough review? Will wait for a confirmation from @jochem-brouwer though (then optimally directly merge).

jochem-brouwer (approved):
Resolved merge conflicts and fixed the example. LGTM (assuming CI passes, will directly merge)

PR #2951, opened 2023-08-09
trie: improve util types and handling

gabrocheleau:
This PR makes a few minor improvements to the trie library util types.

- Improves the `isRawNode` util into a typeguard, which removes the need for typecasting in some instances.
- Slightly simplifies the logic of the `lookupNode` and `decodeNode` methods

### DIFF
--- a/packages/trie/src/trie.ts
+++ b/packages/trie/src/trie.ts
@@ -403,15 +403,16 @@ export class Trie {
     if (isRawNode(node)) {
       return decodeRawNode(node)
     }
-
-    const value = (await this._db.get(node)) ?? null
-
-    if (value === null) {
+    let value = null
+    let foundNode = null
+    value = await this._db.get(node)
+    if (value) {
+      foundNode = decodeNode(value)
+    } else {
       // Dev note: this error message text is used for error checking in `checkRoot`, `verifyProof`, and `findPath`
       throw new Error('Missing node in DB')
     }
-
-    return decodeNode(value) ?? null
+    return foundNode
   }
 
   /**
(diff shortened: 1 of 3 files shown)


packages/trie/src/trie.ts
```diff
@@ -401,18 +401,16 @@ export class Trie {
...
-    } else {
+    const value = (await this._db.get(node)) ?? null
+
+    if (value === null) {
       // Dev note: this error message text is used for error checking in `checkRoot`, `verifyProof`, and `findPath`
       throw new Error('Missing node in DB')
     }
-    return foundNode
+
+    return decodeNode(value) ?? null
```
acolytec3, line 413: Can `decodeNode` ever return undefined?  It's not typed for that.

acolytec3 (changes requested):
Looks good in general.  Left one comment that seems like extraneous code.

packages/trie/src/trie.ts
gabrocheleau: I wondered the same, and there seems to be some erroneous types in either this method or in instances where this method is used. I don't see how this could return undefined/null, but at the same time, in usage, we often check for the truthiness of the `lookupNode` return value, which leads me to think that it can sometimes return a nullish value. I'll investigate further and experiment with removing null completely (seems to me like it should either return a TrieNode or null)

gabrocheleau: I removed `null` from the return value, let's see how the CI runs.

acolytec3 (approved):
LGTM

Issue #2897, opened 2023-07-18
Replace `testdouble` with `vitest` test helpers


acolytec3:
Now that we have replaced `tape` with `vitest` across the monorepo, it would be worthwhile to replace all usage of `testdouble` with vitest equivalents. One notable usage that breaks with vitest is `td.replace` which just flat doesn't work with vitest (due to vitest being ESM based and `testdouble` being CJS based). See the charges in #2797 for some ideas of how this can happen.

PR #2953, opened 2023-08-09
Replace testdouble with vitest test helpers

scorbajio:
This PR partially addresses #2897 by refactoring most of the uses of `td.replace` and other testdouble mock functions with vitest equivalents.


packages/client/test/miner/pendingBlock.spec.ts
```diff
@@ -90,12 +89,15 @@ const setup = () => {
...
+  BlockHeader.prototype['_consensusFormatValidation'] = vi.fn()
+  vi.doMock('@ethereumjs/block', () => {
+    {
+      BlockHeader
+    }
+  })
 
-  const originalSetStateRoot = DefaultStateManager.prototype.setStateRoot
-  DefaultStateManager.prototype.setStateRoot = td.func<any>()
+  // const originalSetStateRoot = DefaultStateManager.prototype.setStateRoot
```
jochem-brouwer, line 99: Comment?

scorbajio: I must have missed that commented-out line. It's been removed.

holgerd77:
Hi Amir,
could you give a short in-between summary how this is going and what your experiences are so far? Is the vitest API for mocking nicer in any way? Or is both ok to use?

Also, just to note: we neither do need to necessarily complete on this, at least short-term nor do we need to do this in one go. So very well possible to also just do this in several steps are also only do to some certain level or something, whatever fits and makes sense and seems appropriate.

scorbajio:
> Hi Amir, could you give a short in-between summary how this is going and what your experiences are so far? Is the vitest API for mocking nicer in any way? Or is both ok to use?
> 
> Also, just to note: we neither do need to necessarily complete on this, at least short-term nor do we need to do this in one go. So very well possible to also just do this in several steps are also only do to some certain level or something, whatever fits and makes sense and seems appropriate.

It's been going good, not much trouble although I have run into a situation with the "should ban peer" test in `rlpxserver.spec.ts` that I was hoping @acolytec3 could take a look at, where the input to a mocked function is slightly different than expected even though while debugging I am able to confirm that the inputs are as expected immediately before the mocked function is called.

I'd say the vitest and testdouble API's are similar and can be interchangeable, but I do have a slight preference for vitest since I don't run into problems as often with e.g. matchers like I do with testdouble.

I'll go ahead and finalize this PR. My goal was to refactor tests that use `td.replace`. The only ones I didn't get to was `dns.spec.ts` and the `libp2p` `.old` files which I assume are going to get dropped before they need to be refactored.

holgerd77:
Thanks a lot, sounds great and super helpful! 🙏 🙂

acolytec3 (approved):
I think this looks good overall.  I don't love that Typescript now thinks the mocked `Peer` import is broken in `peerpool` but I'm not sure how best to fix it beyond adding `any` to lots of places.

holgerd77:
Cool! That’s really interesting work 👀, enjoyed going through here and let the new (and often really more intuitive) syntax sink in.

PR #2954, opened 2023-08-09
client: fix newPayloadV2 having PayloadV3 params

jochem-brouwer:
This fixes a mismatch in `newPayloadV2` of the Shanghai spec: https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#specification

It states:

Client software **MUST** return `-32602: Invalid params` error if the wrong version of the structure is used in the method call.

(structure = the payload).  

It fixes these 3 hive tests:

image

### DIFF
--- a/packages/client/src/rpc/modules/engine.ts
+++ b/packages/client/src/rpc/modules/engine.ts
@@ -806,6 +806,21 @@ export class Engine {
           message: 'ExecutionPayloadV2 MUST be used after Shanghai is activated',
         }
       }
+      const payloadAsV3 = params[0] as ExecutionPayloadV3
+      const { excessBlobGas, blobGasUsed } = payloadAsV3
+
+      if (excessBlobGas !== undefined && excessBlobGas !== null) {
+        throw {
+          code: INVALID_PARAMS,
+          message: 'Invalid PayloadV2: excessBlobGas is defined',
+        }
+      }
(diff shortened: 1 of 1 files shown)


packages/client/src/rpc/modules/engine.ts
```diff
@@ -806,6 +806,21 @@ export class Engine {
...
+      const payloadAsV3 = params[0] as ExecutionPayloadV3
+      const { excessBlobGas, blobGasUsed } = payloadAsV3
+
+      if (excessBlobGas !== undefined && excessBlobGas !== null) {
+        throw {
+          code: INVALID_PARAMS,
+          message: 'Invalid PayloadV2: excessBlobGas is defined',
+        }
+      }
+      if (blobGasUsed !== undefined && excessBlobGas !== null) {
```
spencer-tb, line 818:
```suggestion
      if (blobGasUsed !== undefined && blobGasUsed !== null) {
```
All the current pyspec tests pass with this change in in hive :D

spencer-tb (approved):

packages/client/src/rpc/modules/engine.ts
jochem-brouwer: Thanks for this change :smile: Have just committed!

acolytec3 (approved):
LGTM - verified that the specified hive tests now pass

PR #2957, opened 2023-08-14
fixed kademlia kbucket add

simone1999:
in the kbucket implementation the dontSplit attribute of KBucketNode is typed as bool, but was checked wit "!== undefined" which always was true. This caused KBucket.add to never split the leaf node and therefore kept the kbucket as a single level tree with a maximum of _numberOfNodesPerKBucket entrys in total.
tiny modification with huge impact :D

### DIFF
--- a/packages/devp2p/src/ext/kbucket.ts
+++ b/packages/devp2p/src/ext/kbucket.ts
@@ -149,7 +149,7 @@ export class KBucket {
     }
 
     // the bucket is full
-    if (node.dontSplit !== undefined) {
+    if (node.dontSplit) {
       // we are not allowed to split the bucket
       // we need to ping the first this._numberOfNodesToPing
       // in order to determine if they are alive


holgerd77:
Ugh. 🤯

Fantastic find! How did you discover that?

simone1999:
thx. playing around with the devp2p package and was quite confused to always see exactly 16 nodes in the kbucket. Was digging in the logic that sorts the leaves just to find out that the solution was much simpler haha

holgerd77:
OMG what a bug! 🤯 😆

Thanks a thousand times, great find!!! 🥳

This will likely be super-helpful for our client!

holgerd77:
Yes, can confirm, here two screenshots from Sepolia sync:

Before (staying on 16)
grafik

After:
grafik

Thanks a lot, will merge!

jochem-brouwer:
Thanks a lot! :smile: :+1:

PR #2958, opened 2023-08-14
DevP2P: ETH: only create DEBUG msgs if debugging

jochem-brouwer:
Currently, we always construct DEBUG messages on devp2p/ETH even if we are not debugging. This PR ensures these strings are only built when actually debugging. (Slight optimization)


holgerd77 (approved):
Nice, LGTM!

PR #2959, opened 2023-08-14
common: handle forkHash on timestamp == genesis timestamp

jochem-brouwer:
Fixes: 

`./hive --sim "ethereum/engine" --client ethereumjs --sim.parallelism 16 --sim.limit="engine-cancun/Request Blob Pooled Transactions"`

If fork is at timestamp genesisTimestamp then this fork is supposed to be active at timestamp = 0 and not participate in the forkHash calculation of the devp2p/ETH STATUS message.

See also:

https://github.com/hyperledger/besu/issues/5744
https://github.com/ethereum/go-ethereum/pull/27895
https://eips.ethereum.org/EIPS/eip-6122

### DIFF
--- a/packages/common/src/utils.ts
+++ b/packages/common/src/utils.ts
@@ -153,7 +153,7 @@ function parseGethParams(json: any, mergeForkIdPostMerge: boolean = true) {
     .filter((fork) => fork.block !== null || fork.timestamp !== undefined) as ConfigHardfork[]
 
   for (const hf of params.hardforks) {
-    if (hf.timestamp !== undefined && hf.timestamp === genesisTimestamp) {
+    if (hf.timestamp === genesisTimestamp) {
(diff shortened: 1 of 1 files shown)


g11tech (approved):
lgtm

holgerd77:
Great find! This is a relatively sensitive change and can have side effects on other girl hash calculations, I would feel better with 1-2 dedicated test cases for it.

jochem-brouwer:
Yes, you are right, especially these "small" changes can be tested quickly with small tests. I will add tests. 

For a general comment though, some hive tests are very complex and would take long to integrate - what we could do (mid-term) is to create a hive CI runner and then run these specific tests to cover these test cases (and for a bonus: also include this into codecov, but that would be complex since we need to hook into hive's docker image of our client)

PR #2960, opened 2023-08-14
Update statemanager tests

scorbajio:


jochem-brouwer (approved):
LGTM

holgerd77:
One or two sentences on the PR itself what is actually being done and why would be good, so that one does not need to search for this information implicitly in the charts or wherever.

PR #2965, opened 2023-08-15
Snap Sync: Storage Fetcher Highest-Known-Hash Optimization

scorbajio:
This change implements the highest-known-hash optimization discussed in #2941 for the storage fetcher.


packages/client/test/miner/miner.spec.ts
```diff
@@ -703,5 +703,5 @@ describe('[Miner]', async () => {
     })
     await (miner as any).queueNextAssembly(0)
     await wait(10000)
-  }, 80000)
+  }, 100000)
```
g11tech, line 706: i don't think this helps some times its just stuck, will undo this change

g11tech (approved):

PR #2966, opened 2023-08-16
Trie: decouple walk test from demo

### LINKED ISSUE #2964: Trie: Verbose WalkController Test Output
Since the introduction of the new walk controller in #2904 there is now a lot of verbose test output in the trie library which shouldn't be there:

grafik

This is due to the usage of `console.log()` in the `logNode()` method in `asyncWalk.spec.ts`. While the output is pretty nice we should replace this with something which is not "always on". I am not sure, is there something natively in Vitest? 🤔 Did a quick search but did not find anything immediately.

Not sure if our DEBUG logger might be something for this, a bit sceptical though if this makes sense, so just a first-round alternative idea.

ScottyPoi:
Fixes #2964

The test written for the new async walk iterator also contained a demo which logged to console during testing.

This PR decouples the test from the demo.  
The test should run without display the extra logging.  

TODO: Reintroduce demo as an example script.

### DIFF
--- a/packages/trie/test/util/asyncWalk.spec.ts
+++ b/packages/trie/test/util/asyncWalk.spec.ts
@@ -1,61 +1,10 @@
-/* eslint-disable no-console */
 import { bytesToHex, equalsBytes, hexToBytes, utf8ToBytes } from '@ethereumjs/util'
 import { assert, describe, it } from 'vitest'
 
-import { BranchNode, ExtensionNode, LeafNode, Trie } from '../../src/index.js'
+import { LeafNode, Trie } from '../../src/index.js'
 import { _walkTrie } from '../../src/util/asyncWalk.js'
 import { bytesToNibbles } from '../../src/util/nibbles.js'
 import trieTests from '../fixtures/trietest.json'
-
-import type { TrieNode } from '../../src/index.js'
-
-function getNodeType(node: TrieNode): string {
-  if (node instanceof BranchNode) {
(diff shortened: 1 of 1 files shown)


holgerd77:
Ok, thanks for addressing! I wouldn't want to merge yet in the current state just removing the logic, would be too much a pity and chances would be too high that this gets forgotten over time I guess. So would be nice if can transform this a bit and move this as an example into the `examples` directory.

Side note: maybe do not put into the `merkle_patricia_trees` subfolder - that's a somewhat specific tutorial from @gabrocheleau (Gabriel: and the examples are also outdated now, maybe a good side task?) on MPTs. So rather put directly in `examples/walkController.cts`.

Another side note: I think it has some specific reasons that *these* examples are JavaScript, normally (and: better) to have TypeScript examples, see `evm/examples` for a better "example for an example" 😋.

Examples are also kept still in CommonJS for now and get the `.cjs` ending. Easiest to stick to that for now, I think this would have side effects if changed and generally rather to be addressed separately if tackled.

holgerd77:
Did edits on the previous comment, please read on site.

Also short info: there is a dedicated `examples` script to run existing/new examples in each `package.json` file from a package.

acolytec3 (commented):
Rebased and added some tricks to the example to turn on the logs only while running this script.

acolytec3 (approved):
Looks great!

PR #2967, opened 2023-08-17
client: include parent beacon block root for proposal payload uniquness

g11tech:
Lighthouse seems to have an optimization where they send a fcU before updating head as well as after updating head. this seems to result into two calls with different parentBeaconBlockRoot (which is the CL head on which proposal is being build) 

Since we have not been using parent beacon block root for uniqueness of payload being build, we were returning same payload if to second fcU call (which has an updated parent beacon block root). 

This was leading to invalid evaluation of the block hash later in new payload where the correct parent beacon block root supplied by the CL is injected into payload

This PR attempts to fix the same

- [x] TODO : test on the dencun network with lighthouse: successful block production
  image

### DIFF
--- a/packages/client/src/miner/pendingBlock.ts
+++ b/packages/client/src/miner/pendingBlock.ts
@@ -97,7 +97,7 @@ export class PendingBlock {
     withdrawals?: WithdrawalData[]
   ) {
     const number = parentBlock.header.number + BigInt(1)
-    const { timestamp, mixHash } = headerData
+    const { timestamp, mixHash, parentBeaconBlockRoot } = headerData
     const { gasLimit } = parentBlock.header
 
     // payload is uniquely defined by timestamp, parent and mixHash, gasLimit can also be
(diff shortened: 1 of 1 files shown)


holgerd77:
Can't completely judge this change, so just to make sure: the last time I checked the [fork monitor from devnet 8](https://forkmon.dencun-devnet-8.ethpandaops.io/) lighthouse was failing for all three EL clients (now everything seems to be in the green again?).

So I was just wondering if it makes sense to adjust to the lighthouse behavior here, or if this would rather be a fix on the lighthouse side? Or is this fix here so generic that we can take anyhow?

g11tech:
> Can't completely judge this change, so just to make sure: the last time I checked the [fork monitor from devnet 8](https://forkmon.dencun-devnet-8.ethpandaops.io/) lighthouse was failing for all three EL clients (now everything seems to be in the green again?).
> 
> So I was just wondering if it makes sense to adjust to the lighthouse behavior here, or if this would rather be a fix on the lighthouse side? Or is this fix here so generic that we can take anyhow?

generic fix, we should have had parentBeaconBlockRoot in payload uniqueness as CL can freely send a new FcU if its head changes at any moment for the proposal slit

holgerd77 (approved):
Ok, thanks for the clarification, LGTM

PR #2973, opened 2023-08-19
client: fixes for new engine api method validations for hive pr-834

g11tech:
addresses the new validations added in 
 - https://github.com/ethereum/hive/pull/834

passes all `engine-cancun` tests:

image


packages/client/src/rpc/modules/engine.ts
```diff
@@ -1090,6 +1126,23 @@ export class Engine {
...
+    if (payloadAttributes !== undefined && payloadAttributes !== null) {
+      if (Object.keys(payloadAttributes).length > 3) {
+        throw {
+          code: INVALID_PARAMS,
+          message: 'PayloadAttributesV1 MUST be used for forkchoiceUpdatedV2',
+        }
+      }
+      validateHardforkRange(
+        this.chain.config.chainCommon,
+        1,
```
jochem-brouwer, line 1139: I think it is somewhat implicit, but this check for `forkchoiceUpdatedV1` does not seem to be in the engine api specs. It states that it must be updated for `engine_forkchoiceUpdatedV2` (so if you call it >= Cancun it throws Unsupported fork) but by spec currently it does not list `engine_forkchiceUpdatedV1`, see https://github.com/ethereum/execution-apis/blob/main/src/engine/cancun.md#update-the-methods-of-previous-forks

packages/client/src/rpc/modules/engine.ts
```diff
@@ -1178,22 +1259,40 @@ export class Engine {
...
+        case 1:
+          checkNotBeforeHf = null
+          checkNotAfterHf = Hardfork.Paris
+          break
+
+        default:
+          throw Error(`Invalid payloadVersion=${payloadVersion}`)
       }
 
+      validateHardforkRange(
```
jochem-brouwer, line 1286: Same here with `engine_getPayloadV1`

packages/client/src/rpc/modules/engine.ts
```diff
@@ -1202,12 +1301,12 @@ export class Engine {
   }
 
   async getPayloadV1(params: [Bytes8]) {
-    const { executionPayload } = await this.getPayload(params)
+    const { executionPayload } = await this.getPayload(params, 1)
```
jochem-brouwer, line 1304: Same comment here

jochem-brouwer (commented):
3 comments in total but they have a question about current engine-API spec.

packages/client/src/rpc/modules/engine.ts
g11tech: hmmm yes, should we PR to update specs because obviously lower version can't be used in higher version fork because of missing fields

jochem-brouwer: I think there are some small-ish problems wiht the current spec and those changes:

1. We have 3 files, Paris / Shanghai / Cancun. In Cancun we introduce new rules which override the rules of the other files (this becomes clear once you read all files). The rule which it overrides if for example this Shanghai rule: `Client software MUST return -32602: Invalid params error if the wrong version of the structure is used in the method call.` (of `engine_newPayloadV2`)
2. And building upon the previous: in Cancun we should use `INVALID_PARAMS` if we call `engine_newPayloadV1` (if we are on Shanghai), but this thus implies we are aware (and ready) for Cancun. If we don't know about Cancun and we want to change the spec, then now the behavior changes: if we are on Shanghai, and we call `engine_newPayloadV1`, then we should return `UNSUPPORTED_FORK`. 

So if we change the spec this starts to get somewhat confusing. I do think we should change the spec though because it is more logical. I am also not aware what happens if we change the behavior, are there CLs relying on exact error codes to change behavior, or do they treat each error as an error in general?

g11tech: may be we can address those issues separately, for now we can may be merge this as it passes cancun fully

holgerd77:
Updated this via UI

holgerd77:
Updated this via UI

holgerd77 (approved):
LGTM

PR #2974, opened 2023-08-19
client: grab latest 2 blocks on vm performance sync [no ci]

jochem-brouwer:
Do not merge this PR! This is just used to track some progress here.

The goal here is to setup client in such way that we setup the client DB such that we can run a recent block in the exact same way what would happen if we had all of the state (for benchmarking purposes).


jochem-brouwer:
Will close this PR for now, EVM profiling work using custom tests is the way to optimize it at this point.

PR #2975, opened 2023-08-19
vm: add performance test script using EthersStateManager

jochem-brouwer:
This PR is not intended to be merged since it hacks into some packages to get around not having to download all blocks in order to figure out what Common should be used.

To use it, the main script is `./packages/vm/test/runPerformance.ts`. Setup the right provider, blockTag, and common in this script. Ensure that a node is running which has access to the requested storage / account / code items for the blockTag. Note: blockTag should be set to the block you want to run minus one.

This script first setups the local caches of state manager and then re-runs the block to check the run time. I noticed that the reported block gas is not correct, but this is somewhat usable to start benchmarking EVM against realistic blocks.

Note: this current setup ONLY tests EVM. It does not test using a disk DB (all is read from memory) and it also does not use a Trie, so lookup items in memory are a single operation while trie lookups on disk (or in memory as well) usually take multiple operations to get the actual value.

### DIFF
--- a/packages/blockchain/src/blockchain.ts
+++ b/packages/blockchain/src/blockchain.ts
@@ -490,17 +490,17 @@ export class Blockchain implements BlockchainInterface {
         // save header/block to the database
         dbOps = dbOps.concat(DBSetBlockOrHeader(block))
 
-        let commonAncestor: undefined | BlockHeader
-        let ancestorHeaders: undefined | BlockHeader[]
+        //let commonAncestor: undefined | BlockHeader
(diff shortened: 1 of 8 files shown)


jochem-brouwer:
Sepolia block 4094287:

```
Setup took 13.31
Gas used: 28738369 expected gas used: 29850949
Running from cache took 2.775
Gas used: 28738369 expected gas used: 29850949
```

jochem-brouwer:
Will close this PR for now, EVM profiling work using custom tests is the way to optimize it at this point.

PR #2977, opened 2023-08-21
Use trie verifyRangeProof for no-proof range verification

scorbajio:
Currently, the storage fetcher puts received data into a trie and checks if the root is as expected in order to perform a no-proof range verification. This change moves over to using the `trie.verifyRangeProof` method for no-proof range verification.


holgerd77:
Could you expand a bit here for the unknown what no-proof range verification actually is?

scorbajio:
> Could you expand a bit here for the unknown what no-proof range verification actually is?

Sure! No-proof range verification is a type of "proof" verification where you don't actually need a proof because the full set of leaves are present. This means you can put all of the data in an MPT and check if the root matches the expected state root. I've switched over from using an extra helper function that was building the MPT and checking the root over to using `verifyRangeProof`, which can handle no-proof verification if the first and last key inputs as well as the proof input are all passed in as `null`.

scorbajio:
There are errors with using the `verifyRangeProof` function to perform no-proof verification because the trie used for no-proof verification uses key hashing, and so results in a root mismatch because the input keys are already hashed (and so are being double-hashed). This is a +1 for thinking of a design for `verifyRangeProof` that is static and general enough to allow for this use-case, but for now it seems more fitting to use a helper function that performs the check inside of the fetcher.

holgerd77:
Can't you just add this new `skipKeyTransform` parameter to the `verifyRangeProof()` function and then pass this through to the `trie.put()` calls? 🤔

In case that makes sense and we want to do: I am correct in the assumption that `verifyRangeProof()` has not yet been released, is this correct (PR merge was pretty recent, right? So after breaking releases?). If so: let's please switch to put the `useKeyHashingFunction` parameter + the then added `skipKeyTransform` parameter into an options dictionary with proper typing, parameters there should all be optional and default to something otherwise.

Then we are forward compatible and do can add eventual additional options easily and also use them separately and do not have to chain and chain more and more as single parameters to the method.

holgerd77:
(side note: I am not so deep in the topic though, maybe my suggestion doesn't make sense in the first place)

scorbajio:
I actually took another look at the all-elements proof implementation, and it seems like the problem was my understanding of the return type. I thought the returned boolean indicates whether the proof and data match and are valid, whereas it actually indicates whether or not there are more elements remaining in the trie to the right of the ending key provided as a part of the proof check. After updating the fetcher logic to accommodate this, all that is left is to fix a failing test that fails for a separate reason, and then this change should be good to review and merge.

g11tech:
awesome @scorbajio will review shortly

holgerd77:
Updated this via UI

g11tech:
> awesome @scorbajio will review shortly

started testing and reviewing this :+1:

packages/client/src/sync/fetcher/storagefetcher.ts
```diff
@@ -119,20 +121,9 @@ export class StorageFetcher extends Fetcher<JobTask, StorageData[][], StorageDat
...
-        // ensure the range is monotonically increasing
-        if (bytesToBigInt(slots[i].hash) > bytesToBigInt(slots[i + 1].hash)) {
-          throw Error(
-            `Storage hashes not monotonically increasing: ${i} ${slots[i].hash} vs ${i + 1} ${
-              slots[i + 1].hash
-            }`
-          )
-        }
-      }
-      const trie = new Trie()
```
g11tech, line 132: i think just keep using the `_proofTrie` will cause memory blat as its db will keep getting build as more and more elemts get put to trie during verification

g11tech (changes requested):
need to check if recurring use of proofTrie to verify range will cause memory bloat or not

packages/client/src/sync/fetcher/storagefetcher.ts
scorbajio: Looking at the code for the `verifyProof` and `verifyRangeProof` functions, it seems like they create a new, disposable trie when they need to for checking proofs, so there shouldn't be bloat. Also see #2932 where the dedicated `_proofTrie` pattern was introduced in statemanager.

g11tech (approved):

PR #2980, opened 2023-08-22
Flatten DB

scorbajio:
The aim of this change is to implement a flat database design for keeping state data in. This work will be integrated with the snap sync implementation for keeping snapshots of the latest 128 states. This work picks up where #780 left off. The goal is to first implement the `StackTrie` structure, then to implement the necessary `Snapshot` and database wrapper necessary for data persistence.

### DIFF
--- a/packages/vm/src/flat/stackTrie.ts
+++ b/packages/vm/src/flat/stackTrie.ts
@@ -1,0 +1,256 @@
+import { RLP } from '@ethereumjs/rlp'
+import { matchingNibbleLength } from '@ethereumjs/trie'
+import { KECCAK256_RLP, addHexPrefix, bytesToHex } from '@ethereumjs/util'
+import * as assert from 'assert'
+import { keccak256 } from 'ethereum-cryptography/keccak'
+
+import type { Nibbles } from '@ethereumjs/trie'
+
+// can potentially keep the stack trie in the trie package since it uses trie package types like nibbles and helpers,
+// and import it in vm flat directory for integration with snapshot and statemanager
+
+// TODO have to rename this function or the other since there are two nibblesToBytes being exported from trie package
+function nibblestoBytes(arr: Nibbles): Uint8Array {
+  const buf = new Uint8Array(arr.length / 2)
+  for (let i = 0; i < buf.length; i++) {
+    let q = i * 2
+    buf[i] = (arr[q] << 4) + arr[++q]
+  }
+  return buf
+}
+
+// TODO have to rename this function or the other since there are two bytesToNibbles being exported from trie package
+function bytesToNibbles(key: Uint8Array): Nibbles {
+  const bkey = toBytes(key)
+  const nibbles = [] as Nibbles
+
+  for (let i = 0; i < bkey.length; i++) {
+    let q = i * 2
+    nibbles[q] = bkey[i] >> 4
+    ++q
+    nibbles[q] = bkey[i] % 16
+  }
+
+  return nibbles
+}
+
+///
+
+/**
+ * Given a list of leaves with keys sorted in the ascending order,
+ * re-creates the trie structure hashing up simultaneously the
+ * branches that won't be visited again.
+ * Implementation based on:
+ * https://github.com/holiman/go-ethereum/blob/b7e737bdd87ba51631837f337d128282ff066d24/trie/stacktrie.go
+ */
+export function merkleizeList(leaves: Uint8Array[][]): Uint8Array {
+  let root = new EmptyNode()
+
+  for (const kv of leaves) {
+    const key = bytesToNibbles(kv[0])
+    root = root.insert(key, kv[1])
+  }
+
+  return root.hash()
+}
+
+abstract class BaseNode {
+  abstract insert(key: Nibbles, value: Uint8Array): BaseNode
+  abstract raw(): any
+  serialize(): Uint8Array {
+    return RLP.encode(this.raw())
+  }
+  hash(): Uint8Array {
+    return keccak256(this.serialize())
+  }
+}
+
+export class EmptyNode extends BaseNode {
+  insert(key: Nibbles, value: Uint8Array): BaseNode {
+    return new LeafNode(key, value)
+  }
+
+  raw(): any {
+    return null
+  }
+
+  serialize(): Uint8Array {
+    return new Uint8Array()
+  }
+
+  hash(): Uint8Array {
+    return KECCAK256_RLP
+  }
+}
+
+export class LeafNode extends BaseNode {
+  _key: Nibbles
+  _value: Uint8Array
+
+  constructor(key: Nibbles, value: Uint8Array) {
+    super()
+    this._key = key
+    this._value = value
+  }
+
+  insert(key: Nibbles, value: Uint8Array): BaseNode {
+    const commonLen = matchingNibbleLength(this._key, key)
+    assert(commonLen < this._key.length, 'replacing leaf is not supported')
+
+    const newLeaf = new LeafNode(key.slice(commonLen + 1), value)
+    const branch = new BranchNode()
+    branch.setChild(key[commonLen], newLeaf)
+    branch.setChild(this._key[commonLen], this)
+
+    let root: BaseNode = branch
+    // Need extension node for common path
+    if (commonLen > 0) {
+      root = new ExtensionNode(this._key.slice(0, commonLen), branch)
+    }
+
+    // Slice common part from current leaf's key
+    this._key = this._key.slice(commonLen + 1)
+
+    return root
+  }
+
+  raw(): [Uint8Array, Uint8Array] {
+    const encodedKey = addHexPrefix(this._key.slice(0), true) // the true is for adding terminator
+    return [nibblesToBuffer(encodedKey), this._value]
+  }
+}
+
+export class ExtensionNode extends BaseNode {
+  _key: Nibbles
+  _child: BaseNode
+
+  constructor(key: Nibbles, child: BaseNode) {
+    super()
+    this._key = key
+    this._child = child
+  }
+
+  insert(key: Nibbles, value: Uint8Array): BaseNode {
+    const commonLen = matchingNibbleLength(this._key, key)
+    // If matching prefix, recurse into child node
+    if (this._key.length === commonLen) {
+      this._child = this._child.insert(key.slice(commonLen), value)
+      return this
+    }
+
+    // Otherwise we'll need a new leaf, a branch and possibly an extension
+    const newLeaf = new LeafNode(key.slice(commonLen + 1), value)
+    const branch = new BranchNode()
+    branch.setChild(key[commonLen], newLeaf)
+
+    if (commonLen < this._key.length - 1) {
+      branch.setChild(this._key[commonLen], this)
+      this._key = this._key.slice(commonLen + 1)
+    } else {
+      branch.setChild(this._key[commonLen], this._child)
+    }
+
+    let root: BaseNode = branch
+    // Need extension for common prefix
+    if (commonLen > 0) {
+      root = new ExtensionNode(key.slice(0, commonLen), branch)
+    }
+
+    return root
+  }
+
+  raw(): any {
+    const childRaw = this._child.raw()
+    const childSerialized = RLP.encode(childRaw)
+    const value = childSerialized.length < 32 ? childRaw : keccak256(childSerialized)
+    const encodedKey = addHexPrefix(this._key.slice(0), false)
+    return [nibblesToBuffer(encodedKey), value]
+  }
+}
+
+export class BranchNode extends BaseNode {
+  _children: (BaseNode | null)[]
+  _value: Uint8Array | null
+
+  constructor() {
+    super()
+    this._children = new Array(16)
+    for (let i = 0; i < 16; i++) this._children[i] = null
+    this._value = null
+  }
+
+  setChild(i: number, child: BaseNode) {
+    this._children[i] = child
+  }
+
+  insert(key: Nibbles, value: Uint8Array): BaseNode {
+    const i = key[0]
+    if (!this._children[i]) {
+      this._children[i] = new EmptyNode()
+    }
+    // Hash previously explored branches. We can do this because we assume
+    // leaves are being inserted in the ascending key order.
+    for (let j = 0; j < i; j++) {
+      if (!this._children[j] || this._children[j] instanceof HashNode) continue
+      this._children[j] = new HashNode(this._children[j]!)
+    }
+    this._children[i] = this._children[i]!.insert(key.slice(1), value)
+    return this
+  }
+
+  raw(): any {
+    const raw = []
+    for (let i = 0; i < 16; i++) {
+      const child = this._children[i]
+      if (!child) {
+        raw.push(null)
+      } else if (child instanceof HashNode && (<HashNode>child).embedded) {
+        raw.push(child.raw())
+      } else {
+        const childRaw = child.raw()
+        const childSerialized = RLP.encode(childRaw)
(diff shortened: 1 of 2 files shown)


holgerd77:
Side note: VM is definitely the wrong package, that's clear I guess? Just to point out, so that this doesn't "settle at the wrong place" 😋, I know though that this is very very early.

holgerd77:
Cool work! 🤩

Is there any way this can be practically tested?

scorbajio:
> Cool work! 🤩
> 
> Is there any way this can be practically tested?

I will update with some details on this as I am currently still working on debugging and integrating it with the vm blockchain and state tests with the goal of making it possible to use for e.g. block execution 🙂

packages/util/package.json
```diff
@@ -90,10 +90,11 @@
     "tsc": "../../config/cli/ts-compile.sh"
   },
   "dependencies": {
-    "@ethereumjs/rlp": "^5.0.1",
-    "ethereum-cryptography": "^2.1.2"
+    "@ethereumjs/rlp": "^5.0.0",
+    "abstract-level": "^1.0.3",
```
acolytec3, line 94: Can we make both of the level deps optional peer dependencies like kzg instead of required?  Otherwise, we're shipping these deps to essentially everyone who uses any component of our stack?

packages/trie/src/db/flat/stackTrie.ts
```diff
@@ -0,0 +1,251 @@
...
+    // If matching prefix, recurse into child node
+    if (this._key.length === commonLen) {
+      this._child = this._child.insert(key.slice(commonLen), value)
+      return this
+    }
+
+    // Otherwise we'll need a new leaf, a branch and possibly an extension
+    const newLeaf = new STLeafNode(key.slice(commonLen + 1), value)
+
+    //eslint-disable-next-line
```
acolytec3, line 120: Is it not possible to just reorder the declarations so we don't need to have these eslint comments?

packages/trie/src/db/flat/stackTrie.ts
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  serialize(): Uint8Array {
+    throw new Error('Cant serialize hashnode')
+  }
+
+  hash(): Uint8Array {
+    return keccak256(this._serialized)
+  }
+}
```
acolytec3, line 181:
```suggestion
    return this._hash
```

packages/trie/src/db/flat/stackTrie.ts
```diff
@@ -0,0 +1,251 @@
...
+//       to save on iterating already visited sub-tries.
+export class HashNode extends BaseNode {
+  _ref: BaseNode
+  _serialized: Uint8Array
+
+  constructor(ref: BaseNode) {
+    super()
+    this._ref = ref
+    this._serialized = ref.serialize()
+  }
```
acolytec3, line 159:
```suggestion
    this._hash = keccak256(this._serialized)
```

packages/trie/src/db/flat/stackTrie.ts
```diff
@@ -0,0 +1,251 @@
...
+    const childRaw = this._child.raw()
+    const childSerialized = RLP.encode(childRaw)
+    const value = childSerialized.length < 32 ? childRaw : keccak256(childSerialized)
+    const encodedKey = addHexPrefix(this._key.slice(0), false)
+    return [nibbleTypeToPackedBytes(encodedKey), value]
+  }
+}
+
+// TODO: Either remove HashNode or precompute hash
+//       to save on iterating already visited sub-tries.
```
acolytec3, line 150: Added a suggested way to do this (just hash when constructing the node and return that hashed value)

packages/trie/src/db/flat/stackTrie.ts
```diff
@@ -0,0 +1,251 @@
...
+    super()
+    this._ref = ref
+    this._serialized = ref.serialize()
+  }
+
+  get embedded(): boolean {
+    return this._serialized.length < 32
+  }
+
+  // eslint-disable-next-line
```
acolytec3, line 165: Can be removed
```suggestion
```

packages/trie/src/db/flat/stackTrie.ts
```diff
@@ -0,0 +1,251 @@
...
+    // leaves are being inserted in the ascending key order.
+    for (let j = 0; j < i; j++) {
+      if (!this._children[j] || this._children[j] instanceof HashNode) continue
+      this._children[j] = new HashNode(this._children[j]!)
+    }
+    this._children[i] = this._children[i]!.insert(key.slice(1), value)
+    return this
+  }
+
+  raw(): any {
```
acolytec3, line 212:
```suggestion
  raw(): Uint8Array[] {
```

packages/trie/src/db/flat/stackTrie.ts
```diff
@@ -0,0 +1,251 @@
...
+    let root: BaseNode = branch
+    // Need extension for common prefix
+    if (commonLen > 0) {
+      root = new STExtensionNode(key.slice(0, commonLen), branch)
+    }
+
+    return root
+  }
+
+  raw(): any {
```
acolytec3, line 140:
```suggestion
  raw(): [Uint8Array,Uint8Array] {
```

packages/trie/src/db/flat/stackTrie.ts
```diff
@@ -0,0 +1,251 @@
...
+  get embedded(): boolean {
+    return this._serialized.length < 32
+  }
+
+  // eslint-disable-next-line
+  insert(key: Nibbles, value: Uint8Array): BaseNode {
+    throw new Error("Can't insert into hash node")
+  }
+
+  raw(): any {
```
acolytec3, line 170:
```suggestion
  raw(): Uint8Array[] {
```

packages/statemanager/benchmarks/index.ts
```diff
@@ -0,0 +1,54 @@
+import { Account, Address, MapDB } from '@ethereumjs/util'
+// @ts-ignore - package has no types...
+import { run, mark, logMem } from 'micro-bmark'
+import { DefaultStateManager, FlatStateManager } from '@ethereumjs/statemanager'
```
acolytec3, line 4: Just a suggestion but we already have `benchmark` used in a couple of places in the monorepo so not sure if it makes sense to use a different benchmarking tool here.  No strong opinion but just a thought (especially since I think the APIs are close to identical)

packages/statemanager/src/flat/snapshot.ts
```diff
@@ -0,0 +1,277 @@
...
+  const concatenatedArray = new Uint8Array(l)
+
+  let offset = 0
+  for (const array of arrays) {
+    concatenatedArray.set(array, offset)
+    offset += array.length
+  }
+
+  return concatenatedArray
+}
```
acolytec3, line 40: I believe this functionality is already covered by `concatBytes` from `ethereum-cryptography/utils`

packages/statemanager/src/flat/snapshot.ts
```diff
@@ -0,0 +1,277 @@
...
+   */
+  async clearAccount(address: Address): Promise<void> {
+    const rawAccount = await this.getAccount(address)
+    if (rawAccount === undefined) return
+
+    await this.delAccount(address)
+
+    const account = Account.fromRlpSerializedAccount(rawAccount)
+    if (!equalsBytes(account.codeHash, KECCAK256_NULL)) {
+      // await this.delCode(account.codeHash)
```
acolytec3, line 112: Why is this commented out?  Are we optimistically thinking other accounts might have the same code deposited?

packages/statemanager/src/flat/snapshot.ts
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  async putStorageSlot(address: Address, slot: Uint8Array, value: Uint8Array): Promise<void> {
+    const key = concatenateUint8Arrays([STORAGE_PREFIX, keccak256(address.bytes), keccak256(slot)])
+    await this._saveCachePreState(key)
+    await this._db.put(key, value)
+  }
+
+  async getStorageSlot(address: Address, slot: Uint8Array): Promise<Uint8Array | undefined> {
+    const key = concatenateUint8Arrays([STORAGE_PREFIX, keccak256(address.bytes), keccak256(slot)])
```
acolytec3, line 125:
```suggestion
    const key = concatBytes(STORAGE_PREFIX, keccak256(address.bytes), keccak256(slot))
```

packages/statemanager/src/flat/snapshot.ts
```diff
@@ -0,0 +1,277 @@
...
+    return this._db.get(key)
+  }
+
+  async getStorageSlots(address: Address): Promise<[Uint8Array, Uint8Array | undefined][]> {
+    const prefix = concatenateUint8Arrays([STORAGE_PREFIX, keccak256(address.bytes)])
```
acolytec3, line 130:
```suggestion
    const prefix = concatBytes(STORAGE_PREFIX, keccak256(address.bytes))
```

packages/statemanager/src/flat/snapshot.ts
```diff
@@ -0,0 +1,277 @@
...
+
+    return this._db.byPrefix(prefix, { keyEncoding: KeyEncoding.Bytes })
+  }
+
+  async delStorageSlot(address: Address, slot: Uint8Array): Promise<void> {
+    const key = concatenateUint8Arrays([STORAGE_PREFIX, keccak256(address.bytes), keccak256(slot)])
```
acolytec3, line 136:
```suggestion
    const key = concatBytes(STORAGE_PREFIX, keccak256(address.bytes), keccak256(slot))
```

packages/statemanager/src/flat/snapshot.ts
```diff
@@ -0,0 +1,277 @@
...
+    await this._saveCachePreState(key)
+    await this._db.del(key)
+  }
+
+  async clearAccountStorage(address: Address): Promise<void> {
+    const prefix = concatenateUint8Arrays([STORAGE_PREFIX, keccak256(address.bytes)])
```
acolytec3, line 142:
```suggestion
    const prefix = concatBytes(STORAGE_PREFIX, keccak256(address.bytes))
```

packages/statemanager/src/flat/snapshot.ts
```diff
@@ -0,0 +1,277 @@
...
+    const keys = await this._db.keysByPrefix(prefix, { keyEncoding: KeyEncoding.Bytes })
+    for (const key of keys) await this._saveCachePreState(key)
+    await this._db.delByPrefix(prefix, { keyEncoding: KeyEncoding.Bytes })
+  }
+
+  async putCode(address: Address, code: Uint8Array): Promise<void> {
+    const key = concatenateUint8Arrays([CODE_PREFIX, keccak256(address.bytes)])
```
acolytec3, line 149:
```suggestion
    const key = concatBytes(CODE_PREFIX, keccak256(address.bytes))
```

packages/statemanager/src/flat/snapshot.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  async getCode(address: Address): Promise<Uint8Array | undefined> {
+    const rawAccount = await this.getAccount(address)
+    if (!rawAccount) return undefined
+    const account = Account.fromRlpSerializedAccount(rawAccount)
+    if (equalsBytes(account.codeHash, KECCAK256_NULL)) {
+      return new Uint8Array(0)
+    }
+
+    const key = concatenateUint8Arrays([CODE_PREFIX, keccak256(address.bytes)])
```
acolytec3, line 174:
```suggestion
    const key = concatBytes(CODE_PREFIX, keccak256(address.bytes))
```

packages/statemanager/src/flat/flatStateManager.ts
```diff
@@ -0,0 +1,422 @@
...
+
+/**
+ * Default StateManager implementation for the VM.
+ *
+ * The state manager abstracts from the underlying data store
+ * by providing higher level access to accounts, contract code
+ * and storage slots.
+ *
+ * The default state manager implementation uses a
+ * `@ethereumjs/trie` trie as a data backend.
```
acolytec3, line 39: Just a reminder to update this doc before we merge

packages/statemanager/test/flat/snapshot.spec.ts
```diff
@@ -0,0 +1,258 @@
...
+    assert.equal(res, null)
+  })
+
+  it('should put/get account', async () => {
+    const snapshot = new Snapshot()
+    const addr = new Address(hexToBytes('0x' + '3'.repeat(40)))
+    const val = new Account()
+    await snapshot.putAccount(addr, val)
+    const res = await snapshot.getAccount(addr)
+    assert.equal(JSON.stringify(res), JSON.stringify(val.serialize()))
```
acolytec3, line 33:
```suggestion
    assert.deepEqual(res, val.serialize())
```

acolytec3: This same format can be applied wherever you are comparing bytes but stringifying first.

acolytec3 (changes requested):
I didn't review every line of the tests in the `statemanager` as I'm assuming the diffs are just where you added another state manager type so it would loop over all the tests with both variants.  Looks directionally great!  Left a fair amount of relatively minor feedback (basically clean-up items).

acolytec3 (changes requested):
I didn't review every line of the tests in the `statemanager` as I'm assuming the diffs are just where you added another state manager type so it would loop over all the tests with both variants.  Looks directionally great!  Left a fair amount of relatively minor feedback (basically clean-up items).

packages/util/package.json
scorbajio: Fixed.

packages/statemanager/src/flat/snapshot.ts
scorbajio: Good catch! I've switched over.

packages/trie/src/db/flat/stackTrie.ts
scorbajio: Since all but the `BaseNode` type refer to each other, I wasn't able to remove all the eslint comments, but I was able to get it down to just two.

packages/trie/src/db/flat/stackTrie.ts
scorbajio: It's necessary at this point because if it is removed, the trie package CI run will fail in the linting step and the tests won't be run: https://github.com/ethereumjs/ethereumjs-monorepo/actions/runs/7341554101/job/19989439562?pr=2980

packages/statemanager/test/flat/snapshot.spec.ts
scorbajio: Fixed this.

packages/statemanager/benchmarks/index.ts
scorbajio: I originally put this benchmark with the trie benchmarks, but later moved it over here since it was more related to the `statemanager` package. It uses the same benchmark tools as the trie benchmarks do (`micro-bmark`). The goal is to show the performance difference between the flat and default statemanagers.

packages/statemanager/src/flat/snapshot.ts
scorbajio: Yes, that was my main concern, so I have it commented out until I can uncomment or remove it with confidence. I'll add a TODO comment above it as well, for clarity.

gabrocheleau:
Since this is mostly Amir's work and is outdated (lots of merge conflicts etc.), will close this. If we were to pursue this again then we likely would want to start from scratch or cherry pick from this anyway

PR #2982, opened 2023-08-23
Add Holesky testnet

jochem-brouwer:
See: https://github.com/eth-clients/holesky

I have asked for confirmation of fork hashes on R&D (they should be right).

Also tested client startup (which works) with the `--network holesky` flag.


packages/common/src/chains.ts
```diff
@@ -448,4 +448,122 @@ export const chains: ChainsDict = {
...
+        forkHash: '0x01d65a8c',
+      },
+      {
+        name: 'london',
+        block: 0,
+        forkHash: '0x01d65a8c',
+      },
+      {
+        // The forkHash will remain same as mergeForkIdTransition is post merge,
+        // terminal block: https://sepolia.etherscan.io/block/1450408
```
acolytec3, line 528: Is this comment (or at least this link) still applicable?

g11tech (approved):
lgtm

acolytec3 (dismissed):
LGTM.  One minor question but not a barrier to merging

packages/common/src/chains.ts
jochem-brouwer: No this is not applicable in this case, thanks :smile:

g11tech (approved):

PR #2984, opened 2023-08-23
Add error handling for async errors in client

### LINKED ISSUE #2983: Client shutdown issues if RLPx port is in use
The following setup makes it impossible (?) to gracefully shutdown the client:

1. Ensure port 30303 is in use (start Geth for instance)
2. Now start the client.
3. Client will complain that 30303 is in use.
4. If you `CTRL-C` the client then it will not shutdown
5. If you now shutdown Geth, RLPx server will start
6. Trying to shutdown the client now also fails to work (RLPx server will start, start importing blocks, and VM will keep executing blocks)

Note: in this example the client was executing past blocks which were imported, but not yet executed.

acolytec3:
Fixes #2983 as well as adds an additional layer of error handling to the client top-level `run` function for errors thrown in eventemitter functions (that normally bypass the `run().catch...` call).

You can test the fix for #2983 by starting geth with default settings and then try to start our client with same defaults.  You will observe an error message like `Server error - EADDRINUSE...` or something indicating that geth is already listening on port 30303.  

On master branch, once this error occurs, you will not be able to shut down the client using Ctrl + C because the client startup function hangs mid promise and never resolves.  This handles the error gracefully and starts the client up.  Oddly enough, the client seems to be able to use the port in contention and still be able to sync mainnet.

I also add some additional error handling for other (as yet unknown) errors thrown during async processes like `EventEmitter` events.  This will log these errors and their stack trace to the console and then call the client shutdown function.  If the client did not successfully start up and the `clientStartPromise` has hung, it will ungracefully terminate the client after 30 seconds.

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -670,6 +670,14 @@ function generateAccount(): Account {
 const stopClient = async (config: Config, clientStartPromise: any) => {
   config.logger.info('Caught interrupt signal. Obtaining client handle for clean shutdown...')
   config.logger.info('(This might take a little longer if client not yet fully started)')
+  let timeoutHandle
+  if (clientStartPromise.toString().includes('Promise') === true)
+    // Client hasn't finished starting up so setting timeout to terminate process if not already shutdown gracefully
+    timeoutHandle = setTimeout(() => {
+      config.logger.warn('Client has become unresponsive while starting up.')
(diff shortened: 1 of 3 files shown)


holgerd77 (approved):
Ah, I find this pretty great, so from my side we can merge! ☺️

holgerd77:
Rebased this via UI

PR #2985, opened 2023-08-23
Tx: Cache sender

jochem-brouwer:
I did a test profile run and noticed that especially `ecrecover` takes relatively long. We should only do this once for frozen txs, since the sender cannot change.

Note: this halves the amount of `ecrecover` calls in VM's `runBlock` (two calls: first in `block.validateData` which `tx.getValidationErrors` which checks the signature, and then in `runTx` because we need to see if the sender has enough data, and of course have to setup EVMs `ORIGIN` / `CALLER` for the transaction).


ScottyPoi (approved):
Nice catch :+1: 

LGTM

holgerd77:
Yes, agree, great catch! 👍

Will merge.

PR #2988, opened 2023-08-24
Add EVM profiler

jochem-brouwer:
This PR adds the feature for an EVM profiler. It features, once enabled, tracking:

- The amount of calls, the time it takes, and the gas it consumes for each opcode
- The same "inside" the precompiles

For precompiles we have the special case that the gas there is not tracked as in normal EVM calls, but instead depends upon what precompile is called. If a precompile is called, the "inner" gas used by that precompile (so without the overhead of the initial CALL opcode to the precompile) is tracked.

The VM gets extra options: `profilerOpts`, which can `reportProfilerAfterBlock` and `reportProfilerAfterTx`. Note that if a profiler is reported, the profiler in the EVM is cleared, so you cannot use both options.

EVM itself has an extra option as well, the `profiler` (should rename this to `profilerOpts` as well for consistency) which currently only has the `enable` flag. This will later support reporting the gas used for different hardforks (while the actual EVM runs on the hardfork it expects for that call) such that we can simulate running Shanghai gas on older blocks (since EVM had many underpriced situations back then).

To run and test:

Add this to `BlockchainTestRunner`: 

```typescript
  let vm = await VM.create({
    stateManager,
    blockchain,
    common,
    setHardfork: true,
    profilerOpts: {
      reportProfilerAfterBlock: true
    }
  })
```

Now run a blockchain test: `npm run test:blockchain -- --fork=Shanghai --test=coinbaseWarmAccountCallGasFail_d3g0v0_Shanghai`

Example output: 

image

TODOs

- [x] Gas reported for CALL situations is incorrect, this encompasses setting up the CALL, but also all time consumed within that call frame
- - To test: `DEBUG=ethjs,evm:profiler npm run test:blockchain -- --fork=Shanghai --test=ContractCreationSpam_d0g0v0_Shanghai`
- - The total time of each opcode consumed is higher than the total test run
- [x] Test if precompile reporting is correct

TODOs for subsequent PRs

- Add option to set custom hardfork for gas reporting to EVM
- Add option to also profile the pre-EVM-call setup in VM, and also the post-EVM-call setup in VM

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -344,6 +344,16 @@ const args: ClientOpts = yargs(hideBin(process.argv))
       'To run client in single node configuration without need to discover the sync height from peer. Particularly useful in test configurations. This flag is automically activated in the "dev" mode',
     boolean: true,
   })
+  .option('vmProfileBlocks', {
+    describe: 'Report the VM profile after running each block',
+    boolean: true,
+    default: false,
+  })
+  .option('vmProfileTxs', {
+    describe: 'Report the VM profile after running each block',
+    boolean: true,
+    default: false,
+  })
   .option('loadBlocksFromRlp', {
     describe: 'path to a file of RLP encoded blocks',
     string: true,
@@ -816,6 +826,8 @@ async function run() {
     mine,
     minerCoinbase: args.minerCoinbase,
     isSingleNode,
+    vmProfileBlocks: args.vmProfileBlocks,
+    vmProfileTxs: args.vmProfileTxs,
     minPeers: args.minPeers,
     multiaddrs,
     port: args.port,
--- a/packages/evm/src/interpreter.ts
+++ b/packages/evm/src/interpreter.ts
@@ -18,15 +18,9 @@ import { Stack } from './stack.js'
 
 import type { EVM } from './evm.js'
 import type { Journal } from './journal.js'
+import type { EVMPerformanceLogger, Timer } from './logger.js'
 import type { AsyncOpHandler, OpHandler, Opcode } from './opcodes/index.js'
-import type {
-  Block,
-  Blockchain,
-  EVMPerformanceLogs,
-  EVMProfilerOpts,
-  EVMResult,
-  Log,
-} from './types.js'
+import type { Block, Blockchain, EVMProfilerOpts, EVMResult, Log } from './types.js'
 import type { Common, EVMStateManagerInterface } from '@ethereumjs/common'
 import type { Address } from '@ethereumjs/util'
 const { debug: createDebugLogger } = debugDefault
@@ -141,7 +135,7 @@ export class Interpreter {
   private opDebuggers: { [key: string]: (debug: string) => void } = {}
 
   private profilerOpts?: EVMProfilerOpts
-  private preformanceLogs: EVMPerformanceLogs
+  private performanceLogger: EVMPerformanceLogger
 
   // TODO remove gasLeft as constructor argument
   constructor(
@@ -151,7 +145,7 @@ export class Interpreter {
     env: Env,
     gasLeft: bigint,
     journal: Journal,
-    performanceLogs: EVMPerformanceLogs,
+    performanceLogs: EVMPerformanceLogger,
     profilerOpts?: EVMProfilerOpts
   ) {
     this._evm = evm
@@ -183,7 +177,7 @@ export class Interpreter {
       returnValue: undefined,
       selfdestruct: new Set(),
     }
-    ;(this.profilerOpts = profilerOpts), (this.preformanceLogs = performanceLogs)
+    ;(this.profilerOpts = profilerOpts), (this.performanceLogger = performanceLogs)
   }
 
   async run(code: Uint8Array, opts: InterpreterOpts = {}): Promise<InterpreterResult> {
@@ -272,18 +266,10 @@ export class Interpreter {
   async runStep(): Promise<void> {
     const opInfo = this.lookupOpInfo(this._runState.opCode)
 
-    let timer: number
+    let timer: Timer
 
     if (this.profilerOpts?.enabled === true) {
-      if (this.preformanceLogs.opcodes[opInfo.name] === undefined) {
-        this.preformanceLogs.opcodes[opInfo.name] = {
-          time: 0,
-          calls: 0,
-          gasUsed: 0,
-          gasPerSecond: 0,
-        }
-      }
-      timer = performance.now()
+      timer = this.performanceLogger.startTimer(opInfo.name)
     }
 
     let gas = BigInt(opInfo.fee)
@@ -326,13 +312,7 @@ export class Interpreter {
       }
     } finally {
       if (this.profilerOpts?.enabled === true) {
-        const delta = (performance.now() - timer!) / 1000
-
-        const field = this.preformanceLogs.opcodes[opInfo.name]
-        field.calls++
-        field.gasUsed += Number(gas)
-        field.time += delta
-        // gas per second is updated when it is being read (i.e. evm.getPerformanceLogs())
+        this.performanceLogger.stopTimer(timer!, Number(gas), 'opcodes')
       }
     }
   }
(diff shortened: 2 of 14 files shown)


packages/vm/src/types.ts
```diff
@@ -140,6 +140,12 @@ export interface VMOpts {
    * Use a custom EVM to run Messages on. If this is not present, use the default EVM.
    */
   evm?: EVMInterface
+
+  profilerOpts?: {
+    //evmProfilerOpts: EVMProfilerOpts
+    reportProfilerAfterTx?: boolean
+    reportProfilerAfterBlock?: boolean
```
holgerd77, line 147: Two quick nits: I would drop the `Profiler` here from the two options, so just `reportAfterTx` and `reportAfterBlock`, context is already clear and option gets shorter.

jochem-brouwer: I have explicitly added this profiler to VM, because we want to (?) profile VM pre-setup-before EVM call and post-setup-after-EVM call too. So after the EVM is ran, then VM will flush the StateManager and this can take some time in some situations, which we want to profile to (?)

jochem-brouwer:
image

packages/client/bin/cli.ts
```diff
@@ -344,6 +344,16 @@ const args: ClientOpts = yargs(hideBin(process.argv))
...
+  .option('vmProfileBlocks', {
+    describe: 'Report the VM profile after running each block',
+    boolean: true,
+    default: false,
+  })
+  .option('vmProfileTxs', {
+    describe: 'Report the VM profile after running each block',
+    boolean: true,
+    default: false,
+  })
```
holgerd77, line 356: Oh this is cool! 🤩

holgerd77: Didn't know that you had integrated so deeply into client already...

jochem-brouwer: Yes, the only thing not to forget is; it does not work if you don't set the `DEBUG=ethjs,evm:profiler` env, so this is something we can improve (but not sure how to do this cleanly)

acolytec3: You should just be able to set the debug variable programmatically with this `process.env.DEBUG = 'ethjs,evm:profiler'`.  So just add a code block to the `run` command in the client that checks for either of these args and then sets the debug accordingly.  Or better, do something like this:
```js
if (args.vmProfileBlocks || args.vmProfileTxs) {
  let debugString = process.env.DEBUG
  if debugString.startsWith('ethjs') {
    debugString = debugString + ',evm:profiler'
  } else {
    debugString = 'ethjs,evm:profiler'
  }
  process.env.DEBUG=debugString
}   
```

acolytec3: Note, I haven't tested the above code but I do something like this in one of the `trie` demos to force it to output certain debug logs.

jochem-brouwer:
HEAD at 2e06d8790340ec0bee16cedaf93f25bdfb91ae73

packages/evm/src/interpreter.ts
```diff
@@ -259,41 +266,54 @@ export class Interpreter {
...
-    if (opInfo.name === 'INVALID') {
-      throw new EvmError(ERROR.INVALID_OPCODE)
-    }
+    let gas = BigInt(opInfo.fee)
 
-    // Reduce opcode's base fee
-    this.useGas(gas, `${opInfo.name} fee`)
-    // Advance program counter
-    this._runState.programCounter++
+    try {
```
acolytec3, line 277: Why does this require a `try`?  Can't we just check if the timer exists and stop it at the end with out the `try/finally` syntax?

acolytec3 (commented):
This looks good, though left a couple of questions about syntax I'm not quite tracking.

packages/evm/src/interpreter.ts
```diff
@@ -171,6 +177,7 @@ export class Interpreter {
       returnValue: undefined,
       selfdestruct: new Set(),
     }
+    ;(this.profilerOpts = profilerOpts), (this.performanceLogger = performanceLogs)
```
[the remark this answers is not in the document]
jochem-brouwer: No idea :joy: I definitely did not write this, I think I made a weird syntax mistake and then the linter did this. Will fix

packages/evm/src/interpreter.ts
jochem-brouwer: It is not `try/finally` because of (internal) timer errors.

The reason is that without try/catch, the inner method can error, for instance, see:

```typescript
      // Check for invalid opcode
      if (opInfo.name === 'INVALID') {
        throw new EvmError(ERROR.INVALID_OPCODE)
      }
```

If this thus throws, the timer is never stopped. (This can also happen within other opcodes, e.g. not enough gas, stack underflow, etc.)

packages/evm/src/interpreter.ts
```diff
@@ -259,41 +266,54 @@ export class Interpreter {
...
-      opFn.apply(null, [this._runState, this.common])
+      if (this._evm.events.listenerCount('step') > 0 || this._evm.DEBUG) {
+        // Only run this stepHook function if there is an event listener (e.g. test runner)
+        // or if the vm is running in debug mode (to display opcode debug logs)
+        await this._runStepHook(gas, gasLimitClone)
+      }
+
+      // Check for invalid opcode
+      if (opInfo.name === 'INVALID') {
+        throw new EvmError(ERROR.INVALID_OPCODE)
```
jochem-brouwer, line 297: @acolytec3 see here, this is inside the try/catch

I don't know why this git diff is weird, because I think I only added the try catch blocks and the timers, nothing else, but here it seems like I added a whole bunch of EVM logic (which I did not do)

packages/evm/src/interpreter.ts
acolytec3: Is there intentionally no `catch` block here and if an error is thrown, does that `finally` block still get executed to stop the timer?

jochem-brouwer: Yes, `finally` always runs (I had to check this as well)

jochem-brouwer:
```
try { throw new Error() } finally { console.log(1) }
1
Uncaught Error
```

jochem-brouwer:
```
try { throw new Error() } catch(e) { throw(e) } finally { console.log(1) }
1
Uncaught Error
```

acolytec3: Interesting, learned something new today.

packages/evm/src/interpreter.ts
acolytec3: I'm not sure, I don't see all of this as a diff on my local but it looks fine.

packages/evm/src/interpreter.ts
jochem-brouwer: So this is very nice, you can thus add code which always runs even if it errors or not (even if catch block re-raises the error). So this is a nice pattern if you setup something which you **always** want to cleanup after. Geth has this natively with `defer`, I wanted this in JS too but apparently this is the way to do that :smile:

acolytec3 (approved):
LGTM

packages/vm/src/types.ts
jochem-brouwer: Ah you are right, will update

holgerd77: I would actually still be a fan of this name changes (can be along some follow-up work), so that we get to e.g. `this._opts.profilerOpts?.reportAfterBlock` (still long enough 🙂).

jochem-brouwer: > I would actually still be a fan of this name changes (can be along some follow-up work), so that we get to e.g. `this._opts.profilerOpts?.reportAfterBlock` (still long enough 🙂).

Yes, added this :)

holgerd77 (dismissed):
LGTM

holgerd77 (approved):
Still looks good to me 🙂

holgerd77:
🎉

PR #2937, opened 2023-08-03
Reintegrate Skipped vm Tests

scorbajio:
See #2929.

PR #2991, opened 2023-08-24
VM: Reactivate more slow tests

holgerd77:
Follow-up on #2937 

I had the impression that the static call tests currently still skipped are also fast enough to be reintegrated.

This is a first test run on all HFs.

WIP, will look into some more additional tests.


g11tech (approved):
lgtm

jochem-brouwer (approved):

jochem-brouwer:
Istanbul and MuirGlacier are failing because the test amount is incorrect

Istanbul: 38340

jochem-brouwer:
I think due to https://github.com/ethereumjs/ethereumjs-monorepo/pull/2971 tests decreased.

Small note: if we update ethereum-tests to a new version, we should always use `test all hardforks` as label (did not notice this at that PR).

The CI should pass now, but will wait for approval until it does.

jochem-brouwer (approved):

PR #2993, opened 2023-08-26
tx: consolidate generic tx capabilities

### LINKED ISSUE #2987: Tx: Consolidate generic Functionality in Shared Static Helpers
Just stumbled upon this when having a look at the "Tx cache sender" PR #2985 from @jochem-brouwer (great work! 🎉):

Due to our still not super-optimal tx library design we have *a lot* of code redundancy in there and this worsens with every new code additions like in the PR above.

We should mitigate this - independently from the overall design questions - by introducing static helper methods in a separate file taking in the tx (so: `this`) as a first input + other necessary values and then combines common logic.

I would cautiously say that it makes sense to associate respectively structure these helpers by the `Capabilities` we have for the txs (see `types.ts`) and I would suggest a following structure with a subfolder `capabilities`:

- `capabilities/eip1559FeeMarket.ts`
- `capabilities/eip2718TypedTransaction.ts`
- maybe also: `capabilities/generic.ts`, things which are the same for all txs (would cautiously say `hash()` might be a candidate?)
- ah, and maybe also: `capabilities/legacy.ts` (that's a bit deviating from the enum, but we have things which have been introduced by legacy tx - like the respective `getUpfrontCost()` calculation, which then *changes* with e.g. EIP1559 (so this wouldn't fit into the `generic` category

Just to point out respectively reiterate: the difference between an `EIP1559Transaction` and a `Capability.EIP1559FeeMarket` is that the first encompasses only the respective one transaction type while the latter (so: the capability) can be shared between different tx types which also make this capability (introduced first by the tx type) available.

It would be generally good if we get the assocations right (so: put a method in the "correct" capability file).

Then we can import in the respective tx files (e.g. `eip1559Transaction.ts`) with something like:

```typescript
import * as EIP2718TypedTransaction from 'capabilities/eip2718TypedTransaction.ts'
import * as Legacy from 'capabilities/legacy.ts`
```

(this would be my naming suggestion after giving this some thought)

and then one can call into that with e.g.:

```typescript
public hash(): Uint8Array {
  return Generic.hash(this)
}
```

Also after giving this some thought I would cautiously say that it should be ok for methods where the implementation differs (but the name is the same), so e.g. `getUpfrontCost()` to just have the same name in the different capability files and respectively call into the one used, so e.g.

```typescript
getUpfrontCost(baseFee: bigint = BigInt(0)): bigint {
   return EIP1559FeeMarket.getUpfrontCost(this, baseFee)
}
```

One note: we might get *some* problems with **accessing internal methods**, cache seems to be a particularly reoccuring thing. I think for cache it might e.g. be ok to expose, directly or via a method or something. Generally I think that this whole (so: the whole issue) approach might not work for all and every method but after having a look it should work for many/most and a consolidation in this way would already mean a huge improvement.

Another note: A bit related: I think it is likely advisable to *start* with transforming/moving methods with low or no internal calls or property accesses and then slowly work ones way from there onwards.

And third note: I do think we should even do this transform for one-liner methods, e.g. one example:

```typescript
// old
serialize(): Uint8Array {
  const base = this.raw()
  return concatBytes(TRANSACTION_TYPE_BYTES, RLP.encode(base))
}

// new
serialize(): Uint8Array {
  return EIP2718TypedTransaction.serialize(this, TRANSACTION_TYPE_BYTES)
}
```

...since this will generally help us to avoid consistency mistakes if we change something, even in this one-line implementation.

gabrocheleau:
This PR resolves #2987

### DIFF
--- a/packages/tx/src/capabilities/eip1559.ts
+++ b/packages/tx/src/capabilities/eip1559.ts
@@ -1,0 +1,13 @@
+import type { Transaction, TransactionType } from '../types.js'
+
+type TypedTransactionEIP1559 =
+  | Transaction[TransactionType.BlobEIP4844]
+  | Transaction[TransactionType.FeeMarketEIP1559]
+
+export function getUpfrontCost(this: TypedTransactionEIP1559, baseFee: bigint): bigint {
+  const prio = this.maxPriorityFeePerGas
+  const maxBase = this.maxFeePerGas - baseFee
+  const inclusionFeePerGas = prio < maxBase ? prio : maxBase
+  const gasPrice = inclusionFeePerGas + baseFee
+  return this.gasLimit * gasPrice + this.value
+}
--- a/packages/tx/src/capabilities/eip1559.ts
+++ b/packages/tx/src/capabilities/eip1559.ts
@@ -4,10 +4,10 @@ type TypedTransactionEIP1559 =
   | Transaction[TransactionType.BlobEIP4844]
   | Transaction[TransactionType.FeeMarketEIP1559]
 
-export function getUpfrontCost(this: TypedTransactionEIP1559, baseFee: bigint): bigint {
-  const prio = this.maxPriorityFeePerGas
-  const maxBase = this.maxFeePerGas - baseFee
+export function getUpfrontCost(tx: TypedTransactionEIP1559, baseFee: bigint): bigint {
+  const prio = tx.maxPriorityFeePerGas
+  const maxBase = tx.maxFeePerGas - baseFee
   const inclusionFeePerGas = prio < maxBase ? prio : maxBase
   const gasPrice = inclusionFeePerGas + baseFee
-  return this.gasLimit * gasPrice + this.value
+  return tx.gasLimit * gasPrice + tx.value
 }
--- a/packages/tx/src/capabilities/generic.ts
+++ b/packages/tx/src/capabilities/generic.ts
@@ -1,4 +1,4 @@
-import { bigIntToUnpaddedBytes, ecrecover } from '@ethereumjs/util'
+import { SECP256K1_ORDER_DIV_2, bigIntToUnpaddedBytes, ecrecover } from '@ethereumjs/util'
 import { keccak256 } from 'ethereum-cryptography/keccak.js'
 
 import { Capability, type TypedTransaction } from '../types.js'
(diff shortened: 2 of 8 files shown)


acolytec3:
General question on this PR.  What is the advantage of declaring these helpers as 
`export function serialize(this: TypedTransactionEIP2930): Uint8Array` and then using
`return EIP2930.serialize.bind(this)()` inside each Transaction subclass over just declaring the functions as generic pure functions like:
`export function serialize(tx: TypedTransactionEIP2930): Uint8Array`
and then calling them with
`return EIP2930.serialize(this)`

If there's some SO thread of something that explains the motivation for this, I'd be happy to read if that's easier to do.

packages/tx/src/capabilities/generic.ts
```diff
@@ -0,0 +1,85 @@
+import { SECP256K1_ORDER_DIV_2, bigIntToUnpaddedBytes, ecrecover } from '@ethereumjs/util'
+import { keccak256 } from 'ethereum-cryptography/keccak.js'
+
+import { Capability, type TypedTransaction } from '../types.js'
```
acolytec3, line 4: Ah, I didn't know you could import a type inside a general import statement like this.  Neat!

acolytec3 (approved):
This looks great!

packages/tx/src/capabilities/generic.ts
gabrocheleau: Found that out recently too!

holgerd77:
Super great work, thanks a lot! 🤩

I would really like to put an emphasis here that we get the associations right, so to expand: that we a) name the capability files correctly and b) associate/integrate the correct methods, so that this whole things directly structure things in a right way along, I have some remarks on this, will do along the code.

The most basic specific thing here: [EIP-2930](https://eips.ethereum.org/EIPS/eip-2930) is really only this specific (very much unused) transaction type with access lists. So this should - if including anything - only include access list specific functionality (which we have already extracted in a similar way in former work in `util.ts` and - while it would likely make sense to align this - we likely want/need to skip for now due to backwards compatibility reasons).

All the generic tx type specific functionality has been introduced in [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) and I would therefore strongly lean towards also labeling the capability file this way. Your local transaction type at the beginning of the file is actually the very definition of EIP-2718 typed transactions 🙂:

```typescript
type TypedTransactionEIP2930 = Exclude<TypedTransaction, Transaction[TransactionType.Legacy]>
```

Will do some more comments on the files.

packages/tx/src/capabilities/generic.ts
```diff
@@ -0,0 +1,85 @@
...
+    )
+    if (Object.isFrozen(tx)) {
+      tx['cache'].senderPubKey = sender
+    }
+    return sender
+  } catch (e: any) {
+    const msg = errorMsg(tx, 'Invalid Signature')
+    throw new Error(msg)
+  }
+}
```
holgerd77, line 85: -> legacy.ts

holgerd77 (commented):
Ok, gone through the list of capability methods with various comments.

packages/tx/src/capabilities/eip1559.ts
```diff
@@ -0,0 +1,13 @@
+import type { Transaction, TransactionType } from '../types.js'
+
+type TypedTransactionEIP1559 =
+  | Transaction[TransactionType.BlobEIP4844]
+  | Transaction[TransactionType.FeeMarketEIP1559]
```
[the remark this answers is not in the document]
gabrocheleau: I'll think about this a bit more about how to architect this. You're right that these "capabilities" are currently tightly paired with "transaction types", whereas they shouldn't. I'm currently wondering if we make use of the existing tx Capability enum. Will reflect a bit more and follow up...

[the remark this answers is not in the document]
gabrocheleau: Here's what I feel is a more sound approach: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3010, curious to hear your thoughts!

PR #2994, opened 2023-08-29
Common: Cache Parameter Values + activated EIPs for current Hardfork / SM copy() fix

holgerd77:
First-round performance tests suggests that `Common` `param()` reads take a significant amount of time, see:

```typescript
> import { Chain, Common, Hardfork } from '@ethereumjs/common'
undefined
> let c = new Common({ chain: Chain.Mainnet })
undefined
> console.time('Total'); for (let i = 0; i < 100000; i++) { let p = c.param('gasPrices', 'push') }; console.timeEnd('Total');
Total: 151.637ms
```

(a similar issue was already mentioned years ago by Patricio from HardHat in some separate peformance tracking issue, just to mention)

Time used in the 100s of miliseconds is significant, since the iteration set of 100000 is relatively close to real world conditions and an amount realistic to be called in e.g. the EVM during a block execution (since the `param()` method is used for every opcode gas calculation e.g.).

Looking at the code confirms that this is likely inefficiently coded, since every `param()` read again iterates through both all hardfork files and additionally through all EIP files.

This PR therefore aims to introduce a cache with the parameters at hand for currently active set of both hardforks and EIPs. This cache will be rebuild on a hardfork change or an EIP addition (which both either doesn't happen at all or at least doesn't happen often during a Common object livecycle).

### DIFF
--- a/packages/common/src/common.ts
+++ b/packages/common/src/common.ts
@@ -58,6 +58,7 @@ export class Common {
   protected _customChains: ChainConfig[]
 
   protected _paramsCache: ParamsCacheConfig = {}
+  protected _activatedEIPsCache: number[] = []
 
   protected HARDFORK_CHANGES: [HardforkSpecKeys, HardforkSpecValues][]
 
@@ -248,6 +249,7 @@ export class Common {
     }
     if (Object.keys(this._paramsCache).length === 0) {
       this._buildParamsCache()
+      this._buildActivatedEIPsCache()
     }
   }
 
@@ -295,6 +297,7 @@ export class Common {
         if (this._hardfork !== hardfork) {
           this._hardfork = hardfork
           this._buildParamsCache()
+          this._buildActivatedEIPsCache()
           this.events.emit('hardforkChanged', hardfork)
         }
         existing = true
@@ -470,6 +473,12 @@ export class Common {
           `${eip} cannot be activated on hardfork ${this.hardfork()}, minimumHardfork: ${minHF}`
         )
       }
+    }
+    this._eips = eips
+    this._buildParamsCache()
+    this._buildActivatedEIPsCache()
+
+    for (const eip of eips) {
       if ((EIPs as any)[eip].requiredEIPs !== undefined) {
         for (const elem of (EIPs as any)[eip].requiredEIPs) {
           if (!(eips.includes(elem) || this.isActivatedEIP(elem))) {
@@ -478,8 +487,6 @@ export class Common {
         }
       }
     }
-    this._eips = eips
-    this._buildParamsCache()
   }
 
   /**
@@ -540,6 +547,18 @@ export class Common {
 
       this._mergeWithParamsCache(EIPs[eip])
     }
+  }
+
+  protected _buildActivatedEIPsCache() {
+    this._activatedEIPsCache = []
+
+    for (const hfChanges of this.HARDFORK_CHANGES) {
+      const hf = hfChanges[1]
+      if (this.gteHardfork(hf['name']) && 'eips' in hf) {
+        this._activatedEIPsCache = this._activatedEIPsCache.concat(hf['eips'] as number[])
+      }
+    }
+    this._activatedEIPsCache = this._activatedEIPsCache.concat(this._eips)
   }
 
   /**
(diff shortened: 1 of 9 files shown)


acolytec3 (commented):
This looks great overall. What is the significance of changing all the `private` elements to `protected`?

packages/common/src/common.ts
```diff
@@ -238,12 +247,16 @@ export class Common {
...
+      this._buildParamsCache()
+      this._buildActivatedEIPsCache()
+    }
   }
 
   /**
    * Sets the chain
    * @param chain String ('mainnet') or Number (1) chain representation.
-   *              Or, a Dictionary of chain parameters for a private network.
+   *              Or, a Dictionary of chain parameters for a protected network.
```
jochem-brouwer, line 259:
```suggestion
   *              Or, a Dictionary of chain parameters for a private network.
```

packages/common/src/common.ts
```diff
@@ -466,14 +487,85 @@ export class Common {
...
+  protected _buildActivatedEIPsCache() {
+    this._activatedEIPsCache = []
+
+    for (const hfChanges of this.HARDFORK_CHANGES) {
+      const hf = hfChanges[1]
+      if (this.gteHardfork(hf['name']) && 'eips' in hf) {
+        this._activatedEIPsCache = this._activatedEIPsCache.concat(hf['eips'] as number[])
+      }
+    }
+    this._activatedEIPsCache = this._activatedEIPsCache.concat(this._eips)
```
jochem-brouwer, line 561: For even more efficiency for activated EIPs we might want to use a `Set` instead of an object.

packages/vm/test/api/index.spec.ts
```diff
@@ -225,8 +225,9 @@ describe('VM -> setHardfork, state (deprecated), blockchain', () => {
       const contractAddress = Address.fromString('0x00000000000000000000000000000000000000ff') // contract address
       // setup the vm
       const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Istanbul })
+      const common2 = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Istanbul })
       const vmNotActivated = await VM.create({ common })
-      const vmActivated = await VM.create({ common, activatePrecompiles: true })
+      const vmActivated = await VM.create({ common: common2, activatePrecompiles: true })
```
jochem-brouwer, line 230: I'm not sure here, why is the extra `common2` necessary here?

holgerd77:
> This looks great overall. What is the significance of changing all the `private` elements to `protected`?

That's our "standard" for the libraries, so preferring `protected` over `private`. This makes it easier for people to subclass Common (or another library) and adopt stuff (e.g. overwriting a then `protected` method). This is not possible if things are declared `private`. And from the outer visibility there's no difference.

packages/statemanager/src/stateManager.ts
```diff
@@ -926,6 +926,9 @@ export class DefaultStateManager implements EVMStateManagerInterface {
    * 2. Cache values are generally not copied along
    */
   shallowCopy(): DefaultStateManager {
+    const common = this.common.copy()
+    common.setHardfork(this.common.hardfork())
```
jochem-brouwer, line 930: We need to `setHardfork` here to initialize the cache, right?

packages/vm/test/tester/index.ts
```diff
@@ -263,6 +263,9 @@ async function runTests() {
         t.ok(assertCount >= expectedTests, `expected ${expectedTests} checks, got ${assertCount}`)
       }
 
+      console.log()
+      console.timeEnd('Total (including setup)')
```
jochem-brouwer, line 267: Do we want to keep the profile timers here?

jochem-brouwer (commented):
Some remarks and questions :) LGTM overall!

packages/statemanager/src/stateManager.ts
holgerd77: Yes, I think it's at least a bit safer like this (not 1000% sure it's necessary).

packages/vm/test/tester/index.ts
holgerd77: Yes, I would really like to keep. I need this information repeatedly and it's actually not so easy to place this in "on the fly".

packages/vm/test/api/index.spec.ts
holgerd77: Tested, not necessary any more after the latest initialization fix. 🙂 Will remove.

packages/common/src/common.ts
jochem-brouwer: (this is not necessary for this PR)

holgerd77:
@acolytec3 @jochem-brouwer thanks for the reviews, have answered and/or applied changes. Ready for re-review once tests pass! 🙂

jochem-brouwer (approved):
LGTM!

PR #2996, opened 2023-08-29
EVM optimizations

jochem-brouwer:
This PR:

Optimizing `PUSH` opcode 
------
- Cache PUSH opcode. Each time PUSH is called, code is being re-sliced each time (this is done upon JUMPDEST analysis)
- Update PUSH opcode such that it only calls `isActivatedEIP` in case program counter goes out-of-bounds after PUSH

Micro-managing each overhead before actually executing the opcode (each `runStep`)
-------
- Add `feeBigInt` to the `Opcode` class such that we don't have to convert the `opcode.fee` field to bigint each time
- Add `isInvalid` boolean flag to `Opcode` class such that we can quickly check if the opcode is INVALID (instead of doing a string comparison)
- Remove creating debug strings each time if we call `useGas` in `runStep`
- Put all opcode info into one map. Therefore, looking up an opcode is now done inside EVMs interpreter only once (previously at least two: lookup opcode info in one map, and the fn handler in the other)
- In case of `JUMPDEST` analysis, cache all opcode data of the contract in an Array (this saves Map lookups)

Note: all these changes can be cherry-picked, we should carefully analyze if these changes indeed improve the performance.


jochem-brouwer:
This change: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2996/commits/478f48a019438118d68f8eb0bc152102e6f0822b

Makes `isActivatedEIP` called 1 million times faster, before: `1.586` seconds, after: `0.533` seconds

jochem-brouwer:
(notes for self)

557625575bdb7e6bc264c2b7e483bac5e924cf53
0daf59339cb2cac02a453d7385555e484a7ecbc9
9d02a391f9f8a3392b0873e1534bd81a20cb3bbf
79a0ec81a8895497ddaa817fb872978560f72b18
1c8db86734ad4871d40561e9c16d00f566cd5833

after: 911b03e96ecfa0e1f675a43017aa034555c5b9f7 "opcode tests"
b2b68b11189f19e9fb48d676377ad11716cac4e7 short circuit isActivatedEIP

packages/evm/src/interpreter.ts
```diff
@@ -234,7 +236,9 @@ export class Interpreter {
         (opCode === 0x56 || opCode === 0x57 || opCode === 0x5e)
       ) {
         // Only run the jump destination analysis if `code` actually contains a JUMP/JUMPI/JUMPSUB opcode
-        this._runState.validJumps = this._getValidJumpDests(this._runState.code)
+        const { jumps, pushs } = this._getValidJumpDests(this._runState.code)
```
scorbajio: `pushs` should be named `pushes` so that it's clear that it's the plural form of "push" (gotta love english). Same thing with `cachedPushs`.

scorbajio (commented):
I looked through each optimization, and they all make sense to me. It would be helpful to have benchmarks for each one of them to quantify the performance gain, something like the profiler you shared in #3000.

holgerd77:
Nice, only read the descriptions yet, will have a look at the code changes tomorrow morning, but this app sounds great! 👍🤩🙏

packages/evm/src/interpreter.ts
holgerd77: Have pushed a commit on this.

holgerd77:
Ok, I did some comparison runs with:

```shell
git checkout evm-opts && cd ../evm && npm run build && cd ../client && npm run client:start -- --sync=none --vmProfileBlocks --executeBlocks=1009000-1009302
```

And:

```shell
git checkout master && cd ../evm && npm run build && cd ../client && npm run client:start -- --sync=none --vmProfileBlocks --executeBlocks=1009000-1009302
```

(as a side note: I can very much confirm that this drastically helps to have such a command "in one go" since otherwise chances are such so high that one is forgetting to rebuild or to checkout or whatever. By having this "atomic" and can exclude that these errors are happening otherwise causing a lot of confusion)

grafik

So, these is one result run, I did several (~10 per branch or so). Very much confirmed that things are overall improving, with some more significant gains for the opcodes dedicatedly optimized (PUSH in particular! 🤩).

All great!

For `SLOAD` I discovered a very particular behavior, just to mention here as well: results are just absolutely not reliable (respectively, likely the better word: repeatable). Sometimes the one is faster, sametimes the other, sometimes results are totally off (also in a better direction).

Guess `SLOAD` is just very much a special case which we need to analyze separately. We should just make sure that we do not accidentally introduce side-deteriorations here.

holgerd77:
Some commit-by-commit observations:

85d44953b4f02c0107073200344f669e5c6b403e (PUSH cache) : This gives PUSH a significant jump (for me) from 11 to 16 (always Mgas/s)
4d40bbe9d6533821cd0164baaa7ce657afb57a68 (conditional order switch) : difference not noticeable, expected
c283730a8ab8d0eff1745a928c6da1702909203e (do not convert number -> bigint each run step) : impressed me the most yet: this seems very well noticeable. 🤯

Here are two runs (left is the newer commit), have repeatedly similar results:

grafik

So this really boosts *every* opcode by a noticeable level. So this might be an anchor point for other optimizations as well, respectively we might want to keep looking where type conversions take place and optimally remove. Also brings thoughts back to all the stack conversions. So this will likely remain an experimentation field. Might also be worth to think if bigint is really needed *everywhere* or if there are also places where a simple number still does the job.

af8de00181ab219524107a4db869f4180cdd9875 (add isInvalid flag to ops): seems to not do much

7e3ed048e74b31c139651c85eeeed2bbc5de7619 (ensure useGas runStep does not create strings each time) : This impressively also seems to bring a noticeable improvement on all opcodes. Again 🤯 to me (maybe even a bit more than the other one, since this is simply caching the ``${opInfo.name} fee`` creation. Wow.

17cd10634d99f1e08fb567f77e67cdbbf612ca36 (put all opcode info into one map ) : again, noticeable for all opcodes

0b2461331e21ff5e01a4a4552baf9b883e6b55d2 (change map to array, cache opcodes on jumpdest analysis ) : not so clear of an effect

holgerd77 (approved):
One of the most remarkable performance works I have seen in a very long time, also a lot of inspirations for further work! 🙏 🙂

Really well done, cool! 🎉

PR #3001, opened 2023-09-01
trie: use equalsBytes for root check

ScottyPoi:
# `'/trie.ts'`   
#### changing
```
    if (this.root() === this.EMPTY_TRIE_ROOT ... ) 

```
#### to

```
    if (equalsBytes(this.root(), this.EMPTY_TRIE_ROOT)  ... ) 
```

### DIFF
--- a/packages/trie/src/trie.ts
+++ b/packages/trie/src/trie.ts
@@ -771,7 +771,11 @@ export class Trie {
       } as PutBatch
     })
 
-    if (this.root() === this.EMPTY_TRIE_ROOT && opStack[0] !== undefined && opStack[0] !== null) {
+    if (
+      equalsBytes(this.root(), this.EMPTY_TRIE_ROOT) &&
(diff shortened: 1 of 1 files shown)


ScottyPoi:
This PR fixes the bug
Opening issue  #3002 because the bug wasn't causing issues, and that's probably an issue

acolytec3 (approved):
LGTM! Do you think there are existential issues with `fromProof`?

ScottyPoi:
> LGTM! Do you think there are existential issues with `fromProof`?

no not really.  

in practice the methods worked without this fix because an empty trie root is usually set by 
`trie.root(trie.EMPTY_TRIE_ROOT)`.
  So `trie.root() === trie.EMPTY_TRIE_ROOT`  tended to be `true`, but not for the right reason.

g11tech (approved):

PR #3005, opened 2023-09-01
statemanager: c8 ignore DEBUG conditionals

ScottyPoi:
### statemanager

Both `DefaultStateManager` and `EthersStateManager` use a performance feature that skips calls to the `debug` logger if the log is disabled.

Coverage stats for the `statemanager` package suffered due to these conditionals / lines of code, as the `coverage` tests are run without DEBUG enabled. 

We can ask `v8` to ignore lines of code using `c8 ignore` comments:
` /* c8 ignore next */ `
  or
` /* c8 ignore next n */ `

https://github.com/bcoe/c8/blob/main/README.md#ignoring-uncovered-lines-functions-and-blocks

----

While the coverage improvements may be minor, this also helps to reduce the noise in the coverage reports for these 2 files.

* note that the `c8 ignore` API ignores a set count of lines.  If the debug messages are edited in the future, the line count would need to be edited accordingly.


holgerd77 (commented):
Ok, that’s an interesting find.

I am a bit undecided on this change, regarding if we want to bind us to on every debug call add this + the need to adopt line counts in the future. I would have a slight “take it” tendency, but would like to hear from others first what their tendency is before merging.


Some general formatting nit: can we then always do these comments on top of the affected lines, so that the actual code is not indented to the right? Thanks!

ScottyPoi:
I would prefer a solution that did not require a comment for each individual case, but I was not able to find one.

I suppose another option would be to run `coverage` with DEBUG enabled  :shrug:

holgerd77:
Superseeded by #3027 (alternative solution), will close here, @ScottyPoi thanks again for investigating! 🙏

Issue #3006, opened 2023-09-02
`rlp` CLI throws `ERR_UNKNOWN_FILE_EXTENSION`


ardislu:
Steps to reproduce the issue:
1. Install the `rlp` CLI: `npm i -g @ethereumjs/rlp`
2. Run: `rlp encode '5'`

Expected result: command runs successfully and returns `0x05`

Actual result: command fails with the below error

```
TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension "" for /.../node_modules/@ethereumjs/rlp/bin/rlp. Loading extensionless files is not supported inside of "type":"module" package.json contexts. The package.json file /.../node_modules/@ethereumjs/rlp/package.json caused this "type":"module" context. Try changing /.../node_modules/@ethereumjs/rlp/bin/rlp to have a file extension. Note the "bin" field of package.json can point to a file with an extension, for example {"type":"module","bin":{"rlp":"./bin/rlp.js"}}
    at new NodeError (node:internal/errors:405:5)
    at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:99:9)
    at defaultGetFormat (node:internal/modules/esm/get_format:142:36)
    at defaultLoad (node:internal/modules/esm/load:91:20)
    at DefaultModuleLoader.load (node:internal/modules/esm/loader:263:26)
    at DefaultModuleLoader.moduleProvider (node:internal/modules/esm/loader:179:22)
    at new ModuleJob (node:internal/modules/esm/module_job:63:26)
    at #createModuleJob (node:internal/modules/esm/loader:203:17)
    at DefaultModuleLoader.getJobFromResolveResult (node:internal/modules/esm/loader:156:34)
    at DefaultModuleLoader.getModuleJob (node:internal/modules/esm/loader:141:17) {
  code: 'ERR_UNKNOWN_FILE_EXTENSION'
}
```

---

As suggested by the error message, this looks like it could be a simple fix (**rename `bin/rlp` to `bin/rlp.js`**). However, not sure if there are any dependencies that will break by changing this file name. Also looks like the CLI code hasn't been touched in a while so there may be more updates that need to be made to make it compatible with modules.

acolytec3:
Thanks for reporting this!  I've opened a PR to fix so will hopefully be able to include it in  new point release in the near future.

gabrocheleau:
Closing this since it was fixed by #3007

PR #3007, opened 2023-09-02
Fixes the RLP CLI

acolytec3:
Fixes the `rlp` CLI as reported in #3006.  Also adds a couple of simple CLI tests to verify that it works.


holgerd77:
Great improvement 🤩, browser tests still failing though. Guess we simply do not want to run CLI tests in the browser, right?

acolytec3:
> Great improvement 🤩, browser tests still failing though. Guess we simply do not want to run CLI tests in the browser, right?

🤦 I didn't even think about that. But yes, this test definitely won't run in the browser.

acolytec3:
I'll have to add a vitest config there and have it skip that test for `npm run test:browser`

scorbajio (approved):
utACK

PR #3010, opened 2023-09-05
tx: improve tx capability handling

gabrocheleau:
This PR addresses comments and suggestions for improvements that were made on #2993


packages/tx/src/types.ts
```diff
@@ -152,6 +171,36 @@ export interface TransactionInterface<T extends TransactionType> {
   errorStr(): string
 }
 
+export interface EIP2718CompatibleTxInterface<T extends TransactionType = TransactionType>
+  extends TransactionInterface<T> {
+  readonly chainId: bigint
+  readonly accessList: AccessListBytes
+  readonly AccessListJSON: AccessList
```
holgerd77, line 178: Sorry to be a bit pedantic here, but access lists have really nothing to do with the generic introduction of typed transactions - and so with EIP2718 - and if we have this here, we still have something structurally wrong (e.g. using this type from here while we should have used the EIP2930Compatible type).

Generally this looks great though with these new interfaces! 🤩

packages/tx/src/capabilities/eip2718.ts
```diff
@@ -0,0 +1,26 @@
...
+import { errorMsg } from './legacy.js'
+
+import type { EIP2718CompatibleTxInterface, TypedTransaction } from '../types'
+import type { Input } from '@ethereumjs/rlp'
+
+export function getHashedMessageToSign(tx: EIP2718CompatibleTxInterface): Uint8Array {
+  return keccak256(tx.getMessageToSign())
+}
+
+export function serialize(tx: EIP2718CompatibleTxInterface, base?: Input): Uint8Array {
```
holgerd77, line 16: Maybe we drop the `Interface` from these names to get it a bit shorter? 🤔 No super strong preference, but we also mostly (exception: these very important reused interfaces in Common) do not explicitly name interfaces with `Interface` or `I`or the like in the monorepo.

holgerd77: (but the shortening would be the greater argument here, especially if there are more arguments to follow, then this likely often moves over unnecessarily into two lines)

packages/tx/src/capabilities/eip2718.ts
```diff
@@ -0,0 +1,26 @@
...
+  return concatBytes(txTypeBytes(tx.type), RLP.encode(base ?? tx.raw()))
+}
+
+export function validateYParity(tx: TypedTransaction) {
```
holgerd77, line 20: Is it not possible to type with `EIP2718CompatibleTxInterface` here?

packages/tx/src/types.ts
```diff
@@ -152,6 +171,32 @@ export interface TransactionInterface<T extends TransactionType> {
   errorStr(): string
 }
```
holgerd77, line 172: What is somewhat missing here which would complement the picture would be a `LegacyTxInterface`.

I know this is likely the hardest one with eventual conflicts on re-/cross-used functionality or property access.

But did you try? 

That would be great if we would/could achieve that. Since here we - again - at least to some extend replicate this old hierarchy structure which we would optimally want to completely leave behind.

packages/tx/src/eip1559Transaction.ts
```diff
@@ -42,7 +44,10 @@ const TRANSACTION_TYPE_BYTES = hexToBytes(
  * - TransactionType: 2
  * - EIP: [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)
  */
-export class FeeMarketEIP1559Transaction extends BaseTransaction<TransactionType.FeeMarketEIP1559> {
+export class FeeMarketEIP1559Transaction
+  extends BaseTransaction<TransactionType.FeeMarketEIP1559>
+  implements EIP1559CompatibleTxInterface<TransactionType.FeeMarketEIP1559>
```
holgerd77, line 43: I think this is not structurally correct to add this `implements` note respectively is both redundant and not complete at the same time. So I think we should remove.

If we would want to do this correctly/completely we would need to implement from several compatibility interfaces (here e.g. also from the EIP2930 interface), but multiple inheritance unfortunately doesn't exist in JavaScript (I thought a bit along these lines too some 1-2 years ago when thinking about how we could refactor this stuff).

packages/tx/src/eip1559Transaction.ts
```diff
@@ -306,7 +311,7 @@ export class FeeMarketEIP1559Transaction extends BaseTransaction<TransactionType
    * Returns the public key of the sender
    */
   public getSenderPublicKey(): Uint8Array {
-    return Generic.getSenderPublicKey(this)
+    return Legacy.getSenderPublicKey(this)
```
holgerd77, line 309: Yeah, but generally this reads super-great, to always already have the heritage of the feature read out of these calls, so that one can associate how the tx type is composed and where the functionality comes from! 🥳

packages/tx/src/eip4844Transaction.ts
```diff
@@ -82,7 +84,10 @@ const validateBlobTransactionNetworkWrapper = (
  * - TransactionType: 3
  * - EIP: [EIP-4844](https://eips.ethereum.org/EIPS/eip-4844)
  */
-export class BlobEIP4844Transaction extends BaseTransaction<TransactionType.BlobEIP4844> {
+export class BlobEIP4844Transaction
+  extends BaseTransaction<TransactionType.BlobEIP4844>
+  implements EIP4844CompatibleTxInterface<TransactionType.BlobEIP4844>
```
holgerd77, line 83: Here again (and for the other tx types too), I think we should remove.

This blurs again the capability concept.

holgerd77 (changes requested):
Thanks a lot, looks already pretty great respectively forms to a great picture and gets us a lot closer to a modular architecture! 🤩

Some structural comments to address.

packages/tx/src/types.ts
gabrocheleau: That's true, EIP2930 capabilities should be left within their own scope, will adjust.

packages/tx/src/capabilities/eip2718.ts
gabrocheleau: Shortened by removing the unnecessary Interface suffix.

packages/tx/src/capabilities/eip2718.ts
gabrocheleau: Forgot to adjust after refactoring, fixed.

packages/tx/src/types.ts
gabrocheleau: Good point. Added a LegacyTxInterface, makes this more consistent and complete.

packages/tx/src/eip1559Transaction.ts
gabrocheleau: Yeah, I actually went back and forth on this. I agree about the point concerning multiple inheritance, but since all of the interfaces are nested within each other (eip1559 inherits from eip2930, and eip4844 inherits from eip1559), I figured this wasn't a problem. However I agree that this would not be "correct", especially if we want to treat capabilities as independent from one another.

There are no type errors even if we remove this `ìmplements` (typescript automatically figures out whether or not the class fits the necessary interface when we use the Capabilities, regardless of if we explicitly mention that it `implements` said interface. I'll go ahead and remove this.

packages/tx/src/eip4844Transaction.ts
gabrocheleau: reomved

gabrocheleau:
Everything should be addressed @holgerd77, can be re-reviewed! :)

holgerd77 (approved):

holgerd77 (approved):
Thanks a lot, this looks really fantastic 😍, I didn't imagine this to turn out SO structurally clean on all fronts, totally love this! 🤩

Will merge.

PR #3011, opened 2023-09-05
VM: Small Profiler UX improvements / Fix client reportAfterTx option propagation to EVM

holgerd77:
PR includes some small improvements to the VM block/tx profiler UX to better locate on what blocks/txs profiler is applicable and has been running.

Additionally `reportAfterTx` option has been fixed to now propagate correctly to the EVM, since before the nullish coalescent operator chaining was not working correctly and the option omitted. This also affects the `--vmProfileTxs` CLI option in the client.

### DIFF
--- a/packages/vm/src/vm.ts
+++ b/packages/vm/src/vm.ts
@@ -127,8 +127,13 @@ export class VM {
     if (opts.evm) {
       this.evm = opts.evm
     } else {
-      const enableProfiler =
-        this._opts.profilerOpts?.reportAfterBlock ?? this._opts.profilerOpts?.reportAfterTx ?? false
+      let enableProfiler = false
+      if (
+        this._opts.profilerOpts?.reportAfterBlock === true ||
+        this._opts.profilerOpts?.reportAfterTx === true
+      ) {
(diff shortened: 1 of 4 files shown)


packages/vm/src/vm.ts
```diff
@@ -127,8 +127,13 @@ export class VM {
     if (opts.evm) {
       this.evm = opts.evm
     } else {
-      const enableProfiler =
-        this._opts.profilerOpts?.reportAfterBlock ?? this._opts.profilerOpts?.reportAfterTx ?? false
+      let enableProfiler = false
+      if (
+        this._opts.profilerOpts?.reportAfterBlock === true ||
+        this._opts.profilerOpts?.reportAfterTx === true
```
jochem-brouwer, line 133: Aha! I just learned that I have interpreted `??` wrong, I thought it would continue the chain if the left statement was `false`. But it is not.

jochem-brouwer (approved):
LGTM!

Learned some more about `??` operator here :)

holgerd77:
Admin merged after exchanged with Jochem (unrelated test failures)

PR #3016, opened 2023-09-06
Bug fix and LMDB benchmark

scorbajio:
This change adds benchmarks to the `trie` package for using LMDB as the backing db for a trie. It also fixes a race condition that was discovered when running the client tests at commit hash `ce1a289948f05f0b944b7261f13f6e955dd1c441` where LMDB is being used as the `stateDB` for the client and tests in `test/rpc/eth/getTransactionReceipt.spec.ts` fail because `saveReceipts` calls are not being awaited in `vmexecution`.


holgerd77:
I would also close this PR here with the logic that we didn't have the usage of LMDB completely right with this async/sync usage and so benchmarks are relatively meaningless.

@scorbajio let me know if you disagree and would rather want to bring this to some close or into the code base in some way or the other.

PR #3017, opened 2023-09-07
Trie: findPath without `slice()`

ScottyPoi:
# Trie
### `findPath`

This PR aims to improve performance of `findPath` by eliminating the use of `slice()` operations, which create a shallow copy.

For every node processed in `findPath`, the `onFound` function would compare the current `keyProgress` with a slice of the orignal target key (a shallow copy made with `slice()`).

This also meant that it would recheck all of the previously checked nibbles with each new key.  This alone adds up to a lot of extra work that doesn't need to be done.

In the case of a BranchNode, a shallow copy would be made using `slice`, which we would only use to look at the first nibble in the array.

--

This PR refactors `findPath` to use destructive operations `splice()` and `shift()` on the `targetKey` array, removing nibbles as `findPath` progresses, rather than using shallow copies with `slice()` 

Both `shift()` and `splice()` delete elements from an array and return the deleted element(s).

When `findPath` travels through a branch node, it will look for a node on `targetKey.shift()`, or return itself if the targetKey has been consumed.

When `findPath` hits an ExtensionNode, it will compare the node's keyNibbles to `targetKey.splice(0, node.key().length)`.

When `findPath` hits a LeafNode, it will compare the `node.key()` to what is left of `targetKey`.

--
This PR improves `findPath` by:
1.  Eliminating the use of `slice()` entirely.
2.  Reducing memory with destructive array transformations
3.  Eliminating redundant nibble comparisons.
 
-- 
**Update 9/27/23**
4. Replaced push calls:  `stack.push(node)` with `stack[progress] = node`

Rather than allocate new memory with a `push` to the stack array, we create the stack array with `length: keyLength` (the maximum possible stack size).

The tradeoff is that it does require the final step of removing `undefined` indices from the array before the final return.

Simply using **npm run benchmarks** with the `trie` does not produce consistent enough results to measure these optimizations, unfortunately.

### DIFF
--- a/packages/trie/src/trie.ts
+++ b/packages/trie/src/trie.ts
@@ -326,9 +326,6 @@ export class Trie {
     let result: Path | null = null
 
     const onFound: FoundNodeFunction = async (_, node, keyProgress, walkController) => {
-      // If we already have a result, exit early
-      if (result) return
-
       stack.push(node!)
 
       if (node instanceof BranchNode) {
--- a/packages/trie/src/trie.ts
+++ b/packages/trie/src/trie.ts
@@ -352,10 +352,10 @@ export class Trie {
         )
 
       if (node instanceof BranchNode) {
-        if (keyRemainder.length === 0) {
+        if (targetKey.length === 0) {
           result = { node, remaining: [], stack }
         } else {
-          const branchIndex = keyRemainder[0]
+          const branchIndex = targetKey.shift()!
           this.DEBUG &&
             this.debug(`Looking for node on branch index: [${branchIndex}]`, [
               'FIND_PATH',
@@ -372,16 +372,16 @@ export class Trie {
               ['FIND_PATH', 'BranchNode', branchIndex.toString()]
             )
           if (!branchNode) {
-            result = { node: null, remaining: keyRemainder, stack }
+            result = { node: null, remaining: [branchIndex, ...targetKey], stack }
           } else {
             walkController.onlyBranchIndex(node, keyProgress, branchIndex)
           }
         }
       } else if (node instanceof LeafNode) {
-        if (doKeysMatch(keyRemainder, node.key())) {
+        if (doKeysMatch(targetKey, node.key())) {
           result = { node, remaining: [], stack }
         } else {
-          result = { node: null, remaining: keyRemainder, stack }
+          result = { node: null, remaining: targetKey, stack }
         }
       } else if (node instanceof ExtensionNode) {
         this.DEBUG &&
@@ -395,11 +395,12 @@ export class Trie {
             `,
             ['FIND_PATH', 'ExtensionNode']
           )
-        if (doKeysMatch(keyRemainder.slice(0, node.key().length), node.key())) {
+        const consumed = targetKey.splice(0, node.key().length)
+        if (doKeysMatch(consumed, node.key())) {
           this.DEBUG && this.debug(`NextNode: ${node.value()}`, ['FIND_PATH', 'ExtensionNode'])
           walkController.allChildren(node, keyProgress)
         } else {
-          result = { node: null, remaining: keyRemainder, stack }
+          result = { node: null, remaining: [...consumed, ...targetKey], stack }
         }
       }
     }
--- a/packages/trie/src/trie.ts
+++ b/packages/trie/src/trie.ts
@@ -329,26 +329,7 @@ export class Trie {
       // If we already have a result, exit early
       if (result) return
 
-      if (node === null) {
-        result = { node: null, remaining: [], stack }
-        return
-      }
-
-      this.DEBUG &&
-        this.debug(
-          `${node.constructor.name} FOUND${'_nibbles' in node ? `: [${node._nibbles}]` : ''}`,
-          ['FIND_PATH', keyProgress.toString()]
-        )
-
-      this.DEBUG && this.debug(`[${keyRemainder}] Remaining`, ['FIND_PATH', keyProgress.toString()])
-      stack.push(node)
-      this.DEBUG &&
-        this.debug(
-          `Adding ${node.constructor.name} to STACK (size: ${stack.length}) ${stack.map(
-            (e, i) => `${i === stack.length - 1 ? '\n|| +' : '\n|| '}` + e.constructor.name
-          )}`,
-          ['FIND_PATH']
-        )
+      stack.push(node!)
 
       if (node instanceof BranchNode) {
         if (targetKey.length === 0) {
--- a/packages/trie/src/trie.ts
+++ b/packages/trie/src/trie.ts
@@ -26,7 +26,7 @@ import { verifyRangeProof } from './proof/range.js'
 import { ROOT_DB_KEY } from './types.js'
 import { _walkTrie } from './util/asyncWalk.js'
 import { Lock } from './util/lock.js'
-import { bytesToNibbles, doKeysMatch, matchingNibbleLength } from './util/nibbles.js'
+import { bytesToNibbles, matchingNibbleLength } from './util/nibbles.js'
 import { TrieReadStream as ReadStream } from './util/readStream.js'
 import { WalkController } from './util/walkController.js'
 
@@ -323,16 +323,17 @@ export class Trie {
     const stack: TrieNode[] = []
     const targetKey = bytesToNibbles(key)
     this.DEBUG && this.debug(`Target (${targetKey.length}): [${targetKey}]`, ['FIND_PATH'])
+    const keyLen = targetKey.length
     let result: Path | null = null
-
+    let progress = 0
     const onFound: FoundNodeFunction = async (_, node, keyProgress, walkController) => {
       stack.push(node!)
 
       if (node instanceof BranchNode) {
-        if (targetKey.length === 0) {
+        if (progress === keyLen) {
           result = { node, remaining: [], stack }
         } else {
-          const branchIndex = targetKey.shift()!
+          const branchIndex = targetKey[progress]
           this.DEBUG &&
             this.debug(`Looking for node on branch index: [${branchIndex}]`, [
               'FIND_PATH',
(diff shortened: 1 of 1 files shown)


acolytec3:
Have you found any notable difference in performance?
I get these results on this branch from the benchmarks:
```
Benchmarking
1k-3-32-ran x 3 ops/sec @ 314ms/op
1k-5-32-ran x 2 ops/sec @ 354ms/op
1k-9-32-ran x 2 ops/sec @ 350ms/op
1k-1k-32-ran x 2 ops/sec @ 352ms/op
1k-1k-32-mir x 2 ops/sec @ 359ms/op
Checkpointing: 100 iterations x 3,717 ops/sec @ 268μs/op ± 28.11% (min: 76μs, max: 3ms)
Checkpointing: 500 iterations x 3,155 ops/sec @ 316μs/op ± 11.67% (min: 177μs, max: 6ms)
Checkpointing: 1000 iterations x 2,794 ops/sec @ 357μs/op ± 6.43% (min: 195μs, max: 4ms)
Checkpointing: 5000 iterations x 2,164 ops/sec @ 462μs/op ± 3.63% (min: 206μs, max: 10ms)
RAM: rss=647.2mb heap=536.9mb used=505.8mb ext=3.8mb arr=2.7mb
1k-3-32-ran x 0 ops/sec @ 4919ms/op
1k-5-32-ran x 2 ops/sec @ 369ms/op
1k-9-32-ran x 2 ops/sec @ 410ms/op
1k-1k-32-ran x 2 ops/sec @ 404ms/op
1k-1k-32-mir x 2 ops/sec @ 433ms/op
Checkpointing: 100 iterations x 3,542 ops/sec @ 282μs/op ± 15.92% (min: 97μs, max: 2ms)
Checkpointing: 500 iterations x 2,703 ops/sec @ 369μs/op ± 4.99% (min: 210μs, max: 2ms)
Checkpointing: 1000 iterations x 2,436 ops/sec @ 410μs/op ± 3.39% (min: 250μs, max: 2ms)
Checkpointing: 5000 iterations x 2,094 ops/sec @ 477μs/op ± 1.56% (min: 267μs, max: 7ms)
RAM: rss=399.5mb heap=248.0mb used=169.8mb ext=57.4mb arr=55.7mb
```

And this on master
```
Benchmarking
1k-3-32-ran x 3 ops/sec @ 315ms/op
1k-5-32-ran x 3 ops/sec @ 317ms/op
1k-9-32-ran x 2 ops/sec @ 336ms/op
1k-1k-32-ran x 2 ops/sec @ 343ms/op
1k-1k-32-mir x 2 ops/sec @ 375ms/op
Checkpointing: 100 iterations x 4,118 ops/sec @ 242μs/op ± 24.08% (min: 81μs, max: 3ms)
Checkpointing: 500 iterations x 2,948 ops/sec @ 339μs/op ± 9.75% (min: 164μs, max: 4ms)
Checkpointing: 1000 iterations x 2,578 ops/sec @ 387μs/op ± 6.37% (min: 197μs, max: 5ms)
Checkpointing: 5000 iterations x 2,178 ops/sec @ 458μs/op ± 3.43% (min: 208μs, max: 6ms)
RAM: rss=638.6mb heap=539.5mb used=505.8mb ext=5.5mb arr=4.4mb
1k-3-32-ran x 0 ops/sec @ 4896ms/op
1k-5-32-ran x 2 ops/sec @ 357ms/op
1k-9-32-ran x 2 ops/sec @ 393ms/op
1k-1k-32-ran x 2 ops/sec @ 408ms/op
1k-1k-32-mir x 2 ops/sec @ 398ms/op
Checkpointing: 100 iterations x 3,527 ops/sec @ 283μs/op ± 14.53% (min: 92μs, max: 1ms)
Checkpointing: 500 iterations x 2,873 ops/sec @ 347μs/op ± 5.51% (min: 202μs, max: 2ms)
Checkpointing: 1000 iterations x 2,469 ops/sec @ 404μs/op ± 3.98% (min: 243μs, max: 3ms)
Checkpointing: 5000 iterations x 2,095 ops/sec @ 477μs/op ± 1.48% (min: 246μs, max: 7ms)
RAM: rss=389.0mb heap=238.3mb used=179.4mb ext=58.5mb arr=57.4mb
```

which doesn't feel like a real visible difference anywhere

ScottyPoi:
> Have you found any notable difference in performance? I get these results on this branch from the benchmarks:


No, not really.  I get about the same.

acolytec3:
I spent some time trying to think through it but couldn't figure out exactly how but is there a way to just pass the branch index or else an index that indicates the starting position of the remaining key instead of doing any array operations when working through the smaller bits of the key? I suspect that slicing an array vs destructively manipulating it are both big-ish performance drains. If you just take the key and pass around a reference to it along with a number that indicates the index you're starting from as you walk down child nodes, that would probably be faster, at least to my way of thinking.

Sep 6, 2023 11:49:58 PM Scotty ***@***.***>:

holgerd77:
Thanks for the tests here, while might not directly lead to results definitely interesting experiments. 👀

ScottyPoi:
> I spent some time trying to think through it but couldn't figure out exactly how but is there a way to just pass the branch index or else an index that indicates the starting position of the remaining key instead of doing any array operations when working through the smaller bits of the key? I suspect that slicing an array vs destructively manipulating it are both big-ish performance drains. If you just take the key and pass around a reference to it along with a number that indicates the index you're starting from as you walk down child nodes, that would probably be faster, at least to my way of thinking.

Yeah - that was my other idea. I'll put that together and compare benchies

ScottyPoi:
New solution in latest commit uses a `progress` tracker to track to current index of the search.
---- 
```
-------
Benchmarking
-------
Benchmarking
1k-3-32-ran x 2 ops/sec @ 407ms/op
1k-5-32-ran x 2 ops/sec @ 380ms/op
1k-9-32-ran x 2 ops/sec @ 401ms/op
1k-1k-32-ran x 2 ops/sec @ 429ms/op
1k-1k-32-mir x 1 ops/sec @ 550ms/op
Checkpointing: 100 iterations x 3,816 ops/sec @ 262μs/op ± 29.04% (min: 112μs, max: 4ms)
Checkpointing: 500 iterations x 2,369 ops/sec @ 422μs/op ± 13.43% (min: 176μs, max: 7ms)
Checkpointing: 1000 iterations x 2,330 ops/sec @ 429μs/op ± 8.34% (min: 209μs, max: 7ms)
Checkpointing: 5000 iterations x 1,805 ops/sec @ 553μs/op ± 4.28% (min: 221μs, max: 10ms)
RAM: rss=631.5mb heap=537.7mb used=504.0mb ext=6.8mb arr=4.3mb
1k-3-32-ran x 0 ops/sec @ 5985ms/op
1k-5-32-ran x 2 ops/sec @ 417ms/op
1k-9-32-ran x 2 ops/sec @ 438ms/op
1k-1k-32-ran x 2 ops/sec @ 442ms/op
1k-1k-32-mir x 1 ops/sec @ 500ms/op
Checkpointing: 100 iterations x 3,569 ops/sec @ 280μs/op ± 16.80% (min: 89μs, max: 2ms)
Checkpointing: 500 iterations x 2,630 ops/sec @ 380μs/op ± 5.91% (min: 191μs, max: 2ms)
Checkpointing: 1000 iterations x 2,261 ops/sec @ 442μs/op ± 4.60% (min: 235μs, max: 3ms)
Checkpointing: 5000 iterations x 1,836 ops/sec @ 544μs/op ± 1.92% (min: 247μs, max: 7ms)
RAM: rss=454.2mb heap=306.4mb used=181.7mb ext=60.7mb arr=57.6mb
```

acolytec3:
This change certainly "feels" like it should be faster.  Nice work on figuring out this variant.  I wish I could figure out why it's not actually faster

packages/trie/src/trie.ts
```diff
@@ -280,45 +280,49 @@ export class Trie {
...
-      if (result) return
-
-      if (node === null) {
-        result = { node: null, remaining: [], stack }
-        return
-      }
-
-      const keyRemainder = targetKey.slice(matchingNibbleLength(keyProgress, targetKey))
-      stack.push(node)
+      stack.push(node!)
```
acolytec3, line 287: The only other thing I can come up with for potential performance improvement is to rework this bit here.  In theory, the stack could be created at the maximum possible length for the path of the node, so whatever the depth of the trie is (maybe 16)?  This `push` operation is still another destructive/constructive array operation that theoretically is more expensive than just storing something in the array since we're re-allocating memory here. So theoretically, you could pre-allocate the `stack` to be the maximum possible length, add items to it as you traverse the path of the key, and then return a slice of the path array equivalent to the length of the key, right?  That's the only other thing I can come up with unless @holgerd77 or @jochem-brouwer have other ideas.

ScottyPoi: Are you saying that `stack.push(node)` is more expensive than if we start with  `stack = new Array(32)` and add via  `stack[i] = node` 

How would we know where to put a node in the stack array?

Somewhere along the way we will do:

`stack[ n ] = extension_node`

If the extension_node has nibble_length 3...

Would the next node in the stack belong at `stack[ n + 1 ]` or `stack[ n + 3 ]`?

To find `n + 1` we have to keep track of the index, and then slice the return stack.

`n + 3` would be equal to `progress` -- 
However we would have to iterate over the whole return stack to filter out `undefined`.,,

acolytec3: I'm not 100% sure on the first point but push is having to allocate more memory every time you add to it and I think just copying a value is faster than allocating new memory and then copying it. Will need to do some simple tests to verify though. Let me think on it some more. We can always take this bit by bit

ScottyPoi:
@acolytec3 --

I tried what you suggested with the stack array.
Here are the benchmarks (commit 3e2ab27).



```
Benchmarking
1k-3-32-ran x 3 ops/sec @ 307ms/op
1k-5-32-ran x 2 ops/sec @ 342ms/op
1k-9-32-ran x 2 ops/sec @ 419ms/op
1k-1k-32-ran x 2 ops/sec @ 386ms/op
1k-1k-32-mir x 2 ops/sec @ 401ms/op
Checkpointing: 100 iterations x 4,183 ops/sec @ 239μs/op ± 19.71% (min: 92μs, max: 2ms)
Checkpointing: 500 iterations x 2,729 ops/sec @ 366μs/op ± 12.07% (min: 168μs, max: 5ms)
Checkpointing: 1000 iterations x 2,722 ops/sec @ 367μs/op ± 7.53% (min: 183μs, max: 5ms)
Checkpointing: 5000 iterations x 2,104 ops/sec @ 475μs/op ± 4.00% (min: 198μs, max: 8ms)
RAM: rss=629.6mb heap=530.2mb used=501.8mb ext=5.8mb arr=3.4mb
1k-3-32-ran x 0 ops/sec @ 5167ms/op
1k-5-32-ran x 2 ops/sec @ 379ms/op
1k-9-32-ran x 2 ops/sec @ 395ms/op
1k-1k-32-ran x 2 ops/sec @ 428ms/op
1k-1k-32-mir x 2 ops/sec @ 442ms/op
Checkpointing: 100 iterations x 2,845 ops/sec @ 351μs/op ± 13.25% (min: 112μs, max: 2ms)
Checkpointing: 500 iterations x 2,634 ops/sec @ 379μs/op ± 7.60% (min: 177μs, max: 3ms)
Checkpointing: 1000 iterations x 2,488 ops/sec @ 401μs/op ± 4.03% (min: 232μs, max: 2ms)
Checkpointing: 5000 iterations x 2,055 ops/sec @ 486μs/op ± 1.88% (min: 232μs, max: 3ms)
```

ScottyPoi:
Refreshed this PR and reopened as #3066

PR #3019, opened 2023-09-08
Trie: Debug logging

ScottyPoi:
This introduces `debug` logging to the `Trie` class.

Like other packages, all `debug` calls will be wrapped inside of a conditional:
`if (this.DEBUG) { ... } `

`this.DEBUG` will be set to `true` only if the `ethjs` flag has been set in `process.env.DEBUG`

If active, the debugger will log with the namespace: `ethjs:Trie:`

The debugger is available by calling `this.debug(message: string, namespaces: string[])`

`trie.debug()`  will take 1 or 2 parameters.  The first being the string to be logged, the second being a list of added namespaces for that specific message.

So calling `trie.debug('Trie Created')`  will log 
> // **ethjs:trie:**  Trie Created

And calling `trie.debug('Adding value for key: 0xabcd', ['PUT'] )`  will log:
> // **ethjs:trie:PUT** Adding value for key: 0xabcd

----

Also included is `test/util/log.spec.ts`

This runs a minimal trie script with `this.DEBUG` set to true.
The debugger will remain inactive, but `coverage` will touch the code inside of the DEBUG conditionals.
This way the added lines of code for logging do not artificially hurt the `coverage` statistic.

### DIFF
--- a/packages/trie/README.md
+++ b/packages/trie/README.md
@@ -265,6 +265,62 @@ npm run profiling
 
 0x processes the stacks and generates a profile folder (`<pid>.0x`) containing [`flamegraph.html`](https://github.com/davidmarkclements/0x/blob/master/docs/ui.md).
 
+## Debugging
+
+The `Trie` class features optional debug logging.. Individual debug selections can be activated on the CL with `DEBUG=ethjs,[Logger Selection]`.
+
+`ethjs` **must** be included in the `DEBUG` environment variables to enable **any** logs.
+Additional log selections can be added with a comma separated list (no spaces). Logs with extensions can be enabled with a colon `:`, and `*` can be used to include all extensions.
+
+`DEBUG=ethjs,thislog,thatlog,otherlog,otherlog:sublog,anotherLog:* node myscript.js`
+
+The following options are available:
+
+| Logger            | Description                                    |
+| ----------------- | ---------------------------------------------- |
+| `trie`            | minimal info logging for all trie methods      |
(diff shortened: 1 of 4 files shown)


ScottyPoi:
[trieLogs.pdf](https://github.com/ethereumjs/ethereumjs-monorepo/files/12555032/trieLogs.pdf)

holgerd77:
Thanks a lot, this looks already pretty great! 🤩

What is actually a good and easy way to test (and see) the debug logger?

The following does not yet work: `DEBUG=ethjs,trie:* npm run test:node` while an analogue test on statemanager does.

Didn't have such a close look yet, but this example above is actually following our standard naming pattern for debug logging. So if this is not yet working this should be aligned.

Also, please add a quick README docs section like e.g. for blockchain:
https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/blockchain#developer

holgerd77:
Ah, ok, got it, you took the namespace `ethjs:trie:*`. Yeah, that should be updated to the `ethjs,trie:*` version, since this otherwise don't matches the other libraries.

Decision to have this comma-version was I think to have things shorter when activating several dedicated debuggers from the same library (e.g. `ethjs,devp2p:rlpx:*,devp2p:eth:*`, something like that).

And so we should keep this consistent over the libs.

holgerd77:
Updated this via UI

holgerd77:
Updated this via UI

holgerd77 (approved):
Very cool, great PR and docs 🤩! Will merge.

Issue #3021, opened 2023-09-08
StateManager: add (persistent) Code Cache


holgerd77:
StateManager (SM) caches have gotten a large rewrite in #2630 (maybe not such a good point of reference since a bit intimidating due to the sheer size of code changes, so just for eventually getting some loose overview on code parts touched eventually), followed by several PRs with bug fixes.

This work introduced more intelligent caches for accounts and storage data, which can be used in a persistent way in the client, reaching over the span of a single tx or block. Caches are now also more resistent to attacks like where a lot of data (e.g.: new accounts) is added throughout EVM execution and at the end things are reverted. This is by adding a new `diffCache` layer which only stores the proper differences between StateManager checkpoints and on a commit things are persistet in the upper layer and on a revert things are thrown away.

The respective cache tests in SM provide a lot of scenarios and examples in which the diff cache is used.

While we now have these two proper caches, one cache is still not updated and exists still in its rudimentary form in `stateManager.ts`: the code cache.

This issue is for triggering a change here. 😋

The code cache should be added in a similar way like the other two caches, so that this complements our cache suite, and then so having a new `cache/code.ts` or so file with the cache which is being called into from `stateManager.ts` and (I guess?) EthersStateManager as well in a second round.

The account cache should be structurally more similar than the storage cache and is therefore likely best suited as a boilerplate for the new cache (storage cache is more complex).

So this cache element can likely remain very much unchanged (apart from renaming things obviously):

```typescript
type AccountCacheElement = {
  accountRLP: Uint8Array | undefined
}
```

One main thing to notice is the double role of `undefined` in this definition together with the two cache typings:

```typescript
_lruCache: LRUCache<string, AccountCacheElement> | undefined
_orderedMapCache: OrderedMap<string, AccountCacheElement> | undefined
```

`undefined` in the above type means: for this account the cache in the fitting point in time knows that the element is *not in the state* (and if that respective account is accessed with `getAccount()` or something `undefined` can be returned).

`undefined` for caches typings below means (particularly in diff cache context): at this point in time (so, e.g.: after checkpoint number x) it is known that the element was not in the cache (therefore: the cache - at that point - don't know anything about this account (and a respective account access like the above would need to trigger a trie read).

To understand this difference is somewhat key for the implementation.

[...]

PR #3022, opened 2023-09-11
Implement a more performant code cache

scorbajio:
This PR looks into implementing a more performant code cache for the `statemanager`. Also see #3021.

Some followup work that is left to do after this change is merged:
- [ ] ~~Refactor the `statemanager` caches to reduce code redundancies~~
- [x] Add an option to cli for enabling/disabling the code cache, similar to the other caches
- [ ] ~~Integrate the code cache with `EthersStateManager`~~


jochem-brouwer:
This commit: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3022/commits/105403ab4c143f9aa6e7a224e063552b62dd1505 should fix (some?) of the failing tests.

I have tried to merge master (this had merge conflicts) - not sure if I solved those correctly (if not, rollback to https://github.com/ethereumjs/ethereumjs-monorepo/pull/3022/commits/105403ab4c143f9aa6e7a224e063552b62dd1505)

jochem-brouwer:
Reverted back to old state due to too many tests failing (note: CI run is thus not up to date with current state since it will not re-run due to merge conflicts)

packages/statemanager/src/cache/code.ts
```diff
@@ -0,0 +1,277 @@
...
+      }
+      this._diffCache[this._checkpoints].set(cacheKeyHex, oldElem ?? { code: val })
+    }
+  }
+
+  /**
+   * Puts code into the cache under its hash.
+   * @param codeHash - Hash of the code.
+   * @param code - Bytecode or undefined if code doesn't exist.
+   */
```
jochem-brouwer, line 79: TODO: docs. I am not entirely sure what `currentCode` does as opposed to `code`.

packages/statemanager/src/cache/code.ts
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    this._diffCache.push(new Map<string, CodeCacheElement | undefined>())
+    this._debug = createDebugLogger('statemanager:cache:code')
+  }
+
+  _saveCachePreState(
+    cacheKeyHex: string,
+    currentPutCode?: Uint8Array | undefined,
+    currentCode?: Uint8Array | undefined
```
jochem-brouwer, line 54: Also maybe add docs here?

packages/statemanager/src/stateManager.ts
```diff
@@ -235,6 +236,11 @@ export class DefaultStateManager implements EVMStateManagerInterface {
         type: this._storageCacheSettings.type,
       })
     }
+
+    this._codeCache2 = new CodeCache({
+      size: 20000,
+      type: CacheType.ORDERED_MAP,
```
jochem-brouwer, line 242: Type is always a `ORDERED_MAP`, do we need LRU cache support in code cache then?

packages/statemanager/src/stateManager.ts
```diff
@@ -155,6 +155,7 @@ export class DefaultStateManager implements EVMStateManagerInterface {
   protected _debug: Debugger
   protected _accountCache?: AccountCache
   protected _storageCache?: StorageCache
+  protected _codeCache2?: CodeCache
```
jochem-brouwer, line 158: Rename to `_codeCache`?

packages/statemanager/src/stateManager.ts
```diff
@@ -335,21 +344,17 @@ export class DefaultStateManager implements EVMStateManagerInterface {
    * @param value - The value of the `code`
    */
   async putContractCode(address: Address, value: Uint8Array): Promise<void> {
+    const currentCode = await this.getContractCode(address)
+    this._codeCache2?.put(address, value, currentCode)
```
jochem-brouwer, line 348: Hmm, interesting. It is not possible that `currentCode` is non-empty and then we try to put a different `value` there (it is not possible to overwrite contracts). Can we optimize further here? :thinking:

jochem-brouwer (commented):
Some comments and questions.

packages/statemanager/src/cache/code.ts
scorbajio: fixed

packages/statemanager/src/cache/code.ts
scorbajio: Fixed. Let me know if I can further clarify.

packages/statemanager/src/stateManager.ts
scorbajio: I removed the old cache and renamed the new one in its place.

packages/statemanager/src/stateManager.ts
scorbajio: Yes, I simplified it a bit since all we care about is if the account has a contract or not, which is used for reverting code.

packages/statemanager/src/stateManager.ts
scorbajio: It's the same with the account and storage caches too. Although the value is defaulted, the cache type option is given to the user. Furthermore, all of the caches will be refactored and unified in a future PR, so I'm not opposed to dropping support and only using e.g. `ORDERED_MAP` if it outperforms `LRU` in most/all relevant use cases.

holgerd77: These are two datastructures with different properties, we need both, also for the code cache (so this should be added here as well).

`ORDERED_MAP` performs somewhat better, but is not bound in size, it also works better (at all) for smaller caches. `LRU_CACHE` manages the cache size and drops out older items and the like (this is where the overhead is coming from). We use `LRU` in the client, since if we would not caches would grow indefinitely and the client will go out of memory. `ORDERED_MAP` in return is used in short-lived contexts like the VM blockchain and state tests.

Also just a a word of caution: I have "unified" state and account cache as much as I ever could when I reworked the caches early this year, so many of the design decisions are relatively strongly grounded. Doesn't mean things can not be improved likely, but let's please align closely here before new refactorings of the caches are started. Best to open one or some issues on that clearly describing what you plan to do.

holgerd77: So code cache should also get an LRU option and the client should instantiate the cache with this option and then there should be also an additional flag `codeCache` for the client analog to e.g. `accountCache` or `storageCache`.

holgerd77:
Cool, but this looks great in general! 🤩

packages/statemanager/src/stateManager.ts
jochem-brouwer: Regarding the refactor of the cache: I think both storage/account (and now code) cache look rather similar, so if we could combine them in a base class and then further pick out any differences for the specific caches then this means we could re-use the cache mechanism in the future if we need a new cache somewhere :smile:

Regarding `LRU_CACHE`: ah great, OK we definitely need this for client.

holgerd77:
Ok, doing a few client runs here:

1. `npm run client:start -- --sync=none --transports --executeBlocks=1600000-1600100`
First: works at all, that's great (mean this totally unironically, since for me a lot of things still broke along client runs even I had the tests going (for the other-caches-implementation)! 🤩
Time:(for such a short range period there should be no difference I would assume?)
Seconds for every repeated run: 32 25 24 (often a first run is slower, wonder if this is because Node is then caching some code or whatever? This is a very general observation I have repeatedly)

Oh.

But the code cache is also not building up any size. 🤔 If I inject `console.log(this._codeCache?.size())` in `StateManager` `getContractCode()`, this is always `0`.

I guess there is still something missing on initialization?

Will submit here and have a look.

holgerd77:
Have updated the branch via UI

holgerd77:
Update: ah, I think this is due to the exact same thing which I stumbled upon and am solving here https://github.com/ethereumjs/ethereumjs-monorepo/pull/3063 , so that the --executeBlocks option in client is not properly initialized with the caches.

Will wait until this is merged and then update the branch here and then try again (testing this without the --executeBlocks option is basically not possible, since one then cannot get repeated runs on the same (higher-number) block ranges).

holgerd77:
Ah, and then we'll also still need to see if the cache "holds" for live-higher-range-client blocks.

This is always an exciting moment. 🙂 😋
(I would guess 50-50 chance from my own experiences...)

holgerd77:
Ok, I now updated this branch with the changes from the PR (remember to pull your local branch), will now test again. 🙂

holgerd77:
Just so that I do not forget:

There now needs to be a `downlevelCaches` check in the following code part in `StateManager.shallowCopy()` (as being done for the other caches):

```typescript
if (!this._codeCacheSettings.deactivate) {
      codeCacheOpts = { ...codeCacheOpts, type: CacheType.ORDERED_MAP }
    }
```

Generally code cache is not active yet for the `--executeBlocks` flag still (also with the above change), will have a closer look.

holgerd77:
Haha, it was because the `size()` method is not complete regarding `ORDERED_MAP` 😋:

```typescript
size(): number {
    if (this._lruCache) {
      return this._lruCache.size
    }
    return 0
  }
  ```
  
  Please update, maybe you can have another general look if all methods work for both cache types.

holgerd77:
Ok, great, then for the above example cache grows to 246.

And I have got 588 reads from cache, 4027 from DB.

Interesting, that's something, already for such a relatively small block range.

Needs to stop, unfortunately can't do any more performance comparisons, but will continue to experiment tomorrow a bit.

holgerd77:
Updated this via UI

holgerd77:
Some more performance results, if several values this is for consecutive repetitions of runs:

npm run client:start -- --sync=none --executeBlocks=1600000-1600100 (100 blocks)
Seconds (total): 24 24 24

npm run client:start -- --sync=none --executeBlocks=1600000-1600100 --codeCache=0
25 24 25

npm run client:start -- --sync=none --executeBlocks=1600000-1601000 (1000 blocks)
Minutes (total): 2:07  1:55  1:54

npm run client:start -- --sync=none --executeBlocks=1600000-1601000 --codeCache=0
Minutes (total): 1:55  1:54. 1:57

Ok, this for the most part matches. One outlier with this 2:07, but this can very well be the first run.

holgerd77 (approved):
LGTM

PR #3023, opened 2023-09-12
Trie/StateManager: Prefix Storage Trie Node Keys with Account Hash Substring

holgerd77:
LevelDB performance for sequential reads are significantly faster than random reads, see [here](http://www.lmdb.tech/bench/microbench/benchmark.html).

This PR tests to store the storage trie subtree nodes with a shared prefix based on the account hash (this should also work in all SNAP contexts, might need some confirmation though).

Since SLOAD opcode executions are often performed sequentially on the same account the hope is that this change significantly improves SLOAD performance.

We'll see. 🙂 

I'll post some performance comparisons here using Jochems EVM profiler.

If this works this at least also needs a constructor option for StateManager to use or not use prefix (default: not use) to remain backwards compatible.


acolytec3:
Just to make sure I'm interpreting this correctly. The first four numbers in each block report are from the PR? If so, this is a remarkable improvement

holgerd77:
> Just to make sure I'm interpreting this correctly. The first four numbers in each block report are from the PR? If so, this is a remarkable improvement

Yes, that's correct. 🙂

g11tech (approved):
amazing perf improvement! :heart:

scorbajio:
I ran the snap sync simulated tests on this branch and the snap sync works as expected.

holgerd77:
> I ran the snap sync simulated tests on this branch and the snap sync works as expected with all tests passing.

Cool, thanks a lot for testing! 🤩

packages/statemanager/test/stateManager.storage.spec.ts
```diff
@@ -20,133 +20,135 @@ describe('StateManager -> Storage', () => {
...
-    it(`should throw on storage values larger than 32 bytes`, async () => {
-      const stateManager = new DefaultStateManager({ storageCacheOpts })
-      const address = Address.zero()
-      const account = createAccount()
-      await stateManager.putAccount(address, account)
-
-      const key = zeros(32)
-      const value = hexToBytes('0x' + 'aa'.repeat(33))
-      try {
+    for (const prefixStorageTrieKeys of [false, true]) {
```
holgerd77, line 23: This is the only changed line here (lol).

Diff is showing up as somewhat crazy. Don't get confused! 😆

holgerd77:
I added an option here for SM so that prefixes can be activated (default: false), this is now ready for review. 🙂 

Some additional exchange with @g11tech on the following points:

1. Since this is breaking (on the DB level for client) and devnet 9 is on the horizon (next week or so), we would like to take this in relatively soon (definitely still during the week or optimally next 1-2 days).
2. We agreed that we would both prefer not to prefix with the full account address bytes since this would make for very long keys and practical benefits are limited at the same time (+ it is not clear if this has negative impact on performance). I have now therefore taken a 7 bytes prefix from the address.

packages/trie/test/index.spec.ts
```diff
@@ -17,336 +17,365 @@ import { bytesToNibbles } from '../src/util/nibbles.js'
...
-
-    it('save a value', async () => {
-      await trie.put(utf8ToBytes('test'), utf8ToBytes('one'))
-    })
-
-    it('should get a value', async () => {
-      const value = await trie.get(utf8ToBytes('test'))
-      assert.equal(bytesToUtf8(value!), 'one')
-    })
+for (const keyPrefix of [undefined, hexToBytes('0x1234')]) {
```
g11tech, line 20: looks like here as well this is the only diff and rest is formatting change :+1:

g11tech (approved):
lgtm

PR #3024, opened 2023-09-13
Client: Fetcher Small Bugfixes and Log Improvements

holgerd77:
This PR enhances the fetcher logs which makes it easier to follow jobs on fetcher running and generally allows for easier debugging. Additionally the PR does some output optimizations (a bit less frequent cache and memory logs e.g.).

Beyond some guards for property access (length) is added on re-enqueuing since this occasionally lead to re-enqueue failures.

Ready for review. 🙂


packages/client/src/sync/fetcher/fetcher.ts
```diff
@@ -330,19 +336,21 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable
...
-        }, skip next job execution.`
+        }, skip job ${jobStr} execution.`
       )
       return false
     }
-    if (job.index > this.processed + this.maxQueue) {
-      this.debug(`Job index greater than processed + max queue size, skip next job execution.`)
+    if (job.index > this.finished + this.maxQueue) {
+      this.debug(`Job index greater than finished + max queue size, skip job ${jobStr} execution.`)
+      return false
```
holgerd77, line 350: Ah, and this is worth to mention as well: this switches from using the processed (before writing to BC) to finished jobs for deciding when to skip.

This avoids (to some extend) that there are too big heaps of yet-to-write-to-chain tasks heaping up (like 16 jobs or so still outstanding). Effect is also not too noticeable though, just gradually helps. No strong opinion and if not wanted we can also switch back.

g11tech (dismissed):
looks gtg :fire: retriggered the client run as it didn't show any apparent reason for it to fail except timeouts

scorbajio (dismissed):
ACK

ScottyPoi:
> looks gtg 🔥 retriggered the client run as it didn't show any apparent reason for it to fail except timeouts

i disagree.  `integration/lightsync.spec.ts` consistently fails for me locally, even with long timeouts.  i do think that it is an effect of these changes.

packages/client/src/sync/fetcher/fetcher.ts
```diff
@@ -225,17 +231,25 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable
...
-      reenqueue = true
+    if (result !== undefined) {
+      if ('length' in (result as any)) {
+        if ((result as any).length === 0) {
+          resultSet = 'empty'
+          reenqueue = true
+        }
+      } else {
+        resultSet = 'unknown'
+        reenqueue = true
```
ScottyPoi, line 251: This line here is what breaks `integration/lightsync.spec.ts`

ScottyPoi: the test is fixed by deleting `reenqueue = true`

ScottyPoi: the "unknown" result that the test deals with looks like:
```
{
    reqId: BigInt,
    bv: BigInt,
    headers: BlockHeader[]
}
```

ScottyPoi: Which is a `BlockHeaderResult` from the `HeaderFetcher`.

So the test is also fixed like:

```
 } else if ('headers' in (result as any)) {
        // is a header fetcher job
      } else {
        resultSet = 'unknown'
        reenqueue = true
      }
```

ScottyPoi: Actually --  `result` only ever shows up in the integration tests as `undefined` or as a `BlockHeaderResult`.

So we could delete the whole `if (result !== undefined)` bit all together and and all of our tests would pass.

The timeline on that file only goes back as far as 3 months ago when `lib` became `src`.  I assume that it was written that way for a good reason, and that we just aren't testing it?  Or it is totally unnecessary :shrug:

ScottyPoi: Ok, I tracked the `if (result as any).length === 0` line to a commit by ryan from a couple years ago called **sync stability fixes** #1543.

holgerd77:
Oh, this was actually a nice PR, didn't realize that we never get this merged. 😳

We should really try to get into a merge ready state, this makes debugging the fetcher a lot more fun!

holgerd77:
(actually this is half a note to self)

packages/client/src/sync/fetcher/fetcher.ts
holgerd77: Hmm, the guide should here what behavior we would want "in reality" and not what fixes the tests. But you are correct, if the `reenqueue` line is deleted, the tests pass again.

For devp2p communication it is always good to try to make robust choices. So if the other side delivers something strange, just have a case for that. Here it would be that the structure `(result as any)` does not have a `length` property. The logic I applied here was that this would mean that the other side delivered something unepected (useless) which should therefore be thrown away and the task reenqueued (since it didn't resolve in a successful way). I would still think this should make sense. 🤔

Will have a look into the test case, maybe the assumptions are wong there. Also, the lightclient part of the client is really not maintained (and not even working). Maybe the conclusion here might also just be to throw these tests away.

holgerd77: Have now added a lightsync specific hotfix in form of an extra conditional with a comment to eventually remove in the future.

packages/client/src/sync/fetcher/fetcher.ts
```diff
@@ -226,17 +232,30 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable
...
     let resultSet = ''
     if (result === undefined) {
       resultSet = 'undefined'
       reenqueue = true
     }
-    if (result !== undefined && (result as any).length === 0) {
-      resultSet = 'empty'
-      reenqueue = true
+    if (result !== undefined) {
+      if ('length' in (result as any)) {
```
acolytec3, line 245: Any objection to re-writing this to something like `if (result !== null and result.length !== undefined)`?  I know its pedantic but I really don't like `as any` anywhere if we can help it.

packages/client/src/sync/fetcher/fetcher.ts
```diff
@@ -254,6 +273,7 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable
       job.result = this.process(job, result)
       jobStr = this.jobStr(job, true)
       if (job.result) {
```
acolytec3, line 275: not sure why the linter isn't complaining here but this should be `if (job.result !== undefined)`
```suggestion
      if (job.result !== undefined) {
```

packages/client/src/sync/fetcher/fetcher.ts
```diff
@@ -321,7 +341,7 @@ export abstract class Fetcher<JobTask, JobResult, StorageItem> extends Readable
       if (this.finished !== this.total) {
```
acolytec3, line 341: Not this line but the one above should be `if (job === undefined)`

packages/client/src/sync/fetcher/fetcher.ts
acolytec3: I like this.  Makes more sense in context.

acolytec3 (approved):
LGTM

acolytec3 (approved):
LGTM

PR #3026, opened 2023-09-14
StateManager: some devDependencies move to dependencies

Hann:
Hi I am using hardhat with yarn berry.
I met this error a few days ago whenever i try to compile and test. 

```
Error: @nomicfoundation/ethereumjs-statemanager tried to access @nomicfoundation/ethereumjs-trie, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

Required package: @nomicfoundation/ethereumjs-trie
Required by: @nomicfoundation/ethereumjs-statemanager@npm:2.0.2 (via <my-repository-path>/.yarn/cache/@nomicfoundation-ethereumjs-statemanager-npm-2.0.2-5465bc8cbc-3ab6578e25.zip/node_modules/@nomicfoundation/ethereumjs-statemanager/dist/)
```

it means that some packages are using as "dependencies" but these declared as "devDependencies".
i think this pr is a very tiny thing, but it would help for lots of yarn users 😂

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15724,7 +15724,8 @@
       },
       "devDependencies": {
         "@ethereumjs/block": "^5.0.0",
-        "@ethereumjs/genesis": "^0.1.0"
+        "@ethereumjs/genesis": "^0.1.0",
+        "@types/debug": "^4.1.8"
       }
     },
     "packages/statemanager/node_modules/lru-cache": {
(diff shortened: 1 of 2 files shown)


acolytec3:
Thanks for the PR!  Definitely one of those things we missed along the way.

acolytec3 (approved):
LGTM.  Will merge pending CI

PR #3028, opened 2023-09-15
client: decouple skeleton from beacon sync

g11tech:
skeleton needs to be shared between beacon sync and snap sync to keep the engine api working as well as to keep updating snapsync with latest stateroot.

this PR decouples skeleon from beacon sync


holgerd77:
Updated this via UI

acolytec3:
I updates a fair amount of the typing in the tests that were changed.  Looks good overall but would be nice to finish cleaning up the typing in the tests before we merge.

g11tech:
@acolytec3  i could do separate PR for increasing the typing of the sim utils as followup

acolytec3 (approved):
LGTM

holgerd77:
Still wonder if we could (and would want to) transfer skeleton to become a sub entity or similar of Blockchain? 🤔

PR #3029, opened 2023-09-15
Hybrid EVM Stack Experiment

holgerd77:
This is a pretty extensive experiment switching the current BigInt based EVM stack with a hybrid stack which stores either the Bytes or the BigInt values and - if matches - serves the respetive values without conversion.

Results are unfortunately a lot more mixed than I anticipated. One opcode significantly increases in performance (`MLOAD`, going from 2.9 to 4.4 or so), but other values are mixed throughout the range, some somewhat faster, some somewhat slower.

Might come back to this with some distance, atm I am bit baffled that there is "not more to see".

Here is e.g. a resultset on the same block (left is with the hybrid stack).

grafik

My initial idea was to keep the old stack, but introduce a flag which would allow to switch to the new one. If results remain this mixed this might be obsolete though.

This work might nevertheless be useful for some things, e.g. it is now extremely easily possible to switch e.g. arithmetic operations from being done bigint based to byte based and see how performance reacts.

### DIFF
--- a/packages/evm/src/hybridStack.ts
+++ b/packages/evm/src/hybridStack.ts
@@ -1,0 +1,314 @@
+import { bigIntToBytes, bytesToBigInt } from '@ethereumjs/util'
+
+import { ERROR, EvmError } from './exceptions.js'
+
+import type { EVMStack } from './types.js'
+
+/**
+ * Hybrid stack implementation for the EVM, containing either bytes (preferred) or bigint values.
+ *
+ * This stack implementation is significantly more performant than the single type stack. Atm it is however
+ * not activated by default for backwards compatibility reasons.
+ *
+ * In the current EVM version this stack performs significantly worse than the hybrid type stack on the EVM
+ * `step` event (if used), since there is mapping to a single type stack needed on each step.
+ *
+ * Developer notes:
+ *
+ * 1. If there are bigint/bytes conversions in opcode implementations before or after accessing the stack,
+ * direct calling into the type-named methods here should be preferred and the conversion removed from the
+ * opcode implementation.
+ * 2. If a byte-based opcode implementation has no direct performance penalties towards the bigint version,
+ * the bytes version should be used (respectively: replace the bigint version)
+ */
+
+type StackElement = [bigint | undefined, Uint8Array | undefined]
+type Store = StackElement[]
+
+export class HybridStack implements EVMStack {
+  // This array is initialized as an empty array. Once values are pushed, the array size will never decrease.
+  private _store: Store
+  private _maxHeight: number
+
+  private _len: number = 0
+
+  constructor(maxHeight?: number) {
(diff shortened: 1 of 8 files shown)


holgerd77:
> * This stack implementation is significantly more performant than the single type stack. Atm it is however
> * not activated by default for backwards compatibility reasons.

I was so confident in the outcome here that I had even already written class documentation before I had first results mentioning the increased performance. 😆

packages/evm/src/hybridStack.ts
```diff
@@ -0,0 +1,314 @@
...
+
+    // Length is checked above, so pop shouldn't return undefined
+    // First decrease current length, then read the item and return it
+    // Note: this does thus not delete the item from the internal array
+    // However, the length is decreased, so it is not accessible to external observors
+    const elem = this._store[--this._len]
+    if (elem[0] !== undefined) {
+      return elem[0]
+    }
+    return bytesToBigInt(elem[1]!)
```
acolytec3, line 107: Does this scenario ever happen?  Since the stack is always used internally and never exposed directly, can we safely assume if `popBigInt` is called that `elem[1]` will alaywas be undefined?

packages/evm/src/hybridStack.ts
```diff
@@ -0,0 +1,314 @@
...
+
+    // Length is checked above, so pop shouldn't return undefined
+    // First decrease current length, then read the item and return it
+    // Note: this does thus not delete the item from the internal array
+    // However, the length is decreased, so it is not accessible to external observors
+    const elem = this._store[--this._len]
+    if (elem[1] !== undefined) {
+      return elem[1]
+    }
+    return bigIntToBytes(elem[0]!)
```
acolytec3, line 127: Same question as above.  Shouldn't we just return `elem[1]` here and leave it up to the consumer to know that it's bytes vs bigint?

acolytec3 (commented):
I don't have a lot of deep insight into this yet but left a couple of initial comments.  It's an interesting idea for sure.

packages/evm/src/hybridStack.ts
```diff
@@ -0,0 +1,314 @@
...
+import type { EVMStack } from './types.js'
+
+/**
+ * Hybrid stack implementation for the EVM, containing either bytes (preferred) or bigint values.
+ *
+ * This stack implementation is significantly more performant than the single type stack. Atm it is however
+ * not activated by default for backwards compatibility reasons.
+ *
+ * In the current EVM version this stack performs significantly worse than the hybrid type stack on the EVM
+ * `step` event (if used), since there is mapping to a single type stack needed on each step.
```
jochem-brouwer, line 14: This is fine. We might want to note that if one listens to `step` then EVM performance will be much slower due to the objects it has to generate. `step` event should be somewhat seen as a `debug` mode event.

PR #3031, opened 2023-09-18
client: integrate snapsync on experimental basis

g11tech:
this PR aims to integrate snap sync on experimental basis (and has quite a good number fixes/refacs all over client beyond snapsync)

How to test it:

from `packages/client`
1. run a single node local network with static state
`
NETWORK=mainnet NETWORKID=1337903 ELCLIENT=geth EXTRA_CL_PARAMS="--params.CAPELLA_FORK_EPOCH 0"  DATADIR=/usr/app/ethereumjs/packages/client/data test/sim/single-run.sh
`
let it run for `32+` slots so that finalization happens, else you might end up waiting on step 2 anyway
2. run snapsync sim (you can run this as many times as you want while 1. is still running)
`rm -rf ./datadir; EXTERNAL_RUN=true NETWORK=mainnet NETWORKID=1337903 ELCLIENT=geth SNAP_SYNC=true DATADIR=/usr/app/ethereumjs/packages/client/data npx vitest run test/sim/snapsync.spec.ts`

OR if you are feeling lucky then just try it on holesky (modify your data and jwtsecret args accordingly):
`npm run client:start:ts -- --dataDir /data/lodestar-quickstart/holedata/ethereumjs --rpcEngineAddr 0.0.0.0 --jwtSecret /data/lodestar-quickstart/holesky-data/jwtsecret  --network holesky --rpcEngine --snap`
i.e. just add `--snap` additonal arg to whatever is your usual run command


packages/client/src/sync/fetcher/accountfetcher.ts
```diff
@@ -218,6 +154,129 @@ export class AccountFetcher extends Fetcher<JobTask, AccountData[], AccountData>
...
+
+    try {
+      this.fetcherDoneFlags.accountFetcher.started = true
+      const accountFetch = !this.fetcherDoneFlags.accountFetcher.done ? super.fetch() : null
+      // wait for all accounts to fetch else storage and code fetcher's doesn't get us full data
+      this.config.superMsg(`Snapsync: running accountFetch=${accountFetch !== null}`)
+      await accountFetch
+      this.fetcherDoneFlags.accountFetcher.started = false
+      if (this.fetcherDoneFlags.accountFetcher.done !== true) {
+        // @scorbajio need to see the reason for this here
```
scorbajio, line 168: I collected logs for several test runs but I have not yet been able to reproduce this error case. If there is a better way to reproduce it, let me know. Or if you can reproduce it, it would help if you could share logs for a full run.

packages/client/src/sync/fetcher/accountfetcher.ts
```diff
@@ -218,6 +154,129 @@ export class AccountFetcher extends Fetcher<JobTask, AccountData[], AccountData>
...
+        throw Error('accountFetcher finished without completing the sync')
+      }
+
+      const storageFetch = !this.fetcherDoneFlags.storageFetcherDone
+        ? this.storageFetcher.fetch().then(
+            // we should not be doing this, fetcher itself should mark completion
```
scorbajio, line 174: An alternative to using callbacks for calling `snapFetchersCompleted` is making `snapFetchersCompleted` available in the `storagefetcher` by either exporting it or passing it into it as a parameter. Then, it can call it in store when it receives the termination signal and is setting `destroyWhenDone` to `true`:
```
if (JSON.stringify(result[0]) === JSON.stringify(Object.create(null))) {
        this.debug(
          'Empty result detected - Associated range requested was empty with no elements remaining to the right'
        )
        this.destroyWhenDone = true
        return
}
```
However, the code and trienode fetchers don't have a clear terminating condition. That's why they have the `setDestroyWhenDone` function implemented so that when the account fetcher is done and all other necessary nodes and codes have been queued for fetching, those two fetchers will be destroyed as soon as they complete their last job.

packages/statemanager/src/stateManager.ts
```diff
@@ -406,19 +406,28 @@ export class DefaultStateManager implements EVMStateManagerInterface {
    * cache or does a lookup.
    * @private
    */
-  protected _getStorageTrie(address: Address, account: Account): Trie {
-    // from storage cache
-    const addressHex = bytesToUnprefixedHex(address.bytes)
-    const storageTrie = this._storageTries[addressHex]
+  // TODO PR: have a better interface for hashed address pull?
+  protected _getStorageTrie(addressOrHash: Address | Uint8Array, account?: Account): Trie {
```
g11tech, line 410: @holgerd77 for your attention and feedback :pray:

holgerd77: Hmm, since this is an internal method I would assume this should be fine for now.

Just FYI: we experimented though lately to get rid of this method (and generally: mechanism) completely here #3117 to see if we can get rid of creating all these trie copies.

That's likely (?) another story though.

jochem-brouwer:
Have merged `master` and pushed a fix for the conflicts (am branching off this one as suggested to do reorg logic)

g11tech:
> Have merged `master` and pushed a fix for the conflicts (am branching off this one as suggested to do reorg logic)

added the handling for keeping  unfinalized non canonical blocks around in skeleton to handle some reorgs without backfill
reccomend to add that commit into your hive reorg fixes branch as well

holgerd77:
Have updated this via UI

holgerd77 (approved):
Ok, have had another high level look here, mainly from a how-is-the-main codebase affected. Also did two sanity run checs, one mainnet sync up to block 80.000 or so and one Holesky sync.

Seems things are working fine. 🙂 Can't oversee all the changes in the Skeleton class, e.g., but I do think we should take this code base in. Then we are getting more familiar with this over time, so this is very much a thing where one needs to grow some understanding/feeling of the behavior over time anyhow.

Will merge. 🎉 (thanks @scorbajio for the base stack and @g11tech for the integration! 😍)

PR #3032, opened 2023-09-18
EVM: Initialize memory with CONTAINER_SIZE bytes

holgerd77:
This is a small performance optimization improving `MLOAD` and `MSTORE` performance by 10-20%.

Performance results (always last block):

cd ../evm && npm run build && cd ../client && npm run client:start -- --sync=none --vmProfileBlocks --executeBlocks=962960-962971 (MLOAD 48, MSTORE 100 times)
MLOAD
New 2.8  2.7 2.8 2.8
Old. 2.65 2.8 2.7 2.5
MSTORE
New 1.85 1.8 1.9 1.8
Old. 1.65  1.8 1.7 1.75

cd ../evm && npm run build && cd ../client && npm run client:start -- --sync=none --vmProfileBlocks --executeBlocks=962960-962970 (MLOAD 1224, MSTORE 2550 times)
MLOAD
New 2.8 2.75 2.7 2.65
Old.  2.85 2.8 2.7 2.55
MSTORE
New 1.85 1.85 1.8 1.7
Old. 1.75. 1.7 1.7 1.65

Ok. So I would think that results are clear enough on this and we can take this in. Ready for review. 🙂 
We should nevertheless really do this normalized MGas/s (as discussed with @jochem-brouwer), still think this would make comparison significantly easier (high level idea: normalize by constant factor to factor out runtime differences).

Might take this on in a separate PR.

Think this should nevertheless be fine.

### DIFF
--- a/packages/evm/src/memory.ts
+++ b/packages/evm/src/memory.ts
@@ -19,7 +19,7 @@ export class Memory {
   _store: Uint8Array
 
   constructor() {
-    this._store = new Uint8Array(0)
+    this._store = new Uint8Array(CONTAINER_SIZE)
   }
 
   /**
@@ -34,10 +34,8 @@ export class Memory {
     const newSize = ceil(offset + size, 32)
     const sizeDiff = newSize - this._store.length
     if (sizeDiff > 0) {
-      this._store = concatBytes(
-        this._store,
-        new Uint8Array(Math.ceil(sizeDiff / CONTAINER_SIZE) * CONTAINER_SIZE)
-      )
+      const expandBy = Math.ceil(sizeDiff / CONTAINER_SIZE) * CONTAINER_SIZE
+      this._store = concatBytes(this._store, new Uint8Array(expandBy))
     }
   }
 
--- a/packages/evm/test/memory.spec.ts
+++ b/packages/evm/test/memory.spec.ts
@@ -2,19 +2,21 @@ import { assert, describe, it } from 'vitest'
 
 import { Memory } from '../src/memory.js'
 
+const CONTAINER_SIZE = 8192
+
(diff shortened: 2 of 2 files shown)


jochem-brouwer (approved):
I think this is fine, I might want to add a specific attack test for this one later (but this will be done in a separate PR). However, your description does not seem to match, you seem to run the exact same test (except the second test is one block less than the first one) - but the amount of MSTORE/MLOADS differ.

holgerd77:
@jochem-brouwer Thanks for approval will merge! 👍 

> However, your description does not seem to match, you seem to run the exact same test (except the second test is one block less than the first one) - but the amount of MSTORE/MLOADS differ.

TBH I cannot follow yet what you mean here (there are only two tests?).

jochem-brouwer:
You run:

`cd ../evm && npm run build && cd ../client && npm run client:start -- --sync=none --vmProfileBlocks --executeBlocks=962960-962971 (MLOAD 48, MSTORE 100 times)`

and

`cd ../evm && npm run build && cd ../client && npm run client:start -- --sync=none --vmProfileBlocks --executeBlocks=962960-962970 (MLOAD 1224, MSTORE 2550 times)`

(per OP)

These are almost the same except the second test runs until 962970 (so one block less), yet is has somehow more MLOADs and MSTOREs? :thinking:

holgerd77:
Results (and MLOAD / MSTORE counts) are always only for the last block. I'm always testing like this, running some blocks before, to avoid interference with client startup stuff.

jochem-brouwer:
Oh right, this makes a lot of sense (to warmup the client). Thanks for clarifying :smile:

PR #3034, opened 2023-09-18
EVM: BigInt Constants and EXP Optimization

holgerd77:
This is another smallish performance PR, does the following:

Low number and reused bigints (0, 1, 2,...) in central places are replaced by a one-time initialized constant.

Overall gain is hard to measure, but this is noticeable for opcodes which do not do much else - like the `EQ` opcode.

E.g. running `cd ../evm && npm run build && cd ../client && npm run client:start -- --sync=none --vmProfileBlocks --executeBlocks=962960-962970` on master and this PR gives a repeated new total ms value around 0.41 while the old one is at around 0.46.

So the opcode is just this:

```typescript
// 0x14: EQ
  [
    0x14,
    function (runState) {
      const [a, b] = runState.stack.popN(2)
      const r = a === b ? BIGINT_1 : BIGINT_0
      runState.stack.push(r)
    },
  ],
```

So good testbed for the gain, since two bigint constants are used and not much else is happening.

PR further caches some EXP values, since I realized that this opcode is very often run with same values.

This brings down/up EXP dramatically, from something like 9 Mgas/s to something around 30 Mgas/s.

This is nice, overall gain is limited, should be around 1ms per block though. So that is something.

Another fix: hash function on sha256 precompile (not keccak!) was called twice (sigh), this precompile is seldomly called though.

Open for review. 🙂 (everyone feel invited)

### DIFF
--- a/packages/evm/src/opcodes/functions.ts
+++ b/packages/evm/src/opcodes/functions.ts
@@ -36,6 +36,16 @@ import type { Common } from '@ethereumjs/common'
 
 const EIP3074MAGIC = hexToBytes('0x03')
 
+const BIGINT_0 = BigInt(0)
+const BIGINT_1 = BigInt(1)
+const BIGINT_2 = BigInt(2)
+const BIGINT_96 = BigInt(96)
+const BIGINT_160 = BigInt(160)
+const BIGINT_224 = BigInt(224)
+const BIGINT_2EXP96 = BigInt(79228162514264337593543950336)
+const BIGINT_2EXP160 = BigInt(1461501637330902918203684832716283019655932542976)
+const BIGINT_2EXP224 = BigInt(26959946667150639794667015087019630673637144422540572481103610249216)
+
 export interface SyncOpHandler {
   (runState: RunState, common: Common): void
 }
@@ -88,8 +98,8 @@ export const handlers: Map<number, OpHandler> = new Map([
     function (runState) {
       const [a, b] = runState.stack.popN(2)
       let r
-      if (b === BigInt(0)) {
-        r = BigInt(0)
+      if (b === BIGINT_0) {
+        r = BIGINT_0
       } else {
         r = mod(a / b, TWO_POW256)
       }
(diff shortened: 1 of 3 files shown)


packages/evm/src/opcodes/functions.ts
```diff
@@ -36,6 +36,16 @@ import type { Common } from '@ethereumjs/common'
 
 const EIP3074MAGIC = hexToBytes('0x03')
 
+const BIGINT_0 = BigInt(0)
+const BIGINT_1 = BigInt(1)
+const BIGINT_2 = BigInt(2)
```
jochem-brouwer, line 41: For constants other than `0` and `1`, could you add a comment why these are relevant?

jochem-brouwer (commented):
I think for constants which we re-use in multiple files we should export this from our `util` library for cleani-ness. We might do a monorepo-wide review of `BigInt(` to see where these are created and just cache those values each time.

One extra comment. For EXP, I am not super sure about the changes but if there is a downside of the impact it should be very small.

packages/evm/src/opcodes/functions.ts
```diff
@@ -171,12 +181,25 @@ export const handlers: Map<number, OpHandler> = new Map([
     0x0a,
     function (runState) {
       const [base, exponent] = runState.stack.popN(2)
-      if (exponent === BigInt(0)) {
-        runState.stack.push(BigInt(1))
+      if (base === BIGINT_2) {
```
scorbajio, line 184: Is this conditional added to pre-compute some commonly used exponents of base 2?

acolytec3 (commented):
Another general question here, if we're going to use these constants across functions/gas, should apply it across the rest of the EVM?  There are a number of places where we use `BigInt(0)`.  Should be easy enough to move these constants into their own file and do it this way.

jochem-brouwer:
@acolytec3 I will do this monorepo-wide in a new PR

acolytec3 (approved):
LGTM

PR #3035, opened 2023-09-18
evm: add blobgasfee eip 7516

g11tech:
for devnet 9 (included in cancun) - https://github.com/ethereum/pm/issues/857

EIP https://eips.ethereum.org/EIPS/eip-7516

### DIFF
--- a/packages/common/src/eips.ts
+++ b/packages/common/src/eips.ts
@@ -471,4 +471,17 @@ export const EIPs: EIPsDict = {
     minimumHardfork: Hardfork.London,
     requiredEIPs: [],
   },
+  7516: {
+    comment: 'BLOBBASEFEE opcode',
+    url: 'https://eips.ethereum.org/EIPS/eip-7516',
+    status: Status.Draft,
+    minimumHardfork: Hardfork.London,
+    requiredEIPs: [4844],
+    gasPrices: {
+      blobbasefee: {
+        v: 2,
+        d: 'Gas cost of the BLOBBASEFEE opcode',
+      },
+    },
+  },
 }
--- a/packages/common/src/eips.ts
+++ b/packages/common/src/eips.ts
@@ -475,7 +475,7 @@ export const EIPs: EIPsDict = {
     comment: 'BLOBBASEFEE opcode',
     url: 'https://eips.ethereum.org/EIPS/eip-7516',
     status: Status.Draft,
-    minimumHardfork: Hardfork.Cancun,
+    minimumHardfork: Hardfork.Paris,
     requiredEIPs: [4844],
     gasPrices: {
       blobbasefee: {
--- a/packages/common/src/eips.ts
+++ b/packages/common/src/eips.ts
@@ -475,7 +475,7 @@ export const EIPs: EIPsDict = {
     comment: 'BLOBBASEFEE opcode',
     url: 'https://eips.ethereum.org/EIPS/eip-7516',
     status: Status.Draft,
-    minimumHardfork: Hardfork.Shanghai,
+    minimumHardfork: Hardfork.Cancun,
     requiredEIPs: [4844],
     gasPrices: {
       blobbasefee: {
--- a/packages/common/src/eips.ts
+++ b/packages/common/src/eips.ts
@@ -475,7 +475,7 @@ export const EIPs: EIPsDict = {
     comment: 'BLOBBASEFEE opcode',
     url: 'https://eips.ethereum.org/EIPS/eip-7516',
     status: Status.Draft,
-    minimumHardfork: Hardfork.London,
+    minimumHardfork: Hardfork.Shanghai,
     requiredEIPs: [4844],
     gasPrices: {
       blobbasefee: {
--- a/packages/common/src/hardforks.ts
+++ b/packages/common/src/hardforks.ts
@@ -834,7 +834,7 @@ export const hardforks: HardforksDict = {
   cancun: {
     name: 'cancun',
     comment:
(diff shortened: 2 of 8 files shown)


packages/evm/src/interpreter.ts
```diff
@@ -760,6 +760,18 @@ export class Interpreter {
     return baseFee
   }
 
+  /**
+   * Returns the Base Fee of the block as proposed in [EIP-3198](https;//eips.etheruem.org/EIPS/eip-3198)
```
acolytec3, line 764:
```suggestion
   * Returns the Blob Base Fee of the block as proposed in [EIP-7516](https;//eips.etheruem.org/EIPS/eip-7516)
```

acolytec3 (changes requested):
I'm assuming there may be more to come here but left a couple of cleanup notes related to code comments and casing.

g11tech:
> I'm assuming there may be more to come here but left a couple of cleanup notes related to code comments and casing.

thanks for the code comments, updated

check casing response. should be good for review/merge now

packages/common/src/eips.ts
```diff
@@ -471,4 +471,17 @@ export const EIPs: EIPsDict = {
     minimumHardfork: Hardfork.London,
     requiredEIPs: [],
   },
+  7516: {
+    comment: 'BLOBBASEFEE opcode',
+    url: 'https://eips.ethereum.org/EIPS/eip-7516',
+    status: Status.Draft,
+    minimumHardfork: Hardfork.London,
```
holgerd77, line 478: London is too low, should be Shanghai

packages/common/src/hardforks.ts
```diff
@@ -834,9 +834,9 @@ export const hardforks: HardforksDict = {
   cancun: {
     name: 'cancun',
     comment:
-      'Next feature hardfork after the shanghai having proto-danksharding EIP 4844 blobs (still WIP hence not for production use), transient storage opcodes, parent beacon block root availability in EVM and selfdestruct only in same transaction',
+      'Next feature hardfork after the shanghai having proto-danksharding EIP 4844 blobs (still WIP hence not for production use), transient storage opcodes, parent beacon block root availability in EVM, selfdestruct only in same transaction and blob base fee opcode',
```
scorbajio, line 837:
```suggestion
      'Next feature hardfork after shanghai, includes proto-danksharding EIP 4844 blobs (still WIP hence not for production use), transient storage opcodes, parent beacon block root availability in EVM, selfdestruct only in same transaction, and blob base fee opcode',
```

packages/common/src/eips.ts
```diff
@@ -471,4 +471,17 @@ export const EIPs: EIPsDict = {
...
+    requiredEIPs: [4844],
+    gasPrices: {
+      blobbasefee: {
```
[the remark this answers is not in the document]
acolytec3: You're right.  I loked across the `eips.ts` file and there's no consistency here.  Just a silly thought but maybe a follow-up PR is to just make all of the `gasPrices` fields in `common/src/eips` lowercase so we can drop that `toLowerCase` call here.  That's certainly a small performance gain for every single time we get the base fee.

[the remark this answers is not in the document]
jochem-brouwer: @acolytec3 Note that the line which @g11tech references to is in `codes.ts` of EVM (the opcode builder). The opcode builder expects all to be opcode names in common to be lowercase. Note: this method is only called if we get new opcodes (at EVM startup + fork transition) so this is not called each time for all opcodes if one is called.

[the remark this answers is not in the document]
holgerd77: Won't die on this hill but in most cases we *do* have the camel case and I do somewhat like it for readability, but it is also true that this is a lot more inconsistent than I anticipated when looking through Common naming. Atm we can't update anyhow since this would be a breaking change.

So for here I guess we can keep both ways.

packages/common/src/eips.ts
g11tech: updated min hf to `Cancun` as blobs will be coming `Cancun` onwards

packages/evm/src/interpreter.ts
```diff
@@ -760,6 +760,18 @@ export class Interpreter {
...
+   * Returns the Blob Base Fee of the block as proposed in [EIP-7516](https;//eips.etheruem.org/EIPS/eip-7516)
+   */
+  getBlobBaseFee(): bigint {
+    const blobBaseFee = this._env.block.header.getBlobGasPrice()
+    if (blobBaseFee === undefined) {
+      // Sanity check
+      throw new Error('Block has no Blob Base Fee')
+    }
+    return blobBaseFee
+  }
```
holgerd77, line 773: Not so deep in the blob pricing topic: so is "blobBaseFee === blobGasPrice" here?

If so: I think we then should consider still renaming all `blobGasPrice` related naming in `Block` to `blobBaseFee`? (we can still do since all 4844 functionality is still non-final)

packages/common/src/eips.ts
holgerd77: Ah, no, that's always the same discussion: the definition of this is "the minimum HF where this EIP (+ eventually other EIPs which are necessary for this one, so this would be 4844, but this part is covered by the EIP dependency setting) can be run in isolation". So this would be Shanghai.

packages/evm/src/interpreter.ts
holgerd77: Guess we might want to wait with merging here until this is settled? 🤔 Since this is the very last question on ACD channel and still un-answered?

packages/common/src/eips.ts
g11tech: updated to Paris post discussion to confirm with `4844` settings :+1:

holgerd77 (approved):
LGTM

PR #3036, opened 2023-09-18
Clean up CLI arg passing

### LINKED ISSUE #3025: Client doesn't throw on unknown CLI params
As reported on Discord, the client does not throw when unknown commands are passed to it, despite having the option enabled [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/8ca49a1c346eb7aa61acf550f8fe213445ef71ab/packages/client/bin/cli.ts#L362) in `yargs` to do so.  This issue should be researched and resolved.

acolytec3:
Various CLI fixes
- Fixes #3025 (by fixing how `yargs` is imported)
- Fixes check in `--dev` to ensure can't be passed with no value
- Replaces 2 instances of hand-crafted checks with `yargs` builtin checks for params that conflict with one another or else require another
- Adds a new experimental autocompletion feature for the CLI that only works on the built version of the client

To enable autocomplete:
Run `dist/bin/cli.js -- completion >> ~/.zshrc` (or `.bashrc` if you use bash).  This adds the yargs autocomplete script into your terminal configuration.

Then, any time you run the client, if you run the compiled client, if you type `dist/bin/cli.js --` and then hit `tab` or start to type a parameter and hit `tab`, it should show you a list of param options that start with that letter.


holgerd77:
Updated this via UI

holgerd77 (changes requested):
Can we have the autocomplete installation "guide" a bit more permanently written down, otherwise this will likely get lost again.

Maybe a short section "Autocompletion (Experimental)" or similar after the [Source Installation](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/client#source-installation) section in the client README or so?

packages/client/bin/cli.ts
```diff
@@ -79,10 +80,12 @@ const args: ClientOpts = yargs(hideBin(process.argv))
   .option('customChain', {
     describe: 'Path to custom chain parameters json file (@ethereumjs/common format)',
     coerce: (arg: string) => (arg ? path.resolve(arg) : undefined),
+    implies: 'customGenesisState',
   })
   .option('customGenesisState', {
     describe: 'Path to custom genesis state json file (@ethereumjs/common format)',
     coerce: (arg: string) => (arg ? path.resolve(arg) : undefined),
+    implies: 'customChain',
```
holgerd77, line 88: Just to make sure: is this really true that both arguments always need to be used in conjunction?

acolytec3: That's what the original error message said and it makes logical sense to me, that you need to provide some genesis state data for a custom chain.  [Here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/0606d3cf8c316ebc5d5433c90a34632e3acefb3d/packages/client/bin/cli.ts#L754) is the original code I took this from.

holgerd77:
Updated this via UI

holgerd77:
Updated this via UI

packages/client/test/cli/cli.spec.ts
```diff
@@ -300,21 +292,22 @@ describe('[CLI]', () => {
...
+          port: 8563,
         })
         try {
           await clientNoConnection.request('web3_clientVersion', [], 2.0)
           assert.fail('should have thrown on invalid client address')
         } catch (e: any) {
-          assert.equal(e.code, 'ECONNREFUSED', 'failed to connect to RPC on invalid address')
+          assert.ok(e !== undefined, 'failed to connect to RPC on invalid address')
           child.kill(9)
           resolve(undefined)
```
gabrocheleau, line 305: Can't we just `resolve()` ?

gabrocheleau: For this and other instances.

packages/client/bin/cli.ts
g11tech: well custom genesis can be provided without custom chain --- and/or viceversa, not sure if it actually works

g11tech (commented):
looks amazing especially autocomplete feature, tried it out!
just need one bit of discussion left to resolve

packages/client/test/cli/cli.spec.ts
gabrocheleau: Nevermind, a lot of these resolve are typed with `unknown` as arg type which doesn't allow for optionality.

gabrocheleau (approved):
Nice work!

packages/client/bin/cli.ts
acolytec3: The error messages indicated you can't do one without the other which I why I just copied it over. Do we want to allow one without the other?

g11tech: ok, i guess we are good for now, when we support and test then we can remove this :+1:

g11tech (approved):

scorbajio (approved):
tested ACK

acolytec3:
@holgerd77, github won't let me merge this until you update your review.  Is there anything else outstanding you'd like changed/updated/fixed before we merge?

holgerd77 (approved):

PR #3038, opened 2023-09-18
Return '0x' values as '0x0'

scorbajio:
This change changes the return value format for `statemanager` rpc `getProof` calls from `0x` to `0x0` for a quantity-encoded RPC representation.


packages/statemanager/src/stateManager.ts
jochem-brouwer, line 1: This will not work if the account exists and:

- The balance is zero 
- (Or) The nonce is zero

In practice I don't think an account can exist with nonzero balance and nonzero nonce, however it is definitely possible to have zero balance and nonzero nonce.

Changes should be at line 653, around:

```typescript
    const returnValue: Proof = {
      address: address.toString(),
      balance: bigIntToHex(account.balance),
      codeHash: bytesToHex(account.codeHash),
      nonce: bigIntToHex(account.nonce),
```

jochem-brouwer (changes requested):
See the comment why this might fail on some places

Note: we might want to add a general method maybe, `bigIntoToRPCHex` and `bytesToRPCHex` such that we return `0x0` in case it evaluates to `0x`? (This is not necessary for this PR, but would be nice in general)

packages/statemanager/src/stateManager.ts
holgerd77: I can't ad hoc following here, could you expand on why this is not working if the balance is zero? Wouldn't this then just return 0 and that's it?

jochem-brouwer: I'll add this case as a commit, I think that makes it easier to see :smile:

jochem-brouwer: Never mind, it uses `bigIntToHex` which correctly pads it.

jochem-brouwer: Have added extra test case for this to verify.

jochem-brouwer (approved):

PR #3039, opened 2023-09-19
client/vm: fix block builder london transition

jochem-brouwer:
This fixes the construction of the initial London block. We take the gasLimit from the previous block (this is incorrect in London and should be (approx.) twice the gasLimit of the previous pre-London block at the initial London block) and also do not set the initial base fee correctly.

Found in: https://github.com/ethereum/execution-apis/pull/398

To run: 

1. Checkout the branch of the PR above (it is not merged yet)
2. Run: `./hive --client ethereumjs --sim ethereum/engine --sim.limit "engine-withdrawals/Pre-Merge Fork Number" --docker.output`


packages/vm/src/buildBlock.ts
```diff
@@ -85,7 +85,11 @@ export class BlockBuilder {
       this.vm.common.isActivatedEIP(1559) === true &&
       typeof this.headerData.baseFeePerGas === 'undefined'
     ) {
-      this.headerData.baseFeePerGas = opts.parentBlock.header.calcNextBaseFee()
+      if (this.headerData.number === vm.common.hardforkBlock(Hardfork.London)) {
+        this.headerData.baseFeePerGas = vm.common.param('gasConfig', 'initialBaseFee')
```
jochem-brouwer, line 89: This could also be moved back into client, but I think the place for this change is better here?

Note: it doesn't seem that we have access to the parent block in this scope, so cannot check parent gas limit here... (?)

g11tech: `opts.parentBlock` could be available here

acolytec3:
I've confirmed the specified test above passes so this LGTM.  That said, why does that test not appear in the official hive test results?  I spent a while looking for it in the results but I don't see it listed in the most recent [`engine-withdrawals`](https://hivetests2.ethdevops.io/suite.html?suiteid=1695199314-62503c04b9e4700a299e275d574da5d8.json&suitename=engine-withdrawals#) results.

packages/vm/src/buildBlock.ts
g11tech: added a commit to address this

acolytec3 (approved):
LGTM

jochem-brouwer:
@acolytec3 I suspect it is because there are no "official" Cancun tests yet? (Except on some non-master branches)

PR #3040, opened 2023-09-19
EVM: Simple One-Item Peek Cache Experiment

holgerd77:
This is a quick experiement with a very simple one-item peek cache - following up a bit on the work on the hybrid stack from #3029 - to see if this somewhat mitigates our cleaner but a bit heavier double access stack structure through the separation of gas calculation (`gas.ts`) and opcode execution (`functions.ts`).

Results are not so clear, might bring a performance gain, might not.

This would need some profiler additions I guess to be more on the save side (will talk through with @jochem-brouwer).

### DIFF
--- a/packages/evm/src/stack.ts
+++ b/packages/evm/src/stack.ts
@@ -1,5 +1,10 @@
 import { ERROR, EvmError } from './exceptions.js'
 
+type PeekCache = {
+  num: number
+  peekArray: bigint[]
+}
+
 /**
  * Implementation of the stack used in evm.
  */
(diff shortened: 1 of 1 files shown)


packages/evm/src/stack.ts
```diff
@@ -1,5 +1,10 @@
 import { ERROR, EvmError } from './exceptions.js'
 
+type PeekCache = {
+  num: number
```
jochem-brouwer, line 4: This `num` should be the same as `peekArray.length` - right?

holgerd77: Yes, maybe both wrongly named and not necessary on second thought. 😂

jochem-brouwer:
Interesting, `randomStatetest437` in state tests fails? :thinking: (and similar in blockchain tests?)

PR #3041, opened 2023-09-19
VM/EVM profiler: add summary, add Blocks per Slot normalized field

jochem-brouwer:
image

This PR adds an overall summary of the profiler: the amount of opcodes (calls), the total gas spent, the MGas/S, and the new column Blocks per Slot.

If Blocks per Slot is `>= 1` then we can execute one block within one beacon chain slot (12 seconds). If it is below, we are executing less than one block per slot, i.e. we cannot keep up with the chain because we are too slow.

This BpS is also added for each opcode: if it is below 1, it is too slow. (Note, that if the amount of calls is too low, think less than 10, this might not give a good indication if the opcode is too slow or not)


holgerd77:
Rebased this via UI

holgerd77 (approved):
Cool, LGTM!

PR #3047, opened 2023-09-20
Snap sync: use zero-element proof for checking validity of final, empty range result

scorbajio:
During snap sync, if the client requests a range that is both empty and has no elements remaining to the right of it, the peer responds with just a proof. This proof can be checked as part of a zero-element proof to validate that there really are no remaining elements to be fetched past the range.

### DIFF
--- a/packages/client/src/sync/fetcher/accountfetcher.ts
+++ b/packages/client/src/sync/fetcher/accountfetcher.ts
@@ -237,6 +237,12 @@ export class AccountFetcher extends Fetcher<JobTask, AccountData[], AccountData>
     }
 
     const trie = new Trie()
+    // if (accounts.length === 0) {
+    //   // no-elements proof to check if final range is truly empty
+    //   this.debug('dbg200')
+    //   return trie.verifyRangeProof(stateRoot, origin, null, [], [], <any>proof)
+    // }
+
     const keys = accounts.map((acc: any) => acc.hash)
     const values = accounts.map((acc: any) => accountBodyToRLP(acc.body))
     // convert the request to the right values
@@ -311,6 +317,23 @@ export class AccountFetcher extends Fetcher<JobTask, AccountData[], AccountData>
     ) {
       // TODO have to check proof of nonexistence -- as a shortcut for now, we can mark as completed if a proof is present
       if (rangeResult.proof.length > 0) {
+        this.debug('dbg100')
+        const t = new Trie()
+        try {
+          const isMissingRightRange = await t.verifyRangeProof(
+            this.root,
+            origin,
+            null,
+            [],
+            [],
+            <any>rangeResult.proof
+          )
+        } catch (e) {
+          console.error(e)
+        }
+
+        this.debug(isMissingRightRange)
+
         this.debug(`Data for last range has been received`)
         // response contains empty object so that task can be terminated in store phase and not reenqueued
         return Object.assign([], [Object.create(null)], { completed: true })
--- a/packages/client/src/sync/fetcher/storagefetcher.ts
+++ b/packages/client/src/sync/fetcher/storagefetcher.ts
@@ -244,6 +244,17 @@ export class StorageFetcher extends Fetcher<JobTask, StorageData[][], StorageDat
       // TODO have to check proof of nonexistence since we are statically partitioning ranges --
       // as a shortcut for now, we can mark as completed if a proof is present
       if (rangeResult.proof.length > 0) {
+        this.debug('dbg100')
+        const isMissingRightRange = await this._proofTrie.verifyRangeProof(
+          task.storageRequests[0].storageRoot,
+          origin,
+          null,
+          [],
+          [],
+          <any>rangeResult.proof
+        )
+        this.debug(isMissingRightRange)
+
         this.debug(`Empty range was requested - Terminating task`)
         // response contains empty object so that task can be terminated in store phase and not reenqueued
         return Object.assign([], [Object.create(null) as any], { completed: true })


holgerd77:
Updated this via UI

Can this get a review at some point? 🙏

holgerd77:
Will put this back in WIP since it has failing tests.

holgerd77:
@scorbajio What is the status of this PR?

packages/trie/src/proof/range.ts
```diff
@@ -443,28 +446,28 @@ export async function verifyRangeProof(
...
       rootHash,
-      nibblestoBytes(firstKey),
-      proof,
+      nibblestoBytes(firstKey as any),
+      proof as any,
       useKeyHashingFunction
     )
 
-    if (value !== null || (await hasRightElement(trie, firstKey))) {
+    if (value !== null || (await hasRightElement(trie, firstKey as any))) {
```
jochem-brouwer, line 458: Why does the `as any` has to be added here? This was not necessary before (also in other occurrences of this file)

packages/trie/src/proof/range.ts
```diff
@@ -443,28 +446,28 @@ export async function verifyRangeProof(
...
       throw new Error('invalid zero element proof: value mismatch')
     }
 
     return false
   }
 
+  if (proof === null || firstKey === null || lastKey === null) {
+    throw new Error(
+      'invalid all elements proof: proof, firstKey, lastKey must be null at the same time'
```
jochem-brouwer, line 467: This is new, but has to do with an all elements proof :thinking: (looks correct to me, but maybe add a test :smile: )

jochem-brouwer (commented):
Two comments.

The implementation looks correct to me, but it seems that it lacks tests of the "unhappy" case that a peer gives us a zero element proof when there are clearly elements left in the trie (for both account + storage fetcher). Plus a case of the all-elements proof which is not covered :smile: :+1:

holgerd77:
@scorbajio This is a somewhat strange order of things with some likelihood to cause confusion with Jochem *doing* a review and leave some comments and then this PR being *set* to "Needs review" *without* the things mentioned actually being adressed. 🤔 

Maybe as some general suggestion always good to add some short comment when one is changing the PR state, this always helps to avoid misunderstandings and additionaly avoid of PR state changes getting lost since GitHub does not notify.

Thanks 🙏

packages/trie/src/proof/range.ts
scorbajio: This check is actually not new, I've only moved it bellow where zero-element proof checking happens. It was previously like this:
```
  if (proof === null || firstKey === null || lastKey === null) {
    throw new Error(
      'invalid all elements proof: proof, firstKey, lastKey must be null at the same time'
    )
  }

  // Zero element proof
  if (keys.length === 0) {
    const { trie, value } = await verifyProof(
      rootHash,
      nibblestoBytes(firstKey),
      proof,
      useKeyHashingFunction
    )

    if (value !== null || (await hasRightElement(trie, firstKey))) {
      throw new Error('invalid zero element proof: value mismatch')
    }

    return false
  }
```
The reason I did that was because zero-element proof checking was failing when I was passing in just the `root`, `firstKey`, and `proof`, with `null` for `lastKey` and empty arrays for `keys` and `values`, as such:
```
          const isMissingRightRange = await Trie.verifyRangeProof(
            this.root,
            origin,
            null,
            [],
            [],
            <any>rangeResult.proof,
            { useKeyHashingFunction: keccak256 }
          )
```
The error resulting from this was confusing a zero-element request to be a failed all-elements request:
```
2024-02-16T00:22:13.531Z client:AccountFetcher Error: invalid all elements proof: proof, firstKey, lastKey must be null at the same time
    at Module.verifyRangeProof (/Users/indigoalpha/code/ethjs/0/ethereumjs-monorepo/packages/trie/src/proof/range.ts:450:11)
    at Function.verifyRangeProof (/Users/indigoalpha/code/ethjs/0/ethereumjs-monorepo/packages/trie/src/trie.ts:208:12)
    at AccountFetcher.request (/Users/indigoalpha/code/ethjs/0/ethereumjs-monorepo/packages/client/src/sync/fetcher/accountfetcher.ts:400:50)
    at /Users/indigoalpha/code/ethjs/0/ethereumjs-monorepo/packages/client/test/sync/fetcher/accountfetcher.spec.ts:313:17
    at runTest (file:///Users/indigoalpha/code/ethjs/0/ethereumjs-monorepo/node_modules/@vitest/runner/dist/index.js:486:11)
    at runSuite (file:///Users/indigoalpha/code/ethjs/0/ethereumjs-monorepo/node_modules/@vitest/runner/dist/index.js:594:15)
    at runSuite (file:///Users/indigoalpha/code/ethjs/0/ethereumjs-monorepo/node_modules/@vitest/runner/dist/index.js:594:15)
    at runFiles (file:///Users/indigoalpha/code/ethjs/0/ethereumjs-monorepo/node_modules/@vitest/runner/dist/index.js:645:5)
    at startTests (file:///Users/indigoalpha/code/ethjs/0/ethereumjs-monorepo/node_modules/@vitest/runner/dist/index.js:654:3)
    at file:///Users/indigoalpha/code/ethjs/0/ethereumjs-monorepo/node_modules/vitest/dist/entry.js:278:7
```

packages/trie/src/proof/range.ts
scorbajio: See my other comment bellow (https://github.com/ethereumjs/ethereumjs-monorepo/pull/3047#discussion_r1491827795). It seems like the implementation wasn't allowing the zero-element parameter combination of an existing `firstKey`, a `null` `lastKey`, and an existing `proof` to continue to the zero-element proof verification section of the `verifyRangeProof` implementation because it assumed if `firstKey`, `lastKey`, and `proof` aren't all non-`null` values after moving past the all-elements proof verification section of the code, that there is an input error, when this is not the case for a zero-element proof.

scorbajio:
> Two comments.
> 
> The implementation looks correct to me, but it seems that it lacks tests of the "unhappy" case that a peer gives us a zero element proof when there are clearly elements left in the trie (for both account + storage fetcher). Plus a case of the all-elements proof which is not covered 😄 👍

That's a good point 🙂. I've added a test for the storage fetcher and account fetcher to test if they will reject a zero-element proof if an element to the right exists, but the all-elements proofs are already being tested in the trie package (e.g. https://github.com/ethereumjs/ethereumjs-monorepo/blob/5d522f731f6bee1b8626a3558fa8c5d5c0f6ebbd/packages/trie/test/proof/range.spec.ts#L212).

packages/trie/src/proof/range.ts
scorbajio: I've added a check to make sure the parameters are as expected when running a zero-element proof, so the casting to `any` has been removed.

holgerd77:
Updated this via UI

@jochem-brouwer can this get a review update?

holgerd77:
Updated this via UI

PR #3050, opened 2023-09-20
Bigint constants

jochem-brouwer:
This PR:

- Caches most (not all) constants of BigInts monorepo-wide
- Caches ALL constants in EVM
- The monorepo-wide constants are now exported from `util`

Note: please review https://github.com/ethereumjs/ethereumjs-monorepo/pull/3034 first (and merge in master if OK, or merge this on top if OK and then merge into master)

### DIFF
--- a/packages/block/src/block.ts
+++ b/packages/block/src/block.ts
@@ -3,6 +3,7 @@ import { RLP } from '@ethereumjs/rlp'
 import { Trie } from '@ethereumjs/trie'
 import { BlobEIP4844Transaction, Capability, TransactionFactory } from '@ethereumjs/tx'
 import {
+  BIGINT_0,
   KECCAK256_RLP,
   Withdrawal,
   bigIntToHex,
@@ -468,7 +469,7 @@ export class Block {
    */
   getTransactionsValidationErrors(): string[] {
     const errors: string[] = []
-    let blobGasUsed = BigInt(0)
+    let blobGasUsed = BIGINT_0
     const blobGasLimit = this.common.param('gasConfig', 'maxblobGasPerBlock')
     const blobGasPerBlob = this.common.param('gasConfig', 'blobGasPerBlob')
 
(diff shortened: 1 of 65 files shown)


holgerd77:
Great stuff, thanks for taking this one step further! 🙏

Looks generally great and good to merge from my side (won’t do myself since I opened the initial PR, would be great if someone else could take upon that). Some tests still failing though.

holgerd77:
Updated both the base branch and this branch and resolved a merge conflict, since I wanted to get the latest DB-layout (storage trie prefix) changes in to be able to run the profiler on this and compare.

holgerd77:
Puh puh puh.

It is so hard to get reliable deterministic results, I know realized that having two different CLI tabs (one with the PR and one with master and switching between them) might/seems also to influence performance since the OS might assign different ressources (?) to the specific tab or something. 🤔

I've now switched to staying in one tab and getting the different full paths with the Linux `pwd` command and then running `cd FULL_PATH_PR && npm run client:start ...` and `cd FULL_PATH_MASTER && npm run client:start ...`(having a copied repository there) in alternation, that seems to produce the most "honest" results to me.

Just "live-writing" down the results here, I am not sure yet if they show some clear trend:

Running `npm run client:start -- --sync=none --vmProfileBlocks --executeBlocks=2005567-2005590` in the way above and then again looking at the last block and noting here the BpS value (already totally fluent in our new terminology 😂):

PR        : 11.32. 10.16. 9.613. 14.971. 15.277. 9.521.  15.092. 10.277
Master : 9.651. 10.23. 10.07. 10.463. 9.971. 16.087. 15.641. 10.241

For some reasons GitHub sometimes thinks to auto-insert a dot after these numbers makes sense, I'll leave it there.

Hmm. So these results are so utterly diverse, I guess one can't just read anything out of this.

I'll run the same command as above but compare the on one block (number) lower (so 2005589):

PR.       : 32.583. 32.516. 31.984. 32.448. 30.251. 31.406. 32.788. 31.789
Master : 30.077. 31.789.  29.79.   30.251. 31.66.    31.596. 32.448. 31.853

Ok. Better.

Interesting that some blocks are *so* diverse.

Hmm. I guess one can't completely say for sure nevertheless if this improves things or not out of this results. Slight tendency eventually that it does.

So my conclusion would be somewhat: safe enough to say that this doesn't *worsen* things and likely enough that this somewhat improve things that we can merge. 🙂

acolytec3 (approved):
LGTM.  Going ahead and approving pending CI (though not sure if all will pass).

holgerd77 (approved):
LGTM

PR #3056, opened 2023-09-23
trie: update mpt tutorial

gabrocheleau:


packages/trie/examples/merkle_patricia_trees/README.md
```diff
@@ -209,145 +205,129 @@ Creating a branch node is a bit more complicated. For a branch to exist, we need
...
-[ <Buffer 31>, <Buffer 74 65 73 74 56 61 6c 75 65 41> ] // going down the path of "testKeyA"
+[path:  0x30  | value:  0x7465737456616c756530] // going down the path of "testKey0"
+[path:  0x31  | value:  0x7465737456616c756541] // going down the path of "testKeyA"
 ```
 
-I'm getting ahead of myself here, but here's a spoiler: these are leaf nodes. We see that each line contains two "buffers" (buffers are sequences of bytes). And indeed: leaf nodes are arrays containing two items. The first item is the remaining path (we'll get back to what the hex values stand for in a minute), while the second buffer is the value at that path. We can confirm that the values are what we'd expect:
+I'm getting ahead of myself here, but here's a spoiler: these are leaf nodes. We see that each line contains two "bytes" (bytes are sequences of bytes). And indeed: leaf nodes are arrays containing two items. The first item is the remaining path (we'll get back to what the hex values stand for in a minute), while the second bytes array is the value at that path. We can confirm that the values are what we'd expect:
```
scorbajio, line 326: I think "bytes are sequences of bytes" can be confusing to understand. Maybe rewording it would work better.
```suggestion
I'm getting ahead of myself here, but here's a spoiler: these are leaf nodes. We see that each line contains two byte arrays (byte arrays are sequences of bytes). And indeed: leaf nodes are arrays containing two items. The first item is the remaining path (we'll get back to what the hex values stand for in a minute), while the second bytes array is the value at that path. We can confirm that the values are what we'd expect:
```

packages/trie/examples/merkle_patricia_trees/README.md
```diff
@@ -395,15 +374,7 @@ hex char    bits    |        node type         path length
...
-In our code above, the first hexadecimal characters for the encodedPath of both nodes (`<30>` and `<31>`) was `3`. This correctly indicates that the node is of type `leaf` and that the path that follows is of odd length. The second character ( `<0>` and `<1>` respectively) indicates the last part of the path (the last hex character of our key).
-
-```jsx
-//      <------ same ------> <-> (different)
-<Buffer 74 65 73 74 4b 65 79 30>           // "testKey0", note that "0" = 30 in bytes
-<Buffer 74 65 73 74 4b 65 79 41>           // "testKeyA", note that "A" = 41 in bytes
-```
-
-But what happened with the second to last hex character (`3` and `4` respectively)? Recall that these are not necessary: **they were already provided by the indexes of the previous branch node!**
+In our code above, the first hexadecimal characters for the encodedPath of both nodes (`0x30` and `0x31`) was `3`. This correctly indicates that the node is of type `leaf` and that the path that follows is of odd length. The second character ( `0x0` and `0x1` respectively) indicates the last part of the path (the last hex character of our key).
```
scorbajio, line 377:
```suggestion
In our code above, the first hexadecimal character for the encodedPath of both nodes (`0x30` and `0x31`) is `3`. This correctly indicates that the node is of type `leaf` and that the path that follows is of odd length. The second character ( `0x0` and `0x1` respectively) indicates the last part of the path (the last hex character of our key).
```

packages/trie/examples/merkle_patricia_trees/README.md
```diff
@@ -412,80 +383,99 @@ With this in mind, let's take a closer look at leaf and extension nodes.
...
-<Buffer 74 65 73 74 4b 65 79>
-<Buffer 74 65 73 74 4b 65 79 30 30 30 31>
-<Buffer 74 65 73 74 4b 65 79 30 30 30 41>
+0x746573744b6579
+0x746573744b657930303031
+0x746573744b657930303041
 ```
 
-As you can see, the bytes `<30 30 30>` (standing for `000`> are common to both keys. We therefore should assume an extension that begins at index `3` of the branch node at "testKey". Let's see:
+As you can see, the bytes `303030>` (standing for `000`> are common to both keys. We therefore should assume an extension that begins at index `3` of the branch node at "testKey". Let's see:
```
scorbajio, line 420:
```suggestion
As you can see, the bytes `303030` (standing for `000` are common to both keys. We therefore should assume there exists an extension that begins at index `3` of the branch node at "testKey". Let's see:
```

packages/trie/examples/merkle_patricia_trees/README.md
```diff
@@ -496,43 +486,43 @@ You might have noticed that this child node is a "hash", while in the previous e
...
 
 // RESULT
-<Buffer 74 65 73 74 4b 65 79 30 30 30 31>
-<Buffer 74 65 73 74 4b 65 79 30 30 30 41>
+0x746573744b657930303031
+0x746573744b657930303041
 ```
 
-The first part of the path ("testKey" = `<74 65 73 74 4b 65 79>`) led us to the branch node. Next, taking the branch at index `3` (for hex value `<3>`) led us to our extension node, which automatically leads us down the path `[ 0, 3, 0, 3, 0 ]`. Using only two nodes (branch + extension), we are therefore able to "shortcut" the whole `<30 30 30>` part of the path! With a standard trie, this would have required 6 successive branch nodes!
+The first part of the path ("testKey" = `0x746573744b6579`) led us to the branch node. Next, taking the branch at index `3` (for hex value `3`) led us to our extension node, which automatically leads us down the path `03030`. Using only two nodes (branch + extension), we are therefore able to "shortcut" the whole `303030>` part of the path! With a standard trie, this would have required 6 successive branch nodes!
```
scorbajio, line 497:
```suggestion
The first part of the path ("testKey" = `0x746573744b6579`) led us to the branch node. Next, taking the branch at index `3` (for hex value `3`) led us to our extension node, which automatically leads us down the path `03030`. Using only two nodes (branch + extension), we are therefore able to "shortcut" the whole `303030` part of the path! With a standard trie, this would have required 6 successive branch nodes!
```

packages/trie/examples/merkle_patricia_trees/README.md
```diff
@@ -601,48 +596,68 @@ Easy, wasn't it?
...
+  _value: Uint8Array(0) []
 }
 
 ```
 
 Now let's suppose that I'm provided with various pieces of information, some that I trust, and some that I don't. Here's what I know from a trustworthy source:
 
 - The tree contains key-value pair "testKey": "testValue"
-- There also exists a branch node with hash: `<Buffer 70 b3 d0 20 ad 85 8f d6 00 28 a4 23 e9 8f 1d 99 c5 37 cd b9 1f 27 49 16 40 06 6f ea c7 9c 2f 72>`.
+- There also exists a branch node with hash (the value of the extension node in hex): `0x70b3d020ad858fd60028a423e98f1d99c537cdb91f27491640066feac79c2f72`.
```
scorbajio, line 658: I'm not sure what "the value of the extension node in hex" is referring to, here. Could you clarify or reword it?

gabrocheleau: I've improved the comment, hopefully that clarifies it.

holgerd77:
Cool that this got updated, just seeing this right now! 🙂 🎉

Issue #3044, opened 2023-09-19
Add configuration option for RPC to return actual error message when an RPC call fails


acolytec3:
Currently, if an error occurs when an RPC call is made, unless there is specific handling for a given error (either in input validation or elsewhere inside specific RPC method handlers), the RPC returns a generic `internal error` message that gives no hints as to what went wrong. We should figure out how to have the RPC return the actual error that caused the issue and a stack trace and then setup a new CLI flag to turn on this RPC Debug feature (or else have this driven by the `--rpcDebug` flag).

jochem-brouwer:
Should we not enable this by default? :thinking:

holgerd77:
> Should we not enable this by default? 🤔

Yes, would think as well, any downsides of this?

acolytec3:
The main one I can think of is if an RPC consumer expects the `internal error` message versus the very debug-oriented "Here's the error and its complete stack trace" response proposed above.

scorbajio:
I've dealt with this before, can take a look. 

One downside I can think of for keeping the rpc debug option enabled by default is that client users could forget to disable it in production, which could result in security issues since the stack trace would reveal details to an attacker.

PR #3059, opened 2023-09-24
Add Option to Return Actual Caught Error Message and Stack Trace when RPC Call Fails

scorbajio:
This change catches errors that may happen in RPC calls and returns them as well as a stack trace if the `rpcDebug` option is set to `true`. Also see #3044.


packages/client/src/rpc/modules/eth.ts
```diff
@@ -385,10 +387,12 @@ export class Eth {
       const { execResult } = await vm.evm.runCall(runCallOpts)
       return bytesToHex(execResult.returnValue)
     } catch (error: any) {
-      throw {
-        code: INTERNAL_ERROR,
-        message: error.message.toString(),
+      const e: any = {
```
acolytec3, line 390: This all looks great.  How hard is it to add a type to the `e` value that's thrown by each handler?

acolytec3 (commented):
This is looking good overall.  Did you explore any alternatives that don't require repeating essentially the same try/catch in every function call?  Would it be possible to put a new `callWithStackTrace` method in each class and then have it wrapp the underlying RPC method handlers in a try/catch and do that for every function call?  I guess the way I'm thinking about it, you'd have `eth.callWithStackTrace(this.getBlockByHash(...` or something like that.  And in the body of it, you just do: 
```
callWithStackTrace = (handler: async () => {}, debug: boolean) {
try {
  await   handler()
}
catch (err) {
  // Throw the error with or without stack trace
}
}
```
{

jochem-brouwer:
@acolytec3 I was thinking the same, I think this is possible in the middleware? If you catch it there you could generally report stack traces (I think)

packages/client/src/rpc/modules/eth.ts
scorbajio: Fixed

jochem-brouwer (commented):
The changes in the methods of debug/eth (besides constructor) are there because now the `callWithStackTrace` handles the errors (if `INTERNAL_ERROR`) right?

Also, the CI of client fails :thinking: 

In general LGTM but I think something goes wrong with error handling since client CI fails (have not looked in detail)

acolytec3 (approved):
LGTM.  Tests should all pass now.

PR #3063, opened 2023-09-27
Client: use same Cache Setup for normal and executeBlocks-triggered Execution

holgerd77:
This PR aligns the cache setup for the client for "normal" execution along sync or standalone and re-execution with the `--executeBlocks` flag so that the StateManager and trie cache setups match the cache setup announced by client CLI, react on cache flag usage (especially trie cache) and generally allows for better cache behavior debugging (so that `--executeBlocks` "mimics" the normal execution behavior and not somewhat behaves differently).

Implementation is done by  getting generally more flexible (with additional `shallowCopy()` method options), so that we but also users in general can choose if they want to have their shallow copy for rather short term or long term use. 

For trie I additionally generally generalized the options take in by using an options dict which replaces the plain keyPrefix parameter (still possible since not yet released). Then we can eventually add other options too there.

Open for review. 🙂


packages/trie/src/trie.ts
```diff
@@ -1043,19 +1044,20 @@ export class Trie {
    * Note on db: the copy will create a reference to the
    * same underlying database.
    *
-   * Note on cache: for memory reasons a copy will not
-   * recreate a new LRU cache but initialize with cache
-   * being deactivated.
+   * Note on cache: for memory reasons a copy will by default
+   * not recreate a new LRU cache but initialize with cache
+   * being deactivated. This behavior can be overwritten by
+   * expicitly setting `cacheSize` as an option on the method.
```
acolytec3, line 1050:
```suggestion
   * explicitly setting `cacheSize` as an option on the method.
```

packages/statemanager/test/stateManager.spec.ts
```diff
@@ -67,5 +71,23 @@ describe('StateManager -> General', () => {
       CacheType.ORDERED_MAP,
       'should switch to ORDERED_MAP storage cache on copy()'
     )
+    assert.equal((smCopy as any)._trie._opts.cacheSize, 0, 'should set trie cache size to 0')
```
acolytec3, line 74:
```suggestion 
 assert.equal(smCopy['_trie']._opts.cacheSize, 0, 'should set trie cache size to 0')
```
Can you replace these `any` usages with index keys so we retain the typing here and in the other test cases?

acolytec3 (changes requested):
Small nits but looks great overall.

holgerd77:
@acolytec3 thanks for the review 🙏, ok, have updated!

packages/statemanager/src/stateManager.ts
```diff
@@ -948,27 +948,36 @@ export class DefaultStateManager implements EVMStateManagerInterface {
...
    * instead, since copied instantances are mostly used in
    * short-term usage contexts and LRU cache instantation would create
    * a large overhead here.
-   * 2. Cache values are generally not copied along
+   * 2. The underlying trie object is initialized with 0 cache size
+   *
+   * Both adoptions can be deactivated by setting `downlevelCaches` to
+   * `false`.
+   *
+   * Cache values are generally not copied along.
```
acolytec3, line 965: What does this comment mean in context now?  It feels imprecise and doesn't make it clear in which circumstances (if any) the cache values actually would be copied.

Would it be better to say: `The underlying trie objet is initialized with 0 cache size and cache values are not copied over?`

packages/vm/src/vm.ts
```diff
@@ -239,13 +239,19 @@ export class VM {
   /**
    * Returns a copy of the {@link VM} instance.
    *
-   * Note that the returned copy will share the same db as the original for the blockchain and the statemanager
+   * Note that the returned copy will share the same db as the original for the blockchain and the statemanager.
+   *
+   * Associated state and trie caches will be deleted and caches will be re-initialized for a more short-term
+   * focused usage, being less memory intense. To fine-tune this behavior (if the shallow copy has a longer life
+   * span e.g.) you can set the `downlevelCaches` option to `false`.
```
acolytec3, line 246:
```suggestion
   * Unless `downlevelCaches` is set to `false`, the state and trie caches will not be copied over to the new `vm` instance and will be re-initialized using the `ORDERED_MAP` 
   * cache type which is intended for a more short-term focused usage, being less memory intense. To have the original caches copied over (if the shallow copy has a longer life
   * span e.g.) you can set the `downlevelCaches` option to `false`.
```

acolytec3: Feel free to ignore or modify this language but the original language feels a little unclear to me as to what we're actually doing with this param (at least for the uninitiated dev)

acolytec3 (commented):
Left a couple of more comments on the docs.  Feel free to take or leave them.

One more general question, do we ever actually copy the cache values over?  I don't see that option anywhere in the code and our docs seem to indicate it's possible to do a copy where the state of the cache is actually preserved.

packages/statemanager/src/stateManager.ts
holgerd77: Hmm, this does not only count for point 2.

I have adopted locally to: "Cache values are generally not copied along regardless of the `downlevelCaches` setting."

Will push soon.

packages/vm/src/vm.ts
holgerd77: Hmm, I had the argumentation a bit the other way around: I didn't want to go too much into the details because I would assume that someone dealing with the VM type is not aware about the cache datastructure type options over in the trie library. 🤔

I will do a hybrid and keep my sentence but add some details in brackets.

holgerd77:
@acolytec3 ok, updated 🙂, ready for re-review!

jochem-brouwer (approved):
LGTM

PR #3064, opened 2023-09-27
VM: Complement Profiler with out-of-EVM Measurements

holgerd77:
This PR complements the VM profiler by adding some selected output around the activities happening before and after block/tx execution. This is to get closer to the full picture of where time is is lost and not fall into side-optimization traps hunting for miliseconds in one place while seconds are lost somewhwer else when looking at the "whole thing" [tm]. 😜

Side goals where to both not overload the comprise and nice profiler output - I hope (and think) this should be ok with 3-4 extra lines per tx or block - also not overload the `runTx()` and `runBlock()` code modules with too much additional code, and at the same time be differentiated enough to give solid extra first hints where to look further.

Following are two example outputs from the client.

This one is running with tx profiling activated:

Bildschirmfoto 2023-09-27 um 12 35 34

And this one has activated block profiling:

Bildschirmfoto 2023-09-27 um 13 26 19

Open for review. 🙂


packages/vm/src/runBlock.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
jochem-brouwer, line 1: Can you disable eslint per-line? If we do this for the whole file then it is likely that (at least for me) `console.log ` leftovers get pushed to a PR.

packages/vm/src/runBlock.ts
```diff
@@ -42,17 +43,35 @@ const parentBeaconBlockRootAddress = Address.fromString(
   '0xbEac00dDB15f3B6d645C48263dC93862413A222D'
 )
 
+let enableProfiler = false
+let stateRootCPLabel = ''
+let processTxsLabel = ''
+let withdrawalsRewardsCommitLabel = ''
```
jochem-brouwer, line 49: I think we can directly initialize these to the constant (?) strings they use?

packages/vm/src/runBlock.ts
```diff
@@ -42,17 +43,35 @@ const parentBeaconBlockRootAddress = Address.fromString(
...
   const setHardfork = opts.setHardfork ?? false
   let { block } = opts
   const generateFields = opts.generate === true
 
+  if (enableProfiler) {
+    const title = `Profiler run - Block ${block.header.number} (${bytesToHex(block.hash())} with ${
+      block.transactions.length
+    } txs`
+    console.log(title)
+    stateRootCPLabel = 'New state root, DAO HF, checkpoints, block validation'
```
jochem-brouwer, line 71: Can we also time the entire `runBlock` method? So it can be seen what relatively takes the most time.

packages/vm/src/runTx.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
jochem-brouwer, line 1: Same comment here regarding disabling eslint no-console for the entire file.

packages/vm/src/runTx.ts
```diff
@@ -40,6 +41,12 @@ const { debug: createDebugLogger } = debugDefault
 const debug = createDebugLogger('vm:tx')
 const debugGas = createDebugLogger('vm:tx:gas')
 
+let enableProfiler = false
+let initLabel = ''
+let balanceNonceLabel = ''
+let logsGasBalanceLabel = ''
+let cleanupAndReceiptsLabel = ''
```
jochem-brouwer, line 48: Initialize those here too?

packages/vm/src/runTx.ts
```diff
@@ -61,6 +68,17 @@ function execHardfork(
...
 export async function runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
+  if (this._opts.profilerOpts?.reportAfterTx === true) {
+    enableProfiler = true
+  }
+
+  if (enableProfiler) {
+    const title = `Profiler run - Tx ${bytesToHex(opts.tx.hash())}`
+    console.log(title)
+    initLabel = 'EVM journal init, address/slot warming, fee validation'
+    console.time(initLabel)
```
jochem-brouwer, line 79: Also here, I think it would be nice to profile the entire `runTx` method too.

jochem-brouwer (commented):
Some comments

jochem-brouwer (approved):
Will approve, assuming CI passes. Addressed the review.

holgerd77:
Updated this via UI

PR #3066, opened 2023-09-28
FindPath-optimize

ScottyPoi:
replacing #3017

#### `findPath()`

This PR refactors `trie.findPath` with the goal of performance optimization.

* Reduces use of `slice` and other array mutations
  * The key nibbles are now only parsed once, and the array is never mutated.
  * Progress towards the target is now tracked with an index pointer
* Reduces redundant nibble comparisons 
  * Each key nibble is now checked only once, in order.
* Reduces memory reallocation  
  * The stack array is now allocated to the max length at construction.
  * Nodes are written to stack[progress] instead of pushed with stack.push()
  * Final returned array is filtered once to remove `undefined`


holgerd77:
Oh. Unfortunately discovered this late my day, but would love to have a look! 🤩

ScottyPoi:
Ran a straightforward bechmark comparison between this branch and master.  
1. Built some very large Tries with keys of different sizes, and hashing turned on/off.
2. Ran `findPath` on each Trie with a very large set of keys, with 50% present / 50% not-present.
3. Record the `process.hrtime` nanoseconds for each Trie processing the set of `findPath` calls.
4. Compare the duration for each example performed in `master` vs. `branch`

This sort of approach consistently showed **1%-4%** performance increase from this PR :rocket:

holgerd77:
Updated this via UI

holgerd77:
Another branch update via UI

holgerd77:
Very cool, yes, I just run this on the same mainnet blocks I used for testing #3022 and while results are not 100% clear (always difficult in these ranges of eventual optimizations), there is at least some indicication to confirm your results! 🤩

So this is what I compare with:

npm run client:start -- --sync=none --executeBlocks=1600000-1601000 (1000 blocks, master)
Minutes (total): 2:07 1:55 1:54

And on this branch:
npm run client:start -- --sync=none --executeBlocks=1600000-1601000 (1000 blocks, master)
Minutes (total): 1:51 1:56  1:53

The 2:07 might be a first run outlier, but the other results give some indication that results remain 1-3 seconds faster. Not a 100% conclusion, but I wouldn't want to add here with more runs.

Think we can very safely take this in, thanks a lot! 🙏

holgerd77 (approved):
LGTM

PR #3069, opened 2023-09-28
Simplify `client` transports

acolytec3:
This PR entirely removes the `transports` CLI parameter and has a number of related changes:
- Networking (i.e. RLPx server and discv4/dns discovery) is now not activated when `--sync=none` is passed in the command line
- Converts the `servers` property of the `Config` type into a single `server` and adjusts related code
- Removes the `transports` property from `Config` cleans up boatloads of tests 
where the `transports` property was set


holgerd77 (approved):
Nice, super clean implementation and so much removed code, love this! 🤩

Will merge.

PR #3070, opened 2023-09-28
Update dependency to fix high severity vulnerability

scorbajio:


acolytec3 (approved):
LGTM.  FYI - @holgerd77 this only impacted vitest, and not our production deps so shouldn't require a release.

holgerd77:
Nice, nevertheless thanks for addressing! 🙏🙂

Issue #3071, opened 2023-09-28
Missing "@types/debug" dependency in packages that require it, like `evm`


davidmurdoch:
Since one of the packages does have it in its package.json everything works fine from within the monorepo, but when a package that exports the `debug` module's types is used on its own the type will be missing and invalid in some scenarios.

Example of a problematic type is the `PrecompileInput` type, which exports the `debug` type on the `_debug` property.

If you want to export the `debug` type anywhere I think you'll need to add the `@types/debug` package as a (production) dependency.

acolytec3:
Thanks for raising this.  Looks like this happens in a number of places across the monorepo.  Will take a look and see if we can address.

acolytec3:
I've reviewed all the places where a property of type `debug` or `debugger` (or other references to the `debug` package in general are used) and it looks like you've found the one place where it's not a `private` or `protected` property of one of our classes.  It just so happens that in this `PrecompileInput` interface, someone might actually need to provide a debugger to print debug logs if implementing a custom precompile.  Thanks again for raising this.

jochem-brouwer:
Closed by #3072 (?)

PR #3072, opened 2023-09-29
Add `debug` types to `evm`

acolytec3:
As noted in #3071, the `PrecompileInput` interface exposes a `_debug` property which is of type `debug`.  Since this `PrecompileInput` is be used when somebody is implementing a custom precompile, it makes sens to add the `debug` types to the production dependencies so Typescript can resolve these types.

It also updates the dependency version in two other packages where it is a `devDependency`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2475,10 +2475,9 @@
       "dev": true
     },
     "node_modules/@types/debug": {
-      "version": "4.1.8",
-      "resolved": "https://registry.npmjs.org/@types/debug/-/debug-4.1.8.tgz",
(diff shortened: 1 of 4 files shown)


acolytec3:
cc @davidmurdoch

holgerd77:
Rebased this via UI

holgerd77 (approved):
Cool, makes sense, thanks for adressing so quicky!

holgerd77:
@davidmurdoch merged, we'll do a release round next week, so this will get in relatively quickly! 🙂

PR #3074, opened 2023-09-29
genesis: fix holesky genesis state

g11tech:
as also noted by a discord user (Sea Monkey) , the client wasn't working on `--network holesky` with state trie error of not finding stateroot on execution of block 1.

on debugging realized that the genesis state and the expected stateroot doesn't match, on further inspecton and debugging saw beacon contract was missing from genesis. 

this PR adds the same as well as modify genesis spec to test expected stateroots of all known networks

#### tested with lodestar for holesky, seems to sync blocks now


acolytec3 (approved):
LGTM!

PR #3075, opened 2023-09-29
Make browser binaries optional

acolytec3:
Makes browser binaries optional in package.json


g11tech (approved):
lgtm

holgerd77:
Can we see that we add some dev docs (in the root monorepo install docs I guess?) for something like this so that we don't forget ourselves and others remain able to reproduce? Otherwise great PR! 😃👍

PR #3076, opened 2023-09-30
client: make the newpayload execution of big blocks non blocking

g11tech:
on devnet9 there is  a range of blocks 1700-2000 where blocks have txs ~1400+. This causes newpayload api to timeout for the CLs causing them to be stuck and/or retry the same blocks over and over again.

This PR, responds `SYNCING` to heavy blocks (for ethereumjs) which can then be independently executed while letting the CL sync moveforward.

tested on devnet-9 with lodestar leading to fast resolution of sync from genesis

### DIFF
--- a/packages/client/test/rpc/engine/newPayloadV1.spec.ts
+++ b/packages/client/test/rpc/engine/newPayloadV1.spec.ts
@@ -299,7 +299,7 @@ describe(method, () => {
     }
 
     // set the newpayload limit to 100 for test
-    chain.config.engineNewpayloadMaxTxsExecute = 100
+    ;(chain.config as any).engineNewpayloadMaxTxsExecute = 100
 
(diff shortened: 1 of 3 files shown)


packages/client/test/rpc/engine/newPayloadV1.spec.ts
```diff
@@ -256,6 +256,67 @@ describe(method, () => {
...
+      engine: true,
+    })
+
+    const transactions = Array.from({ length: 101 }, (_v, i) => {
+      const tx = FeeMarketEIP1559Transaction.fromTxData(
+        {
+          nonce: i,
+          maxFeePerGas: '0x7',
+          value: 6,
+          gasLimit: 53_000,
```
acolytec3, line 284: I didn't realize JS allowed underscores in numbers

acolytec3 (approved):

acolytec3 (approved):
LGTM

PR #3077, opened 2023-09-30
client: track and respond to invalid blocks in engine api and other hive engine-cancun fixes

g11tech:
in cancun-engine hive tests, there are lots of `Invalid` scenario tests where the new payload evaluates to `INVALID` but we don't respond accordingly in fcU (rather end up throwing error instead of `INVALID` status)

this PR caches invalids and use it in fcU responses.

What scenario this PR changes/Fixes
 - add invalid block cache populated in new payload and use it in fcu
 - skip adding the payloads generated in getPayload to the chain without going through new payload
 - add a cli option to run hive without engine pruning via `--pruneEngineCache false` as some tests require deep cache
 - fix validHash lookup
 - cache remoteblocks early in newpayload after successful block and parent block validations
 

TODO
- [x] some related invalid responses for hive are still being debugged and fixed

#### Progress
From 69 failing cases 
image

this PR reduces to 61 failing cases
image

#### Clarifications requested
Requested some clarifications for the cases where ethereumjs seems to be behaving correctly:
- [ ] the fields validations happen before the call is actually processed, this test case expects to process fcUv3 even if the parentBeaconBlockRoot is missing (https://discord.com/channels/595666850260713488/1155879392732520578/1157960831204597761)
- [ ] for another test case: Test: Invalid Transaction Nonce NewPayload - Syncing (Cancun) (ethereumjs_cancun-git)the block is rejected with this error: Error: nonce cannot equal or exceed MAX_UINT64 (2^64-1) (https://discord.com/channels/595666850260713488/1155879392732520578/1157963080035209256)
  - Further discussion on this lead to acknowledgment that INVALID should be acceptabled response, ALSO this would resolve most of our invalid transaction failure cases 
- [ ] for "engine-cancun/Replace Blob Transactions the 1st and 3rd txs are correctly selected along with their blobs, while the test expects to only get 1 blob(for just 1 tx) (https://discord.com/channels/595666850260713488/1155879392732520578/1158056029091811438)

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -362,6 +362,11 @@ const args: ClientOpts = yargs(hideBin(process.argv))
     describe: 'path to a file of RLP encoded blocks',
     string: true,
   })
+  .option('pruneEngineCache', {
+    describe: 'Enable/Disable pruning engine block cache (disable for testing against hive etc)',
+    boolean: true,
+    default: true,
+  })
   .completion()
   // strict() ensures that yargs throws when an invalid arg is provided
   .strict().argv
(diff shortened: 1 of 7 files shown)


acolytec3 (approved):
LGTM!  We can add more fixes as future hive tests come on line.

holgerd77:
Pretty cool PR, just seeing! 🙏

PR #3078, opened 2023-10-02
client: fix canonical reset of the chain by the skeleton

g11tech:
on devnet9, because of some reasons skeleton was resetting the chain and these are the remedial fixes that have been deployed on the nodes there

BUILD on top of 
 - https://github.com/ethereumjs/ethereumjs-monorepo/pull/3077

TODO
 - [ ] monitor and see if this fixes the issues there


acolytec3 (approved):
LGTM.

PR #3079, opened 2023-10-03
client: Add eth_coinbase RPC implementation

scorbajio:
This change implements the `eth_coinbase` rpc endpoint. Also see the [JSON RPC API](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_coinbase).

### DIFF
--- a/packages/client/src/rpc/modules/eth.ts
+++ b/packages/client/src/rpc/modules/eth.ts
@@ -524,7 +524,7 @@ export class Eth {
         message: 'Coinbase must be explicitly specified',
       }
     }
-    return cb
+    return cb.toString()
   }
 
   /**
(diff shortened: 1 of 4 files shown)


packages/client/src/rpc/modules/eth.ts
```diff
@@ -509,6 +511,22 @@ export class Eth {
     return bigIntToHex(account.balance)
   }
 
+  /**
+   * Returns the currently configured chain id, a value used in replay-protected transaction signing as introduced by EIP-155.
+   * @param _params An empty array
```
scorbajio, line 516:
```suggestion
```

packages/client/src/rpc/modules/eth.ts
```diff
@@ -509,6 +511,22 @@ export class Eth {
     return bigIntToHex(account.balance)
   }
 
+  /**
+   * Returns the currently configured chain id, a value used in replay-protected transaction signing as introduced by EIP-155.
```
scorbajio, line 515:
```suggestion
   * Returns the currently configured coinbase address.
```

PR #3080, opened 2023-10-04
Code cache should be able to save prestate without reading from statemanager DB

scorbajio:


jochem-brouwer:
Is it correct that only tests have been changed here?

scorbajio:
> Is it correct that only tests have been changed here?

After some time debugging the code cache `_saveCachePreState` function to figure out how to remove the `codeExists` parameter dependency, I discovered that the account cache is also failing some test cases that weren't being caught because some promises weren't being awaited when being passed into assert statements. When awaiting them, it can be seen that the account cache fails some of the same test cases as the code cache. Both caches have a flaw where the revert doesn't revert to undefined when multiple puts are performed for the same address in the same checkpoint.

To my current understanding the issue with both these caches seems to be that after saving prestate for an account or code that isn’t in the cache, the prestate gets overwritten when looking up if the item is in the diffCache the second time the same address gets put because the diffCache  returns undefined for the key, not because it doesn’t exist in the diffCache but because it’s recorded value in the diffCache is undefined.

jochem-brouwer:
I find it very surprising that the tests now fail, we use the account cache in the client and there do not seem to be any consensus bugs there :thinking:

scorbajio:
> I find it very surprising that the tests now fail, we use the account cache in the client and there do not seem to be any consensus bugs there 🤔

Yes, I also remember that before implementing the `codeExists` fix to address the failing tests cases, the code cache was passing all other tests, which I also find surprising. It's potentially a rare case to e.g. put an account or code for a key twice back-to-back right at the beginning of a checkpoint and then revert.

scorbajio:
It seems like the issue that was discovered with both the account and code cache was that the `revert` and `_saveCachePreState` functions were checking for an `undefined` value when they should have been checking a cache element field (`code` for the code cache and `accountRLP` for the account cache) that is set when an item is initially recorded as missing from the cache in `_saveCachePreState`. Both these issues have been resolved, and account tests that were previously silently failing have been updated to fail explicitly. The code cache also does not read from the database anymore on a put for saving prestate.

packages/statemanager/src/cache/code.ts
```diff
@@ -53,13 +53,8 @@ export class CodeCache extends Cache {
    *
    * @param cacheKeyHex Account key for which code is being modified.
    * @param currentPutCode New value of code that is being saved for the account.
```
scorbajio, line 55:
```suggestion
```

holgerd77:
Updated this via UI

packages/statemanager/src/cache/account.ts
```diff
@@ -57,7 +57,7 @@ export class AccountCache extends Cache {
       } else {
         oldElem = this._orderedMapCache!.getElementByKey(cacheKeyHex)
       }
-      this._diffCache[this._checkpoints].set(cacheKeyHex, oldElem)
+      this._diffCache[this._checkpoints].set(cacheKeyHex, oldElem ?? { accountRLP: undefined })
```
holgerd77, line 60: I would still think this is a fix "in the wrong direction", and I wonder why this is working now.

See the comment/differentiation in the code docs above:

>    * If the whole cache element is undefined (in contrast
>   * to the account), the element didn't exist in the cache
>  * before.

As I wrote in some explanations before, the diff cache should only be about the diff *in the cache*. We now artifically set here the information that the account does not exist (at all, so in the state) for the prestate.

We just do not know this (and also don't want to know this at this point). We are just reading the old cache element and if this returns `undefined` the pre state for the cache is just `undefined` as well, transporting the information that we do not know anything about this accout at this pre state point.

packages/statemanager/src/cache/account.ts
```diff
@@ -172,7 +172,7 @@ export class AccountCache extends Cache {
     for (const entry of diffMap.entries()) {
       const addressHex = entry[0]
       const elem = entry[1]
-      if (elem === undefined) {
+      if (elem?.accountRLP === undefined) {
```
holgerd77, line 175: So the wrong fix from above might actually work when here all the `elem?.accountRLP` entries are deleted.

This has a bad side effect on the cache though since it deletes for all accounts where `elem?.accountRLP === undefined` the information from the cache that this account does not exist in the state (which is *also* a state information).

holgerd77 (changes requested):
On this first look I am pretty sure this is not yet the correct fix.

Maybe you can see on first round if you can fix your code along the lines I outlined here, otherwise I can also have another look early next week if I detect the overall flaw in the applied code logic.

packages/statemanager/src/cache/account.ts
scorbajio: I changed to a more intuitive fix, since the overwriting of the `undefined` value was the real issue at the heart of the problem.

holgerd77:
Updated this via UI

holgerd77 (approved):
LGTM, thanks a lot! 🙏

Issue #2938, opened 2023-08-04
Investigate hive problems with enabling websocket


jochem-brouwer:
In this PR https://github.com/ethereum/hive/pull/813 at hive we fix the timeout on startup since hive keeps trying to connect to RPC, but it cannot reach it. However, with the newly introduced arguments, we have to disable the websocket server, because it throws if we enable it. We should fix this.

holgerd77:
Can you paste a concrete CL client start command which throws so to make sure I (others too of course) am getting this right?

jochem-brouwer:
Yes, we had to disable websocket on hive in order to make hive / our client not hang on tests, see;

https://github.com/ethereum/hive/blob/44510de419e0f58dba60a39d27636362f731a184/clients/ethereumjs/ethereumjs.sh#L51

Note that we launch `--ws false`. If you remove this (so just use `--ws`, remove the `false`) then the client will start but immediately complain about reused ports. (If you want I can post the error here as well, but this should turn up if you remove the `false` from the ws and then just run any hive test, e.g. ./hive --sim "ethereum/engine" --client ethereumjs --sim.limit="engine-cancun/Request Blob Pooled Transactions")

holgerd77:
Ok, so, to clarify further: so this is only a specific Hive thing and we do not have a very generic websocket activation bug in the client right now? Is this correct?

jochem-brouwer:
Yes. Try this locally (after rebuilding monorepo)

`node ./dist/bin/cli.js --rpc --rpcEngine --rpcAddr 0.0.0.0 --rpcEngineAddr 0.0.0.0 --rpcEnginePort 0.0.0.0 --rpcEnginePort 8551 --ws --dataDir=./data`

Crashes after a few seconds:

```
08-14|14:34:30] INFO Started JSON RPC Server address=ws://localhost:8545 namespaces=eth,web3,net,admin,txpool,debug 
[08-14|14:34:30] ERROR Error starting client The argument 'options' must have the property "port" or "path". Received [ NaN, 8551 ]
node:events:491
      throw er; // Unhandled 'error' event
      ^

Error: listen EADDRINUSE: address already in use 0.0.0.0:8545
    at Server.setupListenHandle [as _listen2] (node:net:1740:16)
    at listenInCluster (node:net:1788:12)
    at doListen (node:net:1937:7)
    at process.processTicksAndRejections (node:internal/process/task_queues:83:21)
[...]

PR #3083, opened 2023-10-04
Guard against rpc port collisions

acolytec3:
Address #2938 
 - Adds new default port settings for websocket RPC to avoid colliding with HTTP RPC defaults
 - Adds new guards in CLI arg parsing to ensure that colliding port values aren't set between WS and HTTP RPCs


jochem-brouwer (approved):
Assuming CI passes, LGTM

jochem-brouwer (approved):

holgerd77:
So will this go into Hive automatically or do we need to do something for it? 🤔 And if so, which tests are affected?

acolytec3:
I can't envision it causing any issues for hive.  Is Hive using websockets
anywhere in its tests?

holgerd77:
Ah, sorry I read from the chat that this would fix some Hive tests, have likely misread your guys' exchange.

PR #3085, opened 2023-10-05
client: skeleton reorg head fixes and log improvements

g11tech:
fixes to deal with a bugs:
 - new payload annouces a new reorging block but fcU can't figure that out since newpayload always used to write the block to skeleton canonical chain, however it would only write forward new payload annoucements i.e. blocks > current head so that fcU can detect the reorg
 - fixes to not backstep the canonical fill because of any tail reorgs by setHead/putBlocks and exit current fill if canonicalHead is requested to be reset (skeleton tail resets)

TODO:
- [ ] some testing on devnet9/holesky


acolytec3:
Can you revert the `ethereum-tests` commit hash change?  I'm assuming that was an accident.

acolytec3 (dismissed):
LGTM

acolytec3 (approved):
LGTM

packages/client/src/execution/vmexecution.ts
```diff
@@ -528,7 +528,7 @@ export class VMExecution extends Execution {
                 this.config.execCommon.gteHardfork(Hardfork.Paris) === true
                   ? ''
                   : `td=${this.chain.blocks.td} `
-              this.config.logger.info(
+              this.config.logger.verbose(
```
holgerd77, line 531: What is this change doing? 🤔 I am not even finding information about this log level (when googling)?

packages/client/src/rpc/modules/engine.ts
```diff
@@ -548,6 +551,27 @@ export class Engine {
...
+        peers: (this.service.beaconSync as any)?.pool.size,
+      })
+
+      // void this.skeleton.isLastAnnoucement().then((lastAnnouncement) => {
+      //   if (lastAnnouncement || ) {
+      //     if (lastAnnouncement) {
+      //       void this.skeleton.logSyncStatus('status', true)
+      //     }
+      //   }
+      // })
```
holgerd77, line 574: This code is commented out?

packages/client/src/execution/vmexecution.ts
holgerd77: Ah, just testing myself and seeing that this seems to really removing the log message for default setup. So this is one of the most central logging output I am using all the time.

I would urgently want to get this in again, otherwise it is not really possible anymore to follow on a "classic sync" (mainnet), especially since import and execution deviates so much for our client.

Will re-introduce along some other follow-up client log work.

holgerd77: (so for pre-Merge this is definitely necessary)

holgerd77: Another update: ok, but going a bit back here, my client just switched over to now doing the execution and for this situation I also definitely want to see what is going on on the execution side:

grafik

So I would rather want to re-introduce, at least for most contexts.

packages/client/src/service/skeleton.ts
```diff
@@ -850,14 +889,100 @@ export class Skeleton extends MetaDBManager {
...
+    } else {
+      this.config.logger.debug(
+        `${logPrefix} ${status} linked=${
+          this.status.linked
+        } subchains=${this.status.progress.subchains
+          .map((s) => `[head=${s.head} tail=${s.tail} next=${short(s.next)}]`)
+          .join(',')} reset=${this.status.canonicalHeadReset} ${chainHead}`
+      )
+    }
+    return status
```
holgerd77, line 985: Great sync improvements! 🤩

Some formatting aspects: this line is (far) too long though now, doesn't fit for normal laptop setups any more.

Can we split this in two lines or alternatively do we need *all* the info there? Maybe also a way to present some information only periodically? Just options.

So line should really not go much further than what is now `linked=true`.

PR #3086, opened 2023-10-05
Client: Small UX Improvements

holgerd77:
This is a series of small UX improvements and fixes to make the client CLI output more digestable.

Ready for review. 🙂

### DIFF
--- a/packages/client/src/config.ts
+++ b/packages/client/src/config.ts
@@ -653,6 +653,13 @@ export class Config {
     return key
   }
 
+  superMsg(msg: string, meta?: any) {
+    const len = msg.length
+    this.logger.info('-'.repeat(len), meta)
+    this.logger.info(msg, meta)
+    this.logger.info('-'.repeat(len), meta)
+  }
+
   /**
    * Returns specified option or the default setting for whether DNS-based peer discovery
    * is enabled based on chainName. `true` for goerli
--- a/packages/client/src/rpc/util/CLConnectionManager.ts
+++ b/packages/client/src/rpc/util/CLConnectionManager.ts
@@ -38,8 +38,8 @@ type ForkchoiceUpdate = {
 
 type PayloadToPayloadStats = {
   blockCount: number
-  minBlockNumber?: BigInt
-  maxBlockNumber?: BigInt
+  minBlockNumber?: bigint
+  maxBlockNumber?: bigint
   txs: { [key: number]: number }
 }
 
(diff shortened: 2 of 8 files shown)


packages/client/src/config.ts
```diff
@@ -653,6 +653,13 @@ export class Config {
     return key
   }
 
+  superMsg(msg: string, meta?: any) {
```
acolytec3, line 656: Super Message?  :-)

acolytec3: I'd like some more of these.

acolytec3 (commented):
LGTM!

acolytec3 (approved):
Have added some simple tests to bump the coverage in teh `CLConnectionManager` class so hopefully we hit the code coverage requirement.  LGTM!

g11tech:
ahh i might have done some of my own logging improvement here: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3085

packages/client/src/config.ts
holgerd77: Inventing new names and words is one of my favorite parts of being a developer! 😂

> I'd like some more of these.

Yes, that super simple change has an impressive effect on the possibility to orientate, right?

I was a bit cautious to not directly overload the output, also only did the first very very apparent ones, but we will likely stumble upon more "inflection points" which we want to take in here.

holgerd77: ("Client synced" or whatever we have here is for sure another one e.g.)

packages/client/src/rpc/util/CLConnectionManager.ts
```diff
@@ -248,7 +251,7 @@ export class CLConnectionManager {
     if (
       [ConnectionStatus.Disconnected, ConnectionStatus.Uncertain].includes(this.connectionStatus)
     ) {
-      this.config.logger.info('Consensus client connection established', { attentionCL: 'CL' })
+      this.config.superMsg('Consensus client connection established', { attentionCL: 'CL' })
```
holgerd77, line 254: Side note: this is appearing relatively often now as a super msg, but this is rather a problem with the disconnect-detection-logic where we have some flaw or missing some connected-timestamp update (so it's not so simple to just make the threshold larger, since the current one (30 seconds or so) already does not match at all with client interaction in some occasions.

Might have a quick look here too on some follow up PR though.

Issue #3087, opened 2023-10-05
Noisy `client` logs


acolytec3:
We've got some broken/noisy client logs and we should fix this:
Issue 1- New hardfork reached message is being logged on I guess every new block when we are at the head of the chain, like this:
![image](https://github.com/ethereumjs/ethereumjs-monorepo/assets/17355484/e6ca61d7-c3ff-45a7-a77b-704627a57505)

2) These Payload Stats logs should probably be debug or verbose level logs.  They're logged by the `CLConnectionManager` and they don't seem to provide any meaningful info for understanding basic node operations
![image](https://github.com/ethereumjs/ethereumjs-monorepo/assets/17355484/6c91854b-c483-43c3-96d8-e3ab832387bf)

holgerd77:
Here are some traces for the HF back-and-forth switch:

```shell
[10-06|14:15:40] INFO [ CL ] ---------------------------------------
[10-06|14:15:42] INFO [ CL ] Executed blocks count=1 first=51763 hash=0x7403…6e5d baseFee=7 hardfork=shanghai last=51764 hash=0x051b…821a txs=2
Trace
    at EventEmitter.<anonymous> (ethereumjs-monorepo/packages/client/dist/src/blockchain/chain.js:125:21)
    at EventEmitter.emit (node:events:513:28)
    at Common.setHardfork (ethereumjs-monorepo/packages/common/dist/cjs/common.js:220:33)
    at Common.setHardforkBy (ethereumjs-monorepo/packages/common/dist/cjs/common.js:340:14)
    at Blockchain.checkAndTransitionHardForkByNumber (ethereumjs-monorepo/packages/blockchain/dist/cjs/blockchain.js:1043:21)
    at Chain.putBlocks (ethereumjs-monorepo/packages/client/dist/src/blockchain/chain.js:279:39)
    at async ethereumjs-monorepo/packages/client/dist/src/execution/vmexecution.js:219:13
    at async VMExecution.runWithLock (ethereumjs-monorepo/packages/client/dist/src/execution/vmexecution.js:91:27)
    at async Engine.forkchoiceUpdated (ethereumjs-monorepo/packages/client/dist/src/rpc/modules/engine.js:900:17)
[10-06|14:15:42] INFO [ CL ] ----------------------------------------------------------------
[10-06|14:15:42] INFO [ CL ] New hardfork reached 🪢 ! hardfork=mergeForkIdTransition block=0
[...]

PR #3090, opened 2023-10-06
client: fix synced log consistency and add short status summary

g11tech:
- the SYNCED log was shuffling between SYNCING and SYNCED because on the race between fcU and fillCanonicalChain, now the fcu awaits the fill if chain segment to fill is small (2 blocks)
 - add a small explanation to whats going on in the sync for user to comprehend the data 
image

### DIFF
--- a/packages/client/src/service/skeleton.ts
+++ b/packages/client/src/service/skeleton.ts
@@ -1009,9 +1009,7 @@ export class Skeleton extends MetaDBManager {
         let left = this.bounds().tail - BIGINT_1 - this.chain.blocks.height
         if (this.status.linked) left = BIGINT_0
         if (left > BIGINT_0) {
-          if (this.pulled === BIGINT_0) {
-            this.config.logger.info(`Beacon sync starting left=${left}`)
-          } else {
+          if (this.pulled !== BIGINT_0 && fetching === true) {
             const sinceStarted = (new Date().getTime() - this.started) / 1000
             beaconSyncETA = `${timeDuration((sinceStarted / Number(this.pulled)) * Number(left))}`
             this.config.logger.debug(
--- a/packages/client/src/service/skeleton.ts
+++ b/packages/client/src/service/skeleton.ts
@@ -81,6 +81,8 @@ export class Skeleton extends MetaDBManager {
   private status: SkeletonStatus
 
   private started: number /** Timestamp when the skeleton syncer was created */
+  private lastFcuTime: number = 0
+  private syncstarted: number = 0
   private pulled = BIGINT_0 /** Number of headers downloaded in this run */
   private filling = false /** Whether we are actively filling the canonical chain */
 
@@ -333,6 +335,9 @@ export class Skeleton extends MetaDBManager {
       if (this.started === 0) {
         throw Error(`skeleton setHead called before being opened`)
       }
+      if (!init) {
+        this.lastFcuTime = Date.now()
+      }
 
       this.config.logger.debug(
         `New skeleton head announced number=${head.header.number} hash=${short(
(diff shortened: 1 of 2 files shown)


packages/client/src/service/skeleton.ts
```diff
@@ -914,7 +938,63 @@ export class Skeleton extends MetaDBManager {
...
+    const status = isValid
+      ? 'VALID'
+      : isSynced
+      ? executing === true
+        ? `EXECUTING`
+        : `SYNCED`
+      : `SYNCING`
+
+    if (peers === undefined || peers === 0) {
+      this.syncstarted = 0
```
holgerd77, line 951: Can we do camel case here?

holgerd77 (commented):
Cool, some client test(s) failing though, so need an additional push/CI run.

Can we take the occasion and once go through this new log line? As stated in the other PR, in the current form the line is far too long, and should be basically cut into half.

I will post some questions/suggestions in a separate post.

holgerd77:
grafik

Will repost one log line here, then I can easier reference.

So, I'll try to both understand all and go through in a detailed way:

1. This `cl=51686` information seems redundant, at least at this status, so not sure. Can we remove one of these?
2. `vm=15` is this a block number?
3. `beaconsync=filling` might this be redundant too and we can remove?
4. I would log the peer stats separately, this is something different and this wants too much for one line (I know Lodestar has it like this too, nevertheless)
5. What is `reorg` and `chain=false`? Is this needed?
6. I think the `subchains` info might be a good candidate for a separate log message so that this one gets reduced and shorter? Since this is also some very separate info?

holgerd77:
Ok, I think I am seeing better where you are heading with this new status line, since in these after-merge-scenarios we are heading more to a world (and usage of our client) where we are (especially when we are hopefully soon deploying snap sync) already close to the head of the chain and things become less of a linear work-down-the-whole-chain thing (import, execute, import, execute,...) but something where one would want a comprise status update on where the client state is.

Also seeing that this "execute blocks" line is distracting in this context. 🙂 Nevertheless think there should also not be "one-monster-line-to-rule-them-all". 🙂 

So maybe as an refined/further devoloped suggestion:

1. Post-merge we switch to log the "Execute blocks" message under `debug`
2. Eventually we can already shorten the log line by some portion by applying some of the questions/suggestions from above
3. We log the subchain status separately
4. I guess this delimiter line `[ CL ] ---------------------------------------` is then also not fitting in so well, would be good to delete

Does this make sense? 🤔

holgerd77:
grafik

Since this always comes in the form of a double message anyhow: might it make sense to move the "New consensus payload received..." msg to `debug` log level and only keep `New chain head` 🤔?

Then we should take over (in the form of: logging in both msgs) minimally `number=...` and `txs=...` I would say.

g11tech:
> Also seeing that this "execute blocks" line is distracting in this context. 🙂 Nevertheless think there should also not be "one-monster-line-to-rule-them-all". 🙂

yes we need a short analysis of whats going on, i have tried to add stuff that i want to see if someone is saying their client isn't working/syncing with this short summary

>     1. Post-merge we switch to log the "Execute blocks" message under `debug`
>
yes makes sense
>     2. Eventually we can already shorten the log line by some portion by applying some of the questions/suggestions from above
> 
>     3. We log the subchain status separately

i have made it multiline and should address 2,3 both 
image


>     4. I guess this delimiter line `[ CL ] ---------------------------------------` is then also not fitting in so well, would be good to delete
 how about just an empty line to give space
> 
> Does this make sense? 🤔

holgerd77:
> how about just an empty line to give space

Hmm, the problem with a one line separator only below is that this structurally separates *all* above the line from the stuff below the line, so this leads to a somewhat confusing layout. 

Not sure if this is the solution: but maybe really be gracious and do above *and* below and for that generalize my superMsg() method to take in an array of log messages to then do the line delimiter above, print out all log messages from the array separately and then do the line below?

Or will this get *too* spacious? 🤔

g11tech:
> And finally some traces for the HF back and forth switches: [#3087 (comment)](https://github.com/ethereumjs/ethereumjs-monorepo/issues/3087#issuecomment-1750623477)
> 
> Maybe you can directly make something of it.

thanks fixed it :)

PR #3096, opened 2023-10-07
monorepo: add wallet to readme

### LINKED ISSUE #3051: README does not mention wallet dependency
There are two sections of README that can mention wallet: `packages` and `package dependency relationship`. Also, GitHub label.

The question is whether it needs to be mentioned at all.

gabrocheleau:
Resolves #3051 by adding mention of the wallet package to the general README. 

Concerning the dependency graph, the wallet package only imports from util, which we never seem to mention in that graph, so I left that out.


gabrocheleau:
Just noticed that I need to fix one of the links, on my phone now but will do when back home.

acolytec3 (approved):
LGTM

PR #3097, opened 2023-10-07
monorepo: add genesis to readme

gabrocheleau:
Adds the genesis package to the readme

PR #3098, opened 2023-10-08
Add genesis workflow

scorbajio:
This change adds a workflow for the genesis package so that the package's tests are automatically run as a part of CI workflows.


acolytec3:
Closing this as commit was included in #3097

PR #3099, opened 2023-10-10
Client: hive Cancun fixes

jochem-brouwer:
This PR is WIP.

The goal of this PR is to fix most (all?) of failing hive cancun tests. Before this PR: 55 failing tests. After this PR: 30 failing tests

To run:

`./hive --client ethereumjs --sim ethereum/engine --sim.parallelism 8 --sim.limit="engine-cancun"`

(change parallelism and ensure hive is at latest `master`)

Fixes:

- `./hive --client ethereumjs --sim ethereum/engine --sim.parallelism 8 --sim.limit="engine-cancun/Bad Hash on NewPayload"` 104e6e88cdd87bf4c7bbaa8f6ee0955be5fc2649, this fixes 3 tests.
- `Inconsistent Head in ForkchoiceState` in b525804a5e93f64429a1b7ab0e85118defd31c4d, this commit fixes 10 tests. If `recursivelyFindParents` hits genesis block then we should stop finding more parents, we now have a full reorg back to genesis (?)
- `Pre-Merge Fork Number > 0 `: Off by one check in common f2aedcac4ca5cda066733b884a58d0e81ded17ed
- `Unique PayloadID - Withdrawals`: Ensure that different withdrawals result in a different payload ID 14b60a4943e56d09bf4e9975305c4674dc02c6fd

Open failing tests (with notes):

- `Invalid PayloadAttributes: Missing BeaconRoot` -> This test seems wrong, it FCUs without a `parentBlockBeaconRoot` but then expects us to report that blockhash as head hash
- `Invalid Missing Ancestor ReOrg` -> have to do with reorgs, will open a separate PR for this.


jochem-brouwer:
A client test will fail, this expects us to set finalized/safe to the genesis block if it is not yet set. I am not sure if this is correct. Remainder of this is ready for review, other failing hive tests have to do with reorgs, will open a separate PR for that.

packages/common/src/common.ts
```diff
@@ -384,7 +384,7 @@ export class Common {
       if (mergeIndex >= 0 && td !== undefined && td !== null) {
         if (hfIndex >= mergeIndex && BigInt(hfs[mergeIndex].ttd!) > td) {
           throw Error('Maximum HF determined by total difficulty is lower than the block number HF')
-        } else if (hfIndex < mergeIndex && BigInt(hfs[mergeIndex].ttd!) <= td) {
+        } else if (hfIndex < mergeIndex && BigInt(hfs[mergeIndex].ttd!) < td) {
```
jochem-brouwer, line 387: Ok, this change triggers a Common test. I am pretty sure this is ok, we should check for `<` not `<=`?

acolytec3 (approved):
LGTM.  I've previously verified that the specified hive tests do indeed pass.

packages/common/src/common.ts
jochem-brouwer: Ah and I forgot to note: this change was triggered by a hive test with thus this exact scenario of hitting the TTD exactly.

jochem-brouwer:
Hmm, I have reran the client tests twice now since I thought this would be a random timeout, but maybe this is not the case (have just retriggered, if it does not pass then I should investigate)

jochem-brouwer:
Ok, tests seems failing into a context which I thought was not relevant, will take a look!

packages/client/src/blockchain/chain.ts
```diff
@@ -294,18 +294,26 @@ export class Chain {
...
+    blocks.finalized = (await this.getCanonicalFinalizedBlock()) ?? null
+    blocks.safe = (await this.getCanonicalSafeBlock()) ?? null
+    blocks.vm = await this.getCanonicalVmHead()
+
+    headers.latest = blocks.latest.header
     // finalized and safe are always blocks since they have to have valid execution
     // before they can be saved in chain
-    headers.finalized = (await this.getCanonicalFinalizedBlock()).header
-    headers.safe = (await this.getCanonicalSafeBlock()).header
+    if (blocks.finalized !== null) {
```
g11tech, line 305:
```suggestion
   headers.finalized = blocks.finalized?.header ?? null
```
should be cleaner and simlarly for `safe`

packages/client/src/blockchain/chain.ts
```diff
@@ -297,13 +297,13 @@ export class Chain {
     headers.latest = await this.getCanonicalHeadHeader()
     // finalized and safe are always blocks since they have to have valid execution
     // before they can be saved in chain
-    headers.finalized = (await this.getCanonicalFinalizedBlock()).header
-    headers.safe = (await this.getCanonicalSafeBlock()).header
+    headers.finalized = blocks.finalized?.header ?? null
+    headers.safe = blocks.safe?.header ?? null
     headers.vm = (await this.getCanonicalVmHead()).header
 
     blocks.latest = await this.getCanonicalHeadBlock()
```
jochem-brouwer, line 304: Ok, so in this file if one first does gets all the blocks and then the headers, then a sync test fails. If one updates this to the original (see 586868ba51d9d3beddd5961fa44d667c4d547d8d ) then it passes. Test in question is `integration/lightclient.spec.ts`

holgerd77:
Reading from the thread it seems this can be merged and will therefore do. If I accidentally merged with some todo still left please eventually just do in a small follow-up PR, but seems convenient to have this in.

holgerd77 (approved):
LGTM

holgerd77:
(so: pretty cool work with so many Hive fixes! 🤩 🎉)

jochem-brouwer:
Great, thanks :smile: :pray:

packages/client/src/blockchain/chain.ts
```diff
@@ -513,17 +513,17 @@ export class Chain {
   /**
    * Gets the latest block in the canonical chain
    */
-  async getCanonicalSafeBlock(): Promise<Block> {
+  async getCanonicalSafeBlock(): Promise<Block | undefined> {
```
holgerd77, line 516: Just writing releases notes and stumbled upon this: so this is a (type) breaking change.

I first tended to just leave and hope not too many people use this and the method below yet but now when thinking about it twice I would prefer of going (a small) extra round here and introduce two new methods `getCanonicalSafeBlockSafe()` and `getCanonicalFinalizedBlockSafe()` and keep the behavior of the current methods as is, also considering that both methods are just used once in client and can be easily updated.

jochem-brouwer: Ah wait, I just stumbled upon this again, but this is into our client package (and not in our blockchain package). So I think we can leave this as-is? It is very unlikely someone uses our client as an actual package dependency?

holgerd77: Ah, you are right. I only saw `blockchain/*`. 🤦‍♂️

Sorry for the non-issue.

PR #3100, opened 2023-10-10
Client: More UI Improvements / Fix TxPool not being started along FCU

holgerd77:


packages/client/src/rpc/modules/engine.ts
```diff
@@ -1226,11 +1226,9 @@ export class Engine {
       }
       this.service.txPool.removeNewBlockTxs(blocks)
 
-      const isPrevSynced = this.chain.config.synchronized
-      this.config.updateSynchronizedState(headBlock.header)
```
jochem-brouwer, line 1230: Should this line be removed? :thinking:

jochem-brouwer, line 1230: Should this line be removed? :thinking:

jochem-brouwer, line 1230: Should this line be removed? :thinking:

jochem-brouwer, line 1230: Should this line be removed? :thinking:

jochem-brouwer, line 1230: Should this line be removed? :thinking:

holgerd77: Yes.

holgerd77: It's was not triggering anything anyhow and is now replaced by a direct setting of `synchronized` on FCU.

Testing eventual side effects right now by running a full Holesky sync (seems nothing so far).

Point of all this is that the tx pool never started.

jochem-brouwer: I don't think we should remove this line. The txpool thing is fine. Now each time FCU is called it will set state to synchronized. There is an interval in `sync.ts` which calls it also, which might override it back to syncing. I think this has undesired side effects..?

holgerd77:
Ah, no side effects (ran full Holesky and tx pool is properly starting at the end of execution and no other "stuff" happening or triggered).

grafik

Then I guess this is ready for review! 🙂

packages/client/src/rpc/modules/engine.ts
acolytec3: I guess i"m not 100% clear on what this `synchronized` state represents then.  Does it represent the current head that the CL knows of or is it the current canonical head that the `client` is aware of or yet again, is it the current executed head of the client?

cc @g11tech

g11tech: i will lookinto fhis and make a fix shortly, as discussed with @holgerd77 the new logic might open the txpool without us being in the "synchronized" state

g11tech: added the fix:
image

g11tech (approved):
lgtm

Issue #3101, opened 2023-10-12
Client: Improve --rpcDebug Flag


holgerd77:
The `--rpcDebug` flag in the client is hardly usable in its current form post-Merge, since there is so much output from engine API calls that it's close to not possible to spot single `eth_` namespace posts one is interested in.

We should relatively urgently improve here. My suggestion would be that it get's at least possible to provide and activate a namespace here (so by using `--rpcDebug=eth` e.g.), while the original boolean behavior should be preserved.

I am not 100% sure if this is cleanly achievable with Winston though. Yet another flag should be avoided though as well.

Another thing is that from my perspective the condition: flag only activates when used in combination with `--logLevel=debug` is really useless and this should be removed if possible (I have some vague rememberance that there was *some* reasoning behind that when introduced during Greece interop summit 2 years or so ago, but can't remember exactly what it was).

It would be great if someone can already tackle this during the day, so that we can use for final Holesky testing (also see my "tx sending on Holesky not working" post/thread on Discord).

holgerd77:
Closed by #3102

PR #3102, opened 2023-10-13
Change rpcDebug option to allow enabling/disabling specific rpc module logs

scorbajio:
This change modifies the cli `rpcDebug` option to allow for enabling only the `eth` or `engine` rpc module for log collection so that there is less noise during debugging. Also see #3101.

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -205,9 +205,8 @@ const args: ClientOpts = yargs(hideBin(process.argv))
   })
   .option('rpcDebug', {
     describe:
-      'Additionally log complete RPC calls on log level debug (i.e. --loglevel=debug). Can indicate a specific module or a boolean.',
-    choices: ['true', 'false', 'eth', 'engine'],
-    default: 'false',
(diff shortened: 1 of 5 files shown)


holgerd77 (changes requested):
Thanks a lot for quickly jumping in here! 🤩

I have to say: when seeing this now I do not find it clean enough to merge in as production code, even if it is "only" a debug flag. So this mixture of pseudo boolean and strings just doesn't cut it and is getting too ugly with behavior deviating from the real boolean (so e.g. `--rpcDebug` now doesn't work any more, could be fake-fixed though, but also not so pretty).

I would therefore suggest here that we go the full round and go "string-only", maybe with the addition that we also add `all` as an option.

And then on this go that we also simply fully generalize this, which would now be easy doable, and just allow for all sorts of strings (separated by ,) to filter for.

That would really be great, then we can simply do by-message as well e.g..

You know what? I will just directly do and push here. Seems like a no-brainer to me I realize along writing. 🙂

holgerd77 (approved):
LGTM

PR #3106, opened 2023-10-18
Change execution stats intervals

scorbajio:
This change switches `vmexecution.ts` stat reporting over to using time instead of block intervals.


scorbajio:
I tested this by syncing the client executing blocks starting from block `68100` on `mainnet`. The stats seems to be getting reported in 40 seconds intervals and the stats are updating and progressing as expected.

holgerd77:
Updated this via UI

packages/client/src/execution/vmexecution.ts
```diff
@@ -676,10 +687,11 @@ export class VMExecution extends Execution {
       })
 
       if (txHashes.length === 0) {
+        this._statsVm = vm
+
         // we are skipping header validation because the block has been picked from the
         // blockchain and header should have already been validated while putBlock
         const beforeTS = Date.now()
-        this.stats(vm)
```
scorbajio, line 682: One thing to note about this change is that now stat collection and logging isn't being included in the time measurement of `runBlock`, so there will potentially be a difference in the `time` value being logged for block execution.

holgerd77:
Updated this via UI

holgerd77 (approved):
Thanks a lot, this looks great! 🙂 

grafik

I think we can (and should) further reduce the frequency, the above is a run on mainnet e.g. and stats still occuring pretty often and still distract from the main progress output.

Have set down from 40 to 90 secs and pushed, will otherwise merge.

holgerd77:
Updated this via UI

PR #3109, opened 2023-10-21
trie: refactor lock class

gabrocheleau:
This PR refactors the trie package to use the Lock class from util instead of the duplicated one in the trie package.


acolytec3 (approved):
LGTM.

PR #3112, opened 2023-10-23
Util internal tests and fix

scorbajio:
This change adds tests for the `util` package `internal.ts` functions and fixes a bug in the `fromUtf8` function that double-adds the hex prefix to the returned value.


packages/util/src/internal.ts
```diff
@@ -135,10 +135,13 @@ export function toAscii(hex: string): string {
...
  * @returns hex representation of input string
  */
-export function fromUtf8(stringValue: string) {
-  const str = utf8ToBytes(stringValue)
+export function fromUtf8(utf8String: string): string {
+  if (typeof utf8String !== 'string') {
+    throw new Error(`[fromUtf8] input must be a string, received type ${typeof utf8String}`)
+  }
 
-  return `0x${padToEven(bytesToHex(str)).replace(/^0+|0+$/g, '')}`
```
jochem-brouwer, line 141: I'm not sure what the original regex did, remove preceding 0s and trailing 0s? Can't we use the original and instead use `bytesToUnprefixedHex` instead of `bytesToHex`?

packages/util/test/internal.spec.ts
```diff
@@ -67,4 +69,62 @@ describe('internal', () => {
...
+    })
+  })
+
+  describe('arrayContainsArray', () => {
+    it('should return true when the first array contains all elements of the second', () => {
+      assert.isTrue(arrayContainsArray([1, 2, 3, 4, 5], [3, 4]))
+    })
+
+    it('should return false when the first array does not contain all elements of the second', () => {
+      assert.isFalse(arrayContainsArray([1, 2, 3, 4, 5], [6, 7]))
```
jochem-brouwer, line 105: Maybe one extra test here to have it partially contain the array, for instance:

```suggestion
      assert.isFalse(arrayContainsArray([1, 2, 3, 4, 5], [5, 6]))
```

packages/util/test/internal.spec.ts
```diff
@@ -67,4 +69,62 @@ describe('internal', () => {
...
+    })
+  })
+
+  describe('fromUtf8', () => {
+    it('should convert a UTF-8 string to a hex string', () => {
+      assert.equal(fromUtf8('Hello, World!'), '0x48656c6c6f2c20576f726c6421')
```
jochem-brouwer, line 111: This string only contains ASCII characters. Can we also test an UTF8 string here? Ref: https://www.fileformat.info/info/charset/UTF-8/list.htm 

(So this visually adds one character, but actually adds two bytes in the string ref)

jochem-brouwer (commented):
Some comments, tests look great in general! :smile:

holgerd77:
Great new tests and great review! 🤩

packages/util/src/internal.ts
scorbajio: Yes, I've changed over to this.

packages/util/test/internal.spec.ts
scorbajio: Added.

packages/util/test/internal.spec.ts
scorbajio: Created a new test that includes UTF8 2-byte characters.

holgerd77:
Updated this via UI

holgerd77 (approved):
Nice, LGTM

jochem-brouwer (approved):
LGTM

PR #3118, opened 2023-10-25
vm: better error message to avoid confusion with EIP1559 base fee

jochem-brouwer:
If the gas limit of a tx is lower than the minimum gas limit (this consists of: tx upfront gas of 21000, the tx data cost, and optionally more gas needed for contract creation), previously we had this error message: `base fee exceeds gas limit`. This is confusing with EIP1559s base fee and is completely unrelated. This PR updates the error message.


holgerd77:
Updated this via UI

holgerd77 (approved):
LGTM, thanks for adressing so quickly

PR #3119, opened 2023-10-25
Client: pendingBlock fix if FCU is called with withdrawals=null pre-cancun

jochem-brouwer:
Stumbled upon this one in a hive test.

I have explicitly checked client for more of these cases where we test for `!== undefined` but do not have the other check that it should also not be `null`. This was the only occurrence I could find.

### DIFF
--- a/packages/client/src/miner/pendingBlock.ts
+++ b/packages/client/src/miner/pendingBlock.ts
@@ -133,7 +133,7 @@ export class PendingBlock {
 
     let withdrawalsBuf = zeros(0)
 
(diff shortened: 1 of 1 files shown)


holgerd77:
Updated this via UI

jochem-brouwer:
(we can also keep this PR open so I can commit any more of these micro-changes which I might find, otherwise we might end up with a lot of these mini PRs? I stumbled upon this when doing the reorg hive tests research today)

holgerd77 (approved):
LGTM

PR #3120, opened 2023-10-26
Client Discovery Improvements

holgerd77:
This PR improves DiscV4 discovery in various ways.

First it adds an explicit logging message for discovery on startup. This is helpful since it remains often somewhat unclear if and what form of discovery is running.

Second it expands the `--bootnodes` option for the client to also accept a `bootnode.txt` file (e.g. from [EthPandaOps devnet-10 config](https://github.com/ethpandaops/dencun-testnet/tree/master/network-configs/devnet-10).

Third it introduces a new confirmed-peers mechanism for devp2p DPT. This allows to set a higher quality bar for sendNeighbour requests by only send to peers which have been manually confirmed. Otherwise discovery is hammering out requests like crazy to - mainly - mainnet peers not finding any needle in this haystack which rendered the feature more or less useless for all networks except mainnet (in fact we had deactivated in client for everything except mainnet).

Now only peers who are on the correct network and where a network connection was/is possible get the `findNeighbour` requests. This makes discV4 discovery requests acceptable again.

I was able to re-activate peer discovery by default for small networks - here devnet-10 - and have a slowly growing peer base (here the yellow numbers)! 🎉

grafik

Still need to do some testing - also write some tests - but I am confident that this approach works adequately and we can likely merge in.


holgerd77:
Ok, I did some additional improvements and tested under different network conditions (small devnet, testnet, mainnet), following results:

- testnets (like Holesky): previously discV4 deactivated (too much noise), now with `onlyConfirmed`: -> helps finding peers (tested with DNS deactivated)
- small testnets (like devnet-10): previously no discovery, now with `onlyConfirmed`: most useful case, helps finding new peers *at all*
- mainnet: reduces peer discovery speed (since most peers are mainnet anyhow): -> have deactivated

I have also added a descent new test suite with mockups, see `dpt.spec.ts`.

Open for review. 🙂

packages/devp2p/src/dpt/dpt.ts
```diff
@@ -148,6 +157,20 @@ export class DPT {
...
+   * Add peer to a confirmed list of peers (peers meeting some
+   * level of quality, e.g. being on the same network) to allow
+   * for a more selective findNeighbours request and sending
+   * (with activated `onlyConfirmed` setting)
+   *
+   * @param id Unprefixed hex id
+   */
+  confirmPeer(id: string) {
+    if (this._confirmedPeers.size < 5000) {
+      this._confirmedPeers.add(id)
```
jochem-brouwer, line 170: Peers are never removed from `_confirmedPeers`, so at some point we cannot confirm new peers since the set is full? Should the peer id be removed on `removePeer`?

jochem-brouwer (commented):
One comment, and am still tracking down where the filter for the network is :thinking: (see discord, have to track down a long event queue and likely take a wrong path somewhere there)

jochem-brouwer:
Also, the CI fails on the new tests :thinking:

packages/devp2p/src/dpt/dpt.ts
holgerd77: Yeah, that was the lazy version assuming a very slow growth. But removing on `removePeer` definitely make sense, will add. 👍

acolytec3 (approved):

acolytec3 (commented):
LGTM

jochem-brouwer (approved):

PR #3122, opened 2023-10-27
EVM: Prevent Address Creation for EIP-2929 Address Tracking

holgerd77:


jochem-brouwer (approved):
LGTM

Side note: we can maybe further optimize by actually also dropping most `addressToBytes`. We can internally also just handle the BigInt instead of the Bytes. Then we can likely also change the `journal` to use BigInt instead of string, and then we can drop `bytesToUnprefixedHex` in `journal.ts` too :smile:

holgerd77:
> Side note: we can maybe further optimize by actually also dropping most addressToBytes. We can internally also just handle the BigInt instead of the Bytes. Then we can likely also change the journal to use BigInt instead of string, and then we can drop bytesToUnprefixedHex in journal.ts too 😄

I wonder though if we should still do any changes into the BigInt direction, if we think about going full-bytes at some point. 🤔

jochem-brouwer:
Oh right, good point, we should decide on that first (likely shift to Bytes directly)

PR #3123, opened 2023-10-27
EVM: Precalc Reoccuring Word Length Sub Mem Usage Cost

holgerd77:
On the blocks I tested this is *always* iterating between 0 and 273.


packages/evm/src/opcodes/util.ts
```diff
@@ -173,12 +173,17 @@ export function subMemUsage(runState: RunState, offset: bigint, length: bigint,
...
   const newMemoryWordCount = divCeil(offset + length, BIGINT_32)
   if (newMemoryWordCount <= runState.memoryWordCount) return BIGINT_0
-
   const words = newMemoryWordCount
-  const fee = common.param('gasPrices', 'memory')
-  const quadCoeff = common.param('gasPrices', 'quadCoeffDiv')
-  // words * 3 + words ^2 / 512
-  let cost = words * fee + (words * words) / quadCoeff
+
+  let cost = BIGINT_0
```
jochem-brouwer, line 178: I don't think this needs to be instantiated with `BIGINT_0`, this should work 
```suggestion
  let cost: BigInt
```

TypeScript can infer after this if/else statement that `cost` is set (and thus not `undefined`) (AFAIK, but I think I had this same pattern today in another PR)

packages/evm/src/opcodes/util.ts
```diff
@@ -274,3 +279,280 @@ export function exponentiation(bas: bigint, exp: bigint) {
   }
   return t
 }
+
+const SUB_MEM_USAGE_COST_PRECALC_CONSTS = [
+  BigInt(0),
```
jochem-brouwer, line 284: I am not really a fan of this. If one of these constants are wrong, then we have to be "lucky" that the test suite catches this. Can these constants be initialized by calling into `subMemUsage`? (Probably need to just extract the formula `cost = words * fee + (words * words) / quadCoeff` and call into this with the right constants - note: they have not been changed by any hardfork so far and these are `3` for `fee` and 512 for `quadCoeff`).

jochem-brouwer (commented):
I am not entirely sure about this one. This will break any chain which uses different gas cost for memory :thinking: Also, two comments.

packages/evm/src/opcodes/util.ts
holgerd77: I have created the list programmatically by running the formula and programmatically creating a string of the array, so risk should be alread pretty low, I would not know what should have gone wrong there.

What makes this a non issue though is that - on real world blocks - this is simply iterating from 0 to 273 and then starting all over again endlessly. 😋 (totally didn't know about this behavior)

So this is fully tested by a small mainnet run with contract execution.

packages/evm/src/opcodes/util.ts
holgerd77: Yes, works! Will update! 🙂

holgerd77:
Ok. I've pushed my fix for completeness (a reverse-constant-check let's call it), which I think would be worth it for substantial performance gains.

I measured this though now on blocks 699200-700200 on mainnet, so 1000 blocks, and both are in the range of taking 0.15/0.16 secs in total (so: all computation through this function during the block period).

Always hard to say how/where to put this in perspective (from my performance experiences I would now - very roughly say - that a gain by 0.1 sec / 1000 blocks and would be worth it). But so this seems *not* worth to be implemented.

So will close. 🙂 Thanks for reviewing.

PR #3125, opened 2023-10-30
Client: fix rpc debug

jochem-brouwer:
The rpc debug is inconsistent and this PR fixes it. Currently it seems to always truncate RPC logs.

It adds a new flag `rpcDebugVerbose` which allows for a comma-separated list of methods to log in full. Same for `rpcDebug`, this logs the commands but will truncate those. If both `rpcDebugVerbose` and `rpcDebug` filters match the method, the log is not truncated. The necessity to `--loglevel debug` is removed to log the methods in full (the logs are written to INFO).

TODO: explicitly test this (but this should work)

Note: it is also possible to skip adding this new flag, and instead adding the word "verbose" to `rpcDebug` as suggested by Holger. However, this would imply that then all methods which match the filter are logged in verbose and I think this extra flag option gives us more freedom to explicitly verbose-log these methods.


holgerd77:
RPC log truncating has a reason being mainly that from some RPC calls - particularly from the engine API - are so large that without truncating it is not possible any more to follow on the logging.

Can't completely oversee the changes from this PR respectively what it does and what not, but we need to find a good balance here to for one allow that it *becomes* possible - if one is interested in the outcome of one particular call - to see the full output. And on the other hand that RPC logging also allows for this "overview log", so that one has a constant ouptut of calls without getting overwhelmed.

jochem-brouwer:
Just some clarification, without this PR it seems that `rpcDebug` does not work, we launch hive with these flags:

`--gethGenesis ./genesis.json --rpc --rpcEngine --saveReceipts --rpcAddr 0.0.0.0 --rpcEngineAddr 0.0.0.0 --rpcEnginePort 8551 --ws false --logLevel debug --rpcDebug all --isSingleNode`

Note: `--rpcDebug all`. This still truncates logs (without this PR). This PR fixes this.

holgerd77:
Yes, not sure if my comment was clear: rpcDebug logs *needs* to be truncated at least as an option, otherwise this becomes completely unreadable for certain use cases.

I wonder if we eventually want to bite the sour bullet or whatever and *DO* another explicit option here?? `--truncateRPCLogs` or whatever? Since this is definitely otherwise totally intransparent and also gave me headaches as well in the past. But one definitely needs both.

Side note: beyond there are some RPC log inconsistencies in the client, e.g. parts of the (core) engine API (I think FCU and newPayload() ?) is not logged through the option but on normal debug mode, which is confusing and we should fix as well at some point.

packages/client/bin/cli.ts
```diff
@@ -207,7 +207,8 @@ const args: ClientOpts = yargs(hideBin(process.argv))
   .option('rpcDebug', {
     describe:
       'Additionally log complete RPC calls filtered by name (prefix), e.g.: "eth,engine_getPayload" (use "all" for all methods).',
-    default: '',
+    coerce: (arg: string) => (arg === '' ? 'all' : arg),
```
scorbajio, line 210: When I initially changed `rpcDebug` to allow for indicating a specific package for full request/response log output, I included a description with details about the need for `logLevel` being set to `debug` and also included choices which allowed for boolean options or a specific module to be enabled by name, as well as a default of `false`: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3102/commits/b196a131e92c80a30814031d4885b9233e7dfdef. But now that we can pass in a comma separated string that could also include specific function names (e.g. `eth,engine_getPayload`) I think that we can look into using a yargs `check` function for `rpcDebug` to first validate the input and throw helpful error messages so users aren't confused as well as extend our description to include more usage details and maybe even some examples.

packages/client/bin/cli.ts
```diff
@@ -207,7 +207,8 @@ const args: ClientOpts = yargs(hideBin(process.argv))
   .option('rpcDebug', {
     describe:
       'Additionally log complete RPC calls filtered by name (prefix), e.g.: "eth,engine_getPayload" (use "all" for all methods).',
-    default: '',
```
scorbajio, line 210: Using a default value can be more intuitive than removing it and instead having to hand `undefined` as a value.

holgerd77:
> Ah ok good point, I just re-checked, we have this (also in this PR - it is unchanged), see `client/src/util/rpc.ts`.
> 
> ```ts
>       if (show) {
>         if (logger?.level === 'debug') {
>           msg += `${request.method} called with params:\n${inspectParams(request.params)}`
>         } else {
>           msg += `${request.method} called with params: ${inspectParams(request.params, 125)}`
>         }
>         logger?.info(msg)
>       }
> ```
> 
> Therefore, if `--rpcDebug` is enabled but `--loglevel` is not set to `debug`, the logs are truncated (by 125 chars) by default. So to get full logs, one has to explicitly set `--loglevel` to `debug`. (However this might raise the question: do we want an extra flag to actually `--loglevel debug` but not dump the rpcDebug completely (so truncate it)? (Especially for blob txs these might indeed get very large!))

Yes, I was aware of this: debug -> full RPC log, otherwise -> truncated behavior.

And, yes, I think it would be worth to be able to tackle this separately (I can vaguely remember that I might have already tried on the sideline at some point?).

So an alternative to yet-another-flag might be that we add a keyword `verbose` to the existing option (important to add a note to CLI help as well then).

If this is added we do full logging (`--rpcDebug=all,verbose`, `--rpcDebug=eth,verbose`,...) otherwise we truncate.

jochem-brouwer:
Have tested using some hive tests and can report that it works like the top post suggests. CI should also pass.

packages/client/src/util/rpc.ts
```diff
@@ -63,51 +86,23 @@ export function createRPCServer(
...
-      }
-      if (show) {
-        if (logger?.level === 'debug') {
-          msg += `${request.method} called with params:\n${inspectParams(request.params)}`
-        } else {
-          msg += `${request.method} called with params: ${inspectParams(request.params, 125)}`
-        }
-        logger?.info(msg)
-      }
+    if (checkFilter(request.method, rpcDebugVerbose)) {
```
scorbajio, line 91: To note, the options seem to be:
* both `rpcDebugVerbose` and `rpcDebug` are true, in which case the full version will be logged
* both `rpcDebugVerbose` and `rpcDebug` are false, in which case nothing will be logged on requests and responses
* `rpcDebugVerbose` is true and `rpcDebug` is false, in which case the full version will be logged
* `rpcDebugVerbose` is false and `rpcDebug` is true, in which case the truncated version will be logged

scorbajio (approved):
This looks good to me!

packages/client/src/util/rpc.ts
jochem-brouwer: Yes, so for extra for clarification: the `true` and `false` are the returned values from this method, and are thus applied on the filter. If there is no filter (the filter is the empty string by default, so "no filter") then it always returns `false`.

jochem-brouwer:
Have opened a PR for hive to update as well: https://github.com/ethereum/hive/pull/939

holgerd77:
Maybe I missed this: what was the reasoning here again to go for the yet-another-flag solution in favor to an existing-flag using version like `--rpcDebug=verbose,eth` or similar?

jochem-brouwer:
The reasoning is that you can now for instance log verbose logs for `engine_newPayload`, but choose to only log truncated `eth_sendRawTransaction` (especially blob txs are huge). We could also settle for a very specific format inside the `rpcDebug` flag if that is desired and drop the new flag.

holgerd77:
Ah ok, no, makes sense, cool, thanks!

PR #3127, opened 2023-10-31
Add rpcDebug stack collection to all rpc modules

scorbajio:
Currently, the ability to collect stack traces for rpc modules is not applied to all modules. This change adds it to all rpc modules. Also see #3125 which will make the `rpcDebug` functionality and cli option more consistent in behavior.


acolytec3 (approved):

acolytec3 (approved):
LGTM

PR #3131, opened 2023-10-31
handle an edge case in newpayload block execution

g11tech:
handle an edgecase noticed during holesky sync

### DIFF
--- a/packages/client/src/rpc/modules/engine.ts
+++ b/packages/client/src/rpc/modules/engine.ts
@@ -910,20 +910,42 @@ export class Engine {
         }
       }
     } catch (error) {
-      const errorMsg = `${error}`.toLowerCase()
-      if (errorMsg.includes('block') && errorMsg.includes('not found')) {
-        throw {
-          code: INTERNAL_ERROR,
-          message: errorMsg,
-        }
-      }
-      const validationError = `Error verifying block while running: ${errorMsg}`
-      this.config.logger.error(validationError)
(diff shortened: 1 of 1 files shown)


packages/client/src/rpc/modules/engine.ts
```diff
@@ -910,20 +910,42 @@ export class Engine {
...
       const errorMsg = `${error}`.toLowerCase()
       if (errorMsg.includes('block') && errorMsg.includes('not found')) {
-        throw {
-          code: INTERNAL_ERROR,
-          message: errorMsg,
+        if (blocks.length > 1) {
+          // this error can come if the block tries to load a previous block yet not in the chain
+          // error coding of the evm errors should be a better way to handle this
+          //
+          // at Blockchain.getBlock (/home/gajinder/ethereumjs-monorepo/packages/blockchain/src/blockchain.ts:763:15)
```
acolytec3, line 925: Do we need this entire stack trace in the logs?

acolytec3 (changes requested):

acolytec3 (commented):
Generally looks good.  Do we need a specific test for this edge case or is this covered by existing tests?

holgerd77 (approved):
I would really really like to get things prepared for a release round tomorrow, at least if things work on a somewhat descent level (just re-running Holesky again right now).

I would therefore merge here, maybe we can add an eventual test later on.

Thanks! 🙏

PR #3132, opened 2023-10-31
blockchain: ignore older putblocks/header for determining the chain canonicality

g11tech:
currently if an older number block is put into chain for pos, this leads to change in canonicalization (infact resets it, even if the older block is just the repeat of a  previus canonical block). This resetting behavior is causing some `pyspec` to fail (`pyspec/cancun/eip4844_blobs/point_evaluation_precompile/point_evaluation_precompile_before_fork/000-fork=ShanghaiToCancunAtTime15k-correct_proof-point_evaluation_precompile_before_fork`) since the chain is attempted to fill by more than 1 async processes

however we should allow older blocks to be put unless they extends the chain. 

incase someone wants to reset the chain to some older block, they should first reset  it to height < older block and then add the older block to make chain canonical along older block. thats how client skeleton does the chain reset 

TODO
 - [ ] right now older block/header is just ignored but if we keep building the chain past the current head header, there should be canonical number => hash change for all the blocks <= current head header, not sure how to best handle it
 - [ ] test cases


g11tech:
@acolytec3 @holgerd77 @jochem-brouwer 

do you this this PR is along the current resolution? an older putBlockHeader (along canonical chain or not) shouldn't result into resetting and change in canonicalization of the chain

i think we has this rule of number > current header number to make/shift the canonicality along new head, not sure why we changed it, but if we want to keep it this way, need to modify the assumptions/conditions we have in client for putBlock

holgerd77:
Can we for now solve this in a more backwards-compatible way, e.g. by adding a flag `safe` or similar to `putBlockOrHeader()` or similar/related methods? 🤔

I think that should work as well and then we can just switch over in the client to use the flag?

holgerd77:
Will re-prepare releases for tomorrow, let's please not take  this in any more regardless if we find a solution or not.

(so: please do not merge before releases!)

jochem-brouwer:
This might resolve some hive tests also, because if we FCU back to a canonical block, the spec states that EL "may" ignore this (and I think we do this here now, right? So we will not "rollback" to a canonical block if we put the block?

jochem-brouwer:
Have pushed a lint fix and merged master.

holgerd77:
What is the status of this PR? Have lost a bit track, even after skimming over the discussion! Is this integrated in some alternative form somewhere else? Still in discussion? Or decided not to integrate?

Would be good if we can settle here.

g11tech:
i think we solved it another way via fixes in client, closing for now

PR #3135, opened 2023-11-02
vm: add missing gasPrice param to readme example tx

gabrocheleau:
This PR:

- Fixes the Readme example tx that was missing a gasPrice param
- Improves an error message that made it seem like pre-eip1559 transactions were providing an invalid maxFeePerGas param.


acolytec3 (approved):
Confirmed that this now works.  LGTM

PR #3137, opened 2023-11-02
client: Patch fcu skeleton blockfill process to avoid chain reset

g11tech:
to avoid the chain being reset by getting interfered with sksleton's fill Canonicaland fcU's sethead's putblocks of executed blocks

now skeleton attempts blocking fill for the max number of  executed blocks which fcu wants to put via execution's setHead

This PR should fix the hive pytests failing because of this issue

ps:
this is a patch but a proper solution needs to happen at blockchain layer where blocks shouldn't reset the chain unless explicitly asked to or via a proper rule that they should extend the max height of the chain.
Also skeleton 's canonical block handling/tracking needs to be up streamed to the chain , but that could be a little more involved work to address separately

### DIFF
--- a/packages/client/src/rpc/modules/engine.ts
+++ b/packages/client/src/rpc/modules/engine.ts
@@ -1231,12 +1231,6 @@ export class Engine {
         }
       }
       this.service.txPool.removeNewBlockTxs(blocks)
-
-      const isPrevSynced = this.chain.config.synchronized
-      this.config.updateSynchronizedState(headBlock.header)
-      if (!isPrevSynced && this.chain.config.synchronized) {
-        this.service.txPool.checkRunState()
-      }
     } else if (!headBlock.isGenesis()) {
       // even if the vmHead is same still validations need to be done regarding the correctness
       // of the sequence and canonical-ity
--- a/packages/client/src/service/skeleton.ts
+++ b/packages/client/src/service/skeleton.ts
@@ -1309,7 +1309,7 @@ export class Skeleton extends MetaDBManager {
     }
     this.filling = false
     this.config.logger.debug(
-      `Successfully put=${fillLogIndex} skipped=${skippedLogIndex} blocks start=${start} end=${canonicalHead} skeletonHead=${subchain.head} from skeleton chain to canonical syncTargetHeight=${this.config.syncTargetHeight}`
+      `Successfully put=${fillLogIndex} skipped (because already inserted)=${skippedLogIndex} blocks start=${start} end=${canonicalHead} skeletonHead=${subchain.head} from skeleton chain to canonical syncTargetHeight=${this.config.syncTargetHeight}`
     )
   }
 
--- a/packages/client/src/service/skeleton.ts
+++ b/packages/client/src/service/skeleton.ts
@@ -110,8 +110,6 @@ export class Skeleton extends MetaDBManager {
   private lastFcuTime = 0
   private lastsyncedAt = 0
 
-  private fillLogIndex = 0
-
   private STATUS_LOG_INTERVAL = 8000 /** How often to log sync status (in ms) */
 
   /**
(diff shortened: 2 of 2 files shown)


packages/client/src/service/skeleton.ts
```diff
@@ -1201,72 +1208,82 @@ export class Skeleton extends MetaDBManager {
...
-          )} parentHash=${short(block.header.parentHash)}from skeleton chain to canonical`
-        )
-        // Lets log some parent by number and parent by hash, that may help to understand whats going on
-        let parent = null
+      let numBlocksSkipped = 0
+      // chain height has to be <= block number as we will skip putting this block as it might currently
+      // cause chain reset. This can happen if any other async process added a batch of blocks like
+      // execution's setHead. If that caused this chain to be not canonical anymore than the next
+      // putblocks should fail causing the fill to exit with skeleton stepback
+      if (this.chain.blocks.height <= block.header.number) {
```
g11tech, line 1216: this is the only real change here, i.e. skip putting if chain has been advanced by some other putBlock process. But in the hive tests i ran locally, there would still be  a random race where this check would pass but still when actual writing the block, a concurrent putblock wins the race causing an issue 

(so we eventually needs to fix it since the only actual locking happens in the blockchain class)

packages/client/src/rpc/modules/engine.ts
```diff
@@ -1231,12 +1231,6 @@ export class Engine {
         }
       }
       this.service.txPool.removeNewBlockTxs(blocks)
-
-      const isPrevSynced = this.chain.config.synchronized
-      this.config.updateSynchronizedState(headBlock.header)
-      if (!isPrevSynced && this.chain.config.synchronized) {
-        this.service.txPool.checkRunState()
-      }
     } else if (!headBlock.isGenesis()) {
```
acolytec3, line 1234: What does removing this section do?

packages/client/src/service/skeleton.ts
acolytec3: Maybe this is a silly question but do we need to check `block.header.number` and ensure its only one block ahead of `height`?

packages/client/src/service/skeleton.ts
```diff
@@ -1283,16 +1300,16 @@ export class Skeleton extends MetaDBManager {
...
           `Skeleton canonical chain fill status: canonicalHead=${canonicalHead} chainHead=${this.chain.blocks.height} subchainHead=${subchain.head}`
         )
-        this.fillLogIndex = 0
+        fillLogIndex = 0
       }
     }
     this.filling = false
     this.config.logger.debug(
-      `Successfully put blocks start=${start} end=${canonicalHead} skeletonHead=${subchain.head} from skeleton chain to canonical syncTargetHeight=${this.config.syncTargetHeight}`
+      `Successfully put=${fillLogIndex} skipped=${skippedLogIndex} blocks start=${start} end=${canonicalHead} skeletonHead=${subchain.head} from skeleton chain to canonical syncTargetHeight=${this.config.syncTargetHeight}`
```
acolytec3, line 1312:
```suggestion
      `Successfully put=${fillLogIndex} skipped (because already inserted)=${skippedLogIndex} blocks start=${start} end=${canonicalHead} skeletonHead=${subchain.head} from skeleton chain to canonical syncTargetHeight=${this.config.syncTargetHeight}`
```

Is it true that we should never be able to put a block and then skip one higher than it, right?  So, if we're doing canonical fill, we might skip the first `x` blocks because some other process already put them in the chain, but we would then fill any remaining in this current iteration, right?

packages/client/src/rpc/modules/engine.ts
g11tech: it seems to be a redundant txpool start snippet, just below this if/else there is a similar snippet

packages/client/src/service/skeleton.ts
g11tech: yes, except that because of any race condition, it might not be just `first` x and can happen anywhere mid segment because of the async / concurrent nature of fcU calls

packages/client/src/service/skeleton.ts
g11tech: if its not, the putBlock will fail anyway

acolytec3 (commented):
LGTM

PR #2643, opened 2023-04-15
StateManager: stateless verkle state manager

gabrocheleau:
Supersedes #2328 by rebasing the existing verkle work on v7

### Resources 

Latest test data: https://gist.github.com/gballet/6b67e9e52fe833e1801681488ba70bc3

PR #3139, opened 2023-11-03
StateManager: stateless verkle state manager

gabrocheleau:
Supersedes #2643 by rebasing the work on the current master branch w/ the verkle package. 

run kaustinen2 sample network tests
`DEBUG=ethjs,vm:*,evm:*,sm:* npx vitest run test/rpc/engine/kaustinen2.spec.ts `

OR run actual kaustinen2 network:
1. `DEBUG=ethjs,vm:*,evm:* npm run client:start:ts -- --dataDir /data/teme --network kaustinen --rpc --rpcEngine --rpcEngineAuth false`
2. clone https://github.com/ChainSafe/lodestar-quickstart , do vi kaustinen.vars and remove the checkpointSyncUrl  from lodestar args input and run `./setup.sh --dataDir k2data --network kaustinen --justCL` to run lodestar to feed ethereumjs blocks


holgerd77:
Re-posting instructions here how to run the Kaustinen testnet with a manual Lodestar setup (Lodestar quickstart not always working in all conditions (Mac e.g.) and sometimes not up-to-date, also easier to test and adopt thing with a manual setup).

1. Clone Lodestar with `git clone https://github.com/ChainSafe/lodestar.git`, fetch and switch to `g11tech/verge` branch and install and build with `yarn install` and `yarn build`
2. Clone EthPandaOps [verkle-testnets](https://github.com/ethpandaops/verkle-testnets) repo with: `git clone https://github.com/ethpandaops/verkle-testnets.git`
3. Copy the kaustinen config folder to the Lodestar repo with `cd lodestart && cp -Rf ../verkle-testnets/network-configs/kaustinen-testnet .`

Now the EthereumJS client (from this branch) can be run with:

```shell
cd ../vm && npm run build && cd ../client && rm -Rf kaustinen && DEBUG=ethjs,vm:*,evm:*,statemanager:*,sm:* npm run client:start:ts -- --dataDir=kaustinen --network kaustinen --rpc --rpcEngine --rpcEngineAuth false
```

(that's my full command with a) rebuilding a specific package, b) resetting the kaustinen datadir and c) do debug calls, adopt accordingly)

And run Lodestar with:

```shell
rm -Rf kaustinen && ./lodestar beacon --paramsFile kaustinen-testnet/config.yaml --genesisStateFile kaustinen-testnet/genesis.ssz --eth1.depositContractDeployBlock 0 --bootnodesFile kaustinen-testnet/boot_enr.yaml --dataDir=kaustinen
```

This should make it possible for everyone in the team (or outside) to join the testnet! 🙂

packages/statemanager/src/statelessVerkleStateManager.ts
```diff
@@ -0,0 +1,611 @@
...
+  EVMStateManagerInterface,
+  Proof,
+  StorageDump,
+  StorageRange,
+} from '@ethereumjs/common'
+import type { Address, PrefixedHexString } from '@ethereumjs/util'
+
+const { debug: createDebugLogger } = debugDefault
+
+const debug = createDebugLogger('sm:verkle')
```
holgerd77, line 37: Short nit: this should use `statemanager:` and not `sm:` so that we do not use another debug namespace than we do in the other state managers.

packages/block/src/block.ts
```diff
@@ -208,7 +221,18 @@ export class Block {
...
-    return new Block(header, transactions, uncleHeaders, withdrawals, opts)
+    // executionWitness are not part of the EL fetched blocks via eth_ bodies method
+    // they are currently only available via the engine api constructed blocks
+    let executionWitness
+    if (executionWitnessBytes !== undefined) {
+      executionWitness = JSON.parse(bytesToUtf8(RLP.decode(executionWitnessBytes) as Uint8Array))
+    } else {
+      // don't assign default witness if eip 6800 is implemented as it leads to incorrect
+      // assumptions while executing the block. if not present in input implies its unavailable
+      executionWitness = null
```
holgerd77, line 232: I am not completely getting the `null` vs `undefined` case here. Is `null` only for the 6800 activated case? If there is no distinction: can we use only `undefined` and skip the `null` case (as we do for withdrawals)?

If there is a distinction: can we for the non-6800 case only use `undefined` (which is not the case here)?

packages/block/src/block.ts
```diff
@@ -366,13 +400,34 @@ export class Block {
     transactions: TypedTransaction[] = [],
     uncleHeaders: BlockHeader[] = [],
     withdrawals?: Withdrawal[],
+    executionWitness?: VerkleExecutionWitness | null,
```
holgerd77, line 403: Ah, this needs to change for backwards compatibility reasons.

So for now we need to put `executionWitness?` after `opts`, as we did before on similar additions, even if not so beautiful.

Even if discouraged, there are always people (actually I guess: a lot) using the constructor directly and this would break things for them.

packages/block/src/types.ts
```diff
@@ -71,6 +71,32 @@ export interface BlockOptions {
...
+    currentValue: PrefixedHexString | null
+    newValue: PrefixedHexString | null
+    suffix: number
+  }[]
+}
+
+export interface VerkleExecutionWitness {
+  stateDiff: VerkleStateDiff[]
+  verkleProof: VerkleProof
+}
```
holgerd77, line 98: Can you please add some general comments here on the datastructure types and also notes that these are experimental?

packages/blockchain/src/blockchain.ts
```diff
@@ -45,6 +45,37 @@ import type { BlockData } from '@ethereumjs/block'
...
+  genesisState: GenesisState,
+  common: Common
+): Promise<Uint8Array> {
+  const genCommon = common.copy()
+  genCommon.setHardforkBy({
+    blockNumber: 0,
+    td: BigInt(genCommon.genesis().difficulty),
+    timestamp: genCommon.genesis().timestamp,
+  })
+  if (genCommon.gteHardfork(Hardfork.Prague)) {
```
holgerd77, line 64: If I am not mistaken I think it should work also here to check for 6800 EIP activation with `common.isActivatedEIP()` (maybe double check). This is likely the cleaner and a bit more future proof solution to avoid direct Prague checks (maybe also check for other occurrences) but always do 6800 checks. Then we remain HF independent and things keep working if 6800 moves from Prague to whatever HF.

packages/client/bin/cli.ts
```diff
@@ -382,6 +387,21 @@ const args: ClientOpts = yargs(hideBin(process.argv))
     boolean: true,
     default: true,
   })
+  .option('statelessVerkle', {
+    describe: 'Run verkle+ hardforks using stateless verkle stateManager',
+    boolean: true,
+    default: true,
+  })
```
holgerd77, line 394: Please add some experimental notes at least for the two verkle flags.

(are the two others meant to make it into the main code base?)

packages/client/src/execution/vmexecution.ts
```diff
@@ -39,6 +43,8 @@ export class VMExecution extends Execution {
   private _lock = new Lock()
 
   public vm: VM
+  public merkleVm: VM | null = null
```
holgerd77, line 46: Small nit and also only if you clearly agree: for me this `Vm` reads strange and I would relatively clearly prefer `VM`, so e.g. `merkleVM` (I know there is `statsVm` below e.g. as well, maybe you can update along if you agree).

So if you feel the same maybe update. 🙂 No biggie though.

packages/client/src/execution/vmexecution.ts
```diff
@@ -39,6 +43,8 @@ export class VMExecution extends Execution {
...
+  public verkleVm: VM | null = null
```
holgerd77, line 47: Do we need these `null` assignments here both for the verkle and the merkle VM objects and have this ambiguity with `undefined` and `null` here again or can we just stick with `undefined`?

packages/client/src/execution/vmexecution.ts
```diff
@@ -133,21 +124,93 @@ export class VMExecution extends Execution {
...
+    }
+
+    this.config.logger.info(`Setting up verkleVm`)
+    const stateManager = new StatelessVerkleStateManager()
+    this.verkleVm = new (VM as any)({
+      common: this.config.execCommon,
+      blockchain: this.chain.blockchain,
+      stateManager,
+      profilerOpts: this.config.vmProfilerOpts,
+    })
```
holgerd77, line 183: Hmm, now that I am looking at this merkle vs verkle VM thing here again: both VMs are initialized in exactly the same way. I think we can remove this distinction completely and just go (back) to `this.vm` and only setup the respective different state manager.

Can you refactor a bit here?

holgerd77: Update: ok ok, I see this is for this transition logic. So I see where this is coming from, still can't oversee if this is the optimal/cleanest solution though. Maybe you can also have another look at this code part with the comments here in mind and see if you still think this is logically best structured and either keep or update.

packages/client/src/execution/vmexecution.ts
```diff
@@ -133,21 +124,93 @@ export class VMExecution extends Execution {
...
+  }
+
+  async transitionToVerkle(merkleStateRoot: Uint8Array, assignToVm: boolean = true): Promise<void> {
+    if (this.vm instanceof StatelessVerkleStateManager) {
+      return
     }
```
holgerd77, line 189: This check doesn't seem right. 🤔

packages/client/src/execution/vmexecution.ts
```diff
@@ -179,21 +242,80 @@ export class VMExecution extends Execution {
...
       this.config.execCommon.setHardforkBy({ blockNumber: number, td, timestamp })
       this.hardfork = this.config.execCommon.hardfork()
-      this.config.logger.info(`Initializing VM execution hardfork=${this.hardfork}`)
-      if (number === BIGINT_0) {
-        const genesisState =
-          this.chain['_customGenesisState'] ?? getGenesis(Number(this.vm.common.chainId()))
-        if (!genesisState) {
-          throw new Error('genesisState not available')
+
+      if (this.config.execCommon.gteHardfork(Hardfork.Prague)) {
```
holgerd77, line 246: For the client I am actually not sure, maybe also double-think about it: do we here also want to switch to 6800 EIP checks or stick? Does this have advantages/more degrees of freedom as well? Or do we stick here with Prague?

packages/client/src/execution/vmexecution.ts
```diff
@@ -179,21 +242,80 @@ export class VMExecution extends Execution {
...
+        if (!this.config.statelessVerkle) {
+          throw Error(`Currently stateful verkle execution not supported`)
+        }
+        this.config.logger.info(`Skipping Vm verkle statemanager genesis hardfork=${this.hardfork}`)
```
holgerd77, line 250: At least here in the logging: Vm -> VM. 🙂

packages/client/src/execution/vmexecution.ts
```diff
@@ -429,7 +611,25 @@ export class VMExecution extends Execution {
                       td,
                       timestamp,
                     })
+                    const isPostPrague = this.config.execCommon.gteHardfork(Hardfork.Prague)
+                    if (wasPrePrague && isPostPrague) {
+                      await this.transitionToVerkle(parentState!)
+                      clearCache = false
+                    }
```
holgerd77, line 618: Ok, but seeing all this Prague-centered logic here: maybe we keep Prague, right?

packages/common/src/eips.ts
```diff
@@ -471,6 +471,22 @@ export const EIPs: EIPsDict = {
...
+    gasConfig: {},
+    gasPrices: {
+      tx: {
+        v: 34300,
+        d: 'Per transaction. NOTE: Not payable on data of calls between transactions',
+      },
+    },
+    vm: {},
+    pow: {},
+  },
```
holgerd77, line 489: Can you add various `experimental` notes here, so to 6800, Kaustinen and Prague (always in the `comment` field)?

packages/statemanager/src/statelessVerkleStateManager.ts
```diff
@@ -0,0 +1,634 @@
...
+const HEADER_STORAGE_OFFSET = 64
+const CODE_OFFSET = 128
+const VERKLE_NODE_WIDTH = 256
+const MAIN_STORAGE_OFFSET = 256 ** 31
+
+const PUSH_OFFSET = 95
+const PUSH1 = PUSH_OFFSET + 1
+const PUSH32 = PUSH_OFFSET + 32
+
+export class StatelessVerkleStateManager implements EVMStateManagerInterface {
```
holgerd77, line 122: Please add code docs here with a class description/overview and an `experimental` note.

holgerd77 (changes requested):
Looks really great 🤩, some things to address.

packages/block/src/block.ts
g11tech: yes the distinction is there:

null means there is a non default witness and a default witness shouldn't be assigned in the constructor
undefined means that constructor can assign a default witness which is handy in block constructions for tests etc 

at this point we don't know the the common's hardfork because there could be setHardfork params to it. but what it esentially says to the block constructor is if the construction comes from  fromValues don't assign a deafult witness

packages/block/src/block.ts
g11tech: so we would reorder on breaking changes? in that case whenever that happens i would prefer we pass here the {header,transactions,...} as the first arg rather than the args array

holgerd77: Yes, I thought so too 😳. We should rearrange that along next breaking releases. This is otherwise always a bit annoying.

packages/block/src/block.ts
holgerd77: Ok, thanks for the clarification

holgerd77: Update: ok, I am also not completely overseeing the whole "working together", but we should minimally make sure that for the (all) 6800-deactivated cases there is never a default witness assigned to the block (not sure if this is still the case).

Even if "only internal" this would be wrong internal data and could (too easy) lead to side effects e.g. in serialization/output (JSON RPC) cases or whatever (accidentally also printing out witnesses).

packages/block/src/types.ts
gabrocheleau: Done

packages/client/src/execution/vmexecution.ts
gabrocheleau: done

packages/client/src/execution/vmexecution.ts
gabrocheleau: Updated to `this.vm.stateManager`

packages/client/src/execution/vmexecution.ts
gabrocheleau: Am also not sure why we're going with `null` instead of undefined. @g11tech ?

packages/block/src/block.ts
g11tech: yes we don't assign default witness if 6800 is not activated (see L414)

packages/blockchain/src/blockchain.ts
g11tech: ahh yes this makes sense as fork based checks are better in client

packages/client/bin/cli.ts
g11tech: yes, we can include other two as well as they are handy for debugging for merkle/stateful verkle (whenever that happens) and don't really have anything specific to stateless verkle

packages/client/src/execution/vmexecution.ts
g11tech: my opinion is that we keep client with HF centric logic while downstream libs should should EIP enabled centic logic

packages/statemanager/src/statelessVerkleStateManager.ts
gabrocheleau: Done

packages/block/src/block.ts
gabrocheleau: Done

packages/client/src/execution/vmexecution.ts
holgerd77: makes sense

packages/statemanager/package.json
```diff
@@ -52,6 +52,7 @@
     "@ethereumjs/rlp": "^5.0.1",
     "@ethereumjs/trie": "^6.0.1",
     "@ethereumjs/util": "^9.0.1",
+    "@ethereumjs/verkle": "^0.0.1",
```
holgerd77, line 55: Hmm, maybe also for the protocol: now we have the verkle package hardcoded and included into production code. Is this something we can (want to) live with? 🤔 Maybe we should give potential side effects some thought.

Respectively when thinking about it:

What we *DO* side-add here is an additional 500KB+ weight for the VM (in particular, SM of course as well) on dependencies through the current `rust-verkle-wasm` dependency (`577077 12 Nov 16:14 rust_verkle_wasm_bg.wasm`).

I think we really *NOT* want to do that, now thinking about that twice. 🤔 Might (will likely) be necessary at some point down the line, but for now it's far far too early to justify this addition for the production users (and maybe up till then we might be able to further optimize on the wasm library or ship with JS).

So I think we need to come up with a bit more manual solution here and pushing this to the optional peer dependencies for now.

packages/client/src/execution/vmexecution.ts
holgerd77: Can you maybe directly trie to update (aka: remove `null`) if Gajinder does not intervene in the meantime 🙂 and you also see no counter-arguments as well?

g11tech (commented):
looks good to me for merging in main codebase after we resolve the peer dependancy for the wasm

packages/statemanager/package.json
gabrocheleau: Yeah exactly, it's used for pedersenhash. Will update to peerDependency in stateManager and normal dep in the client.

gabrocheleau: pushed

g11tech (approved):
lgtm

Issue #614, opened 2019-11-11
Optionally dump post state after running state test


s1na:
It'd be often helpful for debugging if the state test runner would optionally (e.g. given a `--dumpState` flag) print either the full post state (all accounts in trie and their storages) or the state diff between `pre` and `post`.

There currently exists a `dumpState` in `tests/util.js` which is not being used. One challenge is that when traversing the trie after the fact we only have access to the hash of addresses (`SecureTrie` stores leaves under the hash of the key), and cannot infer the preimage of the hash (i.e. account/storage address) directly.

I tried inheriting `SecureTrie` and `CheckpointTrie` to add a map of form `hash: address` which would allow us to recover addresses from their hashes. But because of how the old version of `merkle-patricia-tree` does inheritance it's hard to get it right and I ended up copying code to VM directly.

After the `merkle-patricia-tree` release this can be done cleanly.

holgerd77:
Update: is this still a thing? Eventually @ryanio has some opinion as the "state expert" in the team? 😄

holgerd77:
When implementing this, also consider #560 along the way (not sure if this is meant here), will close #560 to consolidate on issues.

ryanio:
seems doable! we could implement this as suggested in OP by keeping track of a `hash: address` mapping and then iterate over that list to dump the entire state or a diff between the `pre` state.

@s1na do you have a fresh perspective on if this would still be valuable for you today?

@jochem-brouwer does this sound reasonable to you, would you prefer a different mechanism or would you find this helpful?

holgerd77:
Just an additional note here: we should be careful to not introduce any performance penalties when the flag is not used if we decide to implement.

s1na:
> @s1na do you have a fresh perspective on if this would still be valuable for you today?

I think I ran into this when needing to debug when some state tests were failing, but don't have a use for it now. If you've found yourself in a similar situation when the bug is just evading you then might be worth it :)

holgerd77:
@gabrocheleau can you take responsibilty for this issue, so either close or act upon? Thanks! 🙏

PR #3143, opened 2023-11-05
client: save preimages feature

gabrocheleau:
This PR will add a savePreimages option to the client, which will save a key-value map of keyHashes to their preimages (pre-hashed keys). This is a required preliminary step to eventually implement the merkle->verkle transition logic. 

Could also be somewhat related to #614


jochem-brouwer:
I would actually add this flag into Trie. Then, before you hash the key, save it into another DB (as preimage). This would only require an extra option for the DB to write too, though. But adding it into Trie means that we capture all trie items which are ever put, and we can be 100% (?) sure that we capture all preimages.

gabrocheleau:
> I would actually add this flag into Trie. Then, before you hash the key, save it into another DB (as preimage). This would only require an extra option for the DB to write too, though. But adding it into Trie means that we capture all trie items which are ever put, and we can be 100% (?) sure that we capture all preimages.

So I haven't pushed the whole WIP, as I was still playing around with some lower-level (vm/statemanager/trie) handling. I don't think we can strictly have that at the trie level, as it needs to be at least one-level higher up to handle the transition logic from trie->verkle-trie. Also, after discussing with gajinder, we though that the metaDB from the client seemed like a good place to put this. Nothing set in stone though, happy to hear your thoughts/ideas 


My current plan is to have the following structure:
- the stateManager stores every preimage in memory. If the option is activated at the stateManager level, then anytime there is a putAccount operation, we add the preimage to the stateManager key/value map. I think that should cover all "touched accounts", right? 
- then, at the runBlock level, once all the txs from the block have been executed, we retrieve those stored preimages from the statemanager and dump them. We would return the array of hashedKeys->preimages in `RunBlockResult`
- RunBlockResult can then be accessed by the client, and the preimages can be added to the metadb.

jochem-brouwer:
> the stateManager stores every preimage in memory. If the option is activated at the stateManager level, then anytime there is a putAccount operation, we add the preimage to the stateManager key/value map. I think that should cover all "touched accounts", right?

Yes, upon first thoughts I think this is right. What about the idea of Verkle where touched-but-deleted is different than in Merkle where touched-but-deleted does not change the trie? I think I recall this is dropped?

To test if all preimages are indeed stored in the StateManager I think we can add an extra test to the blockchain/state tests, what we do is we indeed copy all the MetaDB pre-image items, then after `runBlock` we copy the Trie and set the root to the root before the `runBlock` was ran. Then, we re-dump all these preimages in the copied trie and check if the root is the same reported root after `runBlock`.

packages/client/test/rpc/engine/preimages.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+    coinbase,
+  }
+  const blockData = { header: headerData, transactions: txs, withdrawals }
+  const executeBlock = Block.fromBlockData(blockData, { common })
+  const executePayload = blockToExecutionPayload(executeBlock, BigInt(0)).executionPayload
+  const res = await rpc.request('engine_newPayloadV2', [executePayload])
+  assert.equal(res.result.status, 'VALID', 'valid status should be received')
+  return executePayload
+}
+
```
holgerd77, line 99: Just putting this a bit random in the test file, comment is for the whole test setup though:

Are these tests in any form preimage related? 🤔 If not then I would suggest to move to `verkle.spec.ts` (but I might be missing something).

The whole test setup code is pretty extensive, is there a way to consolidate this in an existing `util.ts` file or - otherway around - use existing setup functionality for getting the setup, so that we do not create that much code bloat here?

packages/client/src/execution/preimage.ts
```diff
@@ -0,0 +1,21 @@
+import { DBKey, MetaDBManager } from '../util/metaDBManager'
+
+export class PreimagesManager extends MetaDBManager {
```
holgerd77, line 3: TBH: I still have my problems to understand what a preimage even is. Can we give this some explicit documentation/explanation? A link to some source also wound't hurt.

Maybe starting here and give the class some line of code docs where an explanation is included.

packages/common/src/interfaces.ts
```diff
@@ -83,6 +83,7 @@ export interface StateManagerInterface {
   getProof?(address: Address, storageSlots: Uint8Array[]): Promise<Proof>
   hasStateRoot(root: Uint8Array): Promise<boolean> // only used in client
   shallowCopy(downlevelCaches?: boolean): StateManagerInterface
+  getAppliedKey(address: Uint8Array): Uint8Array
```
holgerd77, line 86: Ah, this is not possible, since breaking. We need to make this optional for now.

packages/evm/src/journal.ts
```diff
@@ -69,6 +73,14 @@ export class Journal {
     this.accessList = new Map()
   }
 
+  /**
+   * Clears the internal `preimages` map, and marks this journal to start reporting
+   * the images (hashed addresses) of the accounts that have been accessed
+   */
+  startReportingPreimages() {
```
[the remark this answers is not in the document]
holgerd77: Another update: generally hanging a bit on this start/stop(?) design. Is this actually something which needs to be started/stopped again? (just an open question) This also reaches over to the VM where I wonder if the option is well placed for `runTx()` or should rather move to the VM initialization (and then should likely/eventually be passed over to EVM initialization?).

Maybe @jochem-brouwer has also some good additional comments on that.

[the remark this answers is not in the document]
holgerd77: Another another update: ok ok, just seeing that the access list feature is implemented in a similar semantics. So maybe we can also just keep the whole structure, only some mild suggestions/questions above.

holgerd77 (commented):
Ok, some comments. Please see through all comments before starting to change/adopt something, rather procedural comments where my line of thinking is changing during comment writing.

packages/client/src/execution/preimage.ts
jochem-brouwer, line 3: Nit: I would prefer the name `PreImagesManager` I think :thinking:

packages/evm/src/journal.ts
```diff
@@ -85,6 +97,13 @@ export class Journal {
   }
 
   private touchAccount(address: string) {
+    // If preimages are being reported, add the address to the preimages map
+    if (this.preimages !== undefined) {
+      const bytesAddress = unprefixedHexToBytes(address)
+      const hashedKey = this.stateManager.getAppliedKey(bytesAddress)
```
jochem-brouwer, line 103: I am confused. We want to store the preimage right? This stores the hashed key? :thinking:

packages/evm/src/journal.ts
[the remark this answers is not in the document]
jochem-brouwer: I agree that the name of `startReportingPreimages` is a bit confusing, but you are right, the access lists are implemented in a similar way. (might need to refactor it a bit to make it more clear)

packages/evm/src/journal.ts
```diff
@@ -85,6 +97,13 @@ export class Journal {
...
+      this.preimages.set(bytesToHex(hashedKey), bytesAddress)
```
jochem-brouwer, line 104: Oh wait, of course, you want a map `hashedKey -> originalKey` :man_facepalming:

packages/client/src/execution/preimage.ts
```diff
@@ -0,0 +1,21 @@
...
+  /**
+   * Returns the preimage for a given hashed key
+   * @param key the hashed key
+   * @returns the preimage of the hashed key
+   */
+  async getPreimage(key: Uint8Array): Promise<Uint8Array | null> {
+    return this.get(DBKey.Preimage, key)
```
jochem-brouwer, line 10: I now realize that this class "only" maps `get -> getPreimage` and `set -> setPreimage` :thinking: I'm not super sure if this is necessary, since it is clear what we are trying to get (`get(DBKey.Preimage, key)`). But not strong feelings.

packages/client/test/rpc/engine/preimages.spec.ts
jochem-brouwer: I agree here, I do not see how this tests the preimages.

packages/evm/src/journal.ts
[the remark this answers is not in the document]
jochem-brouwer: I think the `runTx` place is correct though, the end user can specify in the `runBlock` / `runTx` options if they want preimages or not.

packages/vm/test/api/runTx.spec.ts
```diff
@@ -317,6 +319,24 @@ describe('runTx() -> API parameter usage/data errors', () => {
...
+    const acc = createAccount()
+    await vm.stateManager.putAccount(caller, acc)
+
+    const res = await vm.runTx({ tx, reportPreimages: true })
+
+    const hashedCallerKey = vm.stateManager.getAppliedKey(caller.bytes)
+
+    const retrievedPreimage = res.preimages?.get(bytesToHex(hashedCallerKey))
+
+    assert.ok(retrievedPreimage !== undefined && equalsBytes(retrievedPreimage, caller.bytes))
```
jochem-brouwer, line 337: This tests preimages for accounts, I would also add a test here where storage slot preimages are saved. Do we also need it for the code hashes? (I don't think so - right? In context of the Verkle transition)

jochem-brouwer (commented):
Some comments

packages/client/test/rpc/engine/preimages.spec.ts
g11tech: ahh yes, the comparison with generated and expected pre0images is missing , can add that as well cc @gabrocheleau

packages/common/src/interfaces.ts
gabrocheleau: Made this optional, and adjusted the code.

holgerd77:
Updated this via UI

packages/client/src/execution/preimage.ts
gabrocheleau: I believe that Preimage is actually the correct casing. Preimage is the correct mathematical term (i.e. not pre-image). See for e.g. https://en.wiktionary.org/wiki/preimage

packages/evm/src/journal.ts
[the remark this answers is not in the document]
gabrocheleau: I agree that the naming and general architecture of this is not super intuitive. However, I opted to go for consistency with accessList, as I believe these two (preimages and accessList) handling should definitely be consistent (imho inconsistency would be even more confusing than this arguably unclear design choice).

Perhaps we can mark it as a task for v7? wdyt @holgerd77  https://github.com/ethereumjs/ethereumjs-monorepo/issues/3216

packages/vm/test/api/runTx.spec.ts
gabrocheleau: We have added some additional test vectors from the client-side. Notably we also needed to test blockWitnesses and coinbase, and I think the txs that are being tested at the client level would also indirectly test storage slot preimages. I feel like it's fine to leave that runTx test relatively minimal (basically just testing that the reportPreiamges option results in preimages being added to the journal and then to the txResult), wdyt?

packages/client/test/rpc/engine/preimages.spec.ts
gabrocheleau: I believe this has been addressed with the latest changes.

packages/client/test/rpc/engine/preimages.spec.ts
```diff
@@ -0,0 +1,258 @@
...
+        '0x4000000000000000000000000000000000000000',
+        '0x4100000000000000000000000000000000000000',
+        '0x4200000000000000000000000000000000000000',
+        '0x4300000000000000000000000000000000000000',
+        '0x4400000000000000000000000000000000000000',
+        '0x4500000000000000000000000000000000000000',
+        '0x4600000000000000000000000000000000000000',
+        '0x4700000000000000000000000000000000000000',
+      ],
+      // no txs
```
jochem-brouwer, line 228: Are there no txs in this block? It seems that there are?

jochem-brouwer: Should there also be a test where a contract is invoked and a storage slot is written to?

packages/vm/src/types.ts
```diff
@@ -274,12 +274,30 @@ export interface RunBlockOpts {
...
+   */
+  reportPreImages?: boolean
 }
 
 /**
- * Result of {@link runBlock}
+ * Result of {@link applyBlock}
  */
-export interface RunBlockResult {
+export interface ApplyBlockResult {
```
jochem-brouwer, line 288: Hmm, I  think changing the name of this interface is breaking?

jochem-brouwer (commented):
Some comments

packages/vm/src/types.ts
gabrocheleau: It shows up weird in the git diff, but I just added an ApplyBlockResult interface, and modified RunBlockResult so that it extends it.

packages/client/test/rpc/engine/preimages.spec.ts
gabrocheleau: @g11tech would know more about those cases than me, would you mind clarifying gajinder?

g11tech: yes there are, we need to run the test with debug enabled for evm and add the list of things accessed here

gabrocheleau: I logged the accesses and added some tx preimages to the two blocks that have transactions.

packages/vm/test/api/runTx.spec.ts
g11tech: yes i believe the client test has a tx accessing storage

g11tech (approved):
lgtm

jochem-brouwer (dismissed):
LGTM

jochem-brouwer (approved):
LGTM!

Great work have removed my preimage -> preImage stuff :)

holgerd77:
Seems browser tests are not going through here?

packages/evm/src/types.ts
```diff
@@ -150,9 +150,11 @@ export interface EVMInterface {
     putAccount(address: Address, account: Account): Promise<void>
     deleteAccount(address: Address): Promise<void>
     accessList?: Map<string, Set<string>>
+    preimages?: Map<string, Uint8Array>
     addAlwaysWarmAddress(address: string, addToAccessList?: boolean): void
     addAlwaysWarmSlot(address: string, slot: string, addToAccessList?: boolean): void
     startReportingAccessList(): void
+    startReportingPreimages(): void
```
holgerd77, line 157: Oh, just seeing on release notes writing: this needs to be optional for backwards compatibility.

PR #3146, opened 2023-11-06
Client/Blockchain: receipt reorg logic

jochem-brouwer:
This PR attempts to resolve hive failures in:

`./hive --client ethereumjs --sim ethereum/engine --sim.limit="engine-cancun/Transaction Re-Org"` and
`./hive --client ethereumjs --sim ethereum/engine --sim.limit="engine-cancun/Invalid Missing Ancestor Syncing ReOrg"`

This PR:

- Adds events to `blockchain`: once blocks get reorged and deleted, `deletedCanonicalBlocks` event is emitted
- Adds an extra guard on `eth_getTransactionReceipt`: this verifies that the receipt is in the canonical chain
- Deletes receipts in client once a block gets reorged out
- Ensures INVALID is not returned on `newPayload` / `forkchoiceUpdated` in  case the block is lower than the currently known invalid blocks.

To test this, an extra diff has to be added. This hacks in a way that a peer is always returned, to ensure that the syncing process is completed. Diff for `./packages/client/src/sync/beaconsync.ts`:

```diff
       const latest = await this.latest(peer)
       if (latest) {
         const { number } = latest
-        if ((!best && number >= this.chain.blocks.height) || (best && best[1] < number)) {
+        // if ((!best && number >= this.chain.blocks.height) || (best && best[1] < number))
+        {
           best = [peer, number]
         }
       }
```

When running **all** tests, this brings the failing tests down to just 3 (note, if more fail, re-run these individual tests, they might  be flakey which sometimes happens)

```
Inconsistent Head in ForkchoiceState
Valid NewPayload->ForkchoiceUpdated on Syncing Client
Invalid PayloadAttributes, Missing BeaconRoot, Syncing=False
```


packages/blockchain/src/blockchain.ts
```diff
@@ -1127,34 +1152,48 @@ export class Blockchain implements BlockchainInterface {
...
-      // reset stale iterator heads to current canonical head this can, for
-      // instance, make the VM run "older" (i.e. lower number blocks than last
-      // executed block) blocks to verify the chain up to the current, actual,
-      // head.
-      for (const name of Object.keys(this._heads)) {
-        if (equalsBytes(this._heads[name], hash)) {
-          this._heads[name] = headHash
+        if (this.events.listenerCount('deletedCanonicalBlocks') > 0) {
+          const block = await this.getBlock(blockNumber)
+          this._deletedBlocks.push(block)
```
g11tech, line 1164: i think this is only real change, rest is just formatting change because of try catch

g11tech (approved):
lgtm,

PR #3147, opened 2023-11-06
tsconfig: fix references

jochem-brouwer:
When trying to super-optimize hive, I noted that if you partially copy our monorepo and then try to build a package (all dependencies are in the folder) then for `evm` and `genesis` there are references to packages which are not used. I think we can remove those references?


acolytec3 (commented):
LGTM.  All the tests still pass

acolytec3 (approved):

PR #3148, opened 2023-11-07
client: add hive instructions

jochem-brouwer:
This PR adds instructions to run hive (also in the new "fast development" mode). It would be great if someone could also test the instructions to ensure these are correct (and readable :smile: )

### DIFF
--- a/packages/client/DEVELOPER.md
+++ b/packages/client/DEVELOPER.md
@@ -1,0 +1,23 @@
+# Client Developer docs
+
+## Running Hive Tests
+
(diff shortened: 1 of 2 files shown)


acolytec3 (approved):
Looks great.  I've stepped through both sections and everything works well.  That said, it would be nice to incorporate the revised docker file into hive proper under `Dockerfile.local` and maybe add the instructions from the PR to Hive (or else in this document).  This can be taken up in a future PR though.

PR #3151, opened 2023-11-08
vm: fix nested tests

gabrocheleau:
This PR "un-nests" some nested `it` vitest tests. Nested `it` functions do not get executed, and therefore "silently pass".

### DIFF
--- a/packages/vm/test/api/runBlock.spec.ts
+++ b/packages/vm/test/api/runBlock.spec.ts
@@ -27,7 +27,7 @@ import type { TypedTransaction } from '@ethereumjs/tx'
 const testData = require('./testdata/blockchain.json')
 const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Berlin })
 describe('runBlock tests', () => {
-  it('runBlock() -> successful API parameter usage', async () => {
+  describe('runBlock() -> successful API parameter usage', async () => {
     async function simpleRun(vm: VM) {
       const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.London })
       const genesisRlp = toBytes(testData.genesisRLP)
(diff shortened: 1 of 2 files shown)


packages/vm/test/api/runBlock.spec.ts
```diff
@@ -27,7 +27,7 @@ import type { TypedTransaction } from '@ethereumjs/tx'
 const testData = require('./testdata/blockchain.json')
 const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Berlin })
 describe('runBlock tests', () => {
-  it('runBlock() -> successful API parameter usage', async () => {
+  describe('runBlock() -> successful API parameter usage', async () => {
```
acolytec3, line 30: What would you say to just removing the nested `describe` anyway?  I don't think it gives us any advantage in this case.

gabrocheleau: Yeah sure, just wanted to minimally change what was already there, but agree they don't do much, will remove.

packages/vm/test/api/runTx.spec.ts
```diff
@@ -44,7 +44,7 @@ const TRANSACTION_TYPES = [
 const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.London })
 common.events.setMaxListeners(100)
 describe('runTx tests', () => {
-  it('runTx() -> successful API parameter usage', async () => {
+  describe('runTx() -> successful API parameter usage', async () => {
```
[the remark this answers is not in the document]
gabrocheleau: Removed the top-level one.

packages/vm/test/api/runBlock.spec.ts
gabrocheleau: Removed the top-level one.

jochem-brouwer (approved):
LGTM. I have not compared the diffs but have instead compared the new file to the old file.

PR #3153, opened 2023-11-10
client: ensure executed block does not get pruned if head=final FCU

jochem-brouwer:
This PR fixes:

`./hive --client ethereumjs --sim ethereum/engine --sim.limit="engine-cancun/Valid NewPayload->ForkchoiceUpdated on Syncing Client`

In this test, first FCU is called (after receiving the VALID newPayload):

```
  {
    headBlockHash: A
    safeBlockHash: A
    finalizedBlockHash: A
  }
```
We reply VALID  

  Then, FCU is called again, but now with:

```
  {
    headBlockHash: A
    safeBlockHash: B
    finalizedBlockHash: C
  }
```

B,C are known and VALID. However, before this PR we replied Syncing, since the `pruneCachedBlocks` will actually deleted the just-executed block A from the executed blocks. So if we FCU again (without newPayload which executes the block if this is possible) then it sees that the block is not executed (we thus start executing it) but reply early here with SYNCING.

Note: PR still draft because might fix other hive tests here too.


holgerd77:
@jochem-brouwer what is the status of this PR?

jochem-brouwer:
Ah right I intended this to include more hive fixes. We can use this PR to fix the Cancun failing hive tests, which I assigned myself to.

holgerd77:
@jochem-brouwer ah, can we then please rather take this one in and open a new one instead of re-cycle?

This re-purposing of PRs is always a bit problematic.

I would then merge this in here.

jochem-brouwer:
Sure, that is fine!

holgerd77:
Updated this via UI

holgerd77:
Updated this via UI

holgerd77 (approved):
LGTM

PR #3154, opened 2023-11-12
Some Verkle Clean-Ups

holgerd77:


packages/verkle/LICENSE
jochem-brouwer, line 1: I am fine with MIT, just note: our other packages all seem to be MPL-2

packages/verkle/src/types.ts
```diff
@@ -60,7 +60,7 @@ export interface Point {
 
 export type Proof = Uint8Array[]
 
-export interface VerkleTrieOpts {
+export interface VerkleTreeOpts {
```
jochem-brouwer, line 63: I have no strong opinion about this, but just noting: our MPT is a merkle Trie, so the Trie would be more in-line with that. But I don't mind changing it to a Tree.

jochem-brouwer (commented):
2 comments

packages/verkle/src/types.ts
jochem-brouwer: Actually the technical term is `Tree` (Merkle Tree), so it would actually make sense to rename this to a `Tree` :smile:

gabrocheleau: Yes we discussed that a bit yesterday, and decided to go with Tree. The "trie" terminology (even though technically correct for data structures where to retrieve a value for a given key, you follow the "path" of that key from node to node) tends to confuse people anyway

packages/verkle/test/verkle.spec.ts
```diff
@@ -45,7 +45,7 @@ const absentKeys = [
 
 describe('Verkle trie', () => {
```
gabrocheleau, line 46:
```suggestion
describe('Verkle tree', () => {
```

gabrocheleau (approved):

gabrocheleau:
Did an additional round of trie->tree renaming throughout the package to be consistent in the comments and some helper functions (e.g. walkTrie -> walkTree).

packages/verkle/LICENSE
holgerd77: That's a historical burden (one of the strongest we have, though also not a totally grave one), and e.g. the reason Richard (Ethers) never user our libraries. So all new ones should be the less restrictive MIT (we can't change the old ones any more unfortunately since we would need permission from all contributors).

jochem-brouwer: Ok that is indeed a pity. It might be worthwhile to actually try to contact older contributors to see if we can change the license, especially if it means that we get more users.

holgerd77: Ah, this train is gone for sure. 😆 It is just not possible to contact 50+ contributors - likely for just one library - who did some work over the years. Won't even be possible to reach all of them, and it needs only one or a few persons not wanting the change to stop this.

But: it is also not big of a practical problem. After Richard complained years ago we never ever had any license issues any more.

PR #3158, opened 2023-11-16
fix: Check that code is not a precompile when checking code

roninjin10:
I noticed a footgun that if your precompile takes no arguments the entire function will fail to recognize the code exists. This is because it checks code.length and code.length returns the number of arguments in the case of a function. Fix by checking if the code is a function


jochem-brouwer:
How did you end up finding this bug? Did you create a custom precompile?

roninjin10:
Yea I found it while debugging why [this test wasn't failing](https://github.com/evmts/evmts-monorepo/pull/670/files#diff-bf6818a43572aeb70e72b3b8be250639c5bfc97b7b421c8007c0c60c9e52aeeaR57)

I'm building a app building framework called EVMts. It allows you to import solidity files directly into typescript files and then execute them with ethereumjs in the browser or node. [a user asked for the feature of adding precompiles](https://twitter.com/norswap/status/1724944291204665439) so he could use it to encode client functionality in an onchain contract is why I added the feature.

packages/evm/src/evm.ts
```diff
@@ -251,7 +251,7 @@ export class EVM implements EVMInterface {
     // Load code
     await this._loadCode(message)
     let exit = false
-    if (!message.code || message.code.length === 0) {
+    if (!message.code || (typeof message.code !== 'function' && message.code.length === 0)) {
```
roninjin10, line 254: Happy to add a test if we want one

holgerd77: Really interesting find 🙏 and usage! Yes, it would be great if you can take the relevant parts of the test case over.

holgerd77:
Just approved the test run so that we can see that this generally works.

jochem-brouwer:
Have linted and added a test

packages/evm/src/evm.ts
roninjin10: Apologies GitHub notifications are shambles I didn’t see this comment

acolytec3 (approved):
LGTM

holgerd77:
Thanks for the continuation! 🙏🙂

holgerd77:
This will be fixed in #3261

PR #3161, opened 2023-11-16
Export originalStorageCache

### LINKED ISSUE #3160: `OriginalStorageCache` not exported
In the `StateManager` interface, there is a required property `OriginalStorageCache` and we have working implementation of it within our codebase [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/3b2e97223d3a5b541007dc9b0da66afd2b3c8f7d/packages/statemanager/src/cache/originalStorageCache.ts#L7) and this should be addressed by adding `originalStorageCache.ts` to the list of exports in the `statemanager/src/cache/index.ts` file.  Others implementing their own state manager like [evmts](https://github.com/evmts/evmts-monorepo/blob/main/vm/src/stateManager/Cache.ts) and Ultralight have had to clone our implementation unnecessarily.

acolytec3:
Fixes #3160 and removes pain point for implementing one's own storage manager.

### DIFF
--- a/packages/statemanager/src/cache/index.ts
+++ b/packages/statemanager/src/cache/index.ts
@@ -1,4 +1,5 @@
 export * from './account.js'
 export * from './code.js'
+export * from './originalStorageCache.js'
 export * from './storage.js'
 export * from './types.js'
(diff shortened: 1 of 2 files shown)


jochem-brouwer (approved):
LGTM

PR #3162, opened 2023-11-17
vm: update tests to v13 (cancun tests)

jochem-brouwer:
Updates to https://github.com/ethereum/tests/releases/tag/v13

Some gotcha's:

Test data had to be updated to ensure that the RLP-encoded 4844 txs have a `to` field. Some test data has `to` field with empty bytes, this is not possible per spec.

The `stateRoot` of some tests updates because we implicitly create the beacon root address. If there is no code there (if there is storage there it does not matter) it should be deleted anyways. The change in `runBlock.ts` changes the state root in case that the beacon root address has no code (it is now wiped from the Trie instead of creating a semi-empty account (it has storage, but per definition if it is touched should be deleted after a block run).


jochem-brouwer:
On hold, blockchain tests do seem to ignore EIP 4788.

jochem-brouwer:
Never mind, it's the touched logic again. Should pass now, have included some 4844 spec fixes.

jochem-brouwer:
Nvm, the blockchain/state tests pass locally but it seems that tx now errors out

packages/tx/src/constants.ts
```diff
@@ -3,7 +3,7 @@
 export const MAX_CALLDATA_SIZE = 16777216 // 2 ** 24
 export const MAX_ACCESS_LIST_SIZE = 16777216 // 2 ** 24
 export const MAX_VERSIONED_HASHES_LIST_SIZE = 16777216 // 2 ** 24
-export const LIMIT_BLOBS_PER_TX = 16777216 // 2 ** 24
+export const LIMIT_BLOBS_PER_TX = 6 // 786432 / 2^17 (`MAX_BLOB_GAS_PER_BLOCK` / `GAS_PER_BLOB`)
```
g11tech, line 6: i think we should move this constant to the teser/runner since this would be hf dependant anyway

g11tech (commented):
mostly looks good to me can approve once CI passes

packages/tx/src/constants.ts
jochem-brouwer: We can move `MAX_BLOB_GAS_PER_BLOB` and `GAS_PER_BLOB` into common?

jochem-brouwer:
Updated the opening comment on this PR in case there is some confusion why some test files had to be updated (explicitly checked all)

packages/tx/src/constants.ts
jochem-brouwer: (and then derive max amount of blobs from that)

holgerd77 (approved):
LGTM

Issue #3157, opened 2023-11-15
Client: Mitigate BoundProtocol Insanities


holgerd77:
It has become tradition in our team to rant about the BoundProtocol mechanism connecting networking protocols like ETH or SNAP to a client peer. For a very good reason: to call an ETH protocol `GET_BLOCK_HEADERS` message e.g. goes through many (many) hops and it is close to not possible to (easily) follow the code path in a static way (static in a way of: having concrete TypeScript types available which will guide you to the message call implementation).

The original intent of all of this (written by Vinay who did the first client version in JavaScript) was to remain flexible on adding new protocols and therefore bind protocols and runtime and also do the code in a structurally open way (all protocols possible). Especially the latter was more fitting for a JavaScript world but it now hindering in a TypeScript world, since it prevents the most basic type associations (that an RLPX peer (naturally) has an eventual ETH binding e.g.).

I do think having a flexible architecture here has benefits and we should not completely give up on this, neither on the "application" protocol layer (it was e.g. easily possible to add SNAP alongside ETH with this) nor on the transport layer (so: devp2p, since we might *also* want to expand here, and (who knows?) add Portal Network as a network provider here.

When looking at the code I nevertheless think we can nevertheless take various steps to reduce the call chain and make the code readable/followable again. So we can take it there in a first round and see where we go from there. Note that the number of "hops" (getting from the inititial call trigger e.g. for a `GET_BLOCK_HEADERS` message to the actual execution) is really insane (6?? steps or so?) and this is therfore only to *reduce* the number of steps in a first round (to let's say: 2-3) and will not get to 1 in a first round. It should be nevertheless very much possible to do the steps in isolation and get to a point where the (client networking) code becomes readable again.

Following suggested steps:

### Binding

1. The `BoundProtocol` (`net/protocol/boundprotocol.ts`) is handling too much logic to directly replace or refactor (if necessary, also no full picture). We can nevertheless improve typing here by adding very small subclasses "per protocol" (best in the file itself) like the following with several better typed parameters:

```typescript
export class BoundETHProtocol {
  public name: 'ETH' // public name: string
  private protocol: EthProtocol // private protocol: Protocol
  private peer: RlpxPeer // private peer: Peer
  private sender: RlpxSender // private sender: Sender
}
```

This adds little boilerplate which I think is very much worth it.

2. (the more important one)
[...]

PR #3164, opened 2023-11-21
Mitigate boundprotocol insanities

jochem-brouwer:
Addresses #3157

Type safety:

image

### DIFF
--- a/packages/client/src/net/peer/peer.ts
+++ b/packages/client/src/net/peer/peer.ts
@@ -1,13 +1,9 @@
 import { EventEmitter } from 'events'
 
+import { BoundEthProtocol, BoundLesProtocol, BoundSnapProtocol } from '../protocol'
+
 import type { Config } from '../../config'
-import type {
-  BoundEthProtocol,
-  BoundLesProtocol,
-  BoundProtocol,
-  BoundSnapProtocol,
-  Protocol,
-} from '../protocol'
+import type { BoundProtocol, Protocol, Sender } from '../protocol'
 import type { Server } from '../server'
 
 export interface PeerOptions {
@@ -99,6 +95,42 @@ export abstract class Peer extends EventEmitter {
    */
   handleMessageQueue() {
     this.boundProtocols.map((e) => e.handleMessageQueue())
+  }
+
+  async addProtocol(sender: Sender, protocol: Protocol): Promise<void> {
+    let bound: BoundProtocol
+    const boundOpts = {
+      config: protocol.config, // TODO: why cant we use `this.config`?
+      protocol,
+      peer: this,
+      sender,
+    }
+
+    if (protocol.name === 'eth') {
+      bound = new BoundEthProtocol(boundOpts)
+    } else if (protocol.name === 'les') {
+      bound = new BoundLesProtocol(boundOpts)
+    } else if (protocol.name === 'snap') {
+      bound = new BoundSnapProtocol(boundOpts)
+    } else {
+      throw new Error(`addProtocol: ${protocol.name} protocol not supported`)
+    }
+
+    // Handshake only when snap, else
+    if (protocol.name !== 'snap') {
+      await bound!.handshake(sender)
+    } else {
+      if (sender.status === undefined) throw Error('Snap can only be bound on handshaked peer')
+    }
+
+    if (protocol.name === 'eth') {
+      this.eth = <BoundEthProtocol>bound
+    } else if (protocol.name === 'snap') {
+      this.snap = <BoundSnapProtocol>bound
+    } else if (protocol.name === 'les') {
+      this.les = <BoundLesProtocol>bound
+    }
+    this.boundProtocols.push(bound)
   }
 
   toString(withFullId = false): string {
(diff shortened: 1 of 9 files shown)


packages/client/src/net/peer/peer.ts
```diff
@@ -38,17 +37,22 @@ export interface PeerOptions {
  * Network peer
  * @memberof module:net/peer
  */
-export class Peer extends EventEmitter {
+export abstract class Peer extends EventEmitter {
```
acolytec3, line 40: One general question here is what advantage do we have with an abstract `peer` class and then a concrete `RLPXPeer` subclass?  We don't have multiple network options so this feels like abstraction without a purpose.

acolytec3 (commented):
Looks like a good start on a first review.  Will examine some of the test output later today.

packages/client/src/net/peer/peer.ts
jochem-brouwer: Not super sure here, I just made it abstract since we instantiate it nowhere and it had some obvious abstract methods which should be implemented (empty method which just throws if you call it).

jochem-brouwer:
The tests should now pass. Also, the client can sync again. Open for a first round of reviews.

image

packages/client/test/net/peer/peer.spec.ts
```diff
@@ -1,10 +1,7 @@
-import { EventEmitter } from 'events'
-import * as td from 'testdouble'
 import { assert, describe, it } from 'vitest'
 
 import { Config } from '../../../src/config'
 import { Peer } from '../../../src/net/peer'
-import { Event } from '../../../src/types'
 
 describe('[Peer]', () => {
   const config = new Config({ accountCache: 10000, storageCache: 1000 })
```
acolytec3, line 7: It still feels very weird to me that we now have tests for an abstract class that Typescript doesn't think you can invoke.  In the compiled Javascript, this `Peer` class is just a normal class extending `EventEmitter` but we now have these ugly red squiggles to deal with.  Why is Typescript even letting this pass?

packages/client/test/net/peer/rlpxpeer.spec.ts
```diff
@@ -29,6 +29,8 @@ describe('[RlpxPeer]', async () => {
       id: 'abcdef0123',
       host: '10.0.0.1',
       port: 1234,
+      address: 'address0',
```
acolytec3, line 32: This is one of those weird artifacts of previously having had a libp2p transport option.  The RlpxPeer doesn't actually use the `address` string at all and I don't think it's a used element anywhere in our code.  I think we can safely delete this line since the `RlpxPeer` constructor creates this value and passes it to the `Peer` constructor.

acolytec3 (approved):
I think this can be addressed in a follow-up PR but we should deal with this abstract `Peer` class and it associated tests.  If we're going to convert it to an abstract class, we should migrate those tests to `RlpxPeer` in some form.

Otherwise, looks great!

acolytec3:
Well done, ser.

packages/client/src/net/peer/peer.ts
```diff
@@ -38,17 +33,22 @@ export interface PeerOptions {
...
-  public bound: Map<string, BoundProtocol>
   protected transport: string
   protected protocols: Protocol[]
+  protected boundProtocols: BoundProtocol[] = []
   private _idle: boolean
 
+  // TODO check if this should be moved into RlpxPeer
+  public eth?: BoundEthProtocol
+  public snap?: BoundSnapProtocol
+  public les?: BoundLesProtocol
```
holgerd77, line 50: Yes, not totally urgent, but these should be moved to RlpxPeer

packages/client/src/net/peer/peer.ts
```diff
@@ -38,17 +33,22 @@ export interface PeerOptions {
...
+export abstract class Peer extends EventEmitter {
   public config: Config
   public id: string
   public address: string
   public inbound: boolean
   public server: Server | undefined
-  public bound: Map<string, BoundProtocol>
   protected transport: string
   protected protocols: Protocol[]
+  protected boundProtocols: BoundProtocol[] = []
```
holgerd77, line 44: Is this still an intermediate step to keep this generic `boundProtocols` structure? Otherwise we loose many of the typing benefits here again? 🤔

I thought we can call on `peer.eth.....` and so forth directly?

holgerd77 (commented):
Cool to see this merged, some questions, nothing urgent though.

packages/client/src/net/peer/peer.ts
jochem-brouwer: We can call `peer.eth` directly. The reason why I moved this `boundProtocols` here is because it is used to keep track of what protocols we have added. We can actually remove this, and just do simple checks: `if (this.eth !== undefined)` then we know that we have support for eth protocol.

(Not entirely sure if this answers your question / point?)

The only places where it is used is in `toString()` and into `handleMessageQueue`, this `boundProtocols` can easily be removed (and maybe should do so already?).

holgerd77: If it's only used on these few occasions that it is no big deal for sure. You are a lot deeper into it, then just decide yourself if you'd rather want to keep or not.

jochem-brouwer: I'll open a cleanup PR and will also deal with the Peer/RlpxPeer stuff

PR #3165, opened 2023-11-26
Remove debug loggers devp2p

jochem-brouwer:
This PR removes debug calls in the devp2p package if no DEBUG flag is set. This ensures that we do not create debug strings which we later throw away if no logger is attached. This was already the case in the eth protocol, but this is now package-wide. the only place where it is not added is in `assertEq` in `util.ts`


holgerd77 (approved):
Nice, LGTM

PR #3169, opened 2023-11-28
evm: fix modexp edge cases

### LINKED ISSUE #3168: Precompile contract modexp returns incorrect value in corner case
# Bug description

The precompile modexp is bias from the standard impementation.

When the `Bsize`  is zero, [the TypeScript EVM returns a BINGINT_0 directly](https://github.com/ethereumjs/ethereumjs-monorepo/blob/0170906f2880e01951bc56cc2c86fc263da21cc3/packages/evm/src/precompiles/05-modexp.ts#L154-L159), failed to account for the scenario of  $0^0$. In this case, the modexp is expected to return 1 instead of 0.

Furthermore, if the `Msize` is 0, the return value should be empty instead of 0. 

The following bytecode invokes the modexp with `Bsize = 0, Esize = 0, Msize = 1`, the expected output should be `01ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff`, however  the TypeScript EVM returns `00ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff`: 

```
60006000526000602052600160405260ff60605360ee60615360dd6062537fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff608052602060806080600060055afa60206080f3
```

Another example is when the modexp is invoked with `Bsize = 0, Esize = 0, Msize = 0`. Geth returns `ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff`, while the TypeScript EVM returns `00ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff`:

```
60006000526000602052600060405260ff60605360ee60615360dd6062537fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff608052602060806080600060055afa60206080f3
```

jochem-brouwer:
Closes #3168 

This PR fixes the edge cases in ModExp. Specifically, ModExp has 6 inputs, and the inputs for [0,1,2] are all tested on all inputs. These are cross-referenced against infura. (File here: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3169/commits/144963affd0bfb72c139ac17210449747abf5539)

Test files are here. The format is [input, expectedOutput]

[tests.txt](https://github.com/ethereumjs/ethereumjs-monorepo/files/13489041/tests.txt)

Note: one of the causes is that `bigIntToBytes(BigInt(0))` returns `Uint8Array(1) [ 0 ]` (length 1) instead of `Uint8Array(0) []` (length 0). If you then `setLengthLeft(r, 0)` then it still outputs the array with length 1. This raises 2 questions:

- Should `bigIntToBytes` on `BigInt(0)` return the empty array? (Note: this is super breaking so we cannot do it, we could introduce a new method?)
- Should `setLengthLeft` throw if the expected length is lower than the current length? (It is unknown if we should trim left or right :thinking: )


jochem-brouwer:
Have added a fuzzer which I will run now. It finds more error cases.

jochem-brouwer:
Fuzzer tests: run inputs [0,1,2] on the modExp inputs (6 fields: `bLen, eLen, mLen, B, E, M`) (729 tests). Failures: 35.

jochem-brouwer:
[tests.txt](https://github.com/ethereumjs/ethereumjs-monorepo/files/13489041/tests.txt)

Test cases: array of `[input, expectedOutput]` retrieved from infura for all combinations [0,1,2] on the modExp inputs (6 fields: `bLen, eLen, mLen, B, E, M`)

acolytec3 (commented):
Generally looks good.  I think we should definitely consider whether these utility methods (like `setLengthLeft`) should throw if there is under/overflow but that can be solved separately.

One nit suggestion too.

Will approve once the "fuzzer" tests are added.

acolytec3 (approved):
LGTM!

PR #3170, opened 2023-11-29
Boundprotocol cleanup

jochem-brouwer:
This PR is a follow-up cleanup PR of #3164:

1. Removes `boundProtocols` property from Peer. See: 01b048cc02f1d2fcec5bdc612813968843624cbd
2. Moves Peer into RlpxPeer (?)

Discuss: the second item here is not trivial and requires many changes to files. I also need to fix MockPeer and MockServer in the tests it seems. Do we really want this? It changes the type all over the client, and if we ever want to introduce some new peer we will have to refactor back such that we use an abstract class again.

I have dumped current work into 4ba9bf97fc8f81c6bec480daca4eefb836d9cc2d, but it looks like it needs some more non-trivial work on the tests so I've stopped here. Do we really want to collapse Peer into RlpxPeer?


acolytec3:
I'm a big fan of collapsing the peer classes.  That said, I would go with the class name `Peer` and not `RlpxPeer`.  In our idealized future state, we won't have "rlpxPeers" and "otherTransportPeers".  We'll just have "Peers" and their contact info in the routing table will identify what transports they support (i.e. Rlpx and that's basically it).

jochem-brouwer:
Cool, I will go ahead with the rename `RlpxPeer` -> `Peer` and then removing the old `peer.ts`.

PR #3171, opened 2023-11-30
chore: Improve error handling of fetchFromProvider

roninjin10:
Small improvements to fetchFromProvider
- Add an example jsdoc
- handle non ok status codes


roninjin10:
Gonna have to update some mocks that are failing from the change

acolytec3:
This looks great. Updates should be pretty minimal. I'll take a look at them this afternoon.

acolytec3:
CI is failing on the lint step. Can you run the linter? Fixes look like things that `npm run lint:fix` will address automatically. Will approve and merge once the CI passes

roninjin10:
Oh I didn't push is why they were failing lol

acolytec3:
One more request if you don't mind is could you add the below test to the `util/test/provider.spec.ts` test file so we get updated code coverage?
```
it('should throw a formatted error when an error is returned from the RPC', async () => {
    const realFetch = global.fetch
    //@ts-expect-error -- Typescript doesn't like us to replace global values
    global.fetch = async (_url: string, _req: any) => {
      return {
        text: () => {
          return 'ERROR'
        },
        ok: false,
      }
    }
    try {
      await fetchFromProvider(providerUrl, {
        method: 'eth_getBalance',
        params: ['0xabcd'],
      })
      assert.fail('should throw')
    } catch (err: any) {
      assert.ok(err.message.includes('ERROR'), 'received a formatted RPC error')
      assert.ok(err.message.includes('eth_getBalance'), 'error is for correct method')
    }
    global.fetch = realFetch
  })
```

roninjin10:
Definitely. Btw if you know how to enable pushes let me know I couldn't figure it out clicking around github for a bit

packages/util/src/provider.ts
```diff
@@ -27,6 +37,15 @@ export const fetchFromProvider = async (url: string, params: rpcParams) => {
     method: 'POST',
     body: data,
   })
+  if (!res.ok) {
+    throw new Error(
+      `JSONRpcError: ${JSON.stringify(
+        { method: params.method, status: res.status, message: res.text() },
```
roninjin10, line 43: There is a corner case where this res.text() will throw too. This is when the network fails before the response is fully recieved in this case res.text() will throw. I think this is fine though since it's such a unique corner case

roninjin10: fixed

packages/util/test/provider.spec.ts
```diff
@@ -40,4 +41,26 @@ describe('fetchFromProvider', () => {
       assert.ok(err.message.includes('fetch'), 'tried to fetch and failed')
     }
   })
+
+  it('should throw a formatted error when an error is returned from the RPC', async () => {
+    vi.stubGlobal('fetch', async (_url: string, _req: any) => {
```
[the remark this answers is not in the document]
acolytec3: Ah, cool, I didn't know about this method in the `vi` API. Might be able to use that elsewhere in the tests

roninjin10:
Hmm not obvious to me why codecove thinks these lines aren't covered anymore 
image

roninjin10:
Code cov was still failing beause of lack of happy path test case existing. Added a happy path test case

acolytec3:
> Definitely. Btw if you know how to enable pushes let me know I couldn't figure it out clicking around github for a bit

I thought there was a way to do so in the PR itself.  Something like a setting to "allow commits from maintainers" or something.

holgerd77:
@roninjin10 have invited you to the monorepo, feel free to accept and directly push PRs on the repo if this eases the process! 🙏

PR #3167, opened 2023-11-27
Experiments with `EtherStateManager` blockchain supplemental work

acolytec3:

PR #3172, opened 2023-11-30
Add runCallOpts and runCodeOpts to `evm` exports

acolytec3:
While working on #3167, I noticed that we do not export `EVMRunCallOpts` and `EVMRunCodeOpts` from the `evm` package even though these are the type definitions for the `evm.runCall` and `evm.runCode` public methods of the EVM.  This fixes that.


jochem-brouwer (approved):
LGTM, thanks!

PR #3173, opened 2023-11-30
Replace `superagent` with direct RPC calls in RPC tests

acolytec3:
Removes the `superagent` testing dependency and revises the client's `rpc` tests to make direct RPC calls using a `jayson` rpc client instead.  
- Remove `supertest` usage throughout `client` and replace with `jayson.httpClient` to query the RPC
- Removes a lot of unnecessary mocking/unmocking from RPC tests (left over from the `tape` test days when each test was run in the same environment so mocked/stubbed modules could be carried over from test to test - no longer an issue with `vitest`)
- Removes several now obsolete helpers/utility methods
- Updates args parsing logic in `cli.js` to allow for only the websocket RPC endpoint to be started


gabrocheleau:
Looks great and much more readable.

packages/client/test/rpc/rpc.spec.ts
```diff
@@ -7,53 +8,48 @@ import { METHOD_NOT_FOUND } from '../../src/rpc/error-code'
...
-      .post('/')
-      .send(req)
-      .expect(401)
-      .end((err: any) => {
-        closeRPC(server)
-        assert.notOk(err)
-      })
+    try {
+      await rpc.request(req, [])
+      assert.notOk(undefined, 'should error when request not authenticated by JWT')
```
holgerd77, line 32: Short nit: I think this can be shortened to `assert.fail('should error...')`

acolytec3: Good catch.

holgerd77:
Updated this via UI

scorbajio (commented):
Great work with this one! More readable tests and almost 1000 less lines of code.

packages/client/test/rpc/engine/getPayloadBodiesByRangeV1.spec.ts
```diff
@@ -190,18 +173,14 @@ describe(method, () => {
...
-    const expectRes = (res: any) => {
-      assert.equal(
-        res.body.result[0].withdrawals,
-        null,
-        'withdrawals field is null for pre-shanghai blocks'
-      )
-    }
+    const res = await rpc.request(method, ['0x1', '0x4'])
+
+    assert.equal(
```
gabrocheleau, line 178:
```suggestion
    assert.isNull(
```

packages/client/test/rpc/eth/chainId.spec.ts
```diff
@@ -1,60 +1,37 @@
...
-    const expectRes = (res: any) => {
-      const msg = 'chainId should be a string'
-      assert.equal(typeof res.body.result, 'string', msg)
-    }
-    await baseRequest(server, req, 200, expectRes)
+    const { rpc } = baseSetup()
+
+    const res = await rpc.request(method, [])
+    const msg = 'chainId should be a string'
+    assert.equal(typeof res.result, 'string', msg)
```
gabrocheleau, line 14:
```suggestion
    assert.equal(typeof res.result, 'string', 'chainId should be a string')
```

packages/client/test/rpc/eth/chainId.spec.ts
```diff
@@ -1,60 +1,37 @@
...
-      const msg = 'should return chainId 1'
-      assert.equal(res.body.result, '0x1', msg)
-    }
-    await baseRequest(server, req, 200, expectRes)
+    const { rpc } = baseSetup()
+
+    const res = await rpc.request(method, [])
+
+    const msg = 'should return chainId 1'
+    assert.equal(res.result, '0x1', msg)
```
gabrocheleau, line 23:
```suggestion
    assert.equal(res.result, '0x1', 'should return chainId 1')
```

packages/client/test/rpc/eth/chainId.spec.ts
```diff
@@ -1,60 +1,37 @@
...
-  })
+    const rpc = getRpcClient(startRPC(manager.getMethods()))
+
+    const res = await rpc.request(method, [])
 
-  it('reset TD', () => {
-    BlockHeader.prototype['_consensusFormatValidation'] = originalValidate
-    td.reset()
+    const msg = 'should return chainId 5'
+    assert.equal(res.result, '0x5', msg)
```
gabrocheleau, line 35:
```suggestion
    assert.equal(res.result, '0x5', 'should return chainId 5')
```

packages/client/test/rpc/eth/getBalance.spec.ts
```diff
@@ -59,53 +56,48 @@ describe(
...
-      req = params(method, [address.toString(), 'earliest'])
-      expectRes = (res: any) => {
-        const msg = "should return the correct balance with 'earliest'"
-        assert.equal(res.body.result, bigIntToHex(genesisBalance), msg)
-      }
-      await baseRequest(server, req, 200, expectRes, false)
+      res = await rpc.request(method, [address.toString(), 'earliest'])
+
+      msg = "should return the correct balance with 'earliest'"
+      assert.equal(res.result, bigIntToHex(genesisBalance), msg)
```
gabrocheleau, line 68: minor nit, but unless the msgs are reused, I'd personally just in-line them rather than declaring them as a separate variable.

packages/client/test/rpc/eth/getBlockByNumber.spec.ts
```diff
@@ -68,142 +67,122 @@ const method = 'eth_getBlockByNumber'
...
-      assert.equal(res.body.result.number, '0x0', msg)
-      msg = 'should return only the hashes of the transactions'
-      assert.equal(typeof res.body.result.transactions[0], 'string', msg)
-    }
-    await baseRequest(server, req, 200, expectRes)
+    const rpc = getRpcClient(startRPC(manager.getMethods()))
+
+    const res = await rpc.request(method, ['0x0', false])
+
+    let msg = 'should return a valid block'
```
gabrocheleau, line 84: Same comment about msg, applies to most other files in this directory.

packages/client/test/rpc/eth/getBlockByNumber.spec.ts
```diff
@@ -68,142 +67,122 @@ const method = 'eth_getBlockByNumber'
...
-    const server = startRPC(manager.getMethods())
-    const req = params(method, ['latest', true])
+    const rpc = getRpcClient(startRPC(manager.getMethods()))
+    const res = await rpc.request(method, ['latest', true])
 
-    const expectRes = (res: any) => {
-      assert.equal(typeof res.body.result.transactions[0], 'object', 'should include tx objects')
-    }
-    await baseRequest(server, req, 200, expectRes)
+    assert.equal(typeof res.result.transactions[0], 'object', 'should include tx objects')
```
gabrocheleau, line 164: Could we check for something else (e.g. a property, or length of transactions) instead of typeof object? One gotcha with a `typeof myVar === 'object'` is that `typeof null === 'object'` equals true .

packages/client/test/rpc/validation.spec.ts
```diff
@@ -15,33 +17,22 @@ describe(prefix, () => {
...
-      jsonrpc: '2.0',
-      method: mockMethodName,
-      id: 1,
-    }
-    const expectRes = (res: any) => {
-      assert.equal(res.body.error, undefined, 'should not return an error object')
-    }
-    await baseRequest(server, req, 200, expectRes)
+    const res = await rpc.request(mockMethodName, [])
+    assert.equal(res.error, undefined, 'should not return an error object')
```
gabrocheleau, line 23:
```suggestion
    assert.isUndefined(res.error, 'should not return an error object')
```

packages/client/test/rpc/web3/clientVersion.spec.ts
```diff
@@ -1,36 +1,34 @@
...
+    const [actualClientTitle, actualPackageVersion, actualPlatform, actualNodeVersion] =
+      result.split('/')
+    msg = 'client title should be correct'
+    assert.equal(actualClientTitle, expectedClientTitle, msg)
+    msg = 'package version should be correct'
+    assert.equal(actualPackageVersion, expectedPackageVersion, msg)
+    msg = 'platform should be correct'
+    assert.equal(actualPlatform, expectedPlatform, msg)
+    msg = 'Node.js version should be correct'
+    assert.equal(actualNodeVersion, expectedNodeVersion, msg)
```
gabrocheleau, line 32: Again I think the code would be cleaner if we didn't redeclare msg on every other line.

gabrocheleau (commented):
Looks good to me, left some minor comments

packages/client/test/rpc/web3/clientVersion.spec.ts
acolytec3: Good point.  Will clean this up later today.  Thanks for the review!

packages/client/test/rpc/engine/getPayloadBodiesByRangeV1.spec.ts
```diff
@@ -175,7 +175,7 @@ describe(method, () => {
 
     const res = await rpc.request(method, ['0x1', '0x4'])
 
-    assert.equal(
+    assert.isNull(
       res.result[0].withdrawals,
       null,
```
gabrocheleau, line 180:
```suggestion
```

gabrocheleau (approved):
Looks good!

packages/client/test/rpc/eth/estimateGas.spec.ts
```diff
@@ -102,19 +101,19 @@ describe(
...
-      await baseRequest(server, reqWithoutBlockOpt, 200, expectRes, false)
+      assert.equal(
+        res.result,
+        '0x' + totalGasSpent.toString(16),
+        'should return the correct gas estimate'
+      )
 
+      // Test without blockopt as its optional and should default to latest
+      const res2 = await rpc.request(method, [{ ...estimateTxData, gas: estimateTxData.gasLimit }])
+      assert.equal(res2.result, '0x' + totalGasSpent.toString(16), msg)
```
gabrocheleau, line 116:
```suggestion
      assert.equal(res2.result, '0x' + totalGasSpent.toString(16), 'should return the correct gas estimate')
```

packages/client/test/rpc/eth/estimateGas.spec.ts
```diff
@@ -143,35 +142,34 @@ describe(
...
-      await baseRequest(server, EIP1559reqNoGas, 200, expect1559Res, false)
+      assert.equal(EIP1559reqNoGas.result, '0x' + totalGasSpent.toString(16), 'should return the correct gas estimate for EIP1559 tx')
 
       // Test legacy tx with London head block
-      const legacyTxNoGas = params(method, [
+      const legacyTxNoGas = await rpc.request(method, [
         { ...estimateTxData, maxFeePerGas: undefined, gasLimit: undefined },
       ])
-      await baseRequest(server, legacyTxNoGas, 200, expect1559Res)
+      assert.equal(legacyTxNoGas.result, '0x' + totalGasSpent.toString(16), msg)
```
gabrocheleau, line 164:
```suggestion
      assert.equal(legacyTxNoGas.result, '0x' + totalGasSpent.toString(16), 'should return the correct gas estimate for EIP1559 tx')
```

g11tech (commented):
looks good to me, nice cleanup , can be approved once CI passes

jochem-brouwer:
Have re-linted, should hopefully now pass.

PR #3174, opened 2023-12-02
feat: Add nx for monorepo management

roninjin10:
Putting up a pr instead of opening an issue because it better illustrates nx than talking hypothetically.

image

## Description

Nx provides better monorepo management.  Some features:

## Graph

Nx automatically generates the ethereumjs dependency graph. This is what it uses to figure out which order to run tasks in.

```bash
nx graph
```

image

## Task scheduling

Because nx knows the dependency graph it knows whenever you need to run dependant tasks. For example, if vm depends on evm package and you run `npm run nx build` in the evm package nx will double check that evm is built before building vm.

## Caching

Nx can cache build outputs based on source code and thus reuse the build outputs from the cache. I configured it to cache the dist dir when `build` is ran.

Note: I am being extremely conservitive in the cache here which means it will get a lot of false cache misses. It is possible to [configure nx more granularly](https://github.com/evmts/evmts-monorepo/blob/main/nx.json). For example don't rebuild types if dependency types didn't change. Or don't rebuild dist if only a spec.ts file changed. The current setup will rebuild when any files in any dependent package changes. Better to have false cache misses than false cache hits but this could be optimized in future.
image

Cache is stored in the `.nx` folder and can be disabled by adding `--skip-nx-cache` flag.

## Nice monorepo api

Targeting packages with nx is very easy. For example, I can run the tests in vm without cding to it

**Note this requires configuring an npm package to leave out the @ethereumjs-monorepo/. Something I haven't done yet in this pr**

```bash
# Can be ran from anywhere in the monorepo to build ethereumjs/vm
nx build vm
```

## Remote caching (optional)

Because ethereumjs is an open source monorepo it would qualify for free cloud support from nx. This would make it so permissioned maintainers and github actions could read/write from nx cache and even contributors would be able to read from the cache. This means the cache is shared among everybody using this repo and if you pulled master branch everything would build and test instantly.

## Fast CI (optional)

Nx has [really fast ci](https://nx.dev/ci/features/distribute-task-execution) with nx cloud. The remote caching makes best case times instant. The scheduling and parallization make worst case times (where cache misses) very good too.

image

## Vscode plugin

There is a nice vscode plugin for using nx for those who use vscode. I used it some before I switch to nvim and it was nice to be able to run nx commands from vscode command pilot as well as the gui

## Alternatives

[Turborepo](https://turbo.build/) is a great alternative. It works almost the same as what I set up in this pr in terms of configuration and how it works. Turborepo is way less mature and feature rich than nx though and the free open source cloud offering nx has is super useful to take advantage of. 

## Npm run all command

I addded an npm run all command that runs test build and lint in every package. This is very nice to have and use prepush to make sure pr passes checks

image


nx.json
```diff
@@ -0,0 +1,119 @@
+{
+  "targetDefaults": {
```
roninjin10, line 2: These targetDefaults are the targets one can target with `nx run-many` or `nx affected`

nx.json
```diff
@@ -0,0 +1,119 @@
...
+    "build": {
+      "dependsOn": ["^build"],
```
roninjin10, line 4: This ^build means this command depends on the build of all dependent packages

nx.json
```diff
@@ -0,0 +1,119 @@
...
+      "outputs": ["{projectRoot}/dist"],
+      "cache": true
```
roninjin10, line 6: This target is configured to cache the output of dist

nx.json
```diff
@@ -0,0 +1,119 @@
...
+    "lint:diff": {
+      "cache": true
+    },
+    "lint:fix": {
+      "cache": true
+    }
+  },
+  "affected": {
+    "defaultBase": "master"
+  },
```
roninjin10, line 117: nx affected will run based on what has changed since the latest master branch

nx.json
```diff
@@ -0,0 +1,119 @@
...
+  "nxCloudAccessToken": "NDU2NDU3NmMtZjBiYy00MzBiLTkwMjEtNzczOGZhY2NiMDUzfHJlYWQtd3JpdGU="
```
roninjin10, line 118: This nx cloud token is doing nothing atm until someone claims it. It's meant to be public and a readonly token. Note: ethereumjs as an OSS project qualifies for free nx cloud access.

It doesn't require a remote cache and is optional. Without it you still benifit from your local cache

package.json
```diff
@@ -5,16 +5,20 @@
     "packages/*"
   ],
   "scripts": {
+    "build": "nx run-many --target=build",
```
roninjin10, line 8: this will run build in parallel and in the correct order for every package in the repo

package.json
```diff
@@ -5,16 +5,20 @@
...
     "e2e:inject": "node ./scripts/e2e-inject-resolutions.js",
     "e2e:publish": "./scripts/e2e-publish.sh",
     "e2e:resolutions": "node ./scripts/e2e-resolutions.js",
     "examples": "npm run examples --workspaces --if-present",
+    "lint": "nx run-many --target=lint",
     "preinstall": "npm run checkNpmVersion",
     "postinstall": "npm run build --workspaces",
     "prepare": "git config --local core.hooksPath .githooks",
     "prettier": "prettier --write \"./**/*.{js,json,md,ts,yml}\"",
+    "test": "nx run-many --target=test",
```
roninjin10, line 20: Since I configured test to depend on build it will build any dependent packages and also run tests in parallel. By default it does 3 threads

nx.json
```diff
@@ -0,0 +1,119 @@
+{
```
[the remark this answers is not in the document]
roninjin10: Some more advanced configuration can be seen in EVMts repo https://github.com/evmts/evmts-monorepo/blob/main/nx.json

roninjin10:
@jochem-brouwer I don't think we need to get rid of the tests. It looks like [genesisStateRoot function](https://github.com/ethereumjs/ethereumjs-monorepo/blob/f9525af55b4c2068a003e8035184913c0adcd980/packages/trie/src/util/genesisState.ts#L12) function being moved to the genesis package fixes the circular dependency

roninjin10:
Fixed circular dependency is fixed and nx looks to be working great now. I added a `npm run all` command to root of repo that runs test lint and build on every package 
image

roninjin10:
Moving out of draft since everything but there is definitely more that could be done here depending on if we wanted to use nx more or keep it minimal as is

roninjin10:
Everything seems to be working great. `npm run all` will cache the correct stuff while running lint test and build on every package in parallel 
image

nx.json
roninjin10: Based on the comment about not wanting dependency I'm guessing that this repo would not want to take advantage of nx cloud and thus we would remove this configuration here
```suggestion
```

holgerd77:
Thanks for all the work being put into this! 🙏 ❤️

I guess we need some (more) time to get a bit "warm" here and form a picture about the consequences of an integration, but I have the impression this looks pretty promising and likely would make (a lot of) sense to have such a tool integration, just linking the [Nx](https://nx.dev/) tool here for completion! 🤩

Yeah, also very much agree on the circular dependencies, we should have a closer look. Everytime in the past when we had circular dependencies (even in the tests) something turned out to be structurally wrong or suboptimal (as you mentioned as well).

nx.json
roninjin10: Want to remove this because some folks using ethereumjs monorepo won't want the network requests happening that nx cloud removes. Better to just use the local only cache

roninjin10 (commented):
Happy to break the circular dependency out into it's own pr

roninjin10:
Closed the issue but forgot to close the pr

PR #3179, opened 2023-12-06
client: kaustinen syncing

gabrocheleau:
This PR continues the work from #3139. 

In addition to other fixing issues that will arise during syncing.

To run the actual kaustinen2 network:
1. `DEBUG=ethjs,vm:*,evm:*,statemanager:* npm run client:start:ts -- --dataDir data/testing --network kaustinen --rpc --rpcEngine --rpcEngineAuth false`
2. clone https://github.com/ChainSafe/lodestar-quickstart , do vi kaustinen.vars and remove the checkpointSyncUrl  from lodestar args input and run `./setup.sh --dataDir k2data --network kaustinen --justCL` to run lodestar to feed ethereumjs blocks


packages/evm/src/opcodes/gas.ts
```diff
@@ -209,6 +250,16 @@ export const dynamicGasHandlers: Map<number, AsyncDynamicGasHandler | SyncDynami
         if (common.isActivatedEIP(2929) === true) {
           gas += accessStorageEIP2929(runState, keyBuf, false, common)
         }
+
+        if (common.isActivatedEIP(6800) === true) {
```
holgerd77, line 254: Short nit: for Boolean evaluations the explicit ˋ=== trueˋ is not necessary and you are in fact also using both versions in the code atm. Can you remove for all occurrences?

packages/vm/src/runBlock.ts
```diff
@@ -558,16 +559,30 @@ export function calculateMinerReward(minerReward: bigint, ommersNum: number): bi
 }
 
 export async function rewardAccount(
+  common: Common,
```
holgerd77, line 562: Short nit: can we make this „a bit less breaking“ (function might’ve in direct usage) and put common last and make it optional (and the do „if common and common.isActivated…“ checks)?

Then we can mention this in release not docs (that people should have a look if they use this method directly and optimally switch to pass a common there) and code is not directly breaking.

packages/evm/src/opcodes/gas.ts
jochem-brouwer: We should check if we can add this to the linter rules :+1: :smile:

gabrocheleau:
As discussed with Gajinder, we believe that this PR is in a state where it can be reviewed and merged in order to pave the ground for further work that will build on top of this.

Concerning the Kaustinen2 network syncing, the testing efforts have uncovered some bugs with the network (notably thanks to EthereumJS syncing, kudos to @g11tech for finding and troubleshooting that) that will lead to a relaunch, namely Kaustinen3. It is therefore not worth at this stage making additional efforts to sync Kaustinen2, as we would have to work around known issues that will be fixed in future iterations of the network.

I will fix the failing tests, do a first round of review, and then this could be up for a more general review. 🙏

packages/vm/src/runBlock.ts
gabrocheleau: Just seeing this now. I agree. I will put `common` last and make it optional.

packages/evm/src/evm.ts
```diff
@@ -268,7 +328,7 @@ export class EVM implements EVMInterface {
       return {
         execResult: {
           gasRefund: message.gasRefund,
-          executionGasUsed: BIGINT_0,
+          executionGasUsed: message.gasLimit - gasLimit,
```
holgerd77, line 331: Just reviewing the PR a bit in the light of non-verkle (side) changes to see what the overall effects on our libraries are/would be.

What is this change? (and also some additional gas related changes below)

Is this a general bug fix for the EVM? To now pass some dynamic value instead of `BIGINT_0`?

packages/evm/src/evm.ts
```diff
@@ -574,7 +721,9 @@ export class EVM implements EVMInterface {
       codeAddress: message.codeAddress,
       gasRefund: message.gasRefund,
       containerCode: message.containerCode,
+      chargeCodeAccesses: message.chargeCodeAccesses,
       blobVersionedHashes: message.blobVersionedHashes ?? [],
+      accessWitness: message.accessWitness,
```
holgerd77, line 726: Can we make these (so: `chargeCodeAccesses` and `accessWitness` for `env`, and `accessWitness` for `Message` below) optional for now, and then do `!` within `if (Verkle EIP activated) { ... }` blocks ?

I know this is rather internal, but people are muddling a lot with and go deep into our EVM code, and this would have a higher chance to not break anything for them.

holgerd77: Update: ah, seeing that for Message it (at least partly? not sure about `chargeCodeAccesses`) seems to be. Still for `env` would be also nice if easily doable.

packages/statemanager/src/accessWitness.ts
```diff
@@ -0,0 +1,433 @@
...
+// Since stem is pedersen hashed, it is useful to maintain the reverse relationship
+type StemMeta = { address: Address; treeIndex: number | bigint }
+type RawAccessedState = {
+  address: Address
+  treeIndex: number | bigint
+  chunkIndex: number
+  chunkKey: PrefixedHexString
+}
+
+export enum AccessedStateType {
```
holgerd77, line 55: Can all exported members of this file please get a code code that these code parts are part of verkle implementation and that this is experimental and code/API can change (or something like that)?

packages/verkle/src/util/crypto.ts
```diff
@@ -39,13 +40,17 @@ export function verifyUpdate(
  * @param treeIndex The index of the tree to generate the key for. Defaults to 0.
  * @return The 31-bytes verkle tree stem as a Uint8Array.
  */
-export function getStem(address: Address, treeIndex: number = 0): Uint8Array {
+export function getStem(address: Address, treeIndex: number | bigint = 0): Uint8Array {
```
holgerd77, line 43: Can all exported members/functions an experimental note here as well?

packages/statemanager/src/accessWitness.ts
holgerd77: (maybe a generic two-liner like

> Experimental. This code is part of the Verkle [EIP-23423](link) implementation and the exposed API will change over time. Do not use in production.

or something like that to be repeated everywhere we need this, than this can be easily removed again by search-and-replace later on)

packages/statemanager/src/accessWitness.ts
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    debug(
+      `touchAddressAndChargeGas=${gas} address=${address} treeIndex=${treeIndex} subIndex=${subIndex}`
+    )
+
+    return gas
+  }
+
+  touchAddress(
```
holgerd77, line 249: Just to note:

there is an insane amount of new "Codecov - not covered by tests" messages in this PR.

Know about the experimental state of all this, maybe it would nevertheless make sense to put some emphasis on this relatively soon, to have some internal unit tests to get all this better covered (potentially also a good learning task for others from the team to contribute).

holgerd77 (commented):
Some additional notes from a library-compatibility point of view. 🙂

packages/evm/src/evm.ts
g11tech: this is because even in this early return case, access gas could be consumed (which is dedcted from this function scoped `gasLimit`, so the diff is the whats used)

packages/statemanager/src/accessWitness.ts
jochem-brouwer: I will take this on to get a bit more deeper in this code! :+1:

packages/vm/src/runTx.ts
```diff
@@ -207,6 +210,20 @@ export async function runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
...
+      throw Error(`StatelessVerkleStateManager needed for execution of verkle blocks`)
+    }
+    stateAccesses = (this.stateManager as StatelessVerkleStateManager).accessWitness
+  } else {
+    // assign an empty witness to prevent unnecessary code spagetti of null/undefined checks while usage
+    // but its totally possible that merkle can also accumulate witnesses if such feature was to be made
+    // available there
+    stateAccesses = new AccessWitness()
+  }
+  let txAccesses = stateAccesses?.shallowCopy()
```
jochem-brouwer, line 225: Currently this code block either points to the `this.stateManager` as VerkleStateManager and then shallow copy it (this is the same as a `new AccessWitness`) or it creates a new witness and then copies it (so does the same thing twice) :thinking:

packages/evm/src/interpreter.ts
```diff
@@ -234,10 +238,30 @@ export class Interpreter {
     let doJumpAnalysis = true
     // Iterate through the given ops until something breaks or we hit STOP
     while (this._runState.programCounter < this._runState.code.length) {
+      const programCounter = this._runState.programCounter
       let opCode: number
       let opCodeObj: OpcodeMapEntry
       if (doJumpAnalysis) {
-        opCode = this._runState.code[this._runState.programCounter]
+        opCode = this._runState.code[programCounter]
```
jochem-brouwer, line 245: This is inside `doJumpAnalysis`, is this correct? So this 0xfe check is only done if we have not yet done the jumpdest analysis :thinking:

jochem-brouwer: I think this is problematic if we have already done the jumpdest analysis (so, after we have done a JUMP opcode). Then `doJumpAnalysis` is `false`, so we dont do the invalid check. However, if we then hit an invalid opcode we do not know if this was the real invalid opcode or if we dont have this chunk.

(Also, this assumes that if `this._runState.code[programCounter]` returns 0xfe if we are outside the code chunk - is this correct and is this done somewhere? I guess I'm missing something at this point :) )

packages/vm/src/runTx.ts
```diff
@@ -207,6 +210,20 @@ export async function runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
...
   const state = this.stateManager
 
+  let stateAccesses
+  if (this.common.isActivatedEIP(6800)) {
+    if (!(this.stateManager instanceof StatelessVerkleStateManager)) {
+      throw Error(`StatelessVerkleStateManager needed for execution of verkle blocks`)
+    }
+    stateAccesses = (this.stateManager as StatelessVerkleStateManager).accessWitness
+  } else {
+    // assign an empty witness to prevent unnecessary code spagetti of null/undefined checks while usage
```
jochem-brouwer, line 220: So currently in this code, in `message.actionWitness` is always defined. For testing in EVM it is handy to also make this the default `new ActionWitness` in `message.ts` - is this ok?

packages/evm/src/evm.ts
g11tech: i am confused, both are optional in env and message

packages/evm/src/interpreter.ts
g11tech: ahh yes this needs to be outside

packages/vm/src/runTx.ts
g11tech: yes i don't see the issue but probably this could be removed, lemme try removing it

packages/vm/src/runTx.ts
g11tech: yes because this witness is just used to establish an upfront fee and then thrown away (rewrttiten by a new shallowCopy)

packages/evm/src/interpreter.ts
g11tech: > (Also, this assumes that if this._runState.code[programCounter] returns 0xfe if we are outside the code chunk - is this correct and is this done somewhere? I guess I'm missing something at this point :)

yes code array is intialized with `0xfe` and then code chunks provided are filled in

g11tech (approved):
lgtm to be ready for kaustinen 3

holgerd77 (approved):
Ok, I will now take this in, under the assumption that we will/want to definitely write and define more test cases on this. Also - if there is an eventual performance degradation in the EVM (4% would be a lot!) that we further investigate here!

LGTM

PR #3184, opened 2023-12-08
Add test for ecrecover precompile

scorbajio:
This change aims to increase test coverage for the evm package `01-ecrecover` precompile.

### DIFF
--- a/packages/evm/test/precompiles/01-ecrecover.spec.ts
+++ b/packages/evm/test/precompiles/01-ecrecover.spec.ts
@@ -14,6 +14,8 @@ const address = '0xa6fb229e9b0a4e4ef52ea6991adcfc59207c7711'
 
 describe('Precompiles: ECRECOVER', () => {
   it('ECRECOVER', async () => {
+    // Test reference: https://github.com/ethereum/go-ethereum/issues/3731#issuecomment-293866868
+
     const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Petersburg })
     const evm = new EVM({
       common,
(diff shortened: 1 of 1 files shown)


packages/evm/test/precompiles/01-ecrecover.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Petersburg })
+    const evm = new EVM({
+      common,
+    })
+    const addressStr = '0000000000000000000000000000000000000001'
+    const ECRECOVER = getActivePrecompiles(common).get(addressStr)!
+
+    const prefixedMessage = keccak256(hexToBytes(`0x${prefix}${_hash}`))
+    const data = hexToBytes(`0x${prefixedMessage}${_v}${_r}${_s}`)
+    const result = await ECRECOVER({
```
acolytec3, line 26: This doesn't fix the test per se but at a minimum you need to replace above with the below.
```suggestion
    const data = hexToBytes(`${bytesToHex(prefixedMessage)}${_v}${_r}${_s}`)
```

jochem-brouwer (approved):

holgerd77:
Small process remark here: it's always good (also as a reviewer) to finally move the `PR state: WIP` label to `PR state: needs review` (even if the review is already done). For me this is always a strong and helpful indicitator (along releases or in between clean-ups) that things have been worked through "till the end". 🙂

holgerd77:
(also in case you already wondered why I occasionally update this even after PR merge. Alternatively we could also add an additional label "PR state: done" if you'd prefer, not sure if it would be over-engineered though, so for me personally having everything on "needs review" is enough to keep oversight 🙂)

PR #3186, opened 2023-12-10
Trie/StateManager: create partial state tries / state managers from proofs

### LINKED ISSUE #3178: Reconstructing a partial state `trie`
One question that has been raised in our Discord server recently was how to copy some or all of the state stored in the state manager's `trie` to another instance of the state manager.  This is different from the `shallowCopy` our API currently supports in that the way we would copy this state is to provide the EIP-1186 proofs provided by our `statemanager.getProof` to the State Manager and it would take the necessary steps to put these in the trie.  To this end, it would be nice to add a `fromProof` method on the `stateManager` similar to the `trie.fromProof` method that accepts an EIP-1186 proof and then inserts the account data into the account trie and constructs a storage trie from any storage proofs.  The gotcha here is the storage trie needs to be constructed by storing the nodes of the merkle storage proof directly in the DB as key/value pairs rather than using `trie.put` since our `statemanager` instantiates its underlying trie using hashed keys.

jochem-brouwer:
Closes #3178 

This PR:

- Changes Trie fromProof to updateTrieFromProof and adds a static constructor to do this
- Adds a way to verify that the provided ways are on the correct trie root (by default this is not checked)
- Adds feature to StateManager to import proof(s) from its `getProof` format (together with trie root verification)
- Fixes a bug where `getProof` reports incomplete proofs if a storage item is added but not flushed


packages/statemanager/test/proofStateManager.spec.ts
```diff
@@ -132,7 +147,6 @@ describe('ProofStateManager', () => {
       await trie._db.put(key, bufferData)
     }
     trie.root(stateRoot!)
-    await stateManager.putAccount(address, new Account())
```
jochem-brouwer, line 135: This had to be deleted because this actually creates an empty account in the trie (thus changing the state root) - this test has been wrong the entire time, it seems :thinking:

packages/statemanager/test/proofStateManager.spec.ts
```diff
@@ -28,6 +29,20 @@ describe('ProofStateManager', () => {
     assert.equal(proof.nonce, '0x0', 'Nonce is in quantity-encoded RPC representation')
   })
 
+  it(`should correctly return the right storage root / account root`, async () => {
```
jochem-brouwer, line 32: This tests that the trie is flushed in`getProof`. If the trie is not flushed, then the proof will report storage values but it will not report the right storage hash (storage root).

packages/statemanager/src/stateManager.ts
```diff
@@ -748,6 +749,91 @@ export class DefaultStateManager implements EVMStateManagerInterface {
     return returnValue
   }
 
+  /**
+   * Create a StateManager and initialize this with proof(s) gotten previously from getProof
+   * This generates a (partial) StateManager where one can retrieve all items from the proof
+   * @param proof Either a proof retrieved from `getProof`, or an array of those proofs
+   * @param safe Wether or not to verify that the roots of the proof items match the reported roots
+   * @param opts DefaultStateManagerOpts
```
acolytec3, line 757: Shouldn't we rename this and expand the doc to something like: 

```suggestion
   * @param verifyRoot verify that all proof root nodes match statemanager's stateroot - should be set to `false` when constructing a state manager where the underlying trie has proof nodes from different state roots
```

packages/statemanager/src/stateManager.ts
```diff
@@ -748,6 +749,91 @@ export class DefaultStateManager implements EVMStateManagerInterface {
...
+        storageProof[i].proof.map((e) => hexToBytes(e)),
+        safe
+      )
+    }
+  }
+
+  /**
+   * Add proof(s) into an already existing trie
+   * @param proof The proof(s) retrieved from `getProof`
+   * @param safe Whether or not to verify if the reported roots match the current state root
```
acolytec3, line 816: Same comment as `fromProof`

packages/statemanager/src/stateManager.ts
```diff
@@ -748,6 +749,91 @@ export class DefaultStateManager implements EVMStateManagerInterface {
...
+   */
+  async addProofItems(proof: Proof | Proof[], safe: boolean = false) {
```
acolytec3, line 818: This name feels weird to me.  Can we call it something like `addProofData` or `addTrieData`?  That feels more accurate than `addProofItems`

packages/statemanager/src/stateManager.ts
```diff
@@ -748,6 +749,91 @@ export class DefaultStateManager implements EVMStateManagerInterface {
...
+      return DefaultStateManager.fromProof([proof])
+    }
+  }
+
+  /**
+   * Adds a storage proof to the state manager
+   * @param storageProof The storage proof
+   * @param storageHash The root hash of the storage trie
+   * @param address The address
+   * @param safe Whether or not to verify if the reported roots match the current storage root
```
acolytec3, line 795: Would we ever not one to verify the storage root in this case?  I guess this makes sense in the archive node use case or where we want to verify the before and after state of some contract storage?

packages/trie/src/trie.ts
```diff
@@ -883,8 +883,28 @@ export class Trie {
...
+      this.root(rootKey)
+      await this.persistRoot()
+    }
+    return
+  }
+
+  /**
+   * Updates a trie from a proof
+   * @param proof The proof
+   * @param safe If `true`, verifies that the root key of the proof matches the trie root. Throws if this is not the case.
```
acolytec3, line 904:
```suggestion
   * @param shouldVerifyRoot If `true`, verifies that the root key of the proof matches the trie root. Throws if this is not the case.
```

packages/trie/src/trie.ts
```diff
@@ -896,18 +916,26 @@ export class Trie {
...
     })
 
-    if (
-      equalsBytes(this.root(), this.EMPTY_TRIE_ROOT) &&
-      opStack[0] !== undefined &&
-      opStack[0] !== null
-    ) {
-      this.DEBUG && this.debug(`Setting Trie root from Proof Node 0`, ['FROM_PROOF'])
-      this.root(opStack[0].key)
+    if (safe) {
```
acolytec3, line 919:
```suggestion
    if (shouldVerifyRoot) {
```

acolytec3 (changes requested):

scorbajio (commented):
Good work! A few comments and suggestions.

packages/trie/src/trie.ts
```diff
@@ -883,8 +883,28 @@ export class Trie {
...
+    return
+  }
+
+  /**
+   * Updates a trie from a proof
+   * @param proof The proof
+   * @param safe If `true`, verifies that the root key of the proof matches the trie root. Throws if this is not the case.
+   * @returns The root of the proof
+   */
+  async updateTrieFromProof(proof: Proof, safe: boolean = false) {
```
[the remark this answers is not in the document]
jochem-brouwer: I don't think this works, because that might alter the state root of the trie. You have to input raw nodes here. If you would put key/values into a partial trie, I think it either throws (but thats an edge case) or it alters the state root (and the latter is likely not what we want). (Have to verify this). For practical purposes it might work though to put the key/values directly, but I'm not sure if you want to change the root there as well.

(I have to verify these claims though :smile: :+1: )

acolytec3 (approved):
LGTM

holgerd77:
Great PR, congrats! 🎉🙂👍

PR #3187, opened 2023-12-11
Client: Async VM Initialization

holgerd77:
This moves to a proper async VM initialization in the client. Lot's of tests to fix and adopt, but already fixed a lot, so getting there.

Open to be continued during the day if someone feels like it, otherwise I'll continue tomorrow.

### DIFF
--- a/packages/client/test/miner/miner.spec.ts
+++ b/packages/client/test/miner/miner.spec.ts
@@ -72,6 +72,7 @@ describe('[Miner]', async () => {
         validateDifficulty: () => undefined,
       },
       validateHeader: () => {},
+      getIteratorHead: () => {},
       // eslint-disable-next-line no-invalid-this
       shallowCopy: () => this.blockchain,
       _init: async () => undefined,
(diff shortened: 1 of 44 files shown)


holgerd77:
@scorbajio @acolytec3 thanks a lot for the continued test fixes, that helps a lot, both practically and on my motivation to continue here (since it always a bit daunting to "cause" so many new test failures with the need to fix "just" for a somewhat simple cleanup)! 🤩

packages/client/test/miner/miner.spec.ts
```diff
@@ -259,8 +265,8 @@ describe('[Miner]', async () => {
     // no skipHardForkValidation
     const miner = new Miner({ config: goerliConfig, service })
     const { txPool } = service
+    await service.execution.setupMerkleVM()
```
holgerd77, line 268: There are several (most except for 1) cases where this also works with `await service.execution.open()`, just tested, will replace since we should rather use the "public" (internal) API for things like that for setup which then closer resembles the real situation.

holgerd77:
Tested mainnet sync on this up to a point with a proven block execution (so: a contract execution). This worked fine!

jochem-brouwer (approved):
LGTM (assuming CI passes)

acolytec3 (approved):
LGTM.  Tests should pass now.

PR #3188, opened 2023-12-12
Switch from `ts-node` to `tsx`

acolytec3:
Switches our monorepo's usage of `ts-node` over to `tsx` 

- Replace `ts-node` dep with `tsx` globally
- Replace all references to `ts-node` with `tsx`
- Updates several namespace imports to default imports to make `tsx` happy
- Update `lru-cache` dep in `devp2p` to get rid of `= require('lru-cache')` syntax
- Adds a new `install-browser-deps` npm script to the monorepo root to install pinned versions of the browser testing dependencies (pinned to versions known to work with our code, should be updated if #3191 is merged


jochem-brouwer:
There are some files added to this PR, I think they should be `git mv`d instead of manual renames? (I don't see the originals being deleted in this PR though so thats a bit weird)

packages/vm/test/tester/index.ts
```diff
@@ -45,7 +45,8 @@ import type { Common } from '@ethereumjs/common'
  * --profile                              If this flag is passed, the state/blockchain tests will profile
  */
 
-const argv = minimist(process.argv.slice(2))
+console.log(minimist) //@ts-ignore
```
jochem-brouwer, line 48: Do we want this `console.log`?

jochem-brouwer (commented):
One comment, and I have to test this locally but looks great to me in general.

packages/vm/test/retesteth/transition-child.ts
jochem-brouwer, line 1: Should this be renamed from `transition-child.cts` to here? (Same with transition-cluster)

jochem-brouwer:
I checked out retesteth, I think for now we should skip with the original instructions (use ts-node) or we should figure out how to use `spawn` which then spawns tsx processes :thinking:

acolytec3 (commented):
> I checked out retesteth, I think for now we should skip with the original instructions (use ts-node) or we should figure out how to use `spawn` which then spawns tsx processes :thinking:  

The transition cluster/child run just fine with `tsx` in my testing. If I do `curlie -v POST 127.0.0.1:3000` when the transition cluster is running, I can see logs from the `transition-child` thread running if I add console logs so it's not an issue of the ethjs side of the setup running. It seems like the retesth instructions are out of date though as I'm not able to get it to initiate a test.  I think I'm just not configuring my retesteth setup correctly.

packages/vm/test/retesteth/transition-child.ts
acolytec3: I don't think it matters.  I was just removing that as it's no longer imperative to have these file extensions with `tsx`.  That said, it shouldn't have any impact on it running since `tsx` handles CJS and ESM just fine.

packages/vm/test/tester/index.ts
acolytec3: Nope, removed.

packages/client/test/cli/cli.spec.ts
```diff
@@ -18,6 +18,7 @@ export function clientRunHelper(
   return new Promise((resolve) => {
     child.stdout.on('data', async (data) => {
       const message: string = data.toString()
+      console.log(message)
```
holgerd77, line 21: Intended?

packages/devp2p/package.json
```diff
@@ -62,7 +62,7 @@
     "@scure/base": "1.1.1",
     "debug": "^4.3.3",
     "ethereum-cryptography": "^2.1.2",
-    "lru-cache": "^7.18.3",
+    "lru-cache": "^10.0.0",
```
holgerd77, line 65: Without having checked if this is the case or not, it would be good if all out lru-cache deps would be Version aligned throughout the packages.

packages/vm/test/tester/index.ts
```diff
@@ -45,7 +45,8 @@ import type { Common } from '@ethereumjs/common'
  * --profile                              If this flag is passed, the state/blockchain tests will profile
  */
 
-const argv = minimist(process.argv.slice(2))
+//@ts-expect-error Typescript thinks there isn't a default export on minimist but there is
+const argv = minimist.default(process.argv.slice(2))
```
holgerd77, line 49: Side note: minimist is also causing problems with bun and we should consider to replace (should be simple).

holgerd77 (commented):
Some tests failing and two minor comments, otherwise looks good!

packages/vm/test/tester/index.ts
acolytec3: Okay, I think I might address this in a separate PR.  Not sure how much of a rewrite that will be to switch to yargs.

packages/devp2p/package.json
acolytec3: 10.0.0 is the correct version to match everything else in our usage so we should be good here.  Some of our dev deps use older versions but that shouldn't impact us.

acolytec3:
Looks like everything is finally passing now so should be ready for review.

packages/client/test/cli/cli.spec.ts
```diff
@@ -13,27 +13,23 @@ export function clientRunHelper(
...
-    child.on('close', (code) => {
-      if (typeof code === 'number' && code > 0) {
-        assert.fail(`child process exited with code ${code}`)
-      }
-    })
+    // child.on('close', (code) => {
+    //   if (typeof code === 'number' && code > 0) {
+    //     assert.fail(`child process exited with code ${code}`)
+    //   }
+    // })
```
jochem-brouwer, line 32: Should this be commented out?

packages/client/test/cli/cli.spec.ts
```diff
@@ -47,7 +43,7 @@ describe('[CLI]', () => {
           message.includes('network=sepolia chainId=11155111'),
           'client is using custom inputs for network and network ID'
         )
-        child.kill(9)
+        child.kill(15)
```
jochem-brouwer, line 46: Why does this exit code change?

packages/client/test/cli/cli.spec.ts
```diff
@@ -74,7 +70,7 @@ describe('[CLI]', () => {
         // if http endpoint startup message detected, call http endpoint with RPC method
         await wait(600)
         const client = Client.http({
-          port: 8562,
+          port: 8569,
```
jochem-brouwer, line 73: Out of interest why does the port have to change?

jochem-brouwer (commented):
Some tiny comments

packages/vm/test/tester/index.ts
holgerd77: Yeah, sure, definitely nothing for this PR and generally just a mention.

packages/devp2p/package.json
holgerd77: Ah ok, but it (generally) wouldn't hurt to align the dev dep versions too here (also just a mention)

packages/client/test/cli/cli.spec.ts
acolytec3: Probably didn't have to.  I was trying to figure out why the tests were hanging.  Will change back.

packages/client/test/cli/cli.spec.ts
acolytec3: Yes, should be removed entirely.  The child processes error out some times but it's related to how we shut things down, not the startup conditions that are being tested in these tests.

packages/client/test/cli/cli.spec.ts
acolytec3: I think I had a port collision where multiple child processes (i.e individual tests) were started by `vitest` simultaneously and they both tried to grab the same port.

holgerd77:
Cool, congrats! 🎉🙂

holgerd77:
Ugh, just skimmed through the changes, wasn't aware that this has gone *that* extensive. 🤯

acolytec3:
> Ugh, just skimmed through the changes, wasn't aware that this has gone _that_ extensive. 🤯

:grimacing: It was mostly just find and replace (except for the `lru-cache` update)

PR #3189, opened 2023-12-12
Evm precompile tests

scorbajio:
This change increases coverage of evm precompiles by adding tests for the ripemd160 and blake2f precompile contracts as well as updating the ecrecover precompile tests.


holgerd77:
Some general feedback here: yeah, I do think that it is really good if we have at least 1-2 simple test cases for each precompile in addition to the cross-client testing, also agree with @acolytec3 though that these particular precompiles have not such a high relevance (though they should nevertheless still not break consensus for sure), so we shouldn't take this too far.

Changing error messages in the EVM is always a breaking changes and I think it would not really be worth it here. So as a practical measure I would suggest that we delete the error message testing from the test cases and just concentrate on value testing here and simply see and take this as a small useful addition to the test setup (without discovering anything new).

jochem-brouwer:
I agree changing the error messages is not worth it. I would change the tests to check that there is an error and fail if it does not error out. For reference, in `ethereum/tests` (so our BlockchainTests and StateTests in VM) there is also an "expected error" string field, but we do not check this. It is also not worth it, since these errors are likely not checked among other clients as well: if there are 2 checks to be done (for instance, input length should be X, and gas should be higher than Y) then the order of these checks will thus change what error gets generated (if both checks would fail), but the end result should be the same: in most cases, all gas forwarded is used.

acolytec3 (approved):
LGTM

PR #3190, opened 2023-12-13
evm/ci: add evm state tests for better coverage

jochem-brouwer:


jochem-brouwer:
I experimented a bit to make vitest run our state test runner, but vitest does not seem to (currently) accept passing any `argv` to the test environment, so we can not directly run the state tests using vitest (runner will complain that no `state` or `blockchain` flag is provided). Vitest will also complain in the runner that there is no test suite.

I won't pursue this further, let's bump EVM coverage with a solid coverage/performance framework.

PR #3191, opened 2023-12-13
repo: update vitest

jochem-brouwer:
Vitest has published their first non-beta release a few days ago: https://www.npmjs.com/package/vitest/v/1.0.0

This PR updates vitest. It seems like the browser tests hang :thinking:

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,8 +23,8 @@
     "@types/tape": "4.13.2",
     "@typescript-eslint/eslint-plugin": "5.33.1",
     "@typescript-eslint/parser": "5.33.1",
-    "@vitest/coverage-v8": "0.32.2",
-    "@vitest/ui": "0.32.2",
+    "@vitest/coverage-v8": "1.0.4",
+    "@vitest/ui": "1.0.4",
     "c8": "7.12.0",
     "eslint": "8.45.0",
     "eslint-config-prettier": "8.8.0",
@@ -48,12 +48,12 @@
     "typescript": "4.7.4",
     "vite-plugin-top-level-await": "^1.3.1",
     "vite-plugin-wasm": "^3.2.2",
-    "vitest": "0.32.2"
+    "vitest": "1.0.4"
   },
   "peerDependencies": {
     "playwright": "^1.35.1",
     "webdriverio": "^8.10.7",
-    "@vitest/browser": "0.32.2"
+    "@vitest/browser": "1.0.4"
   },
   "peerDependenciesMeta": {
     "playwright": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,8 +15,8 @@
         "@types/tape": "4.13.2",
         "@typescript-eslint/eslint-plugin": "5.33.1",
         "@typescript-eslint/parser": "5.33.1",
-        "@vitest/coverage-v8": "0.32.2",
-        "@vitest/ui": "0.32.2",
+        "@vitest/coverage-v8": "1.0.4",
+        "@vitest/ui": "1.0.4",
         "c8": "7.12.0",
         "eslint": "8.45.0",
         "eslint-config-prettier": "8.8.0",
@@ -40,14 +40,14 @@
         "typescript": "4.7.4",
         "vite-plugin-top-level-await": "^1.3.1",
         "vite-plugin-wasm": "^3.2.2",
-        "vitest": "0.32.2"
+        "vitest": "1.0.4"
       },
       "engines": {
         "node": ">=18",
         "npm": ">=7"
       },
       "peerDependencies": {
-        "@vitest/browser": "0.32.2",
+        "@vitest/browser": "1.0.4",
         "playwright": "^1.35.1",
         "webdriverio": "^8.10.7"
       },
@@ -307,9 +307,9 @@
       }
     },
     "node_modules/@babel/helper-string-parser": {
-      "version": "7.22.5",
-      "resolved": "https://registry.npmjs.org/@babel/helper-string-parser/-/helper-string-parser-7.22.5.tgz",
-      "integrity": "sha512-mM4COjgZox8U+JcXQwPijIZLElkgEpO5rsERVDJTc2qfCDfERyob6k5WegS14SX18IIjv+XD+GrqNumY5JRCDw==",
+      "version": "7.23.4",
+      "resolved": "https://registry.npmjs.org/@babel/helper-string-parser/-/helper-string-parser-7.23.4.tgz",
+      "integrity": "sha512-803gmbQdqwdf4olxrX4AJyFBV/RTr3rSmOj0rKwesmzlfhYNDEs+/iOcznzpNWlJlIlTJC2QfPFcHB6DlzdVLQ==",
       "dev": true,
       "engines": {
         "node": ">=6.9.0"
@@ -433,9 +433,9 @@
       }
     },
     "node_modules/@babel/parser": {
-      "version": "7.23.0",
-      "resolved": "https://registry.npmjs.org/@babel/parser/-/parser-7.23.0.tgz",
-      "integrity": "sha512-vvPKKdMemU85V9WE/l5wZEmImpCtLqbnTvqDS2U1fJ96KrxoW7KrXhNsNCblQlg8Ck4b85yxdTyelsMUgFUXiw==",
+      "version": "7.23.6",
+      "resolved": "https://registry.npmjs.org/@babel/parser/-/parser-7.23.6.tgz",
+      "integrity": "sha512-Z2uID7YJ7oNvAI20O9X0bblw7Qqs8Q2hFy0R9tAfnfLkp5MW0UH9eUvnDSnFwKZ0AvgS1ucqR4KzvVHgnke1VQ==",
       "dev": true,
       "bin": {
         "parser": "bin/babel-parser.js"
@@ -669,12 +669,12 @@
       }
     },
     "node_modules/@babel/types": {
-      "version": "7.23.0",
-      "resolved": "https://registry.npmjs.org/@babel/types/-/types-7.23.0.tgz",
-      "integrity": "sha512-0oIyUfKoI3mSqMvsxBdclDwxXKXAUA8v/apZbc+iSyARYou1o8ZGDxbUYyLFoW2arqS2jDGqJuZvv1d/io1axg==",
+      "version": "7.23.6",
+      "resolved": "https://registry.npmjs.org/@babel/types/-/types-7.23.6.tgz",
+      "integrity": "sha512-+uarb83brBzPKN38NX1MkB6vb6+mwvR6amUulqAE7ccQw1pEl+bCia9TbdG1lsnFP7lZySvUn37CHyXQdfTwzg==",
       "dev": true,
       "dependencies": {
-        "@babel/helper-string-parser": "^7.22.5",
+        "@babel/helper-string-parser": "^7.23.4",
         "@babel/helper-validator-identifier": "^7.22.20",
         "to-fast-properties": "^2.0.0"
       },
@@ -744,10 +744,12 @@
       "cpu": [
         "arm"
       ],
+      "dev": true,
       "optional": true,
       "os": [
         "android"
       ],
+      "peer": true,
       "engines": {
         "node": ">=12"
       }
@@ -759,10 +761,12 @@
       "cpu": [
         "arm64"
       ],
+      "dev": true,
       "optional": true,
       "os": [
         "android"
       ],
+      "peer": true,
       "engines": {
         "node": ">=12"
       }
@@ -774,10 +778,12 @@
       "cpu": [
         "x64"
       ],
+      "dev": true,
       "optional": true,
       "os": [
         "android"
       ],
+      "peer": true,
       "engines": {
         "node": ">=12"
       }
@@ -789,10 +795,12 @@
       "cpu": [
         "arm64"
       ],
+      "dev": true,
       "optional": true,
       "os": [
         "darwin"
       ],
+      "peer": true,
       "engines": {
         "node": ">=12"
       }
@@ -804,10 +812,12 @@
       "cpu": [
         "x64"
       ],
+      "dev": true,
       "optional": true,
       "os": [
         "darwin"
       ],
+      "peer": true,
       "engines": {
         "node": ">=12"
       }
@@ -819,10 +829,12 @@
       "cpu": [
         "arm64"
       ],
+      "dev": true,
       "optional": true,
       "os": [
         "freebsd"
       ],
+      "peer": true,
       "engines": {
         "node": ">=12"
       }
@@ -834,10 +846,12 @@
       "cpu": [
         "x64"
       ],
+      "dev": true,
       "optional": true,
       "os": [
         "freebsd"
       ],
+      "peer": true,
       "engines": {
         "node": ">=12"
       }
@@ -849,10 +863,12 @@
       "cpu": [
         "arm"
       ],
+      "dev": true,
       "optional": true,
       "os": [
         "linux"
       ],
+      "peer": true,
       "engines": {
         "node": ">=12"
       }
@@ -864,10 +880,12 @@
       "cpu": [
         "arm64"
       ],
+      "dev": true,
       "optional": true,
       "os": [
         "linux"
       ],
+      "peer": true,
       "engines": {
         "node": ">=12"
       }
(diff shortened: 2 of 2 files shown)


holgerd77:
What is the status here? @ScottyPoi can you give a summary what changes are necessary to get this working? Everything "expected" or some unexpected stuff here?

ScottyPoi:
@holgerd77 the main issue here in the browsers is with core node modules.  especially anywhere we use EventEmitters.

Using a plugin called `vite-plugin-node-polyfill` is the recommended solution, which works for tests that do not directly use an event emitter, but does not fix tests that use things like `Common` or `AsyncEventEmitter` which are extensions of the node EventEmitter class.

This is the current error from those tests:
```
SyntaxError: The requested module '.../node_modules/events/events.js?v=09391bae' does not provide an export named 'EventEmitter'
```

ScottyPoi:
:grimacing: 
been working on this all day and can't seem to solve the EventEmitter thing, with polyfills or otherwise.  I will poke at it some more tomorrow.

ScottyPoi:
the wallet test:node passes locally.  not sure why they timeout on here

ScottyPoi:
https://github.com/vitest-dev/vitest/releases

They released v1.2 today which includes some fixes to browser mode.
Going to upgrade with fingers crossed that something in there is helpful to me!

ScottyPoi:
Browser Tests Pass!  

Using this as the base config for browser mode handles most issues:
```
defineConfig({
  test: {
    browser: {
      enabled: true,
      name: 'chrome',
      headless: true,
    },
  },
  resolve: {
    alias: {
      events: 'eventemitter3',
    },
  },
  plugins: [
    nodePolyfills({
      include: ['util', 'fs', 'buffer'],
    }),
  ],
})
```

Additional configs in individual packages handle any specific extra requirements.

jochem-brouwer:
Wow the CI passes, super great! I will give this a review tomorrow!

ScottyPoi:
I'm trying to figure out why coverage went down on certain files, especially in `/client`.  I think it has to do with some `await import(...)` statements, or some of the `vi.mock` things that we do in some tests.

ScottyPoi:
Browser-tests passed in [dc8b4f6](https://github.com/ethereumjs/ethereumjs-monorepo/pull/3191/commits/dc8b4f644ccbdc10beba63513700739bc0801a05)

I don't know why they aren't passing now.  The test that fails on here passes locally

ScottyPoi:
Update: 
- The only blocker from this being fully passing is  `browser / block`, which fails on github, but passes locally.
- I started working out the bugs in the `browser / VM` tests. and have 22/28 test files passing.
  - VM tests are not a part of the github CI "test-all-browser" workflow.
  - I believe I have identified some bugs and false positives in the VM tests.  I'd have to dig a bit more to know whether they were valid tests before the vitest upgrade or not, so I don't know if those fixes fall in the scope of this PR, or whether to finish that work separately.  
  - For now I will continue working on it on this branch, but won't push those changes for now, because it will be a lot of additional commits

jochem-brouwer:
Ah this looks similar to what we had a few weeks back with pinning the vitest dependency. @acolytec3 could you maybe take a look here?

ScottyPoi:
> Ah this looks similar to what we had a few weeks back with pinning the vitest dependency. @acolytec3 could you maybe take a look here?

what do you mean by "pinning"?

acolytec3:
> > Ah this looks similar to what we had a few weeks back with pinning the vitest dependency. @acolytec3 could you maybe take a look here?
> 
> what do you mean by "pinning"?
> 

Just putting the dependency to a specific version in package.json rather than `^0.2.0`

jochem-brouwer:
Ah pinning might not have been the exact right term, I mean changes introduced in for instance this commit: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3188/commits/fe1a1dc110702486ba51024520bca6ed22569def (note: from the tsx PR)

packages/wallet/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": "../../config/tsconfig.json",
-  "include": ["src/**/*.ts", "test/**/*.ts"]
+  "include": ["src/**/*.ts", "test/**/*.ts"],
+  "compilerOptions": {
+    "moduleResolution": "Node16"
```
acolytec3, line 5: Should we add this elsewhere?

acolytec3 (dismissed):
This looks great and I can verify that everything is working correctly.  Left one comment on `tsconfig` (though I think it could be addressed in a future PR.

packages/wallet/tsconfig.json
ScottyPoi: I'm going to do some cleanup on this branch.  some of these config edits might be unnecessary.

acolytec3: Sounds good.

ScottyPoi:
Cool, thanks!  
I'm still bothered by the `codecov/project` stats going down.  Can make that a separate issue, though, i suppose.

ScottyPoi:
@jochem-brouwer 

Not sure why the codecov still fails, but all of the tests ran successfully on here again finally.

A lot of guesswork went into finding a configuration that allowed all of the browser tests to pass on here, even though they all passed locally about 50 commits back...  

Personally I'd prefer to merge this as is rather than make any more commits to this branch.
This was a weird one, with some unanswered questions that I want to tackle in a separate PR.

jochem-brouwer:
I will give this a review later today. Codecov failing is nothing new (but we should fix it anyways). However it seems to fail within 1 second now, so I'm not sure that's right :thinking:

packages/block/package.json
```diff
@@ -16,7 +16,7 @@
   },
   "license": "MPL-2.0",
   "author": "mjbecze (mb@ethdev.com)",
-  "type": "commonjs",
+  "type": "module",
```
acolytec3, line 19: Why is this changed here?  This property is set to `commonjs` in the other package level `package.json` files and overridden to `module` in the ESM build step.

.github/workflows/browser.yml
```diff
@@ -31,11 +31,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: 'npm'
 
-      - run: npm ci
-      - run: npm run install-browser-deps
-
-      #     Install playwright dependencies
-      - run: npx playwright install-deps
```
jochem-brouwer, line 38: Is this playwright thing removed?

package.json
```diff
@@ -16,17 +16,18 @@
...
   "devDependencies": {
     "@types/estree": "^1.0.1",
     "@types/node": "18.11.9",
     "@types/tape": "4.13.2",
     "@typescript-eslint/eslint-plugin": "5.33.1",
     "@typescript-eslint/parser": "5.33.1",
-    "@vitest/coverage-v8": "0.32.2",
-    "@vitest/ui": "0.32.2",
-    "c8": "7.12.0",
+    "@vitest/browser": "^1.2.0",
```
jochem-brouwer, line 27: Should this be the caret version of 1.2.0 or `1.2.0`? 
```suggestion
    "@vitest/browser": "1.2.0",
```

packages/block/examples/4844.ts
```diff
@@ -2,11 +2,14 @@ import { Common, Chain, Hardfork } from '@ethereumjs/common'
 import { Block } from '@ethereumjs/block'
 import { BlobEIP4844Transaction } from '@ethereumjs/tx'
 import { Address, initKZG } from '@ethereumjs/util'
-import * as kzg from 'c-kzg'
+import kzg from 'c-kzg'
 import { randomBytes } from 'crypto'
+import path from 'path'
+import { fileURLToPath } from 'url'
 
+const __dirname = path.dirname(fileURLToPath(import.meta.url)).slice(0, -14)
```
jochem-brouwer, line 10: Why should this be sliced?

packages/block/test/eip4844block.spec.ts
```diff
@@ -20,92 +19,91 @@ import type { TypedTransaction } from '@ethereumjs/tx'
...
   hardfork: Hardfork.Cancun,
 })
 const blobGasPerBlob = common.param('gasConfig', 'blobGasPerBlob')
 
-describe('EIP4844 header tests', () => {
-  it('should work', () => {
-    if (isBrowser() === false) {
-      const earlyCommon = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Istanbul })
+console.log({
+  isBrowser: isBrowser(),
```
jochem-brouwer, line 29: Unremoved console.log

packages/evm/test/blobVersionedHashes.spec.ts
```diff
@@ -2,14 +2,15 @@ import { Common, Hardfork } from '@ethereumjs/common'
...
+import genesisJSON from '../../client/test/testdata/geth-genesis/eip4844.json'
+import { EVM } from '../src/evm.js'
 
 import type { EVMRunCallOpts } from '../src/types.js'
 
 describe('BLOBHASH / access blobVersionedHashes in calldata', () => {
   it('should work', async () => {
     // setup the evm
-    const genesisJSON = require('../../client/test/testdata/geth-genesis/eip4844.json')
+    // const genesisJSON = require('../../client/test/testdata/geth-genesis/eip4844.json')
```
jochem-brouwer, line 13: This comment can be removed

packages/evm/test/precompiles/05-modexp.spec.ts
jochem-brouwer, line 1: Nit: changes in this test file is fine, but once one of the tests fails it will be slightly hard to figure out which specific test failed.

jochem-brouwer (commented):
Some comments

packages/block/examples/4844.ts
ScottyPoi: `__dirname` as a global variable failed in the browser run, but worked in this way.  There's likely a more efficient way, but this slices off `/block/examples`  from the full file path, leaving you at `.../packages`

packages/block/package.json
ScottyPoi: Vitest will run, but give a warning about deprecating their commonjs support.
changing this in all the package.json files is a fix for this, but we can't do that to every package yet.

holgerd77: Yeah, agree with Andrew that we should handle such a change with very much care and give this some thought (likely Andrew best to discuss this), since it could affect our final builds + internal tooling.

packages/client/test/net/peerpool.spec.ts
```diff
@@ -65,24 +71,24 @@ describe('[PeerPool]', async () => {
   })
 
   it('should check contains', () => {
-    const peer = new Peer('abc')
+    const peer: any = new Peer('abc')
```
holgerd77, line 74: So I assume that these (and some other) any castings are necessary to get things working?

Please really double-check if we really need to have, adding `any` is always not so nice (but if needed, so be it).

packages/evm/test/precompiles/01-ecrecover.spec.ts
```diff
@@ -3,7 +3,8 @@ import { bytesToHex, bytesToUnprefixedHex, hexToBytes, utf8ToBytes } from '@ethe
 import { keccak256 } from 'ethereum-cryptography/keccak.js'
 import { assert, describe, it } from 'vitest'
 
-import { EVM, getActivePrecompiles } from '../../src/index.js'
+import { EVM } from '../../src/evm.js'
+import { getActivePrecompiles } from '../../src/precompiles/index.js'
```
holgerd77, line 7: All these import changes for `EVM` and `getActivePrecompiles` and some others are not so nice, also can't totally see why this is necessary or what this is doing. Is this really needed? We should really try to avoid. Not seeing why especially here the normal re-export we have everywhere from `index.js` should not also work?

holgerd77 (commented):
Thanks, looks already pretty good, also left a few comments!

Just a general note to not completely get this out of side: the *wider* goal should always be to have our libraries run easier/better in the browser and not primarily to adopt them to some tooling.

So if there are cases where vitest is only *not* working so well because we have suboptimal and not so browser friendly code in the *main* code base and this is somewhat easily fixable, this fixing should for sure get precedense before adopting to tooling.

Not sure if this might apply here in some cases.

It would also be good if this PR gets a short (!) list of the type of adoptions needed, to have some justification for this already pretty large change set.

packages/block/package.json
acolytec3: I don't think we should make this change anywhere for now.  I know our configs are not perfect but we should be able to set up a separate package.json just for vitest that specifies module if that's what we want to do longer term.

packages/evm/test/precompiles/01-ecrecover.spec.ts
acolytec3: @ScottyPoi, is this because vitest can't find the imports if you import from `../../index.js`?  I've seen that sort of thing before where a test that's in a `test/[subdirectory]` can't find imports that are in `../../src/[subdirectory]/someImport.js`

acolytec3: Ok, I was never able to resolve that module not found thing.  It's something in vitest and I always had to specify complete relative paths to modules that were in subdirectories.  Vitest has some trouble resolving re-exported modules where the importing module and the original export are in different subdirectories at different layers from the root.

I asked on the Vitest discord once and just got a shrug.

ScottyPoi:
Update:  I am rolling back changes where possible.

I suspect that there is an issue with the `test-all-browsers` CI action.  When the CI runs, the browser tests will fail on `block` with a `dynamic import` error.  Run it a second time, and `block` passes, but `blockchain` fails with the same error.  run it a third time and **everything passes**.   

I didn't discover this phenomenon until last night, so I am retracing my steps to undo changes that may have actually had no affect :roll_eyes:

packages/trie/test/trie/secure.spec.ts
```diff
@@ -6,8 +6,8 @@ import {
   hexToBytes,
   utf8ToBytes,
 } from '@ethereumjs/util'
-import { createHash } from 'crypto'
 import { keccak256 } from 'ethereum-cryptography/keccak.js'
+import { sha256 } from 'ethereum-cryptography/sha256.js'
```
ScottyPoi, line 10: Fixes error: 
`Error: Module "crypto" has been externalized for browser compatibility. Cannot access "crypto.createHash" in client code.  See https://vitejs.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility for more details.`

ScottyPoi: for some things, the `crypto` polyfill works, but `crypto.createHash` is not available.  
the `ethereum-cryptography` packages do work in the browser, though.

packages/evm/test/precompiles/01-ecrecover.spec.ts
ScottyPoi: I think I solved some of this by making changes to  `index.ts` files in `evm` itself.

The solution seems to be to make sure there is a consistency with `import` vs `import type` and `export` vs `export type` for each thing exported from an `index.ts` file.

packages/client/test/net/peerpool.spec.ts
ScottyPoi: The test will pass without the `as any`.
We get a ts error flag here, but it doesn't actually stop the test from running.

ScottyPoi:
Update:

Still around and around with the same issue today.  Not sure what else to try.

**The `codecov` failure --**

There is a mismatch between the number of uploaded coverage reports between the base and head being compared.  The documentation for Codecov mentions this as a reason for unexpected results, but offers no explicit solution.

**The `test-all-browsers` failure --**

The results are inconsistent.  All tests pass locally, with the same environment setup and everything.  Sometimes all the tests will run and pass.  Most of the time, either `block` or `blockchain` throws an import error which stops the tests, and often the same test will then pass on rerun.

I have tried various solutions, including editing the workflow file, but have not escaped this particular issue, nor figured out an explanation for why this might happen.

One possibility I have not explored is perhaps running **all** of the browser tests in one workflow causes an issue with resource constraints in github actions.  We could try breaking the browser tests up into more than one workflow.

holgerd77:
Maybe put this aside until our call on Wednesday and we can talk some things through there.

holgerd77:
Will close here, since this PR has gotten too extensive.

Scotty has already extracted some pre-work to #3266, we can continue/pick up this work here at some point in #3281.

Note that it will likely nevertheless be useful to come back to this PR for inspiration/guidance for changes, eventually it might also still make sense to cherry-pick selected commits.

PR #3192, opened 2023-12-14
Client/Monorepo: Use WASM Crypto (keccak256) for Hashing / Consistent Hash Function Overwrite

holgerd77:
This was actually the original triggger for #3187, hehe. 😁

So this PR tests the [@polkadot/wasm-crypto](https://github.com/polkadot-js/wasm/tree/master/packages/wasm-crypto) library as a replacement for the trie key hashing for the VM state manager in the client. Had this long on my plate to analyse this a bit better, since theoretically (re-)using a more native hash function should have a significant performance imact.

So I tested with the following early mainnet block (now starting directly with the block to avoid state caches):

```shell
npm run client:start -- --vmProfileBlocks --executeBlocks=227715
```

Block has 79 simple value transfer txs, so not much else is happening and mainly the sender and receiver state changes are written and come into play here.

Following results (with just the first two commits, so only exchanging for trie in execution):

Before (Entire Block):  280ms  277ms  276ms  283ms. 285ms  283ms. 280ms. 279ms
After.   (Entire Block):  264ms. 270ms. 265ms. 268ms. 266ms. 262ms. 263ms. 265ms

Results are significant enough that I would continue here, especially since these are results for *only* exchanging things for the execution state trie, I expect this to get even substantially further if we do this holistically and exchange for all other places where keccak256 hashing is applied (tx trie validations, keccak256 precompile + opcode (there are these two, right?), guess some other things as well).

I will work a bit on this since I would like to get this done in a clean way. If I am not mistaken our whole "replace the hash function" functionality/API is generally not complete (so pretty sure e.g. for VM this is neither possible in the precompile or anywhere + things (the changed hash function) would at the end also needed to passed down to the lower level libraries.


acolytec3:
I'm liking this new approach. I wonder if we should do the same with KZG and move away from the global singleton

packages/client/bin/cli.ts
```diff
@@ -402,6 +414,11 @@ const args: ClientOpts = yargs(hideBin(process.argv))
       'Skip executing blocks in new payload calls in engine, alias for --engineNewpayloadMaxExecute=0 and overrides any engineNewpayloadMaxExecute if also provided',
     boolean: true,
   })
+  .option('useJsCrypto', {
+    describe: 'Use pure Javascript cryptography functions',
+    boolean: true,
+    default: false,
+  })
```
holgerd77, line 421: Hey there, looks great overall! 🤩

One question here: was this intentional to add such an option for the JS/WASM crypto here or did you just misread my suggestion to add a flag *to the EVM/VM state/blockchain test runner* (maybe re-read the original message)? 😂

Intentional or not, also not against such an option in the client as well and since we have it now we might also keep. I do think it will likely still make sense (or: we should do) to add an option to the test runner (so: here https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/test/tester/index.ts#L21 ), so that we can run the VM/EVM tests with or without the WASM crypto.

acolytec3: 🤦 yes, completely misread your original message. Will update the test runner in my next set of changes.

holgerd77: Haha, option is super handy to do performance comparisions though. 😆

I hope it's ok if I play around with this branch and do some updates and testing stuff? Or do you have something local or are working on it right now?

packages/trie/src/trie.ts
```diff
@@ -93,6 +96,8 @@ export class Trie {
         throw new Error('`valueEncoding` can only be set if a `db` is provided')
       }
       this._opts = { ...this._opts, ...opts }
+      this._opts.useKeyHashingFunction =
+        opts.common?.customCrypto.keccak256 ?? opts.useKeyHashingFunction ?? keccak256
```
holgerd77, line 100: Just remember that recently in an exchange with @jochem-brouwer we stumbled upon a problem (unexpected evaluation) when using this kind of operator chaining, can't completely recall the constallation specifics though. 🤔

Did some basic tests and these went as expected:

grafik

So just dropping this here for awareness.

(maybe though it was something with the `?` operator from TypeScript? So a problem might occur when - in the case above - `common` is not defined? Maybe things would then not evaluate properly? Not sure if Jochem can give more insight here (If I recall correctly a solution for *our* problem were adding some brackets)

packages/tx/src/capabilities/legacy.ts
```diff
@@ -45,14 +45,16 @@ export function hash(tx: LegacyTxInterface): Uint8Array {
     throw new Error(msg)
   }
 
+  const keccackFunction = tx.common.customCrypto.keccak256 ?? keccak256
+
   if (Object.isFrozen(tx)) {
     if (!tx.cache.hash) {
-      tx.cache.hash = keccak256(tx.serialize())
+      tx.cache.hash = keccackFunction(tx.serialize())
```
holgerd77, line 52: Really happy here that our capability consolidation in tx is already paying off. 🤩

//cc @gabrocheleau

holgerd77:
Just compiling a list here on what is still missing, in case its helpful (no matter who continues on this):

- [ ] Client: pass `common` to various trie instantiations, particularly the new Snap fetchters, search for `new Trie(` (no additional `Trie.create(` instance apart from VM execution (already updated))
- [ ] Client: bytecode/trienode fetcher hashing (if easily doable, so in `sync/fetcher`)
- [ ] Block: pass `common` to withdrawals and tx trie (this static `genTransactionsTrieRoot()` needs a closer look here)
- [ ] Block: use keccak for hashing (block/uncle hash)
- [ ] StateManager: pass `common` to internal `_trie` object
- [ ] StateManager: `keccak256` usages (various)
- [ ] VM: receipt trie (in `runBlock()` and `buildBlock()`), also `_genTxTrie()` in `runBlock()`
- [ ] EVM: `AUTH` opcode (optional but very easy)
- [ ] Trie: pass to new instantiation in `verifyProof()`
- [ ] Trie: I think `shallowCopy()` does not work yet, might need confirmation
- [ ] Devp2p: various `keccak256` usages (all over the place 🤯, totally wasn't aware actually)
- [ ] All touched: generall give `shallowCopy()` a quick look (eventually, optimally: a test 😬)

And finally, tricky one but I guess important: `Account` in the Util package. Ugh. Likely again with an interface as in trie and then apply where used (mainly EVM?).


Ugh. This list went longer than I expected. 😆 🙂

But on the other hand: if WASM is really that much faster this also REALLY will have an impact if applied so broadly!

packages/trie/src/trie.ts
jochem-brouwer: I can't recall the specifics, do you maybe have some more context when we stumbled upon this? I do slightly remember a "gotcha" with these operators though, but I don't recall in which context :sweat_smile:

packages/tx/src/capabilities/legacy.ts
```diff
@@ -45,14 +45,16 @@ export function hash(tx: LegacyTxInterface): Uint8Array {
     throw new Error(msg)
   }
 
+  const keccackFunction = tx.common.customCrypto.keccak256 ?? keccak256
```
jochem-brouwer, line 48:
```suggestion
  const keccakFunction = tx.common.customCrypto.keccak256 ?? keccak256
```

packages/tx/src/capabilities/legacy.ts
jochem-brouwer, line 52:
```suggestion
      tx.cache.hash = keccakFunction(tx.serialize())
```

packages/tx/src/capabilities/legacy.ts
```diff
@@ -45,14 +45,16 @@ export function hash(tx: LegacyTxInterface): Uint8Array {
...
   if (Object.isFrozen(tx)) {
     if (!tx.cache.hash) {
-      tx.cache.hash = keccak256(tx.serialize())
+      tx.cache.hash = keccackFunction(tx.serialize())
     }
     return tx.cache.hash
   }
 
-  return keccak256(tx.serialize())
+  return keccackFunction(tx.serialize())
```
jochem-brouwer, line 57:
```suggestion
  return keccakFunction(tx.serialize())
```

packages/tx/src/capabilities/legacy.ts
```diff
@@ -82,7 +84,8 @@ export function getSenderPublicKey(tx: LegacyTxInterface): Uint8Array {
   validateHighS(tx)
 
   try {
-    const sender = ecrecover(
+    const erecoverFunction = tx.common.customCrypto.ecrecover ?? ecrecover
+    const sender = erecoverFunction(
```
jochem-brouwer, line 88:
```suggestion
    const ecrecoverFunction = tx.common.customCrypto.ecrecover ?? ecrecover
    const sender = ecrecoverFunction(
```

jochem-brouwer (commented):
Just noticed those spelling mistakes, decided to pack them in a review :) Have not looked anything deeper though (stumbled upon this PR when I thought about swapping ecrecover out too)

packages/block/src/block.ts
```diff
@@ -62,14 +63,15 @@ export class Block {
...
+  private static keccakFunction: ((msg: Uint8Array) => Uint8Array) | undefined
 
   /**
    * Returns the withdrawals trie root for array of Withdrawal.
    * @param wts array of Withdrawal to compute the root of
    * @param optional emptyTrie to use to generate the root
    */
-  public static async genWithdrawalsTrieRoot(wts: Withdrawal[], emptyTrie?: Trie) {
-    const trie = emptyTrie ?? new Trie()
+  public static async genWithdrawalsTrieRoot(wts: Withdrawal[], emptyTrie: Trie) {
```
holgerd77, line 73: This changes our API in a backwards-incompatible way and will break people's code, please revert.

packages/block/src/block.ts
```diff
@@ -81,8 +83,8 @@ export class Block {
    * @param txs array of TypedTransaction to compute the root of
    * @param optional emptyTrie to use to generate the root
    */
-  public static async genTransactionsTrieRoot(txs: TypedTransaction[], emptyTrie?: Trie) {
-    const trie = emptyTrie ?? new Trie()
+  public static async genTransactionsTrieRoot(txs: TypedTransaction[], emptyTrie: Trie) {
```
holgerd77, line 86: Same here.

packages/block/src/block.ts
```diff
@@ -103,6 +105,8 @@ export class Block {
       withdrawals: withdrawalsData,
       executionWitness: executionWitnessData,
     } = blockData
+    this.keccakFunction = this.keccakFunction ?? opts?.common?.customCrypto.keccak256 ?? keccak256
```
holgerd77, line 108: These assignments shouldn't be done in the static constructors but only one-time in the main constructors from `Block` and `BlockHeader`, otherwise this produces unnecessary code redundancy.

(and does `this` even work here already since we have no object context yet??)

Optimally also *after* the `common` assignment, then the `?` after `common` becomes unnecessary.

Also: the `this.keccakFunction ?? ` should be unnecessary if I am not overlooking something since it should/is by design the first (and only) assignment.

holgerd77: (or let me know if there is a reason why this is put here, I was a bit in doubt in between but now again I am not seeing any)

packages/block/src/block.ts
```diff
@@ -340,10 +354,13 @@ export class Block {
       }
     }
 
-    const transactionsTrie = await Block.genTransactionsTrieRoot(txs)
+    const transactionsTrie = await Block.genTransactionsTrieRoot(
+      txs,
+      new Trie({ common: opts?.common })
+    )
```
holgerd77, line 360: Ah, ok, seeing a bit where this was coming from with the `trie` option made mandatory, should nevertheless not be necessary. Here we pass over the new trie and then that's it.

packages/block/src/block.ts
holgerd77: (and, to add on this: if there *is* a reason we should really try to remove that reason, since this is atm really not beautiful and this reason might generally also point to some suboptimality in the (existing) code)

packages/block/src/header.ts
```diff
@@ -65,6 +65,9 @@ export class BlockHeader {
 
   public readonly common: Common
 
+  private keccakFunction?: Function
+  private static keccakFunction: ((msg: Uint8Array) => Uint8Array) | undefined
```
holgerd77, line 69: Ok, all this weird static/not-static doubling should really go away. Not sure why this is happening, but this should really be only the member property like:

```typescript
protected keccakFunction: ((msg: Uint8Array) => Uint8Array)
```

Then all `!` TypeScript operators should also be not necessary.

Side thing: can you also please leave make this `protected` instead of `private`? This leaves more room for community users who want to subclass the `BlockHeader` class (so `Block` the same). Would be nice if you can also update the other member declarations (e.g. in tx) for `keccakFunction` along this line.

packages/block/src/header.ts
```diff
@@ -144,16 +150,17 @@ export class BlockHeader {
...
-      this.common = options.common.copy()
+  constructor(headerData: HeaderData, opts: BlockOptions = {}) {
+    if (opts.common) {
+      this.common = opts.common.copy()
     } else {
       this.common = new Common({
         chain: Chain.Mainnet, // default
       })
     }
+    this.keccakFunction = this.keccakFunction ?? opts?.common?.customCrypto.keccak256 ?? keccak256
```
holgerd77, line 161: You can use `this.common` here already and spare all the `?` checks.

holgerd77 (commented):
Thanks for the continuation 🙏, some comments to address.

holgerd77:
Also, side note: before we do any further functional additions, all tests should be working again.

holgerd77:
Update: I think I will just work a bit on this myself, maybe I will get a better grasp why this static/non-static thing was done or if it was just an overlook.

Will push my work (if something useful comes out).

packages/block/src/block.ts
holgerd77: Seeing a bit where this is coming from, so the `new Trie()` would obviously not take a common here (and so not be able to leverage the WASM crypto if trie is not provided. But that's something we need to live with right now, then one just needs to pass the trie if one wants the crypto.

We can add this to #3216, have done so.

Otherwise have tackled here (push later).

packages/block/src/block.ts
holgerd77: I am now done with this and think this was really some logic error on your side. Do not see any limitations or things malfunctioning by this.

holgerd77:
Ok, pushed my changes.

To be continued. 🙂

holgerd77:
(and update on the tests: ah, ok, this is "only" this package resolution thing, then discard my non-continue message from above)

holgerd77:
Gave devp2p another look.

This might be rather a thing for a new PR. 

This seems to need some various structural changes and this would likely overload the PR here.

holgerd77:
Updated this via UI.

@acolytec3 feel free to merge this in whenever you think it’s ready!

holgerd77:
(from my side additional commits look good)

packages/client/bin/cli.ts
jochem-brouwer, line 1: Nit: with `loadBlocksFromRlp` the custom crypto arg is not respected.

packages/client/bin/cli.ts
```diff
@@ -828,6 +890,8 @@ async function run() {
       chain: chainName,
       mergeForkIdPostMerge: args.mergeForkIdPostMerge,
     })
+    //@ts-ignore
+    common.customCrypto = cryptoFunctions
```
jochem-brouwer, line 894: Nit: should likely add this `customCrypto` to the `fromGethGenesis` constructor also?

packages/client/test/util/wasmCrypto.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+      },
+    })
+    const common = new Common({ chain: 'mainnet' })
+
+    const pk = randomBytes(32)
+    const tx = LegacyTransaction.fromTxData({}, { common }).sign(pk)
+    const tx2 = LegacyTransaction.fromTxData({}, { common: commonWithCustomCrypto }).sign(pk)
+
+    assert.deepEqual(tx.getSenderPublicKey(), tx2.getSenderPublicKey())
+    assert.deepEqual(tx.hash(), tx2.hash())
```
jochem-brouwer, line 55: Same here as commented on cli.ts, these wrapper methods for `ecsign` and `ecrecover` should move somewhere else. (And thus also these tests)

packages/devp2p/src/rlpx/rlpx.ts
```diff
@@ -50,6 +51,8 @@ export class RLPx {
   protected _refillIntervalId: NodeJS.Timeout
   protected _refillIntervalSelectionCounter: number = 0
 
+  protected _keccakFunction: (msg: Uint8Array) => Uint8Array
```
jochem-brouwer, line 54: Not sure, why is this added? It's used nowhere?

packages/evm/test/customCrypto.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { Chain, Common } from '@ethereumjs/common'
+import { assert, describe, it } from 'vitest'
+
+import { EVM } from '../src/evm.js'
+import { getActivePrecompiles } from '../src/index.js'
+
+describe('custom crypto', () => {
+  it('should use custom sha256 function', async () => {
+    const customSha256 = (msg: Uint8Array) => {
```
jochem-brouwer, line 9: Nit: there is no similar test for `ecrecover`

packages/trie/src/trie.ts
```diff
@@ -143,13 +148,15 @@ export class Trie {
...
+    const keccakFunction =
+      opts?.common?.customCrypto.keccak256 ?? opts?.useKeyHashingFunction ?? keccak256
     let key = ROOT_DB_KEY
 
     const encoding =
       opts?.valueEncoding === ValueEncoding.Bytes ? ValueEncoding.Bytes : ValueEncoding.String
 
     if (opts?.useKeyHashing === true) {
-      key = (opts?.useKeyHashingFunction ?? keccak256)(ROOT_DB_KEY) as Uint8Array
+      key = keccakFunction.call(undefined, ROOT_DB_KEY) as Uint8Array
```
jochem-brouwer, line 159: Not sure, why do we have to `.call` here?

jochem-brouwer (dismissed):
Left a few nits, ready to merge once CI passes

packages/evm/test/customCrypto.spec.ts
acolytec3: Just added.

packages/trie/src/trie.ts
acolytec3: :shrug: I think this was left over from when @holgerd77 was trying to introduce a global crypto thing or something.

packages/devp2p/src/rlpx/rlpx.ts
acolytec3: Not sure, but I'll sure use it in future devp2p PRs.

packages/client/bin/cli.ts
acolytec3: Good point.  Maybe the next PR.

packages/client/bin/cli.ts
```diff
@@ -783,15 +805,54 @@ async function run() {
...
+    ) =>
+      secp256k1Expand(
+        secp256k1Recover(
+          msgHash,
+          concatBytes(setLengthLeft(r, 32), setLengthLeft(s, 32)),
+          Number(calculateSigRecovery(v, chainID))
+        )
+      ).slice(1)
+    cryptoFunctions.sha256 = wasmSha256
+    cryptoFunctions.ecsign = (msg: Uint8Array, pk: Uint8Array, chainId?: bigint) => {
```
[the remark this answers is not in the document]
acolytec3: Why is this?  We don't want the wasm crypto dep to be in any other package than `client` so we don't mess up library consumers who have challenges with WASM.  Every wrapper function will likely be custom since the underlying crypto functions may have different signatures.

acolytec3 (approved):
Approving this.  @jochem-brouwer or @holgerd77, this can be merged if everyone is satisfied.  Plenty of work to add in future PRs but I think this one is fine as is.

packages/client/bin/cli.ts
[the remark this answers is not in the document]
jochem-brouwer: Ah ok, this makes sense :smile:

jochem-brouwer:
I'm fine to merge this when CI passes :)

holgerd77:
👍

PR #3195, opened 2023-12-14
evm: fix CALL(CODE) gas

### LINKED ISSUE #3194: CALLCODE calculates gas incorrectly
# Bug description

The `CALLCODE`'s gas calculation is bias from the standard implementation.

Specifically, if the left gas is not enough to cover the gas of `CALLCODE`,  the EVM should [raise the out of gas error and abort the call](https://github.com/ethereum/go-ethereum/blob/v1.13.5/core/vm/interpreter.go#L214-L216). However, the typescript EVM continues execution in some cases, if the amount of gas to be consumed [does not exceed the remaining gas by 2300 gas](https://github.com/ethereumjs/ethereumjs-monorepo/blob/2b5f86ab8fda734b4b7af6ceaf49b73cb62bce40/packages/evm/src/opcodes/gas.ts#L429-L433), which happens to be the call stipend constant.

It is important to emphasize, the call stipend should only be [added to the gas limit](https://github.com/ethereum/go-ethereum/blob/v1.13.5/core/vm/instructions.go#L719-L722) if the current left gas can cover the gas in `CALLCODE/CALL` opcode, otherwise the EVM should raise the OOG error and abort the call.

We found that the [similar code snippet existed in CALL](https://github.com/ethereumjs/ethereumjs-monorepo/blob/2b5f86ab8fda734b4b7af6ceaf49b73cb62bce40/packages/evm/src/opcodes/gas.ts#L387-L389) is missing in `CALLCODE`, which could fix the bug.

Below is a poc of this bug, we can trigger it by calling the following contract with 50000 tx gas limit under shanghai upgrade: 

```bash
# Bytecode
6000600060006000600130611a90f2
# Mnemonic
PUSH1 0x00
PUSH1 0x00
PUSH1 0x00
PUSH1 0x00
PUSH1 0x01
ADDRESS
PUSH2 0x1a90
CALLCODE
```

Explain: the contract recursively "callcode"s itself. In the second time the `CALLCODE` is invoked, the geth raises out of gas error, but Typescript EVM continues execution because it includes the call stipend.

jochem-brouwer:
Closes #3194

Ready for review, assuming CI passes

### DIFF
--- a/packages/evm/src/opcodes/functions.ts
+++ b/packages/evm/src/opcodes/functions.ts
@@ -1025,7 +1025,7 @@ export const handlers: Map<number, OpHandler> = new Map([
   // 0xf1: CALL
   [
     0xf1,
-    async function (runState: RunState) {
+    async function (runState: RunState, common: Common) {
       const [_currentGasLimit, toAddr, value, inOffset, inLength, outOffset, outLength] =
         runState.stack.popN(7)
       const toAddress = new Address(addresstoBytes(toAddr))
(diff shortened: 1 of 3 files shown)


acolytec3 (approved):

acolytec3 (approved):
LGTM

PR #3196, opened 2023-12-14
EIP 4444 explorations

acolytec3:


acolytec3:
Before I can proceed further with integrating `ultralight` within the `client` so we can serve ancient block history, we need to resolve the fundamental challenge that the `client` is CJS and `ultralight` is ESM and has a number of dependencies that are ESM only.  I've done some research and haven't yet found a bundler that is able to rewrite ESM dependencies to CJS where the ESM dependency uses the `export` key (currently not rewriteable by `esbuild` or `swc` from what I can tell).  
So, the current options I see are:
1) Have the client start the Ultralight `cli` app in a child process and make RPC calls to it in order to access block history.  Probably the "easiest" from a purely implementation perspective but it might add some undue complexity when troubleshooting issues.
2) Rewrite the client and really the whole monorepo to use ESM only so we can easily import all the upstream ESM dependencies.  This feels like a longer term correctly solution but debatable until if/when Node 20 or 22 entirely drops CJS support (which I don't think is currently on the table)
3) Add some new build step (maybe using something like esbuild + a custom plugin that will convert upstream dependencies to CJS in a way the client can import them) - probably the hardest

holgerd77:
Thanks for the update and the good summary of potential paths forward! 🙏

Issue #3140, opened 2023-11-04
Browser examples break for libraries that use `debug`


acolytec3:
The browser examples that are provided with our libraries will no longer work out of the box for all packages that use `process.env` to detect the debug log topics like [this](https://github.com/ethereumjs/ethereumjs-monorepo/blob/6d64162c847c7cd145202ad7c0af276a286138cc/packages/trie/src/trie.ts#L105).  This can be resolved by using a bundler to build the library and replace references to `process.env` with a suitable alternative (or just default it to `false`) but we will need to revisit how we enable debug logs if we want these examples to start working again. 

One option, which Ultralight uses, is to set the debug topics via `debug.enableLog('yourPreferredDebugTopic,anotherDebugTopic')` but this requires a compile time change to the topics at present so probably not ideal for our purposes.

acolytec3:
This set of vite plugins will resolve the issue:
![image](https://github.com/ethereumjs/ethereumjs-monorepo/assets/17355484/3fedf8f3-b983-423b-a39e-ccee008b6eb7)

We should at a minimum update our docs to show this

roninjin10:
I think some of these we could get rid of. For example, I think topLevelAwait comes from a 1 specific dependency awaiting wasm that doesn't need to be. I prefer `vite-plugin-node-polyfills` over `vite-plugin-ngmi-poolyfilll`

acolytec3:
Good to know.  Will experiment with your suggested alternatives and see how it goes.

holgerd77:
Closed by #3197

PR #3197, opened 2023-12-14
statemanager: Fix statemanager Browser Example

scorbajio:
The `statemanager` package browser example has two errors when it's run. The first error comes from referencing `process` in the trie package:
```
trie.ts:105 Uncaught (in promise) ReferenceError: process is not defined
    at new Trie (trie.ts:105:18)
    at new DefaultStateManager (stateManager.ts:200:23)
    at run (browser.html:19:34)
    at browser.html:31:9
```
The second error results from using `hexStringToBytes` instead of `hexToBytes`. This change fixes both of these issues so that the example successfully runs in the browser.

### DIFF
--- a/packages/statemanager/examples/browser.html
+++ b/packages/statemanager/examples/browser.html
@@ -10,7 +10,7 @@
 
         const run = async () => {
             const stateManager = new DefaultStateManager()
-            const address = new Address(hexStringToBytes('a94f5374fce5edbc8e2a8697c15331677e6ebf0b'))
+            const address = new Address(hexToBytes('0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b'))
             const account = new Account(BigInt(0), BigInt(1000))
             await stateManager.checkpoint()
             await stateManager.putAccount(address, account)
--- a/packages/trie/src/trie.ts
+++ b/packages/trie/src/trie.ts
@@ -102,7 +102,8 @@ export class Trie {
       valueEncoding = ValueEncoding.Bytes
     }
 
-    this.DEBUG = process.env.DEBUG?.includes('ethjs') === true
+    this.DEBUG =
+      typeof window === 'undefined' ? process?.env?.DEBUG?.includes('ethjs') ?? false : false
     this.debug = this.DEBUG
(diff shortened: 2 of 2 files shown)


scorbajio:
The `process.env` issue is noted in #3140 as well.

holgerd77 (changes requested):
We should really see that we fix this at the source level and make it easier for people to use the libraries in a browser context rather then adding new polyfills here or something (in fact we should also rather try to remove the other two vite plugins). Especially if it is about something as unimportant (for the end user) as our `DEBUG` functionality.

I just re-read the issue and discussion: so the `process` thing is really a non-issue and should be easy to fix at the source.

So @acolytec3 mentioned that we might want to disable this for browser (set it to `false` somewhere). This is actually already the case in the very most libraries (except for trie, guess it just got forgotten or something).

So this is the code:

```typescript
this.DEBUG =
      typeof window === 'undefined' ? process?.env?.DEBUG?.includes('ethjs') ?? false : false
```

So this translates to: if `window` object is not there (aka: not in browser), try use debugging. Otherwise don't.

We should just make sure that this snippet is consistently used within all libraries and then modfiy it on all occasions to not break in the browser.

One is actually already questioning why this is not working. 🤔 Likely something to do with the chaining of `?` operators or something.

So, when looking at the ESM build (by simply going into the `dist/esm` folder of `statemanager`), one can see that this translates to:

```typescript
this.DEBUG =
            typeof window === 'undefined' ? process?.env?.DEBUG?.includes('ethjs') ?? false : false;
```

This can easily tested in the browser by just going into a browser debug console - e.g. 
in Firefox by going to "Extras -> Browser Tools (or similar) -> Tools for Web Developers" (also some guide for @jochem-brouwer who mentioned he has not so much browser experience, really encourage you to give this a very short try, *absolutely* no rocket science!). 😄

grafik

As the screenshot shows the first version (coming from `statemanager`) evaluates properly. The second version (from `trie`) breaks.

So we should simply look through the libraries and just take the common pattern (from e.g. `statemanager`) and the problem should. be already solved.

holgerd77:
Updated this via UI

holgerd77 (approved):
LGTM

PR #3198, opened 2023-12-15
EVM: bump opcode coverage + add performance tester

jochem-brouwer:
The goal of this PR is to:

- Add basic test coverage of most opcodes
- Use the code ran in these tests to generate a profile report of the EVM

WIP


acolytec3:
This looks great but we should merge #3198 before merging this one. It's not obvious to me but are these opcode tests also run by CI?

packages/evm/test/opcodes/profiler.ts
```diff
@@ -0,0 +1,103 @@
...
+  for (const testSet in opcodeTests) {
+    const testCases = opcodeTests[testSet]
+    for (const opcodeName in testCases) {
+      const tests = testCases[opcodeName]
+      for (const test of tests) {
+        const testName = getOpcodeTestName(opcodeName, test.stack, test.name)
+        const code = makeLoopCode(
+          createBytecode([createOpcodeTest(test.stack, opcodeName, 'none'), POP])
+        )
+        const evm = new EVM({ common, profiler: { enabled: true } })
```
acolytec3, line 60:
```suggestion
        common.events.removeAllListeners()
```

Since we're reusing the same `common` each time, we should clear listeners because the `evm` adds a new listener [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/15f7e1770cf3ea92ccee225b67d23cb28e96ff70/packages/evm/src/evm.ts#L190) and I see memory leak errors when I run the profiler script.

holgerd77:
@jochem-brouwer What is the status of this PR?

jochem-brouwer:
I intend to finish this one this month. The goal is to write basic tests for all opcodes and provide a test framework.

PR #3200, opened 2023-12-15
client: Snap Sync Debugging

scorbajio:
The goal of this PR is to debug the current snap sync implementation and verify that the following scenarios from @g11tech are handled as expected:
1. eoa state inflated pre-sync with post sync all kind of txs
2. storage state inflated pre-sync with post sync all kind of txs
3. bytecode inflated pre-sync with post sync all kind of txs
4. Static snap sync with all of the above inflated pre-sync with post sync all kind of txs
5. Moving state snap sync with continuous eoa txs
6. Moving state snap sync with continuous contract storage txs
7. Moving state snap sync with continuous code deployment txs


holgerd77:
Updated this via UI.

@scorbajio what is the status of this PR?

scorbajio:
> Updated this via UI.
> 
> @scorbajio what is the status of this PR?

This is ready for review. 

> Updated this via UI.
> 
> @scorbajio what is the status of this PR?

This is ready for review.

cc @g11tech

packages/client/src/sync/fetcher/storagefetcher.ts
```diff
@@ -27,7 +27,7 @@ import type { Job, SnapFetcherDoneFlags } from './types'
 import type { Debugger } from 'debug'
 const { debug: createDebugLogger } = debugDefault
 
-const TOTAL_RANGE_END = BIGINT_2 ** BIGINT_256 - BIGINT_1
+const TOTAL_RANGE_END = BigInt(BIGINT_2 ** BIGINT_256 - BIGINT_1)
```
g11tech, line 30: wouldn't this be bigint anyway?

packages/client/src/sync/fetcher/storagefetcher.ts
```diff
@@ -151,26 +152,34 @@ export class StorageFetcher extends Fetcher<JobTask, StorageData[][], StorageDat
...
-    } else {
-      origin = bigIntToBytes(first + BIGINT_1)
+    let origin: Uint8Array
+
+    try {
+      const { task, partialResult } = job
+
+      if (task.multi === true) {
+        // peer does not respect origin or limit for multi-account storage fetch
+        return setLengthLeft(bigIntToBytes(BIGINT_0), 32)
```
g11tech, line 162: even if peer doesn't respect the origin or limit, but if we provide it, does it give an issue?

packages/client/src/sync/fetcher/storagefetcher.ts
```diff
@@ -151,26 +152,34 @@ export class StorageFetcher extends Fetcher<JobTask, StorageData[][], StorageDat
...
-    let origin = undefined
-    if (partialResult) {
-      const lastSlotArray = partialResult[partialResult.length - 1]
-      const lastSlot = lastSlotArray[lastSlotArray.length - 1]
-      origin = bigIntToBytes(bytesToBigInt(lastSlot.hash) + BIGINT_1)
-    } else {
-      origin = bigIntToBytes(first + BIGINT_1)
+    let origin: Uint8Array
+
+    try {
```
g11tech, line 157: could you add a comment  to code which particular error scenario motivated you to put this in try/catch

packages/client/src/sync/fetcher/storagefetcher.ts
scorbajio: The parsing of `partialResult` was resulting in indexing and undefined errors since `lastSlot` was an array in certain instances instead of a slot object.

packages/client/src/sync/fetcher/storagefetcher.ts
scorbajio: The parsing of partialResult was resulting in indexing and undefined errors since lastSlot was an array in certain instances instead of a slot object. I will add a comment about it to the code.

packages/client/src/sync/fetcher/storagefetcher.ts
scorbajio: Yes, I think this was something I did while debugging and forgot to clean up. Fixed.

holgerd77:
Updated this via UI

PR #3201, opened 2023-12-15
Fix Blake2F gas + output calculation on non-zero aligned inputs

kchojn:
Revised the DataView initialization to correctly account for the `byteOffset` in Uint8Array, ensuring accurate data alignment.
fixed:
- rounds, hRaw, mRaw, tRaw

testes with:
- tx_hash: `0x4f2e13a0a3f14033630ab2b8cdad09d316826375f761ded5b31253bb42e0a476`
- data: `0x0000000c28c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b3dd8338ed89de6791854126751ac933302810c04147014e9eb472e4dbc09d3c96abb531c9ae39c9e6c454cb83913d688795e237837d30258d11ea7c75201003000454cb83913d688795e237837d30258d11ea7c752011af5b8015c64d39ab44c60ead8317f9f5a9b6c4c01000000000100ca9a3b000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000`


jochem-brouwer:
This fixes a consensus bug when the original input data into Blake2F is not at offset 0 of the current memory of the calling contract. Have added a test which shows this (fails on master). Not only is the gas output different, also the result differs.

acolytec3 (approved):
LGTM

holgerd77:
This will be released with #3261

PR #3205, opened 2023-12-18
Add missed `.js file endings

acolytec3:
Adds a few missing `.js` file endings


jochem-brouwer (approved):
LGTM

Side note: the linter throws for me if I don't add these .js extensions, so why does it not throw on CI? (I have not checked these packages though, but at least in VM/EVM it threw when I forgot to add those extensions)

PR #3206, opened 2023-12-18
evm: ensure modexp right-pads input data

### LINKED ISSUE #3204: Modexp should pad shorter input with zeros
This is another corner difference between the standard implementation and Typescript EVM. 

Whenever the input is too short, the missing bytes are considered to be zero for modexp input. While the typescript EVM does pad the input when reading values of the three words: 

https://github.com/ethereumjs/ethereumjs-monorepo/blob/a58c6cb1c5f034dfbb2bffd0bb369803b69c5109/packages/evm/src/precompiles/05-modexp.ts#L175-L177

it doesn't pad when reading lengths of the words: 

https://github.com/ethereumjs/ethereumjs-monorepo/blob/a58c6cb1c5f034dfbb2bffd0bb369803b69c5109/packages/evm/src/precompiles/05-modexp.ts#L113-L115

Here is the implementation of geth for reference: https://github.com/ethereum/go-ethereum/blob/v1.13.5/core/vm/contracts.go#L318-L320

Below is a simple poc of this bug. It invokes the modexp with input `[0x41, ]`.  After padding with zeros, the gas cost is very large so the EVM is expected to abort the call.  

```
# Bytecode
0x7f4100000000000000000000000000000000000000000000000000000000000000600052604060006001600060006005618d8ef100
# Mnemonic
PUSH32 0x4100000000000000000000000000000000000000000000000000000000000000
PUSH1 0x00
MSTORE
PUSH1 0x40
PUSH1 0x00
PUSH1 0x01
PUSH1 0x00
PUSH1 0x00
PUSH1 0x05
PUSH2 0x8d8e
CALL
STOP
```

jochem-brouwer:
Closes #3204 

I have reviewed EVM to see if there are more cases of doing "not-safe-slices" of any input. The only other candidate which could have impact which I could find are the ECMUL and ECADD precompiles. We should investigate this. (To be addressed in another PR).

### DIFF
--- a/packages/block/src/header.ts
+++ b/packages/block/src/header.ts
@@ -28,7 +28,7 @@ import { keccak256 } from 'ethereum-cryptography/keccak.js'
 import { CLIQUE_EXTRA_SEAL, CLIQUE_EXTRA_VANITY } from './clique.js'
 import { fakeExponential, valuesArrayToHeaderData } from './helpers.js'
 
-import type { BlockHeaderBytes, BlockOptions, HeaderData, JsonHeader } from './types'
+import type { BlockHeaderBytes, BlockOptions, HeaderData, JsonHeader } from './types.js'
 import type { CliqueConfig } from '@ethereumjs/common'
 import type { BigIntLike } from '@ethereumjs/util'
 
--- a/packages/common/src/chains.ts
+++ b/packages/common/src/chains.ts
@@ -1,4 +1,4 @@
-import type { ChainConfig } from './types'
+import type { ChainConfig } from './types.js'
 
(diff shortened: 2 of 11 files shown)


holgerd77:
Rebased this via UI

holgerd77:
Tests failing, not sure what happened (simply haven’t looked).

jochem-brouwer:
Hmm the modexp tests on blockchain/state are failing so that's interesting, I thought this fix would be somewhat trivial :thinking:

acolytec3 (approved):
LGTM.  Verified that the data length gets appropriately set

Issue #3134, opened 2023-11-01
CodeCov challenges


acolytec3:
We are currently having some challenges with CodeCov reports being incomplete or not loading correctly.

- CodeCov is unable to process the coverage reports for `rlp` and  `ethash`  and gives an error like below:
![image](https://github.com/ethereumjs/ethereumjs-monorepo/assets/17355484/37eadd06-30e3-4fa9-a9e8-f9bd91ff2025)
I've inspected the reports and they appear to follow the same format as our other coverage reports so it's not obvious what the issue is here.

- `genesis` and `wallet` are not populating coverage reports at all so we should add this to their CI jobs and see if we can get CodeCov to include them in its reporting.

holgerd77:
This seems to be getting adressed by latest codecov work from @scorbajio and generally issue is somewhat old, will close.

PR #3207, opened 2023-12-19
Codecov tuning

scorbajio:
The goal of this PR is to look into and fix the issues mentioned in #3134.

### DIFF
--- a/.github/workflows/genesis-build.yml
+++ b/.github/workflows/genesis-build.yml
@@ -39,4 +39,11 @@ jobs:
         working-directory: ${{github.workspace}}
 
       - run: npm run lint
+      
+      - run: npm run coverage
+      - uses: codecov/codecov-action@v3
+        with:
+          files: ${{ env.cwd }}/coverage/lcov.info
+          flags: genesis
+
       - run: npm run test:node # Only run node tests for now until vitest browser test issues are sorted out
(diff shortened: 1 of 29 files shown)


holgerd77:
That's nice, that this gets tackled a bit again, would be really good if we get coverage a bit more integrated in our "everyday thinking" again! 😁

Everytime you get something working what was not working before, would be nice if you drop a short note here, I would be really interested to follow up on where and why this was actually hanging so much.

scorbajio:
Now that their workflow and `package.json` files have been updated, the `wallet` and `genesis` packages are reporting coverage as expected. Both the `rlp` and `genesis` package coverage uploads fail codecov-side with only a `unusable report` error message. [This](https://docs.codecov.com/docs/error-reference) doc page gives four different possible reasons that could be causing the error, however, none of these reasons seem to be the causing the issue.

holgerd77:
Updated this via UI

holgerd77 (approved):
Nice, LGTM!

scorbajio:
I've created a ticket on the codecov-api repo about the `unusable report` error being too generic: https://github.com/codecov/codecov-api/issues/319.

holgerd77:
> I've created a ticket on the codecov-api repo about the `unusable report` error being too generic: [codecov/codecov-api#319](https://github.com/codecov/codecov-api/issues/319).

Cool, not sure if anyone cares there, but let’s be brave! 🙂😅🤩

PR #3209, opened 2023-12-20
Fix broken code path in client readme

timqian:


packages/client/README.md
```diff
@@ -328,7 +328,7 @@ ethereumjs --rpc --maxPeers=0
 ```
 
 Currently only a small subset of `RPC` methods are implemented.(\*) You can have a look at the
-[./lib/rpc/modules/](./lib/rpc/modules) source folder or the tracking issue
+[./src/rpc/modules/](./src/rpc/modules) source folder or the tracking issue
```
holgerd77, line 331: Ah, that's actually a good one, we completely didn't think about the README when we aligned this to match the other libraries (all having a `src` folder). Thanks! 🙏

holgerd77 (approved):
LGTM, thanks for the quick update!

packages/client/README.md
timqian: Thank you for your work on ethereumjs! It is super valuable for me as a JS enthusiast who would like to dive deep into understanding how Ethereum works

PR #3210, opened 2023-12-20
Fix npm command

timqian:


holgerd77:
Hi there, thanks a lot for your contribution!

Can you please submit these two small doc fixes in one single PR (so: e.g. push here and close the other)?

We have a very expensive (in the sense of time/compute ressources) CI run for every PR go through, and it is otherwise too wasteful to have this run for every line of doc change.

Thanks a lot! 🙏

timqian:
@holgerd77 Thanks for the note, as the other PR contains more updates. I have committed the change to the other PR

PR #3212, opened 2023-12-28
Dependency Updates

scorbajio:
This change updates the following dependencies from the following packages:

#### client: ####
* `fs-extra` 10.1.0 to 11.2.0
* `@types/fs-extra` 9.0.13 to 11.0.4
#### ~~trie:~~ #### 
(major version update is breaking due to circular dependency breaking browser tests for trie)
* ~~`@types/readable-stream` 2.3.15 to 4.0.10~~
* ~~`readable-stream` 3.6.2 to 4.5.2~~
#### ~~devp2p:~~ ####
(being handled in #3188)
* ~~`lru-cache` 7.18.3 to 10.1.0~~


holgerd77:
Maybe take the `lru-cache` update out here for now, lru is always hairy, since this is such a central dependency, also behavior changed substantially (especially around the usage of `null`, `undefined` as cache values) so a simple update should be handled with care and often also requires changes on types and code logic.

I think also @acolytec3 did some parallel update on this in some parallel PR, so that's another remark, but that should also be aligned and coordinated.

So, but for here: guess buest to leave `lru-cache` out and keep this to some other few dependencies.

acolytec3 (approved):
LGTM

acolytec3 (dismissed):
LGTM

acolytec3 (approved):
LGTM

acolytec3 (approved):
Goodbye obsolete deps (and may our installs be faster now that we're nearly 100 deps lighter)

acolytec3 (approved):

PR #3214, opened 2024-01-05
Update callcode test 

acolytec3:
This updates the `CALLCODE` test added in #3194 with clearer execution results (and mirrors a new test we're proposing to add to the execution-api-tests).

### DIFF
--- a/packages/evm/test/runCall.spec.ts
+++ b/packages/evm/test/runCall.spec.ts
@@ -725,12 +725,16 @@ describe('RunCall tests', () => {
       common,
     })
 
-    // Callcode
-    for (const [opcode, gas] of [
-      ['f1', 36600], // 36600 is CALL fee
-      ['f2', 11600], // 11600 is CALLCODE fee
+    for (const [opcode, gas, expectedOutput] of [
+      ['f1', 36600, '0x'], // 36600 is CALL fee
+      ['f2', 11600, '0x'], // 11600 is CALLCODE fee
+      ['f1', 36600 + 7 * 3, '0x01'], // 36600 is CALL fee + 7 * 3 gas for 7 PUSH opcodes
+      ['f2', 11600 + 7 * 3, '0x01'], // 11600 is CALLCODE fee + 7 * 3 gas for 7 PUSH opcodes
     ]) {
       // Code to either CALL or CALLCODE into AACC empty contract, with value 1
+      // If enough gas is provided, then since nonzero value is sent, the gas limit
+      // in the call(coded) contract will get the "bonus gas" stipend of 2300
+      // Previously, we added this gas stipend to the current gas available (which is wrong)
 
       /***
        * Bytecode for AAAA contract (used to check CALL/CALLCODE execution when gas is less than required)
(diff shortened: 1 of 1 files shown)


packages/evm/test/runCall.spec.ts
```diff
@@ -725,37 +725,67 @@ describe('RunCall tests', () => {
...
-      await evm.stateManager.putContractCode(contractAddress, contractCode)
+      // Code to either CALL or CALLCODE into AACC empty contract, with value 1
 
-      const account = await evm.stateManager.getAccount(contractAddress)
-      account!.balance = BIGINT_100
+      const callCodeAddress = Address.fromString('0x000000000000000000000000000000000000aaaa')
+      const callCode = hexToBytes(`0x6000600060006000600161AACC611a90${opcode}`)
 
-      await evm.stateManager.putAccount(contractAddress, account)
+      const gas = (11600).toString(16).padStart(4, '0') // Enough gas so less than required for CALL/CALLCODE before adding call stipend
```
jochem-brouwer, line 733: Although we were not vulnerable to this bug in CALL previously, the amount of gas necessary for CALL to trigger this is different than the gas for CALLCODE (which is 11600 for Shanghai)

packages/evm/test/runCall.spec.ts
```diff
@@ -725,37 +725,67 @@ describe('RunCall tests', () => {
...
-      let jumpdestCalls = 0
+      const callCodeAddress = Address.fromString('0x000000000000000000000000000000000000aaaa')
+      const callCode = hexToBytes(`0x6000600060006000600161AACC611a90${opcode}`)
 
-      evm.events.on('step', (e) => {
-        if (e.opcode.name === 'JUMPDEST') {
-          jumpdestCalls++
-        }
-      })
+      const gas = (51600).toString(16).padStart(4, '0') // Enough gas to execute CALL/CALLCODE before adding call stipend
```
jochem-brouwer, line 763: I think this variant should use the exact amount of gas necessary to correctly perform the CALL(CODE) for explicitness (so when the CALL(CODE) in the callCodeAddress finishes there is exactly 0 gas left, which means it still has enough gas left to finish the call frame)

packages/evm/test/runCall.spec.ts
jochem-brouwer: (Likeliky you can do `for [opcode, gas] of [['f1', GasCall], ['f2', 11600]] {`)

jochem-brouwer (commented):
Some small comments. I think in general a short note what the bytecode does would also be a great addition for the future.

jochem-brouwer (dismissed):
LGTM!

packages/evm/test/runCall.spec.ts
```diff
@@ -724,38 +724,96 @@ describe('RunCall tests', () => {
...
-      const contractAddress = Address.fromString('0x000000000000000000000000636F6E7472616374')
-      await evm.stateManager.putContractCode(contractAddress, contractCode)
 
-      const account = await evm.stateManager.getAccount(contractAddress)
-      account!.balance = BIGINT_100
-
-      await evm.stateManager.putAccount(contractAddress, account)
+    // Callcode
+    for (const [opcode, gas] of [
+      ['f1', 36600], // 36600 is CALL fee
```
jochem-brouwer, line 730: Cool, same number I got :smile:

jochem-brouwer (approved):
Assuming CI passes LGTM

PR #3219, opened 2024-01-08
Generalized Runtime Docs / (First round) Bun Support

holgerd77:
This PR adds a generalized "Runtimes" information section to the docs combining info for Node.js and browser (rewritten to be more specific) and then also adds a new "Bun" sub section there (we might want to add Deno as well).

I've done this in an examplaric way for one library, open for a first round of feedback, then I would expand (the same section/structure) to the other libraries (except client).

I've also added a `simple.ts` example, I would want to have one for each package, since this just makes sense in general.

For `Bun` I have also added a new GH Actions script. This is meant to be run nightly. After a first round I would comment out all non-working packages. For the working ones we can add the new "Bun Support" runtime section.


.github/workflows/bun-support.yml
```diff
@@ -0,0 +1,83 @@
...
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - uses: oven-sh/setup-bun@v1
```
roninjin10, line 24:
```suggestion
      - uses: oven-sh/setup-bun@v1
        with:
          bun-version: 1.0.24
```
Might want to pin the version so there is no ambiguity if a pr caused a regression or bun version changed it.

Alternatively consider just echoing the bun version in the next step of this test so if you are debugging a workflow you know exactly which version of bun it used

.github/workflows/bun-support.yml
```diff
@@ -0,0 +1,83 @@
...
+        run: bunx vitest run test
+        working-directory: packages/blockchain
+
+      # Client not working yet in Bun (expected), 2024-01-08
+      # - name: Test Client
+      #   run: bunx vitest run test
+      #   working-directory: packages/client
+
+      - name: Test Common
+        run: bunx vitest run test
```
roninjin10, line 42: I don't think this is actually using bun. Because vitest is an executable bun should default to using node here. I think you need to pass in the `--bun` flag to use bun and you might have issues related more to vitest than bun.
```suggestion
        run: bun --bun run vitest run test
```

roninjin10 (commented):
This is very nice to see this happening! I've been implicitly using bun in [tevm](https://github.com/evmts/tevm-monorepo) for the evm blockchain and state packages for a while without issues

.github/workflows/bun-support.yml
roninjin10: Docs on this are here https://bun.sh/docs/cli/run#bun

holgerd77: Oh, that is the "tip of the year"! 🙏 🙂 

I was already wondering why things are so smooth. 😂 That is the missing link I needed!

holgerd77:
@roninjin10 Thanks a lot for the feedback, that helps a lot!

I wanted to wait for our "self-contained examples" work (see some other now merged in PRs), so we basically have made all examples self-contained, so that they can now run "on their own" with all necessary dependencies included and stuff, the examples are now also included in a dedicated `examples` folder per package.

So this makes `bun` EthereumJS API testing easier. The examples work is now done, so I will pick up this PR sometime soon (next week or the week after I guess). Sufficient basis for this now, also with your `--bun` flag tip (thanks again for that!).

.github/workflows/bun-support.yml
holgerd77: Just to note that this is not yet working (using bun 1.0.26). Tested with `common` (counter test with the original version worked), new version just hangs before any output.

Not dropping here for you to debug, might just be our homework to do with the bun updates adjusting our code base. But just to drop the information here.

holgerd77:
Resolved merge conflict here and updated this via UI

holgerd77:
Update: gave this another try.

There is an issue open on Vitest support over on Bun which is still open: https://github.com/oven-sh/bun/issues/4145 and states that there is something in Bun still missing for this to work:

grafik


So seems it is still too early to try on this route.

holgerd77:
What was an interesting experiment though: I managed to switch over our [transaction test runner](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/test/transactionRunner.spec.ts) (which takes quite some time, >30 sec) to use the Bun native test runner (see [docs](https://bun.sh/docs/test/writing)) by doing some few replacements:

```typescript
// import { assert, describe, it } from 'vitest'
import { describe, expect, test } from 'bun:test'

// assert.ok(!txIsValid, `Transaction should be invalid on ${forkName}`)
expect(!txIsValid).toBeTruthy()

// assert.ok(hashAndSenderAreCorrect && txIsValid,`Transaction should be valid on ${forkName}`)
expect(hashAndSenderAreCorrect && txIsValid).toBeTruthy()

// assert.ok(shouldBeInvalid, `Transaction should be invalid on ${forkName}`)
expect(shouldBeInvalid).toBeTruthy()

//assert.fail(`Transaction should be valid on ${forkName}`) (no replacement)
```

This produces the following output:

grafik

Test execution times are 31.11s with Bun and 32.97s with Vitest.

PR #3224, opened 2024-01-09
Internalize `crc`

acolytec3:
This duplicates the `crc32` code from the `crc` dep into an internal utility.


packages/common/src/crc.ts
```diff
@@ -0,0 +1,52 @@
...
+  0xbdbdf21c, 0xcabac28a, 0x53b39330, 0x24b4a3a6, 0xbad03605, 0xcdd70693, 0x54de5729, 0x23d967bf,
+  0xb3667a2e, 0xc4614ab8, 0x5d681b02, 0x2a6f2b94, 0xb40bbe37, 0xc30c8ea1, 0x5a05df1b, 0x2d02ef8d,
+]
+
+if (typeof Int32Array !== 'undefined') {
+  TABLE = new Int32Array(TABLE)
+}
+
+const crc = (current: Uint8Array, previous?: number) => {
+  let crc = previous === 0 ? 0 : ~~previous! ^ -1
```
acolytec3, line 41: The one thing I don't get is how this works if `previous` is undefined.  The original version of this code just asserts it's not undefined but I'm not sure .  Will work on some tests.

.vscode/settings.json
```diff
@@ -1,6 +1,6 @@
 {
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
```
jochem-brouwer, line 3: What does this do?

packages/common/src/crc.ts
jochem-brouwer: `~~undefined` evaluates to `0` it seems. (`!` is a typescript operator (?))

packages/common/src/crc.ts
jochem-brouwer, line 1: I am assuming this file is copied from the original package, I assume a license/source code pointer should be added?

packages/common/src/crc.ts
scorbajio: Could you link to the original version to see how it was being done.

scorbajio (commented):
Always nice to drop unnecessary dependencies 🙂. Might be worth checking the results of using a WASM version.

.vscode/settings.json
acolytec3: I have no idea.  It just keeps getting auto updated in my local vscode settings.  Lemme remove it.  Didn't mean to commit but it just keeps showing up.

packages/common/src/crc.ts
acolytec3: Ugh, sorry, I had that in here but somehow deleted it.  Will re-add.

packages/common/src/crc.ts
acolytec3: https://github.com/alexgorbatchev/crc/blob/master/src/calculators/crc32.ts  It's this file, I copied basically unmodified other than for adding some type annotations that were missing.

scorbajio (approved):
ACK

packages/common/src/crc.ts
jochem-brouwer: Just saw this got merged wanted to add a super small nit: now the reference points to a GH file, but we should have used a permalink to that file instead (if it moves or changes it will not match what is there in the future))

PR #3226, opened 2024-01-10
Monorepo: Embed Code Examples in README Files

holgerd77:
This is a PR to replace our manual example writing in README files - which is error prone and produces outdated examples over time - with a systematic approach to put all examples in a self-contained way (so: with all necessary imports mainly) in the `examples` folder and then embed the examples in the README files with a tool.

As a tool https://github.com/zakhenry/embedme is used, which seems to work well judging after a first test, see 6602424c8c793bbfaa29a7f4e8b98006ae53ecb8 for the first embed (full view is now a bit bloated with all the code block format changes). There is then a new `package.json` `examples:build` script added for each package. This script can be used to re-generate the `README.md` file after an example is altered in the `examples` folder.

Note that from now on examples should generally be only altered (in a first round) in the `examples` folder, followed by a `npm run examples:build` run.

Side note: PR also changes the `typescript` marking for TypeScript code examples to `ts` - which should work in GitHub as well (and is generally a bit more convenient to write) - since `embedme` was not detecting the long version.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,7 @@
         "@vitest/coverage-v8": "0.32.2",
         "@vitest/ui": "0.32.2",
         "c8": "7.12.0",
+        "embedme": "1.22.1",
         "eslint": "8.45.0",
         "eslint-config-prettier": "8.8.0",
         "eslint-config-typestrict": "1.0.5",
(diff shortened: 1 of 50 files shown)


acolytec3 (approved):
LGTM. Will merge and continue today

acolytec3 (approved):
LGTM. Will merge and continue today

PR #3230, opened 2024-01-12
Use permalink to avoid future breakage of link

scorbajio:
Use permalink in `crc.js` reference to avoid future breakage of link.


holgerd77 (approved):
LGTM

Issue #3218, opened 2024-01-08
Trie: Analyze `readable-steam` Dependency Removal/Replacement Options


holgerd77:
The [readable-stream](https://www.npmjs.com/package/readable-stream) package is used relatively low in our stack (in trie [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/trie/package.json)) and has a significant amount of dependencies (5!, including beauties like `buffer`). I followed all and the situation seems not totally out of control, all end relatively soon in a non-dependency subdependency.

Nevertheless: suboptimal from a security perspective and at least close to a needs-somewhat-urgent-action situation.

We should analyze and weight our options here. Functionality is directly exposed by trie in the `createReadStream()` method and then used in the state manager in `dumpStorage()`. So, relatively isolated, complete removal is nevertheless problematic.

There is a [Web Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) existing, Node.js is also [supporting this](https://nodejs.org/api/webstreams.html#web-streams-api) since I think v16+.

Maybe it is therefore an option to use web streams as a replacement. Not sure on both 1) how long the API already exists in the JavaScript/Browser world and 2) what happens on transpilation/compilation with an import like `const { ReadableStream } = require('node:stream/web')` ? 🤔 Will this automatically produce cross-plattform compatible code? (likely not I would say? but maybe one can integrate in such a way?).

Anyhow. This as a first problem description. Would be good if this gets picked up at some point soon (or someone at least contributes by bringing the analysis further down the road).

Issue #3216, opened 2024-01-08
Breaking/Deprecation Tracking Issue


holgerd77:
Tracking issue for breaking changes and deprecation tasks for an eventual round of breaking releases (nothing planned yet). A release round would follow the `v7` releases from #2945 (Summer 2023).

### @ethereumjs/common

- [ ] Get rid of `networkId` (superseeded by `chainId`)

jochem-brouwer:
To align the JSON RPC names we should consider the official (?) spec for the names: https://github.com/ethereum/execution-apis/tree/main/src/eth

jochem-brouwer:
There is something which I keep noticing, and keep getting confused about:

https://github.com/ethereumjs/ethereumjs-monorepo/blob/d8f7544bf86231781b178f936459d96ddaf3e3a1/packages/vm/src/runTx.ts#L264-L267

This is NOT the base fee as in EIP-1559, but it is the "base fee" of the transaction. This is the cost of the tx (21000) plus the data fee (depends on calldata) and extra fee if this is a CREATE transaction. However we also have this `tx.getBaseFee()` method (which returns this). We should definitely rename this to something else, because this is super confusing. (Will add to tx)

jochem-brouwer:
Linking https://github.com/ethereumjs/ethereumjs-monorepo/issues/3479 (make `getContractStorage` not throw for non-existing accounts)

acolytec3:
> There is something which I keep noticing, and keep getting confused about:
> 
> https://github.com/ethereumjs/ethereumjs-monorepo/blob/d8f7544bf86231781b178f936459d96ddaf3e3a1/packages/vm/src/runTx.ts#L264-L267
> 
> This is NOT the base fee as in EIP-1559, but it is the "base fee" of the transaction. This is the cost of the tx (21000) plus the data fee (depends on calldata) and extra fee if this is a CREATE transaction. However we also have this `tx.getBaseFee()` method (which returns this). We should definitely rename this to something else, because this is super confusing. (Will add to tx)

I think this is what EIPs usually refer to as `intrinsic gas` or something like that.  That name isn't any more obvious but at least if we use that term, it would align us with broader core dev usage.

jochem-brouwer:
Ok, I will open a PR to make these terms more clear :) I'll check the EIPs what terms they use but I think I indeed remember `instrinsic gas`.

PR #3231, opened 2024-01-12
Readable stream analysis refactor

scorbajio:
This PR looks into options for the issues discussed in #3218.


packages/statemanager/src/stateManager.ts
```diff
@@ -962,21 +962,14 @@ export class DefaultStateManager implements EVMStateManagerInterface {
...
     const trie = this._getStorageTrie(address, account)
+    const storage: StorageDump = {}
+    const stream = trie.createReadStream()
 
-    return new Promise((resolve, reject) => {
-      const storage: StorageDump = {}
-      const stream = trie.createReadStream()
+    for await (const chunk of stream) {
+      storage[bytesToHex(chunk.key)] = bytesToHex(chunk.value)
+    }
```
holgerd77, line 970: I haven't looked too deeply, a general thing though:

We need to keep the API intact (aka: backwards compatible), so from `createReadStream()`, so the return type is therefore not allowed to change.

If I interpret the code above correctly (showing a usage example of `createReadStream()` it *does* here though (it does, right? so the new stream API is realized with async calls, the old one with events, if I see correctly?).

Just to clarify: I also like the new API a lot better.

We nevertheless need to find a way to remain compatibilty, otherwise things will break for users using the method in the old way.

Two eventual options:

1. Can we use the web stream API in the way the old one worked (just asking for completeness, I would otherwise assume: likely not)?
2. If not, we likely need to bite the sour apple and keep the old method "as is", keep the dependency for now, add a new method with an eventually temporary name (`createWebReadStream()`, `createAsyncReadStream()`,?) and then we add this whole thing to #3216 to remember to fully remove the dependency along the next breaking release round and eventually rename the new method name to the old one.

If we go 2. (guess that's likely?) please also add a `@deprecated` note to the code docs of the old method.

packages/trie/src/trie.ts
```diff
@@ -43,6 +43,8 @@ import type {
 import type { OnFound } from './util/asyncWalk.js'
 import type { BatchDBOp, DB, PutBatch } from '@ethereumjs/util'
 import type { Debugger } from 'debug'
+// eslint-disable-next-line implicit-dependencies/no-implicit
+import type { ReadableStream } from 'node:stream/web'
```
holgerd77, line 47: Maybe also for others to chime in (@jochem-brouwer @acolytec3, everyone): will this type import affect browser and or other runtime support (bun, deno) in any way?

We generally have the policy to not add Node.js specific package imports. Does this also go for types? I am a bit unsure here. For web it shouldn't make a difference since web is using JavaScript (?). For bun I tested a trie package test run with `bunx vitest run test`. This does work, not sure if vitest "shields" something here.

I am totally undecided. Would it eventually make sense to spare out this import and rather go for `any`? Or are we on the safe side here?

holgerd77 (commented):
Awesome, thanks for exploring so quickly, this generally looks great and will be a substantial improvement on the library. :-) (my new MacOS update killed the emojy usage, lol, at least in Firfox).

Two very general things to discuss around backwards compatibility and Node.js specific imports, more for the whole team (at least the second one).

packages/trie/src/trie.ts
jochem-brouwer: I am no browser expert here but I would assume that bun / deno do not have access to this `node:stream/web` package, however your `bunx vitest run test` indeed seems to pass. However I would assume that this would fail on browser builds, which does not seem to be the case (and the CI browser tests also pass). I have tried `bunx vitest run --config=./vitest.config.browser.ts --browser.name=chrome --browser.headless`, this fails all tests:

`Error: Module "events" has been externalized for browser compatibility. Cannot access "events.EventEmitter" in client code.  See https://vitejs.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility for more details.
`

However this does not seem to be related to `ReadableStream` :thinking:

holgerd77: The point here is: since this is a web API, this import should just not be needed in the browser, and the code work nevertheless. I’m am just not sure how these cases are handled in practice. It would (eventually?) e.g. make sense if browsers just „overlook“ (aka: ignore) such an import.

In this case it is even a type import which should go away for JavaScript anyhow (re-read my message above). So generally I would assume we are on the safe side here.

@scorbajio maybe do add a comment above this import shortly explaining why this can go as an exception and stating that we normally do not do Node.js imports in our libs (not that this serves other devs as a reference to conclude: Node.js allowed). A date for the comment also can’t hurt to get some context.

(and the alternative is still that we go for „any“ here, might also be a justified choice in this case)

packages/statemanager/src/stateManager.ts
scorbajio: Yes, it seems like the web API differs from `readable-stream` in that it uses async calls instead of events. I just took a deeper look and can confirm that there isn't an option to use events with the new web API, so I added the `@deprecated` note and made both options available to eventually update it in the future with a breaking release.

packages/trie/src/trie.ts
scorbajio: I added a note detailing the choice to deviate from our import policy and kept the typing.

packages/statemanager/src/stateManager.ts
holgerd77: 👍

packages/trie/src/trie.ts
holgerd77: Thanks, great comment! 🙏

holgerd77: Also have added to the deprecation issue #3216 as well

holgerd77 (approved):
Nice, LGTM!

packages/trie/src/trie.ts
yann300: Hello @scorbajio @holgerd77 , we current have an issue building for the browser using webpack:
```
ERROR in node:stream/web
Module build failed: UnhandledSchemeError: Reading from "node:stream/web" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
```
could this be related to this change and if yes what would you recommend to make it build for the browser?
Thanks very much!

holgerd77: @yann300 ah, we were arguing if this might bite us or not 😬 😛 (as seen above), seems we now have the confirmation.

@scorbajio can you prepare a short PR where we replace this with `any` and remove the import, as discussed above? That would be nice!

@yann300 Guess we'll make some bugfix releases relatively soon, maybe give us a couple more days, there might be 2-3 other things we would want to include as well.

scorbajio: I've raised #3280 to address this.

packages/trie/src/util/readStream.ts
```diff
@@ -1,3 +1,5 @@
+// eslint-disable-next-line implicit-dependencies/no-implicit
+import { ReadableStream } from 'node:stream/web'
```
yann300, line 2: fyi @scorbajio @holgerd77 the import is also present here.

packages/trie/src/trie.ts
scorbajio: @yann300 Would you be able to provide the steps to reproduce this webpack build error so we can explore the problem and potential fixes?

PR #3232, opened 2024-01-12
Devp2p: Use WASM Crypto

holgerd77:
This branch is targeted towards the main WASM crypto branch, extracting the changes on the devp2p library since they are a bit more complex than the rest.

WIP, to be merged (into the WASM branch) at any time sometime feels comfortable with or needs this for continued work.

This will give a significant speed up to the devp2p workings and make the networking "layer" in the client lighter.


acolytec3 (approved):
LGTM

Issue #3234, opened 2024-01-16
Monorepo examples rewrite


acolytec3:
The work of rewriting our code examples in the various READMEs as actual scripts that are accurate that then get embedded into the readmes using `embedme` was begun with #3226 and #3228 but should be finished up.  Below is the list of what's been done and what's left to be finished.  

For each package, we need to add a new script to `package.json` for `"examples:build": "npx embedme README.md",` and then update all the code blocks in the main README.md for that package to use `embedme`.  Reference previous PRs for ways that we've modified/updated examples to reflect actual usage.

- [x] `block` #3226
- [x] `blockchain` #3228
- [x] `client` - NA
- [x] `common`  #3228
- [x] `devp2p` #3256 
- [x] `ethash` #3256 
- [x] `evm` #3242
- [x] `genesis` #3242
- [x] `rlp` #3242
- [x] `statemanager` #3235
- [x] `trie` #3235
- [x] `tx` #3242
- [x] `util` #3226
~~- [ ] `verkle` -- may not be needed at this point~~
- [x] `vm` #3250
- [x] `wallet` #3250

PR #3235, opened 2024-01-17
Embedme integration: Update StateManager and Trie examples

scorbajio:
Updates examples and README files for the `trie` and `statemanager` packages as part of #3234.


packages/statemanager/README.md
```diff
@@ -13,7 +13,7 @@
 
 To obtain the latest version, simply require the project using `npm`:
 
-```shell
+```bash
```
holgerd77, line 16: I can't follow the logic why this should be updated for `embedme` TBH, since these small install notes have totally nothing to do with this conversion.

My tendency would be revert this, this is only causing (small) harm by cluttering find-and-replace opportunities in the future, also `shell` is the more generic term I would think (since `bash` is just one certain shell flavor).

(eventually you can completely kill out the full commit to make it cleaner, by doing a `git rebase -i master` and then deleting out the commit and force push the branch)

holgerd77 (commented):
Can this PR please get a reference to the new issue Andrew opened, and also a small initial description what the PR is doing + a more descriptive title not so focusing on the (at the end random) tool but what is really "done" here? Thanks! (otherwise it's always tedious later on to associate PRs with the work that has been done + one always need to re-visit the PR and go into the code base to see).

Also the example files itself - being the main part of the integration - are still missing, but that's like because it's still in WIP? So just mentioning.

packages/statemanager/README.md
scorbajio: Fixed

holgerd77: Ah, yes, `sh` should be good, but then do it monorepo-wide with a search-and-replace.

holgerd77:
@scorbajio you have seen my other comments from above, right?

scorbajio:
> @scorbajio you have seen my other comments from above, right?

Ah yes, fixed. And I will follow up on this one tomorrow I spent most of today on Snap Sync 🙂.

packages/statemanager/README.md
scorbajio: I realized since we're not embedding any of the shell examples, it isn't entirely necessary to change shell annotations to an annotation that embedme understands.

packages/statemanager/examples/basicUsage.ts
```diff
@@ -0,0 +1,14 @@
...
+import { hexToBytes } from '@ethereumjs/util'
+
+const main = async () => {
+  const stateManager = new DefaultStateManager()
+  const address = new Address(hexToBytes('0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b'))
+  const account = new Account(BigInt(0), BigInt(1000))
+  await stateManager.checkpoint()
+  await stateManager.putAccount(address, account)
+  await stateManager.commit()
+  await stateManager.flush()
```
acolytec3, line 12: can you add a console log after this flush where you do something like `console.log(`Balance for ${address.toString()} is - ${(await stateManager.getAccount(address)).balance})``.  That may not be exactly right but retrieve the account from the state manager and then log out it's balance or something to show full usage?

packages/statemanager/examples/evm.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const main = async () => {
+  try {
+    const provider = 'https://path.to.my.provider.com'
+    const blockchain = new RPCBlockChain(provider)
+    const blockTag = 1n
+    const state = new RPCStateManager({ provider, blockTag })
+    const evm = new EVM({ blockchain, stateManager: state })
+  } catch (e) {
+    console.log(e)
```
acolytec3, line 12: Maybe log just the error message here and add some comments in the example explaining why it failed (i.e. `fetch should fail unless you replace with a real provider url string`)

packages/statemanager/examples/rpcStateManager.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const main = async () => {
+  try {
+    const provider = 'https://path.to.my.provider.com'
+    const stateManager = new RPCStateManager({ provider, blockTag: 500000n })
+    const vitalikDotEth = Address.fromString('0xd8da6bf26964af9d7eed9e03e53415d37aa96045')
+    const account = await stateManager.getAccount(vitalikDotEth)
+    console.log('Vitalik has a current ETH balance of ', account?.balance)
+  } catch (e) {
+    console.log(e)
```
acolytec3, line 12: like `evm.ts`, can you log out just the error message and not the entire stacktrace and add a comment about "please replace `provider` with a valid rpc url string"?

packages/trie/examples/customLevelDB.ts
```diff
@@ -0,0 +1,130 @@
...
+    return new LevelDB<TKey, TValue>(this._leveldb)
+  }
+
+  open() {
+    return this._leveldb.open()
+  }
+}
+
+async function main() {
+  const trie = new Trie({ db: new LevelDB(new Level('MY_TRIE_DB_LOCATION') as any) })
```
acolytec3, line 128: can you add some simple console log to show this works (maybe log out a DB property or something with the db's storage path or something)?

packages/trie/examples/staticCreate.ts
```diff
@@ -0,0 +1,11 @@
+import { Trie } from '@ethereumjs/trie'
```
acolytec3, line 1: What is the difference between this and `basicUsage.ts`?  Seems duplicative and maybe we can just remove this one?

acolytec3 (changes requested):
Looks good overall.  Have noted a couple of places where it would be nice to provide friendlier console logging explaining what's happening.

packages/statemanager/README.md
```diff
@@ -62,33 +67,63 @@ Have a loot at the extended `CacheOptions` on how to use and leverage the new ca
...
+  )
+
+  await stateManager.putContractCode(contractAddress, byteCode)
+  await stateManager.putContractStorage(contractAddress, storageKey1, hexToBytes('0x01'))
+  await stateManager.putContractStorage(contractAddress, storageKey2, hexToBytes('0x02'))
+
+  const proof = await stateManager.getProof(contractAddress)
+  const proofWithStorage = await stateManager.getProof(contractAddress, [storageKey1, storageKey2])
+
+  const partialStateManager = await DefaultStateManager.fromProof(proof)
```
jochem-brouwer, line 95: (Inner check): if I would instantiate `fromProof(proofWithStorage)` then I should be able to retrieve the contract + 2 storage slots... right?

packages/statemanager/README.md
```diff
@@ -62,33 +67,63 @@ Have a loot at the extended `CacheOptions` on how to use and leverage the new ca
...
+  // To add more proof data, use `addProofData`
+  await partialStateManager.addProofData(proofWithStorage)
+  const accountFromNewSM = await partialStateManager.getAccount(contractAddress)
+  const accountFromOldSM = await stateManager.getAccount(contractAddress)
+  console.log(accountFromNewSM, accountFromOldSM) // should match
+
+  const slot1FromNewSM = await stateManager.getContractStorage(contractAddress, storageKey1)
+  const slot2FromNewSM = await stateManager.getContractStorage(contractAddress, storageKey1)
```
jochem-brouwer, line 103: I think I wrote this example, but this line is exactly the same as the line above, so it will be equal anyways

packages/statemanager/README.md
```diff
@@ -62,33 +67,63 @@ Have a loot at the extended `CacheOptions` on how to use and leverage the new ca
...
+  console.log(slot1FromNewSM, slot2FromNewSM) // should also match
+}
+main()
 ```
 
 ### `RPCStateManager`
 
 First, a simple example of usage:
```
jochem-brouwer, line 111: Is there a ref of what RPCStateManager actually does?

jochem-brouwer: (it is unclear: we know that over RPC we now request unknown data, but via this example this is not clear)

packages/statemanager/README.md
```diff
@@ -101,14 +136,20 @@ The `RPCStateManager` can be be used with any JSON-RPC provider that supports th
...
-const blockchain = new RPCBlockChain({}, provider)
-const blockTag = 1n
-const state = new RPCStateManager({ provider, blockTag })
-const evm = new EVM({ blockchain, stateManager: state })
+const main = async () => {
+  const provider = 'https://path.to.my.provider.com'
+  const blockchain = new RPCBlockChain(provider)
+  const blockTag = 1n
+  const state = new RPCStateManager({ provider, blockTag })
+  const evm = new EVM({ blockchain, stateManager: state })
```
jochem-brouwer, line 150: Likely we need some note around here that now EVM is ready to run BLOCKHASH opcodes (over RPC)

packages/statemanager/examples/fromProofInstantiation.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  const partialStateManager = await DefaultStateManager.fromProof(proof)
+  // To add more proof data, use `addProofData`
+  await partialStateManager.addProofData(proofWithStorage)
+  const accountFromNewSM = await partialStateManager.getAccount(contractAddress)
+  const accountFromOldSM = await stateManager.getAccount(contractAddress)
+  console.log(accountFromNewSM, accountFromOldSM) // should match
+
+  const slot1FromNewSM = await stateManager.getContractStorage(contractAddress, storageKey1)
+  const slot2FromNewSM = await stateManager.getContractStorage(contractAddress, storageKey1)
```
jochem-brouwer, line 32: as noted: line here is the same as the line above, so these two are equal

packages/trie/README.md
```diff
@@ -95,14 +110,22 @@ Starting with the v6 release there is a new API for walking and iterating a trie
...
+async function main() {
+  const trie = await Trie.create()
+  await trie.put(utf8ToBytes('key'), utf8ToBytes('val'))
+  const walk = trie.walkTrieIterable(trie.root())
 
-for await (const { node, currentKey } of walk) {
-  // ... do something i.e. console.log( { node, currentKey } )
+  for await (const { node, currentKey } of walk) {
+    // ... do something
+    console.log({ node, currentKey })
```
jochem-brouwer, line 125: I am unable to deduce what this walk does... what does it do? It logs some nodes and key values..?

packages/trie/README.md
```diff
@@ -120,12 +143,12 @@ If you want to use an alternative database, you can integrate your own by writin
 As an example, to leverage `LevelDB` for all operations then you should create a file with the [following implementation from our recipes](./recipes//level.ts) in your project. Then instantiate your DB and trie as below:
```
jochem-brouwer, line 143: Unrelated, but is this URL correct?

packages/trie/README.md
```diff
@@ -137,11 +160,18 @@ By default, the deletion of trie nodes from the underlying database does not occ
...
 
-const trie = await Trie.create({
-  useRootPersistence: true,
-})
+async function main() {
+  const trie = await Trie.create({
+    useRootPersistence: true,
+  })
+  console.log(bytesToHex(trie.root())) // 0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421
+}
```
jochem-brouwer, line 173: This example does not explain what this actually does (it saves the trie root, which in this example is the empty trie root, but it could also be some "weird other value")

packages/trie/README.md
```diff
@@ -155,66 +185,47 @@ The `createProof` and `verifyProof` functions allow you to verify that a certain
...
-  await trie.put(utf8ToBytes('test'), utf8ToBytes('one'))
-  const proof = await trie.createProof(utf8ToBytes('test'))
-  const value = await trie.verifyProof(trie.root(), utf8ToBytes('test'), proof)
-  console.log(value ? bytesToUtf8(value) : 'not found') // 'one'
-}
+// ./examples/proofs.ts#L12-L16
 
-test()
+// proof-of-inclusion
+await trie.put(t1, v1)
```
jochem-brouwer, line 191: The `t1, v1` values are defined in the example but I do not see them here? Is my rendering wrong?

packages/trie/README.md
jochem-brouwer: It is

packages/trie/README.md
jochem-brouwer: image
I get this when I check GH

packages/trie/README.md
acolytec3: Paging @ScottyPoi on this one.  He wrote this example (but I believe the short answer is yes)

packages/trie/README.md
acolytec3: If you look at the `// ./examples/proofs.ts#L...`, we're just doing a snippet here, I would be of the opinion that this is okay if we add some explanatory comments on what the values are.  Otherwise, we are repeating the example in full multiple times.

packages/statemanager/README.md
scorbajio: I am currently checking this. The storage is present but the bytecode is not included in the proof so you cannot retrieve the contract code.

packages/statemanager/examples/evm.ts
scorbajio: Fixed

packages/statemanager/examples/rpcStateManager.ts
scorbajio: fixed

packages/trie/examples/customLevelDB.ts
scorbajio: fixed

packages/trie/examples/staticCreate.ts
scorbajio: fixed

packages/statemanager/README.md
```diff
@@ -34,17 +34,22 @@ It also includes a checkpoint/revert/commit mechanism to either persist or rever
 #### Usage example
 
 ```ts
+// ./examples/basicUsage.ts
```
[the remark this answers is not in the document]
scorbajio: I added a note to the example.

packages/statemanager/README.md
scorbajio: The second line should be looking up `storageKey2`. Fixed.

packages/statemanager/README.md
scorbajio: I agree it's unclear, so I moved the explanation for `RPCStateManager` to before the example.

packages/statemanager/README.md
scorbajio: Added.

packages/statemanager/examples/fromProofInstantiation.ts
scorbajio: The second line should be looking up `storageKey2`. Fixed.

packages/trie/README.md
scorbajio: Yes, it seems like it's a short example of how a trie can be walked. The trie only has a single node that is logged.

packages/trie/README.md
scorbajio: I added some more details on what's happening in the example.

packages/trie/README.md
scorbajio: I've switched all variable names to reflect key-values, so it's now `k1` and `v1` instead. This should rely on the classic key-value convention of naming and the details of what exactly the values assigned are can be omitted.

holgerd77 (approved):
🙏 This will so so help on the stability of our API to have all these examples running CI and being self-contained, also help users dramatically to just have copy-paste examples, thanks a lot for all this work, couldn't be happier! 🤩 

I will now take this in, enough work and updates being done here given the fact that these are "only" examples (which mostly never worked before 😂 ).

If there is anything left to be done please either open a small issue (whoever complains 🙂) so we can take this in in a separate step.

Thanks a lot Amir for the persistance and working this through!

PR #3238, opened 2024-01-17
tx: add `addSignature` method

jochem-brouwer:
This PR adds `addSignature` method. It is essentially the old `protected _processSignature` method.

This allows users to add raw values of the signature, without having to use `tx.sign` (where the pkey is necessary). This is, for instance, handy for hardware wallets.

Please take care reviewing the Legacy transactions, since those are somewhat exceptional.

Ref: https://github.com/ethereumjs/ethereumjs-monorepo/discussions/3229


acolytec3 (approved):
LGTM!

holgerd77:
@vdamle would be great if you can get some feedback here if you feel good with the API and implementation! 🙏

vdamle:
This looks great, @jochem-brouwer @holgerd77 ! Love the handy `convertV` flag in the API - thanks for the quick turnaround! 🙏🏼

PR #3239, opened 2024-01-18
block: validate transactions are signed

jochem-brouwer:
For some reason we have had for many years (?) now that blocks are accepted with unsigned txs.

I have added this to the `block.validateData`.

However, I'm not sure here, should we instead add this to `baseTransaction.getValidationErrors`? We have:

```typescript
  /**
   * Validates the transaction signature and minimum gas requirements.
   * @returns {string[]} an array of error strings
   */
  getValidationErrors(): string[] {
    const errors = []

    if (this.isSigned() && !this.verifySignature()) {
      errors.push('Invalid Signature')
    }

    if (this.getBaseFee() > this.gasLimit) {
      errors.push(`gasLimit is too low. given ${this.gasLimit}, need at least ${this.getBaseFee()}`)
    }

    return errors
  }
```

Shouldn't this throw for unsigned txs? Or is there use somewhere for the simple check to see if the `this.gasLimit` is at least the `this.getBaseFee()` (not to be confused with EIP1559 base fee, this is the tx base fee of 21k for normal txs or 53k for contract-creation txs)

### DIFF
--- a/.github/workflows/client-build.yml
+++ b/.github/workflows/client-build.yml
@@ -33,7 +33,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: 'npm'
 
-      - run: npm ci --omit=peer
+      - run: npm ci
         working-directory: ${{github.workspace}}
 
       - run: npm run coverage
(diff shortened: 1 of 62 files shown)


holgerd77:
I think we should not add to tx but stick with block. This very much sounds like one of these situations where we do not know how others are using this validation method, and the risk of breaking something for users is too high.

jochem-brouwer:
Ok, let's stick to block. Seems like in client/vm we have some blocks which have unsigned txs, so I will fix that.

acolytec3 (approved):

acolytec3 (approved):
LGTM

PR #3242, opened 2024-01-18
Update more examples

acolytec3:
Updates examples/README.mds for `tx`, `rlp`, and `evm`, as part of #3234 

Notes:

- In `tx`, I removed the section about `fake transactions` as this doesn't really make any sense to me.
- In `rlp`, I removed the section regarding conversion between `buffers` and `Uint8Arrays` since these methods have been removed from `util`.


holgerd77 (approved):
Cool, looks great :-), you had forgotten to build the tx examples, have done that and pushed an additional commit, guess we can then merge once CI passes!

holgerd77:
Update: ah, but example CI is still failing, we should fix that before merge.

(so please ignore my approval for now)

packages/evm/examples/4844.ts
```diff
@@ -0,0 +1,3 @@
+import { Common, Chain, Hardfork } from '@ethereumjs/common'
+
+const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Shanghai, eips: [4844] })
```
jochem-brouwer, line 3: Shouldn't we point this to `hardfork: Hardfork.Cancun?`
```suggestion
const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Cancun })
```

packages/evm/README.md
jochem-brouwer, line 1: Missing: `decode-opcodes.cts`, `runCode.cts`

packages/rlp/examples/simple.ts
jochem-brouwer, line 1: This is copied from `browser.html` (which is fine), is it also possible to auto-integrate these examples in `browser.html`? (Not necessarily in this PR)

packages/tx/README.md
```diff
@@ -359,7 +389,7 @@ const eth = new Eth(transport)
 const common = new Common({ chain: Chain.Sepolia })
 
 let txData: any = { value: 1 }
-let tx: Transaction | FeeMarketEIP1559Transaction
+let tx: LegacyTransaction | FeeMarketEIP1559Transaction
```
jochem-brouwer, line 392: Should this be moved into an example file as well?

packages/tx/README.md
jochem-brouwer, line 1: Missing in this README:

- `custom-chain.tx.cts`
- `ropsten-tx.cts` (should be removed? Ropsten is offline)
- `transactions.cts`

jochem-brouwer (commented):
Some nits

packages/evm/examples/4844.ts
acolytec3: I don't think so.  The idea is to activate an EIP that isn't active on that hardfork and 4844 should be active on Cancun.

packages/tx/README.md
acolytec3: Ideally, yes, but in order to make it work we need to install that ledger dependency.  Let me add it as a dev dependency and see if it even works (given that the whole idea of this example would require someone to have a ledger handy)

packages/tx/README.md
acolytec3: These weren't part of the README examples to begin with though I will look at them and see if we should incorporate them for whatever reason.

packages/evm/README.md
acolytec3: Have renamed.

packages/rlp/examples/simple.ts
acolytec3: That's a really interesting idea, though agree it's beyond the scope of this PR.

packages/tx/README.md
acolytec3: Okay, so digging into this further means a ton of work to make this example work out of the box. Having looked at the code in the Ledger dep, I think the example is sound as is.  I'd need to implement an entire fake transport class to mimic a Ledger device in order for this to work "as-is".  That might be worth it, but it's probably beyond the scope of this PR.

packages/tx/README.md
acolytec3: So my thinking on these three after looking at them is that they are fine as is, but I don't think it makes sense to add them to the readme.  I'd rather just leave them as is and people can play with them if they want.  The Ropsten one might be work removing though.

packages/evm/examples/4844.ts
jochem-brouwer: I see, but then the description in the README should change I think, since now it is:

```
#### Initialization

To run EVM related EIP-4844 functionality you have to active the EIP in the associated `@ethereumjs/common` library:
```

This should then be changed to something like: if you have an EIP that is not active on this hardfork, but want it activated anyways, then use this `eips` param

packages/tx/README.md
jochem-brouwer: Ok, agreed :+1:

packages/tx/README.md
jochem-brouwer: Ah ok, just looked at the file, it doesnt actually instantiate it at ropsten but just instantiates it as `Custom.common({ chainId: 3})` which then maps to the chain name: `custom-chain`. Maybe rename the example to `custom-chain-id-tx` or something (I know this file is not really in the scope of this PR though)

acolytec3: Sounds good.  Have gone ahead and switched all of these examples from `.cts` to `.ts` and renamed this one.

packages/evm/examples/4844.ts
acolytec3: Fair point, have updated the language in the README on this point.

jochem-brouwer (approved):
LGTM

packages/evm/examples/4844.ts
holgerd77: At the end no big deal and for sure not a hill to die upon, but I’ve also updated similar examples in a way Jochem suggested.

PR #3245, opened 2024-01-19
Additional `wasm-crypto` replacements in devp2p

acolytec3:
This adds additional cryptographic primitives from `@polkadot/wasm-crypto`, mainly for use in devp2p.  Partially addresses #3246

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -30,6 +30,7 @@ import {
 } from '@polkadot/wasm-crypto'
 import * as kzg from 'c-kzg'
 import { keccak256 } from 'ethereum-cryptography/keccak'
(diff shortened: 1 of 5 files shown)


acolytec3:
One question here is do we need to continue to write somewhat duplicative tests for each optional custom crypto function in `common` and then in subsequent consuming libraries like `devp2p`?

holgerd77:
> One question here is do we need to continue to write somewhat duplicative tests for each optional custom crypto function in `common` and then in subsequent consuming libraries like `devp2p`?

I think maybe one test in Common for one crypto primitive that this general mechanism (that Common is keeping the thing) is working should be enough (for Common).

holgerd77 (approved):
After a first look I think this seems ready for review, so will do so, let me know if this was missing something.

LGTM, will merge.

PR #3247, opened 2024-01-20
evm: make profiler more accurate

jochem-brouwer:
This PR makes the EVM profiler more accurate.

In particular, this accounts for entering new CALL frames within the EVM (all variants, also CREATE). It now pauses the timer right before entering the opcode loop and unpauses it when exiting it. This is different than before. It now accounts for the setup time of the whole new call frame (setup new Interpreter, load code, etc.)

Old for `npm run test:state -- --profile --test=ContractCreationSpam`: Execution reports 9603ms, total opcode time 3247ms (so off by more than 6k ms)

New: 11000 ms for execution, total opcode time 10388. (so off by only 612ms)

There are 3 things which are either uncovered or can simply not be measured:

- (can be measured) the initial setup of the first call frame. So, when VM.runTx calls into EVM.runCall, it setups the entire call stack before it actually start measuring opcodes. When measuring has started, call frames internally should now be reported much more precise.
- (can be measured), this code: 

```typescript
      let opCode: number
      let opCodeObj: OpcodeMapEntry
      if (doJumpAnalysis) {
        opCode = this._runState.code[this._runState.programCounter]
        // Only run the jump destination analysis if `code` actually contains a JUMP/JUMPI/JUMPSUB opcode
        if (opCode === 0x56 || opCode === 0x57 || opCode === 0x5e) {
          const { jumps, pushes, opcodesCached } = this._getValidJumpDests(this._runState.code)
          this._runState.validJumps = jumps
          this._runState.cachedPushes = pushes
          this._runState.shouldDoJumpAnalysis = false
          cachedOpcodes = opcodesCached
          doJumpAnalysis = false
        }
      } else {
        opCodeObj = cachedOpcodes![this._runState.programCounter]
        opCode = opCodeObj.opcodeInfo.code
      }
      this._runState.opCode = opCode!

      try {
        await this.runStep(opCodeObj!)
 ```

(which thus calls into `runStep`)

... until we actually start the timer:

```typescript
  async runStep(opcodeObj?: OpcodeMapEntry): Promise<void> {
    const opEntry = opcodeObj ?? this.lookupOpInfo(this._runState.opCode)
    const opInfo = opEntry.opcodeInfo

    let timer: Timer

    if (this.profilerOpts?.enabled === true) {
      timer = this.performanceLogger.startTimer(opInfo.name)
```

I measured this overhead for the example, and I got: execution time 11055ms, opcode time 10218ms and this overhead of 112ms.

- The final thing which is not covered: the actual "measuring" overhead (which we could try to measure, but still, if you measure things this will make things slower)

Added 2 new profiler flags for opcodes:

- Initialization: this is the time taken to setup the first call frame up until it starts looping in the opcodes
- Overhead: this is the overhead as mentioned above in the opcode loop

Picture explaining the timers:

image


jochem-brouwer:
Ok, now also added timers for the initialization and the overhead. However, the reported opcode time is now longer than the reported time that `evm.runCall` should use :thinking: I'm not sure because the Logger has explicit logic that only one timer can be running at a time :thinking:

jochem-brouwer:
I should add the timers to precompiles too, `npm run test:state -- --profile --test=CALLBlake2f_MaxRounds` fails with profiler enabled.

jochem-brouwer:
Tested on client and seems to work.

holgerd77 (approved):
LGTM

Issue #3244, opened 2024-01-19
client: TrieNodeFetcher crashes due to undefined pathString


scorbajio:
During the heal phase of snap sync, the `TrieNodeFetcher` crashes due to an `undefined` `pathString`:
```
2024-01-19T01:13:27.794Z client:TrieNodeFetcher All requests for current heal phase have been filled
2024-01-19T01:13:27.795Z client:TrieNodeFetcher Stored accountTrie with root actual=3372b525c0074182d1743177ce733675dc7262fc0fca13fd664e99b5ca0d5ec3 expected=3372b525c0074182d1743177ce733675dc7262fc0fca13fd664e99b5ca0d5ec3
2024-01-19T01:13:27.795Z client:TrieNodeFetcher At start of store phase
2024-01-19T01:13:27.797Z client:TrieNodeFetcher TypeError: Cannot read properties of undefined (reading 'split')
    at TrieNodeFetcher.store (/home/indigoomega021/code/projects/eth/ethjs-lodestar/ethereumjs-monorepo/packages/client/dist/src/sync/fetcher/trienodefetcher.js:110:63)
    at _write (/home/indigoomega021/code/projects/eth/ethjs-lodestar/ethereumjs-monorepo/packages/client/dist/src/sync/fetcher/fetcher.js:275:32)
    at Writable.writev [as _writev] (/home/indigoomega021/code/projects/eth/ethjs-lodestar/ethereumjs-monorepo/packages/client/dist/src/sync/fetcher/fetcher.js:306:24)
    at doWrite (node:internal/streams/writable:409:12)
    at clearBuffer (node:internal/streams/writable:564:5)
    at onwrite (node:internal/streams/writable:464:7)
    at Writable._write (/home/indigoomega021/code/projects/eth/ethjs-lodestar/ethereumjs-monorepo/packages/client/dist/src/sync/fetcher/fetcher.js:278:17)
2024-01-19T01:13:27.797Z client:TrieNodeFetcher At start of store phase
2024-01-19T01:13:27.798Z client:TrieNodeFetcher TypeError: Cannot read properties of undefined (reading 'split')
    at TrieNodeFetcher.store (/home/indigoomega021/code/projects/eth/ethjs-lodestar/ethereumjs-monorepo/packages/client/dist/src/sync/fetcher/trienodefetcher.js:110:63)
[...]

PR #3252, opened 2024-01-23
client: Check for undefined pathStrings

ScottyPoi:
# package: **client**
## Fix for issue #3244

Also fixes the same error, thrown during sync from the `checkFiilter` method in `client/src/util/rpc.ts`.

- ### `trieNodeFetcher.store` 
  - A possible `undefined` was ignored with a `!` and an `as string`.
  - Fixed by assigning an empty string in place of the possible `undefined` string.
- ### `util/rpc.ts -- checkFilter(method: string, filterStringCSV: string)`
  - Function expects a string value for `filterStringCSV`, but can potentially be fed an `undefined` 
  - This probably happens from the same `as string` issue above, or the same bug somewhere else
  - Fixed by checking for `undefined` and treating it the same as an empty string.

Running the script from the Issue with these two fixes resulted in the elimination of that error, and allowed the `sync` process to run smoothly.

### DIFF
--- a/packages/client/src/sync/fetcher/trienodefetcher.ts
+++ b/packages/client/src/sync/fetcher/trienodefetcher.ts
@@ -211,8 +211,8 @@ export class TrieNodeFetcher extends Fetcher<JobTask, Uint8Array[], Uint8Array> 
       for (const nodeData of result[0]) {
         const node = decodeNode(nodeData as unknown as Uint8Array)
         const nodeHash = bytesToHex(this.keccakFunction(nodeData as unknown as Uint8Array))
-        const pathString = this.requestedNodeToPath.get(nodeHash) as string
-        const [accountPath, storagePath] = pathString!.split('/')
+        const pathString = this.requestedNodeToPath.get(nodeHash) ?? ''
+        const [accountPath, storagePath] = pathString.split('/')
         const nodePath = storagePath ?? accountPath
         const childNodes = []
         let unknownChildNodeCount = 0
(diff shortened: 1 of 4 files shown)


ScottyPoi:
To recreate the error: 

1. clone the lodestar-quickstart repo:
  * `git clone https://github.com/ChainSafe/lodestar-quickstart.git`
  * `cd lodestar-quickstart`
2. run `./setup.sh --network ethjsdev0 --dataDir edev0-data --elClient ethereumjs`

To run the sync script using this PR branch:

1. checkout this branch
  * `git checkout `checkFilter-bug`
2. build a local Docker image with the changes
  * i usually name a local image something like: `ethereumjsmonorepo:debugging` 
  * run `docker build --pull --rm -f "Dockerfile" -t ethereumjsmonorepo:debugging "."`
3. clone the lodestar-quickstart repo:
  * `git clone https://github.com/ChainSafe/lodestar-quickstart.git`
  * `cd lodestar-quickstart`
4. Change the variable `ETHEREUMJS_IMAGE` from `ethpandaops/ethereumjs:master` to the name of the local image (`ethereumjsmonorepo:debugging`)
  * `import-images.sh`
  * `ethjsdev0.vars`
5. In: `setup.sh` - skip the call to `docker pull`
  * comment out this code: 
  * ```    elif [ "$elClient" == "ethereumjs" ]
    then
      $dockerExec pull $ETHEREUMJS_IMAGE ```
6. run `./setup.sh --network ethjsdev0 --dataDir edev0-data --elClient ethereumjs`

acolytec3 (approved):
Great fixes!

PR #3256, opened 2024-01-24
Update `devp2p` and `ethash` examples

acolytec3:
Part of #3234 

This updates the `devp2p` and `ethash` examples to use `embedme` and real, runnable code.  In doing so, I've made the following additional changes:

- Modified an import for a `multiaddr` utility in `enr.ts` so `tsx` can find it since the underlying method isn't exported properly (and we use an archaic version of `multiaddr`)
- Removed some browser `dns` stuff and references in the documentation since we've deprecated client usage in the browser (and therefore completely removed the only use case for running devp2p in the browser).


packages/devp2p/README.md
```diff
@@ -61,22 +61,37 @@ includes node discovery ([./src/dpt/server.ts](./src/dpt/server.ts))
...
+const main = async () => {
+  const dpt = new DPT(PRIVATE_KEY, {
+    endpoint: {
+      address: '0.0.0.0',
+      udpPort: null,
+      tcpPort: null,
+    },
+  })
+  console.log(`DPT is active and has id - ${bytesToHex(dpt.id!)}`)
+  // 0xcd80bb7a768432302d267729c15da61d172373ea03679f3c8156b55453b68668a482d183a27139d7f37cc0c83eb7f414a74d6256d0b2bb20655041b9a3c8e0ed
```
scorbajio, line 79: A bit of a nit but being explicit that this comment is the logged output of the above statement or just moving the comment to the same line since it still flows off the screen when this is rendered as markdown can potentially clear up some confusion about what the hash is.

packages/devp2p/README.md
```diff
@@ -217,20 +241,29 @@ Send the initial status message with `sendStatus()`, then wait for the correspon
...
-eth.events.on('message', async (code, payload) => {
-  if (code === devp2p.ETH.MESSAGE_CODES.NEW_BLOCK_HASHES) {
-    // Do something with your new block hashes :-)
-  }
-})
+// ./examples/peer-communication.ts#L116-L118
+
+eth.events.on('message', async (code: ETH.MESSAGE_CODES, payload: any) => {
+  if (code in ETH.MESSAGE_CODES) {
+    requests.msgTypes[code] = code + 1
```
scorbajio, line 266: If a brief comment about what is happening and why in this line where `code` is being incremented and reassigned in `msgTypes` it would clarify what's happening in response to the message as part of the example.

scorbajio (commented):
This looks great! Left some suggestions for improving clarity at some points.

packages/ethash/examples/miner.ts
```diff
@@ -0,0 +1,20 @@
...
+  },
+  { setHardfork: true, skipConsensusFormatValidation: true }
+)
+
+const cacheDB = new MapDB<number, DBObject>()
+
+const e = new Ethash(cacheDB)
+const miner = e.getMiner(block.header)
+const solution = await miner.iterate(-1) // iterate until solution is found
+console.log(bytesToHex(solution!.mixHash))
```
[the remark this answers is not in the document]
acolytec3: Good point.  Have added.

packages/devp2p/README.md
acolytec3: Fair, I've updated this to truncate the logging and hopefully clarify what's going on.

scorbajio (approved):
ACK

PR #3257, opened 2024-01-25
Add tests for verkle statemanager

scorbajio:
The goal of this PR is to increase coverage for the `StatelessVerkleStateManager`. Also see #3179.


scorbajio:
The only test that fails is one that asserts that getting an empty account from the `StatelessVerkleStateManager` should return and `undefined` value. @gabrocheleau should this test be changed or is this something that should be addressed in the implementation?

packages/statemanager/test/statelessVerkleStateManager.spec.ts
```diff
@@ -43,4 +52,115 @@ describe('StatelessVerkleStateManager: Kaustinen Verkle Block', () => {
...
   })
+
+  it('put/delete/modify account', async () => {
+    const stateManager = new StatelessVerkleStateManager({ common })
+    stateManager.initVerkleExecutionWitness(block.executionWitness)
+
+    const address = new Address(randomBytes(20))
+
+    let check = await stateManager.getAccount(address)
+    assert.isUndefined(check, 'should return undefined for nonexistent account')
```
gabrocheleau, line 63: In our current implementation, we are throwing whenever the account is not found in the statelessStateManager (i.e. if it's not in the cache, and not in the witness). An "non-existent" account would be specified as `null` (or `0x`) in the witness.

packages/statemanager/test/statelessVerkleStateManager.spec.ts
```diff
@@ -43,4 +52,115 @@ describe('StatelessVerkleStateManager: Kaustinen Verkle Block', () => {
...
+    await stateManager.modifyAccountFields(address, {
+      nonce: BigInt(2),
+    })
+    check = await stateManager.getAccount(address)
+    assert.deepEqual(check, account, 'should return correct account')
+
+    await stateManager.deleteAccount(address)
+
+    check = await stateManager.getAccount(address)
+    assert.isUndefined(check, 'should return undefined for deleted account')
```
gabrocheleau, line 82: Again, this will throw. We could modify the tests so that we expect throw. 

Could also be nice to add a test case for the following scenario:

- Account A does not exit
- `null` witness is provided for non-existent account
- getAccount(accountA) should return undefined

packages/statemanager/test/statelessVerkleStateManager.spec.ts
```diff
@@ -43,4 +52,115 @@ describe('StatelessVerkleStateManager: Kaustinen Verkle Block', () => {
...
+        type: CacheType.ORDERED_MAP,
+      },
+      storageCacheOpts: {
+        type: CacheType.ORDERED_MAP,
+      },
+      common,
+    })
+    sm.initVerkleExecutionWitness(block.executionWitness)
+
+    const smCopy = sm.shallowCopy()
```
gabrocheleau, line 127: Our current shallowCopy implementations seems to have been overlooked, as it basically does nothing else than create a new StateManager with the same witness. 

I think we should implement `shallowCopy` similarly than what we have in `stateManager`. I'll tackle that this week

gabrocheleau (commented):
Thanks for working on this!

Rebased this on top of client/kaustinen-syncing so that it uses the most up-to-date implementation of the verkle stateless state manager.

holgerd77:
What is the state of this PR?

holgerd77:
(and can this get a branch update with conflict resolution?)

packages/statemanager/test/statelessVerkleStateManager.spec.ts
scorbajio: I've modified the test to catch the initial thrown error when getting a random address that doesn't exist in the witness, but this assert after a deleted account is got actually is not throwing but returning an `undefined`, currently. 

I will work on more test cases including the suggested one in future PR's as I'm still working on collecting witness data for it.

scorbajio:
> What is the state of this PR?

We can merge this PR if @gabrocheleau agrees, or we could keep working on it to add the mentioned extra test case.

holgerd77:
Updated this via UI

packages/statemanager/test/statelessVerkleStateManager.spec.ts
```diff
@@ -1,9 +1,18 @@
...
+import {
+  Account,
+  Address,
+  bytesToBigInt,
+  bytesToHex,
+  hexToBytes,
+  randomBytes,
+} from '@ethereumjs/util'
+import { getStem } from '@ethereumjs/verkle'
+import { skip } from 'tape'
```
gabrocheleau, line 12: Can we remove the `tape` usage?

packages/statemanager/test/statelessVerkleStateManager.spec.ts
```diff
@@ -43,4 +52,123 @@ describe('StatelessVerkleStateManager: Kaustinen Verkle Block', () => {
...
+    await stateManager.modifyAccountFields(address, {
+      nonce: BigInt(2),
+    })
+    check = await stateManager.getAccount(address)
+    assert.deepEqual(check, account, 'should return correct account')
+
+    await stateManager.deleteAccount(address)
+
+    check = await stateManager.getAccount(address)
+    assert.isUndefined(check, 'should return undefined for deleted account')
```
gabrocheleau, line 90: Minor nit, but instead of re-assigning these to a `check` variable, I would find this more readable if we just inlined the `getAccount` inside of the test comparison for all of these. So something like:

```
 assert.isUndefined(await stateManager.getAccount(address), 'should return undefined for deleted account')
```

packages/statemanager/test/statelessVerkleStateManager.spec.ts
gabrocheleau: Can be replaced by the vitest equivalent: https://vitest.dev/api/#test-skip

packages/statemanager/test/statelessVerkleStateManager.spec.ts
```diff
@@ -43,4 +52,123 @@ describe('StatelessVerkleStateManager: Kaustinen Verkle Block', () => {
...
+    assert.equal(
+      account.nonce,
+      bytesToBigInt(hexToBytes(nonceRaw), true),
+      'should have correct nonce'
+    )
+    assert.equal(bytesToHex(account.codeHash), codeHash, 'should have correct codeHash')
+  })
+
+  it(`copy()`, async () => {
+    const sm = new StatelessVerkleStateManager({
```
gabrocheleau, line 124: Nit, but imo let's name this consistently throughout tests (so stateManager instead of sm) 
```suggestion
    const stateManager = new StatelessVerkleStateManager({
```

packages/statemanager/test/statelessVerkleStateManager.spec.ts
```diff
@@ -43,4 +52,123 @@ describe('StatelessVerkleStateManager: Kaustinen Verkle Block', () => {
...
+      nonce: BigInt(2),
+    })
+    await stateManager.putAccount(address, account)
+
+    let check = await stateManager.getAccount(address)
+    assert.deepEqual(check, account, 'should return correct account')
+
+    await stateManager.modifyAccountFields(address, {
+      nonce: BigInt(2),
+    })
```
gabrocheleau, line 83: I am not sure I understand the purpose of this `modifyAccountFields` and subsequent check (since we're modifying with the value it was already created with). Do we actually want to check that "modifying with the same value does not return a different account"? If so can we describe that, either in the assert description or in a comment?

packages/statemanager/test/statelessVerkleStateManager.spec.ts
scorbajio: fixed

packages/statemanager/test/statelessVerkleStateManager.spec.ts
scorbajio: Fixed.

packages/statemanager/test/statelessVerkleStateManager.spec.ts
scorbajio: Fixed in a separate commit.

packages/statemanager/test/statelessVerkleStateManager.spec.ts
scorbajio: I was actually testing to see if updating by using `modifyAccountFields` is working as expected by updating an arbitrary account field. I have fixed it.

gabrocheleau (approved):

PR #3260, opened 2024-01-27
tx: better chain id error

jochem-brouwer:
Some error messages I found in #3259 were not super helpful, so this PR attempts to improve it.

WIP, will push along more errors which I find

### DIFF
--- a/packages/tx/src/baseTransaction.ts
+++ b/packages/tx/src/baseTransaction.ts
@@ -370,7 +370,9 @@ export abstract class BaseTransaction<T extends TransactionType>
       const chainIdBigInt = bytesToBigInt(toBytes(chainId))
       if (common) {
         if (common.chainId() !== chainIdBigInt) {
-          const msg = this._errorMsg('The chain ID does not match the chain ID of Common')
+          const msg = this._errorMsg(
+            `The chain ID of ${chainIdBigInt} does not match the chain ID of Common (${common.chainId()})`
+          )
           throw new Error(msg)
         }
         // Common provided, chain ID does match


packages/tx/src/baseTransaction.ts
```diff
@@ -370,7 +370,9 @@ export abstract class BaseTransaction<T extends TransactionType>
       const chainIdBigInt = bytesToBigInt(toBytes(chainId))
       if (common) {
         if (common.chainId() !== chainIdBigInt) {
-          const msg = this._errorMsg('The chain ID does not match the chain ID of Common')
+          const msg = this._errorMsg(
+            `The chain ID of ${chainIdBigInt} does not match the chain ID of Common (${common.chainId()})`
```
holgerd77, line 374: In these cases we always tried to keep the original string as before for backwards compatibility reasons (if someone does string comparison, and put the extra info at the end (with „(expected: x, actual: y)“ or similar e.g.).

Think that’s a good practice and we should stick to it.

holgerd77: (So this is not a 100% thing, but it helps if someone e.g. does:)
If „chain ID does not match“ in errorMsg do something

jochem-brouwer: Ah ok I see! I'll update. I will (as a kind of "chore task") go over all our error messages to see if we can improve the error strings (at the end)

holgerd77: Thanks, it’s a bit of a „blurry“ rule, but I remember that people complained in the past when we changed error msgs too aggressively.

packages/blockchain/src/blockchain.ts
```diff
@@ -529,7 +529,9 @@ export class Blockchain implements BlockchainInterface {
         let dbOps: DBOp[] = []
 
         if (block.common.chainId() !== this.common.chainId()) {
-          throw new Error('Chain mismatch while trying to put block or header')
+          throw new Error(
+            `Chain mismatch while trying to put block or header. Block chain ID: ${block.common.chainId}, blockchain chain ID: ${this.common.chainId}`
```
scorbajio, line 533: Very small nit, but `Block chain ID` and `blockchain chain ID` can be confusing and sound like the same thing.
```suggestion
            `Chain mismatch while trying to put block or header. Chain ID of block: ${block.common.chainId}, chain ID of blockchain : ${this.common.chainId}`
```

holgerd77:
Updated this via UI

jochem-brouwer:
This one is ready for review (still)

holgerd77:
Updated this via UI

packages/statemanager/src/stateManager.ts
```diff
@@ -902,7 +902,7 @@ export class DefaultStateManager implements EVMStateManagerInterface {
         32
       )
       if (!equalsBytes(reportedValue, storageValue)) {
-        throw new Error('Reported trie value does not match storage')
+        throw new Error(`Reported trie value does not match storage, key: ${stProof.key}`)
```
acolytec3, line 905: Should we also report the hex values of the reportedValue and storageValue here?

acolytec3 (changes requested):
Looks good over all.  One minor suggestion and a question before approving though.

packages/statemanager/src/stateManager.ts
jochem-brouwer: Yup, will add

holgerd77 (approved):
LGTM

PR #3263, opened 2024-01-30
Remove proofTrie patern

scorbajio:
This change removes the `proofTrie` pattern where a trie is either created just for proof verification or is kept around for multiple proof verifications in the statemanager and snap sync fetchers used in the client. This pattern is no longer necessary since static versions of proof verification functions are available from work done in #2949.


packages/client/src/sync/fetcher/storagefetcher.ts
```diff
@@ -129,13 +127,17 @@ export class StorageFetcher extends Fetcher<JobTask, StorageData[][], StorageDat
...
       const values = slots.map((slot: any) => slot.body)
-      return await this._proofTrie.verifyRangeProof(
+      return await Trie.verifyRangeProof(
         stateRoot,
         origin,
         keys[keys.length - 1],
         keys,
         values,
-        <any>proof
+        <any>proof,
```
acolytec3, line 136: Is there a strong reason we need to cast this to `any` instead of just type casting it as we expect (either as `Uint8Array[]` if we always expect it to be bytes or else `Uint8Array[] | Undefined`)?

acolytec3 (dismissed):
Looks good over all but would prefer to not have the `<any>proof` pattern.  Ideally we cast this to the expected type.

holgerd77:
Would be good if finalizing this gets some priority, would be nice if we take this clean-up in into the releases.

packages/client/src/sync/fetcher/storagefetcher.ts
scorbajio: I've changed it so that the value is converted to the expected type instead of casting as any.

holgerd77:
Updated this via UI

holgerd77:
Removed some remaining `any` castings in 23b924cbb6259fe01b877dbc03242580a9dceed7

holgerd77 (approved):
LGTM

PR #3266, opened 2024-02-01
Client: fix test nesting and other vitest issues

ScottyPoi:
This PR addresses the issue of "nesting" in Vitest test designs, as well as issues regarding test assertions inside of event callbacks.

1. Test Nesting:
- The update to vitest v1.x.x introduces stricter rules around nested tests.
- "Nesting" refers to a test case within a test case
  - *describe* inside of *describe*
  - *it* inside of *it*

**The solution for this is usually to break apart the nested test cases into separate test suites**
i.e.
```ts
describe('All the tests', () => {
  it('test 1', () => {
       it('test 1a', () => {
          assert(...)
        })
       it('test 1b', () => {
          assert(...)
        })
    })
   it('test 2', () => {
      assert(...)
    })
})
```
would be refactored to:
```ts
describe('test 1', () => {
       it('test 1a', () => {
          assert(...)
        })
       it('test 1b', () => {
          assert(...)
        })
})
describe('test 2', () => {
    it('test case', () => {
        assert(...)
    })
})
```

2. "Event Based" test assertions

- Test assertions that occur inside of an event callback MUST be wrapped in a test case (`it`) within the callback function.  Otherwise, the test will pass whether or not the event callback happened or not.  

example: 
```
it('Event Test', () => {
    events.on('test', () => {
            assert.ok(2 + 2 === 5)
        })
    events.emit('test')
})
```
This test will **PASS** when it ***SHOULD* FAIL* 
This test should fail when the 'test event' is triggered, and we assert **2 + 2 === 5**.
However it will pass because the test case will exit without waiting for the assertion.

*We use this pattern *A LOT***
**This means many tests report *passing* when *no actual test assertions were checked**!!!!*

Vitest will exit a test case (**it**) irregardless of whether all or any assertions are called.  
*however*
Vitest will not exit a test suite (**describe**) without all of the test cases (**it**) having run.  

Solution:
```ts
describe('Event Test', () => {
    events.on('test', () => {
        it('runs a test case', () => {
            assert.ok(2+2===5)
        })
    })
})
```ts
This test should and will FAIL (because 2 + 2 is not 5).  By isolating the test case (**it**) inside of the event callback, the test case will be required to run before the test suite exits.

### DIFF
--- a/packages/client/test/integration/beaconsync.spec.ts
+++ b/packages/client/test/integration/beaconsync.spec.ts
@@ -7,88 +7,96 @@ import genesisJSON from '../testdata/geth-genesis/post-merge.json'
 
 import { destroy, setup, wait } from './util'
 
-describe('[Integration:BeaconSync]', () => {
-  const common = Common.fromGethGenesis(genesisJSON, { chain: 'post-merge' })
-  common.setHardforkBy({ blockNumber: BigInt(0), td: BigInt(0) })
+const common = Common.fromGethGenesis(genesisJSON, { chain: 'post-merge' })
+common.setHardforkBy({ blockNumber: BigInt(0), td: BigInt(0) })
 
-  it('should sync blocks', async () => {
-    BlockHeader.prototype['_consensusFormatValidation'] = vi.fn()
-    vi.doMock('@ethereumjs/block', () => {
-      {
-        BlockHeader
-      }
-    })
-
-    const [remoteServer, remoteService] = await setup({ location: '127.0.0.2', height: 20, common })
-    const [localServer, localService] = await setup({ location: '127.0.0.1', height: 0, common })
-    const next = await remoteService.chain.getCanonicalHeadHeader()
-    ;(localService.synchronizer as any).skeleton.status.progress.subchains = [
-      {
-        head: BigInt(21),
-        tail: BigInt(21),
-        next: next.hash(),
-      },
-    ]
-    await localService.synchronizer!.stop()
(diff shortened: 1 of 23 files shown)


ScottyPoi (commented):
Starting with `Client` package, and will refactor other packages in subsequent PR's.
All packages likely need some test refactoring, but `client`, `vm`, and `evm` will need the most.

jochem-brouwer:
Just want to chime in here, but if we manage to update vitest to the latest version, then these "nesting tests" are already built in and then vitest will throw if you try to do that :thinking: I remember having to fix many of those when trying to update :thinking:

jochem-brouwer:
Yes, sorry for the misunderstanding: if this PR manages to reach a (sub)goal of getting to the new vitest version this is all fine! (And just to clarify: this PR looks fine too!). Sorry for any misunderstanding.

ScottyPoi:
It's important to clarify that with the upcoming update to Vitest, the framework will indeed enforce stricter rules regarding **test nesting**. This PR lays the groundwork for the vitest upgrade, addressing broader nesting issues to streamline the subsequent upgrade process, which faced delays due to browser test result inconsistencies.

Regarding the event-related changes, I encourage you to conduct a simple test by substituting a test assertion with assert.fail(). This will demonstrate that certain assertions were not being executed as expected, as the test case will still pass despite the failure assertion.  

The decision to address both the nesting adjustments and the event-related issues within a single PR was due to their interconnected impact on numerous tests.

PR #3268, opened 2024-02-03
Implement EIP 2935

jochem-brouwer:
This PR is based upon the EIP 2935 version of: https://github.com/ethereum/EIPs/pull/8166, which is currently not merged.

Note: this also adds functionality to Common which adds the ability to specify new hardforks, where users can describe when they want these forks to be activated (handy for EIP transitioning tests which have not yet been assigned to a hardfork)


packages/common/src/common.ts
```diff
@@ -779,6 +780,26 @@ export class Common {
     return null
   }
 
+  /**
+   * Returns the hardfork change block for eip
+   * @param eip EIP number
+   * @returns Block timestamp or null if unscheduled
+   */
+  eipTimestamp(eip: number): bigint | null {
+    // NOTE: this is not elegant, what if the eip got activated on a block and not timestamp?
```
acolytec3, line 789: Don't the older hardfork block headers still have a timestamp in them that we can report here?  I'm not sure if this is really a concern.  Or, maybe better yet, we just return `null` for any EIP that's not scheduled or pre-Shanghai.

packages/common/test/customChains.spec.ts
```diff
@@ -151,6 +152,56 @@ describe('[Common]: Custom chains', () => {
       'customChains, should allow to switch custom chain'
     )
   })
+
+  it('customHardforks parameter: initialization and transition tests', () => {
```
acolytec3, line 156: Great test!

packages/vm/src/runBlock.ts
```diff
@@ -376,6 +385,42 @@ async function applyBlock(this: VM, block: Block, opts: RunBlockOpts) {
...
+  if (parentBlock.header.timestamp < forkTime) {
+    let ancestor = parentBlock
+    const range = this.common.param('vm', 'minHistoryServeWindow')
+    for (let i = 0; i < Number(range) - 1; i++) {
+      if (ancestor.header.number === BIGINT_0) {
+        break
+      }
+
+      ancestor = await this.blockchain.getBlock(ancestor.header.parentHash)
+      await putBlockHash.bind(this)(ancestor.hash(), ancestor.header.number)
```
acolytec3, line 419: Just because I hate `.bind`, is there a reason we can't just pass the `vm` in here?  Does it make the code more clunky to do so?

acolytec3 (commented):
Left some initial feedback but looks good in the main.  WIll need to dig into the tests more tomorrow.

packages/common/test/customChains.spec.ts
jochem-brouwer: Thanks :smile:

packages/vm/src/runBlock.ts
jochem-brouwer: Hmm I don't know why I added this :thinking: Let me check if I can make it not use bind :)

packages/common/src/common.ts
jochem-brouwer: Oh right, this is indeed an interesting line of thought. However this would make the method `async` and we would need to get the block from `blockchain` in order to retrieve the timestamp :thinking: Not sure I like that. Especially gets a bit hairy if it reorgs (you save the timestamp, chain reorgs, and now the timestamp differs). I will update the docs to remove the "block" part and just write down that it returns the timestamp when this EIP is scheduled (therefore, on missed slots the first block might actually be later than the scheduled timestamp)

jochem-brouwer: Ahh! Wait! I now remember the point I wanted to make here. The same applies to `eipBlock` actually. If `eipTimestamp` returns `null` (for instance, for `Dao`), then we know it is either (1) not scheduled at a timestamp (but at a block number) or (2) it is not even scheduled at all (so the fork will never happen). I think this is beyond the scope of this PR though and it is also an edge case.

packages/vm/src/runBlock.ts
jochem-brouwer: Ok, I think I added it because `this` as argument in your method apparently does not count as argument and therefore you need to use `bind`, I added it for reading clearlity but the method is so small, I will remove it.

packages/common/test/customChains.spec.ts
holgerd77: Yeah, agree! 🙂

packages/vm/src/runBlock.ts
```diff
@@ -376,6 +385,42 @@ async function applyBlock(this: VM, block: Block, opts: RunBlockOpts) {
   return blockResults
 }
 
+export async function accumulateParentBlockHash(this: VM, block: Block) {
```
holgerd77, line 388: Can this method get some code docs please?

packages/common/src/eips.ts
```diff
@@ -190,6 +191,23 @@ export const EIPs: EIPsDict = {
       },
     },
   },
+  2935: {
+    comment: 'Save historical block hashes in state',
```
holgerd77, line 195: Can we add some addition like `...state (Verkle related usage, likely not stable yet)` or so?

(let me know if the comment does not make sense "as is")

packages/evm/src/opcodes/functions.ts
```diff
@@ -593,19 +594,35 @@ export const handlers: Map<number, OpHandler> = new Map([
...
-        return
-      }
+      if (common.isActivatedEIP(2935)) {
+        if (number >= runState.interpreter.getBlockNumber()) {
+          runState.stack.push(BIGINT_0)
+          return
+        }
 
-      const block = await runState.blockchain.getBlock(Number(number))
+        const historyAddress = Address.fromString(
```
holgerd77, line 606: Very optional and not directly PR related, just since I noted that many conversion steps:
Would it make sense to add a `Address.fromBigInt()` additional constructor to `Address` where we can skip one (the to hex string one) conversion step? 🤔 

(decide for yourself if a) yes or no and b) if to add here)

packages/vm/src/runBlock.ts
holgerd77: Do we already want to expose this method? Respectively is this generally necessary?

Maybe we keep internal until the EIP has a more stable state at least?

packages/vm/src/runBlock.ts
```diff
@@ -407,6 +416,51 @@ async function applyBlock(this: VM, block: Block, opts: RunBlockOpts): Promise<A
...
+    const key = setLengthLeft(bigIntToBytes(number), 32)
+    await vm.stateManager.putContractStorage(historyAddress, key, hash)
+  }
+  await putBlockHash(this, block.header.parentHash, block.header.number - BIGINT_1)
+
+  // Check if we are on the fork block
+  const forkTime = this.common.eipTimestamp(2935)
+  if (forkTime === null) {
+    throw new Error('EIP 2935 should be activated by timestamp')
+  }
```
holgerd77, line 447: I don't think that we need this extra `eipTimestamp()` method. Isn't this the normal `isActivatedEIP()` call??

Also: such a check should likely be at the beginning of the method (?).

packages/vm/src/runBlock.ts
```diff
@@ -407,6 +416,51 @@ async function applyBlock(this: VM, block: Block, opts: RunBlockOpts): Promise<A
...
+  const parentBlock = await this.blockchain.getBlock(block.header.parentHash)
+
+  if (parentBlock.header.timestamp < forkTime) {
+    let ancestor = parentBlock
+    const range = this.common.param('vm', 'minHistoryServeWindow')
+    for (let i = 0; i < Number(range) - 1; i++) {
```
holgerd77, line 453: I would cautiously think that this `<` usage together with ` - 1` might produce an "off by one" error?

Since the code from the EIP states `for i in range(MIN_HISTORY_SERVE_WINDOW - 1)`?

(maybe confirm with edge case tests?)

holgerd77 (commented):
Some comments to addresss...

packages/evm/src/opcodes/functions.ts
jochem-brouwer: Note: I had to convert bigInt -> Address because in our Common the only allowed `v` values for the `param`s of EIPs (such as `historyStorageAddress`) is bigint. I am not sure if we want to add the `fromBigInt` method because it seems a bit error prone to me: if you delete some numbers, you get a number which in hex string would be less than 20 bytes / 40 characters, but it is still a correct `bigint` - it sounds slightly unsafe to me. (I think we can discuss this but we should address this in a different PR if we want this)

packages/common/src/eips.ts
holgerd77: 👍 

Take whatever text you like, just that people know this is not "one of the normal EIPs they can integrate" (and DRAFT status is not *that* much of a warning, even EIPs close to hardfork often still in DRAFT 😂 ).

packages/evm/src/opcodes/functions.ts
holgerd77: 👍

packages/vm/src/runBlock.ts
jochem-brouwer: I have added this because I need to detect we are on the first fork block. The reason is that on the first fork block, we have to store all the `MIN_HISTORY_SERVE_WINDOW` hashes. I don't know if there is another way to detect if we are on the fork block? I think we need to fork timestamp and then the exact condition `if (parentBlock.header.timestamp < forkTime) {` (this only happens once and is on the fork block)

packages/vm/src/runBlock.ts
holgerd77: 👍

packages/vm/src/runBlock.ts
holgerd77: Ah, you are right, sorry, please dismiss! 🙂 

Should then the whole method get a check at the beginning if 2935 is activated and throw otherwise?

packages/vm/src/runBlock.ts
jochem-brouwer: I think this is correct (but this is definitely off-by-one error prone! :smile: )

From EIP: `for i in range(MIN_HISTORY_SERVE_WINDOW - 1):` this is python code. This loops the for loop `MIN_HISTORY_SERVE_WINDOW - 1` time. Note that `MIN_HISTORY_SERVE_WINDOW` is 256 so we loop this 255 times. In the current TS code we loop `i < 256 - 1` = 255 times also. Note that for each loop, we store the ancestor block. Also note that previously we have already stored the parent block hash of the current block. So, in total we have stored 256 blocks and therefore the history contract has access to 256 slots.

packages/vm/test/api/EIPs/eip-2935-historical-block-hashes.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+    await vm.blockchain.putBlock(block)
+    await vm.runBlock({ block, generate: true })
+
+    const storage = await vm.stateManager.getContractStorage(
+      historyAddress,
+      setLengthLeft(bigIntToBytes(BigInt(0)), 32)
+    )
+    assert.ok(equalsBytes(storage, genesis.hash()))
+  })
+  it('should ensure blocks older than 256 blocks can be retrieved from the history contract', async () => {
```
jochem-brouwer, line 110: @holgerd77 had to re-check, but this test *should* actually test for the off-by-one error.

packages/vm/src/runBlock.ts
jochem-brouwer: I am comparing it with `accumulateParentBeaconBlockRoot`, this also does not have a check for `common.isActivatedEIP(4788)`, but for consistency I can add (but then I will add it to both methods)

packages/vm/src/runBlock.ts
holgerd77: 👍

packages/vm/src/runBlock.ts
holgerd77: Yes, maybe do (for both).

holgerd77 (approved):
LGTM

PR #3276, opened 2024-02-10
Properly apply statemanager `opts` in `fromProof`

acolytec3:
This adjusts the `fromProof` static constructor in `statemanager` to apply `opts` passed into the constructor.


acolytec3:
Once this PR is finished, this script below should not error and should produce the correct storage slot value:
```
import { DefaultStateManager } from './src/index'
import { Trie } from '@ethereumjs/trie'
import { Account, Address, bytesToHex, hexToBytes, intToBytes, randomBytes, setLengthLeft } from '@ethereumjs/util'

const main = async () => {
    const trie = await Trie.create({ useKeyHashing: false })
    const sm = new DefaultStateManager({ trie })
    const pk = hexToBytes('0x9f12aab647a25a81f821a5a0beec3330cd057b2346af4fb09d7a807e896701ea')
    const address = Address.fromPrivateKey(pk)
    const account = new Account()
    await sm.putAccount(address, account)
    await sm.putContractStorage(address, setLengthLeft(intToBytes(0),32), intToBytes(32))
    const storage = await sm.dumpStorage(address)
    const keys = Object.keys(storage)
    const proof = await sm.getProof(address, keys.map(key => hexToBytes(key)))
    const newTrie = (await Trie.create(
        { useKeyHashing: false }
      ))
    const partialSM = await DefaultStateManager.fromProof(proof, false,  { trie: newTrie})
    console.log(await partialSM.getAccount(address))
    console.log(await partialSM.getContractStorage(address, hexToBytes(keys[0])))
}

main()
```

packages/statemanager/README.md
```diff
@@ -71,7 +71,9 @@ Have a loot at the extended `CacheOptions` on how to use and leverage the new ca
...
 
-The `DefaultStateManager` has a static constructor `fromProof` that accepts one or more [EIP-1186](https://eips.ethereum.org/EIPS/eip-1186) [proofs](./src/stateManager.ts) and will instantiate a `DefaultStateManager` with a partial trie containing the state provided by the proof(s). See below example:
+The `DefaultStateManager` has a static constructor `fromProof` that accepts one or more [EIP-1186](https://eips.ethereum.org/EIPS/eip-1186) [proofs](./src/stateManager.ts) and will instantiate a `DefaultStateManager` with a partial trie containing the state provided by the proof(s). Be aware that this constructor accepts the `StateManagerOpts` dictionary as a second parameter (i.e. `stateManager.fromProof(proof, opts)`). Therefore, if you need to use a customized trie (e.g. one that does not use key hashing) or specify caching options, you can pass them in here. If you do instantiate a trie and pass it into the `fromProof` constructor, you also need to instantiate the trie using the corresponding `fromProof` constructor to ensure the state root matches when the proof data is added to the trie. See [this test](./test/stateManager.spec.ts#L287-L288) for more details.
```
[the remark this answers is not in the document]
jochem-brouwer: Ah wait, never mind, that is actually tested :smile:

packages/statemanager/test/stateManager.spec.ts
```diff
@@ -281,4 +284,36 @@ describe('StateManager -> General', () => {
...
+      const keys = Object.keys(storage)
+      const proof = await sm.getProof(
+        address,
+        keys.map((key) => hexToBytes(key))
+      )
+      const newTrie = await Trie.createFromProof(
+        proof.accountProof.map((e) => hexToBytes(e)),
+        { useKeyHashing: false }
+      )
+      const partialSM = await DefaultStateManager.fromProof(proof, true, { trie: newTrie })
```
jochem-brouwer, line 307: I see here, but we already instantiated the `newTrie` with the proof, right? So does this actually add anything? (Except that it checks that we indeed keep the `useKeyHashing` false?). Maybe for completeness add another key/value pair here (a `proof2`) and insert this in the state manager here. Then one can check for both these key/value pairs that they are correct.

packages/statemanager/test/stateManager.spec.ts
```diff
@@ -281,4 +284,36 @@ describe('StateManager -> General', () => {
     zeroAccount = await newPartialStateManager2.getAccount(Address.zero())
     assert.ok(zeroAccount!.nonce === zeroAddressNonce)
   })
+  it.skipIf(isBrowser() === true)(
+    'should create a statemanager fromProof with opts preserved',
+    async () => {
+      const trie = await Trie.create({ useKeyHashing: false })
```
jochem-brouwer, line 290: Ok, so this `trie` uses `useKeyHashing: false` in order to check that we can retrieve the original data, right? (Just checking). If we would use `useKeyHashing: true` and then use `useKeyHashing: false` in the trie and state manager below here then this would lead to incorrect results.

packages/statemanager/test/stateManager.storage.spec.ts
```diff
@@ -14,14 +14,15 @@ import { DefaultStateManager } from '../src/index.js'
...
+const isBrowser = new Function('try {return this===window;}catch(e){ return false;}')
 describe('StateManager -> Storage', () => {
   for (const storageCacheOpts of [
     { deactivate: false },
     { deactivate: true },
     { deactivate: false, size: 0 },
   ]) {
     for (const prefixStorageTrieKeys of [false, true]) {
-      it(`should dump storage`, async () => {
+      it.skipIf(isBrowser() === true)(`should dump storage`, async () => {
```
jochem-brouwer, line 25: (why is this added here? :thinking: )

jochem-brouwer (commented):
Ok this looks good in general, have some comments though.

packages/statemanager/test/stateManager.spec.ts
acolytec3: Correct.

packages/statemanager/test/stateManager.spec.ts
acolytec3: It adds the storage values from the proof and the test then verifies that they are returned and match the expected result.  I will add a second account and verify that as well.

packages/statemanager/test/stateManager.storage.spec.ts
acolytec3: `dumpStorage` crashes in browser because the trie stream issues we have.

jochem-brouwer (commented):
One comment

jochem-brouwer (approved):
LGTM thanks!

PR #3280, opened 2024-02-13
Remove nodejs import from trie.ts

scorbajio:
This change fixes an import error that is coming up in webpack browser builds resulting from a nodejs-specific import being made in the trie package as a result of the readable-stream refactor done in #3231. See https://github.com/ethereumjs/ethereumjs-monorepo/pull/3231#discussion_r1486061100.

### DIFF
--- a/packages/statemanager/src/stateManager.ts
+++ b/packages/statemanager/src/stateManager.ts
@@ -967,14 +967,21 @@ export class DefaultStateManager implements EVMStateManagerInterface {
       throw new Error(`dumpStorage f() can only be called for an existing account`)
     }
     const trie = this._getStorageTrie(address, account)
-    const storage: StorageDump = {}
-    const stream = trie.createAsyncReadStream()
 
-    for await (const chunk of stream) {
-      storage[bytesToHex(chunk.key)] = bytesToHex(chunk.value)
-    }
+    return new Promise((resolve, reject) => {
+      const storage: StorageDump = {}
+      const stream = trie.createReadStream()
 
-    return storage
+      stream.on('data', (val: any) => {
+        storage[bytesToHex(val.key)] = bytesToHex(val.value)
+      })
+      stream.on('end', () => {
+        resolve(storage)
+      })
+      stream.on('error', (e) => {
+        reject(e)
+      })
+    })
   }
 
   /**
@@ -997,35 +1004,44 @@ export class DefaultStateManager implements EVMStateManagerInterface {
       throw new Error(`Account does not exist.`)
     }
     const trie = this._getStorageTrie(address, account)
-    let inRange = false
-    let i = 0
 
-    /** Object conforming to {@link StorageRange.storage}. */
-    const storageMap: StorageRange['storage'] = {}
-    const stream = trie.createAsyncReadStream()
-    for await (const chunk of stream) {
-      if (!inRange) {
-        // Check if the key is already in the correct range.
-        if (bytesToBigInt(chunk.key) >= startKey) {
-          inRange = true
-        } else {
-          continue
+    return new Promise((resolve, reject) => {
+      let inRange = false
+      let i = 0
+
+      /** Object conforming to {@link StorageRange.storage}. */
+      const storageMap: StorageRange['storage'] = {}
+      const stream = trie.createReadStream()
+
+      stream.on('data', (val: any) => {
+        if (!inRange) {
+          // Check if the key is already in the correct range.
+          if (bytesToBigInt(val.key) >= startKey) {
+            inRange = true
+          } else {
+            return
+          }
         }
-      }
-      if (i < limit) {
-        storageMap[bytesToHex(chunk.key)] = { key: null, value: bytesToHex(chunk.value) }
-        i++
-      } else if (i === limit) {
-        return {
+
+        if (i < limit) {
+          storageMap[bytesToHex(val.key)] = { key: null, value: bytesToHex(val.value) }
+          i++
+        } else if (i === limit) {
+          resolve({
+            storage: storageMap,
+            nextKey: bytesToHex(val.key),
+          })
+        }
+      })
+
+      stream.on('end', () => {
+        resolve({
           storage: storageMap,
-          nextKey: bytesToHex(chunk.key),
-        }
-      }
-    }
-    return {
-      storage: storageMap,
-      nextKey: null,
-    }
+          nextKey: null,
+        })
+      })
+      stream.on('error', (e) => reject(e))
+    })
   }
 
   /**
--- a/packages/trie/src/trie.ts
+++ b/packages/trie/src/trie.ts
@@ -30,10 +30,7 @@ import { verifyRangeProof } from './proof/range.js'
 import { ROOT_DB_KEY } from './types.js'
 import { _walkTrie } from './util/asyncWalk.js'
 import { bytesToNibbles, matchingNibbleLength } from './util/nibbles.js'
-import {
-  TrieReadStream as ReadStream,
-  asyncTrieReadStream as asyncReadStream,
-} from './util/readStream.js'
+import { TrieReadStream as ReadStream } from './util/readStream.js'
 import { WalkController } from './util/walkController.js'
 
 import type {
@@ -1198,19 +1195,10 @@ export class Trie {
 
   /**
    * The `data` event is given an `Object` that has two properties; the `key` and the `value`. Both should be Uint8Arrays.
-   * @deprecated Use `createAsyncReadStream`
    * @return Returns a [stream](https://nodejs.org/dist/latest-v12.x/docs/api/stream.html#stream_class_stream_readable) of the contents of the `trie`
    */
   createReadStream(): ReadStream {
     return new ReadStream(this)
-  }
-
-  /**
-   * Use asynchronous iteration over the chunks in a web stream using the for await...of syntax.
-   * @return Returns a [web stream](https://nodejs.org/api/webstreams.html#example-readablestream) of the contents of the `trie`
-   */
-  createAsyncReadStream(): any {
-    return asyncReadStream(this)
   }
 
   /**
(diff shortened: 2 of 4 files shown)


yann300:
fyi I posted a message in the previous PR: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3231/files#r1487361314

holgerd77:
@yann300 @scorbajio Ok, I'll put this back in a "WIP" state unless we have that solved. So if the main file (non type) import is *also* a problem, then we eventually have to roll back the whole feature (?).

TBH this completely let's my head explode, that a web standard implementation in Node.js is *then* breaking web compatibility. 🤯 Is there really no way around that? I wrote at some other occasion when we implemented I guess that normally all browsers would need to do is to *just ignore* this specific import (since things are just *meant* to work any in the browser without this Node.js import).

Hmm. Hmm. Hmm.

holgerd77:
(I tried but I was personally not able to dig something out by googling on this)

scorbajio:
@yann300 could you provide steps to reproduce the build error so we can look into possible solutions?

holgerd77:
> @yann300 could you provide steps to reproduce the build error so we can look into possible solutions?

@scorbajio Just an additional note: I am *sure* this can be solved on the build side, for us this is just not an option, we would cripple our browser compatibility with that.

@yann300 maybe wait with going too deep here, we might just do a follow-up release removing the new functionality again.

yann300:
I had a look at `node:stream/web` and `ReadableStream`, that's indeed somehow weird and annoying...
You might find interesting to look at https://github.com/nodejs/readable-stream ... but I understand adding a new dep can be frustrating even if it comes form a legit org ..

yann300:
> @yann300 could you provide steps to reproduce the build error so we can look into possible solutions?

you could clone https://github.com/ethereum/remix-project and `yarn && yarn build`

holgerd77 (approved):
LGTM, thanks for adressing so quickly!

Bye bye web streams API, it was nice with you. 😑

Have added a note to #3216 so that we do not forget about the code base and might be able to re-integrate along next breaking releases.

Will still wait a couple of days to see if more stuff drops in but otherwise we would do a fixing release round. Maybe next Thursday or so.

yann300:
@holgerd77 @scorbajio thanks for the PR! do you have guys a planned date for getting a release which include this PR?

holgerd77:
@yann300 Hi there, yes, we plan to do releases early next week, want to bundle this with some other stuff (WASM kzg)!

roninjin10:
Just ran into this issue after upgrading ethereumjs libraries in [tevm](https://tevm.sh). Thanks for the fix!

holgerd77:
> Just ran into this issue after upgrading ethereumjs libraries in [tevm](https://tevm.sh). Thanks for the fix!

Thanks for reporting, good to have this confirmed! 🙏 🙂 

Releases on this (and some other stuff) are planned for next Tuesday (since relatively urgent).

Ah, there is also the KZG -> WASM work in there (that will actually be the main feature, so from #3294), so @acolytec3 has done a WASM port of the `c-kzg` library and all 4844 things (so mainly blob txs and the point evaluation precompile in the EVM)). That will likely also be relevant for `Tevm`? That would also be valuable if you can give feedback on the releases from this angle! 😄 

@yann300 so guess for you guys this is relevant as well (?)

holgerd77:
> Let me know if you would like me to add a lint rule that checks for this issue

Yeah, that might be helpful! We are generally aware of this though, this one was a somewhat special case, mental blocker on my side, I couldn't really imagine that using a WEB streams API would break browser compatibilty. 😂

holgerd77:
@yann300 @roninjin10 Another quick question, since I "already have you here" 🙂 

We are a bit drawn back and forth how to handle KZG in EVM and VM.

Basically the thing is: post Cancun, 4844 related functionality becomes an integral part of the EVM/VM with the addition of the new 4844 point evaluation precompile. So that means that the KZG setup is basically necessary for all EVM/VM runs.

This means that we now have a somewhat solid amount of boilerplate for EVM/VM initialization:

```typescript
import { createKZG } from 'kzg-wasm'
import { Common, Chain, Hardfork } from '@ethereumjs/common'
import { EVM } from '@ethereumjs/evm'
import { initKZG } from '@ethereumjs/util'

const main = async () => {
  const kzg = await createKZG()
  initKZG(kzg)
  const common = new Common({
    chain: Chain.Mainnet,
    hardfork: Hardfork.Cancun,
    customCrypto: { kzg: kzg },
  })
  const evm = new EVM({ common })
 }
 ```
 
 We could (likely) avoid this by bundling the `kzg-wasm` library with the EVM/VM. For lower level libraries (like tx) we have users who have a strict no-WASM policy where we would therefore prevent this usage if we do bundle in WASM dependencies.
 
 For higher level libraries like EVM/VM we are not so sure.
 
 What are you guys personal opinions (respectively team based policies) on that? Do you have some problem with bundled in WASM? As a side note it might be worth to mention that the WASM file we have here is somewhat large (~500KB). So this also has the effect that this definitely adds additional boilerplate to the EVM bundle. This is a general problem though, also if the manual self-installation option is used.
 
 Some feedback would be valuable! 🙏

**Comment by holgerd77** [comment]
Side note: we might be lucky and get some JS KZG library written in the next 1-3 months, so this might be generally a temporary situation with the boilerplate code (and this might be an argument to for now skip the deep WASM integration).

**Comment by roninjin10** [comment]
> all EVM/VM runs

Really? Even EVM runs that don't do any blob related things need it? Why is that?

## Wasm bundle size
> the WASM file we have here is somewhat large (~500KB)

Oh wow this is highly problematic. I'm definitely gonna need to have to find a workaround for dealing with that. What makes it so big and is there any potential to make it smaller?  

Definitely would prefer not bundling it in on this bundle size thing alone as it would hurt my ability to work around the large bundle hit

## Initialization code

```typescript
  const kzg = await createKZG()
  initKZG(kzg)
  const common = new Common({
    chain: Chain.Mainnet,
    hardfork: Hardfork.Cancun,
    customCrypto: { kzg: kzg },
  })
  const evm = new EVM({ common })
```

I think it's fine. 

From end user of tevm point of view all this gets abstracted into a `client` instance that has a `ready` method that resolves when initialization is done.  

From my point of view maintaining tevm, I actually currently am following a convention of always explicitly passing in every single parameter into VM, EVM, Blockchain, etc.   E.g. below is how I[ initialize a VM](https://github.com/evmts/tevm-monorepo/blob/main/packages/base-client/src/createBaseClient.js#L151). It's already pretty verbose intentionally

```typescript
const stateManager = createTevmStateManager(getStateManagerOpts())
const common = new TevmCommon({
		chain: 1,
		hardfork: options.hardfork ?? 'shanghai',
		eips: /**@type number[]*/ (options.eips ?? [1559, 4895]),
	})
		const blockchain = await createBlockchain({ common })
		const evm = createEvm({
			common,
			stateManager,
			blockchain,
			allowUnlimitedContractSize: options.allowUnlimitedContractSize ?? false,
			customPrecompiles: options.customPrecompiles ?? [],
			profiler: options.profiler ?? false,
		})
		const vm = await createVm({
			stateManager,
			evm,
			blockchain,
			common,
		})
````

**Comment by acolytec3** [comment]
You should be able to run the EVM just fine, but if you haven't instantiated a KZG instance and try to execute the point evaluation precompile, you'll get an error about `kzg not loaded`.  Otherwise, shouldn't be an issue.

In terms of the size, I think it mainly comes down to the size of the trusted setup.  I'm not an expert on compiling C to WASM, but I turned on all the optimizations that `emscripten` advised and it still was in the 500k range.  If I had to guess, even if/when we have a pure TS/JS version of the KZG library, you're still going to have a big bundle size because there's no way around loading that trusted setup if you want to do KZG crypto operations (required for anything related to 4844).

**Comment by roninjin10** [comment]
> You should be able to run the EVM just fine, but if you haven't instantiated a KZG instance and try to execute the point evaluation precompile, you'll get an error about kzg not loaded. Otherwise, shouldn't be an issue.

This is very nice. This means implementing some kind of lazy loading might be an option.  Putting the KZG implementation on a backend server instead of browser bundle might also be an option. Will dig into it more next week.

> In terms of the size, I think it mainly comes down to the size of the trusted setup. I'm not an expert on compiling C to WASM, but I turned on all the optimizations that emscripten advised and it still was in the 500k range. If I had to guess, even if/when we have a pure TS/JS version of the KZG library, you're still going to have a big bundle size because there's no way around loading that trusted setup if you want to do KZG crypto operations (required for anything related to 4844).

This makes sense. I was lowkey hoping the wasm was compiled from golang in which case there would be a lot of room for improvement.

Looking at the [kzg-wasm](https://github.com/ethereumjs/kzg-wasm/blob/master/package.json) code I'm happy to see it's initialized asyncronously. I opened [this issue in the rustbn repo](https://github.com/ethereumjs/rustbn-wasm/issues/10) that is causing issues from awaiting in the top level of the module.

**Comment by roninjin10** [comment]
Looks like the kzg committment file alone is about 437kb so definitely not much  headroom to make it better https://github.com/ethereum/go-ethereum/blob/00905f7dc406cfb67f64cd74113777044fb886d8/crypto/kzg4844/trusted_setup.json

**Comment by roninjin10** [comment]
With 7zip I was able to compress the 437kb file down to 210kb

```typescript
7z a -mx=9 compressed-file.7z kzg.json

7-Zip [64] 17.05 : Copyright (c) 1999-2021 Igor Pavlov : 2017-08-28
p7zip Version 17.05 (locale=utf8,Utf16=on,HugeFiles=on,64 bits,10 CPUs LE)

Scanning the drive:
1 file, 447355 bytes (437 KiB)

Creating archive: compressed-file.7z

Items to compress: 1


Files read from disk: 1
Archive size: 214362 bytes (210 KiB)
Everything is Ok
```

roninjin10:
btw @acolytec3 @holgerd77 is there a telegram chat or a similar channel for Ethereumjs chat?  Feel free to dm me if there is but you prefer to keep it private: https://twitter.com/FUCORY?ref_src=twsrc%5Egoogle%7Ctwcamp%5Eserp%7Ctwgr%5Eauthor

acolytec3:
You're welcome to jump on our discord for more real-time discussion - https://discord.gg/JCsNTaVw

holgerd77:
Ok, so we have settled to not bundle WASM KZG in as a direct dependency.

PR #3281, opened 2024-02-14
Update Vitest to v1 (New)

holgerd77:
Follow-up respectively replacement for #3191 (will close)

This is a fresh take on the Vitest test runner update from `v0.32.0` to a v1 version (version on PR submission was `v1.2.2`, see Vitest [Releases](https://github.com/vitest-dev/vitest/releases)), also following the work from @ScottyPoi to extract necessary test structure changes in #3266.

This is not necessarily to be immediately picked up but just to get an impression on "where we are". It might eventually make sense to other extractable pre-work and then rebase this PR here and do the Vitest update itself only at the very end of things.

PR should also serve as some discussion ground on some observations in between.

First thing I noted that `Vitest` now has excluded all Node.js primitives and one need to manually add `vite-plugin-node-polyfills`. Have done here now, needed to add this for `events` in Common.

This leads me to the question: how problematic *are* actually Node.js events regarding browser? 🤔 Is this something to address (we can take this over to the chat)?


acolytec3:
Have gone through and updates this based on current master.  Everything passes (most of the time) locally.  I still get seemingly random errors like this for browser tests so still need to research further on that front.

image

acolytec3:
One interesting thing I found is that we do not need to alias `events` to `eventemitter3` and that actually introduces new problems.  Have removed that from my attempts for now.

acolytec3:
So we're down to `wallet` tests failing because of timeouts (they pass locally but are really slow on CI) and then the random failures I described above where browser tests fail because of random "
`failed to load dynamically imported module`

packages/client/test/testdata/blocks/kaustinen4.json
```diff
@@ -1,6 +1,6 @@
 {
-  "353":{
```
scorbajio, line 2: I have come across these random spacing changes myself with larger PR's that required several updates from master. I think it happens with each new verkle testnet update?

packages/client/test/testdata/gethk5vecs/statediffs.json
```diff
@@ -1 +1,457 @@
...ffffffff166381ca91d36040518163ffffffff1660e0"},{"suffix":133,"currentValue":null,"newValue":"0x001b815260040160206040518083038186803b1580156100b857600080fd5b50"},{"suffix":134,"currentValue":null,"newValue":"0x005afa1580156100cc573d6000803e3d6000fd5b505050506040513d601f1960"},{"suffix":135,"currentValue":null,"newValue":"0x011f820116820180604052508101906100f0919061010a565b905090565b6000"},{"suffix":136,"currentValue":null,"newValue":"0x008151905061010481610170565b92915050565b600060208284031215610120"},{"suffix":137,"currentValue":null,"newValue":"0x005761011f61016b565b5b600061012e848285016100f5565b91505092915050"},{"suffix":138,"currentValue":null,"newValue":"0x00565b61014081610161565b82525050565b600060208201905061015b600083"},{"suffix":139,"currentValue":null,"newValue":"0x000184610137565b92915050565b6000819050919050565b600080fd5b610179"},{"suffix":140,"currentValue":null,"newValue":"0x0081610161565b811461018457600080fd5b5056fea2646970667358221220a6"},{"suffix":141,"currentValue":null,"newValue":"0x00a0e11af79f176f9c421b7b12f441356b25f6489b83d38cc828a701720b41f1"},{"suffix":142,"currentValue":null,"newValue":"0x1064736f6c634300080700330000000000000000000000000000000000000000"}]}]]
```
scorbajio, line 1: Same thing here with linting updates from verkle testnet data.

scorbajio (commented):
I checked all of the browser tests locally. They all pass except the block browser tests which consistently fail with `Failed to fetch dynamically imported module`. The vm browser tests are also all consistently passing locally for me, even though they are commented out with a comment saying several are failing (maybe we could uncomment and reintroduce those in case they aren't being problematic). 

All round great update!

packages/client/test/testdata/blocks/kaustinen4.json
acolytec3: I'm assuming so.  I didn't touch this file in my work.

packages/client/test/testdata/gethk5vecs/statediffs.json
acolytec3: Would assume this is the same, just linting stuff as I didn't touch it directly.

scorbajio (approved):
tested ACK

package.json
```diff
@@ -75,5 +77,8 @@
   "engines": {
     "node": ">=18",
     "npm": ">=7"
+  },
+  "dependencies": {
```
holgerd77, line 81: Do we still need this after Amir removed the respective code?

acolytec3: You know, good point, I hadn't considered that.  I thought that was a separate dependency but will check.

Issue #3278, opened 2024-02-12
VM Nightly Tests fail


holgerd77:
The VM nightly tests complely fail for some time, see [here](https://github.com/ethereumjs/ethereumjs-monorepo/actions/runs/7866075404/job/21459884145) for a latest run.

The error sounds relatively generic, eventually "only" TypeScript releated, but nevertheless very much into the source code, here is an extract:

<img width="1001" alt="grafik" src="https://github.com/ethereumjs/ethereumjs-monorepo/assets/931137/6a6551fd-43ba-4a02-b0f3-f51dba938a94">

This should get some attention and should be adressed soon, this might point to something which should also be fixed for end users.

jochem-brouwer:
I will post my findings here.

I had trouble reproducing the CI. I tried first `rm package-lock.json` and then `npm i`. Then I tried `npm run clean` and `rm package-lock.json`. In order to reproduce (do not run this!) I had to delete my entire global npm cache using `npm cache clean --force`. I now get the same error. I isolated the CI in #3285 and will now try to fix it.

jochem-brouwer:
Side question: I think we should add the "clean build" run to our CI. It does not have to run any tests, we simply just check if we can do a clean build. (That way we likely find this issue earlier than we did now, since I still don't get the notifications of these failing runs :( ). The downside of this (which we should discuss) is that once we have this "clean build" error, then all our PRs will fail by default until we fix it.

jochem-brouwer:
The problem is caused by: https://github.com/isaacs/node-lru-cache/commit/1fb116729f28fb610285f5ae455cb383f2d3d3bf and the fact that we use `@types/core-js`.

`@types/core-js` defines the interface for Map as:

```
interface Map<K, V> {
    toJSON(): any;
}
```

However, `Map` does actually not have this `toJSON` method. Due to us loading in devDependencies the builder now thinks that the `LRU-Cache` should have a `toJSON` method (which it does not) since I assume that this new Map interface now overrides the actual interface of Map (if I comment this out in node the Map type does also not get the `toJSON` method in VSCode).

Not sure how to proceed, I will pin the LRU-Cache to the working version in the package for now. Should I report a bug at `@types/core-js`?

holgerd77:
Great analysis 😃, thanks! 👍

PR #3285, opened 2024-02-15
Fix vm nightly tests

jochem-brouwer:
Closes https://github.com/ethereumjs/ethereumjs-monorepo/issues/3278

TODO:

- [x] Restore the old CI jobs
- [x] Ensure build now works. (It does, see this run: https://github.com/ethereumjs/ethereumjs-monorepo/actions/runs/7923279400/job/21632661144?pr=3285. It runs on a cleaned-out nightly VM test which only builds the packages in a clean environment)

The root build problem is: https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/68655

This is because `lru-cache` got this commit: https://github.com/isaacs/node-lru-cache/commit/1fb116729f28fb610285f5ae455cb383f2d3d3bf which got introduced in v10.2.0. The problem is that the map problem of `@types/node` is wrong (see the discussion linked above). Have pinned lru-cache to the previous version which works.


holgerd77 (approved):
LGTM

Issue #3089, opened 2023-10-06
Feature Request: Prometheus Metrics


SeaMonkey82:
The ability to expose a Prometheus metrics endpoint would be a welcome addition, as it is a standard feature in all other Ethereum clients.

acolytec3:
I haven't looked at what other clients populate within their metrics. What would be a useful starting point for identifying what metrics would be helpful?

holgerd77:
I think we should really do this. This will be also pretty valuable for ourselves to analyze stuff.

Maybe we can start a bit small here?

Guess we can take whatever we feel we would like to have here, no imminent need to copy over behavior from other clients (but likely good as a source for inspiration). So:

- Blocks executed per second
- Tx pool size
- Number of peer connections
- Memory load
- …
- 

There is likely a lot of useful stuff to display.

Maybe we just pick the easiest out (tx pool size?) to get things started 🙂

holgerd77:
https://tohidhaghighi.medium.com/add-prometheus-metrics-in-nodejs-ce0ff5a43b44

PR #3287, opened 2024-02-17
Add prometheus to client and begin implementing custom metrics

scorbajio:
This PR begins the Prometheus integration work by first adding a Prometheus client server to the ethereumjs client and then adding custom metrics discussed in #3089.


packages/client/bin/cli.ts
```diff
@@ -59,6 +59,10 @@ import type { CustomCrypto } from '@ethereumjs/common'
 import type { GenesisState } from '@ethereumjs/util'
 import type { AbstractLevel } from 'abstract-level'
 
+const promClient = require('prom-client')
```
acolytec3, line 62: I know this is a little nitpicky but is there a specific motiviation for using `require` syntax here?  This just always ends up in painful places down the line since we will hopefully one day be ESM only and no longer ever need this.  You should be able to use `import * as promClient from 'prom-client'` here.  We do so in Ultralight and should hopefully work the same here.

scorbajio: Good point. Fixed.

packages/client/src/service/txpool.ts
```diff
@@ -362,7 +369,11 @@ export class TxPool {
       add.push({ tx, added, hash })
       this.pool.set(address, add)
       this.handled.set(hash, { address, added })
+
       this.txsInPool++
+      if (this.txGauge !== undefined) {
+        this.txGauge.inc()
+      }
```
acolytec3, line 376:
```suggestion
      this.txGauge?.inc()
```

This syntax works great in ultralight and is a lot code than the `if metrics then metrics.doSomething` as we add more to the system

packages/client/src/service/txpool.ts
```diff
@@ -399,7 +410,13 @@ export class TxPool {
     const poolObjects = this.pool.get(address)
     if (!poolObjects) return
     const newPoolObjects = poolObjects.filter((poolObj) => poolObj.hash !== txHash)
+
     this.txsInPool--
+
+    if (this.txGauge !== undefined) {
+      this.txGauge.dec()
+    }
```
acolytec3, line 418:
```suggestion
    this.txGauge?.dec()
```

packages/client/src/service/txpool.ts
```diff
@@ -840,7 +857,13 @@ export class TxPool {
   close() {
     this.pool.clear()
     this.handled.clear()
+
     this.txsInPool = 0
+
+    if (this.txGauge !== undefined) {
+      this.txGauge.set(0)
+    }
+
```
acolytec3, line 866:
```suggestion
    this.txGauge?.set(0)
```

acolytec3 (changes requested):
Looks good as a starting point.  Can we add a simple unit test to verify that the metrics are being collected?  Might be as simple as starting the metrics client in the `txPool` tests and then fetching from it once a transaction is added to the pool to verify that the txGauge gets incremented.

packages/client/src/service/txpool.ts
```diff
@@ -177,6 +182,8 @@ export class TxPool {
 
     this.opened = false
     this.running = false
+
+    this.txGauge = options.txGauge
```
acolytec3, line 186: Could we add a gauge for each transaction type instead of one general one?  That would be nice for seeing how many blob transactions are currently in the mempool (versus other types)

scorbajio: I've added gauges for each different transaction type. Fixed 🙂.

packages/client/src/client.ts
```diff
@@ -59,6 +59,8 @@ export interface EthereumClientOptions {
 
   /* if client can be run stateless post verkle, defaults to true for now */
   statelessVerkle?: boolean
+
+  prometheusMetrics?: PrometheusMetrics | undefined
```
acolytec3, line 63: Let's add some comments here explaining this option

acolytec3:
Changes I have made to the original PR
- Add a new CLI flag for `prometheusPort` 
- Moved `prometheusMetrics` to a new optional property on the `config` object that is passed around within the client.  This should make for less repetitive code when we add new metrics in other places within the `client`
- Add better typing to `stopClient` parameters in `bin/cli.ts`
- Add minimal docs for how to start metrics server

acolytec3 (dismissed):
I'm going to approve since I requested changes (but I made most of those myself).  Will leave to @scorbajio or someone else to merge so I don't approve my own changes.

g11tech (dismissed):
lgtm

packages/client/bin/cli.ts
```diff
@@ -1027,6 +1048,41 @@ async function run() {
...
+      const reqUrl = new url.URL(req.url, `http://${req.headers.host}`)
+      const route = reqUrl.pathname
+
+      if (route === '/metrics') {
+        // Return all metrics in the Prometheus exposition format
+        res.setHeader('Content-Type', register.contentType)
+        res.end(await register.metrics())
+      }
+    })
+    // Start the HTTP server which exposes the metrics on http://localhost:8080/metrics
```
scorbajio, line 1081: Should this comment be changed to not include the port or to generalize the port number so developers/users don't get confused by this comment?

packages/client/bin/cli.ts
```diff
@@ -227,6 +232,16 @@ const args: ClientOpts = yargs
...
   })
+  .option('prometheus', {
+    describe: 'Enable the Prometheus metrics server with HTTP endpoint',
+    boolean: true,
+    default: false,
+  })
+  .option('prometheusPort', {
+    describe: 'Enable the Prometheus metrics server with HTTP endpoint',
+    number: true,
+    default: 8000,
```
scorbajio, line 243: Should this be defaulted to port 8080 or 8000?

packages/client/bin/cli.ts
scorbajio: Decided to keep the prometheus server defaulting to port 8000 and updating the comment to include a general port number instead of port 8080.

acolytec3 (approved):
LGTM

PR #3288, opened 2024-02-19
Remove devnet6 trusted setup

acolytec3:
Updates all references to the `devnet6.txt` trusted setup to point to the `official.txt` trusted setup for mainnet.  

Also regenerates test data where necessary using official trusted setup.


acolytec3 (commented):
Note, I removed the nethermind blob transaction in the `tx` eip4844 tests because it was created on `devnet6` and I have no way of regnerating that test data using the mainnet trusted setup.  I will try and find a full blob transaction from geth or another client that we can verify using the new settings.

jochem-brouwer (commented):
If you could leave some documentation so we can verify that the output data is correct, then I think all is fine :)

acolytec3:
> If you could leave some documentation so we can verify that the output data is correct, then I think all is fine :)

Okay, I basically just updated the expected outputs to match what is produced when you use the official trusted setup instead of the devnet 6 one.

holgerd77 (approved):
Great work, looks good! 👍 🙂 

Will merge.

PR #3291, opened 2024-02-21
Client: Refactor Engine API

holgerd77:
The Engine API grown like crazy over the last months and years and our single `engine.ts` file doesn't hold all the code any more.

This is a much needed basic structural refactoring, putting things into its own files, adding (some) additional code docs and generally making the overloaded `engine.ts` file leaner and better readable again.

Still WIP, will push some additional clean-ups, likely ready for review at the end of the day though.


g11tech (approved):
lgtm

packages/client/src/rpc/modules/engine/util/getPayloadBody.ts
```diff
@@ -0,0 +1,14 @@
+import { bytesToHex } from '@ethereumjs/util'
```
acolytec3, line 1: Any objection to including under `getPayload.ts` since it and the `getPayload` methods are logically connected and this doesn't have many lines of code to it.

PR #3292, opened 2024-02-21
trie: export "Path" interface

ScottyPoi:
Users of the `@ethereumjs/trie` library will see this error when using `trie.findPath` in their tooling, specifically when writing a function which returns the result of `trie.findPath`:

> Return type of public method from exported class has or is using private name 'Path'.ts(4055)

This happens because the `Path` interface was not previously exported.  Users were unable to import it directly from the Trie library to declare a return type for their functions.  The Typescript compiler would also be unable to properly infer the return type for the user function, because the expected return type `Path` could not actually be found from the outside.

Fix:
`interface Path {...} ` > `export interface Path {...}`

### DIFF
--- a/packages/trie/src/trie.ts
+++ b/packages/trie/src/trie.ts
@@ -47,7 +47,7 @@ import type { OnFound } from './util/asyncWalk.js'
 import type { BatchDBOp, DB, PutBatch } from '@ethereumjs/util'
 import type { Debugger } from 'debug'
 
-interface Path {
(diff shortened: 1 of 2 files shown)


holgerd77 (approved):
Thanks for the PR and the great explanation along! 🙂 

I think we should then move the interface over to the `types.ts` file (have directly done) for consistency, otherwise looks good!

Will merge.

PR #3295, opened 2024-02-27
Implement eth_feeHistory

jochem-brouwer:
Follow-up PR of https://github.com/ethereumjs/ethereumjs-monorepo/pull/2773

Note: tests fail, have to figure out how to migrate those to our new format.

### DIFF
--- a/packages/client/src/rpc/modules/eth.ts
+++ b/packages/client/src/rpc/modules/eth.ts
@@ -1,9 +1,13 @@
+import { Hardfork } from '@ethereumjs/common'
+import { RLP } from '@ethereumjs/rlp'
 import { BlobEIP4844Transaction, Capability, TransactionFactory } from '@ethereumjs/tx'
 import {
   Address,
   BIGINT_0,
   BIGINT_1,
   TypeOutput,
+  // eslint-disable-next-line import/named
+  bigIntMax,
   bigIntToHex,
   bytesToHex,
   equalsBytes,
@@ -201,6 +205,97 @@ const jsonRpcReceipt = async (
 })
 
 /**
+ * Get block by option
+ */
+const getBlockByOption = async (blockOpt: string, chain: Chain) => {
+  if (blockOpt === 'pending') {
+    throw {
+      code: INVALID_PARAMS,
+      message: `"pending" is not yet supported`,
+    }
+  }
+
+  let block: Block
+  const latest = chain.blocks.latest ?? (await chain.getCanonicalHeadBlock())
+
+  switch (blockOpt) {
+    case 'earliest':
+      block = await chain.getBlock(BigInt(0))
+      break
+    case 'latest':
+      block = latest
+      break
+    case 'safe':
+      block = chain.blocks.safe ?? (await chain.getCanonicalSafeBlock())
+      break
+    case 'finalized':
+      block = chain.blocks.finalized ?? (await chain.getCanonicalFinalizedBlock())
+      break
+    default: {
+      const blockNumber = BigInt(blockOpt)
+      if (blockNumber === latest.header.number) {
+        block = latest
+      } else if (blockNumber > latest.header.number) {
+        throw {
+          code: INVALID_PARAMS,
+          message: 'specified block greater than current height',
+        }
+      } else {
+        block = await chain.getBlock(blockNumber)
+      }
+    }
+  }
+  return block
+}
+
+const calculateRewards = async (
+  block: Block,
+  receiptsManager: ReceiptsManager,
+  priorityFeePercentiles: bigint[]
+) => {
+  const blockRewards: bigint[] = []
+  const txGasUsed: bigint[] = []
+  const baseFee = block.header.baseFeePerGas
+  const receipts = await receiptsManager.getReceipts(block.hash())
+
+  if (receipts.length > 0) {
+    txGasUsed.push(receipts[0].cumulativeBlockGasUsed)
+    receipts.shift()
+  }
+
+  for (const receipt of receipts) {
+    txGasUsed.push(receipt.cumulativeBlockGasUsed - txGasUsed[txGasUsed.length - 1])
+  }
+
+  const txs = block.transactions
+  const txsWithGasUsed = txs.map((tx, i) => ({
+    ...tx,
+    gasUsed: txGasUsed[i],
+    effectivePriorityFee: tx.getEffectivePriorityFee(baseFee),
+  }))
+
+  const txsWithGasUsedSorted = txsWithGasUsed.sort((currTx, nextTx) =>
+    currTx.effectivePriorityFee > nextTx.effectivePriorityFee ? 1 : -1
+  )
+
+  let totalGasUsed = 0n
+  let rewardPercentileIndex = 0
+  for (const tx of txsWithGasUsedSorted) {
+    totalGasUsed += tx.gasUsed
+
+    while (
+      rewardPercentileIndex < priorityFeePercentiles.length &&
+      (totalGasUsed / block.header.gasUsed) * 100n >= priorityFeePercentiles[rewardPercentileIndex]
+    ) {
+      blockRewards.push(tx.effectivePriorityFee)
+      rewardPercentileIndex++
+    }
+  }
+
+  return blockRewards
+}
+
+/**
  * eth_* RPC module
  * @memberof module:rpc/modules
  */
@@ -366,6 +461,16 @@ export class Eth {
     )
 
     this.gasPrice = middleware(callWithStackTrace(this.gasPrice.bind(this), this._rpcDebug), 0, [])
+
+    this.feeHistory = middleware(
+      callWithStackTrace(this.feeHistory.bind(this), this._rpcDebug),
+      2,
+      [
+        [validators.either(validators.hex, validators.integer)],
+        [validators.either(validators.hex, validators.blockOption)],
+        [validators.array(validators.integer)],
+      ]
+    )
   }
 
   /**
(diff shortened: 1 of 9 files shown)


packages/client/src/rpc/modules/eth.ts
```diff
@@ -1110,4 +1170,72 @@ export class Eth {
...
+      rewards = await Promise.all(
+        requestedBlocks.map((b) =>
+          calculateRewards(b, this.receiptsManager!, priorityFeePercentiles)
+        )
+      )
+    }
+
+    return {
+      baseFeePerGas: baseFees.map((f) =>
+        f !== undefined ? bigIntToHex(f) : bigIntToHex(BigInt(0))
```
acolytec3, line 1234: Is there a reason to not just put the string literal here and save ourselves a couple of conversion steps?
```suggestion
        f !== undefined ? bigIntToHex(f) : '0x'
```

acolytec3 (dismissed):
Left some formatting comments and questions

packages/client/src/rpc/modules/eth.ts
```diff
@@ -200,6 +203,53 @@ const jsonRpcReceipt = async (
...
+  const baseFee = block.header.baseFeePerGas
+  const receipts = await receiptsManager.getReceipts(block.hash())
+
+  if (receipts.length > 0) {
+    txGasUsed.push(receipts[0].cumulativeBlockGasUsed)
+    receipts.shift()
+  }
+
+  for (const receipt of receipts) {
+    txGasUsed.push(receipt.cumulativeBlockGasUsed - txGasUsed[txGasUsed.length - 1])
```
acolytec3, line 222: Shouldn't we subtract `receipts[n-1].cumulativeBlockGasUsed` here and not just the gas for the previous tx?

packages/tx/src/eip1559Transaction.ts
```diff
@@ -206,6 +206,21 @@ export class FeeMarketEIP1559Transaction extends BaseTransaction<TransactionType
     return EIP2930.getDataFee(this)
   }
 
+  /**
+   * Returns the minimum of calculated priority fee (from maxFeePerGas and baseFee) and maxPriorityFeePerGas
+   *
+   * * @param baseFee Base fee retrieved from block
+   */
+  getEffectivePriorityFee(baseFee: bigint | undefined): bigint {
```
acolytec3, line 214: Feels like this could be a generic utility function rather than having to expand the class API or else a capability in eip1559.  I dunno, just feels like we have a lot of duplicate code across the txn classes.  I understand that legacy txns have a slightly different process to identify the effective tip so happy to let this sit as is if you feel its a cleaner approach.

packages/client/src/rpc/modules/eth.ts
```diff
@@ -200,6 +203,53 @@ const jsonRpcReceipt = async (
...
+    ...tx,
+    gasUsed: txGasUsed[i],
+    effectivePriorityFee: tx.getEffectivePriorityFee(baseFee),
+  }))
+
+  const txsWithGasUsedSorted = txsWithGasUsed.sort((currTx, nextTx) =>
+    currTx.effectivePriorityFee > nextTx.effectivePriorityFee ? 1 : -1
+  )
+
+  let totalGasUsed = 0n
```
acolytec3, line 236:
```suggestion
  let totalGasUsed = BIGINT_0
```

packages/client/src/rpc/modules/eth.ts
```diff
@@ -200,6 +203,53 @@ const jsonRpcReceipt = async (
...
+  let rewardPercentileIndex = 0
+  for (const tx of txsWithGasUsedSorted) {
+    totalGasUsed += tx.gasUsed
+
+    while (
+      rewardPercentileIndex < priorityFeePercentiles.length &&
+      (totalGasUsed / block.header.gasUsed) * 100n >= priorityFeePercentiles[rewardPercentileIndex]
```
acolytec3, line 243:
```suggestion
      (totalGasUsed / block.header.gasUsed) * BigInt(100) >= priorityFeePercentiles[rewardPercentileIndex]
```

packages/tx/src/eip1559Transaction.ts
jochem-brouwer: I think I can clean this up a bit :) I agree that this bloats the code.

packages/client/src/rpc/modules/eth.ts
```diff
@@ -1110,4 +1170,72 @@ export class Eth {
...
+    ).header
+
+    const oldestBlockNumber = bigIntMax(lastRequestedBlockNumber - blockCount - 1n, BigInt(0))
+
+    const requestedBlockNumbers = Array.from(
+      { length: Number(blockCount) },
+      (_, i) => oldestBlockNumber + BigInt(i)
+    )
+
+    const requestedBlocks = await Promise.all(
```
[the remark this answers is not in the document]
jochem-brouwer: I think if this throws then the RPC will return an internal error. We could wrap it to make it more descriptive? I'll add a test to see what happens.

packages/client/src/rpc/modules/eth.ts
jochem-brouwer: Yup good one. Note: output of `bigIntToHex(0n)` = `0x0`. So will put that there :)

packages/tx/src/eip1559Transaction.ts
```diff
@@ -206,6 +206,21 @@ export class FeeMarketEIP1559Transaction extends BaseTransaction<TransactionType
...
+    if (baseFee === undefined || baseFee >= this.maxFeePerGas) {
+      return 0n
```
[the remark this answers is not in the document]
jochem-brouwer: Yes and actually we should use the cached constants :) `BIGINT_0`

packages/tx/src/eip1559Transaction.ts
jochem-brouwer: Have addressed

packages/client/src/rpc/modules/eth.ts
jochem-brouwer: Yes you are right :)

packages/client/src/rpc/modules/eth.ts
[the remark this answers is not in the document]
acolytec3: Actually, you know, if we have the debug turned on, it will return a useful error with the stacktrace anyway.

packages/client/test/rpc/eth/getFeeHistory.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+  it(`${method}: should return 12.5% decreased base fee if the block is empty`, async () => {
+    const { chain, server, execution } = await setupChain(gethGenesisStartLondon(pow), 'powLondon')
+    const gasUsed = BigInt(0)
+
+    await produceFakeGasUsedBlock(execution, chain, gasUsed)
+    const rpc = getRpcClient(server)
+
+    const res = await rpc.request(method, ['0x2', 'latest', []])
+    const [, previousBaseFee, nextBaseFee] = res.result.baseFeePerGas as [string, string, string]
+    const increase =
```
acolytec3, line 92: This is nitpicky but can we call this `decrease` instead since we're looking for a decrease in base fee?

packages/client/src/rpc/modules/eth.ts
```diff
@@ -200,6 +204,66 @@ const jsonRpcReceipt = async (
...
+      if (priorityFeeIndex >= priorityFeePercentiles.length) {
+        // prevent out-of-bounds read
+        break
+      }
+      const priorityFeePercentile = priorityFeePercentiles[priorityFeeIndex]
+      targetCumulativeGasUsed = (block.header.gasUsed * BigInt(priorityFeePercentile)) / BIGINT_100
+    }
+  }
+
+  return blockRewards
```
acolytec3, line 264: Before we return, don't we need to divide the total fees for a percentile by the # of transactions that were included in that percentile to get the median fee in that set of transactions?

packages/client/src/rpc/modules/eth.ts
jochem-brouwer: We care about the tx gas used and not about the cumulative gas used (this is fixed).

packages/client/src/rpc/modules/eth.ts
jochem-brouwer: I think this is now resolved :)

packages/client/src/rpc/validation.ts
```diff
@@ -303,6 +303,22 @@ export const validators = {
     }
   },
 
+  /**
+   * number validator to check if type is integer
+   * @param params parameters of method
+   * @param index index of parameter
+   */
+  get integer() {
```
ScottyPoi, line 311: Strictly, an **integer** is:
> A member of the set of positive whole numbers {1, 2, 3, … }, negative whole numbers {−1, −2, −3, … }, and zero {0}. 

`Number.isInteger(n: unknown)` will return **false** *unless* `typeof n === 'number'` *and* **n** is an **integer** 

```suggestion
  
  get integer() {
    return (params: any[], index: number) => {
      if (!Number.isInteger(params[index])) {
        return {
          code: INVALID_PARAMS,
          message: `invalid argument ${index}: argument is not a number`,
        }
      }
    }
  },
```

packages/client/src/rpc/modules/eth.ts
```diff
@@ -1110,4 +1192,92 @@ export class Eth {
...
+    }
+
+    return {
+      baseFeePerGas: baseFees.map(bigIntToHex),
+      gasUsedRatio: gasUsedRatios,
+      baseFeePerBlobGas: baseFeePerBlobGas.map(bigIntToHex),
+      blobGasUsedRatio,
+      oldestBlock: bigIntToHex(oldestBlockNumber),
+      reward: rewards.map((r) => r.map((n) => bigIntToHex(n))),
+    }
```
ScottyPoi, line 1281: just to match the pattern above :smile
```suggestion
      reward: rewards.map((r) => r.map(bigIntToHex)),
```

ScottyPoi:
With [a33dc8e](https://github.com/ethereumjs/ethereumjs-monorepo/pull/3295/commits/a33dc8e6b6785db5b04f1e9c9d3ee61478c4fd5e) I broke up the `rewardPercentile` validator into two separate validators that match the naming schema in **fee_market.yaml**.  

`rewardPercentile`  = Floating point value between 0 and 100
`rewardPercentiles` = A monotonically increasing list of percentile values

I personally like to avoid using plural namings (percentile / percentiles), but that is how they are named in the spec.

acolytec3 (approved):
LGTM!

ScottyPoi:
LGTM!  :+1:

PR #3298, opened 2024-03-01
evm/vm: ensure  backwards compatability reportPreimages

jochem-brouwer:
This PR guards `startReportingPreimages` of EVM

This is OK because:

- The only way this is called is within VM `reportPreimages` -> therefore, we already assume that `startReportingPreimages` is available


holgerd77 (approved):
LGTM

Issue #3299, opened 2024-03-02
 [EIP-6110] Supply validator deposits on chain


scorbajio:
Changes in the Execution Layer (from the [slides](https://docs.google.com/presentation/d/1Q2X8sNzbhdGjOo0IVmjxYfMEdznTRArKGj2DznYCbLk/edit#slide=id.g255342db7c9_0_17)):
* Defining the deposit structure for engine API request and response, RLP encoding and decoding
* Define new `ExecutionPayload` in engine API to include deposits
* Introduce `depositsRoot` and deposits to `BlockHeader` and `BlockBody` respectively
* Update `BlockCreator` to extract and decode deposits from transaction receipts
* Update `BlockValidator` to validate deposits
* Introduce `depositContractAddress` to genesis config
* Design a series of acceptance test for EIP-6110

References:
* https://eips.ethereum.org/EIPS/eip-6110
* https://github.com/ethereum/consensus-specs/tree/dev/specs/_features/eip6110
* https://www.youtube.com/watch?v=tRTBgCN9VgY
* https://github.com/mkalinin/execution-apis/blob/eip6110/src/engine/experimental/eip6110.md#executionpayloadv6110
* https://github.com/hyperledger/besu/pull/5055
* https://github.com/hyperledger/besu/pull/5295

PR #3303, opened 2024-03-05
EIP-6110: Supply Validator Deposits on Chain

scorbajio:
This PR implements EIP-6110 and addresses issue #3299.


jochem-brouwer:
I've pointed https://github.com/ethereumjs/ethereumjs-monorepo/pull/3336 towards this branch :)

Before we merge this one into `master`, we should ensure that we can can still use Verkle, and possibly also use Verkle together with 6110 + 7002.

holgerd77:
Ugh.

Now having read this EIP for the first time I’m realizing what a complex beast this is, always considered this more as some „side EIP“. 🤯

@scorbajio Congrats on that, this looks already pretty great! 🤩 Really blown away by the amount of code already touched here!

packages/block/src/types.ts
```diff
@@ -168,12 +177,26 @@ export type BlockBytes =
       WithdrawalsBytes,
       ExecutionWitnessBytes
     ]
+  | [BlockHeaderBytes, TransactionsBytes, UncleHeadersBytes, WithdrawalsBytes, DepositsBytes]
+  | [
+      BlockHeaderBytes,
+      TransactionsBytes,
+      UncleHeadersBytes,
+      WithdrawalsBytes,
+      DepositsBytes,
```
scorbajio, line 186: Not sure if this is going to be before or after `ExecutionWitnessBytes` in Prague.

holgerd77: ExecutionWitness is for Verkle which is not scheduled for Prague, so Deposits can always be placed before ExecutionWitness.

packages/common/src/eips.ts
```diff
@@ -505,6 +505,13 @@ export const EIPs: EIPsDict = {
       },
     },
   },
+  6110: {
+    comment: 'Supply validator deposits on chain',
+    url: 'https://eips.ethereum.org/EIPS/eip-6110',
+    status: Status.Draft,
+    minimumHardfork: Hardfork.Cancun,
```
holgerd77, line 512: The `minimumHardfork` is not the inclusion HF, but the first HF from which this EIP can be run in isolation (so: all pre-requisits for the EIP are met). So we can (should) switch to `Cancun` here, just as some additional context.

scorbajio:
Implementation of EIP-6110 is being continued at #3390.

Smangele87 (commented):
packages/common/src/eips.ts

PR #3304, opened 2024-03-05
Update rustbn-wasm usage 

acolytec3:
This PR updates the `rustbn-wasm` verion to use the new API of an async init function, following `kzg-wasm`.  

It also adds a new `evm.create` async constructor.


packages/evm/src/evm.ts
```diff
@@ -137,9 +137,15 @@ export class EVM implements EVMInterface {
 
   protected readonly _emit: (topic: string, data: any) => Promise<void>
 
-  constructor(opts: EVMOpts = {}) {
-    this.events = new AsyncEventEmitter()
+  private bn128: {
```
roninjin10, line 140: Does it make sense to create a `EVM.create` method similar to `VM.create`?

roninjin10:
Another benifit of this change is you should see slightly faster startup time because wasm compilation won't block module resolution

packages/evm/src/evm.ts
acolytec3: That thought definitely occurred to me while I was doing this. Let me take a look at our current usage of the EVM constructor and see if it makes sense. It would definitely simplify the API

packages/evm/test/precompiles/06-ecadd.spec.ts
```diff
@@ -6,9 +6,10 @@ import { EVM, getActivePrecompiles } from '../../src/index.js'
 describe('Precompiles: ECADD', () => {
   it('ECADD', async () => {
     const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Petersburg })
-    const evm = new EVM({
+    const evm = await EVM.create({
       common,
     })
+    console.log(evm['bn128'])
```
jochem-brouwer, line 12: `console.log`

packages/evm/src/evm.ts
```diff
@@ -137,7 +138,23 @@ export class EVM implements EVMInterface {
...
   protected readonly _emit: (topic: string, data: any) => Promise<void>
 
-  constructor(opts: EVMOpts = {}) {
+  private bn128: {
+    ec_pairing: (input_str: string) => string
+    ec_add: (input_str: string) => string
+    ec_mul: (input_hex: string) => string
+  }
+
+  static async create(opts: EVMOpts) {
```
jochem-brouwer, line 147: I think we should make `opts` optional.
```suggestion
  static async create(opts?: EVMOpts) {
```

jochem-brouwer (commented):
Preliminary review with some thoughts :smile: :+1:

packages/evm/src/evm.ts
```diff
@@ -137,7 +138,23 @@ export class EVM implements EVMInterface {
...
+    if (opts.bn128 === undefined) {
+      opts.bn128 = await initRustBN()
+    }
+    return new EVM(opts)
+  }
+  constructor(opts: EVMOpts) {
+    if (opts.bn128 === undefined) {
```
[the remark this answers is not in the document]
acolytec3: What is breaking here? I guess it's not obvious to me why we would want to allow the EVM run without this dep?  We've never allowed it before so doesn't make sense to allow it now to my way of thinking.

packages/evm/src/evm.ts
```diff
@@ -137,7 +138,23 @@ export class EVM implements EVMInterface {
...
-  constructor(opts: EVMOpts = {}) {
+  private bn128: {
+    ec_pairing: (input_str: string) => string
+    ec_add: (input_str: string) => string
+    ec_mul: (input_hex: string) => string
+  }
+
+  static async create(opts: EVMOpts) {
+    if (opts.bn128 === undefined) {
+      opts.bn128 = await initRustBN()
```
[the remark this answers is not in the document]
acolytec3: I thought about that but this isn't really "custom crypto".  This is just our existing dependency but since it now requires an async constructor, I'm having to pass it in.  In this case, I don't think it makes sense to instantiate it in Common or even attach it there. There isn't an alternative implementation of the crypto anywhere that I'm aware of aside from the one older ASM version we were previously using.

holgerd77 (approved):
LGTM! 🎉 

@roninjin10 FYI: we are now actually planning to do out of schedule breaking releases for EVM/VM on this (which is also an experiment for us).

We would not want to do this too often, but for this very general fix to make and the new `EVM.create()` constructor it seems justified here.

If you do opinions and/or remarks on this let us know! 🙏 🙂

holgerd77:
(so this will slightly postpone the schedule for this release round, pushing this to next week)

PR #3311, opened 2024-03-09
Update "--dev" to use the Cancun hardfork.

Amxx:
Currently, calling `ethereumjs --dev=true` starts a network using the london hardfork. This is quite different from mainnet, and IMO is not representative of what dev would expect.

With DenCun hitting mainnet in the next days, I believe that this should be updated to Cancun.

### DIFF
--- a/packages/client/CHANGELOG.md
+++ b/packages/client/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 (modification: no type change headlines) and this project adheres to
 [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+Update the devnet setting (enabled using `--dev`) to use the Cancun hardfork.
+
 ## 0.10.0 - 2024-02-08
 
 This client release now comes with official Dencun hardfork support 🎉 and by default uses WASM for crypto primitives for faster block execution times.
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -654,6 +654,9 @@ async function setupDevnet(prefundAddress: Address) {
       istanbulBlock: 0,
       berlinBlock: 0,
       londonBlock: 0,
+      mergeForkBlock: 0,
+      shanghaiTime: 0,
+      cancunTime: 0,
       ...consensusConfig,
     },
     nonce: '0x0',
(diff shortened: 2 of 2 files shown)


holgerd77:
Will circle in @g11tech in here, most experienced with these dev flags: Gajinder, can we update here, or does this have side effects on tests (maybe we'll directly see on CI run, not yet finished while I am writing here)?

g11tech (commented):
LGTM, not approving if we want the unreleased changelog edit removed

g11tech:
Also as mentioned in the PR, let's just schedule it to dencun

The only reason we may want it to be on London however is to have a runnable chain without a CL

holgerd77:
Updated this via UI

holgerd77:
Updated this via UI

acolytec3 (commented):
Unfortunately, I don't think its as simple as changing the starting hardfork to start on Cancun.  If you run `npx tsx bin/client.ts --dev=true` you get this sort of output:

```
[03-12|20:17:18] INFO ---------------------------
[03-12|20:17:18] INFO Client started successfully
[03-12|20:17:18] INFO ---------------------------
[03-12|20:17:28] INFO txsByPriceAndNonce selected txs=0, skipped byNonce=0 byPrice=0 byBlobsLimit=0 
[03-12|20:17:28] INFO Miner: Assembling block from 0 eligible txs (baseFee: 7)
[03-12|20:17:28] INFO Miner: Sealed block with 0 txs (in turn)
[03-12|20:17:28] WARN Execution of block number=1 hash=0x8c93…dc62 hardfork=cancun failed:
Error: invalid block stateRoot, got: 0x5e2514a6fc32e2f9b31cbe99d78a5caa2e5e8fbd51461faa8d3f0cc59714ab37, want: 0x874ad777665ba3859329672df99b4bfee3ee775b36379d773ae69dc6d8612db2 (vm hf=cancun -> block number=1 hash=0x8c93e4329f71ba3f7fd33c6f001572b1e31bd02b8ba7c7a044d2438e6523dc62 hf=cancun baseFeePerGas=7 txs=0 uncles=0) 
[03-12|20:17:28] INFO Executed blocks count=1 first=0 hash=0xa8e0…b110 td=3 baseFee=7 hardfork=cancun last=0 hash=0xa8e0…b110 txs=0 
[03-12|20:17:36] INFO Stopped synchronization.
```

It's trying to start a PoA network with post-merge hardforks which is an invalid combination with the current `dev` genesis config.  I agree with the spirit of the PR but we need to think about something a little more fully orbed before we take this in, maybe provide something like the `lodestar-quickstart` setup so it comes paired with a CL client so it can produce valid blocks.

holgerd77:
@acolytec3 can you take responsibilty for this PR and see if it is worth to pursue or close? Thanks! 🙏

holgerd77:
👋 Hope, everything is fine. Outdated, client deprecated, will close.

PR #3312, opened 2024-03-09
WIP: implement EIP-5806

Amxx:
See https://github.com/Amxx/EIP-5806/tree/main to track the other implementation effort.

### DIFF
--- a/packages/client/bin/cli.ts
+++ b/packages/client/bin/cli.ts
@@ -654,6 +654,9 @@ async function setupDevnet(prefundAddress: Address) {
       istanbulBlock: 0,
       berlinBlock: 0,
       londonBlock: 0,
+      mergeForkBlock: 0,
+      shanghaiTime: 0,
+      cancunTime: 0,
       ...consensusConfig,
     },
     nonce: '0x0',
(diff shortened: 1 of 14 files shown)


holgerd77:
Hi there,
super interesting work! 🤩 To be honest, we cannot promise at the point that we will take this into the main code base, so you might need to live with the situation to reference this as a PR or your own fork.

We'll see. We do not have super transparent ruling what to take in and what not (also to be honest here), at the moment we mostly integrate EIPS already included in HFs and occassionally we do some exceptions for "promising EIPs" which "doesn't bloat the code base too much". 😝 

Admittedly these are very blurry "criterias". We'll discuss internally how we want to proceed here, maybe also worth to document this in a more explicit way.

We might also want to think along stuff like...

- See if we can make our stack (even more) modular, to make these kind of integrations easier and/or
- Make it easier to fork and publish the whole stack

(both rather for the cases that we do not integrate)

Again, we'll see. Will report back here, happy to take opinions/suggestions on that in between! 🙂

Amxx:
> super interesting work! 🤩 To be honest, we cannot promise at the point that we will take this into the main code base, so you might need to live with the situation to reference this as a PR or your own fork.

That is ok.

holgerd77:
Not considered for integration, will close

PR #3315, opened 2024-03-11
EVM/VM create() Static Constructor Reworking

holgerd77:


holgerd77:
Please wait, I have forgotten the VM part.

packages/vm/src/vm.ts
```diff
@@ -78,107 +77,57 @@ export class VM {
...
-    this.events = new AsyncEventEmitter<VMEvents>()
-
-    this._opts = opts
-
-    if (opts.common) {
-      this.common = opts.common
-    } else {
+    // Add common, SM, blockchain, EVM here
+    if (opts.common === undefined) {
       const DEFAULT_CHAIN = Chain.Mainnet
```
acolytec3, line 82: Do we really need this?  It feels unnecessary to declare this constant instead of just passing `Chain.Mainnet` in the next line to the `common` constructor.

packages/evm/src/evm.ts
```diff
@@ -138,47 +139,54 @@ export class EVM implements EVMInterface {
...
-    this._optsCached = opts
+    const bn128 = await initRustBN()
 
-    this.transientStorage = new TransientStorage()
-
-    if (opts.common) {
-      this.common = opts.common
-    } else {
+    if (opts.common === undefined) {
       const DEFAULT_CHAIN = Chain.Mainnet
```
acolytec3, line 156: As in `vm`, this feels duplicative.  Can we just pass `Chain.Mainet` to the `common` constructor?

acolytec3 (dismissed):
Looks good in general though left a couple of notes aboute code that feels unnecessarily wordy.

jochem-brouwer:
Have pushed blockchain `_init` method removal in https://github.com/ethereumjs/ethereumjs-monorepo/pull/3315/commits/47f235db5e81daf72940e14e55a80325c2d382cc (remove from this PR if we want to merge it without the removal of blockchain `_init` and proceed in new PR)

Client tests fail, but only 2: to run: `npx vitest run ./test/net/protocol/ethprotocol.spec.ts` (in client)

holgerd77:
> npx vitest run ./test/net/protocol/ethprotocol.spec.ts

Very cool! 🥳 

I would cautiously think we can keep this `_init()` removal in here? Would be nice if we have the "full thing" merged and that's it. Will have a look at the failing test(s), maybe I can spot something.

I would target to do the breaking releases on Thursday then, should be realistic. Last chance to last-minute get some manageably sized local refactoring of stuff (EVM/VM?) in, if you have spotted something along the work. 😋

holgerd77:
Ok. I've now traced this down to happen during the following step respectively along this trace:

```shell
Trace:
    at Common.setHardfork (/ethereumjs-monorepo/packages/common/src/common.ts:316:13)
    at Common.setHardforkBy (/ethereumjs-monorepo/packages/common/src/common.ts:451:10)
    at Blockchain.checkAndTransitionHardForkByNumber (/ethereumjs-monorepo/packages/blockchain/src/blockchain.ts:1346:17)
    at Function.create (/ethereumjs-monorepo/packages/blockchain/src/blockchain.ts:194:24)
    at Function.create (/ethereumjs-monorepo/packages/client/src/blockchain/chain.ts:165:8)
    at /ethereumjs-monorepo/packages/client/test/net/protocol/ethprotocol.spec.ts:122:19
    at runTest (file:///ethereumjs-monorepo/node_modules/@vitest/runner/dist/index.js:486:11)
    at runSuite (file:///ethereumjs-monorepo/node_modules/@vitest/runner/dist/index.js:594:15)
    at runSuite (file:///ethereumjs-monorepo/node_modules/@vitest/runner/dist/index.js:594:15)
    at runFiles (file:///ethereumjs-monorepo/node_modules/@vitest/runner/dist/index.js:645:5)
```

So the outer `chainCommon` from client is changing from `london` to `chainstart` along this blockchain init code in the client:

```ts
console.log(`chain common before: ${options.config.chainCommon.hardfork()}`)
    options.blockchain =
      options.blockchain ??
      (await Blockchain.create({
        db: new LevelDB(options.chainDB),
        common: options.config.chainCommon,
        hardforkByHeadBlockNumber: true,
        validateBlocks: true,
        validateConsensus,
        genesisState: options.genesisState,
        genesisStateRoot: options.genesisStateRoot,
      }))
    console.log(`chain common after: ${options.config.chainCommon.hardfork()}`)
```

I am not fully sure why this didn't happen like this before. I nevertheless thought longer about this and would say that this is the behavior which we want since for blockchain - where the `common` is dynamic (in contrast to tx, block) the `common` (e.g. also the `chainCommon` from client) should move along with the inner Blockchain object.

I have cross-checked in the VM, if this e.g. has side-effects for the case that we instantiate a VM with a certain `common` and with a default blockchain created inside the VM (so, side effect I feared was that the VM common switches over to `chainstart` as well in this case).

This seems not to be the case, have found the following test in the VM in `test/api/index.ts`:

```ts
it('should succeed when common is set to a supported hardfork', async () => {
    const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Byzantium })
    const vm = await VM.create({ common })
    assert.equal(vm.common.hardfork(), Hardfork.Byzantium)
  })
```

So this is basically the described scenario.

Long story short 🙂: I think it is therefore justified to adjust the failing test by adding e.g. an additional `common` for the tx or explicitly reset the `chainCommon`. Will do so and push.

jochem-brouwer:
@holgerd77 Thanks for the investigation and the fix 😃!

holgerd77:
> @holgerd77 Thanks for the investigation and the fix 😃!

Ok, yeah, have pushed, also had another look at the whole code changes, so from my side - if tests pass - we can merge this (cannot approve, started the PR)!

jochem-brouwer (dismissed):
LGTM

Regarding any further small refactoring: at this point I do not have anything to add

jochem-brouwer:
Never mind, CI fails. Does it pass for you locally?

holgerd77:
Ah no, it does not, just tested. This is another test, didn't realized that there were 2 affected files.

jochem-brouwer (approved):
LGTM!

jochem-brouwer (approved):
LGTM!

PR #3317, opened 2024-03-13
Block: use trie.batch

ScottyPoi:
Follow-up to #3313
Continuing work towards optimizations in #3293

### block.genTransactionsTrie()

*Returns the txs trie root for array of `TypedTransaction`*

This method is used during **block building** and **block validation**

An empty trie is contructed from a list of serialized transactions (with the tx index as key)

The root of this trie is the `TransactionsTrieRoot` of a block.

### trie.batch()

The refactored `trie.batch` method in #3313 demonstrated up to **33%** efficiency improvements over the naive `for (tx of txs) {...}` approach.

### Changes:

`block.genTransactionsTrieRoot()` will now construct a trie from a batch input of serialized transactions.

```
    const batchOp: BatchDBOp[] = txs.map((tx, i) => {
      return {
        type: 'put',
        key: RLP.encode(i),
        value: tx.serialize(),
      }
    })
    await trie.batch(batchOp)
```

**This should improve performance during block building and validation**

**TODO:**
- merge #3313
- rebase
- benchmark


holgerd77:
Cool that you even already started on this one 👍, also cool that this is separated into two PRs! 💯 

Is this already "so ready" that I could test this in the client setup I had and see how it goes? Or does this need more work before?

PR #3320, opened 2024-03-13
Add test for checkpoint/revert

yann300:
This test the checkpoint/revert against the contract storage: contract storages don't seem to be reverted as it should.
~~Also this test might reveal an issue with the account storage: in order to send the third tx, the nonce has to be set to 2 although this should be 1 after the revert.~~

### DIFF
--- a/packages/vm/test/api/runTx.spec.ts
+++ b/packages/vm/test/api/runTx.spec.ts
@@ -80,6 +80,179 @@ describe('runTx() -> successful API parameter usage', async () => {
     await simpleRun(vm, 'goerli (PoA), london HF, default SM - should run without errors')
   })
 
+  it('simple run and revert', async () => {
+    const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.London })
+    const vm = await VM.create({
+      common,
+      blockchain: await Blockchain.create({ validateConsensus: false, validateBlocks: false }),
+    })
+    /*
+    pragma solidity >=0.8.2 <0.9.0;
+    contract Storage {
+
+        uint256 public number;
+
+        function incr() public {
+            number++;
+        }
+    }
+    */
+    const bytecode =
+      '0x608060405234801561001057600080fd5b50610164806100206000396000f3fe608060405234801561001057600080fd5b50600436106100365760003560e01c8063119fbbd41461003b5780638381f58a14610045575b600080fd5b610043610063565b005b61004d61007c565b60405161005a9190610091565b60405180910390f35b600080815480929190610075906100b6565b9190505550565b60005481565b61008b816100ac565b82525050565b60006020820190506100a66000830184610082565b92915050565b6000819050919050565b60006100c1826100ac565b91507fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff8214156100f4576100f36100ff565b5b600182019050919050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052601160045260246000fdfea26469706673582212209406edd37c93969bf043c9479e8dd9d97cb54f94deb6c31567a70e62288f9df764736f6c63430008070033'
+    let tx = getTransaction(
+      vm.common,
+      TransactionType.FeeMarketEIP1559,
+      true,
+      '',
+      true,
+      0,
+      bytecode,
+      undefined,
+      200000,
+      1
+    )
+
+    const caller = tx.getSenderAddress()
+    const acc = createAccount(BigInt(0), BigInt(1000000000))
+    await vm.stateManager.putAccount(caller, acc)
+
+    let block = Block.fromBlockData(
+      {
+        header: {
+          number: 0,
+          gasLimit: 8000000,
+          baseFeePerGas: '0x1',
+        },
+        transactions: [tx],
+      },
+      { common }
+    )
+    // 1 - contract deployment
+    let resBlock = await vm.runBlock({
+      block,
+      generate: true,
+      skipNonce: true,
+      skipBlockValidation: true,
+      skipBalance: false,
+    })
+    let res = resBlock.results[0]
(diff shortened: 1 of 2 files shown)


packages/vm/test/api/runTx.spec.ts
```diff
@@ -80,6 +80,141 @@ describe('runTx() -> successful API parameter usage', async () => {
...
+      generate: true,
+      skipNonce: true,
+      skipBlockValidation: true,
+      skipBalance: false,
+    })
+    res = resBlock.results[0]
+
+    assert.equal(
+      bytesToHex(res.execResult.returnValue),
+      '0x0000000000000000000000000000000000000000000000000000000000000000'
```
holgerd77, line 214: Could you state by adding a "should do x..." comment along `assert()` what you actually want to test here?

holgerd77:
Ok, just looked a bit deeper.

One can get additional library-specific logging with something like:

```shell
DEBUG=ethjs,statemanager:statemanager,statemanager:cache:storage,vm:* npx vitest test/api/runTx.spec.ts
```

This one already gives a somewhat good overview of what is happening when, so loggin all the state actions by the `statemanager` logger and the general flow with the `vm` logger (if you add e.g. `evm:*` it quickly gets too noisy).

This gives an output similar to:

```shell
2024-03-14T12:59:50.578Z vm:tx Sender's pre-tx balance is 998573135
2024-03-14T12:59:50.578Z statemanager:statemanager Save account address=0xbe862ad9abfe6f22bcb087716c7d89a26051f74c nonce=1 balance=997583135 contract=no empty=no
2024-03-14T12:59:50.578Z vm:tx Update fromAccount (caller) balance (-> 997583135))
2024-03-14T12:59:50.578Z vm:tx Running tx=0xc1b81dae58f220ddbe27ed0529c4af43db72afe882b48b13c181a3e96e6101c4 with caller=0xbe862ad9abfe6f22bcb087716c7d89a26051f74c gasLimit=68936 to=0x61de9dc6f6cff1df2809480882cfd3c2364b28f7 value=0 data=0x119fbbd4
2024-03-14T12:59:50.578Z statemanager:statemanager Save account address=0xbe862ad9abfe6f22bcb087716c7d89a26051f74c nonce=2 balance=997583135 contract=no empty=no
2024-03-14T12:59:50.578Z statemanager:cache:storage New checkpoint 4
2024-03-14T12:59:50.578Z statemanager:statemanager Save account address=0xbe862ad9abfe6f22bcb087716c7d89a26051f74c nonce=2 balance=997583135 contract=no empty=no
2024-03-14T12:59:50.578Z statemanager:statemanager Save account address=0x61de9dc6f6cff1df2809480882cfd3c2364b28f7 nonce=1 balance=0 contract=yes empty=no
2024-03-14T12:59:50.579Z statemanager:cache:storage Get storage for 61de9dc6f6cff1df2809480882cfd3c2364b28f7
2024-03-14T12:59:50.579Z statemanager:cache:storage Put storage for 61de9dc6f6cff1df2809480882cfd3c2364b28f7: 0000000000000000000000000000000000000000000000000000000000000000 -> 80
2024-03-14T12:59:50.580Z statemanager:cache:storage Get storage for 61de9dc6f6cff1df2809480882cfd3c2364b28f7
2024-03-14T12:59:50.580Z statemanager:cache:storage Get storage for 61de9dc6f6cff1df2809480882cfd3c2364b28f7
2024-03-14T12:59:50.580Z statemanager:cache:storage Put storage for 61de9dc6f6cff1df2809480882cfd3c2364b28f7: 0000000000000000000000000000000000000000000000000000000000000000 -> 01
2024-03-14T12:59:50.580Z statemanager:cache:storage Commit to checkpoint 3
2024-03-14T12:59:50.580Z statemanager:statemanager state checkpoint committed
2024-03-14T12:59:50.580Z vm:tx Update fromAccount (caller) nonce (-> 1)
2024-03-14T12:59:50.580Z vm:tx ----------------------------------------------------------------------------------------------------
2024-03-14T12:59:50.580Z vm:tx Received tx execResult: [ executionGasUsed=22382 exceptionError=none returnValue=0x gasRefund=0 ]
2024-03-14T12:59:50.580Z vm:tx Generated tx bloom with logs=0
```

As some basic comment: our usage of the checkpointing mechanism is basically - at least somewhat implicitly - based on the assumption that things start before block execution at checkpoint level 0 and then go back down there and then things are flushed or whatever. So we have code (in state manager like `if (this._checkpointCount === 0) { await this.flush() }` or similar). That does not mean that you can absolutely not use an additional outer checkpointing layer but it is already some kind of a hack with our system not fully build for the purpose. 😋 

I have some suspicions where things might go wrong (we have e.g. an `OriginalStorageCache` which is re-initialized with new block storage values and where things might already be updated when storage being reverted on 2nd or 3rd block), just to get the broader picture: from what EVM/VM versions are you actually upgrading? So that one get a feeling what your previous code base was. Especially since we updated the state manager caches significantly along the way.

holgerd77:
What you can definitely experiment with is setting the `caches` (storage, account, code) in `StateManager` to 0 respectively deactivate and see if this changes anything. So this needs a custom `StateManager` to passed to the VM (maybe you are doing this already anyhow). Eventually additionally deactivating the Trie caches might also be worth a test (this additionally needs a Trie manually passed to the StateManager 😋 ).

jochem-brouwer:
I have looked into this but I think what is at least missing is proper documentation for EVMs journal (what it actually does).

Also, note that `runTx` calls into EVMs `cleanJournal`. This will clean the journal (reset it). So, your checkpoints will actually get deleted :thinking: The journal is thus not fit for this purpose if `runTx` is used (which logically gets used for your use case).

I will take a deeper look tomorrow (also see discord for some additional questions).

holgerd77:
Will close this, feel free to reopen if felt that there is some need/demand to finalize on this respectively integrate.

PR #3297, opened 2024-03-01
4844 Browser Readiness Releases

holgerd77:
Follow-up on #3261 

This is a (relatively big) release round following relatively shortly after the main Dencun hardfork releases.

Biggest addition is the integration of the [kzg-wasm](https://github.com/ethereumjs/kzg-wasm) library (together with an also browser-compatibility-affecting trie bug) which now makes all 4844 functionality fully browser compatible! 🎉

PR #3321, opened 2024-03-14
Integrate Simplified kzg-wasm Version / Deprecate Util.initKZG()

holgerd77:
This PR integrates a new `kzg-wasm` version from https://github.com/ethereumjs/kzg-wasm/pull/6 providing a simplified interface with a direct setup loading and additionally removes the usage of `Util.loadKZG()` on the monorepo side and deprecates the respective method.

Currently `kzg-wasm` is referenced by Git commit, so this still needs a release before merging here.

### DIFF
--- a/packages/util/src/kzg.ts
+++ b/packages/util/src/kzg.ts
@@ -23,19 +23,6 @@ export interface Kzg {
   ): boolean
 }
 
-function kzgNotLoaded(): never {
-  throw Error('kzg library not loaded')
-}
-
-// eslint-disable-next-line import/no-mutable-exports
-export let kzg: Kzg = {
-  loadTrustedSetup: kzgNotLoaded,
-  blobToKzgCommitment: kzgNotLoaded,
-  computeBlobKzgProof: kzgNotLoaded,
-  verifyKzgProof: kzgNotLoaded,
-  verifyBlobKzgProofBatch: kzgNotLoaded,
-}
-
 /**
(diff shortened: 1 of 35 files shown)


holgerd77:
Additional note: this PR *does not* update the docs. I would do this along #3297 after this has been merged.

g11tech (commented):
looks way cleaner :zap:

holgerd77:
Ok, ready for re-review respectively a final review! 🙂

packages/util/src/kzg.ts
```diff
@@ -32,10 +37,13 @@ export let kzg: Kzg = {
...
+ * @deprecated This initialization method is deprecated since trusted setup loading is done directly in the reference KZG library
+ * initialization or should othewise be assured independently before KZG libary usage.
+ *
  * @param kzgLib a KZG implementation (defaults to c-kzg)
- * @param trustedSetupPath the full path (e.g. "/home/linux/devnet4.txt") to a kzg trusted setup text file
+ * @param a dictionary of trusted setup options
  */
-export function initKZG(kzgLib: Kzg, trustedSetupPath?: string) {
+export function loadKZG(kzgLib: Kzg, _trustedSetupPath?: string) {
   kzg = kzgLib
```
g11tech, line 47:
```suggestion
```
i don't think we need this, as well as the exported kzg object because it was earlier only used in helpers which you have updated

g11tech (approved):
lgtm

PR #3323, opened 2024-03-16
Fix RPCStateManager Inconsistincy

scorbajio:
This change fixes a bug in the `RPCStateManager` that results in an empty account being returned when provider is responding that account does not exist.


jochem-brouwer (commented):
What about the case if the account is in the state, but it is empty? If you would request the account on the DefaultStateManager it will return the empty account, but the RPC will now return `undefined`. 

We can also decide not to fix this, since these accounts got wiped years back with EIP-158 (I guess I am fine with not fixing this)?

holgerd77 (approved):
Ok, since @jochem-brouwer already expressed being ok with not fully fixing towards the old „empty account“ blocks, I will take this in so it can still make it into todays releases! 🤩

PR #3325, opened 2024-03-17
chore(common): Update Optimism hardfork

roninjin10:
Optimism is now on the latest hardfork


acolytec3 (dismissed):

holgerd77 (changes requested):
While this is technically correct, it introduces an inconsistency in behavior towards the other custom (and official) chains.

So, the thing is: we do have a default HF, which is at the moment on `Shanghai` for all forks. If we move *this* single one over to `Cancun`, we basically make an Optimism-only exception.

So the correct thing to do here, would be to just simply remove this hardfork adjustment to berlin, since it's just not necessary anymore. This would the move Optimism (default) to the general (Shanghai) default.

(I know this might be partly unsatisfactory, but consistency should go over one-case-convenience for these kind of cases)

roninjin10:
Interesting can you tell me more about these inconsistent behaviors?  Tevm atm defaults to setting Cancun on all chains would you recommend I revert that back to Shanghai?

holgerd77:
> Interesting can you tell me more about these inconsistent behaviors? Tevm atm defaults to setting Cancun on all chains would you recommend I revert that back to Shanghai?

Ah no, definitely no need to set Tevm to Shanghai.

We just need to set our libraries (respectively mainly the Common library) to *some* default hardfork (respectively the alternative would be to completely skip and always request from the user to provide one. This is not so convenient though).

Unfortunately we cannot update this default HF during normal releases, since this is a breaking change and would break peoples code in various places. So we cannot do these HF updates very often (we roughly do one breaking release round per year to not overload people).

And atm the default HF is just set to Shanghai. The deviation for the Optimism setting to Berlin was just because we could not go as high as Shanghai because Shanghai was just not available during the time of implementation. If we would now go up to Cancun we would introduce an inconsistency regarding the other chains, which would still be on Shanghai (which in my opinion is never a good thing, to make these kind of singular exceptions, even if for this one case things might then be better).

So: that's the story. 🙂

roninjin10:
Ah makes sense thanks for sharing

> This is not so convenient though

Believe it or not I have a rule/convention in tevm that I have to explicitly pass in all params into ethereumjs components even if they are optional.

packages/common/src/common.ts
```diff
@@ -157,8 +157,7 @@ export class Common {
             chainId: 69,
             networkId: 69,
           },
-          // Optimism has not implemented the London hardfork yet (targeting Q1.22)
-          { hardfork: Hardfork.Berlin, ...opts }
+          { hardfork: Hardfork.Cancun, ...opts }
```
roninjin10, line 160:
```suggestion
          { hardfork: Hardfork.Shanghai, ...opts }
```

roninjin10:
```suggestion
          opts
```

acolytec3 (approved):
LGTM

PR #3330, opened 2024-03-19
Replace native `events` with `eventemitter3`

acolytec3:
This PR is an exploration of replacing `eventEmitter` with `eventemitter3` to improve browser compatibility since the native nodejs Event Emitter seems to have some typing issues/interface variance from the browser equivalent.  

- Known impacts - requires removing get/`setMaxListeners` and `prependListener/Once`

- There is some type difference in the generics where our current use of `<E extends keyod T>` in `asyncEventEmitter` that causes Typescript to complain

- The `once` function on the async event emitter does not appear to work properly now.  It appears that the step in the process where it removes the listener after executiong the calback never occurs.


holgerd77:
Starting to think that we might want to move this `eventemitter` thing to the next breaking release round. This is absolutely not urgent and we would cause less hazzle here.

I am loosely targeting a breaking release round in autumn, September or so, then we can collect some more stuff and start a condensed update period after the Summer holidays.

acolytec3:
That would be fine. I'm still not 100% convinced this is the right approach anyway. There are some odd typing issues I'm trying to sort out at the moment to at if I can make it all fit together.

acolytec3:
After merging the vitest 1.0 PR, doesnt't seem like it makes sense to pursue this further since `eventemitter3`'s API isn't a dropin replacement for native events as claimed (see the changes I had to make to `asyncEventEmitter`

PR #3334, opened 2024-03-24
fix: Export EVMOpts

roninjin10:
EVMOpts are part of the EVM class but not publically available. Making it public has a few benefits:
1. Makes it so typedoc can properly link to docs for EVMOpts
2. Makes it so external users extend these options

EVMOpts are the options passed into `EVM.create`


acolytec3 (approved):
LGTM

acolytec3:
Can you run `npm lint:fix` and make the linter happy?

holgerd77 (approved):
LGTM

PR #3335, opened 2024-03-25
New Byte Type Aliases with Byte Length Indicators

holgerd77:
This is on my plate for a very long time, PR adds simple byte aliases for both `Uint8Array` and `PrefixedHexString` with byte length indication.

As discussed already quite some time back ago, this implementation has none of the eventual backlashes as stricter solutions (like performance drawbacks on actual length checks everywhere in the code base) and should at the same time already bring a lot of the benefits, mostly being a lot better documented function signatures and length hints. VSC mouseover of the one exemplaric type in Block e.g. now look like this:

Bildschirmfoto 2024-03-25 um 17 10 36

For the `PrefixedHex*` thing I tested (and "stared" for quite some time 😂) at the different flavors of a possible naming scheme:

- `PrefixedHexString20`
- `PrefixedHexStr20`
- `PrefixedHex20`

I came to the conclusion that the last version should be sufficiently expressive and is at the same time the easiest to grasp, with the need to only digest 3 word parts instead of 4 (which is a lot), and after 3-4 rounds of usage latest it should be very intrinsic that this is about a string (if not clear from the beginning). So this version would be my personal favorite.

I would give this open for a first look and feedback and then continue a bit with it.

We do not need to get to completion here and merge this in a selectively applied state. Then we can just replace types over time, this is otherwise too tedious to accomplish in one go.

### DIFF
--- a/packages/util/src/types.ts
+++ b/packages/util/src/types.ts
@@ -20,10 +20,30 @@ export type BytesLike =
   | TransformabletoBytes
   | PrefixedHexString
 
+/**
+ * Byte type aliases for Uint8Array with a Byte length indicator
+ *
+ * Every type added here should have a corresponding PrefixedHex[NUM] type
+ * (see below).
+ */
+export type Bytes8 = Uint8Array
+export type Bytes20 = Uint8Array
+export type Bytes32 = Uint8Array
+
 /*
  * A type that represents a `0x`-prefixed hex string.
  */
 export type PrefixedHexString = string
+
+/**
+ * Prefixed hex string type aliases for Uint8Array with a Byte length indicator
+ *
+ * Every type added here should have a corresponding Bytes[NUM] type
+ * (see above).
+ */
+export type PrefixedHex8 = string
+export type PrefixedHex20 = string
+export type PrefixedHex32 = string
 
 /**
  * A type that represents an input that can be converted to an Address.
--- a/packages/block/src/types.ts
+++ b/packages/block/src/types.ts
@@ -6,6 +6,9 @@ import type {
   BigIntLike,
   BytesLike,
   JsonRpcWithdrawal,
+  PrefixedHex20,
+  PrefixedHex32,
+  PrefixedHex8,
   PrefixedHexString,
   WithdrawalBytes,
   WithdrawalData,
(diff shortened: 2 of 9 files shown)


acolytec3:
This is lovely.  We should definitely do this across the monorepo.

gabrocheleau:
I like this!

One thing that came up while searching the monorepo, is that we have some exported Bytes32 and other similar types in the client, but in that case they served as aliases for `string`. It could be helpful to remove that ambiguity (in my view Bytes should be Uint8Array, while PrefixedHex should be strings).  Link: https://github.com/ethereumjs/ethereumjs-monorepo/blob/new-byte-type-aliases/packages/client/src/rpc/modules/engine/types.ts#L40

packages/util/src/types.ts
```diff
@@ -20,11 +20,31 @@ export type BytesLike =
...
 
+/**
+ * Prefixed hex string type aliases for Uint8Array with a Byte length indicator
+ *
+ * Every type added here should have a corresponding Bytes[NUM] type
+ * (see above).
+ */
+export type PrefixedHex8 = string
+export type PrefixedHex20 = string
+export type PrefixedHex32 = string
```
gabrocheleau, line 46: Should these be updated to PrefixedHexString rather than string?

holgerd77: Do not see the advantage of such alias chaining TBH.

holgerd77:
> One thing that came up while searching the monorepo, is that we have some exported Bytes32 and other similar types in the client, but in that case they served as aliases for `string`. It could be helpful to remove that ambiguity (in my view Bytes should be Uint8Array, while PrefixedHex should be strings). Link: https://github.com/ethereumjs/ethereumjs-monorepo/blob/new-byte-type-aliases/packages/client/src/rpc/modules/engine/types.ts#L40

Yes, that makes definitely sense to align in client, would also agree with the Bytes == Uint8Array stuff.

The whole PR is currently (and maybe permanently block). Context is that the type aliases do not propagate through in the editor (Visual Studio Code), so hovering over the code then shows `Uint8Array` again instead of e.g. `Bytes32` (see chat exchange with Jochem somewhere). This makes the whole thing too brittle and half-useless and we should see if we find something both more stable and robust and still non-breaking (might not be a solution here).

Jochem actually tried the same thing as here already some time ago (didn't know that before) and stumbled upon the very same thing (and therefore drew the same conclusion here).

We'll see. Would be really really nice to have something like this.

Maybe also worth to have a look at other low-level libraries for some inspiration.

packages/util/src/types.ts
gabrocheleau: Ah you're right, but I was thinking in the context of that other PR (where PR was updated to `0x${string}`).

holgerd77:
(and once we have found a robust approach, it should be quick to realize with AI)

gabrocheleau:
> @gabrocheleau I still think that giving byte length hints (or better: checks) would be one of the most valuable structural improvements we could bring to the monorepo. This is so central to the well-behavior of the whole system, especially the EVM.
> 
> This approach does not seem to work though, agree, will close.

Agreed that this would be great to have when the tooling/TS picks up the feature

PR #3336, opened 2024-03-26
Implement EIP 7002 (exits)

jochem-brouwer:
https://eips.ethereum.org/EIPS/eip-7002

Ref: https://github.com/ethereum/EIPs/commit/35589d35c40576d932923542b31a4d8f7812c3e7

PR #3338, opened 2024-03-28
block: make withdrawals trie check consistent with txtrie

jochem-brouwer:
When working on #3336 I noticed some inconsistent things unrelated to EIP 7002.

This PR fixes that. (Have currently set it up to not run CI yet since I might want to add other things I find)

Currently:

- Adds withdrawalsTrie cache (we do not want to re-calculate the withdrawals trie each time)
- Speeds up withdrawals trie calculation if there are no withdrawals
- Speeds up uncle hash calculation if there are no uncles (now each time there are no uncles we return `return equalsBytes(this.keccakFunction(RLP.encode([])), this.header.uncleHash)`, this is unnecessary especially since we currently have no uncles)


acolytec3 (approved):
LGTM

acolytec3 (approved):
LGTM

PR #3344, opened 2024-04-04
Vitest browser cleanup

acolytec3:
Removes a couple of unnecessary deps that were previously used for vitest browser testing and missed during #3281


jochem-brouwer (approved):
LGTM

PR #3345, opened 2024-04-06
modify rpcStataManager getAccount

rodrigoherrerai:
When getting an account from the provider that has never been touched, the returned codeHash is not keccak(""), but just empty bytes (0...0).  Therefore, the getAccount function will not return "undefined" but an account with empty bytes. 

This creates a problem in some edge cases. For example, when forking a network and creating a contract within a contract, the evm will revert with collision error, due to this implementation.


scorbajio:
This looks good to me. Would be useful to capture some data and add a unit test.

acolytec3:
> Would be useful to capture some data and add a unit test.

This should already be covered in the existing tests.  I've also updated our `mockProvider` to return a response that mirrors the Infura API response to `getProof` for non-existent accounts.

acolytec3:
Have also updated the check of the `codeHash` field to compare the underlying bytes instead of `.every(...` notation per some internal discussion.

acolytec3 (approved):
LGTM!  Thanks for the PR.

PR #3347, opened 2024-04-06
Fix ts-doc typo

pldespaigne:
The ts-doc example code block was not closed with ` ``` `, as a result the [docs page](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/docs/README.md#fetchfromprovider) is wrongly rendered.

I just fixed it by adding the missing ` ``` ` 👍


acolytec3 (approved):
LGTM!  Thanks for fixing this.

PR #3348, opened 2024-04-07
util: update prefixedHexString to use literal string template

gabrocheleau:
TypeScript 4.1 introduced template literal types: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html 

We can use those to restrict the type of some strings further with no runtime performance downside. This PR updates PrefixedHexString from `string` to `0x{string}`


holgerd77:
Oh, that’s actually a hyper-great one, love this! 🤩🙏❤️

holgerd77:
(feel that I should really start looking into TypeScript (updates) a bit more, maybe/likely there is other new base functionality we can leverage in a useful and very targeted way. Also, we can actually generally look into updating our Typescript version (what is a bit tricky and what I haven't given much thought yet: what is actually with TypeScript backwards compatibility, so if we update to a new version and use some of the very latest features. Then we likely also break other people's TypeScript code. 🤔 So guess this also needs at least some basic thinking and following guidelines we set for ourselves (maybe in a first round: just not get too hyper-progressive with this stuff and not fully jump on the latest stuff from 2 weeks ago (atm we are not in danger with our `4.7.4` usage from (just googled...) mid 2022. 😂).

acolytec3 (approved):
LGTM

PR #3349, opened 2024-04-07
client: fix misc tsc issues

gabrocheleau:
This PR addresses misc TypeScript compile issues that were being flagged up by `tsc`.


holgerd77:
Updated this via UI

packages/client/test/sync/fetcher/accountfetcher.spec.ts
```diff
@@ -280,7 +280,7 @@ describe('[AccountFetcher]', async () => {
   })
 
   it('should reject zero-element proof if elements still remain to right of requested range', async () => {
-    const config = new Config({ transports: [], accountCache: 10000, storageCache: 1000 })
+    const config = new Config({ accountCache: 10000, storageCache: 1000 })
```
holgerd77, line 283: Might be off with this, but I think setting the `transports` here to empty was done to not have e.g. the RLPx layer started on tests here (and other occurrences as well). So this change might have the side-effect to reintroduce this for the test runs which is not what we want. We should double check here.

(I generally find this a bit problematic to do an eventual functional change just for typing reasons. What was the complaint here? Is there maybe a solution possible to adjust the type or something? 🤔 )

gabrocheleau: The actual TS issue in this case is the following:

```
Object literal may only specify known properties, and 'transports' does not exist in type 'ConfigOptions'.ts(2353)
(property) transports: never[]
```
It seems like the `transports` property was removed in this PR: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3069 

Not sure why this should be kept?

acolytec3: Second what @gabrocheleau said.  `transports` no longer exists as a config option since we only support RPLx and not libp2p.

acolytec3 (approved):
LGTM!

packages/client/test/sync/fetcher/accountfetcher.spec.ts
holgerd77: 👍

PR #3350, opened 2024-04-08
Bring 2537 back

acolytec3:
Brings back our original EIP implementation of 2537.

- Reverts commit 31aa264 
- Updates `mcl-wasm` to latest versions
- Removes browser exceptions
- Adds tests using test vectors from [EIP](https://github.com/ethereum/EIPs/blob/master/assets/eip-2537/test-vectors.md)


packages/evm/src/evm.ts
```diff
@@ -129,6 +131,15 @@ export class EVM implements EVMInterface {
     return this._opcodes
   }
 
+  protected _isInitialized: boolean = false
```
jochem-brouwer, line 134: This initializer should be removed, the constructor is protected and to create an EVM you `await EVM.create`

packages/evm/src/evm.ts
```diff
@@ -237,6 +252,22 @@ export class EVM implements EVMInterface {
       typeof window === 'undefined' ? process?.env?.DEBUG?.includes('ethjs') ?? false : false
   }
 
+  protected async init(): Promise<void> {
```
jochem-brouwer, line 255: Can we remove this method and put this into `evm.create`?

packages/evm/src/precompiles/index.ts
```diff
@@ -138,6 +147,88 @@ const precompileEntries: PrecompileEntry[] = [
     precompile: precompile0a,
     name: 'KZG (0x0a)',
   },
+  // 0x00..0b: beacon block root, see PR 2810
```
jochem-brouwer, line 150: Where did the beacon block root precompile go to?

packages/evm/src/precompiles/util/bls12_381.ts
```diff
@@ -0,0 +1,344 @@
...
+import { ERROR, EvmError } from '../../exceptions.js'
+
+// base field modulus as described in the EIP
+const fieldModulus = BigInt(
+  '0x1a0111ea397fe69a4b1ba7b6434bacd764774b84f38512bf6730d2a0f6b0f6241eabfffeb153ffffb9feffffffffaaab'
+)
+
+// gas discount pairs taken from EIP-2537 `Bls12381MultiExpGasDiscount` parameter
+export const gasDiscountPairs = [
+  [1, 1200],
```
jochem-brouwer, line 19: Hmm, this is also in common, what is it doing here? :thinking:

jochem-brouwer (commented):
Some comments. We should also add more tests! (Or import / test locally by using old ethereum/tests vectors). Now the precompiles are not tested :thinking:

packages/evm/src/precompiles/util/bls12_381.ts
acolytec3: Not 100% sure but I'd honestly rather have it here than in `common` since it throws off the typing for the `common` params dictionary.  WDYT?

jochem-brouwer: That's also fine, but then it should be removed from common.

packages/evm/src/precompiles/index.ts
acolytec3: Good question :thinking:  Will have to look

packages/evm/src/precompiles/util/bls12_381.ts
holgerd77: Yeah, I thought the same yesterday when having a quick look and wondered if it might make sense to move this *out of Common*, to not have the generic Common configuration bloated too much and getting the typing unnecessarily extensive.

jochem-brouwer (approved):
LGTM

PR #3351, opened 2024-04-09
devp2p: increase timeout for failing CI test

gabrocheleau:
This PR increases the timeout for a devp2p test that sometimes led to failures in the CI due to timeout.


holgerd77 (approved):
Not sure if we have a „promise“ that this deterministically solves things, will nevertheless merge since improvement is likely and in the other direction can’t hurt.

PR #3357, opened 2024-04-16
monorepo: PrefixedHexString related type fixes

gabrocheleau:
This PR addresses misc. type issues that made their way into the monorepo after #3348


holgerd77 (approved):
Have updated the PR, and tests are passing.

Nice, this looks super great! 🤩

A general note: I DO think that we will likely introduce some update problems for upstream libraries/users with this. This is somewhat breaking on the type level, both if input and output values of API calls are tightened.

I will nevertheless merge this in now, so much work and really worth the additional security guarantees coming with it. This is likely nevertheless at its limits of what we can/should do along non-breaking releases.

Two thoughts here:

1. I wonder if we want to make this an "official" policy (so: write down somewhere prominently (likely: root monorepo docs) in the docs + announce in Twitter, that we will be doing minor release versions which might be breaking on the TypeScript level and if people want to be safe there they should use `~` version ranges. Would this make sense?

2. I also wonder if we would want to downgrade here again and re-type at least prominent API calls (most obvious one: `hexToBytes`) from `PrefixedHexString` to `string` and add this instead to a task to #3216. We now with this already have all the pre-conditions in place, so still 95% of the way done, and with this we would ease this a bit for our users.

I think I am somewhat in favor to do.

Nevertheless: great stuff! 👍 🙂

PR #3359, opened 2024-04-18
Add ESM build to client

acolytec3:
This replaces the current client build process with one that is ESM only.

Notable changes
- Adds a `"type":"module"` to `package.json`
- Adds a new ESM build step for `npm run build`
- Removes all `require(...` usage in `client`
- Revises imports of current CommonJS imports to use default imports and then destructuring as needed
- Removes `fs-extra` dependency since not needed (and was causing issues with the ESM build)
- Adds some typing in `bin/cli.ts` for the `servers` bit

TODOs:
- Completely remove the now non-functional CJS build


Dockerfile
```diff
@@ -14,13 +14,13 @@ WORKDIR /usr/app
 COPY --from=build /usr/app .
 
 # Sanity check
-#RUN node /usr/app/packages/client/dist/bin/cli.js --help
-RUN node /usr/app/node_modules/.bin/ethereumjs --help
+RUN node /usr/app/packages/client/dist/esm/bin/cli.js --help
+#RUN node /usr/app/node_modules/.bin/cjs/ethereumjs --help
```
jochem-brouwer, line 18: Why is this commented out?

packages/client/bin/cli.ts
```diff
@@ -488,17 +487,23 @@ function initDBs(config: Config): {
...
+  mkdirSync(stateDataDir, {
+    recursive: true,
+  })
   const stateDB = new Level<string | Uint8Array, string | Uint8Array>(stateDataDir)
 
   // Meta DB (receipts, logs, indexes, skeleton chain)
   const metaDataDir = config.getDataDirectory(DataDirectory.Meta)
-  ensureDirSync(metaDataDir)
+  mkdirSync(metaDataDir, {
+    recursive: true,
```
jochem-brouwer, line 505: Why is the recursive now necessary?

packages/client/src/util/rpc.ts
```diff
@@ -222,7 +224,7 @@ export function createWsRPCServerListener(opts: CreateWSServerOpts): HttpServer
     if (typeof rpcCors === 'string') app.use(cors({ origin: rpcCors }))
     httpServer = createServer(app)
   }
-
+  //@ts-ignore
```
jochem-brouwer, line 227: Why is this necessary?

jochem-brouwer (commented):
Some initial comments

packages/client/bin/cli.ts
acolytec3: I copied this code exactly from what `fs-extra` was doing.  Probably not necessary so will remove.

packages/client/src/util/rpc.ts
acolytec3: Looks like it isn't.  Have removed.

packages/client/bin/cli.ts
acolytec3: So have added these back to the PR.

packages/client/archive/libp2p/index.ts
```diff
@@ -8,9 +8,9 @@ import { Config } from '../src/config'
 import { LevelDB } from '../src/execution/level'
 import { parseMultiaddrs } from '../src/util'
 
-import { getLogger } from './logging'
+import { getLogger } from './logging.js'
 // Blockchain
-export * from '../src/blockchain/chain'
+export * from '../src/blockchain/chain/index.js'
```
gabrocheleau, line 13: Why are we adding that `.js` suffix in some places but not all?

packages/client/bin/startRpc.ts
```diff
@@ -176,7 +176,7 @@ export function startRPCServers(client: EthereumClient, args: RPCArgs) {
       rpcDebug,
       rpcDebugVerbose,
       logger,
-    })
+    }) //@ts-ignore
```
gabrocheleau, line 179: Why is this necessary? Can we just typecast instead?

packages/client/scripts/postBuildFixes.sh
```diff
@@ -0,0 +1,54 @@
...
+    printf "${DIM}$1${NOCOLOR}"
+}
+
+post_build_fixes() {
+    blue "[Post Build Fixes]"
+    if [ -f ./dist/esm/bin/cli.js ];
+    then
+        echo "Adding ./dist/cjs/package.json"
+        rm -f ./dist/cjs/package.json
+        cat <<EOT >> ./dist/cjs/package.json
```
gabrocheleau, line 36: Why do we need to add a cjs package.json if we're removing cjs support in the client?

packages/client/src/sync/fetcher/storagefetcher.ts
```diff
@@ -15,15 +15,15 @@ import {
 import debugDefault from 'debug'
 import { keccak256 } from 'ethereum-cryptography/keccak'
 
-import { short } from '../../util'
+import { short } from '../../util/index.js'
```
gabrocheleau, line 18:
```suggestion
import { short } from '@ethereumjs/util'
```

packages/client/src/util/rpc.ts
```diff
@@ -22,13 +23,13 @@ type CreateRPCServerOpts = {
   logger?: Logger
 }
 type CreateRPCServerReturn = {
-  server: RPCServer
+  server: any
```
gabrocheleau, line 26: Can't we type those two as jayson.Server?

gabrocheleau (dismissed):
Left a couple of questions/comments. Looks great

packages/client/archive/libp2p/index.ts
acolytec3: I shouldn't have addit it all in this `archive` folder. Just find/replace errors.  We can safely disregard anything in this folder.

packages/client/bin/startRpc.ts
acolytec3: I think this was an earlier left over.  Looks like we don't need it at all.  Will remove.

packages/client/scripts/postBuildFixes.sh
acolytec3: I just hadn't gone this far yet.  Will remove this entire section of the build script if everyone is good with dropping a CJS build.

packages/client/src/util/rpc.ts
acolytec3: Indeed we can.  Thanks for catching this!

gabrocheleau (approved):

PR #3361, opened 2024-04-22
monorepo: cleanup ts-ignore and ts-expect-error

gabrocheleau:
This PR cleanups several instances where ts-ignore or ts-expect-errors was used, instead replacing them with less intrusive TypeScript solutions.

Notably, these TypeScript tags were very often used to ignore TS errors related to access of protected object properties. Instead, these properties can be accessed (without error) by the string literal. 

Here's an example:
```
 TS error:
if (this._eciesSession._gotEIP8Auth) {

No TS error:
if (this._eciesSession['_gotEIP8Auth']) {
 ```

### DIFF
--- a/packages/vm/test/api/runBlock.spec.ts
+++ b/packages/vm/test/api/runBlock.spec.ts
@@ -23,6 +23,7 @@ import type {
   PreByzantiumTxReceipt,
   RunBlockOpts,
 } from '../../src/types'
+import type { DefaultStateManager } from '@ethereumjs/statemanager'
 import type { TypedTransaction } from '@ethereumjs/tx'
 
 const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Berlin })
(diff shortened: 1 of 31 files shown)

### CODE REVIEW

File: packages/vm/test/api/runBlock.spec.ts [hunk 1]
```diff
@@ -464,15 +459,14 @@ describe('runBlock() -> tx types', async () => {
     const blockRlp = toBytes(testData.blocks[0].rlp)
     const block = Block.fromRLPSerializedBlock(blockRlp, { common, freeze: false })
 
-    //@ts-ignore overwrite transactions
+    //@e transactions
     block.transactions = transactions
 
     if (transactions.some((t) => t.supports(Capability.EIP1559FeeMarket))) {
-      // @ts-ignore overwrite read-only property
+      // @e read-only property
```
holgerd77, line 466: Won’t make it a blocker, but this and above seems like a leftover?

holgerd77 (approved):
Nice, LGTM! 🙂👍

PR #3363, opened 2024-04-22
Client: Better --execution Flag Guard

holgerd77:
When I just tested the client on the ESM build with `npm run client:start -- --execution=false` I realized that the execution flag is ignored and execution started nevertheless.

This is likely since we have various execution entrypoints atm and it is likely that at least one gets triggered in an non-expected context.

This small PR adds a better guard in the execution `run()` method directly to avoid this kind of scenarios. Should be safe to merge since the `--execution` flag is used with very much a clear intention and so this cannot have side effects on beacon sync or whatsoever on normal runs.


g11tech (approved):
lgtm although we could just not even allocate/start vmexecution (can be done separately )

jochem-brouwer (approved):

PR #3367, opened 2024-04-24
Return custom hardforks properly

axic:


packages/common/src/common.ts
```diff
@@ -962,7 +962,7 @@ export class Common {
   hardforks(): HardforkTransitionConfig[] {
     const hfs = this._chainParams.hardforks
     if (this._chainParams.customHardforks !== undefined) {
-      this._chainParams.customHardforks
+      return this._chainParams.customHardforks
```
axic, line 965: I'm not 100% sure what the goal of this code was, but it seems to be wrong without the `return`.

axic: Or is this meant to do concatenation?

jochem-brouwer: I added this but I have no idea what it was supposed to do :thinking: Have to check, maybe this indeed should be concatenation :thinking:

axic: Good inert code then 😅

PR #3372, opened 2024-04-25
Implement EIP 7685

acolytec3:
Implements [EIP-7685](https://eips.ethereum.org/EIPS/eip-7685)


packages/block/src/block.ts
```diff
@@ -51,6 +51,7 @@ export class Block {
   public readonly transactions: TypedTransaction[] = []
   public readonly uncleHeaders: BlockHeader[] = []
   public readonly withdrawals?: Withdrawal[]
+  public readonly requests?: CLRequestType<any>[]
```
jochem-brouwer, line 54: Isn't `ELRequestType` be more appropriate here? :thinking:

packages/block/src/block.ts
```diff
@@ -92,6 +94,24 @@ export class Block {
...
+   * @returns a 32 byte Uint8Array representing the requests trie root
+   */
+  public static async genRequestsTrieRoot(requests: CLRequest[], emptyTrie?: Trie) {
+    const sortedRequests = requests?.sort((a, b) => {
+      if (a.type !== b.type) return a.type - b.type
+      return a.greaterThan(b) === true ? 1 : -1
+    })
+    const trie = emptyTrie ?? new Trie()
+    for (const [i, req] of sortedRequests.entries()) {
+      await trie.put(RLP.encode(i), req.serialize())
```
jochem-brouwer, line 110: It seems that this only sorts the types, but not the order of the requests of the same operation. Is this correct?

jochem-brouwer: It is to be defined by the operation, but I think the ordering of the same-type requests depend on whatever sorting algorithm is being used internally - right?

jochem-brouwer (commented):
Left some comments. I think we need a sorting mechanism for any EIP which implements this. 

For 6110:
`Each deposit accumulated in the block must appear in the EIP-7685 requests list in the order they appear in the logs`

For 7002:
`It should be in the order it is pulled out of the queue (FIFO)`

So this is not lexicographic, but in the order of how they are executed or in whatever order they get added to the FIFO queue.

packages/block/src/block.ts
acolytec3: We expect each request type to define a `greaterThan` function that we then use in a simple array `sort` method to take care of sorting.

jochem-brouwer: Oh right :man_facepalming: I failed to see that this was already in the sort method.

packages/block/src/block.ts
```diff
@@ -414,7 +447,8 @@ export class Block {
     uncleHeaders: BlockHeader[] = [],
     withdrawals?: Withdrawal[],
     opts: BlockOptions = {},
-    executionWitness?: VerkleExecutionWitness | null
+    executionWitness?: VerkleExecutionWitness | null,
+    requests?: CLRequest[]
```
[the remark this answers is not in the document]
acolytec3: Good point.  Lemme move it.

packages/block/src/block.ts
acolytec3: My weakly held opinion is that this should be `CLRequests` since we're making a request to the Consensus Layer, more so than "from" the EL.

holgerd77: Yeah, would agree after reading the EIP

packages/common/src/eips.ts
```diff
@@ -542,4 +542,13 @@ export const EIPs: EIPsDict = {
       },
     },
   },
+  7685: {
+    comment: 'General purpose execution layer requests',
+    url: 'https://eips.ethereum.org/EIPS/eip-7685',
+    status: Status.Draft,
+    // TODO: Set correct minimum hardfork
+    minimumHardfork: Hardfork.Shanghai,
```
holgerd77, line 550: Guess this also can be kept as "Shanghai".

acolytec3: I'm updating it to Cancun for now until we decide how to handle the potentially undefined fields for `blobGasUsed`, `excessBlobGas`, and `parentBeaconBlockHash`, that exist in the blockheader values array if we don't activate 4844 and 4788 alongside 7685.

packages/block/src/block.ts
g11tech: yes CL requests sound more appropriate, or we can call it, ELTriggeredCLRequests if we want to be more verbose

g11tech:
> Left some comments. I think we need a sorting mechanism for any EIP which implements this.

We discussed this on call (@acolytec3 and me), 7685 should have a code block to gather the requests from other EIPs in the sequence ordered by type and then there should be an exact match of requests in blockbody/request root.

packages/vm/src/runBlock.ts
```diff
@@ -285,6 +285,12 @@ export async function runBlock(this: VM, opts: RunBlockOpts): Promise<RunBlockRe
     preimages: result.preimages,
   }
 
+  if (this.common.isActivatedEIP(7685)) {
```
g11tech, line 288: we can do request field existence check even before we run the block

Also we need to pop requests out before we verify stateroot (and call verkle verify state)

g11tech (approved):
lgtm

g11tech (dismissed):

g11tech (dismissed):

packages/block/test/eip7685block.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+    )
+
+    assert.ok(await block.requestsTrieIsValid())
+
+    // Throws when requests are not ordered correctly
+    await expect(async () =>
+      Block.fromBlockData(
+        {
+          requests: [request1, request3, request2],
+          header: { requestsRoot },
```
jochem-brouwer, line 85: For completeness maybe also calculate the correct `requestsRoot` for this requests array? (I am assuming that the `toThrow` thing checks for the correct error message, though?)

packages/common/src/eips.ts
```diff
@@ -542,4 +542,13 @@ export const EIPs: EIPsDict = {
       },
     },
   },
+  7685: {
+    comment: 'General purpose execution layer requests',
+    url: 'https://eips.ethereum.org/EIPS/eip-7685',
+    status: Status.Draft,
+    // TODO: Set correct minimum hardfork
```
jochem-brouwer, line 549: TODO left here, I would assume that this can be done from Merge.

packages/util/src/requests.ts
```diff
@@ -0,0 +1,28 @@
...
+  type: number
+  bytes: Uint8Array = new Uint8Array()
+  constructor(type: number, bytes: Uint8Array) {
+    if (type === undefined) throw new Error('request type is required')
+    this.type = type
+    this.bytes = bytes
+  }
+
+  serialize() {
+    return concatBytes(Uint8Array.from([this.type]), this.bytes)
```
jochem-brouwer, line 26: Nit: this does not work for types > 255. (But since we are super far from having to support this it is fine)

packages/vm/src/buildBlock.ts
```diff
@@ -316,6 +320,14 @@ export class BlockBuilder {
       blobGasUsed = this.blobGasUsed
     }
 
+    let requests
+    let requestsRoot
+    if (this.vm.common.isActivatedEIP(7685)) {
+      const requests = await accumulateRequests(this.vm)
```
jochem-brouwer, line 326: Doesn't this `const requests` now flag the `let requests` above so the requests are always undefined?

packages/vm/src/buildBlock.ts
```diff
@@ -338,7 +351,9 @@ export class BlockBuilder {
       header: headerData,
       transactions: this.transactions,
       withdrawals: this.withdrawals,
+      requests,
     }
+    console.log(blockData)
```
jochem-brouwer, line 356: `console.log`

packages/vm/src/runBlock.ts
```diff
@@ -208,66 +215,89 @@ export async function runBlock(this: VM, opts: RunBlockOpts): Promise<RunBlockRe
...
-    if (equalsBytes(result.receiptsRoot, block.header.receiptTrie) === false) {
-      if (this.DEBUG) {
-        debug(
-          `Invalid receiptTrie received=${bytesToHex(result.receiptsRoot)} expected=${bytesToHex(
-            block.header.receiptTrie
-          )}`
-        )
+  } else {
+    if (this.common.isActivatedEIP(7685)) {
+      const valid = await block.requestsTrieIsValid()
```
jochem-brouwer, line 233: Shouldn't we do this check before running all txs?

packages/vm/src/runBlock.ts
```diff
@@ -192,6 +192,13 @@ export async function runBlock(this: VM, opts: RunBlockOpts): Promise<RunBlockRe
     throw err
   }
 
+  let requestsRoot
+  let requests: CLRequest[] | undefined
+  if (block.common.isActivatedEIP(7685)) {
+    const requests = await accumulateRequests(this)
```
jochem-brouwer, line 198: I think here the `const requests` again flag the `let requests` above (I actually just checked and it does). So the `const` should be removed. (`let requests` stays undefined always)

packages/vm/src/runBlock.ts
```diff
@@ -918,3 +950,24 @@ const DAOConfig = {
...
+ * @returns an list of CL requests in ascending order by type
+ */
+export const accumulateRequests = async (_vm: VM): Promise<CLRequest[]> => {
+  const requests: CLRequest[] = []
+
+  // TODO: Add in code to accumulate deposits (EIP-6110)
+
+  // TODO: Add in code to accumulate partial withdrawals (EIP-7002)
+
+  if (requests.length > 1) {
```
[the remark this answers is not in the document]
jochem-brouwer: I agree that this check is not necessary if we check the requests in the right order above.

jochem-brouwer (commented):
Looks good so far! I got a few comments. I like the tests :+1:

packages/block/src/block.ts
```diff
@@ -139,11 +171,28 @@ export class Block {
     }
 
     const withdrawals = withdrawalsData?.map(Withdrawal.fromWithdrawalData)
+
+    // Requests should be sorted in monotonically ascending order based on type
+    // and whatever internal sorting logic is defined by each request type
+    if (clRequests !== undefined && clRequests.length > 1) {
+      for (let x = 1; x < clRequests.length; x++) {
+        if (clRequests[x].type < clRequests[x - 1].type)
+          throw new Error('requests are not sorted in ascending order')
```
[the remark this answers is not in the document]
acolytec3: Fair point.  Will move

packages/block/src/types.ts
```diff
@@ -286,4 +302,5 @@ export type ExecutionPayload = {
   parentBeaconBlockRoot?: PrefixedHexString | string // QUANTITY, 64 Bits
   // VerkleExecutionWitness is already a hex serialized object
   executionWitness?: VerkleExecutionWitness | null // QUANTITY, 64 Bits, null implies not available
+  // TODO: Determine if we need the requestsRoot here
```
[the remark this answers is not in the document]
acolytec3: @g11tech, agree?

packages/block/test/eip7685block.spec.ts
acolytec3: This is happening in line 66 above (assuming the requests are in the correct order). There is no valid `requestsRoot` for a requests array that isn't in the correct order.

The `toThrow` check examines the error message to confirm that the error is regarding the order of the requests (i.e. throws if the request array isn't monotonically increasing)

packages/common/src/eips.ts
acolytec3: This one isn't obvious to me.  Unless we do some more structured work around `raw()` function in the `Blockheader` class, not sure we can have less than Cancun here.  See https://github.com/ethereumjs/ethereumjs-monorepo/pull/3372#discussion_r1581698444

packages/util/src/requests.ts
acolytec3: Not disagreeing with you but the EIP specified the first byte as the type byte, so this is according to spec.

packages/vm/src/buildBlock.ts
acolytec3: Good catch.  Thanks!

packages/vm/src/runBlock.ts
acolytec3: If the `generate` option is called, the `requestsRoot` passed in the `block` will be disregarded and the correct one derived so checking before we validate the `generate` flag doesn't make sense.  We validate it here if we are re-executing a block to ascertain its validity.

packages/block/test/eip7685block.spec.ts
jochem-brouwer: Ok, since `toThrow` checks for the correct error this is fine :smile:

packages/common/src/eips.ts
jochem-brouwer: Hmm right this is a good point, I think for `raw` it will work: 

```typescript
    if (this.common.isActivatedEIP(4844)) {
      rawItems.push(bigIntToUnpaddedBytes(this.blobGasUsed!))
      rawItems.push(bigIntToUnpaddedBytes(this.excessBlobGas!))
    }
    if (this.common.isActivatedEIP(4788)) {
      rawItems.push(this.parentBeaconBlockRoot!)
    }
    if (this.common.isActivatedEIP(7685) === true) {
      rawItems.push(this.requestsRoot!)
    }
```

However we get in trouble in `valuesArrayToHeaderData` since this assumes that 4788, 4844 is active in case we activate 7685. So you are correct, in our current codebase the minimum is Cancun.

packages/util/src/requests.ts
jochem-brouwer: Ah good point, the EIP itself does not support types > 255. Will resolve.

packages/vm/src/runBlock.ts
jochem-brouwer: If `generate` is called we generate the `requestsRoot` and put that in the block (this is correct). However, this check is done in case we are not using `generate` (so `generate` is `false`). The `block.requestsTrieIsValid` is called on whatever block is provided - this itself, should not change (so this can be done before running any txs). However what you also want to check is that the `block.genRequestsTrieRoot(requests)` (so not `block.requests`, but the `requests` we just have accumulated in `accumulateRequests(this)`) equals the provided `block.header.requestsRoot`. If this is not the case, then the block thus has a different requests root and the block is incorrect.

packages/block/src/types.ts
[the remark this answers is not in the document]
jochem-brouwer: I found this in the CL specs: https://github.com/ethereum/consensus-specs/blob/7bf43d1bc4fdb91059f0e6f4f7f0f3349b144950/specs/electra/beacon-chain.md#executionpayload

It is not updated yet (it is on old 6110 / 7002 which did not yet use 7685), but I would strongly assume that thus the `requestsRoot` is added.

packages/vm/src/runBlock.ts
```diff
@@ -918,3 +950,24 @@ const DAOConfig = {
...
+ * This helper method generates a list of all CL requests that can be included in a pending block
+ * @param _vm VM instance from which to derive CL requests
+ * @returns an list of CL requests in ascending order by type
+ */
+export const accumulateRequests = async (_vm: VM): Promise<CLRequest[]> => {
+  const requests: CLRequest[] = []
+
+  // TODO: Add in code to accumulate deposits (EIP-6110)
+
+  // TODO: Add in code to accumulate partial withdrawals (EIP-7002)
```
holgerd77, line 964: Simply for my understand: so here a call into `read_withdrawal_requests()` from https://eips.ethereum.org/EIPS/eip-7002#execution-layer ("System Call" section) would take place, right?

Side question for @acolytec3, since you likely have better imagination on this: how does this affect execution with e.g. RPCStateManager? Does this still work (do not yet get the pieces together in my head)?

packages/vm/test/api/EIPs/eip-7685.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    const res = await vm.runBlock({
+      block: emptyBlock,
+      generate: true,
+    })
+    assert.equal(res.gasUsed, 0n)
+  })
+  it('should error when an invalid requestsRoot is provided', async () => {
+    const vm = await setupVM({ common })
+    const emptyBlock = Block.fromBlockData(
+      { header: { requestsRoot: randomBytes(32) } },
```
holgerd77, line 41: Nit: does this randomness improve anything here or should we rather go with a static value here (same below)?

holgerd77 (commented):
Looks absolutely amazing 🤩, also agree with Jochem, test cases added are really fabulous and already covering such an amazingly broad scope of the added functionality!

Some small comments but nothing major to be added from my review here.

packages/util/src/requests.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export interface CLRequestType {
+  readonly type: number
+  readonly bytes: Uint8Array
+  serialize(): Uint8Array
+}
+
+export class CLRequest implements CLRequestType {
+  type: number
+  bytes: Uint8Array = new Uint8Array()
```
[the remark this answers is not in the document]
holgerd77: Guess comment is misplaced here? 😅

packages/common/src/eips.ts
acolytec3: It doesn't really even make sense in `raw` since we aren't checking to verify that the items we're pushing into the array are in the correct position.  The blob gas fields should always come before the `parentBeaconBlockRoot` but if you have 4788 activated and not 4844, you would get the parentBeaconBlockRoot in the position where our code expects the blob related fields to be

packages/vm/src/runBlock.ts
acolytec3: Ah, that makes sense.  Will push an update shortly.

packages/vm/src/runBlock.ts
acolytec3: To the question for me, I don't "think" this should have any impact.  We already track contract state updates in a local cache in the RPCStateManager so it should keep track of any changes we make in the processing of requests.

packages/vm/test/api/EIPs/eip-7685.spec.ts
acolytec3: Will replace with a constant value.

packages/vm/src/runBlock.ts
jochem-brouwer: @holgerd77 Yes exactly here the call to `read_withdrawal_requests` is made. For 6110 I am not sure if all the logic is in here or not (but I am not super deep into 6110, @scorbajio might have better insights here)

packages/util/src/requests.ts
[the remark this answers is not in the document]
acolytec3: Indeed, meant to put that one on your question about `randomBytes`.  This is why I shouldn't respond to code reviews on mobile.

holgerd77 (approved):
Will take this in now to "break the PR dependency cascade", guess we can eventually fine tune last things along both practical integrations.

packages/vm/src/runBlock.ts
scorbajio: For 6110, deposits are retrieved from the block receipt logs for any logs whose address is `DEPOSIT_CONTRACT_ADDRESS` and also from the block body requests field for any requests that are typed as deposit requests.

PR #3373, opened 2024-04-25
common: add spec test for 2935 contract code and update history storage address

g11tech:
test the 2935 contract code as per the eip
 - https://eips.ethereum.org/EIPS/eip-2935
 
and update history save address, however old address is till overriden for kaustine6 (tested and verified) 

its likely that the code and hence address will be modified at which point the address and the code for testing will be updated

### DIFF
--- a/packages/vm/test/api/EIPs/eip-2935-historical-block-hashes.spec.ts
+++ b/packages/vm/test/api/EIPs/eip-2935-historical-block-hashes.spec.ts
@@ -1,12 +1,18 @@
+import { Block } from '@ethereumjs/block'
 import { Blockchain } from '@ethereumjs/blockchain'
 import { Common, Hardfork } from '@ethereumjs/common'
+import { LegacyTransaction } from '@ethereumjs/tx'
 import {
+  Account,
   Address,
   BIGINT_1,
   bigIntToBytes,
   bigIntToHex,
+  bytesToHex,
   equalsBytes,
+  privateToAddress,
   setLengthLeft,
+  unprefixedHexToBytes,
   zeros,
 } from '@ethereumjs/util'
 import { hexToBytes } from 'ethereum-cryptography/utils'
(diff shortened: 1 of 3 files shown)


jochem-brouwer (commented):
LGTM. Will not approve since the code + addr will (indeed) likely get updated.

packages/vm/test/api/EIPs/eip-2935-historical-block-hashes.spec.ts
```diff
@@ -166,5 +180,60 @@ describe('EIP 2935: historical block hashes', () => {
...
+      }).sign(callerPrivateKey)
+
+      await vm.stateManager.putAccount(callerAddress, new Account())
+      const account = await vm.stateManager.getAccount(callerAddress)
+      account!.balance = PREBALANCE
+      await vm.stateManager.putAccount(callerAddress, account!)
+      await vm.stateManager.putContractCode(historyAddress, contract2935Code)
+
+      const result = await vm.runTx({ tx, block, skipHardForkValidation: true })
+      const blockHashi = result.execResult.returnValue
```
holgerd77, line 213: Small nit: above 15 lines can be easily DRYied by setting up a small method taking in i and getting back the blockhash or something similar and then reused in the test below. This will likely also ease the writing of additional test cases.

holgerd77: (can also quickly do if you answer in the next 20 min)

g11tech: ahh seeing this now, will do and update

g11tech:
> LGTM. Will not approve since the code + addr will (indeed) likely get updated.

this will not change until after devnet0, lets take this in and then re-PR when changes come in spec

holgerd77 (approved):
Ok, LGTM

PR #3377, opened 2024-04-27
Remove unnecessary boolean comparisons when using isActivatedEIP

scorbajio:
This change removes some unnecessary explicit boolean comparisons where `isActivatedEIP` is called in order to make the code cleaner and more consistent.


acolytec3 (approved):
LGTM

holgerd77:
Hi there, cool, thanks, that’s useful and a nice cleanup! 🙂👍 Just to add: I would assume we have 2-3 more heavily used methods from Common where such a cleanup might be nice, if you want another task for some in between sideline work. 🙂

Issue #3333, opened 2024-03-22
'Error: Uint8Array expected' when trying to delete from a trie with 'useNodePruning' enabled


bogdan-rosianu:
`package.json`: `"@ethereumjs/trie": "^6.2.0",`

I have isolated the issue within an unit test:
```
it("test trie deletion", async () => {
    const trie = new Trie({ db: new MapDB(), useRootPersistence: true, useNodePruning: true });
    
    await trie.put(utf8ToBytes("key1"), utf8ToBytes("value1"));
    await trie.put(utf8ToBytes("key2"), utf8ToBytes("value2"));
    await trie.del(utf8ToBytes("key1"));
    expect(bytesToHex(trie.root())).toBe("0x9fabdc1cccb1e3a1fa5d991a12087b54360ea47811b36acc1193e2d9ede0cdef");
  });
```

This will fail when trying to delete with this error:
```
Error: Uint8Array expected

    at bytesToHex (.../node_modules/@noble/hashes/src/utils.ts:49:30)
    at .../node_modules/@ethereumjs/trie/src/db/checkpoint.ts:247:36
    at Array.map (<anonymous>)
    at CheckpointDB.batch (.../node_modules/@ethereumjs/trie/src/db/checkpoint.ts:240:36)
    at Trie.saveStack (.../node_modules/@ethereumjs/trie/src/trie.ts:1079:20)
    at Trie._deleteNode (.../node_modules/@ethereumjs/trie/src/trie.ts:1033:18)
```

Switching to `useNodePruning: false` fixes this issue.

Not sure if expected or not.

gabrocheleau:
Thanks for creating the issue and coming up with a test case! 

Will take a look and report back.

gabrocheleau:
Hi @bogdan-rosianu 

The issue has been addressed by #3378, which should be reviewed and merged in shortly. Thanks a lot for uncovering it and coming up with a straightforward reproduction test case. We will be adding your test case to the monorepo as well.

holgerd77:
Hi @bogdan-rosianu, just to let you know: we have a very very late release round this time - since we had (close to) no release pressure - so that issues like this only get in now (or: in 1-2 weeks on release). 😬

Apologies!

PR #3378, opened 2024-04-27
trie: fix del stack operation key formatting

gabrocheleau:
This PR addresses #3333.

This uncovered an issue where we were typecasting EmbeddedNodes as `Uint8Array` (whereas they could also be Uint8Array[]) before trie/database operations. This resulted in issues when the node was < 32 bytes, as in MPTs, nodes are not serialized when they can directly fit in < 32 bytes. This only occurs in the context of non-hashed tries (as hashing normalizes everything to 32 bytes anyways), so I assume this is why this did not get caught.


packages/trie/src/trie.ts
```diff
@@ -1055,24 +1064,24 @@ export class Trie {
...
+      }
+      if (node instanceof BranchNode && lastRoot !== undefined) {
+        const branchKey = key.pop()
+        node.setBranch(branchKey!, lastRoot)
       }
       lastRoot = this._formatNode(node, stack.length === 0, opStack) as Uint8Array
     }
 
-    if (lastRoot) {
+    if (lastRoot !== undefined) {
```
jochem-brouwer, line 1084: Hmm, interesting, why did this not get caught by our linter before?

jochem-brouwer (approved):
LGTM, left one small question regarding the linter.

packages/trie/src/trie.ts
gabrocheleau: I'm not really sure, this linting sometimes behaves in ways that I don't understand, for e.g. by flagging straight booleans (e.g. isActivatedEIP) and not flagging instances like this.

PR #3382, opened 2024-04-27
monorepo: revert and adjust some prefixedHexTypes

gabrocheleau:
This PR reverts some updates that had been made to `string` in #3348  and #3357, where string types were updated to `PrefixedHexString` when that was relevant, but this can lead to breaking changes in certain contexts. I've reverted what seemed like the most obvious common usages (e.g. block or common types that can be used externally). Instead of actually reverting to `string`, I updated the types to `PrefixedHexString | string` so that we can easily eventually remove `string` fully. I've also added TODO's above each of these instances so we can quickly retieve these instances and do that eventually.

### DIFF
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -40,14 +40,15 @@ export interface ChainConfig {
   consensus: ConsensusConfig
 }
 
+// TODO: Remove the string type and only keep PrefixedHexString
 export interface GenesisBlockConfig {
-  timestamp?: PrefixedHexString
-  gasLimit: number | PrefixedHexString
-  difficulty: number | PrefixedHexString
-  nonce: PrefixedHexString
-  extraData: PrefixedHexString
-  baseFeePerGas?: PrefixedHexString
-  excessBlobGas?: PrefixedHexString
+  timestamp?: PrefixedHexString | string
+  gasLimit: number | PrefixedHexString | string
+  difficulty: number | PrefixedHexString | string
+  nonce: PrefixedHexString | string
+  extraData: PrefixedHexString | string
+  baseFeePerGas?: PrefixedHexString | string
+  excessBlobGas?: PrefixedHexString | string
 }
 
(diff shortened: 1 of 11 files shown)


acolytec3 (approved):

acolytec3 (approved):
LGTM

holgerd77:
Cool 🎉, can you also add a note to #3216 ?

packages/common/src/types.ts
```diff
@@ -157,10 +158,11 @@ export interface GethConfigOpts extends BaseOpts {
...
 }
 
+// TODO: Deprecate the string type and only keep BigIntLike
 export interface HardforkByOpts {
-  blockNumber?: BigIntLike
-  timestamp?: BigIntLike
-  td?: BigIntLike
+  blockNumber?: BigIntLike | string
+  timestamp?: BigIntLike | string
+  td?: BigIntLike | string
```
holgerd77, line 165: Oh. This seems wrong I guess?

holgerd77:
I could instantly spot too missed instances, `hexToBytes()` and `intToHex()` which are still missed, so I think we need to do a closer search here what is still missing.

(or are these not breaking for some reasons?)

gabrocheleau:
> I could instantly spot too missed instances, `hexToBytes()` and `intToHex()` which are still missed, so I think we need to do a closer search here what is still missing.
> 
> (or are these not breaking for some reasons?)

Circling back to this. 

I don't think that methods that output `PrefixedHexString` are a problem (since PrefixedHexstring can be assigned to string, it shouldn't break anything). However taking PrefixedHexString as input is definitely breaking. I can revert hexToBytes. We'll have the best of both worlds in a way, as we have already updated all the monorepo. We can perhaps consider then re-switching to PrefixedHexString only in a future release.

gabrocheleau:
Addressed in https://github.com/ethereumjs/ethereumjs-monorepo/pull/3427 @holgerd77

PR #3385, opened 2024-04-30
Implement EIP 7002 (withdrawal requests)

jochem-brouwer:
https://eips.ethereum.org/EIPS/eip-7002

Ref: https://github.com/ethereum/EIPs/blob/3b5fcad6b35782f8aaeba7d4ac26004e8fbd720f/EIPS/eip-7002.md


packages/vm/src/runBlock.ts
```diff
@@ -951,17 +957,143 @@ const DAOConfig = {
...
+      bigIntToBytes(newExcess)
+    )
+
+    // Reset the withdrawals count
+    await vm.stateManager.putContractStorage(
+      withdrawalsAddress,
+      withdrawalsCountSlot,
+      new Uint8Array()
+    )
+  }
```
holgerd77, line 1096: Can we move this whole code block to an internal helper method ˋ_accumulate7002WithdrawalsRequests()ˋ or something?

packages/vm/src/runBlock.ts
```diff
@@ -951,17 +957,143 @@ const DAOConfig = {
...
+    )
+
+    const maxWithdrawalsPerBlock = common.param('vm', 'maxWithdrawalRequestsPerBlock')
+    const targetWithdrawalRequestsPerBlock = common.param('vm', 'targetWithdrawalRequestsPerBlock')
+    const offset = common.param('vm', 'withdrawalsRequestQueueStorageOffset')
+
+    // Dequeue withdrawal requests
+
+    const queueHeadIndex = bytesToBigInt(
+      await vm.stateManager.getContractStorage(withdrawalsAddress, queueHeadSlot)
```
holgerd77, line 1015: Maybe for user convenience to wrap this first withdrawals system contract access in a try/catch together with a meaningful error msg? („….failed, likely withdrawals system contract not deployed“ or similar)

packages/vm/test/api/EIPs/eip-7002.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+
+const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Cancun, eips: [7685, 7002] })
+
+const deploymentTxData = {
+  nonce: BigInt(0),
+  gasLimit: BigInt('0x3d090'),
+  gasPrice: BigInt('0xe8d4a51000'),
+  data: hexToBytes(
+    '0x61049d5f5561013280600f5f395ff33373fffffffffffffffffffffffffffffffffffffffe146090573615156028575f545f5260205ff35b366038141561012e5760115f54600182026001905f5b5f82111560595781019083028483029004916001019190603e565b90939004341061012e57600154600101600155600354806003026004013381556001015f3581556001016020359055600101600355005b6003546002548082038060101160a4575060105b5f5b81811460dd5780604c02838201600302600401805490600101805490600101549160601b83528260140152906034015260010160a6565b910180921460ed579060025560f8565b90505f6002555f6003555b5f548061049d141561010757505f5b60015460028282011161011c5750505f610122565b01600290035b5f555f600155604c025ff35b5f5ffd'
+  ),
```
holgerd77, line 30: Can this code block get some docs? 😜

jochem-brouwer: Yes I need to add docs, the tests are in a super "local" state since it does not what I want it to do so did not yet bother to clean that all up.

For reference: it is the "Deployment" code from the EIP https://eips.ethereum.org/EIPS/eip-7002#deployment

holgerd77: Ah ok, so the requests itself are not yet done, right?

packages/vm/src/runBlock.ts
jochem-brouwer: Yes, good point!

packages/vm/test/api/EIPs/eip-7002.spec.ts
jochem-brouwer: The code itself *should* generate requests, however, there is some unspecified behavior in the EIP which prevents me from actually submitting a withdrawal. I will update by throwing out this "mocked system call" (a la 4788) and instead perform the actual system call :) This will enable the requests.

jochem-brouwer: Done! Tests now cover the requests also! :smile:

packages/vm/src/runBlock.ts
jochem-brouwer: Have addressed!

packages/vm/src/runBlock.ts
jochem-brouwer: Have addressed!

packages/vm/src/runBlock.ts
```diff
@@ -971,3 +986,48 @@ export const accumulateRequests = async (_vm: VM): Promise<CLRequest[]> => {
...
+  })
+
+  const resultsBytes = results.execResult.returnValue
+  if (resultsBytes.length > 0) {
+    const withdrawalRequestType = Number(vm.common.param('vm', 'withdrawalRequestType'))
+    // Each request is 76 bytes
+    for (let startByte = 0; startByte < resultsBytes.length; startByte += 76) {
+      const slicedBytes = resultsBytes.slice(startByte, startByte + 76)
+      const sourceAddress = slicedBytes.slice(0, 20)
+      const validatorPubkey = slicedBytes.slice(20, 20 + 48)
```
acolytec3, line 1026: Are we just breaking this out to show the bytes size?  Just seems excessively verbose.

packages/vm/src/runBlock.ts
```diff
@@ -971,3 +986,48 @@ export const accumulateRequests = async (_vm: VM): Promise<CLRequest[]> => {
   }
   return requests
 }
+
+const _accumulateEIP7002Requests = async (vm: VM, requests: CLRequest[]): Promise<void> => {
+  // TODO PERFORM LOGIC TO CHECK IF CONTRACT EXISTS
```
acolytec3, line 991: Is this todo still valid?  Doesn't your code below check if the contract exists?

packages/vm/test/api/EIPs/eip-7002.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+    await vm.runBlock({
+      block: block3,
+      skipHeaderValidation: true,
+      skipBlockValidation: true,
+      generate: true,
+    })
+
+    // Note: generatedBlock is now overridden with the new generated block (this is thus block number 3)
+    // Ensure there are 2 requests
+    assert.ok(generatedBlock!.requests!.length === 2)
```
acolytec3, line 158: This is a fantastic test.  I didn't even think about being able catch the generated block from teh `afterBlock` event.

acolytec3 (approved):
Going to approve because this looks fantastic.  Nits can be disregarded if not applicable.

packages/vm/src/runBlock.ts
jochem-brouwer: Yes I explicitly did this, this is to explicitly match the EIP spec (validatorPubkey is 48 bytes). However I do agree it is verbose, can change it to just normal numbers and comment it out that its 48 bytes.

packages/vm/src/runBlock.ts
jochem-brouwer: Yup, I forgot to remove the TODO

packages/vm/test/api/EIPs/eip-7002.spec.ts
jochem-brouwer: Yes, actually I want to raise a point about this, isn't the entire point of `generate` to actually retrieve the generated block (i.e. build the block?). I was somewhat surprised that this block was not returned from `runBlock` so had to figure out if I would be able to extract it (if that was even possible). So this is something we could take a look at in the future (not part of this PR)

packages/vm/src/runBlock.ts
jochem-brouwer: I removed it, 6dbb5f2990bc1b96fcfe89aeb78c579b77ad67ec

packages/vm/src/runBlock.ts
jochem-brouwer: I changed it back to normal numbers and instead went for a comment, see 6dbb5f2990bc1b96fcfe89aeb78c579b77ad67ec

jochem-brouwer:
Just pushed a commit for the final cleanups, thanks for the review @acolytec3. If you could re-approve (after CI passes) then from my side this is OK to merge :)

acolytec3 (approved):

holgerd77 (commented):
Totally agree with Andrew, this looks fantastic! 🤩

Some comments on docs, but totally and solely on the "would make this even better" level, since I think think this is already *so* close on the self-readability level that we could even jump over "the last hurdle to perfect" (lol). 😝

One side bonus question:

Would it make sense to also add one test to trigger this behavioral part of the withdrawal request contract? Then we would have all the functionality from the contract "covered"?

grafik

Or is this over the top since this is normally only called on the user side?

(would nevertheless be a somewhat nice bonus feature - especially in these early stages - that we would then *test* the withdrawal request contract itself (to some extend).)

g11tech (commented):
lgtm (post merge review)

PR #3386, opened 2024-05-01
fix(tx): Bug in error message

roninjin10:
Error message had a typo


acolytec3 (approved):
Thanks for the catch!

PR #3390, opened 2024-05-02
Revise EIP 6610 

acolytec3:
Rework of EIP 6110 based on #3303


packages/vm/src/runBlock.ts
```diff
@@ -968,11 +969,16 @@ export class ValidatorWithdrawalRequest extends CLRequest implements CLRequestTy
...
+export const accumulateRequests = async (
+  vm: VM,
+  txResults: RunTxResult[]
+): Promise<CLRequest[]> => {
   const requests: CLRequest[] = []
   const common = vm.common
 
-  // TODO: Add in code to accumulate deposits (EIP-6110)
+  if (common.isActivatedEIP(6110)) {
+    await accumulateDeposits(txResults, requests)
```
g11tech, line 980: accumulate7002 is with `_` while this is not, may be we take a consistent approach

g11tech (approved):
lgtm

PR #3392, opened 2024-05-03
block: fix the block body parsing as well as save/load from blockchain

g11tech:
fixes the block body parsing and removes the body save optimization making the body load simple enough

tested kaustinen6 network and confimed working (so in general all post merge networks) as well as mainnet pow first few thousand blocks

### DIFF
--- a/packages/block/src/block.ts
+++ b/packages/block/src/block.ts
@@ -597,6 +597,12 @@ export class Block {
     if (withdrawalsRaw) {
       bytesArray.push(withdrawalsRaw)
     }
+
+    const requestsRaw = this.requests?.map((req) => req.serialize())
+    if (requestsRaw) {
+      bytesArray.push(requestsRaw)
+    }
+
     if (this.executionWitness !== undefined && this.executionWitness !== null) {
       const executionWitnessBytes = RLP.encode(JSON.stringify(this.executionWitness))
       bytesArray.push(executionWitnessBytes as any)
--- a/packages/block/src/block.ts
+++ b/packages/block/src/block.ts
@@ -30,10 +30,13 @@ import type {
   BlockData,
   BlockOptions,
   ExecutionPayload,
+  ExecutionWitnessBytes,
   HeaderData,
   JsonBlock,
   JsonRpcBlock,
+  RequestsBytes,
   VerkleExecutionWitness,
+  WithdrawalsBytes,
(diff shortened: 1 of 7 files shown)


packages/blockchain/src/db/helpers.ts
```diff
@@ -43,14 +43,7 @@ function DBSetBlockOrHeader(blockBody: Block | BlockHeader): DBOp[] {
 
   const isGenesis = header.number === BIGINT_0
 
-  if (
```
g11tech, line 46: I think the optimization of not saving the body when there aren't non zero fields is not worth it compared to the body load complexity where we would push empty `[]` to resconstruct the body

packages/blockchain/src/db/manager.ts
```diff
@@ -107,40 +104,6 @@ export class DBManager {
     if (hash === undefined || number === undefined) return undefined
     const header = await this.getHeader(hash, number)
     const body = await this.getBody(hash, number)
-    if (body[0].length === 0 && body[1].length === 0) {
```
g11tech, line 110: we don't need this reconstruction now because we now just simply write the body array as is

packages/block/src/block.ts
```diff
@@ -217,10 +220,22 @@ export class Block {
 
     // First try to load header so that we can use its common (in case of setHardfork being activated)
     // to correctly make checks on the hardforks
-    const [headerData, txsData, uhsData, withdrawalBytes, requestBytes, executionWitnessBytes] =
```
g11tech, line 220: we can't do a simple assignment because this sequence is dependent on the eip activation

packages/blockchain/src/blockchain.ts
```diff
@@ -375,8 +375,8 @@ export class Blockchain implements BlockchainInterface {
   async getCanonicalHeadHeader(): Promise<BlockHeader> {
     return this.runWithLock<BlockHeader>(async () => {
       if (!this._headHeaderHash) throw new Error('No head header set')
-      const block = await this.getBlock(this._headHeaderHash)
-      return block.header
+      const header = await this._getHeader(this._headHeaderHash)
```
g11tech, line 378: various fixes in this class to not refer block for header based functionality

holgerd77 (approved):
LGTM

PR #3394, opened 2024-05-03
blockchain: handle nil blockbodies for backward compatibility

g11tech:
handle nil blockbodies for backward compatibility

validated things working fine for kaustinen6 as well as mainnet pow few thousand blocks


holgerd77 (approved):
LGTM

PR #3396, opened 2024-05-03
chore: Small cleanup to vm._emit

roninjin10:
- Move it to an instance method. There shouldn't be any issues with it being an instance method since the underlying event emitter is still per object.
- Make it more typesafe by taking keyof Events rather than string
- TODO meant to do this pr to master not vitest-update

### DIFF
--- a/packages/vm/src/vm.ts
+++ b/packages/vm/src/vm.ts
@@ -59,7 +59,9 @@ export class VM {
    * set to public due to implementation internals
    * @hidden
    */
-  public readonly _emit: (topic: string, data: any) => Promise<void>
+  public _emit(topic: keyof VMEvents, data: any): Promise<void> {
+    return new Promise((resolve) => this.events.emit(topic, data, resolve))
+  }
 
   /**
    * VM is run in DEBUG mode (default: false)
@@ -167,10 +169,6 @@ export class VM {
     this._opts = opts
 
     this._setHardfork = opts.setHardfork ?? false
-
-    this._emit = async (topic: string, data: any): Promise<void> => {
-      return new Promise((resolve) => this.events.emit(topic as keyof VMEvents, data, resolve))
-    }
 
     // Skip DEBUG calls unless 'ethjs' included in environmental DEBUG variables
     // Additional window check is to prevent vite browser bundling (and potentially other) to break


acolytec3:
I switched it over to master and looks like it's accidentally grabbed all of the commits from my PR.  Maybe just cherry-pick to a new PR?

roninjin10:
Done! This pr is in draft only because I haven't ran any tests or typecheckers yet and my type change might break some downstream usages of the function since it's stricter now. 

If pr is green I'll move it out of draft.
If pr is not green I'll fix issues then move it out of draft

packages/vm/src/vm.ts
```diff
@@ -59,7 +59,9 @@ export class VM {
    * set to public due to implementation internals
    * @hidden
    */
-  public readonly _emit: (topic: string, data: any) => Promise<void>
+  public readonly _emit: (topic: keyof VMEvents, data: any): Promise<void> {
```
roninjin10, line 62:
```suggestion
  public readonly _emit: (topic: keyof VMEvents, data: unknown): Promise<void> {
```
THis is slightly more typesafe and best practice to be unknown but 🤷🏽

roninjin10, line 62:
```suggestion
  public readonly _emit(topic: keyof VMEvents, data: any): Promise<void> {
```
syntax error

packages/vm/src/vm.ts
```diff
@@ -59,7 +59,9 @@ export class VM {
    * set to public due to implementation internals
    * @hidden
    */
-  public readonly _emit: (topic: string, data: any) => Promise<void>
+  public readonly _emit(topic: keyof VMEvents, data: any): Promise<void> {
```
acolytec3, line 62: The linter is complaining about this being an invalid use of `readonly` now that you've included the implementation alongside the declaration.  I don't have an opinion about this in general but can we either remove `readonly` or else put the implementation back down below to make the linter happy?

roninjin10: Oh yea readonly cannot be used in this spot in ts. Sorry was making the changes in github editor instead of real editor. Will pull the pr and make it good later today

acolytec3 (approved):
LGTM

PR #3398, opened 2024-05-04
enhance typing of cl requests and do beacon/execution payload handling

g11tech:


packages/block/src/block.ts
```diff
@@ -459,9 +464,25 @@ export class Block {
       requestsRoot: reqRoot,
     }
 
+    const hasDepositRequests = depositRequests !== undefined && depositRequests !== null
+    const hasWithdrawalRequests = withdrawalRequests !== undefined && withdrawalRequests !== null
+    const requests =
+      hasDepositRequests || hasWithdrawalRequests ? ([] as CLRequest<CLRequestType>[]) : undefined
```
jochem-brouwer, line 470: Shouldn't this just check if there is a `requestsRoot`?

packages/block/src/block.ts
```diff
@@ -978,6 +999,29 @@ export class Block {
...
       parentBeaconBlockRoot: header.parentBeaconBlockRoot,
       executionWitness: this.executionWitness,
+
+      // lets add the  request fields first and then iterate over requests to fill them up
+      depositRequests: this.common.isActivatedEIP(6110) ? [] : undefined,
+      withdrawalRequests: this.common.isActivatedEIP(7002) ? [] : undefined,
+    }
+
+    if (this.requests !== undefined) {
+      for (const request of this.requests) {
```
jochem-brouwer, line 1009: These requests should be ordered. I'm not sure if a `switch / case` structure is the "nicest" way to handle this :thinking:

jochem-brouwer (commented):
I got two small comments, but I don't think I'm deep enough in this code (yet) to give an in-depth review, so I think better wait until @acolytec3 takes a look :smile:

packages/block/src/block.ts
g11tech: the sorting is verified as part of block validations so we don't need to worry and keep things simple here

packages/block/src/block.ts
g11tech: well it can check that as well but I thought this check to be more thorough. though i have no strong opinions and can do requestroot check

holgerd77:
This is really a great one! 👍❤️🙂🙏

PR #3406, opened 2024-05-06
Add verkle execution support to executeBlocks

acolytec3:


packages/client/src/execution/vmexecution.ts
```diff
@@ -989,8 +989,36 @@ export class VMExecution extends Execution {
...
+    // Setup VM with verkle state manager if Osaka is active
+    if (
+      this.config.execCommon.hardforkGteHardfork(startExecutionHardfork, Hardfork.Osaka) &&
+      this.config.statelessVerkle
+    ) {
+      const verkleSM = await StatelessVerkleStateManager.create({
+        common: this.vm.common,
+        verkleCrypto: (this.vm.stateManager as StatelessVerkleStateManager).verkleCrypto,
+      })
+      verkleSM.initVerkleExecutionWitness(parentBlock.header.number, parentBlock.executionWitness!)
```
g11tech, line 1013: we don't need to call this with parentBlock as curren't blocks witnesses are inited and used in runBlock

acolytec3: So should I call it with `block.executionWitness` instead of `parentBlock.executionWitness`?  It errors out with ` ERROR Error starting client startExecutionHardfork is not defined` if I don't call that `initVerkleExecutionWitness` function before instantiating the VM.

g11tech (approved):
lgtm

PR #3407, opened 2024-05-08
Verkle checkpoint db tests

scorbajio:
This change adds some unit testing to the `verkle` package and fixes an issue where db stats aren't being updated in the commit function when write operations are being written to the db.


packages/verkle/src/db/checkpoint.ts
```diff
@@ -109,6 +109,7 @@ export class CheckpointDB implements DB {
             opts: { keyEncoding: KeyEncoding.Bytes, valueEncoding: ValueEncoding.Bytes },
           })
         }
+        this._stats.db.writes += 1
```
gabrocheleau, line 112: In the trie package checkpoint, we only add writes for `del` operations and `put` operations. Why do we add it here to `commit`?

gabrocheleau (commented):
Nice! Really like the additional tests, and it's great to see you guys jumping in and getting your hands dirty with the verkle work :) 

Left one small question concerning the db.writes stats, as there seems to be an inconsistency in the way that we are doing this here vs. in the trie package. 

The actual behavior seems to be consistent, but at the trie level, we are incrementing the db.writes count whenever there is a put operations and `hasCheckpoints` is false. See https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/trie/src/db/checkpoint.ts#L185-L200 

I have limited familiarity with that section (i.e. the stats handling) of the code, but shouldn't this be done as low-level as possible (at the level where the actual db writes are happening), to avoid the need to manually increment in multiple places? Curious to hear your thoughts. 

Can be merged once that's discussed/resolved.

scorbajio:
> Nice! Really like the additional tests, and it's great to see you guys jumping in and getting your hands dirty with the verkle work :)
> 
> Left one small question concerning the db.writes stats, as there seems to be an inconsistency in the way that we are doing this here vs. in the trie package.
> 
> The actual behavior seems to be consistent, but at the trie level, we are incrementing the db.writes count whenever there is a put operations and `hasCheckpoints` is false. See https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/trie/src/db/checkpoint.ts#L185-L200
> 
> I have limited familiarity with that section (i.e. the stats handling) of the code, but shouldn't this be done as low-level as possible (at the level where the actual db writes are happening), to avoid the need to manually increment in multiple places? Curious to hear your thoughts.
> 
> Can be merged once that's discussed/resolved.

I did some more digging into the discrepancy and added some more tests for the `CheckpointDB` implementation in the `trie` package as well: It seems like the issue is coming from the `batch` function, since if checkpoints already exist, each batch operation is performed using the `CheckpointDB`'s `put`/`del` functions, which already increment the `write` stat, but if checkpoints don't exist, the wrapped `db`'s own `batch` function implementation is used, which doesn't have access to and does not increment the wrapping class' stats property. I've changed both to now increment that stats in the batch function if checkpoints don't exist.

packages/verkle/src/db/checkpoint.ts
scorbajio: See bellow for full response: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3407#issuecomment-2103267297

gabrocheleau (approved):
thanks for addressing the review! lgtm

PR #3410, opened 2024-05-08
client: fix the getpayloadv4 with a deposit tx and expected deposit requests

g11tech:
issue:

deposit tx is not yielding deposit requests in get payload v4

run the spec which adds a deposit tx and expects deposits in response:

`DEBUG=ethjs,evm:*,vm:* npx vitest run test/rpc/engine/newPayloadV4.spec.ts`


acolytec3 (approved):
LGTM

PR #3411, opened 2024-05-08
Clean up access to deposit address in `common`

acolytec3:
Small cleanup item for the `depositContractAddress` used in 6110 - 

- Adds a test that demonstrates how to set a custom `depositContractAddress` and access it from `common`
- Revises how the `requests` module accesses the `depositContractAddress` parameter on `common`


g11tech (approved):
lgtm!

PR #3412, opened 2024-05-08
feat: Make Address.toString() checksum

roninjin10:
Address.toString() I believe is more useful if it returns a checksum version of teh address.

More tests would need to be updated but I figured I'd put up the pr now to get feedback on if this is something we want.

- Would this be considered a breaking change?
- Should this be a seperate method instead of changing toString()?

### DIFF
--- a/packages/util/src/address.ts
+++ b/packages/util/src/address.ts
@@ -1,3 +1,5 @@
+import { keccak256 } from 'ethereum-cryptography/keccak.js'
+
 import {
   generateAddress,
   generateAddress2,
@@ -129,7 +131,20 @@ export class Address {
    * Returns hex encoding of address.
    */
   toString(): PrefixedHexString {
-    return bytesToHex(this.bytes)
+    const unprefixedString = bytesToHex(this.bytes).substring(2).split('')
+
+    const hash = keccak256(this.bytes)
+
+    for (let i = 0; i < 40; i += 2) {
+      if (hash[i >> 1] >> 4 >= 8 && unprefixedString[i]) {
+        unprefixedString[i] = unprefixedString[i].toUpperCase()
+      }
+      if ((hash[i >> 1] & 0x0f) >= 8 && unprefixedString[i + 1]) {
+        unprefixedString[i + 1] = unprefixedString[i + 1].toUpperCase()
+      }
+    }
+
+    return `0x${unprefixedString.join('')}`
   }
 
   /**
--- a/packages/util/test/address.spec.ts
+++ b/packages/util/test/address.spec.ts
@@ -40,7 +40,7 @@ describe('Address', () => {
     const pubKey = hexToBytes(
       '0x3a443d8381a6798a70c6ff9304bdc8cb0163c23211d11628fae52ef9e0dca11a001cf066d56a8156fc201cd5df8a36ef694eecd258903fca7086c1fae7441e1d'
     )
-    const str = '0x2f015c60e0be116b1f0cd534704db9c92118fb6a'
+    const str = '0x2f015c60E0BE116b1f0cd534704Db9c92118FB6a'
     const addr = Address.fromPublicKey(pubKey)
     assert.equal(addr.toString(), str)
   })
@@ -55,7 +55,7 @@ describe('Address', () => {
   it('should instantiate from private key', () => {
     // prettier-ignore
     const privateKey = Uint8Array.from([234, 84, 189, 197, 45, 22, 63, 136, 201, 58, 176, 97, 87, 130, 207, 113, 138, 46, 251, 158, 81, 167, 152, 154, 171, 27, 8, 6, 126, 156, 28, 95])
-    const str = '0x2f015c60e0be116b1f0cd534704db9c92118fb6a'
+    const str = '0x2f015c60E0BE116b1f0cd534704Db9c92118FB6a'
     const addr = Address.fromPrivateKey(privateKey)
     assert.equal(addr.toString(), str)
   })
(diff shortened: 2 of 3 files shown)


jochem-brouwer:
This is definitely breaking (see the CI, a lot of tests fail because we likely check for the unchecksummed address there).

Could you rename this to `toChecksumString()` or something similar?

roninjin10:
@jochem-brouwer makes sense do you prefer a new method or would you prefer a `formatAsChecksum = false` optional arg added to toString()?

roninjin10:
TIL about this EIP https://github.com/ethereum/ercs/blob/master/ERCS/erc-1191.md

So IMO we should make the seperate `toChecksumString` method and have it optionally take a chainId too

gabrocheleau:
> TIL about this EIP [ethereum/ercs@`master`/ERCS/erc-1191.md](https://github.com/ethereum/ercs/blob/master/ERCS/erc-1191.md?rgh-link-date=2024-05-09T22%3A44%3A53Z)
> 
> So IMO we should make the seperate `toChecksumString` method and have it optionally take a chainId too

Seconding @jochem-brouwer comments. It is definitely too breaking to modify the existing `.toString()` method. 

We already have a toChecksumAddress function that accepts a eip-1191 chainId param. However its usage is discouraged due to lack of backwards compatibility. See https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/src/account.ts#L438 

Since we already have that feature in the utils package, I don't see the necessity of adding it as a new method to the Address class, I will therefore close this issue. Feel free to re-open the PR if you disagree.

roninjin10:
@gabrocheleau this is great! thanks for linking I will use this util and this util is preferred over sticking it on a class method for my own use case.

Semi related, I believe a significant amount of the functionality ethereumjs has should be util methods rather than class methods.  Ethereumjs does not support much code splitting and seperating a lot of the util methods from the core data structures is the biggest thing that would help that issue.

gabrocheleau:
> @gabrocheleau this is great! thanks for linking I will use this util and this util is preferred over sticking it on a class method for my own use case.
> 
> Semi related, I believe a significant amount of the functionality ethereumjs has should be util methods rather than class methods. Ethereumjs does not support much code splitting and seperating a lot of the util methods from the core data structures is the biggest thing that would help that issue.

Thanks for sharing! Do you have examples of util methods that you feel should be refactored into separate utils, rather than baked in core data structures?

roninjin10:
The comment about discouraging usage is interesting. I just tweeted encouraging everybody to use EIP-1191. I think the UX benifits of it are huge and thus disagree with that comment

roninjin10:
> > @gabrocheleau this is great! thanks for linking I will use this util and this util is preferred over sticking it on a class method for my own use case.
> > Semi related, I believe a significant amount of the functionality ethereumjs has should be util methods rather than class methods. Ethereumjs does not support much code splitting and seperating a lot of the util methods from the core data structures is the biggest thing that would help that issue.
> 
> Thanks for sharing! Do you have examples of util methods that you feel should be refactored into separate utils, rather than baked in core data structures?

## Small example of something that should not be on the class instance

One recent example is the tx class which includes a static method to create one from an ethers provider (pretty niche likely mostly unused functionality.

https://github.com/ethereumjs/ethereumjs-monorepo/blob/ded3c6bc1cb589c04246c787cde7cf130f913441/packages/tx/src/transactionFactory.ts#L105 .

## My personal feelings

Personally, I would actually prefer ALL methods are util methods. For example, in the VM package `runTx` `buildBlock`, etc. are all seperate methods that take the base class as a param (in this case as `this`). This is how viem is structured for maximum tree shakability and also how tevm is structured. Would love to see this too but I also understand that this issue is mostly an issue for my specific use case and anybody running ethereumjs outside of the browser won't care about this.

## Bigger painpoint wrt tree shaking

In terms of what is the biggest painpoint it's actually the static methods that have defaults. [example](https://github.com/ethereumjs/ethereumjs-monorepo/blob/ded3c6bc1cb589c04246c787cde7cf130f913441/packages/evm/src/evm.ts#L180)

Because these static methods include defaults like DefaultBlockchain and DefaultStateManager they are impossible to treeshake. If I use a different state manager like RpcStateManager or Verkle state manager the default state manager will still end up bloating the code bundle.

## Example of the patterns that

[tevm](https://tevm.sh) implements it's own `blockchain`, `block`, and `stateManager` package that implements the same interface as ethereumjs so it is a decent example of how to best structure something like ethereumjs for the browser (again I recognize browser is currently a niche use case but just sharing for informational purposes.)

- The BaseBlockchain interface defines the internal data structure: https://github.com/evmts/tevm-monorepo/blob/main/packages/blockchain/src/BaseChain.ts
- The createBaseChain factory internally creates a base chain https://github.com/evmts/tevm-monorepo/blob/main/packages/blockchain/src/createBaseChain.js
- Methods like getBlock  are implemented as decorators in their own files https://github.com/evmts/tevm-monorepo/blob/main/packages/blockchain/src/actions/getBlock.js
- createChain factory puts everything together https://github.com/evmts/tevm-monorepo/blob/main/packages/blockchain/src/createChain.js. Currently it just adds every method but you could imagine in - - future we only add the methods we need for better code splitting. It returns a Chain object that matches the ethjs blockchain interface (with minor differences for methods like shallowCopy returns a Tevm Chain specifically https://github.com/evmts/tevm-monorepo/blob/main/packages/blockchain/src/Chain.ts

This is maximally tree shakable because one could choose to only use the BaseBlockchain and only pull in the instance methods they need for their specific use. I don't currently take advantage of this but will in future

roninjin10:
@gabrocheleau let me know if it would be helpful to open an issue for tree shaking and bundle size concerns. If there was a desire to make ethereumjs better for the browser wrt bundle size it's something I'd be happy to discuss what the best compromises are and help make happen.

We wouldn't necessarily need to go as far as deleting classes as I am doing in tevm to make a decent impact we could keep ethereumjs mostly as is and find the lightest touch changes that make the biggest impact

roninjin10:
Ricmoo from ethers.js explained why EIP-1193 is not supported by web3.js and ethers and it made sense. Now I get why it's discouraged

PR #3413, opened 2024-05-09
vm: fix the gathering of stateroot only after requests have been accumulated

g11tech:
stateRoot was collected before requests were accumulated which actually changes the state. so the build blocks would have the wrong stateroot

this pr fixes the same

tested on electra with lodestar integration


acolytec3 (approved):
LGTM

jochem-brouwer (approved):

jochem-brouwer:
Have approved, not sure why this is not caught by the latest fixtures though.

PR #3415, opened 2024-05-10
Add support for `pending` in `getTransactionCount`

acolytec3:
This updates the response handler for `eth_getTransactionCount` to support the `pending` block option.  This includes pending transactions currently in the local transaction pool in the transaction count.

This is needed to allow our client to work with the [tx fuzzer](https://github.com/MariusVanDerWijden/tx-fuzz/)

### DIFF
--- a/packages/client/src/rpc/modules/eth.ts
+++ b/packages/client/src/rpc/modules/eth.ts
@@ -770,7 +770,9 @@ export class Eth {
    */
   async getTransactionCount(params: [string, string]) {
     const [addressHex, blockOpt] = params
-    const block = await getBlockByOption(blockOpt, this._chain)
+    let block
+    if (blockOpt !== 'pending') block = await getBlockByOption(blockOpt, this._chain)
+    else block = await getBlockByOption('latest', this._chain)
 
     if (this._vm === undefined) {
       throw new Error('missing vm')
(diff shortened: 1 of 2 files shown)


packages/client/src/rpc/modules/eth.ts
```diff
@@ -784,7 +786,16 @@ export class Eth {
     if (account === undefined) {
       return '0x0'
```
jochem-brouwer, line 787: Was about to say that this should include pending txpool msgs from this account, but this is impossible since accounts do need to exist because otherwise they cannot pay the min basefee of 7 wei.

packages/client/test/rpc/eth/getTransactionCount.spec.ts
```diff
@@ -74,15 +74,34 @@ describe(method, () => {
...
-  it('call with unsupported block argument', async () => {
+  it('call with pending block argument', async () => {
     const blockchain = await Blockchain.create()
 
     const client = await createClient({ blockchain, includeVM: true })
     const manager = createManager(client)
+    const service = client.services.find((s) => s.name === 'eth') as FullEthereumService
     const rpc = getRpcClient(startRPC(manager.getMethods()))
 
-    const res = await rpc.request(method, ['0xccfd725760a68823ff1e062f4cc97e1360e8d997', 'pending'])
```
jochem-brouwer, line 84: Not sure, maybe we want to keep this test in it but now instead of using `pending` we use some invalid tag like `unsupported` or something.

jochem-brouwer: Meh I guess this is tested in the `getBlock` anyways

jochem-brouwer (approved):
LGTM

PR #3416, opened 2024-05-10
fix estimateGas

acolytec3:


packages/client/src/rpc/modules/eth.ts
```diff
@@ -28,7 +29,7 @@ import type { EthereumClient } from '../../index.js'
 import type { EthProtocol } from '../../net/protocol/index.js'
 import type { FullEthereumService, Service } from '../../service/index.js'
 import type { RpcTx } from '../types.js'
-import type { Block, JsonRpcBlock } from '@ethereumjs/block'
+import type { JsonRpcBlock } from '@ethereumjs/block'
```
jochem-brouwer, line 32: How come this is added but nothing is done with it in this PR? I don't see anything in the diff :thinking:

acolytec3: It should have moved block from a type import to a regular import

jochem-brouwer:
I have further cleaned up some default values (the estimateGas tests were failing) such that the right base fee is used on the block.

packages/client/src/rpc/modules/eth.ts
jochem-brouwer: Oh right yes this makes sense, thanks

jochem-brouwer: (its actually in the diff :sweat_smile: - time to call it a day I guess :smile: )

jochem-brouwer (approved):
I will approve here, but before merging could you check my further changes, and also check that we can now still run kurtosis with tx-fuzz? :smile:

g11tech (commented):
post merge review lgtm

PR #3418, opened 2024-05-13
evm,vm: remove the hacks to prevent account cleanups of system contracts

g11tech:
remove some cleanup prevention hacks and add 158 cleanups to beacon block root and block root contract which would cleanup state changes if contracts are not yet deployed

(a genesis state with 4788 activated and no contract deployed for e.g. was causing consistences with block root mismatch issues)


jochem-brouwer (commented):
LGTM. However, a chain where 4788 is active but the contract is not deployed: this is undefined behavior and is not covered by the EIP. This should never be allowed. I assume this was a custom devnet?

yes custom devnet

g11tech:
> LGTM. However, a chain where 4788 is active but the contract is not deployed: this is undefined behavior and is not covered by the EIP. This should never be allowed. I assume this was a custom devnet?

yes correct

acolytec3 (approved):
LGTM.  Have merged latest master so lets merge when CI passes.

jochem-brouwer (approved):

PR #3419, opened 2024-05-13
statemanager: verify prestate accesses

gabrocheleau:
This PR adds a verkle feature to verify that all the provided witness keys have been accessed during block execution. As per the spec, blocks that include extra witnesses are considered invalid and should be rejected.


gabrocheleau:
After discussing this, it turns out that the postState exhaustiveness verification implicitly also acts as a satisfactory verification for the preState. Initially did not realize this was the case, but this means that we can now close this. 

As for the other fixes and increased test coverage that were part of this PR, I will port them over to #3421

PR #3420, opened 2024-05-13
Move tree key computation to verkle and simplify

acolytec3:
This migrates the `getTreeKey*` methods (and related constants) from `statemanager` to `verkle` where they are more properly housed.

- Replaces `getTreeKeyFor*` methods with a single `getTreeKey` that accepts a `leafType` as a second input and returns the correct tree key corresponding to the value type


packages/verkle/src/types.ts
```diff
@@ -123,3 +123,22 @@ export type FoundNodeFunction = (
 export const ROOT_DB_KEY = utf8ToBytes('__root__')
 
 export type VerkleCrypto = VerkleFFI
+
+export enum leafType {
```
holgerd77, line 127: Short nit: We have our `enum`s capitalized by convention, so `leafType` -> `LeafType`.

gabrocheleau: I've searched through the repo and the keys of the existing enums are also capitalized, will update both.

gabrocheleau (approved):

gabrocheleau (approved):

PR #3421, opened 2024-05-14
vm: missing beaconroot account verkle fix 

gabrocheleau:
This PR addresses an issue where the account at the beaconRootAddress would be missing from the verkle witness, leading to an error since the statelessStateManager throws whenever there is an account missing in the witness. With this PR, we are fixing this in the same way that we are addressing the same issue that appeared with EIP-2935.

We are discussing with the verkle folks whether or not this should be included in the witness. We (Gajinder and I) are thinking that it should, as it would be nice for a stateless client not to depend on "local state" or "special cases handling" to execute a block. Eventually, the verkle behavior could change so that this becomes unnecessary. 

This also updates the verkle test data to include a bigger, more complex verkle block, and fixes the eip-6800 (verkle) test by actually running the block rather than the individual transactions.


g11tech (approved):
lgtm

holgerd77:
Ok, so this one is "on hold" then until things are settled? Will give this a "Blocked" label for now due to this assumption, feel free to remove for any reason.

gabrocheleau:
> Ok, so this one is "on hold" then until things are settled? Will give this a "Blocked" label for now due to this assumption, feel free to remove for any reason.

Just need to figure out why the postState mismatches, otherwise this should be OK.

g11tech (approved):
lgtm

PR #3423, opened 2024-05-14
verkle: implement verkle proof verification 

gabrocheleau:


packages/statemanager/src/statelessVerkleStateManager.ts
```diff
@@ -903,21 +896,26 @@ export class StatelessVerkleStateManager implements EVMStateManagerInterface {
   async flush(): Promise<void> {}
 
   /**
-   * Gets the verkle root.
-   * NOTE: this needs some examination in the code where this is needed
-   * and if we have the verkle root present
-   * @returns {Promise<Uint8Array>} - Returns the verkle root of the `StateManager`
+   * Gets the cache state root.
+   * This is used to persist the stateRoot between blocks, so that blocks can retrieve the stateRoot of the parent block.
```
acolytec3, line 900: This will only work for block execution when you're doing it sequentially, right?

acolytec3 (approved):
This looks great so far.  Are we going to do a separate PR to add proof verification in the client?

packages/statemanager/src/statelessVerkleStateManager.ts
gabrocheleau: Indeed. We could come up with a more elaborate scheme, however Guillaume confirmed to me that the verkle block format would be updated to also contain the preStateRoot. We therefore will not need this in the next testnet, so I think that this naive approach is sufficient for testing purposes.

packages/client/src/config.ts
```diff
@@ -542,8 +544,7 @@ export class Config {
     this.statelessVerkle = options.statelessVerkle ?? true
     this.startExecution = options.startExecution ?? false
     this.ignoreStatelessInvalidExecs = options.ignoreStatelessInvalidExecs ?? false
-
-    this.metrics = options.prometheusMetrics
+    this.initialVerkleStateRoot = options.initialVerkleStateRoot ?? new Uint8Array()
```
gabrocheleau, line 547:
```suggestion

    this.metrics = options.prometheusMetrics
    this.initialVerkleStateRoot = options.initialVerkleStateRoot ?? new Uint8Array()
```

packages/vm/src/runBlock.ts
```diff
@@ -125,26 +125,39 @@ export async function runBlock(this: VM, opts: RunBlockOpts): Promise<RunBlockRe
...
-    ;(this._opts.stateManager as StatelessVerkleStateManager).initVerkleExecutionWitness(
-      block.header.number,
-      block.executionWitness
-    )
+    // Update the stateRoot cache
+    await stateManager.setStateRoot(block.header.stateRoot)
+
+    // Populate the execution witness
+    stateManager.initVerkleExecutionWitness(block.header.number, block.executionWitness)
+
```
acolytec3, line 147: Can we move this below the error condition checks or do they need to happen before we check for parent stateRoot?

acolytec3 (approved):
Going to go ahead and approve.  One nit but I don't know if it's even a valid concern.

packages/vm/src/runBlock.ts
gabrocheleau: Good point, I moved this up

acolytec3 (approved):

g11tech (commented):
lgtm, post merge review

PR #3427, opened 2024-05-14
util: revert some PrefixedHexString breaking changes

gabrocheleau:
This PR re-adds support for type `string` as an input type for some util helpers, to preserve backwards compatibility. 

I also realized that isHexPrefixed was redundant (we already had an internal helper that implements that behavior) so I removed it.

### DIFF
--- a/packages/util/README.md
+++ b/packages/util/README.md
@@ -117,7 +117,7 @@ Genesis related interfaces and helpers.
 
 ### Module: [internal](src/internal.ts)
 
-Internalized simple helper methods like `isHexPrefixed`. Note that methods from this module might get deprectared in the future.
+Internalized simple helper methods like `isHexString`. Note that methods from this module might get deprectared in the future.
 
 ### Module: [kzg](src/kzg.ts)
 
@@ -257,7 +257,7 @@ The following methods are available by an internalized version of the [ethjs-uti
 - arrayContainsArray
 - getBinarySize
 - stripHexPrefix
-- isHexPrefixed
+- isHexString
 - isHexString
 - padToEven
 - fromAscii
--- a/packages/util/README.md
+++ b/packages/util/README.md
@@ -117,7 +117,7 @@ Genesis related interfaces and helpers.
 
 ### Module: [internal](src/internal.ts)
 
-Internalized simple helper methods like `isHexString`. Note that methods from this module might get deprectared in the future.
+Internalized simple helper methods like `isHexString`. Note that methods from this module might get deprecated in the future.
 
 ### Module: [kzg](src/kzg.ts)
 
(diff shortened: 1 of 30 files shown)


packages/util/README.md
```diff
@@ -117,7 +117,7 @@ Genesis related interfaces and helpers.
 
 ### Module: [internal](src/internal.ts)
 
-Internalized simple helper methods like `isHexPrefixed`. Note that methods from this module might get deprectared in the future.
+Internalized simple helper methods like `isHexString`. Note that methods from this module might get deprectared in the future.
```
scorbajio, line 120:
```suggestion
Internalized simple helper methods like `isHexString`. Note that methods from this module might get deprecated in the future.
```

packages/util/src/internal.ts
```diff
@@ -27,17 +27,17 @@ import { bytesToUnprefixedHex, utf8ToBytes } from './bytes.js'
 import type { PrefixedHexString } from './types.js'
 
 /**
- * Returns a `Boolean` on whether or not the a `String` starts with '0x'
- * @param str the string input value
- * @return a boolean if it is or is not hex prefixed
- * @throws if the str input is not a string
+ * Returns a boolean on whether or not the the input starts with '0x' and if matches the optional length
```
scorbajio, line 30:
```suggestion
 * Returns a boolean on whether or not the the input starts with '0x' and matches the optional length
```

packages/util/src/internal.ts
```diff
@@ -27,17 +27,17 @@ import { bytesToUnprefixedHex, utf8ToBytes } from './bytes.js'
...
+ * @param {string} value the string input value
+ * @param {number|undefined} length the optional length of the hex string in bytes
+ * @returns {boolean} Whether or not the string is a valid PrefixedHexString matching the optional length
  */
-export function isHexPrefixed(str: string): str is PrefixedHexString {
-  if (typeof str !== 'string') {
-    throw new Error(`[isHexPrefixed] input must be type 'string', received type ${typeof str}`)
-  }
+export function isHexString(value: string, length?: number): value is PrefixedHexString {
+  if (typeof value !== 'string' || !value.match(/^0x[0-9A-Fa-f]*$/)) return false
```
scorbajio, line 36: It seems like this regex would consider `0x` a valid hex string. The regex could be changed to expect at least one hexadecimal digit to follow the prefix: `/^0x[0-9A-Fa-f]+$/`.

scorbajio (changes requested):
This looks great! Left a few suggestions and comments.

packages/util/src/internal.ts
gabrocheleau: I get your point, but I don't think we can make that call, as I've seen `0x` used in many places, and I think expecting a length of at least one hex would be a breaking change.

scorbajio (approved):
ACK

PR #3428, opened 2024-05-14
util: remove isHexString length arg

gabrocheleau:
This PR removes the isHexString length arg.

My reasons for removing: 

- It's usage is confusing because it expects bytes length, and not hex length. This is not obvious at all from the api, only realized through looking at the code
- It is not used anywhere in our repo
- It's in the internal folder, so I assume intended for internal usage.


gabrocheleau:
Since this is breaking and might be used externally, we have decided to leave this out. Closing.

PR #3430, opened 2024-05-15
Verkle Implementation: Build out Trie Processing

acolytec3:
Foundations for a full verkle trie implementation:  
1) Completely reworks the structure of the `VerkleTrie` class to remove unnecessary code that was borrowed from `trie` as well as `go-verkle`
2) Simplifies the public surface of the `InternalNode` and `LeafNode` classes to make them easier to work with (e.g. commitments are updated internally when you update a value/child in the node)
3) Updates the structure of the `children` array within `InternalNode` so that each element of `children` stores both the commitment for that child node as well as the path to that node (which is either a partial stem - if the child node is an Internal Node, or a full stem if the child node is a Leaf Node)
4) Update `findPath` to use the new structure of the internal nodes described above so that walking the trie is merely a matter of retrieving the child node at each level of the `path` key provided as an input to `findPath`
5) Removes `depth` as a property on the `VerkleNode` class (since no obvious use case)


electra.yaml
```diff
@@ -0,0 +1,21 @@
+participants:
+  - el_type: ethereumjs
+    el_image: ethereumjs:local
+    cl_type: grandine
+    cl_image: ethpandaops/grandine:feature-electra
```
gabrocheleau, line 5: Doesn't seem like that should belong to this PR?

packages/verkle/src/node/internalNode.ts
```diff
@@ -63,19 +64,30 @@ export class InternalNode extends BaseVerkleNode<VerkleNodeType.Internal> {
...
+    resolver: () => void,
+    verkleCrypto: VerkleCrypto
+  ): void {
     // Index of the child pointed by the next byte in the key
     const childIndex = stem[this.depth]
 
     const child = this.children[childIndex]
 
     if (child instanceof LeafNode) {
+      // TODO: Understand the intent of what cowChild is suppoded to do
```
gabrocheleau, line 90: cow stands for "copy-on-write". As far as I recall cowChild is used as a record to mark the children that have been modified, in order to stack the commitment updates at the end since that is way more efficient.

packages/verkle/src/node/internalNode.ts
```diff
@@ -93,19 +105,26 @@ export class InternalNode extends BaseVerkleNode<VerkleNodeType.Internal> {
...
-      const leafNode = LeafNode.create(stem, values)
+      // TODO: Determine if this is how to create the correct commitment
+      let commitment = verkleCrypto.zeroCommitment
+      for (const [idx, value] of values.entries()) {
+        if (bytesToBigInt(value) > BIGINT_0)
+          commitment = verkleCrypto.updateCommitment(commitment, idx, new Uint8Array(32), value)
+      }
+      const leafNode = LeafNode.create(stem, values, this.depth + 1, commitment)
 
+      // TODO - Why is the leaf node set at depth + 2 instead of + 1)?
```
gabrocheleau, line 121: I think that this is because the leaf is inserted in the position of the suffix level commitment. So we have:

depth `n` : internal node:
depth `n+1`: the extension level commitment
depth `n+2`: the suffix level commitment 

See this picture for how that corresponds to the structure:
image

gabrocheleau (commented):
Had a little bit of time and quickly glanced through, nice work! left some comments on some of your questions/todos

electra.yaml
acolytec3: It doesn't.  Accidentally included and will remove.

packages/verkle/src/node/internalNode.ts
```diff
@@ -29,11 +29,17 @@ export class InternalNode extends BaseVerkleNode<VerkleNodeType.Internal> {
     // Not implemented yet
   }
 
-  setChild(index: number, child: VerkleNode) {
+  // Updates the commitment value for a child node at the corresponding index
+  setChild(index: number, child: Uint8Array) {
     this.children[index] = child
+    // TODO: Update commitment as well
```
gabrocheleau, line 35: Isn't `child` the commitment that is assigned to `children[index]` ? At least that's what I understand from line 13-14:

```
  // Array of commitments to child nodes
  public children: Uint8Array[]
```

packages/verkle/src/node/internalNode.ts
```diff
@@ -47,35 +53,47 @@ export class InternalNode extends BaseVerkleNode<VerkleNodeType.Internal> {
...
+  getChildren(index: number): Uint8Array | null {
     return this.children?.[index] ?? null
   }
 
-  insert(key: Uint8Array, value: Uint8Array, resolver: () => void): void {
+  insert(
+    key: Uint8Array,
+    value: Uint8Array,
+    resolver: () => void,
+    verkleCrypto?: VerkleCrypto
```
gabrocheleau, line 77: Aren't we able to access verkleCrypto from this.verkleCrypto?

packages/verkle/src/node/internalNode.ts
```diff
@@ -47,35 +53,47 @@ export class InternalNode extends BaseVerkleNode<VerkleNodeType.Internal> {
...
+  ): void {
     const values = new Array<Uint8Array>(NODE_WIDTH)
     values[key[31]] = value
-    this.insertStem(key.slice(0, 31), values, resolver)
+    this.insertStem(key.slice(0, 31), values, resolver, verkleCrypto!)
```
gabrocheleau, line 81: I think using `verkleCrypto` from `this.verkleCrypto` should also remove the need for the `!`

packages/verkle/src/node/internalNode.ts
```diff
@@ -47,35 +53,47 @@ export class InternalNode extends BaseVerkleNode<VerkleNodeType.Internal> {
...
+    resolver: () => void,
+    verkleCrypto: VerkleCrypto
+  ): void {
     // Index of the child pointed by the next byte in the key
     const childIndex = stem[this.depth]
 
     const child = this.children[childIndex]
 
     if (child instanceof LeafNode) {
+      // TODO: Understand the intent of what cowChild is suppoded to do
```
gabrocheleau, line 96: It seems like we have no way of "stacking commitments" (perhap because of what we have access to through the verkle cryptography wasm API), so I guess we can do without `cowChild` for the time being? We can always re-add this at some point if we have the need for it.

packages/verkle/src/node/internalNode.ts
```diff
@@ -85,35 +103,64 @@ export class InternalNode extends BaseVerkleNode<VerkleNodeType.Internal> {
...
       newBranch.cowChild(nextByteInExistingKey)
-      this.children[childIndex] = newBranch
+      this.children[childIndex] = newBranch.commitment
       newBranch.children[nextByteInExistingKey] = child
       child.depth += 1
 
       const nextByteInInsertedKey = stem[this.depth + 1]
       if (nextByteInInsertedKey === nextByteInExistingKey) {
-        return newBranch.insertStem(stem, values, resolver)
+        return newBranch.insertStem(stem, values, resolver, verkleCrypto)
```
gabrocheleau, line 114:
```suggestion
        return newBranch.insertStem(stem, values, resolver)
```

packages/verkle/src/node/internalNode.ts
```diff
@@ -85,35 +103,64 @@ export class InternalNode extends BaseVerkleNode<VerkleNodeType.Internal> {
...
       }
 
       // Next word differs, so this was the last level.
       // Insert it directly into its final slot.
-      const leafNode = LeafNode.create(stem, values)
+      // TODO: Fix this following `trie.put` logic
+      let leafCommitment = verkleCrypto.zeroCommitment
+      let c1 = verkleCrypto.zeroCommitment
+      let c2 = verkleCrypto.zeroCommitment
+      for (const [idx, value] of values.entries()) {
```
gabrocheleau, line 123: What does `idx` stand for? Could be helpful to add a line or two of comments that describe what is going on in that loop. I have a rough idea but not super clear on all the parts.

packages/verkle/src/node/internalNode.ts
```diff
@@ -85,35 +103,64 @@ export class InternalNode extends BaseVerkleNode<VerkleNodeType.Internal> {
...
+      )
+
+      // TODO - Why is the leaf node set at depth + 2 instead of + 1)?
       leafNode.setDepth(this.depth + 2)
       newBranch.cowChild(nextByteInInsertedKey)
       newBranch.children[nextByteInInsertedKey] = leafNode
     } else if (child instanceof InternalNode) {
       this.cowChild(childIndex)
-      return child.insertStem(stem, values, resolver)
+      return child.insertStem(stem, values, resolver, verkleCrypto)
```
gabrocheleau, line 156:
```suggestion
      return child.insertStem(stem, values, resolver)
```

packages/verkle/src/node/internalNode.ts
```diff
@@ -85,35 +103,64 @@ export class InternalNode extends BaseVerkleNode<VerkleNodeType.Internal> {
...
     } else {
       throw new Error('Invalid node type')
     }
   }
 
   // TODO: go-verkle also adds the bitlist to the raw format.
```
gabrocheleau, line 162:
```suggestion
```

We can do without the bitlist as it was only used in the geth serialization implementation for their own purposes, and will be supserseded by an updated, more efficient serialization technique, see: 
- https://github.com/ethereum/go-verkle/pull/433 
- https://github.com/ethereum/go-verkle/pull/432

packages/verkle/src/node/leafNode.ts
```diff
@@ -39,11 +54,11 @@ export class LeafNode extends BaseVerkleNode<VerkleNodeType.Leaf> {
     const stem = rawNode[1]
     // TODO: Convert the rawNode commitments to points
```
gabrocheleau, line 55:
```suggestion
```

packages/verkle/src/node/types.ts
```diff
@@ -16,14 +16,15 @@ export type VerkleNode = TypedVerkleNode[VerkleNodeType]
 
 export interface VerkleNodeInterface {
   commit(): Uint8Array
-  hash(): any
+  hash(verkleCrypto: VerkleCrypto): any
```
gabrocheleau, line 19: This should not be necessary now, right?

packages/verkle/src/types.ts
```diff
@@ -8,6 +8,7 @@ import type { VerkleCrypto as VerkleFFI } from 'verkle-cryptography-wasm'
 // Field representation of a commitment
 export interface Fr {}
 
+// TODO: Decide if we still need this interface.  Commitments are now all bytes from the JS perspective
```
gabrocheleau, line 11: I think we can do without this. We are not interacting with elliptic curve point representations directly, and always interfacing through verklecrypto

packages/verkle/src/types.ts
```diff
@@ -142,3 +143,6 @@ export const HEADER_STORAGE_OFFSET = 64
 export const CODE_OFFSET = 128
 export const VERKLE_NODE_WIDTH = 256
 export const MAIN_STORAGE_OFFSET = BigInt(256) ** BigInt(31)
+
+export const zeroValues = new Array(256).fill(new Uint8Array())
+export const zeroCValues = new Array(128).fill(new Uint8Array())
```
gabrocheleau, line 148: To me "CValue" is a bit unclear. Do you think it would be clearer if we named this "zeroLeafValues"?

packages/verkle/src/node/internalNode.ts
```diff
@@ -1,23 +1,23 @@
...
+  // Array of commitments to child nodes
+  public children: Uint8Array[]
   public copyOnWrite: Record<string, Uint8Array>
   public type = VerkleNodeType.Internal
 
-  /* TODO: options.children is not actually used here */
   constructor(options: VerkleNodeOptions[VerkleNodeType.Internal]) {
     super(options)
-    this.children = options.children ?? new Array(NODE_WIDTH).fill(null)
+    this.children = options.children ?? new Array(NODE_WIDTH).fill(new Uint8Array())
```
gabrocheleau, line 20: We could use the `zeroValues` or `zeroCValues` constant that has been declared in one of the other files, right?

packages/verkle/src/node/internalNode.ts
acolytec3: Yes indeed.  Thanks for catching that.  Will adjust.

packages/verkle/src/node/util.ts
```diff
@@ -28,3 +29,30 @@ export function decodeNode(raw: Uint8Array) {
 export function isRawNode(node: Uint8Array | Uint8Array[]): node is Uint8Array[] {
   return Array.isArray(node) && !(node instanceof Uint8Array)
 }
+
+/***
+ * Converts 128 32byte values of a leaf node into 16 byte values for generating a commitment for half of a
+ * leaf node's values
+ * @param values - an array of Uint8Arrays representing the first or second 128 values stored by verkle trie leaf node
```
gabrocheleau, line 36:
```suggestion
 * @param values - an array of Uint8Arrays representing the first or second set of 128 values stored by verkle trie leaf node
```

packages/verkle/src/node/internalNode.ts
acolytec3: I would tend to agree.  We probably will be able to improve our performance down the line but would prefer to do everything naively for starters.

packages/verkle/src/util/keys.ts
```diff
@@ -94,3 +94,7 @@ export const getTreeKeyForStorageSlot = async (
 
   return concatBytes(getStem(verkleCrypto, address, treeIndex), toBytes(subIndex))
 }
+
+export const verifyKeyLength = (key: Uint8Array) => {
+  if (key.length !== 32) throw new Error(`expected key with length 32l; got ${key.length}`)
```
gabrocheleau, line 99:
```suggestion
  if (key.length !== 32) throw new Error(`expected key with length 32; got ${key.length}`)
```

packages/verkle/src/node/util.ts
```diff
@@ -28,3 +29,30 @@ export function decodeNode(raw: Uint8Array) {
...
+ * @param deletedValues - an array of booleans where a value of true at a given position indicates a value
+ * that is being deleted - should always be false if generating C2 values
+ * Returns an array of 256 16byte UintArrays with the leaf marker set for each value that is deleted
+ */
+export const createCValues = (values: Uint8Array[], deletedValues = new Array(128).fill(false)) => {
+  if (values.length !== 128 || deletedValues.length !== 128)
+    throw new Error(`got wrong number of values, expected 128, got ${values.length}`)
+  const expandedValues: Uint8Array[] = new Array(256)
+  for (let x = 0; x < 128; x++) {
+    // We add 16 trailing zeros to each value since all commitments are to an array of 32 byte values
```
gabrocheleau, line 46:
```suggestion
    // We add 16 trailing zeros to each value since all commitments are padded to an array of 32 byte values
```

packages/verkle/src/types.ts
acolytec3: Not really sure.  It's not precisely `zeroLeafValues` but more like the `zeroHalfLeafValues`.  What do you think?

packages/verkle/src/util/walkController.ts
```diff
@@ -80,7 +80,7 @@ export class WalkController {
     for (const child of children) {
       if (child.nodeRef !== null) {
         const childKey = new Uint8Array([...key, child.keyExtension])
-        this.pushNodeToQueue(child.nodeRef.hash(), childKey)
+        this.pushNodeToQueue(child.nodeRef.hash(undefined as any), childKey)
```
gabrocheleau, line 83: https://github.com/ethereum/go-verkle/pull/433 et https://github.com/ethereum/go-verkle/pull/432

packages/verkle/src/verkleTree.ts
```diff
@@ -149,18 +159,28 @@ export class VerkleTree {
...
-    }
+  async get(key: Uint8Array): Promise<Uint8Array | null> {
+    verifyKeyLength(key)
+    const nodeRLP = await this._db.get(key)
+    if (nodeRLP === undefined) return null
+    // We first retrieve the node and decode it
+    const rawNode = RLP.decode(nodeRLP) as Uint8Array[]
+    const node =
+      rawNode[0][0] === VerkleNodeType.Leaf
+        ? // TODO: Figure out how to determine depth from key
```
gabrocheleau, line 172: I don't think Depth can be directly inferred from the key, as the same key would have different depths depending on which verkle trie structure they are a part of. Shouldn't that information be contained in the serialized node that is stored in the db? 

I just checked the go code and they don't store that in their (legacy) serialization format..  maybe I'm missing something.

packages/verkle/src/verkleTree.ts
```diff
@@ -149,18 +159,28 @@ export class VerkleTree {
...
+          LeafNode.fromRawNode(rawNode, 1, this.verkleCrypto)
+        : InternalNode.fromRawNode(rawNode, 1, this.verkleCrypto)
+
+    if (node instanceof LeafNode)
+      if (node instanceof LeafNode) {
+        const keyLastByte = key[key.length - 1]
+
+        // The retrieved leaf node contains an array of 256 possible values.
+        // The index of the value we want is at the key's last byte
+        return node.values?.[keyLastByte] ?? null
```
gabrocheleau, line 182: extra if?

packages/verkle/src/verkleTree.ts
```diff
@@ -173,14 +193,82 @@ export class VerkleTree {
...
+      const commitmentIndex = suffix < 128 ? suffix * 2 : (suffix - 128) * 2
+      let cCommitment = this.verkleCrypto.zeroCommitment
+      // Update the commitment for the first 16 bytes of the value
+      cCommitment = this.verkleCrypto.updateCommitment(
+        cCommitment,
+        commitmentIndex,
+        new Uint8Array(32),
+        cValues[commitmentIndex]
+      )
+      // Update thecommitment for the second 16 bytes of the value
```
gabrocheleau, line 222:
```suggestion
      // Update the commitment for the second 16 bytes of the value
```

packages/verkle/src/verkleTree.ts
```diff
@@ -173,14 +193,82 @@ export class VerkleTree {
...
+        key.slice(0, 31),
+        values,
+        leafNode.length,
+        commitment,
+        c1,
+        c2,
+        this.verkleCrypto
+      )
+      await this._db.put(leafNode.hash(), leafNode.serialize())
     }
```
gabrocheleau, line 273: Maybe I'm not following this, but where is the part of the code where if the leafNode exists, we update it? I'm just seeing us create a new leaf if it does not exist in the trie.

gabrocheleau (commented):
Amazing work! Nice to finally dive in and properly go through the code. I've added a bunch of comments and suggestions .

packages/verkle/src/node/internalNode.ts
gabrocheleau: agreed!

packages/verkle/src/verkleTree.ts
acolytec3: Yup, this is still very incomplete.  This is where I'm currently focused.  We need to actually pull parent nodes out of the DB and update the child commitment value in their values array and then update their commitment.

packages/verkle/src/verkleTree.ts
acolytec3: Okay, I think that makes sense to me.  We can ensure that's included in the serialized node.

packages/verkle/src/node/internalNode.ts
```diff
@@ -1,39 +1,49 @@
...
+  // Array of uncompressed commitments (i.e. 64 byte Uint8Arrays) to child nodes
+  public children: Uint8Array[]
   public type = VerkleNodeType.Internal
 
-  /* TODO: options.children is not actually used here */
   constructor(options: VerkleNodeOptions[VerkleNodeType.Internal]) {
     super(options)
-    this.children = options.children ?? new Array(NODE_WIDTH).fill(null)
-    this.copyOnWrite = options.copyOnWrite ?? {}
+    this.children = options.children ?? new Array(64).fill(new Uint8Array(64))
```
gabrocheleau, line 19: Just reading up the conversation with Kev. I get why children are 64-bytes in length (uncompressed version), but I'm not sure why the length of the array is itself 64, shouldn't it be 256 since an internal node can have 256 children?

packages/verkle/src/verkleTree.ts
```diff
@@ -162,7 +162,9 @@ export class VerkleTree {
    * @returns A Promise that resolves to `Uint8Array` if a value was found or `null` if no value was found.
    */
   async get(key: Uint8Array): Promise<Uint8Array | null> {
-    verifyKeyLength(key)
+    // TODO: Determine if we actually need to do this check.  The key for an internal node will necessarily
+    // be less than 32 bytes since it's the path to that node in the trie
+    // verifyKeyLength(key)
```
gabrocheleau, line 167: Wouldn't `get` always be used for full 32-byte keys? The way I'm thinking of it is:
- `trie.get` is the high-level method that we use for 32-byte keys (just like we `put` values using 32-byte keys)
- Get only returns actual "end values"
- The "path to the node in the trie" which is referred to in the comment wouldn't be directly accessed with the `get` method (even though the logic would navigate through these internal node swhen walking the trie) 

There I think we should verifyKeyLength

packages/verkle/src/verkleTree.ts
```diff
@@ -277,10 +278,23 @@ export class VerkleTree {
     let currentDepth = leafNode.depth
 
     while (currentDepth > 0) {
-      const parentKey = currentKey.slice(0, -1)
       const parentIndex = currentKey[currentKey.length - 1]
-      const parentNode = InternalNode.create(currentDepth, this.verkleCrypto)
-      parentNode.children[parentIndex] = currentNode
+      // Parent key should be at least one shorter than the current key length
+      const parentKey = currentKey.slice(0, currentKey.length - 1)
+      const rawNode = await this.get(parentKey)
```
gabrocheleau, line 284: I'm not sure that this is a proper use of `this.get`? Shouldn't we use findPath instead for e.g.?

packages/verkle/src/verkleTree.ts
```diff
@@ -277,10 +278,23 @@ export class VerkleTree {
...
+      let parentNode
+      // TODO: Determine how to decide if we need to go further up the tree or create a new internal node here
+      // Currently, this code would insert a new internal node at every step up the trie (which we don't want)
```
gabrocheleau, line 287: The way I'm thinking of this, we could find the nearest existing path to the leaf node that we want to insert, while keeping track of all the tries that have led to that, and update them sequentially, rather than coming from the child and looking up each parent. Does that make sense?

packages/verkle/src/node/internalNode.ts
acolytec3: That's a bug. Should be 256.

packages/verkle/src/verkleTree.ts
acolytec3: So there's never an instance where you might want to `get` an internal node?

packages/verkle/src/verkleTree.ts
acolytec3: Oh, you know, I think i meant to do `trie._db.get` here. Maybe that's the difference I was missing before

packages/verkle/src/verkleTree.ts
gabrocheleau: I think there are instances where we'd want to retrieve an internal node, but not using the trie.get method

acolytec3: Sounds good.  Will revert this change.

holgerd77:
Great work going on here, really cool to see! 🤩

Have added the following TODO to the breaking release tracking issue #3216! Not fully sure if feasible but that would be great if this would be possible and guess timing would fit well (with Verkle progress + breaking release timeline).

> - [ ] See if we can come up with a generic tree interface for both Trie and Verkle which we can put in Common (as we do for `StateManager` e.g.) which abstracts away to a greater extend the need for a Verkle/Trie distinction (at least for some places/usages, might also turn out to be only a subset of the functionality)

packages/verkle/src/node/internalNode.ts
```diff
@@ -1,119 +1,91 @@
...
   }
 
-  setChild(index: number, child: VerkleNode) {
-    this.children[index] = child
+  // Updates the commitment value for a child node at the corresponding index
+  setChild(childIndex: number, child: ChildNode) {
+    // Get previous child commitment at `index`
+    const oldChildReference = this.children[childIndex]
+    // Updates the commitment to the child node at `index`
+    this.children[childIndex] = { ...child }
```
gabrocheleau, line 28: Why are we spreading into an object? To clone the object and not just copy the reference?

acolytec3: Correct.  Because objects are reference types in Javascript, I need to clone it here because the `child` was being passed by reference in the call to `setChild` and then getting garbage collected when `setChild` returned.  I think this is probably essentially what the `copyOnWrite` functionality is doing in `go-verkle`

packages/verkle/src/types.ts
gabrocheleau: Hmm yeah. I'm not convinced about what the naming should actually be, but I don't find zeroValues/zeroCValues insightful/clear either. I'll try to think of alternatives..

gabrocheleau: Maybe just 
```
export const emptyByteArrays = new Array(256).fill(new Uint8Array());
export const emptyByteArraysHalfSize = new Array(128).fill(new Uint8Array());
```
 ?

acolytec3:
This is ready for final review.

packages/verkle/src/node/internalNode.ts
```diff
@@ -1,119 +1,94 @@
...
+      // The hashed child commitments are used when updating the internal node commitment
+      this.verkleCrypto.hashCommitment(oldChildReference.commitment),
+      this.verkleCrypto.hashCommitment(child.commitment)
+    )
   }
 
-  static fromRawNode(rawNode: Uint8Array[], depth: number): InternalNode {
+  static fromRawNode(
+    rawNode: Uint8Array[],
+    depth: number,
```
gabrocheleau, line 41: Can we remove depth from this API?

packages/verkle/src/node/leafNode.ts
```diff
@@ -21,11 +23,78 @@ export class LeafNode extends BaseVerkleNode<VerkleNodeType.Leaf> {
...
+   * @param stem the 31 byte stem corresponding to the where the leaf node should be placed in the trie
+   * @param values the 256 element array of 32 byte values stored in the leaf node
+   * @param depth the depth of the leafnode
+   * @param verkleCrypto the verkle cryptography interface
+   * @returns an instantiated leaf node with commitments defined
+   */
+  static async create(
+    stem: Uint8Array,
+    values: Uint8Array[],
+    depth: number,
```
gabrocheleau, line 37: Can we remove depth here too?

packages/verkle/src/node/util.ts
```diff
@@ -1,30 +1,64 @@
...
 import { LeafNode } from './leafNode.js'
 import { type VerkleNode, VerkleNodeType } from './types.js'
 
-export function decodeRawNode(raw: Uint8Array[]): VerkleNode {
+import type { VerkleCrypto } from '@ethereumjs/util'
+
+// TODO: Make depth optional since we don't always know what it is
+export function decodeRawNode(
+  raw: Uint8Array[],
+  depth: number,
```
gabrocheleau, line 13: Remove depth?

packages/verkle/src/node/util.ts
```diff
@@ -1,30 +1,64 @@
...
+  verkleCrypto: VerkleCrypto
+): VerkleNode {
   const nodeType = raw[0][0]
-  const depth = 0
   switch (nodeType) {
     case VerkleNodeType.Internal:
-      return InternalNode.fromRawNode(raw, depth)
+      return InternalNode.fromRawNode(raw, depth, verkleCrypto)
```
gabrocheleau, line 19: Won't repeat for all cases but I'd remove all instances of depth

packages/verkle/src/verkleTree.ts
```diff
@@ -209,193 +412,150 @@ export class VerkleTree {
...
-            resolve({ node, remaining: new Uint8Array(0), stack })
-          } else {
-            // reached leaf but keys don't match
-            resolve({ node: null, remaining: keyRemainder, stack })
-          }
-        }
-      }
+  async findPath(key: Uint8Array): Promise<Path> {
+    this.DEBUG && this.debug(`Path (${key.length}): [${bytesToHex(key)}]`, ['FIND_PATH'])
+    // TODO: Decide if we should allow keys longer than 31 bytes (since a verkle stem can never be longer than that)
```
gabrocheleau, line 417: Can we remove this Todo? We've already discussed this right?

packages/verkle/src/verkleTree.ts
```diff
@@ -209,193 +412,150 @@ export class VerkleTree {
...
+    while (!finished) {
+      rawNode = await this._db.get(child.path)
+      // We should always find the node if the path is specified in child.path
+      if (rawNode === undefined) throw new Error(`missing node at ${bytesToHex(child.path)}`)
+      const decodedNode = decodeNode(rawNode, result.stack.length, this.verkleCrypto)
+
+      // Calculate the index of the last matching byte in the key
+      const matchingKeyLength = matchingBytesLength(key, child.path)
+      const foundNode = equalsBytes(key, child.path)
+      if (foundNode || child.path.length >= key.length || decodedNode instanceof LeafNode) {
```
gabrocheleau, line 452: Is the last condition necessary? If I'm thinking about this correctly, whenever the child path length is longer than the key, the found node would necessarily be a leaf node?

packages/verkle/src/node/baseVerkleNode.ts
```diff
@@ -2,24 +2,22 @@ import { RLP } from '@ethereumjs/rlp'
 
 import { type VerkleNodeInterface, type VerkleNodeOptions, type VerkleNodeType } from './types.js'
 
+import type { VerkleCrypto } from 'verkle-cryptography-wasm'
```
holgerd77, line 5: This should now use the import from `verkle` in `Util`, just opened a related issue here https://github.com/ethereumjs/verkle-cryptography-wasm/issues/54

I would go as far and argue that we should *also* remove the `verkle-cryptography-wasm` dependency from this package as well and dependency inject, and so to have this wider strategic goal over the head that this should be replaceable and also here optimally a JS implementation being used.

We then also automatically continue to evolve a generic interface around this and keep thinking about what's needed and what not (I e.g. still like the idea that we take the more hashing related parts out *directly* and move over to the pedersen/poseidon hash usage from Pauls libraries https://github.com/paulmillr/scure-starknet (if applicable) and with this slowly shrink down the surface of what we need to compile over into the WASM stuff.

Feels to me that there is still too much in. 

Practical note: if depency injection feels a bit impractical for now for testing and the like we could also decide to do at a later stage.

holgerd77: Ah, but the main argument for direct dependency injection (so, this would mean: having a mandatory `crypto` constructor argument) is actually that we avoid having the same bundle problems again once/if we have Verkle added somewhere as a dependency.

holgerd77 (commented):
Couldn't really dive deep into the code base this morning, so just some quick broader comment on how we want to integrate the crypto package.

But pretty cool to see this evolving so strongly, looking forward to go deeper! 🤩

packages/verkle/src/types.ts
acolytec3: Okay, lets go with this for now. Will update.

acolytec3: On second thought, I don't see these used anywhere so no real point I don't think. I'm just going to remove entirely until we actually use them.

gabrocheleau: Fair enough!

packages/verkle/src/node/internalNode.ts
```diff
@@ -58,13 +58,16 @@ export class InternalNode extends BaseVerkleNode<VerkleNodeType.Internal> {
...
     })
-    return new InternalNode({ commitment, depth, verkleCrypto, children })
+    return new InternalNode({ commitment, verkleCrypto, children })
   }
 
-  static create(depth: number, verkleCrypto: VerkleCrypto): InternalNode {
+  /**
+   * Generates a new Internal node with default commitment
+   */
+
```
gabrocheleau, line 67:
```suggestion
```

packages/verkle/src/verkleTree.ts
```diff
@@ -214,202 +214,196 @@ export class VerkleTree {
...
-      // We're done so return early
-      return
-    }
+  // TODO: Rewrite following logic in verkle.spec.ts "findPath validation" test
+  async put(_key: Uint8Array, _value: Uint8Array): Promise<void> {
+    throw new Error('not implemented')
+    // verifyKeyLength(key)
+    // const stem = key.slice(0, 31)
+    // this.DEBUG && this.debug(`Key: ${bytesToHex(key)}`, ['PUT'])
+    // this.DEBUG && this.debug(`Value: ${bytesToHex(value)}`, ['PUT'])
```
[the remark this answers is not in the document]
acolytec3: Sure, have stashed it locally

gabrocheleau (dismissed):

gabrocheleau (approved):

PR #3433, opened 2024-05-17
client: kaustinen7 verkle testnet preparation

gabrocheleau:
This PR makes the following updates to ready our client for Kaustinen stateless syncing:

- [x] Removes the stateRoot handling and caching for the previous block. The parent block's stateRoot will now be provided in the verkle execution witness directly, so we can retrieve it from it. 
- [x] Updates the Verkle leaf structure. The version, nonce, balance and codeSize fields have been packed into a single `BASIC_DATA` field. See https://github.com/ethereum/EIPs/pull/8550 for more details.
- [ ] Update test data to Kaustinen7 test data
- [ ] Test Kaustinen7 syncing


g11tech (commented):
yes we look mostly uptodate, i have an open PR regarding system accounts that we needs to be finalized

acolytec3 (commented):
Have updated this based on latest master and cleaned up the references to stateroot in `verifyVerkleProof`

packages/statemanager/src/statelessVerkleStateManager.ts
```diff
@@ -520,13 +500,13 @@ export class StatelessVerkleStateManager implements StateManagerInterface {
    * @param {Uint8Array} stateRoot - The stateRoot to verify the executionWitness against
    * @returns {boolean} - Returns true if the executionWitness matches the provided stateRoot, otherwise false
    */
-  verifyVerkleProof(stateRoot: Uint8Array): boolean {
```
scorbajio, line 523: Renaming `verifyVerkleProof` on the `StatelessVerkleStateManager` to  `verifyProof` would be more in line with the changes in #3557 and with other current implementations of statemanagers that we have.

gabrocheleau: This might be something that we'd want to extract to separate helpers, in line with what we've been doing with some other classes (extracting methods into standalone helpers)

packages/util/src/verkle.ts
```diff
@@ -215,3 +222,46 @@ export const getVerkleTreeKeyForStorageSlot = async (
...
+  const version = bytesToInt32(versionBytes, true)
+  const nonce = bytesToBigInt(nonceBytes, true)
+  const codeSize = bytesToInt32(codeSizeBytes, true)
+  const balance = bytesToBigInt(balanceBytes, true)
+
+  return { version, nonce, codeSize, balance }
+}
+
+export function encodeVerkleLeafBasicData(basicData: VerkleLeafBasicData): Uint8Array {
+  const encodedVersion = setLengthRight(
```
g11tech, line 250: littleending=true and setlength right combo might not be correct

acolytec3:
@gabrocheleau could we skip the failing tests for now and then continue work for the kaustinen7 changes (notably the fix I have for the `encodeBasicData` function) in the statefulVerkleStateManager  branch?

packages/util/src/verkle.ts
```diff
@@ -215,3 +222,46 @@ export const getVerkleTreeKeyForStorageSlot = async (
 
   return concatBytes(getVerkleStem(verkleCrypto, address, treeIndex), toBytes(subIndex))
 }
+
+export function decodeVerkleLeafBasicData(encodedBasicData: Uint8Array): VerkleLeafBasicData {
```
holgerd77, line 226: Just to note that we should at some point (best still before breaking releases even if not particularly about Verkle) give the Verkle docs a bit more love and e.g. complete and clean up the docs in this `Util` module (and e.g. give methods like `decodeVerkleLeafBasicData()` some code docs).

Also the Verkle [README](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/verkle#ethereumjsverkle) should giv a somewhat more complete picture on and introduction to the library.

packages/statemanager/src/statelessVerkleStateManager.ts
acolytec3: Will open a new issue for this one since we can update it for both of the verkle SM implementations at once.

acolytec3 (approved):
LGTM - will address kaustinen7 test data needs in a future PR

packages/statemanager/src/statelessVerkleStateManager.ts
scorbajio: Not sure if one has already been created, but created #3629 for this.

packages/util/src/verkle.ts
scorbajio: I can give this a look and get some changes started.

scorbajio: Got a basic description and example set up at #3630.

PR #3435, opened 2024-05-21
Fix tx status in `eth_getTransactionReceipt`

acolytec3:
While running `assertoor` tests in kurtosis, we identified a bug in our code where we were not returning a `status` field as part of the response from `eth_getTransactionReceipt`.  This PR brings our code into line with the [spec](https://ethereum.github.io/execution-apis/api-documentation/)


gabrocheleau (approved):
Lgtm

PR #3437, opened 2024-05-23
Add benchmark for bytecode execution

JacekGlen:
This PR adds an EVM benchmark that can be run for individual bytecodes. 

The main motivation for the tool is to be used in [gas-cost-estimator](https://github.com/imapp-pl/gas-cost-estimator) project, but its purpose goes beyond it. You can learn more about the project from this short [presentation](https://streameth.org/devconnect/watch?session=65b8f8d5a5b2d09b88ec1838), or visit the wiki [Stage 4 Scope](https://github.com/imapp-pl/gas-cost-estimator/wiki/Stage-4-Scope).

The benchmark implementation differs slightly as it does not add to the existing `Benchmark.Suite`. Instead, it creates a new `Benchmark` object for each run and handles async resolution. This was proven to produce more consistent results.


packages/vm/benchmarks/bytecode.ts
```diff
@@ -0,0 +1,93 @@
+import { readFileSync } from 'fs'
+import Benchmark from 'benchmark'
```
roninjin10, line 2: Curious if there is a reason to use this benchmark package rather than vitest built in benchmarking features?

packages/vm/benchmarks/bytecode.ts
```diff
@@ -0,0 +1,93 @@
...
+
+type EvmCreator = () => Promise<EVM>
+
+async function runBenchmark(bytecode: string, createEvm: EvmCreator): Promise<Stats> {
+  const evm = await createEvm()
+
+  let promiseResolve: any
+  const resultPromise: Promise<Stats> = new Promise((resolve, reject) => {
+    promiseResolve = resolve
+  })
```
roninjin10, line 24:
```suggestion
  let promiseResolve: any
  const resultPromise: Promise<Stats> = new Promise((resolve, reject) => {
    promiseResolve = resolve
  })
  const {resolve: promiseResolve, promise: resultPromise} = Promise.withResolvers()
```
Should be able to do this

packages/vm/benchmarks/bytecode.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  for (let i = 0; i < (numSamples ?? 1); i++) {
+    if (suite) {
+      let results = await runBenchmark(bytecode!, createEvm)
+      console.log(
+        `${i}, ${results.iterationsCount}, ${results.totalTimeNs}, ${results.stdDevTimeNs}`
+      )
+    } else {
+      let evm = await createEvm()
+      evm.runCode({
```
roninjin10, line 87: Should this have an await?
```suggestion
      await evm.runCode({
```

packages/vm/benchmarks/bytecode.ts
acolytec3: I haven't had a chance to review the whole PR yet but our use of `benchmark` long predates our use of `vitest`. We don't use benchmarking of this sort all that much in our current development process but there may be a reason to switch at some point.

JacekGlen: It's consistent with the existing implementation, see `run.ts` and `mainnetBlocks.ts`. I don't mind migrating to  `vitest` if that's the preferred way.

packages/vm/benchmarks/bytecode.ts
JacekGlen: I could be wrong, but I think that `Promise.withResolvers()` has been introduced in TS v5.4.0, and this repo's minimal requirement is `"typescript": "4.7.4"`. But I need someone with a better knowledge of this to confirm.

packages/vm/benchmarks/bytecode.ts
JacekGlen: Of course! Good catch

packages/vm/benchmarks/run.ts
```diff
@@ -1,52 +1,58 @@
 import Benchmark from 'benchmark'
 import { BenchmarksType } from './util'
 import { mainnetBlocks } from './mainnetBlocks'
+import { bytecode } from './bytecode'
+import { ArgumentParser } from 'argparse'
```
acolytec3, line 5: I think I'd rather use `yargs` here since we already have it in other places in the monorepo rather than add yet another args parsing package to our overall monorepo.  I'd be happy to pick this up if you're not familiar with the API.

packages/vm/benchmarks/bytecode.ts
```diff
@@ -0,0 +1,93 @@
...
+  const createEvm: EvmCreator = async () => {
+    const profiler = { enabled: suite ? false : true }
+    return await EVM.create({ stateManager, common, profiler })
+  }
+
+  for (let i = 0; i < (numSamples ?? 1); i++) {
+    if (suite) {
+      let results = await runBenchmark(bytecode!, createEvm)
+      console.log(
+        `${i}, ${results.iterationsCount}, ${results.totalTimeNs}, ${results.stdDevTimeNs}`
```
acolytec3, line 83: Can you add some sort of labeling that defines what each of the columns indicates?  The output as currently logged is hard to understand unless you inspect the code of the benchmark.

packages/vm/benchmarks/bytecode.ts
acolytec3: @JacekGlen how much work would it be to switch all of this to the `vitest` benchmarking tools? Looks like it produces some [nice output](https://vitest.dev/guide/features.html#benchmarking) and might be easier to use over all.

packages/vm/benchmarks/bytecode.ts
roninjin10: I would just bump the typescript version up 4.7.4 is very old. Feel free to ignore though this is just a stylistic nit.

packages/vm/benchmarks/bytecode.ts
roninjin10: If helpful, I am doing vitest benchmark testing in the [test/bench](https://github.com/evmts/tevm-monorepo/tree/main/test/bench) folder

JacekGlen: @acolytec3 @roninjin10 I have updated the benchmarks to use `tinybench` (via `vitest`). 
The reason for using the benchmark library directly, rather than unit-test style preferred by `vitest`, is that this is more command-line tool. Using the benchmark library directly gives better control over parameters and output.

packages/vm/benchmarks/run.ts
JacekGlen: Done

packages/vm/benchmarks/bytecode.ts
JacekGlen: Done

holgerd77:
Hi there,
sorry that this takes such a somewhat solid amount of time to review here and comment, everyone on the team is pretty busy with relatively heavy implementational stuff regarding EIPs and the like.

Generally this is pretty interesting work though, we are really eager to continue to have the tooling around benchmarking improved!

@jochem-brouwer recently also worked on some benchmarking tooling regarding the performance of bytecodes for the VM - see https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/DEVELOPER.md#benchmarks - and maybe a reason for a slow start here is also that we are (at least: I am) not really sure how these two tools relate and if we are producing some redundant tooling here or not.

Can you have some word on this? That would help for sure on the process! 🙏 🙂

PR #3438, opened 2024-05-25
vm: updated 2935 tests with the new proposed bytecode and corresponding config

g11tech:
update the test with new proposed bytecode and deployment  configs

ref: 
 - https://github.com/ethereum/EIPs/pull/8596
 
 Note: this test setup can/will be used to test various flavours of the 2935 contract as well as can be used as inspirations for other system contracts


holgerd77:
Is the discussion around this spec/code change settled(enough)?

holgerd77 (approved):
Nevermind, just re-read your introduction and let it sink in, will approve and merge.

PR #3440, opened 2024-05-29
Implement Mega EOF

jochem-brouwer:
https://github.com/ipsilon/eof/blob/main/spec/eof.md

https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7692.md

Not ready for review yet, there is still a lot left to do. 

I have some local changes which are super dirty and I did not push yet, but will do once I've cleaned that up (mainly in execution).

To run tests: (note: this setups both blockchain/state and header tests)

1. Download fixtures from latest release in `execution-spec-tests`, e.g. https://github.com/ethereum/execution-spec-tests/releases/tag/eip7692%40v1.0.4
2. Untar this and put everything inside `./packages/ethereum-tests`, for instance inside the "eof" folder
3. Now run the **blockchain** tests from `./packages/vm`: `npm run test:blockchain -- --fork=Prague --dir=../eof/fixtures/blockchain_tests`
4. Now run the **state** tests from `./packages/vm`: `npm run test:blockchain -- --fork=Prague --dir=../eof/fixtures/state_tests`
5. For the eof header tests, first change `./packages/evm/test/eips/eof-header-validation.spec.ts` to make `testDir` point to the right directory, change line 14 to: `const testDir = path.resolve('../ethereum-tests/eof/fixtures/eof_tests')`. Rename the file to `eof-header-validation.spec.ts`
6. Run the header tests: `cd ./packages/evm`: `npx vitest run ./test/eips/eof-header-validation.spec.ts`

Gotchas:

- The execution tests ALWAYS run on verified EOF code. The execution tests thus ONLY cover the execution part of EOF
- There are also `eof-tests` in the above listed fixtures. These are actual header tests, and these verify if the EOF containers are correct

Code gotchas:

- The "eof code exectution" is somewhat hacked into the current EVM, by setting the program counter to the first part of the code section. Any code section called to (via `CALLF`) will save the correct pc once returned, and will update the pc to the start of the called code section
- There are some special cases for `EXT*CALL` which should put `1` on the stack, this sets a special flag for `messageGasLimit` of `-BIGINT_1` (yes, negative), this is also somewhat a hack (we can provide an extra flag)
- The code is super unoptimized and will likely need multiple PRs on top to optimize it
- This code will also likely impact the speed of legacy contracts (since calls such as EXTCODEHASH will now first read the complete EXTCODE to check if it is EOF code)
- The `EOFContainer` will give access to the current container. However, it will not pre-load the subcontainers of the container. This is ONLY done into `validateEOF` which will recursively load these subcontainers. 

TODO

- [ ] The validation of EOF now has a lot of flags (two enums, and a boolean flag). This should be cleaned up, it should be investigated what situations there are (initcode, runtime code, and what type of containers there are: init container and deployment container?). It is now very messy and it can likely be cleaned up to 1 or 2 flags.


packages/evm/src/eof/stackDelta.ts
```diff
@@ -0,0 +1,472 @@
...
+  OpcodeInfo.unallocatedOpcode(0xfc),
+  OpcodeInfo.terminalOpcode('REVERT', 0xfd, 2, 0, 1),
+  OpcodeInfo.terminalOpcode('INVALID', 0xfe, 0, 0, 1),
+  OpcodeInfo.invalidOpcode('SELFDESTRUCT', 0xff),
+]
+
+console.log(JSON.stringify(stackDelta))
+
+
+*/
```
holgerd77, line 472: Maybe we put this generation code into `evm/scripts` and then reference from here? 🤔

jochem-brouwer: Yup, good point, will do!

jochem-brouwer:
As of https://github.com/ethereumjs/ethereumjs-monorepo/pull/3440/commits/28feb483c8db4035eff1b1e1bc6fbb65a74c1e0f

We pass https://github.com/ethereum/execution-spec-tests/releases/tag/eip7692%40v1.0.4 :tada:

jochem-brouwer:
As of https://github.com/ethereumjs/ethereumjs-monorepo/pull/3440/commits/1e97ba19f17f33cd0f4efd6f442b08e93c53d0c9, we pass the EOF container validation tests of https://github.com/ethereum/execution-spec-tests/releases/tag/eip7692%40v1.0.5

We also pass all 1.0.5 state and blockchain tests now!

packages/evm/src/eof/stackDelta.ts
holgerd77: Still open task

packages/evm/src/exceptions.ts
```diff
@@ -35,6 +37,39 @@ export enum ERROR {
...
+  INVALID_EOF_OUTPUTS = 'invalid EOF type section: too much outputs',
+  INVALID_EOF_MAX_STACK_HEIGHT = 'invalid EOF type section: too high max stack height',
+
+  INVALID_EOF_CODE_OPCODE_UNDEFINED = 'invalid EOF bytecode: undefined opcode found',
+
+  INVALID_EOF_DATA_SECTION_SIZE = 'invalid EOF data section size: data section too large',
+
+  INVALID_EOF_FIRST_CODE_SECTION_TYPE = 'invalid EOF: first code section must have 0 inputs and should have 0x80 (non-returning) outputs',
+
+  INVALID_EOF_FINAL_SECTION_OPCODE = 'invalid EOF: code section does not have a valid final (terminating) instruction opcode',
```
holgerd77, line 72: Just throwing in one-off questions and not doing in the form a full review yet: was is the difference/distinction between the errors here and the ones in `eof/errors.ts`? This partly looks redundant on first look?

packages/evm/src/eof/container.ts
```diff
@@ -0,0 +1,367 @@
...
+import { EOFError, validationError } from './errors.js'
+import { ContainerSectionType, verifyCode } from './verify.js'
+
+import type { EVM } from '../evm.js'
+
+export enum EOFContainerMode {
+  Default, // Default container validation
+  Initmode, // Initmode container validation (for subcontainers pointed to by EOFCreate)
+  TxInitmode, // Tx initmode container validation (for txs deploying EOF contracts)
+}
```
holgerd77, line 32: Not fully sure if this is only about validation or not.

If it is I would suggest the following renaming for better clarity:

```ts
export enum EOFContainerValidationMode {
  Default, // Default container validation
  EOFCreateSubcontainerInit, // Initmode container validation (for subcontainers pointed to by EOFCreate)
  TxEOFContractDeployment, // Tx initmode container validation (for txs deploying EOF contracts)
}
```

(or similar, maybe take the bits out you like or evolve 🙂)

The current version is some mix out of being an enum for a "container type" (in the wider sense) and a "validation mode", which is a bit ambiguous.

packages/evm/src/eof/container.ts
```diff
@@ -0,0 +1,367 @@
...
+// TODO add initcode flags (isEOFContract)
+// TODO validation: mark sections as either initcode or runtime code to validate
+
+class EOFHeader {
+  typeSize: number
+  codeSizes: number[]
+  containerSizes: number[]
+  dataSize: number
+  dataSizePtr: number // Used to edit the dataSize in RETURNCONTRACT
+  buffer: Uint8Array
```
holgerd77, line 98: What is this `buffer` in `EOFHeader`? Can we name this eventually a bit more expressively?

jochem-brouwer: The `buffer` is the entire buffer which the Header consists of (I will rename)

packages/evm/src/eof/container.ts
```diff
@@ -0,0 +1,367 @@
...
+
+  constructor(
+    buf: Uint8Array,
+    header: EOFHeader,
+    eofMode: EOFContainerMode = EOFContainerMode.Default,
+    dataSectionAllowedSmaller = false // Only for validation: Deployment containers are allowed to have smaller data section size
+  ) {
+    const stream = new StreamReader(buf)
+    const typeSections: TypeSection[] = []
+    for (let i = 0; i < header.typeSize / 4; i++) {
```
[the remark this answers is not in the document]
jochem-brouwer: > Maybe also connect more between the code/implementation and the spec, with occasional deep links like https://eips.ethereum.org/EIPS/eip-3540#container into the EIP.

Right, sorry, I realize the code is not that well documented. I should indeed have linked towards the EIP links deep inside the docs.

Just for a quick reply: the entire container file is essentially 3540 (container validation) / stack validation (5450), depending on various contexts (initcode / tx init code). Note that already-deployed containers do not need to be validated - this is done at contract deployment, where the published EOF container with the subcontainers are recursively validated.

Note: there are checks all over the place in the EIPs, for a complete spec in one file, read this file: https://github.com/ipsilon/eof/blob/main/spec/eof.md. It has links to the EIPs themselves too.

packages/evm/src/eof/container.ts
```diff
@@ -0,0 +1,367 @@
...
+  typeSize: number
+  codeSizes: number[]
+  containerSizes: number[]
+  dataSize: number
+  dataSizePtr: number // Used to edit the dataSize in RETURNCONTRACT
+  buffer: Uint8Array
+
+  private codeStartPos: number[]
+
+  constructor(input: Uint8Array) {
```
holgerd77, line 102: I guess adding a simple debug logger `evm:eof` with a few strategic occasions for debug output like "Creating EOF header with 4 code sections and ..." would help a lot for making things more explicit and debug and analyze on various (test) situations and scenarios.

Doesn't need to be extensive on first round (so, e.g. 4-6 debug messages already a good start).

packages/evm/src/eof/container.ts
jochem-brouwer: Yes those flags are confusing, I wanted to look if I can compress these in a single flag instead of 2 flags with 3 options (this leads to 9 unique combinations).

packages/evm/src/eof/verify.ts
```diff
@@ -0,0 +1,503 @@
...
+        validationError(EOFError.StackUnderflow)
+      }
+
+      const delta = opcodeOutputs - opcodeInputs
+
+      let minStackNext = minStackCurrent + delta
+      let maxStackNext = maxStackCurrent + delta
+
+      if (maxStackNext > 1023) {
+        // TODO verify if 1023 or 1024 is the right constant
```
acolytec3, line 200: EOF spec seems to says this should be 1023 (assuming this code section is for a complete EOF function)

packages/evm/src/opcodes/functions.ts
```diff
@@ -984,6 +1011,346 @@ export const handlers: Map<number, OpHandler> = new Map([
...
+
+        const gasLimit = runState.messageGasLimit!
+        runState.messageGasLimit = undefined
+
+        let data = new Uint8Array(0)
+        if (inputSize !== BIGINT_0) {
+          data = runState.memory.read(Number(inputOffset), Number(inputSize), true)
+        }
+
+        runState.programCounter++ // Jump over the immediate byte
```
acolytec3, line 1291: Is this byte the `EOFCREATE` opcode?

packages/evm/src/opcodes/functions.ts
```diff
@@ -984,6 +1011,346 @@ export const handlers: Map<number, OpHandler> = new Map([
...
+        const originalDataSize = deployContainer.header.dataSize
+        const preDeployDataSectionSize = deployContainer.body.dataSection.length
+        const actualSectionSize = preDeployDataSectionSize + Number(auxDataSize)
+
+        if (actualSectionSize < originalDataSize) {
+          trap(ERROR.INVALID_RETURNCONTRACT_DATA_SIZE)
+        }
+
+        if (actualSectionSize > 0xffff) {
+          // Data section size is now larger than the max data section size
```
acolytec3, line 1336: Could this ever happen in EOF?  I thought the code validation would prevent this?

packages/evm/src/index.ts
```diff
@@ -1,4 +1,4 @@
-import { EOF } from './eof.js'
+//import { EOF } from './eof.js'
```
acolytec3, line 1: Can we remove this?

packages/evm/src/interpreter.ts
```diff
@@ -11,7 +12,11 @@ import {
 } from '@ethereumjs/util'
 import debugDefault from 'debug'
 
-import { EOF } from './eof.js'
+//import { EOF } from './eof/eof.js'
```
acolytec3, line 15: Same, can we remove?

packages/evm/src/interpreter.ts
```diff
@@ -184,40 +190,54 @@ export class Interpreter {
   }
 
   async run(code: Uint8Array, opts: InterpreterOpts = {}): Promise<InterpreterResult> {
-    if (!this.common.isActivatedEIP(3540) || code[0] !== EOF.FORMAT) {
+    if (!this.common.isActivatedEIP(3540) || code[0] !== FORMAT) {
       // EIP-3540 isn't active and first byte is not 0xEF - treat as legacy bytecode
       this._runState.code = code
     } else if (this.common.isActivatedEIP(3540)) {
-      if (code[1] !== EOF.MAGIC) {
+      if (code[1] !== MAGIC) {
```
acolytec3, line 197: Just occurred to me, will we ever have other kinds of magic (e.g. authcall tx call data MAGIC bytes or something like that)?  Would hate to have to rename this again later if we end up with multiple types of magic bytes to check for.

acolytec3 (commented):
I need to go over all the validation rules in detail again with the spec next to me but have left some initial comments.

packages/evm/src/opcodes/functions.ts
jochem-brouwer: No, it jumps over the immediate byte, which is the container the EOFCREATE points to. You can compare it with a PUSH1 operation, in that case it would "jump over" the byte which is being pushed to stack.

packages/evm/src/opcodes/functions.ts
jochem-brouwer: This can happen in the data section. The data section of a not-yet-deployed contract can be lower than, or equal to, the data size in the contract header. When the contract is being deployed, the RETURNCONTRACT will actually point to a memory region of data to append. 

I don't know why I added "Temp: trap OOG?" there, maybe this is not in the spec (but it still seems to pass the reference tests). So, you are right, this is actually one of the cases where an invalid container might be deployed.

packages/evm/src/index.ts
jochem-brouwer: Yes I should remove this, however we should consider if we want to expose the `EOFContainer` :thinking:

packages/evm/src/interpreter.ts
jochem-brouwer: Oh wait this might actually be a leftover of old code (will check)

packages/evm/src/interpreter.ts
jochem-brouwer: I have used the terminology of EOF, but you are right, the current 7702 spec essentially uses the magic of 0x01 to designate some particular behavior :sweat_smile:

packages/evm/src/eof/verify.ts
jochem-brouwer: I should likely remove this TODO, I think I changed it to 1024 and then reference tests started to fail.

packages/evm/src/eof/container.ts
```diff
@@ -0,0 +1,461 @@
...
+      }
+    } else {
+      dataSection = stream.readRemainder()
+    }
+
+    // Write all data to the object
+    this.typeSections = typeSections
+    this.codeSections = codes
+    this.containerSections = containers
+    this.entireCode = entireCodeSection
```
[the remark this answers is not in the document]
jochem-brouwer: Good point, we likely don't need it anywhere (will check). This might be from an old version of EOF where it was still possible (I think) to inspect own code

packages/evm/src/opcodes/functions.ts
jochem-brouwer: Ah, we can likely remove this. If you try to deploy a data section of size `0xffff`, you try to deploy a contract with higher byte length than the maximum, so it will go OOG anyways.

packages/evm/src/eof/stackDelta.ts
jochem-brouwer: Have moved this to a script!

packages/evm/src/exceptions.ts
jochem-brouwer: I have moved all errors, except INVALID EOF CONTAINER inside the EOF errors. The reason that the prior stays in exceptions is that it is an EvmError and therefore needs to stay there. (This error is also caused by EIP-3541, it is CREATE/CREATE2 trying to deploy EOF containers)

packages/evm/src/eof/container.ts
jochem-brouwer: I have to revisit renaming those validation modes. For the other flag, I have added a long comment explaining what it does.

packages/evm/src/eof/container.ts
jochem-brouwer: I have added a comment, should I still rename?

packages/evm/src/eof/container.ts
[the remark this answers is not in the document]
jochem-brouwer: I have added much more comments and documentation!

packages/evm/src/eof/container.ts
```diff
@@ -0,0 +1,461 @@
...
+
+  /**
+   * Create an EOF header. Performs various validation checks inside the constructor
+   * @param input The input should either be a raw header, or a complete container
+   */
+  constructor(input: Uint8Array) {
+    if (input.length > MAX_HEADER_SIZE) {
+      throw new Error('err: container size more than maximum valid size')
+    }
+    const stream = new StreamReader(input)
```
acolytec3, line 152: I get why you called it "stream" and `streamReader` but this title is throwing me.  Can we call it something like `eofBytesReader` or `eofBytesParser` and then move it to its own file?  It adds a bunch of lines to this file and makes it kinda overwhelming to me.

packages/evm/src/eof/container.ts
```diff
@@ -0,0 +1,461 @@
...
+    // Verify that the header starts with 0xEF0001
+    stream.verifyUint(FORMAT, EOFError.FORMAT)
+    stream.verifyUint(MAGIC, EOFError.MAGIC)
+    stream.verifyUint(VERSION, EOFError.VERSION)
+    if (input.length < 15) {
+      throw new Error('err: container size less than minimum valid size')
+    }
```
acolytec3, line 159: I would do this check first (so put it above verifying the header preamble).

packages/evm/src/eof/container.ts
```diff
@@ -0,0 +1,461 @@
...
+    // Verify that the types section is present, and verify that the type section length is valid
+    stream.verifyUint(KIND_TYPE, EOFError.KIND_TYPE)
+    const typeSize = stream.readUint16(EOFError.TypeSize)
+    if (typeSize < TYPE_MIN) {
+      validationError(EOFError.InvalidTypeSize, typeSize)
+    }
+    if (typeSize % TYPE_DIVISOR !== 0) {
+      validationError(EOFError.InvalidTypeSize, typeSize)
+    }
```
acolytec3, line 168: Total nit but can we combine these since the validation error is the same?
```suggestion
    if ((typeSize < TYPE_MIN) || (typeSize % TYPE_DIVISOR !== 0))  {
      validationError(EOFError.InvalidTypeSize, typeSize)
    }
```

packages/evm/src/eof/container.ts
```diff
@@ -0,0 +1,461 @@
...
+    if (nextSection === KIND_CONTAINER) {
+      // The optional container section is present, validate that the size is within bounds
+      const containerSectionSize = stream.readUint16(EOFError.ContainerSize)
+
+      if (containerSectionSize < CONTAINER_MIN) {
+        validationError(EOFError.ContainerSectionSize)
+      }
+      if (containerSectionSize > CONTAINER_MAX) {
+        validationError(EOFError.ContainerSectionSize)
+      }
```
acolytec3, line 203: As above, consolidate since error is the same
```suggestion
      if ((containerSectionSize < CONTAINER_MIN) || (containerSectionSize > CONTAINER_MAX)) {
        validationError(EOFError.ContainerSectionSize)
      }
```

packages/evm/src/eof/container.ts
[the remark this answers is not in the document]
acolytec3: Can we go ahead and remove this (and the assignment for it above)?

packages/evm/src/eof/container.ts
```diff
@@ -0,0 +1,461 @@
...
+    for (const [i, codeSize] of header.codeSizes.entries()) {
+      try {
+        const code = stream.readBytes(codeSize)
+        codes.push(code)
+      } catch {
+        validationError(EOFError.CodeSection, i)
+      }
+    }
+    // Write the entire code section to the entireCodeSection
+    const entireCodeSection = buf.slice(codeStartPtr, stream.getPtr())
```
acolytec3, line 337: As below, can we remove this or do we still need it?

packages/evm/src/eof/container.ts
```diff
@@ -0,0 +1,461 @@
...
+
+    // Write all data to the object
+    this.typeSections = typeSections
+    this.codeSections = codes
+    this.containerSections = containers
+    this.entireCode = entireCodeSection
+    this.dataSection = dataSection
+    this.buffer = buf
+  }
+  sections() {
```
acolytec3, line 381: Can you add a little bit more documentation explaining what this returns?  I think I get it but still hazy on the `typeSections`

packages/evm/src/eof/container.ts
```diff
@@ -0,0 +1,461 @@
...
+  size() {
+    return {
+      typeSize: this.typeSections.length,
+      codeSize: this.codeSections.length,
+      dataSize: this.dataSection.length,
+    }
+  }
+  sectionSizes() {
+    return [
+      this.typeSections.map(() => 4),
```
acolytec3, line 393: Feels like a magic number here so would be nice to either add the below comment (if correct) or replace `4` with a constant that's declared elsewhere (in `constants.ts`?)
```suggestion
      // A type section size is always 4 bytes long
      this.typeSections.map(() => 4),
```

acolytec3 (commented):
Have added more comments on `container.ts`.  Mostly nits and organizational stuff but looks good so far and code generally quite clear.

packages/evm/src/eof/verify.ts
```diff
@@ -0,0 +1,519 @@
...
+  function addIntermediate(location: number) {
+    if (jumpLocations.has(location)) {
+      // When trying to add an intermediate to a location already JUMPed to: this is invalid
+      validationError(EOFError.InvalidRJUMP)
+    }
+    intermediateBytes.add(location)
+  }
+
+  // TODO (?) -> stackDelta currently only has active EOF opcodes, can use it directly (?)
+  // (so no need to generate the valid opcodeNumbers)
```
acolytec3, line 89: Would second this thought if possible.  This feels like redoing the same thing everytime we are doing container validation.

packages/evm/src/eof/verify.ts
```diff
@@ -0,0 +1,519 @@
...
+  // Add all reachable code sections
+  const reachableSections: { [key: number]: Set<number> } = {}
+
+  let codeSection = -1
+  for (const code of container.body.codeSections) {
+    codeSection++
+
+    reachableSections[codeSection] = new Set()
+
+    const returningFunction = container.body.typeSections[codeSection].outputs === 0x80
```
acolytec3, line 162: The naming of this variable is opposite of what is actually contained.  Only non-returning functions have an outputs section of 0x80

```suggestion
    // 0x80 indicates that this code section is a function that does not return a value
    const nonReturningFunction = container.body.typeSections[codeSection].outputs === 0x80
```

packages/evm/src/eof/verify.ts
acolytec3: Does this check cover both overflow for `CALLF` and `JUMPF`?

packages/evm/src/eof/verify.ts
```diff
@@ -0,0 +1,519 @@
...
+
+      let minStackNext = minStackCurrent + delta
+      let maxStackNext = maxStackCurrent + delta
+
+      if (maxStackNext > 1023) {
+        // TODO verify if 1023 or 1024 is the right constant
+        validationError(EOFError.StackOverflow)
+      }
+
+      if (returningFunction && opcode === 0xe4) {
```
acolytec3, line 220: Per above, the naming of this variable `returningFunction` is opposite of what is actually checked in the code (i.e. should be "nonReturningFunction")

packages/evm/src/eof/verify.ts
```diff
@@ -0,0 +1,519 @@
...
+        const tableSize = code[ptr + 1] + 1
+
+        if (tableSize === undefined) {
+          validationError(EOFError.OpcodeIntermediatesOOB)
+        } else if (tableSize === 0) {
+          validationError(EOFError.RJUMPVTableSize0)
+        }
+
+        if (ptr + tableSize * 2 + 2 >= code.length) {
+          // Fall-through case
```
acolytec3, line 262: Can we explain what the fall-through case is?  If I understand the spec correctly, this is where the `case` in the stack, is greater than the highest possible index to jump to so it just skips over the intermediates to the next opcode, right?

packages/evm/src/eof/verify.ts
```diff
@@ -0,0 +1,519 @@
...
+          const targetNonReturning = targetOutputs === 0x80
+
+          if (targetOutputs > currentOutputs && !targetNonReturning) {
+            // Spec rule:
+            // JUMPF operand must point to a code section with equal or fewer number of outputs as
+            // the section in which it resides, or to a section with 0x80 as outputs (non-returning)
+            validationError(EOFError.InvalidJUMPF)
+          }
+
+          if (returningFunction && targetOutputs <= 0x7f) {
```
acolytec3, line 331: Is this invalid?  As I've noted before, I'm pretty sure that `returningFunction` is actually "false" if the function returns anything so not sure if this check is correct or not.

packages/evm/src/eof/verify.ts
```diff
@@ -0,0 +1,519 @@
...
+            // Current function is returning, but target is not, cannot jump into this
+            validationError(EOFError.InvalidReturningSection)
+          }
+
+          if (targetNonReturning) {
+            // Target is returning
```
acolytec3, line 337: I'm confused by this comment.  If a code section has 0x80 as the outputs, doesn't that mean it's **not returning**?

packages/evm/src/eof/verify.ts
```diff
@@ -0,0 +1,519 @@
...
+            const expectedStack = currentOutputs + targetInputs - targetOutputs
+            if (!(minStackCurrent === maxStackCurrent && maxStackCurrent === expectedStack)) {
+              validationError(EOFError.InvalidStackHeight)
+            }
+          }
+          if (
+            maxStackCurrent + container.body.typeSections[target].maxStackHeight - targetInputs >
+            1024
+          ) {
+            //console.log(maxStackCurrent, targetOutputs, targetInputs, targetNonReturning)
```
acolytec3, line 352: Can we remove this and any other commented out console logs?

packages/evm/src/eof/verify.ts
```diff
@@ -0,0 +1,519 @@
...
+      } else if (opcode === 0xe6) {
+        // DUPN
+        const toDup = code[ptr + 1]
+        if (toDup + 1 > minStackCurrent) {
+          validationError(EOFError.StackUnderflow)
+        }
+      } else if (opcode === 0xe7) {
+        // SWAPN
+        const toSwap = code[ptr + 1]
+        // TODO: EVMONEs test wants this to be `toSwap + 2`, but that seems to be incorrect
```
acolytec3, line 372: Do we keep the stack 1 based or zero based in our code?  That would inform this I think?  Guessing maybe EVMOne is 0 based?

packages/evm/src/eof/verify.ts
```diff
@@ -0,0 +1,519 @@
...
+        reachableOpcodes.add(ptr)
+
+        // Add next opcode to successorSet
+        // NOTE: these are all opcodes except RJUMP
+        if (opcode !== 0xe0) {
+          successorSet.add(ptr)
+        }
+      }
+
+      // TODO here validate stack / reachability and stack overflow check
```
acolytec3, line 454: Is this still a TODO?  Feels like we've covered it here.

acolytec3 (commented):
Have finished a first review of `verify.ts`

packages/evm/src/eof/errors.ts
```diff
@@ -0,0 +1,254 @@
...
+  MinCodeSections = `should have at least 1 code section`,
+  MaxCodeSections = `can have at most 1024 code sections`,
+  CodeSection = `expected a code section`,
+  DataSection = `Expected data section`,
+
+  // Container section
+  ContainerSection = 'expected a container section',
+  ContainerSectionMin = 'container section should be at least 1 byte',
+  InvalidEOFCreateTarget = 'EOFCREATE targets an undefined container',
+  InvalidRETURNContractTarget = 'RETURNCONTRACT targets an undefined container',
```
acolytec3, line 47: This makes more sense if we make the full RETURNCONTRACT opcode name capitalized.

```suggestion
  InvalidRETURNCONTRACTTarget = 'RETURNCONTRACT targets an undefined container',
```

acolytec3, line 47: This makes more sense if we make the full RETURNCONTRACT opcode name capitalized.

```suggestion
  InvalidRETURNCONTRACTTarget = 'RETURNCONTRACT targets an undefined container',
```

packages/evm/src/eof/verify.ts
jochem-brouwer: This cannot be done at the verification level and has to be done at runtime. This is done (see their definitions in `functions.ts`)

packages/evm/src/eof/verify.ts
jochem-brouwer: Ah right, I should fix this. The problem is that these 0x80 functions usually call a RETURN opcode :sweat_smile: Which makes it confusing (but it is non-returning in the sense of: it does not return to whatever CALLFd it)

packages/evm/src/eof/verify.ts
jochem-brouwer: No, the problem here is that RJUMPV is either the final opcode (this is not allowed) or, if it would "fall through" (the `case` on stack is not within the table indices) the program counter would end up out-of-bounds i.e. outside the code. Sorry, this comment is confusing.

General note: in `verify.ts` we do not have access to any stack, so we cannot reason about it, and thus have to consider all values

packages/evm/src/eof/verify.ts
jochem-brouwer: Sorry the terminology is confusing. We are currently in a code section which has 0x80 outputs (so it is "non returning" by the EOF spec). However, the target of the JUMPF is now returning. This means that if you CALLF a 0x80 code section, and this JUMPFs into this <= 0x7F (returning code section), then thus CALLF did not call into a non-returning function, but it actually called into a function, and therefore this invalidates the container.

packages/evm/src/eof/verify.ts
jochem-brouwer: Definitely!

packages/evm/src/eof/verify.ts
jochem-brouwer: We want to keep the stack as-is (I think we are zero-based). There were many off-by-one bugs which I have to fix (those are often also confusing). However I think this comment is a reply on one of EVMONEs internal tests, which seemed wrong to me (it tried to swap the wrong values, at least that's what I thought)! (I will have to check back later on this)

packages/evm/src/eof/verify.ts
```diff
@@ -0,0 +1,519 @@
...
+          validationError(EOFError.InvalidRETURNContractTarget)
+        }
+        if (containerTypeMap.has(target)) {
+          if (containerTypeMap.get(target) !== ContainerSectionType.DeploymentCode) {
+            validationError(EOFError.ContainerDoubleType)
+          }
+        }
+        containerTypeMap.set(target, ContainerSectionType.DeploymentCode)
+      } else if (opcode === 0xd1) {
+        // DATALOADN
```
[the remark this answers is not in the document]
jochem-brouwer: We cannot do this, in this context we do not have access to the stack. The value on the stack could be anything. We thus have to verify this at runtime.

Side-note: this could be made more clear in the EIP itself, the difference between container validation and runtime validation!

packages/evm/src/eof/verify.ts
jochem-brouwer: Yes I think you are right!

holgerd77 (approved):
I have gotten the ok from Jochem to take this in now, I do think this makes sense since Jochem is away and all we can do here is cause new dramatic merge conflicts with our breaking-change hacking. 😂 

So I will merge now. We and/or Jochem can address remaining issues in follow-up PRs. We should generally also go back to this original PR here and continue to review over time.

PR #3441, opened 2024-05-30
verkle: Add tests for verkle bytes helper

scorbajio:
This change helps increase test coverage of the `verkle` package bytes helper by adding some unit tests.

### DIFF
--- a/packages/verkle/src/util/bytes.ts
+++ b/packages/verkle/src/util/bytes.ts
@@ -8,15 +8,29 @@
  */
 export function matchingBytesLength(bytes1: Uint8Array, bytes2: Uint8Array): number {
   let count = 0
-
-  // The minimum length of both arrays
   const minLength = Math.min(bytes1.length, bytes2.length)
 
-  for (let i = 0; i < minLength; i++) {
+  // Unroll the loop for better performance
+  for (let i = 0; i < minLength - 3; i += 4) {
+    // Compare 4 bytes at a time
+    if (
+      bytes1[i] === bytes2[i] &&
+      bytes1[i + 1] === bytes2[i + 1] &&
+      bytes1[i + 2] === bytes2[i + 2] &&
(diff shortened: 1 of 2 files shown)


holgerd77:
I wonder if this method is rather something for our Util package or is even already present there in some form? 🤔

scorbajio:
I was also able to optimize it a bit:
```
Current Approach x 630 ops/sec ±1.08% (93 runs sampled)
Optimized Approach x 1,059 ops/sec ±4.31% (93 runs sampled)
Fastest is Optimized Approach
```
Benchmark file: [bench.ts.tar.gz](https://github.com/user-attachments/files/15507336/bench.ts.tar.gz)

holgerd77:
@acolytec3 sure, then let's keep! 👍

packages/verkle/src/util/bytes.ts
```diff
@@ -8,15 +8,29 @@
  */
 export function matchingBytesLength(bytes1: Uint8Array, bytes2: Uint8Array): number {
   let count = 0
-
-  // The minimum length of both arrays
   const minLength = Math.min(bytes1.length, bytes2.length)
 
-  for (let i = 0; i < minLength; i++) {
+  // Unroll the loop for better performance
```
acolytec3, line 13: Did you do performance testing to show that this is faster?  It's not obvious to me why this is any faster than iterating element by element.

scorbajio: Yes the benchmark file is in the above comment: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3441#issuecomment-2141083701. Since multiple bytes are being compared at a time, the loop runs less times and handles the remaining comparisons outside of the first loop.

acolytec3: Ah, sorry, didn't see this.  Thanks for the comparison!

acolytec3 (approved):
LGTM!

PR #3445, opened 2024-06-06
feat: Make a POC of pureInterpreter

roninjin10:
This is just me hacking around to see what it would take to get a pure interpreter that executes bytecode with no state or can be injected with any state

I think we can make this way cleaner just trying stuff out atm


roninjin10:
Generally don't like this compared to making the EVM tree shake better https://github.com/ethereumjs/ethereumjs-monorepo/issues/3446

PR #3447, opened 2024-06-07
util,vm,evm: update the 2935 latest bytecode in test and fix address conversion issues 

g11tech:
two things:

1. update the latest bytecode as per fixes in https://github.com/ethereum/EIPs/pull/8596
2. fix the evm/vm code where bigint to address conversion was breaking for the addresses with leading bytes as 0


packages/util/src/bytes.ts
```diff
@@ -430,6 +430,10 @@ export const bigIntToUnpaddedBytes = (value: bigint): Uint8Array => {
   return unpadBytes(bigIntToBytes(value))
 }
 
+export const bigIntToAddressBytes = (value: bigint): Uint8Array => {
+  return setLengthLeft(bigIntToBytes(value), 20).slice(-20)
+}
```
holgerd77, line 435: Seems to me such a new method might be reasonable and justified. If we want to explicitly add we would need 2-3 explicit test cases though. Also I guess this should (?) throw if the big int value exceeds the 20 bytes.

holgerd77 (commented):
One thing for clarification respectively to address around the new util method.

packages/util/src/bytes.ts
holgerd77: (less-work-alternative: do not add new method but use code from method directly)

g11tech: good idea to add some test cases :+1: .

not so sure about throwing because we were already doing slice(-20) in the evm code where i also replaced this method. I mean would we error in the EVM execution if the stack input of the address would > 20 bytes?

holgerd77: I guess that's a question for @jochem-brouwer 🙂

jochem-brouwer: Good point, I actually think that EVM does **not** throw here, especially because it will throw in the new EOF cases in case that you put more than 20 bytes on the stack. (So, in current CALLs we slice the stack with 20 bytes, but in EOF if the stack has more than 20 bytes it throws).

This should likely fail some ethereum/tests cases in case where we start throwing on these CALL situations.

g11tech: we can add an extra flag : `strict: boolean` , if passed true (for e.g. EOF activated), then we can throw else do the current behavior

holgerd77: Can we do it other way around and have the strict case as the default and so set the flag to true and only deviate (to the old behavior) if false is passed? (so this obviously would only work for the new method due to backwards compatibility needs of existing Util methods)

g11tech: sounds good, will impl that way :+1:

g11tech: modified :+1:

holgerd77 (approved):
LGTM

Issue #3450, opened 2024-06-11
Client: Reduce Simple Dependencies


holgerd77:
Our client is up for some more serious production use cases with the rise of a stateless Ethereum world on the horizon and the need for lighter (browser) client (modules) coming along!

This is a warm-up task regarding bringing client dependencies to a bearable level, which is a necessary pre-requisite if we want the client being deployed and used in mission-critical environments!

There are likely various dependencies which can be easily removed and either replaced by some own code or "internalized" as we did in the past for other small libraries where the license permits it (e.g. MIT), e.g. `crc32` in Common, see `crt.ts`.

On a rough first look these dependencies might be candidates (but definitely needs a second look/judgement, very quickly writing this together):

- bodyparser
- connect
- cors
- it-pipe
- jwt-simple
- qheap

(note that - for sure - yargs and winston *will* be candidates too to be handled, but these are more challening (might be solutions like the ability to replace the logger or something), and should not be part of this "warm-up session" 🙂 )

(another note: some dependencies (like level will likely/eventually also "fall away" coming with some client refactoring/modularization (so where e.g. a stateless client might not need a full level DB. But also something for later and more strucutred thoughts preparing for this 🙂 )

holgerd77:
Ah, also to note: please rather to be tackled "one by one" (so: one dependency, one PR, rare exceptions on this)

foufrix:
Hey, @holgerd77, I was taking a quick look at this. Most of the time, it depends on external packages as well. How do you want to proceed:  start internalizing those and add new external or everything in one PR? But I suppose some have a lot of nested and sub-nested packages.

Example : 

- `connect` [use 4 other externals](https://github.com/senchalabs/connect/blob/0a71c6b139b4c0b7d34c0f3fca32490595ecfd60/package.json#L19-L24)

- `body-parser` [use 10 externals](https://github.com/expressjs/body-parser/blob/998b265db57a80ae75ea51c55f6a191e2d168a60/package.json#L11-L22)

holgerd77:
Hey @foufrix, cool that you were having a look! 😄 Just to clarify: this was just a *very* quick write-up, I didn't have a closer look if removing the specifically cited dependencies is *really* an option respectively how we use these in the client code.

See the dependencies references rather as some "examples", this would still need a closer look one-by-one before doing anything (and it might very well turn out that the first look was too superficial and removing a certain dependency is not realistic).

[...]

PR #3451, opened 2024-06-12
Experiment with internalizing QHeap dependency

scorbajio:
This pull request looks into the feasibility of replacing the `qheap` dependency with a local implementation. Also see #3450.


holgerd77:
Hi there, thanks for picking this up! 👍 Just to clarify on how I used some terminology, so I used "internalize" to refer to "just copying over the code" and add the license (if the license permits, which - just confirmed for myself - is the case for the Apache license being used for the `qheap` package). As an example I referred to the [crc.ts](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/crc.ts) where we just did that, see also the comment header.

For `qheap` this is definitely also a viable option I guess, this is just a single file [qheap.js](https://github.com/andrasq/node-qheap/blob/master/qheap.js) which we can copy over, add the license, and also add some typing (judging from former experiences with other internalizations, this is amittedly the most tricky parts and this leaves *some* room for errors).

Yet another option might be to *first* look for a more modern but still simple, dependency-free and performant enough TypeScript alternative and then copy over the code rather from there. Not sure, if there is one this will likely be better, since - again - adding the typing is a bit hairy.

holgerd77:
👍 

One thing I stumbled upon when coming back here along reviewing the `cors` internalization suggestion from @foufrix from here https://github.com/ethereumjs/ethereumjs-monorepo/issues/3450#issuecomment-2163107743 : we should not put the internalized code "somewhere" in the code but have a central and clearly marked place for this.

Can you please move over to a dedicated `ext` folder as we have done e.g. in `devp2p` [here](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/devp2p/src) and place the `qheap.ts` file there together with an additional `index.ts` folder file? That has become somewhat our "standard" for these integrations. Thanks! 🙏

holgerd77 (approved):
🎉

PR #3456, opened 2024-06-16
Internalize jwt-simple dependency

scorbajio:
This pull request looks into internalizing the `jwt-simple` dependency in the `client` package. Also see https://github.com/ethereumjs/ethereumjs-monorepo/issues/3450.

### DIFF
--- a/packages/client/src/ext/index.ts
+++ b/packages/client/src/ext/index.ts
@@ -1,3 +1,4 @@
 'use strict'
 
+export * from './jwt-simple.js'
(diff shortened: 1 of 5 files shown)


holgerd77:
Oh, this is double/parallel work right now with @foufrix also started on this on #3458 

grafik

I guess @foufrix was first to "officially announce" 😋, @scorbajio is it okay if we close here and you work together with @foufrix to bring this over the finish line?

PR #3458, opened 2024-06-17
internalization of jwt-simple

foufrix:
Related to #3450

I'm almost done, but I need some guidance on how to migrate from `Buffer.from` to `Uint8Array` on those 2 functions to follow Ethereum linting rules.

`No Buffers - use Uint8Array insteadeslint(ethereumjs/noBuffer)`

```ts
function base64urlDecode(str: string) {
  // eslint-disable-next-line ethereumjs/noBuffer
  return Buffer.from(base64urlUnescape(str), 'base64').toString()
}

function base64urlEncode(str: string) {
  // eslint-disable-next-line ethereumjs/noBuffer
  return base64urlEscape(Buffer.from(str).toString('base64'))
}
```

I've refactored a lot of parts of base jwt-simple to follow all linting rules and changed the way it checked if the algorithm is supported to fit TS way better: 

```ts
function assertAlgorithm(alg: any): asserts alg is Algorithm {
  if (!['HS256', 'HS384', 'HS512', 'RS256'].includes(alg)) {
    throw new Error('Algorithm not supported')
  }
}
```


foufrix:
@holgerd77 I've tried reaching you trough warpcast, not sure you see your message there, do you have some guidance on how to migrate the function I've sent before from `Buffer.from` to `Uint8Array ` to follow the repo linting rules?

Thanks 🙏

holgerd77:
Hi there,
cool that you already opened a PR 🙂, so it would be definitely good here if we get `Buffer` out here, this was one of our central refactoring goals some time ago already, since `Buffer` is Node.js only and we would want to get more Node.js independent (mid-term also fro client).

I am not fully sure about the conversion from the `jwt-simple` package, so from [this](https://github.com/hokaccha/node-jwt-simple/blob/master/lib/jwt.js#L194) line:

```ts
return Buffer.from(base64urlUnescape(str), 'base64').toString();
```

I googled a bit and I would cautiously think that using this [atob](https://developer.mozilla.org/en-US/docs/Web/API/Window/atob) method here should work and should be ok, this seems to be available both in Node.js and the browser.

Hmm, I am not fully sure though, I just tested both versions like this and this does not seem to be fully doing the same, or I am doing something wrong:

grafik

holgerd77:
@foufrix we talked this through in the team: eventually it is also an option to use the `scure/base` package here since we already have this in `devp2p` and also somewhere else, so this would be no extra dependency even if we add to the client. Just make sure then you use the same version as [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/devp2p/package.json#L63).

acolytec3:
> @foufrix we talked this through in the team: eventually it is also an option to use the `scure/base` package here since we already have this in `devp2p` and also somewhere else, so this would be no extra dependency even if we add to the client. Just make sure then you use the same version as [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/devp2p/package.json#L63).

The issue with `btoa` and `atob` is that they don't exist natively in NodeJS but only in the browser.  Since we don't use `buffer` in our libraries anymore, we need to go with the `scure/base` dep (my preference) or else implement an internal base64 decoder.
image

In using `scure/base`, this should allow you to remove the Buffer references and just do simply:
```js
const data = Uint8Array.from([1, 2, 3]);
base64.decode(base64.encode(data));
```
use the `scure/base` API.

foufrix:
Thanks for the quick answer, @holgerd77 @acolytec3 
I've migrated the test file from jwt-simple repo to Ethereum using vitest. I still encounter an issue with the decode, trying to dig into what I'm doing wrong (I've pushed the file, and the part I added that does not work has been commented out). If you have an idea, that would be helpful. `jwt-simple.ts->base64urlDecode()`
Encoding works well!

acolytec3:
I'm not seeing any test failures locally.  Will see how CI does.

packages/client/src/ext/jwt-simple.ts
```diff
@@ -0,0 +1,213 @@
...
+  HS256: 'hmac',
+  HS384: 'hmac',
+  HS512: 'hmac',
+  RS256: 'sign',
+}
+
+/**
+ * expose object
+ */
+export const jwt = module.exports
```
acolytec3, line 40: We can't use `module.exports` here since we have an ESM build and don't use bundlers to transpile CJS code.

You can replace this with something like below where each property is the necessary encode/decode function used within our code:
```js
export const jwt = {
  decode: decode,
  encode: encode
}
```

Also ideally move this to the end of the file so the logic makes sense of instantiating the individual functions and then exporting several of them as the named export object `jwt`.  Let me know if that doesn't make sense.

packages/client/src/ext/jwt-simple.ts
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * version
+ */
+jwt.version = '0.5.6'
```
acolytec3, line 45: We don't need this anymore since it's going to be part of our code base now.

packages/client/src/ext/jwt-simple.ts
```diff
@@ -0,0 +1,213 @@
+/*
+ * jwt-simple
+ *
+ * JSON Web Token encode and decode module for node.js
+ *
+ * Copyright(c) 2011 Kazuhito Hokamura
```
acolytec3, line 6: Can you replace these references with something like I did for the `asyncEventEmitter` code [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/src/asyncEventEmitter.ts)

packages/client/src/ext/jwt-simple.ts
acolytec3: Follow-up, it looks like the only code we use from the JWT API is the [`decode` function](https://github.com/ethereumjs/ethereumjs-monorepo/blob/52c31c0451e150e3ef773b8de468114cf18900b4/packages/client/src/util/rpc.ts#L16) so you can probably just export that one function (though I guess also export `encode`) and be done with it.

foufrix: we also use `base64urlDecode` in the test file `jwt-simple.spec.ts`

foufrix: done

packages/client/src/ext/jwt-simple.ts
foufrix: done

packages/client/src/ext/jwt-simple.ts
foufrix: done

foufrix:
> I'm not seeing any test failures locally. Will see how CI does.

Currently, I've commented out what is not working and let in place `Buffer.from` in `base64urlDecode()`
I still have trouble understanding what the issue is. I have some padding length error, I'm not 100% familiar with encoding decoding
I've seen this in `scure-base` about [padding](https://github.com/paulmillr/scure-base/issues/4).

image
 
seems to be related

acolytec3:
Will take a look.

acolytec3 (approved):
LGTM pending CI passing

foufrix:
@acolytec3 Thanks 🙏 

I was not properly converting it to Uint8Arr <-> utf8 then.

I will be careful about using `bytesToUtf8` and `utf8ToBytes` next time!

acolytec3:
Base64 is tricky and `@scure/base` is more intended for bytes <-> string rather than handling string -> differently formatted string stuff done in this library.  Thanks for your contribution!

holgerd77:
Thanks a lot to both of you for bringing this over the finish line! ❤️🙂

Sorry, this was substantially more tedious than I initially anticipated. 😬

Issue #3446, opened 2024-06-07
Improve tree shaking for smaller bundle sizes


roninjin10:
To make Ethereumjs better for frontend, it's important we maximize code splitting. 

### What is tree shaking?

Tree shaking is when a bundler such as vite or esbuild are able to detect and delete unused code. 

```typescript
export function foo() {}
export function bar() {}
export function baz() {}

if (process.env.NODE_ENV !== 'production') {
   console.log('dev mode')
}
```

If the above code is a library and we only import `baz`, a bundler will be smart enough to delete foo and bar as unused. It will also be able to detect `production` !== `production` and delete that if block.

### When can a bundler tree shake

A bundler can only tree shake when it can safely and reliably know code is unused. The following cannot be tree shaken notably

- static methods
- methods on a class
- commonjs (e.g. code using `require` and `module.exports` rather than esm)

All types treeshake so you can use a type (for example if a class is dependency injected) as long as you don't provide a default.

# Package analysis

We will go package by package analyzing tree shaking opportunities. The tool used to do this is [bundle.js](https://bundlejs.com/?q=%40ethereumjs%2Fcommon%404.3.0&treeshake=%5B*%5D)

## `@ethereumjs/common`

| Import    | Gzip size |
| -------- | ------- |
| import * from '@ethereumjs/common'  | 30.2kb    |
| import {Common} from '@ethereumjs/common' | 30.2 kB     |

Doesn't tree shake at all as we can tell from the specific import being same size as entire package

### Unnecessary static methods

Problem: All static methods on the Common class cannot tree shake
Solution: Move static methods to exports

This cannot tree shake even if custom is completely unused

```typescript
class Common {
  static custom() {}
}
```

This would tree shake

```typescript
// option 1 export a method not attatched to class
export const createCustomCommon() {}

// option 2 move it to a new class
class CustomCommon {
  static custom() {}
}
```

### Class methods cannot treeshake

Problem: All class methods cannot be tree shaken
Solution: Consider moving class methods to a util function or seperate class. Especially if it's niche or unlikely to be used
Recommendation: I personally don't see problematic methods so I would skip this one

```typescript
// this is what a maximally tree shakeable library like Viem or Tevm would do
// instead of class method seperate data and methods and pass data in as first argument
export const consensusAlgorithm = (common: Common): string | ConsensusAlgorithm {}
```

### Individual chains and hardforks are not tree shakeable

[...]

PR #3462, opened 2024-06-18
EVM/Monorepo: Verkle Decoupling

holgerd77:
This PR works towards reducing the EVM bundle size by reducing the Verkle coupling.

The one direct casting to `StatelessVerkleStateManager` is removed by using the `StateManager` interface here by adding the Verkle methods as optional methods there. I would generally like to expand on this since it reduces the direct coupling to the SVSM class.

Additionally an interface `AccessWitness` is added (also to `Common`), which we can use instead of the direct import.

And, third measure: constants and methods not depending on Verkle cryptography have been moved to a dedicated `verkle` module in `Util`. This should make sense as well and was necessary in the EVM case, since it allowed to remove the `@ethereumjs/verkle` dependency completely.

---

I would think the above measures should "do the job". I could not yet prove it though. When running the `esbuild` command from https://github.com/ethereumjs/ethereumjs-monorepo/issues/3446#issuecomment-2172947170 build size stayed unexpectedly on the 2.6 MB level (for EVM).

Some googling brought up that `esbuild` just seems to not properly tree shake on unused imports from external modules, see e.g. https://github.com/evanw/esbuild/issues/1794. 🤔

I discovered the `--analyze=verbose` flag along (full command: `npm run build && esbuild --bundle 10EVM.mjs --outfile=out1.mjs --analyze=verbose`) - generally useful - which helps to see where things are still drawn in:

```shell
   ├ ../verkle/dist/esm/node/types.js ────────────────────────────────────────────────────── 226b ──── 0.0%
   │  └ ../verkle/dist/esm/node/index.js
   │     └ ../verkle/dist/esm/index.js
   │        └ ../statemanager/dist/esm/statelessVerkleStateManager.js
   │           └ ../statemanager/dist/esm/index.js
   │              └ dist/esm/evm.js
   │                 └ dist/esm/index.js
   │                    └ 10EVM.mjs
```

So this is for the `@ethereumjs/verkle` dependency, which should not be there anymore, and which brought me to googling for the `esbuild` issue.

I then started thinking about using another bundler (so: if `esbuild` might generally not deliver valid tree shaking results) trying `rollup` with the following:

```shell
npm i -g rollup
npm i -g @rollup/plugin-node-resolve
rollup --format=cjs --file=out1.mjs --plugin=node-resolve -- 10EVM.mjs
```

This gives a bunch of other warning and (at least) one error (where things break).

grafik

Not fully sure if it's worth to resume on that path (or try e.g. another bundler).

Maybe our tree shaking & bundling expert @roninjin10 can have a comment! 🙂 
rollup --format=cjs --file=out1.mjs --plugin=node-resolve -- 10EVM.mjs


packages/evm/src/interpreter.ts
```diff
@@ -272,9 +271,7 @@ export class Interpreter {
         const contract = this._runState.interpreter.getAddress()
 
         if (
-          !(await (
-            this._runState.stateManager as StatelessVerkleStateManager
```
roninjin10, line 276: This generally won't affect tree shaking. TypeScript types are always removed even if they are values or classes as long as they are used as a type as long as we use `import type` syntax

roninjin10:
@holgerd77 wierd rollup didn't work but since you are already using vitest using vite is likely the easiest. I added a bundle analyzer setup you can use in this pr https://github.com/ethereumjs/ethereumjs-monorepo/pull/3463

g11tech (commented):
refactor looks superclean and modular. don't have insight into tree shaking however as of now, so for that will delegate to experts :)

holgerd77:
> refactor looks superclean and modular. don't have insight into tree shaking however as of now, so for that will delegate to experts :)

This is still in the pre-tree-shaking era, so just removing dependencies so that they are not there. 😂

packages/util/src/verkle.ts
```diff
@@ -0,0 +1,228 @@
...
+
+/**
+ * @dev Returns the tree key for a given verkle tree stem, and sub index.
+ * @dev Assumes that the verkle node width = 256
+ * @param stem The 31-bytes verkle tree stem as a Uint8Array.
+ * @param subIndex The sub index of the tree to generate the key for as a Uint8Array.
+ * @return The tree key as a Uint8Array.
+ */
+
+export const getKey = (stem: Uint8Array, leaf: LeafType | Uint8Array) => {
```
gabrocheleau, line 161: Minor note on a few of these helpers. They were initially named quite minimally `getKey` rather than `getVerkleTreekey` because the "verkle" was obvious from it being imported from the verkle package. I feel like we should likely rename these helpers more explicitly as "verkle" since they are now exported in utils (`getKey` could be so many things). We could perhaps do the same with `VERSION_LEAF_KEY` => `VERKLE_VERSION_LEAF_KEY`

acolytec3 (commented):
> Minor note on a few of these helpers. They were initially named quite minimally `getKey` rather than `getVerkleTreekey` because the "verkle" was obvious from it being imported from the verkle package. I feel like we should likely rename these helpers more explicitly as "verkle" since they are now exported in utils (`getKey` could be so many things). We could perhaps do the same with `VERSION_LEAF_KEY` => `VERKLE_VERSION_LEAF_KEY`  

100% agree.

gabrocheleau (approved):
Looks great! as noted on the call, we will leave the verkle namespacing updates to a future call in order not to interfere with pending verkle work

PR #3463, opened 2024-06-18
feat: Add bundle visualizer

roninjin10:
Adds a recomended setup for measuring bundle size.

- Uses vite because the repo is already using vitest
- Uses `safest` option for the most conservative worst case tree shaking
- Adds a visualizer for visualizing bundle size

Run `npm run visualize:bundle` to vie a visualization of bundle for evm package

To test a specific import modify the index.ts to only export that import

image

### DIFF
--- a/packages/evm/vite.config.ts
+++ b/packages/evm/vite.config.ts
@@ -13,7 +13,7 @@ export default defineConfig({
     lib: {
       entry: 'src/index.ts',
       name: '@ethereumjs/evm',
-      fileName: (format) => `my-library.${format}.js`,
+      fileName: (format) => `ethereumjs-evm-bundle.${format}.js`,
       // only build for es
       formats: ['es'],
     },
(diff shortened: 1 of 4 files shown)


packages/evm/vite.config.ts
```diff
@@ -1,7 +1,21 @@
...
+  plugins: [wasm(), topLevelAwait(), visualizer({ open: true, gzipSize: true })],
+  build: {
+    rollupOptions: {
+      // We choose safest to get worst case values
+      treeshake: 'safest',
+    },
+    lib: {
+      entry: 'src/index.ts',
+      name: '@ethereumjs/evm',
+      fileName: (format) => `my-library.${format}.js`,
```
roninjin10, line 16:
```suggestion
      fileName: (format) => `ethereumjs-evm-bundle.js`,
```

acolytec3 (approved):
This is super neat.

roninjin10:
Btw you can manually play around with tree shaking via changing index.ts to export only what you want to measure and see how big the difference is vs measuring everything

holgerd77:
Ok, yeah, that's super great! Will merge!

PR #3465, opened 2024-06-19
Expand bundle visualizer to monorepo

holgerd77:
I started this work this morning since I would want to expand the new bundle visualizer from @roninjin10 to all monorepo packages.

However when I add a new `vite.config.ts` file the `tsconfig.json` file turns red along and gives out the following error message:

grafik

We should find out what is happening here before we continue here, maybe someone already has some insight.

Maybe this vite config additions directly to the `build` command have side effects (eventually already for EVM), and we would want to make sure that this does not affect other processes.

### DIFF
--- a/packages/block/.eslintignore
+++ b/packages/block/.eslintignore
@@ -1,0 +1,2 @@
+vite.config.ts
+typedoc.cjs
--- a/packages/block/vite.config.ts
+++ b/packages/block/vite.config.ts
@@ -1,0 +1,19 @@
+import { defineConfig } from 'vitest/config'
+import { visualizer } from 'rollup-plugin-visualizer'
+
+export default defineConfig({
+  plugins: [visualizer({ open: true, gzipSize: true })],
+  build: {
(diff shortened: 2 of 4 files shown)


acolytec3:
Line 6-7 already shows up in VScode for me (has been for a while). Not sure about the `node` types issue but the one about module resolution can be fixed by changing our dev build environment from CJS to ESM.

roninjin10:
For the second error that looks like an actual misconfiguration of tsconfig. I think for a library like Ethjs `module` and `moduleResolution` being set to `NodeNext` is the ideal configuration.

The first error message maybe it just wants you to npm install @types/node?

acolytec3:
This might be solved with our recent of config changes. Will take another look

gabrocheleau:
Is this still an issue? We've done some related work in the past year, perhaps the issue has been fixed through other PRs. I am not getting this issue locally, but perhaps you are? Lmk and we can close this accordingly

PR #3467, opened 2024-06-21
switch rpc.ts to `itty-router` instead of `connect` `cors` `body-parser`

foufrix:
Following discussion in #3450

Attempt to move `client/src/util/rpc.ts` to `itty-router` to remove `body-parser` `connect` `cors`

It's not working now, from what I understood, we will need to add an adapter `@whatwg-node/server`, if someone can confirm that @holgerd77 @acolytec3 

Documentation about that can be found [here](https://itty.dev/itty-router/guides/node#runtimes-node-js-itty-router)

Still trying to understand what I'm doing wrong here


acolytec3:
Sorry for not commenting sooner but if it turns it we need that `whatwyg/node` dependency for `itty-router` to work, we're not gaining anything as that brings in another 5-6 dependencies of it's own. I took an initial look and seems like itty-router might not be the right choice after all. I will take another look this week but we might want to look for alternatives that work with less hassle

foufrix:
Hey @acolytec3, sorry for the late reply. I was away.

I'm not 100% sure that we need it, but the adapter seems necessary to use the HTTP `createServer` method. I don't know if we can do without this HTTP `createServer` method that we pass to Node.

acolytec3:
> Hey @acolytec3, sorry for the late reply. I was away.
> 
> I'm not 100% sure that we need it, but the adapter seems necessary to use the HTTP `createServer` method. I don't know if we can do without this HTTP `createServer` method that we pass to Node.

Understood.  If that's the case, I don't think we should go with `itty-router`.  The whole idea was to reduce our dependency subtree for `connect`/`cors`/`body-parser`.  If there isn't something similar that doesn't require shims and additional deps to make it work, we might as well just stick with what we have now or internalize it (though I don't love that idea all that much).

holgerd77:
Needs another approach and no follow-up, will close

PR #3468, opened 2024-06-21
verkle: rename verkle utils and refactor

gabrocheleau:
This PR updates the verkle utils namespace so that they are clearer in the context of being exported from the util package rather than the verkle package. It also ports over some tests from verkle to util.


packages/util/src/verkle.ts
```diff
@@ -26,7 +26,7 @@ import type { PrefixedHexString } from './types.js'
 // (basically everything generic not depending on Verkle cryptography)
 //
 // TODO 2: Basically all these constants and methods need a Verkle or (VERKLE) prefix since
-// atm names are too generic to be grasped in a non-Verkle-limited context (e.g. `getKey()`)
+// atm names are too generic to be grasped in a non-Verkle-limited context (e.g. `getVerkleKey()`)
```
acolytec3, line 29: Can we remove this todo with this PR?

gabrocheleau: indeed, removed!

acolytec3 (approved):

holgerd77:
Cool, thanks for addressing so quickly, great to now have this "round"! 🙂 ❤️ 

I think I would be a fan of also moving `matchingBytesLength()` over to Util `bytes`, even if it might not be totally apparent if there will be other usages. But this is *such* a generic function that it's worth to "offer as a service" (who knows what other users are doing) and then we have this out of Verkle.

At along maybe we can even (already) move the `PrioritizedTaskExecutor` over and also integrate into `Trie` to make the round-trip. Would also make sense and then we can simply kill off the whole `util` folder from `Verkle`.

Issue #3460, opened 2024-06-18
Client Modularization Part I: Client as a Library


holgerd77:
This is a meta issue for a first round of work preparing for a more modular client world as discussed in the roadmap call, with the goal of having one or more client modules with a minimal dependency set which can then be instantiated in different security and application contexts (browser, verkle, eventually Portal,...).

First step here is to not think about modularization/refactoring yet but get to a cleaner separation between the core package as a library and a lightweight "packaging" or instantiation script. We have this separation already with the `bin` folder (instantiation of things like DB, logging, ...) and the `src` folder containing the "library" and at some later stage these things might go into separate packages, `@ethereumjs/client-lib` and `@ethereumjs/client` or `@ethereumjs/client-el` or however we reason about things, no need to decide yet and rather for a later stage.

So this issue is basically to get from 70% to 95% on a clean `bin` and `src` separation, following TODOs (not necessarily exhaustive):

- [ ] Adds things which feel "extra" and draw in extra dependencies as an option (e.g. Prometheus) (so that the dependency can later on be easily removed on a `bin` and `src` package separation
- [ ] Decide what to do with the logger in a library context (how are other clients actually handling this? 🤔 I am pretty sure I stumbled upon examples here that other clients can run in "library mode"?), so what should happen with the logging output, is the current output even a good format or does this needs adoption on frequency, format and so on?
- [ ] The former also leads to the case that the logger urgently needs to be replaceable *generally*, `Winston` draws in an [awful lot of dependencies](https://github.com/winstonjs/winston/blob/master/package.json) and it should minimally be possible to replace by a simple no-dependency logger (have a look at `package.json` linked, basically with `Winston` in we cannot responsibly provide anything other than a research client)
- [ ] There should be a closer look at the code in `bin` - this has substantially grown over the years - with the eye upon the question if some code is a) initialization code (then it can stay) or b) core logic code (then it should move over to `src`)

Guess there might be other things, feel free to add below. 🙂

holgerd77:
Also to add here, regarding the "looking at the `bin` code" TODO: this can actually also encompass to structure the remaining code better, so e.g. to split out some separate files in `bin` where/if it makes sense to make everything more readable and overall better structured.

PR #3469, opened 2024-06-22
client: make prom client optional

gabrocheleau:
This PR addresses a task that was mentioned in #3460, namely making the prometheus dependency optional.


holgerd77 (changes requested):
Ah, very cool! 🤩

For now I would rather like to not make the depedency optional (yet) for comfortability reasons but keep in the normal set of dependencies, can we do that?

This is rather a conceptual thing within the code base so that we are prepared later on to easily remove (If/when we split packages out), and the code changes do count for this here.

Side question out of interest: I found/find this super interesting, that it is possible to type with `promClient` but only have the dependency in the `optionalDependencies` set, since this is at least *touching* one of the core conceptual problems I always have with this that the type *would* be nice to have but the dependency *should* be out.

Is this - to your knowlege - the minimal setup to achieve this? 🤔 Or is there any way to get the dependency out completely but keep the typing??

Or are there other ways to approach this? There is no way (in `package.json`) to "only reference the types" of a dependency, or something like that, right? Or would it be a way to "auto-generate" types for these cases from the original library (to get this done quickly) and include these??

Still one of the biggest structurally unsolved problems for me. 🤓

acolytec3:
> Ah, very cool! 🤩
> 
> For now I would rather like to not make the depedency optional (yet) for comfortability reasons but keep in the normal set of dependencies, can we do that?
> 
> This is rather a conceptual thing within the code base so that we are prepared later on to easily remove (If/when we split packages out), and the code changes do count for this here.
> 
> Side question out of interest: I found/find this super interesting, that it is possible to type with `promClient` but only have the dependency in the `optionalDependencies` set, since this is at least _touching_ one of the core conceptual problems I always have with this that the type _would_ be nice to have but the dependency _should_ be out.
> 
> Is this - to your knowlege - the minimal setup to achieve this? 🤔 Or is there any way to get the dependency out completely but keep the typing??
> 
> Or are there other ways to approach this? There is no way (in `package.json`) to "only reference the types" of a dependency, or something like that, right? Or would it be a way to "auto-generate" types for these cases from the original library (to get this done quickly) and include these??
> 
> Still one of the biggest structurally unsolved problems for me. 🤓

I don't remember exactly how it all worked (and don't have the energy to do the git archaeology this early in the morning) but this is basically what we did with our DB back when we had the specific `level` dep instead of the abstracted class we have now.  We just had an interface defined that imported the `level` types and then just expected an object passed in that conformed to those types.  I think maybe the difference there is that `level` did/(does?) has a specific types package that is separate from the main `level.js` dep that we used.  Not sure if something similar exists for `prometheus` or not.

gabrocheleau:
> Ah, very cool! 🤩
> 
> For now I would rather like to not make the depedency optional (yet) for comfortability reasons but keep in the normal set of dependencies, can we do that?
> 
> This is rather a conceptual thing within the code base so that we are prepared later on to easily remove (If/when we split packages out), and the code changes do count for this here.
> 
> Side question out of interest: I found/find this super interesting, that it is possible to type with `promClient` but only have the dependency in the `optionalDependencies` set, since this is at least _touching_ one of the core conceptual problems I always have with this that the type _would_ be nice to have but the dependency _should_ be out.
> 
> Is this - to your knowlege - the minimal setup to achieve this? 🤔 Or is there any way to get the dependency out completely but keep the typing??
> 
> Or are there other ways to approach this? There is no way (in `package.json`) to "only reference the types" of a dependency, or something like that, right? Or would it be a way to "auto-generate" types for these cases from the original library (to get this done quickly) and include these??
> 
> Still one of the biggest structurally unsolved problems for me. 🤓

We certainly can leave that as a dependency, and eventually decide to make it optional. It does make the conditionality checks irrelevant (since it will always be present), but at least that part of the work will be done. Wouldn't it make sense for this to be done "all at once" though rather than in 2 steps? 

Concerning the optionalDependencies and their related types, I've researched this a bit today and it seems like the proper approach would be:
- Make the actual dep an optional dependency
- Install the types as a devDependency

Otherwise we would have (from what I'm reading) to do some conditional magic for the types as well in case they have not been installed.

holgerd77:
Hmm, but how would this work for TypeScript users of the library with types as a dev dependency? 🤔 Wouldn’t this break their code?

gabrocheleau:
> Hmm, but how would this work for TypeScript users of the library with types as a dev dependency? 🤔 Wouldn’t this break their code?

Not sure I understand what you mean. Aren't types always just used as dev dependency since they're only needed during development and not in the production runtime environment?

holgerd77:
> > Hmm, but how would this work for TypeScript users of the library with types as a dev dependency? 🤔 Wouldn’t this break their code?
> 
> Not sure I understand what you mean. Aren't types always just used as dev dependency since they're only needed during development and not in the production runtime environment?

No, at least as far as my understanding goes (correct me if I have some thought error along the way), types are shipped as these `d.ts` files for people who want to work with a dependency within their own TypeScript dev environments. And these type files reference/import the respective types from the respective libraries we are talking about.

So here is e.g. a similar example from `Blockchain` (since I am working on that right now), where the `_ethash` property is typed with `Ethash`:

grafik

If we would now directly remove the dependency (or make it optional) we will break people's TypeScript code from the default installation, right?

That's why I have e.g. replaced this with a minimal interface here:
https://github.com/ethereumjs/ethereumjs-monorepo/pull/3504/files#diff-0d1fc3cbd7363c27c9d15a33f9271bbf4fa0d4b4671f7f0b4be5faca3241031aR7

But that's also where I am hanging a bit. So this is simple to be done for dependencies with a very low/small surface API but otherwise things get hairy quickly.

Or is the way to go here to include the definitely-typed type package (so here e.g.: for `prom`) to the production dependencies?

gabrocheleau:
> That's why I have e.g. replaced this with a minimal interface here:

I think the example you mention is a different scenario, where we were actually instantiating the class here: `new Ethash(this.blockchain!.db as any)`, hence it was necessary to include the full dependency. Now this has been replaced with a constructor arg, which is typed with the minimal interface. A middle-ground more similar to what I'm suggesting iwth prom could have been to:
- Include the types only as a devDependency
- Add the constructor arg with the `EthHash` type, removing the need to instantiate the actual class 

For something like prometheus, I think the API surface is more substantial, so creating a minimal interface might not be feasible. I'll look into it again.

gabrocheleau:
Closing this as these client-deps cleanups are not our best use of time rn, and the best course of action would likely be to tackle this from scratch (merge conflicts etc.)

PR #3470, opened 2024-06-23
Implement EIP 7702

jochem-brouwer:
Implements 7702 according to https://github.com/ethereum/EIPs/blob/14400434e1199c57d912082127b1d22643788d11/EIPS/eip-7702.md

Note that there are many open PRs in the EIP repo:

https://github.com/ethereum/EIPs/pulls?q=is%3Apr+is%3Aopen+7702

So, the spec is likely to change.

Some other items I found:

- EIP 1559 extends the logic of EIP 2930, but the code seems to be copied
- I had to do the same with EIP 7702
- In a later refactor of tx we should consider making these txs "stackable" (I tried this, but this is somewhat a hassle with the typing).
- So we currently now have a lot of copied code regarding 2930 / 1559 / 7702 which we should address soon(ish)
- The problem is that you cannot do `EOFCodeEIP7702Transaction extends FeeMarketEIP1559Transaction` is that the return types of some methods change, i.e. the new txs now have these mandatory `authorityLists` fields, so we should likely think of something the same as how we have created `BaseTransaction` now (these new txs now also accept a return type, so we can "stack" the tx types)

Still WIP, main things to do:

- [x] Add logic to VM
- [x] Write tests
- [ ] Watch the spec for updates


holgerd77:
So this is touching the whole topic of a fitting structural setup for the tx library which we litterally have for years, so since tx types have been introduced.

I don't think "stacking" or inheritance is a good design choice here, also cc-ing @gabrocheleau here since you had some related exchange in the chat, since this is just tighly bundling functionality together which now "accidentally" matches but will likely bite us in the future.

Tx types are designed to be independent of each other (for the good and the bad), so there is no such thing as "this tx type inherits its properties from tx type x" or similar stated in EIPs, and in the future arbitrary things can (and likely will) change, be it the way access lists are structured or the fee market is organzied (to pick up the 2930/1559 example). If you think this along the lines of a potentially introduced inheritance, things get messy pretty quickly (or: first one is starting with some still-working work-arounds (inherited `oldAccessLists` boolean flags or whatever), and *then* things get messy in round 2 😋).

This actually also goes for the `BaseTransaction`, I was rather always thinking in my head that we should *get rid* of this at some point, since this is also some relatively random mix of "features" for txs grown out of historic reasons, atm mainly the signature scheme, which might also change at some point.

I think what describes a desired design goal best is the concept of "composability", so that one can plug the different "features" togehter, as independently as possible. That's why we came up with this `Capabilities` concept which Gabriel implemented recently, which already matches this well (after exploring (and failing) more sophisticated solutions like multiple inheritance or the like).

When looking at the current code I find this actually not "quite as bad" already. 🙂 (things would be a lot worse without our `Capabilities` stuff)

So: I wouldn't count these one liner integrations like:

```ts
  getMessageToSign(): Uint8Array {
    return EIP2718.serialize(this, this.raw().slice(0, 9))
  }
```

actually as redundancy. This is more "by design" and rather has its beauty by stating explicitly what this respective tx type composes off (takes the base seralization functionality from generic tx types).

What is admittedly a bit suboptimal here is this code doc doubling, maybe we can look into this a bit specifically at some point and find a targeted solution here (not sure, is it possible to "programmatically reference existing code docs" to be included? 🤔).

I generally think to improve here needs specific looks into what is doubled/getting redundant. I think we just have different "more subtle" problem sets still in which needs to be addressed separately and are a bit a left-over (respectively just: weren't addressed yet) along this capability refactor (since this was the big first step) and are just "step 2".

I will try to tear this a bit apart, will see how far I'll come 🙂:

### raw() method

So the `raw()` method is admittedly just a very specifc method which is different for more or less every tx type.

What we could do here is to put the 1559 inner method code over to the 1559 capabilites and call the capability code from the 1559 tx and then also call the 1559 capability code from the 7702 tx and then use `splice()` or so to rebuild the array.

Might be worth it for code redundancy reduction, and might also not have relevant performance implications.

**Note:** I guess these kind of things I write down here are structurally present throughout the wider code base and will likely be able to applied on other tx types as well (if we decide we want that).

### Static constructors

```ts
    if (
      equalsBytes(serialized.subarray(0, 1), txTypeBytes(TransactionType.EOACodeEIP7702)) === false
    ) {
      throw new Error(
        `Invalid serialized tx input: not an EIP-7702 transaction (wrong tx type, expected: ${
          TransactionType.EOACodeEIP7702
        }, received: ${bytesToHex(serialized.subarray(0, 1))}`
      )
    }

    const values = RLP.decode(serialized.subarray(1))

    if (!Array.isArray(values)) {
      throw new Error('Invalid serialized tx input: must be array')
    }
 ```
 
 I think this can go into a `2718` capability helper method taking in the tx type and an EIP string (`EIP-7702`) and we should do that (for all tx classes where applicable).
 
 Guess `fromValuesArray()` is *so* type specific that we want to live with that (?).
 
 ### main construtor
 
 Not sure if it's possible to put the following into a one-liner in the 2930 capabilites, might be worth it:
 
 ```ts
 // Populate the access list fields
    const accessListData = AccessLists.getAccessListData(accessList ?? [])
    this.accessList = accessListData.accessList
    this.AccessListJSON = accessListData.AccessListJSON
    // Verify the access list format.
    AccessLists.verifyAccessList(this.accessList)
```

```ts
this.maxFeePerGas = bytesToBigInt(toBytes(maxFeePerGas))
    this.maxPriorityFeePerGas = bytesToBigInt(toBytes(maxPriorityFeePerGas))

    this._validateCannotExceedMaxInteger({
      maxFeePerGas: this.maxFeePerGas,
      maxPriorityFeePerGas: this.maxPriorityFeePerGas,
    })

    BaseTransaction._validateNotArray(txData)

    if (this.gasLimit * this.maxFeePerGas > MAX_INTEGER) {
      const msg = this._errorMsg('gasLimit * maxFeePerGas cannot exceed MAX_INTEGER (2^256-1)')
      throw new Error(msg)
    }

    if (this.maxFeePerGas < this.maxPriorityFeePerGas) {
      const msg = this._errorMsg(
        'maxFeePerGas cannot be less than maxPriorityFeePerGas (The total must be the larger of the two)'
      )
      throw new Error(msg)
    }
```

I think most of this can be put into a 1559 capability check method? 🤔

### addSignature

I have the impression this is just a bit ugly because there is generally something a bit suboptimally designed around here. Not sure, maybe this can be generally made more elegant throughout the library (might or might not need some breaking changes).

Yeah, so I would say that's already mostly it.

I would agree that if we "fix" the doc redundancy problem (generally), that would be a somewhat big win for the library.

But otherwise this fits well into our current design without the need for very deep reaching refactorings.

packages/tx/src/util.ts
```diff
@@ -116,6 +129,98 @@ export class AccessLists {
...
+      const nonceList = authorizationListItem[2]
+      const yParity = authorizationListItem[3]
+      const r = authorizationListItem[4]
+      const s = authorizationListItem[5]
+      validateNoLeadingZeroes({ address, yParity, r, s })
+      if (address.length !== 20) {
+        throw new Error('Invalid EIP-7702 transaction: address length should be 20 bytes')
+      }
+      if (nonceList.length > 1) {
+        throw new Error('Invalid EIP-7702 transaction: nonce list should consist of at most 1 item')
```
acolytec3, line 211: This is super weird to me.  Why not just use a single integer, 0 or greater?  Also, is it dangerous to let nonce be null?  Does this allow for unbounded use of some code?

packages/tx/src/util.ts
```diff
@@ -116,6 +129,98 @@ export class AccessLists {
...
+    }
+
+    return {
+      AuthorizationListJSON,
+      authorizationList: bufferAuthorizationList,
+    }
+  }
+
+  public static verifyAuthorizationList(authorizationList: AuthorizationListBytes) {
+    for (let key = 0; key < authorizationList.length; key++) {
```
acolytec3, line 199: The EIP says `If any of the above steps fail, immediately stop processing that tuple and continue to the next tuple in the list. It will In the case of multiple tuples for the same authority, set the code specified by the address in the first occurrence.` about validating each tuple in the authorization list.  It doesn't indicate we should throw an error but just proceed to the next tuple.  That's a little obtuse but I'm assuming this means that when we are processing one of these txns in `vm`, we should only set the code for that authority if the tuple is valid.  Is this method intended to serve that purpose or is this just a sanity check?

packages/tx/test/eip7702.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+        {
+          authorizationList: [
+            {
+              chainId: '0x',
+              address: `0x${'20'.repeat(20)}`,
+              nonce: ['0x1'],
+              yParity: '0x1',
+              r: ones32,
+              //@ts-ignore
+              s: undefined,
```
acolytec3, line 86:
```suggestion
              s: undefined as never,
```

Would prefer to cast to `never` over `ts-ignore` but just a nit.

packages/tx/test/eip7702.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+
+    assert.throws(() => {
+      EOACodeEIP7702Transaction.fromTxData(
+        {
+          authorizationList: [
+            {
+              chainId: '0x',
+              address: `0x${'20'.repeat(20)}`,
+              // @ts-ignore
+              nonce: undefined,
```
acolytec3, line 140:
```suggestion
              nonce: undefined as never,
```

packages/tx/src/util.ts
jochem-brouwer: I understand the EIP as: there has to be a list, the length of this list can either be 0 or 1, and it is used to represent a certain nonce which the account should have in order to set the code to it.

packages/tx/src/util.ts
jochem-brouwer: No, it is used to verify that the authority list is properly formatted. It does not check the tuples, this is done in VM. For instance, it checks that the encoded fields are there, and they have no leading zeros in them.

packages/tx/test/eip7702.spec.ts
jochem-brouwer: Aha, does that work? I did not know, will change.

packages/tx/src/eip7702Transaction.ts
```diff
@@ -0,0 +1,397 @@
...
+type TxValuesArray = AllTypesTxValuesArray[TransactionType.EOACodeEIP7702]
+
+/**
+ * Typed transaction with the ability to set codes on EOA accounts
+ *
+ * - TransactionType: 4
+ * - EIP: [EIP-7702](https://github.com/ethereum/EIPs/blob/62419ca3f45375db00b04a368ea37c0bfb05386a/EIPS/eip-7702.md)
+ */
+export class EOACodeEIP7702Transaction extends BaseTransaction<TransactionType.EOACodeEIP7702> {
+  // implements EIP7702CompatibleTx<TransactionType.EOCodeEIP7702>
```
gabrocheleau, line 45:
```suggestion
```

packages/vm/src/runTx.ts
```diff
@@ -619,6 +671,20 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
     }
   }
 
+  /**
+   * Cleanup code of accounts written to in a 7702 transaction
+   */
+
+  for (const str of writtenAddresses) {
+    const address = Address.fromString(str)
+    // Special case: in StateManager if we try to set code to the empty code, we ignore this
```
acolytec3, line 680: Not directly part of this PR but why is this that we don't update the code hash in the state manager if it's `KECCAK_NULL`?

packages/vm/src/runTx.ts
```diff
@@ -440,6 +447,51 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
     fromAccount.balance = BIGINT_0
   }
   await this.evm.journal.putAccount(caller, fromAccount)
+
+  const writtenAddresses = new Set<string>()
+  if (tx.supports(Capability.EIP7702EOACode)) {
+    const authorizationList = (<EIP7702CompatibleTx>tx).authorizationList
```
acolytec3, line 453:
```suggestion
    // Add contract code for authroity tuples provided by EIP 7702 tx
    const authorizationList = (<EIP7702CompatibleTx>tx).authorizationList
```

Would like something like this comment to give some minimal explanation of what's happening here for future devs

packages/vm/src/runTx.ts
```diff
@@ -440,6 +447,51 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
...
+    const MAGIC = new Uint8Array([5])
+    for (let i = 0; i < authorizationList.length; i++) {
+      const data = authorizationList[i]
```
acolytec3, line 456:
```suggestion
      // Authority tuple validation
      const data = authorizationList[i]
```

packages/vm/src/runTx.ts
```diff
@@ -440,6 +447,51 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
...
+      }
+      if (nonceList.length !== 0 && account.nonce !== bytesToBigInt(nonceList[0])) {
+        continue
+      }
+
+      const addressConverted = new Address(address)
+      const addressCode = await this.stateManager.getContractCode(addressConverted)
+      await this.stateManager.putContractCode(authority, addressCode)
+
+      writtenAddresses.add(authority.toString())
```
acolytec3, line 490: Don't we also need to check and see if the address provided in the authority tuple has already been set (or does the above `isContract` check take care of that)?

packages/vm/src/runTx.ts
```diff
@@ -619,6 +671,20 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
...
+   */
+
+  for (const str of writtenAddresses) {
+    const address = Address.fromString(str)
+    // Special case: in StateManager if we try to set code to the empty code, we ignore this
+    // Do not do this and explicitly override
+    // TODO maybe use modifyAccount?
+    const account = (await this.stateManager.getAccount(address)) ?? new Account()
+    account.codeHash = KECCAK256_NULL
+    await this.stateManager.putAccount(address, account)
```
acolytec3, line 685: This should work just fine (`modifyAccountFields` already generates a new account if it doesn't exist in the state).
```suggestion
    await this.stateManager.modifyAccountFields(address, { codeHash: KECCAK256_NULL })
```

packages/vm/src/runTx.ts
jochem-brouwer: I think the reasoning is that if we try to set code to `KECCAK_NULL` this is a no-op, since the code of an account is empty by default. Note that once code is set, the only way to (currently) remove it is to use SELFDESTRUCT which calls another method `deleteAccount`.

However, thus now this invariant is broken by this EIP, since we first set code, and then unset it, and this is not done by SELFDESTRUCT.

packages/vm/src/runTx.ts
jochem-brouwer: Yes, this should be clear from the comment under `account.isContract()` but it is not super clear, have added some extra clarification

packages/vm/src/runTx.ts
jochem-brouwer: I don't understand why I did not do this directly :thinking:

jochem-brouwer: Ah the original reason was that I wanted to `putContractCode` to empty, which is thus ignored. But I just added a test and it still fails :thinking:

packages/vm/src/runTx.ts
```diff
@@ -619,6 +673,19 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
...
   }
 
+  /**
+   * Cleanup code of accounts written to in a 7702 transaction
+   */
+
+  for (const str of writtenAddresses) {
+    const address = Address.fromString(str)
+    await this.stateManager.putContractCode(address, new Uint8Array())
+    // TODO verify if this is necessary
```
acolytec3, line 683: I think we absolutely have to do this.  Otherwise the `codeHash` won't be null and we can/will have other downstream issues if somebody else tries to `getContractCode`, they will get a non `null` and `KECCAK_NULL` value for an account that shouldn't have code.

acolytec3 (dismissed):
LGTM - we should remove the TODO and merge.

packages/vm/src/runTx.ts
jochem-brouwer: In my test cases, if I remove the `modify` stuff, it still passes. But I cannot think of a scenario where it would not pass - but I could imagine if there is a complex scenario then there might be an edge case where we need it :thinking:

packages/tx/src/eip7702Transaction.ts
```diff
@@ -0,0 +1,396 @@
...
+      this.v !== undefined ? bigIntToUnpaddedBytes(this.v) : new Uint8Array(0),
+      this.r !== undefined ? bigIntToUnpaddedBytes(this.r) : new Uint8Array(0),
+      this.s !== undefined ? bigIntToUnpaddedBytes(this.s) : new Uint8Array(0),
+    ]
+  }
+
+  /**
+   * Returns the serialized encoding of the EIP-7702 transaction.
+   *
+   * Format: `0x02 || rlp([chainId, nonce, maxPriorityFeePerGas, maxFeePerGas, gasLimit, to, value, data,
```
g11tech, line 270: authoritylist missing here in comment

packages/tx/src/eip7702Transaction.ts
```diff
@@ -0,0 +1,396 @@
...
+   * @param baseFee The base fee of the block (will be set to 0 if not provided)
+   */
+  getUpfrontCost(baseFee: bigint = BIGINT_0): bigint {
+    return EIP1559.getUpfrontCost(this, baseFee)
+  }
+
+  /**
+   * Returns a Uint8Array Array of the raw Bytes of the EIP-7702 transaction, in order.
+   *
+   * Format: `[chainId, nonce, maxPriorityFeePerGas, maxFeePerGas, gasLimit, to, value, data,
```
g11tech, line 239: authorizationList, missing in comment

packages/tx/src/eip7702Transaction.ts
```diff
@@ -0,0 +1,396 @@
...
+   *
+   * Note: in contrast to the legacy tx the raw message format is already
+   * serialized and doesn't need to be RLP encoded any more.
+   *
+   * ```javascript
+   * const serializedMessage = tx.getMessageToSign() // use this for the HW wallet input
+   * ```
+   */
+  getMessageToSign(): Uint8Array {
+    return EIP2718.serialize(this, this.raw().slice(0, 9))
```
g11tech, line 293:
```suggestion
    return EIP2718.serialize(this, this.raw().slice(0, 10))
```

packages/tx/src/eip7702Transaction.ts
```diff
@@ -0,0 +1,396 @@
...
+    const accessListJSON = AccessLists.getAccessListJSON(this.accessList)
+    const baseJson = super.toJSON()
+
+    return {
+      ...baseJson,
+      chainId: bigIntToHex(this.chainId),
+      maxPriorityFeePerGas: bigIntToHex(this.maxPriorityFeePerGas),
+      maxFeePerGas: bigIntToHex(this.maxFeePerGas),
+      accessList: accessListJSON,
+      authorizationList: this.AuthorizationListJSON,
```
g11tech, line 374: why are we saving json upfront in the

packages/tx/src/types.ts
```diff
@@ -99,6 +117,25 @@ export function isAccessList(input: AccessListBytes | AccessList): input is Acce
...
+    return true
+  }
+  const firstItem = input[0]
+  if (Array.isArray(firstItem)) {
+    return true
+  }
+  return false
+}
+
+export function isAuthorizationList(
```
g11tech, line 133: i think typescript should be able to infer the other type by itself if you use !isAuthorizationList

but am not 100% sure of typescript shenanigans

packages/tx/src/util.ts
```diff
@@ -116,6 +129,98 @@ export class AccessLists {
...
+export class AuthorizationLists {
+  public static getAuthorizationListData(
+    authorizationList: AuthorizationListBytes | AuthorizationList
+  ) {
+    let AuthorizationListJSON
+    let bufferAuthorizationList
+    if (isAuthorizationList(authorizationList)) {
+      AuthorizationListJSON = authorizationList
+      const newAuthorizationList: AuthorizationListBytes = []
+      const jsonItems = ['chainId', 'address', 'nonce', 'yParity', 'r', 's']
```
g11tech, line 141: would just reading the properities and comparing them to undefined with a `||` be simpler ?

packages/vm/src/runTx.ts
g11tech: if we have a extcodehash for EOA in a following transaction of th, its likely it will get the old codeshash.
should be possible to construct a testcase

packages/tx/src/eip7702Transaction.ts
jochem-brouwer: We also do this in 1559 / 2929, I suggest we factor this out in another PR which also covers those txs

packages/tx/src/types.ts
jochem-brouwer: I have copied this pattern from the access lists :)

packages/tx/src/util.ts
jochem-brouwer: I'm not sure what you mean here?

acolytec3 (approved):
LGTM

holgerd77:
🎉

packages/common/src/interfaces.ts
```diff
@@ -65,6 +65,30 @@ export type AccessListBytesItem = [Uint8Array, Uint8Array[]]
...
+  Uint8Array,
+  Uint8Array,
+  Uint8Array[],
+  Uint8Array,
+  Uint8Array,
+  Uint8Array
+]
+export type AuthorizationListBytes = AuthorizationListBytesItem[]
+export type AuthorizationList = AuthorizationListItem[]
+
```
holgerd77, line 91: There is not a good reason that we have these `AuthorizationList` types in the Common interface file (these should really be reserved for the VIIs, to the "Very Important Interfaces" 😂 used to type on other libraries.

Same goes I guess for the access list interfaces (for both: unless I am unaware of some below-tx-library usages).

We should move them over to tx in a small clean-up PR.

packages/tx/test/eip7702.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+          authorizationList: [
+            {
+              chainId: '0x',
+              address: undefined as never,
+              nonce: ['0x1'],
+              yParity: '0x1',
+              r: ones32,
+              s: ones32,
+            },
+          ],
```
holgerd77, line 159: These repeated test data lists make this whole test setup pretty expansive. Can we clean this up by taking a base authorization list item data set (or the full authorization list, whatever fits better), and then only overwrite the data item to test (can maybe be done in one PR togehter with the interface movings to tx, see above).

Guess that's something to //cc @jochem-brouwer in.

packages/tx/test/eip7702.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+        { common }
+      )
+    }, 'chainId is not defined')
+
+    assert.doesNotThrow(() => {
+      EOACodeEIP7702Transaction.fromTxData(
+        {
+          authorizationList: [
+            {
+              chainId: '0x',
```
holgerd77, line 188: This is valid? To have the `chainId` set to `0x`??

holgerd77: Respectively going further (without having read the EIP): shouldn't there be a check on our side if the chainId here matches the one from Common and otherwise (we) throw?

packages/vm/src/runTx.ts
```diff
@@ -440,6 +447,53 @@ async function _runTx(this: VM, opts: RunTxOpts): Promise<RunTxResult> {
...
+    const authorizationList = (<EIP7702CompatibleTx>tx).authorizationList
+    const MAGIC = new Uint8Array([5])
+    for (let i = 0; i < authorizationList.length; i++) {
+      // Authority tuple validation
+      const data = authorizationList[i]
+      const chainId = data[0]
+      const chainIdBN = bytesToBigInt(chainId)
+      if (chainIdBN !== BIGINT_0 && chainIdBN !== this.common.chainId()) {
+        // Chain id does not match, continue
+        continue
```
holgerd77, line 463: Ok, just seeing this here.

So I would interpret this (still without reading the EIP): txs with non-matching chain IDs are valid (could be included in mainnet), but execution just "does nothing" if not matching.

Is this correct/intended?

packages/tx/test/eip7702.spec.ts
jochem-brouwer: Yes, that would be a better approach in general, will address.

packages/tx/test/eip7702.spec.ts
jochem-brouwer: Yes, 0x means it is valid on all chains. From the EIP:

`2. Verify the chain id is either 0 or the chain’s current ID.`

packages/vm/src/runTx.ts
jochem-brouwer: Yes, from the EIP:

`If any of the above steps fail, immediately stop processing that tuple and continue to the next tuple in the list. `

The list contains point (2) the chain id check, it should either be 0 or the current chain id. If it thus does not match (and it is nonzero) then we continue processing, but it thus does not change any code. (Point 7 is changing the code of the authority).

See: https://eips.ethereum.org/EIPS/eip-7702#behavior

packages/common/src/interfaces.ts
jochem-brouwer: I just checked, for both (so also `AccessList`) the only packages which import those are tx and vm, and since vm has a tx dependency we can indeed move these into common.

I don't recall why we put AccessList in common (but to be consistent I also put AuthorizationList in here). I do slightly recall we maybe did this for package consumers to import these types (?) but then they would only need the `common` package and not import `tx`? However this does not sound like a very good reason for me, I don't know if there is any use case for those types which would not depend or use `tx`.

We should indeed clean-up and remove both `AuthorizationList` and `AccessList` from common (in same PR)

packages/tx/test/eip7702.spec.ts
holgerd77: Ah, interesting.

holgerd77: Ok, but then I guess at least Common (and therefore signed tx through chain ID in signature) and chain ID here differing should throw on the tx side?

Because atm it does not, tested with mainnet Common and chain ID 2.

holgerd77: Had another look into the EIP, this (chain ID check) is only mentioned in execution, but would this make sense to allow such a tx in the network (so: in a block?).

If you haven't heard anything specific about that we should likely ask for confirmation/explanation/re-evaluation.

jochem-brouwer: Yes, it is thus a valid tx and it is thus allowed in the block, we should not throw. If the chain id does not match, we thus "skip" the tuple and do not invalidate the tx.

holgerd77: What was the reasoning to allow this during the EIP discussion/specification?

holgerd77: (or: *was* this a decision?)

jochem-brouwer: I don't know the exact reason, but I think it is among the lines just like with access lists, you are allowed to put the same address twice. This does not invalidate the tx, you just have to pay more gas (so no one wil do it). The same reasoning works here, you can input any authority tuple, but if it does not match the chain id it will do absolutely nothing, but you have to pay gas.

PR #3471, opened 2024-06-24
EVM: Generic BLS Interface / Use JS Implementation (@noble/curves) as Default

holgerd77:
Follow-up on #3350 

Goal of this PR is to move over to use a JavaScript BLS implementation for the BLS precompiles ([EIP-2537](https://eips.ethereum.org/EIPS/eip-2537)). For this the direct usage of the MCL library in the precompile files as well as the tight MCL library coupling (by adding as a property to the EVM) is replaced by using a generic `EVMBLSInterface`. It will likely be possible later on to fully allow switching the implementation, not sure if we want to go that far, we'll see.

This first push is for testing with the official test suite after setting up the base structures (as well as some temporary double structures like the doubled mcl code for gradual implementation which will go away at the end).


holgerd77:
grafik

I've done something wrong with my fork builds of `@noble/curves` from here https://github.com/holgerd77/noble-curves/tree/build-5fcd71a, I battled with this already the first 1 1/2 hours of the morning and ended up manually copying over the code to my `node_modules` folder.

This is a known issue though and should go away once `@noble/curves` gets an official update with the changes.

holgerd77:
So if one is running

```shell
npx vitest run test/precompiles/eip-2537-bls.spec.ts
```

from the `evm` folder locally these are the more meaningful test runs (currently 8/131 tests failing).

jochem-brouwer:
Instructions on how to run the `ethereum-tests` BLS tests (please note, we had a discussion on that PR in April, see https://github.com/ethereum/tests/pull/1366)

Ensure `ethereum-tests` is on `v14`:

`cd ./packages/ethereum-tests && git checkout tags/v14.0`

To run the state tests:

`cd ./packages/vm && npm run test:state -- --fork=Prague --dir=../EIPTests/StateTests/stEIP2537`

Note: blockchain tests cannot be ran, it throws some error about a header field missing, so it runs on partial Prague.

State tests output: 

```
1..1403
# tests 1403
# pass  1217
# fail  186
```

Note: also not sure if the tests are correct :thinking: Might have to see if execution-spec-tests has some tests also.

jochem-brouwer:
Oh wait, nvm, I can of course test using `mcl-wasm`. If I re-run the tests with MCL, then all pass. So we definitely have bugs!

packages/evm/src/index.ts
```diff
@@ -30,4 +35,14 @@ export type {
...
+export {
+  EOF,
+  EVM,
+  EvmError,
+  EVMErrorMessage,
+  getActivePrecompiles,
+  getOpcodesForHF,
+  MCLBLS,
+  Message,
+  NobleBLS,
```
jochem-brouwer, line 47: I'm not sure if we want to export these semi-internal `NobleBLS` and `MCLBLS`?

packages/evm/src/precompiles/bls12_381/util.ts
```diff
@@ -0,0 +1,86 @@
...
+      )
+    }
+    return false
+  }
+  return true
+}
+
+export const zeroByteCheck = (
+  opts: PrecompileInput,
+  zeroByteRanges: number[][],
```
jochem-brouwer, line 69: Note: this will thus slice the array **exclusive** of `zeroByteRanges[x][1]`. So for instance the array `[[0,1]]` will only check the 0th byte, not the byte at position 1. Is this correct? (And can docs be added for this method?)

packages/evm/src/precompiles/bls12_381/util.ts
```diff
@@ -0,0 +1,86 @@
...
+  pName: string,
+  pairStart = 0
+) => {
+  for (const index in zeroByteRanges) {
+    const slicedBuffer = opts.data.subarray(
+      zeroByteRanges[index][0] + pairStart,
+      zeroByteRanges[index][1] + pairStart
+    )
+    if (!(equalsBytes(slicedBuffer, ZERO_BYTES_16) === true)) {
```
jochem-brouwer, line 78: Why is this constant `ZERO_BYTES_16` here and not `zeros(x)` where `x` is the expected length of the zero byte array?

packages/evm/src/types.ts
```diff
@@ -256,6 +256,22 @@ export interface EVMOpts {
...
+   * is used (see `noble.ts` file in the `precompiles` folder).
+   *
+   * To use an alternative implementation this option can be used by passing
+   * in a wrapper implementation integrating the desired library and adhering
+   * to the `EVMBLSInterface` specification.
+   *
+   * An interface for the MCL WASM implementation https://github.com/herumi/mcl-wasm
+   * is shipped with this library, see the following test file for an example on
+   * how to use:
+   * https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/evm/test/precompiles/eip-2537-bls.spec.ts
```
jochem-brouwer, line 271: This test file does not have tests which specifically use MCL. Did you intend to use a permalink here?

jochem-brouwer: I think you intend to permalink to the test which specifically runs on this PR (and later thus likely on master). Could you also point to the specific lines in this permalink? Especially the mcl init and the mcl curve settings should be initialized before being passed into the EVM.

packages/evm/test/precompiles/eip-2537-bls.spec.ts
```diff
@@ -35,48 +36,63 @@ const precompileMap: { [key: string]: string } = {
...
-            await precompile(callData)
-            assert.fail('The precompile should have thrown')
-          } catch {
-            // If the precompile throws, catch it here, but this is fine, since it is expected to fail
+// MCL Instantiation
+await mcl.init(mcl.BLS12_381)
+mcl.setMapToMode(mcl.IRTF) // set the right map mode; otherwise mapToG2 will return wrong values.
+mcl.verifyOrderG1(true) // subgroup checks for G1
+mcl.verifyOrderG2(true) // subgroup checks for G2
+const mclbls = new MCLBLS(mcl)
```
jochem-brouwer, line 44: I think we can maybe improve here (I also now see why we export MCLBLS / NobleBLS). Can we add a static method to the interfaces which inits these BLSs? Then we can init the MCL using `MCLBLS.init(mcl)` and write those settings there

jochem-brouwer (commented):
Some comments, but this looks extremely good in general, especially how MCL / Noble is now decoupled, I love it!

packages/evm/src/index.ts
jochem-brouwer: Nvm, I now see, we want this to init MCLBLS if users want to use it. (Read other comment for some more suggestions)

packages/evm/test/precompiles/eip-2537-bls.spec.ts
holgerd77: That's a great idea, will directly add! 👍 🙂

holgerd77: I now went for a non-static init() version, I find it a bit nicer if the out asny initialization (so `await mcl.init(mcl.BLS12_381)`) is not also wrapped and can be done separately by the user, guess both ways would work and be fine though, for this one I find the typing also a bit nicer. The optional `init()` method from the interface is the automatically called from within the EVM constructor if present.

```ts
import * as mcl from 'mcl-wasm'

await mcl.init(mcl.BLS12_381)
const evm = await EVM.create({ mcl: new MCLBLS(mcl) })
```

packages/evm/src/precompiles/bls12_381/util.ts
holgerd77: Yes, I think this should be all correct, actually this only copies your former code into a new function 😂, see this commit https://github.com/ethereumjs/ethereumjs-monorepo/pull/3471/commits/62f15f2c9da06ef81e2ff92fea0d72d344c9f076.

So this is a BLS specific zero byte check, applying this part of the [EIP](https://eips.ethereum.org/EIPS/eip-2537#field-elements-encoding):

grafik

And specificall this "Due to the size of p, the top 16 bytes are always zeroes." part, see that the input subarrays always have a length of 16 bytes.

I'll give this and the other methods some documentation.

packages/evm/src/precompiles/bls12_381/util.ts
holgerd77: See answer above.

packages/evm/src/precompiles/bls12_381/util.ts
```diff
@@ -0,0 +1,141 @@
...
+ * ]
+ * ```
+ *
+ * @param opts
+ * @param zeroByteRanges
+ * @param pName
+ * @param pairStart
+ * @returns
+ */
+export const zeroByteCheck = (
```
jochem-brouwer, line 122: Nit: this does zero byte checks, but only if the range of these zero bytes is 16 zero bytes, so the name is a bit misleading.

packages/vm/test/tester/index.ts
```diff
@@ -45,6 +45,7 @@ import type { bn128 } from '@ethereumjs/evm'
  * --expected-test-amount: number.        If passed, check after tests are ran if at least this amount of tests have passed (inclusive)
  * --verify-test-amount-alltests: number. If passed, get the expected amount from tests and verify afterwards if this is the count of tests (expects tests are ran with default settings)
  * --reps: number.                        If passed, each test case will be run the number of times indicated
+ * --bls: string.                         Use MCL as a bls library if passed (default: Noble) (state tests only)
```
jochem-brouwer, line 48: Could we also add this for blockchain tests?

jochem-brouwer (commented):
This review is on top of my previous review, got 2 small comments, LGTM in general :smile:

packages/evm/src/precompiles/bls12_381/noble.ts
```diff
@@ -0,0 +1,348 @@
...
+
+  const G1 = bls12_381.G1.ProjectivePoint.fromAffine({
+    x,
+    y,
+  })
+
+  return G1
+}
+
+// input: a mcl G1 point
```
acolytec3, line 55: Do we need all these `mcl` references?  I'm assuming these are obsolete in this file, right?  Can they be replaced with something like `noble/curves`

packages/vm/test/tester/index.ts
acolytec3: Following on this, is there a reason we wouldn't want to use MCL by default in the test runner?  I don't know how many BLS related tests there are but anything that slows down the tests seems undesirable to me since these are the main bottleneck in our CI right now, right?

acolytec3 (changes requested):
Looks great overall.  One item regarding comments in the `noble.ts` that should be clarified.

packages/evm/src/precompiles/bls12_381/util.ts
holgerd77: Good point, will rename to `leading16ZeroBytesCheck`

packages/vm/test/tester/index.ts
holgerd77: Was a bit lazy and I feel that every new code is this messy test runner is a bit of "lost code" 😂, so I decided to skip. But I can very well add, makes sense for practical reasons and shouldn't be so much extra work.

holgerd77: @acolytec3 Yes, guess we can do that for the VM as long as we keep running tests for both implementations in the EVM.

packages/evm/src/precompiles/bls12_381/noble.ts
holgerd77: Don't think that a `noble/curves` reference is needed here, will just completely remove

jochem-brouwer (dismissed):
LGTM!

jochem-brouwer (approved):
EVM lint failed, should pass now, still LGTM :smile: :+1:

PR #3473, opened 2024-06-24
Implement `trie.put`

acolytec3:
This builds out `trie.put` for the `VerkleTrie` class


packages/verkle/src/verkleTree.ts
```diff
@@ -213,9 +215,149 @@ export class VerkleTree {
    * @param value - the value to store
    * @returns A Promise that resolves once value is stored.
    */
-  // TODO: Rewrite following logic in verkle.spec.ts "findPath validation" test
-  async put(_key: Uint8Array, _value: Uint8Array): Promise<void> {
-    throw new Error('not implemented')
+  async put(key: Uint8Array, value: Uint8Array): Promise<void> {
```
ScottyPoi, line 218: This function is very long.  I find functions more difficult to read and debug when I have to scroll in order to see different parts of it.

I appreciate the way you use the comments to outline how things are happening step by step.  I wonder if these different steps could be externalized into separate functions, so that the `put` function itself can fit on a screen?

acolytec3: This is a 100% valid criticism.  Lemme look at the logic and see if there's a way to break it down.  I think you're right that it can be done in smaller chunks.

packages/verkle/src/verkleTree.ts
```diff
@@ -213,11 +215,169 @@ export class VerkleTree {
...
+    let leafNode: LeafNode
+    // First see if leaf node already exists
+    if (foundPath.node !== null) {
+      // Sanity check to verify we have the right node type
+      if (foundPath.node.type !== VerkleNodeType.Leaf) {
+        throw new Error(
+          `expected leaf node found at ${bytesToHex(stem)}. Got internal node instead`
+        )
+      }
+      leafNode = foundPath.node as LeafNode
```
gabrocheleau, line 243: I can create some type guards for those, I think this will be useful in qute a few places.

gabrocheleau: Just did: [`9b00803` (#3473)](https://github.com/ethereumjs/ethereumjs-monorepo/pull/3473/commits/9b00803c1708ce30839ac2bfd28be87cef768c00)

acolytec3: Sounds good!

packages/verkle/src/verkleTree.ts
```diff
@@ -213,11 +215,168 @@ export class VerkleTree {
...
+        `Updating value for suffix: ${suffix} at leaf node with stem: ${bytesToHex(stem)}`,
+        ['PUT']
+      )
+    putStack.push([leafNode.hash(), leafNode])
+
+    // `path` is the path to the last node pushed to the `putStack`
+    let lastPath = leafNode.stem
+
+    // Step 2) Determine if a new internal node is needed
+    if (foundPath.stack.length > 1) {
```
gabrocheleau, line 274: Just to make sure I am getting this correctly, could you confirm that this scenario is correct:
- Let's say we start with a root node
- We put a key/value pair where the key starts with `0x01`. 
- `findPath` returns the rootNode.
- We create a leafNode, and add a pointer to the newly created leaf node at index `2` of the root node. 
- We put another key/value pair that also starts with `0x01`. 
- `findPath` now returns the `leafNode` that was created in the previous put operation, along with the `remaining` values where the paths branch off. 
- We then create an internal node where the paths branch off, commit to the two children to it, and then commit the internal node to the root node (at index `2`).
- rootHash gets updated.
-

acolytec3: That should be how it works. Not guaranteeing I got it exactly right but the tests seem to indicate so

packages/verkle/src/verkleTree.ts
```diff
@@ -213,11 +215,168 @@ export class VerkleTree {
...
+        ['PUT']
+      )
+    this.DEBUG && this.debug(`Updating root node hash to ${bytesToHex(this._root)}`, ['PUT'])
+    putStack.push([this._root, rootNode])
+    await this.saveStack(putStack)
   }
 
+  /**
+   * Helper method for updating or creating the parent internal node for a given leaf node
+   * @param leafNode the reference leaf node
```
gabrocheleau, line 328: By reference leaf node I assume you mean the "new leaf node". Do you think that "newLeafNode/new leaf node" would be a clearer naming?

packages/verkle/test/verkle.spec.ts
```diff
@@ -247,4 +221,39 @@ describe('findPath validation', () => {
...
+    await trie.put(hexToBytes(keys[0]), hexToBytes(values[0]))
+    await trie.put(hexToBytes(keys[1]), hexToBytes(values[1]))
+    await trie.put(hexToBytes(keys[2]), hexToBytes(values[2]))
+    await trie.put(hexToBytes(keys[3]), hexToBytes(values[3]))
+    const val1 = await trie.get(hexToBytes(keys[0]))
+    assert.deepEqual(val1, hexToBytes(values[0]))
+    const val3 = await trie.get(hexToBytes(keys[2]))
+    assert.deepEqual(val3, hexToBytes(values[2]))
+    const val4 = await trie.get(hexToBytes(keys[3]))
+    assert.deepEqual(val4, hexToBytes(values[3]))
```
gabrocheleau, line 257: Imo the `val1, val2` naming makes this a bit tedious to read through. Could we just inline the `await trie.get(...)` inside the assert.deepEqual ?

packages/verkle/src/verkleTree.ts
gabrocheleau: Cool, I didn't infer the above from how the function/tests were structured, but wrote it based on my understanding of verkle trees, good to hear it matches

gabrocheleau (approved):
Excellent work! Pushed and implemented some typeguards, and left some minor nitpicks that you might want to consider. Othewise this can be merge imo!

packages/verkle/src/verkleTree.ts
acolytec3: Hmm, it could be a new leaf node but it could also just be an existing leaf node where we've updated a value.  In this instance, we wouldn't be creating a new internal node above but would still update the parent.  Maybe I can call it "child leaf node"?

packages/verkle/src/verkleTree.ts
acolytec3: Looking at your comment again, let me confirm something.  What you call "index 2" here would be `rootNode.children[1]` right?  These are zero indexed arrays and so inserting a node that starts with `0x01` means you check `rootNode.children[1]` to see if there's an existing reference to some child node and if not, create a new leaf node with the value for the key that starts with `0x01` and then put the commitment from that new leaf node in `rootNode.children[1]`

packages/verkle/test/verkle.spec.ts
acolytec3: Yup!

packages/verkle/src/verkleTree.ts
gabrocheleau: Ah that's fair. I'm not actually sure I find child leaf node clearer. I think the comment makes it clear enough, thanks for addressing.

gabrocheleau (approved):

packages/verkle/src/verkleTree.ts
gabrocheleau: Ah yes that's what I meant, sorry for the confusing indexing

PR #3475, opened 2024-06-27
common: update eip 2935 as per latest devnet1 spec and add related eip 7709

g11tech:
for devnet 1 of prague (spec: https://notes.ethereum.org/@ethpandaops/pectra-devnet-1)

eip 2935 got added with `set` leading to bytecode change as well as got broken down into two parts with `BLOCKHASH` reading from state moved to 7709 to be activated in verkle (without changing the BLOCKHASH window)

- https://eips.ethereum.org/EIPS/eip-7709
- https://eips.ethereum.org/EIPS/eip-2935

### DIFF
--- a/packages/client/test/rpc/engine/newPayloadV4.spec.ts
+++ b/packages/client/test/rpc/engine/newPayloadV4.spec.ts
@@ -50,7 +50,10 @@ describe(`${method}: call with executionPayloadV4`, () => {
     const { pragueJson, pragueTime } = readyPragueGenesis(genesisJSON)
     const { service, server } = await setupChain(pragueJson, 'post-merge', { engine: true })
     const rpc = getRpcClient(server)
-    // const res1 = await rpc.request(`eth_getBlockByNumber`, ["0x0",false])
+    let res
+
+    res = await rpc.request(`eth_getBlockByNumber`, ['0x0', false])
+    assert.equal(res.result.hash, validForkChoiceState.headBlockHash)
 
     const validBlock = {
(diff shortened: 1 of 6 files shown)


acolytec3:
Looks fine as far as it goes.  Lemme know once ready for final review.

packages/client/test/rpc/engine/newPayloadV4.spec.ts
```diff
@@ -50,6 +50,8 @@ describe(`${method}: call with executionPayloadV4`, () => {
     const { pragueJson, pragueTime } = readyPragueGenesis(genesisJSON)
     const { service, server } = await setupChain(pragueJson, 'post-merge', { engine: true })
     const rpc = getRpcClient(server)
+    // const res1 = await rpc.request(`eth_getBlockByNumber`, ["0x0",false])
```
acolytec3, line 53: Should this be commented out or can we remove it?

acolytec3 (commented):
One commented out code line we should either remove or uncomment.

packages/client/test/rpc/engine/newPayloadV4.spec.ts
g11tech: i want to retain it to get the genisis data to build out the test if/when something changes which causes genesis change like what happened here.

g11tech: uncommented and added a new assert

acolytec3 (approved):

PR #3476, opened 2024-06-28
Optimize storage of default values in `VerkleNode`

acolytec3:
Optimizes how we store default values for `InternalNode` and `LeafNode` so memory footprint is shrunk.

For `InternalNode`, if an element of `internalNode.children` has not been written to, the value is stored as `null`.

For `LeafNode`, if an element of the values array is 0 and has not been written before, it's stored as `0`.  If it's 0 and has been deleted, it's stored as `1`.  Otherwise, it's stored in its 32 byte `Uint8Array` representation.

Have also added some helpers to create default values (empty leafnode value, deleted leafnode value)


holgerd77 (dismissed):
Nice, LGTM!

acolytec3:
Need to fix some bugs locally before we merge. Just noticed I wasn't handling some of the changes correctly in the `put` method.

packages/verkle/src/node/leafNode.ts
```diff
@@ -1,15 +1,21 @@
...
+} from './types.js'
 import { createCValues } from './util.js'
 
 import type { VerkleNodeOptions } from './types.js'
 import type { VerkleCrypto } from '@ethereumjs/util'
 
 export class LeafNode extends BaseVerkleNode<VerkleNodeType.Leaf> {
   public stem: Uint8Array
-  public values: Uint8Array[] // Array of 256 possible values represented as 32 byte Uint8Arrays
+  public values: (Uint8Array | 0 | 1)[] // Array of 256 possible values represented as 32 byte Uint8Arrays or 0 if untouched or 1 if deleted
```
gabrocheleau, line 18: Could we, instead of numbers, use an enum? I think that would be clearer in usage

gabrocheleau: I've gone ahead and done that, let me know what you think

packages/verkle/src/node/types.ts
```diff
@@ -45,3 +51,15 @@ export interface VerkleNodeOptions {
...
+
+export const createZeroLeaf = () => new Uint8Array(32)
+
+export const createDeletedLeafValue = () =>
+  setLengthLeft(bigIntToBytes(bytesToBigInt(new Uint8Array(32)) + BigInt(2 ** 128)), 32)
+
+export const DELETED_LEAF_VALUE = setLengthLeft(
+  bigIntToBytes(bytesToBigInt(new Uint8Array(32)) + BigInt(2 ** 128)),
+  32
+)
```
gabrocheleau, line 63: Why do we need a constant in addition to a function? Also, could that be made more efficient by manipulating bytes directly rather than doing BigInt operations and multiple conversions?

acolytec3: I did the constant since we compare the bytes from a serialized node that's stored in the db to the constant when reinstating it. I guess we might not need to now that you mention it though. Just need a clean way to represent the deleted value type in the db.

acolytec3: And yes, doing this with byte arithmetic is the more efficient way. Just hadn't quit decided on the best way to do it. The more I think about it, the spec is a little goofy here. Seems like there should be a more efficient way to represent deleted value

acolytec3: And looks like you already took care of it. I like your edits.

packages/verkle/src/node/leafNode.ts
acolytec3: These edits look great! Thanks!

packages/verkle/src/node/leafNode.ts
```diff
@@ -133,16 +133,23 @@ export class LeafNode extends BaseVerkleNode<VerkleNodeType.Leaf> {
...
+   * Set the value at the provided index from the values array and update the node commitments
+   * @param index the index of the specific leaf value to be updated
+   * @param value the value to insert into the leaf value at `index`
+   */
+  setValue(index: number, value: Uint8Array | VerkleLeafNodeValue): void {
+    let val
+    // `val` is a bytes representation of `value` used to update the cCommitment
+    if (value instanceof Uint8Array) val = value
+    else
+      val = value === VerkleLeafNodeValue.Untouched ? new Uint8Array(32) : createDeletedLeafValue()
```
gabrocheleau, line 146: Could use `createUntouchedLeafValue() ` instead of new Uint8Array(32)

acolytec3: Done.  Anything else you want to add before we merge?

gabrocheleau: All good lgtm

gabrocheleau (approved):

PR #3478, opened 2024-06-30
vm: remove backfill of block hashes on 2935 activation

g11tech:
remove the backfill of the last 256 hashes as it is already removed from the eip following  kenya discussions

### DIFF
--- a/packages/vm/src/runBlock.ts
+++ b/packages/vm/src/runBlock.ts
@@ -493,8 +493,6 @@ export async function accumulateParentBlockHash(
   )
   const historyServeWindow = this.common.param('vm', 'historyServeWindow')
 
-  const forkTime = this.common.eipTimestamp(2935)
-
   // getAccount with historyAddress will throw error as witnesses are not bundeled
   // but we need to put account so as to query later for slot
   try {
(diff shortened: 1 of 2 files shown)


acolytec3 (approved):

packages/vm/src/runBlock.ts
```diff
@@ -524,28 +522,6 @@ export async function accumulateParentBlockHash(
   }
   await putBlockHash(this, parentHash, currentBlockNumber - BIGINT_1)
```
jochem-brouwer, line 523: I think we can now remove this function, since it is only called once.

packages/vm/test/api/EIPs/eip-2935-historical-block-hashes.spec.ts
```diff
@@ -211,9 +211,9 @@ describe('EIP 2935: historical block hashes', () => {
     })
     it('should ensure blocks older than 256 blocks can be retrieved from the history contract', async () => {
       // Test: build a chain with 256+ blocks and then retrieve BLOCKHASH of the genesis block and block 1
-      const blocksActivation = 256 // This ensures that block 0 - 255 all get stored into the hash contract
+      const blocksActivation = 256 // This ensures that only block 255 gets stored into the hash contract
       // More than blocks activation to build, so we can ensure that we can also retrieve block 0 or block 1 hash at block 300
-      const blocksToBuild = 300
+      const blocksToBuild = 500
```
jochem-brouwer, line 216: Do we now need 500 blocks, or can we also "just" create a few blocks (say 5)? We only need to test that the last blockhash is put into the contract at fork activation time.

jochem-brouwer (commented):
Two questions/comments. It also seems that CI is not passing :thinking:

packages/vm/test/api/EIPs/eip-2935-historical-block-hashes.spec.ts
g11tech: yes the test tests various scenarios of blockhash as well as the contract code windows as well as activation time

holgerd77:
Short note: there is a nice type label `type: spec updates` which sums up these kind of changes nicely.

PR #3482, opened 2024-07-04
EVM/Common: SimpleStateManager

holgerd77:
~This is a proof of concept if it is possible~ This PR implements a `SimpleStateManager` with the goal to replace the default state manager implementation and depdendency from the EVM with a very simple < 5KB state manager located in `Common` without any further depdencies.

This would allow to ship with a significantly trimmed down base EVM, both with a substantially reduced dependency set and bundle size, going from 1.23 MB to 850 KB.

Before:

evm-bundle_before_SM_removal

After (also note the "number of boxes": statemanager out, trie out, js-sdsl out, readable-stream out, Buffer out, lru-cache out). So if you click to enlarge, this is so much more a representation of what the EVM is about than before:

evm-bundle_after_StateManager_removal

This would ease browser usage for simple EVM use cases and integrations.

Some more comments to make here, short on time right now.

I have integrated this into the general SM test suite, all checkpoint tests are passing.

For the EVM this also already gives pretty solid results, with adhoc (without any modification) only 4 / 1107 tests failing.

grafik


holgerd77:
Ok, ready! 🎉 

All EVM tests passing, the one thing I needed to do for this is to also move the `OriginalStorageCache` short implementation to `Common` and integrate, since otherwise the pre-state setting use/test cases were not working.

So: I think for the EVM this is a big deal! So this lifts a weight of close to 400KB to the EVM in addition to the other reductions, and so brings this package in the realm of becoming a "normal" browser library which people just side-integrate among other dependencies. The overall code adition "on the other side" is still < 8 KB.

Want to nevertheless some more cases here respectively give some additional context/explanation about design choices, so to make things more transparent and "followable".

### Why Common ?

So, this is admittedly relatively low in the stack. This was mainly guided by the following design "needs":

1. The implementation should implement the `EVMStateManager` interface being located in Common (also for the same "independent from the `statemanager` package reason)
2. Being "lower" than `statemanager` made it possible to relatively easily and without broader moving things around re-use the tests from `statemanager` and run them directly

We could also think about moving this to the `EVM` package directly instead. This would make 2. more complicated. I also find the idea relatively charming to just have a simple state manager available "everywhere" which can be used to manage Ethereum state on a basic level for whatever use cases, so for me it feels a bit as a general Utility which fits well into the chosen setting. Comparing the total `Common` bundle size of 178 KB this feels worth and in-relation to me to add. I would therefore pledge for this solution.

I think this new state manager will be generally useful, also for our own use cases, mainly e.g. regarding testing, stumbled upon this thought yesterday when seeing you guys battling with the `DefaultStateManager` bug. So with this simpler state manager makes it really handy to dig into state changes during EVM execution, since one can plainly see the state evolving in the simple map structures. Used this yesterday during a debugging session within VSCode, and this feels really handy to have cache/trie and so on out of the mix.

Anyhow: ready for review! 🙂 

Note that this PR builds upon the BLS PR (for bundle comparison reasons). I think it's better to merge BLS first since these are conceptually unrelated.

packages/common/src/state/simple.ts
```diff
@@ -0,0 +1,185 @@
...
+    this.originalStorageCache = new OriginalStorageCache(this.getContractStorage.bind(this))
+  }
+
+  topA() {
+    return this.accountStack[this.accountStack.length - 1]
+  }
+  topC() {
+    return this.codeStack[this.codeStack.length - 1]
+  }
+  topS() {
```
jochem-brouwer, line 51: Should these methods be exported publicly? They should also get a more descriptive name e.g. `topStorageStack()`

packages/common/src/state/simple.ts
```diff
@@ -0,0 +1,185 @@
...
+      const accountCopy =
+        account !== undefined
+          ? Object.assign(Object.create(Object.getPrototypeOf(account)), account)
+          : undefined
+      newTopA.set(address, accountCopy)
+    }
+    this.accountStack.push(newTopA)
+    this.codeStack.push(new Map(this.topC()))
+    this.storageStack.push(new Map(this.topS()))
+  }
```
jochem-brouwer, line 67: What does this method do? It seems to copy the account, but it shallow copies the code and storage stack?

jochem-brouwer: (is this a checkpoint?)

packages/common/src/state/simple.ts
```diff
@@ -0,0 +1,185 @@
...
+  async getContractStorage(address: Address, key: Uint8Array): Promise<Uint8Array> {
+    return this.topS().get(`${address.toString()}_${bytesToHex(key)}`) ?? new Uint8Array(0)
+  }
+
+  async putContractStorage(address: Address, key: Uint8Array, value: Uint8Array): Promise<void> {
+    this.topS().set(`${address.toString()}_${bytesToHex(key)}`, value)
+  }
+
+  async checkpoint(): Promise<void> {
+    this.add()
```
jochem-brouwer, line 119: Ah ok, `add` is an alias for `checkpoint`

packages/common/src/state/simple.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+  async commit(): Promise<void> {
+    const topA = this.accountStack.pop()
+    const topC = this.codeStack.pop()
+    const topS = this.storageStack.pop()
+    this.accountStack.pop()
+    this.codeStack.pop()
+    this.storageStack.pop()
+    this.accountStack.push(topA!)
```
jochem-brouwer, line 128: `pop` followed by `push` is the same as overwriting the topmost index, I suggest changing to that.

`this.accountStack[this.accountStack.length - 1] = topA!`

packages/common/src/state/simple.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  async revert(): Promise<void> {
+    this.accountStack.pop()
+    this.codeStack.pop()
+    this.storageStack.pop()
+  }
+
+  async flush(): Promise<void> {}
+  clearCaches(): void {}
```
jochem-brouwer, line 140: These empty methods should throw a "Not implemented" error.

packages/common/src/state/simple.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  shallowCopy(): EVMStateManagerInterface {
+    const copy = new SimpleStateManager()
+    for (let i = 0; i < this.accountStack.length; i++) {
+      copy.accountStack.push(new Map(this.accountStack[i]))
+      copy.codeStack.push(new Map(this.codeStack[i]))
+      copy.storageStack.push(new Map(this.storageStack[i]))
```
jochem-brouwer, line 147: So this shallow copy will initialize the shallow copied state manager with the current state of the stack of this state manager.

For the `accountStack`, if you edit the account on the original state manager, then it will also be edited on the copied state manager, I do not think that this is desired? (So need to create copies of the accounts of the `accountStack`).

Side-note: if you would raw-edit the bytes on the `codeStack` or the `storageStack` items these will also be changed in both state managers. However I don't think we ever edit raw bytes so this should not be a problem.

packages/common/src/state/simple.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+  setStateRoot(): Promise<void> {
+    throw new Error('Method not implemented.')
+  }
+  hasStateRoot(): Promise<boolean> {
+    throw new Error('Method not implemented.')
+  }
+
+  // Only goes for long term create situations, skip
+  async clearContractStorage(): Promise<void> {}
```
jochem-brouwer, line 164: Should also throw.

I just glanced in EVM, I see one call to this method, but I do not think this would ever actually "clear" any storage (however there might be some exotic cases which I have not thought of)

packages/statemanager/README.md
```diff
@@ -69,6 +69,20 @@ Caches now "survive" a flush operation and especially long-lived usage scenarios
...
 Have a loot at the extended `CacheOptions` on how to use and leverage the new cache system.
 
+### `SimpleStateManager`
+
+There is a dependency-free simple state manager implementation available shipped with the [@ethereumjs/common](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/common) package. While this state manager implementation lacks the implementations of some non-core functionality as well as proof related logic (e.g. `setStateRoot()`) it is suitable for a lot use cases where things like sophisticated caching or state root handling is not needed.
+
+This state manager can be used as follows and otherwise being used as shown in the basic example from above:
+
+```ts
+import { SimpleStateManager } from '@ethereumjs/common'
```
jochem-brouwer, line 79: I suggest we re-export `SimpleStateManager` from this package.

jochem-brouwer: However there is of course also an argument to not do this, to make users aware that the simple state manager lacks certain functionality.

packages/statemanager/src/cache/originalStorageCache.ts
```diff
@@ -4,6 +4,17 @@ import type { Address } from '@ethereumjs/util'
...
+/**
+ * Helper class to cache original storage values (so values already being present in
+ * the pre-state of a call), mainly for correct gas cost calculation in EVM/VM.
+ *
+ * TODO: Usage of this class is very implicit through the injected `getContractStorage()`
+ * method bound to the calling state manager. It should be examined if there are alternative
+ * designs being more transparent and direct along the next breaking release round.
+ *
+ * @deprecated This helper class has been moved to the @ethereumjs/common package
+ * and will be removed from this package along the next breaking release round.
```
jochem-brouwer, line 16: Can't we technically remove this here by re-exporting this `OriginalStorageCache` from common? :thinking:

jochem-brouwer (commented):
Some comments, LGTM in general (and indeed, also super simple yet elegant! :smile: )

For the state manager tests, some now run on the DefaultStateManager and the SimpleStateManager - but only the checkpoint tests. Are there other tests which we want to incorporate here?

Also, small side note, for code coverage the tests from StateManager -> Common will not get tracked, so our `common` coverage will likely go down, since likely (?) `SimpleStateManager` is not covered by the tests.

packages/common/src/state/simple.ts
```diff
@@ -0,0 +1,185 @@
...
+  async getContractCode(address: Address): Promise<Uint8Array> {
+    return this.topC().get(address.toString()) ?? new Uint8Array(0)
+  }
+
+  async putContractCode(address: Address, value: Uint8Array): Promise<void> {
+    this.topC().set(address.toString(), value)
+    if ((await this.getAccount(address)) === undefined) {
+      await this.putAccount(address, new Account())
+    }
+    await this.modifyAccountFields(address, { codeHash: keccak256(value) })
```
acolytec3, line 102: I do not have a super strong opinion on this but it seems like we should follow the same pattern we follow through the rest of the libraries where we allow users to pass in their own crypto and not require use of this `keccak256` function (like we do [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/d701dfacaa5654cfda05bf21faedda5a081debe5/packages/statemanager/src/stateManager.ts#L211)) in the `DefaultStateManager`.

acolytec3 (commented):
Awesome implementation, love the `simple`ness of this :-).  That said, I've read through the explanation of why this should be in `Common` and I'm not tracking.  This is only used in `evm` and nowhere else so doesn't make sense to me include this in `common`.   What is the driving goal for not putting this in `statemanager`?  If it implements the `EVMstateManagerInterface`, I don't see any reason that tree shaking wouldn't take care of any excess bundle size concerns if a user imports it from `@ethereumjs/statemanager`.

packages/common/src/state/simple.ts
holgerd77: I would like to not export publicly to not further enlarge the API surface we expose, then we have yet another three methods to care for. 🙂 

TBH I really do like the short names, since they are so unintrusive and I feel both from reading the methods itself as well as usages like:

```ts
this.codeStack.push(new Map(this.topC()))
```

It becomes pretty obvious what they do.

So I personally would like to keep.

Won't die on this hill though. 😂

packages/common/src/state/simple.ts
holgerd77: Ah, I think I can directly call `this.commit()` from within the constructor and put this code there.

Will try, then we can skip the extra method.

packages/common/src/state/simple.ts
holgerd77: Will do/integrate. First thought in terms of "simplification", but this might not be one actually. 😆

packages/common/src/state/simple.ts
holgerd77: Yeah, seem comment above.

packages/common/src/state/simple.ts
holgerd77: I battled a bit with this, will try your version.

packages/common/src/state/simple.ts
holgerd77: Ah, yeah, as you seemed to have noted somewhere below, these methods are called from within the EVM so they need to be callable.

General note: One thing that this state manager work further exposes is that our whole interface design around the state manger is not optimal, so what is generally in and what not, what is optional, around what principles the different interfaces are structured. This needs some refactoring along the breaking releases, likely together with the implementations itself. Will write up some issue with some evolved thinking about this at some point.

packages/common/src/state/simple.ts
holgerd77: > For the accountStack, if you edit the account on the original state manager, then it will also be edited on the copied state manager, I do not think that this is desired?

This is the definition of `shallowCopy()` and that's also the behavior from the `DefaultStateManager` and that's why we introduced this new name, to align with this behavior.

So if we want a full copy method, we should likely (?) call this `copy()` and do in addition. This would draw in interface changes though, so I would think we should not do here. My goal here was to get as close to the existing implementation as possible.

packages/common/src/state/simple.ts
holgerd77: Have answered above.

I would think that reasoning about what `clearContractStorage()` is doing when called from within the `EVM` and eventually do changes on the EVM side is outside the scope of this PR.

packages/statemanager/README.md
holgerd77: If we re-export, then we can as well directly integrate. So the whole initial point of this PR was to avoid that people need to draw in the somewhat heavy state manager dependency.

Guess this fully falls into the realm of the `Common` or `StateManager` (or `EVM`) discussion going on on Discord and we can take it from there.

packages/common/src/state/simple.ts
jochem-brouwer: Should we mark these `private`? If we decide to keep those as now (so those are `public` methods), then a description should be added, otherwise there are public methods `topS` with no description.

packages/statemanager/src/cache/originalStorageCache.ts
holgerd77: It's not clear if people are importing this from `statemanager` and things would otherwise break for them. Therefore only the deprecation for now.

This might not be a thing anyhow depending on how we decide on where to place the whole thing.

packages/common/src/state/simple.ts
jochem-brouwer: Ah ok, good point that these need to be callable, and also good point regarding the interface.

packages/common/src/state/simple.ts
jochem-brouwer: Ah right, I already knew I was confused, you are right regarding the shallow copy, so we can keep. We should indeed not deep copy.

packages/common/src/state/simple.ts
holgerd77: Ah, sorry, I always forget this "public as default" concept.

Yes, will do.

packages/common/src/state/originalStorageCache.ts
```diff
@@ -0,0 +1,57 @@
+import { bytesToUnprefixedHex } from '@ethereumjs/util'
+
+import type { Address } from '@ethereumjs/util'
```
gabrocheleau, line 3: nbd, but just pointing out that there is a more concise syntax for this: 

```suggestion
import { type Address, bytesToUnprefixedHex } from '@ethereumjs/util'
```

packages/common/src/state/simple.ts
```diff
@@ -0,0 +1,185 @@
...
+  EVMStateManagerInterface,
+  Proof,
+  StorageDump,
+  StorageRange,
+} from '../interfaces.js'
+import type { Address, PrefixedHexString } from '@ethereumjs/util'
+
+/**
+ * Simple and dependency-free state manager for basic state access use cases
+ * where a trie or verkle backed state manager is too heavy-weight.
```
gabrocheleau, line 17:
```suggestion
 * Simple and dependency-free state manager for basic state access use cases
 * where a merkle-patricia or verkle tree backed state manager is too heavy-weight.
```

packages/common/src/state/simple.ts
gabrocheleau: Personally not a fan of non-explicit names. At least to me it's not immediately obvious what they do, and I don't see the benefit of shortening/acronyms in general. Less of a concern since they're private though.

packages/common/src/state/simple.ts
gabrocheleau: I was going to suggest the same and then saw Andrew's comment. I agree that it would be preferable not to "hardcode" this.

packages/common/src/state/simple.ts
```diff
@@ -0,0 +1,185 @@
...
+    await this.modifyAccountFields(address, { codeHash: keccak256(value) })
+  }
+
+  async getContractCodeSize(address: Address): Promise<number> {
+    const contractCode = await this.getContractCode(address)
+    return contractCode.length
+  }
+
+  async getContractStorage(address: Address, key: Uint8Array): Promise<Uint8Array> {
+    return this.topS().get(`${address.toString()}_${bytesToHex(key)}`) ?? new Uint8Array(0)
```
gabrocheleau, line 111: Perhaps we could create a helper for `storageKeyFormatting` that parsed an address + key into the expected string? This could be helpful notably if we need to switch around the format (pre and post verkle comes to mind), then we don't have to modify it in multiple places.

packages/common/src/state/simple.ts
gabrocheleau: This could be even simpler as long as we know there are at least 2 elements in the stack:
```
this.accountStack.splice(-2, 1);

```

packages/statemanager/README.md
```diff
@@ -69,6 +69,20 @@ Caches now "survive" a flush operation and especially long-lived usage scenarios
 
 Have a loot at the extended `CacheOptions` on how to use and leverage the new cache system.
 
+### `SimpleStateManager`
+
+There is a dependency-free simple state manager implementation available shipped with the [@ethereumjs/common](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/common) package. While this state manager implementation lacks the implementations of some non-core functionality as well as proof related logic (e.g. `setStateRoot()`) it is suitable for a lot use cases where things like sophisticated caching or state root handling is not needed.
+
+This state manager can be used as follows and otherwise being used as shown in the basic example from above:
```
gabrocheleau, line 76:
```suggestion
This state manager can be used as follows and otherwise being used as shown in the basic example from above:
```
I don't understand this sentence

gabrocheleau (commented):
Wow, that's some great work! Really like how this removes the statemanager package dependency from the evm. Left some minor comments.

packages/common/src/state/originalStorageCache.ts
jochem-brouwer: Doesn't / shouldn't the linter do this? :thinking:

packages/common/src/state/simple.ts
holgerd77: Ok, generally totally on your side, so let's not make a because-of-the-context exception, will rename to something more expressive

holgerd77: (will go for `protected` for extensibility reasons)

packages/common/src/state/simple.ts
holgerd77: Ah, now I remember, I added this since I otherwise couldn't call the async `checkpoint()` method from the constructor.

I changed to a more expressive name (`checkpointSync()`), if there is a more elegant solution let me know.

packages/common/src/state/simple.ts
holgerd77: Updated with Gabriel's version, thanks 🙏 (I looked into splice() but didn't make it to the negative indexes 😂 )

packages/common/src/state/simple.ts
holgerd77: Yeah, there are various small optimizations/changes we can make to the interface which in sum will make them significantly more solid, so in these two cases e.g. it makes a lot of sense to make both methods optional in the interfaces and only call with TypeScript's '?'.

Small thing admittedly, but these kind of single things add up.

jochem-brouwer:
The git diff is looking messed up :thinking: It tries to re-import the changes to BLS?

packages/common/src/state/simple.ts
holgerd77: Yeah, but I honestly also really was confused first round to somewhat be forced to add the worse option here (we introduced `shallowCopy()` because for a full-blow level DB state it's just not realistic to have a full copy, but in most use cases a full copy would actually be *better*).

I guess what we can do (mid term) is that we add a `copy()` method to the EVM/VM additionally to the now `shallowCopy()` implementations which then checks if the current active `StateManager` and `Blockchain` instances have `copy()` methods (we would e.g. for this SM then also add an additional `copy()`) method and if they do `copy()` is done and if not an Error is thrown.

Guess that would be a pretty nice additional functionality for use cases where people *want* a full copy (of the EVM e.g.), e.g. to have isolated test runs or whatever.

packages/common/src/state/simple.ts
holgerd77: Have added

packages/common/src/state/originalStorageCache.ts
holgerd77: Yeah, I would like to stick to what the linter proposes (and my original version was what was proposed)

packages/common/src/state/simple.ts
holgerd77: Will keep for now, we can still easily do later on

packages/statemanager/README.md
holgerd77: Have clarified

holgerd77:
> The git diff is looking messed up 🤔 It tries to re-import the changes to BLS?

Yeah, I've cherry picked the commits out locally, will force-push

packages/evm/src/types.ts
```diff
@@ -279,7 +279,13 @@ export interface EVMOpts {
...
 
   /*
-   * The StateManager which is used to update the trie
+   * The EVM comes with a basic dependency-free `SimpleStateManager` implementation
+   * which serves most code execution use cases and which is included in the
+   * `@ethereumjs/common` package.
+   *
+   * The `@ethereumjs/statemanager` package provides a variety of state manager
+   * implementations for different needs (tree backed, RPC, experimental verkle)
+   * which can be used by this option as a replacement.
```
holgerd77, line 288: Can you please update the docs to now reflect the new SSM location?

holgerd77 (commented):
Looks otherwise good, feel free to merge (also ok to spare out this one doc change from my side, just saw it's really this one very mention, can also update later, not sure if worth to trigger whole CI)

packages/statemanager/src/simpleStateManager.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  async modifyAccountFields(address: Address, accountFields: AccountFields): Promise<void> {
+    let account = await this.getAccount(address)
+    if (!account) {
+      account = new Account()
+    }
+    account.nonce = accountFields.nonce ?? account.nonce
+    account.balance = accountFields.balance ?? account.balance
+    account.storageRoot = accountFields.storageRoot ?? account.storageRoot
+    account.codeHash = accountFields.codeHash ?? account.codeHash
```
jochem-brouwer, line 104: I just checked, even if `account` is returned `undefined`, and we provide the `accountFields` which is `Partial<Account>` this still works fine (we default the account to default values)

packages/statemanager/src/stateManager.ts
```diff
@@ -155,6 +160,11 @@ export interface DefaultStateManagerOpts {
  *
  * The default state manager implementation uses a
  * `@ethereumjs/trie` trie as a data backend.
+ *
+ * Note that there is a `SimpleStateManager` dependency-free state
+ * manager implementation available shipped with the `@ethereumjs/common`
```
jochem-brouwer, line 165: Docs change also here! Shipped by `@ethereumjs/statemanager` now

jochem-brouwer (commented):
3 small-ish comments

packages/statemanager/README.md
```diff
@@ -17,17 +17,20 @@ To obtain the latest version, simply require the project using `npm`:
...
 
 ### Introduction
 
 The `StateManager` provides high-level access and manipulation methods to and for the Ethereum state, thinking in terms of accounts or contract code rather then the storage operations of the underlying data structure (e.g. a [Trie](../trie/)).
 
-The library includes a TypeScript interface `StateManager` to ensure a unified interface (e.g. when passed to the VM), a concrete Trie-based `DefaultStateManager` implementation, as well as an `RPCStateManager` implementation that sources state and history data from an external JSON-RPC provider.
+This library includes several different implementations that all implement the `StateManager` interface which is accepted by the `vm` library. These include:
+
+- [`SimpleStateManager`](./src/simpleStateManager.ts) -a minimally functional (and dependency minimized) version of the state manager suitable for most basic EVM bytecode operations
+- [`DefaultStateManager`](./src//stateManager.ts) - a Merkle-Patricia Trie-based `DefaultStateManager` implementation that is used by the `@ethereumjs/client`
```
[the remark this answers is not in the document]
acolytec3: Indeed.  Updated.

packages/statemanager/src/simpleStateManager.ts
acolytec3: Is there a question here?  I'm not sure what you're getting at?

jochem-brouwer (approved):
LGTM

PR #3486, opened 2024-07-05
Add `trie.del`

acolytec3:
Adds `trie.del` as a simple extension of `trie.put`

- Adds a new `trie.del` function which wraps `trie.put`
- Add special handling in `trie.put` for the deleted leaf value
  1. If no leaf node exists at the specified key (i.e. `findPath` returns no leaf node), return early (since you can't delete a value in a leaf node that doesn't exist)
  2. Writes the special `VerkleLeafNodeValues.Deleted` when inserting the value into the leaf node.

### DIFF
--- a/packages/verkle/src/verkleTree.ts
+++ b/packages/verkle/src/verkleTree.ts
@@ -14,7 +14,7 @@ import { loadVerkleCrypto } from 'verkle-cryptography-wasm'
 import { CheckpointDB } from './db/checkpoint.js'
 import { InternalNode } from './node/internalNode.js'
 import { LeafNode } from './node/leafNode.js'
-import { type VerkleNode } from './node/types.js'
+import { VerkleLeafNodeValue, type VerkleNode, createDeletedLeafValue } from './node/types.js'
 import { decodeNode, isLeafNode } from './node/util.js'
 import {
   type Proof,
@@ -250,11 +250,28 @@ export class VerkleTree {
         )
       }
     } else {
+      if (equalsBytes(value, createDeletedLeafValue())) {
+        // Special case for when the deleted leaf value is passed to `put`
+        // You can't delete a value on a leaf node that doesn't exist
+        this.DEBUG &&
+          this.debug(`Leaf node with stem: ${bytesToHex(stem)} not found in trie`, ['DEL'])
+        return
+      }
       // Leaf node doesn't exist, create a new one
       leafNode = await LeafNode.create(stem, this.verkleCrypto)
       this.DEBUG && this.debug(`Creating new leaf node at stem: ${bytesToHex(stem)}`, ['PUT'])
     }
     // Update value in leaf node and push to putStack
+    if (equalsBytes(value, createDeletedLeafValue())) {
+      // Special case for when the deleted leaf value is passed to `put`
+      // Writing the deleted leaf value to the suffix indicated in the key
+      this.DEBUG &&
+        this.debug(
+          `Deleting value at suffix: ${suffix} in leaf node with stem: ${bytesToHex(stem)}`,
+          ['DEL']
+        )
+      leafNode.setValue(suffix, VerkleLeafNodeValue.Deleted)
+    }
     leafNode.setValue(suffix, value)
     this.DEBUG &&
       this.debug(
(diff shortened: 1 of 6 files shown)


packages/verkle/src/verkleTree.ts
```diff
@@ -250,11 +250,28 @@ export class VerkleTree {
         )
       }
     } else {
+      if (equalsBytes(value, createDeletedLeafValue())) {
```
gabrocheleau, line 253: Should this not throw an error instead of silently returning?

packages/verkle/test/verkle.spec.ts
```diff
@@ -211,7 +211,7 @@ describe('Verkle tree', () => {
     assert.deepEqual(val2, hexToBytes(values[2]), 'confirm values[2] can be retrieved from trie')
   })
 
-  it('should put values and find them', async () => {
+  it('should put, find, delete, and the put values (again)', async () => {
```
gabrocheleau, line 214: I'm not sure I get the part that says `and the put values (again)`. Could you clarify?

gabrocheleau: From looking at the test I adjusted the description of the test, feel free to adjust.

packages/verkle/test/verkle.spec.ts
```diff
@@ -211,7 +211,7 @@ describe('Verkle tree', () => {
     assert.deepEqual(val2, hexToBytes(values[2]), 'confirm values[2] can be retrieved from trie')
   })
```
gabrocheleau, line 213: We could also add a test for the expected behavior when attempting to delete a value that does not exist? Left a related comment in the other file (i.e. should this throw or return?)

gabrocheleau (commented):
Nice work, cool to see so much progress on the verkle tree front lately! 
Left some comments, can be merged once addressed.

packages/verkle/src/verkleTree.ts
acolytec3: Now that I've re-read the spec, I'm not sure what the expected behavior is.  If it's just writing `0` to the trie, then this behavior is actually invalid and we should just let it create a new leaf node and write what I'm calling the `deleted leaf value` here.  Will ask kev

packages/verkle/test/verkle.spec.ts
acolytec3: Above is fine.  I just wanted to make sure that "deleting" and then putting in the same spot again still worked correctly.

packages/verkle/src/verkleTree.ts
acolytec3: My instinct here is that we should actually just create a new leaf node and write zeroes to it.  There is no "deletion" in verkle tries so writing zeroes to a leaf value is the same as "deleting" a value in the trie.  Will update this section of code accordingly (i.e. just remove this special handling and allow `trie.del to write zeroes)

packages/verkle/src/verkleTree.ts
```diff
@@ -255,6 +255,16 @@ export class VerkleTree {
...
+      // Special case for when the deleted leaf value is passed to `put`
+      // Writing the deleted leaf value to the suffix indicated in the key
+      this.DEBUG &&
+        this.debug(
+          `Deleting value at suffix: ${suffix} in leaf node with stem: ${bytesToHex(stem)}`,
+          ['DEL']
+        )
+      leafNode.setValue(suffix, VerkleLeafNodeValue.Deleted)
+    }
     leafNode.setValue(suffix, value)
```
gabrocheleau, line 268: Isn't line 268 overwriting what has been done in line 266 within the if block?

acolytec3: Indeed.  Will address in the next commit.

acolytec3:
> One general Q. We currently return `undefined` for both VerkleNodeLeafValues (Untouched and Deleted) when calling `getValue` on a leaf node. Is this what we want? Are there ever cases where blurring the distinction between Ontouched & Deleted would be problematic (in which case we might want to return undefined & null respectively, although I know you're not a big fan of treating those differently?)

Good point.  We should returns an array of zeroes for the "deleted" case.  This comes back to the underlying point that there really is no concept of "deleted" in a verkle trie.  There's just untouched, and then touched (with maybe a special flag for zeroes).  For the moment, I think it makes the most sense to update `trie.get` to return an array of 0s unless the leaf value requested is untouched.

gabrocheleau:
> > One general Q. We currently return `undefined` for both VerkleNodeLeafValues (Untouched and Deleted) when calling `getValue` on a leaf node. Is this what we want? Are there ever cases where blurring the distinction between Ontouched & Deleted would be problematic (in which case we might want to return undefined & null respectively, although I know you're not a big fan of treating those differently?)
> 
> Good point. We should returns an array of zeroes for the "deleted" case. This comes back to the underlying point that there really is no concept of "deleted" in a verkle trie. There's just untouched, and then touched (with maybe a special flag for zeroes). For the moment, I think it makes the most sense to update `trie.get` to return an array of 0s unless the leaf value requested is untouched.

Makes sense! Thanks for addressing.

gabrocheleau (dismissed):

gabrocheleau (approved):

Mikhai56 (approved):

packages/verkle/src/verkleTree.ts
Mikhai56: Ok

Issue #3487, opened 2024-07-09
Monorepo -> Tree Shaking Refactor: Replace Static Class Constructors


holgerd77:
Part of #3446 

**Earliest start together with official breaking release work!**

As stated by Will in the original issue static constructors of classes cannot be tree-shaked out, see e.g. this Rollup issue https://github.com/rollup/rollup/issues/349 which was closed at the end with "won't fix" after a long discussion with various explanations on why JavaScripts dynamic nature makes it difficult to realize.

We have static constructors all over the place and for e.g. block or tx I looked through the code and the code from these constructors make up up to 30-50% of the overall class code base, e.g. for the various `from*` constructors for the tx types. So while this is - a bit - tedious for users and one of the more invasive changes, I would make a strong pledge that we should adopt here. For users this is a one-time simple search-and-replace adjustment and this would significantly improve our structural setup.

Solution here would be to simply do normal/simple static top level methods replacing the constructors, guess we can go with a naming scheme just simply "switching around the wording a bit":

- `1559FeeMarketTx.fromValuesArray()` -> `create1559TxFromValuesArray()` (not sure if to keep `FeeMarket` or not)
- `Block.fromBlockData()` -> `createBlockFromBlockData()`
- `Block.fromExecutionPayload()` -> `createBlockFromExecutionPayload()`

So these two libraries are the super obvious ones.

The main `create()` constructors for (more or less) all libraries might deserve a second round of thinking, but on first round I would think: why not?

So:

- `EVM.create()` -> `createEVM()`
- `Blockchain.create()` -> `createBlockchain()`

And so on.

Guess we can just do a first round PR on Block/Tx and then we already have some more in-depth view if there might be some side-effects we didn't anticipate.

A monorepo-wide search on the words `public static` close to plainly reveals all static method instances.

Guess we can actually replace them all, since there is simply no benefit apart form this syntactic sugar and at the same time having these concrete downsides.

Some more candidates: `AccessLists` in tx, `Withdrawals`, `Requests`, `Accounts`, `Wallet` constructors (why not? not much additional work.)

jochem-brouwer:
Monorepo-wide naming conventions: (from: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3524#pullrequestreview-2190327056)

- `createAccountFromAccountData()` -> `createAccount()`
- `create1559FeeMarketTxFromTxData()` -> `create1559FeeMarketTx()`
[...]

PR #3489, opened 2024-07-10
Block: replace static constructors

ScottyPoi:
Works on #3487

Replaces **all**  static Block methods with independently defined functions

### Block constructor methods:
- Block.fromBlockData
- Block.fromRLPSerializedBlock
- Block.fromValuesArray
- Block.fromRPC
- Block.fromJsonRPCProvider
- Block.fromExecutionPayload
- Block.fromBeaconPayloadJson

moved to `packages/block/src/blockConstructor.ts` and renamed:

- createBlockFromBlockData
- createBlockFromRLPSerializedBlock
- createBlockFromValuesArray
- createBlockFromRPC
- createBlockFromJsonRPCProvider
- createBlockFromExecutionPayload
- createBlockFromBeaconPayloadJson

### Trie Root methods:

- Block.genWithdrawalsTrieRoot
- Block.genTransactionsTrieRoot
- Block.genRequestsTrieRoot

moved to `packages/block/src/helpers.ts` and renamed: 

- genWithdrawalsTrieRoot
- genTransactionsTrieRoot
- genRequestsTrieRoot


ScottyPoi:
1 statemanager test failing with this message:
```
 FAIL  test/rpcStateManager.spec.ts [ test/rpcStateManager.spec.ts ]
Error: Missing "./dist/cjs/helpers.js" specifier in "@ethereumjs/block" package
```

packages/block/src/blockConstructor.ts
```diff
@@ -0,0 +1,442 @@
+import { RLP } from '@ethereumjs/rlp'
```
holgerd77, line 1: I would like to establish a general naming convention for this file name that we re-use throughout all libraries, so that it's just clear where to look for, same a bit as our `types.ts` files.

Can we take `constructors.ts` here in this case? That would be my pledge! 🙂 

So without `block` (redundant, I would also pledge for just always having one file for this per library), and also using plural (`constructors`), since several constructors (let's always use this name, and not fall back to singular if for the moment there is only one constructor present for a library).

holgerd77: (so these comments go more or less 1:1 for the other PR for blockchain)

packages/block/src/block.ts
```diff
@@ -82,455 +56,6 @@ export class Block {
...
-    payload: BeaconPayloadJson,
-    opts?: BlockOptions
-  ): Promise<Block> {
-    const executionPayload = executionPayloadFromBeaconPayload(payload)
-    return Block.fromExecutionPayload(executionPayload, opts)
-  }
-
   /**
    * This constructor takes the values, validates them, assigns them and freezes the object.
    * Use the static factory methods to assist in creating a Block object from varying data types and options.
```
holgerd77, line 61: One thing which has gotten a bit worse with this extraction is discoverability, so it has gotten harder to find/auto-complete what constructors are available.

Can we make it a habit to therefore to always provide a list with all the (linked, so clickable) static constructor methods, either here or along the class (`Block`) code docs (can't fully decide, where does one stumble more upon this? If one imports a `Block` and then hovers over the class import or if one types in `new Block()` and then sees the code docs for the constructor? Guess maybe the class docs?).

So that would be great and pretty helpful!

Also: can you add a `@deprecated` note here for this constructor as well, like we have for other constructors (tx? evm?) already?

holgerd77 (commented):
Ok, had a look, great, some things around general conventions for this kind of PRs, otherwise I guess we can also already take this in, had some closer look at the EVM changes, these are actually not that vast so I guess merging won't be as much as a problem as I anticipated.

There are 3 merge conflicts in this PR to fix though first.

holgerd77:
Ah, and one other thing: I would also find it great if we make it a habit to always post a comparable pre- and post-PR bundle size respectively the diff, so that we get a feeling about the changes and have this "measured".

Won't be dramatically big here, but nevertheless.

Can you maybe do this along the instructions from my presentation (slide 5 😋), and so maybe do a super-small script creating a block using one of the constructors (best the simplest to use one) with the old `Block.from` and see how the bundle size goes (so: pointing in the - now called - `vite.config.bundle.ts` file in EVM to your new small script and then run `npm run visualize:bundle`) and then post the numbers here? 🙂 

Guess this will generally be pretty motivating!

packages/block/src/block.ts
ScottyPoi: I agree.  It is a nicer experience to type in `Block.` and have the constructors pop up.

holgerd77 (approved):
LGTM

PR #3493, opened 2024-07-11
EVM: Vite visualize:bundle Adjustments (maintenance-v8)

holgerd77:
Taking over a subset of #3492 (Vite visualize:bundle adjustments) to prevent release corruption over on `maintenance-v8`.


acolytec3 (approved):

holgerd77:
Ah, this is still from the old setup of required CI jobs, vaguely remember that we once switched over with something. Will see if I can easily adjust to current `master` setup.

PR #3497, opened 2024-07-12
Monorepo: Set sideEffects to false for all Packages

holgerd77:
I've now read a bit on side effects [here](https://sgom.es/posts/2020-06-15-everything-you-never-wanted-to-know-about-side-effects/) and have gotten some base feeling for potential situations.

My overall stance here is that it is...

## a) generally unlikely that we have side effects in our libraries.

Putting something in the global scope which then constitutes the state for some other library is generally a pattern we try to fully avoid. And if something is passed on via an object (`Common` being the most prominent one), this is then no problem again. The one thing which might came closest is this global KZG initialization we had before we discoverd ourself that this bad (also for other reasons) and where we switched to this `Common.customCrypto`-encapsuled way of passing things around. If I understand side effects correctly I would also think that the `debug` usage does not apply here. It's not that we re-use a `debug` object e.g. in `statemanager` which we before globally instantiated in `util` (or whatever).

## b) chances are high that we discover over time.

Other thing is: we will now have quite some time playing around with the libraries and the changes. No production releases in sight. We we also do alpha/dev releases where people can test things. So for the unlikely chance that we'll get a hidden side effect missed it is pretty likely that this will be discovered along the way.

I will nevertheless go one-by-one library here. Not necessarily by pushing, but at least along switching and give every library some quick thought if there might be mechanisms/code parts in which could be sensitive regarding introducing some side effect risk.

### RLP

Generally do not think so. Might the `bin` folder need some thought here?

### Util

Looked once over the modules we have there, do not think so.

### Common

Full no. All encapsulated in the `Common` main class.

### Tx

Again a no. Tx classes or TxFactory class.

### Trie

No. Trie class.

### Block

Noddeldy noodeldy no. Block/header class.

### Blockchain

No. Blockchain class.

### Verkle

No.

### StateManager

No.

### Wallet

No, Wallet class.

### Genesis

Would not think so.

### Devp2p

A bit more complex, but also would not think so.

### Ethash

No. Ethash object/class.

### EVM

Also here: all in EVM I would say.

### VM

No.

### Client

I will leave client out for now. Not in a library mode for now. We might want to switch on later on.

Ok, that's actually it. 🙂 

Will now push with all the above enabled.

Guess we can merge if CI passes!

### DIFF
--- a/packages/block/package.json
+++ b/packages/block/package.json
@@ -17,6 +17,7 @@
   "license": "MPL-2.0",
   "author": "mjbecze (mb@ethdev.com)",
   "type": "module",
+  "sideEffects": false,
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
   "exports": {
--- a/packages/blockchain/package.json
+++ b/packages/blockchain/package.json
@@ -17,6 +17,7 @@
   "license": "MPL-2.0",
   "author": "mjbecze <mjbecze@gmail.com>",
   "type": "module",
+  "sideEffects": false,
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
   "exports": {
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -27,6 +27,7 @@
     }
   ],
   "type": "module",
+  "sideEffects": false,
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
   "exports": {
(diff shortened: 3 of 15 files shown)


holgerd77:
Ah, and this even safed us 1/2 a KB for the StateManager. 😂 

At least that's I proof I guess that the change *did something* and the setting works. That's something.

Before:

grafik

After:

grafik

jochem-brouwer:
I think for RLP it is fine also, the `bin` is just a script so I think we can safely also set `sideEffects` there.

acolytec3 (approved):

PR #3498, opened 2024-07-12
Fix Prague test runner

jochem-brouwer:
This PR solves some bugs in https://github.com/ethereum/execution-spec-tests/releases/tag/devnet-1%40v1.3.0

Note: some bugs were in the state runner itself, others were in the VM, mainly the one in `requests.ts`.

A question we might have to answer here is if our 7702 txs are too strict on loading values, for instance an `yParity` of `0x00` will not be accepted (leading zeros in the hex string) and neither will a `v` value (old alias for `yParity`).

This solves both the state and blockchain tests.

Note: also ensures the correct EIPs are on Prague's common


acolytec3 (approved):
LGTM.  Ran all the devnet-1 tests and everything passes

jochem-brouwer:
I might need to dive a bit further into this point:

> A question we might have to answer here is if our 7702 txs are too strict on loading values, for instance an yParity of 0x00 will not be accepted (leading zeros in the hex string) and neither will a v value (old alias for yParity).

We use the `v` var for the other tx types :thinking:

Issue #3448, opened 2024-06-10
Common: Separate Chain configs for Tree Shaking


holgerd77:
Part of #3446 

**Note: This issue is breaking, please do not start until breaking release work has officially started!**

Picking up on the following from @roninjin10 :

> ### Individual chains and hardforks are not tree shakeable
> 
> Problem: many uses of Ethereumjs common will use a single chain or a single hardfork. But all chain or hardfork specific data is not tree shakable Solution: Seperate all chain data into a single chain config
> 
> This is IMO the biggest opportunity the common package has. [example](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/chains.ts), another [example](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/common.ts#L112)
> 
> There are many solutions as long as the chain data ends up seperate (never on same class, never on same object). One is an abstract class
> 
> ```ts
> export abstract class Common {
>   abstract readonly chainId: number
>   ...
> }
> 
> // Then seperately
> 
> export class Optimism extends Common {
>   chainId: 10
>   ...
> }
> 
> export class Mainnet extends Common {
>   chainId: 1
>  ...
> }
> ```

Ok, I had some somewhat larger battle with this and went through several iterations of thinking if feasible or not 😋 . Here is my latest stance, happy to further discuss.

### Hardforks

This is basically all config values from [hardforks.ts](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/hardforks.ts). There is heavy one-way dependency here and values are taken throughout all the HF configurations, analyzing the changes. I do not see enough value here respectively too much effort respectively trade-offs to be done to justify to tear this apart, unless someone comes up with a very brilliant and non-invasive idea.

What we can and very much should do is simplify the structure overall from:

```ts
dup: {
  v: 3,
  d: 'Base fee of the DUP opcode',
}
```

To:

```ts
dup:  3, // Base fee of the DUP opcode
}
```

Im pretty sure no-one is actively using these comments programmatically (even had a quick look at https://www.evm.codes/ and their mouseovers and the like). Not sure about the tree shaking effect here (so these `d` values might (?) even be tree shaked out, but apart from that the above change will already cut the code size of the whole thing basically into half.

### Chains

The main thing to consider here: Common is deeply into integrated into our code base respectively the most exposed auxiliary code part we have in the library suite and code like the following is all over the place:

```ts
[...]

PR #3500, opened 2024-07-12
common: move description string from being an object field to a comment

scorbajio:
This change is inspired by #3448 and refactors the following files by removing any description string fields and instead including them as comments to reduce code size:
* `eips.ts`
* `hardforks.ts`

### DIFF
--- a/packages/common/src/common.ts
+++ b/packages/common/src/common.ts
@@ -414,7 +414,7 @@ export class Common {
       (this._paramsCache as any)[topic] !== undefined &&
       (this._paramsCache as any)[topic][name] !== undefined
     ) {
-      value = (this._paramsCache as any)[topic][name].v
+      value = (this._paramsCache as any)[topic][name]
     }
     return BigInt(value ?? 0)
   }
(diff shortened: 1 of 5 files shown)


holgerd77 (commented):
Ah, one step further please, so the `v` (+ the surrounding dict) should be killed completely! 🙂

holgerd77:
Can you also please post a set of pre and post work numbers here so that we always have these comparison values in the PRs themselves? (Simply the two files sizes in KB pre/post should be enough in this case I guess)

scorbajio:
It seems like some tests that directly access the `v` field are failing, will clean those up, but the refactor seems to be successful, with the following results for the following test code:
```

const c = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.Prague })
console.log(c.chainId())
```
Before: out1.mjs  162.0kb
After: out1.mjs  187.4kb

I wasn't able to import the `EIPs` dictionary since it's not being exported in the index file, although since it's being accessed indirectly by other imports, results should still include the size reduction it's gone through (although not entirely sure on this, may need confirmation).

scorbajio:
I wasn't able to locally recreate any of the failing vm-api or test-wallet tests, so I just triggered a rerun to double check the CI results.

holgerd77 (approved):
Nice, cool, that we have a fix here!

Will merge.

PR #3501, opened 2024-07-12
tx/common: remove yParity from tx json interface

jochem-brouwer:
This removes the alias `yParity` from the `tx` library, I introduced it for 7702 (thinking that we also used this on other txs - but we don't). This removes the alias, since there was (before this PR) either the value `v` in non-7702 txs, or `yParity` on 7702 txs.

Note: the tests fixed in https://github.com/ethereumjs/ethereumjs-monorepo/pull/3498 still pass (just checked).


holgerd77 (approved):
LGTM

ScottyPoi (dismissed):
LGTM!

can confirm that tests fixed in #3495 still pass

jochem-brouwer:
Have marked to draft, did not read the spec too well, it seems for txs in RPC we should indeed have `yParity`. For 4844, 1559, 2930, (7702?) this field is mandatory, but these txs can optionally have a `v` field. Note: for Legacy txs the `v` field is mandatory and `yParity does not seem to be an option. Source: https://ethereum.github.io/execution-apis/api-documentation/ (check `eth_getTransactionByHash`)

jochem-brouwer:
I will close this PR, since this actually degrades matching the spec instead of increasing it, thanks for catching this @ScottyPoi :smile: :+1:

PR #3503, opened 2024-07-14
Client: fix more hive RPC-Compat tests

jochem-brouwer:
To run, setup local hive with EthereumJS

1. Ensure in hive, `clients/ethereumjs` has a cloned `ethereumjs-monorepo`
2. Ensure that in `clients`, you rename `Dockerfile.local` to `Dockerfile`
3. To run: `./hive --sim ethereum/rpc --client ethereumjs --sim.limit rpc-compat`

Before: 16/90 failed, after: 13/90 failed

Side note: in order to run EthereumJS in hive with TS-node from the source, change `ethereumjs-local.sh`:

`ethereumjs="node /ethereumjs-monorepo/packages/client/dist/esm/bin/cli.js"`

To

`ethereumjs="tsx /ethereumjs-monorepo/packages/client/bin/cli.ts"`

And ensure that in `Dockerfile` (the renamed `Dockerfile.local`) you install `tsx`:

```
RUN npm i -g ts-node
RUN npm i -g tsx
```

**Failing tests overview**

- `eth_call/call-callenv` - the test expects us to return 0 on the `BASEFEE` opcode it seems, but this does not seem correct to me
- `eth_call/call-contract` - tests calls into a contract which loads the `CALLDATASIZE`, and if its nonzero it reverts (it seems to push some string on the stack). The test however expects us to return a completely different value (and I do not understand why)

I investigated the other tests but it is not clear why those fail to me (or they are simply too complex to implement, for instance decode the ABI revert data into human readable strings)

### DIFF
--- a/packages/client/src/rpc/modules/eth.ts
+++ b/packages/client/src/rpc/modules/eth.ts
@@ -507,6 +507,7 @@ export class Eth {
       gasPrice: toType(gasPrice, TypeOutput.BigInt),
       value: toType(value, TypeOutput.BigInt),
       data: data !== undefined ? hexToBytes(data) : undefined,
+      block,
     }
     const { execResult } = await vm.evm.runCall(runCallOpts)
     return bytesToHex(execResult.returnValue)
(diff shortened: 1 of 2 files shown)


holgerd77 (approved):
Nice, LGTM

PR #3504, opened 2024-07-15
Blockchain: More Modern and Flexible Consensus Layout / Tree Shaking Optimization

holgerd77:
This PR moves to a more flexible consensus layout for blockchain allowing for an optional `consensusDict` to be passed in where all consensus objects which should be used for consensus validation must be present.

This basic setup allows for full tree shaking regarding Clique and Etash code and it will - step 2 - also allow to remove the Ethash dependency.

Since we are now in a Proof-of-Stake world where consensus validation is the exception the PR also moves to having consensus validation deactivated by default. Semantics of the `consensusValidation` flag is "merged" with the `consensusDict` object ("consensus" and "consensus validation" is basically the same), this allows for a substantial simplification on otherwise useless "combinations" (this and that object passed, but `consensusValidation` set to `false`) and also have a lot cleaner deactivated-by-default structure.

Deactivated-by-default also allows for removing a lot of `consensusValidation: false` passing, which we lately did more and more on tests since we just didn't want to have/do any more.

Consensus transitions - if wished - still fully work, as a side note (even more flexible than before).

WIP, several tests failing. But on the way.

### DIFF
--- a/packages/blockchain/src/blockchain.ts
+++ b/packages/blockchain/src/blockchain.ts
@@ -380,7 +380,7 @@ export class Blockchain implements BlockchainInterface {
         }
 
         if (this._validateConsensus) {
-          if (!(this.common.consensusAlgorithm() in this._consensusDict)) {
+          if (this.consensus === undefined) {
             throw new Error(
               `Consensus object for ${this.common.consensusAlgorithm()} consensus not available in (see consensusDict option)`
             )
--- a/packages/blockchain/src/blockchain.ts
+++ b/packages/blockchain/src/blockchain.ts
@@ -77,6 +77,7 @@ export class Blockchain implements BlockchainInterface {
   public readonly common: Common
   private _hardforkByHeadBlockNumber: boolean
   private readonly _validateBlocks: boolean
+  private readonly _validateConsensus: boolean
   private _consensusDict: ConsensusDict
 
   /**
@@ -110,6 +111,7 @@ export class Blockchain implements BlockchainInterface {
 
     this._hardforkByHeadBlockNumber = opts.hardforkByHeadBlockNumber ?? false
     this._validateBlocks = opts.validateBlocks ?? true
+    this._validateConsensus = opts.validateConsensus ?? true
     this._customGenesisState = opts.genesisState
 
     this.db = opts.db !== undefined ? opts.db : new MapDB()
@@ -375,7 +377,9 @@ export class Blockchain implements BlockchainInterface {
           await this.validateBlock(block)
         }
 
-        await this.consensus?.validateConsensus(block)
+        if (this._validateConsensus) {
+          await this.consensus?.validateConsensus(block)
+        }
 
         // set total difficulty in the current context scope
         if (this._headHeaderHash) {
--- a/packages/blockchain/src/blockchain.ts
+++ b/packages/blockchain/src/blockchain.ts
@@ -14,7 +14,6 @@ import {
   equalsBytes,
 } from '@ethereumjs/util'
 
-import { CasperConsensus } from './consensus/index.js'
 import {
   DBOp,
   DBSaveLookups,
@@ -120,7 +119,6 @@ export class Blockchain implements BlockchainInterface {
     this.events = new AsyncEventEmitter()
 
     this._consensusDict = {}
-    this._consensusDict[ConsensusAlgorithm.Casper] = new CasperConsensus()
 
     if (opts.consensusDict !== undefined) {
       this._consensusDict = { ...this._consensusDict, ...opts.consensusDict }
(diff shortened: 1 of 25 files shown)


holgerd77:
Ah, and to say: so this *does* complete on its goals. So `ethash` dependency is fully removed from Blockchain (together with its problematic [bigint-crypto-utils](https://github.com/juanelas/bigint-crypto-utils) dependency and tree shaking is done for `EthashConsensus` and `CliqueConsensus`.

Before:

grafik

(blockchain size + ethash still present)

After:

grafik

packages/blockchain/src/consensus/casper.ts
```diff
@@ -20,11 +18,6 @@ export class CasperConsensus implements Consensus {
 
   public async validateConsensus(): Promise<void> {}
 
-  public async validateDifficulty(header: BlockHeader): Promise<void> {
-    if (header.difficulty !== BIGINT_0) {
-      const msg = 'invalid difficulty.  PoS blocks must have difficulty 0'
-      throw new Error(`${msg} ${header.errorStr()}`)
-    }
-  }
+  public async validateDifficulty(): Promise<void> {}
```
jochem-brouwer, line 21: Where is the `difficulty` must be 0 check now?

packages/blockchain/src/blockchain.ts
```diff
@@ -390,7 +380,13 @@ export class Blockchain implements BlockchainInterface {
         }
 
         if (this._validateConsensus) {
-          await this.consensus.validateConsensus(block)
+          if (!(this.common.consensusAlgorithm() in this._consensusDict)) {
```
jochem-brouwer, line 383: For cleani-ness should this not use `this.consensus()`?

packages/blockchain/test/customConsensus.spec.ts
```diff
@@ -150,21 +153,5 @@ describe('consensus transition checks', () => {
         `checkAndTransitionHardForkByNumber should not throw with custom consensus, error=${err.message}`
       )
     }
-
-    blockchain.consensus = new EthashConsensus()
-    blockchain.common.consensusAlgorithm = () => 'fibonacci'
-
-    try {
-      await blockchain.checkAndTransitionHardForkByNumber(5n)
```
jochem-brouwer, line 158: This part of the test got deleted, is this correct?

jochem-brouwer (commented):
This looks super clean, some questions though

packages/blockchain/src/consensus/casper.ts
holgerd77: Hey there,
thanks a lot for the review, all good points I guess, will address one by one:

So this one originated from a former more radical version, where I fully deleted the `CasperConsensus`, and as an in-between step deleted this difficulty check. This is not really a consensus check but just a simple format check and is in fact also done in the BlockHeader: https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/block/src/header.ts#L502

I think it is beneficial to remove it here since it otherwise "tricks" us into thinking we *would* do some kind of consensus validation here for PoS (must be, there is some code! 😋 ), while in reality this is not the case (PoS consensus validation is done on the CL side).

No super strong opinion, but I would tend to keep this removed.

packages/blockchain/src/blockchain.ts
holgerd77: You are right, that works as well 🙂, will update!

packages/blockchain/src/consensus/casper.ts
jochem-brouwer: I should have checked upon reviewing because I suspected this check was in our Block Header as well (this would actually be the more logical place!). I agree it should not be in blockchain. This is actually a great change/removal, let's keep it :smile: :+1:

packages/blockchain/test/customConsensus.spec.ts
holgerd77: Could readd by making the checks more consistent.

packages/blockchain/test/pos.spec.ts
```diff
@@ -89,29 +88,6 @@ describe('Proof of Stake - inserting blocks into blockchain', () => {
         BigInt(1313601),
         'should have calculated the correct post-Merge total difficulty'
       )
-
-      const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.London })
-      const powBlock = createBlockFromBlockData(
-        {
-          header: {
-            number: 16,
-            difficulty: BigInt(1),
```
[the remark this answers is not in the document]
holgerd77: Ok.

Needed to think three times (or so 😋) about this.

So this is where this actually interflicts with our removed difficulty check.

So: a PoW block for Londond is created: works. Block is added to Blockchain: now (with removed check) works as well.

I wanted to suggest that we instead do a hardfork check on the Commons (block + blockchain).

This gets hairy quickly though (might still be the better thing to do), since we have this `_hardforkByHeadBlockNumber` setting in Blockchain and this generally reaches deep into the blockchain update code.

I will capitulate here for now and re-add the difficulty check for now, together with a comment that this is not really a consensus check. That should enable to re-activate the test case.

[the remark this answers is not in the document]
holgerd77: Yes, works (locally), and I guess for this PR it is ok (maybe: better), since this PR is about something else at the end.

But to admit, this difficulty check is just picking out a very selected case and this problem that it should not be possible to add a non-HF matching block (with eventually different formatting or properties) to a blockchain is a lot broader and should likely be solved on a more generic level.

jochem-brouwer (approved):
LGTM!

Issue #3505, opened 2024-07-15
Custom common: support for new / different params


jochem-brouwer:
It currently does not seem possible to create a custom common where one can create / override params. This would be a nice feature, for instance for chains which have different gas prices for certain opcodes.

PR #3506, opened 2024-07-15
Common: allow overriding params

jochem-brouwer:
Closes https://github.com/ethereumjs/ethereumjs-monorepo/issues/3505

This PR expands the types in Common such that one can override params in hardforks.

An example is given in the 2935 test on how to use it.

### DIFF
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -39,6 +39,11 @@ export interface ChainConfig {
   dnsNetworks?: string[]
   consensus: ConsensusConfig
   depositContractAddress?: PrefixedHexString
+  params?: {
+    [topic: string]: {
+      [param: string]: number | bigint
+    }
+  }
 }
 
 // TODO: Remove the string type and only keep PrefixedHexString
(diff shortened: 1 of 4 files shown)


packages/common/src/types.ts
```diff
@@ -196,6 +201,22 @@ export type HardforkConfig = {
...
+  pow?: {
+    [key: string]: number | bigint | null
+  }
+  sharding?: {
+    [key: string]: number | bigint | null
+  }
+  vm?: {
+    [key: string]: number | bigint | null
+  }
 } & EIPOrHFConfig
```
holgerd77, line 220: I think you don't need to do this, this is already in through `&EIPOrHFConfig`

jochem-brouwer: Hey, you are right, I think this got introduced in #3500

holgerd77 (commented):
Can we maybe have a test case for the `Common` part? Or do we already have one?

jochem-brouwer:
I will add a test case!

holgerd77 (approved):
Cool, thanks, LGTM!

PR #3507, opened 2024-07-15
Replace stream usage with async iterators & pipelines

scorbajio:
This change continues the work started in #3231 but with a different approach discussed in [this article](https://darkmannn.dev/posts/nodejs-in-2023-streams-streams-vs-async-generators#replace-a-readable-stream). The idea is to use async iterators and [the Nodejs `pipeline` function](https://nodejs.org/api/stream.html#streampipelinesource-transforms-destination-callback) to replace standard stream usage and phase out the dependency. Also see #3216 for a list of breaking changes being proposed/pursued.

### DIFF
--- a/packages/block/test/eip4844block.spec.ts
+++ b/packages/block/test/eip4844block.spec.ts
@@ -1,4 +1,4 @@
-import { Chain, Common, Hardfork } from '@ethereumjs/common'
+import { Chain, Common, Hardfork, createCommonFromGethGenesis } from '@ethereumjs/common'
 import { BlobEIP4844Transaction } from '@ethereumjs/tx'
 import {
   blobsToCommitments,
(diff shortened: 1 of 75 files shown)


holgerd77:
🙏 

Thanks for trying!

Can you already say if you have a somewhat good feeling with this approach? Or is it too early too ask?

scorbajio:
> 🙏
> 
> Thanks for trying!
> 
> Can you already say if you have a somewhat good feeling with this approach? Or is it too early too ask?

Yes, it seems like the approach is workable for us. I've got the tests passing in the trie package, and I'll come back to clean this up today. I may need a second look at how I'm typing things, but otherwise things seems to be working.

holgerd77:
> > 🙏
> > Thanks for trying!
> > Can you already say if you have a somewhat good feeling with this approach? Or is it too early too ask?
> 
> Yes, it seems like the approach is workable for us. I've got the tests passing in the trie package, and I'll come back to clean this up today. I may need a second look at how I'm typing things, but otherwise things seems to be working.

That would be pretty cool! 🤩

acolytec3:
This should be superseded by #3519.  @scorbajio please reopen if you feel otherwise.

PR #3510, opened 2024-07-16
refactor: restrict PrefixedHexString types

gabrocheleau:
This PR addresses one of the points from #3216, namely:

- Go through the bytes module, and related functions that accept both PrefixedHexString | string as types, and restrict the type to only accept PrefixedHexString.


holgerd77 (approved):
Nice, LGTM! 🙂

Issue #3488, opened 2024-07-09
EVM/Common -> Tree Shaking Refactor: Separate Common EVM Gas Prices from other Params


holgerd77:
Part of #3446 

**Earliest start together with official breaking release work!**

Common remains to be a thing with its 179 KB in size, which is a lot for a configuration library, also referencing the other Common issue #3448 here along. So it will be hard to substantially bring the 300 KB or so from Tx - other numbers but somewhat similar with Block - if we do not substantially bring Common down.

<img width="427" alt="grafik" src="https://github.com/ethereumjs/ethereumjs-monorepo/assets/931137/146ef5cb-ec87-48e9-869e-3ad93a9d79ac">

The vast chunk of the parameters actually make up the gas prices for the EVM (>70% on first estimate). I first thought that we should fully move to EVM, turns out to be too difficult on second look though with all the which-gas-price-at-which-hardfork logic baked into Common which we would otherwise need to replicate.

What we can do instead though is to separate VM/EVM and non-VM/EVM params in Common itself a bit more by doing two separate `eips.ts` and `eipsEVM.ts` (or so) as well as `hardforks.ts` and `hardforksEVM.ts` files (or, alternative, maybe also enough: at least do separate data structures `EIPsDict` and `EIPsDictEVM` or so).

Then we can do separate constructors like `createCommon()` (#3487 already baked in here! 🙂) and `createCommonWithEVM()`.

This would already allow to tree-shake things out for Block/Tx/Blockchain users and so cut out 20-30KB from the 40KB eips.ts and hardforks.ts together.

There might be better solutions out there, open for suggestions (one would e.g. need to be a bit more careful that a non-EVM common does not make it into the EVM, I would suggest a boolean flag in `Common` `isEVMCommon` or so and a direct check everytime a `Common` is set in EVM/VM with directly throwing so that this is in doubt detected early).

But the proposed solution would already substantially improve the situation.

holgerd77:
After seeing how much this is condensed in #3500  from @scorbajio (see e.g. https://github.com/ethereumjs/ethereumjs-monorepo/blob/7bf3771210e9c353e1d44c65c289c6428fffb0d6/packages/common/src/hardforks.ts) I wonder if this change I proposed here is really worth the effort (respectively I guess I am rather on the not worth it side already), since this would introduce significant double structures - also with some additional bloat - and the "new bloat to win" ratio would still be positive but also not sooo super big.

[...]

PR #3512, opened 2024-07-17
Common: Simplify EIP and Hardfork Config Structure

holgerd77:
Small PR to move the `comment`, `url` and `status` properties of the Common EIP and hardfork configuration files over to the code docs, since there is no active programmatical use and this unnecessarily increases bundle sizes (as discussed in #3488).


jochem-brouwer (approved):
LGTM

PR #3513, opened 2024-07-17
Common/Monorepo: Remove NetworkId

holgerd77:
This PR removes `networkId` from Common and upstream usages and configuration files. Client is keeping for now as a CLI option (but deprecated) for a smoother transition on things like Docker and stuff.

Ready for review! 🙂


packages/client/bin/cli.ts
```diff
@@ -923,7 +929,7 @@ async function run() {
   // TODO sharding: Just initialize kzg library now, in future it can be optimized to be
   // loaded and initialized on the sharding hardfork activation
   // Give network id precedence over network name
```
gabrocheleau, line 931:
```suggestion
  // Give chainId priority over networkId
  // Give networkId precedence over network name
```

packages/common/test/customChains.spec.ts
```diff
@@ -63,12 +61,12 @@ describe('[Common]: Custom chains', () => {
 
   it('custom() -> behavior', () => {
     let common = createCustomCommon({ chainId: 123 })
-    assert.deepEqual(common.networkId(), BigInt(1), 'should default to mainnet base chain')
+    assert.equal(common.consensusAlgorithm(), 'casper', 'should default to mainnet base chain')
```
gabrocheleau, line 64: Is this right? Why are we not checking chainID?

holgerd77: Yes, that's correct, the goal of this test is to *not* check for the value passed (`chainId`), but to check that the other values remain the mainnet-specific ones.

holgerd77:
Have addressed, ready for re-review!

packages/client/bin/cli.ts
```diff
@@ -85,8 +85,14 @@ const args: ClientOpts = yargs
...
   })
+  .option('chainId', {
+    describe: 'Chain ID',
+    choices: networks.map((n) => parseInt(n[0])).filter((el) => !isNaN(el)),
+    default: undefined,
+    conflicts: ['customChain', 'customGenesisState', 'gethGenesis'], // Disallows custom chain data and chainId
+  })
   .option('networkId', {
-    describe: 'Network ID',
+    describe: 'Network ID (deprecated, use chainId instead)',
```
ScottyPoi, line 95:
```suggestion
    describe: 'Network ID',
    deprecated: true,
    deprecate: 'use --chainId instead',
```

jochem-brouwer (approved):
LGTM!

PR #3515, opened 2024-07-17
Trie: replace static methods

ScottyPoi:
Moving ahead with #3487

### constructor methods:

Moved to `constructors.ts`

- `Trie.create`
  - becomes `createTrie`
- `Trie.createFromProof`
  - becomes `createTrieFromProof`
- (deprecated) ~~`Trie.fromProof`~~
  - absorbed into `createTrieFromProof`

### static proof methods

Moved to `proof/index.ts`

- `Trie.verifyProof`
  - becomes `verifyTrieProof`
- `Trie.verifyRangeProof`
  - becomes `verifyTrieRangeProof`

--

### small refactor in `createTrieFromProof` 

new function `createTrieFromProof` replaces `Trie.createFromProof` *and* deprecated method `Trie.fromProof`

A small refactor was made, removing the optional parameter `shouldVerifyRoot` parameter, which defaulted to `false`.  The new function will infer this intention by the presence of a `root` value passed into the (also optional) `trieOpts` parameter.

```
const shouldVerifyRoot = trieOpts?.root !== undefined
```


holgerd77 (approved):
LGTM, nice! 🙂

PR #3516, opened 2024-07-18
EVM: replace static constructor

ScottyPoi:
Works on #3487

Replaces `EVM.create` with function `createEVM`

`createEVM` is defined in `constructors.ts`


holgerd77 (approved):
LGTM

PR #3517, opened 2024-07-18
Common: Remove HF names from Params Dict

holgerd77:
Sorry, I was confused yesterday. So the names in `HardforkTransitionConfig` are something different, so HTC is a different structure. These are describing the HF order, while we are talking about the HF paramter dict here.

So we can safely remove and do this small "refactoring"/simplification.

Ready for review and merge.


acolytec3 (approved):
LGTM

PR #3518, opened 2024-07-18
Monorepo: Removal all TTD Logic from Libraries

holgerd77:
Too radical? 😂 

Could anyone already give this a first look, removed all TTD stuff from Commoon, lots of tests falling off the roof, but I would say: so what?

(so Common would be just a first step, this would go up the libraries then)


packages/common/src/common.ts
```diff
@@ -779,7 +723,8 @@ export class Common {
       const blockOrTime = hf.timestamp ?? hf.block
       if (
         (hf.forkHash === null || hf.forkHash === undefined) &&
-        ((blockOrTime !== null && blockOrTime !== undefined) || typeof hf.ttd !== 'undefined')
```
g11tech, line 782: Q: how are we preventing paris from not adding to the forkhash, may be we should check for paris here?

holgerd77: This is actually handled in the `_calcForkHash()` method, and we do have a `paris` check already in there, just looked it up (and tested via REPL):

```ts
if (
  typeof blockOrTime === 'number' &&
  blockOrTime !== 0 &&
  blockOrTime !== prevBlockOrTime &&
  name !== Hardfork.Paris
) {
  const hfBlockBytes = hexToBytes(`0x${blockOrTime.toString(16).padStart(16, '0')}`)
  hfBytes = concatBytes(hfBytes, hfBlockBytes)
  prevBlockOrTime = blockOrTime
}
```

So this should not be an issue.

packages/client/src/miner/pendingBlock.ts
```diff
@@ -107,10 +107,8 @@ export class PendingBlock {
     if (typeof vm.blockchain.getTotalDifficulty !== 'function') {
       throw new Error('cannot get iterator head: blockchain has no getTotalDifficulty function')
     }
```
acolytec3, line 109: Do we still need this check?

holgerd77: I would find it nice if we are still able to at least *run* ethash blockchains, and this check might be necessary for it?

acolytec3: You are probably right.  On a related note:  should this still work `npm run client:start:ts -- --dev=pow`?
image

holgerd77: No idea about this dev flags

packages/block/test/header.spec.ts
```diff
@@ -107,18 +107,6 @@ describe('[Block]: Header functions', () => {
...
-    assert.throws(
-      () =>
-        createHeaderFromRLP(rlpHeader, {
-          common,
-          freeze: false,
-          setHardfork: 1n, // Added to bypass defaulting setHardfork to true in static constructor
-        }),
-      'A base fee',
-      undefined,
-      'throws when RLP serialized block with no base fee on default hardfork (london) and setHardfork left undefined',
```
jochem-brouwer, line 119: This test is removed but I don't think this has anything to do with TTD, so we should likely keep this in?

packages/client/src/rpc/modules/engine/engine.ts
```diff
@@ -292,26 +289,6 @@ export class Engine {
       () => this.connectionManager.updateStatus(),
     )
 
-    /**
-     * exchangeTransitionConfiguration
-     */
-    this.exchangeTransitionConfigurationV1 = cmMiddleware(
```
jochem-brouwer, line 298: Had to check, but we can indeed remove this method https://github.com/ethereum/execution-apis/blob/b7432c54d2faa5f282db7b3202c3c2ef907103ad/src/engine/cancun.md#deprecate-engine_exchangetransitionconfigurationv1

packages/client/src/rpc/modules/engine/engine.ts
```diff
@@ -1485,41 +1424,6 @@ export class Engine {
   async getPayloadV4(params: [Bytes8]) {
     return this.getPayload(params, 4)
   }
-  /**
-   * Compare transition configuration parameters.
-   *
-   * V1 (Paris HF), see:
-   * https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#engine_exchangetransitionconfigurationv1
-   *
-   * Note: This method is deprecated starting with the Cancun HF
```
jochem-brouwer, line 1494: Ah right its also noted here that its deprecated.

packages/client/test/integration/merge.spec.backup.ts
jochem-brouwer, line 1: I think we can remove this file, as it also needs the PoW -> PoS transition

packages/common/src/common.ts
```diff
@@ -690,6 +630,7 @@ export class Common {
         typeof blockOrTime === 'number' &&
         blockOrTime !== 0 &&
         blockOrTime !== prevBlockOrTime &&
+        // TODO: Decide if we should remove this since we no longer support TTD merge switch
```
jochem-brouwer, line 633: Remaining TODO

packages/common/src/constructors.ts
```diff
@@ -35,22 +35,14 @@ export function createCustomCommon(
 /**
  * Static method to load and set common from a geth genesis json
  * @param genesisJson json of geth configuration
- * @param { Mainnet, eips, genesisHash, hardfork, mergeForkIdPostMerge } to further configure the common instance
+ * @param \{ Mainnet, eips, genesisHash, hardfork, mergeForkIdPostMerge } to further configure the common instance
```
jochem-brouwer, line 38:
```suggestion
 * @param { Mainnet, eips, genesisHash, hardfork, mergeForkIdPostMerge } to further configure the common instance
```

packages/common/test/utils.spec.ts
```diff
@@ -57,122 +55,24 @@ describe('[Utils/Parse]', () => {
     assert.equal(params.genesis.timestamp, '0x10', 'timestamp parsed correctly')
   })
 
-  it('should successfully parse kiln genesis and set forkhash', async () => {
-    const common = createCommonFromGethGenesis(gethGenesisKilnJSON, {
```
jochem-brouwer, line 61: I don't think this one should be removed as well? Correctly setting the hardfork order? Maybe the `mergeForkIdTransition` should be removed here?

packages/common/test/utils.spec.ts
```diff
@@ -57,122 +55,24 @@ describe('[Utils/Parse]', () => {
...
-        'london',
-        'paris',
-        'shanghai',
-      ],
-      'hardfork parse order should be correct',
-    )
-
-    assert.equal(common.getHardforkBy({ blockNumber: 0n }), Hardfork.London, 'london at genesis')
-    assert.equal(
-      common.getHardforkBy({ blockNumber: 1n, td: 2n }),
```
jochem-brouwer, line 149: Same for this test, we can remove all checks with have to do with `td` checks, but the `getHardforkBy` via either block or timestamp should still work.

jochem-brouwer (commented):
Some small-ish comments.

The `mergeForkIdTransition`: should we remove this? This is a special fork id, which specifically has to do with the PoW -> PoS transition logic which we essentially have removed here (see: https://eips.ethereum.org/EIPS/eip-3675#eip-2124-fork-identifier). Should we remove this hardfork as well? It has lead numerous times to confusion :sweat_smile: Since we don't support PoW -> PoS transition I think it is fine to remove it as well (can be done in a new PR)

Also side note: we will now thus fail any tests which have to do with PoW -> PoS transitions (there likely are not many, but there likely will be some in the `ethereum/tests` and in hive)

holgerd77:
> Some small-ish comments.
> 
> The `mergeForkIdTransition`: should we remove this? This is a special fork id, which specifically has to do with the PoW -> PoS transition logic which we essentially have removed here (see: https://eips.ethereum.org/EIPS/eip-3675#eip-2124-fork-identifier). Should we remove this hardfork as well? It has lead numerous times to confusion 😅 Since we don't support PoW -> PoS transition I think it is fine to remove it as well (can be done in a new PR)

Let's be semantically precise here: we are *only* removing the *Live* transition from PoW -> PoS (at least that's my understanding from the basis of what we are doing and that's also what I would be suggesting), so that is this first-time transition where the fork block is not yet known and need to be determined by TTD in a complicated manner.

We still do support a static transition (post Merge PIT, merge block known), and I personally would strongly pledge to keep it this way, since the remaining logic is easy and not a substantial overhead and otherwise we would e.g. finally loose on a theoretical/practical mainnet full sync capability (still dreaming of this 😂).

For this this HF is also needed, otherwise one would branch off with a missing `mergeForkId` HF during sync.

Ah, and while writing about it: the subsequent fork hashes would also be wrong, so we would loose the correct fork hash calculation (since in-between HFs are calculated in into newer ones). So we should definitely keep "as is" and live with some mild technical debt here (can remember a somewhat longer discussion/thread from Martin Holst Swende where he also complained about this, that this forkk would now be "in forever", but that's just how it is now. 🙂 )

> Also side note: we will now thus fail any tests which have to do with PoW -> PoS transitions (there likely are not many, but there likely will be some in the `ethereum/tests` and in hive)

So, since `ethereum/tests` are passing here, this does not seem to be a problem at least for `ethereum/tests` (?).

packages/common/test/utils.spec.ts
holgerd77: I tried to re-add the two tests you mentioned, but I am getting a "Error parsing parameters file: nonzero terminal total difficulty is not supported" error now when trying to re-add the Kiln config file, so I guess this is not so easy to re-create in this setup.

I would generally pledge that there are more simple test setups for HF order as well as `getHardforkBy()` functionality and this is sufficiently tested in other places and we therefore shouldn't stick to this super-outdated Kiln setup (I was somewhat happy to throw the Kiln config out) and just throw it out here.

(side mention: guess nothing urgent to ad hoc re-add, but I do wonder if we would rather want to keep the "Geth config with post-Merge transition but a dedicated Merge HF block included" case (so this would also match what we keep in other places) instead of fully throwing this out. Guess *THAT* case also should not produce significant overhead and still migth be useful. Again, but in doubt we can "re-add on demand" I guess).

packages/common/test/utils.spec.ts
holgerd77: See comment above

packages/common/src/constructors.ts
holgerd77: Will do locally.

packages/common/src/common.ts
holgerd77: This needs to stay in, same argument as the broader fork hash comment, will just remove the TODO.

packages/client/test/integration/merge.spec.backup.ts
holgerd77: Have removed.

packages/client/src/rpc/modules/engine/engine.ts
holgerd77: Yes, also looked this up before removal (not sure why I stumbled initially upon that, likely because it used the functionality we are removing here).

packages/common/src/common.ts
holgerd77: (so, to summarize: having the Paris check here is not about a TTD merge switch but a simple by-block-number merge switch + still needed for correct post-Merge fork hash calculation)

acolytec3 (approved):
LGTM.  Lets fix that bracket later.

jochem-brouwer:
> So, since ethereum/tests are passing here, this does not seem to be a problem at least for ethereum/tests (?).

@holgerd77 This is because we do not run certain forks on the CI, in this case that would be the fork which actually has the Merge transition. (Just like we don't run `ethereum/tests` for Chainstart for instance on the CI)

PR #3519, opened 2024-07-18
Stream dependency replacement with walk all value nodes

scorbajio:
This change picks up where #3507 left off, but with a change of plan to use `walkAllValueNodes` to implement storage-dump functionality, and remove the stream dependency currently being used.


packages/statemanager/src/stateManager.ts
```diff
@@ -1017,6 +1024,15 @@ export class DefaultStateManager implements EVMStateManagerInterface {
    The object will also contain `nextKey`, the next (hashed) storage key after the range included in `storage`.
    */
   async dumpStorageRange(address: Address, startKey: bigint, limit: number): Promise<StorageRange> {
+    function nibblestoBytes(arr: Nibbles): Uint8Array {
```
acolytec3, line 1027: I don't love that we're duplicating code here that already exists in `trie` but I get why we're doing it.  It's fine for purposes of this PR but a takeaway is to clean up our `trie` utility methods and then come back and remove this.

acolytec3:
Checked the bundle size with this change and it looks like it reduces the `trie` bundle by about 30% with removing the `readable-stream` dependency.

acolytec3 (dismissed):
LGTM!  Great work here.

packages/statemanager/src/stateManager.ts
acolytec3: Nevermind, found the appropriate function in `trie` and have imported it instead of duplicating the code here.

acolytec3 (dismissed):

packages/statemanager/src/stateManager.ts
```diff
@@ -1026,44 +1027,48 @@ export class DefaultStateManager implements EVMStateManagerInterface {
...
+              i++
+            } else if (i === limit) {
+              resolve({
+                storage: storageMap,
+                nextKey: bytesToHex(keyBytes),
+              })
+            }
+          }
+        })
+        .then((_) =>
```
holgerd77, line 1066: We are exposing a lot from `Trie` to the `StateManager`, also with all these imports above. Can we abstract this away one level more by adding an async method `getStorageMap(): StorageMap` to the `Trie` class and call that instead?

holgerd77 (commented):
One thing I would at least like to discuss, otherwise this looks great! 🤩

holgerd77:
Yeah, actually this looks really, really great! 🎉 

So this is for the VM.

Before:

grafik

After:

grafik

Note that *both* the `readable-stream` code *and* the `Buffer` polyfill are gone!!!

packages/statemanager/src/stateManager.ts
holgerd77: To reiterate here: I would really like to update here!

I am still not sure e.g. if it is (was) a good idea that we instantiate a new object for every trie node walk throuh (and should rather go bare metal and use the plain Uint8Array or something for performance reasons), and if we so deep-use the trie internals it would make it harder to encapsule respectively to refactor.

I also do think that regarding verkle statefull integration it will be beneficial if we stay 1:1 with as many methods as possible and this one would also be a candidate for a 1:1 replacement if we add this higher level method.

(let me know if there are downsides I might not see atm)

acolytec3: I'm not 100% following your thoughts here.  Is your basic concern that we're calling `walkAllValueNodes` here and have to use `TrieNode` and other related types/helpers from `trie` inside `dumpStorageRange`?  I'm not seeing where we instantiate a heavy new `trie` object (unless you're talking about further down in the process of walking the trie - so maybe `lookupNode`?)  If so, that feels like something we should handle as a separate PR.  Otherwise, yes, seems like we could certainly encapsulate the logic here in `getStorageMap` as you described and add a range filter as a parameter.

holgerd77: > Is your basic concern that we're calling walkAllValueNodes here and have to use TrieNode and other related types/helpers from trie inside dumpStorageRange

Yes, that's my main concern.

We make heavy use of API parts from Trie reaching deep into Trie and which might not have been a good idea to have been exposed in the first place, and if we officially use these ourselves we are stuck with these structures until the next breaking release round.

So I would strongly prefer here to use/introduce a higher level and more abstracted API call into trie (and, again: which then might be the same for Verkle) which does not make so much use of Trie (half-)internals.

holgerd77: And, to re-iterate on the last part: a higher level `getStorageMap()` method already lays the ground for more shared state manager code by stateful verkle and MPT, while this specifc Trie code drives us further apart from that. So I think that is the better abstraction.

acolytec3: You're definitely right on this last part.  That said, the **only** trie walking mechanism I have in the verkle trie module right now is `trie.get`.  Since there is no separate `storageTrie` with verkle it will be a completely different beast to retrieve a storage map(and probably a lot easier to implement since I guess all the storage values should roughly be in the same stem).

holgerd77: Are you ok with switching here? Relatively strong opinion and ready to argue further. 😂

acolytec3: Yes as far as MPT goes.  As far as the Verkle API goes, lets hold it loosely.  The Verkle Trie structure isn't as closely tied to Ethereum state definitions as the MPT is (in the sense that VKT doesn't have internal distinctions between account/storage values) so I dunno if we want to create this `getStorageMap` function as part of the verkle trie API or not.  I may be thinking too much about this since verkle tries only exist for Ethereum today but if we ever did want to make `verkleTrie` a generalized data structure available for non-Ethereum specific uses, `getStorageMap` would no longer make sense. In any case, that's something that can be decided in the future once I've actually built out a `StatefulVerkleStateManager`.

holgerd77: Ok, thanks, yes, we doesn't need to decide/commit to verkle at this point

scorbajio: So the idea would be having a single `getStorageMap` function that allows you to pass in a range and get a storageMap in return, if no range is passed in default to the full range, which is to say a full dump of storage?

holgerd77: Yes, I guess so, do not have this fully on the plate, but basically a method in the `Trie` class which can then cover all the use cases here, and each optimally with a one-line-code call into trie.

You can also comment if this makes sense or if I am overlooking something. If you think this is feasible it would be great if you can update on the PR!

acolytec3: >So the idea would be having a single getStorageMap function that allows you to pass in a range and get a storageMap in return, if no range is passed in default to the full range, which is to say a full dump of storage?

Yes, this is exactly it. Just copy your code from `dumpStorageRange` and put it in its own function on the `trie` class with optional start and ends for the range..

holgerd77: Can someone please actually DO it? 🙂😆 This is not more that an hours work and it would be really good if we get this one merged.

acolytec3 (approved):
LGTM

PR #3523, opened 2024-07-19
TX: Replace static constructors

ScottyPoi:
Continues work on #3487 

Replaces static constructors from transaction classes with functions exported from `constructors.ts`

### from_tx_data
`<TransactionClass>.fromTxData` becomes `txFromTxData.<TransactionClass>`

example: `LegacyTransaction.fromTxData(...)`  is now `txFromTxData.LegacyTransaction(...)`

### from_values_array
`<TransactionClass>.fromValuesArray` becomes `txFromValuesArray.<TransactionClass>`

example: `LegacyTransaction.fromValuesArray(...)`  is now `txFromValuesArray.LegacyTransaction(...)`

### from_serialized_tx
`<TransactionClass>.fromSerializedTx` becomes `txFromSerializedTx.<TransactionClass>`

example: `LegacyTransaction.fromSerializedTx(...)`  is now `txFromSerializedTx.LegacyTransaction(...)`


holgerd77 (changes requested):
Hi Scotty,
thanks for taking this on! 🙂 

I am relatively sure that this solution takes away the tree shaking benefits, respectively makes tree shaking worse than before.

So these generic `txFromTxData` dict structures always reference all the tx classes, and if people only deal with a specific tx type (e.g. want to fromValueCreate a 1559 tx), all the code from the other classes will be drawn in.

Can we therefore please stick to the simple scheme we started with (please also make sure to stay in the `create*` naming scheme)? This is just the simplest and proven thing to do.

Please also make sure to add these overview comments we started to add with the `Block` class here: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3489/files#diff-749e831c793c4a66ab671adfcadb65ba7a8266efe92ab4856e9ea4bd890100cbL61

Thanks! 🙏 🙂

holgerd77:
Update: ah, and I guess what we do want to have for this one is having the constructors a bit more structured. I would suggest that we do it as a folder in this case with the following structure:

- `constructors/index.ts`
- `constructors/1559.ts`
-  and so on

Alternatively we can also do the folders by tx type (?). Maybe that's even better?

- `1559/index.ts`
- `1559/constructors.ts`
- `1559/tx.ts` (not sure, or `main.ts`, other?)

???

jochem-brouwer:
I think my preference is folders if we have multiple files per tx type to keep it all organized!

holgerd77:
Jochem and I would vote for the `1559/*.ts` folder structure 🙂

ScottyPoi:
Reviewing the comments and conversation.
It will probably be easier to start fresh with a new PR and close this one, eh?

acolytec3:
Yes, I think the replacement PR already got merged. Will close here

PR #3524, opened 2024-07-19
util: Replace account static constructors

ScottyPoi:
continues with #3487 

Replaces static constructors for `Account` class with functions  exported from `account.ts`

- `Account.fromRlpSerializedAccount(...)` > `accountFromRlp(...)`
- `Account.fromRlpSerializedPartialAccount(...)` > `accountFromPartialRlp(...)`
- `Account.fromValuesArray(...)` > `accountFromValuesArray(...)`
- `Account.fromAccountData(...)` > `accountFromAccountData(...)`
- `Account.fromPartialAccountData(...)` > `accountFromPartialAccountData(...)`


holgerd77 (changes requested):
Please stay in the nameing scheme we started with, so `createAccountFromAccountData()`.

### Possible Monorepo Wide Naming Simplifications

We can discuss about nameing simplifications (also just realizing that the above one actually is pretty long). But if we do we should do it consistently accross all libraries.

I would e.g. first round suggest that we generally switch over for the "base" (so: the `*from*Data()` constructors to a simplified naming scheme and drop the `from*Data()` part, so .e.g.:

- `createAccountFromAccountData()` -> `createAccount()`
- `create1559FeeMarketTxFromTxData()` -> `create1559FeeMarketTx()`
(here I would value some additional opinion also from others if we would want to drop or keep the `FeeMarket` naming addition, the names get somewhat long, on the other hand this helps on expressiveness and code readability. I have some slight tendency to keep, but would have me overvoted)
- `createBlockFromBlockData()` -> `createBlock()`

I think these more basic names are still "expressive" enough since this (passing in separate values/parameters) is just the natural thing to do/expect.

### RLP Name

- `Account.fromRlpSerializedAccount` -> `createAccountFromRlp`

I do like this naming simplification, think still expressive enough. If we do we should also do monorepo wide.

@ScottyPoi I would suggest that you finish this PR with the old naming conventions and then we eventual monorepo-wide name changes on the stuff we discussed in a separate follow-up PR.

holgerd77:
Ok, I talked this through with Jochem, not sure if you have seen the #work channel discussion, and we have some agreement on the renamings. 🙂 

So to summarize here (so this goes monorepo-wide then):

1) We rename/shorten the `from[Library]Data` names, example: `createAccountFromAccountData()` -> `createAccount()`
2) We shorten the `fromRlpSerialized[Library]` names, let's please also do: `Rlp` -> `RLP` as we have in Block already, example: `createAccountFromRlpSerializedAccount` -> `createAccountFromRLP`
3) We rename the `fromValuesArray` names to a more expressive `fromBytesArray`, example : `createAccountFromValuesArray` -> `createAccountFromBytesArray`

If you go along, from my side you/we can then also directly do for Account within this PR. We should just remember to then also do a follow-up PR and do it monorepo-wide.

jochem-brouwer:
I have done this for the partial accounts: `createPartialAccount` and `createPartialAccountFromRLP`. Not entirely sure if this naming is intuitive/correct though. It was `accountFromPartialAccountData` before I changed it.

holgerd77 (approved):
LGTM, thanks @ScottyPoi and @jochem-brouwer! 🙂

PR #3528, opened 2024-07-22
Switch `js-sdsl` to `js-sdsl/orderedMap` sub package

### LINKED ISSUE #3526: Monorepo: Use Isolation Packages for js-sdsl
Part of #3446 

Just stumbled upon this: for `js-sdsl`, where we use the `OrderedMap` for state manager and maybe stuff in 1-2 other places (needs a check), one can also use the [isolation packages](https://github.com/js-sdsl/js-sdsl?tab=readme-ov-file#-download) which we should definitely do.

grafik

The package makes up for a non-trivial part in the overall bundle size, here for/from VM:

grafik

(ah, just seeing: it's actually client and SM (version numbers should be aligned along))

acolytec3:
Fixes #3526

Reduces the statemanager package bundle by 4kb


acolytec3:
Reduces the bundle size from :
image
to
image

So not a huge savings but still an improvement.

jochem-brouwer (approved):
LGTM!

holgerd77 (approved):
LGTM

(yeah, interesting, thought it would be more. 🤔 Wonder if it's worth to generally keep this dependency, this was a bit thought of that we would expand usage over time but at the end we ended up just only using this one single data structure)

Issue #3522, opened 2024-07-19
VM -> Tree Shaking: Decouple `runBlock()`, `runTx()` and `buildBlock()`


holgerd77:
Part of #3446

In the VM there are these three central methods `runBlock()`, `runTx()` and `buildBlock()` which have separate files on their own (like [runBlock.ts](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/runBlock.ts) and are already build for taking in the `VM` as an object (`export async function runBlock(this: VM, opts: RunBlockOpts): Promise<RunBlockResult>`). These functions are then bound in the [vm.ts](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/vm.ts#L194C3-L196C4) file to the VM object with code like this:

```ts
async runBlock(opts: RunBlockOpts): Promise<RunBlockResult> {
  return runBlock.bind(this)(opts)
}
```

This basically does nothing useful except effectively chop off tree shaking and we can easily improve the situation by removing those bindings and switch calling into these like this without loosing any comfort:

```ts
vm.runBlock(opts: RunBlockOpts) // old
runBlock(vm, opts) // new
```

(I would have a tendency to keep the simple names)

In fact these methods are very much functional/procedural by nature (so just: wrapping some flow around some objects (EVM, StateManager, Blockchain) and not so much about "doing manipulations on the VM object" (what is a "VM" anyway?), so this might generally even be the more fitting design. But that just as a side note.

Bundle for this part looks like this:

![grafik](https://github.com/user-attachments/assets/66c4050f-6282-45b4-961e-28905f1a289f)

With the decoupling the respectively not-used parts (so if one is e.g. only calling into `runTx()` this will be: `runTx()`, `buildBlock()` and even the `requests` stuff (used in `runBlock()` and `buildBlock()`) will go away.

holgerd77:
Note: we can generally further think about where we are with `VM` (was e.g. wondering in between if we still need/want a `VM` class at all?), but I would suggest to first round start with this work, since it's pretty simple and already brings us along for some significant part of the way.

scorbajio:
Addressed by #3530.

PR #3530, opened 2024-07-22
 VM: runTx(), runBlock(), buildBlock() Standalone Methods

holgerd77:
Addresses #3522

First round on replacing the VM core class based methods with function calls by remove the bindings.

Feel free to pick up if wanted!


jochem-brouwer:
I think this should be ready for review now.

Small side note, upon checking `vm.ts` to see whats left, we have a huge `emitEVMProfile` method there, which does NOT call into `this` anywhere, so we can also remove that!

holgerd77:
Can you do in a separate PR please?

jochem-brouwer:
Ok, went ahead and also moved `emitEVMProfile` to a separate file. 

This will only help if the end user only uses `buildBlock` (which is likely the least popular method of VM), but this was a simple optimization so I think this is OK.

holgerd77 (commented):
Very cool! 👏 

(will admin-merge)

PR #3529, opened 2024-07-22
evm: move map -> array for opcodes/gas

jochem-brouwer:
This PR changes the internal Opcodes / Dynamic Gas type from Map -> Array. Array lookups seem faster than Map.

Also super small code cleanup, such as removing unused arguments (`_common`)

PR #3531, opened 2024-07-22
evm: opcode benchmarks

jochem-brouwer:
Originally intended for #3529, but this can be used as standalone.

This PR creates a very simple "opcode benchmark". It reads all stack information from `stackDelta` from EOF. Note: this stack delta only contains opcodes available in EOF, so opcodes as `CALL` are not in there.

The benchmark simply setups the stack with the right amount of inputs (all zeros), then performs the operation, and then pops any outputs if the opcode has them. Then it jumps back (it loops forever). This is obviously a very naive test, for instance for the ADD opcode it will thus keep performing `0 + 0` the entire time.


gabrocheleau:
Do you think this is worth keeping + rebasing + merging @jochem-brouwer ?

jochem-brouwer:
No, the https://github.com/ethereum/execution-specs/tree/forks/amsterdam/tests/benchmark tests can be used for this, as well as the tooling to run those: https://github.com/ethpandaops/benchmarkoor/tree/master which will generate reports more sophisticated than those attempted to generate here :smile: :+1: 

The fill tests can be used with `--gas-benchmark-values 1` which limits block gas limit to 1M. The benchmarks are specifically designed to target the worst case scenarios (not average case) and this differs per implementation (also EthereumJS / JavaScript) and on the hardware (+ interpreter) it runs on.

Will close here :smile: :+1:

PR #3532, opened 2024-07-23
Common Refactor

holgerd77:


package.json
```diff
@@ -12,6 +12,11 @@
     "e2e:publish": "./scripts/e2e-publish.sh",
     "e2e:resolutions": "node ./scripts/e2e-resolutions.js",
     "examples": "npm run examples --workspaces --if-present",
+    "lint": "npm run lint --workspaces --if-present",
+    "lint:fix": "npm run lint:fix --workspaces --if-present",
+    "test": "npm run test --workspaces --if-present",
+    "test:node": "npm run test:node --workspaces --if-present",
```
acolytec3, line 18: I'm fully in support of these but should they be part of this PR?  If so, can we also update devp2p so it has a `test:node` script as an alias of `test`?

acolytec3 (commented):
Looks great so far!  Left a couple of questions for us to confirm before we merge.

package.json
holgerd77: I would pledge yes, this is such a tiny side change not affecting anything apart of having a bit more comfort. I will add the alias.

packages/common/src/common.ts
```diff
@@ -410,95 +386,88 @@ export class Common {
...
-        ) {
-          value = (hfChanges[1] as any)[topic][name]
+        const hfValue = hfChanges[1].params?.[name]
+        if (hfValue !== undefined) {
+          value = hfValue
         }
       }
       if (hfChanges[0] === hardfork) break
     }
+    if (value === undefined) {
```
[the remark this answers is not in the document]
holgerd77: Stumbled upon this too, so `value == null` is officially part of the mechanism/logic, meaning that a value from a former HF/EIP is erased again, this is being used e.g. in `netSstoreNoopGas: null, // Removed along EIP-1283`.

Then this goes through here and along `return BigInt(value ?? 0)` below the `null` value gets transitioned into `0`.

acolytec3 (approved):

jochem-brouwer:
Could the PR description be updated? This is handy for when we write the breaking release notes later :smile: :+1:

holgerd77:
@jochem-brouwer you're right, have done!

jochem-brouwer:
The description is still empty for me though?

PR #3534, opened 2024-07-23
Refactor trie util helpers

scorbajio:
This change takes a look at and refactors some of the following utility functions in the `trie` package:
* `nibblesToBytes`: This function is modified to return a `Uint8Array` bytes array instead of taking the byte array that will be getting the final nibbles copied over to as input, in order to conform more closely to its inverse, `bytesToNibbles`.
* `nibblestoBytes`: `nibblestoBytes` is replaced by the identical `nibblesTypeToPackedBytes` function, and is different from `nibblesToBytes` in that its purpose is to take in a `Nibble`-typed input and return a `Uint8Array`-typed output that contains one nibble per byte (the lower end is padded with 0's).
* `nibbleTypeToPackedBytes`: This function is renamed to `nibblesTypeToPackedBytes` and is moved to `nibbles.ts`, in place of where `nibblestoBytes` was.  
* `bytesToNibbles`: This function is left unchanged.


scorbajio:
This change has a merge conflict with #3519 and should be merged after it has already been merged.

acolytec3 (approved):
LGTM

PR #3535, opened 2024-07-23
Tx: clarify method names

jochem-brouwer:
This PR:

- Renames tx's `getBaseFee` to `getIntrinsticGas` to avoid confusion with 1559s `baseFee`
- Renames tx's `getDataFee` to `getDataGas` to be more explicit that the unit of this method is gas units, not "fee" which could be interpreted as "wei" (actual Eth value, which for instance would be `tx.gasLimit * tx.gasPrice` for the maximum cost of that tx)
- Makes the `getUpfrontCost` documentation a bit more specific


packages/tx/src/7702/tx.ts
```diff
@@ -119,8 +119,8 @@ export class EOACodeEIP7702Transaction extends BaseTransaction<TransactionType.E
   /**
    * The amount of gas paid for the data in this tx
    */
-  getDataFee(): bigint {
-    return EIP7702.getDataFee(this)
+  getDataGas(): bigint {
```
acolytec3, line 122: Can someone remind me why we wrap this helper function instead of just having the code here?

holgerd77: To avoid redundancy, since we have this code in various tx types.

holgerd77 (approved):
LGTM, thanks! 🙏 🙂

PR #3537, opened 2024-07-24
Common: Remove Parameters

holgerd77:
THIS is actually the reason I did all this Common refactoring during the last days. 😂 

This PR fully removes all parameters from Common and moves them over to the packages into dedicated `params.ts` file. Nevertheless the full Common hardfork/EIP switching logic is preserved with Common still providing the central binding ground for parameter selection, so there is are no API changes for the users whatsoever.

Initially this work was mainly motivated by our current tree shaking work. I now find other benefits rather prevailing.

This new structure makes it significantly easier to fully adjust and customize parameter sets for libraries in a robust way along all hardforks, which was previously extremely complicated. I added new `params` options for the affected libraries along (tx, block, evm, vm), see associated test case and/or code docs for usage.

There are somewhat significant tree shaking benefits too though.

So this is the `londonTx.ts` example before:

grafik

This is with the PR:

grafik

Benefits are mainly due to the large EVM parameter set not being unnecessarily pulled over to the tx library. This now goes for all libraries though with the new structure, so evm is free-d up from tx, block and vm parameters.

Some additional note:

I had some problems with side effects along plainly using the param dicts as a base for customization (manipulation of the imported dict directly affected the original dictionary itself). I added explicit notes to the docs. Also - to make deep cloning easier - I removed all BigInt usages from params and replaced it with strings. I think this is a general improvement, I had some problems with this already earlier, can't fully remember the context though. Here it is for easier deep cloning, which does not work for BigInts (depending what option you choose). Change has no drawbacks, since there is a `BigInt(param)` call in the `Common.param()` method anyhow and so there was no logic code change necessary and this rather eliminated this double-BigInt instantiation (which worked).

Ready for review! 🙂


holgerd77:
This is now ready for review, added a description in the initial post section.

packages/block/src/types.ts
```diff
@@ -46,6 +46,23 @@ export interface BlockOptions {
    * Default: `false` (HF is set to whatever default HF is set by the {@link Common} instance)
    */
   setHardfork?: boolean | BigIntLike
+  /**
+   * Block parameters sorted by EIP can be found in the exported `paramsBlock` dictionary,
+   * which is internally passed to the associate `@ethereumjs/common` instance which
+   * manages parameter selection based on the hardfork and EIP settings.
+   *
+   * This option allows to provide a set of own parameters. Note that parameters
+   * get fully overwritten, so you need to blend off from the default params dict
```
scorbajio, line 55: What does "blend off" mean to say here?

packages/block/src/types.ts
```diff
@@ -46,6 +46,23 @@ export interface BlockOptions {
    * Default: `false` (HF is set to whatever default HF is set by the {@link Common} instance)
    */
   setHardfork?: boolean | BigIntLike
+  /**
+   * Block parameters sorted by EIP can be found in the exported `paramsBlock` dictionary,
+   * which is internally passed to the associate `@ethereumjs/common` instance which
+   * manages parameter selection based on the hardfork and EIP settings.
+   *
+   * This option allows to provide a set of own parameters. Note that parameters
```
scorbajio, line 54:
```suggestion
   * This option allows providing a custom set of parameters. Note that parameters
```

packages/common/test/data/paramsTest.ts
```diff
@@ -0,0 +1,52 @@
...
+  /**
+.  * Constantinope HF Meta EIP
+.  */
+  1013: {
+    // gasPrices
+    netSstoreNoopGas: 200, // Once per SSTORE operation if the value doesn't change
+    // pow
+    minerReward: BigInt('2000000000000000000'), // The amount a miner gets rewarded for mining a block
+  },
+
```
scorbajio, line 29:
```suggestion
```
A bit of a nit but maybe spacing can be kept consistent.

packages/common/test/data/paramsTest.ts
```diff
@@ -0,0 +1,52 @@
...
+  /**
+.  * Petersburg HF Meta EIP
+.  */
+  1716: {
+    // gasPrices
+    netSstoreNoopGas: null, // Removed along EIP-1283
+  },
+
```
scorbajio, line 37:
```suggestion
```

packages/evm/src/types.ts
```diff
@@ -227,6 +233,24 @@ export interface EVMOpts {
    */
   allowUnlimitedInitCodeSize?: boolean
 
+  /**
+   * EVM parameters sorted by EIP can be found in the exported `paramsEVM` dictionary,
+   * which is internally passed to the associate `@ethereumjs/common` instance which
```
scorbajio, line 238:
```suggestion
   * which is internally passed to the associated `@ethereumjs/common` instance which
```

packages/evm/src/types.ts
```diff
@@ -227,6 +233,24 @@ export interface EVMOpts {
...
+   * manages parameter selection based on the hardfork and EIP settings.
+   *
+   * This option allows to provide a set of own parameters. Note that parameters
```
scorbajio, line 241:
```suggestion
   * This option allows providing a set of custom parameters. Note that parameters
```

packages/evm/src/types.ts
```diff
@@ -227,6 +233,24 @@ export interface EVMOpts {
...
+   * get fully overwritten, so you need to blend off from the default params dict
```
scorbajio, line 242: Again here, not sure what "blend off" intends to say?

packages/tx/src/types.ts
```diff
@@ -82,6 +83,23 @@ export interface TxOptions {
    * Current default hardfork: `istanbul`
    */
   common?: Common
+  /**
+   * Tx parameters sorted by EIP can be found in the exported `paramsTx` dictionary,
+   * which is internally passed to the associate `@ethereumjs/common` instance which
+   * manages parameter selection based on the hardfork and EIP settings.
+   *
+   * This option allows to provide a set of own parameters. Note that parameters
+   * get fully overwritten, so you need to blend off from the default params dict
```
scorbajio, line 92: Again here: "blend off".

packages/tx/src/types.ts
```diff
@@ -82,6 +83,23 @@ export interface TxOptions {
    * Current default hardfork: `istanbul`
    */
   common?: Common
+  /**
+   * Tx parameters sorted by EIP can be found in the exported `paramsTx` dictionary,
+   * which is internally passed to the associate `@ethereumjs/common` instance which
+   * manages parameter selection based on the hardfork and EIP settings.
+   *
+   * This option allows to provide a set of own parameters. Note that parameters
```
scorbajio, line 91:
```suggestion
   * This option allows providing a set of custom parameters. Note that parameters
```

packages/vm/src/types.ts
```diff
@@ -153,6 +153,23 @@ export interface VMOpts {
    * Default: `false` (HF is set to whatever default HF is set by the {@link Common} instance)
    */
   setHardfork?: boolean | BigIntLike
+  /**
+   * VM parameters sorted by EIP can be found in the exported `paramsVM` dictionary,
+   * which is internally passed to the associate `@ethereumjs/common` instance which
+   * manages parameter selection based on the hardfork and EIP settings.
+   *
+   * This option allows to provide a set of own parameters. Note that parameters
```
scorbajio, line 161:
```suggestion
   * This option allows providing a set of custom parameters. Note that parameters
```

packages/vm/src/types.ts
```diff
@@ -153,6 +153,23 @@ export interface VMOpts {
...
+   * get fully overwritten, so you need to blend off from the default params dict
```
scorbajio, line 162: "blend off"

packages/block/src/types.ts
holgerd77: Have rewritten, thought that would be somewhat of an english word. 😂

packages/block/src/types.ts
holgerd77: Will change locally.

packages/common/test/data/paramsTest.ts
holgerd77: Updated

packages/common/test/data/paramsTest.ts
holgerd77: Updated

packages/evm/src/types.ts
holgerd77: updated

ScottyPoi (approved):
LGTM!

packages/common/src/common.ts
```diff
@@ -89,6 +89,55 @@ export class Common {
...
+   * ```
+   *
+   * @param params
+   */
+  updateParams(params: ParamsDict) {
+    for (const [eip, paramsConfig] of Object.entries(params)) {
+      if (!(eip in this._params)) {
+        this._params[eip] = { ...paramsConfig } // copy
+      } else {
+        this._params[eip] = { ...this._params[eip], ...params[eip] }
```
jochem-brouwer, line 114: Does / should this deep copy?

jochem-brouwer (commented):
One question.

I love this idea in general to move out all the package-specific params into their own params file. I had never thought about this. Great!! :smile: :+1:

packages/common/src/common.ts
holgerd77: Ah, yes, likely, thanks for catching. I actually thought this would (and I still think it does to some extend, but maybe not on the second level dicts? 🤔) But then I re-stumbled upon this in the test writing, where I tried the same and it actually didn’t.

I will keep this here to save CI but do in subsequent PR today (also about Common).

holgerd77:
> One question.
> 
> I love this idea in general to move out all the package-specific params into their own params file. I had never thought about this. Great!! 😄 👍

This was a long thought process, and not a pop-up idea out of nowhere, by the way! 😂

packages/common/src/common.ts
```diff
@@ -75,6 +73,8 @@ export class Common {
         (this._chainParams.customHardforks && this._chainParams.customHardforks[hf.name]),
     ])
     this._hardfork = this.DEFAULT_HARDFORK
+    this._params = { ...(opts.params ?? {}) } // copy
```
jochem-brouwer, line 76: If the shallow copy should be removed, then this one should likely also be addressed :smile:

packages/common/src/common.ts
```diff
@@ -89,6 +89,55 @@ export class Common {
...
+   *   1559: {
+   *     initialBaseFee: 1000000000,
+   *   }
+   * }
+   * ```
+   *
+   * @param params
+   */
+  resetParams(params: ParamsDict) {
+    this._params = { ...params } // copy
```
jochem-brouwer, line 137: Here is also a copy

packages/common/src/common.ts
jochem-brouwer: That's fine! I have addressed two other places where the shallow copy seems to happen with comments.

packages/common/src/common.ts
```diff
@@ -326,10 +375,10 @@ export class Common {
   /**
    * Internal helper for _buildParamsCache()
    */
-  protected _mergeWithParamsCache(params: HardforkConfig | EIPConfig) {
+  protected _mergeWithParamsCache(params: ParamsConfig) {
     this._paramsCache = {
       ...this._paramsCache,
-      ...params['params'],
+      ...params,
```
jochem-brouwer, line 381: Here also a possible shallow copy.

packages/common/src/common.ts
holgerd77: Oh yes, good point! Thanks!

packages/common/src/common.ts
holgerd77: Guess you mean "deep copy" but: yes!

Will all do in subsequent PR (not the state manager one)

Issue #3455, opened 2024-06-14
Trie -> Tree Shaking Refactor: ProofTrie Subclass


holgerd77:
Part of #3446 

**Earliest start together with official breaking release work!**

### Description

This is the code size situation for trie:

![grafik](https://github.com/ethereumjs/ethereumjs-monorepo/assets/931137/41b5a5fe-9f62-45da-bdec-bca63e136587)

The most effective TSR (Tree Shaking Refactor 😂) by far and the I guess lowest hanging fruit is to separate out the proof code and do a subclass `ProofTrie` inheriting from `Trie` which can be used if one does need proof functionality. Everyone else can stick to the `Trie` class.

This substantially reduces the size of the main `trie.ts` file (~ estimate by manually deleting: 48 KB -> 37 KB) and fully draws out the huge (22 KB) `range.ts` file.

So this is close to cutting things in half.

### Other

Two other things worth to consider at first glance (respectively: think more about if there are feasible refactoring strategies):

1. Refactor CheckpointDB passing and usage in a way that checkpointing code can be tree shaked out if not needed
2. Create an own `Read(Only)Trie` class as a super class of current `Trie` class and remove all trie *generation* class from there (not fully sure if there are compelling enough respectively real-world use cases)

### Bundle Size Testing

The following is a one-liner which has prooven to be able to (re-)test bundle size:

```shell
npm i -g esbuild
cd packages/trie
```

Create a file `t1.mjs` with a code path to be tested (e.g. a simple `Trie` instantiation) like:

```ts
import { Trie } from '@ethereumjs/trie'

const t = new Trie()
console.log(t.root())
```

The bundle can then be (re-)created with:

```shell
npm run build && esbuild --bundle t1.mjs --outfile=out1.mjs
```

This gives output like the following (which is already impressive, 380 KB for trie along is actually quite a lot 😬):

![grafik](https://github.com/ethereumjs/ethereumjs-monorepo/assets/931137/0c1db89b-56a4-4314-bb94-7d5d0b0d83d8)

holgerd77:
Closed by #3551

PR #3538, opened 2024-07-24
Move proof functions into ProofTrie subclass

scorbajio:
This change is inspired by #3455 and aims to make the trie package more modular and it's bundle size smaller by separating proof functionality from the `Trie` class and putting it in a `ProofTrie` subclass.


holgerd77:
I wonder if a sub class is really the right concept here, I already had my doubts when writing this issue. 

The big downside then is again it's that we then have again two Trie classes and one needs to choose which one to use, also don't like the typing with ProofTrie in StateManager that much. 

A better approach might be to just make dedicated standalone functions from the proof methods in a separate file and taking in a Trie object as the first parameter. 

This should also make tree shaking possible while preserving this only-one-Trie structure. 

Let's please further discuss before you proceed.

scorbajio:
Closing this change, work is being continued in #3551.

PR #3539, opened 2024-07-25
Update linting

acolytec3:
This PR updates Eslint to v8.57 (last legacy version) and Prettier to 3.3.3. 
- Updates eslint to use top level `tsconfig.lint.json` so that examples are included in linting in all packages
- Disables `no-console` and `no-unused-vars` rules in the examples
- Cleans up obsolete lint configs in `client`
- Removes various odd rules in place in specific packages 
- Fix all the lint issues


package.json
```diff
@@ -34,20 +34,20 @@
     "@vitest/ui": "^v2.0.0-beta.12",
     "c8": "7.12.0",
     "embedme": "1.22.1",
-    "eslint": "8.45.0",
-    "eslint-config-prettier": "8.8.0",
-    "eslint-config-typestrict": "1.0.5",
+    "eslint": "8.57.0",
+    "eslint-config-prettier": "^9.1.0",
+    "eslint-config-typestrict": "^1.0.5",
```
gabrocheleau, line 39: Are we comfortable moving to unpinned versions? I recall previous discussions where we had decided to pin them down for security reasons.

packages/block/.eslintrc.cjs
```diff
@@ -1 +1,15 @@
...
+    parserOptions: {
+      project: ['./tsconfig.lint.json'],
+    },
+    overrides: [
+      {
+        files: ['examples/**/*'],
+        rules: {
+          'no-console': 'off',
+          '@typescript-eslint/no-unused-vars': 'off',
+        },
```
gabrocheleau, line 12: Why do we need to overwrite the examples folder for every package? Isn't there a way to specify that generally?

gabrocheleau (approved):

packages/block/.eslintrc.cjs
acolytec3: Let me experiment and see.

package.json
acolytec3: I'm not anxious about it since these are dev dependencies but will happily defer to other opinions.

holgerd77: I am also ok with unpinned versions

gabrocheleau: good!

gabrocheleau (approved):

PR #3540, opened 2024-07-25
client: ENGINE_NEWPAYLOAD_MAX_TXS_EXECUTE constant

ScottyPoi:
Updates ENGINE_NEWPAYLOAD_MAX_TXS_EXECUTE constant to 200.

This fixes the failing eip-6110 hive test brought to our attention in the chat.


holgerd77:
@g11tech what do you think, can we do this change? Guess might be not too dramatic if we risk a bit of lagging, but can’t fully judge!

jochem-brouwer:
I think this is fine, note that even if we would limit it to 1 tx, then you can still make the tx consume 30M gas (block gas limit) which might still run over 12s if they hit the unoptimized/slow parts of our code.

acolytec3 (approved):

acolytec3:
Have verified this fixes the failing hive test.  Will merge here.

g11tech (approved):
Yes should be harmless

PR #3543, opened 2024-07-26
StateManager: Interface Refactor/Simplification

holgerd77:
This PR refactors and simplifies the `StateManager` interfaces located in `Common`. Especially the double structure with the more generic `StateManagerInterface` and the inheriting `EVMStateManagerInterface` turned out to be too inflexible and more and more non-EVM methods had been added there over the years, since one could not "escape the inheritance structure".

grafik

After some discussion of some more complex solutions (multiple interfaces, by case API definitions citing the various interfaces needed), this PR now goes for a more simple solution, which nevertheless should suit the practical needs. The two interfaces are unified again, and instead not so central functionality is made optional (`getProof()`, `dumpStorage(),..). This allows to free the various state manager implementations from most of the "dummy methods" which simply throw when called (which is inherently dangerous).

Instead it should be made sure (mostly) on the constructor level of consuming classes that a state manager is passed in which provides the respective functionality for some applied configuration (e.g. the EVM checking for the Verkle-needed methods when 6800 is activated in the constructor).

The whole refactoring is accompanied by a stronger grouping and documentation of the various state manager properties in the `StateManagerInterface`. This should additionally help to not get lost again what functionality is used where and for what and to make proper decisions in the future on how to place a new state manager method and if to make optional or not.

Best reviewed commit-by-commit (I guess), ready for review! 🙂

### DIFF
--- a/packages/client/src/execution/vmexecution.ts
+++ b/packages/client/src/execution/vmexecution.ts
@@ -296,7 +296,7 @@ export class VMExecution extends Execution {
           this.chain['_customGenesisState'] ?? getGenesis(Number(blockchain.common.chainId()))
         if (
           !genesisState &&
-          (!('generateCanonicalGenesis' in this.vm) || !this.config.statelessVerkle)
+          (!('generateCanonicalGenesis' in this.vm.stateManager) || !this.config.statelessVerkle)
         ) {
           throw new Error('genesisState not available')
         } else {
--- a/packages/client/src/execution/vmexecution.ts
+++ b/packages/client/src/execution/vmexecution.ts
@@ -296,11 +296,11 @@ export class VMExecution extends Execution {
           this.chain['_customGenesisState'] ?? getGenesis(Number(blockchain.common.chainId()))
         if (
           !genesisState &&
-          (this.vm instanceof DefaultStateManager || !this.config.statelessVerkle)
+          (!('generateCanonicalGenesis' in this.vm) || !this.config.statelessVerkle)
         ) {
           throw new Error('genesisState not available')
         } else {
-          await this.vm.stateManager.generateCanonicalGenesis(genesisState)
+          await this.vm.stateManager.generateCanonicalGenesis!(genesisState)
         }
       }
 
--- a/packages/common/src/interfaces.ts
+++ b/packages/common/src/interfaces.ts
@@ -184,11 +184,10 @@ export interface StateManagerInterface {
   /*
    * Proof Functionality
    */
-  // Only client, should not be mandatory
+  // Only client (should not be mandatory)
   getProof?(address: Address, storageSlots: Uint8Array[]): Promise<Proof>
 
   shallowCopy(downlevelCaches?: boolean): StateManagerInterface
-  getAppliedKey?(address: Uint8Array): Uint8Array
 
   /*
    * EVM/VM Specific Functionality
@@ -197,14 +196,15 @@ export interface StateManagerInterface {
     get(address: Address, key: Uint8Array): Promise<Uint8Array>
     clear(): void
   }
+  generateCanonicalGenesis?(initState: any): Promise<void> // TODO make input more typesafe
   // only Verkle/EIP-6800 (experimental)
   checkChunkWitnessPresent?(contract: Address, programCounter: number): Promise<boolean>
+  getAppliedKey?(address: Uint8Array): Uint8Array // only for preimages
 }
 
 export interface EVMStateManagerInterface extends StateManagerInterface {
   dumpStorage(address: Address): Promise<StorageDump> // only used in client
   dumpStorageRange(address: Address, startKey: bigint, limit: number): Promise<StorageRange> // only used in client
-  generateCanonicalGenesis(initState: any): Promise<void> // TODO make input more typesafe
 
   shallowCopy(downlevelCaches?: boolean): EVMStateManagerInterface
 }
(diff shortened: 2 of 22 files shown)


jochem-brouwer (commented):
One general comment (should be disregarded) and one question

packages/evm/src/evm.ts
```diff
@@ -94,7 +94,7 @@ export class EVM implements EVMInterface {
   public readonly common: Common
   public readonly events: AsyncEventEmitter<EVMEvents>
 
-  public stateManager: EVMStateManagerInterface
+  public stateManager: StateManagerInterface
```
acolytec3, line 97: I didn't realize just how minimal this change was.  Nice!

packages/common/src/interfaces.ts
```diff
@@ -164,35 +167,46 @@ export interface StateManagerInterface {
...
+  getAppliedKey?(address: Uint8Array): Uint8Array // only for preimages
 
-  dumpStorage(address: Address): Promise<StorageDump> // only used in client
-  dumpStorageRange(address: Address, startKey: bigint, limit: number): Promise<StorageRange> // only used in client
-  generateCanonicalGenesis(initState: any): Promise<void> // TODO make input more typesafe
-  getProof(address: Address, storageSlots?: Uint8Array[]): Promise<Proof>
-
-  shallowCopy(downlevelCaches?: boolean): EVMStateManagerInterface
+  /*
+   * Generic
```
acolytec3, line 208: Can we rename this to something like `Miscellaneous` or `Utility`?  "Generic" seems like the wrong term here.

packages/common/src/interfaces.ts
```diff
@@ -164,35 +167,46 @@ export interface StateManagerInterface {
...
   commit(): Promise<void>
   revert(): Promise<void>
 
-  // State root methods
+  /*
+   * State Root Functionality
+   */
   getStateRoot(): Promise<Uint8Array>
   setStateRoot(stateRoot: Uint8Array, clearCache?: boolean): Promise<void>
   hasStateRoot(root: Uint8Array): Promise<boolean> // only used in client
```
acolytec3, line 182: Can these three be made optional?  I'm assuming we can't implement them in either the `SimpleStateManager` or the `RpcStateManager`.

packages/statemanager/src/simpleStateManager.ts
```diff
@@ -125,6 +118,8 @@ export class SimpleStateManager implements EVMStateManagerInterface {
     this.topStorageStack().set(`${address.toString()}_${bytesToHex(key)}`, value)
   }
 
+  async clearStorage(): Promise<void> {}
```
acolytec3, line 121: Why is there no implementation of this one?

packages/statemanager/src/simpleStateManager.ts
```diff
@@ -163,24 +158,4 @@ export class SimpleStateManager implements EVMStateManagerInterface {
   hasStateRoot(): Promise<boolean> {
     throw new Error('Method not implemented.')
   }
```
acolytec3, line 160: What do you think about making these three optional since 2 out of our 4 state managers don't implement them.

acolytec3 (changes requested):
Love the overall simplification here.  Left a few questions we should address before merging.

packages/client/src/execution/vmexecution.ts
```diff
@@ -296,11 +296,11 @@ export class VMExecution extends Execution {
           this.chain['_customGenesisState'] ?? getGenesis(Number(blockchain.common.chainId()))
         if (
           !genesisState &&
-          (this.vm instanceof DefaultStateManager || !this.config.statelessVerkle)
```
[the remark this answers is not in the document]
holgerd77: Yes, fixed!

packages/vm/src/types.ts
```diff
@@ -137,11 +136,6 @@ export interface VMOpts {
    * Default: `false`
    */
   activatePrecompiles?: boolean
-  /**
-   * A genesisState to generate canonical genesis for the "in-house" created stateManager if external
-   * stateManager not provided for the VM, defaults to an empty state
-   */
-  genesisState?: GenesisState
```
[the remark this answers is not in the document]
holgerd77: I have checked, we have 12-15 places where we do internally and only 1 where we use this `genesisState` functionality and otherwise we simply do:

```ts
this.vm.stateManager.generateCanonicalGenesis(genesisState)
```

So I thought this is as convenient (for a rarely used functionality) and from a `StateManagerInterface` perspective it had the advantage that we can lower the requirements for the `generateCanonicalGenesis()` implementations (since not necessary in VM) and do not need extra VM checks.

packages/common/src/interfaces.ts
holgerd77: Sure.

packages/statemanager/src/simpleStateManager.ts
holgerd77: There was non before, so there is no status quo change here, but we might add (just would not want to do it now, have this other PR on the plate as well). I tested just re-setting the internal storage dict, but *this* actually then made tests fail in VM/EVM so I removed again.

acolytec3 (approved):
LGTM

packages/vm/src/types.ts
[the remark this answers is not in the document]
jochem-brouwer: Great, agreed

jochem-brouwer (approved):
LGTM

PR #3544, opened 2024-07-26
util: refactor bigint to address helper into Address.fromBigInt method

gabrocheleau:
This PR refactors the Address constructors into separate helpers.


packages/util/src/address.ts
```diff
@@ -37,6 +38,14 @@ export class Address {
     return new Address(zeros(20))
   }
 
+  /**
+   * Returns an Address object from a bigint address (they are stored as bigints on the stack)
+   * @param value The bigint address
+   */
+  static fromBigInt(value: bigint): Address {
+    return new Address(setLengthLeft(bigIntToBytes(value & MASK_160), 20))
```
jochem-brouwer, line 46: I don't think we should put the mask here. This has the undesired side effect that if an user passes a bigint > 20 bytes, it does not throw but silently throws away the upper bytes. (We should thus do this explicitly in EVM, but not in this method). This method should thus also throw if you pass a bigint > 20 bytes.

jochem-brouwer: However, it should still `setLengthLeft` after checking that the bigint <= 20 bytes. So you can `Address.fromBigInt(BigInt(1))` without problems.

gabrocheleau: Ok thanks for the note. So you are suggesting that we:
- Add length validation to the `fromBigInt` helper so that it throws if length > 20
- Remove the bitwise mask operation and instead do it in usage. 
- Keep setLengthLeft(value,20) as we do now

Since fromBigInt is strictly used (in our monorepo) for the evm, I'm not sure how to best architect this. Will experiment

jochem-brouwer: Yes, exactly as you have noted here. So the EVM should call `fromBigInt` with the masked variant, `fromBigInt` should throw if bytes > 20 and it should always perform the `setLengthLeft` (otherwise Address constructor will throw on bytes < 20 bytes)

gabrocheleau: I made the adjustment above and decided to create a helper method in the evm that does the masking, curious to hear what you think!

packages/evm/src/opcodes/util.ts
```diff
@@ -20,9 +20,20 @@ import { EvmError } from '../exceptions.js'
 import type { ERROR } from '../exceptions.js'
 import type { RunState } from '../interpreter.js'
 import type { Common } from '@ethereumjs/common'
+import type { Address } from '@ethereumjs/util'
 
 const MASK_160 = (BIGINT_1 << BIGINT_160) - BIGINT_1
 
+/**
+ * Create an address from a 160-bit integer.
```
jochem-brouwer, line 28:
```suggestion
 * Create an address from a stack item (256 bit integer).
```

packages/util/src/address.ts
```diff
@@ -139,3 +70,89 @@ export class Address {
...
+/**
+ * Returns an Address object from a bigint address (they are stored as bigints on the stack)
+ * @param value The bigint address
+ */
+export function createAddressFromBigInt(value: bigint): Address {
+  const bytes = bigIntToBytes(value)
+  if (bytes.length > 20) {
+    throw new Error(`Invalid address, too long: ${bytes.length}`)
+  }
+  return new Address(setLengthLeft(bigIntToBytes(value), 20))
```
jochem-brouwer, line 90:
```suggestion
  return new Address(setLengthLeft(bytes, 20))
```

No need to bigint -> bytes twice ;)

jochem-brouwer (commented):
Two comments.

Could you also add some tests for the new `createAddressFromBigInt`?

packages/util/src/address.ts
gabrocheleau: ah, of course, thanks

gabrocheleau:
> Two comments.
> 
> Could you also add some tests for the new `createAddressFromBigInt`?

done!

jochem-brouwer (approved):
LGTM

PR #3545, opened 2024-07-26
Common: Chain Refactor

holgerd77:
Ok, this is a first PoC on the Common simplification with directly passing in (and only allowing) a `ChainConfig` object (so, simply: `Mainnet` e.g.) to the `chain`, as discussed with @acolytec3 some time ago!

This is for tree shaking, but also leads to a dramatic simplification of code! Conversion helpers, all gone, lots of typing simplifications, also a strongly streamlined `creatCustomCommon`.

I also killed `setChain()` along the way, since we never had any use case for runtime-chain switching, as some side note.

This is absurdly intrusive though regarding the code changes, hope I get this under control! 😝 Not so intrusive luckily regarding the final outer API changes.

Uff. 🤯 

Ah, and the `Common` tests are already passing, that's something. None of the upstream tests though yet.

-- Also added `getPresetChainConfig` as a reasonable replacement for `getInitializedChains` since we no longer allow multiple chain configurations in a single instance of `common`. (note from @acolytec3)


packages/common/src/types.ts
```diff
@@ -27,7 +27,7 @@ type ConsensusConfig = {
 
 export interface ChainConfig {
   name: string
-  chainId: number | bigint
+  chainId: number | string
```
holgerd77, line 30: This is also a change for serialization and is not yet reflected in the code base.

packages/common/src/common.ts
```diff
@@ -746,7 +710,7 @@ export class Common {
   /**
    * Returns an eth/64 compliant fork hash (EIP-2124)
    * @param hardfork Hardfork name, optional if HF set
-   * @param genesisHash Genesis block hash of the chain, optional if already defined and not needed to be calculated
+   * @param genesisHash Genesis block hash of the Mainnet, optional if already defined and not needed to be calculated
```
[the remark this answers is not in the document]
acolytec3: I believe so.  Will update the comment.

packages/common/src/chains.ts
```diff
@@ -1,655 +1,653 @@
...
-        port: 30303,
-        id: '4aeb4ab6c14b23e2c4cfdce879c04b0748a20d8e9b59e25ded2a08143e265c6c25936e74cbc8e641e3312ca288673d91f2f93f8e277de3cfa444ecdaaf982052',
-        location: 'eu-central-1-001',
-        comment: 'bootnode-hetzner-fsn',
-      },
-    ],
-    dnsNetworks: [
-      'enrtree://AKA3AM6LPBYEUDMVNU3BSVQJ5AD45Y7YPOHJLEF6W26QOE4VTUDPE@all.mainnet.ethdisco.net',
-    ],
+export const Goerli: ChainConfig = {
```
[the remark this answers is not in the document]
acolytec3: Hmm, good point.  Can we save that for a future PR so I don't have to risk breaking tests with hardcoded Goerli values?

packages/common/src/chains.ts
```diff
@@ -1,655 +1,653 @@
...
+      location: '',
+      comment: 'bootnode 2',
+    },
+  ],
+  dnsNetworks: [
+    'enrtree://AKA3AM6LPBYEUDMVNU3BSVQJ5AD45Y7YPOHJLEF6W26QOE4VTUDPE@all.holesky.ethdisco.net',
+  ],
+}
+
+export const Kaustinen6: ChainConfig = {
```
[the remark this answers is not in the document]
acolytec3: Same point as Goerli.  Yes, you're correct that Kaustinen7 is the next one, but I would prefer to defer on that work until Kaustinen7 actually launches.

packages/common/src/chains.ts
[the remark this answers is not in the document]
jochem-brouwer: Yes that's fine!

jochem-brouwer (approved):
Great teamwork, LGTM :pray: :smile:

packages/vm/test/api/customChain.spec.ts
```diff
@@ -50,10 +50,10 @@ const genesisState: GenesisState = {
   [contractAddress]: accountState,
 }
 
-const common = new Common({
+// @ts-ignore PrefixedHesString type is too strict
```
holgerd77, line 52: Was this PrefixedHexString comment something one could not "fix on the fly"? 🤔 

If something is "structurally" too strict, we should maybe un-stricten it?

packages/vm/test/api/runBlock.spec.ts
```diff
@@ -58,10 +58,10 @@ const common = new Common({ chain: Mainnet, hardfork: Hardfork.Berlin })
 describe('runBlock() -> successful API parameter usage', async () => {
   async function simpleRun(vm: VM) {
     const common = new Common({ chain: Mainnet, hardfork: Hardfork.London })
-    const genesisRlp = hexToBytes(testData.genesisRLP as PrefixedHexString)
+    const genesisRlp = hexToBytes(testData.default.genesisRLP as PrefixedHexString)
     const genesis = createBlockFromRLPSerializedBlock(genesisRlp, { common })
 
-    const blockRlp = hexToBytes(testData.blocks[0].rlp as PrefixedHexString)
+    const blockRlp = hexToBytes(testData.default.blocks[0].rlp as PrefixedHexString)
```
holgerd77, line 64: Are all these now commonly applied `default` accesses future proof? (so: not only work in our specific current tooling setup)

packages/common/src/common.ts
```diff
@@ -60,7 +60,7 @@ export class Common {
   constructor(opts: CommonOpts) {
     this.events = new EventEmitter()
 
-    this._chainParams = JSON.parse(JSON.stringify(opts.chain))
+    this._chainParams = opts.chain
```
holgerd77, line 63: This `JSON.parse(...` stuff was to do a deep copy of the original structure, I otherwise had the cases where the original data got modified when being modified in the `Common` object.

So, this e.g. concretely played out somewhere in the EVM/VM tests where a test broke when simply `Mainnet` or so was used as in input two times in a row to a `Common` object and in the first round the associated `_chainParams` got modified.

(I am doing a commit-by-commit review, bear with me if some of the stuff is getting fixed on later commits)

packages/client/test/rpc/eth/sendRawTransaction.spec.ts
```diff
@@ -31,6 +31,12 @@ describe(method, () => {
       return this
     }
     const common = new Common({ chain: Mainnet })
+    common
+      .hardforks()
+      .filter((hf) => hf.timestamp !== undefined)
+      .map((hf) => {
+        hf.timestamp = undefined
+      })
```
holgerd77, line 39: I am not fully seeing through this change, but do we really need this hacky stuff here? 🤔 

Or is otherwise something wrong with our typing/data structure/whatever?

holgerd77 (commented):
@acolytec3 @jochem-brouwer thanks a lot for working through this 🙏, some additional post-merge comments, but nothing grave!

One-step-back question: are you guys generally satisfied/happy with the direction of the changes within this PR, now that you have worked your way through this?

packages/common/src/common.ts
holgerd77: ( @jochem-brouwer even mentioned that we should do this in yet another place, can't ad hoc remember where (maybe along the `copy()` method))

holgerd77:
Also, to complete here:

This DOES have its desired effect on the bundle size, so `Common` is slowly getting to sane(r) levels.

Before (on `tx/examples/londonTx.ts`):

grafik

grafik

After:

grafik

grafik

Or simply by "looking at the box sizes" if the above is too noisy:

Before:

grafik

After:

grafik

Slowly getting there.

roninjin10:
Very nice!

packages/common/src/common.ts
roninjin10: Btw, I learned just this week there is a better way to deep copy an object in js called structuredClone https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

acolytec3: That'll be a nice thing to switch to if/when NodeJS brings it out from behind the `experimental` flag.  Looks like it's in Node 22 but not fully supported yet. https://nodejs.org/api/globals.html#structuredclonevalue-options

packages/vm/test/api/runBlock.spec.ts
holgerd77: @acolytec3 maybe a final comment on this one?

acolytec3: I think it's fine, though honestly not 100% sure. I think it was a knock-on effect of switching our dev setup to `module` in our package.json across the monorepo.

packages/vm/test/api/customChain.spec.ts
acolytec3: When we take JSON and pass it directly to the various constructors that have the `PrefixedHexString` type, Typescript now complains because we're passing in `string` s. We either need to loosen the type on the opts that can be passed to the various `*Data` constructors or else type check all these strings ahead of time to confirm they're prefixed correctly.

packages/vm/test/api/runBlock.spec.ts
holgerd77: Ok, thanks, then let's keep/take it, though I find all these kind of things honestly a bit spooky

PR #3548, opened 2024-07-27
blockchain: move chain,skeleton to blockchain

g11tech:
move skeleton,chain from client to blockchain class

first cut after a few iterations
 - doesn't yet merges db ops of skeleton to dbmanager, could be done in followup
 - need to ensure that skeleton to main db movement is minimzed (or even eliminated?)

aim of PR is to get and keep everything working

TODO:
 - [ ] readd chain updated event back
 - [ ] find the correct structure for status logging and its bookkeeping variables
 - [ ] fix the tests and all

### DIFF
--- a/packages/blockchain/src/chain.ts
+++ b/packages/blockchain/src/chain.ts
@@ -25,6 +25,7 @@ export type ChainConfig = {
   skeletonFillCanonicalBackStep: number
   maxPerRequest: number
   chainCommon: Common
+  logger?: Logger
 }
 
 /**
@@ -34,7 +35,6 @@ export interface ChainOptions {
   /**
    * Client configuration instance
    */
-  logger?: Logger
   config: ChainConfig
 
   /**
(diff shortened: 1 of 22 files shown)


holgerd77:
Hi Gajinder, had some first look, great that you started with this!

I honestly wonder a bit if this approach will work, to move all this stuff over that we have in client now. There is so much bloat and stuff we do not want to have in a generic blockchain class at the end, since all too client-specific.

So I wonder if it would be better to start the other way around: to try to expand Blockchain (somewhat from scratch) to allow for a less strict Skeleton version (not thinking about client at all), which might e.g. be the superclass of blockchain and just allows for a "non full" blockchain (likely by missing out some checks which are then done in the subclass) and *then* one could write some tests so that everything works and integrate this into the client instead of the current Skeleton. 🤔 

However you do (there might be other ways), we should really make sure that we not totally bloat the Blockchain class. That should still remain a generic package - not tailored in any way for our client - and also serve as a generic package for whatever usages as a first-class use case.

g11tech:
Understood @holgerd77 will rethink the approach and rework

g11tech:
closing in favor of https://github.com/ethereumjs/ethereumjs-monorepo/pull/3584 which aims to re-imagines this as per the feedback

PR #3551, opened 2024-07-29
Proof function util helpers refactor

scorbajio:
This change is an alternative approach to the work started in #3538. It looks into refactoring the trie class to remove proof functionality from the main class and move them into standalone functions in order to improve tree shaking for the trie package and decrease bundle size.


scorbajio:
Using a minimal code path for the `Trie` class:
```
import { Trie } from '@ethereumjs/trie'

const t = new Trie()
console.log(t.root())
```
Results:
```
before: 223.6kb
after: 205.7kb
delta: -17.9kb
```

packages/trie/src/proof/index.ts
```diff
@@ -64,4 +66,124 @@ export function verifyTrieRangeProof(
...
+}
+
+/**
+ * Create a trie from a given (EIP-1186)[https://eips.ethereum.org/EIPS/eip-1186] proof. An EIP-1186 proof contains the encoded trie nodes from the root
+ * node to the leaf node storing state data. This function does not check if the proof has the same expected root. A static version of this function exists
+ * with the same name.
+ * @param proof an EIP-1186 proof to update the trie from
+ * @deprecated Use `updateFromProof`
+ */
+export async function fromProof(trie: Trie, proof: Proof): Promise<void> {
```
acolytec3, line 176: This function feels duplicative of `createTrieFromProof` in the `constructors.ts`.  Can you think of a reason we'd need it here as well?

packages/trie/src/proof/index.ts
```diff
@@ -1,10 +1,12 @@
...
 import { createTrieFromProof } from '../constructors.js'
-import { verifyRangeProof } from '../index.js'
+import { Trie, verifyRangeProof } from '../index.js'
 import { bytesToNibbles } from '../util/nibbles.js'
 
 import type { Proof, TrieOpts } from '../index.js'
+import type { PutBatch } from '@ethereumjs/util'
 
 /**
  * Static version of verifyProof function with the same behavior. An (EIP-1186)[https://eips.ethereum.org/EIPS/eip-1186] proof contains the encoded trie nodes
```
acolytec3, line 12: This phrasing is obsolete
```suggestion
 * An (EIP-1186)[https://eips.ethereum.org/EIPS/eip-1186] proof contains the encoded trie nodes
```

packages/trie/src/proof/index.ts
```diff
@@ -64,4 +66,124 @@ export function verifyTrieRangeProof(
...
+
+  await trie['_db'].batch(opStack)
+  if (opStack[0] !== undefined) {
+    return opStack[0].key
+  }
+}
+
+/**
+ * Verifies a proof by putting all of its nodes into a trie and attempting to get the proven key. An (EIP-1186)[https://eips.ethereum.org/EIPS/eip-1186] proof
+ * contains the encoded trie nodes from the root node to the leaf node storing state data. A static version of this function exists with the same name.
```
acolytec3, line 121:
```suggestion
 * contains the encoded trie nodes from the root node to the leaf node storing state data.
```

packages/trie/test/util/log.spec.ts
```diff
@@ -34,7 +35,7 @@ describe('Run Trie script with DEBUG enabled', async () => {
   process.env.DEBUG = ''
   const trie2 = new Trie({})
   trie2['DEBUG'] = true
-  await trie2.fromProof(proof)
+  await fromProof(trie2, proof)
```
acolytec3, line 38:
```suggestion
  await createTrieFromProof(trie2, proof)
```

packages/trie/test/util/log.spec.ts
```diff
@@ -1,6 +1,7 @@
 import { utf8ToBytes } from '@ethereumjs/util'
 import { assert, describe, it } from 'vitest'
 
+import { createProof, fromProof, verifyProof } from '../../src/index.js'
```
acolytec3, line 4:
```suggestion
import { createProof, createTrieFromProof, verifyProof } from '../../src/index.js'
```

acolytec3 (changes requested):
Looks great overall.  Some comments need cleanup and I think we can remove that `fromProof` function entirely.

packages/trie/src/proof/index.ts
scorbajio: I'd agree

packages/trie/test/util/log.spec.ts
scorbajio: Fixed locally

packages/trie/test/util/log.spec.ts
scorbajio: Fixed locally

packages/trie/src/proof/index.ts
```diff
@@ -64,4 +65,104 @@ export function verifyTrieRangeProof(
...
+  const p = stack.map((stackElem) => {
+    return stackElem.serialize()
+  })
+  trie['DEBUG'] && trie['debug'](`Proof created with (${stack.length}) nodes`, ['CREATE_PROOF'])
+  return p
+}
+
+/**
+ * Updates a trie from a proof by putting all the nodes in the proof into the trie. If a trie is being updated with multiple proofs, {@param shouldVerifyRoot} can
+ * be passed as false in order to not immediately throw on an unexpected root, so that root verification can happen after all proofs and their nodes have been added.
```
gabrocheleau, line 86: This comment seems outdated, mentioning that `{@param shouldVerifyRoot} can be passed as false` seems to imply that it does not default to false (which it does). Imo we should clarify this, and add a mention the the `shouldVerifyRoot` param natspec that it defaults to false (I would naturally assume that root is verified).

acolytec3 (approved):
LGTM

PR #3553, opened 2024-08-03
Add EOF container validation script

jochem-brouwer:
This PR adds a CLI to enable EOF container checks via the command line, see https://github.com/ipsilon/eof/issues/146

Additionally, this exports `EOFContainer` and `validateEOF` from EVM to expose this externally.

Please do not merge until we get a :+1: that this works with the fuzzer.

Usage:

Ensure all libraries are build, cd into the monorepo and then `npm i`. 

Now `cd ./packages/evm` and then `tsx ./scripts/eofContainerValidator.ts`. The file will read hex strings from the stdin and will either output OK if the container is valid, or `err: <REASON>` if there is a validation error.


shemnon:
It seems like it should work but when I try and run a process from within Java I never get any output from the subprocess.  Is there a way to force flushes?  My fuzzer hangs reading from the stream.

Here's the equivelant Java code starting the process.
```
Process p =
    new ProcessBuilder()
        .redirectErrorStream(true)
        .command(
            "/opt/homebrew/bin/tsx",
            "/Users/shemnon/git/ethereumjs/ethereumjs-monorepo/packages/evm/scripts/eofContainerValidator.ts")
        .start();
```

jochem-brouwer:
Could you retry with my latest commit? Otherwise I might have to figure out how to setup the Java process so I can test it myself :sweat_smile:

shemnon:
No dice.  I think readline package may be the issue.  Java processes are just pipes, there is no higher level TTY interactions.

Something like this needs to work:
```
echo ef00010100040200010001040000000080000000 > simple.eof
tsx scripts/eofContainerValidator.ts < simple.eof 
```
and output `OK`

jochem-brouwer:
This latest push could fix it?

shemnon:
That works.  I'll post some findings over at https://github.com/ipsilon/eof/issues/146#issuecomment-2266718916 in a bit.

jochem-brouwer:
Cool, thanks a lot!

holgerd77 (approved):
Cool, nice! 🎉❤️ Love this! 😃

From my side we could merge too (won’t do myself).

acolytec3 (approved):

PR #3556, opened 2024-08-05
Adjust ttd checks in `parseGethParams`

acolytec3:
In #3518, we removed almost all of the TTD related logic in the code base and with that tightened up what we accept in our parsing of geth genesis parameters.  This caused an issue with Hive where the hive genesis parameters would provide a nonzero TTD but still expecting that merge would be at genesis (since the difficulty of the genesis block is higher than the provided TTD).  This adjusts our logic to compare genesis block difficulty vs TTD to accomodate this quirk in hive.


holgerd77 (approved):
LGTM

PR #3569, opened 2024-08-07
statemanager: cache and other refactors

gabrocheleau:
This reverts some of the changes from #3554 and rebuilds upon them.


holgerd77 (commented):
Looks great so far! 🤩 

The one major thing which I would still do structurally differently is that I think we should pass in the cache options directly to the `Caches` class and - for the state managers - take an initialized `Caches` object as the option input (and no cache options at all).

The we can remove the default initialization `this._caches = new Caches(....)` and go with no-cache-as-default.

Default initializations are major tree shaking killers. If we remove, a huge block of code goes away (if not used), so *all* caching code + the large LRU and js-sdsl dependencies.

It is *a bit* more to do for users on the UX side (so if caches should be used one needs to pass over: `{ caches: new Caches() }` (or with some options, but I would think that this is very much worth it.

gabrocheleau:
> Looks great so far! 🤩
> 
> The one major thing which I would still do structurally differently is that I think we should pass in the cache options directly to the `Caches` class and - for the state managers - take an initialized `Caches` object as the option input (and no cache options at all).
> 
> The we can remove the default initialization `this._caches = new Caches(....)` and go with no-cache-as-default.
> 
> Default initializations are major tree shaking killers. If we remove, a huge block of code goes away (if not used), so _all_ caching code + the large LRU and js-sdsl dependencies.
> 
> It is _a bit_ more to do for users on the UX side (so if caches should be used one needs to pass over: `{ caches: new Caches() }` (or with some options, but I would think that this is very much worth it.

Just did that! 

It did involve some other changes, notably in tests and in the default VM.create() state manager instantiation. I have decided to default to using Caches() (so, same behavior as before) for the default VM as a lot of vm packages tests seem to rely on that.

gabrocheleau:
Made some more progess with making cache optional and fixing tests where necessary.

With regards to your comment:
"I think you can likely simplify many of the above if-causes in a same manner and simply remove." ([link](https://github.com/ethereumjs/ethereumjs-monorepo/commit/2439219be9fe00800c6c337e031a8e7196cee4f8#r145227194))

I'm actually having a broader reflection about the handling of the undefined caches. I'm not a fan of the complex `if` clauses at the moment, I think the dual checking of `undefined` and  `deactivate` is confusing.

I don't have a clear full picture right now. We currently instantiate each cache (account, storage, code) with defaults, unless `deactivate` has been set to true. Now since `_caches` itself is optional, and all caches are optional on their own, I don't think it's necessary to check against `deactivate` in usage as we are doing here. Perhaps we still need to keep `deactivate` as a CacheStateManagerOption, but I think we could/should refrain from using it to check whether or not a particular cache is active.

acolytec3:
> Made some more progess with making cache optional and fixing tests where necessary.
> 
> With regards to your comment: "I think you can likely simplify many of the above if-causes in a same manner and simply remove." ([link](https://github.com/ethereumjs/ethereumjs-monorepo/commit/2439219be9fe00800c6c337e031a8e7196cee4f8#r145227194))
> 
> I'm actually having a broader reflection about the handling of the undefined caches. I'm not a fan of the complex `if` clauses at the moment, I think the dual checking of `undefined` and `deactivate` is confusing.
> 
> I don't have a clear full picture right now. We currently instantiate each cache (account, storage, code) with defaults, unless `deactivate` has been set to true. Now since `_caches` itself is optional, and all caches are optional on their own, I don't think it's necessary to check against `deactivate` in usage as we are doing here. Perhaps we still need to keep `deactivate` as a CacheStateManagerOption, but I think we could/should refrain from using it to check whether or not a particular cache is active.

If the caches are optional, can't you just do 
```ts
this.account?.del(key)
this.storage?.del(key)
this.storage?.get(key)
```

gabrocheleau:
> > Made some more progess with making cache optional and fixing tests where necessary.
> > With regards to your comment: "I think you can likely simplify many of the above if-causes in a same manner and simply remove." ([link](https://github.com/ethereumjs/ethereumjs-monorepo/commit/2439219be9fe00800c6c337e031a8e7196cee4f8#r145227194))
> > I'm actually having a broader reflection about the handling of the undefined caches. I'm not a fan of the complex `if` clauses at the moment, I think the dual checking of `undefined` and `deactivate` is confusing.
> > I don't have a clear full picture right now. We currently instantiate each cache (account, storage, code) with defaults, unless `deactivate` has been set to true. Now since `_caches` itself is optional, and all caches are optional on their own, I don't think it's necessary to check against `deactivate` in usage as we are doing here. Perhaps we still need to keep `deactivate` as a CacheStateManagerOption, but I think we could/should refrain from using it to check whether or not a particular cache is active.
> 
> If the caches are optional, can't you just do
> 
> ```ts
> this.account?.del(key)
> this.storage?.del(key)
> this.storage?.get(key)
> ```

Pretty much what I have ended up doing, yes!

holgerd77:
Yes, agree

packages/client/src/execution/vmexecution.ts
```diff
@@ -160,21 +165,23 @@ export class VMExecution extends Execution {
...
-        type: CacheType.LRU,
-        size: this.config.storageCache,
-      },
-      codeCacheOpts: {
-        deactivate: false,
-        type: CacheType.LRU,
-        size: this.config.codeCache,
-      },
+      caches: new Caches({
+        accountCacheOpts: {
```
holgerd77, line 169: I think we can fully drop the `CacheOpts` from this key, makes it nicer to digest/read and remains fully expressive.

holgerd77: (same for others of course)

packages/statemanager/src/cache/types.ts
```diff
@@ -1,6 +1,58 @@
...
   type: CacheType
 }
+
+export type CacheSettings = {
+  deactivate: boolean
+  type: CacheType
+  size: number
+}
+
+export interface CacheStateManagerOpts {
```
holgerd77, line 17: Maybe drop the `StateManager` here from the name

packages/statemanager/src/rpcStateManager.ts
```diff
@@ -16,26 +16,24 @@ import {
...
   protected _provider: string
+  protected _contractCache: Map<string, Uint8Array>
+  protected _storageCache: StorageCache
   protected _blockTag: string
-
-  _accountCache: AccountCache
-  _storageCache: StorageCache
-  _contractCache: Map<string, Uint8Array>
-
+  protected _accountCache: AccountCache
```
holgerd77, line 34: The caches in `RPCStateManager` will still be transfered to the new interface, right?

packages/statemanager/src/cache/types.ts
gabrocheleau: If we remove StateManager, we'll have a naming conflict with `CacheOpts` declared above. I am not 100% satisfied with the general naming, and there is some overlap between these interfaces/types (which makes me want to unify them somehow), but there does not seem to be sufficient overlap to really unify them. Open to suggestions!

packages/client/src/execution/vmexecution.ts
gabrocheleau: done!

packages/statemanager/src/rpcStateManager.ts
gabrocheleau: The `contractCache` seems different in the rpcStateManager than in the others (i.e. not the `ContractCache` class). It would indeed be nice to refactor in a similar fashion here. @acolytec3, why do we use a custom Map<string, Uint8Array> here rather than the `ContractCache` class?

holgerd77: I am pretty sure that just some of the inconsistencies we have in state manager and that was rather to have some simple version in and so you can just fully and 1:1 with the original version replace.

packages/statemanager/src/cache/types.ts
holgerd77: Maybe give this a second thought, but I would kill off the `deactivate` part and take the `size === 0` as the one and only way to deactivate and the remove the `CacheOpts` and only use the `CachesStateManagerOpts` (then renamed to `CacheOpts` 😋) as the one thing for all.

packages/statemanager/src/rpcStateManager.ts
acolytec3: I'm 99% sure that is just because the original implementation from the RPCstatemanager just didn't have caches at all. No compelling reason though and the cache will do the same thing.

gabrocheleau: Just update to use the same _caches property and type as the other 2

packages/statemanager/src/cache/types.ts
gabrocheleau: Interesting, just seeing this comment I'll give that a try, would really like to simplify further

acolytec3 (commented):
LGTM.  One TODO to be cleaned up separately

acolytec3 (approved):

holgerd77:
🎉

Issue #3563, opened 2024-08-06
Block -> Tree Shaking: Separate Consensus Methods


holgerd77:
Sorry, I am still not willing to fully throw Ethash and Clique out. 🫣 😆 What we can do though is to (easily) further separate.

For `Block` the consensus related methods should go into separate methods additionally taking in the block object as a first parameter being placed into:

- `consensus/ethash.ts`
- `consensus/clique.ts`
- `consensus/index.ts` (for export as usual, all methods should also be available from root)

Following methods:

- `ethashCanonicalDifficulty()`
- `_requireClique()` (should also go out from `Block` and used with additional block parameter from the new location, but not being exposed to users)
- `cliqueSigHash()`
- `cliqueIsEpochTransition()`
- `cliqueExtraVanity()`
- `cliqueExtraSeal()`
- NOT `cliqueSealBlock()` (separate topic, at least not for this round of work (bound to a constructor option, at least substantially more difficult to substitute/move/replace))
- `cliqueEpochTransitionSigners()`
- `cliqueVerifySignature()`
- `cliqueSigner()`

I would cautiously think we can just keep the method names.

If there are method checks like `if Block.ethashCanonicalDifficulty !== undefined` in the code base these checks can be removed.

Bundle size for a block example before/after should be provided especially on this one since this should be pretty satisfactory! 😂

PR #3571, opened 2024-08-07
Block consensus methods refactoring

scorbajio:
This change reforms certain consensus-related `Block` and `BlockHeader` class methods into standalone functions, as proposed in #3563.


scorbajio:
Testing the bundle size with a standard code path:
```

const block = Block.fromBlockData(
  {
    header: {
      baseFeePerGas: BigInt(10),
      gasLimit: BigInt(100),
      gasUsed: BigInt(60),
    },
  },
  { common }
)

console.log(Number(block.header.calcNextBaseFee())) // 11
```
Results:

before: 325.8kb
after: 322.9kb
delta: -2.9kb

scorbajio:
> Testing the bundle size with a standard code path:
> 
> ```
> import { Block } from '@ethereumjs/block'
> import { Chain, Common, Hardfork } from '@ethereumjs/common'
> const common = new Common({ chain: Chain.Mainnet, hardfork: Hardfork.London })
> 
> const block = Block.fromBlockData(
>   {
>     header: {
>       baseFeePerGas: BigInt(10),
>       gasLimit: BigInt(100),
>       gasUsed: BigInt(60),
>     },
>   },
>   { common }
> )
> 
> console.log(Number(block.header.calcNextBaseFee())) // 11
> ```
> 
> before: 325.8kb after: 322.9kb delta: -2.9kb

Taking a look at the output bundle, it seems like some of the functions are still being included in the final output bundle, will have to look into why this is happening as the code path doesn't seem to use them.

packages/evm/package.json
```diff
@@ -56,6 +56,7 @@
     "visualize:bundle": "npx vite build --config=./vite.config.bundler.ts --emptyOutDir=false --outDir ."
   },
   "dependencies": {
+    "@ethereumjs/block": "^5.2.0",
```
holgerd77, line 59: Oh, this is very very very unlucky, and might become a blocker for the whole PR.

We fought really hard to get Block out of EVM, e.g. by adding a `defaultBlock()` method faking a default block. It would definitely not be worth it to re-add just for this refactor. This would bloat the (non-tree-shaked) bundle size and so this would be an improved-on-one-end-worsened-on-the-other situation. 

Scratching my head around this, but haven't found a good solution yet. 🤔

holgerd77: Update: we could pass in the method `cliqueSigner` as an `EVMOpts` and check on instantiation - if clique is active (via Common) - that the option is passed (and otherwise throw).

Not fully sure if its worth it but also would go along if you decide to do. Maybe it's on the "ok" side. 

Yeah, maybe it's worth it. Go ethereum has its clique code also still in https://github.com/etclabscore/core-geth/blob/master/consensus/clique/clique.go , I would also really like to keep, othewise we are so super-dependent on the Consensus side of things for "everything" [tm].

And at the end not so much work, just a simple option-addition.

packages/evm/src/types.ts
```diff
@@ -346,6 +346,11 @@ export interface EVMOpts {
    *
    */
   profiler?: EVMProfilerOpts
+
+  /**
+   * Must be present if consensus type is clique/poa, else error will be thrown
+   */
+  cliqueSigner?: Function
```
holgerd77, line 353: Can you maybe fully type this out so that we have the input and output typed?

packages/block/src/consensus/index.ts
```diff
@@ -0,0 +1,12 @@
...
+  CLIQUE_EXTRA_SEAL,
+  CLIQUE_EXTRA_VANITY,
+  cliqueEpochTransitionSigners,
+  cliqueExtraSeal,
+  cliqueExtraVanity,
+  cliqueIsEpochTransition,
+  cliqueSigHash,
+  cliqueSigner,
+  cliqueVerifySignature,
+} from './clique.js'
```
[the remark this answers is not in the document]
scorbajio: The only function we don't want imported/used by users is (`_requireClique`)[https://github.com/ethereumjs/ethereumjs-monorepo/blob/41ffe667f6d5247f7e1ed39522bb22d6babb57f5/packages/block/src/consensus/clique.ts#L22]. Could export all functions, but this function too will be exported since it gets used in a separate file internal to the block package that would need to use it.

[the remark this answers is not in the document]
holgerd77: Ok, makes sense, no big deal, than just leave it as it is.

packages/block/src/header.ts
```diff
@@ -811,65 +765,6 @@ export class BlockHeader {
     return extraData
   }
```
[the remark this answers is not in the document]
scorbajio: The `cliqueSigner` option is a byte array containing the private key to be used for sealing, so not sure if we can remove it from the opts as it's not just the `cliqueSigner` function being passed. There is code in the client `miner.ts` file that passes account keys using that option field. I was able to create a constructor and move some of those functions though.

acolytec3 (changes requested):
Looks good overall.  Left a couple of comments for cleanup but great job!

packages/block/src/constructors.ts
```diff
@@ -515,3 +526,31 @@ export async function createBlockFromBeaconPayloadJson(
...
+  const signature = ecSignFunction(cliqueSigHash(header), cliqueSigner)
+  const signatureB = concatBytes(signature.r, signature.s, bigIntToBytes(signature.v - BIGINT_27))
+
+  const extraDataWithoutSeal = header.extraData.subarray(
+    0,
+    header.extraData.length - CLIQUE_EXTRA_SEAL,
+  )
+  const extraData = concatBytes(extraDataWithoutSeal, signatureB)
+  return extraData
+}
```
holgerd77, line 556: Ah, no, this is not a constructor (and like this is also has no tree shaking effects, since this method is then used again in the main class).

So a constructor here would be a something with a signature like:

```ts
createSealedCliqueBlock(blockData: BlockData = {}, cliqueSigner: Uint8Array, opts?: BlockOptions): Block
```

And this method would calculate the signature (like above) and integrate this into `extraData` and then likely (?) overwrites the `extraData` from `blockData` and then creates and returns a block with this.

Respectively there should likely be a corresponding `createSealedCliqueBlockHeader()` method where all this stuff happens.

Just looking at the other constructors: we definitely do not want to copy-paste the `createBlock()` or so code here. So if we find a solution it should be one which has not significantly more code lines than the code above.

So I guess what we can do is:

1. Do this `extraData` calculation from above and...
2. then relatively directly hand this over to the generic `createBlock()` (`createHeader()`) method.

Guess that should work? 

Can you experiment with that?

packages/evm/src/evm.ts
```diff
@@ -1150,7 +1150,6 @@ export function defaultBlock(): Block {
   return {
     header: {
       number: BIGINT_0,
-      cliqueSigner: () => createZeroAddress(),
```
holgerd77, line 1153: Is this save to remove?

packages/evm/src/types.ts
```diff
@@ -346,6 +346,11 @@ export interface EVMOpts {
    *
    */
   profiler?: EVMProfilerOpts
+
+  /**
+   * Must be present if consensus type is clique/poa, else error will be thrown
```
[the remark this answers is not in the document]
holgerd77: Yes, agree.

Otherwise this seems to well-work, nice! 🙂

packages/evm/src/types.ts
```diff
@@ -346,6 +346,11 @@ export interface EVMOpts {
...
+   */
+  cliqueSigner?: (header: Block['header']) => Address
```
holgerd77, line 353: Ah, nice trick!

holgerd77 (dismissed):

packages/block/src/consensus/clique.ts
```diff
@@ -0,0 +1,129 @@
...
+import type { BlockHeader } from '../index.js'
+import type { CliqueConfig } from '@ethereumjs/common'
+
+// Fixed number of extra-data prefix bytes reserved for signer vanity
+export const CLIQUE_EXTRA_VANITY = 32
+// Fixed number of extra-data suffix bytes reserved for signer seal
+export const CLIQUE_EXTRA_SEAL = 65
+
+// This function is not exported in the index file to keep it internal
+export function _requireClique(header: BlockHeader, name: string) {
```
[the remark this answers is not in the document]
scorbajio: Fixed

packages/evm/src/types.ts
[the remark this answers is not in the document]
scorbajio: Fixed

packages/block/src/constructors.ts
```diff
@@ -533,7 +533,10 @@ export async function createBlockFromBeaconPayloadJson(
  * @param cliqueSigner clique signer key to be used for creating sealed block
  * @returns the block constructed block
```
jochem-brouwer, line 534: These docs are incorrect, it returns the extraData, not a block (currently)

scorbajio: Will update the docs, just copy pasted them.

holgerd77:
Nice, this looks great now! 🤩

packages/block/src/constructors.ts
```diff
@@ -515,3 +516,34 @@ export async function createBlockFromBeaconPayloadJson(
...
+): Block {
+  const sealedCliqueBlock = createBlock(blockData, { ...opts, ...{ freeze: false } })
+  ;(sealedCliqueBlock.header.extraData as any) = retrieveCliqueBlockExtraData(
+    sealedCliqueBlock.header,
+    cliqueSigner,
+  )
+  if (opts?.freeze === true) {
+    // We have to freeze here since we can't freeze the block when constructing it since we are overwriting `extraData`
+    Object.freeze(sealedCliqueBlock)
+  }
```
holgerd77, line 533: Ah, yes, that's a good solution, important to keep the `freeze` option intact! 🙏

packages/evm/src/types.ts
scorbajio: typing wizard @acolytec3 ftw

acolytec3: Can I add this to my official job title?

acolytec3 (approved):
LGTM

holgerd77:
Ah, looking through this I have to say, this feels *really* good to have this consensus code and this specific methods out of core Block, also changes all look good, had a final look over the code changes! Thanks Amir, nice work! 👍 🙂

PR #3583, opened 2024-08-13
Prometheus cleanup

scorbajio:
This change does two things:
* clean up the Prometheus metrics server documentation for the client
* adds a default case of returning a `404` not found response in the case of navigating to an unkown/unhandled route


holgerd77 (approved):
Nice, LGTM! 🙂

PR #3584, opened 2024-08-13
blockchain: allow optimistic block insertion in blockchain

g11tech:
re-imagining the previous work  (https://github.com/ethereumjs/ethereumjs-monorepo/pull/3548) of extending blockchain for skeleton use based on the feedback


packages/blockchain/src/blockchain.ts
```diff
@@ -367,15 +367,10 @@ export class Blockchain implements BlockchainInterface {
...
             'Cannot put a different genesis block than current blockchain genesis: create a new Blockchain',
           )
+          // genesis block is not optimistic
+          optimistic = false
         }
 
-        const { header } = block
-        const blockHash = header.hash()
-        const blockNumber = header.number
-        let td = header.difficulty
```
g11tech, line 375: this diff has primarily shifted into the else part of the optimistic condition on L414 with some basic initializations still out side the if {} else {} block

packages/blockchain/src/db/operation.ts
```diff
@@ -14,6 +14,8 @@ import {
 
 import type { CacheMap } from './manager.js'
 
+export type OptimisticOpts = { fcUed: boolean; linked?: boolean }
+
```
holgerd77, line 18: These options are not really "understandable" in a blockchain context respectively bring in concepts from the client (FcU), and a blockchain "should not need to know what an FcU is". 😛 

Can we rename this more to the point, so with what this is doing here from the blockchain perspective (I honestly also cannot read this out of the name and bring the two things together)?

Also `linked` I also can't directly read what is meant here.

These options should - if we keep - also move to `types.ts` since they become part of the official API eventually.

holgerd77: (maybe then directly give this some code docs)

Jamiecaty (approved):

packages/blockchain/examples/4444.ts
```diff
@@ -0,0 +1,87 @@
+import { createBlock } from '@ethereumjs/block'
+import { createBlockchain } from '@ethereumjs/blockchain'
+import { Common, Hardfork, Mainnet } from '@ethereumjs/common'
+import { bytesToHex, hexToBytes } from '@ethereumjs/util'
+
+const main = async () => {
+  const common = new Common({ chain: Mainnet, hardfork: Hardfork.Shanghai })
+  // Use the safe static constructor which awaits the init method
+  const blockchain = await createBlockchain({
+    validateBlocks: false, // Skipping validation so we can make a simple chain without having to provide complete blocks
```
holgerd77, line 10: Ok, so the `validateBlocks` flag now also allows to insert blocks not being in the canoncial order, is this correct? (or was this behavior already there bofore) I generally do like this! 🙂 

So, and - otherway around - if `validateBlocks: true`, then it remains forbidden to not go the normal path, right (or not), so 1,2,3,4,...?

holgerd77: Or, not sure if I am fully like it: can't we just fully validate again after this one not validated block (15537393n + 500n). So isn't it only this one block which can't be validated (and therefore would be something like a checkpoint) and then we can (and should) turn validation on again? Or can we do this with the current API already in this example?

g11tech: i just picked the flag from other example, but i think yes we can enable the flag, and yes if the parent is missing then it won't be validated, so optimistic inserts will happen

but when you have put an anchor and move the head forward (which again happens through the forward putblocks along the backfilled chain), then they should be validated

i will remove this from the example and see if it works (most probably should)

holgerd77: Ah, forget about this, this is just simplification for the example, right?

packages/blockchain/src/blockchain.ts
```diff
@@ -369,89 +371,177 @@ export class Blockchain implements BlockchainInterface {
...
-          await this.validateBlock(block)
-        }
+        const { header } = block
+        const blockHash = header.hash()
+        const blockNumber = header.number
 
-        if (this._validateConsensus) {
-          await this.consensus!.validateConsensus(block)
-        }
+        // check if head is still canonical i.e. if this is a block insertion on tail or reinsertion
```
g11tech, line 384: fyi this function is in need to cleanup so ignore for now

packages/blockchain/src/db/operation.ts
```diff
@@ -13,6 +13,8 @@ import {
 
 import type { CacheMap } from './manager.js'
 
+export type PutOpts = { canonical?: boolean; parentTd?: bigint }
```
holgerd77, line 16: Can you move this over to `types.ts` and give both option some docs?
(also for (my) understanding)

jochem-brouwer (commented):
Hi Gajinder, I gave this a look and had a few questions/comments :smile: :+1: 

In general, for both documentation and for review purposes, could you document what this PR is trying to do, and what these new flags are? By the filenames I think this is in the context of EIP-4444. Could you document the new flags in the blockchain, as well as what an Optimistic / Complete block is here in this context? I understand that you are moving the skeleton part away from client and into blockchain, I think that makes sense. However I'm trying to think of how this changes the current behavior of blockchain with these new flags. For instance, what if I insert two optimistic blocks with the same number? What would `getBlock(number)` then return? In client by backfill we would then create two different subchains - right? How would that work here? :thinking:

packages/blockchain/examples/4444.ts
```diff
@@ -0,0 +1,85 @@
...
+  headBlock = await blockchain.getCanonicalHeadBlock()
+  console.log(
+    `Blockchain ${blockchain.consensus.algorithm} Head: ${headBlock.header.number} ${bytesToHex(headBlock.hash())}`,
+  )
+  // Blockchain casper Head: 0 0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3
+
+  // 1. We can put any post merge block as 4444 anchor by using TTD as parentTD
+  // 2. For pre-merge blocks its prudent to supply correct parentTD so as to respect the
+  //    hardfork configuration as well as to determine the canonicality of the chain on future putBlocks
+  await blockchain.putBlock(block1, { parentTd: chainTTD })
```
holgerd77, line 64: I still do not fully understand this anchor concept. What is anchored and how?

packages/blockchain/examples/optimistic.ts
```diff
@@ -0,0 +1,143 @@
...
+  hasBlock3 = (await blockchain.getBlock(block3.header.number).catch((e) => null)) ? true : false
+  console.log(
+    `canonicality: head=${headBlock.header.number}, 0 ... ${block1.header.number}=${hasBlock1} ${block2.header.number}=${hasBlock2} ${block3.header.number}=${hasBlock3} `,
+  )
+  // canonicality: head=0, 0 ... 15537893=false 15537894=true 15537895=true
+
+  // 1. We can put any post merge block as 4444 anchor by using TTD as parentTD
+  // 2. For pre-merge blocks its prudent to supply correct parentTD so as to respect the
+  //    hardfork configuration as well as to determine the canonicality of the chain on future putBlocks
+  await blockchain.putBlock(block1, { parentTd: chainTTD })
```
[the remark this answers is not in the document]
g11tech: blockvalidation depends on parent availability in the blockchain and not on any other factor, for optimistic blocks it doesn't happen.

so lets say this was an optimistic block i.e. its parent was not in blockchain, validate blocks will be skipped. if you add canonical=true here, it will add the number => hash index and will run the head update rules (depending on if this is pow block or pos block) since parentTd has been specified and it will assume that it parent chain is confirmed and doesn't need to be added to blockchain

packages/blockchain/src/blockchain.ts
```diff
@@ -369,89 +371,177 @@ export class Blockchain implements BlockchainInterface {
...
-        const parentTd = await this.getParentTD(header)
-        if (!block.isGenesis()) {
-          td += parentTd
-        }
+        // 1. if canonical is explicit false then just dump the block
+        // 2. if canonical is explicit true then apply that even for the pow/poa blocks
+        //    if they are optimistic, i.e. can't apply the normal rule
+        // 3. if canonical is not defined, then apply normal rules
+        if (opts?.canonical === false || (opts?.canonical === undefined && !isComplete)) {
+          if (parentTd !== undefined) {
```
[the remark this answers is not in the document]
g11tech: in this case: `(opts?.canonical === undefined && !isComplete)` would evaluate to true and the block will just be added to the blockchain as optimistic block, and no head update rules will be run or number => hash index modified

packages/blockchain/examples/4444.ts
g11tech: block1 is the 4444 anchor i.e. the parent chain to block1 doesn't need to be added to the blockchain and blockchain can be build forward from here

it acts as a re-genesis anchor, one specifies an anchor by providing parentTD, for a PoS block parentTD can just be chainTTD, for a PoW block an actual parentTD should be added so as to eventually and correctly matchup with the terminal block TTD when forward chain blocks will be added on top of it

packages/blockchain/examples/optimistic.ts
```diff
@@ -0,0 +1,143 @@
...
+  console.log(
+    `Blockchain ${blockchain.consensus.algorithm} Head: ${headBlock.header.number} ${bytesToHex(headBlock.hash())}`,
+  )
+  // Blockchain casper Head: 0 0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3
+
+  // allows any block > head + 1 to be put as non canonical block
+  await blockchain.putBlock(block3, { canonical: true })
+  headBlock = await blockchain.getCanonicalHeadBlock()
+  blockByHash = await blockchain.getBlock(block3.hash()).catch((e) => null)
+  blockByNumber = await blockchain.getBlock(block3.header.number).catch((e) => null)
```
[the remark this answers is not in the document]
g11tech: yes, its just a pattern used to log the status, sometimes blockByNumber will not be found depending upon the scenario

packages/blockchain/src/blockchain.ts
```diff
@@ -369,89 +371,177 @@ export class Blockchain implements BlockchainInterface {
...
+          let updatesHead = block.isGenesis()
+          let updatesHeadBlock
+          if (parentTd === undefined) {
+            // if the block is pow and optimistic, and has not been explicity marked canonical, then
+            // throw error since pow blocks can't be optimistically added without expicit instruction about
+            // their canonicality
+            if (
+              !block.isGenesis() &&
+              block.common.consensusType() !== ConsensusType.ProofOfStake &&
+              opts?.canonical === undefined
```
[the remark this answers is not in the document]
g11tech: if parentTD is undefined, and this is not an explicity specified canonical PoS block, throw error because we can't determine the canonicality of this block

ideally such a scenario will not happen becuase we will not end up in else part of condition of L424, i will do a cleanup and see if we can come to a simplified logic, it could be redundant

packages/blockchain/test/4444Optimistic.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+    const blockchain = await createBlockchain({ genesisBlock, common, validateBlocks: false })
+
+    // randomly insert the blocks with gaps will be inserted as non canonical
+    for (const block of [block4, block3PoS, block7, block2]) {
+      // without explicit flag their insertion will fail as parent not present and canonicality
+      // can't be established
+      await blockchain.putBlock(block).catch((_e) => {
+        undefined
+      })
+      dbBlock = await blockchain.getBlock(block.hash()).catch((_e) => undefined)
```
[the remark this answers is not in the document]
g11tech: true but since the condition below checks for it to be not undefined, the reason doesnt matter  for the purposes of the test and will need to be debugged

packages/blockchain/test/4444Optimistic.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+      // however the block by number should still not fetch it
+      dbBlock = await blockchain.getBlock(block.header.number).catch((_e) => undefined)
+      assert.equal(
+        equalsBytes(dbBlock.hash(), block.hash()),
+        false,
+        'block number index should not exit',
+      )
+
+      const td = await blockchain.getTotalDifficulty(block.hash()).catch((_e) => undefined)
+      assert.equal(td, undefined, 'block should be not be marked complete')
```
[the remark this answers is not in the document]
g11tech: a complete block is the one which is confirmed to be linked all the way to the genesis via chain of parent blocks.

in our blockchain db, this is marked via block having the totalDifficulty written onto the chain. a boolean flag could also be used, but total difficulty serves well for both PoW and PoS blocks since PoW total difficulty is actually need to make head and canonicality decisions, while for PoS it just acts an an indicator.

while putting a block, if block's parent is in the blockchain and the parent also has totalDifficulty, then the new block is considered complete and it can move head

packages/blockchain/examples/4444.ts
holgerd77: These are one of the things which are too implicit, at least on the API, maybe the DB level. I am still not fully grasping if the `parentTd` is also needed (for calculating/validating something) or is only set in some way to indicate the anchor status.

No matter how it is: this is unnecessarily implicit and we should think about the "anchor" (or call it checkpoint?) status and give this its own API and database representation.

I think "Checkpoint" might actually be the more common and established terminolgy.

So I think we definitely want to have some external relation in the DB where we store all the checkpoints (and not only have this in some very implicit form), and can e.g. extend our API with things like:

```ts
blockchain.getCheckpoints()
blockchain.isCheckpoint(5n)
```

(or first there would be the question if there can be only one of these, but guess there can be several respectively would make sense?)

(wonder if we can even lign in the genesis block in this concept, guess this is exactly the same as a checkpoint?)

And then for setting these checkpoints, this should be very explicitly named. So if `parentTd` is still necessary, even if we store a `Checkpoints -> [ Block Numbers ]` relation, then this should minimally be provided with naming the option `setCheckpoint: 560934n` (and so provide the parentTD still, but set the very emphasis on this checkpoint setting, so one can read from the API call what is happening here).

packages/blockchain/examples/optimistic.ts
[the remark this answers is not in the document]
holgerd77: > blockvalidation depends on parent availability in the blockchain

This is too implicit behavior. If we have set `validateBlocks = true`, blocks should be validated. Period. There shouldn't be inner auto-triggered exceptions from this rule (except maybe for the very explicit task of setting a checkpoint), otherwise one could not rely on the blockchain functionality anymore.

If one wants to deactivate block validation, we should add an option `blockchain.setBlockValidation(false)` or similar.

Guess this should then be accompanied by a flag for blockchain (if we go this route) to indicate that we might deal with a not-fully validated blockchain.

At least all these things (the "state" of the blockchain) should be transparent and requestable, and not just be "implicitly there".

packages/blockchain/test/4444Optimistic.spec.ts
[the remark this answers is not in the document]
holgerd77: We should really really define (write down!) very clearly what we mean with both `canonical` and `complete` here. Again, maybe we are also "not there" with these names yet (e.g. `complete` for a block with this kind of definition is very misleading, one is just thinking in the realm of that the block has all txs, fields, .... We need to urgently find some better and more precise terminology here).

packages/blockchain/src/db/operation.ts
holgerd77: This is still one of the crucial tasks so that we can work on some better understanding and documentation for these options and eventually come to better names.

So, if we have the definitions as comments over the options, this can be reviewed and alternatives/improvements can be suggested.

packages/blockchain/src/blockchain.ts
```diff
@@ -369,89 +371,177 @@ export class Blockchain implements BlockchainInterface {
...
+            await this.validateBlock(block)
+          }
         }
 
-        // calculate the total difficulty of the new block
-        const parentTd = await this.getParentTD(header)
-        if (!block.isGenesis()) {
-          td += parentTd
-        }
+        // 1. if canonical is explicit false then just dump the block
```
acolytec3, line 420: What does `canonical === false` mean here?  Just that we got a block that we know isn't canonical but we want to hold on to it "for future reference"?

packages/blockchain/src/blockchain.ts
```diff
@@ -369,89 +371,177 @@ export class Blockchain implements BlockchainInterface {
...
+            const td = header.difficulty + parentTd
+            dbOps = dbOps.concat(DBSetTD(td, blockNumber, blockHash))
+          }
+          await this.dbManager.batch(dbOps)
+        } else {
+          let updatesHead = block.isGenesis()
+          let updatesHeadBlock
+          if (parentTd === undefined) {
+            // if the block is pow and optimistic, and has not been explicity marked canonical, then
+            // throw error since pow blocks can't be optimistically added without expicit instruction about
```
acolytec3, line 435: I don't think I grasp what this means.  How do we know if a PoW block is canonical if we don't have the chain of blocks with successively greater TD leading up to it?  Presumably we aren't communicating with a CL client in this instance so not sure where this notion of canonicality would come from external to the `blockchain` class.

PR #3585, opened 2024-08-13
Example test runner: make script more readable / ensure examples do not `process.exit`

jochem-brouwer:
The examples test runner tries to run each example in a package, and if one of them throws, the `npm run examples` will fail (and will also throw on CI). However, some examples have `process.exit`, if this is hit, the process exits with a successful code and any errors which might have thrown *could* be discarded (depending if the error gets handled first, or the `process.exit`)

### DIFF
--- a/packages/devp2p/examples/simple.ts
+++ b/packages/devp2p/examples/simple.ts
@@ -43,5 +43,5 @@ for (const bootnode of BOOTNODES) {
 }
 
 setTimeout(() => {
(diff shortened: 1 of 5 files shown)


acolytec3 (approved):
LGTM

holgerd77:
Just to confirm: so this is really no problem on my side (regarding examples creation or whatever)! 🙂

Issue #3560, opened 2024-08-06
Monorepo -> Tree Shaking: Remaining Static Constructor Tasks & Clean-Up


holgerd77:
Ok, this is a issue to finalize on #3487, will go through all the libaries and give a write-up of the remaining constructor refactors + some additional clean-up/refactoring tasks:

### Block

- Kill off `from-rpc.ts`, `header-from-rpc.ts` files, move code within function blocks of `constructors.ts`
- Create `block` and `header` directories, move `block.ts` and `header.ts` over, create `index.ts` file, split `constructors.ts` into the block and header methods and move over respectively (if this full TODO item is controversial we can also give a quick discussion, I would find it a useful better organizing structure)
- Rename `createBlockFromValuesArray` -> `createBlockFromBytesArray`
- Rename `createBlockHeaderFromValuesArray` -> `createBlockHeaderFromBytesArray`

### Util

Withdrawal still missing, accustomed naming scheme:

- `createWithdrawal()`
- `createWithdrawalFromBytesArray()`

No extra file, all in `withdrawal.ts`

Same with Request (rename `requests.ts` to `request.ts` along for consistency, separate commit):

- `createDepositRequest()`
- `createDepositRequestFromJSON()`
- `createWithdrawalRequest()`
- `createWithdrawalRequestFromJSON()`

### Verkle

- New `constructors.ts` file
- `createVerkleTree()`

### VM

Yes, I know, only one (same as Verkle). But we should nevertheless do now for both consistency and extensibility reasons, so the moment we want to add a second constructor we would otherwise be stuck with a mixed scheme.

That should be it. One or separate PRs, as preferred, should both be ok! 🙂

ScottyPoi:
Closing with all tasks completed

holgerd77:
🎉

PR #3586, opened 2024-08-13
Block: treeshaking tasks

ScottyPoi:
Block Treeshaking task cleanup from #3560

- [x] Kill off `from-rpc.ts`, `header-from-rpc.ts` files, move code within function blocks of `constructors.ts`
- [x] Create `block` and `header` directories, move `block.ts` and `header.ts` over, create `index.ts` file, split `constructors.ts` into the block and header methods and move over respectively (if this full TODO item is controversial we can also give a quick discussion, I would find it a useful better organizing structure)
- [x] Rename `createBlockFromValuesArray` -> `createBlockFromBytesArray`
- [x] Rename `createBlockHeaderFromValuesArray` -> `createBlockHeaderFromBytesArray`


packages/block/src/block/constructors.ts
```diff
@@ -57,71 +59,25 @@ import type {
...
-/**
- * Static constructor to create a block header from a header data dictionary
- *
- * @param headerData
- * @param opts
- */
-export function createBlockHeader(headerData: HeaderData = {}, opts: BlockOptions = {}) {
-  return new BlockHeader(headerData, opts)
-}
+export function normalizeTxParams(_txParams: any) {
```
acolytec3, line 62: Two things on this function:
1) It isn't a constructor so ideally we put it in a helpers/utils file.  
2) We have essentially the same function [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/659d4280609943effe5322dcec7747a0cb6e50f8/packages/tx/src/fromRpc.ts#L5).  Can we harmonize these two and consolidate in one location (maybe in `@ethereumjs/util/helpers.ts` or `internal.ts`)  The only real difference I see is converting `0x0` to `0x` in the transaction signatuere values.

acolytec3 (dismissed):
I generally really like this approach.  One nit about `normalizeTxParams` that we should address (though can be in a different PR if you want).

holgerd77 (approved):
LGTM

packages/block/src/block/constructors.ts
```diff
@@ -10,9 +10,9 @@ import {
   CLRequestFactory,
   ConsolidationRequest,
```
Jamiecaty, line 11: - [ ]

Jamiecaty (changes requested):

Jamiecaty (approved):

PR #3592, opened 2024-08-14
Speedup CI with cache

jochem-brouwer:
Fork of https://github.com/ethereumjs/ethereumjs-monorepo/pull/2709

Dumps a CI cache after building the entire monorepo, so further CI jobs do not have to rebuild the repo

TODOs

- [x] Add new packages to use the cache also (wallet, verkle, maybe more?)
- [x] Verify that the CI does the expected job
- [x] Ensure that the cache is updated for each PR, and does not pull old versions (essentially running "old tests" on "old monorepo" thus giving the impression that all is fine, while it runs on old code / old monorepo which passed CI)


holgerd77:
What is the relevant CI job one should look for here then? Would this be just *one* single CI job? Or are there still the single per-package-jobs listed?

acolytec3:
The master job is the `build.yml` file which then invokes the individual module CI test runs. We will still want some separate ones for linting and nightly stuff

holgerd77:
"And after climbing 537 hills, our hero finally succeeded!" 🤩😂

(honestly, totally admire your dedication here)

acolytec3:
I may be aspiring too highly here but it really should work I think. If I do one general cache for the PR, it works fine but we have to get it right. I might just go with the deps optimization for starters because I thought this would work for caching the build step.

.github/workflows/build.yml
```diff
@@ -0,0 +1,167 @@
...
+          cache: 'npm'
+
+      - run: npm ci
+        working-directory: ${{github.workspace}}
+        
+      - name: Cache build
+        uses: actions/cache/save@v4
+        with:
+          path: ${{github.workspace}}
+          key: npm-${{ hashFiles('package-lock.json') }}
```
jochem-brouwer, line 43: I don't think this key is correct, this will restore cache (so including the build items) if the package-lock.json matches. So, if I push any change, the cache will thus start at the old version of the code? If that code version passes all tests, and if I then introduce a bug, the CI would then not flag it - right? (Since it runs tests on the old, valid, code?)

acolytec3: Should be fixed by 824345f2604ec6257ecf5d2db9f389e8f93c1b1c

holgerd77:
Could this get a quick summary (but going beyond the initial description) what has been done and what has been achieved? Thanks! 🙂 🙏

holgerd77:
Also: still in Draft?

acolytec3 (approved):

jochem-brouwer:
LGTM

PR #3597, opened 2024-08-16
tx: rename method names (remove EIP word)

jochem-brouwer:
This PR does one thing and that is to remove the EIP word from the Tx library methods and classes. (Also renames constructors as `create7702EOACodeTx` to `createEOACode7702Tx`)

I have only used "find-and-replace" so if CI passes this should be fine.

To quickly test the renames, build the library, and then in node `require('@ethereumjs/tx')` and verify the output names of the exported methods/classes.


holgerd77 (approved):
Looks and reads absolutely great! 🤩 

Will merge.

PR #3599, opened 2024-08-16
Remove almost all `cspell:ignore`

jochem-brouwer:
This PR removes almost all `cspell:ignore` on top of script files.

Notes:

- Excluded `./packages/client/archive` for cspell. Once we reuse it, it will automatically get into the spell check pool
- Did not remove `cspell:ignore` in wallet, because I don't know what those values mean, and I don't want to touch the wallet code

TODO

- [x] I realized that there might be `// cspell` comments into the docs of methods/classes, this will thus also end up in the end doc. Need to figure out how to fix this, since wrapping the entire comment in `cspell:disable` an then `cspell:enable` will effectively disable the spell check for those docs which is not what we want.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "checkNpmVersion": "./scripts/check-npm-version.sh",
     "clean": "./config/cli/clean-root.sh",
     "cspell": "npm run cspell:ts && npm run cspell:md",
-    "cspell:ts": "npx cspell --gitignore -e \"./packages/ethereum-tests\" -e \"./packages/wallet/test\" -c ./config/cspell-ts.json \"./packages/**/*.ts\" --cache --show-suggestions --show-context",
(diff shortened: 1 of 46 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "checkNpmVersion": "./scripts/check-npm-version.sh",
     "clean": "./config/cli/clean-root.sh",
     "cspell": "npm run cspell:ts && npm run cspell:md",
-    "cspell:ts": "npx cspell --gitignore -e \"./packages/ethereum-tests\" -e \"./packages/wallet/test\" -c ./config/cspell-ts.json \"./packages/**/*.ts\" --cache --show-suggestions --show-context",
+    "cspell:ts": "npx cspell --gitignore -e \"./packages/ethereum-tests\" -e \"./packages/wallet/test\" -e \"./packages/client/archive\" -c ./config/cspell-ts.json \"./packages/**/*.ts\" --cache --show-suggestions --show-context",
```
holgerd77, line 11: Nit: can't more of these options put in the configuration file so that we can keep it simpler here?

PR #3601, opened 2024-08-17
VM: Bundle Optimizations

holgerd77:
WIP

### DIFF
--- a/packages/vm/test/api/copy.spec.ts
+++ b/packages/vm/test/api/copy.spec.ts
@@ -18,12 +18,6 @@ describe('VM Copy Test', () => {
       'account exists before copy',
     )
 
-    const vmCopy = await vm.shallowCopy()
-    assert.isUndefined(
-      await vmCopy.stateManager.getAccount(address),
-      'non-committed checkpoints will not be copied',
-    )
-
     await vm.stateManager.checkpoint()
     await vm.stateManager.commit()
 
--- a/packages/statemanager/src/stateManager.ts
+++ b/packages/statemanager/src/stateManager.ts
@@ -202,8 +202,14 @@ export class DefaultStateManager implements StateManagerInterface {
    * @param value - The value of the `code`
    */
   async putCode(address: Address, value: Uint8Array): Promise<void> {
-    this._caches?.code?.put(address, value)
     const codeHash = this.keccakFunction(value)
+
+    if (this._caches?.code !== undefined) {
+      this._caches!.code!.put(address, value)
+    } else {
+      const key = this._prefixCodeHashes ? concatBytes(CODEHASH_PREFIX, codeHash) : codeHash
+      await this._getCodeDB().put(key, value)
+    }
 
     if (this.DEBUG) {
       this._debug(`Update codeHash (-> ${short(codeHash)}) for account ${address}`)
--- a/packages/statemanager/src/statelessVerkleStateManager.ts
+++ b/packages/statemanager/src/statelessVerkleStateManager.ts
@@ -25,10 +25,11 @@ import debugDefault from 'debug'
 import { keccak256 } from 'ethereum-cryptography/keccak.js'
 
 import { AccessWitness, AccessedStateType, decodeValue } from './accessWitness.js'
-import { Caches, OriginalStorageCache } from './cache/index.js'
+import { OriginalStorageCache } from './cache/index.js'
 import { modifyAccountFields } from './util.js'
 
 import type { AccessedStateWithAddress } from './accessWitness.js'
+import type { Caches } from './cache/index.js'
 import type { StatelessVerkleStateManagerOpts, VerkleState } from './index.js'
 import type { DefaultStateManager } from './stateManager.js'
 import type { AccountFields, Proof, StateManagerInterface } from '@ethereumjs/common'
@@ -206,9 +207,9 @@ export class StatelessVerkleStateManager implements StateManagerInterface {
    * at the last fully committed point, i.e. as if all current
    * checkpoints were reverted.
    */
-  shallowCopy(): StatelessVerkleStateManager {
+  shallowCopy(downlevelCaches = true): StatelessVerkleStateManager {
     const stateManager = new StatelessVerkleStateManager({
-      caches: this._caches !== undefined ? new Caches() : undefined,
+      caches: this._caches?.shallowCopy(downlevelCaches),
       verkleCrypto: this.verkleCrypto,
     })
     return stateManager
--- a/packages/statemanager/test/stateManager.code.spec.ts
+++ b/packages/statemanager/test/stateManager.code.spec.ts
@@ -15,162 +15,167 @@ import type { AccountData } from '@ethereumjs/util'
 
 describe('StateManager -> Code', () => {
   for (const accountCacheOpts of [{ size: 1000 }, { size: 0 }]) {
-    it(`should store codehashes using a prefix`, async () => {
-      /*
-        This test is mostly an example of why a code prefix is necessary
-        I an address, we put two storage values. The preimage of the (storage trie) root hash is known
-        This preimage is used as codeHash
+    for (const codeCacheOpts of [{ size: 1000 }, { size: 0 }]) {
+      it(`should store codehashes using a prefix`, async () => {
+        /*
+          This test is mostly an example of why a code prefix is necessary
+          I an address, we put two storage values. The preimage of the (storage trie) root hash is known
+          This preimage is used as codeHash
+  
+          NOTE: Currently, the only problem which this code prefix fixes, is putting 0x80 as contract code
+          -> This hashes to the empty trie node hash (0x80 = RLP([])), so keccak256(0x80) = empty trie node hash
+          -> Therefore, each empty state trie now points to 0x80, which is not a valid trie node, which crashes @ethereumjs/trie
+        */
 
-        NOTE: Currently, the only problem which this code prefix fixes, is putting 0x80 as contract code
-        -> This hashes to the empty trie node hash (0x80 = RLP([])), so keccak256(0x80) = empty trie node hash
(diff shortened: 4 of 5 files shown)


holgerd77:
Ah, and here for completeness also the final bundle image. So the interesting thing is actually what is *missing*. So: full blockchain is missing, (more or less) all caches code from StateManager, the stateless Verkle SM, the tx types (actually, most of: there are some still in (4844), but this has other (code) reasons, the block `block.js` code: 🙂 

grafik

packages/evm/src/precompiles/util.ts
```diff
@@ -1,5 +1,33 @@
...
 
+/**
+ * Checks that the gas used remain under the gas limit.
+ *
+ * @param opts
+ * @param gasUsed
+ * @param pName
+ * @returns
+ */
+export const gasLimitCheck = (opts: PrecompileInput, gasUsed: bigint, pName: string) => {
```
acolytec3, line 13: I really like this.

packages/evm/src/opcodes/codes.ts
```diff
@@ -52,159 +52,173 @@ export type OpcodeList = Map<number, Opcode>
 type OpcodeEntry = { [key: number]: { name: string; isAsync: boolean; dynamicGas: boolean } }
 type OpcodeEntryFee = OpcodeEntry & { [key: number]: { fee: number } }
 
+// Default: sync and no dynamic gas
```
acolytec3, line 55: Can you explain what all of this is doing?  I'm not anxious about the change per se but what is this achieving for the purposes of tree-shaking?

packages/evm/src/eof/verify.ts
```diff
@@ -101,31 +101,21 @@ function validateOpcodes(
...
 
   // Remove CODESIZE, CODECOPY, EXTCODESIZE, EXTCODECOPY, EXTCODEHASH, GAS
-  opcodeNumbers.delete(0x38)
-  opcodeNumbers.delete(0x39)
-  opcodeNumbers.delete(0x5a)
-  opcodeNumbers.delete(0x3b)
-  opcodeNumbers.delete(0x3c)
-  opcodeNumbers.delete(0x3f)
-
+  const a1 = [0x38, 0x39, 0x5a, 0x3b, 0x3c, 0x3f]
```
acolytec3, line 104: What is all of this doing?  Are we just saving Lines of Code/character counts to make this change?  

Bigger picture, is there a more comprehensible way to organize our lists of valid opcodes per EVM context (i.e. what opcodes are valid under EOF only and what opcodes are valid in legacy EVM context)?  This whole code section is kind of hard to understand as to why we are deleting certain opcodes from the valid list.  A "better way" in my opinion would be to have two configurations, one valid for EOF and one for legacy EVM mode.

packages/evm/src/precompiles/index.ts
```diff
@@ -55,12 +55,12 @@ interface PrecompileAvailabilityCheckTypeEIP {
   type: PrecompileAvailabilityCheck.EIP
   param: number
 }
-
-const ripemdPrecompileAddress = '0000000000000000000000000000000000000003'
+const BYTES_19 = '00000000000000000000000000000000000000'
```
acolytec3, line 58: I'm not 100% sure I love this as an optimization.  Feels a tiny bit like a bridge too far for tree-shaking at the expense of readability but can live with it if it gains that much for us.

acolytec3 (approved):
I'm going to cautiously approve this one, though a little ambivalent that we are getting close to the point where we're taking things too far (see my comment on readability vs compactness).

packages/evm/src/opcodes/codes.ts
holgerd77: This is not tree shaking, this is simply code compactification chopping of a few bytes. I would also argue here that this reads *better* than before than going through this endless list since it more prominently points out the deviations, e.g. `dynamicGasOp('MLOAD')`.

Will keep this one.

packages/evm/src/eof/verify.ts
holgerd77: Have removed the commit

acolytec3 (approved):

packages/evm/src/opcodes/codes.ts
```diff
@@ -52,159 +52,173 @@ export type OpcodeList = Map<number, Opcode>
...
-  0x34: { name: 'CALLVALUE', isAsync: true, dynamicGas: false },
-  0x35: { name: 'CALLDATALOAD', isAsync: true, dynamicGas: false },
-  0x36: { name: 'CALLDATASIZE', isAsync: true, dynamicGas: false },
-  0x37: { name: 'CALLDATACOPY', isAsync: true, dynamicGas: true },
-  0x38: { name: 'CODESIZE', isAsync: false, dynamicGas: false },
-  0x39: { name: 'CODECOPY', isAsync: false, dynamicGas: true },
-  0x3a: { name: 'GASPRICE', isAsync: false, dynamicGas: false },
-  0x3b: { name: 'EXTCODESIZE', isAsync: true, dynamicGas: true },
-  0x3c: { name: 'EXTCODECOPY', isAsync: true, dynamicGas: true },
+  0x30: asyncOp('ADDRESS'),
```
jochem-brouwer, line 103: Wow, just noticing this here, it seem that we have marked some opcodes (also in the original) with the wrong type. `ADDRESS` is definitely not async! I will address typing those right in a new PR.

packages/evm/src/types.ts
```diff
@@ -343,9 +343,13 @@ export interface EVMOpts {
   stateManager?: StateManagerInterface
 
   /**
+   * The EVM comes with a basic mock blockchain interface and implementation for
+   * non-block containing use cases.
    *
+   * For block-containing setups use the full blockchain implementation from the
```
jochem-brouwer, line 349: The only use case in the EVM for Blockchain is to use the BLOCKHASH opcode. So, in a context where BLOCKHASH is not being used / ran, using the mockchain is fine

packages/vm/examples/test.ts
```diff
@@ -0,0 +1,6 @@
+import { StatelessVerkleStateManager } from '@ethereumjs/statemanager'
+import { loadVerkleCrypto } from 'verkle-cryptography-wasm'
+
+const verkleCrypto = await loadVerkleCrypto()
+const sm = new StatelessVerkleStateManager({ verkleCrypto })
+console.log(sm)
```
jochem-brouwer, line 6: This file should likely be removed

packages/vm/src/runTx.ts
```diff
@@ -52,6 +51,8 @@ import type {
 const debug = debugDefault('vm:tx')
 const debugGas = debugDefault('vm:tx:gas')
 
+const DEFAULT_HEADER = createBlockHeader()
```
jochem-brouwer, line 54: Doesn't this import the block header in runTx? Why do we need this? Could we not instantiate this with the appropriate fields here, so we don't import the whole BlockHeader class? The header needs: `coinbase`, `baseFeePerGas`, `gasLimit`, `getBlobGasPrice()`, `gasUsed` (can mock this one in another PR if wanted)

packages/vm/test/api/copy.spec.ts
```diff
@@ -18,12 +18,6 @@ describe('VM Copy Test', () => {
       'account exists before copy',
     )
 
-    const vmCopy = await vm.shallowCopy()
-    assert.isUndefined(
-      await vmCopy.stateManager.getAccount(address),
-      'non-committed checkpoints will not be copied',
```
jochem-brouwer, line 24: I'm not entirely sure why this is deleted, am I missing something?

holgerd77: The behavior just changed here, non-committed checkpoints *are* now copied in the default case of non caches used (which is ok, the checkpointing is in this case fully done by the trie checkpointing mechanism).

Since this check is not testing a required property but just a it-was-like-this-in-this-case property which now changed I removed the test (the alternative would have been to setup a full with-caches-without-caches setup respetively introduce some differentiation here, which I totally found overblown - again, since this is just not testing something which needs to have the tested property.

Hope this is clear. 🙂

packages/evm/src/opcodes/codes.ts
holgerd77: Yeah, so I *do* still thing that this new more compact notation is useful (beyond small size gains), since it more points the properties out (I would assume that you actually spotted due to this new form factor).

jochem-brouwer: Yup, that's true. I actually thought you converted the opcodes wrong until I realized that it was wrong in the first place :joy: This is definitely better and makes it much more readable! :smile:

packages/evm/src/types.ts
holgerd77: Yeah, I would nevertheless not want to write something like "if you use BLOCKHASH" here since in many cases this is just not known before.

So I would think this phrasing is fine, people will know if they have/need a blockchain containing real data at hand.

packages/vm/examples/test.ts
holgerd77: Will remove, accidentally committed

jochem-brouwer (approved):
LGTM (assuming CI passes)

packages/evm/src/opcodes/codes.ts
holgerd77: Side question (out of interest): does this have any effect (on performance e.g.) if changed back to `sync`? Or is this more for correctness?

jochem-brouwer: I think it will have a tiny impact since if it is async, it will likely end up in a promise queue, so this would at least "save" a push and pop operation (but this depends on the inner workings of node, so I have no idea if this is actually the case). But it should be changed nevertheless for correctness.

PR #3603, opened 2024-08-18
Implement t8ntool to use for execution-spec-tests

### LINKED ISSUE #3602: Implement T8N support (tracking issue)
EthereumJS currently does not T8N support, which is the "transition-tool". This would be incredibly helpful, especially regarding filling tests on the `execution-spec-tests` side to help filling the tests. If the tests on their side are written, but no client implementation is ready (or: does not have t8n support), then the tests cannot be filled and thus other clients cannot consume those tests (in the form of fixtures, i.e. state, blockchain (and hive) tests).

T8Ntool docs: https://ethereum-tests.readthedocs.io/en/latest/t8ntool-ref.html

`execution-spec-tests` will need a EthereumJS t8n wrapper: https://github.com/ethereum/execution-spec-tests/tree/main/src/evm_transition_tool

If done rather quickly we could use this to fill the EIP-7702 tests, as in this `execution-spec-tests` PR: https://github.com/ethereum/execution-spec-tests/pull/733. This will need https://github.com/ethereumjs/ethereumjs-monorepo/pull/3581

We have retesteth support which looks like t8n-like: https://github.com/ethereumjs/ethereumjs-monorepo/pull/2818 (sample PR for fixups, but `./packages/vm/test/retesteth`)

Add t8ntool to client docs from `ethereum-tests` https://ethereum-tests.readthedocs.io/en/develop/t8ntool.html

NOTE: this issue is WIP

CC @spencer-tb

jochem-brouwer:
Resolves #3602 

This PR implements T8NTool, used for `execution-spec-tests` to fill (and in the future: write) tests. To see how it works, check the t8ntool readme in `vm/test/t8n/README.md`.


packages/vm/test/t8n/t8n.ts
```diff
@@ -0,0 +1,328 @@
...
+  }
+}
+
+const outputAlloc = alloc
+
+const outputResultFilePath = join(args.output.basedir, args.output.result)
+const outputAllocFilePath = join(args.output.basedir, args.output.alloc)
+
+writeFileSync(outputResultFilePath, JSON.stringify(output))
+writeFileSync(outputAllocFilePath, JSON.stringify(outputAlloc))
```
holgerd77, line 328: Ugh. So much code to write here. 🤯

jochem-brouwer: Yes, but it works :D

jochem-brouwer:
Do not spend time to review this, the branch is extremely messy

jochem-brouwer:
@gabrocheleau Yes, this tool allows us to "fill" those tests. This includes checks which are in those tests, for instance it can test if the post-state of something (for instance a storage slot, or a balance) matches the expected case. The output is the `fixtures` format, which we can also directly consume using blockchain/state tests. 

I also noticed there are verkle releases of those tests, these can be directly consumed by our blockchain/state test runner. The only thing we have to do there is to ensure we are on the right fork (i.e. Verkle). Those tests can be found here: https://github.com/ethereum/execution-spec-tests/releases/tag/verkle%40v0.0.3

packages/vm/test/t8n/transitionTool.ts
```diff
@@ -0,0 +1,381 @@
...
+      // Hash not found, so return the zero hash
+      return {
+        hash() {
+          return zeros(32)
+        },
+      }
+    }
+
+    await mcl.init(mcl.BLS12_381)
+    const bls = new MCLBLS(mcl)
```
holgerd77, line 82: Don't know, do we want to use MCL here? Or do we rather also want to use the pure/default JS version?

packages/vm/src/buildBlock.ts
```diff
@@ -242,7 +245,11 @@ export class BlockBuilder {
 
       // Guard against the case if a tx came into the pool without blobs i.e. network wrapper payload
       if (blobTx.blobs === undefined) {
-        throw new Error('blobs missing for 4844 transaction')
+        // TODO: verify if we want this, do we want to allow the block builder to accept blob txs without the actual blobs?
+        // (these must have at least one `blobVersionedHashes`, this is verified at tx-level)
```
acolytec3, line 249: What is the context for this?  Do we ever build blocks with the versioned hash but no blob or is that specific to t8n?

jochem-brouwer: Should maybe update the comment here. This is if we allow to build a block in the block builder where the blob itself is not available, but the `blobVersionedHashes` in the 4844 tx is. (This is the case in t8ntool, the hashes are there, the actual data is not there. Note that this does not really matter, since EVM does not have access to this data anyways)

acolytec3: Ah, ok, then yes, this should be fine (though I agree that we should only allow this behind a guard so it's not run "in production" this way).

packages/vm/test/t8n/README.md
```diff
@@ -0,0 +1,40 @@
...
+## Using T8Ntool to fill `execution-spec-tests`
+
+To fill `execution-spec-tests` (or write own tests, and test those against the monorepo), follow these steps:
+
+1. Clone <https://github.com/ethereum/execution-spec-tests/>.
+2. Follow the installation steps: <https://github.com/ethereum/execution-spec-tests?tab=readme-ov-file#quick-start>.
+
+To fill tests, such as the EIP-1153 TSTORE/TLOAD tests, run:
+
+- `fill -vv -x --fork Cancun tests/cancun/eip1153_tstore/ --evm-bin=../ethereumjs-monorepo/packages/vm/test/t8n/ethereumjs-t8ntool.sh`
```
acolytec3, line 14: Should this command work correctly?  I get validation errors from the CLI when I run this particular test and it's not obvious to me if I mangled something trying to install the execution tests or if this is a flaw on our side?

packages/vm/test/t8n/ethereumjs-t8ntool.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+if [[ "$1" == "--version" ]]; then
+    echo "ethereumjs t8n v1"
+    exit 0
+fi
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+export NODE_OPTIONS="--max-old-space-size=4096" 
+tsx "$SCRIPT_DIR/launchT8N.ts" "$@"
```
acolytec3, line 8:
```suggestion
npx tsx "$SCRIPT_DIR/launchT8N.ts" "$@"
```

This way it will run even if I only have `tsx` installed locally within ethjs and not globally.

packages/vm/test/t8n/README.md
acolytec3: Disregard.  This was user error.

acolytec3 (dismissed):
LGTM.  Was able to get the fill process to run.

acolytec3 (approved):

jochem-brouwer:
Thanks a lot all, this was a huge process, but super happy to get this merged now :smile: 

Together with https://github.com/ethereum/execution-spec-tests/pull/752 we are now ready "out-of-the-box" to fill tests :smile: :rocket:

holgerd77:
🎉🎉🎉

packages/vm/test/t8n/transitionTool.ts
holgerd77: I wonder though if we should rather generally move over to WASM with this (so basically the setup we have in the client), since the tool atm still is relatively slow and this is already noticeable, e.g. on filling the 1153 tests, this already takes a solid amount of time.

Maybe we also want to decide where it's worth to switch - e.g. for very basic and settled stuff like `keccak` or `ecrecover` we can very likely do (?) - and for other things (e.g. Noble/MCL for BLS might be a case here) we might rather want to test/fill with our native code.

Thoughts? Opinions?

jochem-brouwer:
Hi @holgerd77, the slow stuff of t8n has not much to do with using MCL. It is simply the load time of all our libraries. It is done each time t8n is called, which can be done multiple times per test (in fact, for each block t8n is called). 

I have on the agenda a separate PR to open up a server daemon which will then remove this "cold load" time and instead "hot load" the VM. This will drastically speed up the runtime of this tool (for multiple tests).

You could think of this when running the Blockchain tests, except now you dont run it with the default settings but instead you run `npm run test:blockchain -- --test=TEST`. This is essentially what is happening right now with t8n, which is obviously very slow.

Note that for BLS tests MCL is indeed somewhat faster (we have seen in the Blockchain tests a decrease from ~7 seconds to something like ~6 seconds). However that is for all tests, this ~1 second change is already what would be saved if we would "hot load" the libraries.

I plan to add this server daemon asap!

holgerd77:
> Hi @holgerd77, the slow stuff of t8n has not much to do with using MCL. It is simply the load time of all our libraries. It is done each time t8n is called, which can be done multiple times per test (in fact, for each block t8n is called).
> 
> I have on the agenda a separate PR to open up a server daemon which will then remove this "cold load" time and instead "hot load" the VM. This will drastically speed up the runtime of this tool (for multiple tests).
> 
> You could think of this when running the Blockchain tests, except now you dont run it with the default settings but instead you run `npm run test:blockchain -- --test=TEST`. This is essentially what is happening right now with t8n, which is obviously very slow.
> 
> Note that for BLS tests MCL is indeed somewhat faster (we have seen in the Blockchain tests a decrease from ~7 seconds to something like ~6 seconds). However that is for all tests, this ~1 second change is already what would be saved if we would "hot load" the libraries.
> 
> I plan to add this server daemon asap!

Ah ok, makes a lot of sense, thanks for the insight! 🤯 

Also thanks for planning ahead on this already! 🙏

Issue #3593, opened 2024-08-15
Monorepo: Nightly VM Tests fail


holgerd77:
Our nightly VM tests fail for weeks now, see e.g. [here](https://github.com/ethereumjs/ethereumjs-monorepo/actions/runs/10397207940).

It would really be good, if someone would look into that. Might be a small-ish thing (fails directly on installation with some dependency resolution problem).

jochem-brouwer:
Tests now run, but we have a blockchain test runner problem, e.g. job https://github.com/ethereumjs/ethereumjs-monorepo/actions/runs/10428382061

```
2024-08-17T02:57:33.8041160Z Errors thrown in file: TransitionTests/bcBerlinToLondon/londonUncles.json test: londonUncles_BerlinToLondonAt5:
2024-08-17T02:57:33.8044471Z 	Error: calcNextBaseFee() can only be called with EIP1559 being activated (block header number=3 hash=0x8dee0dbbf4b9a228734395e0be8021808664d8a9ee02e9e7f12d6067d260e476 hf=berlin baseFeePerGas=none)
2024-08-17T02:57:33.8047137Z 	Error: value for 0x0eab710deceb909e553d95bcc14065cc81a6197f334cb545065f7280e634a15c not found in DB
2024-08-17T02:57:33.8048840Z 	Error: value for 0x6148adba5f62e3b924cdc1630cd7b9b1d52ef623eefa395f9761c670980ec774 not found in DB
2024-08-17T02:57:33.8050240Z 	Error: value for 0x574dfd05aff119f7869aae03a845921a77e109c2b2c948d10f07417cb82fd1d8 not found in DB
2024-08-17T02:57:33.8051285Z 	Error: value for 0x575e520c911fcad10b7d437f68256c760e91c81b697bd70afd83c0ea39838fbd not found in DB
2024-08-17T02:57:33.8052925Z 	Error: value for 0xc0570502136f2b0d3f1bf76b22e99614b9e48ede6c478ffc8408d3ecb4510a60 not found in DB
2024-08-17T02:57:33.8054642Z 	Error: value for 0x9b973bf69e337207759cda9905a2b8c498430f2cd8bae9a4a6b7f3cc7e439f14 not found in DB
2024-08-17T02:57:33.8055684Z 	correct last header block
[...]

Issue #3572, opened 2024-08-09
Common: Re-add Deep Copy for Params


holgerd77:
Just quickly dropping here so that it does not get forgotten, we need to re-add deep copy of the params in Common here https://github.com/ethereumjs/ethereumjs-monorepo/pull/3545#discussion_r1703738945 and somewhere Jochem mentioned in one of these PRs (guess for the copy()-type method).

Otherwise this will have side effects in some cases.

(e.g. Common initialized with some param set, changing a param from the original param set for whatever reasons will change the param in Common if we do not deep-copy).

(might do this myself, just not in "dev mode" right now)

PR #3608, opened 2024-08-20
Common deep copy refactor

scorbajio:
This change refactors the `Common` class to deep copy parameters on calls to `copy`. It aims to address issue #3593 and #3572.


acolytec3 (dismissed):
LGTM

packages/common/test/chains.spec.ts
```diff
@@ -23,6 +24,33 @@ describe('[Common/Chains]: Initialization / Chain params', () => {
...
+      console.log('dbg101')
+      console.log(cCopy)
+      assert.equal(cCopy.chainName(), 'mainnet', 'should return original chain name')
+      assert.equal(cCopy.chainId(), BigInt(1), 'should return original chain Id')
+    }
+
+    await setCommon()
+    await resetCommon()
+  })
+
```
holgerd77: I am honestly very confused by this whole setup, so these tests trigger the issue but were later *removed*?? 😂

I will copy (cherry-pick) them over in a new PR and add the changes I initially intended.

holgerd77: (and see if tests are passing)

packages/common/src/common.ts
```diff
@@ -69,7 +71,7 @@ export class Common {
         (this._chainParams.customHardforks && this._chainParams.customHardforks[hf.name]),
     ])
     this._hardfork = this.DEFAULT_HARDFORK
-    this._params = { ...(opts.params ?? {}) } // copy
+    this._params = opts.params ? JSON.parse(JSON.stringify(opts.params)) : {} // copy
```
holgerd77: Ok, so this is actually the single line change I initially intended, likely with the ones from `updateParams()` and `resetParams()` (good catch!), no changes intended for `copy()` whatsoever.

Will tests these lines in a new PR (maybe again, eventually you've been to this point already), then this should already be enough for this (or a replacing) PR!

(we also should not mix this with #3593 , it's absolutely not clear that these are related)

packages/common/src/common.ts
```diff
@@ -58,6 +59,7 @@ export class Common {
   public events: EventEmitter
 
   constructor(opts: CommonOpts) {
+    this._opts = opts
     this.events = new EventEmitter()
 
     this._chainParams = opts.chain
```
holgerd77: Ah, but you are right, we need additional treatment for `chain`. But we should do here, then we can avoid doing any `copy()` changes.

Also (more important): if we do *not* here, then the initial `Mainnet` (or whatever chain) structure gets modified *in the first place*. That's bad and we should very much avoid (same for the `params` structs, I stumbled upon this kind of failure).

packages/common/test/chains.spec.ts
```diff
@@ -23,6 +24,33 @@ describe('[Common/Chains]: Initialization / Chain params', () => {
     )
   })
 
+  it('Deep copied common object should have parameters that are independent of the original copy', async () => {
+    let chainConfig: ChainConfig
+    let c: Common
+    const setCommon = async () => {
+      chainConfig = Mainnet
```
holgerd77: Haha, so funny, I am pretty sure (or: could very much imagine) that this is where you turned wrong: so, by associating `chainConfig = Mainnet` here and *then* manipulating this further down the line with `      chainConfig.chainId = 2` this then also manipulates `Mainnet` itself.

And *that* causes another test below in line 55 to fail. 😂 And the error looks pretty much like a "still failing" message if one is not looking pretty closely. 🙂

holgerd77:
Have replaced with #3613, will close here.

Ok, but this work was definitely not "wasted", since this discovered various additional necessary deep-copy cases + could re-use the test setup! 🙏

PR #3589, opened 2024-08-14
Util: treeshaking tasks

ScottyPoi:
Util treeshaking task cleanup from #3560 

- `withdrawal.ts`
  - `Withdrawal` static constructors extracted to functions
- `requests.ts` renamed `request.ts`
  - static constructors extracted to functions
    - `WithdrawalRequest`
    - `ConsolidationRequest`
    - `DepositRequest`

PR #3611, opened 2024-08-22
Update readmes and examples

scorbajio:
The goal of this change is to update the package READMEs to reflect the latest changes to the examples files that are being embedded in them, and also to check and update the READMEs with more examples and details where necessary to document newly breaking changes to packages.


holgerd77 (approved):
Thanks for tackling 🙂! This is touching already so many code lines, I will take this in, hope that's ok (but I don't see any downsides also on second thought).

Please then just do eventual subsequent work in follow-up PRs! Thanks! 🙏

scorbajio:
> Thanks for tackling 🙂! This is touching already so many code lines, I will take this in, hope that's ok (but I don't see any downsides also on second thought).
> 
> Please then just do eventual subsequent work in follow-up PRs! Thanks! 🙏

The only thing to note is that there is one remaining breaking change that is not yet merged (#3589). So that PR should also include an update to the `util` package docs and any other packages that may need their docs updated with new example embeddings.

holgerd77:
> The only thing to note is that there is one remaining breaking change that is not yet merged (#3589). So that PR should also include an update to the `util` package docs and any other packages that may need their docs updated with new example embeddings. Unless it's going to be done in a separate PR.

Ah, yeah, ok, I think there are just plently of areas in the docs where there are still some updates needed. Guess this just generally needs a good final look at the end. 🙂 (but I think it's likely ok to do semi-structured updates in between, not sure if even possible to do in a super-consistent way)

PR #3614, opened 2024-08-26
VM: exit early on non-existing system contracts

jochem-brouwer:
T8NTool / test fix: we touch system accounts and leave them empty once generating tests (even if those do not exist): this PR exits early and does not write non-existing (as opposed to - *SIGH* empty accounts) to the trie


jochem-brouwer:
Not ready, creates weird results using t8n. Will mark ready for review once ready

jochem-brouwer:
PR is done, this:

1. Exits early if there is no code deposited in the system contract (so this is not 0xffff..fffe, but rather the system contract, usually an account starting with a lot of zeros)
2. The system address is restored to the original state of the system address after the system call. (EVM bumps nonce, this should be rolled back)
3. Note before, we threw if code does not exist at the system contract. However, there are t8ntool tests (Prague tests) which expects the pre-state to be non-filled with system contracts, and therefore the output should also not contain those (we could however not run these blocks, since these threw).

acolytec3 (approved):
LGTM

PR #3616, opened 2024-08-27
verkle: extract static constructor

ScottyPoi:
ref issue: #3560

**VerkleTree:**
moves state `VerkleTree.create` to 'constructors.ts' 
renames it to `createVerkleTree`

### DIFF
--- a/packages/verkle/src/verkleTree.ts
+++ b/packages/verkle/src/verkleTree.ts
@@ -9,8 +9,6 @@ import {
 } from '@ethereumjs/util'
 import debug from 'debug'
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-import { createVerkleTree } from './constructors.js'
 import { CheckpointDB } from './db/checkpoint.js'
 import { InternalNode } from './node/internalNode.js'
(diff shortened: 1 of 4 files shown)


acolytec3 (approved):
LGTM

packages/verkle/src/verkleTree.ts
```diff
@@ -24,6 +21,8 @@ import {
   type VerkleTreeOptsWithDefaults,
 } from './types.js'
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import type { createVerkleTree } from './constructors.js' // Imported so intellisense can display docs
```
holgerd77, line 25: Ah, yeah, I stumbled over this as well at some point. It would really be good if this can be solved on a more global scale (all such imports auto-added or so), this problem is all over the place and it would be good if we could use this `@link` feature from the docs more reliably.

(just as some PR-unrelated side note)

acolytec3: Yup, have done some research on this but haven't quite solved it yet. We need to somehow link via something like `{@link module#referebcedType }` but I haven't figured out specifically how to implement yet.

PR #3617, opened 2024-08-27
Update vitest deps to v2

acolytec3:
Updates the vitest dependency set to v2 from the v2 beta releases.


gabrocheleau (approved):

holgerd77:
Out of interest: did this final version bump from beta bring any additional changes? (relevant to us)

gabrocheleau:
> Out of interest: did this final version bump from beta bring any additional changes? (relevant to us)

Just checked, no there are no breaking changes between 2.0 beta12 and 2.0 that are relevant to us. Btw, between beta12 and 2.0 there was only beta13, so the changes with 2.0 are very minimal.

PR #3550, opened 2024-07-27
Block: extract static Header constructors

ScottyPoi:
Continues work on #3487 

Extract Header static constructors into `block/src/constructors.ts`

`BlockHeader.fromHeaderData` > `createHeader`
`BlockHeader.fromRLPSerializedHeader` > `createHeaderFromRLP`
`BlockHeader.fromValuesArray` > `createHeaderFromValuesArray`

PR #3619, opened 2024-08-27
Update block docs with header renamings

scorbajio:
This change updates the `block` package documentation with new breaking changes from #3550. 

While working on this change, it was discovered that the `block` package README uses code examples that are not being embedded from source example files from the examples folder. These examples need to be moved into their own executable files and embedded into the README with `npm run examples:build`. This will be done in a followup PR.

### DIFF
--- a/packages/block/README.md
+++ b/packages/block/README.md
@@ -445,7 +445,7 @@ For sealing a block on instantiation you can use the `cliqueSigner` constructor 
 
 ```ts
 const cliqueSigner = Buffer.from('PRIVATE_KEY_HEX_STRING', 'hex')
-const block = Block.fromHeaderData(headerData, { cliqueSigner })
+const block = createSealedCliqueBlock(blockData, cliqueSigner)
 ```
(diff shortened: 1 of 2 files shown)


acolytec3 (approved):

packages/block/README.md
```diff
@@ -60,11 +60,12 @@ Properties of a `Block` or `BlockHeader` object are frozen with `Object.freeze()
 API Usage Example:
 
 ```ts
+// ./examples/1559.ts#L43-L47
+
```
holgerd77, line 64: @scorbajio Thanks for the different doc updates! 🙂 

One note on the per-line example references: on the one hand I find this elegant. On the other hand I fear a bit that we wil rather overlook this mid-term and it will relatively likely happen that we will do some example updates (and e.g. add a few lines to some examples), globally rebuild the docs example integration and then overlook that the lines are not matching anymore.

Maybe let's therefore be a bit careful with this.

PR #3622, opened 2024-08-29
Trie static constructors doc updates

scorbajio:
This change update the trie package README to use the updated standalone functions instead of old static constructors that were refactored in #3515.


acolytec3 (approved):
LGTM

Issue #3609, opened 2024-08-21
Noble BLS fails BLS tests on Prague, while MCL does not


jochem-brouwer:
When working on t8n, we failed to build fixtures for BLS (EIP-2537). This is because our EVM defaults to NobleBLS, but, in the blockchain/state test runner we:

```typescript
  let bls: EVMBLSInterface
  if (argv.bls !== undefined && argv.bls.toLowerCase() === 'noble') {
    console.log('BLS library used: Noble (JavaScript)')
    bls = new NobleBLS()
  } else {
    await mcl.init(mcl.BLS12_381)
    bls = new MCLBLS(mcl)
    console.log('BLS library used: MCL (WASM)')
  }
```

default to MCL. Using MCL, the tests are generated.

The fixtures below have been generated by geth (I have done this locally) and are included as zip in this issue. To run tests, unzip the fixtures inside `ethereum-tests` and now run blockchain/state tests: (from vm package)

- `npm run test:blockchain -- --fork=Prague --dir=../fixtures/blockchain_tests` (pass)
- `npm run test:blockchain -- --fork=Prague --dir=../fixtures/blockchain_tests --bls=noble` (fail)
- `npm run test:state -- --fork=Prague --dir=../fixtures/state_tests` (pass)
- `npm run test:state -- --fork=Prague --dir=../fixtures/state_tests --bls=noble` (fail)

TODOs:

- [ ] Figure out why NobleBLS fails
- [ ] Add a specific state/blockchain test which will also test the other dependency (so not the default) to the CI action. This should only run the 2537 tests, no need to run all other tests
- [ ] Also, the choice to use MCL is likely made in the CI for the performance benefit, but we should likely switch to the default here in case no BLS is passed?

[fixtures.zip](https://github.com/user-attachments/files/16695309/fixtures.zip)

holgerd77:
No, we should never make MCL default again, and rather work our heads off to get Noble passing. 😂 A WASM free VM is one of the core value propositions of this release round (security, especially in the light of 7702,).

holgerd77:
(in doubt ask @paulmillr )

paulmillr:
I will dive into that.

To ease the ramp up, could you please share a specific code sample & vector that fails?

jochem-brouwer:
[...]

PR #3623, opened 2024-09-02
Fix BLS usage

paulmillr:
Closes gh-3609

### DIFF
--- a/packages/evm/src/precompiles/bls12_381/noble.ts
+++ b/packages/evm/src/precompiles/bls12_381/noble.ts
@@ -35,7 +35,7 @@ type Fp2 = {
  * @param input Input Uint8Array. Should be 128 bytes
  * @returns Noble G1 point
  */
-function BLS12_381_ToG1Point(input: Uint8Array) {
+function BLS12_381_ToG1Point(input: Uint8Array, verifyOrder = true) {
   if (equalsBytes(input, BLS_G1_INFINITY_POINT_BYTES)) {
     return bls12_381.G1.ProjectivePoint.ZERO
   }
@@ -47,6 +47,13 @@ function BLS12_381_ToG1Point(input: Uint8Array) {
     x,
     y,
   })
+
+  try {
+    G1.assertValidity()
+  } catch (e) {
+    if (verifyOrder || (e as Error).message !== 'bad point: not in prime-order subgroup')
+      throw new EvmError(ERROR.BLS_12_381_POINT_NOT_ON_CURVE)
+  }
 
   return G1
 }
@@ -66,7 +73,7 @@ function BLS12_381_FromG1Point(input: AffinePoint<bigint>): Uint8Array {
  * @param input Input Uint8Array. Should be 256 bytes
  * @returns Noble G2 point
  */
-function BLS12_381_ToG2Point(input: Uint8Array): any {
+function BLS12_381_ToG2Point(input: Uint8Array, verifyOrder = true): any {
   // TODO: remove any type, temporary fix due to conflicting @noble/curves versions
   if (equalsBytes(input, BLS_G2_INFINITY_POINT_BYTES)) {
     return bls12_381.G2.ProjectivePoint.ZERO
@@ -84,6 +91,13 @@ function BLS12_381_ToG2Point(input: Uint8Array): any {
     x: Fp2X,
     y: Fp2Y,
   })
+
+  try {
+    pG2.assertValidity()
+  } catch (e) {
+    if (verifyOrder || (e as Error).message !== 'bad point: not in prime-order subgroup')
+      throw new EvmError(ERROR.BLS_12_381_POINT_NOT_ON_CURVE)
+  }
 
   return pG2
 }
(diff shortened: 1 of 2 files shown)


jochem-brouwer:
Wow great, thanks a lot @paulmillr!! I will push a few commits to add Noble tests to CI and to fix the linter.

jochem-brouwer:
Yes, will do, actually working on this PR ATM, also flipping out our test runner to use Noble by default, not MCL.

holgerd77:
> Yes, will do, actually working on this PR ATM, also flipping out our test runner to use Noble by default, not MCL.

Wondering a bit: do we really want to fully switch to Noble for the test runner? We are also not doing this for the other pure JS stuff for performance reasons to not slow down the mostly unrelated everyday-PR-CI runs. 🤔 

The alternative would be to separate this and maybe run along the nightly VM tests. I think I would have a somewhat clear preference for that. That would give this a good balance and still ensures that we do not re-introduce new errors here.

jochem-brouwer:
I'm not sure here. So currently (if we use MCL by default on our test runner, but Noble by default in our EVM), then we are thus showing that if one uses the alternative (MCL) the tests pass. However we want to ensure that if people use our EVM with default settings, it should also pass the state/blockchain tests.

I would rather prefer we test our default implementation (we can add a CI job either on PR or on nightly to test the alternatives, so MCL in this case). 

I just tested, the specific state tests here run 6.484 seconds using MCL, and 7.993 using Noble (using `multiplyUnsafe` already). So this is a slowdown for the test runner of ~1.6 seconds.

paulmillr:
Another optimization has been added.

holgerd77:
> I'm not sure here. So currently (if we use MCL by default on our test runner, but Noble by default in our EVM), then we are thus showing that if one uses the alternative (MCL) the tests pass. However we want to ensure that if people use our EVM with default settings, it should also pass the state/blockchain tests.
> 
> I would rather prefer we test our default implementation (we can add a CI job either on PR or on nightly to test the alternatives, so MCL in this case).
> 
> I just tested, the specific state tests here run 6.484 seconds using MCL, and 7.993 using Noble (using `multiplyUnsafe` already). So this is a slowdown for the test runner of ~1.6 seconds.

Yes, argumentation is also totally valid and given the numbers you are citing it is totally out of question to even consider (running MCL as default), agree!

We should also run MCL in some setup (nightly likely). This is not *totally* optional in the sense that we also provide code for this and so it's more like an option to choose from, where the functionality is already baked in. So we should also make sure to have this tested to some minimal extend.

Ah, a bit live-updated thought though: it generally would make sense to *generalize* here, also for the other optional WASM stuff, and make one test round (normal CI e.g.) all with JS and one test round (nightly e.g.) all with WASM. *Then* I am not sure though anymore how performance is affected and if this is still realistic this way around.

Anyhow: let's keep this for now, we can discuss on team call how we generally want to proceed here!

packages/evm/src/precompiles/bls12_381/noble.ts
```diff
@@ -305,30 +321,21 @@ export class NobleBLS implements EVMBLSInterface {
...
-      const G2 = pair[1] as ProjPointType<Fp2>
-
-      if (index === 0) {
-        GT = bls12_381.pairing(G1, G2)
-      } else {
-        GT = bls12_381.fields.Fp12.mul(GT!, bls12_381.pairing(G1, G2))
-      }
-    }
+    // @ts-ignore
+    const FP12 = bls12_381.pairingBatch(pairs, true)
```
holgerd77, line 337: @paulmillr 🙏

holgerd77 (approved):
LGTM, thanks for all the fixes and optimizations! 🙏 🙂 ❤️

paulmillr:
@jochem-brouwer i'm curious about time of execution now after `pairingBatch`. How can one calculate this time?

> I just tested, the specific state tests here run 6.484 seconds using MCL, and 7.993 using Noble (using multiplyUnsafe already). So this is a slowdown for the test runner of ~1.6 seconds.

jochem-brouwer:
Yes, sure! The idea is to run the state tests for both Noble and MCL. We could also run the blockchain tests, but these give blockchain overhead which we don't want.

To do so, first download the test: https://github.com/ethereumjs/ethereumjs-monorepo/issues/3609#issue-2478495412 (download the fixtures, and extract those in `./packages/ethereum-tests`.

Now run the state tests using Noble:

`npm run test:state -- --fork=Prague --dir=../fixtures/state_tests`

You can check it uses Noble, test runner outputs first:

```
BLS library used: Noble (JavaScript)
BN254 (alt_BN128) library used: Noble (JavaScript)
```

At the end of the tests, the total time is shown, for me: `Total (including setup): 7.513s`

Now run it using MCL:

`npm run test:state -- --fork=Prague --dir=../fixtures/state_tests --bls=mcl`

Output: `Total (including setup): 6.503s`

Note, this runs all tests in those fixtures, so these might include test not calling into `pairingBatch`.

paulmillr:
Results on my M2:

```
Noble: Total (including setup): 2.879s
MCL: Total (including setup): 2.728s
```

PR #3628, opened 2024-09-02
Add `statefulVerkleStateManager`

acolytec3:
Implements a stateful version of the `statemanager` interface that uses a verkle trie as a backend.


acolytec3:
Note, this still has extraneous commits that already exist on master in it (from the kaustinen7 readiness PR).  Will purge those out and clean up the commit trie at some point here soon.

packages/statemanager/src/statefulVerkleStateManager.ts
```diff
@@ -0,0 +1,239 @@
...
+        // Delete account
+      }
+    } else {
+      if (account !== undefined) {
+        this._caches?.account?.put(address, account, true)
+      } else {
+        this._caches?.account?.del(address)
+      }
+    }
+  }
```
holgerd77, line 155: Don't want to get yourself too much on the nerve with this refactoring/DRY stuff (but maybe a bit 😂), but maybe you can put such code parts like the `putAccount()` directly in a `src/helpers/verkle.ts` file or similar with an adjusted function signature `putAccount(self/this: SM, address: Address, account?: Account): Promise<void>` taking the code from the stateless verkle SM directly over so that these kind of code parts just never reach this file?

Then we directly have a unified version and do not need to post-DRY this and this should not be significantly ore work (just run the statless tests from `statelessVerkleStateManager.spec.ts` along)!

acolytec3: I guess I'm not 100% sure what you mean since the implementations are not the same.  The `stateless` SM puts things in an in memory `map` via direct `this._state[verkleKey] = serializedBytes` and the `stateful` implementation uses `await this.trie.put(stem, suffixes, values)` to "put" data.  The thing I could do differently is to have `encodeVerkleLeafBasicData` accept an `account` rather than the various fields and that would de-dup a few lines of code.

holgerd77: Ok, yeah, nevermind, you are right, haven't looked/though closely enough. I think I am generally still latently unsatisfied that we always copy over 2/3 of code which is redundant/the same.

Maybe we can also tackle this in a more fine-grained way. We could e.g. centralize the generic DEBUG strings and re-use, things like that (for sure not in this PR).

So don't let me distract you here! 🙂 I will nevertheless not give up on the topic, maybe open some more low-reaching small issues on this after this PR got merged.

packages/statemanager/src/statefulVerkleStateManager.ts
```diff
@@ -0,0 +1,344 @@
...
+      chunkStems[0],
+      chunkSuffixes.slice(0, codeChunks.length <= 128 ? codeChunks.length : 128),
+      codeChunks.slice(0, codeChunks.length <= 128 ? codeChunks.length : 128),
+    )
+    // Put additional chunks under additional stems as applicable
+    for (let stem = 1; stem < chunkStems.length; stem++) {
+      await this._trie.put(
+        chunkStems[stem],
+        chunkSuffixes.slice(
+          128 + (256 * stem - 1),
```
gabrocheleau, line 215: I think we should use the constants for these values for 2 reasons:
- If the spec change, we can easily just update the constant.
- It makes the meaning of these numbers obvious (e.g. CODE_OFFSET, VERKLE_NODE_WIDTH, etc.)

packages/statemanager/src/statefulVerkleStateManager.ts
```diff
@@ -0,0 +1,344 @@
...
+          codeChunks.length <= 128 + 256 * stem ? codeChunks.length : 128 + 256 * stem,
+        ),
+        codeChunks.slice(
+          128 + (256 * stem - 1),
+          codeChunks.length <= 128 + 256 * stem ? codeChunks.length : 128 + 256 * stem,
+        ),
```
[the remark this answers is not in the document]
acolytec3: Where are you seeing this?  In the code below here, it's not `stem - 1` but `(VERKLE_NODE_WIDTH * stem) - 1` since these arrays are zero-indexed.  I'm not referencing a specific chunkId but a section of the overall array of chunks corresponding to the stem we're putting them under.

packages/evm/test/verkle.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+    const account = createAccount({ nonce: 3n, balance: 0xffffffffn })
+    await sm.putAccount(address, account)
+    const evm = await createEVM({ common, stateManager: sm })
+    const code = hexToBytes('0x6001600255')
+    const accessWitness = new AccessWitness({ verkleCrypto })
+    const res = await evm.runCall({
+      code,
+      caller: address,
+      accessWitness,
+      to: createAddressFromString(address.toString()),
```
gabrocheleau, line 37:
```suggestion
      to: address,
```

packages/evm/test/verkle.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+    })
+    assert.deepEqual(res.execResult.returnValue, new Uint8Array())
+    console.log(res.execResult.runState?.interpreter._env.accessWitness)
```
gabrocheleau, line 40:
```suggestion
```

packages/evm/test/verkle.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+    const common = new Common({ chain: Mainnet, eips: [6800], hardfork: Hardfork.Cancun })
+    const trie = await createVerkleTree()
+    const sm = new StatefulVerkleStateManager({ trie, verkleCrypto })
+    const address = createAddressFromString('0x9e5ef720fa2cdfa5291eb7e711cfd2e62196f4b3')
+    const account = createAccount({ nonce: 3n, balance: 0xffffffffn })
+    await sm.putAccount(address, account)
+    const evm = await createEVM({ common, stateManager: sm })
+    const code = hexToBytes('0x6001600255')
+    const accessWitness = new AccessWitness({ verkleCrypto })
+    const res = await evm.runCall({
```
gabrocheleau, line 33: What is happening in this test case is not clear to me from reading the code. Could we add some comments to clarify what is going on with `runCall` (e.g. why are we expecting from running this code, why execResult is an empty Uint8Array, why storage at index 2 is now `1n`)

packages/util/src/verkle.ts
```diff
@@ -293,18 +319,71 @@ export function decodeVerkleLeafBasicData(encodedBasicData: Uint8Array): VerkleL
...
+    if (x < 640) {
+      // Suffixes for chunks 384 - 639 start at 0 and go to 255
+      chunkSuffixes[x] = x - VERKLE_CODE_OFFSET - VERKLE_NODE_WIDTH
+      continue
+    }
+    if (x > 639) {
+      // Suffixes for chunks 640 - 793 start at 0 and go to 153
+      chunkSuffixes[x] = x - VERKLE_CODE_OFFSET - 2 * VERKLE_NODE_WIDTH
+      continue
+    }
```
gabrocheleau, line 357: Is the idea that the number of chunks would never exceed 793? If so, wdyt if we:
- Add a comment above the function making that explicit (
- Add some validation that throws if numChunks > 793
?

packages/util/src/verkle.ts
```diff
@@ -293,18 +319,71 @@ export function decodeVerkleLeafBasicData(encodedBasicData: Uint8Array): VerkleL
...
+  }
+  return chunkSuffixes
+}
+
+export const generateCodeStems = async (
+  numChunks: number,
+  address: Address,
+  verkleCrypto: VerkleCrypto,
+) => {
+  // The maximum number of chunks is 793 (maxCodeSize - 24576) / (bytes per chunk 31) + (round up - 1)
```
gabrocheleau, line 367: Why `-24576`?

packages/verkle/src/verkleTree.ts
```diff
@@ -290,11 +305,9 @@ export class VerkleTree {
...
   }
 
-  async del(key: Uint8Array): Promise<void> {
-    const stem = key.slice(0, 31)
-    const suffix = key[key.length - 1]
-    this.DEBUG && this.debug(`Stem: ${bytesToHex(stem)}; Suffix: ${suffix}`, ['DEL'])
-    await this.put(key, createDeletedLeafValue())
+  async del(stem: Uint8Array, suffixes: number[]): Promise<void> {
+    this.DEBUG && this.debug(`Stem: ${bytesToHex(stem)}; Suffix(es): ${suffixes}`, ['DEL'])
+    await this.put(stem, suffixes, new Array(suffixes.length).fill(createDeletedLeafValue()))
```
gabrocheleau, line 310: I think it would be more efficient to first declare `deletedLeafValue = createDeletedLeafValue()`, and then fill with that declared value `fill(deletedLeafValue)` rather than create each time. Unless we have a reason to want different objects instantiations?

packages/statemanager/test/statefulVerkleStateManager.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+    verkleCrypto = await loadVerkleCrypto()
+  })
+  it('should put/get/delete an account (with no storage/code from the trie', async () => {
+    const trie = await createVerkleTree()
+    const sm = new StatefulVerkleStateManager({ trie, verkleCrypto })
+    const address = createAddressFromString('0x9e5ef720fa2cdfa5291eb7e711cfd2e62196f4b3')
+    const account = createAccount({ nonce: 3n, balance: 0xfffn })
+    await sm.putAccount(address, account)
+    const retrievedAccount = await sm.getAccount(address)
+    assert.equal(retrievedAccount?.balance, account.balance)
```
gabrocheleau, line 33: Let's check the nonce too!

packages/statemanager/test/statefulVerkleStateManager.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+      '0x7faaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',
+    ) // PUSH32 aa.....
+    await sm.putCode(address, bigByteCode)
+    const retrievedBigByteCode = await sm.getCode(address)
+    assert.deepEqual(bigByteCode, retrievedBigByteCode)
+    const reallyBigByteCode = hexToBytes(
+      (await import('./testdata/biggestContractEver.json')).default.bytecode as PrefixedHexString,
+    )
+    // Biggest mainnet contract - 0x10C621008B210C3A5d0385e458B48af05BF4Ec88 (supposedly anyway)
+    await sm.putCode(address, reallyBigByteCode)
```
gabrocheleau, line 56: We could also verify that the account's codeSize and codehash fields have been updated

packages/statemanager/test/statefulVerkleStateManager.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+      reallyBigByteCode.length,
+    )
+  })
+  it('should put and get storage', async () => {
+    const trie = await createVerkleTree()
+    const sm = new StatefulVerkleStateManager({ trie, verkleCrypto })
+    const address = createAddressFromString('0x9e5ef720fa2cdfa5291eb7e711cfd2e62196f4b3')
+    await sm.putAccount(address, new Account(0n, 1n))
+    await sm.putStorage(address, setLengthLeft(bigIntToBytes(0n), 32), hexToBytes('0x1'))
+    const retrievedValue = await sm.getStorage(address, setLengthLeft(bigIntToBytes(0n), 32))
```
gabrocheleau, line 70: We could declare some variables instead of repeatedly inlining them:
- `storageKey = setLengthLeft(bigIntToBytes(0n,32))`
- `storageValue = hexToBytes('0x1')`

packages/util/test/verkle.spec.ts
```diff
@@ -81,3 +85,45 @@ describe('should generate valid tree keys', () => {
...
+
+describe('should encode and decode basic data values', () => {
+  const account = new Account(2n, 123n)
+  it('should encode basicData to 32 bytes', () => {
+    const basicDataBytes = encodeVerkleLeafBasicData(account)
+    assert.equal(basicDataBytes.length, 32)
+    assert.equal(
+      basicDataBytes.slice(4, 12)[7],
+      2,
+      'confirm that last byte of nonce slice is equal to nonce (i.e. coded as bigEndian',
```
gabrocheleau, line 97:
```suggestion
      'confirm that last byte of nonce slice is equal to nonce (i.e. coded as bigEndian)',
```

packages/util/test/verkle.spec.ts
```diff
@@ -81,3 +85,45 @@ describe('should generate valid tree keys', () => {
...
+  it('should encode basicData to 32 bytes', () => {
+    const basicDataBytes = encodeVerkleLeafBasicData(account)
+    assert.equal(basicDataBytes.length, 32)
+    assert.equal(
+      basicDataBytes.slice(4, 12)[7],
+      2,
+      'confirm that last byte of nonce slice is equal to nonce (i.e. coded as bigEndian)',
+    )
+    const decodedData = decodeVerkleLeafBasicData(basicDataBytes)
+    assert.equal(decodedData.balance, 123n)
```
gabrocheleau, line 100: Could also check the nonce

packages/util/test/verkle.spec.ts
```diff
@@ -81,3 +85,45 @@ describe('should generate valid tree keys', () => {
...
+  it('should chunkify code with overflow PUSHDATA', () => {
+    const byteCode = hexToBytes(
+      '0x7faaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',
+    ) // PUSH32 aa.....
+    const chunkifiedCode = chunkifyCode(byteCode)
+    assert.equal(chunkifiedCode.length, 2, 'bytecode of length 33 should be in 2 chunks')
+    assert.equal(
+      chunkifiedCode[1][0],
+      2,
+      'second chunk should have a 2 in first position (for 2 bytes of PUSHDATA overflow from previous chunk',
```
gabrocheleau, line 114:
```suggestion
      'second chunk should have a 2 in first position (for 2 bytes of PUSHDATA overflow from previous chunk)',
```

packages/util/test/verkle.spec.ts
```diff
@@ -81,3 +85,45 @@ describe('should generate valid tree keys', () => {
...
+  it('should chunkify code without overflow PUSHDATA', () => {
+    const byteCode = hexToBytes(
+      '0x70aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',
+    ) // PUSH17 aa.....
+    const chunkifiedCode = chunkifyCode(byteCode)
+    assert.equal(chunkifiedCode.length, 2, 'bytecode of length 33 should be in 2 chunks')
+    assert.equal(
+      chunkifiedCode[1][0],
+      0,
+      'second chunk should have a 0 in first position (for 0 bytes of PUSHDATA overflow from previous chunk',
```
gabrocheleau, line 126:
```suggestion
      'second chunk should have a 0 in first position (for 0 bytes of PUSHDATA overflow from previous chunk)',
```

packages/statemanager/src/statefulVerkleStateManager.ts
```diff
@@ -0,0 +1,467 @@
...
+      account = createPartialAccount({
+        version: basicData.version,
+        balance: basicData.balance,
+        nonce: basicData.nonce,
+        codeHash:
+          accountValues[1] === undefined
+            ? KECCAK256_NULL // codeHash is undefined in trie (i.e. not touched)
+            : equalsBytes(accountValues[1], new Uint8Array(32))
+              ? KECCAK256_NULL // codeHash is deleted in trie (i.e. overwritten with zeroes)
+              : accountValues[1],
```
gabrocheleau, line 122: I think this is equivalent and imo easier to parse: 
```suggestion
        codeHash:
          accountValues[1] === undefined || equalsBytes(accountValues[1], new Uint8Array(32)) ? KECCAK256_NULL : accountValues[1]
```

packages/statemanager/src/statefulVerkleStateManager.ts
```diff
@@ -0,0 +1,467 @@
...
+        codeSize: basicData.codeSize,
+        storageRoot: KECCAK256_NULL, // TODO: Add storage stuff
+      })
+    }
+    // check if the account didn't exist if any of the basic keys are undefined
+    else if (accountValues[0] === undefined || accountValues[1] === undefined) {
```
gabrocheleau, line 128: I think we already know that `accountValues[0] === undefined` right?

packages/statemanager/src/statefulVerkleStateManager.ts
```diff
@@ -0,0 +1,467 @@
...
+      if (account !== undefined) {
+        const stem = getVerkleStem(this.verkleCrypto, address, 0)
+        const basicDataBytes = encodeVerkleLeafBasicData(account)
+        await this._trie.put(
+          stem,
+          [VerkleLeafType.BasicData, VerkleLeafType.CodeHash],
+          [basicDataBytes, account.codeHash],
+        )
+      } else {
+        // Delete account
```
gabrocheleau, line 161: Is this a TODO?

packages/evm/test/verkle.spec.ts
acolytec3: Will do.  All the bytecode does is push the values 1 and 2 onto the stack and then call SSTORE (which pops two values from the stack and stores the second value popped in the storage slot indicated by the first value popped).  In this case, it's "store the value 1 in slot 2".  Therefore, there's no return value since we don't RETURN anything in the bytecode.  It's literally just putting a value in storage and this was a simple way to verify that bytecode could be executed and state updated correctly using the SFVKSM.

packages/statemanager/src/statefulVerkleStateManager.ts
acolytec3: AH, yes, good catch.  Will fix this tomorrow

packages/statemanager/src/statefulVerkleStateManager.ts
acolytec3: Good call.  Thanks!

packages/statemanager/src/statefulVerkleStateManager.ts
```diff
@@ -0,0 +1,467 @@
...
+          accountValues[1] === undefined
+            ? KECCAK256_NULL // codeHash is undefined in trie (i.e. not touched)
+            : equalsBytes(accountValues[1], new Uint8Array(32))
+              ? KECCAK256_NULL // codeHash is deleted in trie (i.e. overwritten with zeroes)
+              : accountValues[1],
+        codeSize: basicData.codeSize,
+        storageRoot: KECCAK256_NULL, // TODO: Add storage stuff
+      })
+    }
+    // check if the account didn't exist if any of the basic keys are undefined
```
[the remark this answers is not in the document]
acolytec3: Yep, copied and didn't cleann up.  Have clarified now

packages/statemanager/src/statefulVerkleStateManager.ts
acolytec3: Good point.

packages/util/src/verkle.ts
acolytec3: That's the maximum allowed code size [see here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/8b7d4b19abf1834d15759a559adeffc85c06a57c/packages/evm/src/params.ts#L132)

packages/util/src/verkle.ts
acolytec3: That's the idea though now that I look at my math, something seems off, since a maximum code size of 24576 should guarantee we never have more than 768 chunks (24576 / 32 = 768).
Let me revalidate the math here and see where I got 793 from and will clean this up.

packages/verkle/src/verkleTree.ts
acolytec3: Well, if you declare it once and then reference it across all the deleted suffixes, I'm pretty sure you just get that references to one Uint8Array of 32 0s.  I'm assuming that is problematic.

packages/util/src/verkle.ts
acolytec3: Okay, I remember why the maxChunks is 793 now.  It's this `// The maximum number of chunks is 793 (maxCodeSize - 24576) / (bytes per chunk 31) + (round up - 1)`
So, if code size is 24756 bytes, chunk size is 31 bytes, then 24756 / 31 = 792 and change, so we round up to 793 to get the max number of chunks.  Will define this as a constant.

packages/evm/test/verkle.spec.ts
gabrocheleau: Nice, thanks for the explainer

packages/util/src/verkle.ts
```diff
@@ -322,48 +325,51 @@ export function decodeVerkleLeafBasicData(encodedBasicData: Uint8Array): VerkleL
...
-      // Suffixes for chunks 384 - 639 start at 0 and go to 255
-      chunkSuffixes[x] = x - VERKLE_CODE_OFFSET - VERKLE_NODE_WIDTH
-      continue
-    }
-    if (x > 639) {
-      // Suffixes for chunks 640 - 793 start at 0 and go to 153
-      chunkSuffixes[x] = x - VERKLE_CODE_OFFSET - 2 * VERKLE_NODE_WIDTH
-      continue
+  if (numChunks === 0) return []
+  const chunkSuffixes: number[] = new Array<number>(numChunks)
```
gabrocheleau, line 345:
```suggestion
  const chunkSuffixes = new Array<number>(numChunks)
```
This type is probably unnecessary? Should be inferred from new Array<number>(numChunks)

packages/util/src/verkle.ts
```diff
@@ -322,48 +325,51 @@ export function decodeVerkleLeafBasicData(encodedBasicData: Uint8Array): VerkleL
...
+  const firstChunksSet = numChunks > VERKLE_CODE_OFFSET ? VERKLE_CODE_OFFSET : numChunks
```
gabrocheleau, line 346:
```suggestion
  const firstChunksSet = Math.min(numChunks, VERKLE_CODE_OFFSET)
  ```
We could do this if you think it's easier to parse mentally

acolytec3: Much better.  Have applied

gabrocheleau (approved):

holgerd77:
🙂 🎉

PR #3633, opened 2024-09-05
Block: Fix Uncle Block Behind During HF Transition Test Runner Case

holgerd77:
This fixes the blockchain test runner failure on `BerlinToLondonAt5` from #3593 having a case where uncle block numbers are behind (which is allowed to some extend) on a block during a HF transition which causes an unintended `calcNextBaseFee()` call.

By activating the `setHardfork` option on block creation the option also propagates to the uncles and so an eventual deviating HF is set for the uncles and not simply taken from the general header Common.


holgerd77:
Some extended blockchain tests still failing, for reference:

```shell
npm run test:blockchain -- --file='lastblockhashException.json' --fork=Istanbul
npm run test:blockchain -- --file='wrongRLPGenesis.json' --fork=Istanbul
```

(cross-tested both also on `master` and tests are failing there as well, so not newly introduced)

holgerd77:
Ok, the RLP genesis thing fails on "correct genesis RLP" (maybe that's already a good result at the wrong place, looking at the name?):

grafik

Byte 41 in the Uint8Array is the first - and only ! - one differing (206 vs 204).

holgerd77:
Ok, another iteration and back to this `wrongRLPGenesis.json` test. So, this fails on `t.deepEquals(genesisBlock.serialize(), rlp, 'correct genesis RLP')` in our blockchain test runner, which I guess is somewhat of an extra check we have added as a sanity check.

When I comment out this line, the rest of the test ("correct last header block") then passes again, which is the "core" test. Seems this test is just somewhat colliding with our test/sanity check setup, but it's for sure not worth to change anything here (or hack in a per-one-test exception) just for this one test.

Not sure how to proceed, @jochem-brouwer, other, any idea/suggestion?

As a side note, I cannot really explain why this test case is now failing and seem to have made no problems in the past. I am not sure if it is really worth the effort to expore though (respectively: I would vote for rather not, this does not seem to bring any new insight or make the libraries any better if we would hunt this down).

holgerd77:
This is the filler for this lastblockhash test:
https://github.com/ethereum/tests/blob/develop/src/BlockchainTestsFiller/InvalidBlocks/bcExpectSection/lastblockhashExceptionCopier.json

Maybe that's a bit a similar case? This tests the exception that the block hash does *not* match per test, and our test setup is just not made up for this?

Should we just add both tests to the exceptions and not run?

jochem-brouwer:
`npm run test:blockchain -- --file='lastblockhashException.json' --fork=Istanbul`

This test points to: 

`ethereumjs-monorepo/packages/ethereum-tests/LegacyTests/Cancun/BlockchainTests/InvalidBlocks/bcExpectSection/wrongRLPGenesis.json`

Which:

```
    "wrongRLPGenesis_Istanbul" : {
        "_info" : {
            "comment" : "This is a unit test for retesteth logic",
```

So we can ignore this one.            

Same for `lastblockhash`, points to `ethereumjs-monorepo/packages/ethereum-tests/LegacyTests/Cancun/BlockchainTests/InvalidBlocks/bcExpectSection/lastblockhashException.json`, which has the same "unit test" comment on it.

jochem-brouwer:
Hi @holgerd77, I have added these tests which fail on our side (due to internal tests of ethereum/tests) to the skipped tests.

I also fixed `npm run test:blockchain -- --fork=HomesteadToDaoAt5` (pushed here).

jochem-brouwer:
Ah wait a min, we likely have to update ethereum/tests to move those internal tests out., see https://github.com/ethereum/tests/pull/1392/files

jochem-brouwer:
Have updated ethereum/tests to v14.1 (from v14.0), this removes the internal tests from the monorepo.

Note: if you check this locally, then `git submodule update --recursive` otherwise LegacyTests will point to an older version of the submodule.

holgerd77 (commented):
Cool, thanks for looking into this! 🤩 

(will admin merge since implicitly reviewed by fixing all the remaining issues by a second team member)

PR #3638, opened 2024-09-05
monorepo: json and rpc casing

### LINKED ISSUE #3632: Monorepo: Json/JSON and Rpc/RPC Namining inconsistencies
Is it fun? Who knows. 😂 

But one of the broadest inconsistencies in naming throughout our libraries is still the more or less randomly chosen use of capitalized vs camel-case JSON/Json and RPC/Rpc.

I would suggest that we eliminate this once and for all and go for one version, I guess (?, cc @jochem-brouwer ) we can then add the non-allowed versions (e.g. `Rpc`, `Json`) even to the spell checker (does the spell checker takes capitalization into account?), which would allow us to fully get rid of this.

(side question: another such pair anyone can think of?)

Current stand (monorepo-wide search is):
- JSON: 749 usages, Json: 285 usages
- RPC: 417 usages, Rpc: 331 usages

I would have a preference to take the capitalized versions here, since both are abbreviations with every letter standing for a separate word ("JavaScript Object Notation", "Remote Procedure Call"), so it's somewhat already camel-case condensed 🙂, and so I think keeping everything capitalized is somewhat more logical.

gabrocheleau:
Resolves #3632

### DIFF
--- a/packages/client/src/rpc/helpers.ts
+++ b/packages/client/src/rpc/helpers.ts
@@ -4,9 +4,9 @@ import { INTERNAL_ERROR, INVALID_BLOCK, INVALID_PARAMS } from './error-code.js'
 
 import type { Chain } from '../blockchain/index.js'
 import type { Block } from '@ethereumjs/block'
-import type { JsonRpcTx, TypedTransaction } from '@ethereumjs/tx'
+import type { JSONRPCTx, TypedTransaction } from '@ethereumjs/tx'
 
-type RpcError = {
+type RPCError = {
   code: number
   message: string
   trace?: string
@@ -19,7 +19,7 @@ export function callWithStackTrace(handler: Function, debug: boolean) {
       const res = await handler(...args)
       return res
     } catch (error: any) {
-      const e: RpcError = {
+      const e: RPCError = {
         code: error.code ?? INTERNAL_ERROR,
         message: error.message,
         data: error.data,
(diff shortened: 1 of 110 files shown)


holgerd77 (approved):
LGTM, great, thanks! 🤩 

Can you please also submit a follow-up PR, where we add this to the spell checker? That seems to be the single most important thing to me now, so that we do not risk of loosing consistency here again (which otherwise *would* happen)!

packages/client/src/rpc/helpers.ts
```diff
@@ -36,7 +36,7 @@ export function callWithStackTrace(handler: Function, debug: boolean) {
 /**
  * Returns tx formatted to the standard JSON-RPC fields
  */
-export const jsonRpcTx = (tx: TypedTransaction, block?: Block, txIndex?: number): JsonRpcTx => {
+export const jsonRPCTx = (tx: TypedTransaction, block?: Block, txIndex?: number): JSONRPCTx => {
```
holgerd77, line 39: Can we maybe rename this one as well so that we do not need to keep this somewhat confusing lowercasing due to having this at the beginning, maybe just simply by renaming to something like `toJSONRPCTx()` or so?

holgerd77:
(so if `json` provides a problem (due to file endings e.g.) we should at least disallow `Json` and `Rpc`)

packages/client/src/rpc/helpers.ts
gabrocheleau: Yes, will do in a follow up

gabrocheleau:
> (so if `json` provides a problem (due to file endings e.g.) we should at least disallow `Json` and `Rpc`)

I've disallowed Json and Rpc (will push shortly in a separate PR). The lower case versions (`json` and `rpc`) end up being quite messy, because of filenames, cli options, etc. Lowercase `json` is definitely a no-go in my view because of the .json file extension. Lowercase `rpc` could be done, but I'm not sure if we want to update all cli arguments, folder names, and other places where this is used.

PR #3640, opened 2024-09-06
wallet: simplify test case

gabrocheleau:


scorbajio (approved):
ACK

holgerd77:
Like that we slowly get some cautious active work on wallet! 👍🙂

PR #3647, opened 2024-09-09
Update block doc examples

scorbajio:
This change removes a standalone code example in the block package README and embeds it from a file in the examples directory, and also refactors some of the block examples to make sure they are working.


acolytec3 (approved):
LGTM

packages/block/examples/6110Requests.ts
```diff
@@ -21,7 +21,7 @@ const main = async () => {
     signature: randomBytes(96),
     index: bytesToBigInt(randomBytes(8)),
   }
-  const request = DepositRequest.fromRequestData(depositRequestData) as CLRequest<CLRequestType>
+  const request = createDepositRequest(depositRequestData) as CLRequest<CLRequestType>
```
holgerd77, line 24: We should investigate why this was not caught by the CI (I very much think we have a GH action testing the example running?)! 🤔

PR #3649, opened 2024-09-10
Monorepo: Fix and Improve Browser Examples

holgerd77:
Our browser examples embedded in the `browser.html` files are the last bits of both redundant and untested TypeScript code in our monorepo (and consequently we also forgot to update most of these examples 😛).

Today I discovered that we can directly reference one of the other example TypeScript files in the `examples` folder. 🙂 This will allow us to have the browser example seemlessly/naturally updated along in the future and we only have to care about the core examples in general.

https://github.com/user-attachments/assets/ed1e5ddb-b1cc-4616-872c-40d599c9d134

I tested all examples in the both the Firefox and Chrome browser debug extensions and things work like a charm! 🤩


packages/statemanager/src/merkleStateManager.ts
```diff
@@ -47,6 +48,16 @@ import type {
...
 
+/**
+ * Prefix to distinguish between a contract deployed with code `0x80`
+ * and `RLP([])` (also having the value `0x80`).
+ *
+ * Otherwise the creation of the code hash for the `0x80` contract
+ * will be the same as the hash of the empty trie which leads to
+ * misbehaviour in the underlying trie library.
+ */
+export const CODEHASH_PREFIX = utf8ToBytes('c')
```
holgerd77, line 59: This just fixed the state manager browser examples and I thought it was worth the change (no side effects)

PR #3650, opened 2024-09-10
Various Verkle Fixes

acolytec3:
Fixes several bugs in our verkle trie implementation and adds a couple of tests from the [python implementation](https://github.com/crate-crypto/verkle-trie-ref/blob/483f40c737f27bc8f059870f862cf6c244159cd4/verkle/verkle_test.py#L63) to verify that basic interop is now possible.

Notable fixes
- Correctly implement little-endian value construction in leaf nodes 
- Update leaf marker to little endian representation 

Note: I'm not sure why most of these commits are showing up.  Most of these are from a previous PR and the actual "files changed" section shows the true set of changes.


holgerd77 (approved):
LGTM

packages/verkle/src/node/util.ts
```diff
@@ -42,18 +42,15 @@ export function isInternalNode(node: VerkleNode): node is InternalNode {
...
- *
- * @returns a 32 byte array of zeroes with the 129th bit set to 1
+ * @returns a 32 byte array of zeroes (optionally with 129th bit set to 1)
  */
-export const createDeletedLeafValue = () => {
+export const createDeletedLeafValue = (setLeafMarker = false) => {
   const bytes = new Uint8Array(32)
-  // Set the 129th bit to 1 directly by setting the 17th byte (index 16) to 0x80
-  bytes[16] = 0x80
+  // Set the 129th bit to 1 directly by setting the 17th byte (index 16) to 1 (since these bytes are little endian)
```
gabrocheleau, line 52: Are we able to add a reference for where this convention comes from?

PR #3651, opened 2024-09-10
monorepo: skip typecheck for node modules

gabrocheleau:
When running typescript's typecheck (npm run `tsc`), some internal errors from node_modules show up, which is a blocker to eventually adding this to the CI. This options skips node_modules internal typechecking


acolytec3 (approved):
LGTM

PR #3652, opened 2024-09-10
client: add missing types as devDep

gabrocheleau:
This adds two missing @ types devdependencies that were giving ts errors.


acolytec3 (approved):
LGTM

Issue #3634, opened 2024-09-05
Monorepo: Consistent DEBUG logger README docs & DEBUG=ethjs,* guarding


holgerd77:
For various packages (EVM, StateManager, Trie,...) we use the debug logger for verbose debug developer output. To not pollute end user debug sessions we guard runs with `DEBUG=ethjs,*` (so: `ethjs` needs to be added for logs to be visible) for EthereumJS library debugging.

As per comment from @acolytec3 in the chat...

<img width="825" alt="grafik" src="https://github.com/user-attachments/assets/98590697-b05b-4446-bdd8-440949e647d3">

...we seem to not have done this consistently. We should go through the libraries where the debug logger is used and complete this to have this consistent. This can generally also be confusing, @g11tech stumbled upon this interpreting the non-logged out stuff as "simply not running", things like that are unlucky side effects.

As some second complementary task: for most debug logger usages we have per-package documentation sections like for [devp2p](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/devp2p/README.md#debugging) or for [EVM](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/devp2p/README.md#debugging), for some we have not though. We should look through the libraries where debug is used and add respective sections, both for ourselves as well as other users of the libraries.

scorbajio:
Addressed by #3663 and #3654 and will be closed when the latter is merged.

PR #3654, opened 2024-09-11
Make logging consistent

scorbajio:
This change checks for the `ethjs` rail guard to be active for logging to be activated in the `client` and `devp2p` packages. This makes them consistent with the rest of the packages that abide by the guard rail, as discussed in #3634.

In order to test this, one can try running `DEBUG='ethjs,*' npm run client:start:dev2` and comparing it's log output to running `DEBUG='*' npm run client:start:dev2`. 

In a followup PR, the documentation of the monorepo packages that support logging will be updated with logging instructions/details.

### DIFF
--- a/packages/blockchain/src/consensus/clique.ts
+++ b/packages/blockchain/src/consensus/clique.ts
@@ -114,7 +114,13 @@ export class CliqueConsensus implements Consensus {
    */
   public _cliqueLatestBlockSigners: CliqueLatestBlockSigners = []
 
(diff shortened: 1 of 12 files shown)


acolytec3 (approved):

acolytec3 (commented):
LGTM

holgerd77:
Also massive! 🤩😎👍

Side note: didn’t even think about client, but this is actually a good preparation for the library-fication of the client! 🙏🙂

PR #3655, opened 2024-09-11
Several EVM Precompile Clean-Ups & Renamings

holgerd77:
This PR does the following:

- Renames the EVM bn254 precompiles (`ec` based names like `ecAdd` too generic, not consistent with bls12)
- Makes usage of the precompile names canonical by only using the versions from `precompiles/index.ts` `precompileEntries` as a source of truth (this comes along with renamings as well as inconsistency fixes, e.g. Identity -> IDENTITY, some hex address additions, KZG -> KZG_POINT_EVALUATION)
- Makes the (not fully exposed) `getPrecompileName()` a bit more flexible and comfortable, by left padding eventually left out 0s as well as return '' (instead of undefined) if no name available (simplifies typing, alternative would have been to just throw but decided that this might be a bit overblown for this type of method)
- `ec*Gas` param names have also be renamed, to `bn254*Gas`
- `Bls12381*Gas` param names have been renamed to `bls12381*Gas` (all our param names start with lower-case)

P.S.: Ignore the in-between hex reordering. I was confused. 🙂 🙃


packages/evm/src/precompiles/01-ecrecover.ts
```diff
@@ -13,13 +13,16 @@ import { OOGResult } from '../evm.js'
 
 import { gasLimitCheck } from './util.js'
 
+import { getPrecompileName } from './index.js'
+
 import type { ExecResult } from '../types.js'
 import type { PrecompileInput } from './types.js'
 
 export function precompile01(opts: PrecompileInput): ExecResult {
+  const pName = getPrecompileName('01')
```
jochem-brouwer, line 22: This constant can be moved outside this method, such that it does not get called each time if this precompile gets called (micro-optimization). (For all precompiles)

packages/evm/src/precompiles/index.ts
```diff
@@ -6,9 +6,9 @@ import { precompile02 } from './02-sha256.js'
 import { precompile03 } from './03-ripemd160.js'
 import { precompile04 } from './04-identity.js'
 import { precompile05 } from './05-modexp.js'
-import { precompile06 } from './06-ecadd.js'
-import { precompile07 } from './07-ecmul.js'
-import { precompile08 } from './08-ecpairing.js'
+import { precompile06 } from './06-bn254-add.js'
+import { precompile07 } from './07-bn254-mul.js'
+import { precompile08 } from './08-bn254-pairing.js'
```
jochem-brouwer, line 11: I am confused now. The EIPs https://eips.ethereum.org/EIPS/eip-196 https://eips.ethereum.org/EIPS/eip-197 mention that these are "alt_bn128"? It seems that `bn254` is an alias for this? :thinking:

jochem-brouwer (commented):
Two small comments

packages/evm/src/precompiles/index.ts
paulmillr: Yes, these are the same things. It was called bn128 when it had 128-bit security, but then researchers decreased security bits. So it's now bn254, because of 254-bit field

holgerd77: Yes, I followed Pauls advice here, also googling a bit and stumbling upon articles like https://hackmd.io/@jpw/bn254 for some confirmation that this is the "official" respectively "cryptographically correct" name now.

This naming confusion is a thing though, to add on this: there is even the name "bn256" (!!) around, this seems to be wrong and there is some explanation what happened to these two bits I forgot again. 😂 Geth is using this (wrong) name though, I just discovered yesterday (which should not bother us).

But let's stop here. This is really not something where we can clean up the mess. So: bn254 seems very much to be the best choice here.

packages/evm/src/precompiles/01-ecrecover.ts
holgerd77: Will do.

holgerd77: (and totally not so "micro", makes a lot of sense)

holgerd77:
Ok, addressed, ready for re-review! 🙂

packages/evm/src/precompiles/01-ecrecover.ts
holgerd77: Ah, this does not seem to work, CI breaks on several occasions:

grafik

When looking at the code, I think this is due to circular imports (this whole precompile code is generally not so optimally organized, but I would not want to expand this into a full round refactor), so the `precompiles/index.ts` file is importing all the precompiles but is then itself *exporting* this `getPrecompileName` method, this does not seem to be able to be resolved.

Can we therefore please keep the non-optimized version? Will delete the respective commit again and re-push. I do not have an easy fix for this.

jochem-brouwer (approved):

PR #3663, opened 2024-09-12
Update readmes with debugging info

scorbajio:
This change updates the documentation of packages that use the `debug` module with details on how to use the debugger, as discussed in https://github.com/ethereumjs/ethereumjs-monorepo/issues/3634.

This change also renames some of the debugger namespaces to make things more uniform.

In the `statemanager` package, the following namespaces have been renamed:
`statemanager:statemanager` -> `statemanager:MerkleStateManager`

`statemanager:rpcStateManager` -> `statemanager:RPCStateManager`

`statemanager:statefulVerkleStatemanager` -> `statemanager:StatefulVerkleStateManager`

`statemanager:verkle` -> `statemanager:StatelessVerkleStateManager`

And in the `evm` package, the following namespaces have been renamed:
`statemanager:statemanager` -> `evm:journal`


jochem-brouwer (approved):
LGTM! Only have to be careful with these logs that these should be properly upper/lower cased

packages/blockchain/README.md
```diff
@@ -239,9 +239,15 @@ The `Blockchain` class has a public property `events` which contains an `EventEm
...
+## Debugging
 
-For debugging blockchain control flows the [debug](https://github.com/visionmedia/debug) library is used and can be activated on the CL with `DEBUG=[Logger Selection] node [Your Script to Run].js`.
+For the debugging output to show up, set the `DEBUG` environment variable, e.g. in Linux/Mac OS:
+`export DEBUG=ethjs,*`
+
+`ethjs` **must** be included in the `DEBUG` environment variables to enable **any** logs.
+Additional log selections can be added with a comma separated list (no spaces). Logs with extensions can be enabled with a colon `:`, and `*` can be used to include all extensions.
+
+`DEBUG=ethjs,thislog,thatlog,otherlog,otherlog:sublog,anotherLog:* node myscript.js`
```
holgerd77, line 250: I think this "thislog, thatlog" is not very helpful, can me make this concrete and use an example from the library itself (also goes for some other docs below)?

It would also be good if all the loggers are just listed here as well (other library as well), even if there are only 2-3, so that users directly get an overview and guidance what to use.

packages/evm/src/journal.ts
```diff
@@ -53,7 +53,7 @@ export class Journal {
     this.DEBUG =
       typeof window === 'undefined' ? (process?.env?.DEBUG?.includes('ethjs') ?? false) : false
 
-    this._debug = debugDefault('statemanager:statemanager')
+    this._debug = debugDefault('evm:journal')
```
holgerd77, line 56: 👍

packages/statemanager/DEVELOPER.md
```diff
@@ -10,3 +10,35 @@ Tests can be found in the `tests` directory. The StateManager can be tested as:
...
+`DEBUG=ethjs,thislog,thatlog,otherlog,otherlog:sublog,anotherLog:* node myscript.js`
+
+The following initial logger is currently available:
+
+| Logger                                     | Description                                              |
+| ------------------------------------------ | -------------------------------------------------------- |
+| `statemanager:MerkleStateManager`          | Operations happening on the `MerkleStateManager`         |
+| `statemanager:RPCStateManager`             | Operations happening on the `RPCStateManager`            |
+| `statemanager:StatefulVerkleStateManager`  | Operations happening on the `StatefulVerkleStateManager` |
+| `statemanager:StatelessVerkleStateManager` | Operations accessing verkle witnesses                    |
```
holgerd77, line 31: Ah, it's generally a good idea to get consistent here! 👍 Can we nevertheless go the otherway around here and *shorten* all names, so `merkle`, `rpc` and for `verkle` it might even be ok to for both stay in the same namespace (since statemanagers aren't really used concurrently in most/all use cases, so no risk of mixing things up)?

In practice it just turned out that it becomes annying pretty quickly if logger names are too long, especially if one wants to concatenate 2,3 or 4 of them.

Also - the thing Jochem mentioned - debug loggers should always be lowercase (so: `statemanager` is also lowercase, just as an example) - I think this was also some discussion in the past and it generally makes sense since then its a lot easier to adhoc remember and not stumble over not casing things right on usage.

packages/verkle/README.md
```diff
@@ -80,6 +80,50 @@ const { EthereumJSClass } = require('@ethereumjs/[PACKAGE_NAME]')
...
+Run with minimal logging:
+
+```shell
+DEBUG=ethjs,verkle npx vitest test/verkle.spec.ts
+```
+
+Run with **put** method logging:
+
+```shell
+DEBUG=ethjs,verkle:PUT npx vitest test/verkle.spec.ts
```
holgerd77, line 112: Ah, ok, just wrote the lowercase argument, now seeing this here: I am pretty sure though, that we already had this as common ground, just some inconsistent introduction here. Can you also please just do `PUT` -> `put` and the like here along this work?

holgerd77 (commented):
Thanks a lot 🙂 👍, some things to address!

packages/blockchain/README.md
```diff
@@ -239,9 +239,15 @@ The `Blockchain` class has a public property `events` which contains an `EventEm
 | ------------------------ | ------------------------------------------- |
 | `deletedCanonicalBlocks` | Emitted when blocks are reorged and deleted |
 
-## Developer
+## Debugging
 
-For debugging blockchain control flows the [debug](https://github.com/visionmedia/debug) library is used and can be activated on the CL with `DEBUG=[Logger Selection] node [Your Script to Run].js`.
+For the debugging output to show up, set the `DEBUG` environment variable, e.g. in Linux/Mac OS:
+`export DEBUG=ethjs,*`
```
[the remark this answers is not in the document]
scorbajio: Fixed.

packages/verkle/README.md
scorbajio: I went ahead and reverted changing the examples to use lowercase for the method names, but I made the namespaces we have control over all uniformly use lowercase letters.

packages/blockchain/README.md
scorbajio: The `blockchain`, `verkle`, and `client` packages don't have enough namespaces available to make a concrete example that demonstrates using both specific namespaces and wildcards, so I reused the `statemanager` example for those packages.

Also collated and added the namespaces available in the client package as well since that one was missed, but the blockchain namespaces are fully listed out here, there's just a single one available.

packages/statemanager/DEVELOPER.md
scorbajio: I made all of the namespaces that we declare lowercase, but as mentioned just bellow, namespaces are case sensitive, and setting the debug to trace e.g. `trie:put` will not work while `trie:PUT` will, so there is going to be some inconsistency here.

packages/verkle/README.md
holgerd77: Ok, nice 🙂

packages/blockchain/README.md
holgerd77: Ok, that might be ("packages don't have enough namespaces available") but we DO actually want to document the usage of the specific packages here and it really is not that helpful to users if they find statemanager examples on the blockchain package (and if there is no usage for wildcards for blockchain so be it, then we do not need to document there and can just remove from the docs). Will make this small update and then merge, otherwise this looks good! 🙂

holgerd77: Update: ah, just seeing that there is also a blockchain example, ok. If possible we should then optimally rather use some other-lib examples which makes some sense in the documented library context, but for blockchain admittedly this is also difficult, guess there is no other logger (from our dependencies) which can really be used. Anyhow: I'll look a bit over the examples and maybe make some small adjustments (e.g. for client it is more of a "real world" example if one is citing the `devp2p` loggers e.g. there, statemanager is not really a common debug need there, and then I'll merge.

packages/client/src/sync/fetcher/bytecodefetcher.ts
```diff
@@ -61,7 +61,7 @@ export class ByteCodeFetcher extends Fetcher<JobTask, Uint8Array[], Uint8Array>
 
     this.keccakFunction = this.config.chainCommon.customCrypto.keccak256 ?? keccak256
 
-    this.debug = debug('client:ByteCodeFetcher')
+    this.debug = debug('client:fetcher:bytecode')
```
holgerd77, line 64: Ah, yes, that's a nicer hierarchy of things! 🙂 👍

packages/statemanager/src/statefulVerkleStateManager.ts
```diff
@@ -84,7 +84,7 @@ export class StatefulVerkleStateManager implements StateManagerInterface {
     this._trie =
       opts.trie ??
       new VerkleTree({ verkleCrypto: opts.verkleCrypto, db: new MapDB<Uint8Array, Uint8Array>() })
-    this._debug = debugDefault('statemanager:statefulVerkleStatemanager')
+    this._debug = debugDefault('statemanager:StatefulVerkleStateManager')
```
holgerd77, line 87: This is still the old name, will directly update

holgerd77: (same for the other verkl SM)

holgerd77 (approved):
LGTM

Issue #3664, opened 2024-09-12
Upgrade ethereum-cryptography to v3


paulmillr:
https://github.com/ethereum/js-ethereum-cryptography/releases/tag/3.0.0

It would allow to de-duplicate BLS, BN.

holgerd77:
Closed by #3668

PR #3668, opened 2024-09-13
Upgrade ethereum-cryptography / Remove Direct EVM @noble/curves Usage

ScottyPoi:
Issue: #3664 - 

All Packages:
Upgrades to ethereum-cryptography v3 (3.0.0)

- Wallet: Removes `await` from *non-Promise* `encrypt` and `decrypt` functions


holgerd77:
Ok, just wanted to test if deduplication of `@noble/curves` for the EVM works (there we have the direct dependency, and then we also have the one from `ethereum-cryptography`).

So this is `master` (with `@noble/curves` being double-present upper-right and lower-left, recognisable e.g. from the `weierstrass.js` inclusion):

grafik

And this is the PR:

grafik

Works! 🎉

Another 80 KB chopped off!

holgerd77:
Update: ah, just seeing this, I guess we can use `bls` and `bn` directly from [ethereum-cryptography](https://github.com/ethereum/js-ethereum-cryptography/releases/tag/3.0.0) - since it is exposed there (Paul also somewhat mentioned in the issue), then we can skip this direct `@noble/curves` dependency alltogether and do not run the risk of re-duplicating again by versions moving apart.

Will directly test (and merge) this in the realm of this PR.

holgerd77 (approved):
Cool, works 🙂, will take this in!

PR #3673, opened 2024-09-16
client/block: ensure BlockBodies requests get properly propagated

jochem-brouwer:
This PR:

- Fixes the BlockBytes type in Block (note: `RequestBytes` and `RequestsBytes` (plural!!) is a mistake hard-to-miss, maybe we should think of better naming for those types to avoid these kind of mistakes?)
- Ensures that BlockBodies now take the Requests sent with the BlockBodies response used to create the relevant blocks.

Will keep this at draft, will test (or someone else) later on devnet-3 if this fixes the issue where we report that clients are not sending us BlockBodies. So please only merge if this has been tested!

### DIFF
--- a/packages/block/src/types.ts
+++ b/packages/block/src/types.ts
@@ -160,7 +160,7 @@ export type BlockBodyBytes = [
   TransactionsBytes,
   UncleHeadersBytes,
   WithdrawalsBytes?,
-  RequestBytes?,
+  RequestsBytes?,
 ]
 /**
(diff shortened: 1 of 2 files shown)


g11tech (approved):

holgerd77:
> Will keep this at draft, will test (or someone else) later on devnet-3 if this fixes the issue where we report that clients are not sending us BlockBodies. So please only merge if this has been tested!

Has this been tested?

g11tech:
deploying this right now to devnet3

g11tech:
confirmed the patch to be working in prsym<>ethereumjs node which was stalled, so we gtg :+1:

PR #3674, opened 2024-09-16
Add `micro-eth-signer/kzg` 

acolytec3:
Takes an initial pass at integrating the kzg module from `micro-eth-signer`, putting it in `tx` for now.  

I've added one of the trusted setups for now so we don't have to build/import `trusted-setups`

Also removes `loadTrustedSetup` from the kzg interface.  We don't use it anywhere and I'm not convinced we will ever need to load another one.


acolytec3:
image

Here are some initial benchmarks for the various kzg functions we use.  Refer to `[root]/packages/tx/test/kzg.bench.ts` to review.

You can run with `npx vitest bench test/kzg.bench.ts` from the `tx` directory

packages/tx/src/jsKZGWrapper.ts
```diff
@@ -0,0 +1,45 @@
...
+
+//@ts-ignore  NOTE: Have placed the "fast" version of the trusted setup here for now till the `trusted-setups` package is published
+import { trustedSetup as fast } from './fast.js'
+
+import type { PrefixedHexString } from '@ethereumjs/util'
+// import { trustedSetup as fast } from 'trusted-setups/fast.js'
+
+const kzg = new KZG(fast)
+
+export const jsKZG: Kzg = {
```
holgerd77, line 12: This should implement our interface from Util

paulmillr:
The package is now live on NPM at `@paulmillr/trusted-setups`

holgerd77:
https://github.com/paulmillr/trusted-setups

packages/tx/package.json
```diff
@@ -57,12 +57,14 @@
     "@ethereumjs/common": "^4.4.0",
     "@ethereumjs/rlp": "^5.0.2",
     "@ethereumjs/util": "^9.1.0",
-    "ethereum-cryptography": "^3.0.0"
+    "ethereum-cryptography": "^3.0.0",
+    "trusted-setups": "github:paulmillr/trusted-setups"
```
holgerd77, line 61: This seems to have been forgotten. Will push a small update.

holgerd77:
Have updated the `package-lock.json` file (there was a conflict) and removed the forgotten depedency.

There are some client tests failing now I also planned to fix, but I am not fully seeing through, so there is some naming/interface thing happening in the tests (e.g. `getRawBlock.spec.ts`, also others) and I am not fully sure how to fix.

packages/client/bin/cli.ts
```diff
@@ -28,6 +28,8 @@ import {
   setLengthLeft,
   short,
 } from '@ethereumjs/util'
+//@ts-ignore trusted-setups doesn't include types
+import { trustedSetup as fast } from '@paulmillr/trusted-setups/fast.js'
```
jochem-brouwer, line 32: Nit: in tests it does not matter, but in production code I would suggest using a bit more descriptive name than `fast`. Something like `kzgSetupFast` or something.

packages/evm/src/opcodes/functions.ts
```diff
@@ -762,7 +762,7 @@ export const handlers: Map<number, OpHandler> = new Map([
     function (runState) {
       const index = runState.stack.pop()
       if (runState.env.blobVersionedHashes.length > Number(index)) {
-        runState.stack.push(bytesToBigInt(runState.env.blobVersionedHashes[Number(index)]))
+        runState.stack.push(BigInt(runState.env.blobVersionedHashes[Number(index)]))
```
jochem-brouwer, line 765: I will address this at some point, bit inside the EVM we should likely not put the strings in, but directly the bigints so we can push those to stack.

jochem-brouwer: Oh, hmm ok originally they were bytes which we kept converting to bigints, which is/was also suboptimal.

packages/evm/vite.config.bundler.ts
```diff
@@ -11,7 +11,7 @@ export default defineConfig({
       treeshake: 'safest',
     },
     lib: {
-      entry: './src/',
+      entry: '../tx/examples/blobTx.ts',
```
jochem-brouwer, line 14: This change should likely be reverted?

packages/util/test/kzg.bench.ts
jochem-brouwer, line 1: Nice :smile:

packages/client/bin/cli.ts
jochem-brouwer: Ah, just using `trustedSetup` here should be fine :) 
```suggestion
import { trustedSetup } from '@paulmillr/trusted-setups/fast.js'
```

jochem-brouwer (commented):
Some super small nits, LGTM otherwise

jochem-brouwer (approved):
LGTM

packages/util/test/kzg.bench.ts
```diff
@@ -0,0 +1,36 @@
...
+  })
+  describe('verifying proof', async () => {
+    bench('wasm verifyProof', () => {
+      kzg.verifyBlobKZGProofBatch([blob], [commit], [proof])
+    })
+    bench('js verifyProof', () => {
+      jsKZG.verifyBlobProofBatch([blob], [commit], [proof])
+    })
+  })
+})
```
holgerd77, line 36: Ah, this is also ok if we directly put this between the tests! 🙂👍

holgerd77: (maybe even better, even more „at hand“)

holgerd77:
🎉

Issue #3671, opened 2024-09-16
Blockchain: Add debug Logger


holgerd77:
When working on the debug logger docs here #3663 we (re-)discovered that the `blockchain` package has not even a basic debug logger for the core functionality.

Since blockchain is also a very procedural library it makes a lot of sense to add at least some basic debug logging to at least the core steps of the longer methods like (that's totally in the center) the `putBlocks` flow and maybe the validation methods. For the others (methods) I would suggest a bit spare and not overdo, maybe one for `delBlock()` or so, surely not for things like `getIteratorHead()` (at least I would think so?).

I think if we have these basic logging functionality in some few additional logging lines not added in this first round will then also add naturally if someone looks into something and then misses some output.

scorbajio:
Closed by #3676

PR #3676, opened 2024-09-17
Blockchain debug additions

scorbajio:
This change adds debug logging to core parts of the `blockchain` package as discussed in #3671


packages/blockchain/README.md
```diff
@@ -245,16 +245,18 @@ This library uses the [debug](https://github.com/visionmedia/debug) debugging ut
...
 The following initial logger is currently available:
 
-| Logger              | Description                                                 |
-| ------------------- | ----------------------------------------------------------- |
-| `blockchain:clique` | Clique operations like updating the vote and/or signer list |
+| Logger                        | Description                                                              |
+| ----------------------------- | ------------------------------------------------------------------------ |
+| `blockchain:core`             | Core blockchain operations like when a block or header is put or deleted |
+| `blockchain:consensus:clique` | Clique consensus operations like updating the vote and/or signer list    |
+| `blockchain:consensus:ethash` | Ethash consensus operations like PoW block or header validation          |
```
holgerd77, line 252: Can we really please keep these logger names short (and not over-namespacify 😂 )? We do not need to be 100% consistent here and not sure if you have used loggers a lot in the past but it really becomes annoying quickly if these logger names are long.

(also this :core addition is a bit inconsistent, since we are not doing/using this on any other logger for the core class)

So let's please simply do `blockchain`, `blockchain:clique` and `blockchain:ethash`. This is mostly for our internal usage and we can make "breaking" changes here anytime.

packages/blockchain/src/blockchain.ts
```diff
@@ -452,6 +466,8 @@ export class Blockchain implements BlockchainInterface {
         await this.dbManager.batch(ops)
 
         await this.consensus?.newBlock(block, commonAncestor, ancestorHeaders)
+        this.DEBUG &&
+          this._debug(`put block: number ${block.header.number} hash ${bytesToHex(block.hash())}`)
```
holgerd77, line 470: Do not have the full picture but if the hash is available beforehand it would be good to not calculate (call) again for performance reasons but simply take an eventually existing assigned variable.

Can you have a look through the `hash()` usages (several below as well) where possible?

holgerd77 (commented):
Ah, thanks for starting, but this is not yet going as deep as I intended with the issue! We can eventually nevertheless merge when the comments are addressed, since it is for sure already consistent and useful on its own.

I meant this more than a simple work down issue though and really go one step deeper and give this some substantial thinking *where* debug messages are useful, what are delicate situations and so on.

Not sure, maybe I am also off here myself, wait, I will just go through the library and write down what I would identify as "interesting and/or debug-worthy events" 🙂 (might mention some which are already in the logs, won't "sync" with the changes here, then simply ignore).

Will number here, just if we want to "talk about one of them", so order has no other meaning:

1. When a genesis block is initialized (so that's in the constructors)
2. How many blocks are put with `putBlocks` or `putHeaders`
3. How many blocks are deleted with `_deleteCanonicalChainReferences`
4. From where to where `resetCanonicalHead` is setting a head
5. What common ancestor `findCommonAncestor` has found and how many ancestor headers
6. How many `staleHeads` have been found (also citing `header`) in `_rebuildCanonical`

Ok. That's what I could identify. 🙂 

I guess that's not too much! Can you therefore directly in this PR?

packages/blockchain/src/blockchain.ts
scorbajio: Fixed

packages/blockchain/README.md
scorbajio: Changed this back to what it was before for now. Will create an issue on rethinking namespaces to allow wildcards like `DEBUG=ethjs,blockchain:*` work for all namespaces in blockchain and not unexpectedly exclude the core namespace, as is happening currently.

holgerd77 (commented):
Some of these debug messages can need a bit more content. Will push an update, then we can merge! 🙂

holgerd77 (approved):
LGTM

PR #3688, opened 2024-09-19
Pin vitest deps

acolytec3:
Fixes the VM nightly jobs

1) Pins vitest to 2.1.0 so we don't get point update mismatches
2. Update `verkle-cryptography-wasm` and the corresponding interface so it's no longer broken


gabrocheleau (approved):

gabrocheleau (approved):

holgerd77:
Nice, thanks! 👍🙂

Definitely fixed the related VM nightly stuff, just had a look!

Issue #3693, opened 2024-09-21
Byte utilities: reduce duplication, use optimized technique


paulmillr:
While browsing [single-file-evm.js](https://gist.github.com/holgerd77/2c032488196b4afee5d976dc85ee70eb), i've noticed there are at least a bunch of duplicate hexToBytes and concatBytes methods. While duplication is not a big deal, it could be avoided.

But the issue is about optimization. noble exposes [hexToBytes in its code](https://github.com/paulmillr/noble-curves/blob/d0a8d2134c5737d9d0aa81be13581cd416ebdeb4/src/abstract/utils.ts#L63-L91). I see ethereumjs implemented custom version using Regular Expressions, which are [2x slower by my benchmarks](https://gist.github.com/paulmillr/b7ab94e3db799eb4166c258e03eac8af). On larger inputs, the difference is 3x.

Overall, noble utilities could be used in a following way:

```js
import { hexToBytes as unprefixedHexToBytes } from '@noble/hashes/utils';
function hexToBytes(hex) {
  if (!hex.startsWith('0x')) throw new Error('hex string must start with 0x');
  return unprefixedHexToBytes(hex.slice(2));
}
```

Let me know if you want it to be exposed from `ethereum-cryptography`, which `utils` are already using.

acolytec3:
Let me do some benchmarking of our version vs noble.  Definitely open to it if it proves out that yours is faster (since we do boatloads of hex <-> bytes conversions.  I think the regex in ours is just to check for non hex characters so could potentially be removed if it was the sole source of drag.

acolytec3:
![image](https://github.com/user-attachments/assets/94b2d2e2-bf64-467b-a350-bc6deac0c41b)
 
Victory - noble

acolytec3:
I added a second test for `bytes2Hex`.

![image](https://github.com/user-attachments/assets/814a07ba-881a-41aa-b975-298673d0f514)

Victory - noble

acolytec3:
Here's the benchmark script (using the vitest benchmark feature) if anybody else wants to run it for comparison.

```ts
import { bytesToHex as nobleB2H, hexToBytes as nobleH2B } from '@noble/curves/abstract/utils'
import { bench, describe } from 'vitest'

import { bytesToHex, randomBytes, unprefixedHexToBytes } from '../src/bytes.js'

describe('h2b benchmarks', () => {
  bench('noble', () => {
[...]

PR #3698, opened 2024-09-23
Use `noble` bytes conversion utilities internally

acolytec3:
Addresses #3693 since the noble hex to bytes conversion utility is about 2x ours.


packages/util/src/bytes.ts
```diff
@@ -1,6 +1,9 @@
 import { getRandomBytesSync } from 'ethereum-cryptography/random.js'
 // eslint-disable-next-line no-restricted-imports
-import { bytesToHex as _bytesToUnprefixedHex } from 'ethereum-cryptography/utils.js'
+import {
+  bytesToHex as _bytesToUnprefixedHex,
```
paulmillr, line 4: it's not unprefixed - it allows 0x, but doesn't mandate it

holgerd77:
Browser tests still failing

holgerd77:
Update: just read up on Discord, let me know if I should still deactivate the browser tests.

packages/util/src/bytes.ts
```diff
@@ -73,6 +54,9 @@ export const bytesToBigInt = (bytes: Uint8Array, littleEndian = false): bigint =
   if (littleEndian) {
     bytes.reverse()
   }
+
+  // Ensure that bytes are actually bytes (and not just an array of numbers)
+  // bytes = Uint8Array.from(bytes)
```
jochem-brouwer, line 59: This is commented out?

packages/client/test/net/protocol/ethprotocol.spec.ts
```diff
@@ -70,7 +70,7 @@ describe('[EthProtocol]', () => {
       'encode status',
     )
     const status = p.decodeStatus({
-      chainId: [0x01],
+      chainId: Uint8Array.from([0x01]),
```
jochem-brouwer, line 73: Why is this change now necessary?

paulmillr: the function accepted arrays before. Now it is limited to u8a

packages/util/src/bytes.ts
acolytec3: Meant to remove it entirely.  It breaks a bunch of VM tests and doesn't solve anything other than a couple of client tests that were already incorrect.

packages/client/test/net/protocol/ethprotocol.spec.ts
acolytec3: Furthermore, this property is supposed to be a Uint8Array if you look at the types.

scorbajio (approved):
LGTM

config/vitest.config.browser.mts
```diff
@@ -22,6 +22,7 @@ const config = defineConfig({
   ],
   optimizeDeps: {
     exclude: ['kzg-wasm'],
+    include: ['vite-plugin-node-polyfills/shims/buffer', 'vite-plugin-node-polyfills/shims/global', 'vite-plugin-node-polyfills/shims/process']
```
holgerd77, line 25: At first look, these additions seem to go against our as-pure-as-possible-browser-support direction strategy. 🤔 

Is this something to further investigate (why needed, what changed, where to adjust)?

(I just created a non-minifed VM (so not: EVM) bundle as a first test and searched for "Buffer" but did not find anything suspicious)

acolytec3: I would agree that this was my surface thought as well.  We have these various shims by default as a part of the `vite-plugin-node-polyfills` dep for running our tests and vitest was having trouble reloading them or something (saw these messages about optimizing these deps over and over again in the logs) and the suggested fix was to add them to the `optimizeDeps` list in our config.  Given the effort on getting rid of polyfills, we might be able to just get rid of that plugin though I'm not 100% sure.  Worth researching.

Issue #3687, opened 2024-09-19
Monorepo: Suggestion for a better (and any-less) Error Handling


holgerd77:
One of the most common `any` typing we still have in the monorepo is for typing of the error parameter in the `try / catch` blocks like [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/evm/src/precompiles/07-bn254-mul.ts#L27) for an EVM precompile:

```ts
try {
  returnData = (opts._EVM as EVM)['_bn254'].mul(input)
} catch (e: any) {
  if (opts._debug !== undefined) {
    opts._debug(`${pName} failed: ${e.message}`)
  }
  return EvmErrorResult(e, opts.gasLimit)
}
```

This is non-beautiful for test code and problematic for production code, since in JavaScript, the type of `e` is just not defined and eventually this error code will not work (if an underlying library e.g. just throws a string). Tested with the following code snippets what will actually happen:

![grafik](https://github.com/user-attachments/assets/0dc57f2e-ac34-4fee-b531-1432ef8eff51)

TypeScript is handling the JS error situation by just applying type `unknown` as a type for `e`, see [here](https://stackoverflow.com/questions/54649465/how-to-do-try-catch-and-finally-statements-in-typescript) (which makes a lot of sense respectively is adequate). I tested:

![grafik](https://github.com/user-attachments/assets/b22e5e50-34ba-4435-afca-6d92777ab68a)

We might nevertheless also want to set the [useUnknownInCatchVariables](https://www.typescriptlang.org/tsconfig/#useUnknownInCatchVariables) flag.

So here is a suggestion for how we can improve the overall situation:

### Production Code

For production code we remove any eventual `any` typing and accept the fact that a type for an error is just not known when being caught in JS. Then we add an exception value normalization as proposed [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#catch_binding) in the Mozilla docs somewhere down the line, like:

```js
try {
  throw "Oops; this is not an Error object";
} catch (e) {
  if (!(e instanceof Error)) {
    e = new Error(e);
  }
  console.error(e.message);
}
```

This should make our (most of our) production code try/catch clauses more robust and safe to rely on.

### Test Code

[...]

PR #3700, opened 2024-09-24
monorepo: Normalize exception values in error handling

ScottyPoi:
Implements suggestion in issue #3687

Replacing this pattern:
```ts
try {
...
} catch (e: any) {
   ...
}
```

with

```ts
try {
...
} catch (e) {
   if (!(e instanceof Error) {
         e = new Error(e)
       }
  ...
}
```


config/eslint.cjs
```diff
@@ -56,7 +56,8 @@ module.exports = {
       'error',
       { argsIgnorePattern: '^_', varsIgnorePattern: '^_' },
     ],
-    '@typescript-eslint/no-redeclare': ['error'],
+    '@typescript-eslint/no-redeclare': 0,
```
holgerd77, line 59: I don't think we have to set this down, when reading on [no-redeclare](https://eslint.org/docs/latest/rules/no-redeclare) we are not re*declaring* anyhting but only *reassiging*, which is totally norma, right?

I did a quick test re-adding the original rule setup and run`npm run lint` from root, still seems to work.

config/eslint.cjs
```diff
@@ -56,7 +56,8 @@ module.exports = {
...
+    'no-ex-assign': 0,
```
holgerd77, line 60: Ok, this rule [no-ex-assign](https://eslint.org/docs/latest/rules/no-ex-assign) seems out of question (as discussed)

config/tsconfig.json
```diff
@@ -14,6 +14,7 @@
     "strict": true,
     "target": "es2020",
     "lib": ["ES2020", "DOM"],
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "useUnknownInCatchVariables": false
```
holgerd77, line 18: Ah, this needs to be the otheway around: we *want* to have `unknown` there, atm this now goes to `any`:

grafik

So and this is the version of when removing the new assignment with `e` still being `any`:

grafik

So no underlying of `e.message` (what we would want. `any` is just saying: use whatever you want, I do not care! 🙂  

With `unknown` is the more strict version - indicating that for `e` we just don't (and can't) know the type.

Ah, just seeing that a plain use in `new Error()` is then not directly possible. 🤔 

grafik
grafik

But for `any` we still have typed as `any`. Also not optimal:

grafik

Hmm. Let me give this some thought, input welcome from the team.

holgerd77:
Will give this a "WIP" label again to indicate that we need to re-think this to some extend and this needs some continued work, so that this won't get merged accidentally.

We nevertheless should already think and reason about this.

jochem-brouwer:
Yes this is correct! Note, this set `useUnknownInCatchVariables` to `true`!

Issue #3701, opened 2024-09-25
Trie: Better Structured & Expanded Checkpointing Tests


holgerd77:
This is a follow-up on #3645, post by @TimDaub.

So the mentioned issue is about trie failures in the context of checkpointing usage. I guess the only realistic way that we can catch this respectively generally make this part of Trie more robust is by expanding on the checkpointing tests from [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/trie/test/trie/checkpoint.spec.ts).

These are not as structured and do not cover as much the complete range of combinations as the lot more structured [StateManager tests](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/statemanager/test/checkpointing.account.spec.ts).

So I guess the best way is to somewhat more align and roughly (trie adopted with key/values for sure) the labelling mechanism from the SM tests (e.g. `A1.1 -> CP -> A1.2 -> Revert -> Flush() (-> A1.1)`) and then inject/integrate the existing trie checkpointing tests (or: move over) and then see which cases are left and fully add these as new tests.

(this would be generally good to have this, since if we identify a new missing combination in the future in one of the libraries, it get's a lot easier to just directly add to the other)

From my intuition I would guess chances are somewhat high that we discover 1-2 bugs by this (> 50%, we'll see).

PR #3705, opened 2024-09-26
Trie checkpoint test expansion

scorbajio:
This change expands the `trie` package check-pointing tests, as suggested in #3701.


scorbajio:
I've started with porting over the statemanager account checkpointing test cases. No tests failed, but we could continue adding cases from there.

holgerd77:
That looks already super great 🤩, can you make delete the old cases in the file that are more double tested, so that we get to one aligned systematic and go out of this double structure? Optimally *all* old test cases go over!

holgerd77 (approved):
Looks great, thanks a lot! 👍🙂🙏

Issue #3704, opened 2024-09-25
Review and improve trie node typing


scorbajio:
Currently, the trie `Node` base type is extended in the extension and leaf node implementations (but unexpectedly, not in the branch node implementation) and it has some typing issues, such as the `value()` function on different node classes being typed to return a `Uint8Array` but it can also return `null` or `Uint8Array[]`. This issue, and other similar typing issues or class design issues should be considered and fixed if possible or notes/documented for future review. Also see [this comment](https://github.com/ethereumjs/ethereumjs-monorepo/pull/3692#issuecomment-2374783456) for more context on the problem.

scorbajio:
In order to recreate a basic example of the issue, check running `npx vitest test/util/log.spec.ts` in the trie package. Also navigate to and look at this code section of the `onFound` function in `trie.ts` in the `src` folder of the `trie` package:
```
        for (const k of node.key()) {
          console.log('dbg100')
          console.log(node.value())
          this.DEBUG && this.debug(`NextNode: ${node.value()}`, ['find_path', 'extension_node'])
          if (k !== targetKey[progress]) {
            result = { node: null, remaining: targetKey.slice(_progress), stack }
            return
          }
          progress++
        }
        walkController.allChildren(node, keyProgress)
```
We debug by printing the `node.value()` of each node being walked.:
```
...
dbg100
Uint8Array(32) [
  212,  59, 135, 253, 205,  66,  23,   1,
   60, 204, 146, 208,  70,  98, 225,  45,
   54, 228, 204,  37, 220, 105,   0, 119,
  205, 130,  26,  25,  86, 252,  62,  54
]
dbg100
[
  Uint8Array(0) [],
  Uint8Array(0) [],
  Uint8Array(0) [],
  Uint8Array(0) [],
  Uint8Array(0) [],
  Uint8Array(0) [],
  [ Uint8Array(1) [ 53 ], Uint8Array(4) [ 99, 111, 105, 110 ] ],
  Uint8Array(0) [],
  Uint8Array(0) [],
  Uint8Array(0) [],
  Uint8Array(0) [],
  Uint8Array(0) [],
  Uint8Array(0) [],
  Uint8Array(0) [],
  Uint8Array(0) [],
  Uint8Array(0) [],
  Uint8Array(5) [ 112, 117, 112, 112, 121 ]
]
...
```
[...]

PR #3707, opened 2024-09-27
Trie node typing exploration

scorbajio:
This PR explores some options with reworking some of the typings in the `trie` package. Also see #3704.

### DIFF
--- a/packages/trie/src/node/branch.ts
+++ b/packages/trie/src/node/branch.ts
@@ -4,11 +4,11 @@ import type { EmbeddedNode } from '../types.js'
 
 export class BranchNode {
   _branches: (EmbeddedNode | null)[]
-  _value: Uint8Array | null
+  _value: Uint8Array
 
   constructor() {
     this._branches = new Array(16).fill(null)
-    this._value = null
+    this._value = new Uint8Array(0)
   }
 
   static fromArray(arr: Uint8Array[]): BranchNode {
@@ -18,12 +18,12 @@ export class BranchNode {
     return node
   }
 
-  value(v?: Uint8Array | null): Uint8Array | null {
-    if (v !== null && v !== undefined) {
+  value(v?: Uint8Array): Uint8Array {
+    if (v !== undefined) {
       this._value = v
     }
 
-    return this._value && this._value.length > 0 ? this._value : null
+    return this._value
   }
 
   setBranch(i: number, v: EmbeddedNode | null) {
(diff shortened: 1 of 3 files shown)


scorbajio:
Currently, commenting out the `EmbeddedNode` type alias and expanding it throughout the `trie` package source code results in a type error, which may be resulting in [inconsistencies discovered in typing](https://github.com/ethereumjs/ethereumjs-monorepo/issues/3704#issuecomment-2375327783).

packages/trie/src/node/branch.ts
```diff
@@ -1,14 +1,12 @@
 import { RLP } from '@ethereumjs/rlp'
 
-import type { EmbeddedNode } from '../types.js'
-
 export class BranchNode {
-  _branches: (EmbeddedNode | null)[]
-  _value: Uint8Array | null
+  _branches: (Uint8Array | Uint8Array[] | null)[]
```
jochem-brouwer, line 4: We can make this even more explicit and type-safe:

1. We know that this array is always of length 16
2. We know that the elements are either `null` (the branch is empty), `Uint8Array` (points to hash of next node) or `[Uint8Array, Uint8Array]` which is of type `RawExtensionNode` or `RawLeafNode` (these are new types), which represent `[keyRemainder, nodeHashOrValue]`. Here `keyRemainder` is whatever key the extension/leaf node extends with, and `nodeHashOrValue` is either a hash (pointed to by the `ExtensionNode`) or a value (pointed to by a `LeafNode`)

packages/trie/src/node/branch.ts
```diff
@@ -1,14 +1,12 @@
...
+  _value: Uint8Array
 
   constructor() {
     this._branches = new Array(16).fill(null)
-    this._value = null
+    this._value = new Uint8Array(0)
```
jochem-brouwer, line 9: This will not work, now all branch nodes have value 0. This is not possible, for instance for the unhashed trie where we have these key/values:

`dog` -> 1
`done` -> 2

If you would index this on `do` (or `d`) it would return value 0, where it should be `null`. This will likely not mess up the hash of the tries, but it will mess up people using the tries (likely the unhashed versions) as key/value store. They would expect empty values to return `null` and not empty Uint8Arrays.

packages/trie/src/trie.ts
```diff
@@ -710,7 +709,7 @@ export class Trie {
     }
 
     if (lastNode instanceof BranchNode) {
-      lastNode.value(null)
+      lastNode.value()
```
jochem-brouwer, line 712: Weird, this doesn't do anything right (not even in the original code)?

scorbajio:
Fixed by #3708

PR #3708, opened 2024-09-28
trie: improve node typings and class architecture 

### LINKED ISSUE #3704: Review and improve trie node typing

gabrocheleau:
This PR resolves #3704 by:
- Renaming the Node class to the more accurate `ExtensionAndLeafNodeBase` class and making it abstract.
- Removing the `encodeKey` method from the extension and leaf node classes as they were not used
- Adding RawExtensionNode and RawLeafNode types
- Removing the EmbeddedNode type and replacing it with the `NodeReferenceOrRawNode` type
- Adding a `BranchNodeBranchValue` type that fits the types of the values of the _branches of the branch node class.

### DIFF
--- a/packages/trie/src/node/extensionOrLeafNodeBase.ts
+++ b/packages/trie/src/node/extensionOrLeafNodeBase.ts
@@ -8,12 +8,12 @@ import type { Nibbles, RawExtensionNode, RawLeafNode } from '../types.js'
 export abstract class ExtensionOrLeafNodeBase {
   _nibbles: Nibbles
   _value: Uint8Array
-  _terminator: boolean
+  _isLeaf: boolean
 
-  constructor(nibbles: Nibbles, value: Uint8Array, terminator: boolean) {
+  constructor(nibbles: Nibbles, value: Uint8Array, isLeaf: boolean) {
     this._nibbles = nibbles
     this._value = value
-    this._terminator = terminator
+    this._isLeaf = isLeaf
   }
 
   static decodeKey(key: Nibbles): Nibbles {
@@ -21,7 +21,7 @@ export abstract class ExtensionOrLeafNodeBase {
   }
 
   encodedKey(): Nibbles {
-    return addHexPrefix(this._nibbles.slice(0), this._terminator)
+    return addHexPrefix(this._nibbles.slice(0), this._isLeaf)
   }
 
   key(k?: Nibbles): Nibbles {
--- a/packages/trie/src/node/extensionOrLeafNodeBase.ts
+++ b/packages/trie/src/node/extensionOrLeafNodeBase.ts
@@ -3,9 +3,9 @@ import { RLP } from '@ethereumjs/rlp'
 import { addHexPrefix, removeHexPrefix } from '../util/hex.js'
 import { nibblesTypeToPackedBytes } from '../util/nibbles.js'
 
-import type { Nibbles } from '../types.js'
+import type { Nibbles, RawExtensionNode, RawLeafNode } from '../types.js'
 
-export class Node {
+export abstract class ExtensionOrLeafNodeBase {
   _nibbles: Nibbles
   _value: Uint8Array
   _terminator: boolean
@@ -18,6 +18,10 @@ export class Node {
 
   static decodeKey(key: Nibbles): Nibbles {
     return removeHexPrefix(key)
+  }
+
+  encodedKey(): Nibbles {
+    return addHexPrefix(this._nibbles.slice(0), this._terminator)
   }
 
   key(k?: Nibbles): Nibbles {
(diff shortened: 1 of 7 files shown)


packages/trie/src/node/extensionOrLeafNodeBase.ts
```diff
@@ -3,23 +3,27 @@ import { RLP } from '@ethereumjs/rlp'
...
 
-import type { Nibbles } from '../types.js'
+import type { Nibbles, RawExtensionNode, RawLeafNode } from '../types.js'
 
-export class Node {
+export abstract class ExtensionOrLeafNodeBase {
   _nibbles: Nibbles
   _value: Uint8Array
-  _terminator: boolean
+  _isLeaf: boolean
```
jochem-brouwer, line 11: With these changes I suggest we remove the `ExtensionNode` and `LeafNode` completely. We could introduce static constructors here which return the right node type. (`_isLeaf` thus needs to be public in order to determine if it is a leaf/extension node)

gabrocheleau: I went back and forth on this earlier today, but figured it was better to have these classes separate just for naming and clarity purposes, just so that we can at least do `new LeafNode` and `new ExtensionNode` instead of something like `new ExtensionOrLeafNode(isLeaf: true)`. What do feel is the value in having them both be the same class?

jochem-brouwer: You are right, the classes are much more clear for readability! Let's not change it.

acolytec3 (dismissed):
This looks great.  Makes so much more sense now.

packages/trie/src/types.ts
```diff
@@ -8,9 +8,17 @@ export type TrieNode = BranchNode | ExtensionNode | LeafNode
 
 export type Nibbles = number[]
 
+// [ encodedPath, key ]
```
jochem-brouwer, line 11: Hmm, I find this documentation somewhat misleading. The `key` in this context would be the pointer to whatever hash this `ExtensionNode` points to. I think this is implied to by "key" but this might be confused with the MPT-keys (so the ones you use as `get` argument). Not sure what a better term here would be though. `referencedHash`? Or maybe `pointerHash`?

gabrocheleau: I've copied what is in the ethereum.org documentation. I agree that it is not immediately obvious what this represents. I will remove these two comments and instead add a link to the docs.

acolytec3: I dunno if I like the updated comment any better. The docs do explain what the extension node is, in that the `key` is the DB key for the next node. 

Can we get rid of the "raw node" terminology and just have extension nodes, which are simply the encoded path to the node and then the DB key for the next lookup, and then lead nodes, which are the encoded path and then the leaf value (which is either an rlp encoded account or null, right)?

gabrocheleau: Yeah that's fair. 

I'm not sure I get the idea behind getting rid of the "raw node" terminology. We could do a renaming, but then we would have to rework the classes since these are already called "ExensionNode" and "LeafNode". Additionally, in comments and various places (e.g. methods like `decodeRawNode`), we have been referring to "raw nodes" quite often, and I would tend not to make changes that are too intrusive. 

I've updated the comment, let me know if that seems clear enough

acolytec3 (approved):
LGTM

jochem-brouwer (approved):
LGTM

packages/trie/src/types.ts
```diff
@@ -8,9 +8,18 @@ export type TrieNode = BranchNode | ExtensionNode | LeafNode
 
 export type Nibbles = number[]
 
+// A raw node refers to the non-serialized, array form of the node
+// A raw extension node is a 2-item node, where the first item is the encoded path to the next node, and the second item is the reference to the next node
+// A raw leaf node is a 2-item node, where the first item is the remaining path to the leaf node, and the second item is the value
+// To learn more: https://ethereum.org/en/developers/docs/data-structures-and-encoding/patricia-merkle-trie/#optimization
```
jochem-brouwer, line 14: Yes, great that the extra docs on top have been added. In the future this external URL could 404, so in general we should add extra comments if we point to external URLs :smile: :+1:

PR #3709, opened 2024-09-28
monorepo: remove redundant fills and zeros function

gabrocheleau:
This PR removes instances where we are needlessly filling Uint8Array's with zeroes with `fill(0)` (this is unnecessary because Uint8Arrays are initialized with zeroes already).

The PR also removes the `zeros` method, which was useful in the Buffer days since that operation was somewhat complex, but now it does not seem worth it to keep it.

### DIFF
--- a/packages/block/src/header/header.ts
+++ b/packages/block/src/header/header.ts
@@ -17,7 +17,6 @@ import {
   equalsBytes,
   hexToBytes,
   toType,
-  zeros,
 } from '@ethereumjs/util'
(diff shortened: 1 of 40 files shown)


jochem-brouwer (approved):
LGTM. Have re-triggered CI because browser tests failed (could not load some dynamic module from some url). Hopefully it now passes.

holgerd77:
Oh, very nice clean up, didn’t notice yet. 🙏

PR #3710, opened 2024-09-29
Clean up dev dependencies

acolytec3:
Removes `vitest-plugin-node-polyfills` from our dev dependencies and fixes related tests
Removes  `@ethersproject/abi` since it was only used to derive calldata for a single test


jochem-brouwer (commented):
Changes LGTM, however the examples also seem to have a dependency on `ethersproject/abi` 

Also, the browser tests fail (re-ran it once, gave same error): 

```
Run npm run install-browser-deps
  
> install-browser-deps
> npm install webdriverio @vitest/browser
npm error Cannot read properties of null (reading 'edgesOut')
npm error A complete log of this run can be found in: /home/runner/.npm/_logs/2024-09-29T21_49_25_024Z-debug-0.log
Error: Process completed with exit code 1.
```

See: https://github.com/ethereumjs/ethereumjs-monorepo/actions/runs/11088816287/job/30825710880?pr=3710

package.json
```diff
@@ -68,7 +67,7 @@
     "@rollup/rollup-linux-x64-gnu": "*"
   },
   "peerDependencies": {
-    "@vitest/browser": "2.1.0",
+    "@vitest/browser": "^2.1.0",
     "webdriverio": "^8.39.0"
```
jochem-brouwer, line 71: If the `install-browser-deps` now installs specific versions of these dependencies, should we then also remove the caret here to ensure this specific version is installed?

packages/vm/test/api/customChain.spec.ts
```diff
@@ -97,13 +96,11 @@ describe('VM initialized with custom state', () => {
     common.setHardfork(Hardfork.London)
     const vm = await createVM({ blockchain, common })
     await vm.stateManager.generateCanonicalGenesis!(genesisState)
-    const sigHash = new Interface(['function retrieve()']).getSighash(
-      'retrieve',
-    ) as PrefixedHexString
+    const calldata = '0x2e64cec1' // This is the call data to trigger the `retrieve` method on the given contract
```
jochem-brouwer, line 99:
```suggestion
    const calldata = new Interface(['function retrieve()']).getFunction('retrieve')!.selector
```

Since we have kept the ethers dependency in I think we should (just like the other example) use the interface to derive the correct selector (for clarity)

jochem-brouwer (commented):
Two nits/questions

packages/vm/test/api/customChain.spec.ts
acolytec3: Let me see if this works without bringing in those node builtins again.  Not sure if Richard got rid of them or not.

jochem-brouwer: Whoops, have to import Interface as well :sweat:

package.json
acolytec3: I tried this but when I run the `npm run install-browser-deps` script, it automatically adds the caret back (no idea why).  It should be okay though since we're installing specific versions from the script.

scorbajio (approved):
ACK

holgerd77:
Cool, great! 🙂 🙏 

That's really a nice step towards better assured browser compatibility, so that one not always have to doubt if/why these plugins/polyfills are needed.

PR #3711, opened 2024-09-29
client: add getBlobsV1 to the client to support CL blob import

g11tech:
ref: 
 - https://github.com/ethereum/execution-apis/pull/559

TODO: add to spec test

### DIFF
--- a/packages/client/src/rpc/modules/engine/engine.ts
+++ b/packages/client/src/rpc/modules/engine/engine.ts
@@ -1523,6 +1523,16 @@ export class Engine {
   }
 
   private async getBlobsV1(params: [[Bytes32]]): Promise<(BlobAndProofV1 | null)[]> {
+    const blobGasLimit = this.config.chainCommon.param('maxblobGasPerBlock')
+    const blobGasPerBlob = this.config.chainCommon.param('blobGasPerBlob')
+    const allowedBlobsPerBlock = blobGasLimit / blobGasPerBlob
+    if (params[0].length > allowedBlobsPerBlock) {
+      throw {
+        code: TOO_LARGE_REQUEST,
+        message: `More than allowedBlobsPerBlock=${allowedBlobsPerBlock} hashes queried`,
+      }
+    }
+
     const blobsAndProof: (BlobAndProofV1 | null)[] = []
     for (const versionedHashHex of params[0]) {
       blobsAndProof.push(this.service.txPool.blobsAndProofsByHash.get(versionedHashHex) ?? null)
--- a/packages/client/src/rpc/modules/engine/engine.ts
+++ b/packages/client/src/rpc/modules/engine/engine.ts
@@ -1522,9 +1522,9 @@ export class Engine {
     return payloads
   }
 
-  private async getBlobsV1(params: [Bytes32]): Promise<(BlobAndProofV1 | null)[]> {
+  private async getBlobsV1(params: [[Bytes32]]): Promise<(BlobAndProofV1 | null)[]> {
     const blobsAndProof: (BlobAndProofV1 | null)[] = []
-    for (const versionedHashHex of params) {
+    for (const versionedHashHex of params[0]) {
       blobsAndProof.push(this.service.txPool.blobsAndProofsByHash.get(versionedHashHex) ?? null)
     }
 
--- a/packages/client/src/rpc/modules/engine/engine.ts
+++ b/packages/client/src/rpc/modules/engine/engine.ts
@@ -52,6 +52,7 @@ import type { Config } from '../../../config.js'
 import type { VMExecution } from '../../../execution/index.js'
 import type { FullEthereumService, Skeleton } from '../../../service/index.js'
 import type {
+  BlobAndProofV1,
   Bytes32,
   Bytes8,
   ExecutionPayloadBodyV1,
@@ -313,6 +314,13 @@ export class Engine {
       middleware(callWithStackTrace(this.getPayloadBodiesByRangeV1.bind(this), this._rpcDebug), 2, [
         [validators.bytes8],
         [validators.bytes8],
+      ]),
+      () => this.connectionManager.updateStatus(),
+    )
+
+    this.getBlobsV1 = cmMiddleware(
+      middleware(callWithStackTrace(this.getBlobsV1.bind(this), this._rpcDebug), 1, [
+        [validators.array(validators.bytes32)],
       ]),
       () => this.connectionManager.updateStatus(),
     )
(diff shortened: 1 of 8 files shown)


packages/client/src/service/txpool.ts
```diff
@@ -379,6 +398,18 @@ export class TxPool {
     }
   }
 
+  pruneBlobsAndProofsCache() {
+    const pruneLength = this.blobsAndProofsByHash.size - BLOBS_AND_PROOFS_CACHE_LENGTH
+    let pruned = 0
+    for (const versionedHash of this.blobsAndProofsByHash.keys()) {
```
jochem-brouwer, line 404: `Map.keys()` returns in insertion order right? Can this get a comment that this deletes on insertion-order? This is not directly clear, it looks like it "randomly" deletes the order, but since `keys()` is sorted by insertion order this is not the case (it actually deletes based on first-seen).

packages/client/src/service/txpool.ts
```diff
@@ -34,6 +35,8 @@ const MIN_GAS_PRICE = BigInt(100000000) // .1 GWei
 const TX_MAX_DATA_SIZE = 128 * 1024 // 128KB
 const MAX_POOL_SIZE = 5000
 const MAX_TXS_PER_ACCOUNT = 100
+// keep for 1 epoch deep to handle reorgs
+const BLOBS_AND_PROOFS_CACHE_LENGTH = 6 * 32 * 1
```
jochem-brouwer, line 39: The comment seems to imply we cache blobs based upon one epoch, but this is not the case, we cache the txpool blob pool based on 6 (max blobs per block) times 32 (epoch length) transactions. This can thus include both pending or included transactions (in the future).

Also slight nit: maybe define (if we don't have any) 32 as epoch length, and pull the `6` from max blobs per block param?

jochem-brouwer (commented):
Strong support for the API since we need to bring bandwidth requirements of a running a node down :smile: 

Two comments.

packages/client/src/service/txpool.ts
g11tech: yes it deletes based on first seen

packages/client/src/service/txpool.ts
g11tech: updated on similar lines

packages/client/src/rpc/modules/engine/engine.ts
```diff
@@ -1513,4 +1521,13 @@ export class Engine {
     }
     return payloads
   }
+
+  private async getBlobsV1(params: [[Bytes32]]): Promise<(BlobAndProofV1 | null)[]> {
+    const blobsAndProof: (BlobAndProofV1 | null)[] = []
```
jochem-brouwer, line 1526: Should we put a check here which throws if `params.length` is too long?

packages/client/src/service/txpool.ts
```diff
@@ -379,6 +398,26 @@ export class TxPool {
     }
   }
 
+  pruneBlobsAndProofsCache() {
+    // currently common.param errors when not achieved cancun hardfork to do the below calculation
+    // const blobGasLimit = this.config.chainCommon.param('maxblobGasPerBlock')
+    // const blobGasPerBlob = this.config.chainCommon.param('blobGasPerBlob')
+    //
+    // A better way could be figured out to get allowedBlobsPerBlock
+    const allowedBlobsPerBlock = 6
```
jochem-brouwer, line 407: Why can't we pull this from common? It does not make sense to call this method pre-Cancun right?

packages/client/test/rpc/engine/getPayloadV3.spec.ts
```diff
@@ -110,6 +110,16 @@ describe(method, () => {
...
     await service.txPool.add(tx, true)
+
+    // check the blob and proof is available via getBlobsV1
+    res = await rpc.request('engine_getBlobsV1', [txVersionedHashes])
+    const blobsAndProofs = res.result
+    for (let i = 0; i < txVersionedHashes.length; i++) {
+      const { blob, proof } = blobsAndProofs[i]
+      assert.equal(blob, txBlobs[i])
+      assert.equal(proof, txProofs[i])
+    }
```
jochem-brouwer, line 121: I suggest that there should also be a test which tests the cache mechanism here.

jochem-brouwer:
The spec says:

```
The fields are encoded as follows:

- `blob`: `DATA` - `FIELD_ELEMENTS_PER_BLOB * BYTES_PER_FIELD_ELEMENT = 4096 * 32 = 131072` bytes (`DATA`) representing a SSZ-encoded `Blob` as defined in [EIP-4844](https://eips.ethereum.org/EIPS/eip-4844).
```

It does however not look like we SSZ-encode the blob? I think this is mandatory here? (per spec)

g11tech:
> The spec says:
> 
> ```
> The fields are encoded as follows:
> 
> - `blob`: `DATA` - `FIELD_ELEMENTS_PER_BLOB * BYTES_PER_FIELD_ELEMENT = 4096 * 32 = 131072` bytes (`DATA`) representing a SSZ-encoded `Blob` as defined in [EIP-4844](https://eips.ethereum.org/EIPS/eip-4844).
> ```
> 
> It does however not look like we SSZ-encode the blob? I think this is mandatory here? (per spec)

this the blobs generated/provided in tx are already in ssz encoded format which is just 131072 bytes, we don't have to do anything here

packages/client/test/rpc/engine/getPayloadV3.spec.ts
g11tech: cache pruning you mean?

packages/client/src/service/txpool.ts
g11tech: somehow its called from some tests, will need to debug

packages/client/src/rpc/modules/engine/engine.ts
g11tech: yea we could add

packages/client/test/rpc/engine/getPayloadV3.spec.ts
jochem-brouwer: Yes

jochem-brouwer:
Ok, I was not aware that the blobs were already SSZ-encoded (do not see this in EIP-4844), but if that's the case that's great, otherwise we implicitly have to take in the SSZ dependency (which we will have to do at some point when we switch to SSZ from RLP in general)

jochem-brouwer (approved):
LGTM!

PR #3716, opened 2024-10-02
vm: adjust test runner with verkle state management

gabrocheleau:
This PR improves the test runner so that a `stateManager` option can be passed in order to use a non-default (default being MerkleStateManager) state manager (for now, the only other option is `stateManager=verkle`, which will instantiate a StatefulVerkleStateManager to run the requested tests.


holgerd77:
What is the status of this?

acolytec3:
I'm still using it to try and run the verkle test fixtures but it's not ready for merging. We still have some blockers when running verkle at genesis that I don't have a solution for. Notably, if verkle is at genesis, we have to figure out how to hack in an execution witness or else relax the requirements for creating verkle blocks. I have ideas but none are really elegant or don't introduce tech debt we don't want

gabrocheleau:
Closing in favor of #3848

PR #3717, opened 2024-10-02
trie: rename trie to mpt

gabrocheleau:
This PR renames the `Trie` class to `MerklePatriciaTrie`. This aligns with the modularization of state management (i.e. MerkleStateManager & VerkleStateManager) and data structure agnosticism. 

I wasn't sure whether to name it "Tree" or "Trie" and decided to go with what was on the ethereum.org docs: https://ethereum.org/en/developers/docs/data-structures-and-encoding/patricia-merkle-trie/

### DIFF
--- a/packages/block/src/block/block.ts
+++ b/packages/block/src/block/block.ts
@@ -1,6 +1,6 @@
 import { ConsensusType } from '@ethereumjs/common'
 import { RLP } from '@ethereumjs/rlp'
-import { Trie } from '@ethereumjs/trie'
+import { MerklePatriciaTrie } from '@ethereumjs/trie'
 import { Blob4844Tx, Capability } from '@ethereumjs/tx'
 import {
   BIGINT_0,
@@ -226,7 +226,10 @@ export class Block {
    * Generates transaction trie for validation.
    */
   async genTxTrie(): Promise<Uint8Array> {
-    return genTransactionsTrieRoot(this.transactions, new Trie({ common: this.common }))
+    return genTransactionsTrieRoot(
+      this.transactions,
+      new MerklePatriciaTrie({ common: this.common }),
+    )
   }
 
   /**
@@ -471,7 +474,7 @@ export class Block {
     if (this.cache.withdrawalsTrieRoot === undefined) {
       this.cache.withdrawalsTrieRoot = await genWithdrawalsTrieRoot(
         this.withdrawals!,
-        new Trie({ common: this.common }),
+        new MerklePatriciaTrie({ common: this.common }),
       )
     }
     result = equalsBytes(this.cache.withdrawalsTrieRoot, this.header.withdrawalsRoot!)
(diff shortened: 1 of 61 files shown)


holgerd77:
So, to make the picture complete: the old pre-monorepo package was called [merkle-patricia-tree](https://github.com/ethereumjs/merkle-patricia-tree), so now we have everything in to confuse people to the max. 😂 

Anyhow: I just read up again a bit on [Trie vs Tree](https://www.google.com/search?q=trie+vs+tree) and came to the conclusion that trie is indeed the more correct term here. So that would be my tendency as well.

I thought about package renaming as well couple of weeks ago, and I think we should consequently do. If aligned with the class name that would be:

- `MerklePatriciaTrie`
- `@ethereumjs/merklepatriciatrie`

One thing where I am not totally out of sympathy is to at least give it some quick thought if `MPT` (and `@ethereumjs/mpt` might also be enough in this case and shorter and easier to write down, this is such a common naming in the Ethereum world that this might be enough as well.

That would be:

- `MPT`
- `@ethereumjs/mpt`

And last but not least. Maybe it's also ok to not fully align class + package name. I personally e.g. do not really like the long package name from above. So one option might also be:

- `MerklePatriciaTree`
- `@ethereumjs/mpt`

(I think this might be my favorite, but have not very strong preferences on all)

gabrocheleau:
> So, to make the picture complete: the old pre-monorepo package was called [merkle-patricia-tree](https://github.com/ethereumjs/merkle-patricia-tree), so now we have everything in to confuse people to the max. 😂
> 
> Anyhow: I just read up again a bit on [Trie vs Tree](https://www.google.com/search?q=trie+vs+tree) and came to the conclusion that trie is indeed the more correct term here. So that would be my tendency as well.
> 
> I thought about package renaming as well couple of weeks ago, and I think we should consequently do. If aligned with the class name that would be:
> 
> * `MerklePatriciaTrie`
> * `@ethereumjs/merklepatriciatrie`
> 
> One thing where I am not totally out of sympathy is to at least give it some quick thought if `MPT` (and `@ethereumjs/mpt` might also be enough in this case and shorter and easier to write down, this is such a common naming in the Ethereum world that this might be enough as well.
> 
> That would be:
> 
> * `MPT`
> * `@ethereumjs/mpt`
> 
> And last but not least. Maybe it's also ok to not fully align class + package name. I personally e.g. do not really like the long package name from above. So one option might also be:
> 
> * `MerklePatriciaTree`
> * `@ethereumjs/mpt`
> 
> (I think this might be my favorite, but have not very strong preferences on all)

I _think_ my favorite is also the latter (explicit MerklePatriciaTree for the class, mpt for the package). Generally not a big fan of acronyms as I find they increase barrier to entry, but in the context of a package name I think it's fine & more concise.

jochem-brouwer (approved):
LGTM

PR #3720, opened 2024-10-03
t8n: do not create empty coinbase account from SpuriousDragon on state.reward == 0

jochem-brouwer:
Fixes a bug in t8ntool filler where we create empty coinbase accounts


ScottyPoi (approved):
LGTM

PR #3726, opened 2024-10-07
mpt: expand getValueMap tests

jochem-brouwer:
Expands getValueMap tests by also testing startKey / limit params

### DIFF
--- a/packages/mpt/test/trie/trie.spec.ts
+++ b/packages/mpt/test/trie/trie.spec.ts
@@ -5,6 +5,7 @@ import {
   KECCAK256_RLP,
   MapDB,
   bigIntToBytes,
+  bytesToBigInt,
   bytesToHex,
   concatBytes,
   equalsBytes,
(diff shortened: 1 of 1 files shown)


ScottyPoi:
I believe I discovered a bug with `getValueMap`

I pushed a change to the test which breaks it to demonstrate this.  

`getValueMap` must assume that all keys are of equal length, because adding a key of a different length than the rest results in a failing test.

with both `0x01` and `0x010a` as keys, `getValueMap()` fails to return `0x01` in the map.

acolytec3:
That seems correct.  Any ideas on a generalized solution?  My initial instinct was to just pad everything to 32 bytes but that doesn't actually solve the issue internally.

ScottyPoi:
Looks like `getValueMap` only returns Leaf nodes.  In the above example, we miss the value stored at `0x01` because it is in a Branch Node

PR #3719, opened 2024-10-03
trie: trie package renaming to mpt

gabrocheleau:
This PR renames the trie package to mpt

PR #3730, opened 2024-10-08
mpt: adjust verifyMPTProof naming and restructure index file

gabrocheleau:
This addresses the comment here: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3719#issuecomment-2399517361


acolytec3 (approved):
LGTM

packages/mpt/src/proof/index.ts
```diff
@@ -1,174 +1 @@
...
-    return value
-  } catch (err: any) {
-    if (err.message === 'Missing node in DB') {
-      throw new Error('Invalid proof provided')
-    } else {
-      throw err
-    }
-  }
-}
+export * from './proof.js'
```
holgerd77, line 1: This should re-export the range proof functionality as well.

packages/mpt/src/proof/proof.ts
```diff
@@ -0,0 +1,174 @@
...
+  return verifyRangeProof(
+    rootHash,
+    firstKey && bytesToNibbles(firstKey),
+    lastKey && bytesToNibbles(lastKey),
+    keys.map((k) => k).map(bytesToNibbles),
+    values,
+    proof,
+    opts?.useKeyHashingFunction ?? keccak256,
+  )
+}
```
holgerd77, line 68: Ah, ok, just seeing that this is included here! 🤯 Then on-top request, can we please:

1. Move this fully over to `range.ts` and have *one* documented method here, likely kill off `verifyRangeProof`
2. And *then* re-export `verifyMerkleRangeProof` from `range.ts` from within `index.ts`

packages/statemanager/src/proofs/merkle.ts
```diff
@@ -240,7 +240,7 @@ export async function verifyMerkleStateProof(
     const storageProof = stProof.proof.map((value: PrefixedHexString) => hexToBytes(value))
     const storageValue = setLengthLeft(hexToBytes(stProof.value), 32)
     const storageKey = hexToBytes(stProof.key)
-    const proofValue = await verifyMPTProof(storageKey, storageProof, {
+    const proofValue = await verifyMerkleProof(storageKey, storageProof, {
       useKeyHashing: true,
     })
     const reportedValue = setLengthLeft(
```
holgerd77, line 246: Path related comment (so not for this file): can we align the proof directories from mpt and statemanager, and so (my suggestion) call this one here also:

- `statemanager/src/proofs` -> `statemanager/src/proof`

(we have a lot of stuff in singular - e.g. also `test` (instead of `tests`), so this feels somewhat more as our "convention" here)

holgerd77 (commented):
Great, thanks for tackling so quickly! 🤩 

Two somewhat additional things, can you maybe do in a follow-up PR if you agree with the updates?

packages/statemanager/src/proofs/merkle.ts
gabrocheleau: Looking at the paths from mpt, it seems like we can actually do without  `src/proof` and `src/db`. 

- Proof only exports two files
- DB only exports one

So I moved them over to the src directory (renamed range.ts to rangeProof.ts). This flattens the repo structure and has the benefit of limiting the amount of nested index files.

PR #3732, opened 2024-10-09
fix vite bundler visualizer

jochem-brouwer:
This PR fixes the vitest bundle visualizer which we used to see tree shaking effects.

In `./packages/evm`:

`npx vite build --config=./vite.config.bundler.ts --emptyOutDir=false --outDir .`

Previously, this ran into an error. 

image

Note: I have not much understanding about this whole vite bundler thing, and also don't know of any side effects :sweat_smile:

### DIFF
--- a/config/vitest.config.browser.mts
+++ b/config/vitest.config.browser.mts
@@ -19,6 +19,7 @@ const config = defineConfig({
   optimizeDeps: {
     
     exclude: ['kzg-wasm'],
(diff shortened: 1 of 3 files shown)


holgerd77:
Blocked since eventually not an issue (please do not merge for now)

jochem-brouwer:
Also converted to draft. If someone else runs into the same issue and can reproduce then we can unblock this. (Will add needs reproduce label)

jochem-brouwer:
Nvm, will close, can reopen if someone else runs into this problem

Issue #3723, opened 2024-10-07
Tx: Class names not fully consistent yet


holgerd77:
Following tx type class names still needs shortening from `Transaction` to `Tx` to be consistent with the other class names. 🙂 

- `AccessList2930Transaction`
- `EOACode7702Transaction`

Correct example: `FeeMarketEIP1559Transaction` -> `FeeMarket1559Tx`

jochem-brouwer:
I will address this among the refactor PR(s). (Might break up the refactor in multiple PRs)

jochem-brouwer:
Ok, interesting how we missed this one, because you are right, for EIP1559 and for EIP4844 the names are correct :thinking:

Issue #3694, opened 2024-09-22
Tx: High Cost of super() call in Constructors / Inheritance


holgerd77:
Related: #3484

Today I looked into tx instantiation (thinking from client perspective where we consecutively instantiate thousands of txs), and results were quite suprising:

<img width="485" alt="grafik" src="https://github.com/user-attachments/assets/a31f39b2-2fd3-405f-9e91-06b32983bfba">
<img width="726" alt="grafik" src="https://github.com/user-attachments/assets/4ed28705-0c04-469b-bfa3-9fb292cfddba">

So these are times measuring the costs of the steps along tx construction (in total  tx instantiation is quite noticeable, ~230ms for 10.000 txs, this comes together quickly). Notice the t1 step: this is for the `super()` call. I first thought: oh, something expensive in `super()`, but: no. The `s*` are actually the super measurements. This doesn't sum up. Turns out the costly thing is the `super()` call itself. I stumbled upon [this](https://netflixtechblog.com/improving-the-performance-of-our-javascript-inheritance-model-af376d75665) article on Netflix Tech Blog about general performance downsides of JS inheritance / super() calls.

I think this information is likely primarily relevant in the context of the tx refactor, I therefore assigned @jochem-brouwer to this issue, more to take notice than to directly act upon. This might be another strong argument to generally get rid of the `BaseTransaction` (@jochem-brouwer, I think you already mentioned something like this? I also thought about this couple of times).

jochem-brouwer:
Hi Holger, thanks for writing this down. I was not aware that `super` methods have such a performance impact!

The goal of the tx-refactor from my side actually has two goals which I want to tackle at once:

1. Remove redundant code
2. Remove `BaseTransaction`

Might look like two separate tasks, but I did some quick-and-dirty experimenting and I think these two work well together. I want to open a proof-of-concept PR this week to show the general idea, it mainly consists expanding the `Capabilities` idea which we already have in Tx to expand this to other redundant code, and to remove `BaseTransaction` along the road altogether!

holgerd77:
Sounds great! 🙂

Issue #3484, opened 2024-07-04
Tx: Some Tx Type Capabilities Extraction Refactor Continuation


holgerd77:
During the #3470 work there were some remaining tasks for extracting redundant functionality into capability methods identified here https://github.com/ethereumjs/ethereumjs-monorepo/pull/3470#issuecomment-2185956766.

These can be worked on and integrated, now that the 7702 work has been merged.

jochem-brouwer:
I will get started on this one!

holgerd77:
Good choice

PR #3733, opened 2024-10-09
TX refactor PR

jochem-brouwer:
Had to add a change to open a PR. I will use this PR to dump in any thoughts / finds upon the TX refactor and will cherry-pick in my commits. I will update this text and likely also post some comments here.

Addresses / fixes https://github.com/ethereumjs/ethereumjs-monorepo/issues/3723, https://github.com/ethereumjs/ethereumjs-monorepo/issues/3694, https://github.com/ethereumjs/ethereumjs-monorepo/issues/3484

The main goal of this is to refactor the Tx library.

Updating from an inheritance pattern to a composition pattern
=======

We have used the inheritance pattern by extending the `BaseTransaction` (this holds the "shared logic" of currently all txs which we support). This does seem to make sense since all transaction types on L1 currently share a lot of logic. However, when we take into account L2 txs:

Arbitrum: https://docs.arbitrum.io/how-arbitrum-works/arbos/geth#transaction-types
Optimism: https://specs.optimism.io/protocol/deposits.html#the-deposited-transaction-type

(two examples, obviously not all) 

These txs could for instance not have a signature, or not have a nonce. However, these still run the VM (and likely are also included in the block, but I did not check this).

We have created `Capabilities` into Tx before (thanks @gabrocheleau https://github.com/ethereumjs/ethereumjs-monorepo/pull/2993, https://github.com/ethereumjs/ethereumjs-monorepo/pull/3010). It seems to make much more sense (especially when keeping in mind support for the L2 txs) to move away from the inheritance pattern and instead use a composition pattern.

That is:

- Transaction objects now do not share the BaseTransaction class. Instead, these call into Capabilities which does this shared logic for them
- Integrating L2 txs should be made more simpler
- Identify what properties/methods are used inside our own libraries (VM, Block (+client)) which should be exposed. This is likely a super small set. For runtime behavior, this already differs on the tx type (legacy txs have no access lists, so no access lists logic there. If there is an access list, trigger the access list logic). For block, this is likely only serializing txs, plus some extra logic (likely) for blobs.
- This logic should be tx-dependent and therefore keep a Capabilities-like mechanism and let consumer libraries call into these to see if a Capability is supported by the tx, and then use this Capability to run specific capability-like logic.
- The composition pattern gets rid of `BaseTransaction` and will likely improve performance, see: https://github.com/ethereumjs/ethereumjs-monorepo/issues/3694


holgerd77:
That sounds all great (I have one small comment but I will go the dive-deeper later)! ⭐ 😄 

Some note: there is only one file (and only types) pushed for now, is this intended?

jochem-brouwer:
Yes, and I actually intended not to push that. I have 2 monorepo folders open to experiment a bit, and sometimes (if I switch branches for instance to test a PR) I stash/commit my experiment so I can continue later on. (The local branches here are super dirty :sweat: ). I accidentally pushed this branch, will force-push and remove it later on and then cherry-pick the relevant commits in to tidy things up :)

holgerd77:
> Side note: I think the earliest I can spin up a PR which would be in a kind-of re-viewable state would be Monday

Yeah, I mean this late pushing - especially on larger work - always bears the risk of stronger discussions and eventual change needs, so it also has some benefits to push earlier on, even if things are not fully cleaned up. But as you want. 🙂

jochem-brouwer:
Yes, good point. I have to note that I have started over from scratch many times (also recently) with new ideas (or finding out that the previous approach did not work). I will push once I have a clean PoC so we can start to get some eyes on there and leave some feedback and open the discussion :smile:

jochem-brouwer:
For phase 2 of tx refactor will push more of the composition pattern to the tx, together with Holgers document and some past ideas I had the following idea:

The 7702 transaction for instance is a composition of the logic of Access Lists (2930), Fee Market (1559) and EOA code deployment (7702). What I was thinking about is along the following lines: currently we have Capabilities logic per EIP, but I think we should swap this out to something more specific. For instance, a tx in the future could have parts of an EIP but also some parts not. So I would propose:

The capabilities get renamed: 2930 -> AccessLists, 1559 -> FeeMarket, 4844 -> Blobs, 7702 -> EOACode.

A `EOA7702Tx` is then a tx composed of: AccessLists, FeeMarket, EOACode. 

If we do this composition pattern, then for instance AccessLists capability will demand the `accessLists` property on the tx, plus it will also allow it in the TxData field. And FeeMarket will reject `gasPrice` and add `maxPriorityFeePerGas` and `maxFeePerGas` to the interface.

Besides this pattern, this new "features" will then also allow is (thinking a bit of L2 support): "Signable" (this allows the `v/r/s` fields) and we could go even further, "CreateContract" (this allows a `to: null` field, this is NOT allowed in 7702 / 4844).

This also allows us to create a SUPER MINIMAL tx interface with only the required fields. We now have some super deep logic hardcoded into the tx interface, for instance such as `toCreationAddress()` (method to figure out if we create a contract, this is also required currently on 7702/4844 but always false. 

Minimal interface would likely only be methods as: `raw()` (before serialization) `serialize()` (RLP serialize), `hash()`, `toJSON`.

jochem-brouwer:
image

Here is the code overview of master as of 287f96076fc52435a04cb70de8e92b45c8d33fb3 (to track if we blow up the package with a lot of lines or not)

PR #3734, opened 2024-10-09
Add Units class for conversions

scorbajio:
This change adds a `Units` class in the `util` package with two conversion functions in it:
```
  static ether(amount: number | bigint): bigint 
  static gwei(amount: number | bigint): bigint
```
Also see #1397, most importantly [this](https://github.com/ethereumjs/ethereumjs-monorepo/issues/1397#issuecomment-2401861664) comment for more details.

### DIFF
--- a/packages/util/src/units.ts
+++ b/packages/util/src/units.ts
@@ -18,3 +18,21 @@ export function formatBigDecimal(
   const zerosPostDecimal = String(maxDecimalFactor).length - 1 - String(fraction).length
   return `${full}.${'0'.repeat(zerosPostDecimal)}${fraction}`
 }
+
+export class Units {
+  static ether(amount: number | bigint): bigint {
+    if (typeof amount === 'number' && !Number.isInteger(amount)) {
+      throw new Error('Input must be an integer number')
+    }
+    const weiPerEther = BigInt(10 ** 18)
+    return BigInt(amount) * weiPerEther
+  }
+
+  static gwei(amount: number | bigint): bigint {
+    if (typeof amount === 'number' && !Number.isInteger(amount)) {
+      throw new Error('Input must be an integer number')
+    }
+    const weiPerGwei = BigInt(10 ** 9)
+    return BigInt(amount) * weiPerGwei
+  }
+}
--- a/packages/util/src/units.ts
+++ b/packages/util/src/units.ts
@@ -1,6 +1,9 @@
 import { BIGINT_0, BIGINT_1 } from './constants.js'
+
 /** Easy conversion from Gwei to wei */
 export const GWEI_TO_WEI = BigInt(1000000000)
+export const WEI_TO_EITHER = BigInt(10 ** 18)
+export const WIE_TO_GWEI = BigInt(10 ** 9)
 
 export function formatBigDecimal(
   numerator: bigint,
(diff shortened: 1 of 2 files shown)


packages/util/src/units.ts
```diff
@@ -18,3 +18,21 @@ export function formatBigDecimal(
   const zerosPostDecimal = String(maxDecimalFactor).length - 1 - String(fraction).length
   return `${full}.${'0'.repeat(zerosPostDecimal)}${fraction}`
 }
+
+export class Units {
+  static ether(amount: number | bigint): bigint {
+    if (typeof amount === 'number' && !Number.isInteger(amount)) {
+      throw new Error('Input must be an integer number')
+    }
+    const weiPerEther = BigInt(10 ** 18)
```
jochem-brouwer, line 27: This const should be moved outside of this method to avoid recomputing the const

packages/util/src/units.ts
```diff
@@ -18,3 +18,21 @@ export function formatBigDecimal(
...
+    return BigInt(amount) * weiPerEther
+  }
+
+  static gwei(amount: number | bigint): bigint {
+    if (typeof amount === 'number' && !Number.isInteger(amount)) {
+      throw new Error('Input must be an integer number')
+    }
+    const weiPerGwei = BigInt(10 ** 9)
```
jochem-brouwer, line 35: Same here :)

packages/util/src/units.ts
```diff
@@ -18,3 +18,21 @@ export function formatBigDecimal(
   const zerosPostDecimal = String(maxDecimalFactor).length - 1 - String(fraction).length
   return `${full}.${'0'.repeat(zerosPostDecimal)}${fraction}`
 }
+
+export class Units {
```
jochem-brouwer, line 22: Could you add docs to be explicit that the input amount is the amount of the unit, and that the returned value is in `wei`?

Also, I think a guard should be added to prevent negative numbers as input

jochem-brouwer (commented):
Some small comments!

holgerd77 (approved):
Cool, that was quick!

Looks great and super clean! 🤩

Will merge.

packages/util/src/units.ts
```diff
@@ -1,6 +1,9 @@
 import { BIGINT_0, BIGINT_1 } from './constants.js'
+
 /** Easy conversion from Gwei to wei */
 export const GWEI_TO_WEI = BigInt(1000000000)
+export const WEI_TO_EITHER = BigInt(10 ** 18)
```
jochem-brouwer, line 5: This is inconsistent, the above constant which was already there is the conversion rate from Gwei to Wei (the multiplier. So this should be `ETHER_TO_WEI` (there also is a typo here `EITHER` -> `ETHER`).

packages/util/src/units.ts
```diff
@@ -1,6 +1,9 @@
...
+export const WIE_TO_GWEI = BigInt(10 ** 9)
```
jochem-brouwer, line 6: This should be GWEI_TO_WEI, a constant which is declared above (although it is more clear to write that constant as `10 ** 9` to make it more readable). Also a typo: `WIE` -> `WEI`.

packages/util/test/units.spec.ts
```diff
@@ -20,3 +20,73 @@ describe('formatBigDecimal', function () {
...
+
+    it('should throw error when a non-integer number is used', () => {
+      assert.throws(function () {
+        Units.gwei(0.5)
+      }, 'Input must be an integer number')
+    })
+
+    it('should throw error when a negative number is used', () => {
+      assert.throws(() => {
+        Units.ether(-1)
```
jochem-brouwer, line 83: This tests negative numbers on `ether`, should be `gwei`.

jochem-brouwer (commented):
I got some comments, addressed those in #3736

PR #3735, opened 2024-10-09
Replace `eventEmitter` with a snail pretending to be a fast, modern event emitter

acolytec3:
Replaces our use of native NodeJS event emitters with the `emittery` package.  

All events are now async by default

Other notable changes
- Removed `AsyncEventEmitter` export from `util`
- Added types for `common`, `blockchain`, and `devp2p` events


packages/devp2p/src/types.ts
```diff
@@ -237,3 +242,57 @@ export interface RLPxOptions {
...
+  status:
+    | LES.Status
+    | {
+        chainId: Uint8Array | Uint8Array[]
+        td: Uint8Array
+        bestHash: Uint8Array
+        genesisHash: Uint8Array
+        forkId?: Uint8Array | Uint8Array[]
+      }
+}
```
holgerd77, line 275: We normally type/name everything in singular, can we do this here too?

Otherwise this all looks super great! 🎉👍🙂

acolytec3: What do you mean by type in singular?  For our current event typing, it's all `VMEvents` or `EVMEvents`.  I have no strong opinion but was trying to be consistent with our existing event typing.

holgerd77: Ah, ok, then I guess `VMEvents` also made no sense in the first place.

So, such a thing is clearly *one* event:

```ts
export interface EVMEvents {
  newContract: { data: NewContractEvent; resolve?: (result?: any) => void }
  beforeMessage: { data: Message; resolve?: (result?: any) => void }
  afterMessage: { data: EVMResult; resolve?: (result?: any) => void }
  step: { data: InterpreterStep; resolve?: (result?: any) => void }
}
```

And we also just type anything else in the monorepo with an anlogue structure with "singular", so these events clearly stand out.

Anyhow: don't want to bring you on a side track then, will open a small refactoring issue on this.

holgerd77: (or maybe you can directly adjust here for `VMEvents` and `EVMEvents`, at the end it's a pretty small update?)

packages/blockchain/src/blockchain.ts
```diff
@@ -346,7 +346,7 @@ export class Blockchain implements BlockchainInterface {
       await this.checkAndTransitionHardForkByNumber(canonicalHead, header.timestamp)
     })
     if (this._deletedBlocks.length > 0) {
-      this.events.emit('deletedCanonicalBlocks', this._deletedBlocks)
+      void this.events.emit('deletedCanonicalBlocks', () => this._deletedBlocks)
       for (const block of this._deletedBlocks)
```
acolytec3, line 350: Do we need this event at all?  I don't see it listened to anywhere in the code base. @g11tech, any ideas?

packages/devp2p/src/types.ts
acolytec3: Yep, it's a simple change and makes sense to do now.

packages/blockchain/src/blockchain.ts
acolytec3: Nevermind.  Found where this is used.

acolytec3:
Please don't close this yet.  I need to reference for my various cleanup PRs

acolytec3:
I've taken care of the mess.  We can wish this nonsense goodbye.

holgerd77:
🎉 😂 🙏

PR #3736, opened 2024-10-09
util: Units follow-up PR

jochem-brouwer:
Follow up on #3734 (see my review comments there, addressed those here)

### DIFF
--- a/packages/util/src/units.ts
+++ b/packages/util/src/units.ts
@@ -1,9 +1,8 @@
 import { BIGINT_0, BIGINT_1 } from './constants.js'
 
-/** Easy conversion from Gwei to wei */
-export const GWEI_TO_WEI = BigInt(1000000000)
-export const WEI_TO_EITHER = BigInt(10 ** 18)
-export const WIE_TO_GWEI = BigInt(10 ** 9)
+/** Conversion constants to wei */
+export const GWEI_TO_WEI = BigInt(10 ** 9) // Multiplier to convert from Gwei to Wei
+export const ETHER_TO_WEI = BigInt(10 ** 18) // Multiplier to convert from Ether to Wei
 
 export function formatBigDecimal(
   numerator: bigint,
@@ -40,7 +39,7 @@ export class Units {
    */
   static ether(amount: number | bigint): bigint {
     Units.validateInput(amount)
-    return BigInt(amount) * WEI_TO_EITHER
+    return BigInt(amount) * ETHER_TO_WEI
   }
 
   /**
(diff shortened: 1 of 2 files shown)


holgerd77:
Ah, sorry, I perceived them as addressed because the conversations were closed.

holgerd77 (approved):
LGTM

holgerd77:
I wondered a bit yesterday: should we also allow the level of floating point up to  what's possible without rounding (counting decimals) instead of directly throwing? Or are we ok ATM with this limitation?

jochem-brouwer:
I like the idea, but we have to be careful here, since the use case of these converters will likely have to do with "real life money". (and we dont want any mistakes where the conversion goes wrong).

There definitely should be a check to figure out if the number provided does not end up having "fractional wei", we indeed cannot have an output which reports "1.3 wei" for instance.

I am not knowledgeable enough about how to do this conversion without any loss, so I'm not sure how to do it. Just took a look, web3 does it here: 

https://github.com/web3/web3.js/blob/f70140693b14edb20891a4c8554f5fcc58dc6dd3/packages/web3-utils/src/converters.ts#L574

But as you can see it is not trivial (and there are also warnings about precision loss). I don't think it is worth it, if people really want this they should use an external library to do so.

jochem-brouwer:
Ah ok, a follow up since the web3 code is rather elegant to solve this problem (for context see https://github.com/web3/web3.js/blob/f70140693b14edb20891a4c8554f5fcc58dc6dd3/packages/web3-utils/src/converters.ts#L574):

```typescript
	// join the value removing `.` from
	// 24.56 -> 2456
	const value = BigInt(`${integer}${fraction}`);

	// multiply value with denomination
	// 2456 * 1000000 -> 2456000000
	const updatedValue = value * denomination;

	// check if whole number was passed in
	const decimals = fraction.length;
	if (decimals === 0) {
		return updatedValue.toString();
	}
```

So what we could do here (I am mainly worried about precision) is to allow string inputs. Via this method we can safely convert the string to Wei, because we can read the integer and the fraction both as BigInt (and throw is the fraction is too large if it does not fit in the requested unit).

However, I just want to note, I do not recall that while programming our library I ever wanted a helper to convert ethers to weis for instance (but this might also be because we work in a testing environment).  I also don't know about external users, there likely already is such converter which they use.

holgerd77:
Ok, that makes sense. Then let's at least not directly do and skip for now or at all. If we stumble upon this too often we can still decide in the future that we want to integrate.

holgerd77:
(and totally falls into my (very present here) keep-this-work-simple base goal 😄 (these old discussions are still in my head, phew, and for such a simple lets-do-live-a-bit-better thing))

holgerd77:
Thanks for the additional notes, yeah, again: let's keep it for now, and not go further into this we-could-do-this-or-that mode.

I personally DO have a use case for this though - just to note - I am not so firm with setting gas prices and the like and so I am a bit eager to get some help here and not "get lost in 0s". 🙂

PR #3738, opened 2024-10-11
Refactor client package to use Units class conversion helpers

scorbajio:
This change looks at places in the monorepo where the new `Units` class and it's static conversion methods can be used, most notably places where transaction constructors are being used where fees need to be input. This is a followup to #3734 and #1397.


holgerd77 (approved):
Wow, wow, wow, I can't say how much I love this! ❤️ 👍 

Actually didn't know before that it would be *that* satifying to see all these 0s replaced with a directly spottable `Units.gwei(1)`!!! (at least I am totally not able to "see" anything out of `1000000000n`.

Will merge. 🙂 

Small nit: let's please be a bit sparse with the word "refactor". I would like to keep/reserve this a bit for the PRs who really do somewhat significant changes on the API, the internal structure of code, etc.. So this one here is clearly *not* a refactor, this is just an application of a new API. Limiting this a bit helps me to re-identify the things to point out on release notes e.g..

PR #3740, opened 2024-10-12
mpt: simplify rangeProof handling & flatten mpt file structure

gabrocheleau:
This addresses [this comment](https://github.com/ethereumjs/ethereumjs-monorepo/pull/3730#discussion_r1793021137) and makes some misc minor improvemnts to the mpt package.


holgerd77 (commented):
For the record: I liked the non-flattened version a bit more, would not make this a blocker though.

What *is* important (necessary) though: have these files been moved by `git` ? (GitHub then normally shows the transition) Otherwise we would loose git history for these files. So if not, this would need an update.

gabrocheleau:
> Client tests not going through

fixed by re-running

packages/mpt/src/proof/range.ts
```diff
@@ -412,17 +415,23 @@ async function hasRightElement(trie: MerklePatriciaTrie, key: Nibbles): Promise<
...
+  firstKeyRaw: Uint8Array | null,
+  lastKeyRaw: Uint8Array | null,
+  keysRaw: Uint8Array[],
   values: Uint8Array[],
   proof: Uint8Array[] | null,
-  useKeyHashingFunction: HashKeysFunction,
+  useKeyHashingFunction: HashKeysFunction = keccak256,
 ): Promise<boolean> {
+  // Convert Uint8Array keys to nibbles
+  const firstKey = firstKeyRaw !== null ? bytesToNibbles(firstKeyRaw) : null
```
jochem-brouwer, line 431: Doesn't the change from Nibbles to Uint8Array decrease the accuracy here? Nibbles are limited from the numbers 0 -0x0f and the uint8array is limited (per entry) from 0x0 to 0xff. (Trying to understand this change here and if it is limiting)

jochem-brouwer: Ah, wait, this method was never exported on package-level in the first place (it was only used internally) so this is fine.

jochem-brouwer (approved):
LGTM

PR #3742, opened 2024-10-13
Add T9N (TransactionTool) test consumption 

jochem-brouwer:
This PR adds T9N test support, to decode tx rlps and then verify:

- the sender
- the hash
- the intrinsic gas (the minimal gas limit the tx has to pay, covering data fee / access lists / tx base fee (21k in most cases, higher when creating contracts))

Or:

- If the RLP is invalid, the tx creation aborts

Runs tests of `ethereum-tests/TransactionTests`.

This currently only allows to consume the tests, not to fill them.

See: https://github.com/ethereum/go-ethereum/tree/master/cmd/evm#transaction-tool

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17760,10 +17760,102 @@
         "@types/node-dir": "^0.0.34",
         "micro-eth-signer": "^0.11.0",
         "minimist": "^1.2.0",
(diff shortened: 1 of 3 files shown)


holgerd77 (approved):
Nice, LGTM! 🙂👍

PR #3743, opened 2024-10-16
Skip t9n in browser

acolytec3:
Adds an edit to the `tx` browser test config to skip `t9n` tests in browser and fix browser tests **again**


jochem-brouwer (approved):
Thanks! LGTM! :smile: :+1:

PR #3744, opened 2024-10-17
Tx: get rid of BaseTransaction

jochem-brouwer:
Part of #3733 

This PR is "phase 1" of the big tx refactor. This PR:

- Gets rid of the inheritance pattern of tx and switches to a composition pattern
- This means that we fully get rid of `src/basetransaction.js`

This PR does not intend to change any logic. Besides that, this PR:

- Aligns the tx interface, it changes `AccessList2930Transaction` to `AccessList2930Tx` and `EOACode7702Transaction` to `EOACode7702Tx`. This closes https://github.com/ethereumjs/ethereumjs-monorepo/issues/3723

The current state is still WIP, need to:

- [x] Better align shared methods (most of them are in Legacy capabilities, should likely move to some shared folder)
- [x] Fix tests
- [x] Fix upstream imports


packages/tx/src/features/util.ts
```diff
@@ -0,0 +1,138 @@
...
+  return common?.copy() ?? new Common({ chain: Mainnet })
+}
+
+/**
+ * Validates that an object with BigInt values cannot exceed the specified bit limit.
+ * @param values Object containing string keys and BigInt values
+ * @param bits Number of bits to check (64 or 256)
+ * @param cannotEqual Pass true if the number also cannot equal one less the maximum value
+ */
+export function valueBoundaryCheck( // TODO: better method name
```
gabrocheleau, line 27: For this and the other util methods:
In the context of composition rather than inheritance, I feel like we should be a bit more explicit with the naming (e.g. with reference to "sharedTxConstructor" rather than "shared constructor"). I'm not sure if this method (and the other method utils) are intended to be used outside of a transaction context, but if they are exported outside of the package, I think we have be particularly explicit with naming. This is actually one of the benefits of class-based methods imo (not that I don't prefer this pattern), is they are naturally namespaced by virtue of the class they belong to.

jochem-brouwer: These are not supposed to be exported outside of the package! These are only supposed to be used internally. These "shared" methods nevertheless have to be named differently. We also have the opportunity now (due to breaking releases) to give everything a more clear name!

packages/tx/src/features/util.ts
```diff
@@ -0,0 +1,138 @@
...
+export function sharedConstructor(
+  tx: Mutable<TransactionInterface>,
+  txData: TxData[TransactionType],
+  opts: TxOptions = {},
+) {
+  // LOAD base tx super({ ...txData, type: TransactionType.Legacy }, opts)
+  tx.common = getCommon(opts.common)
+  tx.common.updateParams(opts.params ?? paramsTx)
+
+  validateNotArray(txData) // is this necessary?
```
acolytec3, line 80: As I recall, this got added because there was some weird edge case where we were accepting `txData` where somehow an array got passed in.

acolytec3 (approved):
LGTM

PR #3745, opened 2024-10-17
Add tests for partial account helper functions

scorbajio:
This change adds unit tests testing a variety of cases for the partial account helper functions:
* `createPartialAccount`
* `createPartialAccountFromRLP`
* `serializeWithPartialInfo`


holgerd77 (commented):
Is this AI generated? If so can we please establish the convention (within the whole team) to always clearly mark/label this as such?

If so, we should clearly exchange on some best practices on how to review such code and what expectation we put on it.

If not: my apologies. 😂

packages/util/test/account.spec.ts
```diff
@@ -810,3 +815,342 @@ describe('Utility Functions', () => {
...
+    const account = createPartialAccount({
+      nonce: 1n,
+      balance: Units.ether(1),
+      storageRoot: KECCAK256_RLP,
+      codeHash: KECCAK256_RLP,
+      codeSize: 10,
+      version: 1,
+    })
+
+    assert.deepEqual(account, new Account(1n, Units.ether(1), KECCAK256_RLP, KECCAK256_RLP, 10, 1))
```
jochem-brouwer, line 845: For this test, could you arrange the used data in a `const` variable and read from there? (This makes the test more robust and clear)

packages/util/test/account.spec.ts
```diff
@@ -810,3 +815,342 @@ describe('Utility Functions', () => {
...
+      storageRoot: undefined,
+      codeHash: null,
+      codeSize: 10,
+      version: undefined,
+    })
+
+    assert.deepEqual(
+      account,
+      new Account(undefined, Units.ether(1), undefined, null, 10, undefined),
+    )
```
jochem-brouwer, line 861: The same for this test, use a `const` data variable to read from.

jochem-brouwer (commented):
The tests themselves look good! However, if you create a partial account from some data and then assert later it is the correct account (with the same, copied data) it is more clean to read these variables from data and then compare those (now the values are copied, so if you change one value you also have to change it on another place. If it's in a const var, only one place needs to be edited)

gabrocheleau (changes requested):
Left some comments

packages/util/test/account.spec.ts
scorbajio: Fixed this.

packages/util/test/account.spec.ts
scorbajio: Fixed

packages/util/test/account.spec.ts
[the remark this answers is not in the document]
scorbajio: I've pulled out the field data into a const variable and am using that to both create the account and check for equality.

packages/util/test/account.spec.ts
```diff
@@ -810,3 +815,342 @@ describe('Utility Functions', () => {
...
+    assert.throws(() => createPartialAccountFromRLP(serialized), /Invalid partial balance encoding/)
+  })
+
+  it('should throw an error for invalid storageRoot encoding (non-array)', () => {
+    const serialized = RLP.encode([[], [], KECCAK256_RLP, [], [], []])
+    assert.throws(
+      () => createPartialAccountFromRLP(serialized),
+      /Invalid partial storageRoot encoding/,
+    )
+  })
```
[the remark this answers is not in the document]
scorbajio: I've switched over to keeping the tests in an array that gets run as a part of a loop. It's not less lines of code, but it is more readable since the data being serialized is very similar. If I had disposed of the test description strings, it would have been much DRYer, but I think the descriptions are useful, so I kept them.

jochem-brouwer (approved):
LGTM!

gabrocheleau (approved):
Thanks for addressing all the comments! lgtm

PR #3746, opened 2024-10-17
Switch events to eventemitter3 - the saga

acolytec3:
After [failed](https://github.com/ethereumjs/ethereumjs-monorepo/pull/2303) [attempt](https://github.com/ethereumjs/ethereumjs-monorepo/pull/2376) [after](https://github.com/ethereumjs/ethereumjs-monorepo/pull/3330) [attempt](https://github.com/ethereumjs/ethereumjs-monorepo/pull/3735), I'm determined to get this right.

`eventemitter3` is the right choice for us on a performance basis since it's 50% faster than native Nodejs events.

~~Now, I just need to take on the final boss of async events with `eventemitter3`.~~

Final boss has been defeated with the help of the robots.


packages/client/src/sync/fullsync.ts
```diff
@@ -70,8 +70,8 @@ export class FullSynchronizer extends Synchronizer {
     const syncEvent: Promise<boolean> = new Promise((resolve) => {
       // This event listener listens for other instances of the fetcher that might be syncing from a different peer
       // and reach the head of the chain before the current fetcher.
-      this.config.events.once(Event.SYNC_SYNCHRONIZED, (height?: number) => {
-        this.resolveSync(height)
+      this.config.events.once(Event.SYNC_SYNCHRONIZED, (chainHeight?: bigint) => {
+        this.resolveSync(Number(chainHeight))
```
jochem-brouwer, line 74: We could also propagate the bigint here?

packages/client/src/types.ts
```diff
@@ -56,33 +54,6 @@ export interface EventParams {
   [Event.PROTOCOL_MESSAGE]: [messageDetails: any, protocolName: string, sendingPeer: Peer]
 }
```
jochem-brouwer, line 55: How does the deletion of this event bus change the typing of the client types?

packages/client/test/net/protocol/rlpxsender.spec.ts
```diff
@@ -52,14 +52,3 @@ describe('should receive message', async () => {
...
 })
-
-describe('should catch errors', async () => {
-  const rlpxProtocol = { events: new EventEmitter() }
-  const sender = new RlpxSender(rlpxProtocol as Devp2pETH)
-  it('throws sendStatus error', () => {
-    assert.throws(() => sender.sendStatus({ id: 5 }), /not a function/, 'sendStatus error')
-  })
-  it('throws sendMessage error', () => {
-    assert.throws(() => sender.sendMessage(1, 5), /not a function/, 'sendMessage error')
```
jochem-brouwer, line 63: These events are deleted and do not seem related (?) to events, is this intended?

packages/client/test/net/protocol/rlpxsender.spec.ts
```diff
@@ -43,7 +43,7 @@ describe('should receive status', async () => {
 
 describe('should receive message', async () => {
   const rlpxProtocol = { events: new EventEmitter() }
-  const sender = new RlpxSender(rlpxProtocol as Devp2pETH)
+  const sender = new RlpxSender(rlpxProtocol as never as Devp2pETH)
```
jochem-brouwer, line 46: Why do we have to cast it as `never` first now?

packages/evm/README.md
```diff
@@ -79,9 +79,10 @@ const main = async () => {
   // Note that numbers added are hex values, so '20' would be '32' as decimal e.g.
   const code = [PUSH1, '03', PUSH1, '05', ADD, STOP]
 
-  evm.events.on('step', function (data) {
+  evm.events.on('step', function (data, resolve) {
     // Note that data.stack is not immutable, i.e. it is a reference to the vm's internal stack object
     console.log(`Opcode: ${data.opcode.name}\tStack: ${data.stack}`)
+    resolve?.()
```
jochem-brouwer, line 85: Isn't the new function sig that `resolve` is always passed? Such that we can just do `resolve()`?

packages/evm/src/evm.ts
```diff
@@ -221,7 +221,16 @@ export class EVM implements EVMInterface {
...
     this._emit = async (topic: string, data: any): Promise<void> => {
-      return new Promise((resolve) => this.events.emit(topic as keyof EVMEvents, data, resolve))
+      const listeners = this.events.listeners(topic as keyof EVMEvent)
+      for (const listener of listeners) {
+        if (listener.length === 2) {
+          await new Promise<void>((resolve) => {
+            listener(data, resolve)
+          })
+        } else {
+          listener(data)
```
jochem-brouwer, line 231: This is the async logic handler now, right? This is not part of `eventemitter3`?

packages/evm/test/eips/eip-5656.spec.ts
```diff
@@ -91,11 +91,13 @@ describe('should test mcopy', () => {
...
       let currentMem = ''
 
-      evm.events.on('step', (e) => {
+      evm.events.on('step', (e, resolve) => {
         if (e.opcode.name === 'STOP') {
           currentMem = bytesToHex(e.memory)
           assert.equal(currentMem, '0x' + situation.post, 'post-memory correct')
         }
+        // we need to call resolve() to avoid the event listener hanging
+        resolve?.()
```
jochem-brouwer, line 100: But we can also remove the `resolve` here, right? (So no change in this file)?

packages/evm/test/runCall.spec.ts
```diff
@@ -627,11 +627,12 @@ describe('RunCall tests', () => {
 
     let verifyMemoryExpanded = false
 
-    evm.events.on('step', (e) => {
+    evm.events.on('step', (e, resolve) => {
       assert.ok(e.memory.length <= 96)
       if (e.memory.length > 0) {
         verifyMemoryExpanded = true
       }
+      resolve?.()
```
jochem-brouwer, line 635: Same here, remove the resolve from the listener and the call?

packages/vm/README.md
```diff
@@ -380,6 +379,22 @@ You can subscribe to the following events:
...
 
+Note, if subscribing to events with an async listener, specify the second parameter of your listener as a `resolve` function that must be called once your listener code has finished.
+
+```ts
+// ./examples/eventListener.ts#L10-L18
+
+// Setup an event listener on the `afterTx` event
+vm.events.on('afterTx', (event, resolve) => {
+  console.log('asynchronous listener to afterTx', bytesToHex(event.transaction.hash()))
+  resolve?.()
```
jochem-brouwer, line 390: An explicit note that `resolve` must be called would be nice here :)

jochem-brouwer (commented):
Looks clean and good, however, I see that for VM and EVM there are now changes where it seems that even on sync events `resolve()` is added and called (in the tests). These should be removed.

Also, the "breaking change" here is that blockchain events do now not support async events. Not sure if we want to put that back in, but if anyone is listening async on `deletedCanonicalBlocks` we definitely should.

I will also test this ASAP :)

packages/evm/README.md
acolytec3: yes

packages/client/test/net/protocol/rlpxsender.spec.ts
acolytec3: Right, they aren't directly related but they're pointless tests and don't actually test anything other than that when you try to access a non-existent property on an object that Javascript throws.  I just stumbled across it because it referenced `EventEmitter`.

packages/evm/src/evm.ts
acolytec3: Correct, if the listener has 2 arguments, we assume it's looking for the event object and then the resolve function (and therefore is async).  Otherwise, we just call the listener.

packages/client/src/sync/fullsync.ts
acolytec3: Will make this change.  Not really sure why we even propagate this other than for debug logging.

packages/client/src/types.ts
acolytec3: `eventEmitter3` doesn't need this additional layer of type expansions and we can just use the basic `EventParams` interface keyed by the properties of the `Event` enum.

packages/client/test/net/protocol/rlpxsender.spec.ts
acolytec3: I've reversed this edit.  For some reason, Typescript wasn't paying nicely with it before.

scorbajio (approved):
LGTM

packages/client/src/sync/sync.ts
```diff
@@ -145,9 +145,9 @@ export abstract class Synchronizer {
 
   abstract syncWithPeer(peer?: Peer): Promise<boolean>
 
-  resolveSync(height?: number) {
+  resolveSync(height?: bigint) {
     this.clearFetcher()
-    const heightStr = typeof height === 'number' && height !== 0 ? ` height=${height}` : ''
+    const heightStr = typeof height === 'bigint' && height !== BIGINT_0 ? ` height=${height}` : ''
```
holgerd77, line 150: Do these things even solve any bugs?? Or did you just also change the type coming in?

PR #3376, opened 2024-04-26
 Toubleshoot codecov failures to upload coverage data

scorbajio:
This PR looks into issues that codecov is having with uploading coverage info. Currently, only the tx package is reporting coverage in codecov.

This PR was originally at #3375 before it was accidentally closed by resetting and force pushing the head commit.

PR #3751, opened 2024-10-17
Further troubleshoot current codecov CI issues

scorbajio:
This change is a followup on #3376 and looks into remaining codecov issues:
- [ ] The master branch isn't having coverage pushed/calculated/visualized, so I'll see if this is just a misconfig or something
- [ ] Some packages (devp2p, wallet, genesis, client) aren't having their coverage files uploaded? Or, atleast they're not showing up on codecov at the moment, so I'll triage and followup here.
- [ ] As mentions by @jochem-brouwer, most of the coverage diffs seem to not be calculates/reported (https://github.com/ethereumjs/ethereumjs-monorepo/pull/3376#issuecomment-2406199634).
- [ ] `rlp` and `ethash` packages are uploading coverage reports, but their reports are erroring out as unusable. I remember working on this before and opening a codecov ticket on it too (in the codecov official repo) that didn't get much activity, but will look into and followup on this on as well.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,6 +2,9 @@ name: Build
 on:
   pull_request:
     types: [opened, reopened, synchronize]
+  push:
+    branches: [master]
+    tags: ['*']
   workflow_dispatch:
 
 env:
--- a/.github/workflows/devp2p-build.yml
+++ b/.github/workflows/devp2p-build.yml
@@ -52,7 +52,7 @@ jobs:
       - uses: codecov/codecov-action@v4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          files: ${{ github.workspace }}/packages/devp2p/coverage/lcov.info
+          files: ${{ env.cwd }}/coverage/lcov.info
           flags: devp2p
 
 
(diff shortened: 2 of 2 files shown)


scorbajio:
I just checked all of the packages (devp2p, wallet, genesis, client) that are showing no coverage in codecov, and it seems like their lcov files are being uploaded, but since their coverage difference is being calculated from the lcov files uploaded from the last commit on `master` that had any uploads at all, the diffs aren't going to be calculated as expected. It may help to resolve the issue that's keeping master from having it's coverage updated in codecov to see if this fixes things here too.

scorbajio:
In order to update codecov coverage, we'd need to merge this in order for coverage to be calculated and uploaded for the `master` branch.

acolytec3 (approved):

holgerd77:
🙏 

Cool! Can you maybe extract the TODO list from above in a separate issue, at least if you plan to do this several iterations (likely, right?). Otherwise this will create this weird referencing of this sub-PR, that will likely get confusing at some point.

But - yeah - totally like respectively feels good that we get a coverage re-integration! 😆 🤩

PR #3752, opened 2024-10-17
genesis: remove mpt devdep

gabrocheleau:
This PR removes the mpt dev dependency from the genesis package.json. It was only used for one check in a test, and it leads to a circular dependency that was spotted by @holgerd77 . If we want to do that root check, we can always port some tests over to another package, but I suspect that this sort of very common stateRoot check is covered by many other tests already.


acolytec3 (approved):
LGTM

holgerd77:
Thanks! 🙏

Issue #3750, opened 2024-10-17
To-do list from #3735


acolytec3:
I tackled a number of additional items in this PR that I'd like to add to the master branch though this PR is no longer viable.

- [x] - Add types for `common` and `devp2p` events
- [x] - Rewrite devp2p `eth-simulator` tests to make sense
~~- [ ] - Remove LES tests from both devp2p and client~~
- [x] - Rewrite `test/integration/miner.spec.ts` in `client` to use `inlineClient` (and related changes) will be closed by #3757

acolytec3:
Missing accomplished

PR #3757, opened 2024-10-21
Use `inlineclient` in miner integration tests

acolytec3:
Cleanup item from #3750

This replaces the current `miner` integration test with one that uses the inline client setup from the `test/sim/simutils` instead of complicated mocks.


g11tech (approved):
lgtm

holgerd77:
Interesting. 🤔 👍

PR #3759, opened 2024-10-21
Remove LES from `client` and `devp2p`

acolytec3:
This removes all LES code and tests from the monorepo

- [x] Remove LES code from `client`
- [x] Remove LES code from `devp2p`
~~- [ ] Collapse `FullEthereumService` into `Service`~~


packages/client/src/sync/fetcher/headerfetcher.ts
jochem-brouwer, line 1: I am confused, why is this file deleted? :thinking:

jochem-brouwer: I think `BlockFetcher` might also auto-download the headers? (Or likely first download the headers and then the bodies?). I nevertheless think that there are (a lot?) of use cases where we dont want the block bodies, just the headers?

jochem-brouwer (commented):
Super clean PR, some small nits. The "big" thing would be the removal of `HeaderFetcher`, not sure if we should do this (might also miss some context here).

packages/client/src/rpc/modules/admin.ts
```diff
@@ -23,7 +23,7 @@ export class Admin {
    * @param client Client to which the module binds
    */
   constructor(client: EthereumClient, rpcDebug: boolean) {
-    const service = client.services.find((s) => s.name === 'eth') as Service
+    const service = client.service as FullEthereumService
```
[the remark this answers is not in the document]
acolytec3: Correct, since we only have one service type.

packages/client/src/sync/fetcher/headerfetcher.ts
acolytec3: Hmm, interesting question.  I thought the header fetcher was explicitly and only used by the client when running the LES protocol and syncing, since that's basically what the old style light clients did.  The header fetcher as currently written relies on this `flowControl` thing which was again an LES specific function.  If we want a header fetcher that isn't dependent on LES stuff, we should rewrite this and then find a proper use case for it.  I know Geth used to sync just headers first as part of its traditional sync process but we haven't adapted that to ours.  We just get a block of headers and then get the corresponding bodies, assemble, and return to the synchronizer.

jochem-brouwer (approved):
LGTM.

If we want a header fetcher in the future, we should indeed (re)write it.

Issue #3755, opened 2024-10-19
`rlpx` simulator tests are not running correctly


acolytec3:
In the midst of other work, I discovered that the `rlpx-simulator` tests in the `devp2p/test/integration` folder are all currently run as synchronous tests even though they are using events and async code.  As such, the tests aren't currently doing anything since the test exits before the async parts of the test execute.  To resolve this, we need to convert all of these tests to something of the form:
```ts
it('should do some async RLPx thing', async () => {
    const basePort = 60661
    const rlpxs = util.getTestRLPXs(3, 1, basePort)
    const peer = { address: util.localhost, udpPort: basePort + 1, tcpPort: basePort + 1 }
    await new Promise((resolve) => {
        rlpxs[0].events.on('peer:added', () => {
            assert.ok('something happened correctly')
            util.destroy(rlpxs)
            resolve(undefined)
        })
    })
    await rlpxs[0].addPeer(peer)
})
```

I did something similar for the `eth-simulator` test so this can be used as a guide for further ideas.  I can't vouch for the validity of any of these tests as I haven't looked deeply into them so be prepared for this to be tricky.

Issue #3754, opened 2024-10-18
Codecov troubleshooting tracker


scorbajio:
- [ ] The master branch isn't having coverage pushed/calculated/visualized, so I'll see if this is just a misconfig or something
- [ ] Some packages (devp2p, wallet, genesis, client) aren't having their coverage files uploaded? Or, atleast they're not showing up on codecov at the moment, so I'll triage and followup here.
- [ ] As mentions by @jochem-brouwer, most of the coverage diffs seem to not be calculates/reported (https://github.com/ethereumjs/ethereumjs-monorepo/pull/3376#issuecomment-2406199634).
- [ ] `rlp` and `ethash` packages are uploading coverage reports, but their reports are erroring out as unusable. I remember working on this before and opening a codecov ticket on it too (in the codecov official repo) that didn't get much activity, but will look into and followup on this on as well.

PR #3760, opened 2024-10-21
Rework rlpx-simulator.spec.ts tests / Fix devp2p/genesis/wallet Coverage Reporting 

scorbajio:
This change looks into the `rlpx-simulator.spec.ts` tests in the `devp2p` package and attempts to understand and fix the test cases and if any expected bugs are found, they will be discussed here and fixed in followup PRs. Also see #3755.


packages/devp2p/test/integration/rlpx-simulator.spec.ts
```diff
@@ -6,94 +6,100 @@ import { DISCONNECT_REASON } from '../../src/types.js'
...
-  it('RLPX: remove node', () => {
-    const rlpxs = util.initTwoPeerRLPXSetup(undefined, undefined, undefined, 40504)
-
-    try {
-      rlpxs[0].events.once('peer:added', (peer) => {
-        rlpxs[0].disconnect(peer['_remoteId'])
+  it('RLPX: remove node', async () => {
+    const { rlpxs, peer } = util.initTwoPeerRLPXSetup(undefined, undefined, undefined, 40504)
+    rlpxs[0]
+      ['_dpt']!.addPeer(peer)
```
scorbajio, line 61: The reason I moved the `addPeer` call out of `initTwoPeerRLPXSetup` is to be able to chain the disconnect here before checking for `peer:removed` events in the awaited promise that follows.

packages/devp2p/test/integration/rlpx-simulator.spec.ts
```diff
@@ -6,94 +6,100 @@ import { DISCONNECT_REASON } from '../../src/types.js'
...
-      assert.equal(
-        peer['_port'],
-        basePort + 1,
-        'should have added peer on peer:added after successful handshake',
-      )
-      assert.equal(rlpxs[0].getPeers().length, 1, 'peer list length should be 1')
-      assert.equal(rlpxs[0]._getOpenSlots(), 9, 'should have maxPeers - 1 open slots left')
-      await util.delay(500)
-      util.destroyRLPXs(rlpxs)
+    const { rlpxs, peer } = util.initTwoPeerRLPXSetup(undefined, undefined, undefined, basePort + 1)
```
scorbajio, line 11: I believe `basePort + 1` is correct here since the following assert is checking if the peer's port is connecting as expected.

packages/devp2p/test/integration/rlpx-simulator.spec.ts
```diff
@@ -6,94 +6,100 @@ import { DISCONNECT_REASON } from '../../src/types.js'
...
+      .then((peer1) => {
+        rlpxs[0].disconnect(peer1['id'])
       })
-      rlpxs[0].events.once('peer:removed', async (_, reason: any) => {
+      .catch((e) => {
+        throw new Error(`Peering failed: ${e}: ${e.stack}`)
+      })
+    await new Promise((resolve) => {
+      rlpxs[0].events.once('peer:removed', (_, reason: any) => {
         assert.equal(
```
scorbajio, line 70: This assert is failing, expecting a `CLIENT_QUITTING` but getting a `TIMEOUT`.

holgerd77:
Can you add a readiness label (+ drop a comment on that)?

holgerd77:
Could you give a summary of what you finally did here and how things played out?

packages/devp2p/test/integration/rlpx-simulator.spec.ts
```diff
@@ -6,76 +6,63 @@ import { DISCONNECT_REASON } from '../../src/types.js'
...
+    })
   })
-
-  it('RLPX: test peer queue / refill connections', () => {
+  it('RLPX: test peer queue / refill connections', async () => {
     const basePort = 60661
     const rlpxs = util.getTestRLPXs(3, 1, basePort)
     const peer = { address: util.localhost, udpPort: basePort + 1, tcpPort: basePort + 1 }
     rlpxs[0]['_dpt']!.addPeer(peer)
-    try {
```
scorbajio, line 77: I reworked this test case to be a lot more similar to how the test was originally written to preserve the test coverage it gives for queuing/refilling connections while also updating it to run asynchronously.

scorbajio (commented):
I left some comments detailing what I did with the last few commits to get the tests running as expected. Other than that, I also fixed some typing errors.

holgerd77:
Thanks for the insights! 🙏 And on the more general level, so this re-activates test running in general, is this correct?

Can we "prove" this is working by coverage already? If not we should get the coverage running again as a first step maybe and then test here with the fix, at least that's what I would suggest if I get everything "right" here and over on the coverage front. 🙂

scorbajio:
> Thanks for the insights! 🙏 And on the more general level, so this re-activates test running in general, is this correct?
> 
> Can we "prove" this is working by coverage already? If not we should get the coverage running again as a first step maybe and then test here with the fix, at least that's what I would suggest if I get everything "right" here and over on the coverage front. 🙂

Yes, this should activate tests that were previously not running fully to show lines covered increase in codecov relative to the base commit. `devp2p` is however one of a number of our packages that has switched over fully to using vitest instead of tape, which results in lines covered not showing up in the uploaded lcov file generated with coverage runs. I'll prioritize getting coverage results for the `devp2p` package as a part of #3754 and come back to test the theory with these tests being fixed and reintroduced 🙂

scorbajio:
I switch `devp2p`, `genesis`, and `wallet` over to using vitest's v8 provider as their coverage provider, it seems to be doing the trick. The [new coverage report for rlpx.ts](https://app.codecov.io/github/ethereumjs/ethereumjs-monorepo/pull/3760/blob/packages/devp2p/src/rlpx/rlpx.ts) can be seen to show lines covered by the fixed tests, than the original (I locally calculated the coverage for the file on master as well to confirm this). This just leaves `client` as the last major package to not be showing any lines covered, likely due to issues with vitest, and a similar issue we have is two packages are uploading coverage files that are erroring as "unreadable," which I wonder if that is also related to changing our testing framework to vitest. You can also see [here](https://app.codecov.io/github/ethereumjs/ethereumjs-monorepo/pull/3760/flags) that the devp2p flag is showing suddenly a large percentage increase in coverage, as further confirmation to switching over to the vitest coverage provider is a working fix for now.

holgerd77 (approved):
Nice! 🙏🎉

Thanks for all the additional info!

PR #3761, opened 2024-10-22
Update verkle docs

acolytec3:
Updates the `verkle` package `README.md` with basic usage guidelines reflecting the current state of the package.

Also, adds a couple of simple example scripts.


packages/verkle/README.md
```diff
@@ -23,32 +23,79 @@ npm install @ethereumjs/verkle
 
 ## Usage
 
-### Initialization and Basic Usage
+### Initialization
+
+To initialize a verkle tree, we provide an async constructor `createVerkleTree` which returns a `VerkleTree` instance that is properly initialized with the required`VerkleCrypto` package that implements the necessary cryptographic primitives used by Verkle trees.
```
jochem-brouwer, line 28:
```suggestion
To initialize a verkle tree, we provide an async constructor `createVerkleTree` which returns a `VerkleTree` instance that is properly initialized with the required `VerkleCrypto` package that implements the necessary cryptographic primitives used by Verkle trees.
```

packages/verkle/README.md
```diff
@@ -23,32 +23,79 @@ npm install @ethereumjs/verkle
...
+
+import { MapDB, bytesToHex } from '@ethereumjs/util'
 import { VerkleTree } from '@ethereumjs/verkle'
-import { bytesToUtf8, utf8ToBytes } from '@ethereumjs/util'
+import { loadVerkleCrypto } from 'verkle-cryptography-wasm'
+
+const verkleCrypto = await loadVerkleCrypto()
+
+const main = async () => {
+  const tree = new VerkleTree({ verkleCrypto, db: new MapDB<Uint8Array, Uint8Array>() })
```
jochem-brouwer, line 50: Q: can't we default to use a MapDB if no db is provided? (obviously not part of this review)

packages/verkle/README.md
```diff
@@ -23,32 +23,79 @@ npm install @ethereumjs/verkle
...
+  await tree['_createRootNode']()
```
jochem-brouwer, line 51: Why are we calling a "private" function here which we should not call externally?

packages/verkle/README.md
```diff
@@ -23,32 +23,79 @@ npm install @ethereumjs/verkle
...
-const tree = new VerkleTree()
+### Getting and Putting Values
+
+Values are stored using a combination of a `stem` obtained through the `getVerkleStem` function exposed by `@ethereumjs/util`.
+
+Following the design goal of verkle trees of allowing efficient reads and writes of multiple values that are "close" to each other, the `get` and `put` methods take a stem as a first argument and then an array of "suffixes" (the 32nd byte) of the key used to access a value.
+
+#### Getting values
+
+When retrieving values given a stem `0x781f1e4238f9de8b4d0ede9932f5a4d08f15dae7` and suffixes `[0, 1]`, the `get` method would access the values stored at `0x781f1e4238f9de8b4d0ede9932f5a4d08f15dae70000` and `0x781f1e4238f9de8b4d0ede9932f5a4d08f15dae70001`.
```
jochem-brouwer, line 66: I might be missing something here, but it concats `0000` or `0001` to a 20-byte hex string? (Not something like 31 or 32 bytes which seems to be  expected here?)

jochem-brouwer: Ah ok, it seems that this is an address. Then the docs should be updated that the suffix is thus added to the stem (concatenated). A stem thus seems to be able to have variable sizes (?)

jochem-brouwer (commented):
Some comments

packages/verkle/README.md
acolytec3: Good call, there are a few things in the initialization process we need to smooth out.  Will add this to a next PR.

packages/verkle/README.md
acolytec3: Leftover from when we had a static constructor.  We can make this public but it's something that has to be performed outside of the actual constructor since it's async.  Will add to the cleanup PR.

packages/verkle/README.md
acolytec3: Oops, I must have copied the address instead of the stem I generated for it here.  Let me generate a valid stem and will update this.

packages/verkle/README.md
```diff
@@ -23,32 +23,79 @@ npm install @ethereumjs/verkle
...
+import { createVerkleTree } from '@ethereumjs/verkle'
 
 async function test() {
-  await tree.put(utf8ToBytes('test'), utf8ToBytes('one'))
-  const value = await tree.get(utf8ToBytes('test'))
-  console.log(value ? bytesToUtf8(value) : 'not found') // 'one'
+  const addrHex = '0x781f1e4238f9de8b4d0ede9932f5a4d08f15dae7'
+  const address = createAddressFromString(addrHex)
+  const tree = await createVerkleTree()
+  const stem = getVerkleStem(tree['verkleCrypto'], address)
```
gabrocheleau, line 84:
```suggestion
  const stem = getVerkleStem(tree['verkleCrypto'], address) // we retrieve the verkle stem associated with the address
```

packages/verkle/README.md
```diff
@@ -23,32 +23,79 @@ npm install @ethereumjs/verkle
...
+  console.log(bytesToHex(tree.root())) // 0x0000000000000000000000000000000000000000000000000000000000000000
+}
+
+void main()
+```
 
-const tree = new VerkleTree()
+### Getting and Putting Values
+
+Values are stored using a combination of a `stem` obtained through the `getVerkleStem` function exposed by `@ethereumjs/util`.
```
gabrocheleau, line 60:
```suggestion
Values are stored using a combination of a `stem` obtained through the `getVerkleStem` function exposed by `@ethereumjs/util`. In the context of Ethereum, to retrieve the data associated with an account at a particular address, we would first compute the verkle stem of that address (`getVerkleStem(verkleCrypto, address)`), and then get the particular pieces of data we're interested in by suffixing the stem with the suffixes corresponding to that data.
```
Perhaps this extra explanation could be useful?

holgerd77 (approved):
Ok, all addressed I guess, nice, LGTM!

PR #3768, opened 2024-10-24
verkle: improvements to verkle tree instantiation

### LINKED ISSUE #3763: Make Verkle instantiation easier
We have a few rough edges in how we initialize `verkle` trees at the moment. As long as you use the `createVerkleTree` constructor function, it's pretty straight forward.  However, we should make a few changes:

- Have the constructor default to instantiate a `MapDB` if none is provided
- The `_createRootNode` function should be made public.
- We should just get rid of the `database` function all together.  There's no need for an independent DB setter/getter as this should be part of the constructor process

gabrocheleau:
This PR closes #3763 by

- Defaulting the constructor to MapDB if no DB is provided. Also cleans up the constructor logic a bit.
- Making the _createRootNode function public.
- Getting rid of the VerkleTree.database function.


gabrocheleau:
One thing I noticed when testing this updated constructor setup. 

We can make verkleCrypto optional and instantiate a verkle-crypto instance from the loadVerkleCrypto ourselves. However, this leads to issues when other verkle-related functions are called with another verkle-crypto instantiation:

this works:
```
    const trie = await createVerkleTree({ verkleCrypto })

    await trie.createRootNode()

    let putStack: [Uint8Array, VerkleNode][] = []
    const stem1 = keys[0].slice(0, 31)
    // Create first leaf node
    const leafNode1 = await LeafVerkleNode.create(stem1, verkleCrypto)
```

this fails:
```
    const trie = await createVerkleTree({ verkleCrypto })

    await trie.createRootNode()

    let putStack: [Uint8Array, VerkleNode][] = []
    const stem1 = keys[0].slice(0, 31)
    // Create first leaf node
    const leafNode1 = await LeafVerkleNode.create(stem1, verkleCrypto)
 ```
 
 I am not actually sure why that is.

File: packages/verkle/src/types.ts [hunk 1]
```diff
@@ -11,11 +11,11 @@ export interface VerkleTreeOpts {
   /**
    * An instantiated Verkle Cryptography interface
    */
-  verkleCrypto: VerkleCrypto | undefined
+  verkleCrypto?: VerkleCrypto | undefined
   /**
    * A database instance.
    */
-  db: DB<Uint8Array, Uint8Array>
+  db?: DB<Uint8Array, Uint8Array>
```
acolytec3, line 18: So a thought here now that I see the code, we should either make these two opts mandatory or else replace the `createRootNode` with a general `init` that requires the db and verkle crypto objects and then initializes the tree with this empty root node.  In doing so, we should also check to see if the tree has the `useRootPersistence` opt set and throw if so and the rootnode already exists (since that would overwrite the root node and be destructive to the existing trie).

To do this, you need to do `tree.findPath(getVerkleStem(this.root))` and see if you get back an existing node.

What do you think?

packages/verkle/src/constructors.ts
```diff
@@ -6,36 +6,33 @@ import { VerkleTree } from './verkleTree.js'
...
-      }
-    else {
-      opts.verkleCrypto = verkleCrypto
-    }
-  }
-
-  const trie = new VerkleTree(opts)
-  await trie['_createRootNode']()
+  const trie = new VerkleTree(parsedOptions)
+  await trie.createRootNode()
```
acolytec3, line 36: You don't want to call this if `useRootPersistence === true` and you can retrieve the `parsedOptions.root` enables you to retrieve an actual node from the db.

packages/verkle/test/proof.spec.ts
```diff
@@ -29,12 +29,9 @@ describe('lets make proofs', () => {
...
       '0x0300000000000000000000000000000000000000000000000000000000000000',
     ].map((key) => hexToBytes(key as PrefixedHexString))
-    const trie = await createVerkleTree({
-      verkleCrypto,
-      db: new MapDB<Uint8Array, Uint8Array>(),
-    })
+    const trie = await createVerkleTree()
 
-    await trie['_createRootNode']()
+    await trie.createRootNode()
```
acolytec3, line 34:
```suggestion
```

This isn't necessary if you use `createVerkleTree`

packages/verkle/test/proof.spec.ts
```diff
@@ -73,7 +70,7 @@ describe('lets make proofs', () => {
   it('should pass for empty trie', async () => {
     const trie = await createVerkleTree({ verkleCrypto, db: new MapDB() })
 
-    await trie['_createRootNode']()
+    await trie.createRootNode()
```
acolytec3, line 73:
```suggestion
```

acolytec3 (approved):

holgerd77:
Nice! 👍

PR #3769, opened 2024-10-24
monorepo: run examples build

gabrocheleau:
Through updating the examples for another PR, I noticed that quite a few were outdated. I've therefore split this into this discinct PR.


acolytec3 (approved):

packages/devp2p/README.md
```diff
@@ -262,7 +261,7 @@ Wait for follow-up messages to arrive, send your responses.
 ```ts
 // ./examples/peer-communication.ts#L116-L119
 
-eth.events.on('message', async (code: ETH.MESSAGE_CODES, payload: any) => {
+eth.events.on('message', async (code: any, payload: any) => {
```
holgerd77, line 264: ? //cc @acolytec3

acolytec3: Oops 😅

holgerd77: Ah, so there is absolutely no "Oops" for you in there 😂 , so the code is from @gabrocheleau, I was just wondering if there is an `any` necessary here (likely not? at least I would hope? 🙂) and wanted some confirmation "from the expert". So if not we should remove again here.

(@acolytec3: if you are working on something then maybe just do along other work, so that we do not need a full PR for this, and just drop a note here, or otherwise @gabrocheleau ).

Update: ah, yes, just tested, `any` can safely be deleted.

PR #3770, opened 2024-10-25
evm: migrate accessWitness file to evm package

### LINKED ISSUE #3749: EVM: Refactor accessWitness from StateManager into EVM
We have recently discussed the possibility of migrating the accessWitness from the StateManager to the EVM. It makes conceptual sense for the EVM to be in charge of this low-level tracking rather than the StateManager.

The accessWitness file itself is fairly minimal, only importing from the `common` and `util` packages, so the migration of the file itself is straightforward.

However there are some additional considerations:

- accessWitness is deeply tied to verkle logic, notably for pricing execution witnesses (ref: https://eips.ethereum.org/EIPS/eip-4762), and we should only enable when verkle (EIP-6800) is also enabled. I think we want this to be hardfork/EIP-inferred rather than manually specified so that we are able to dynamically transition in the context of the merkle->verkle transition. 
- verkle state managers (both stateless and stateful) will have to be adjusted so that they obtain the acccess witness from the EVM somehow for things like postStateVerify. This shouldn't be too tricky, as I think postStateVerify is called by the VM, which could also pass in the accessWitness as an arg.
- We'll need to create an AccessWitness interface or type for use for e.g. in the StateManager (we wouldn't want to add an EVM dependency just for importing the AccessWitness types, enums, etc)

gabrocheleau:
This PR closes #3749 by refactoring the accessWitness from the StateManager to the EVM


packages/evm/src/accessWitness.ts
```diff
@@ -358,16 +357,19 @@ export function decodeAccessedState(treeIndex: number | bigint, chunkIndex: numb
...
   if (value === null) {
     return ''
   }
 
   switch (type) {
-    case AccessedStateType.BasicData:
-    case AccessedStateType.CodeHash:
-    case AccessedStateType.Code:
-    case AccessedStateType.Storage: {
+    case VerkleAccessedStateType.BasicData:
```
acolytec3, line 369: Do we need to use `decodeVerkleBasicData` on this one?

packages/evm/src/accessWitness.ts
```diff
@@ -43,17 +44,15 @@ type RawAccessedState = {
...
-  | { type: AccessedStateType.Storage; slot: bigint }
+  | {
+      type: Exclude<
+        VerkleAccessedStateType,
+        VerkleAccessedStateType.Code | VerkleAccessedStateType.Storage
+      >
+    }
+  | { type: VerkleAccessedStateType.Code; codeOffset: number }
+  | { type: VerkleAccessedStateType.Storage; slot: bigint }
 export type AccessedStateWithAddress = AccessedState & {
```
acolytec3, line 56: Shouldn't these types move to `common` or else `util`?  We need them in the verkle state managers and can't import them from `evm` since that would create a circular dep.

gabrocheleau: Yeah agreed, I really haven't done much work yet (just moved the few obvious parts yesterday before going to bed), probably best to wait a bit before reviewing

packages/statemanager/src/util.ts
```diff
@@ -1,7 +1,32 @@
...
+ */
+export function decodeVerkleAccessWitnessValue(
+  type: VerkleAccessedStateType,
+  value: PrefixedHexString | null,
+): string {
+  if (value === null) {
+    return ''
+  }
+
+  switch (type) {
```
acolytec3, line 21: Is there an expectation that we might change how each one is handled in the future?  This seems like a lot of lines of code for essentially `value === null ? '' : value`

packages/evm/src/verkleAccessWitness.ts
```diff
@@ -276,11 +259,11 @@ export class AccessWitness implements AccessWitnessInterface {
   }
 
   /**Create a shallow copy, could clone some caches in future for optimizations */
-  shallowCopy(): AccessWitness {
-    return new AccessWitness({ verkleCrypto: this.verkleCrypto })
+  shallowCopy(): VerkleAccessWitness {
```
acolytec3, line 262: Do we actually need this? I don't see it referenced anywhere in the codebase?  Doesn't seem like it does anything other than just instantiate a brand new access witness?  Also, the code doc references caches and stuff and the `AccessWitness` class doesn't appear to have any caches.

packages/vm/src/runBlock.ts
```diff
@@ -151,6 +155,8 @@ export async function runBlock(vm: VM, opts: RunBlockOpts): Promise<RunBlockResu
       // Update the stateRoot cache
       await stateManager.setStateRoot(block.header.stateRoot)
 
+      // Initialize the access witness
+      vm.evm.verkleAccessWitness = new VerkleAccessWitness({ verkleCrypto: vm.evm.verkleCrypto })
```
acolytec3, line 159: Could we just borrow the `verkleCrypto` from the verkle state manager rather than having to add it as a property on `evm`/`vm`?  Seems like we're adding some unnecessary overhead for instantiating the `vm`/`evm` for this very isolated use case. If you are doing verkle stuff, you already instantiate it in the verkle state manager classes (either directly or indirectly through the `create*StateManager` constructor functions).  Seems like a better user experience to not have to bother instantiating it.
```suggestion
      vm.evm.verkleAccessWitness = new VerkleAccessWitness({ verkleCrypto: vm.statemanager['verkleCrypto'] })
```

acolytec3 (commented):
Generally looks good but I would love to not have to pass `verkleCrypto` in to the `vm` and `evm` constructors since it's only use is instantiating the witness class and we already have verkleCrypto in the statemanager for this exact purpose.

packages/statemanager/src/util.ts
gabrocheleau: This is really just a relic from the time where we had separate treatment for each. I'll remove it, it brings no value at this time.

gabrocheleau: I've simplified this part and the usage upstream. ty for the suggestion

packages/evm/src/verkleAccessWitness.ts
gabrocheleau: Hmm good point, perhaps something that was at some point used but then left out, will remove

gabrocheleau: Ah, actually we are using it here: https://github.com/ethereumjs/ethereumjs-monorepo/blob/evm/access-witness-refactor/packages/vm/src/runTx.ts#L200

so I won't remove

acolytec3: I think my point here is that `shallowCopy` here isn't copying anything. It's just calling the access witness constructor again. Couldn't we just replace the call in`runTx` with a call to the constructor and then remove this code?

gabrocheleau: Ah yeah that's right. Removed

c4brown1:
Remove Velcro

acolytec3 (approved):
LGTM

packages/vm/src/runBlock.ts
holgerd77: @gabrocheleau @acolytec3 I think with verkleCrypto we should go over Common here as we do with all our custom and shared crypto, that would be by far the cleanest solution, compared to “borrowing implicitly from somewhere”.

holgerd77 (approved):
Nice, this looks like a (relatively) straightforward refactoring on first glance, @gabrocheleau can you confirm this? Or was this more twisted than on first look?

Also, so do you feel good with this new placement here and are you confident, that this was the "right thing to do" [tm]? 🙂 

I really think we should still move `verkleCrypto` to `customCrypto` in `Common`, (so then: `customCrypto.verkle` I guess), having a relatively strong opinion on this, unless someone counter-prooves that this is an unlucky choice for reasons x,y.

For me this would feel a lot more consistent.

packages/vm/src/runBlock.ts
gabrocheleau: Yeah, I think that makes sense to me. Created this issue for it: https://github.com/ethereumjs/ethereumjs-monorepo/issues/3774

gabrocheleau:
> Nice, this looks like a (relatively) straightforward refactoring on first glance, @gabrocheleau can you confirm this? Or was this more twisted than on first look?
> 
> Also, so do you feel good with this new placement here and are you confident, that this was the "right thing to do" [tm]? 🙂
> 
> I really think we should still move `verkleCrypto` to `customCrypto` in `Common`, (so then: `customCrypto.verkle` I guess), having a relatively strong opinion on this, unless someone counter-prooves that this is an unlucky choice for reasons x,y.
> 
> For me this would feel a lot more consistent.

It definitely makes sense to me to extract this from the stateManager, the accessWitness is concerned with the low-level calls of the EVM, and is mostly used there. Putting it in the EVM also allows us to import it in the VM (which wouldn't be the case if for e.g. we had refactored it to the VM, which was another defensible option). I think I agree with you with moving verkleCrypto to common. I'll try it out and we can evaluate from the actual PR.

PR #3706, opened 2024-09-27
pectra devnet4: implement pectra devnet4 spec

g11tech:
refrence:

https://notes.ethereum.org/@ethpandaops/pectra-devnet-4

image

PR #3771, opened 2024-10-25
packages: add requests hash to genesis block

spencer-tb:
# Description
Adds the request hash to the genesis block to enable running tests on hive: `consume engine` and `consume rlp`.

## Reproduce Hive Passes

Using the following hive branch [spencer-tb@prague-devnet-4-ethjs](https://github.com/spencer-tb/hive/tree/prague-devnet-4-ethjs) run either of:
```bash
./hive --sim 'ethereum/eest/consume-block-rlp' --client-file configs/prague.yaml --docker.output --client ethereumjs --docker.nocache .
./hive --sim 'ethereum/eest/consume-engine' --client-file configs/prague.yaml --docker.output --client ethereumjs --docker.nocache .
```

## Empty Requests Hash

Note the keccak256 of the empty requests should account for the type prefixing, hence why:
```
KECCAK256_RLP_RH == 0x6036c41849da9c076ed79654d434017387a88fb833c2856b32e18218b3341c5f
```
and not the default `0x56e8...`.

### DIFF
--- a/packages/util/src/constants.ts
+++ b/packages/util/src/constants.ts
@@ -68,6 +68,17 @@ export const KECCAK256_RLP = hexToBytes(KECCAK256_RLP_S)
 export const SHA256_NULL = sha256(new Uint8Array())
 
 /**
+ * Keccak-256 hash of the RLP of an empty requests hash
+ */
+export const KECCAK256_RLP_RH_S =
+  '0x6036c41849da9c076ed79654d434017387a88fb833c2856b32e18218b3341c5f'
+
+/**
+ * Keccak-256 hash of the RLP of an empty requests hash
+ */
+export const KECCAK256_RLP_RH = hexToBytes(KECCAK256_RLP_RH_S)
+
+/**
  *  RLP encoded empty string
  */
 export const RLP_EMPTY_STRING = Uint8Array.from([0x80])
--- a/packages/util/src/constants.ts
+++ b/packages/util/src/constants.ts
@@ -70,13 +70,7 @@ export const SHA256_NULL = sha256(new Uint8Array())
 /**
  * Keccak-256 hash of the RLP of an empty requests hash
  */
-export const KECCAK256_RLP_RH_S =
-  export const SHA256_EMPTY_RH_S = sha256(new Uint8Array([0, 1, 2]))
-
-/**
- * Keccak-256 hash of the RLP of an empty requests hash
- */
-export const SHA256_EMPTY_RH = hexToBytes(SHA256_EMPTY_RH_S)
+export const SHA256_EMPTY_RH = sha256(new Uint8Array([0, 1, 2]))
 
 /**
  *  RLP encoded empty string
--- a/packages/util/src/constants.ts
+++ b/packages/util/src/constants.ts
@@ -71,7 +71,7 @@ export const SHA256_NULL = sha256(new Uint8Array())
  * Keccak-256 hash of the RLP of an empty requests hash
  */
 export const KECCAK256_RLP_RH_S =
-  '0x6036c41849da9c076ed79654d434017387a88fb833c2856b32e18218b3341c5f'
+  export const SHA256_EMPTY_RH_S = sha256(new Uint8Array([0, 1, 2]))
 
 /**
  * Keccak-256 hash of the RLP of an empty requests hash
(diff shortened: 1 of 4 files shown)


packages/util/src/constants.ts
```diff
@@ -64,6 +64,17 @@ export const KECCAK256_RLP_S = '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cad
  */
 export const KECCAK256_RLP = hexToBytes(KECCAK256_RLP_S)
 
+/**
+ * Keccak-256 hash of the RLP of an empty requests hash
+ */
+export const KECCAK256_RLP_RH_S =
+  '0x6036c41849da9c076ed79654d434017387a88fb833c2856b32e18218b3341c5f'
```
jochem-brouwer, line 71: Could you clarify this "magic constant" by explicitly calling the hash function with the input? I also think this should be SHA256? For the empty hash (or also all other hashes) this is not the RLP of the input (the input are opaque bytes which are concatenated).

```suggestion
  export const SHA256_EMPTY_RH_S = sha256(new Uint8Array([0, 1, 2]))
```

(You will need to also import sha256 likely)

packages/util/src/constants.ts
```diff
@@ -64,6 +64,17 @@ export const KECCAK256_RLP_S = '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cad
...
+
+/**
+ * Keccak-256 hash of the RLP of an empty requests hash
+ */
+export const KECCAK256_RLP_RH = hexToBytes(KECCAK256_RLP_RH_S)
```
jochem-brouwer, line 76:
```suggestion
export const SHA256_EMPTY_RH = hexToBytes(SHA256_EMPTY_RH_S)
```

jochem-brouwer (commented):
This looks great, thanks a lot for this (will have to test locally). One small point regarding correctness and clarity (for devnet-4).

Side-note, in devnet-5 this will get easier, since the empty request hash will be the empty sha256 hash, see:

https://github.com/ethereum/EIPs/pull/8989

(We will nevertheless need this change to import genesis blocks which the requests hash, plus we also need the specific empty hash for devnet-4 :) )

packages/util/src/constants.ts
```diff
@@ -64,6 +65,11 @@ export const KECCAK256_RLP_S = '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cad
  */
 export const KECCAK256_RLP = hexToBytes(KECCAK256_RLP_S)
 
+/**
+ * Keccak-256 hash of the RLP of an empty requests hash
```
jochem-brouwer, line 69:
```suggestion
 * SHA-256 hash of the RLP of an empty requests hash
```

jochem-brouwer (approved):

jochem-brouwer:
What does `docker.nocache .` do? Will it re-build all docker images each time? :thinking:

packages/util/src/constants.ts
```diff
@@ -67,6 +67,11 @@ export const KECCAK256_RLP = hexToBytes(KECCAK256_RLP_S)
 
 export const SHA256_NULL = sha256(new Uint8Array())
 
+/**
+ * SHA-256 hash of the RLP of an empty requests hash
+ */
+export const SHA256_EMPTY_RH = sha256(new Uint8Array([0, 1, 2]))
```
holgerd77, line 73: This here would become part of the official API of the libraries and it would be hard "to get rid of it" if things change, so let's please not do this yet as long as specs are still changing and rather hardcode in blockchain.

jochem-brouwer: Hi Holger, good point, I did somewhat notice this but note that the "empty requests hash" will be gone starting of devnet-5 (see: https://github.com/ethereum/EIPs/pull/8989), where it is the empty sha256 hash and not this thing with encoded constants (so we can get rid of this constant!). But I agree it should not have been put in the util constants (note: this PR is not merged into master, it is merged in our devnet4 branch, so for now this is fine! :smile: )

jochem-brouwer: Should nevertheless not forget to remove this constant :smile: :+1:

holgerd77: These things DO have some tendency to get forgotten, so I would still pledge to just give this a quick update *now* and we are done with it.

jochem-brouwer: Done in https://github.com/ethereumjs/ethereumjs-monorepo/pull/3706/commits/dc93f8f0cec8b0b025d3c393f1b336ee08349e73

PR #3777, opened 2024-10-29
client: fix double execution of a newplayload block on fcu

g11tech:
as discovered by @jochem-brouwer , the newly block to be added is doubly executed by new payload and then on fcu by the vmexecution's forward run() which gets triggered with the chain update event when the block is inserted in the chain by the fcU.

this two prong execution mechanism iis there to handle various edgecases but in this case causes a race. so delaying the update and letting fcU do the update event should take care of this race.


jochem-brouwer:
LGTM! Will merge once CI passes

PR #3778, opened 2024-10-29
Fix handling of storage values

acolytec3:
This PR fixes an issue in the `put/getStorage` handlers in the `statefulVerkleStateManager` where we were RLP encoding storage values (copied from our MPT based statemanager) and this was at variance with the current reference implementation in geth.

This also adds a `dumpLeafValues` helper to dump all the keys and values for leaf values in the tree.


holgerd77 (approved):
TMB

PR #3779, opened 2024-10-30
monorepo: fix npm high severity issues

gabrocheleau:
This PR fixes the high severity issues reported by NPM by manually resolving puppeteer-core
 (which was the root cause of these issues) to a more recent, patched version.

There are still low-severity issues (related to cookie-handling) which should not be a concern for us, and can't be fixed easily right now as we would have to downgrade vitest-browser to 1.6.0 (we're on 2.1.0). I think we should rather wait for them to get this fixed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,7 +53,7 @@
       },
       "peerDependencies": {
         "@vitest/browser": "^2.1.0",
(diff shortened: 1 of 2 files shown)


holgerd77 (approved):
Thanks a lot! 🙏

Issue #3776, opened 2024-10-29
Run `inlineClient` in REPL


acolytec3:
One thing that might make troubleshooting the client on devnets and the like easier would be to be able to run it in a REPL so we can have direct access to the client object as its running, similarly to how geth has their javascript console.

We already have a [createInlineClient](https://github.com/ethereumjs/ethereumjs-monorepo/blob/abb381594f3c5b1dcafdc2bffe9510c7f2de4149/packages/client/test/sim/simutils.ts#L433) function available to us that accepts the entire configuration object available to the client so it shouldn't be too hard to expand this a bit to allow a user to pass in a geth genesis.json and start a client entirely within a REPL.

The piece that isn't built yet would be hooking it up to the RPC.  We have a [startRPCServers](https://github.com/ethereumjs/ethereumjs-monorepo/blob/abb381594f3c5b1dcafdc2bffe9510c7f2de4149/packages/client/bin/startRPC.ts#L75) function that we could pass the instantiated "inline" client to and then just grab a reference to it in the REPL so we can be sure to not lose it.

This could also be an interesting "community showcase" idea.

acolytec3:
This script is a working proof of concept.

```ts
const { readFileSync } = await import('fs')

const { createCommonFromGethGenesis, parseGethGenesis } = await import('@ethereumjs/common')
const { createInlineClient } = await import('./test/sim/simutils.ts')
const { Config } = await import('./src/config.ts')
const { getLogger } = await import('./src/logging.ts')
const { startRPCServers } = await import('./bin/startRPC.ts')
const genesisFile = JSON.parse(readFileSync('./genesis.json', 'utf-8'))
const chainName = 'pectra'
const common = createCommonFromGethGenesis(genesisFile, {
  chain: chainName,
})
const config = new Config({ common, logger: getLogger({ logLevel: 'debug' }), saveReceipts: true, enableSnapSync: true })
const client = await createInlineClient(config, common, {}, '', true)
const servers = startRPCServers(client, {
  rpc: true,
  rpcAddr: '0.0.0.0',
  rpcPort: 8545,
  ws: false,
  wsPort: 0,
  wsAddr: '0.0.0.0',
  rpcEngine: true,
  rpcEngineAddr: '0.0.0.0',
  rpcEnginePort: 8551,
  wsEngineAddr: '0.0.0.0',
  wsEnginePort: 8552,
  rpcDebug: 'eth',
[...]

PR #3781, opened 2024-10-31
Ethjs JavaScript REPL

scorbajio:
This change continues from a PoC created in #3776 to add a repl to the ethereumjs client module.


packages/client/bin/repl.ts
```diff
@@ -0,0 +1,85 @@
...
+
+const setupRepl = async () => {
+  const { client, executionRpc, engineRpc } = await setupClient()
+
+  const replServer = repl.start({
+    prompt: 'EthJS > ',
+    ignoreUndefined: true,
+  })
+
+  // bootstrap contexts or modules
```
scorbajio, line 54: We could potentially make the rpc servers and the API's they serve available as contexts here, but on first look, this wouldn't be very usable. A more user friendly option might be defining commands in the repl that would interact with the client and rpc server objects that could be called like this: `.getBlock 0x01`.

packages/client/bin/repl.ts
```diff
@@ -0,0 +1,85 @@
...
+  replServer.context.client = client
+  replServer.context.executionRpc = executionRpc._methods // TODO modify methods to only include functions and make them usable
+  replServer.context.engineRpc = engineRpc._methods
+
+  replServer.on('exit', () => {
+    console.log('Exiting REPL...')
+    process.exit()
+  })
+
+  // define commands
```
scorbajio, line 64: Commands can be defined as such that would allow use to simplify rpc API function calls and passing input parameters. Any other processing can also be done in the command callback implementation, but we should be careful not to e.g. reimplement rpc functions here.

acolytec3:
Another thing that would be really cool would be to be able to turn on the different log levels dynamically in the client (and also maybe the `debug` stuff too to see lower level debug logging.

holgerd77:
Can we take the `admin_addPeer` RPC implementation out here as a separate PR (and also add a few tests), so that there is not too much heaping up here?

> Another thing that would be really cool would be to be able to turn on the different log levels dynamically in the client (and also maybe the debug stuff too to see lower level debug logging.

This might also be a candidate for a separate PR.

I have the general impression that this might be a good strategic approach, respectively that this here (the REPL PR) is somewhat of a "sub task trigger PR" 🙂 So it's pretty likely that this will let one stumble upon refactoring tasks, clean-up TODOs and the like and then I think it might also be good to continue like this: to then extract these as separate/small pieces of work and keep this PR here lean and small and updated.

holgerd77:
(but then likely do the PRs on top of each other, so that you do not always have to wait a day until one of these is merged. So basically (PR order) from clean (mergeable, e.g. a `admin_addPeer PR with tests) to dirty (likely the REPL PR for some time 😁))

acolytec3:
> (but then likely do the PRs on top of each other, so that you do not always have to wait a day until one of these is merged. So basically (PR order) from clean (mergeable, e.g. a `admin_addPeer PR with tests) to dirty (likely the REPL PR for some time 😁))

Agree.  This has been working well for the verkle stuff.

packages/client/test/rpc/websocket.spec.ts
```diff
@@ -54,7 +54,7 @@ describe('JSON-RPC call', () => {
         resolve(undefined)
       })
     })
-  }, 3000)
+  }, 30000)
```
scorbajio, line 57: This specific test was timing out on this branch in the CI (it was passing locally).

packages/client/src/util/simutils.ts
```diff
@@ -0,0 +1,68 @@
...
+import { MemoryLevel } from 'memory-level'
+
+import { EthereumClient } from '../../src/client.js'
+import { Config } from '../../src/config.js'
+import { LevelDB } from '../../src/execution/level.js'
+
+import type { ConsensusDict } from '@ethereumjs/blockchain'
+import type { GenesisState } from '@ethereumjs/util'
+
+export async function createInlineClient(
```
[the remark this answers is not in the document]
acolytec3: Can we change the name of this file to `inlineClient` or something similar?  It's not "simutils" in this context.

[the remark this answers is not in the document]
scorbajio: Fixed

jochem-brouwer:
If I `npm run repl` this starts up a client and this also gives me a REPL interface. However, the client starting with log outputs interferes with my console interaction. Also, how do I use the console / how am I supposed to use this? How would I add a peer, for instance (and is this part of the goal here?)? I can interact with `client`, but if I start typing this (so type `client` in the REPL), it auto-completes and it interferes with the interaction.

acolytec3 (approved):
LGTM :shipit:

jochem-brouwer (dismissed):
LGTM :tada:

jochem-brouwer (approved):
LGTM :rocket: :shipit:

jochem-brouwer (approved):
LGTM :tada: :rocket: :shipit:

PR #3785, opened 2024-11-03
Implement Verkle in JS

paulmillr:
https://github.com/paulmillr/micro-eth-signer/blob/main/src/verkle.ts
https://github.com/paulmillr/micro-eth-signer/blob/main/test/verkle.test.js

```
commitToScalars
├─wasm x 16,225 ops/sec @ 61μs/op
└─micro x 9,761 ops/sec @ 102μs/op
createProof
├─wasm x 6 ops/sec @ 158ms/op
└─micro x 3 ops/sec @ 266ms/op
getTreeKey
├─wasm x 41,921 ops/sec @ 23μs/op
└─micro x 4,545 ops/sec @ 220μs/op
hashCommitment
├─wasm x 215,100 ops/sec @ 4μs/op
└─micro x 332,778 ops/sec @ 3μs/op
serializeCommitment
├─wasm x 1,154,734 ops/sec @ 866ns/op
└─micro x 377,786 ops/sec @ 2μs/op
updateCommitment
├─wasm x 9,471 ops/sec @ 105μs/op
└─micro x 3,222 ops/sec @ 310μs/op
verifyExecutionWitnessPreState
├─wasm x 336 ops/sec @ 2ms/op
└─micro x 7,418 ops/sec @ 134μs/op
verifyProof
├─wasm x 49 ops/sec @ 20ms/op
└─micro x 196 ops/sec @ 5ms/op
```


gabrocheleau:
Wow! Great that you've re-implemented this library in JS. Thanks a lot for the work + for pushing that PR and providing some performance data. Performance-wise, the one thing I can see that could be a concern is `getTreeKey`, as it's 10x less performant and is quite heavily used in verkle work. Still, we think that moving to a native JS library is preferable, and will be doing some work soon-ish to modularize this so that users can easily specify the cryptography library that they wish to use. Will approve & merge once tests pass.

gabrocheleau (approved):

packages/statemanager/package.json
```diff
@@ -64,7 +64,6 @@
     "@ethereumjs/block": "^5.3.0",
     "@ethereumjs/genesis": "^0.2.3",
     "@types/debug": "^4.1.9",
-    "rustbn-wasm": "^0.4.0",
-    "verkle-cryptography-wasm": "^0.4.8"
+    "rustbn-wasm": "^0.4.0"
```
paulmillr, line 67: This also seems to be not needed anymore?

holgerd77:
Hi Paul, this is absurdly impressive that you ad-hoc come up here with a fully test passing Verkle crypto JS implementation, also taken into account that the [implementation](https://github.com/paulmillr/micro-eth-signer/blob/main/src/verkle.ts) really does not fall under "small addition to KZG" anymore, also linking the [tests](https://github.com/paulmillr/micro-eth-signer/blob/main/test/verkle.test.js) here for completeness!! 🎉 🎉 🤯 

Thanks so much for this!

From my side we can merge this in (with this note from @gabrocheleau "some work soon-ish to modularize this so that users can easily specify the cryptography library" in the back), will wait for @gabrocheleau or @acolytec3 though to do so).

Again: *GREAT WORK!!!* 🙏

holgerd77:
Ah, just to note: so these specs will likely continue to change (occasionally I guess) in the future? Are you willing to do occasional continued updates on the library (relatively short-termed, so: within days up to 1-2 weeks) in the future? That would be important for us since otherwise this would always break our testnet compatibility.

Not sure if this is rather a bit theoretical though and this lower-level part of the spec is rather stable, maybe Gabriel or Andrew can also have a word here.

paulmillr:
Confirming: the plan is to do all necessary changes to align the package with future testnets and spec updates.

packages/statemanager/package.json
holgerd77: Yes, you are right, not sure how this came in here in the first place? Anyhow, have directly removed via GitHub online editor and committed.

acolytec3 (approved):
Thanks for this!  Can't wait to try it out in some other contexts.  I do think it makes sense to move verkle instantiation into `common` and attach it to the `customCrypto` property as that will get rid of this somewhat unnatural `verkleCrypto = await new Promise.resolve(verkle)` thing.  But we can do that in a future PR.

PR #3787, opened 2024-11-03
statemanager: add put/delete test case

gabrocheleau:
This PR adds a test case that does the following:

1. Creates a new verkleTree
2. Puts an account into the verkleTree and gets the stateRoot (stateRoot1)
3. Puts another account into the verkle tree and ensures that the stateRoot (stateRoot2) differs from stateRoot1 
4. Deletes account2 from the verkle tree and ensures that stateRoot now matches stateRoot1

This is currently failing and I suspect that this causes some failures that we are seeing in the test fixtures, most likely related to the handling of Untouched values and account deletion


acolytec3:
Additional changes I've added to this PR
1) new `dumpNodeHashes` helper for viewing the hashes and paths of all nodes in a trie
2) New logic branches in `put` to remove leaf nodes when all leaf values are written as `untouched`
3) New logic to collapse inner nodes when it only has one leaf node value after deleting another leaf node and replace parent node reference with the surviving leaf node

Note, 2-3 should only ever happen when we are reverting an ephemeral state change (i.e. an account gets touched in the course of block execution but the tx reverts or an account is otherwise deleted prior to being written to state)

I'm not 100% confident this is the longterm approach we want since it would preferable to always just do our account updates ephemerally in a cache and only write to the trie once we are ready to "commit" and "flush" values to the db.

packages/verkle/src/util.ts
```diff
@@ -28,9 +34,46 @@ export const dumpLeafValues = async (
...
+ * @returns An array of key-value pairs containing the tree paths and associated hashes
+ */
+export const dumpNodeHashes = async (
+  tree: VerkleTree,
+  startingNode: Uint8Array,
+): Promise<[PrefixedHexString, PrefixedHexString][] | undefined> => {
+  let entries: [PrefixedHexString, PrefixedHexString][] = []
+  // Retrieve starting node from DB
+  const rawNode = await tree['_db'].get(startingNode)
+  if (rawNode === undefined) return
```
gabrocheleau, line 56: Why do we want to return `undefined` here instead of `[]` ?

packages/verkle/src/util.ts
```diff
@@ -28,9 +34,46 @@ export const dumpLeafValues = async (
...
+
+  const node = decodeVerkleNode(rawNode, tree['verkleCrypto'])
+  // If current node is root, push '0x' for path and node hash for commitment
+  equalsBytes(startingNode, tree.root()) && entries.push(['0x', bytesToHex(startingNode)])
+  if (node instanceof LeafVerkleNode) {
+    // Leaf node paths/hashes were added in the previous inner node's iteration
+    return []
```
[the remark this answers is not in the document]
gabrocheleau: I've pushed that change and made some minor simplifications/added some comments. Let me know how you like it.

packages/verkle/src/verkleTree.ts
```diff
@@ -217,8 +217,20 @@ export class VerkleTree {
           ['put'],
         )
     }
-    // Push new/updated leafNode to putStack
-    putStack.push([leafNode.hash(), leafNode])
+    if (leafNode.values.filter((val) => val !== LeafVerkleNodeValue.Untouched).length === 0) {
```
gabrocheleau, line 220: I was wondering, when troubleshooting over the weekend, if there should be another way to mark a node as "to be deleted" that does not require to "put untouched on all leafNode values". 

I think your intuition of a lifecycle where we "commit" and "flush" sounds right. I don't think it's necessarily what we should do now (maybe worth discussing that tomorrow on the call?) as this PR definitely brings enough improvements to be merged (and can get us to iterate on verkle tests quicker), but I also feel like we need, in the end, a better architecture.

acolytec3 (approved):
LGTM

PR #3789, opened 2024-11-06
Monorepo: Simplify coverage Command & Reactivate Reporting

holgerd77:
This simplifies the coverage runs by removing the c8 tool usage and directly use the in-build vitest coverage functionality.

Beside being simpler goal is to reactivate coverage reporting for certain packages.


holgerd77 (commented):
We should clean these things up a bit at some point (not urgent! focus on your Devcon stuff!! 😋), so decide what to do with `coverage.sh` (still needed?), also align `coverage` commands, but I guess we can take in for now, coverage increase seems to be working?

Will do by admin-merge (own PR continued).

PR #3790, opened 2024-11-07
common: refactor verkleCrypto into customCrypto from common

### LINKED ISSUE #3774: Move over verkleCrypto handling into common
We currently are in a situation where we are passing `verkleCrypto` around in the VM/EVM/StateManager. With the latest AccessWitness work from #3770, we are now getting the verkleCrypto instantiation from stateManager, which is a bit roundabout and counterintuitive. As suggested by Holger in this comment: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3770#discussion_r1818569641, we could instead move it into Common.

gabrocheleau:
This PR resolves #3774 by moving the verkleCrypto handling into common

### DIFF
--- a/packages/client/test/rpc/engine/kaustinen6.spec.ts
+++ b/packages/client/test/rpc/engine/kaustinen6.spec.ts
@@ -6,6 +6,7 @@ import {
 } from '@ethereumjs/block'
 import { hexToBytes } from '@ethereumjs/util'
 import { readFileSync } from 'fs'
+import * as verkle from 'micro-eth-signer'
 import * as td from 'testdouble'
 import { assert, describe, it } from 'vitest'
 
@@ -80,6 +81,7 @@ describe(`valid verkle network setup`, async () => {
   const { server, chain, common } = await setupChain(kaustinen6Data, 'post-merge', {
     engine: true,
     genesisStateRoot: genesisVerkleStateRoot,
+    customCrypto: { verkleCrypto: verkle },
   })
   const rpc = getRPCClient(server)
   it('genesis should be correctly setup', async () => {
--- a/packages/statemanager/src/statelessVerkleStateManager.ts
+++ b/packages/statemanager/src/statelessVerkleStateManager.ts
@@ -116,7 +116,7 @@ export class StatelessVerkleStateManager implements StateManagerInterface {
     this._caches = opts.caches
 
     if (opts.common.isActivatedEIP(6800) === false) {
-      throw new Error('EIP-6800 required for verkle state management')
+      throw new Error('EIP-6800 required for stateless verkle state management')
     }
 
     if (opts.common.customCrypto.verkleCrypto === undefined) {
--- a/packages/statemanager/src/statelessVerkleStateManager.ts
+++ b/packages/statemanager/src/statelessVerkleStateManager.ts
@@ -33,6 +33,7 @@ import type { StatelessVerkleStateManagerOpts, VerkleState } from './index.js'
 import type { MerkleStateManager } from './merkleStateManager.js'
 import type {
   AccountFields,
+  Common,
   StateManagerInterface,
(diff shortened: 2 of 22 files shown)


packages/statemanager/src/statelessVerkleStateManager.ts
```diff
@@ -112,14 +115,19 @@ export class StatelessVerkleStateManager implements StateManagerInterface {
 
     this._caches = opts.caches
 
-    this.keccakFunction = opts.common?.customCrypto.keccak256 ?? keccak256
-
-    this._debug = debugDefault('statemanager:verkle:stateless')
+    if (opts.common.isActivatedEIP(6800) === false) {
+      throw new Error('EIP-6800 required for verkle state management')
```
jochem-brouwer, line 119:
```suggestion
      throw new Error('EIP-6800 required for stateless verkle state management')
```

This is the stateless verkle state manager, so I don't think this message is completely correct?

packages/statemanager/test/statefulVerkleStateManager.spec.ts
```diff
@@ -40,7 +46,12 @@ describe('Verkle Tree API tests', () => {
 
   it('should return same stateRoot when putting and then deleting account', async () => {
     const trie = await createVerkleTree()
-    const sm = new StatefulVerkleStateManager({ trie, verkleCrypto })
+    const common = new Common({
+      chain: Mainnet,
+      eips: [6800],
+      customCrypto: { verkleCrypto },
```
jochem-brouwer, line 49: I have a suggestion, can we do the check for the existence of `verkleCrypto` in `Common` inside common? So, if 6800 is active, or it is activated, check if `verkleCrypto` is defined?

Not sure here though, since this would need a check for all cases when 6800 gets active (so when `new Common` constructing it and `common.setActiveEIPs`)

jochem-brouwer (commented):
2 comments. 

I did not check `packages/util.src/verkle.ts`, looks like runtime logical changes but I'm not deep enough into verkle to review that :sweat_smile:

packages/statemanager/test/statefulVerkleStateManager.spec.ts
gabrocheleau: Yeah, I was initially going with that, but I realized that:
- We don't seem to do that for any other EIPs (the constructor logic is quite minimal)
- There are situations where we will initially activate the EIP, but only provide verkleCrypto later on by manually assigning `common.customCrypto.verkleCrypto = verkleCrypto` (e.g. b/c of asynchrony considerations)

So I decided do not do that. We could perhaps think of a "validateActiveEIPs" or something like that where we make sure that some EIPs have what they need in order to work. In the context of verkle though (at least with the way we've structured this) I decided to just validate upon usage.

acolytec3 (approved):
Shippit.

packages/client/test/rpc/engine/kaustinen6.spec.ts
```diff
@@ -80,6 +81,7 @@ describe(`valid verkle network setup`, async () => {
   const { server, chain, common } = await setupChain(kaustinen6Data, 'post-merge', {
     engine: true,
     genesisStateRoot: genesisVerkleStateRoot,
+    customCrypto: { verkleCrypto: verkle },
```
holgerd77, line 84: Some nit here: can we call this `verkle` instead of `verkleCrypto`?

This is otherwise redundant, makes the API (a bit) less elegant and is also not in line with namings like `kzg` (which is also not named `kzgCrypto` e.g.).

Should be some relatively simple and quick search-and-replace task.

holgerd77: Update: it would also simplify a lot of assignments like `customCrypto: { verkleCrypto: verkle }` to `customCrypto: { verkle }` further down the line.

PR #3800, opened 2024-11-18
Run client with stateful verkle

acolytec3:
Changes needed to run client with the `StatefulVerkleStateManager`

- Fixes a bug in `verkletrie.findPath` where we weren't addressing a scenario where an internal node has no node pointing to a non-existent leaf node
- Adds a `generateCanonicalGenesis` method to `StatefulVerkleStateManager`
- Adds a `generateVerkleStateRoot` utility method for the `client`
- Updates `blockchain.createGenesisBlock` to support building a verkle genesis block
- Renames the `Osaka` hardfork to `Verkle` to match the existing EEST usage (since verkle is no longer in Osaka)
- Adds a new `statefulVerkle` flag to the `clientOpts` to support running the client with the `StatefulVerkleStateManager`
- Updates the client startup logic to not default to stateless verkle execution
To test, you can download the kaustinen-7 genesis parameters from `https://verkle-gen-devnet-7.ethpandaops.io/` and then run the following command:
`npm run client:start:ts -- --rpc --gethGenesis=./devnets/kaustinen7/genesis.json --dataDir=datadir/kaust7 --rpcEngine --rpcEngineAuth=false --statefulVerkle=true` to start the client
Then run `./setup.sh --network kaustinen7 --dataDir kaust --justCL` with a cloned instance of the `lodestar-quickstart` repo to get a corresponding Lodestar CL client

The client will fail to validate the first block since we aren't currently in spec with kaustinen 7.


packages/client/bin/cli.ts
```diff
@@ -1211,6 +1238,7 @@ async function run() {
       return { client, servers }
     })
     .catch((e) => {
+      console.log(e)
```
gabrocheleau, line 1241:
```suggestion
```

packages/client/src/execution/vmexecution.ts
```diff
@@ -196,13 +197,19 @@ export class VMExecution extends Execution {
...
-    this.config.logger.info(`Setting up verkleVM`)
-    const stateManager = new StatelessVerkleStateManager({
-      common: this.config.execCommon,
-    })
+    if (this.config.statelessVerkle) {
+      this.config.logger.info(`Setting up verkleVM for stateless verkle execution`)
+      stateManager = new StatelessVerkleStateManager({
+        common: this.config.execCommon,
+      })
+    } else {
```
gabrocheleau, line 209: Would it be best, instead of defaulting to stateful, to require users to explicitly mention either  `this.config.statelessVerkle` or `this.config.statefulVerkle`, and to throw if none of these are met?

packages/common/src/hardforks.ts
```diff
@@ -167,7 +167,7 @@ export const hardforksDict: HardforksDict = {
    * URL        : https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/osaka.md
    * Status     : Final
    */
-  osaka: {
-    eips: [2935, 6800],
+  verkle: {
+    eips: [6800],
```
gabrocheleau, line 171: Shouldn't we have `2935, 4762` too?

acolytec3: Yes on 4762. 2935 is in Prague and the Verkle fork is currently being built on top of Prague.

packages/client/src/execution/vmexecution.ts
acolytec3: Good point, will fix.

PR #3810, opened 2024-12-04
Add simple cache to access witness

acolytec3:
WIP

### DIFF
--- a/packages/evm/src/evm.ts
+++ b/packages/evm/src/evm.ts
@@ -287,10 +287,10 @@ export class EVM implements EVMInterface {
         if (this.DEBUG) {
           debugGas(`callAccessGas charged(${callAccessGas}) caused OOG (-> ${gasLimit})`)
         }
-        message.accessWitness.clearCache()
+        message.accessWitness.revert()
         return { execResult: OOGResult(message.gasLimit) }
       } else {
-        message.accessWitness.flushCache()
+        message.accessWitness.commit()
         if (this.DEBUG) {
           debugGas(`callAccessGas used (${callAccessGas} gas (-> ${gasLimit}))`)
         }
@@ -320,7 +320,7 @@ export class EVM implements EVMInterface {
         )
         gasLimit -= absenceProofAccessGas
         if (gasLimit < BIGINT_0) {
-          message.accessWitness?.clearCache()
+          message.accessWitness?.revert()
           if (this.DEBUG) {
             debugGas(
               `Proof of absence access charged(${absenceProofAccessGas}) caused OOG (-> ${gasLimit})`,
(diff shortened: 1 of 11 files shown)


packages/vm/src/runBlock.ts
```diff
@@ -438,6 +438,8 @@ async function applyBlock(vm: VM, block: Block, opts: RunBlockOpts): Promise<App
     await accumulateParentBlockHash(vm, block.header.number, block.header.parentHash)
   }
 
+  vm.evm.systemVerkleAccessWitness?.flushCache()
```
gabrocheleau, line 441: Out of correctness, could we make that part of a isActivatedEIP(6800) conditional?

packages/vm/src/runTx.ts
```diff
@@ -802,6 +802,7 @@ async function _runTx(vm: VM, opts: RunTxOpts): Promise<RunTxResult> {
     )
   }
 
+  vm.evm.verkleAccessWitness?.flushCache()
```
gabrocheleau, line 805: Same as above (part of `isActivatedEIP(6800)` )

packages/evm/test/verkle.spec.ts
```diff
@@ -42,4 +42,68 @@ describe('verkle tests', () => {
     const retrievedValue = await sm.getStorage(address, setLengthLeft(bigIntToBytes(2n), 32))
     assert.deepEqual(retrievedValue, setLengthLeft(bigIntToBytes(1n), 32))
   })
+  it('should revert and access witness should not contain a write access due to OOG', async () => {
```
gabrocheleau, line 45: Nice tests, simple and to-the-point.

packages/evm/src/evm.ts
```diff
@@ -632,6 +642,7 @@ export class EVM implements EVMInterface {
           }
```
gabrocheleau, line 642: Should we also clearCaches when the code size exceeds maximum code size?

packages/evm/src/evm.ts
```diff
@@ -663,6 +676,7 @@ export class EVM implements EVMInterface {
       )
       gasLimit -= createCompleteAccessGas
       if (gasLimit < BIGINT_0) {
+        message.accessWitness?.clearCache()
         if (this.DEBUG) {
           debug(
             `ContractCreateComplete access gas (${createCompleteAccessGas}) caused OOG (-> ${gasLimit})`,
```
gabrocheleau, line 682: Should we also add a corresponding flushCaches in the else statement for contractCreateComplete?

gabrocheleau (approved):
Looks great! Thanks a lot for tackling this OOG stuff. 
Left a few comments, can be merged once addressed

packages/evm/src/evm.ts
g11tech: we haven't written any code chunks here, so no need

PR #3811, opened 2024-12-05
Implement debug_setHead

am1r021:
This change provides an implementation for `debug_setHead` as a part of the work being done in #3781 in order to integrate, develop, and provide an ethereumjs repl console. This work also relates to #1114 and increasing JSON-RPC API coverage.


packages/client/src/rpc/modules/debug.ts
```diff
@@ -457,4 +460,25 @@ export class Debug {
...
+  /**
+   * Returns an RLP-encoded block
+   * @param blockOpt Block number or tag
+   */
+  async setHead(params: [string]) {
+    const [blockOpt] = params
+    if (blockOpt === 'pending') {
+      throw {
+        code: INVALID_PARAMS,
+        message: `"pending" is not yet supported`,
```
acolytec3, line 473:
```suggestion
        message: `"pending" is not supported`,
```

We should never support pending for this operation.  It doesn't make sense to set our head to a "pending" block that hasn't been included in the chain.

packages/client/src/rpc/modules/debug.ts
```diff
@@ -457,4 +460,35 @@ export class Debug {
...
+        code: INVALID_PARAMS,
+        message: `"pending" is not supported`,
+      }
+    }
+
+    let headHash = await this.service.skeleton?.headHash()
+    const oldHead = headHash ? bytesToHex(headHash!) : undefined
+    const block = await getBlockByOption(blockOpt, this.chain)
+    try {
+      await this.service.skeleton?.setHead(block, true)
```
jochem-brouwer, line 483: This sets the skeleton head. Should the vm execution part also not be updated? :thinking:

packages/client/src/rpc/modules/debug.ts
```diff
@@ -457,4 +460,35 @@ export class Debug {
     this.client.config.logger.configure({ level: logLevels[level] })
     return `level: ${this.client.config.logger.level}`
   }
+
+  /**
+   * Sets the current head of the local chain by block number. Note, this is a
+   * destructive action and may severely damage your chain. Use with extreme
+   * caution.
```
jochem-brouwer, line 467: This looks like the docs from geth, are these also applicable here?

jochem-brouwer (commented):
2 comments.

I would have assumed we have some blockchain generators for tests in the client,  but it appears not? :thinking: I'm not sure if we want to export those from the blockchain package for production use-cases though (for testing - absolutely! :+1: )

packages/client/test/rpc/debug/util.ts
```diff
@@ -0,0 +1,96 @@
+import { Block, createBlock, createBlockHeader } from '@ethereumjs/block'
```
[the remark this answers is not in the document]
scorbajio: While looking more into this, I realized this is not possible since it would introduce circular dependencies.

packages/client/src/rpc/modules/debug.ts
scorbajio: I've updated it to set the execution head as well. 🙂

packages/client/src/rpc/modules/debug.ts
scorbajio: Since it is implemented with `force=true`, I'd say it can be destructive, so it would be good to know in the docs.

packages/client/src/rpc/modules/debug.ts
jochem-brouwer: Cool, can you also add this check in the tests? :)

scorbajio: I've updated the tests to include it.

packages/client/src/rpc/modules/debug.ts
```diff
@@ -457,4 +460,36 @@ export class Debug {
...
+      await this.service.execution.setHead([block])
+    } catch (e) {
+      throw {
+        code: INTERNAL_ERROR,
+      }
+    }
+    headHash = await this.service.skeleton?.headHash()
+    const newHead = headHash ? bytesToHex(headHash!) : undefined
+
+    return `oldHead: ${oldHead} - newHead: ${newHead}`
```
acolytec3, line 493: One thought I had here is it would be worth making our response value match Geth's as well.  Strings like above are pretty non-standard for RPC responses and you can't parse them programmatically without having to do fancy string parsing.

scorbajio: Yea, I think you're right will remove. It was a convenient way to develop, but will change it to match.

acolytec3 (approved):
LGTM

PR #3814, opened 2024-12-17
client: Add more unit tests for setHead function

carmen0208:
Add unit tests for PR: [Implement debug_setHead #3811](https://github.com/ethereumjs/ethereumjs-monorepo/pull/3811)
* Add more unit tests for the `setHead` function
* Add `RPCMethod` as the return type for the middleware function.
* Add the missing name parameters to the function definition comments.

### DIFF
--- a/packages/client/src/rpc/validation.ts
+++ b/packages/client/src/rpc/validation.ts
@@ -1,18 +1,21 @@
 import { INVALID_PARAMS } from './error-code.js'
 
+type RPCMethod = (params: any[]) => Promise<any>
+
 /**
  * middleware for parameters validation
  * @memberof module:rpc
  * @param method function to add middleware
  * @param requiredParamsCount required parameters count
  * @param validators array of validators
+ * @param names Optional parameter names for error messages, length must be equal to requiredParamsCount
  */
 export function middleware(
   method: any,
   requiredParamsCount: number,
   validators: any[] = [],
   names: string[] = [],
-): any {
+): RPCMethod {
   return function (params: any[] = []) {
     return new Promise((resolve, reject) => {
       if (params.length < requiredParamsCount) {
--- a/packages/client/src/rpc/validation.ts
+++ b/packages/client/src/rpc/validation.ts
@@ -1,6 +1,6 @@
 import { INVALID_PARAMS } from './error-code.js'
 
-type RPCMethod = (params: any[]) => any
+import type { RPCMethod } from './types.js'
 
 /**
  * middleware for parameters validation
--- a/packages/client/src/rpc/validation.ts
+++ b/packages/client/src/rpc/validation.ts
@@ -1,6 +1,6 @@
 import { INVALID_PARAMS } from './error-code.js'
 
-type RPCMethod = (params: any[]) => Promise<any>
+type RPCMethod = (params?: any[]) => Promise<any>
 
 /**
  * middleware for parameters validation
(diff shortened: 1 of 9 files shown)


gabrocheleau (approved):
Nice work! Thanks a lot for this submission. Will merge in once CI passes.

packages/client/src/rpc/validation.ts
```diff
@@ -1,18 +1,21 @@
 import { INVALID_PARAMS } from './error-code.js'
 
+type RPCMethod = (params: any[]) => Promise<any>
```
gabrocheleau, line 3: Looks like this is leading to type issues, see this CI run: https://github.com/ethereumjs/ethereumjs-monorepo/actions/runs/12370881134/job/34609917182?pr=3814 
Could you address those, either by improving the type (either this declared type or in usage) or alternatively, remove it?

gabrocheleau (dismissed):
See comment, type issues

scorbajio (changes requested):

packages/client/src/rpc/validation.ts
carmen0208: Thanks, I will fix it. To avoid it happening again, is there any precheck I should do before I submit the PR?

carmen0208: @gabrocheleau Fixed the type to return any. 
I also noticed several functions don't need to be wrapped with middleware because the middleware function is for parameter validation, and those functions have no parameters. So, I just removed it. Let me know if that works.

packages/client/test/rpc/debug/setHead.spec.ts
```diff
@@ -45,4 +45,47 @@ describe(method, async () => {
...
+      validateBlocks: true,
+      validateConsensus: false,
+    })
+    const blocks = await blockchain.getBlocks(0, 6, 0, false)
+    const exec = await testSetup(blockchain)
+    await exec.run()
+    const newHead = await (exec.vm.blockchain as Blockchain).getIteratorHead!()
+    assert.equal(newHead.header.number, BigInt(mainnetData.length), 'should run all blocks')
+
+    const a = await createClient({ blockchain })
```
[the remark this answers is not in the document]
carmen0208: Addressed

packages/client/src/rpc/validation.ts
```diff
@@ -1,18 +1,21 @@
 import { INVALID_PARAMS } from './error-code.js'
 
+type RPCMethod = (params: any[]) => any
```
acolytec3, line 3:
```suggestion
export type RPCMethod = (...params: any) => any
```

Let's move this to `./types.ts` in the same directory and then use it to type the `callWithStackTrace` method for consistency.  Should have the same effect overall.

packages/client/test/rpc/debug/setHead.spec.ts
```diff
@@ -18,31 +18,74 @@ describe(method, async () => {
...
+    ;(client.service.skeleton as any) = {
+      open: async () => {
+        throw new Error('open failed')
+      },
+    }
+
+    const manager = createManager(client)
+    const rpc = getRPCClient(startRPC(manager.getMethods()))
+
+    for (let i = 0; i < blocks.length; i++) {
```
acolytec3, line 85: Can we just check 1 block here and not 6 and also only run one block above (just to make this particular test more efficient since the RPC tests in general are kind of slow and this will add more overall CI time)?

packages/client/src/rpc/validation.ts
carmen0208: @acolytec3 addressed

packages/client/test/rpc/debug/setHead.spec.ts
carmen0208: addressed

acolytec3 (approved):
LGTM

chensongIntel:
Very well done @carmen0208 
😀

PR #3816, opened 2024-12-18
EIP 6690: EVMMAX implementation

am1r021:
This change is a work in progress implementation of the evmmax opcodes as defined in the latest [EIP 6690 spec](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-6690.md) along with the [geth reference implementation](https://github.com/ethereum/go-ethereum/pull/30840/).

TODO:
- [x] implement `FieldContext`
- [ ] integrate `FieldContext` with evm to implement opcodes
- [ ] implement montgomery form optimization on `FieldContext`
- [ ] benchmarking and testing


packages/evm/src/opcodes/codes.ts
```diff
@@ -210,6 +210,14 @@ const opcodes: OpcodeEntry = {
   0xa3: dynamicGasOp('LOG'),
   0xa4: dynamicGasOp('LOG'),
 
+  // '0xf0' range - extended range/width modular arithmetic
+  0xc0: asyncAndDynamicGasOp('SETMODX'),
```
jochem-brouwer, line 214: This will add these opcodes as available from all EVMs at all hardforks (so these opcodes are now available at Frontier / Chainstart)

packages/evm/src/opcodes/codes.ts
```diff
@@ -372,6 +380,19 @@ const eipOpcodes: { eip: number; opcodes: OpcodeEntry }[] = [
...
+    eip: 6690,
+    opcodes: {
+      // control & i/o
+      0xc0: asyncAndDynamicGasOp('SETMODX'),
+      0xc1: asyncAndDynamicGasOp('LOADX'),
+      0xc2: asyncAndDynamicGasOp('STOREX'),
+      // arithmetic
+      0xc3: asyncAndDynamicGasOp('ADDMODX'),
+      0xc4: asyncAndDynamicGasOp('SUBMODX'),
+      0xc5: asyncAndDynamicGasOp('MULMODX'),
```
jochem-brouwer, line 393: Note: for dynamic gas ops, a gas method should be provided in `gas.ts`, otherwise the Interpreter will crash (it expects an entry there if the opcode is marked as dynamic)

packages/common/src/hardforks.ts
```diff
@@ -171,4 +171,7 @@ export const hardforksDict: HardforksDict = {
   verkle: {
     eips: [4762, 6800],
   },
+  evmmax: {
+    eips: [6690],
```
jochem-brouwer, line 175: I don't think we should do an `evmmax` hardfork for this: if EVMMAX should be activated then the 6690 EIP should be activated (so instantiate a common and activate 6990 to use EVMMAX). I might have missed something from the EIP though. It seems to me that the EIP introduces new opcodes?

jochem-brouwer (commented):
Cool! Left three comments with some pointers :)

packages/evm/src/opcodes/codes.ts
scorbajio: I've removed the declarations here and only left the ones in `eipOpcodes`, I think that should fix this so that they only activate via eip activation via common.

packages/common/src/hardforks.ts
scorbajio: Yes, currently 6 new opcodes are added upon evmmax activation.

gabrocheleau:
Will close this as this is not maintained + doesn't seem to have gained traction long-term

PR #3817, opened 2024-12-18
evm: verkle testing extraction

gabrocheleau:
This PR extracts some verkle work from #3716 so that the other PR can remain scoped on running verkle tests.

### DIFF
--- a/packages/evm/src/params.ts
+++ b/packages/evm/src/params.ts
@@ -272,7 +272,7 @@ export const paramsEVM: ParamsDict = {
    */
   2935: {
     // evm
-    historyStorageAddress: '0xfffffffffffffffffffffffffffffffffffffffe', // The address where the historical blockhashes are stored
+    historyStorageAddress: '0x0aae40965e6800cd9b1f4b05ff21581047e3f91e', // The address where the historical blockhashes are stored
     historyServeWindow: 8192, // The amount of blocks to be served by the historical blockhash contract
     systemAddress: '0xfffffffffffffffffffffffffffffffffffffffe', // The system address
   },
--- a/packages/evm/src/params.ts
+++ b/packages/evm/src/params.ts
@@ -272,8 +272,9 @@ export const paramsEVM: ParamsDict = {
    */
   2935: {
     // evm
-    historyStorageAddress: '0x0aae40965e6800cd9b1f4b05ff21581047e3f91e', // The address where the historical blockhashes are stored
+    historyStorageAddress: '0xfffffffffffffffffffffffffffffffffffffffe', // The address where the historical blockhashes are stored
     historyServeWindow: 8192, // The amount of blocks to be served by the historical blockhash contract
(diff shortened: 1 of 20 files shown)


packages/evm/src/params.ts
```diff
@@ -272,8 +272,9 @@ export const paramsEVM: ParamsDict = {
    */
   2935: {
     // evm
-    historyStorageAddress: '0x0aae40965e6800cd9b1f4b05ff21581047e3f91e', // The address where the historical blockhashes are stored
+    historyStorageAddress: '0xfffffffffffffffffffffffffffffffffffffffe', // The address where the historical blockhashes are stored
     historyServeWindow: 8192, // The amount of blocks to be served by the historical blockhash contract
+    systemAddress: '0xfffffffffffffffffffffffffffffffffffffffe', // The system address
```
acolytec3, line 277: Do we need to have both of these parameters since they are identical?

packages/statemanager/src/statefulVerkleStateManager.ts
```diff
@@ -677,6 +680,7 @@ export class StatefulVerkleStateManager implements StateManagerInterface {
         canonicalValue = ZEROVALUE
       }
 
+      this._debug(`computed ${computedValue} canonical ${canonicalValue}`)
```
acolytec3, line 683:
```suggestion
```

I think this is an extra leftover debug log from work I was doing with the access witness cache.

packages/vm/src/params.ts
```diff
@@ -67,10 +68,8 @@ export const paramsVM: ParamsDict = {
    * Ethereum state using a unified verkle tree (experimental)
    */
   6800: {
-    // kaustinen 6 current uses this address, however this will be updated to correct address
-    // in next iteration
     // config
-    historyStorageAddress: '0xfffffffffffffffffffffffffffffffffffffffe', // The address where the historical blockhashes are stored
+    historyStorageAddress: '0x0aae40965e6800cd9b1f4b05ff21581047e3f91e', // The address where the historical blockhashes are stored
```
acolytec3, line 72: Why doesn't this match the EIP 2935 address?

acolytec3 (changes requested):
Few questions/clean up items we should address before merging.

packages/evm/src/params.ts
gabrocheleau: I will admit that I've been a bit confused with regards to those two addresses. I've played around when troubleshooting so this is a remnant of that. I've adjusted it so that it matches the EIP: https://eips.ethereum.org/EIPS/eip-2935

packages/vm/src/params.ts
gabrocheleau: Left is as is but adjusted the eip 2935 one from evm

acolytec3 (approved):
LGTM

gabrocheleau:
> LGTM

Ty! It seems like the PR wasn't approved, can you re-approve? thanks!

acolytec3 (approved):

PR #3819, opened 2024-12-19
chore: fix 404 status URL

goodfirm:
The link has expired, replace it with the latest address


acolytec3:
We don't accept typo PRs.

PR #3827, opened 2024-12-22
Implement EIP-4803 (Limit Tx Gas)

### LINKED ISSUE #3825: Implement EIP-4803
This was raised as a note in the most recent ACDE-E call that all clients should implement [EIP-4803](https://eips.ethereum.org/EIPS/eip-4803) which limits a transaction's gas limit to 2^63-1.  We haven't actually implemented this yet so should go ahead and add it to our transaction classes.  It's purely theoretical since this gas limit exceeds the max gas per block anyway but we should still do it to be spec compliant.  I'm sure some EEST test will come along and find us out otherwise.

jochem-brouwer:
Closes https://github.com/ethereumjs/ethereumjs-monorepo/issues/3825


jochem-brouwer:
Running all hardforks because I think `ethereum/tests` might not have implemented this (and instead checked original limit of `2^64-1`)

jochem-brouwer:
Ok, this is not supported by `ethereum/tests` but rather in the TransactionTests, see: 

https://github.com/ethereumjs/ethereumjs-monorepo/actions/runs/12450447293/job/34757292887?pr=3827

Errors for instance:

```
     → RLP is valid, but decoding failed
   × TransactionWithHighGasLimit64Minus1 > TransactionWithHighGasLimit64Minus1 [shanghai]
```

gabrocheleau:
Wondering if this could get finalized/merged @jochem-brouwer ?

PR #3830, opened 2025-01-07
evm: simplify verkle access witness methods

gabrocheleau:
This PR aims to simplify the verkle access witness method interface by focusing on the actual explicit actions (read/writing different parts of an account) rather than abstract methods (like "create a contract"). This ends up with DRY-er code and makes it eaiser to parse what is going on.


acolytec3 (approved):
Feels like a huge improvement to me.

PR #3832, opened 2025-01-12
evm: verkle updates

gabrocheleau:
This PR extracts some Verkle-related updates and fixes from #3716:
- Adds a systemVerkleAccessWiness to the EVM class.
- Adds a codeHash read access when creating a contract (for collision checking)
- Adds conditions that skips some accesses when interacting with system contracts and precompiles, as specified by the EIP-4762 EIP.
- Clarifies and fixes code chunk handling in the statefulVerkleStateManager
- Adjusts some miner/account reward logic so that it uses the systemVerkleAccessWitness and not the tx-level verkleAccessWitness, and skips access witness for miner when reward is 0

### DIFF
--- a/packages/evm/src/opcodes/EIP2929.ts
+++ b/packages/evm/src/opcodes/EIP2929.ts
@@ -31,6 +31,10 @@ export function accessAddressEIP2929(
     // if verkle not activated
     if (chargeGas && !common.isActivatedEIP(6800)) {
       return common.param('coldaccountaccessGas')
+    } else if (chargeGas && common.isActivatedEIP(6800)) {
+      // If Verkle is active, then the warmstoragereadGas should still be charged
+      // This is because otherwise opcodes will have cost 0 (this is thus the base fee)
+      return common.param('warmstoragereadGas')
     }
     // Warm: (selfdestruct beneficiary address reads are not charged when warm)
   } else if (chargeGas && !isSelfdestruct) {
--- a/packages/evm/src/opcodes/gas.ts
+++ b/packages/evm/src/opcodes/gas.ts
@@ -166,7 +166,8 @@ export const dynamicGasHandlers: Map<number, AsyncDynamicGasHandler | SyncDynami
         let charge2929Gas = true
         if (
           common.isActivatedEIP(6800) &&
-          runState.interpreter._evm.getPrecompile(address) === undefined
+          runState.interpreter._evm.getPrecompile(address) === undefined &&
+          !address.equals(createAddressFromStackBigInt(common.param('systemAddress')))
         ) {
           let coldAccessGas = BIGINT_0
           coldAccessGas += runState.env.accessWitness!.readAccountBasicData(address)
@@ -199,7 +200,8 @@ export const dynamicGasHandlers: Map<number, AsyncDynamicGasHandler | SyncDynami
         let charge2929Gas = true
         if (
           common.isActivatedEIP(6800) &&
-          runState.interpreter._evm.getPrecompile(address) === undefined
+          runState.interpreter._evm.getPrecompile(address) === undefined &&
+          !address.equals(createAddressFromStackBigInt(common.param('systemAddress')))
         ) {
           let coldAccessGas = BIGINT_0
           coldAccessGas += runState.env.accessWitness!.readAccountBasicData(address)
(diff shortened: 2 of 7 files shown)


acolytec3 (commented):
Can you be more explicit about what these changes are targeting?  If I understand it correctly, you're dealing with the following items:

- Checking to see if an accessed address is a system contract/precompile so we don't charge AccessWitness read for the account header fields
- Verifying that we only add the miner to the write portion of the System Verkle Witness if the miner actually get a reward for the transaction

Is that correct or is there more being covered here?

gabrocheleau:
> Can you be more explicit about what these changes are targeting? If I understand it correctly, you're dealing with the following items:
> 
> * Checking to see if an accessed address is a system contract/precompile so we don't charge AccessWitness read for the account header fields
> * Verifying that we only add the miner to the write portion of the System Verkle Witness if the miner actually get a reward for the transaction
> 
> Is that correct or is there more being covered here?

Just clarified the description to be more explicit about the changes

packages/evm/src/opcodes/gas.ts
```diff
@@ -166,7 +166,8 @@ export const dynamicGasHandlers: Map<number, AsyncDynamicGasHandler | SyncDynami
         let charge2929Gas = true
         if (
           common.isActivatedEIP(6800) &&
-          runState.interpreter._evm.getPrecompile(address) === undefined
+          runState.interpreter._evm.getPrecompile(address) === undefined &&
+          !address.equals(createAddressFromStackBigInt(common.param('systemAddress')))
```
acolytec3, line 170: My only concern with this is that should consider the larger picture, mentioned by @g11tech, that it makes sense to store system contract address, codehash, and code, in memory the way we do for precompiles.  I think we should go ahead and merge this as is but then do some work to make the system contract (plus any future ones) available in memory so we don't need to read code from storage for them.

packages/vm/src/runBlock.ts
```diff
@@ -750,24 +752,24 @@ export async function rewardAccount(
...
-    if (evm.verkleAccessWitness === undefined) {
+  if (common.isActivatedEIP(6800) === true && reward !== BIGINT_0) {
+    if (evm.systemVerkleAccessWitness === undefined) {
       throw Error(`verkleAccessWitness required if verkle (EIP-6800) is activated`)
     }
     // use vm utility to build access but the computed gas is not charged and hence free
-    evm.verkleAccessWitness.writeAccountBasicData(address)
-    evm.verkleAccessWitness.readAccountCodeHash(address)
+    evm.systemVerkleAccessWitness.writeAccountBasicData(address)
+    evm.systemVerkleAccessWitness.readAccountCodeHash(address)
```
acolytec3, line 772: Can we consolidate all this down?  Instead of having a whole code block to check to write the account header when we have account non-existence and then a second check, can we just do (this is pseudo code but you get the idea)
```
let accountExists = true
if (account === undefined) {
    accountExists = false
    account = new Account()
}

if (6800 is active and reward > 0)
  accountExists ?  writeAccountData/readCodHash : writeAccountHeader
```

acolytec3 (commented):
This looks good.  I'm going to test out my proposed simplification of the miner address AW in `runTx` logic locally and it the tests are still passing the same way, will add a commit so we can merge.

acolytec3 (approved):
LGTM.  Please review my last commit before merging

packages/evm/src/opcodes/EIP2929.ts
```diff
@@ -31,6 +31,10 @@ export function accessAddressEIP2929(
     // if verkle not activated
     if (chargeGas && !common.isActivatedEIP(6800)) {
       return common.param('coldaccountaccessGas')
+    } else if (chargeGas && common.isActivatedEIP(6800)) {
+      // If Verkle is active, then the warmstoragereadGas should still be charged
+      // This is because otherwise opcodes will have cost 0 (this is thus the base fee)
+      return common.param('warmstoragereadGas')
```
g11tech, line 37: we have a check for warm in L26, so this should just be cold gas segment no?

packages/evm/src/opcodes/gas.ts
g11tech: yea we need to address this both for system calls which will have address equal to system address as well as normal contract calls into system contracts

packages/statemanager/src/statefulVerkleStateManager.ts
```diff
@@ -318,7 +318,7 @@ export class StatefulVerkleStateManager implements StateManagerInterface {
       chunkStems[0],
       chunkSuffixes.slice(
         0,
-        codeChunks.length <= VERKLE_CODE_OFFSET ? codeChunks.length : VERKLE_CODE_OFFSET,
+        chunkSuffixes.length <= VERKLE_CODE_OFFSET ? chunkSuffixes.length : VERKLE_CODE_OFFSET,
```
g11tech, line 321: is there a difference in code chunks and suffixes length?

gabrocheleau: No, but it seemed more correct to actually use chunkSuffixes for this slice rather than implicitly relying that codeChunks.length === codeSuffixes.length

Issue #3822, opened 2024-12-21
Update `fromGethGenesis` to include blob schedule (and related changes)


acolytec3:
There's a new [EIP](https://github.com/ethereum/EIPs/pull/9129/files) with some updates to the genesis config file that we need to support.

It hads a per fork schedule for target and max blobs and we should consume this.

Needed updates
- Add blob schedule to `common.params`
- Maybe update `eth_feeHistory` to incorporate the target and max blobs schedule?
- Not sure what else

PR #3835, opened 2025-01-16
common: allow specifying eip-7840 blobSchedule via geth genesis

g11tech:
allow gethgenesis to specify eip 7840 blobschedule + optional update fraction: https://github.com/ethereum/EIPs/pull/9240

Closes https://github.com/ethereumjs/ethereumjs-monorepo/issues/3822


packages/common/src/common.ts
```diff
@@ -66,8 +66,9 @@ export class Common {
     // Assign hardfork changes in the sequence of the applied hardforks
     this.HARDFORK_CHANGES = this.hardforks().map((hf) => [
       hf.name,
-      hardforksDict[hf.name] ??
-        (this._chainParams.customHardforks && this._chainParams.customHardforks[hf.name]),
+      // Allow to even override an existing hardfork specification
+      (this._chainParams.customHardforks && this._chainParams.customHardforks[hf.name]) ?? 
+      hardforksDict[hf.name],
```
g11tech, line 71: important change here cc @jochem-brouwer @acolytec3 @holgerd77 

(please see comment)

jochem-brouwer: Ok, this works, but the problem is that if you want to correctly override it, you first have to import all the params from the original hardfork and then set that as custom hardfork

jochem-brouwer: My initial thoughts here are: load the original `hardforksDict[hf.name]` and then merge the overridden changes in it? (via copy so via JSON.stringify / JSON.parse)

packages/common/src/utils.ts
```diff
@@ -93,6 +96,33 @@ function parseGethParams(json: any) {
     throw new Error('nonzero terminal total difficulty is not supported')
   }
 
+  let customHardforks: HardforksDict | undefined = undefined;
+  if (config.blobSchedule !== undefined) {
+    customHardforks = {};
+    const blobGasPerBlob = 131072;
```
jochem-brouwer, line 102: Ideally, we do not hard-code this constant and read it from `blobGasPerBlob` (= 2**17 / 131072)

packages/common/src/utils.ts
```diff
@@ -93,6 +96,33 @@ function parseGethParams(json: any) {
...
+    for (const [hfKey, hfSchedule] of Object.entries(config.blobSchedule)) {
+      const hfConfig = hardforksDict[hfKey];
+      if (hfConfig === undefined) {
+        throw new Error(`unknown hardfork=${hfKey} specified in blobSchedule`);
```
jochem-brouwer, line 106: Could you add this case to tests?

packages/common/src/utils.ts
```diff
@@ -93,6 +96,33 @@ function parseGethParams(json: any) {
...
+      }
+      const { target, max, baseFeeUpdateFraction: blobGasPriceUpdateFraction } = hfSchedule as { target?: number, max?: number, baseFeeUpdateFraction?: undefined };
+      if (target === undefined || max === undefined) {
```
jochem-brouwer, line 109: Why is `blobGasPriceUpdateFraction` not mandatory?

jochem-brouwer: The "fallback" logic now works different from target/max and blobGasPriceUpdateFraction, for blobGasPriceUpdateFraction it will do the fallback, but for max/target it has to be explicitly defined.

packages/common/src/utils.ts
```diff
@@ -93,6 +96,33 @@ function parseGethParams(json: any) {
...
+        throw new Error(`undefined target or max specified in blobSchedule for hardfork=${hfKey}`);
+      }
+
+
+      // copy current hardfork info to custom and add blob config
+      const customHfConfig = JSON.parse(JSON.stringify(hfConfig));
+      customHfConfig.params = {
+        ...customHardforks.params,
+        // removes blobGasPriceUpdateFraction key to prevent undefined overriding if undefined
```
jochem-brouwer, line 118: I don't think this works, if `blobGasPriceUpdateFraction` is undefined then it will set the param to `undefined` also.

packages/common/src/utils.ts
```diff
@@ -111,25 +141,26 @@ function parseGethParams(json: any) {
     },
     hardfork: undefined as string | undefined,
     hardforks: [] as ConfigHardfork[],
+    customHardforks,
     bootstrapNodes: [],
     consensus:
       config.clique !== undefined
         ? {
-            type: 'poa',
```
jochem-brouwer, line 118: Why is this code block moved?

packages/common/test/utils.spec.ts
```diff
@@ -108,4 +108,83 @@ describe('[Utils/Parse]', () => {
...
+      "depositContractAddress": "0x4242424242424242424242424242424242424242",
+      "terminalTotalDifficulty": 0,
+      "terminalTotalDifficultyPassed": true,
+      "shanghaiTime": 0,
+      "cancunTime": 0,
+      "blobSchedule": {
+        "prague": {
+          "target": 61,
+          "max": 91,
+          "baseFeeUpdateFraction": 13338477
```
jochem-brouwer, line 137: Could you also add a test where `baseFeeUpdateFraction` is not set (also not explicitly to `undefined`)? I think this will override and set the param in the fork to `undefined` instead of leaving that key alone.

config/cspell-md.json
```diff
@@ -321,6 +321,7 @@
     "beaconroot",
     "Grandine",
     "EVMBN",
-    "kaust"
+    "kaust",
+    "blobschedule"
```
jochem-brouwer, line 325: This should be in `cspell-ts.json`

jochem-brouwer (changes requested):
Ok rather lengthy review :sweat_smile: 

This change: 

```typescript
      // Allow to even override an existing hardfork specification
      (this._chainParams.customHardforks && this._chainParams.customHardforks[hf.name]) ?? 
      hardforksDict[hf.name],
```
      
Is rather big, and I think we should test this more, because I think currently "wrong" inputs can throw away all hardfork params (for that common)?

jochem-brouwer:
I have updated this PR description to mark this issue https://github.com/ethereumjs/ethereumjs-monorepo/issues/3822 to be closed if it gets merged

jochem-brouwer:
Have temp changed the target branch to `update-7702-devnet-6` so we can use this branch for devnet-6 for a client branch (needs 7840)

acolytec3 (changes requested):
Other than removing some obsolete code,  looks good!

acolytec3 (approved):

PR #3836, opened 2025-01-16
Remove unused params from common

jochem-brouwer:
(mini cleanup PR)

Besides cleaning up, this PR also ensures dynamic gas values for 2537 are read from common, not specifically pointing to EIP-2537 (for future possible price updates)

I noticed that some params in common appear to be unused. CoPilot helped me, had to tweak the script but got this:

```bash
# Step 1: List all parameters
grep -ohP '(\w+):' packages/*/src/params.ts | sed 's/:$//' | sort | uniq > all_params.txt

# Step 2: Search for usage
grep -r -ohP "param\('\K[^']+" packages/**/src --include \*.ts | sort | uniq > used_params.txt

# Step 3: Identify unused parameters
comm -23 all_params.txt used_params.txt > unused_params.txt

# Output the unused parameters
cat unused_params.txt
```

Note: this outputs a lot of `*Gas`, these are still used, but EVM indexes those by `opcode.name + "Gas"` and not explicit hardcoded strings. 

NOTE: this also gives false positives, since this does not check the `paramByEIP`

### DIFF
--- a/packages/block/src/params.ts
+++ b/packages/block/src/params.ts
@@ -17,8 +17,6 @@ export const paramsBlock: ParamsDict = {
     minimumDifficulty: 131072, // The minimum that the difficulty may ever be
     difficultyBoundDivisor: 2048, // The bound divisor of the difficulty, used in the update calculations
     durationLimit: 13, // The decision boundary on the blocktime duration used to determine whether difficulty should go up or not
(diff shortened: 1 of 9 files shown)


jochem-brouwer:
All removed params have been explicitly searched under `./packages/**/` with no hits (except in param file)

g11tech (approved):

PR #3837, opened 2025-01-16
evm: update bls g2 msm price

jochem-brouwer:
As discussed on ACD-E 123 the BLS G2 price gets updated such that the formula has no non-integer arithmetic.

Note: this is NOT part of devnet-5, so do not merge (yet)

https://github.com/ethereum/EIPs/pull/9245


jochem-brouwer:
Never mind, I thought we decided yesterday that we wanted this, but we did not.

PR #3840, opened 2025-01-21
Client: fix version logging and hive block imports

jochem-brouwer:
When running `./hive --sim rpc --sim.limit rpc-compat --client ethereumjs --docker.output ethereumjs`, the `rpc-compat` tests it is noticed that the initial blocks fail to import and the client immediately exits. This PR will fix this problem. It also fixes finding the correct package.json file to extract the client version.

The main change is in the blockchain package, which now correctly sets the default value of the requestsHash in case it is not provided for the genesis block. If this is not done (or the incorrect value is used), the blockhash of the genesis block will be different and therefore the blockchain is configured to run a different chain.

### DIFF
--- a/packages/blockchain/src/blockchain.ts
+++ b/packages/blockchain/src/blockchain.ts
@@ -7,6 +7,7 @@ import {
   KECCAK256_RLP,
   Lock,
   MapDB,
+  SHA256_NULL,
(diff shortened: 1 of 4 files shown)


acolytec3 (approved):
LGTM

PR #3841, opened 2025-01-21
block: fix the param resolution for excess gas and blob fee compute

g11tech:
in devnet 5 , we fail at the transition block with mismatching excess gas compute error because target and update fraction changes. 

actually excess gas needs to be calculated with parent with params of current block else the calculations are wrong at the fork transition where these params change.

This PR fixes that by resolving the params in calc next excess gas, and calc next blob fee with the common of the the child block

However the feehistory and blob fee `eth` endpoints might not give the most accurate data as calculations will still use the common of the headBlock since we don't yet know the fork of child block as its timestamp activated and not height activated


packages/block/src/helpers.ts
```diff
@@ -119,6 +119,19 @@ export const fakeExponential = (factor: bigint, numerator: bigint, denominator:
   return output / denominator
 }
 
+/**
+   * Returns the blob gas price depending upon the `excessBlobGas` value
+   * @param excessBlobGas
+   * @param common
+   */
+export const computeBlobGasPrice = (excessBlobGas: bigint, common: Common) => {
```
jochem-brouwer, line 127: Great cleanup :+1:

packages/blockchain/src/blockchain.ts
```diff
@@ -581,7 +581,7 @@ export class Blockchain implements BlockchainInterface {
     }
 
     if (header.common.isActivatedEIP(4844)) {
-      const expectedExcessBlobGas = parentHeader.calcNextExcessBlobGas()
+      const expectedExcessBlobGas = parentHeader.calcNextExcessBlobGas(this.common)
```
jochem-brouwer, line 584: Hmm, but this gets awry at the fork block, right?

If we are currently at the Cancun fork but the next block is Prague, then `this.common` will point to Cancun, but the next block is Prague, so it should use the Prague parameters to calculate the excess blob gas. What is also somewhat annoying here is that the fork used obviously depends upon the timestamp of the next block. It is only certain that the next block will be a fork block if we are exactly 1 slot before the fork block.

packages/client/src/rpc/modules/eth.ts
```diff
@@ -1426,7 +1426,8 @@ export class Eth {
 
     if (this._chain.blockchain.common.isActivatedEIP(4844)) {
       baseFeePerBlobGas.push(
-        requestedBlocks[requestedBlocks.length - 1].header.calcNextBlobGasPrice(),
+        // use the last blocks common for fee estimation
+        requestedBlocks[requestedBlocks.length - 1].header.calcNextBlobGasPrice(requestedBlocks[requestedBlocks.length - 1].header.common),
```
jochem-brouwer, line 1430: If there are two ore more blocks, we can use the common of `requestedBlocks.length -1` and we can call it on `requestedBlocks.length - 2` (this would only be observable at the fork block)

packages/vm/src/buildBlock.ts
```diff
@@ -120,7 +120,7 @@ export class BlockBuilder {
       this.vm.common.isActivatedEIP(4844) &&
       typeof this.headerData.excessBlobGas === 'undefined'
     ) {
-      this.headerData.excessBlobGas = opts.parentBlock.header.calcNextExcessBlobGas()
+      this.headerData.excessBlobGas = opts.parentBlock.header.calcNextExcessBlobGas(this.vm.common)
```
jochem-brouwer, line 123: I think specifically for `buildBlock` we can improve: we can pre-generate the required common for this block (because we have the timestamp) and we should pass this common to calculate the next blob gas.

jochem-brouwer (commented):
Some comments. In general it would be nice to add a test here to see that common is correctly being used.

packages/vm/src/buildBlock.ts
g11tech: buildblock 's common is correct, its vm's common

packages/vm/src/params.ts
```diff
@@ -54,8 +55,14 @@ export const paramsVM: ParamsDict = {
 .  * Shard Blob Transactions
 .  */
   4844: {
+    // gasConfig
+    targetBlobGasPerBlock: 393216, // The target blob gas consumed per block
     blobGasPerBlob: 131072, // The base fee for blob gas per blob
     maxBlobGasPerBlock: 786432, // The max blob gas allowable per block
+    blobGasPriceUpdateFraction: 3338477, // The denominator used in the exponential when calculating a blob gas price
+    // gasPrices
+    simplePerBlobGas: 12000, // The basic gas fee for each blob
```
jochem-brouwer, line 64: Can you remove this one? It is unused. See: https://github.com/ethereumjs/ethereumjs-monorepo/pull/3836

packages/client/src/rpc/modules/eth.ts
g11tech: used getBlobGasPrice instead of calcnext, see if that makes sense

packages/client/src/rpc/modules/eth.ts
```diff
@@ -1425,12 +1425,7 @@ export class Eth {
     baseFees.push(nextBaseFee)
 
     if (this._chain.blockchain.common.isActivatedEIP(4844)) {
-      baseFeePerBlobGas.push(
-        // use the last blocks common for fee estimation
-        requestedBlocks[requestedBlocks.length - 1].header.calcNextBlobGasPrice(
-          requestedBlocks[requestedBlocks.length - 1].header.common,
-        ),
-      )
+      baseFeePerBlobGas.push(requestedBlocks[requestedBlocks.length - 1].header.getBlobGasPrice())
```
jochem-brouwer, line 1428: This should definitely return the blob price of the next block. `eth_feeHistory` says:

"`baseFeePerBlobGas` - An array of base fees per blob gas for blocks. **This includes the next block following the newest in the returned range**, as this value can be derived from the latest block. For blocks before EIP-4844, zeroes are returned."

There are of course exceptions, for instance if the beacon chain skips a few slots, then the next block might suddenly be the fork block but this was not scheduled. 

Anyways, it currently returns the blob price of the latest block in the requested blocks. This is already there (see the reduce "loop" above). It should definitely be the `calcNextBlobGasPrice` be for this entry. For now we can re-use the common that specific block.


Side-note: not sure why this did not fail CI. Checking that there also is the "next blob gas price" in the returned values should be a rather basic return value and test.

jochem-brouwer (approved):
LGTM :rocket: :smile:

jochem-brouwer:
Will merge, just checked: this also passes https://github.com/ethereum/execution-spec-tests/releases/tag/pectra-devnet-5%40v1.3.0 for the fork transition tests `CancunToPragueAtTime15k`.

`master` does not :)

PR #3843, opened 2025-01-21
Run `biome` as part of lint scripts

acolytec3:
This addresses a gap in our `lint` NPM scripts where `biome` was not being run.  This adds biome check and fix scripts to each package.json and ensures that the style checker is run on lint. 

Also applies existing lint fixes from `biome`.


g11tech (approved):
lfg :rocket:

PR #3845, opened 2025-01-22
Switch `c8` to `@vitest/coverage-v8`

acolytec3:
Our code coverage scripts broke with when updating `vitest` to v3. 

This PR upgrades our coverage tool from `c8` to `@vitest/v8` (which is the Vitest specific plugin for this) so coverage runs again.


jochem-brouwer (approved):
LGTM

PR #3847, opened 2025-01-24
Fix some failing hive tests in the devp2p/snap sync test suite

scorbajio:
This change does two things:
* Remove the TTD check in common that results in failed geth genesis parsing when importing a genesis with a `terminalTotalDifficulty` (TTD) field
* Add missing forks to fork map in common

These two changes are necessary in being able to parse and import blockchain data correctly when the deprecated TTD field is being used.

Thanks to @acolytec3 and @jochem-brouwer for their invaluable contributions in debugging this issue.


acolytec3:
Several additional items added to the PR:
1) Updates `parseGethParams` in `common` to use our preferred `forkmap` ordering for hardforks rather than what is provided from the `genesis` parameters file
2) Replace all references to `MergeForkTransitionId` with `MergeNetSplitBlock` (mirroring current geth usage)
3) Replaces the current logic for placing the merge fork block and computing the forkId needed for the merge with the following:
- if genesis params contains `mergeForkBlock` but no `mergeNetSplitBlock`, place the `mergeNetSplitBlock` at the Paris hardfork block  (note: `mergeForkBlock` is not present in current "canonical" testnet genesis parameters
- if genesis params contains no `mergeNetSplitBlock` but does contain `shanghaiTaim`, place mergeNetSplitBlock at 0 and assume a post-merge network
 - Removed now obsolete logging in `CLConnectionManager` about merge block approaching. The merge is ancient history and we aren't going to be replaying it on any live networks going forward

acolytec3 (approved):
I approve but will not merge since a lot of this is my code.

packages/vm/test/api/EIPs/eip-2935-historical-block-hashes.spec.ts
```diff
@@ -45,7 +45,7 @@ function eip2935ActiveAtCommon(timestamp: number, address: bigint) {
     Hardfork.GrayGlacier,
     Hardfork.Shanghai,
     Hardfork.Paris,
-    Hardfork.MergeForkIdTransition,
+    Hardfork.mergeNetsplitBlock,
```
jochem-brouwer, line 48:
```suggestion
    Hardfork.MergeNetsplitBlock,
```

All HFs / enums are uppercase in our repo :)

packages/block/test/testdata/4844-hardfork.ts
```diff
@@ -13,6 +13,7 @@ export const hardfork4844Data = {
     muirGlacierBlock: 0,
     berlinBlock: 0,
     londonBlock: 0,
+    mergeForkBlock: 0,
```
jochem-brouwer, line 16: This addition should do nothing. If there is one fork at a block, or if there are two forks at the same block: this has no impact on the `forkHash` (note: we should rename this in a different PR to `forkID`) (so both configs will have the same `forkHash`)

jochem-brouwer: (but, no need to remove it, since this adds a fork which does nothing)

packages/block/test/testdata/goerliCommon.ts
```diff
@@ -74,14 +74,14 @@ export const Goerli: ChainConfig = {
...
-      // The forkHash will remain same as mergeForkIdTransition is post merge,
+      // The forkHash will remain same as mergeNetsplitBlock is post merge,
       // terminal block: https://goerli.etherscan.io/block/7382818
       name: 'paris',
       block: 7382819,
       forkHash: '0xb8c6299d',
     },
     {
-      name: 'mergeForkIdTransition',
+      name: 'mergeNetsplitBlock',
```
jochem-brouwer, line 84: Side note, since this "fork" has the `null` block do we actually need this fork? Can't we just remove it?

packages/client/bin/utils.ts
```diff
@@ -125,7 +125,7 @@ export function getArgs(): ClientOpts {
       })
       .option('mergeForkIdPostMerge', {
         describe:
-          'Place mergeForkIdTransition hardfork before (false) or after (true) Merge hardfork in the custom gethGenesis',
+          'Place mergeNetsplitBlock hardfork before (false) or after (true) Merge hardfork in the custom gethGenesis',
```
jochem-brouwer, line 128: Nit: should we rename this client option? (And what was this supposed to do? :sweat_smile: )

packages/common/src/enums.ts
```diff
@@ -60,7 +60,7 @@ export enum Hardfork {
   London = 'london',
   ArrowGlacier = 'arrowGlacier',
   GrayGlacier = 'grayGlacier',
-  MergeForkIdTransition = 'mergeForkIdTransition',
+  mergeNetsplitBlock = 'mergeNetsplitBlock',
```
jochem-brouwer, line 63:
```suggestion
  MergeNetsplitBlock = 'mergeNetsplitBlock',
```

packages/common/test/utils.spec.ts
```diff
@@ -108,4 +96,14 @@ describe('[Utils/Parse]', () => {
       'should parse correct address',
     )
   })
+  it('should add mergeNetSplitBlock if not present when Shanghai is present', () => {
+    const genesisJSON = postMergeData
```
jochem-brouwer, line 100: This works for test data (hive), but it will not work if someone explicitly imports mainnet genesis as "custom chain". Then we will end up with wrong fork hashes :thinking:

jochem-brouwer (commented):
Some comments :smile: :+1:

packages/client/bin/utils.ts
acolytec3: We should definitely remove. I missed that.

packages/common/test/utils.spec.ts
acolytec3: As you and I discussed, mainnet it doesn't seem to have a mergeNetsplitBlock but it just be there somewhere. I honestly don't think this is an issue since why would somebody ever run mainnet from a geth genesis file?

packages/block/test/testdata/goerliCommon.ts
acolytec3: Should be able to!

packages/block/test/eip4844block.spec.ts
```diff
@@ -27,6 +27,7 @@ describe('EIP4844 header tests', () => {
     customCrypto: { kzg },
   })
 
+  console.log(common.hardforks())
```
jochem-brouwer, line 30: Console.log

packages/block/test/testdata/goerliCommon.ts
```diff
@@ -74,17 +74,12 @@ export const Goerli: ChainConfig = {
       forkHash: '0xb8c6299d',
     },
     {
-      // The forkHash will remain same as mergeForkIdTransition is post merge,
+      // The forkHash will remain same as MergeNetsplitBlock is post merge,
```
jochem-brouwer, line 77: This does not look correct (anymore)? Should be GrayGlacier?

packages/client/src/rpc/modules/engine/CLConnectionManager.ts
```diff
@@ -115,11 +115,11 @@ export class CLConnectionManager {
       maximumFractionDigits: 1,
     })
 
-    if (this.config.chainCommon.gteHardfork(Hardfork.MergeForkIdTransition)) {
+    if (this.config.chainCommon.gteHardfork(Hardfork.MergeNetsplitBlock)) {
```
jochem-brouwer, line 118: I believe this should be Paris. Will check (currently reviewing from mobile) when at home. But I believe CL will not start on mainnet?

packages/common/script.ts
jochem-brouwer, line 1: I believe this file was unintentionally included

jochem-brouwer (commented):
Some additional comments

packages/client/src/rpc/modules/engine/CLConnectionManager.ts
acolytec3: Why would this be the case?  The only change here is renaming `MergeForkIdTransition` to `MergeNetsplitBlock`.

packages/block/test/testdata/goerliCommon.ts
acolytec3: I'm not tracking your question here?  If I revert the one change (and add the `MergeForkIdTransition` back with null values and just rename to `MergeNetsplitBlock`) the tests still pass as before.

jochem-brouwer: Ah it's in test data I see (could not see that on the mobile UI). My point was that there is no `MergeNetsplitBlock` in this common so this comment does not make much sense. It's fine though, it's in the test data.

packages/client/src/rpc/modules/engine/CLConnectionManager.ts
jochem-brouwer: Good point, this should not have impact.

jochem-brouwer (approved):
LGTM :tada: :smile:

PR #3851, opened 2025-01-27
Remove Goerli from packages

jochem-brouwer:
This PR removes the Goerli chain from our packages.

However, Goerli is used in our tests to run tests for the PoA (clique) consensus, so we cannot remove these from tests (yet). I have put the Goerli config inside all the packages which run these tests. 

It would be somewhat helpful if we could create a shared directly for test data which is used among multiple packages, I've now added the same file for all packages which use Goerli tests.


acolytec3 (approved):
LGTM

acolytec3 (approved):

acolytec3 (approved):

acolytec3 (approved):

PR #3852, opened 2025-01-28
vm: add yargs as dev dep for t8n

jochem-brouwer:
Adds the yarg package to dev dependencies of VM. I just launched t8n and t8n complained about not finding the yargs package. Also adds t8n back to coverage.


acolytec3 (approved):
LGTM

PR #3857, opened 2025-01-29
binarytree: create binarytree package

gabrocheleau:
This PR creates a new binarytree package for implementing binary trees as described in https://eips.ethereum.org/EIPS/eip-7864


packages/binarytree/src/binaryTree.ts
```diff
@@ -0,0 +1,707 @@
...
+        )
+    }
+
+    // If all values are null then we treat this as a deletion.
+    if (stemNode.values.every((val) => val === null)) {
+      if (foundPath.node !== null) {
+        this.DEBUG && this.debug(`Deleting stem node for stem: ${bytesToHex(stem)}`, ['put'])
+        putStack.push([this.merkelize(stemNode), null])
+      } else {
+        return // nothing to delete
```
acolytec3, line 217: We should check if it's a deletion first and only then handle creating or updating the stem node.

packages/binarytree/src/binaryTree.ts
```diff
@@ -0,0 +1,707 @@
...
+    this.root(this.merkelize(initialNode))
+    await this._db.put(this.root(), initialNode.serialize())
+    await this.persistRoot()
+  }
+
+  /**
+   * Saves a stack of nodes to the database.
+   *
+   * @param putStack - an array of tuples of keys (the partial path of the node in the trie) and nodes (BinaryNodes)
+   */
```
acolytec3, line 517: Shouldn't we be writing zeros here instead of null if we don't allow true deletions at this point?

acolytec3 (commented):
Some early feedback.  I need to look at the `put` logic some more.  We might want to think about going to the recursive model they use in the python implementation.

packages/binarytree/LICENSE
```diff
@@ -0,0 +1,22 @@
+The MIT License (MIT)
+
+Copyright (c) 2015
```
jochem-brouwer, line 3:
```suggestion
Copyright (c) 2025 EthereumJS 
```

packages/binarytree/src/binaryTree.ts
```diff
@@ -0,0 +1,750 @@
...
+  /** The root for an empty tree */
+  EMPTY_TREE_ROOT: Uint8Array
+
+  protected _db!: CheckpointDB
+  protected _hashLen: number
+  protected _lock = new Lock()
+  protected _root: Uint8Array
+
+  protected DEBUG: boolean
+  protected _debug: Debugger = debug('binarytree:#')
```
ScottyPoi, line 46: why the "#" on the end here?

packages/binarytree/src/binaryTree.ts
```diff
@@ -0,0 +1,750 @@
...
+    return this._root
+  }
+
+  /**
+   * Checks if a given root exists.
+   */
+  async checkRoot(root: Uint8Array): Promise<boolean> {
+    try {
+      const value = await this._db.get(root)
+      return value !== null
```
ScottyPoi, line 115: As written this always returns `true` because `this._db.get()` returns `undefined`, not `null`

```suggestion
      return value !== undefined
```

packages/binarytree/src/binaryTree.ts
```diff
@@ -0,0 +1,750 @@
...
+    } catch (error: any) {
+      if (error.message === 'Missing node in DB') {
+        return equalsBytes(root, this.EMPTY_TREE_ROOT)
```
ScottyPoi, line 118: Is the idea that `checkRoot(this.EMPTY_TREE_ROOT)` should return true?  as written it will not, because db.get won't throw an error looking for it, but also wont find anything there, so this will return `false`

packages/binarytree/src/binaryTree.ts
acolytec3: This is the new logging convention we use for the base topic for any package (so `mpt:#`, `evm:#`, etc).  There was some screwiness with doing `DEBUG=ethjs,evm*` as opposed to `DEBUG=ethjs,evm:*` where subtopics weren't getting logged or something.

ScottyPoi (approved):
Added a few tests to boost the coverage. 
This looks great!  Giving my approval :+1: 
I gather that there are unanswered questions around how `tree.del` should work?  
And, of course, the proof methods and whatnot need to be filled in.  Are those defined yet in the spec, or is that still being worked on?

packages/binarytree/src/binaryTree.ts
acolytec3: I'm actually not sure here.  I think this might have been copied over from another trie class since I don't know if we use this anywhere.  We can definitely fix though.

gabrocheleau: Anyhow let's keep that change, we can revisit this whole handling once we have more clarity on how to handle deleted values etc.

gabrocheleau (commented):
Lgtm!

acolytec3 (approved):

